PR #105, opened 2015-10-25
MBS-2677: Add a way to filter search by edit count of the editor

gentlecat:
This adds a new condition into [edit search](https://beta.musicbrainz.org/search/edits): "Applied Edit Count of Editor". It allows to filter search by number of applied edits associated with the user (editor). :mag_right:

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
@@ -23,9 +23,9 @@ sub combine_with_query {
     my $sql = "EXISTS (
         SELECT 1
         FROM edit edit_inner
-        WHERE edit_inner.editor = edit.editor AND edit_inner.status = " . $STATUS_APPLIED .
-        "HAVING " . $having .
-        ")";
+        WHERE edit_inner.editor = edit.editor AND edit_inner.status = $STATUS_APPLIED
+        HAVING $having
+    )";
     $query->add_where([ $sql, $self->sql_arguments ]);
 }
 
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
@@ -10,22 +10,18 @@ extends 'MusicBrainz::Server::EditSearch::Predicate::ID';
 sub combine_with_query {
     my ($self, $query) = @_;
 
-    my $having;
+    my $subquery = "(SELECT edits_accepted + auto_edits_accepted FROM editor WHERE editor.id = edit.editor)";
+
+    my $sql;
     given ($self->operator) {
         when ('BETWEEN') {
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
```diff
@@ -0,0 +1,33 @@
...
+            $having = join(' ', 'count(*)', $self->operator, '?');
+        }
+    }
+
+    my $sql = "EXISTS (
+        SELECT 1
+        FROM edit edit_inner
+        WHERE edit_inner.editor = edit.editor AND edit_inner.status = " . $STATUS_APPLIED .
+        "HAVING " . $having .
+        ")";
```
gentlecat: Is this going to perform well? I tested only with a couple of edits.

gentlecat: Also, am I formatting query string right?

lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
```diff
@@ -0,0 +1,33 @@
...
+        }
+        default {
+            $having = join(' ', 'count(*)', $self->operator, '?');
+        }
+    }
+
+    my $sql = "EXISTS (
+        SELECT 1
+        FROM edit edit_inner
+        WHERE edit_inner.editor = edit.editor AND edit_inner.status = " . $STATUS_APPLIED .
```
mwiencek: The double quotes let you interpolate variables directly into the string, without concatenation.

lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
mwiencek: Re performance, it might work to join with the `editor` table instead, and add `editor.edits_accepted + editor.auto_edits_accepted`. They're counts of applied edits, so it should be equivalent, I hope.

root/edit/search_macros.tt
```diff
@@ -299,6 +301,23 @@
   </select>
 [% END %]
 
+[% MACRO predicate_applied_edits(field, field_contents) WRAPPER wfield predicate='id' %]
```
mwiencek: Is there any reason we can't use `predicate_int` instead? Looks about the same but I might have missed something.

gentlecat: Yeah, it fits perfectly there.

lib/MusicBrainz/Server/EditSearch/Predicate/AppliedEdits.pm
mwiencek: Re the query string, it's fine :) We don't format them consistently. I tend to use here-documents and right-align the SELECT/FROM but it's up to personal preference.

gentlecat: OK, query is updated. Using `editor` table now.

mwiencek:
:shipit:👌

PR #107, opened 2015-10-28
MBS-8590: Can't clear relationship credits

mwiencek:


lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
```diff
@@ -191,8 +191,8 @@ role {
 
                 $args{entity0} = $field->{forward} ? $source : $target;
                 $args{entity1} = $field->{forward} ? $target : $source;
-                $args{entity0_credit} = $field->{entity0_credit};
-                $args{entity1_credit} = $field->{entity1_credit};
+                $args{entity0_credit} = $field->{entity0_credit} if exists $field->{entity0_credit};
```
mwiencek: The changes to generic.js make it so that these fields are only submitted if they're changed. So we don't want to "update" any data for fields that don't exist.

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -619,8 +619,8 @@ sub update
     my $new = {};
     $new->{entity0} = $values->{entity0_id} if $values->{entity0_id};
     $new->{entity1} = $values->{entity1_id} if $values->{entity1_id};
-    $new->{entity0_credit} = $values->{entity0_credit} if $values->{entity0_credit};
-    $new->{entity1_credit} = $values->{entity1_credit} if $values->{entity1_credit};
+    $new->{entity0_credit} = $values->{entity0_credit} if defined $values->{entity0_credit};
```
mwiencek: This fixes the bug, since `''` is defined but falsey.

root/static/scripts/relationship-editor/generic.js
```diff
@@ -142,13 +142,6 @@ import {SERIES_ORDERING_TYPE_AUTOMATIC} from '../common/constants';
 
             pushInput(prefix, "target", relationship.target(vm.source).gid);
 
-            _.each(['entity0_credit', 'entity1_credit'], function (prop) {
-                var value = clean(relationship[prop]());
-                if (value) {
```
mwiencek: This is also wrong, since `''` is falsey, so no field would get outputted if you cleared the credit.

But even before this, an edit would still get submitted, even though no field was outputted. How? Well, see the lines I'm changing at https://github.com/metabrainz/musicbrainz-server/pull/107/files#diff-9dca8abb5df5f6c32c4c0ecb6a3c4ea4R194

root/static/scripts/relationship-editor/generic.js
```diff
@@ -168,6 +161,12 @@ import {SERIES_ORDERING_TYPE_AUTOMATIC} from '../common/constants';
                 }
             });
 
+            _.each(['entity0_credit', 'entity1_credit'], function (prop) {
+                if (typeof changeData[prop] === 'string') {
```
mwiencek: This makes the code consistent with its surroundings, by only outputting fields that were changed.

root/static/scripts/tests/relationship-editor.js
```diff
@@ -700,6 +700,8 @@ relationshipEditorTest("hidden input fields are generated for non-release forms"
         "edit-artist.rel.2.period.end_date.year" : "",
         "edit-artist.rel.2.period.ended": "1",
         "edit-artist.rel.2.backward": "1",
+        "edit-artist.rel.2.entity0_credit": "",
+        "edit-artist.rel.2.entity1_credit": "",
```
mwiencek: These now exist because these tests are for new relationships, where every field is considered "changed" to start with.

t/lib/t/MusicBrainz/Server/Controller/URL/Edit.pm
```diff
@@ -7,47 +7,90 @@ use MusicBrainz::Server::Test qw( html_ok );
 with 't::Mechanize', 't::Context';
 
 test all => sub {
+    my $test = shift;
```
mwiencek: Every change under `test all` is just fixing the indentation.

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -619,8 +619,8 @@ sub update
...
+    $new->{entity1_credit} = $values->{entity1_credit} if defined $values->{entity1_credit};
```
gentlecat: So `update_row` below updates values for items that are defined in `$new`, right?

lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
gentlecat: And you use `exists` here because you need to check if this key is actually in a hash? Compared to `defined` below, where you actually know that key is there and want to see if value is actually *defined*?

gentlecat:
LGTM :boat:

lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
mwiencek: Right, FormHandler will not include keys for fields that weren't submitted, which is why `exists` is used there.

The code below is slightly different, 'cause there are multiple places in the code that call `Data::Relationship::update`. I haven't checked thoroughly, but I'm sure some `undef` credits will end up there, so I used `defined` because `undef` isn't valid either way.

lib/MusicBrainz/Server/Data/Relationship.pm
mwiencek: Yup, exactly, it'll create an `UPDATE` query from all the keys in the hash https://github.com/mwiencek/musicbrainz-server/blob/mbs-8590/lib/Sql.pm#L194

PR #108, opened 2015-11-01
More autoedits

mwiencek:
Makes "add release," "add release label," "delete release label," and "add relationship" edits auto-edits, as decided during the summit.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Edit/Relationship/Create.pm
+++ b/t/lib/t/MusicBrainz/Server/Edit/Relationship/Create.pm
@@ -9,7 +9,10 @@ BEGIN { use MusicBrainz::Server::Edit::Relationship::Create }
 
 use DBDefs;
 use MusicBrainz::Server::Context;
-use MusicBrainz::Server::Constants qw( $EDIT_RELATIONSHIP_CREATE );
+use MusicBrainz::Server::Constants qw(
+    $EDIT_RELATIONSHIP_CREATE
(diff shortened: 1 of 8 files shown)


t/lib/t/MusicBrainz/Server/Edit/Relationship/Create.pm
```diff
@@ -185,6 +182,7 @@ sub _create_edit {
         end_date => { year => 1995 },
         attributes => [ ],
         ended => 1,
+        privileges => $UNTRUSTED_FLAG,
```
gentlecat: Untrusted user? What does that do?

gentlecat:
LGTM

t/lib/t/MusicBrainz/Server/Edit/Relationship/Create.pm
mwiencek: Yes, for reference see https://github.com/metabrainz/musicbrainz-server/blob/65f54a1247f0c1003b7dc9cc0a47b96463b97794/lib/MusicBrainz/Server/Data/Edit.pm#L484

This is also how the "make votable" checkbox works: https://github.com/metabrainz/musicbrainz-server/blob/65f54a1247f0c1003b7dc9cc0a47b96463b97794/lib/MusicBrainz/Server/Controller.pm#L83

gentlecat: Thanks for the info!

PR #110, opened 2015-11-04
MBS-8607: JSON-LD corrections

mwiencek:
Fixes the issues brought up in the ticket, plus a couple other things I noticed.

'writer' relationships are returned as http://schema.org/composer if they exist, which may not be 100% correct. We'll have to verify this with the people at Google.


lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
```diff
@@ -11,19 +15,27 @@ extends 'MusicBrainz::Server::WebService::Serializer::JSON::LD';
...
+sub is_person {
+    my $artist = shift;
+    return ($artist->type_id && $artist->type_id == $ARTIST_TYPE_PERSON) ? 1 : 0;
+}
+
 around serialize => sub {
     my ($orig, $self, $entity, $inc, $stash, $toplevel) = @_;
     my $ret = $self->$orig($entity, $inc, $stash, $toplevel);
 
-    $ret->{'@type'} = ($entity->type && $entity->type->name eq 'Person') ? ['Person', 'MusicGroup'] : 'MusicGroup';
```
mwiencek: This check failed if the types weren't loaded; now it checks `type_id` which should always be there.

t/lib/t/MusicBrainz/Server/Controller/Artist/Aliases.pm
```diff
@@ -59,7 +54,6 @@ $mech->content_unlike(qr/Test Alias/, 'other artist pages do not have the alias'
 page_test_jsonld $mech => {
     '@context' => 'http://schema.org',
     '@type' => 'MusicGroup',
-    'alternateName' => [],
```
mwiencek: We don't return empty lists for anything else, so I assume this was just an oversight.

t/sql/mozart.sql
```diff
@@ -0,0 +1,188 @@
+-- Automatically generated, do not edit.
```
mwiencek: The added test data is generated with `script/dump-entities-sql.pl` (which I ran from astro). We should probably change the script so that it prints the script arguments on top in a comment...but it's just "$entity_type $mbid1 $mbid2 ..."

mwiencek: Done in https://github.com/metabrainz/musicbrainz-server/commit/ef4d2597f23d78f8c39b4484b11af2c3cad47703

gentlecat: Is this just a dump from actual database? Where (in code) is it added during testing?

gentlecat:
LGTM :ship:

t/sql/mozart.sql
mwiencek: Right, I ran the dump-entities-sql script on astro, so it's dumped from the live database. It's a much easier way to generate test data when you know there's already an entity in the database that has certain characteristics. Otherwise the test data would have to be tediously written by hand.

For where it gets used in a test, see e.g. https://github.com/metabrainz/musicbrainz-server/pull/110/files#diff-c8e07cb07658abf5d9f49ff27878e7b1R251 The second argument to `prepare_test_database` specifies a file from t/sql/ to load.

PR #112, opened 2015-11-04
MBS-8614: Release merges can fail if a recording is both a merge source and merge target

mwiencek:


lib/MusicBrainz/Server/Data/Medium.pm
```diff
@@ -245,11 +245,19 @@ sub merge
...
+    # target on another track (in which case we should instead use the ID of
+    # the target from the first merge).
+    my %merge_targets;
+
     for my $recording_merge (@recording_merges) {
         my ($new, $old) = @$recording_merge;
         next if $target_count{$old} > 1;
 
-        $self->c->model('Recording')->merge(@$recording_merge);
+        $new = $merge_targets{$new} // $new;
```
gentlecat: *merge_target* is a rowid, right? So this gets new rowid if it was merged before (which means it's in `merge_targets` hash), because old one will no longer be in the database?

gentlecat:
LGTM :shipit:

lib/MusicBrainz/Server/Data/Medium.pm
mwiencek: Yup, that sounds correct, `%merge_targets` is meant to map merged source recording IDs to their (new) target recording IDs.

It's checking if `$new` itself was already merged before, and gets the ID of the recording it was merged into. Then it merges `$old` into that ID instead.

PR #113, opened 2015-11-05
MBS-8615: Non-conflicting mediums appended after a release merge is entered block the merge

mwiencek:


lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -1004,31 +1004,34 @@ sub merge
     );
 
     if ($merge_strategy == $MERGE_APPEND) {
-        my %positions = %{ $opts{medium_positions} || {} }
+        my %positions_by_medium = %{ $opts{medium_positions} || {} }
```
mwiencek: Non-functional change, just renaming the variable to be more clear what it's mapping, given that there's now a `%new_positions` variable.

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -1004,31 +1004,34 @@ sub merge
...
             or confess('Missing medium_positions parameter');
 
+        my %new_positions = map { $_ => 1 } values %positions_by_medium;
```
mwiencek: Creates a "set" of all the new positions, so we can easily check if a position is used.

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -1004,31 +1004,34 @@ sub merge
...
-            'SELECT id FROM medium WHERE release IN (' . placeholders($new_id, @old_ids) . ')',
+        my @existing_mediums = @{ $self->sql->select_list_of_hashes(
+            'SELECT id, position FROM medium WHERE release IN (' . placeholders($new_id, @old_ids) . ')',
             $new_id, @old_ids
         ) };
 
         confess('medium_positions does not account for all mediums in all releases')
-            unless all { exists $positions{$_} } @medium_ids;
+            unless all {
+                exists $positions_by_medium{$_->{id}} || !exists $new_positions{$_->{position}}
```
mwiencek: Here's the main change, since this is where the exception is thrown.

The `exists $positions_by_medium{$_->{id}}` by itself is the same as the previous test. But we also now check if the medium we're looking has a position that doesn't conflict with any the edit wants to take.

mwiencek: Using http://musicbrainz.org/edit/35664377 as an example:

When it checks medium 3 on http://musicbrainz.org/release/269fe168-7393-4fdc-8a94-eaf4e98b33db `%new_positions` will only contain 1, 2, and 4 (see all the `new_position` fields in http://musicbrainz.org/edit/35664377/data) so `!exists $new_positions{$_->{position}}` will succeed and allow the edit to go through.

gentlecat:
LGTM :racehorse:

PR #115, opened 2015-11-05
MBS-8620: Force numeric context on relationship JSON linkTypeID

mwiencek:
I'm not really sure where the strings are coming from, since I can't reproduce this locally (I guess some are coming from the cache), but this is the most sure-fire way to ensure the JSON outputs numbers instead of strings.


gentlecat:
:+1:

PR #118, opened 2015-11-10
MBS-2255: VA reports for releases and release groups

gentlecat:
*http://tickets.musicbrainz.org/browse/MBS-2255*


gentlecat:
Ugh, wrong branch.

mwiencek:
> Ugh, wrong branch.

Yeah. I was wondering if we should mark beta as the default branch.

PR #121, opened 2015-11-10
MBS-8623: Duplicate checker entity list lacks a space between name and disambiguation

mwiencek:


root/static/scripts/common/utility/entityLink.js
```diff
@@ -4,7 +4,7 @@
 // http://www.gnu.org/licenses/gpl-2.0.txt
 
 import React from 'react';
-import {l} from '../../common/i18n';
+import {l} from '../i18n';
```
gentlecat: Why is that different now?

gentlecat:
:shipit:

root/static/scripts/common/utility/entityLink.js
mwiencek: They actually both resolve to the same path, I'm not sure why it's moving out of ./common and then back again, currently.

I'll put this in a separate commit

PR #123, opened 2015-11-11
Store %ENTITIES hash as JSON

mwiencek:
To eventually provide easy access for our JavaScript.

Later, as part of the gulpfile, we'll just generate a file containing something like:

```
const ENTITIES = $json;
export default ENTITIES;
```


gentlecat:
LGTM

PR #124, opened 2015-11-11
MBS-8631: beta: "Language set." banner shows raw HTML on 404 page

mwiencek:


lib/MusicBrainz/Server/Controller/Root.pm
```diff
@@ -145,12 +145,10 @@ sub error_403 : Private
     $c->stash->{template} = 'main/403.tt';
 }
 
-sub error_404 : Private
-{
+sub error_404 : Private {
     my ($self, $c) = @_;
 
     $c->response->status(404);
-    $c->stash->{template} = 'main/404.tt';
```
mwiencek: Unrelated to the ticket but related to the 404 page...I forgot to remove the old root/main/404.tt so I'm doing that in https://github.com/mwiencek/musicbrainz-server/commit/0fd5a2885cb1242d45ca58cf8aaee395a8731b8d

root/layout/index.js
```diff
@@ -112,7 +112,7 @@ const Layout = (props) => {
 
         {!!server.alert &&
           <div className="banner warning-header">
-            <p>{server.alert}</p>
+            <p dangerouslySetInnerHTML={{__html: server.alert}}></p>
```
mwiencek: The same issue applied to the alert message (which banner admins can set, and allows HTML).

gentlecat:
LGTM

mwiencek:
It turns out that root/main/404.tt was actually still being used, since actions that detach to `/error_404` manually don't have a `NewTemplate` attribute. The new rendering stuff was breaking the tests in general, so I had to commit some stuff to fix it since today's a release.

See https://github.com/metabrainz/musicbrainz-server/commit/187f3d39e06b1f0a950e89e572d1108672fed93a and https://github.com/metabrainz/musicbrainz-server/commit/93ffe29f70cd4ffe8b1d113f0f32ddcc0fd4f68e

PR #125, opened 2015-11-12
MBS-8208 and MBS-8271

mwiencek:
cc @Freso @reosarevok

Adds an `$EDITING_DISABLED_FLAG` which when set disables editing/voting on an account, and displays a large banner message on every page (with the assumption that we won't ever set this flag without also contacting the user to explain why).

MBS-8271 banner

Also adds a UI for reporting editors for bad behavior, mostly copied from the UI for sending messages to editors. Except this sends a message to all the "account administrators" (since they're the ones who can toggle account flags).

Please review/comment on the list of options!

MBS-8208 UI

### DIFF
--- a/lib/MusicBrainz/Server/Constants.pm
+++ b/lib/MusicBrainz/Server/Constants.pm
@@ -298,6 +298,7 @@ Readonly our $MBID_SUBMITTER_FLAG      => 64;
 Readonly our $ACCOUNT_ADMIN_FLAG       => 128;
 Readonly our $LOCATION_EDITOR_FLAG     => 256;
 Readonly our $BANNER_EDITOR_FLAG       => 512;
+Readonly our $EDITING_DISABLED_FLAG    => 1024;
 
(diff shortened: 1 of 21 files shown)


gentlecat:
Maybe it would be useful to add a link to our Code of Conduct since it's mentioned in one of the options?

mwiencek:
Agreed. Done in https://github.com/mwiencek/musicbrainz-server/commit/0fdbeb51dc3bdcbb8a22c24f46c1e1b8126136aa

mwiencek:
I removed the "Send a copy to my own email address" option because it would send multiple copies (one for each account admin) and would require some refactoring to fix.

mwiencek:
I'll work on making it only send one email with all the admins in the To field.

Freso:
Looks good to me. :+1:

reosarevok:
Looks pretty good :)

gentlecat:
LGTM

PR #126, opened 2015-11-13
Remove unused cruft from WS::2::Instrument

mwiencek:
Instrument browse requests aren't actually implemented (and I'm not sure any would make sense at the moment).


lib/MusicBrainz/Server/Controller/WS/2/Instrument.pm
```diff
@@ -67,34 +62,9 @@ sub instrument : Chained('load') PathPart('') {
     $c->res->body($c->stash->{serializer}->serialize('instrument', $instrument, $c->stash->{inc}, $stash));
 }
 
-sub instrument_browse : Private {
-    my ($self, $c) = @_;
-
-    my ($resource, $id) = @{ $c->stash->{linked} };
```
mwiencek: As you can see, these params aren't actually used to do anything.

gentlecat:
LGTM :scissors:

PR #127, opened 2015-11-19
MBS-8639: Instrument merge gives ERROR: duplicate key value violates unique constraint "link_attribute_pkey"

mwiencek:
This was really annoying to fix. http://musicbrainz.org/edit/35820353 is stuck because http://musicbrainz.org/release/d769b4ac-cd35-39ab-833c-8933f2f128e5 contains this relationship: "drums, musical saw, other drums [stringdrum], percussion and tambourine: Björn Tollin (1994)."

So, that relationship contains a link attribute to "drums" (the merge target) and "other drums" (the merge source) with a `credited_as` value of "stringdrum." Since the credit makes the attributes differentiable, it tries to keep both post-merge, but you can only have one (link, attribute_type) pair in the `link_attribute` table, so it fails.

The solution here is, if there's only one conflicting attribute (like in this case), to just move the credit into the target's link attribute if it's empty. Otherwise, it creates separate links/relationships to keep the conflicting credits in.

### DIFF
--- a/lib/MusicBrainz/Server/Data/LinkAttributeType.pm
+++ b/lib/MusicBrainz/Server/Data/LinkAttributeType.pm
@@ -1,6 +1,5 @@
 package MusicBrainz::Server::Data::LinkAttributeType;
 
-use JSON;
 use Moose;
 use namespace::autoclean;
 use Sql;
@@ -13,6 +12,7 @@ use MusicBrainz::Server::Data::Utils qw(
     hash_to_row
     generate_gid
     placeholders
+    non_empty
 );
 
 extends 'MusicBrainz::Server::Data::Entity';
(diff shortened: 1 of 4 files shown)


gentlecat:
LGTM

mwiencek:
So after merging this, it fixed the ISE in the ticket, but there was then a *different* ISE preventing the same edit from closing. It's no surprise we haven't run into these issues often, because only 6 instrument merge edits have ever been entered.

I've applied https://github.com/metabrainz/musicbrainz-server/commit/2cfecd8bf26f856906d0e2c431bab7a068a5a013 to beta and ran CheckVotes on pino again, and the merge has finally applied.

gentlecat:
Because it was trying to create a relationship from `credited_as` that already existed?

mwiencek:
Not specifically affected by `credited_as`, this time. I think there were just two relationships like this:

```
X is a member of band Y (other drums)
X is a member of band Y (drums)
```
When it tried to create a new link for the first relationship using "drums" instead of "other drums", it found and used the link for the second relationship, via `Data::Link::find_or_insert`.

Then the `UPDATE l_${entity_type0}_${entity_type1} r1 SET link = ? WHERE link = ?` part failed because an identical relationship using that link already existed.

PR #128, opened 2015-11-19
MBS-8641: Release editor breaks with "error: 0" if a submitted edit throws a NoChanges exception

mwiencek:
If an edit throws `MusicBrainz::Server::Edit::Exceptions::NoChanges`, return a special `errorCode` in the /ws/js/edit response to indicate this, so we can just ignore it.

This is easy to reproduce if you open two instances of the release editor and try to make the same change in both, but it also happens in weird cases which are hard to reproduce, as described by Zas in the chatlog link in the ticket.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
@@ -51,6 +51,7 @@ no if $] >= 5.018, warnings => "experimental::smartmatch";
 
 Readonly our $ERROR_NOT_LOGGED_IN => 1;
 Readonly our $ERROR_NON_EXISTENT_ENTITIES => 2;
+Readonly our $ERROR_NO_CHANGES => 3;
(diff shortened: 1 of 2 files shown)


mwiencek:
I thought about just returning 200 OK in this situation, but it's possible we might want to differentiate them in the future...so I left it as-is.

gentlecat:
I can't reproduce this for some reason. I tried changing name of a release in two separate editor instances, but it submitted same changes without any problems.

mwiencek:
> I can't reproduce this for some reason. I tried changing name of a release in two separate editor instances, but it submitted same changes without any problems.

Hmm, did the changes actually get applied (did you set your user as an auto-editor before testing)? It appears to break without fail for me.

gentlecat:
Ah, I missed the auto-editor part. I see the error now.

gentlecat:
LGTM :shipit:

PR #132, opened 2015-11-20
MBS-1880: Allow an editor to apply edits without voting within a short timeframe after his/her 'add release'

mwiencek:
Makes the following edits auto-edits if entered within 1 hour of the original release add, by the same editor:
Add medium
Edit medium
Edit release

"Add release label" and "Remove release label" were already made auto-edits, but "Edit release label" was not, which doesn't make sense. So I just added the `AlwaysAutoEdit` role to `EditReleaseLabel` too, instead of the new `AllowAmendingRelease` role.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Edit.pm
+++ b/lib/MusicBrainz/Server/Data/Edit.pm
@@ -511,16 +511,16 @@ sub create {
 
     $edit->insert;
 
-    my $now = DateTime->now;
     my $duration = DateTime::Duration->new( days => $conditions->{duration} );
+    my $interval = DateTime::Format::Pg->format_interval($duration);
 
     my $row = {
         editor => $edit->editor_id,
         data => JSON::Any->new( utf8 => 1 )->objToJson($edit->to_hash),
         status => $edit->status,
         type => $edit->edit_type,
-        open_time => $now,
-        expire_time => $now + $duration,
+        open_time => \"now()",
+        expire_time => \"now() + interval '$interval'",
         autoedit => $edit->auto_edit,
         quality => $edit->quality,
         close_time => $edit->close_time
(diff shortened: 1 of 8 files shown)


lib/MusicBrainz/Server/Data/Edit.pm
```diff
@@ -511,16 +511,16 @@ sub create {
...
 
     my $row = {
         editor => $edit->editor_id,
         data => JSON::Any->new( utf8 => 1 )->objToJson($edit->to_hash),
         status => $edit->status,
         type => $edit->edit_type,
-        open_time => $now,
-        expire_time => $now + $duration,
+        open_time => \"now()",
+        expire_time => \"now() + interval '$interval'",
```
mwiencek: See 	01d2ac0. I noticed that locally, auto-edits were "closing" 6 hours after they were opened (I'm in UTC-6), so I couldn't even test this ticket without fixing this first.

lib/MusicBrainz/Server/Edit/Medium/Create.pm
```diff
@@ -175,26 +176,6 @@ override 'to_hash' => sub
     return $hash;
 };
 
-sub allow_auto_edit {
```
mwiencek: This was already implemented for `$EDIT_MEDIUM_CREATE`, so I basically just moved the code it has to a role.

lib/MusicBrainz/Server/Data/Edit.pm
mwiencek: Oh, and the `\` in front of the string creates a scalar reference, which tells `insert_row` to insert the string directly into the query, not to use `?` and parameter substitution, which obviously wouldn't work there.

Something new I learned today. :)

See https://github.com/mwiencek/musicbrainz-server/blob/mbs-1880/lib/Sql.pm#L119

gentlecat:
LGTM

PR #133, opened 2015-11-20
MBS-8636: Adding duplicate entities to a series causes an internal server error

mwiencek:
Ensure that `%relationships_by_link_order` is kept up-to-date based on where we plan to move things, not just based on what's currently in the database, to avoid all conflicts.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Series.pm
+++ b/lib/MusicBrainz/Server/Data/Series.pm
@@ -315,6 +315,7 @@ sub automatically_reorder {
             unless ($conflicting_relationship) {
                 push @from_values, "(?, ?)";
                 push @from_args, $relationship->id, $link_order;
+                push @{$relationships_by_link_order{$link_order}}, $relationship;
             }
 
             $prev_relationship = $relationship;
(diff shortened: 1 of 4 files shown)


gentlecat:
I've been trying to reproduce this by adding same relationship multiple times, but that didn't seem to cause any issues. As I was testing, I got this exception when approving an edit: https://gist.github.com/anonymous/c328ac4eb46f257631c7.

mwiencek:
@gentlecat That looks like the same exception this patch should fix, actually, so I guess you did manage to reproduce it, hopefully not while on this branch. :)

The ticket summary is misleading, as you should be able to add duplicate items to a series without issues...the problem occurs when you edit a series that already has duplicate items, and a re-order is triggered.

The way I can reproduce it is a bit contrived, but...if I (1) add the same item twice to a series, both with the same "number" text value, (2) submit, and (3) add another item with a text value that sorts *directly* before the duplicate items, then I consistently get the ISE.

gentlecat:
Understood. That was on latest *beta* at the time. Edits apply fine with your fix. :shipit:

PR #135, opened 2015-11-22
MBS-8647: "Edits for subscribed entites" doesn't show auto-edits

mwiencek:
cc @reosarevok

This'll show all edits by default, and it reverses the sorting so that newer edits are shown first.

But if you check "Only show open edits (oldest first)." it'll sort them oldest-first, like before.

We'll see what people think about that. See comments on http://tickets.musicbrainz.org/browse/MBS-8647

### DIFF
--- a/lib/MusicBrainz/Server/Data/Role/QueryToList.pm
+++ b/lib/MusicBrainz/Server/Data/Role/QueryToList.pm
@@ -16,13 +16,19 @@ sub query_to_list {
 }
 
 sub query_to_list_limited {
-    my ($self, $query, $args, $limit, $offset, $builder) = @_;
+    my ($self, $query, $args, $limit, $offset, $builder, $dollar_placeholders) = @_;
 
     $builder //= $self->can('_new_from_row');
 
     my @args = @$args;
+    my $count = @args;
+
     if (defined $offset) {
-        $query .= " OFFSET ?";
+        if ($dollar_placeholders) {
+            $query .= ' OFFSET $' . (++$count);
+        } else {
+            $query .= ' OFFSET ?';
+        }
         push @args, $offset;
     }
 
@@ -33,7 +39,11 @@ sub query_to_list_limited {
     };
 
     if (defined $limit) {
-        $wrapping_query .= " LIMIT ?";
+        if ($dollar_placeholders) {
+            $wrapping_query .= ' LIMIT $' . (++$count);
+        } else {
+            $wrapping_query .= ' LIMIT ?';
+        }
         push @args, $limit;
     }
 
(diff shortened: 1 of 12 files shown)


mwiencek:
This now implements http://tickets.musicbrainz.org/browse/MBS-8647?focusedCommentId=37601&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_37601 and I've fixed the conflict.

lib/MusicBrainz/Server/Data/Role/QueryToList.pm
```diff
@@ -16,13 +16,19 @@ sub query_to_list {
...
 
     $builder //= $self->can('_new_from_row');
 
     my @args = @$args;
+    my $count = @args;
+
     if (defined $offset) {
-        $query .= " OFFSET ?";
+        if ($dollar_placeholders) {
+            $query .= ' OFFSET $' . (++$count);
```
gentlecat: Initially `$count` is size of `@args`, right? So you increment it to reference `$offset` and `$limit` below?

gentlecat: Also, what's the difference between using `?` placeholders and `$<value>`? I don't understand why there need to be two ways to pass these arguments.

mwiencek: > Initially $count is size of @args, right? So you increment it to reference $offset and $limit below?

Right, it's to match the position of offset and limit in the parameter list.

> Also, what's the difference between using ? placeholders and $<value>? I don't understand why there need to be two ways to pass these arguments.

The `$` placeholders let you reference a parameter multiple times in the query, by its position. With `?`, you have to pass the same parameter multiple times, in the positions they appear.

A problem is, you can't mix these styles in the same query, or it'll throw an error. Since `query_to_list_limited` is taking an outside query and wrapping it with extra placeholders, it needs to know what style the inner query is using.

The specific reason I used `$` placeholders is, the query construction in `subscribed_entity_edits` is really complicated. When I had to worry about pushing parameters in the correct order, it was even worse.

gentlecat: OK, makes sense. Is there any advantage in using `?` instead of referencing parameters explicitly?

gentlecat:
LGTM

lib/MusicBrainz/Server/Data/Role/QueryToList.pm
mwiencek: Not that I can think of...slightly easier to type and you don't have to escape dollar signs inside double quotes, I guess? Probably zero technical advantages for us.

PR #136, opened 2015-11-23
Always show the collection name to add/remove even if there is only one

alastair:
http://tickets.musicbrainz.org/browse/MBS-8460

It doesn't make sense to show 'your collection' when there is only one
collection, but then show collection names when there is more than one.
Instead, always show collection names in the text 'add to {collection}'

### DIFF
--- a/root/layout/sidebar/shared-entity-sidebar.tt
+++ b/root/layout/sidebar/shared-entity-sidebar.tt
@@ -101,19 +101,11 @@
                     <li>
                     [%~ IF containment.${collection.id} ~%]
                         <a href="[%~ c.uri_for_action("/collection/remove", [collection.gid], { $entity_type => entity.id }) ~%]">
-                            [%~ IF collections.size == 1 ~%]
-                                [%~ l('Remove from my collection') ~%]
-                            [%~ ELSE ~%]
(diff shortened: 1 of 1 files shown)


mwiencek:
This would revert http://tickets.musicbrainz.org/browse/MBS-1768 but I think it definitely makes sense to show the name if someone renames it.

The only problem I see is that the default name "My Collection" will no longer be translated, so I wonder if it makes sense to just change the logic from `collections.size == 1` to `collection.name == 'My Collection'`.

mwiencek:
I'm fine with :ship: ing this as-is, unless others agree with changing the logic to keep the translation

alastair:
I think MBS-1766 made sense when there were only Release collections, but I would argue that it's more confusing now to say "My Collection". Which one of my collections?

mwiencek:
http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-11-23/?msg=3408909&page=3

mwiencek:
(This was merged to the beta branch, but since it was opened against master I have to close it manually.) 

Thanks!

alastair:
thanks! Next time I'll open against beta.

PR #138, opened 2015-11-25
 MBS-8645: Enable searching for edits by an event

uklauer:
Events were missing from edit search due to hard-coded entity lists. Add them to the templates and make the query parser use `entities_with`.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Query.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Query.pm
@@ -5,7 +5,7 @@ use MusicBrainz::Server::CGI::Expand qw( expand_hash );
 use MooseX::Types::Moose qw( Any ArrayRef Bool Int Maybe Str );
 use MooseX::Types::Structured qw( Map Tuple );
 use Moose::Util::TypeConstraints qw( enum role_type );
-use MusicBrainz::Server::Constants qw( $LIMIT_FOR_EDIT_LISTING );
+use MusicBrainz::Server::Constants qw( $LIMIT_FOR_EDIT_LISTING entities_with );
 use MusicBrainz::Server::EditSearch::Predicate::Date;
 use MusicBrainz::Server::EditSearch::Predicate::ID;
 use MusicBrainz::Server::EditSearch::Predicate::Set;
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/EditSearch/Query.pm
```diff
@@ -44,9 +43,12 @@ my %field_map = (
     editor_flag => 'MusicBrainz::Server::EditSearch::Predicate::EditorFlag',
     applied_edits => 'MusicBrainz::Server::EditSearch::Predicate::AppliedEdits',
 
-    map {
-        $_ => 'MusicBrainz::Server::EditSearch::Predicate::' . camelize($_)
-    } qw( area artist instrument label place recording release release_group series work )
+    entities_with(['mbid', 'relatable'],
```
uklauer: Note that this adds not only `event`, but `url`, too. You can only use it if you write a suitable query string by hand (I tested it, it works), but it is a first step towards MBS-7610.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Relationship.pm
```diff
@@ -2,7 +2,6 @@ package MusicBrainz::Server::WebService::Serializer::JSON::2::Relationship;
 use Moose;
 use Hash::Merge qw( merge );
 use List::AllUtils qw( any );
-use String::CamelCase qw( camelize );
```
uklauer: Isn’t used in this file, I found it because I grepped for `camelize`. Otherwise unrelated.

root/edit/search_macros.tt
```diff
@@ -356,7 +357,7 @@
          predicate_subscription(linked_type);
        END %]
 
-    [% FOR linked_type=[ 'area', 'instrument', 'place', 'recording', 'release_group', 'release', 'work', 'editor' ];
+    [% FOR linked_type=[ 'area', 'event', 'instrument', 'place', 'recording', 'release_group', 'release', 'work', 'editor' ];
```
uklauer: This should of course be based on `%ENTITIES` at some point, too; perhaps when it is converted to react.

mwiencek:
LGTM. Seems fine to include in this release.

PR #139, opened 2015-11-25
Fix some typos in banner messages

uklauer:


mwiencek:
Wow. My brain never noticed those.

PR #140, opened 2015-11-26
Run JavaScript tests under Node.js instead of PhantomJS

mwiencek:
This is faster because they don't have to be bundled into a gigantic file first. They can just be executed directly (JSX and all, thanks to Babel). It also removes a huge dependency (PhantomJS).

We have tests that expect a DOM to exist, though, which PhantomJS provided for us. In Node.js, we emulate the DOM using jsdom, which works everywhere except for one Autocomplete test involving element visibility that doesn't seem feasible to fix.

I think in the future we'll either want to be testing some renderer-agnostic output produced by React (and trusting that React does the right thing with it), or splitting the tests that must be run in a browser out from the unit tests.

Running the tests in a browser is still supported; that behavior has been split between the following two files:

root/static/scripts/tests/browser-runner.js
root/static/scripts/tests/node-runner.js

t/qunit.t has been moved to t/js.t, to get rid of the confusing name (we haven't used QUnit in ages).

yarb has been updated to 0.5 to fix some issues discovered where modules require() in JSON files.


Makefile.PL
```diff
@@ -150,7 +150,6 @@ requires 'Locale::Messages' => '1.20';
 author_requires 'Test::NoTabs';
 author_requires 'Test::EOL';
 
-test_requires 'Env::Path';
```
mwiencek: Seems to have only been used in t/qunit.t in relation to phantomjs.

root/static/gulpfile.js
```diff
@@ -151,9 +151,7 @@ function langToPosix(lang) {
 function buildScripts() {
   process.env.NODE_ENV = String(process.env.DEVELOPMENT_SERVER) === '1' ? 'development' : 'production';
 
-  var commonBundle = runYarb('common.js', function (b) {
-    b.expose(path.resolve(STATIC_DIR, 'lib/leaflet/leaflet-src.js'), 'leaflet');
```
mwiencek: The files that need leaflet now require this path directly. `require('leaflet')` didn't work under Node.js because it couldn't resolve that name to anything (it's an alias we only add during the bundling process, at this line).

root/static/scripts/common.js
```diff
@@ -1,8 +1,10 @@
-window.aclass = require("aclass");
-window.ko = require("knockout");
-window.L = require("leaflet");
-window._ = require("lodash");
-window.$ = window.jQuery = require("jquery");
+import global from './global';
```
mwiencek: The name of the global object is different depending on the environment (`global` vs. `window`). I added a helper module to decide which to use.

root/static/scripts/common/MB.js
```diff
@@ -50,7 +52,7 @@ MB.localStorage = function (name, value) {
 
         if (MB.hasLocalStorage) {
             try {
-                localStorage[name] = value;
+                window.localStorage.setItem(name, value);
```
mwiencek: I've added a `node-storage-shim` dep to package.json to emulate local/sessionStorage under node, but it doesn't support this kind of property access, you have to use the class API.

root/static/scripts/edit/MB/CoverArt.js
```diff
@@ -159,7 +159,7 @@ MB.CoverArt.upload_status_enum = {
 
 MB.CoverArt.validate_file = function (file) {
     var deferred = $.Deferred();
-    var reader = new FileReader();
+    var reader = new window.FileReader();
```
mwiencek: We have to be explicit now about things that exist on `window` as opposed to the global object (which may not be `window`). Which seems like a good idea anyway.

root/static/scripts/edit/MB/edit.js
```diff
@@ -290,10 +290,10 @@ import {VIDEO_ATTRIBUTE_GID} from '../../common/constants';
 
 
     function editConstructor(type, callback) {
-        return function (args, orig) {
+        return function (args, ...rest) {
             args = _.extend({ edit_type: type }, args);
 
-            callback && callback.apply(null, arguments);
+            callback && callback.apply(null, [args].concat(rest));
```
mwiencek: Was causing tests to fail in Node. Presumably https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode/Transitioning_to_strict_mode#arguments_doesn't_alias_named_function_arguments

(We reassign `args` above, which in non-strict mode would be an alias for the corresponding `arguments` slot.)

root/static/scripts/tests/autocomplete.js
```diff
@@ -84,13 +84,18 @@ autocompleteTest("clicking on actions should not close the menu (MBS-6912)", fun
...
+    t.plan(isNodeJS ? 1 : 2);
 
     searchAndClick(t, $input, $menu, ':contains(Show more...)');
 
     blurAutocomplete($input);
 
-    t.ok($menu.is(":hidden"), "menu is hidden after blurring the autocomplete");
+    // FIXME: test fails under Node.js
+    if (!isNodeJS) {
+        t.ok($menu.is(":hidden"), "menu is hidden after blurring the autocomplete");
```
mwiencek: Here's the one test that I couldn't fix.

root/static/scripts/tests/guessFeat.js
```diff
@@ -8,7 +8,7 @@ var _ = require('lodash');
 var guessFeat = require('../edit/utility/guessFeat');
 
 test('guessing feat. artists', function (t) {
-    t.plan(18);
+    t.plan(17);
```
mwiencek: This was just wrong...not sure why it didn't cause problems in phantomjs, but it did under node

mwiencek:
The phantomjs tests would frequently fail with unhelpful errors ("Script error") whenever I made large changes (porting something to React, adding new tests, etc.), and was impossible to debug without opening the tests inside a browser (which sometimes wouldn't help because they only failed within phantomjs). This should make development much easier.

gentlecat:
It seems like one of the modules is missing even after `npm install`:

```
~/w/musicbrainz-server> ./script/compile_resources.pl tests
module.js:338
    throw err;
          ^
Error: Cannot find module 'through'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/roman/workspace/musicbrainz-server/node_modules/yarb/node_modules/browser-pack/node_modules/JSONStream/index.js:6:15)
    at Module._compile (module.js:460:26)
    at Module._extensions..js (module.js:478:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/roman/workspace/musicbrainz-server/node_modules/babel-core/lib/api/register/node.js:214:7)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
```

root/static/scripts/tests/GuessCase.js
```diff
@@ -188,8 +188,6 @@ test('Work', function (t) {
     ];
 
     $.each(tests, function (idx, test) {
-        var gc = window.gc;
-
         gc.CFG_UC_ROMANNUMERALS = test.roman;
```
gentlecat: Where's `gc` coming from there?

mwiencek: https://github.com/mwiencek/musicbrainz-server/blob/node-tests/root/static/scripts/guess-case/MB/GuessCase/Main.js#L135

It's still a global, as the (redundant) removed line showed, so I removed the `var gc = window.gc;` line because there's no need to shadow it as a local variable.

mwiencek:
http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-11-27/?msg=3414386&page=3

https://github.com/mwiencek/musicbrainz-server/commit/ba248b707a386af4e98e224bc83c62eccd766257 should fix the buffer issue with newer versions of node, not sure about 0.12.x. :/

root/static/scripts/tests/GuessCase.js
mwiencek: (The line was also wrong, because in Node it's not on the `window` object.)

gentlecat:
Works fine now on node 5.1.0 and npm 3.3.12. Looks like node 0.12 has its own set of problems.

gentlecat:
LGTM :+1:

PR #141, opened 2015-11-29
MBS-8660: Fix removing an editor’s location

uklauer:
It was impossible to remove a location from an editor’s profile. The problem and fix was exactly the same as MBS-7437 (removing the area from places); see commit bc8d2635731f2541d9523dc4cf8c480f643c7876 for a detailed explanation.


mwiencek:
:shipit:

PR #142, opened 2015-11-30
MBS-8661: Fix adding and editing non-ended areas

uklauer:
The edit types for adding and editing areas did not have the new `DatePeriod` role added when the default value for the `ended` field was removed in commit 1616482a0e6681a810dcb7994601d81230dc87bd. With the code to set the `ended` value correctly missing, edits affecting non-ended areas tried to set the `ended` flag in the database to `NULL`, causing a constraint violation and an ISE. Fix this by adding the missing role.


uklauer:
Broken since 2015-07-13, never got noticed. We may really have a problem with location editing …

mwiencek:
:shipit:

mwiencek:
Added some tests in https://github.com/metabrainz/musicbrainz-server/commit/3597076493ab9a99a8bdd7305bede0a0fb14078d since I broke it.

PR #143, opened 2015-12-01
Update package.json deps

mwiencek:
The only upgrade that required any changes, based on my testing, was babel 6, since they moved everything to plugins. (Babel 5 is now deprecated.)

Note that our shrinkwrap file is currently broken with npm 2, even before this patch (it's the same issue Gentlecat had with the node-tests branch, a missing "through" module). Regenerating it (as I did here) fixed the missing-module issue, but introduced another unrelated issue with "fsevents" only on Linux. (npm is really, really buggy.) The conclusion is that we currently require npm 3. See https://github.com/metabrainz/musicbrainz-server/commit/76307d24fd0285543a080a737af5d44af1b5d48c for further explanation.


gentlecat:
LGTM. I agree with you notes about referencing specific versions. :shipit:

PR #144, opened 2015-12-01
Remove root/account/edit_ok.tt

mwiencek:
I don't see any reason we should take the user to an empty page telling them to click to their profile, instead of just redirecting to the profile...


uklauer:
I agree this is annoying and should be changed (I saw it too often when testing for MBS-8660), but I see two issues with the implementation:

- The address in the flash message is not being escaped. Although FormHandler at this point has run it through `Email::Valid`, I’m not sure what the local part still may contain. (Not interested in dissecting a screenful of a regular expression.) So, I’d like to have it escaped just as a precaution.
- There is still an intermediate step here. When the user `POST`s to `/account/edit`, they are redirected to `/account/edit?ok=1`, which (now) immediately redirects them to `/user/foobar`. Can’t we leave out the second step and directly go to the profile page?

uklauer:
Hm, on second thought I believe the changed version won’t even display the email address in the flash message when things are normal, because this is the second step (the `GET` for `/account/edit?ok=1`) where the posted data is no longer available. On the other hand, the `ok` bypasses FormHandler, so I actually could go to `/acount/edit?ok=1&email=<script type="xss-hack">...` and have it appear as-is in the flash message.

Anyway, the solution is to kick out the intermediate step

uklauer:
Hm, I can’t add a line comment to a random line not affected by the pull request. So,

    $c->response->redirect($c->uri_for_action('/account/edit', \%args));

is where the redirect to `/user/foobar` and the flash message should go.

mwiencek:
@uklauer thanks for your feedback, I made the suggested changes.

uklauer:
:shipit: (though I don’t understand how the image relates to shipping).

Freso:
@uklauer https://www.quora.com/GitHub/What-is-the-significance-of-the-Ship-It-squirrel?share=1

PR #145, opened 2015-12-01
MBS-8664: Show last login time in user profile

uklauer:
On the user profile page, show the `last_login_date` (time when a user last entered their name and password) to account administrators and when someone is viewing their own profile.

### DIFF
--- a/root/user/profile.tt
+++ b/root/user/profile.tt
@@ -91,6 +91,12 @@
             [% END %]
         [% END %]
 
+        [% IF c.user_exists AND (viewing_own_profile OR c.user.is_account_admin);
+             WRAPPER property name=l('Last password login:');
+               UserDate.format(user.last_login_date);
+             END;
+           END %]
+
         [% IF user.website && (user.has_confirmed_email_address || c.user.is_account_admin) %]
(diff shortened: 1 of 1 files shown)


root/user/profile.tt
```diff
@@ -91,6 +91,12 @@
             [% END %]
         [% END %]
 
+        [% IF c.user_exists AND (viewing_own_profile OR c.user.is_account_admin);
+             WRAPPER property name=l('Last password login:');
```
gentlecat: "password" part seems unnecessary since the only way to login is with a password. :)

mwiencek: I think the idea is to distinguish between logging in via typing your password, and logging in via an existing session cookie, but I don't have any suggestion for a better wording. :)

uklauer: Yeah, at least in the code we call the other way to get a Catalyst session a “cookie login”. Normal users will understand it like Gentlecat did, so we could probably say just “Last login” and hope that account admins will still remember this can be quite different from “Last activity”.

mwiencek: Let's use "Last login:" then and :shipit:

PR #146, opened 2015-12-02
Small fixes to the “report user” feature

uklauer:
1. Only show user-reporting link when logged in

    The link to the form for reporting misbehaving users was shown even to anonymous users. This was inconsistent with related functionality such as contacting users. Only show the link when a user is logged in.

2. Redirect to profile instead of forwarding silently

    When a user tried to report themselves, this was ignored, and the profile page shown under the `/report` URL. Instead, explicitly redirect to the profile page.


uklauer:
I stumbled over 2. when clicking the link for myself as a not-logged-in user; I was sent to the login screen, logged in, saw the profile, logged out again and didn’t get to the not-logged-in version of the profile, but to the login screen again.

Not really bad, but it annoyed me.

mwiencek:
:shipit:

PR #147, opened 2015-12-03
MBS-8145: Treat timeouts differently from ISEs

uklauer:
Database timeouts used to be treated as an internal server error, i.e. an email was sent to MetaBrainz employees about each timeout, and a rather technical error page with a stack trace shown to
the user. When, on the other hand, a request exceeded the allowable time, the server process was
immediately killed with a notice in the logs. The user was shown a plain nginx error page.

This, instead, suppresses error emails for both kinds of timeouts (but logs them) and shows a dedicated error page suggesting to reload the page, with a 503 status code (service unavailable). Instead of killing the process, an exception is thrown, allowing the normal error processing and cleanup to happen. This also makes the return-to-search-form feature of the edit search less fragile (it only worked when the query timed out before the request did, MBS-8382).

### DIFF
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -396,11 +396,23 @@ around 'finalize_error' => sub {
     my @args = @_;
 
     $c->with_translations(sub {
+        my $errors = $c->error;
+
+        my $timed_out = 0;
+        $timed_out = 1
+            if scalar @$errors == 1 && blessed $errors->[0]
+                && $errors->[0]->does('MusicBrainz::Server::Exceptions::Role::Timeout');
+
+        # don't send mail about timeouts (ErrorCatcher will log instead)
+        local $MusicBrainz::ErrorCatcherEmailWrapper::suppress = 1
+            if $timed_out;
+
         $c->$orig(@args);
 
         if (!$c->debug && scalar @{ $c->error }) {
-            $c->stash->{errors} = $c->error;
-            $c->stash->{template} = 'main/500.tt';
+            $c->stash->{errors} = $errors;
+            $c->stash->{template} = $timed_out ?
+                'main/timeout.tt' : 'main/500.tt';
             $c->stash->{stack_trace} = $c->_stacktrace;
             try { $c->stash->{hostname} = hostname; } catch {};
             $c->clear_errors;
@@ -411,6 +423,8 @@ around 'finalize_error' => sub {
                 $c->res->{body} = 'clear';
                 $c->view('Default')->process($c);
                 $c->res->{body} = encode('utf-8', $c->res->{body});
+                $c->res->{status} = 503
+                    if $timed_out;
             }
         }
     });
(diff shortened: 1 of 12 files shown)


uklauer:
I tested this on a local server where I added `sleep;`, `die "foo";` and `...::StatementTimedOut->throw(...);` in a controller method to cause a request timeout and to simulate ISE-worthy errors and database timeouts, respectively.

root/main/timeout.tt
```diff
@@ -0,0 +1,8 @@
+[% WRAPPER "layout.tt" title=l('Request Timed Out') full_width=1 %]
+    <div id="content">
+        <h1>[% l('Request Timed Out') %]</h1>
+
+        <p><strong>[% l('Processing your request took too long and timed out.') %]</strong></p>
+        <p>[% l('It may help to try again by reloading the page.') %]</p>
+    </div>
```
uklauer: Feel free to comment on the language, of course.

I was also wondering whether we should include the strack trace here, perhaps in an element that is `display: none`? It may still be useful for developers without access to the log.

lib/MusicBrainz/Server.pm
```diff
@@ -367,19 +367,11 @@ around dispatch => sub {
...
-                                   $max_request_time));
-            $c->log->error(Devel::StackTrace->new->as_string);
-            $c->log->_flush;
-
             my $context = $c->model('MB')->context;
             if (my $sth = $context->sql->sth) {
                 $sth->cancel;
             }
-
-            $context->connector->disconnect;
```
mwiencek: No longer needed because it'll make it to the end of `dispatch` now, right?

root/main/timeout.tt
mwiencek: Yeah, that sounds like a good idea. The message otherwise reads fine to me.

mwiencek:
This is wonderful. :) We can merge after the hidden stack trace is added, if you'd like to do that.

lib/MusicBrainz/Server.pm
uklauer: Yes, exactly.

uklauer:
I added a `<div style="display:none">` to the timeout page containing the information that is shown for ISEs. To this end, I took the relevant code blocks from `500.tt` and put them into two new files, `error.tt` with the error string and stack trace and `environment.tt` with things such as time and processing server.

In the process, I moved the request parameters from `error.tt` to `environment.tt` because it seemed more fitting there, and also made `400.tt` use these blocks (if 400s outside of the WS even occur).

root/main/400.tt
```diff
@@ -4,15 +4,7 @@
...
-        <p>
-            <strong>[%- l('Error message: ') -%]</strong>
-            [% IF message %]
-                <code>[%- message | html -%]</code>
-            [% ELSE %]
-                <code>[%- l('(No details about this error are available)') -%]</code>
-            [% END %]
-        </p>
-
+        [% INCLUDE 'main/info/error.tt' %]
```
uklauer: When there is no stack trace in the stash, `error.tt` resolves exactly to the above.

gentlecat:
Is there a way to keep track of number of timeouts that we have apart from looking at the logs directly? Something in http://stats.musicbrainz.org/?

mwiencek:
@gentlecat Hmm, not really. You can see the number of 503s on stats.mb, but those will almost all be from rate-limiting. We at least know exactly what pages tend to time out (or it's obvious while using the site), so the emails weren't all that helpful (and it's easy to run `grep -R 'Request took more than' .` in the service directory if we need examples).

PR #148, opened 2015-12-03
Convert GuessCase/Mode/Base to ES6 module

mwiencek:
In an effort to remove all global variables and be able to run individual test files directly using `babel-cli`.

(Annoyed me again while looking at some guess-case bugs.)


root/static/scripts/common/utility/getBooleanCookie.js
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (C) 2015 MetaBrainz Foundation
+// Licensed under the GPL version 2, or (at your option) any later version:
+// http://www.gnu.org/licenses/gpl-2.0.txt
+
+import getCookie from './getCookie';
+
+export default function getBooleanCookie(name, defaultValue = false) {
+  let value = getCookie(name);
+
+  if (value === '1' || value === 'true') {
```
mwiencek: We're inconsistent about how we store cookies everywhere, so I added a helper function.

root/static/scripts/guess-case/MB/GuessCase/Mode/Base.js
```diff
@@ -1,324 +0,0 @@
...
-
-    // ----------------------------------------------------------------------------
-    // member functions
-    // ---------------------------------------------------------------------------
-
-    self.description = "";
-
-    self.getDescription = function () {
-        /* Work around MBS-5734 */
-        return self.description.replace('<a ', '<a target="_blank" ');
```
mwiencek: MBS-5734 has been fixed for a while so I removed this from the new file, in favor of a correct solution.

root/static/scripts/common/utility/getBooleanCookie.js
mwiencek: This made it more convenient to remove the `CFG_UC_ROMANNUMERALS` variable attached to the `gc` global. We just read the `guesscase_roman` cookie directly where it's needed now.

root/static/scripts/guess-case/MB/Control/GuessCase.js
```diff
@@ -97,7 +97,7 @@ ko.bindingHandlers.guessCase = {
         }
 
         if (!guessCaseOptions.upperCaseRoman.peek()) {
-            guessCaseOptions.upperCaseRoman(global.gc.CFG_UC_ROMANNUMERALS);
+            guessCaseOptions.upperCaseRoman(getBooleanCookie('guesscase_roman'));
```
uklauer: Look, @gentlecat has his own cookie!

gentlecat: :cat2:

root/static/scripts/guess-case/MB/Control/GuessCase.js
```diff
@@ -60,7 +61,7 @@ var mode = ko.computed({
 
         if (modeName !== gc.modeName) {
             gc.modeName = modeName;
-            gc.mode = MB.GuessCase.Mode[modeName];
+            gc.mode = require('../../modes')[modeName];
```
gentlecat: I assume this points to the *root/static/scripts/guess-case/modes.js* script. I'm a bit confused about what `require('../../modes')` function returns and what's referenced by `modeName`.

Is there a good summary of new features out there? I'm always a bit behind on all new JS/node stuff. :(

mwiencek: http://babeljs.io/docs/learn-es2015/ is pretty good I think. But it helps to understand [Node.js's module system](https://nodejs.org/api/modules.html) too, since that's what Babel converts the ES6 `import` statements into.

The `require` function always returns module's `exports` object. Statements like `export let English = ...` are converted into something like `exports.English = ...`.  So `modeName` would be English, French, Sentence, etc.

This should probably be moved to an `import` at the top of the file, anyway.

mwiencek: http://jsmodules.io/cjs.html is also useful.

The biggest different between these module systems is, `import` statements are statically analyzable at compile-time. `require` is just a function executed at run-time like any other, so you can e.g. pass in a variable containing the path to a module, which just isn't possible with `import`.

gentlecat:
LGTM. :shipit: Sorry for delay, still trying to understand how modern JS is supposed to work.

mwiencek:
This was merged to master. The PR will close at next release.

PR #149, opened 2015-12-07
MBS-7012: Join phrase cleanup not running

mwiencek:
The first argument is the input element:
https://github.com/metabrainz/musicbrainz-server/blob/1838455/root/components/forms.tt#L358


gentlecat:
LGTM :ship:

PR #151, opened 2015-12-07
MBS-8635: Untranslatable strings

mwiencek:
There were three quite unrelated issues here. The commits have additional info, but the summaries are:

* Locale::Messages strictly looks up msgids via bytewise comparison
* A msggrep wildcard of ours wasn't doing what was intended
* We were using the wrong po2json `format` option

I *think* my solution to the first bullet point shouldn't cause any unintended effects (those subroutines should only be called from the templates and with raw strings), but it could certainly use some sanity checking.

### DIFF
--- a/lib/MusicBrainz/Server/Plugin/Translation.pm
+++ b/lib/MusicBrainz/Server/Plugin/Translation.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use base 'Template::Plugin';
+use Encode qw( encode );
 use MusicBrainz::Server::Translation ();
 use MusicBrainz::Server::Translation::Statistics ();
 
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/Plugin/Translation.pm
```diff
@@ -23,20 +24,24 @@ sub new {
 sub l {
     my ($self, $msgid, $vars) = @_;
```
gentlecat: `$msgid` is the actual string, right?

mwiencek: Right, it's just called `msgid` in the po files, etc.

gentlecat:
LGTM :shipit:

PR #152, opened 2015-12-07
MBS-8368: Use a better-performing query for collection edits

uklauer:
This was the winner in the performance tests: `SELECT * FROM edit JOIN ($subquery) re ON edit.id = re.edit` It’s the only one where the default selectivity is not applied.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Edit.pm
+++ b/lib/MusicBrainz/Server/Data/Edit.pm
@@ -36,7 +36,7 @@ use aliased 'MusicBrainz::Server::Entity::EditorSubscription';
 
 extends 'MusicBrainz::Server::Data::Entity';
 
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Edit.pm
```diff
@@ -385,6 +391,7 @@ EOSQL
            JOIN editor_subscribe_collection esc ON esc.collection = ${coll_table}.collection
          WHERE esc.available"
     } entities_with('collections'));
+    # FIXME: very similar to $EDIT_IDS_FOR_COLLECTION_SQL, should be generalized
```
uklauer: Should be done ideally, but not now. ;-) Just marking it for future reference (and so that someone grepping for the constant will find this, too).

mwiencek:
:shipit: !

Thanks a lot for working on this! Let's hope it makes beta a bit more usable.

PR #153, opened 2015-12-07
Error message improvements

uklauer:
1. Show the failed statement in case of a database error. I’ve found this quite useful in debugging.
2. Simplify debugging output for when inserting an edit fails
    
  When inserting an edit failed with an exception, a new exception was thrown containing the body (`POST`) parameters of the request and a dump of the original exception. However, the request parameters are shown anyway both on the ISE page and in the error email, and the dump of a complex exception object is far too verbose.
    
  Don’t include the request parameters in the new exception, and just show the stringified original exception together with its class name.

### DIFF
--- a/lib/MusicBrainz/Server/Controller.pm
+++ b/lib/MusicBrainz/Server/Controller.pm
@@ -98,10 +98,7 @@ sub _insert_edit {
         } elsif (ref($_) eq 'MusicBrainz::Server::Edit::Exceptions::DuplicateViolation') {
             $c->stash(duplicate_violation => 1);
         } else {
-            use Data::Dumper;
-            croak "The edit could not be created.\n" .
-                "POST: " . Dumper($c->req->params) . "\n" .
-                "Exception:" . Dumper($_);
+            croak 'The edit could not be created. Exception (' . (ref ne '' ? ref : 'string') . '): ' . $_;
         }
     };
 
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller.pm
```diff
@@ -98,10 +98,7 @@ sub _insert_edit {
         } elsif (ref($_) eq 'MusicBrainz::Server::Edit::Exceptions::DuplicateViolation') {
             $c->stash(duplicate_violation => 1);
         } else {
-            use Data::Dumper;
-            croak "The edit could not be created.\n" .
-                "POST: " . Dumper($c->req->params) . "\n" .
-                "Exception:" . Dumper($_);
+            croak 'The edit could not be created. Exception (' . $_->meta->name . '): ' . $_;
```
uklauer: Of course, the _most important_ fix here is the space after the colon. ;-)

mwiencek:
The repeated post params have been a long-time annoyance. Will this still show the edit data though?

uklauer:
Yes, showing `%opts` is handled elsewhere: https://github.com/uklauer/musicbrainz-server/blob/error_msgs/lib/MusicBrainz/Server/Data/Edit.pm#L511

mwiencek:
This was merged to master. The PR will close at next release.

PR #158, opened 2015-12-08
MBS-7882: Make bracketed('') a no-op

uklauer:
Make the `bracketed` macro return an empty string instead of empty parentheses when it is given an empty string as input. This renders some guards in other macros unnecessary and fixes an issue in the
template for "Edit alias" edits [where such a guard was missing](https://github.com/metabrainz/musicbrainz-server/blob/4020e6faad0f639286aa49238a20bed3e5f7d99c/root/edit/details/edit_alias.tt#L24-L26).

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -6,7 +6,7 @@
 [%~ MACRO replace(text, search, replace) BLOCK; text | replace(search, replace); END -%]
 [%~ MACRO url_escape(url) BLOCK; url | url; END -%]
 [%~ MACRO html_escape(url) BLOCK; url | html; END -%]
-[%~ MACRO bracketed(text) BLOCK; l(' ({text})', { text => text }); END -%]
+[%~ MACRO bracketed(text) BLOCK; text != '' ? l(' ({text})', { text => text }) : ''; END -%]
 [%~ MACRO js_escape(text) BLOCK; text | js; END ~%]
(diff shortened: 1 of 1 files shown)


uklauer:
Just saw that `isolate_text` might cause similar problems in the future (e.g. when calling `bracketed(isolate_text(text))`), so fixed that pre-emptively, too.

uklauer:
Example: http://chirlu.mbsandbox.org/edit/33188900 vs. https://beta.musicbrainz.org/edit/33188900

root/components/common-macros.tt
```diff
@@ -6,7 +6,7 @@
 [%~ MACRO replace(text, search, replace) BLOCK; text | replace(search, replace); END -%]
 [%~ MACRO url_escape(url) BLOCK; url | url; END -%]
 [%~ MACRO html_escape(url) BLOCK; url | html; END -%]
-[%~ MACRO bracketed(text) BLOCK; l(' ({text})', { text => text }); END -%]
+[%~ MACRO bracketed(text) BLOCK; text != '' ? l(' ({text})', { text => text }) : ''; END -%]
```
mwiencek: I wondered if that should also check `text.defined`, but maybe it's better to make it obvious where we have `undef` values creeping around.

mwiencek:
:ship:

PR #159, opened 2015-12-08
Faster subscribed entity edits

mwiencek:


mwiencek:
This is required to fix MBS-8647 on beta, which is currently unusable. See http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-12-08/?msg=3428137&page=3

PR #160, opened 2015-12-08
MBS-8414/MBS-7572: Discrepancies between edit data and database

uklauer:
Two variations of the same problem: We have database triggers that will enforce certain restrictions on the data; in particular, an artist with an end area must have `ended` set to true (MBS-8414) and a search hint alias can’t have locale, dates etc. (MBS-7572). However, the edit types didn’t know about this, and so it was possible that an edit stored an invalid combination (e.g. `end_area_id => 1234, ended => 0`). The edit would be displayed with the information “Ended: No”, but when it was applied, _ended_ would be set to true. I fix this by making sure that the edit data is consistent when an edit is created.

There is a minor UI thing here, too, that is somewhat related; interestingly, nobody has ever complained about it, so no ticket. In the edit form for an alias, when “search hint” is chosen as the type, the input fields that aren’t applicable to search hints are disabled. However, the checkbox for “ended” stayed active, even though search hints can’t have an ended flag.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Alias.pm
+++ b/lib/MusicBrainz/Server/Edit/Alias.pm
@@ -1,0 +1,49 @@
+package MusicBrainz::Server::Edit::Alias;
+use Moose::Role;
+
+use MusicBrainz::Server::Data::Utils qw( partial_date_to_hash non_empty );
+use MusicBrainz::Server::Constants qw( %ENTITIES );
+
+use aliased 'MusicBrainz::Server::Entity::PartialDate';
+
+sub enforce_dependencies {
+    my ($self, $opts) = @_;
+
+    unless (non_empty($opts->{locale})) {
+        # Alias without locale can't be primary
+        $opts->{primary_for_locale} = 0;
+    }
+
+    my $search_hint_type = $ENTITIES{$self->_alias_model->type}->{aliases}{search_hint_type};
+    my $type = $opts->{type_id};
+    if (defined $type && $type == $search_hint_type) {
+        # Search hints can't have any other data
+        $opts->{sort_name} = $opts->{name};
+        $opts->{locale} = undef;
+        @$opts{qw( begin_date end_date )} = (partial_date_to_hash(PartialDate->new)) x 2;
+        @$opts{qw( ended primary_for_locale )} = (0) x 2;
+    }
+}
+
+1;
+
+=head1 COPYRIGHT
+
+Copyright (C) 2015 Ulrich Klauer
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License as
+published by the Free Software Foundation; either version 2 of
+the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+MA 02110-1301, USA.
+
+=cut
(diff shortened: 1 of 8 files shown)


uklauer:
For the UI issue, compare http://chirlu.mbsandbox.org/artist/f27ec8db-af05-4f36-916e-3d57f91ecf5e/alias/2483/edit to https://beta.musicbrainz.org/artist/f27ec8db-af05-4f36-916e-3d57f91ecf5e/alias/2483/edit

lib/MusicBrainz/Server/Edit/Alias/Delete.pm
```diff
@@ -8,7 +8,7 @@ use MusicBrainz::Server::Data::Utils qw( partial_date_to_hash );
 use MusicBrainz::Server::Edit::Types qw( Nullable PartialDateHash );
 
 extends 'MusicBrainz::Server::Edit';
-
+with 'MusicBrainz::Server::Edit::Alias';
```
uklauer: Only for completeness (in case we want to put more alias-specific common functions in there).

lib/MusicBrainz/Server/Edit/Alias.pm
```diff
@@ -0,0 +1,52 @@
+package MusicBrainz::Server::Edit::Alias;
+use Moose::Role;
+
+use MusicBrainz::Server::Data::Utils qw( partial_date_to_hash );
+use MusicBrainz::Server::Constants qw( %ENTITIES );
+
+use aliased 'MusicBrainz::Server::Entity::PartialDate';
+
+sub enforce_dependencies {
```
uklauer: I’m not too happy with the name, so if you have better ideas …

lib/MusicBrainz/Server/Edit/Alias/Edit.pm
```diff
@@ -172,6 +174,8 @@ sub initialize
         $opts{sort_name} = $opts{name} if non_empty($opts{name});
```
mwiencek: Would it make sense to move this chunk of code to `enforce_dependencies` too? Edit::Alias::Add has similar logic, except it doesn't `delete $opts{sort_name}` or check `if non_empty($opts{name})`.

lib/MusicBrainz/Server/Edit/Alias.pm
```diff
@@ -0,0 +1,52 @@
...
+    my ($self, $opts, $alias) = @_;
+
+    if ($opts->{locale} // '' eq '') {
```
mwiencek: perhaps `if (!non_empty` or `unless (non_empty` for consistency?

mwiencek:
LGTM.

uklauer:
I’m going to use `non_empty` in any case and look into whether the `sort_name` code can be generalized.

uklauer:
OK, made both changes.

lib/MusicBrainz/Server/Edit/Alias/Edit.pm
```diff
@@ -167,10 +169,7 @@ sub initialize
     die "You must specify the alias object to edit" unless defined $alias;
     my $entity = delete $opts{entity} or die 'Missing "entity" argument';
 
-    unless (non_empty($opts{sort_name})) {
-        delete $opts{sort_name};
-        $opts{sort_name} = $opts{name} if non_empty($opts{name});
-    }
```
uklauer: I wonder if there was a reason that I don’t understand for that this code was longer than the other version. But `$opts{name}` can’t be empty (the name is a required field), so the inner condition seems redundant to me.

mwiencek: I think it's just because the tests for `Edit::Artist::EditAlias` don't pass in a name when constructing the edits. Those tests fail if I remove that part. I see the name is required by FormHandler, so perhaps this can't be triggered through the UI. I guess I'd lean toward keeping it so we don't have to update the tests, and in case we ever allow this edit type in /ws/js/edit, which I imagine would happily accept only the fields that changed.

PR #165, opened 2015-12-09
Providing links to accounts in emails to admins. http://tickets.musicbrainz.org/browse/MBS-8663

stasszczesniak:
We have an option to report a user that's misbehaving or needs to be looked into for whatever reason. This sends a mail to the admins that includes the user's username, but it would be much nicer if it also included a link to their MusicBrainz profile. So the code that sends the emails is modified now to do that.

GCI task, connected to this pull request.
https://codein.withgoogle.com/dashboard/task-instances/5804155620818944/

### DIFF
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -490,9 +490,9 @@ $subject for the following reason:
 
 “$reason”
 
-Links to related accounts :
-	Reporter’s account: https://musicbrainz.org/user/$reporter_tolink
-	Reported user’s account: https://musicbrainz.org/user/$reported_user_tolink
+Reporter’s account: https://musicbrainz.org/user/$reporter_tolink
+Reported user’s account: https://musicbrainz.org/user/$reported_user_tolink
+
 EOF
 
     if ($opts{reveal_address}) {
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -483,7 +483,7 @@ sub send_editor_report {
     my $reported_user = $opts{reported_user};
     my $subject = 'Editor ' . $reported_user->name . ' has been reported by ' . $reporter->name;
     my $reason = $MusicBrainz::Server::Form::User::Report::REASONS{$opts{reason}};
-    my $reporter_tolink = uri_escape_utf8($reporter)
+    my $reporter_tolink = uri_escape_utf8($reporter);
     my $reported_user_tolink = uri_escape_utf8($reported_user);
     my $body .= <<EOF;
 $subject for the following reason:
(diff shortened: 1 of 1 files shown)


mwiencek:
Just a few small comments:

* Regarding the style, we should lowercase `user's` in `Reported User's account` and move the colons directly after "account" on both lines, without any spaces before them. Since not everyone's email client will use a fixed-width font, the current alignment won't work everywhere and will look strange. :)
* Other parts of the message use a fancy Unicode apostrophe `’` instead of a typewriter apostrophe `'`, so let's use the Unicode one too.
* The URLs may not auto-link correctly if there are any spaces in the usernames. I think we should URI-encode the usernames beforehand, to convert the spaces into `%20`. There's already a `uri_escape_utf8` function imported in this file, so you can just do something like this above:
```Perl
my $reported_user_escaped = uri_escape_utf8($reported_user);
my $reporter_escaped = uri_escape_utf8($reporter);
```

mwiencek:
For bonus points, it'd be great if you could include a reference to the ticket number (MBS-8663) in the commit message, instead of "modified: lib/MusicBrainz/Server/Email.pm"

stasszczesniak:
It should be ok now :+1:

lib/MusicBrainz/Server/Email.pm
```diff
@@ -483,12 +483,16 @@ sub send_editor_report {
...
     my $reason = $MusicBrainz::Server::Form::User::Report::REASONS{$opts{reason}};
-
+    my $reporter_tolink = uri_escape_utf8($reporter);
+    my $reported_user_tolink = uri_escape_utf8($reported_user);
     my $body .= <<EOF;
 $subject for the following reason:
 
 “$reason”
 
+Links to related accounts :
```
mwiencek: It think it was better without this line. The links are already individually labeled.

lib/MusicBrainz/Server/Email.pm
```diff
@@ -483,12 +483,16 @@ sub send_editor_report {
...
+	Reporter’s account: https://musicbrainz.org/user/$reporter_tolink
```
mwiencek: Please remove the tab characters at the start of these lines.

lib/MusicBrainz/Server/Email.pm
```diff
@@ -483,12 +483,16 @@ sub send_editor_report {
...
+	Reported user’s account: https://musicbrainz.org/user/$reported_user_tolink
 EOF
```
mwiencek: There used to be an extra line break before `EOF`, but you removed it. I think it's needed, otherwise the text added below won't be spaced properly.

mwiencek:
Looks perfect now, thanks! :shipit:

mwiencek:
I squashed the commits into https://github.com/metabrainz/musicbrainz-server/commit/91009f62e5f518ed2a933b95d9b21e0d56041983 and merged it to master.

PR #167, opened 2015-12-11
MBS-8626-8628: add icons for area, event, instrument, and series

caroline-gschwend:


reosarevok:
Note: This shouldn't be merged for now, since we might get more versions of the icons and in that case we'll decide once all are in which icons to pick :)

caroline-gschwend:
New design scheme of MB. Currently working on updating all MB entity icons to look better with the new face of MB.

PR #168, opened 2015-12-13
Use Moose::Util::does_role in another place

uklauer:
Use `does_role` in the edit search, too, because `->does` isn’t working there either. Currently, edit searches that time out show the generic timeout page instead of returning to the search form.


uklauer:
Should go into beta despite the freeze because it resolves a beta-only bug.

mwiencek:
:shipit:

PR #170, opened 2015-12-14
MBS-8665 - Auto add space after ; in join phrases

caroline-gschwend:


root/static/scripts/edit/MB/Control/ArtistCredit.js
```diff
@@ -76,7 +76,9 @@ MB.Control.ArtistCreditName = aclass(MB.entity.ArtistCreditName, {
 
         } else if (/^&$/.test(join)) {
             this.joinPhrase(" & ");
-        }
+        }else if(/^;$/.test(join) ){
```
reosarevok: Minor, but you could change the spaces here so that they are like the others (before "else", after "if", after "join)". Similarly, you seem to have more spaces than the rest of the file - it's generally best to make everything have the same style, so please take a look at changing that :)

Freso: s/could/should/ ;)

We generally follow [PEP8](https://www.python.org/dev/peps/pep-0008/) with regards to writing Python code. This is the standard used for the Python project itself and far most Python projects adopt most or all of this guideline as well, so it is useful to be familiar with it and get used to writing in the style it dictates if you plan on contributing to projects using Python code (within MetaBrainz or elsewhere). :)

gentlecat: Python sure looks weird here. :snake:

Freso: Yeah, I think I was in CB mode and didn't pay mind to the `{}`s. I need to do fewer things at the same time…

uklauer:
There is another whitespace issue: The new lines use a tab for indentation, but as a matter of style we only use spaces. So you should replace each of the tab characters with eight regular spaces to get the same indentation.

Then after that is fixed, I think you should combine the (by then probably four) commits into one. Best to ask on IRC about how to do this, because it is a bit more complicated – you’d first need to clean up the rogue commit from your master and beta branches.

PR #172, opened 2015-12-15
Search for edits by edit note author (MBS-4565) plus associated autocomplete fixes (MBS-7873, MBS-8051)

mwiencek:


mwiencek:
In b093629 I removed our own `encode_entities` subroutine in favor of the one from HTML::Entities, since we already use that elsewhere. I'll have to double-check if this will cause any problems, but I don't think it should.

The only reason I did this is because it resolves a circular dependency introduced in f2ae146.

uklauer:
I object to using `HTML::Entities::encode_entities`.

> We already use HTML::Entities all over the place, so it can't possibly be broken.

The function is used in two places exactly (introduced quite recently) and should be replaced with “our” `encode_entities` even there. While it may do exactly what it promises to do and also not technically break anything, it is a horrible idea to inflate something like `西東京市` into `&#x897F;&#x6771;&#x4EAC;&#x5E02;` everywhere in our HTML. It makes the source hard to read when debugging and will probably break some simple scripts that don’t expect these entities. [Here is the W3C’s take on it.](http://www.w3.org/International/questions/qa-escapes#not)

`HTML::Entities::encode_entities` can do what we want if it is told which characters to encode, in a second parameter: `encode_entities($string, q{<>&"'})` So if you think a library function is preferable, we could change the implementation of `MB::S::V::encode_entities` to call the `HTML::Entities` version in that way, but adding the second argument to every `encode_entities` call in the codebase is not a good idea IMO, since it is bound to be forgotten next time a call is added.

Where is this circular dependency BTW? As far as I see, `MB::S::Validation` now imports `MB::S::Constants`, but `Constants` still doesn’t depend on any `MB` modules at all.

t/lib/t/MusicBrainz/Server/Validation.pm
```diff
@@ -187,10 +199,6 @@ test 'Test is_valid_partial_date' => sub {
     };
 };
 
-test 'Test encode_entities' => sub {
-    is(encode_entities('ãñ><"\'&Å'), 'ãñ&gt;&lt;&quot;&#39;&amp;Å');
```
uklauer: With the other `encode_entities`, this expands to `&atilde;&ntilde;&gt;&lt;&quot;&#39;&amp;&Aring;`.

mwiencek:
@uklauer thanks for looking more closely at `HTML::Entities::encode_entities`. The circular import appears to be via `DBDefs`, which `MB::S::Constants` imports. Here's the trace:

```
Error while loading /Users/Michael/code/musicbrainz-server/app.psgi: Can't locate object method "MB_SERVER_ROOT" via package "DBDefs" at lib/MusicBrainz/Server/Constants.pm line 358.
Compilation failed in require at lib/MusicBrainz/Server/Validation.pm line 70.
BEGIN failed--compilation aborted at lib/MusicBrainz/Server/Validation.pm line 70.
Compilation failed in require at lib/MusicBrainz/Server/Translation.pm line 15.
BEGIN failed--compilation aborted at lib/MusicBrainz/Server/Translation.pm line 15.
Compilation failed in require at lib/DBDefs/Default.pm line 34.
BEGIN failed--compilation aborted at lib/DBDefs/Default.pm line 34.
Compilation failed in require at /System/Library/Perl/5.18/parent.pm line 20.
BEGIN failed--compilation aborted at lib/DBDefs.pm line 29.
Compilation failed in require at /Users/Michael/code/musicbrainz-server/app.psgi line 5.
BEGIN failed--compilation aborted at /Users/Michael/code/musicbrainz-server/app.psgi line 5.
```

I'll drop b093629 and figure out a way around it.

uklauer:
Ah, `DBDefs::Default` uses `MusicBrainz::Server::Translation` for translating some banners, which in turn uses `encode_entities`

uklauer:
Hm, `encode_entities` is not really about validation, though. We could perhaps move it elsewhere.

mwiencek:
I was thinking of moving it to `MB::S::Filters`. I see a lot of other functions in `MB::S::Validation` that aren't related to validation, but I'll leave them alone for now.

mwiencek:
Heh, that didn't help, since `MB::S::Filters` also imports `MB::S::Constants`. Instead I just did https://github.com/mwiencek/musicbrainz-server/commit/dbb7b548f190acbb15a0d224337dccc763e52438#diff-e11edc95d90abe8b020956c678bb56afL355 but clearly our modules are not granular enough in places.

gentlecat:
I think now I understand what was causing MBS-7873. The rest looks good as well. :shipit:

PR #174, opened 2015-12-15
MBS-8690: Rejecting add-entity edits with subscriptions

mwiencek:
The ticket was about artists, but the same issue exists for labels and series, so I fixed those too.


gentlecat:
LGTM :ship:

PR #175, opened 2015-12-16
Mandate JSON::XS and kill JSON::Any

mwiencek:
This simplifies our JSON story somewhat, and avoids cases like the one described in the comment in app.psgi. (See: http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-12-03/?msg=3422352&page=5)

The reason I set an environment variable instead of just importing `JSON::XS` everywhere is, `JSON::XS` doesn't export things like `null`, `true`, `false`, etc., and it's also a huge amount of work to update everything, for questionable gain.

I did however use `JSON::XS` directly in various admin scripts, rather than making sure the environment variable is set everywhere.

In the process of fixing things, I converted the deprecated `jsonToObj` and `objToJson` methods to `decode` and `encode`, respectively.

Bizarrely, the `JSON::Any` `utf8` flag has the exact opposite meaning that `JSON` uses. You can see where it's flipped in the `JSON::Any` source code, and it became obvious once I started getting mojibake in certain places while testing. So, everywhere that `utf8 => 1` was passed to `JSON::Any->new`, we now do *not* pass the `utf8` flag, which is to say it should have the exact same behavior as before.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -282,7 +282,6 @@ sub cover_art_uploader : Chained('load') PathPart('cover-art-uploader') Edit {
 sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
     my ($self, $c) = @_;
     my $entity = $c->stash->{$self->{entity_name}};
-    my $json = JSON::Any->new( utf8 => 1 );
 
     $c->model('Release')->load_meta($entity);
 
@@ -307,7 +306,7 @@ sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
         images => \@artwork,
         mime_types => \@mime_types,
         access_key => DBDefs->COVER_ART_ARCHIVE_ACCESS_KEY // '',
-        cover_art_types_json => $json->encode(
+        cover_art_types_json => $c->json->encode(
             [ map {
                 { name => $_->name, l_name => $_->l_name, id => $_->id }
             } $c->model('CoverArtType')->get_all() ]),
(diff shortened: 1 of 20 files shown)


lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -307,7 +306,7 @@ sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
         images => \@artwork,
         mime_types => \@mime_types,
         access_key => DBDefs->COVER_ART_ARCHIVE_ACCESS_KEY // '',
-        cover_art_types_json => $json->encode(
+        cover_art_types_json => $c->json->encode(
```
gentlecat: Is `$c->json` a reference to one of those modules? (I can't find it in Catalyst docs.)

Also, how does it know which module to use? By reading `PERL_JSON_BACKEND` variable?

mwiencek: Yes, `$c` is an instance of `MusicBrainz::Server`, which extends `Catalyst` with our own stuff, and that's where we define `json`: https://github.com/metabrainz/musicbrainz-server/blob/12dfab53283d573d2160698a9c6b9dcccb20597e/lib/MusicBrainz/Server.pm#L431

So, it's just a plain attribute pointing to a `JSON` instance we can reuse. The `JSON` module is what reads the `PERL_JSON_BACKEND` environment variable to decide what backend to use.

gentlecat: Ah, that's why I couldn't find it. Thanks!

gentlecat:
LGTM :shipit:

PR #176, opened 2015-12-16
MBS-8676: Add a page to see all notes left on your own edits

mwiencek:
I hacked in a basic notification banner for when this page changes, using redis. It stores two keys per user; when they last viewed the page, and when it was last updated. (These expire in a month, so we don't store unused data indefinitely.) The latter is bumped in `Data::EditNote::add_note` (so now we have two problems, besides emails :)).

There's also a new option cookie on the page to disable the banner permanently. It can't be toggled by limited users.

One problem is that if someone were to load all the individual edits they received notes on, the banner still wouldn't go away until they loaded `/edit/notes-received`, which is strange, and I'm unsure of a good solution. (Perhaps there isn't one, except to wait for a proper notification system.)

### DIFF
--- a/lib/MusicBrainz/DataStore/Redis.pm
+++ b/lib/MusicBrainz/DataStore/Redis.pm
@@ -48,6 +48,14 @@ sub get {
     return defined $value ? $self->_json->decode($value) : undef;
 }
 
+sub mget {
+    my ($self, @keys) = @_;
+
+    map {
+        defined $_ ? $self->_json->decode($_) : undef
+    } $self->_connection->mget(map { encode('utf-8', $self->prefix . $_) } @keys);
+}
+
 sub set {
     my ($self, $key, $value) = @_;
 
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -278,6 +278,23 @@ sub subscribed_editors : Local RequireAuth {
     load_everything_for_edits($c, $edits);
 }
 
+sub notes_received : Path('/edit/notes-received') RequireAuth {
+    my ($self, $c) = @_;
+
+    my $edit_notes = $self->_load_paged($c, sub {
+        $c->model('EditNote')->find_by_recipient($c->user->id, shift, shift);
+    });
+
+    $c->model('Editor')->load(@$edit_notes);
+    $c->model('Edit')->load_for_edit_notes(@$edit_notes);
+    $c->model('Vote')->load_for_edits(map { $_->edit } @$edit_notes);
+
+    $c->stash(
+        edit_notes => $edit_notes,
+        template => 'edit/notes-received.tt',
+    );
+}
+
 =head2 conditions
 
 Display a table of all edit types, and their relative conditions
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -281,6 +281,14 @@ sub subscribed_editors : Local RequireAuth {
 sub notes_received : Path('/edit/notes-received') RequireAuth {
     my ($self, $c) = @_;
 
+    # Log when the editor loaded the page, so that we know when to notify them
+    # again about new edits (see Data::EditNote::add_note).
+    my $redis = $c->model('MB')->context->redis;
+    my $notes_viewed_key = 'edit_notes_received_last_viewed:' . $c->user->name;
+    $redis->set($notes_viewed_key, time);
+    # Expire the notification in 30 days.
+    $redis->expire($notes_viewed_key, 60 * 60 * 24 * 30);
+
     my $edit_notes = $self->_load_paged($c, sub {
         $c->model('EditNote')->find_by_recipient($c->user->id, shift, shift);
     });
--- a/lib/MusicBrainz/Server/Data/EditNote.pm
+++ b/lib/MusicBrainz/Server/Data/EditNote.pm
@@ -85,6 +85,13 @@ sub add_note
         (map { $_->editor_id } @{ $edit->edit_notes }));
     $self->c->model('Editor')->load_preferences(values %$editors);
 
+    # Inform the edit's author that a note was left.
+    my $edit_author = $editors->{$edit->editor_id}->name;
+    my $notes_updated_key = "edit_notes_received_last_updated:$edit_author";
+    $self->c->redis->set($notes_updated_key, time);
+    # Expire the notification in 30 days.
+    $self->c->redis->expire($notes_updated_key, 60 * 60 * 24 * 30);
+
     my @to_email = grep { $_ != $note_hash->{editor_id} }
         map { $_->id } grep { $_->preferences->email_on_notes }
         map { $editors->{$_->editor_id} }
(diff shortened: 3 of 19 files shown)


root/layout.tt
```diff
@@ -71,6 +71,17 @@
             </div>
         [%- END -%]
 
+        [%- IF new_edit_notes && new_edit_notes_mtime >
+                (c.req.cookies.new_edit_notes_dismissed_mtime.value || 0) -%]
+            <div class="banner new-edit-notes">
+                <p>
+                    [% l('Notes have been left on some of your edits. Please {uri|click here} to read them.',
```
uklauer: `... edits. Read {link|your edits with new notes} now.` or something?

Background: http://www.w3.org/QA/Tips/noClickHere, http://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/

(Regarding `uri`, not all translators might understand it.)

lib/MusicBrainz/Server/Data/EditNote.pm
```diff
@@ -106,6 +113,27 @@ sub add_note
...
+           AND edit IN (SELECT id FROM edit WHERE editor = $1)
+         ORDER BY post_time DESC, edit DESC
+EOSQL
+    my ($edit_notes, $hits) = $self->query_to_list_limited(
+        $query, [$recipient_id], $limit, $offset, undef, 1
+    );
+
+    $self->c->model('Editor')->load(@$edit_notes);
+    $self->c->model('Edit')->load_for_edit_notes(@$edit_notes);
+    $self->c->model('Vote')->load_for_edits(map { $_->edit } @$edit_notes);
```
uklauer: Wouldn’t we normally leave this to the controller?

lib/MusicBrainz/Server/Data/EditNote.pm
```diff
@@ -106,6 +113,27 @@ sub add_note
     }
 }
 
+sub find_by_recipient {
+    my ($self, $recipient_id, $limit, $offset) = @_;
+
+    my $query = <<'EOSQL';
+        SELECT editor, edit, text, post_time
```
uklauer: I’d suggest using `$self->_columns`. Or is there a special reason why you left out `id` here? Seems likely to me that the missing ID will cause issues down the line (e.g. isn’t there an `#anchor` on edit pages that contains the note ID?). – Also `$self->_table` below, probably, although changing the table name is less likely.

lib/MusicBrainz/Server/Data/EditNote.pm
```diff
@@ -106,6 +113,27 @@ sub add_note
...
+          FROM edit_note
+         WHERE editor != $1
+           AND edit IN (SELECT id FROM edit WHERE editor = $1)
+         ORDER BY post_time DESC, edit DESC
```
uklauer: This will be difficult performance-wise for editors with many edits.

Though I guess reosarevok, drsaunde and company will continue to rely on email anyway. ;-)

root/layout.tt
mwiencek: I used `{link|New notes} have been left on some of your edits. Please make sure to read them and respond if necessary.` to sound slightly less commanding, let me know if that can be improved.

lib/MusicBrainz/Server/Data/EditNote.pm
mwiencek: Right, moved.

lib/MusicBrainz/Server/Data/EditNote.pm
mwiencek: :+1: Done.

lib/MusicBrainz/Server/Data/EditNote.pm
mwiencek: This I can't really fix right now, we'll have to wait and see. :|

mwiencek:
I removed "WIP" since this is in a better state now, with an added option to disable the banner permanently. And I made it so that limited users can't toggle that option.

mwiencek:
And now the Reactified 404 page shows the banner (and all the banners can be dismissed on that page, since that was apparently added after it was originally written...).

gentlecat:
I was trying to make the banner show up by adding comments to both closed and open edits with another user, but it didn't work. These notes show up at */edit/notes-received* and notifications are enabled.

lib/MusicBrainz/Server/Controller/Edit.pm
```diff
@@ -278,6 +278,31 @@ sub subscribed_editors : Local RequireAuth {
     load_everything_for_edits($c, $edits);
 }
 
+sub notes_received : Path('/edit/notes-received') RequireAuth {
+    my ($self, $c) = @_;
+
+    # Log when the editor loaded the page, so that we know when to notify them
+    # again about new edits (see Data::EditNote::add_note).
+    my $redis = $c->model('MB')->context->redis;
+    my $notes_viewed_key = $c->user->name . '|edit_notes_received_last_viewed';
```
uklauer: I think we should name the keys a bit differently. Catalyst uses `session:$sid`, `expires:$sid`, `flash:$sid` (where `$sid` stands for the session ID), so perhaps `edit_notes_received_last_viewed:$user` and `edit_notes_received_last_updated:$user`? I know the remember-me tokens use `$user|$token` without indication what they are about, but I’ve thought before that `remember-me:$user:$token` (or something) might be better; it’s just complicated to change without breaking existing sessions, so I haven’t done it yet.

lib/MusicBrainz/DataStore/Redis.pm
```diff
@@ -48,6 +48,14 @@ sub get {
     return defined $value ? $self->_json->decode($value) : undef;
 }
 
+sub mget {
+    my ($self, @keys) = @_;
+
+    map {
+        defined $_ ? $self->_json->decode($_) : ()
```
uklauer: This should be `... : undef`, not `... : ()`. As it is, undefined values are completely dropped from the list; i.e. if `alert_mtime` and `notes_viewed` are `undef`, the value for `notes_updated` will end up in the position for `alert_mtime`.

uklauer:
After the Redis fix, banners were working on my sandbox, but not on the 404 page for some reason.

lib/MusicBrainz/DataStore/Redis.pm
mwiencek: Argh. It was working locally because I had alert, alert_mtime keys in redis, so the array got shifted enough. Thanks for finding that!

lib/MusicBrainz/Server/Controller/Edit.pm
mwiencek: Sure. I think I only looked at the remember-me tokens and copied their pattern.

mwiencek:
Looking on rika, I see that the node process for your user has `--development 0`, which means the modules containing all the components are cached, perhaps prior to switching branches. With `--development 1` it should reload the JS modules on every page refresh.

But, the node process should also be restarting whenever the perl does, due to c1c6f3302e657f0e93ee984aea3a2b98d1dbcb1e, so it sounds like that may not be working there.

Another possibility is that there's another bug I'm not hitting on my local machine. :)

uklauer:
Ah, right, I get `Error: listen EADDRINUSE 127.0.0.1:9009` all the time when the server restarts, so it does seem the node renderer survives.

gentlecat:
OK, seems to work well now. :ship:

PR #173, opened 2015-12-15
#MBS-8482 Renamed "Add Disc" button in release editor to "Add Medium"

stasszczesniak:
MB's button to add another medium to a release is labeled "Add disc", but a lot of our releases don't have discs (they have cassettes, or USBs, or just folders in a download). So i renamed that to "Add medium", which is the more MusicBrainz-y term.

While i was at it  i changed the mentions of "tracklist" to say "medium" too, where it says "search for an existing tracklist", "Existing tracklist" and "To create a new tracklist, use an existing tracklist or".

[Link to the ticket](http://tickets.musicbrainz.org/browse/MBS-8482)
[Link to the GCI task](https://codein.withgoogle.com/dashboard/task-instances/6181327283945472/)

PR #177, opened 2015-12-16
#MBS-8482 Renamed "Add Disc" button in release editor to "Add Medium"

stasszczesniak:
MB's button to add another medium to a release is labeled "Add disc", but a lot of our releases don't have discs (they have cassettes, or USBs, or just folders in a download). So i renamed that to "Add medium", which is the more MusicBrainz-y term.

While i was at it  i changed the mentions of "tracklist" to say "medium" too, where it says "search for an existing tracklist", "Existing tracklist" and "To create a new tracklist, use an existing tracklist or".

[Link to the ticket](http://tickets.musicbrainz.org/browse/MBS-8482)
[Link to the GCI task](https://codein.withgoogle.com/dashboard/task-instances/6181327283945472/)

Sorry for this previous request, i didn't understand what is the role of "po" directory.
I think it should be ok now. :+1:

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -90,7 +90,7 @@
   </table>
 
   <!-- ko if: searchResults() && searchResults().length > 0 -->
-  <p>[% l('Select a disc from the search results below and then click "Add Disc".') %]</p>
+  <p>[% l('Select a disc from the search results below and then click "Add Medium".') %]</p>
   <!-- /ko -->
 
   <!-- ko if: totalPages() > 1 -->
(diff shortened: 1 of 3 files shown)


uklauer:
I think that catches all relevant strings. The internal names (variables, element IDs) probably can stay as they are, since they aren’t user-visible and changing them might break userscripts.

stasszczesniak:
I was thinking also about changing : root/release/discids.tt:  
```
<p>[% l('There are no disc IDs attached to this release; to find out more about how to add one, see {doc|How to Add Disc IDs}.', { doc => doc_link('How_to_Add_Disc_IDs') }) %]</p>

```

uklauer:
Uh, no. https://github.com/metabrainz/musicbrainz-server/pull/173#discussion_r47682354

root/release/edit/tracklist.tt
```diff
@@ -90,7 +90,7 @@
   </table>
 
   <!-- ko if: searchResults() && searchResults().length > 0 -->
-  <p>[% l('Select a disc from the search results below and then click "Add Disc".') %]</p>
+  <p>[% l('Select a disc from the search results below and then click "Add Medium".') %]</p>
```
mwiencek: Perhaps "Select a medium ..." here too.

mwiencek:
This looks good to me, but before merging I think we should fix https://github.com/metabrainz/musicbrainz-server/pull/177#discussion_r47817554 while we're changing that string, so that it's consistent with itself. :)

stasszczesniak:
Definitely this sentence made no sense. I've changed it now so everything should be ok now.

PR #178, opened 2015-12-18
MBS-8241: Highlight mediums with pending edits

uklauer:
The medium header still had the `class="mp"`, but the CSS to turn it orange was missing.

This doesn’t completely match the style of the regular header, which has a bit of 3D. I leave fixing that to someone who cares. ;-) At least this fixes the functionality.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -491,6 +491,7 @@ div.warning img.warning {
 .tooltip { border-bottom: 1px dotted; } /* Useful when adding title attributes to elements */
 
 table.tbl tr.mp td,
+table.tbl tr.mp th, /* Medium headers in release view */
 span.mp,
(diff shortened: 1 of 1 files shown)


uklauer:
Example: http://chirlu.mbsandbox.org/release/4f34e8a2-fc31-4972-acbb-dd5af6d59d10 (CD 4)

jesus2099:
It works great on your sandbox ! Thanks ! :)

Just for what it’s worth (probably nothing) I used the following selector : 

```css
div#content > table.tbl.medium > thead > tr.mp
```

The “`div#content >`” part is most probably useless but I let you judge if it is useful or not to restrict the rest as I did or to leave it more generous as you did. :)

mwiencek:
:shipit:

PR #179, opened 2015-12-18
MBS-8691: Avoid “subscribed editor edits” timeouts

uklauer:
The query for edits by subscribed editors took too long time after the recent changes to it; the cause and the cure are the same as for the subscribed entities (commit 0dc12b9177f97a24cd29b093c2e55226c8ad92a2).

### DIFF
--- a/lib/MusicBrainz/Server/Data/Edit.pm
+++ b/lib/MusicBrainz/Server/Data/Edit.pm
@@ -330,16 +330,7 @@ sub subscribed_entity_edits {
     my $table = $self->_table;
     my @args = ($editor_id); # $1
 
-    if ($only_open) {
-        push @args, $STATUS_OPEN;
-    } else {
-        # $3
-        # XXX This won't handle all cases in which an edit is stuck open, but
-        # it is significantly faster to leave out the "OR status = 1" filter
-        # when showing all recent edits.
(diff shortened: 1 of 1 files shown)


uklauer:
I branched this off from production, so it could be merged first into beta and later separately into production, if it works OK on beta and is decided to be hotfixed.

(I have some more XSS hotfixes in the pipeline, by the way.)

lib/MusicBrainz/Server/Data/Edit.pm
```diff
@@ -330,16 +330,7 @@ sub subscribed_entity_edits {
...
 
-    if ($only_open) {
-        push @args, $STATUS_OPEN;
-    } else {
-        # $3
-        # XXX This won't handle all cases in which an edit is stuck open, but
-        # it is significantly faster to leave out the "OR status = 1" filter
-        # when showing all recent edits.
-        my $max_open_duration = $OPEN_EDIT_DURATION->clone->add_duration($MINIMUM_RESPONSE_PERIOD)->add(hours => 1);
-        push @args, DateTime::Format::Pg->format_interval($max_open_duration);
```
uklauer: Factored out into a function to be used by both methods.

mwiencek:
:shipit:

PR #180, opened 2015-12-19
MBS-8658: Include tag name on user/tag/tagname pages

opatel99:
Related ticket: http://tickets.musicbrainz.org/browse/MBS-8658

### DIFF
--- a/root/user/tag.tt
+++ b/root/user/tag.tt
@@ -1,10 +1,11 @@
 [% WRAPPER 'user/profile/layout.tt' page='tags' %]
 
 [% IF tag_in_use %]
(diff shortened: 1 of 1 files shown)


root/user/tag.tt
```diff
@@ -1,10 +1,11 @@
 [% WRAPPER 'user/profile/layout.tt' page='tags' %]
 
 [% IF tag_in_use %]
+    <h2>[% tag_name %]</h2>
```
mwiencek: This should be escaped like `[% tag_name | html %]` if you don't want arbitrary scripts executing on the page. ;)

mwiencek:
LGTM.

PR #181, opened 2015-12-20
MBS-8692: Set an expire time for all Catalyst session keys

uklauer:
Of the three types of session sub-keys used by Catalyst, only two had an expire time set in order to get them auto-removed from Redis when the session was no longer valid, but `expires` had not. This was introduced in commit e7e15ef, most likely unintentionally and a side effect of `expires` being treated specially for encoding purposes. It has led to the production Redis instance accumulating over 4.8 million keys since early 2013, most of which presumably are `expires` subkeys of long-gone Catalyst sessions.

Ensure that a Redis expire time is set for all session keys equally. This also adds a script that will clean out existing expired session keys, and does some cleanup for `MB::DataStore:Redis`.

### DIFF
--- a/lib/Catalyst/Plugin/Session/Store/MusicBrainz.pm
+++ b/lib/Catalyst/Plugin/Session/Store/MusicBrainz.pm
@@ -30,15 +30,11 @@ sub get_session_data {
 sub store_session_data {
     my ($self, $key, $data) = @_;
 
-    if (my ($sid) = $key =~ /^expires:(.*)/)
-    {
-        $self->_datastore->set($key, $data);
+    unless ($key =~ /^expires:/) {
+        $data = encode_base64(nfreeze($data));
     }
-    else
-    {
-        $self->_datastore->set($key, encode_base64(nfreeze($data)));
-        $self->_datastore->expireat($key, $self->session_expires);
-    }
+    $self->_datastore->set($key, $data);
+    $self->_datastore->expireat($key, $self->session_expires);
 }
 
 sub delete_session_data {
--- a/lib/MusicBrainz/DataStore.pm
+++ b/lib/MusicBrainz/DataStore.pm
@@ -22,25 +22,6 @@ Expire the specified key at (unix) $timestamp.
 
 requires 'expireat';
 
-=method incr
-
-Increment the value for the $key.
-
-=cut
-
-requires 'incr';
-
-=method add
-
-Store the $value on the server under the $key, but only if the key
-doesn't exists on the server.
-
-=cut
-
-requires 'add';
-
-requires 'ttl';
-
 =head1 LICENSE
 
 Copyright (C) 2013 MetaBrainz Foundation
--- a/lib/MusicBrainz/DataStore.pm
+++ b/lib/MusicBrainz/DataStore.pm
@@ -12,7 +12,7 @@ Expire the specified key in $s seconds.
 
 =cut
 
-requires 'expireat';
+requires 'expire';
 
 =method expireat
 
--- a/lib/MusicBrainz/DataStore/Redis.pm
+++ b/lib/MusicBrainz/DataStore/Redis.pm
@@ -40,39 +40,48 @@ has '_json' => (
     default => sub { return JSON->new->allow_nonref->ascii; }
 );
 
+sub _prepare_key {
+    my ($self, $key) = @_;
+    return encode('utf-8', $self->prefix . $key);
+}
+
+sub _decode_value {
+    my ($self, $value) = @_;
+    return defined $value ? $self->_json->decode($value) : undef;
+}
+
 sub get {
     my ($self, $key) = @_;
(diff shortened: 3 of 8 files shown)


uklauer:
`admin/RemoveExpiredSessions.pl` could be run up to three times:

1. (optional) Now, to remove the millions of old keys.
2. At least two hours after the fix was deployed to production, in order to remove the keys created in the meantime.
3. (optional) Some days later, to verify that no new keys lacking an expire time have been created.

lib/Catalyst/Plugin/Session/Store/MusicBrainz.pm
```diff
@@ -30,15 +30,11 @@ sub get_session_data {
...
-    else
-    {
-        $self->_datastore->set($key, encode_base64(nfreeze($data)));
-        $self->_datastore->expireat($key, $self->session_expires);
+    unless ($key =~ /^expires:/) {
+        $data = encode_base64(nfreeze($data));
     }
+    $self->_datastore->set($key, $data);
+    $self->_datastore->expireat($key, $self->session_expires);
 }
```
uklauer: This is the main fix: Call `expireat` for *all* keys.
Plus some removal of duplicate code.

lib/MusicBrainz/DataStore.pm
```diff
@@ -12,7 +12,7 @@ Expire the specified key in $s seconds.
 
 =cut
 
-requires 'expireat';
+requires 'expire';
```
uklauer: Copy-and-paste error, `expireat` was `require`d twice.

lib/MusicBrainz/DataStore.pm
```diff
@@ -22,25 +22,6 @@ Expire the specified key at (unix) $timestamp.
...
-
-Store the $value on the server under the $key, but only if the key
-doesn't exists on the server.
-
-=cut
-
-requires 'add';
-
-requires 'ttl';
-
```
uklauer: These weren’t used except in tests (for `add` see below).

lib/MusicBrainz/DataStore/Redis.pm
```diff
@@ -40,39 +40,48 @@ has '_json' => (
...
+sub _prepare_key {
+    my ($self, $key) = @_;
+    return encode('utf-8', $self->prefix . $key);
+}
+
+sub _decode_value {
+    my ($self, $value) = @_;
+    return defined $value ? $self->_json->decode($value) : undef;
+}
+
```
uklauer: Gave these repeated code snippets their own methods. They are short but not necessarily trivial and might conceivably change at some point.

lib/MusicBrainz/Script/RemoveExpiredSessions.pm
```diff
@@ -0,0 +1,102 @@
...
+sub run {
+    my ($self, @args) = @_;
+    die "Usage error ($0 takes no arguments)" if @args;
+
+    my $r = $self->c->redis;
+    my $now = time + 5; # five seconds grace period for server differences
+
+    printf "Fetching entries from database; prefix used is \"%s\".\n", $r->prefix if $self->verbose;
+    my @keys = $r->_connection->keys($r->prefix . "expires:*");
+        # KEYS is very heavy, but our current Redis doesn't have SCAN
```
uklauer: I had originally written the script to use [SCAN](http://redis.io/commands/scan), which would be a perfect fit here. Then I made a test run and noticed the command is unknown in our Redis version …

So this will read 4.8 million keys in one go. Let’s see if it works. ;-) (6000 keys were no problem in a test run.)

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -612,7 +612,7 @@ sub allocate_remember_me_token {
         my $token = generate_token();
 
         my $key = "$normalized_name|$token";
-        $self->redis->add($key, 1);
+        $self->redis->set($key, 1);
```
uklauer: The difference between `add` ([SETNX](http://redis.io/commands/setnx)) and `set` was that `add` only set the key if it didn’t exist before. However, as the value is always 1 anyway, the only difference would be that setting a value also resets the expire time; but we explicitly re-set it in the next command. So, no added value in having a separate `add`.

mwiencek:
Alright, let's merge this and attempt a dry-run + run on astro. Then I'll report back here. :)

mwiencek:
Looks like it chokes trying to mget 4.8M values.

```
musicbrainz@astro:~/musicbrainz-server$ ./admin/RemoveExpiredSessions.pl --verbose --dry-run
Fetching entries from database; prefix used is "MB:".
Redis::X::Reconnect=SCALAR(0x5b912e98)Retrieved 4755935 keys,musicbrainz@astro:~/musicbrainz-server$
```

I guess we should upgrade redis.

PR #183, opened 2015-12-20
MBS-8167: Use HTML diffs for relationship attributes

uklauer:
Relationship attributes can now contain HTML (instrument links); diffing them as regular text for edit display may lead to broken tags. Use HTML diffs instead.


uklauer:
Example: http://chirlu.mbsandbox.org/edit/31282806 vs. https://beta.musicbrainz.org/edit/31282806

mwiencek:
:shipit:

PR #184, opened 2015-12-20
Add “use utf8” where necessary and missing

uklauer:


mwiencek:
:shipit:

mwiencek:
This broke the `Edit::Recording::Create` tests, and I can't really figure out why. :/

But basically, it appears to be wrapping the SQL in that test file with `Encode::decode('utf-8', ...)`, which throws an exception because it's already decoded? But the file is saved as UTF-8.

uklauer:
Yes, I saw; I think I know why.

PR #185, opened 2015-12-21
MBS-8381: Don’t double-encode area names in inline search

uklauer:
Area names in containments were doubly encoded in the inline search; see the screenshot in the ticket.

### DIFF
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -779,7 +779,7 @@ MB.Control.autocomplete_formatters = {
         if (item.typeName || (item.containment && item.containment.length)) {
             var items = [];
             if (item.typeName) {
-                items.push(_.escape(item.typeName));
+                items.push(item.typeName);
             }
             items.push(renderContainingAreas(item));
             a.append('<br /><span class="autocomplete-comment">' +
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -924,7 +924,7 @@ function getCatalogNumber(releaseLabel) {
 }
 
 function renderContainingAreas(area) {
-    return i18n.commaOnlyList(_(area.containment).pluck('name').map(_.escape).value());
+    return i18n.commaOnlyList(_(area.containment).pluck('name').value());
```
uklauer: I’m now encoding everything immediately before it goes into the DOM, not here.

root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -779,7 +779,7 @@ MB.Control.autocomplete_formatters = {
         if (item.typeName || (item.containment && item.containment.length)) {
             var items = [];
             if (item.typeName) {
-                items.push(_.escape(item.typeName));
+                items.push(item.typeName);
```
uklauer: This was not the visible issue; but if there are (translated) area type names with special characters, they would have been encode here and again on line 786, below.

uklauer:
To test, visit https://beta.musicbrainz.org/artist/create and http://chirlu.mbsandbox.org/artist/create and search for Yekaterinburg in one of the area autocompletes. On the sandbox, you can also test the place autocomplete by searching for “Test Place” (direct search).

mwiencek:
:shipit:

PR #186, opened 2015-12-21
MBS-6407 Add username to account verification email

caroline-gschwend:


lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -526,7 +526,8 @@ sub _send_confirmation_email
         $c->model('Email')->send_email_verification(
             email             => $email,
             verification_link => $verification_link,
-            ip                => $c->req->address
+            ip                => $c->req->address,
+            user_id           => $c->request->params->{userid}
```
uklauer: This won’t work; there normally is no such parameter, and if there were, it would just be some random string.

The information about the user is in `$editor`.

PR #191, opened 2015-12-22
Added a notice showing when an edit changes only relationship credits

stasszczesniak:
User might change only relationship credits and not the actual relationship and it would have looked the same. I've added a notice, which shows only if relationship credits were changed and everything else remains the same.
[Link to the ticket](http://tickets.musicbrainz.org/browse/MBS-8662)
[Link to the GCI task](https://codein.withgoogle.com/dashboard/task-instances/5300055393173504/?sp-page=1)

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -1054,3 +1054,5 @@ END -%]
     </div>
   [%- END -%]
 [%~ END ~%]
+
+[%~ MACRO non_empty(text) GET (text != '' && text.defined) ~%]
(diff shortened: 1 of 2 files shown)


root/components/common-macros.tt
```diff
@@ -1054,3 +1054,5 @@ END -%]
     </div>
   [%- END -%]
 [%~ END ~%]
+
+[%~ MACRO non_empty(text) (text != '' && text.defined) ~%]
```
mwiencek: The syntax is wrong here. :) You're missing at least BLOCK, some semicolons, and END. See for example https://github.com/stasszczesniak/musicbrainz-server/blob/master/root/components/common-macros.tt#L7

PR #192, opened 2015-12-22
Name-variation fixes (includes MBS-7844, MBS-8330)

uklauer:
We wrap entites in `<span class="name-variation">` if the displayed name is not identical to the true entity name so that client-side CSS can highlight them. However, there were some cases where this was done or omitted erroneously:

- If the display name is from a diff and already in HTML form, it is necessary not only to remove any `<span>` tags (for diff highlighting) before the comparison – we already did that – but also to undo the conversion of HTML reserved characters into entities. Otherwise, names containing one of the reserved characters were treated as a variation even when they were, in fact, identical to the entity name. (MBS-8330)
- Artist names in artist credit diffs from the `Diff` plugin were linked inside the plugin, using what was essentially a Perl reimplementation of the `link_artist` template macro. However, this knock-off did not even attempt to create the name-variation marker. (MBS-7844) It has even more bugs for which no separate tickets exist; see the comments on MBS-7844. Instead of trying to fix it, I just made the plugin call the TT macro directly, removing the duplication and ensuring feature parity. For this, I needed to extend `link_artist` slightly so that it can take HTML (like `link_recording` already did).

### DIFF
--- a/lib/MusicBrainz/Server/Plugin/Diff.pm
+++ b/lib/MusicBrainz/Server/Plugin/Diff.pm
@@ -124,26 +124,11 @@ sub _render_side_diff {
 
 sub _link_artist_credit_name {
     my ($self, $acn, $name) = @_;
-    my $comment;
-    if ($acn->artist->comment) {
-        $comment = ' (' . $acn->artist->comment . ')';
-    }
-    else {
-        $comment = '';
-    }
 
-    if ($acn->artist->gid) {
-        return $h->a({
-            href => $self->uri_for_action('/artist/show', [ $acn->artist->gid ]),
-            title => encode_entities($acn->artist->sort_name . $comment),
-        }, $name || encode_entities($acn->name));
-    }
-    else {
-        return $h->span({
-            class => 'deleted tooltip',
-            title => l('This entity has been removed, and cannot be displayed correctly.')
-        }, $name || encode_entities($acn->name));
-    }
+    $name //= encode_entities($acn->name);
+
+    # defer to the template macro
+    return $self->{c}->stash->get('link_artist', [ $acn->artist, 'show', $name, 1 ]);
 }
 
 sub _link_joined {
--- a/lib/MusicBrainz/Server/Plugin/Diff.pm
+++ b/lib/MusicBrainz/Server/Plugin/Diff.pm
@@ -201,7 +201,6 @@ sub diff_artist_credits {
 
 sub diff {
     my ($self, $old, $new) = @_;
-    my $h = HTML::Tiny->new;
 
     my @spans;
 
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Plugin/Diff.pm
```diff
@@ -216,7 +201,6 @@ sub diff_artist_credits {
 
 sub diff {
     my ($self, $old, $new) = @_;
-    my $h = HTML::Tiny->new;
```
uklauer: Not really related. There is already a package-local variable `$h` (created on line 39 of the file) that is used everywhere else except for this function; and there is no need to have a separate instance here.

lib/MusicBrainz/Server/Plugin/Diff.pm
```diff
@@ -124,26 +124,11 @@ sub _render_side_diff {
...
-    else {
-        return $h->span({
-            class => 'deleted tooltip',
-            title => l('This entity has been removed, and cannot be displayed correctly.')
-        }, $name || encode_entities($acn->name));
-    }
+    $name //= encode_entities($acn->name);
+
+    # defer to the template macro
+    return $self->{c}->stash->get('link_artist', [ $acn->artist, 'show', $name, 1 ]);
```
uklauer: This is equivalent to `link_artist(acn.artist, 'show', name, 1)` in a template.

uklauer:
An example demonstrating both issues is https://beta.musicbrainz.org/edit/27510463 vs. http://chirlu.mbsandbox.org/edit/27510463

- The title for track 5 is marked as having a name variation on beta, which is wrong; caused by the encoded apostrophe in the name.
- On the other hand, the artist for track 9 lacks the name-variation marker in the tracklist, even though the name is different (correctly shown in the artist credit part below, which is created in a different way).

root/components/common-macros.tt
```diff
@@ -6,6 +6,7 @@
 [%~ MACRO replace(text, search, replace) BLOCK; text | replace(search, replace); END -%]
 [%~ MACRO url_escape(url) BLOCK; url | url; END -%]
 [%~ MACRO html_escape(url) BLOCK; url | html; END -%]
+[%~ MACRO html_unescape(text) BLOCK; text.replace('&quot;', '"').replace('&lt;', '<').replace('&gt;', '>').replace('&#39;', "'").replace('&amp;', '&'); END -%]
```
mwiencek: Would it work to just put `html_unescape => &HTML::Entities::decode_entities` in the stash?

I've wondered why we don't do the same for `html_escape`, in order to use our own `encode_entities`.

mwiencek:
:shipit:

PR #194, opened 2015-12-27
Fix double-encoding issues after XSS fixes (includes MBS-8702, MBS-8705)

uklauer:
The recent XSS fixes, which included HTML encoding the content of the `<title>` tag by the global layout template, exposed some places in individual templates where this was inconsistently already done, now causing double-encoding. (One of these was MBS-8702, but I did a – hopefully – complete review of all page titles and found some more places where this was happening, such as FreeDB entry pages.)

Similarly, the XSS fixes to `diff_side` revealed a latent bug in that this function was (erroneously) used to diff HTML (release events in “edit release” edits, MBS-8705). This should have been `diff_html_side` all the time.


uklauer:
Compare http://chirlu.mbsandbox.org/recording/d5137fd2-1f6d-45b4-a9ee-d0e639a15a73 to https://beta.musicbrainz.org/recording/d5137fd2-1f6d-45b4-a9ee-d0e639a15a73 for the first issue (page titles), http://chirlu.mbsandbox.org/edit/33191193 to https://beta.musicbrainz.org/edit/33191193 for the second.

root/entity/edits.tt
```diff
@@ -2,8 +2,8 @@
     SET type_for_css = ent_type;
     SET type_for_css = 'rg' IF ent_type == 'release_group';
 
-    SET name_for_title = html_escape(entity.name);
-    SET name_for_title = html_escape(entity.l_name) IF ent_type == 'instrument';
+    SET name_for_title = entity.name;
+    SET name_for_title = entity.l_name IF ent_type == 'instrument';
     SET name_for_title = display_url(entity) IF ent_type == 'url';
```
uklauer: This is actually slightly broken, too, but I’m replacing `display_url` in another branch anyway.

root/entity/edits.tt
```diff
@@ -11,7 +11,7 @@
 
     SET layout_title = all_edits ? l('Edits for {name}', { name => name_for_title }) : l('Open Edits for {name}', { name => name_for_title });
     SET heading = all_edits ? l('Edits for {name}', { name => link_for_heading }) : l('Open Edits for {name}', { name => link_for_heading });
-    SET subheading = html_escape(entity.l_type_name) or format_entity_type_name(ent_type);
+    SET subheading = html_escape(entity.l_type_name or format_entity_type_name(ent_type));
```
uklauer: Vaguely related; missing encoding for the type name (but it’s for the heading, not the title, and current type names don’t contain special characters).

root/release/edit/layout.tt
```diff
@@ -1,6 +1,6 @@
 [%- PROCESS 'release/edit/macros.tt' -%]
 
-[%- WRAPPER 'layout.tt' full_width=1 edit=1 title=page_title -%]
+[%- WRAPPER 'layout.tt' full_width=1 edit=1 -%]
```
uklauer: The `page_title` was never set anywhere. It’s all handled in JavaScript.

root/user/edits.tt
```diff
@@ -27,6 +27,6 @@
     END ~%]
 
 [% WRAPPER 'user/profile/layout.tt' title=page_title %]
-   <h2>[% page_heading %]</h2>
+   <h2>[% page_heading | html %]</h2>
```
uklauer: Because the above `param` is used both for the title and the heading, I had to add the encoding for the heading.

PR #195, opened 2015-12-27
MBS-8694 (beta): Re-allow 0 for edit search “ID” fields, for now

uklauer:
Because the “ID” edit search predicate class is used not only for IDs, but also other numbers such as the Yes/No vote counts, the recently improved validation for database row IDs broke those other criteria.

Re-allow 0 (which does not cause problems when used in a database query; it will simply never find anything) as a temporary solution until the predicate class hierarchy is improved.


uklauer:
I’ve been working on edit search improvements including the inheritance hierarchy, but it’s not ready yet. However, this particular thing should be fixed/worked around before the release.

PR #196, opened 2015-12-28
URL fixes, part 1 (includes MBS-8413)

uklauer:
This is the first part of a number of loosely URL-related improvements. Having only one pull request would make it difficult to review.

The main change here is to give all the link macros the responsability to entity-encode their parameters, instead of the callers having to do that (who will forget it). This affects the hover text (`title` attribute) and the `href` attribute itself.

Other changes are the removal of the `display_url` macro, because it didn’t serve any useful purpose, and of the `generic_link` macro, which two templates had locally to do the same as `simple_link`.

Finally, this has a fix for MBS-8413, where deleted URL entities (in the edit history) were doubly percent-encoded, breaking the link.

### DIFF
--- a/lib/MusicBrainz/Server/Filters.pm
+++ b/lib/MusicBrainz/Server/Filters.pm
@@ -224,6 +224,7 @@ sub format_editnote
          [<a href="//$server/edit/$1">edit #$1</a>]gi;
 
     # links to wikidocs
+    # (only safe because \w doesn't match any of the HTML reserved characters)
     $html =~ s/doc:(\w[\/\w]*)(``)*/<a href="\/doc\/$1">$1<\/a>/gi;
     $html =~ s/\[(\p{IsUpper}[\/\w]*)\]/<a href="\/doc\/$1">$1<\/a>/g;
 
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -10,12 +10,6 @@
 [%~ MACRO bracketed(text) BLOCK; text != '' ? l(' ({text})', { text => text }) : ''; END -%]
 [%~ MACRO js_escape(text) BLOCK; text | js; END ~%]
 [%~ MACRO closing_tag_escape(text) BLOCK; text.replace('</', '<\\/'); END ~%]
-[%~ MACRO display_url(url) BLOCK; PROCESS _unescaped_display_url | html; END -%]
-[%~ BLOCK _unescaped_display_url;
-    IF url.pretty_name; url.pretty_name;
-    ELSE; url | uri_decode;
-    END;
-END -%]
 [%- USE UserDate(c.user.preferences, current_language) -%]
 [%- USE Countdown -%]
 [%~ MACRO doc_link(to) BLOCK -%][% c.uri_for('/doc', to) %][%- END -%]
@@ -361,7 +355,7 @@ END -%]
     END;
 
     IF type == 'url' AND content == '';
-        mod_content = INCLUDE _unescaped_display_url url=entity;
+        mod_content = entity.pretty_name;
         options.unshift('info_link');
         infolink = link;
         link = entity.href_url;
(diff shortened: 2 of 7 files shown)


lib/MusicBrainz/Server/Filters.pm
```diff
@@ -224,6 +224,7 @@ sub format_editnote
          [<a href="//$server/edit/$1">edit #$1</a>]gi;
 
     # links to wikidocs
+    # (only safe because \w doesn't match any of the HTML reserved characters)
```
uklauer: I thought this should be documented, in case someone at some future point intends to change the regex.

root/components/common-macros.tt
```diff
@@ -5,17 +5,11 @@
 [%~ MACRO add_commas(n) GET n.chunk(-3).join(',') -%]
 [%~ MACRO replace(text, search, replace) BLOCK; text | replace(search, replace); END -%]
 [%~ MACRO url_escape(url) BLOCK; url | url; END -%]
-[%~ MACRO html_escape(url) BLOCK; url | html; END -%]
+[%~ MACRO html_escape(text) BLOCK; text | html; END ~%]
```
uklauer: Just a naming issue … The argument to `html_escape` can be a URL, but most often is some other text.

root/components/common-macros.tt
```diff
@@ -5,17 +5,11 @@
...
 [%~ MACRO html_unescape(text) BLOCK; text.replace('&quot;', '"').replace('&lt;', '<').replace('&gt;', '>').replace('&#39;', "'").replace('&amp;', '&'); END -%]
 [%~ MACRO bracketed(text) BLOCK; text != '' ? l(' ({text})', { text => text }) : ''; END -%]
 [%~ MACRO js_escape(text) BLOCK; text | js; END ~%]
 [%~ MACRO closing_tag_escape(text) BLOCK; text.replace('</', '<\\/'); END ~%]
-[%~ MACRO display_url(url) BLOCK; PROCESS _unescaped_display_url | html; END -%]
-[%~ BLOCK _unescaped_display_url;
-    IF url.pretty_name; url.pretty_name;
-    ELSE; url | uri_decode;
-    END;
```
uklauer: Essentially, this distinguished between URL entities (having a `pretty_name` method) and other URI objects. For the former, I just replaced the macro call by `x.pretty_name`; for the latter (of which there only was one case), we can let the URI object do the decoding.

root/components/common-macros.tt
```diff
@@ -249,7 +243,7 @@ END -%]
       ELSIF entity.isa('MusicBrainz::Server::Entity::URL'); link_url(entity, action, text);
       END;
     ELSIF entity.isa('MusicBrainz::Server::Entity::URL');
-      simple_link(url_escape(entity.url), display_url(entity.url));
+      simple_link(entity.href_url, entity.pretty_name);
```
uklauer: This was MBS-8413, the double-encoding issue. Using `url_escape` is practically always wrong (I’ll remove it altogether in another part).

root/relationship/header.tt
```diff
@@ -1,7 +1,6 @@
-[% MACRO generic_link(url, text) BLOCK %]<a href="[% url %]">[% text %]</a>[% END %]
```
uklauer: This file and `root/user/profile/layout.tt` had this macro. You can see on http://chirlu.mbsandbox.org/user/Psychotic%20&%20Q-riosity that the ampersand is correctly encoded as `&amp;` now in the `href`, whereas https://beta.musicbrainz.org/user/Psychotic%20%26%20Q-riosity has illegal HTML.

uklauer:
As an example for MBS-8413, compare http://chirlu.mbsandbox.org/edit/33112104 to https://beta.musicbrainz.org/edit/33112104

mwiencek:
LGTM.

PR #197, opened 2015-12-28
URL fixes, part 2

uklauer:
More about URLs:
- Remove the `url_escape` macro and all uses of the `url` filter (all bugs, the `html` filter was intended).
- Fix report-a-bug links on error pages: They used the wrong JIRA base URL that doesn’t fully accept seeding, had double-encoding issues, and (due to a mistake I made when preparing this patch series) still used the `display_url` macro that was already removed in part 1.
- In some cases, URIs inside an URL entity were changed unintentionally. Use `clone` to operate on copies instead.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/URL.pm
+++ b/lib/MusicBrainz/Server/Entity/URL.pm
@@ -73,13 +73,14 @@ sub url_is_scheme_independent { 0 }
 
 sub href_url {
     my $self = shift;
-    my $url = $self->affiliate_url();
+    my $url = $self->affiliate_url;
 
-    if ($self->url_is_scheme_independent()) {
+    if ($self->url_is_scheme_independent) {
+        $url = $url->clone;
         $url->scheme("");
     }
 
-    return $url;
+    return $url->as_string;
 }
 
 __PACKAGE__->meta->make_immutable;
--- a/lib/MusicBrainz/Server/Entity/URL/ASIN.pm
+++ b/lib/MusicBrainz/Server/Entity/URL/ASIN.pm
@@ -27,14 +27,16 @@ sub sidebar_name { shift->pretty_name }
 override affiliate_url => sub {
     my $self = shift;
 
-    my $url = super()->clone;
+    my $url = super();
     if ($url =~ m{^(?:https?:)?//(?:.*?\.)(amazon\.([a-z\.]+))(?:\:[0-9]+)?/[^?]+$}i) {
         my $ass_id = DBDefs->AWS_ASSOCIATE_ID($1);
         return $url unless $ass_id;
-        return URI->new("$url?tag=$ass_id");
-    } else {
+
+        $url = $url->clone;
+        $url->query_form(tag => $ass_id);
         return $url;
     }
+    return $url;
 };
 
 sub url_is_scheme_independent { 1 }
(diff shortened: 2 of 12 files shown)


uklauer:
Example for the clone issues: https://beta.musicbrainz.org/url/c4b79260-68d5-4bda-8ea2-9c68d8ca5620 has a schema-less URL in the heading, even though what is actually stored is an `http:` URL. Fixed on http://chirlu.mbsandbox.org/url/c4b79260-68d5-4bda-8ea2-9c68d8ca5620

lib/MusicBrainz/Server/Entity/URL.pm
```diff
@@ -73,13 +73,14 @@ sub url_is_scheme_independent { 0 }
 
 sub href_url {
     my $self = shift;
-    my $url = $self->affiliate_url();
+    my $url = $self->affiliate_url;
 
-    if ($self->url_is_scheme_independent()) {
+    if ($self->url_is_scheme_independent) {
```
uklauer: The above is only coding style.

lib/MusicBrainz/Server/Entity/URL/ASIN.pm
```diff
@@ -27,14 +27,16 @@ sub sidebar_name { shift->pretty_name }
 override affiliate_url => sub {
     my $self = shift;
 
-    my $url = super()->clone;
+    my $url = super();
```
uklauer: In this subclass, it wasn’t done wrong; but it’s slightly more efficient to only clone when actually changing the URL.

root/components/common-macros.tt
```diff
@@ -1097,3 +1103,9 @@ note-[%- edit.id -%]-[%- loop.count() -%]
     </div>
   </div>
 [%~ END ~%]
+
+[%~ MACRO bugtracker_url(description) BLOCK;
```
uklauer: I made a macro (two, actually, there is also a JavaScript implementation below) to have less duplication.

root/components/common-macros.tt
```diff
@@ -1097,3 +1103,9 @@ note-[%- edit.id -%]-[%- loop.count() -%]
...
+    'http://tickets.musicbrainz.org/secure/CreateIssueDetails!init.jspa?' _
```
uklauer: Only in this form, JIRA accepts a seeded description. `http://tickets.musicbrainz.org/secure/CreateIssue.jspa` simply ignored everything beyond project ID and issue type.

root/user/profile.tt
```diff
@@ -99,8 +99,8 @@
 
         [% IF user.website && (user.has_confirmed_email_address || c.user.is_account_admin) %]
             [% WRAPPER property name=l("Homepage:") %]
-                <a href="[% user.website | url %]" rel="nofollow">
-                    [%~ user.website | html_entity ~%]
+                <a href="[% user.website | html %]" rel="nofollow">
+                    [%~ user.website | html ~%]
```
uklauer: `html_entity` wasn’t wrong, strictly speaking, but used unnecessarily many `&#123;` entities.

mwiencek:
lgtm.

mwiencek:
In Beta Testing. Not sure if there were any tickets for this one. :)

PR #198, opened 2016-01-01
Split both before and after separator for diffs (fixes MBS-5205)

uklauer:
Please look at the two commits for explanation.

Result: http://chirlu.mbsandbox.org/edit/18645604 vs. https://beta.musicbrainz.org/edit/18645604


mwiencek:
I wish we had tests for this, but lgtm.

PR #201, opened 2016-01-03
MBS-8713 (Block /search and /tags SoundCloud URLs)

opatel99:
I regex'ed a negative lookahead for "search" and "tags" after soundcloud.com to block all URLs calling those endpoints by using (?! ___ ).

Additionally, I added a missing validator script for LINK_TYPES.soundcloud.series

[MBS-8713](http://tickets.musicbrainz.org/browse/MBS-8713)

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -862,13 +862,13 @@ validationRules[LINK_TYPES.songfacts.work] = function (url) {
 };
 
 // allow only Soundcloud pages with the Soundcloud rel
-validationRules[LINK_TYPES.soundcloud.artist] = function (url) {
-  return /soundcloud\.com\//.test(url);
-};
+function validateSoundCloud(url) {
+  return /soundcloud\.com\/(?!(search|tags)[\/?#])/.test(url);
+}
 
-validationRules[LINK_TYPES.soundcloud.label] = function (url) {
-  return /soundcloud\.com\//.test(url);
-};
+_.each(LINK_TYPES.soundcloud, function (id) {
+  validationRules[id] = validateSoundCloud;
+}
 
 // allow only VIAF pages with the VIAF rel
 validationRules[LINK_TYPES.viaf.artist] = function (url) {


uklauer:
Using a negative assertion is good. However, since it is conceivable SoundCloud might allow account names such as `searchingforsomething` or `tagsoup`, it should be a little bit more specific – block only if the `search` or `tags` is followed by `[/?#]`.

The (moderately complex) regex is repeated three times. I think it would be preferrable to have a common validation function for SoundCloud URLs that could then be reused ([see a similar case here](https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/edit/URLCleanup.js#L778-L783)).

Finally, the first commit message has two bugs:
> MB-8713: Blocked all URLs with /search or /tags
The `S` is missing from `MBS`, and not _all_ such URLs are blocked, only the SoundCloud ones. :-)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -862,13 +862,13 @@ validationRules[LINK_TYPES.songfacts.work] = function (url) {
...
+function validateSoundCloud(url) {
+  return /soundcloud\.com\/(?!(search|tags)[\/?#])/.test(url);
+}
 
-validationRules[LINK_TYPES.soundcloud.label] = function (url) {
-  return /soundcloud\.com\//.test(url);
-};
+validationRules[LINK_TYPES.soundcloud.artist] = validateSoundCloud;
+validationRules[LINK_TYPES.soundcloud.label] = validateSoundCloud;
+validationRules[LINK_TYPES.soundcloud.series] = validateSoundCloud;
```
mwiencek: Instead of assigning all of these directly (which is how we (I?) missed series), I think we should just do something like the wikipedia ones: https://github.com/metabrainz/musicbrainz-server/blob/1724da055c5d6485c0e7a9ff2de6c3ebf748810d/root/static/scripts/edit/URLCleanup.js#L828-L830

mwiencek:
lgtm.

uklauer:
:shipit:, but probably squashed.

PR #202, opened 2016-01-04
URL fixes, part 3 (mostly internationalized domain names/MBS-8683)

uklauer:
Percent-encoded UTF-8 in the local part of URLs has long been decoded for display, but punycoded internationalized domain names were shown as raw ASCII.
    
This uses the existing IRI support of the `URI` module to decode all parts of URLs at the same time. Some new attributes (so that the decoding only has to be done once) are added to the URL base class for access to the decoded URL or just the decoded local part and for checking whether the local part was in some non-UTF-8 legacy encoding.
    
Finally, the `pretty_name` method of the base class and overridden display names in the subclasses make use ofthe decoded form, as do two templates that don’t use `pretty_name` but always show the actual URL.

Additional changes:
- One of my previous related pull requests left a spurious `</a>` that is now removed.
- Added an `MB::S::Entity::URL::MediaWiki` role to hold some code that the classes for MediaWiki sites duplicated.
- Ozon.ru was marked as scheme independent, i.e. links from HTTPS MusicBrainz led to their HTTPS site. However, their certificate is broken (only valid for ozone.ru), so let’s stick with HTTP for now.

### DIFF
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -59,7 +59,6 @@ __PACKAGE__->config(
             'format_wikitext' => \&MusicBrainz::Server::Filters::format_wikitext,
             'format_editnote' => \&MusicBrainz::Server::Filters::format_editnote,
             'format_setlist' => \&MusicBrainz::Server::Filters::format_setlist,
-            'uri_decode' => \&MusicBrainz::Server::Filters::uri_decode,
             'language' => \&MusicBrainz::Server::Filters::language,
             'locale' => \&MusicBrainz::Server::Filters::locale,
             'gravatar' => \&MusicBrainz::Server::Filters::gravatar,
--- a/lib/MusicBrainz/Server/Entity/URL.pm
+++ b/lib/MusicBrainz/Server/Entity/URL.pm
@@ -1,42 +1,72 @@
 package MusicBrainz::Server::Entity::URL;
 use Moose;
 
-use Encode 'decode';
 use MooseX::Types::URI qw( Uri );
-use MusicBrainz::Server::Filters;
-use URI::Escape;
-use Try::Tiny;
 
 extends 'MusicBrainz::Server::Entity::CoreEntity';
 with 'MusicBrainz::Server::Entity::Role::Linkable';
 
 has 'url' => (
-    is => 'rw',
+    is => 'ro',
     isa => Uri,
     coerce => 1
 );
 
-=attribute utf8_decoded
+has 'iri' => (
+    is => 'ro',
+    isa => 'Str',
+    lazy => 1,
+    default => sub { shift->url->as_iri },
+);
 
-Returns the URL, with entities unescaped and the string decoded from utf-8 into
-a Perl string. If the decoding fails, then the URL is probably not in utf-8
-encoding, and `undef` is returned.
+=attribute uses_legacy_encoding
+
+Indicates whether the URL contains bytes that can't be interpreted as UTF-8.
 
 =cut
 
-has utf8_decoded => (
+has 'uses_legacy_encoding' => (
     is => 'ro',
+    isa => 'Bool',
+    lazy => 1,
+    default => sub { shift->iri =~ /%[89A-F]/ },
+        # as_iri only leaves bytes with bit 7 set percent-encoded if they
+        # are not part of a valid UTF-8 sequence. ASCII characters may
+        # still be percent-encoded (e.g. %25, the percent sign itself).
+);
+
+=attribute decoded
+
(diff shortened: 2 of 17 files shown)


lib/MusicBrainz/Server.pm
```diff
@@ -59,7 +59,6 @@ __PACKAGE__->config(
             'format_wikitext' => \&MusicBrainz::Server::Filters::format_wikitext,
             'format_editnote' => \&MusicBrainz::Server::Filters::format_editnote,
             'format_setlist' => \&MusicBrainz::Server::Filters::format_setlist,
-            'uri_decode' => \&MusicBrainz::Server::Filters::uri_decode,
```
uklauer: No longer needed. Decoding is mostly handled by `URI::as_iri`, plus some custom code in our URL class.

lib/MusicBrainz/Server/Entity/URL.pm
```diff
@@ -1,42 +1,72 @@
...
-use MusicBrainz::Server::Filters;
-use URI::Escape;
-use Try::Tiny;
 
 extends 'MusicBrainz::Server::Entity::CoreEntity';
 with 'MusicBrainz::Server::Entity::Role::Linkable';
 
 has 'url' => (
-    is => 'rw',
+    is => 'ro',
```
uklauer: The attribute was never set except on construction, so this doesn’t really change anything. However, because I added a number of lazy attributes that cache values computed from the `url` attribute here, I thought I should mark it as constant.

lib/MusicBrainz/Server/Entity/URL/IMSLP.pm
```diff
@@ -44,7 +38,7 @@ IMSLP URLs are only show in the sidebar if the URL can be decoded from utf-8
 
 =cut
 
-sub show_in_sidebar { defined(shift->utf8_decoded) }
+sub show_in_sidebar { !shift->uses_legacy_encoding }
```
uklauer: Interestingly, the string value of the old `utf8_decoded` attribute was used nowhere outside the base class; all subclasses only checked for definedness (was undefined for legacy-encoded URLs) and otherwise did their own decoding.

root/url/index.tt
```diff
@@ -3,7 +3,7 @@
     <table class="details">
         <tr>
             <th>[% l('URL:') %]</th>
-            <td>[% simple_link(url.href_url, url.pretty_name) %]</a></td>
+            <td>[% simple_link(url.href_url, url.pretty_name) %]</td>
```
uklauer: The `</a>` was from before `simple_link`.

t/lib/t/MusicBrainz/Server/Controller/URL/Show.pm
```diff
@@ -15,7 +15,7 @@ my $c    = $test->c;
 MusicBrainz::Server::Test->prepare_test_database($c, '+url');
 
 $mech->get_ok('/url/25d6b63a-12dc-41c9-858a-2f42ae610a7d');
-$mech->content_contains('http://zh-yue.wikipedia.org/wiki/%E7%8E%8B%E8%8F%B2');
+$mech->content_contains('http://zh-yue.wikipedia.org/wiki/王菲');
 $mech->content_contains('zh-yue: 王菲');
```
uklauer: Reversal of the test, as I warned before. :) I left the additional test, though.

BTW, the full URL in decoded form that the test checks for is in the heading [here](http://chirlu.mbsandbox.org/url/25d6b63a-12dc-41c9-858a-2f42ae610a7d), it was never in an `href`.

uklauer:
Examples:
- http://chirlu.mbsandbox.org/url/2de1616a-7ca0-4688-92cc-0a8373190ede vs. https://beta.musicbrainz.org/url/2de1616a-7ca0-4688-92cc-0a8373190ede (Cyrillic IDN)
- http://chirlu.mbsandbox.org/edit/33191804 (made-up URL with Japanese IDN and percent-encoded local part, including a control character)
- http://chirlu.mbsandbox.org/url/25d6b63a-12dc-41c9-858a-2f42ae610a7d (the page also used by the test)

uklauer:
Hm, it seems Ozon.ru fixed their certificate issues; unless they are transient … I’ll drop that commit.

t/lib/t/MusicBrainz/Server/Controller/URL/Show.pm
mwiencek: Yeah, I was looking at the scheme-less one. Woops.

mwiencek:
Nice work. Let's merge now for wider testing.

PR #204, opened 2016-01-07
MBS-8722: Fail edits that would violate group_type_implies_null_gender

mwiencek:


uklauer:
1. This doesn’t catch orchestras and choirs; those can’t have a gender, either.

2. While double sure is often better, the checks at initialization time would actually make this three-and-a-half sure, and I feel that’s unnecessary. On the client side, the “gender” drop-down is disabled for groups (this is the half, because a malicious user can circumvent it); then on the server, form validation makes sure inconsistent data never makes it into an edit (https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Form/Artist.pm#L65); and if this check really failed somehow, there still is the database constraint.

3. For the case from the ticket, I feel a better place for the check would be around the `merge_changes` method; it is responsible for these cases where an entity has had changes between creation and application of an edit, and already raises the alarm on incompatible ones (e.g. edit 1: gender “unknown → male”, edit 2: gender “unknown → female”); it just doesn’t know about entity-specific dependencies between fields, of course. It could look like this:

  ```perl
  around merge_changes => sub {
      my ($orig, $self, @args) = @_;
      my $merged = $self->$orig(@args);
      if (defined $merged->{gender_id} && $self->current_instance->type_id “is group”) {
          MusicBrainz::Server::Edit::Exceptions::FailedDependency->throw('fooo');
      }
      return $merged;
  };
  ```

gentlecat:
How did that edit (https://beta.musicbrainz.org/edit/36566039) even happen? I see in the [edit history](https://beta.musicbrainz.org/artist/1f199bfe-14a6-4d3a-b7bd-c11211b9e402/edits) for this artist that type has been changed a couple of times. Is that because of ones that failed before?

uklauer:
No – here is a (condensed) timeline of events:
1. artist is created without type or gender
2. 2015-12-26T08:06 – edit is made to change type to group; edit stays open for voting
3. 2015-12-26T09:13 – the problematic edit is made, to change gender to male; edit stays open for voting
4. 2016-01-02T09:00 – the first edit is applied, setting the type to group
5. 2016-01-02T10:00 – the second edit is supposed to be applied, setting the gender to male, but fails

The other edits failed without further problem because they tried to change the type from nothing to person, when it no longer was nothing. That they also tried to change the gender didn’t matter.

mwiencek:
Re checking this during initialization, it was less about being extra-sure and more about providing a human-readable error message before it hits the database. It's not a problem now, but it's likely the entity-creation dialogs will use /ws/js/edit eventually. The edit classes are already responsible for making sure they don't insert something illegal in `accept`, so it doesn't seem like a stretch to do the same in `initialize` for edit types that insert things immediately.

uklauer:
In my opinion, `/ws/js/edit` should validate the input, just like FormHandler does … It should also construct new `%opts` hashes for creating edits, based on a whitelist, instead of passing them on mostly as submitted.

But anyway, points 1 and 3 still stand. :)

mwiencek:
@uklauer thanks for the bug report + suggestions. look better?

uklauer:
:shipit:

PR #205, opened 2016-01-07
[WIP] One macro to link them all

chirlu:
_NB. This can’t be merged yet or things will break; there are still `link_area` and `link_artist` calls that need to be replaced. I put it up so that @mwiencek can consider the changed `link_entity` syntax for the JavaScript version._

Up to now, we had a generic `link_entity` macro that can link the core entities, but also individual `link_artist`, `link_label`, `link_recording` etc. macros. The only difference left was that `link_artist` etc. would never show a disambiguation comment; `link_entity` would if it also displayed the entity name (instead of some random text provided as the third argument) **or** if it displayed some random text provided as the _fourth_ argument. To make this syntax clearer, I did the following:
- Have only one “random text to replace the entity name”, as the third argument.
- Make the macro accept a hash with options as the fourth parameter, with a `show_disambiguation` option explicitly controlling whether the disambiguation is shown or not. (No option in the hash = default behaviour, as before.)
- Turn the `no_escape` flag, up to now the fifth parameter, into another option (under the name `already_html`).

I then proceeded to adapt the (few) `link_entity` calls that made use of the fourth/fifth parameters to the new syntax, and to replace the remaining direct calls to `link_label` etc. with `link_entity` calls; if necessary, with `show_disambiguation => 0`. _(Not complete yet.)_ Lastly, I removed all the `link_foo` macros; this saved a considerable amount of duplication  and one level of indirection, which might or might not result in a perceivable performance improvement (macro calls in TT are relatively slow).

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -234,25 +234,19 @@ END -%]
     # - show_disambiguation (omit for default behaviour)
     # - already_html: the text argument is HTML; no entity encoding
     #   will be done.
-    #   already_html is only available for artists and recordings.
     show_disambiguation = text == '';
     SET show_disambiguation = opts.show_disambiguation
       IF opts.show_disambiguation.exists;
     IF entity.gid;
-      IF    entity.isa('MusicBrainz::Server::Entity::Artist'); link_artist(entity, action, text, opts.already_html);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Area'); link_area(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Collection'); link_collection(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Work'); link_work(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Event'); link_event(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Instrument'); link_instrument(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Label'); link_label(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Place'); link_place(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::Release'); link_release(entity, action, text);
-      ELSIF entity.isa('MusicBrainz::Server::Entity::ReleaseGroup'); link_release_group(entity, action, text);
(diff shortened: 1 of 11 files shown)


root/components/common-macros.tt
```diff
@@ -229,27 +229,24 @@ END -%]
...
-      END;
+      ent_type = get_entity_type(entity);
+      INCLUDE _link_mbid_entity entity=entity
+          type=ent_type action=action content=text
+          noescape=opts.already_html
+          hover=ent_type == 'artist' ?
+              artist.sort_name _ bracketed(artist.comment) : ''
+          default_content=(ent_type == 'area' OR ent_type == 'instrument') ?
+              entity.l_name : ''
+          namevar=(ent_type == 'artist' OR ent_type == 'recording');
```
uklauer: I’m wondering (but that would be for a different PR) if we shouldn’t extend name-variation to all entity types (except URLs, probably, which don’t actually have a name). Perhaps @jesus2099 has an opinion about that?

mwiencek:
At a glance I think the changes make sense. What do you think the scope of link_entity should be? Right now it supports entities that have MBIDs, plus editors, for some reason. I'd like not to repeat any design mistakes. Judging by its name, I'd think it should support any entity—but trying to do that all in a single macro might be a mistake if it requires juggling tons of different flags. (I do agree that supporting all the core/MBID entities in a single macro makes sense, though.)

mwiencek:
By "any entity" I mean any linkable entity, of course (the ones that use `_link_other_entity` come to mind).

I don't plan on deviating from what the TT versions support right now, but I thought it was a relevant question.

uklauer:
I’m actually not sure why it’s just editors in addition to the relatable entities. I believe that was done before I first started working on these macros. One thing they do have in common with the relatable entities is that there are autocomplete fields for them, but I don’t think that would require `link_entity` in any way.

Essentially, we have four groups of entities:
- relatable ones
- editors (have avatars for which one can specify a size to `link_editor`, whereas `link_entity` only supports the default)
- other non-MBID entities: tags, ISWCs, CD stubs, etc. (those relying on `_link_other_entity`)
- non-relatable “special” entities with MBID: tracks, link types, link attributes (very rarely linked to, no macro exists for them)

I think we should leave the fourth group alone for now. For the others, I don’t see a particular advantage or disadvantage in having them in one macro.

root/components/common-macros.tt
jesus2099: It sounds good. But it's just my opinion. :)
I haven't thought much about it to know if it is appropriate or not.
For labels I have thought about it already, at least...

mwiencek:
Going to close this, since we won't be using the TT macro much longer. It certainly did influence the React one, though!

PR #206, opened 2016-01-10
MBS-8730: Reordering of sub work parts causes unwanted reordering of main work parts

mwiencek:
This changes `getRelationshipGroup` to consider direction in addition to the link type.

### DIFF
--- a/root/static/scripts/relationship-editor/common/dialog.js
+++ b/root/static/scripts/relationship-editor/common/dialog.js
@@ -511,17 +511,13 @@ const PART_OF_SERIES_LINK_TYPE_GIDS = _.values(PART_OF_SERIES_LINK_TYPES);
     });
 
     function addRelationships(relationships, source, viewModel) {
-        var linkType = relationships[0].linkTypeInfo();
-
         _.each(relationships, function (relationship) {
             if (source.mergeRelationship(relationship)) {
                 return;
             }
 
-            var typeInfo = relationship.linkTypeInfo();
-
-            if (typeInfo.orderableDirection) {
-                var group = source.getRelationshipGroup(typeInfo.id, viewModel);
+            if (relationship.linkTypeInfo().orderableDirection) {
+                var group = source.getRelationshipGroup(relationship, viewModel);
                 var maxLinkOrder = -Infinity;
 
                 _.each(group, function (other) {
--- a/root/static/scripts/relationship-editor/common/entity.js
+++ b/root/static/scripts/relationship-editor/common/entity.js
@@ -6,6 +6,18 @@
 var mergeDates = require('./mergeDates.js');
 var deferFocus = require('../../edit/utility/deferFocus.js');
 
+function getDirection(relationship, source) {
+  let entities = relationship.entities();
+
+  if (source === entities[0]) {
+    return 'forward';
+  }
+
+  if (source === entities[1]) {
+    return 'backward';
+  }
+}
+
 (function (RE) {
 
     MB.entity.CoreEntity.extend({
@@ -64,7 +76,7 @@ var deferFocus = require('../../edit/utility/deferFocus.js');
                 var dialog = RE.UI.AddDialog({
                     source: self,
                     target: MB.entity({}, firstRelationship.target(self).entityType),
-                    direction: self === firstRelationship.entities()[1] ? "backward" : "forward",
+                    direction: getDirection(firstRelationship, self),
                     viewModel: vm
                 });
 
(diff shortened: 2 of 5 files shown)


root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -511,17 +511,13 @@ const PART_OF_SERIES_LINK_TYPE_GIDS = _.values(PART_OF_SERIES_LINK_TYPES);
     });
 
     function addRelationships(relationships, source, viewModel) {
-        var linkType = relationships[0].linkTypeInfo();
```
mwiencek: unused variable.

root/static/scripts/relationship-editor/common/entity.js
```diff
@@ -6,6 +6,18 @@
 var mergeDates = require('./mergeDates.js');
 var deferFocus = require('../../edit/utility/deferFocus.js');
 
+function getDirection(relationship, source) {
```
mwiencek: Direction is relative to the entity you're editing from, which is what `source` is.

root/static/scripts/relationship-editor/common/entity.js
```diff
@@ -6,6 +6,18 @@
...
+  let entities = relationship.entities();
+
+  if (source === entities[0]) {
+    return 'forward';
+  }
+
+  if (source === entities[1]) {
+    return 'backward';
+  }
+}
```
mwiencek: This will return `undefined` if `source` is not used by the relationship, which should never happen, but if it does I'd rather the function not return an untrue value, hence the explicit checks.

root/static/scripts/tests/relationship-editor.js
```diff
@@ -1003,3 +1003,80 @@ relationshipEditorTest("empty date period fields are outputted when cleared", fu
...
+    var relationship = vm.getRelationship(relData[2], vm.source);
+    relationship.moveEntityDown();
+
+    MB.relationshipEditor.prepareSubmission('edit-work');
+
+    t.deepEqual(formData(), {
+        'edit-work.rel.0.backward': '1',
+        'edit-work.rel.0.link_type_id': '281',
+        'edit-work.rel.0.relationship_id': '55536',
+        'edit-work.rel.0.target': 'dce85d48-7ce6-4bd6-b21e-a7c1be6eb1b4',
```
mwiencek: The relationship editor currently outputs fields for all relationships, even ones that weren't changed, but you'll notice the fields for "The Well-Tempered Clavier, Book I" have no changes to the `link_order`.

root/static/scripts/tests/typeInfo.js
```diff
@@ -4785,7 +4785,7 @@ MB.relationshipEditor.exportTypeInfo({
          "cardinality1" : 0,
          "deprecated" : false,
          "id" : 281,
-         "orderableDirection" : 0,
+         "orderableDirection" : 1,
```
mwiencek: This file should be automatically generated, but...meh. This is the only change we need, and it won't be reverted later, because that's how it is in the database now.

gentlecat: What does `orderableDirection` actually specify?

root/static/scripts/tests/relationship-editor.js
gentlecat: I'm a bit confused about the values of `link_order`. Above it's 15, 0, and 0; then you move last one down. Is it going to be 15, 0, 1 after that?

root/static/scripts/tests/typeInfo.js
uklauer: It’s which end of the relationship can be ordered. To understand what that means, assume we have these “is parent of” relationships:
- Mother is parent of A
- Mother is parent of B
- Father is parent of A
- Father is parent of B

We could now choose whether we want the ordering apply to the parents or to the children. In one case, we would see “children: 1. A, 2. B” on the pages for Mother and Father, and “parents: Father, Mother [no explicit order]” on the pages for A and B; in the other case, we would have “children: A, B [no explicit order]” on the pages for Mother and Father and “parents: 1. Mother, 2. Father” on the pages for A and B.

A value of 0 means that neither of the endpoints is ordered.

mwiencek: That just specifies which entity is being ordered. It can be set to either 1=forward (entity1 is orderable) or 2=backward (entity0 is orderable). It only applies to link types in the `orderable_link_type` table, which is where `orderableDirection` comes from.

In this case it's 1, so entity1 is orderable:
WorkA has part WorkB (link_order: 1)
WorkA has part WorkC (link_order: 2)
WorkA has part WorkD (link_order: 3)

Some link types have an `orderableDirection` of 2, e.g. the event<->series part-of link type:
EventA (link_order: 1) is part of SeriesA
EventB (link_order: 2) is part of SeriesA
EventC (link_order: 3) is part of SeriesA

The entities in an l_ table are positioned alphabetically, e.g. l_**E**vent_**S**eries, so in some cases it's just determined by that.

root/static/scripts/tests/relationship-editor.js
mwiencek: It'll be 15, 1, and 2 (see the fields below).

It's a bit smarter than just incrementing the `link_order` for that specific relationship. If you click some arrow in an orderable group (determined by `getRelationshipGroup`), it knows that the other relationships in the group should also be ordered, so it first moves the relationship in the visual position you want (here visual position 2) and then assigns a non-zero `link_order` (because 0 implies unordered) to every other relationship in the group, based on the positions they were in. (If their link orders are all 0, they'd be sorted alphabetically by default.)

That's all done in https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/relationship-editor/common/fields.js#L370

gentlecat: OK, got it. I also see that first item (`dce85d48-7ce6-4bd6-b21e-a7c1be6eb1b4`) doesn't have `link_order` defined here (was 15). What happened there?

mwiencek: In general it only outputs fields that changed, which is what happened in that case: https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/relationship-editor/generic.js#L196

The other `edit-work.rel.0.*` fields happen to be required fields for *new* relationships. I don't think we need to submit those at all for *edited* relationships anymore, so I'll have to look into removing them.

gentlecat: Understood. :+1:

gentlecat:
LGTM

PR #207, opened 2016-01-10
Consistent JS imports

mwiencek:
I guess I've flip-flopped on this, but I think we should avoid using ES6 import/export statements until they're supported natively in Node.js and https://github.com/whatwg/loader is standardized.

Right now we have a mishmash of require expressions and import statements. There are some cases where we *must* use `require` or `module.exports`, though, because we don't know what we're importing or exporting at compile time. So if we want consistent imports, we have to use CommonJS everywhere.

The import/export statements are compiled into CommonJS expressions anyway, but it can be confusing how they work together, e.g. see 3b2cbed. Using the natively-supported syntax directly makes it easier to see what's actually going on.

Some other changes I made, for consistency:

 * Prefixed all imports with `const`
 * Separated external imports from local imports
 * Sorted imports by module name/path
 * Removed files extensions where they aren't needed (pretty much everywhere)


gentlecat:
LGTM. :+1: for consistency.

root/layout/components/Menu.js
```diff
@@ -3,11 +3,12 @@
...
-import React from 'react';
-import {l, lp} from '../../static/scripts/common/i18n';
-import {VARTIST_GID} from '../../static/scripts/common/constants';
-import EditorLink from '../../static/scripts/common/components/EditorLink';
+const _ = require('lodash');
+const React = require('react');
+
+const EditorLink = require('../../static/scripts/common/components/EditorLink');
+const {VARTIST_GID} = require('../../static/scripts/common/constants');
+const {l, lp} = require('../../static/scripts/common/i18n');
```
gentlecat: How do these work (and one above)?

mwiencek: By "destructuring" https://babeljs.io/docs/learn-es2015/#destructuring

It's basically the same as this:
```
var foo = require(...);
var l = foo.l;
var lp = foo.lp
```
(But obviously only makes sense if the module is exporting an object.)

gentlecat: Oh, so kind of like unpacking in Python? But it's referencing by name instead of position.

mwiencek: Yeah, it's the same thing. You can do it by list position too if you use square brackets like it shows in the link, or by name with curly brackets. Very handy.

PR #208, opened 2016-01-10
Make _get_by_keys return a list instead of a hash reference

mwiencek:
The majority of places that use `_get_by_keys` wrap the result in `values %{}`, so it makes more sense to just return a list.

This was part of my MBS-7241 branch, but since it's taking me a while to finish that, I'm submitting this for review now.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Entity.pm
+++ b/lib/MusicBrainz/Server/Data/Entity.pm
@@ -6,7 +6,6 @@ use Class::MOP;
 use Data::Dumper;
 use Devel::StackTrace;
 use List::MoreUtils qw( part uniq );
-use MusicBrainz::Server::Data::Utils qw( placeholders );
 use MusicBrainz::Server::Validation qw( is_guid is_database_row_id );
 use Carp qw( confess );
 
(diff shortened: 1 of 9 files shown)


lib/MusicBrainz/Server/Data/Entity.pm
```diff
@@ -60,11 +48,18 @@ sub get_by_id
     return $result[0];
 }
 
-sub get_by_id_locked
-{
+sub get_by_id_locked {
     my ($self, $id) = @_;
-    my @result = values %{$self->_get_by_keys_append_sql($self->_id_column, 'FOR UPDATE', $id)};
```
mwiencek: This was the only thing `_get_by_keys_append_sql` was used for. But the query it's modifying is too simple to be worth factoring this out, I think.

lib/MusicBrainz/Server/Data/Entity.pm
```diff
@@ -60,11 +48,18 @@ sub get_by_id
     return $result[0];
 }
 
-sub get_by_id_locked
-{
+sub get_by_id_locked {
```
gentlecat: What does `locked` mean in this case?

gentlecat:
LGTM.

lib/MusicBrainz/Server/Data/Entity.pm
uklauer: The function not only gets the IDs, but also puts a row lock on them via the `FOR UPDATE` clause. See http://www.postgresql.org/docs/9.1/static/explicit-locking.html#LOCKING-ROWS

gentlecat: Huh, that's a nice feature.

PR #209, opened 2016-01-10
MBS-6166 once more: Don’t restore deleted users that still have a Catalyst session

uklauer:
Deleted users that still had a Catalyst session, possibly on another computer or in a different browser, could be active on the site until that session expired. (They could not log in again, even if they had set a new password during that time.) When trying to restore a user from a session, check for the deleted flag and deny restoration when indicated.

### DIFF
--- a/lib/MusicBrainz/Server/Authentication/Store.pm
+++ b/lib/MusicBrainz/Server/Authentication/Store.pm
@@ -43,7 +43,9 @@ sub for_session
 sub from_session
 {
     my ($self, $c, $frozen) = @_;
-    return _rebless_editor($c->model('Editor')->get_by_id($frozen->{id}));
+    my $editor = $c->model('Editor')->get_by_id($frozen->{id});
+    return unless ($editor && !$editor->deleted);
+    return _rebless_editor($editor);
 }
 
 sub _rebless_editor {
--- a/lib/MusicBrainz/Server/Controller/Root.pm
+++ b/lib/MusicBrainz/Server/Controller/Root.pm
@@ -199,7 +199,7 @@ sub begin : Private
     $c->stats->enable(1) if DBDefs->DEVELOPMENT_SERVER;
 
     # Can we automatically login?
-    if (!$c->user_exists) {
+    if (!$c->user) {
         $c->forward('/user/cookie_login');
     }
 
--- a/t/lib/t/Mechanize.pm
+++ b/t/lib/t/Mechanize.pm
@@ -8,14 +8,16 @@ has mech => (
     is => 'ro',
     required => 1,
     lazy => 1,
-    default => sub {
-        MusicBrainz::WWW::Mechanize->new( catalyst_app => 'MusicBrainz::Server', quiet => 1 );
-    },
+    builder => 'make_mech',
     clearer => '_clear_mech'
 );
 
 before run_test => sub {
     shift->_clear_mech;
 };
+
+sub make_mech {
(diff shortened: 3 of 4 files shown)


lib/MusicBrainz/Server/Authentication/Store.pm
```diff
@@ -43,7 +43,9 @@ sub for_session
 sub from_session
 {
     my ($self, $c, $frozen) = @_;
-    return _rebless_editor($c->model('Editor')->get_by_id($frozen->{id}));
+    my $editor = $c->model('Editor')->get_by_id($frozen->{id});
+    return unless ($editor && !$editor->deleted);
+    return _rebless_editor($editor);
```
uklauer: This is the main part of this PR, fixing the issue. `from_session` is called with the session data in `$frozen` when a browser has sent a cookie with a valid session ID, and the user returned here is immediately accepted. This avoids all the checks that were added in https://bitbucket.org/metabrainz/musicbrainz-server/pull-requests/448/mbs-6182-delete-editors-with-a-dedicated

lib/MusicBrainz/Server/Controller/Root.pm
```diff
@@ -199,7 +199,7 @@ sub begin : Private
     $c->stats->enable(1) if DBDefs->DEVELOPMENT_SERVER;
 
     # Can we automatically login?
-    if (!$c->user_exists) {
+    if (!$c->user) {
```
uklauer: The `user_exists` method does not actually try to restore the user; if there is a valid session, it will return true, even if the owner of the session has been deleted in the meantime. Checking early for this condition results in a clearer error message than when the issue is detected later inside the Redis access try/catch block.

t/lib/t/Mechanize.pm
```diff
@@ -8,9 +8,7 @@ has mech => (
     is => 'ro',
     required => 1,
     lazy => 1,
-    default => sub {
-        MusicBrainz::WWW::Mechanize->new( catalyst_app => 'MusicBrainz::Server', quiet => 1 );
-    },
+    builder => 'make_mech',
```
uklauer: So that a test can run `make_mech` again to get another Mechanize instance (= browser session).

t/lib/t/MusicBrainz/Server/Controller/Admin/DeleteEditor.pm
```diff
@@ -34,6 +43,11 @@ test 'Delete account as a regular user' => sub {
     html_ok($mech->content);
     $mech->content_contains('You need to be logged in to view this page');
 
+    $second_session->get('/');
+    is($second_session->status, 500, 'restoring deleted user fails');
```
uklauer: If this situation occurs (someone is active from two browsers more or less at the same time, deletes their account from one of them, then continues browsing the website using the other browser before the session has expired), Catalyst throws an exception, resulting in an ISE page. I think this will be so rare that it’s not worth to try and catch it and deal with it separately.

t/lib/t/MusicBrainz/Server/Controller/Admin/DeleteEditor.pm
```diff
@@ -34,6 +43,11 @@ test 'Delete account as a regular user' => sub {
...
+
+    $second_session->get_ok('/');
+    $second_session->content_contains('Log In', 'no longer logged in in second session');
```
uklauer: If indeed someone does it, the session will have been invalidated, so just reloading the page works.

mwiencek:
:shipit:

PR #210, opened 2016-01-10
MBS-8688: Use consistent order for entity types on tag pages

uklauer:
For user tag pages, the entity types were sorted using the `<=>` operator. Since this operator treats its operands as numbers, the strings `"release"`, `"recording"` etc. were all the same to it, having the numerical value 0. As Perl’s sort is not guaranteed to be stable, the resulting order was unpredictable.

Use `cmp` instead, the operator for lexicographic comparison of strings.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User.pm
+++ b/lib/MusicBrainz/Server/Controller/User.pm
@@ -440,7 +440,7 @@ sub tag : Chained('load') PathPart('tag') Args(1)
     my $tag = $c->model('Tag')->get_by_name($tag_name);
     my %tags = ();
     my $tag_in_use = 0;
-    my @entities_with_tags = sort { $a <=> $b } entities_with('tags');
+    my @entities_with_tags = sort { $a cmp $b } entities_with('tags');
 
     # Determine whether this tag exists in the database
     if ($tag) {


lib/MusicBrainz/Server/Controller/User.pm
```diff
@@ -440,7 +440,7 @@ sub tag : Chained('load') PathPart('tag') Args(1)
     my $tag = $c->model('Tag')->get_by_name($tag_name);
     my %tags = ();
     my $tag_in_use = 0;
-    my @entities_with_tags = sort { $a <=> $b } entities_with('tags');
+    my @entities_with_tags = sort { $a cmp $b } entities_with('tags');
```
uklauer: I had considered using a collator for sorting, but the full strings (“Releases” etc.) and their translations are only available in the templates. At least it’s a consistent order now.

jesus2099: Hi Ulrich,
I don’t know what `$a` and `$b` are in `$a cmp $b` but is this sorting the same order, whatever the language is ? :)

uklauer: Yes, the order will be independent of the interface language. It’s the result of sorting the internal (English) type names, so it will be `area`, `artist`, `instrument`, `label`, `place`, `recording`, `release`, `release_group`, `series`, `work` in that order.

mwiencek:
:shipit:

PR #211, opened 2016-01-11
Two string improvements

uklauer:
I fixed a spelling mistake in one string (it’s _proceed_, but _precede_), then used the corrected string in another place where the same thing was explained in less clear words.


lib/MusicBrainz/Server/Form/Field/DatePeriod.pm
```diff
@@ -37,7 +37,7 @@ after 'validate' => sub {
     return 1 unless $begin->{year} && $end->{year};
 
     unless (is_date_range_valid($begin, $end)) {
-        return $self->field('end_date')->add_error(l('The end date must occur on or after the begin date'));
```
uklauer: At least it’s my opinion that this was less clear.

mwiencek:
:shipit:

PR #212, opened 2016-01-14
MBS-7241: Make cache transactional

mwiencek:
This ensures that an entity being deleted can't be re-added to the cache in a concurrent transaction before ours commits.

`pg_advisory_xact_lock` is used before cache deletion, and `pg_try_advisory_xact_lock` (which is non-blocking) before adding things to the cache. If the second case fails to obtain the lock, we don't add anything to the cache. Both locks are automatically released at the end of the transaction.

The "two 32-bit key values" signature is used for both of those lock functions. The first key is an arbitrary number I've assigned to each entity type we store in memcached. Those numbers are stored in entities.json. I've moved the cache prefixes there too, for consistency. The second key is just the id of the entity being added to or deleted from the cache.

Keeping the locking in a central place meant making it so that `EntityCache` and `CoreEntityCache` shared the same `_add_to_cache` method. They were unified by factoring out a `_create_cache_entries` method that could be extended as necessary.

Since an `_add_to_cache` method was the only important thing defined by `EntityCache`, I was able to remove that package entirely, and move `EntityCacheBase` in its place.


entities.json
```diff
@@ -1,4 +1,11 @@
 {
+    "alias_type": {
+        "cache": {
+            "id": 39,
```
mwiencek: This id is the first argument to `pg_advisory_xact_lock`/`pg_try_advisory_xact_lock`, and is unique for every cache prefix. We could've probably generated these somehow from the prefixes, but I preferred being explicit.

lib/MusicBrainz/Server/Data/AreaType.pm
```diff
@@ -6,11 +6,13 @@ use MusicBrainz::Server::Entity::AreaType;
 use MusicBrainz::Server::Data::Utils qw( load_subobjects );
 
 extends 'MusicBrainz::Server::Data::Entity';
-with 'MusicBrainz::Server::Data::Role::EntityCache' => { prefix => 'area_type' };
+with 'MusicBrainz::Server::Data::Role::EntityCache';
 with 'MusicBrainz::Server::Data::Role::SelectAll';
 with 'MusicBrainz::Server::Data::Role::OptionsTree';
 with 'MusicBrainz::Server::Data::Role::Attribute';
 
+sub _type { 'area_type' }
```
mwiencek: `_type` is the key into `%ENTITIES` and is required to look up the cache id/prefix now.

I'm not a fan of adding all these, since I think we store them in the Entity classes too. I tried adding a method that looks them up based on the package name, but it caused some test failures I couldn't figure out, so I gave in.

lib/MusicBrainz/Server/Data/Role/EntityCache.pm
```diff
@@ -1,38 +1,131 @@
...
+
+    @ids = uniq grep { defined } @ids;
+    return unless @ids;
+
+    my $cache_id = $self->_id_cache_id;
+    my $cache_prefix = $self->_id_cache_prefix . ':';
+    my @keys;
+
+    for my $id (@ids) {
+        if (looks_like_number($id)) {
```
mwiencek: Since gids are passed to this method too. I'm pretty sure it's safe that we only lock things based on the row id, since if the gid isn't cached, it has to call `get_by_id` which acquires a lock on the id.

lib/MusicBrainz/Server/Data/Role/EntityCache.pm
```diff
@@ -1,38 +1,131 @@
...
+            push @entries, [$cache_prefix . $id, $data->{$id}];
+        }
+    }
+    @entries;
+}
+
+sub _add_to_cache {
+    my ($self, $cache, %data) = @_;
+
+    $cache->set_multi($self->_create_cache_entries(\%data));
```
uklauer: Is it tested that `set_multi` can handle an empty list?

lib/MusicBrainz/Server/Data/Role/EntityCache.pm
```diff
@@ -1,38 +1,131 @@
...
+        if (looks_like_number($id)) {
+            # MBS-7241
+            $self->c->sql->do('SELECT pg_advisory_xact_lock(?, ?)', $cache_id, $id);
+        }
+        push @keys, $cache_prefix . $id;
+    }
+
+    my $cache = $self->c->cache($self->_id_cache_prefix);
+    my $method = @keys > 1 ? 'delete_multi' : 'delete';
+    $cache->$method(@keys);
```
uklauer: This seems even more likely to break than `set_multi`. :-)

lib/MusicBrainz/Server/Data/Role/EntityCache.pm
mwiencek: Seems to work locally, but I'm gonna add a guard just to be safe.

PR #215, opened 2016-01-16
 MBS-8726: Replicated updates don't invalidate cache entries on slave servers

mwiencek:
This is based on https://github.com/metabrainz/musicbrainz-server/pull/212 to avoid conflicts (and to also simplify the implementation). For just the relevant changes, see https://github.com/metabrainz/musicbrainz-server/commit/b115b490565436382c9c4b28813fc8a39b54a35e

This adds two complementary solutions:

First, suggested by Matthew Glubb in MBS-8732, is a new DBDefs setting called `ENTITY_CACHE_TTL` which defaults to 1 hour on slave servers. It ensures that if all else fails, entities will only remain cached for an hour at most.

Secondly, I added a script that can clear everything in memcached, and changed the sample replication post-process file to call it. In order to have this run after each packet, you can just do:

```bash
mv admin/replication/hooks/post-process.sample admin/replication/hooks/post-process
```


mwiencek:
I don't think we store anything else in memcached that needs a setting like this? I see most of the things using `Data::Role::SelectAll` are stored in the in-memory cache, which already has a TTL, but some things aren't. This seems broken, in fact: https://github.com/metabrainz/musicbrainz-server/blob/2d64080751192bb234ee1327ebc41df6db3530d2/lib/MusicBrainz/Server/Data/Role/SelectAll.pm#L25

uklauer:
Seems broken indeed. I actually think it has happened in the past that changes via the admin interface weren’t visible at first, and the cache had to be cleared manually first. This looks like it could have been the reason.

There is one other use of `$c->cache` without argument in `MB::S::Data::Role::MediaWikiAPI::_get_cache_and_key`. Shouldn’t matter in that case, but still a latent bug. Perhaps we should throw an exception when `MB::S::CacheManager::cache` receives an undefined prefix parameter?

lib/DBDefs/Default.pm
```diff
@@ -271,6 +271,15 @@ sub CACHE_MANAGER_OPTIONS {
...
 }
 
+# Sets the TTL for entities stored in memcached, in seconds. On slave servers,
+# this is set to 1 hour by default, to mitigate MBS-8726. On standalone
+# servers, this is set to 0 (meaning no expiration is set), because cache
+# invalidation is already handled by the server in that case.
+sub ENTITY_CACHE_TTL {
+    return 3600 if shift->REPLICATION_TYPE == RT_SLAVE;
+    return 0;
+}
```
uklauer: Do we still want to keep the separate `default_expires` option in the in-memory cache profile, just above this? I guess it’s not completely redundant because the `default_expires` applies even on standalone servers, but I see how having two settings could be confusing. Hm.

Annoying that `Cache::Memcached::Fast` doesn’t fully implement the `Cache` interface, and in particular not that option.

lib/DBDefs.pm.sample
```diff
@@ -292,6 +292,15 @@ sub WEB_SERVER                { "www.musicbrainz.example.com" }
...
 # }
 
+# Sets the TTL for entities stored in memcached, in seconds. On slave servers,
+# this is set to 1 hour by default, to mitigate MBS-8726. On standalone
+# servers, this is set to 0 (meaning no expiration is set), because cache
+# invalidation is already handled by the server in that case.
+# sub ENTITY_CACHE_TTL {
+#     return 3600 if shift->REPLICATION_TYPE == RT_SLAVE;
+#     return 0;
+# }
```
uklauer: I think we should have a simpler example method here in the sample file, e.g. `# sub ENTITY_CACHE_TTL { 0 }`. The more complex slave/non-slave logic can stay in `DBDefs::Default`.

script/clear_external_cache.pl
```diff
@@ -0,0 +1,15 @@
...
+use warnings;
+
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+use MusicBrainz::Server::Context;
+
+# This script can be used to clear everything in memcached. It assumes that
+# the CACHE_MANAGER_OPTIONS DBDefs setting resembles the default one (so it
+# has an "external" profile pointing to Cache::Memcached::Fast, and
+# default_profile is set to "external").
```
uklauer: Should we have a warning here that the _complete_ cache will be cleared, not just our namespace?

script/clear_external_cache.pl
```diff
@@ -0,0 +1,15 @@
...
+
+my $c = MusicBrainz::Server::Context->create_script_context;
+$c->cache->flush_all;
```
uklauer: If the argument to `$c->cache` becomes mandatory, as I suggested above, this will need changing, too.

UmkaDK: Doesn't this delete all the keys from memcached?? What if memcached is shared between multiple applications?? Maybe, a better approach would be to only delete the keys in MusicBrainz' namespace, as defined in `DBDefs -> MEMCACHED_NAMESPACE`.

uklauer: [Yes, it does.](https://github.com/metabrainz/musicbrainz-server/pull/215#discussion_r49938057) The problem is that memcached doesn’t support this ([docs](https://code.google.com/p/memcached/wiki/NewProgramming#Invalidating_by_Tag)). There is an undocumented [`stats cachedump`](http://blog.elijaa.org/index.php?post/2010/12/24/Understanding-Memcached-stats-cachedump-command) that will list items, but only to a hardcoded limit, so it‘s unreliable.

If we want to selectively remove items from cache, I think a better approach would be to retrieve those entities that have a recent `last_updated` timestamp from Postgres, then remove them from memcached.

mwiencek:
@uklauer Maybe it's time we got rid of the in-memory cache, since we don't even have a way to invalidate it across servers. But I agree it would be a good idea to have `$c->cache` throw an error for an undef parameter—I don't see any reason to allow something error-prone like that.

lib/DBDefs/Default.pm
mwiencek: Yeah, I had checked before if `Cache::Memcached::Fast` supported `default_expires`. Certainly confusing. Like I mentioned below, I think it makes sense to just remove the in-memory cache in a later PR. I'm afraid that setting an expiration of "0" on things in the in-memory cache would cause some items to never be invalidated in some now-known cases.

script/clear_external_cache.pl
mwiencek: Okay, there's a new script in https://github.com/mwiencek/musicbrainz-server/commit/1cc98c5c72ae31f54701802501caba66c6a5cb6f that deletes things updated in the last 2 hours. I removed the old script that clears the entire cache.

mwiencek:
https://github.com/mwiencek/musicbrainz-server/commit/1cc98c5c72ae31f54701802501caba66c6a5cb6f#diff-40c11f28200f921e0df9eaefde5c8cbaR46 and below was problematic for the AliasType models, so I actually ended up generating them from a template in https://github.com/mwiencek/musicbrainz-server/commit/fe2c06326a1664b332c7dc4b0c5ceb058f48526d

I tried generating them using `eval` instead (so, as multiple packages defined in the existing file), but `Module::Pluggable::Object` wasn't finding them. I do think we discussed using templates for our sql before.

lib/DBDefs/Default.pm
uklauer: @zas proposed to drop memcached altogether and store everything in Redis. (This would require two Redis instances, one with key eviction for the cache and one without it for the sessions.) (The advantage being that Redis is better at having multiple cache servers, and of course we would have one less major dependency.) I successfully made my sandbox use Redis for caching, without code changes, as a (partial) proof of concept. `Cache::Redis` does support an expire time parameter, though annoyingly it is called `default_expires_in`.

uklauer:
Can’t say that I like the TT-generated code files particularly, but it should work for now. However, now the master file says “auto-generated, do not edit”, too; I think we should have a TT comment before it saying that actually _this_ is where changes are to be made.

lib/DBDefs/Default.pm
mwiencek: Cool, that's something we should try, I think. I do like memcached but I'm not sure it has a lot of technical advantages for us over redis. With redis we'd at least have a way to delete keys by prefix using SCAN, AFAICT, and maybe be able to set up local, read-only replicated instances on each app server if the latency isn't too bad.

zas: Redis is easier to scale, as it comes with master/slave options. This is the main reason i asked if we can move to it, performance-wise i don't think it will make much difference anyway. Redis has much more options and since it is already used, as you said, it will actually reduce number of dependencies.

mwiencek:
Comment added.

PR #216, opened 2016-01-20
Various /ws/2 fixes, mostly JSON related

mwiencek:
It's better to check the commits separately since they're mostly unrelated.


uklauer:
Other than that I think it should be `relationships_loaded` or `relationships_are_loaded`, not `are_relationships_loaded`, in commit 84b374604cbe449e3c47235f31be035bb6ca0de8, this looks OK to me.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Role/Relationships.pm
```diff
@@ -11,11 +11,16 @@ around serialize => sub
...
+        defined $inc &&
+        $inc->has_rels &&
+        $entity->are_relationships_loaded;
 
     my @rels = map { serialize_entity($_, $inc, $opts) }
-        sort_by { $_->target_key . $_->link->type->name }
-            @{ $entity->relationships };
+        sort_by { $_->link->type->name }
+        sort_by { $_->link_order // 0 }
+        sort_by { $_->target_key }
```
gentlecat: I'm a bit confused about how sorting works there. Is it sorting first by `target_key`, then by `link_order`, and `link->type->name`?

lib/MusicBrainz/Server/Test.pm
```diff
@@ -404,8 +405,7 @@ sub _build_ws_test_json {
             $mech->get_ok($end_point . $url, 'fetching');
             is_valid_json($mech->content, "validating (is_valid_json)");
 
-            is_json($mech->content, $expected);
```
gentlecat: A weird name for comparison function.

gentlecat:
PUID is [this fingerprinting system](https://beta.musicbrainz.org/doc/Fingerprinting#PUID), right?

gentlecat:
LGTM.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Role/Relationships.pm
uklauer: Oh right, that seems broken – it‘s equivalent to only sorting by `$_->link->type->name`, just less efficient.

lib/MusicBrainz/Server/Test.pm
uklauer: I assume it was though of as “the JSON variant of `is`”, not the English statement “is JSON”. Well, it’s gone now.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Role/Relationships.pm
mwiencek: @gentlecat Yeah, they're just functions without the parentheses so it reads RTL. `sort_by({ $_ }, sort_by({ $_ },  sort_by({ $_ }, ...)))`.

@uklauer Can you explain? My understanding is that relationships with equal link type names will then be sorted by link_order (if the sort is stable).

mwiencek:
@gentlecat Right, it's our old proprietary fingerprinting system. Been defunct for over two years now.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Role/Relationships.pm
uklauer: Yes, but the sort isn’t stable. At least it’s not guaranteed to be.

The way it was done previously (i.e. concatenating the keys) should be safe, in particular when using a separator that sorts less than all characters that can appear in the parts (TAB should work). E.g.

    sort_by { join("\t", $_->target_key, (sprintf "%09d", $_->link_order // 0), $_->link->type->name) ) }

mwiencek: Got it, I was wrong to assume that.

uklauer: Alternatively, it would be possible to request stable sort via `use sort 'stable';` (http://perldoc.perl.org/sort.html), but I think concatenating keys is better.

mwiencek:
Re `are_relationships_loaded` I was trying to be consistent with our is_ and has_ variables. But I won't lie, it sounded weird, maybe because it was plural...

uklauer:
`has_loaded_relationships` then, perhaps? The problem to me is that I read it as “some_object has_confirmed_email_address” and “some_object is_limited” and that’s fine, but “some_object are_relationships_loaded” isn’t a sentence.

mwiencek:
`has_loaded_relationships` worksforme too.

PR #217, opened 2016-01-21
Remove CalculateRelatedTags.sh from daily.sh

mwiencek:
Judging by the last_updated timestamps, this script has been broken since 2014-11-17. We can enable it again once it works, and once we have an actual use for it, perhaps when we do something with genres.

I think this is partially responsible for the daily script always emailing us, since it always results in a division by zero error.

### DIFF
--- a/admin/CalculateRelatedTags.sh
+++ b/admin/CalculateRelatedTags.sh
@@ -1,4 +1,8 @@
 #!/bin/sh
 
+# This script is currently broken. Presumably it needs updating for all
+# the entities added in MBS-7551. Once it's fixed, it can be re-added to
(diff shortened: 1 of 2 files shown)


mwiencek:
This script is what populates the tag_relation table, which *I think* tries to find tags frequently used together. But that table isn't actually used by anything.

gentlecat:
Huh, seems ok to remove it then. Maybe it's worth adding a note in the script as well, to prevent confusion?

mwiencek:
Added a short comment to admin/CalculateRelatedTags.sh.

gentlecat:
:shipit:

PR #218, opened 2016-01-25
Replace Algorithm::Merge with Text::Diff3 (MBS-7475, MBS-8752, ...)

mwiencek:
Looks like we finally found a reproducible case where Algorithm::Merge is the culprit for buggy behavior (see rationale in MBS-7475). Using Text::Diff3 instead causes it to not drop tracks in the test I added, and the rest of our Edit:: tests still pass. I haven't done extensive testing outside of that—I'm not sure how many of our Edit:: tests go through CheckForConflicts, actually.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/CheckForConflicts.pm
+++ b/lib/MusicBrainz/Server/Edit/CheckForConflicts.pm
@@ -3,7 +3,7 @@ use 5.10.0;
 use Moose::Role;
 use namespace::autoclean;
 
-use Algorithm::Merge qw( merge );
+use Text::Diff3 qw( merge );
 use MusicBrainz::Server::Edit::Utils qw( merge_value );
 use Try::Tiny;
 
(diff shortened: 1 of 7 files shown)


lib/MusicBrainz/Server/Edit/CheckForConflicts.pm
```diff
@@ -123,23 +123,22 @@ sub merge_changes {
...
-                { CONFLICT => sub {
-                    die bless({
-                        property => $name,
-                        ancestor => $ancestor,
-                        current  => $current,
-                        new      => $new
-                    }, 'Conflict')
-                } }
-            );
+            my $merged_json = merge(map +[ $_->[0] ], ($new, $ancestor, $current));
```
mwiencek: The main things that changed are the parameter order and conflict handling.

lib/MusicBrainz/Server/Edit/Medium/Edit.pm
```diff
@@ -301,15 +301,6 @@ sub build_display_data
             $data->{changed_mbids} = 1;
         }
 
-        $data->{artist_credit_changes} = [
```
mwiencek: This and `hash_artist_credit_without_join_phrases` weren't used outside of tests, so I moved what's here into the test file, but without the join phrase exclusion logic, because that isn't required by what the tests are checking (they basically just check the length of `artist_credit_changes` and the type of changes).

lib/MusicBrainz/Server/Edit/Medium/Edit.pm
```diff
@@ -381,17 +385,17 @@ sub accept {
...
+                [ artist_credit => \@merged_artist_credits, $hash_artist_credit ],
                 [ is_data_track => \@merged_is_data_tracks ]
             ) {
                 my ($property, $container, $key_generation) = @$merge;
-                push @$container, merge(
-                    track_column($property, $self->data->{old}{tracklist}),
-                    track_column($property, $current_tracklist),
-                    track_column($property, $data_new_tracklist),
-                    { CONFLICT => sub { die } },
-                    $key_generation // ()
```
mwiencek: `Text::Diff3::merge` doesn't appear to have key support, which is why we're now feeding it through `track_column` to call it in there, and mapping the artist credit keys to the actual AC instances above in `%hashed_artist_credits`.

gentlecat:
Why is that related to MBS-8736?

mwiencek:
Heh. I must've copied that from the wrong tab. Meant to say MBS-8752

gentlecat:
:)

gentlecat:
LGTM, as far as I managed to understand it.

mwiencek:
Packaged libtext-diff3-perl for trusty and updated chef-cookboks https://github.com/metabrainz/chef-cookbooks/commit/dedf182b0d1bd3fbede30e7c170a74f86813d1d6

PR #219, opened 2016-01-28
Remove uri_for* from JS templates

mwiencek:
I had an idea that we could eventually use `x-accel-redirect` in production for our JS template renderer. Which has the potential to dramatically improve performance. But that can't work if we have these special markers in the response that need to be processed by the perl. This PR removes those.

The benefits of `uri_for` and `uri_for_action` are mainly that (1) the URLs are path-checked and type-checked, in a way, at run-time (but, that also causes problems during deployment), (2) you can change an action's URL without changing the internal path (not that we've done that often), and (3) they can give a decent hint as to what controller the action is in.

The downsides are that the replacement magic prevents us from using `x-accel-redirect`, and that they can't be used on the client anyway, so any components that are shared by both the client and server would have to use hardcoded links regardless.

Pure-JS implementations of `uri_for*` look to be out of the question, since their logic is super complex and they need access to the huge internal map of actions.

### DIFF
--- a/lib/MusicBrainz/Server/View/Node.pm
+++ b/lib/MusicBrainz/Server/View/Node.pm
@@ -9,23 +9,7 @@ use HTML::Entities qw( decode_entities );
 use HTTP::Request;
 use JSON -convert_blessed_universally;
 use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
-use MusicBrainz::Server::Validation qw( encode_entities );
 use URI;
-
-# INFORMATION SEPARATOR ONE
-my $INFO_SEP = "\x{001F}";
-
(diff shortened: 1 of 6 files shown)


gentlecat:
`x-accel-redirect` to delegate template rendering to that node server?

mwiencek:
Yeah. Well, we wouldn't be delegating anything that it isn't already doing (it renders the templates either way), except probably outputting some additional headers. The `x-accel-redirect` thing just determines who waits for the request: perl or nginx. We don't want our perl processes sitting idle waiting for some synchronous request to finish. Nginx is much better suited for that.

gentlecat:
Right, that's how I understood it. LGTM.

PR #220, opened 2016-01-28
MBS-8624: Components for server-side rendering are not equivalent to TT macros

mwiencek:
This patch took a super long time, like a couple months, which I feel bad about, but it was mostly because I had to patch React to do stuff it can't do (in order to make sure the markup matches what TT produces), and had to write a bunch of tests (otherwise we'd have no idea if it does what I say).

The main things I had to change about React were making it so it doesn't wrap text in span elements (which I submitted an upstream PR for), and making it so you can return multiple elements from a component (wrapped in a special `<frag></frag>` element, short for [fragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment)), which is needed in order to generate markup like `<a /> <span class="comment" />` from a single component. Previously we were faking that by having functions that returned an array of elements, which we could embed like `<div>{someFunctionThatReturnsAnArrayOfElements()}</div>`, but that was bad because (1) it's not in a component, so it has to be computed on every render, (2) every element in the array needed to be assigned a `key` prop, as required by React, and (3) it was ugly and out of place with how all other components were handled.

The fragment thing is actually a much-requested feature that they want to implement, so we won't have to maintain a fork forever, but probably for the next few versions at least.

The one thing I changed on the TT side, in order to make sure the markup was identical, was to have the entity macros output relative instead of absolute URIs. I'm not sure if there was an actual technical reason we did that, other than that's what Catalyst forced on us.


root/static/scripts/common/i18n.js
```diff
@@ -137,46 +142,7 @@ exports.expandToArray = function (string, args) {
...
-    var output = '';
-
-    if (!items.length) {
-        return output;
-    }
-
-    output = l('{last_list_item}', {last_list_item: items.pop()});
-    items.reverse();
-
-    _.each(items.reverse(), function (item) {
```
mwiencek: Incorrectly reversed the list twice, which is why that one relationship editor test changed. I added some tests for commaOnlyList.

.babelrc
```diff
@@ -6,6 +6,7 @@
   ],
   "presets": ["es2015", "stage-2"],
   "plugins": [
+    "jsx-control-statements",
```
mwiencek: This is a cool project that lets us do stuff like `<If condition={foo}><div /></If>` without any overhead (it's converted into a ternary expression). I think it improves readability a lot.

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -264,6 +264,7 @@ around TO_JSON => sub {
 
     return {
         %{$self->$orig},
+        email => $self->email,
```
mwiencek: Something to be wary about: we'll eventually want a sanitized version of these TO_JSON methods so the output can be embedded into a page or outputted in /ws/js/edit.

root/static/scripts/common/utility/reactTextContent.js
```diff
@@ -0,0 +1,25 @@
+// This file is part of MusicBrainz, the open internet music database.
+// Copyright (C) 2015 MetaBrainz Foundation
+// Licensed under the GPL version 2, or (at your option) any later version:
+// http://www.gnu.org/licenses/gpl-2.0.txt
+
+const nonEmpty = require('./nonEmpty');
+
+function reactTextContent(reactElement, previous = '') {
```
mwiencek: This can take something like `<span>foo<span>bar</span></span>` and return `foobar`. It's used by the `EntityLink` component to determine if there's a name variation in the content. Right now the TT hacks around this by stripping the HTML.

root/static/scripts/tests/react-macros.js
```diff
@@ -0,0 +1,155 @@
+// This file is part of MusicBrainz, the open internet music database.
```
mwiencek: This evals the React snippets from the Perl tests and compares them to the corresponding TT output. Comparing two HTML strings is far, far easier in JavaScript than in it is Perl (the DOM has a built-in HTML parser).

.babelrc
gentlecat: Yeah. I guess the only way to do this before was to create an element within JS control statements and inserting them into final element, or just splitting into multiple `return`s.

Seems like a good addition. Though tag with an uppercase letter looks a bit unusual. :P

lib/MusicBrainz/Server/Entity/Editor.pm
gentlecat: Which means we'll need to exclude private info, right?

.babelrc
mwiencek: Right now we're mostly doing it like `<div>{!!(condition) && <div />}</div>` (abusing short-circuit evaluation, since React will never output `false` in the HTML) which is kinda gross and annoying to indent. We'll have to convert those later.

I guess it made more sense to use an uppercase later since it's closer in spirit to a custom component than a native element. Except it's not a component at all, so dunno. :)

lib/MusicBrainz/Server/Entity/Editor.pm
mwiencek: Yeah, like emails, for instance ^

If we use these `TO_JSON` representations as our standard format and base /ws/3 on them, we'll need to split them into public/private formats.

root/static/scripts/common/components/DescriptiveLink.js
```diff
@@ -0,0 +1,42 @@
+// This file is part of MusicBrainz, the open internet music database.
+// Copyright (C) 2015—2016 MetaBrainz Foundation
+// Licensed under the GPL version 2, or (at your option) any later version:
+// http://www.gnu.org/licenses/gpl-2.0.txt
+
+const React = require('react');
```
gentlecat: How is this constant used there? I don't see any place that calls it.

mwiencek: So e.g. with something like `<EntityLink entity={entity} {...props} />` below, the JSX gets converted into `React.createElement(EntityLink, {entity: entity, ...})` during compilation. You gotta import React everywhere you use JSX, basically.

root/static/scripts/tests/react-macros.js
```diff
@@ -0,0 +1,155 @@
...
+
+'use strict';
+
+require('babel-core/register');
+
+const document = require('jsdom').jsdom();
+const React = require('react');
+const ReactDOMServer  = require('react-dom/server');
+
+const DescriptiveLink = require('../common/components/DescriptiveLink');
```
mwiencek: These imports are there 'cause on line 117 we `eval` code that the perl sent to us in t/lib/t/TemplateMacros.pm, which references these components.

root/static/scripts/common/utility/reactTextContent.js
gentlecat: Where is that name variation checking used?

root/static/scripts/common/utility/isolateText.js
```diff
@@ -0,0 +1,10 @@
+const React = require('react');
+
+function isolateText(content) {
+  if (content) {
+    return <bdi>{content}</bdi>;
```
gentlecat: Huh, first time I see this tag. :)

root/static/scripts/common/utility/reactTextContent.js
gentlecat: Ah, https://github.com/metabrainz/musicbrainz-server/pull/220/files#diff-26d6e5dfdd7c2a7549283d874f914adcR138.

gentlecat:
LGTM.

root/static/scripts/common/utility/reactTextContent.js
mwiencek: Yeah, as the link shows it's only really used for artists (artist credits) and recordings (which are linked using track names on release pages) right now.

Usually the name variation is just gonna be a string in those cases. But I think we pass in HTML in the edit diff display stuff. Which hasn't been converted to React yet.

mwiencek:
*crosses fingers*

PR #221, opened 2016-02-01
MBS-8770: Load banner message correctly

uklauer:
The banner message (if any) was assigned to a local variable that was shadowing the variable that it should have been assigned to. Consequently, when the inner variable went out of scope, the banner message text was gone and could no longer be shown.


uklauer:
Targetting the production branch because it is intended to set a banner pointing to the messaging blog post.

Warning: @reosarevok set “test” or something as the banner message (not sure whether on the main server or beta). This will become visible when this fix is deployed.

mwiencek:
Thanks :+1:

PR #223, opened 2016-02-04
Update team link to the MeB Team page

reosarevok:
Much less uses than the contact us page so probably not worthy of a constant? 

As before, there's an entry in admin/nginx/mbserver-rewrites.conf that I decided not to touch. Let me know if I should.


mwiencek:
Feel free to update the mbserver-rewrites.conf file too (I already did the same for the contact page).

uklauer:
Why have a schema-neutral URL in the JS template and an https URL in the TT one? Is metabrainz.org even still accessible under http?

reosarevok:
Just because it was /doc before so I didn't think about changing that :) It would probably be reasonable to have both as https, yes - will change.

reosarevok:
Oh, actually, it was because the one over it was done like that. But that's an old link from when we used mediawiki, so I'll update that too (the redirect works, but why redirect I guess).

PR #224, opened 2016-02-04
MBS-8779: Update the footer text to mention MeB

reosarevok:


root/layout.tt
```diff
@@ -130,12 +130,11 @@
...
 
             <p class="right">
-                [% l('Cover Art provided by the {caa|Cover Art Archive}. Hosted by {host|Digital West}. Sponsored by: {url1|Google},
-                      {url2|OSUOSL} and {more|others...}. ',
-                    { host => 'https://www.digitalwest.com/',
-                      url1 => 'https://www.google.com/',
-                      url2 => '//osuosl.org/',
-                      more => 'https://metabrainz.org/sponsors',
+                [% l('Brought to you by {MeB|MetaBrainz Foundation} and our {spon|sponsors} and {supp|supporters}. 
+                      Cover Art provided by the {caa|Cover Art Archive}.',
```
mwiencek: This string is actually slightly different from the one in the .js file, since it has extra spaces after `{supp|supporters}.`. We'll need to put it on a single line like the .js file has so xgettext sees them the same.

mwiencek:
:shipit: after fixing the extra space issue

root/layout.tt
reosarevok: Huh, pretty sure I did it that way because it was different originally too! But that makes sense so changing :)

mwiencek: You're right, it was wrong before you touched it. I never noticed it because string extraction from .js files was broken (oops). I fixed that earlier today in e2a67ab4badf2ce90f7edb376d7e363e40f9d5b8

PR #225, opened 2016-02-04
Remaining collections (MBS-8004)

mwiencek:
Fixes the broken code from 59a19c4. Tested adding/removing entities for all the new collection types.


lib/MusicBrainz/Server/Controller/ReleaseGroup.pm
```diff
@@ -30,6 +30,10 @@ with 'MusicBrainz::Server::Controller::Role::EditRelationships';
 with 'MusicBrainz::Server::Controller::Role::JSONLD' => {
     endpoints => {show => {copy_stash => [{from => 'releases_jsonld', to => 'releases'}]}, aliases => {copy_stash => ['aliases']}}
 };
+with 'MusicBrainz::Server::Controller::Role::Collection' => {
+    entity_name => 'rg',
```
mwiencek: This is a dumb distinction I'd like to remove later. Pretty sure release groups are the only entity type that does this (uses a different name for the entity in the stash).

t/lib/t/MusicBrainz/Server/Data/Role/Collection.pm
```diff
@@ -0,0 +1,105 @@
...
+            SELECT setval('${entity_type}_id_seq', 666, FALSE);
+EOSQL
+
+        my $model = $c->model($ENTITIES{$entity_type}{model});
+        my $entity1 = $model->insert(create_insert($entity_type, 'Test123'));
+        my $entity2 = $model->insert(create_insert($entity_type, 'Test456'));
+
+        create_collection($c, $entity_type, $entity1);
+
+        if ($entity_type eq 'release') {
```
mwiencek: I have another diff that makes the merge signatures consistent, but it's a huge patch, so for now we need to special case things.

PR #226, opened 2016-02-04
MBS-8007: Allow changing the type of an existing collection

mwiencek:
Fixes the issues described in http://tickets.musicbrainz.org/browse/MBS-8007?focusedCommentId=34417&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_34417

### DIFF
--- a/lib/MusicBrainz/Server/Data/Collection.pm
+++ b/lib/MusicBrainz/Server/Data/Collection.pm
@@ -89,6 +89,17 @@ sub contains_entity {
         $collection_id, $id) ? 1 : 0;
 }
 
+sub is_empty {
+    my ($self, $type, $collection_id) = @_;
+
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -89,6 +89,15 @@ sub contains_entity {
         $collection_id, $id) ? 1 : 0;
 }
 
+sub non_empty {
```
uklauer: Given that we already have a generic utility method `non_empty` that tests for not-undef-and-not-the-empty-string, I’d prefer to have a different name for this

mwiencek: For series we use `is_empty`, so I can do that for consistency. I guess I inverted this one based on how it was being used.

PR #228, opened 2016-02-07
MBS-8784: Add TripleJUnearthed to the otherdbs whitelist

reosarevok:
I'm pretty sure this should work, although I couldn't test it because compile_resources won't work in rika and I don't have the permissions to fix it.


mayhem:
Seems sane to me. Thanks for taking care of this.

PR #229, opened 2016-02-08
Design update

gentlecat:
The goal of this update to integrate new branding into MusicBrainz. This is the mockup that I have so far. Some things still need improvements and fixes.


gentlecat:
First version

mwiencek:
* I'd prefer getting rid of the bottom menu in the future. But there's too much crap on the page right now, so I think it's fine to just apply the theme without changing the content in this PR.
* The screenshot looks good to me, overall. I like how the search is more prominent, but the length of the bottom-left menu does make things look more "unbalanced."

gentlecat:
*With some background color (bottom margin is a bit off):*

mayhem:
I like this last one -- the search is prominent, but we're not cramming too many menu items into the toolbar with too little space. I think we can move some of the things that were removed from top level items and put them back. Not sure about the peach background for the menu bar -- we already have too many colors now.

MonkeyDo:
The search bar looks really nice in that last screenshot. Maybe an colourful button would add a little something?
You can try extending the css classes: btn btn-primary navbar-btn

I think if you're going for a colour background for the menu bar (instead of the [thick border underneath](http://metabrainz.github.io/lobes/examples/kitchen-sink.html#navbar) like on the other sites), you should try integrating the new design fully and go with the orange colour (same as the logo) for brand continuity: #eb743b

gentlecat:
We decided not to use Bootstrap. Not at this point anyway.

gentlecat:
*Alternative background color:*

mwiencek:
I like the one with a purple background color. Maybe using a darker magenta that matches the logo more, and with the text white instead of black. Adding more color is a good thing though, with just the line the page looks too plain/white to me.

gentlecat:
*More similar to what we have on other websites:*

MonkeyDo:
I really like that last one.
In my opinion it ties it well with the rest of the family.

I would maybe lower the logo so that the bottom of the MusicBrainz font is aligned with the purple line of the menu.

mayhem:
+1 to monkey's comments! I'd also love to see a full sizes logo on the home page as we have it with other projects.

gentlecat:
I tied moving logo a bit lower, but it still looks out of place:

mwiencek:
Hmm. Maybe the vertical center of the hexagon/text needs to line up with the horizontal purple line.

That looks good to me otherwise. :+1:

uklauer:
I don’t like the grey background that open menus have; perhaps make them white, possibly with a fine border to separate them from the regular page content?

The two rows of menus (user above, About etc. below the horizontal line) aren’t my favourite, either … but the line ending to the right of the logo is good, in any case.

gentlecat:
@mwiencek, can you please check code when you have some time? I think we settled on general layout and colors.

gentlecat:
And then I need to copy this stuff to new templating system.

root/static/styles/extra/banner.less
```diff
@@ -1,9 +1,10 @@
 .banner {
     margin: @form-margin 0;
+    margin-left: 12px;
+    margin-right: 12px;
     text-align: center;
-    background: #FFFFFF;
+    border: #CCC 1px solid;
```
mwiencek: `@medium-border`

I think the banners look a lot better with `background: @effectively-white-bg` too, instead of just a border

root/static/styles/layout.less
```diff
@@ -126,8 +125,183 @@ pre code {
...
+
+                    }
+                }
+
+                .separator {
+                    border-top: 1px dotted #000;
+                }
+
+            }
+
```
mwiencek: Can we remove these newlines?

mwiencek:
I still think we should link to the forums/bug tracker under "Community," since now those are just in the footer. Other than that I think it's good, just the react stuff needs fixing as you noted.

root/static/styles/layout.less
gentlecat: Yep, I'll clean it up.

root/static/styles/layout.less
```diff
@@ -126,8 +125,183 @@ pre code {
...
+    clear: both;
+
+    img.logo {
+        width: @logo-width;
+        height: @logo-height;
+        margin-top: 16px;
+    }
+
+    .right {
+        width: ~"calc(100% - @{logo-width} - 26px)";
```
mwiencek: Probably not available in all our supported browsers. :/ Looks pretty messed up in Opera 12.

gentlecat:
Oh, that's right. I forgot to add those.

root/static/styles/layout.less
gentlecat: Added alternative that you suggested.

root/static/styles/layout.less
gentlecat: Better?

root/static/styles/extra/banner.less
gentlecat: Yep, need to update icon for hiding them though.

root/static/styles/layout.less
```diff
@@ -126,8 +125,168 @@ pre code {
...
+    clear: both;
+
+    img.logo {
+        width: @logo-width;
+        height: @logo-height;
+        margin-top: 16px;
+    }
+
+    .right {
+        width: 98%; // fallback for browsers without support for calc()
```
mwiencek: 85, not 98 :) http://i.imgur.com/Isk1YFL.png

root/static/styles/layout.less
mwiencek: better :+1:

root/static/styles/layout.less
gentlecat: *Someone's copy-pasting from Stack Overfow too much.* :disappointed:

root/static/styles/extra/banner.less
gentlecat: Done.

gentlecat:
OK, new templating is updated as well.

root/layout/components/BottomMenu.js
```diff
@@ -10,7 +10,7 @@ const EditorLink = require('../../static/scripts/common/components/EditorLink');
 const {VARTIST_GID} = require('../../static/scripts/common/constants');
 const {l, lp} = require('../../static/scripts/common/i18n');
 
-function languageLink(language) {
+function languageLink(language, selected) {
```
gentlecat: Not sure if this name is good there. Maybe it should just say `showArrow`?

root/static/scripts/common/MB/Control/Menu.js
```diff
@@ -24,16 +24,16 @@ MB.Control.HeaderMenu = function () {
...
 
-    $('#header-menu > div > ul > li').bind('mouseenter.mb', function (event) {
+    $('.header ul.menu > li').bind('mouseenter.mb', function (event) {
         if (self.timeout) {
             clearTimeout(self.timeout);
-            $('#header-menu ul li ul').css('left', '-10000px');
+            $('.header ul.menu li ul').css('left', '-10000px');
         }
 
         $(this).children('ul').css('left', 'auto');
```
gentlecat: I think I'll leave direction as it is right now. Might be something worth looking into though.

root/static/styles/layout.less
```diff
@@ -126,8 +125,168 @@ pre code {
...
+                        border: 1px solid @outer-border-color;
+                        border-left-color: @inner-border-color;
+                        border-right-color: @inner-border-color;
+                        padding-left: 6px;
+                        padding-right: 22px;
+                        margin-left: -5px;
+                        -webkit-appearance: none;
+                        -moz-appearance: none;
+                        border-radius: 0;
+                        background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1…0px'><text x='0' y='15' fill='black' font-size='15'>&#x25BE;</text></svg>");
```
gentlecat: Kind of ugly just to insert one character. Don't know if there's a better way.

root/layout/components/BottomMenu.js
mwiencek: I like `selected` better. :P

gentlecat: ¯\\_(ツ)_/¯

mwiencek:
lgtm, :shipit: once we're done with final round of feedback?

gentlecat:
Alright. I pulled it in your sandbox: http://bitmap.mbsandbox.org, http://bitmap.mbsandbox.org/404.

root/static/styles/layout.less
MonkeyDo: As far as I know, that's the best solution to style a <select>.
Otherwise you can wrap it in a div and add an :after pseudo element with a unicode arrow content.

Not much cleaner :)

root/static/styles/layout.less
```diff
@@ -126,8 +125,167 @@ pre code {
...
+            }
+            .search-container {
+                @item-height: 25px;
+                @outer-border-color: fade(@musicbrainz-purple, 50%);
+                @inner-border-color: @medium-border;
+
+                float: right;
+                form {
+                    width: 302px;
+                    margin-top: 3px;
```
MonkeyDo: I would change the `margin-top` to `-2px` to align the search form vertically in the whitespace between the top of the page and the menu bar. Looks more balanced IMO.

mwiencek:
I think it's time to merge this. Not everyone will be 100% happy with it, but we can still make tweaks to it afterward.

gentlecat:
Yep. :ship:

MonkeyDo:
Looks good and fairly consistent with the other website.
Great job @gentlecat !

PR #230, opened 2016-02-08
MBS-8787: Enable CORS requests to JSON-LD

uklauer:


lib/MusicBrainz/Server/Controller/Role/JSONLD.pm
```diff
@@ -46,6 +46,7 @@ role
 
             if ($self->should_return_jsonld($c)) {
                 $c->res->content_type('application/ld+json; charset=utf-8');
+                $c->res->header('Access-Control-Allow-Origin' => '*');
```
uklauer: Should be rather obvious. I also tested it like this:
    curl http://chirlu.mbsandbox.org/artist/22a1f4d7-81ff-4216-ac06-3387da6bc118 -H "Accept: application/ld+json" -I

There is a `Role::CORS`, too, but we can’t use it here because it would add the header to all pages, not just the JSON-LD ones.

PR #232, opened 2016-02-10
MBS-8789: fix "{user} has not tagged anything" text

mwiencek:


root/components/tags.tt
```diff
@@ -4,7 +4,7 @@
 
 <h2>[% l("Tags") %]</h2>
 
-[%- IF tags.size OR c.user_exists -%]
```
mwiencek: I probably added that because I figured you'd be able to edit tags on this page, but you can't, so.

PR #233, opened 2016-02-11
MBS-8790: put parens around ended

reosarevok:
This should mean both add and edit rel edits show "(ended)" instead of the former showing "ended". This is also apparently used somewhere in series but I can't see any reason why we'd want a bare "ended" when linking to a rel anyway.

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -746,7 +746,7 @@ END -%]
         l('until {date}', { date => relationship.link.end_date.format });
     ELSIF relationship.link.ended;
         ' ';
-        l('ended');
+        l('(ended)');
     END;
     IF relationship.${extra_attributes_field};
         ' (' _ relationship.${extra_attributes_field} _ ')';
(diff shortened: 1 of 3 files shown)


mwiencek:
Apparently we have 3 different uses, `([% l('ended') %])` in root/components/relationship-editor.tt, `[% l('(ended)') %]` in root/edit/details/macros.tt and `[% l('ended') %]` in root/components/common-macros.tt (the one you're fixing here).

I think all 3 should probably be using the `bracketed` macro for this, like `bracketed(l('ended'))`. Mind changing them all to be consistent?

reosarevok:
All yours!

uklauer:
The way to fix the second commit instead of adding a third one would be to run `git rebase -i HEAD~3`, then changing `pick` into `fixup` on the too-many commit line.

The result has to be force-pushed (prefix the branch name with a `+`).

reosarevok:
I was under the impression force-pushing is bad, which is why I didn't try (I've used rebase a couple times now, even though I'm still kinda confused about it!)

mwiencek:
Force pushing is OK for temporary feature branches like this. The branch will presumably be deleted after this is merged so it hardly matters if you rewrite history on it. :)

reosarevok:
Magic!

PR #234, opened 2016-02-11
MBS-739: Make menu header click open/close dropdowns

reosarevok:
So, the <span> method is probably a huge hack or something, and I should probably be using something else, but I'm putting this up for now so that it can be checked and I can get further feedback / instructions :) 

The stuff generally seems to work, otherwise - available at http://reosarevok.mbsandbox.org/


root/layout/header/top.tt
```diff
@@ -2,16 +2,10 @@
     <ul class="menu">
     [%- IF c.user_exists -%]
         <li class="account">
-            [% link_entity(c.user) %]
+            <span class="menu-header">[%- c.user.name -%]</span>
```
reosarevok: This does lose the gravatar (but we didn't have the gravatar before so it might not be a big deal?). Do we have any code currently that shows the gravatar but does not create a link to the user? If not, should we?

root/layout/header/bottom.tt
```diff
@@ -9,9 +9,12 @@
 
 <ul class="menu">
     <li class="about">
-        <a href="[% doc_link('About') %]">[% l('About') %]</a>
+        <span class="menu-header">[% l('About Us') %]</span>
         <ul>
             <li>
+                <a href="[% doc_link('About') %]">[% l('About MusicBrainz') %]</a>
+            </li>
```
reosarevok: Unsure of the best name for this. "About MusicBrainz"? Just "MusicBrainz"? I didn't want to repeat "About Us".

root/layout/header/bottom.tt
```diff
@@ -44,7 +47,7 @@
         </ul>
     </li>
     <li class="products">
-        <a href="[% doc_link('Products') %]">[%  l('Products') %]</a>
+        <span class="menu-header">[% l('Products') %]</span>
```
reosarevok: This is pretty much "here is a list of all the products linked below" so I don't think it's worth keeping in any way in the menu.

root/layout/header/bottom.tt
```diff
@@ -79,8 +82,11 @@
         </ul>
     </li>
     <li class="search">
-        <a href="[% c.uri_for('/search') %]">[% l('Search') %]</a>
+        <span class="menu-header">[% l('Search') %]</span>
         <ul>
+            <li>
+                <a href="[% c.uri_for('/search') %]">[% l('Search') %]</a>
+            </li>
```
reosarevok: "Search Data"? This page in general is a bit weird and I'm not sure how to call it (I wonder if people know it exists...)

root/layout/header/bottom.tt
```diff
@@ -96,7 +102,7 @@
     </li>
     [%- IF c.user_exists -%]
         <li class="editing">
-            <a href="[% doc_link('How_Editing_Works') %]">[% l('Editing') %]</a>
+            <span class="menu-header">[% l('Editing') %]</span>
```
reosarevok: We dropped the link to the voting guide, and this isn't much more useful - plus the guides are already linked under Documentation in the How-To page, where it'd make more sense for someone to look for them.

root/layout/header/bottom.tt
```diff
@@ -152,6 +158,9 @@
             <li>
                 <a href="[% doc_link('Frequently_Asked_Questions') %]">[% l('FAQs') %]</a>
             </li>
+            <li>
+                <a href="[% doc_link('MusicBrainz_Documentation') %]">[% l('Documentation Index') %]</a>
+            </li>
```
reosarevok: Another case of "not sure what the best name would be" - this seemed as good as any I guess. This page might not be very useful for a newcomer, not sure, but at least it has a lot of links so that's good, I guess.

root/layout/header/top.tt
```diff
@@ -25,7 +19,7 @@
             </ul>
         </li>
         <li class="data">
-            <a href="[% c.uri_for_action('/user/profile', [ c.user.name ]) %]">[% l('My Data') %]</a>
+            <span class="menu-header">[%- l('My Data') -%]</span>
```
reosarevok: My feeling is that we should merge this into the other user menu, since they're kinda too related to be separate. That said, we still would need two lines of space for log in / create account, so we wouldn't win any real navbar space with that merge I guess.

root/static/styles/layout.less
```diff
@@ -241,17 +241,17 @@ pre code {
                 list-style-image: none;
                 &:hover {
                     background-color: @musicbrainz-orange;
-                    > a {
+                    > a, .menu-header {
```
reosarevok: I guess "a" could go, if nothing is a link anymore...

root/layout/header/top.tt
mwiencek: No code for that AFAICT, so would have to be factored out of `link_editor`. I guess it would be nice but isn't a big deal.

root/layout/header/bottom.tt
mwiencek: About MusicBrainz sounds fine to me.

root/static/styles/layout.less
mwiencek: You forgot the admin menu, that's still a link. But speaking of, /admin doesn't seem like a very useful page, it's just the menu in page-form...

But yeah, I guess we can remove the `a, ` here after.

root/static/scripts/common/MB/Control/Menu.js
```diff
@@ -41,6 +41,10 @@ MB.Control.HeaderMenu = function () {
         }, self.timeout_msecs);
     });
 
+    $('.header .menu-header').click(function () {
+        $(this).siblings('ul').toggle();
```
mwiencek: This behaves kinda weirdly in normal browsers, because if you click the menu-header it stays invisible even if you hover over it again. I was experimenting with different code to find something that works in the browser and my iphone, and got this:

```JavaScript
    $('.header .menu-header').on('click touchend', function (event) {
        event.preventDefault();
        event.stopPropagation();
        var ul = $(this).siblings('ul');
        $('.header ul.menu li ul').not(ul).css('left', '-10000px');
        ul.css('left', ul.css('left') === 'auto' ? '-10000px' : 'auto');
    });
```
^ can you see if that behaves how you expect too?

mwiencek:
I think we should add a css rule inside ul.menu like `.menu-header { cursor: default; }` so that the cursor isn't in "text-selection" mode when you hover over them.

root/static/scripts/common/MB/Control/Menu.js
reosarevok: Works fine. Honestly I'd like to get rid of hover completely, but small steps and all that :p

root/static/styles/layout.less
```diff
@@ -241,18 +241,19 @@ pre code {
                 list-style-image: none;
                 &:hover {
                     background-color: @musicbrainz-orange;
-                    > a {
+                    > .menu-header {
                         color: @text-white;
                     }
                 }
-                a {
+                a, .menu-header {
```
reosarevok: This one should still have a since it also applies to the menu entries.

mwiencek:
With the admin link gone I think we can just delete root/admin/index.tt and remove `sub index` from Controller::Admin. Can't find anywhere else that page is linked from.

root/layout/header/top.tt
mwiencek: Oh, and this should be html-escaped like `[%- c.user.name | html -%]`

root/layout/header/bottom.tt
mwiencek: Search Entities maybe?

mwiencek:
lgtm.

PR #235, opened 2016-02-12
MBS-8797: make tabs fit the current style a bit more

reosarevok:
This changes the old tabs to fit a bit more - orange + white text on selected to match the header, white background elsewhere, rounded corners, lighter borders, a purple border on hover because purple and all. It makes the font a bit bigger (or well, no longer @small-text) because it felt a bit too small to me but happy to change that if needed.

Available to check at my sandbox, e.g. http://reosarevok.mbsandbox.org/artist/ebe3e32c-99f5-46a3-8c15-bac2ab5b8e4d

### DIFF
--- a/root/static/styles/entity.less
+++ b/root/static/styles/entity.less
@@ -3,9 +3,8 @@
  */
 
 div.tabs {
-    border-bottom: solid 1px @dark-border;
+    border-bottom: solid 1px @light-border;
     padding-top: 1em;
-    font-size: @small-text;
 
     ul {
         overflow: hidden;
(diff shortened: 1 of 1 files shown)


root/static/styles/entity.less
```diff
@@ -23,23 +22,28 @@ div.tabs {
             a {
                 display: block;
                 padding: 0.1em 0.5em;
-                margin-left: 4px;
+                margin-right: 2px;
```
reosarevok: I never liked the left margin really, and I think with this design it looks even weirder. Switching to right.

root/static/styles/entity.less
```diff
@@ -23,23 +22,28 @@ div.tabs {
...
-                color: @text-black;
-            }
-
             &.sel a {
-                font-weight: bold;
-                border-bottom: solid 1px #fff;
-            }
+                &,
+                &:hover,
+                &:focus {
```
reosarevok: These three bits come from the Monkey css, honestly not sure if they're needed.

root/static/styles/entity.less
```diff
@@ -23,23 +22,28 @@ div.tabs {
...
+                    color: @text-white;
+                    font-weight: bold;
+                    background-color: @musicbrainz-orange;
+                    border: 1px solid @light-border;
+                    border-bottom-color: transparent;
+                    cursor: default;
+      }
+    }
```
mwiencek: ^ The indentation is kinda messed up here.

reosarevok:
Hmm. Seeing it in use, I'm wondering if even that isn't a bit too colorful and orange text on white (like on the RE tabs, https://beta.musicbrainz.org/release/b673c60f-731f-4e93-a29b-6360f9407648/edit ) wouldn't be better.

mwiencek:
Dunno, I think I really like the added color tbh.

PR #236, opened 2016-02-12
MBS-8798: invert paginator colours, make hover text white

reosarevok:
I feel this is more consistent. Also, white on hover makes the number more visible.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -716,11 +716,12 @@ nav .pagination {
         color: @text-black;
 
         &:hover {
-            background-color: @musicbrainz-orange;
+            background-color: @musicbrainz-purple;
+            color: @text-white;
         }
 
         &.sel {
-            background-color: @musicbrainz-purple;
+            background-color: @musicbrainz-orange;
             color: @text-white;
             font-weight: bold;
         }


mwiencek:
:shipit:

mwiencek:
Is this more consistent though? The menus are orange on hover so I'd expect other buttons to be orange on hover too. But the menus don't have any concept of "selected" I guess, just hovered.

reosarevok:
Well, it will be for sure if we merge the tabs PR as is :p As it is, my feel was that on hover = open menu = active (BTW, noticed menus are not orange when clicked on mobile, maybe I should look into that too)

mwiencek:
True, orange = active does make sense for the top-level menu headers. But you'd still have tab hover = white bg, purple border, and paginator hover = purple bg, white text. I'd expect purple bg/white text for the tabs, too, and also the menu links. Could be that I'm crazy, so if you think everything's consistent then we can merge it all.

And strange, the menus turned orange for me (when I was testing the clickable menu branch) on ios yesterday.

reosarevok:
Hmm. I thought purple BG, white text might be a bit too colorful for this case (while on the paginators it didn't feel too distracting) but we could try it if you want!

mwiencek:
Merged the tabs PR (you might be right about it being too bold/colorful there).

Using a purple border for the pagination looks pretty good to me, but we'd also need to add spacing between the buttons like the tabs have (so, stop collapsing the borders), otherwise it's hard to get a consistent outline on all four sides:

spaced paginator buttons with purple border on hover

That looks alright to me though, what do you think?

reosarevok:
Ooooh, I really like that.

gentlecat:
Looks good, I agree.

PR #237, opened 2016-02-13
MBS-739 (II): fix missing style, add 'hover' effects on Android

reosarevok:


root/layout/header/bottom.tt
```diff
@@ -13,7 +13,7 @@
 
 <ul class="menu">
     <li class="about">
-        <span class="menu-header">[% l('About Us') %]</span>
+        <span class="menu-header">[% l('About Us') %]&nbsp;&#x25BE;</span>
```
mwiencek: Can you add that to the react files too? But it'll need to look like `{'\xA0\u25BE'}` 'cause react makes it hard to insert arbitrary HTML.

root/static/styles/layout.less
```diff
@@ -245,9 +245,9 @@ pre code {
                 float: left;
                 list-style-type: none;
                 list-style-image: none;
-                &:hover {
+                &:hover, &:active, &.fake-active {
                     background-color: @musicbrainz-orange;
-                    > .menu-header {
+                    > a, .menu-header {
```
mwiencek: How come that's being added back?

mwiencek: Ah, probably so it applies to the menu items.

reosarevok: I added it back for log in and create account - it feels weird to set those as class menu header because they're not the headers for a menu.

mwiencek: Oh, right. Forgot about that issue.

mwiencek:
lgtm, thanks!

PR #238, opened 2016-02-14
MBS-8801: Fix spurious group-with-gender error

uklauer:
An artist edit can remove the artist’s type or gender by setting it to `NULL` (represented by `undef` in Perl). On the other hand, it can leave fields unchanged (represented in Perl by the absence of a hash entry). Because of a test for definedness instead of existence of a key before falling back to the current state of the artist, it could happen that an edit was rejected for resulting in a group artist with a gender when that was not actually the case.

Includes a test.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Artist/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Artist/Edit.pm
@@ -214,8 +214,10 @@ around merge_changes => sub {
 
     my $merged = $self->$orig(@args);
     my $artist = $self->current_instance;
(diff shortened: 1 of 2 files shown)


uklauer:
The test itself is untested, unfortunately, because the test database on rika is broken.

mwiencek:
Test database should work on rika now. And you should have permission to run create_test_db.sh, if you need it.

I ran the test locally and there was a failure, so needs checking in any case. :)

uklauer:
OK, fixed.

mwiencek:
Thank you :shipit:

PR #240, opened 2016-02-20
Various collection fixes for the web service

mwiencek:
Basically finishes implementing collections for all the new entity types in the WS, and fixes some other issues people have reported regarding them.

Main API changes:
 * Makes /ws/2/collection/mbid behave like other lookups. Previously it behaved like `GET /ws/2/collection` which returns all collections. This is a breaking change, but the existing behavior was not documented in the API, and generally makes no sense.
 * Adds browsing entities by `collection`
 * Adds browsing collections by `editor` (with inc=user-collections added to include private collections).
 * Adds browsing collections by entity (also with inc=user-collections)
 * Adds PUT/DELETE support for all entity types

Should resolve these tickets:
http://tickets.musicbrainz.org/browse/MBS-3125
http://tickets.musicbrainz.org/browse/MBS-5323
http://tickets.musicbrainz.org/browse/MBS-6120
http://tickets.musicbrainz.org/browse/MBS-6152
http://tickets.musicbrainz.org/browse/MBS-8459
http://tickets.musicbrainz.org/browse/MBS-8651

### DIFF
--- a/lib/MusicBrainz/Server/Authentication/User.pm
+++ b/lib/MusicBrainz/Server/Authentication/User.pm
@@ -33,6 +33,8 @@ sub new_from_editor
     return Class::MOP::Class->initialize($class)->rebless_instance($editor);
 }
 
+sub is_authorized { 1 }
+
 __PACKAGE__->meta->make_immutable;
 no Moose;
 1;
--- a/lib/MusicBrainz/Server/Controller/Role/Collection.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Collection.pm
@@ -42,19 +42,21 @@ role
 
         my @collections;
         my %containment;
+        my $entity_collections = $self->_all_collections($c);
+        my %entity_collections_map = map { $_->id => 1 } @$entity_collections;
+
         if ($c->user_exists) {
             # Make a list of collections and whether this entity is contained in them
             @collections = $c->model('Collection')->find_all_by_editor($c->user->id, 1, $entity_type);
             foreach my $collection (@collections) {
-                $containment{$collection->id} = 1
-                  if ($c->model('Collection')->contains_entity($entity_type, $collection->id, $c->stash->{$entity_name}->id));
+                $containment{$collection->id} = 1 if $entity_collections_map{$collection->id};
             }
         }
 
         $c->stash
           (collections => \@collections,
            containment => \%containment,
-           all_collections => $self->_all_collections($c),
+           all_collections => $entity_collections,
           );
     };
 
--- a/lib/MusicBrainz/Server/Controller/Role/Collection.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Collection.pm
@@ -33,10 +33,10 @@ role
 
     method _all_collections => sub {
         my ($self, $c) = @_;
-        my ($collections) = $c->model('Collection')->find_by_entity(
-            $entity_type,
-            $c->stash->{$entity_name}->id,
-        );
+        my ($collections) = $c->model('Collection')->find_by({
+            entity_id => $c->stash->{$entity_name}->id,
+            entity_type => $entity_type,
+        });
         $collections;
     };
 
@@ -51,7 +51,11 @@ role
 
         if ($c->user_exists) {
             # Make a list of collections and whether this entity is contained in them
-            ($collections) = $c->model('Collection')->find_by_editor($c->user->id, 1, $entity_type);
+            ($collections) = $c->model('Collection')->find_by({
+                editor_id => $c->user->id,
+                entity_type => $entity_type,
+                show_private => $c->user->id,
+            });
             foreach my $collection (@$collections) {
                 $containment{$collection->id} = 1 if $entity_collections_map{$collection->id};
             }
--- a/lib/MusicBrainz/Server/Controller/Role/Collection.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Collection.pm
@@ -33,28 +33,32 @@ role
 
     method _all_collections => sub {
         my ($self, $c) = @_;
-        return [ $c->model('Collection')->find_all_by_entity($entity_type, $c->stash->{$entity_name}->id) ];
+        my ($collections) = $c->model('Collection')->find_by_entity(
+            $entity_type,
+            $c->stash->{$entity_name}->id,
+        );
+        $collections;
     };
 
     # Stuff that has the side bar and thus needs to display collection information
     method _stash_collections => sub {
         my ($self, $c) = @_;
 
-        my @collections;
+        my $collections;
         my %containment;
         my $entity_collections = $self->_all_collections($c);
         my %entity_collections_map = map { $_->id => 1 } @$entity_collections;
 
         if ($c->user_exists) {
             # Make a list of collections and whether this entity is contained in them
-            @collections = $c->model('Collection')->find_all_by_editor($c->user->id, 1, $entity_type);
-            foreach my $collection (@collections) {
+            ($collections) = $c->model('Collection')->find_by_editor($c->user->id, 1, $entity_type);
+            foreach my $collection (@$collections) {
                 $containment{$collection->id} = 1 if $entity_collections_map{$collection->id};
             }
         }
 
         $c->stash
-          (collections => \@collections,
+          (collections => $collections,
            containment => \%containment,
            all_collections => $entity_collections,
           );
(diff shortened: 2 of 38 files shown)


mwiencek:
Actually, http://tickets.musicbrainz.org/browse/MBS-6152 isn't fixed until I add browsing collections by release, area, artist, etc. I won't be adding inc=collections to the other entity types.

gentlecat:
Are collections the only thing editable from the WS?

mwiencek:
You can also submit ISRCs, CDStubs, tags, ratings, and barcodes.

lib/MusicBrainz/Server/WebService/XMLSerializer.pm
```diff
@@ -228,9 +228,10 @@ sub _serialize_collection
 
     my $ser = "_serialize_${entity_type}_list";
     my $gen_list = "${entity_type}_list";
+    my $list = $opts->{$plural};
 
-    if ($toplevel) {
-        $self->$ser(\@collection, $gen, $opts->{$plural}, $inc, $stash);
+    if ($toplevel && defined($list->{items}) && @{ $list->{items} }) {
```
gentlecat: What's happening here: `@{ $list->{items} }`?

mwiencek: It's basically checking that the array is not empty, with some Perlisms: `$list->{items}` is an ARRAY reference, `@{ $list->{items} }` dereferences it into an array, and inside the `if` it's evaluated in scalar context which returns the length of the array.

lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -136,33 +136,51 @@ sub get_first_collection {
     return $self->sql->select_single_value($query, $editor_id);
 }
 
-sub find_all_by_editor {
```
mwiencek: I combined `find_all_by_editor` and `find_all_by_entity` into a single `find_by` subroutine that takes a hash ref with params, since they shared a lot of code.

Instead of `$show_private` just being 0/1, it's now the id of the editor to include private collections for. This seems safer since we won't accidentally return other private collections we don't care about. It also won't screw up the query limit/offset behavior.

I had to add an `order_by` param to `query_to_list_limited` to get this returning a consistent order. Since `query_to_list_limited` wraps the given query inside a CTE, it loses any internal ordering, so it has to be appended to the outer query.

lib/MusicBrainz/Server/Authentication/User.pm
```diff
@@ -33,6 +33,8 @@ sub new_from_editor
     return Class::MOP::Class->initialize($class)->rebless_instance($editor);
 }
 
+sub is_authorized { 1 }
```
mwiencek: This is a workaround for when I was testing things in the browser, which I guess uses this class for cookie logins. 

This is called at https://github.com/metabrainz/musicbrainz-server/pull/240/files#diff-5f430d5eefec8efae8f240e8d2b0991fR198

In normal use of the web service `$c->user` should be lib/MusicBrainz/Server/Authentication/WS/User.pm which already has `is_authorized`.

mwiencek:
Added more tests for browsing by entity, including JSON ones.

lib/MusicBrainz/Server/Data/Collection.pm
uklauer: Hm … Why does this work everywhere else, then?

lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -136,33 +136,51 @@ sub get_first_collection {
...
-sub find_all_by_entity {
-    my ($self, $type, $id) = @_;
-    my $query = "SELECT " . $self->_columns . "
-                 FROM " . $self->_table . "
-                    JOIN editor_collection_$type ce
-                        ON editor_collection.id = ce.collection
-                 WHERE ce.$type = ? ";
+    if (my $entity_type = $opts->{entity_type}) {
+        push @joins, 'JOIN editor_collection_type ct ON editor_collection.type = ct.id';
+        push @conditions, 'ct.entity_type = ?';
```
uklauer: A semijoin (`WHERE EXISTS (SELECT 1 FROM editor_collection_type ct WHERE ct.id = editor_collection.type AND ct.entity_type = ?)`) is better in theory; in any case, you could get rid of `@joins`.

lib/MusicBrainz/Server/Controller/Role/Collection.pm
```diff
@@ -33,28 +33,38 @@ role
 
     method _all_collections => sub {
         my ($self, $c) = @_;
-        return [ $c->model('Collection')->find_all_by_entity($entity_type, $c->stash->{$entity_name}->id) ];
+        my ($collections) = $c->model('Collection')->find_by({
+            entity_id => $c->stash->{$entity_name}->id,
+            entity_type => $entity_type,
+        });
+        $collections;
```
gentlecat: This returns it, right?

uklauer: Yes, exactly.

lib/MusicBrainz/Server/Controller/Role/Collection.pm
```diff
@@ -33,28 +33,38 @@ role
 
     method _all_collections => sub {
         my ($self, $c) = @_;
-        return [ $c->model('Collection')->find_all_by_entity($entity_type, $c->stash->{$entity_name}->id) ];
+        my ($collections) = $c->model('Collection')->find_by({
+            entity_id => $c->stash->{$entity_name}->id,
+            entity_type => $entity_type,
+        });
```
gentlecat: Does the new one (`find_by` with `entity_id` and `entity_type` specified) produce the same output as `find_all_by_entity`?

lib/MusicBrainz/Server/Data/Collection.pm
mwiencek: I didn't know that it did. Don't have an immediate answer to that, needs investigating. :)

lib/MusicBrainz/Server/Controller/Role/Collection.pm
gentlecat: Oh, sorry. You updated it, not paying enough attention.

mwiencek: It changed slightly 'cause it used to use `query_to_list` and now uses `query_to_list_limited`. `query_to_list` just returned a list of all results directly, but `query_to_list_limited` returns `($array_ref_of_results, $total_results)`, which is why it's doing `my ($collections) = ` instead of `my @collections) = ` (we don't care about the total in the first case).

gentlecat:
LGTM, as far as I managed to understand it.

alastair:
Part of what you've done here address another of my tickets, http://tickets.musicbrainz.org/browse/MBS-8652
 * Makes /ws/2/collection/mbid behave like other lookups. Previously it behaved like GET /ws/2/collection which returns all collections. 

Does it affect any other things that I reported? (mismatching types, invalid ids, not found)? and also the case of data being returned on a PUT or DELETE request?

lib/MusicBrainz/Server/Data/Collection.pm
mwiencek: Huh. After moving the `ORDER BY` back to the inner query, I can't reproduce whatever ordering issues I was running into. I think assumed that was the case since I couldn't find anywhere in the docs that the ordering was guaranteed to be the same. So either I missed something or postgres's implementation does guarantee that. Not a satisfying conclusion. :)

lib/MusicBrainz/Server/Data/Collection.pm
mwiencek: Fixed.

lib/MusicBrainz/Server/Data/Collection.pm
mwiencek: Reverted `order_by` changes.

mwiencek:
@alastair e496386 should fix the PUT response for unimplemented paths and adds tests for the other cases. But GET on a private collection still 401s.

mwiencek:
I guess I'll put this on beta and see what happens.

lib/MusicBrainz/Server/Data/Collection.pm
uklauer: Oh, I also had the other join in mind (necessary to get rid of `@joins`).

mwiencek: Sorry, thought you just meant here and didn't check the other one. Fixed in https://github.com/metabrainz/musicbrainz-server/commit/cacaf237e110045acbd3106e8530b25fc0702359

jesus2099:
Thanks so much @mwiencek!
It is ultra useful to me but I hope it will be useful to many other people as well. :)

PR #241, opened 2016-02-20
Vertically center the search box

uklauer:
Vertically center the search box between the top of the page and the horizontal line, as suggested by @MonkeyDo.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -169,7 +169,7 @@ pre code {
                 float: right;
                 form {
                     width: 302px;
-                    margin-top: 3px;
+                    margin-top: -1px;
                     position: relative;
                     input#headerid-query {
                         width: 150px;


root/static/styles/layout.less
```diff
@@ -169,7 +169,7 @@ pre code {
                 float: right;
                 form {
                     width: 302px;
-                    margin-top: 3px;
+                    margin-top: -1px;
```
uklauer: The suggestion was made [here](https://github.com/metabrainz/musicbrainz-server/pull/229#discussion_r52579134), where it was actually `-2px`. However, I tested (on FF and Chromium) and `-1px` looks better IMO.

gentlecat:
Yeah, I think it looks better. But now it's not aligned with links on the left, which looks a bit weird to me now.

uklauer:
Those are far away, though, and aren’t the same height anyway.

mwiencek:
Sure, let's merge this as an improvement. :shipit:

PR #242, opened 2016-02-23
[wip] Add .travis.yml

mwiencek:
Just experimenting with travis to see how fast the builds are, this will probably be broken till I work out all the kinks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,173 @@
+sudo: required
+dist: trusty
+
+env:
+  global:
+    - CATALYST_DEBUG=0
+    # no cpan questions
+    - PERL_MM_USE_DEFAULT=1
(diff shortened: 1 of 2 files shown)


mwiencek:
Not quite there yet, but things I've learned so far...

 * We need sudo to install postgres extensions, so we can't use the new "container-based" infrastructure, which doesn't allow sudo
 * Only the container-based stuff supports directory caching, so we have to install all dependencies on each build
 * Running cpanm on every build is super slow, installing from our ppa is much faster
 * Using the perl modules from our ppa means we can't target multiple perl versions using the builtin perlbrew setup

mwiencek:
We're not going to be using Travis anytime soon AFAIK and this is all obsolete anyway.

PR #243, opened 2016-02-29
MBS-6511: List of editors in the area

gentlecat:
I named the tab/page "Users" and not "Editors" to prevent confusion with "Edit" tab. Let me know if it shouldn't be like that.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Controller/Area/Users.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/Area/Users.pm
@@ -1,0 +1,40 @@
+package t::MusicBrainz::Server::Controller::Area::Users;
+use Test::Routine;
+use Test::More;
+use MusicBrainz::Server::Test qw( html_ok );
+
+with 't::Mechanize', 't::Context';
+
+test 'MBS-6511: List of editors in the area' => sub {
+    my $test = shift;
+    my $mech = $test->mech;
+
+    MusicBrainz::Server::Test->prepare_test_database($test->c);
+
+    $mech->get_ok('/area/489ce91b-6658-3307-9877-795b68554c98/users');
+    html_ok($mech->content);
+    $mech->content_contains('There are currently no users in this area');
+
+    $test->c->sql->do(q{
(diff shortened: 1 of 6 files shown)


gentlecat:
I probably should add some tests for all this, right?

mwiencek:
This looks really good, if you wanna try adding some tests (which would be nice), I'd create t/lib/t/MusicBrainz/Server/Controller/Area/Users.pm with the other files in that dir as a guide and just check that the page displays some editors...if you need to insert some test data you can do something like https://github.com/metabrainz/musicbrainz-server/blob/master/t/lib/t/MusicBrainz/Server/Controller/Series/Tags.pm#L14

mwiencek:
> I named the tab/page "Users" and not "Editors" to prevent confusion with "Edit" tab. Let me know if it shouldn't be like that.

That seems fine, I think. We're not very consistent about it, or we mostly use "editors" in the context of editing, and "users" for the site in general.

gentlecat:
Added some tests.

t/lib/t/MusicBrainz/Server/Controller/Area/Users.pm
```diff
@@ -0,0 +1,37 @@
...
+    $mech->content_contains('There are currently no users in this area');
+
+    $test->c->sql->do(q{
+        INSERT INTO editor (area, name, password, ha1)
+        VALUES (222, 'Editor 1', 'hunter2', '');
+    });
+
+    $mech->get_ok('/area/489ce91b-6658-3307-9877-795b68554c98/users');
+    html_ok($mech->content);
+    $mech->content_contains('There is currently 1 user in this area');
```
mwiencek: How about also adding a content_contains for "Editor 1" and "Editor 2" below?

mwiencek:
lgtm.

t/lib/t/MusicBrainz/Server/Controller/Area/Users.pm
gentlecat: Good idea. Done.

PR #244, opened 2016-03-01
MBS-8839: Implement Discourse SSO endpoint

mwiencek:
Tested this with their vagrant machine. It handles logging in, plus syncing the email on Discourse whenever a new one is verified on MB (according to the docs, they use the email to map external users to Discourse users, but I'm still not entirely sure how their system works).

One weird thing is, if you try to log in to Discourse and then create a new MB account through that (which then redirects back to Discourse), MB doesn't store the provided email until you verify it, and it seemed wrong to then send that email back to Discourse before it's verified, so for now it just sets it to "noreply@example.com" until it's verified...


Freso:
Would it be possible to make Discourse reject the login request until the e-mail has been verified?

gentlecat:
According to [their post](https://meta.discourse.org/t/official-single-sign-on-for-discourse/13045) it is very important to validate email addresses before sending them to Discourse. Also,

> Alternatively, if you insist on sending unvalidated emails BE SURE to set `require_activation=true`, this will force all emails to be validated by Discourse.

Not allowing to login to Discourse until user has a verified email address also sounds like a good and less hacky option to me.

mwiencek:
The require_activation=true thing means discourse will send its own activation email too, which I don't think we want, but we can have the MB sso endpoint check if the user has a valid email, that's a good idea.

root/account/registered.tt
```diff
@@ -1,26 +0,0 @@
-[% WRAPPER 'layout.tt' title=l("Account Created") %]
```
mwiencek: This was unused so I guess it can be removed. But I reused some of the email-verification text below.

root/account/registered_discourse.tt
```diff
@@ -0,0 +1,10 @@
+[% WRAPPER 'layout.tt' title=l("Account Created") %]
```
mwiencek: This page will be displayed if the user just created an account and the return uri matches `^/discourse/sso`.

root/unconfirmed_email_discourse.tt
```diff
@@ -0,0 +1,7 @@
+[% WRAPPER 'layout.tt' title=l("Unconfirmed Email Address") %]
```
mwiencek: This page will be displayed if they created an account already, and are already logged in, but don't have a confirmed email. This shouldn't happen often except to confused people.

root/account/registered_discourse.tt
gentlecat: Good idea. :+1:

gentlecat:
LGTM

uklauer:
Multiple MB accounts can share one email address. If Discourse uses the email address for the mapping, how will it react to that?

gentlecat:
BTW, how is the change of email address handled?

mwiencek:
> Multiple MB accounts can share one email address. If Discourse uses the email address for the mapping, how will it react to that?

Just tested this. When I logged in as the second user, it updated the username associated with that email on Discourse. So, posts that the user previously made were associated with whatever name they most recently logged in as.

However, I see that there are option to prevent that from occurring in the admin interface:

```
sso overrides email
Overrides local email with external site email from SSO payload on every login, and prevent local changes. (WARNING: discrepancies can occur due to normalization of local emails)

sso overrides username
Overrides local username with external site username from SSO payload on every login, and prevent local changes. (WARNING: discrepancies can occur due to differences in username length/requirements)

sso overrides name
Overrides local full name with external site full name from SSO payload on every login, and prevent local changes.

sso overrides avatar
Overrides user avatar with external site avatar from SSO payload. If enabled, disabling allow_uploaded_avatars is highly recommended
```

(sso overrides username and email were enabled when I was testing.)

But as for whether someone can have two different Discourse identities where the MB accounts share the same email, I just don't think that's possible.

mwiencek:
> BTW, how is the change of email address handled?

When someone changes their email on MB, we send a verification email to the new one which they have to click. So we continue to store and use the old email until they verify the new one. Once they do, https://github.com/metabrainz/musicbrainz-server/pull/244/files#diff-5053a22c07db9b62eceff446a9b99731R107 should kick in (that's in the verification endpoint) and run https://github.com/metabrainz/musicbrainz-server/pull/244/files#diff-27d09305b51300467c78c397b9a03ddaR66. It POSTs an SSO payload with the new email address to Discourse.

gentlecat:
Ah, OK!

mwiencek:
Just to follow up, according to https://meta.discourse.org/t/official-single-sign-on-for-discourse/13045 it's actually looking up SSO records by `external_id` (in our case the row ID of the MB user). So I imagine that's how "sso overrides email" works (since it only has the new email). But it must also be enforcing the whole email=identity thing too, so SSO record != Discourse user. This is mostly guessing...

gentlecat:
Would've been ideal to test this before making it public, but it's probably too late.

mwiencek:
Well, still time to test this PR if anyone has other ideas of things that might go wrong. :) But yeah, I have no clue how the live OAuth stuff behaves in these scenarios, probably similarly.

uklauer:
As far as I understand it, every email verification will trigger an SSO POST, right? What happens if that particular user didn’t have a Discourse account before? Will one be created without their request?

uklauer:
What do we do when someone deletes their account? (Internally, this means they will be renamed `Deleted Editor #123456`, logged out and prevented from being logged in.) Similarly, what if an admin removes someone’s email address? Shouldn’t be as common now that editing can be disabled explicitly, but might when we learn that the email address is no longer valid (and someone else took it over and is complaining about receiving unwanted notifications from MB).

mwiencek:
> As far as I understand it, every email verification will trigger an SSO POST, right? What happens if that particular user didn’t have a Discourse account before? Will one be created without their request?

Yeah, the current code creates a discourse user even if they don't attempt to log in. Maybe this is reasonable. Since discourse has private messaging, I think we should use this for messaging in MB and not write our own.

mwiencek:
> What do we do when someone deletes their account? (Internally, this means they will be renamed Deleted Editor #123456, logged out and prevented from being logged in.)

Good point, right now nothing happens there. There's a way to sign someone out via the Discourse API, so we should do that, and probably sync their username (Deleted Editor #123456). Their posts will remain but they won't be able to log in again.

> Similarly, what if an admin removes someone’s email address? Shouldn’t be as common now that editing can be disabled explicitly, but might when we learn that the email address is no longer valid (and someone else took it over and is complaining about receiving unwanted notifications from MB).

Hmm. From what I've seen, Discourse requires an email to be set at all times. The best we might be able to do is bring back the `username@example.com` hack for this case.

uklauer:
Discourse it rather spammy in its default settings; it wouldn’t do if it started mailing notifications about every new topic to every user (who signs up/verifies their email address after this goes live).

mwiencek:
This handles editors being deleted and having their emails removed now. Sets their Discourse email to `editor-$id@musicbrainz.invalid` in either case.

PR #245, opened 2016-03-02
MBS-8841: Update links to forums.

Freso:
http://tickets.musicbrainz.org/browse/MBS-8841

Disclaimer: I didn't actually test this.


mwiencek:
If you're fine with the text staying as "Forums" then lgtm :shipit:

Freso:
I figure that the most important thing to change is the URL. I'm not 100% happy with using "Forums", but "Community" also sounds misleading and "Discourse" is not really that intuitive.

PR #246, opened 2016-03-02
MBS-8842: Update Picard logo on frontpage.

Freso:
Exported from "Picard_logo_no_text.svg" in metabrainz-logos and run through `optipng -fix -o7 -strip all -clobber` and `advpng -4 -i 2500 -z`.

Disclaimer: Not actually tested…


mwiencek:
:shipit:

PR #247, opened 2016-03-02
MBS-8843: Require clicking on the header menus to open them

mwiencek:


root/static/scripts/common/MB/Control/Menu.js
```diff
@@ -32,35 +29,31 @@ MB.Control.HeaderMenu = function () {
...
-
-    $('.header ul.menu li').bind('mouseleave.mb', function (event) {
-        var ul = $(this).children('ul');
-
-        self.timeout = setTimeout(function () {
-            ul.css('left', '-10000px');
-        }, self.timeout_msecs);
-    });
-
     $('.header .menu-header').on('click touchend', function (event) {
```
uklauer: Is `touchend` still necessary/advisable? Clicking should work on touchscreens, too. (Same for line 52.)

mwiencek: I see a slight difference in behavior in iOS...with click events there's a noticeable 300 ms or so delay, and with touchend it's instant. The bootstrap menus on CB also have a delay, so it's probably advisable to remove since they're smarter than I am.

root/static/scripts/common/MB/Control/Menu.js
```diff
@@ -32,35 +29,31 @@ MB.Control.HeaderMenu = function () {
...
-            $(this).parent().siblings().removeClass('fake-active');
+        var isClosing = ul.css('left') !== '-10000px';
+        ul.css('left', isClosing ? '-10000px' : getLeft(this.parentNode));
+        $('.header .menu-header').parent().removeClass('fake-active');
+        if (!isClosing) {
             $(this).parent().toggleClass('fake-active');
         }
     });
 
+    $('body').on('click', function (event) {
```
mwiencek: This doesn't work at all on my phone (iOS). Works if I add `touchstart`, but that means I can't close a menu by clicking on the active header again (it closes the menu, then opens it again from the click event). There's definitely ways to work around it but might be more effort than it's worth...

gentlecat: I just checked in Chrome on Android and behavior is the same as on desktop.

gentlecat:
This looks good to me. Seems like it partially solves the problem of dropdowns covering bottom menu, can't do it by accident now.

root/static/scripts/common/MB/Control/Menu.js
mwiencek: Huh, well that's good at least. Thanks for checking. :)

PR #248, opened 2016-03-03
MBS-8837: rare ISE editing event time

mwiencek:
Broken in 6a8e2e6127cbdec3e23590368e78c2490c200768

This ISE can occur if the time being set is already the current value.

The source of the bug is due to the change in parameter order associated with Text::Diff3's merge subroutine.

Before the referenced commit, CheckForConflicts hashed the results of merge_time in the order (ancestor, current, new), which would be something like [['19:30', '19:30'], ['19:00', (DateTime object)], ['19:00', '19:00]], where the first value in each array ref is the key. Since the last two array refs have the same key, the second would override the first.

After 6a8e2e6, those array refs are hashed in the order new, ancestor, current, which means the DateTime object is the result.

The fix here is to revert the change in the order they're hashed (new should be last), and always return a formatted string for the current value, not a DateTime object, since the object isn't stringified to the same format that ancestor and new are in.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/CheckForConflicts.pm
+++ b/lib/MusicBrainz/Server/Edit/CheckForConflicts.pm
@@ -123,7 +123,7 @@ sub merge_changes {
                 );
 
             # Stores a mapping of the JSON serialization to the object value
-            my %hashed = map { @$_ } ($new, $ancestor, $current);
+            my %hashed = map { @$_ } ($ancestor, $current, $new);
 
             # Attempt to merge all JSON values together
             my $merged_json = merge(map +[ $_->[0] ], ($new, $ancestor, $current));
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Edit/CheckForConflicts.pm
```diff
@@ -123,7 +123,7 @@ sub merge_changes {
                 );
 
             # Stores a mapping of the JSON serialization to the object value
-            my %hashed = map { @$_ } ($new, $ancestor, $current);
```
mwiencek: So...ideally the same hash key should map to the exact same value, and these would be equivalent. But that wasn't the case here, and I have to wonder if this resulted in any other bugs.

mwiencek: And to be clear, I had changed the order to be the same as `merge` below, but there was no actual reason to.

gentlecat:
LGTM

PR #249, opened 2016-03-07
Respect area containments for MBS-6511

mwiencek:
This makes it so that if I say I'm from Chicago, I show up in the editors tab for Illinois (and USA) too.


gentlecat:
That's cool! :shipit:

PR #250, opened 2016-03-08
Fixed transparency for older browsers (IE 8, Opera 12)

jesus2099:
Unlike all the other new icons, the MB icon had no transparent background in my Opera 12.18, and it is showing harsh pixels instead of its anti-aliasing. I don’t know why.

I have eventually just opened it in [GIMP](http://gimp.org) and saved.
It fixed the display in both Opera 12.18 and Internet Explorer 8 on XP (antiquities, I know) and 7. :)


mwiencek:
Yes, this seems to work for some reason. Thank you!

PR #251, opened 2016-03-08
Add inc=user-collections to /ws/2/release (MBS-6152)

mwiencek:
`inc=user-collections` will return private (and public) collections for the authenticated user. It does not imply `inc=collections` (which returns all public collections).

Was probably wrong to not do this in the previous PR. Should help fix one or two Picard tickets.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
@@ -79,7 +79,7 @@ sub serialize
     };
 
     $body{collections} = list_of($entity, $inc, $stash, "collections")
-        if $inc && $inc->collections;
+        if $inc && ($inc->collections || $inc->user_collections);
 
     $body{"release-group"} = serialize_entity($entity->release_group, $inc, $stash)
         if $inc && $inc->release_groups;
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
```diff
@@ -79,7 +79,7 @@ sub serialize
     };
 
     $body{collections} = list_of($entity, $inc, $stash, "collections")
-        if $inc && $inc->collections;
+        if $inc && ($inc->collections || $inc->user_collections);
```
gentlecat: Do you need to do something similar to `authenticate` in WS/2 here?

mwiencek: Shouldn't need to, since this is just for serializing the data once it's loaded. Controller::WS::2::Release already runs & authenticates for us, then passes things over to here.

The XMLSerializer didn't need any changes because it just checks if any collections were loaded (and they wouldn't have been if those inc params weren't provided): https://github.com/metabrainz/musicbrainz-server/blob/c707e8e264425d2882396cb05ff7a553276ebe6b/lib/MusicBrainz/Server/WebService/XMLSerializer.pm#L444-L444

The JSON behaves differently 'cause it'll output an empty list even if none were found.

gentlecat:
LGTM :shipit:

PR #252, opened 2016-03-10
Output track artist aliases in /ws/2/release (MBS-7914)

mwiencek:
Doesn't make sense to output aliases for recording artists but not track artists. Note that we already don't output the track AC at all if it's the same as the recording AC.

This will output duplicate data if the track and recording ACs share any artists. But I'd rather do that than have some weird rule that makes parsing the data like a puzzle (e.g., outputting track artist aliases where the artist doesn't appear in the recording AC, or something).


gentlecat:
LGTM

PR #253, opened 2016-03-11
Fail edit medium edits using cached deleted recordings (MBS-8858)

mwiencek:
This is kinda masking the underlying problem, but at least edits won't get stuck anymore.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Medium/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Medium/Edit.pm
@@ -439,7 +439,10 @@ sub accept {
         }
 
         my @final_tracklist;
-        my $existing_recordings = $self->c->model('Recording')->get_by_ids(@merged_recordings);
+        my $existing_recordings = $self->c->sql->select_single_column_array(
+            'SELECT id FROM recording WHERE id = any(?)', \@merged_recordings
+        );
+        my %existing_recordings = map { $_ => 1 } @$existing_recordings;
         while (1) {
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Edit/Medium/Edit.pm
```diff
@@ -439,7 +439,10 @@ sub accept {
         }
 
         my @final_tracklist;
-        my $existing_recordings = $self->c->model('Recording')->get_by_ids(@merged_recordings);
+        my $existing_recordings = $self->c->sql->select_single_column_array(
+            'SELECT id FROM recording WHERE id = any(?)', \@merged_recordings
```
mwiencek: The fix is to bypass the cache and just directly check which ids exist in the table, since that's all it needs to know.

gentlecat:
LGTM. And the underlying problem is that recordings are not removed from the cache as well?

mwiencek:
> And the underlying problem is that recordings are not removed from the cache as well?

Yeah, in a specific case they won't be. We have a `delete_orphaned_recordings` postgres function that runs whenever a track is updated/deleted: https://github.com/metabrainz/musicbrainz-server/blob/c707e8e264425d2882396cb05ff7a553276ebe6b/admin/sql/CreateFunctions.sql#L922-L922

That'll potentially delete a recording, but obviously can't clear it from the cache from within plpgsql (well, it's certainly possible, just not easy :P).

PR #256, opened 2016-03-13
MBS-8848: Own private collections ignored on entity pages

mwiencek:
Missed an arg to load private collections for the current user, when I refactored the `find_by` stuff.


gentlecat:
LGTM

PR #257, opened 2016-03-15
Fix menus so they work without JavaScript: the header-menu ID doesn't seem to be set where it's needed.

ethus3h:


uklauer:
So … unfortunately, this doesn’t work in Chromium for me. Also, if it worked, `:hover` doesn’t sound like it reveals/hides the menus on click instead of mouseover?

General remarks:
- Whatever is being done in the end, the corresponding bit in `root/layout/index.js` needs to be changed in the same way.
- It’d be good if you could follow git commit message general practice: Short headline (“Fix menus so they work without JavaScript” should be OK), blank line, then any other explanations.

ethus3h:
Ok, thanks! I'll try again :)

uklauer:
You can update pull requests by (force-)pushing to the branch.

PR #258, opened 2016-03-15
Make menu work without JavaScript

ethus3h:
This hopefully will make it only open the menu when it is clicked. That
is by using the CSS focus attribute, instead of the hover attribute
that was previously used. I guess I wasn’t paying attention then….

### DIFF
--- a/root/layout/components/TopMenu.js
+++ b/root/layout/components/TopMenu.js
@@ -14,7 +14,7 @@ function userLink(userName, path) {
 }
 
 const AccountMenu = () => (
-  <li className="account">
+  <li className="account" tabindex="-1">
     <span className="menu-header">{$c.user.name}{'\xA0\u25BE'}</span>
     <ul>
       <li>
@@ -39,7 +39,7 @@ const DataMenu = () => {
   let userName = $c.user.name;
 
   return (
-    <li className="data">
+    <li className="data" tabindex="-1">
       <span className="menu-header">{l('My Data')}{'\xA0\u25BE'}</span>
       <ul>
         <li>
@@ -72,7 +72,7 @@ const DataMenu = () => {
 };
 
 const AdminMenu = () => (
-  <li className="admin">
+  <li className="admin" tabindex="-1">
     <span className="menu-header">{l('Admin')}{'\xA0\u25BE'}</span>
     <ul>
       {$c.user.is_location_editor &&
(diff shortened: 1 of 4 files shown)


uklauer:
Still doesn’t work for me in Chromium. (You can try it on http://chirlu.mbsandbox.org/) The menus don’t open.

ethus3h:
It looks like the .header .right ul.menu li ul rule is overriding the left:auto...

ethus3h:
This seems to work:

```
.menu li:focus ul {
	left: auto !important;
}
``` 

but !important is very bad practice. I'll try to increase the specificity of the selector

ethus3h:
Ok, here we go:

```
.header .menu > li:focus > ul {
	left: auto !important;
}
```

I'll try updating the pull request once I finish re-cloning the repository. Thank you for your help @uklauer! :)

ethus3h:
Ugh, I still had the !important in there.
Yet again, I think I've found the solution:
```
.header > .right > .bottom > .menu > li:focus > ul {
	left: auto;
}
```
#fingerscrossed

mwiencek:
This isn't working for me. :( However, it works if I add `tabindex="-1"` to the `<li>` elements. That seems to make them focusable.

ethus3h:
Oh, I guess that is needed. Thanks @mwiencek! I'll do another commit...

uklauer:
> Whatever is being done in the end, the corresponding bit in `root/layout/index.js` needs to be changed in the same way.

(from #257)

ethus3h:
Done. (I assume I shouldn't ask why that code is in two places.) Is there a corresponding JavaScript file where the `tabindex="-1"`s need to be added?

uklauer:
Should be in `root/layout/components/TopMenu.js`. The reason for the duplication is that we have long-term plans to move all templates to JS, but for a long time we’ll have some TT templates and some JS templates. And both need to have the menus.

ethus3h:
Well, ok, I did that, but the list items in TopMenu.js are different from in bottom.tt (e.g. only the former has `<li class="data"`, and only the latter has `<li class="about"`…). Is that OK anyway?

mwiencek:
Forgot to update BottomMenu.js and top.tt (and the .js files should use `tabIndex`, not `tabindex`, since the names there correspond to the DOM API). I went ahead and made those changes and merged this. Thanks!

mwiencek:
(Sigh, I intended to squash the commits, but somehow it got messed up and I didn't see.)

ethus3h:
Cool, thanks! :)

mwiencek:
I didn't see a JIRA ticket (which we use to generate release notes), so I created http://tickets.musicbrainz.org/browse/MBS-8886

If you have a JIRA account there, you can assign yourself to that ticket.

PR #259, opened 2016-03-16
MBS-8838: Add gids to all *_type* tables

gentlecat:
*http://tickets.musicbrainz.org/browse/MBS-8838*


admin/sql/CreateTables.sql
```diff
@@ -21,6 +21,7 @@ CREATE TABLE application
 
 CREATE TABLE area_type ( -- replicate
     id                  SERIAL, -- PK
+    gid                 uuid NOT NULL,
```
mwiencek: I'm not sure this would practically cause us any problems, but since `ALTER TABLE` in the upgrade script will add the column to the end, I wonder if we should do the same here for consistency.

mwiencek: Well, actually, I guess it'd break data dump imports if we didn't.

mwiencek:
Some missing pieces I see. :)

Could do with some unique indexes in admin/sql/CreateIndexes.sql and admin/sql/caa/CreateIndexes.sql like https://github.com/metabrainz/musicbrainz-server/blob/ad6f5aa090bc25fecea48b4530e40625dc5c9014/admin/sql/CreateIndexes.sql#L321-L321

upgrade.json in the root of the repository should be updated with a key for schema version "23" (then you just need to add "20160516-mbs-8838.sql" to slave and standalone sub-keys, meaning it'll run on both types of servers when someone runs upgrade.sh).

mwiencek:
After adding indexes, run `./admin/GenerateSQLScripts.pl admin/sql/` to automatically update all the DropIndexes.sql files too.

admin/sql/CreateTables.sql
uklauer: Yes, unfortunately Postgres (even in 9.5) still doesn’t allow freely ordering columns. :-/

gentlecat: Ah, unfortunate. I pushed an update.

gentlecat:
OK, indexes added and *upgrade.json* is updated.

Is WS update the only thing left there?

mwiencek:
> Is WS update the only thing left there, assuming I got the rest right?

Yeah, I think so. :100: So now you've got the annoying part for last.

mwiencek:
But I don't see any change to upgrade.json here. :P

gentlecat:
Forgot to push. :scream:

gentlecat:
*http://chatlogs.metabrainz.org/brainzbot/metabrainz/2016-03-18/?msg=3539231&page=1*

gentlecat:
I updated `XMLSerializer` and most things seem to be working. However, I couldn't figure out how to integrate `gid` columns from the following tables:
* cover_art_archive.art_type
* editor_collection_type
* series_ordering_type
* work_attribute_type_allowed_value

mwiencek:
Looks good so far. For the tables you mentioned:

* cover_art_archive.art_type - I guess we don't currently output detailed type information for cover art in the web service, just "front"/"back", which seems odd in retrospect. So let's leave that alone for now.
* editor_collection_type - Check https://github.com/metabrainz/musicbrainz-server/blob/ff74a51/lib/MusicBrainz/Server/WebService/XMLSerializer.pm#L219-L219
* series_ordering_type - That's not outputted in the web service either, but IIRC it was intentional since it's mostly an internal thing. So let's ignore that too.
* work_attribute_type_allowed_value - Annoying since it's just text data, like "E minor" in http://beta.musicbrainz.org/ws/2/work/aec2489b-051c-4892-8df5-6c601877e5f6 Let's just add a `value-id` to the attribute node. (Only where `$work_attribute_type->free_text` is false, since free-text values don't have any gid.)

lib/MusicBrainz/Server/WebService/XMLSerializer.pm
```diff
@@ -281,19 +283,27 @@ sub _serialize_release_group
...
 
-        my ($fallback) =
+        my ($fallback_name) =
             nsort_by { $fallback_type_order{$_} }
                 grep { exists $fallback_type_order{$_} }
                     map { $_->name }
                         $release_group->all_secondary_types;
+        my ($fallback_gid) =
+            nsort_by { $fallback_type_order{$_} }
+                grep { exists $fallback_type_order{$_} }
```
mwiencek: `%fallback_type_order` doesn't have gids for keys, so nothing will exist there. You can try removing the `map { $_->name }` above, use `$_->name` as the key into `%fallback_type_order`, and then do `$fallback->name` and `$fallback->gid` below.

lib/MusicBrainz/Server/WebService/XMLSerializer.pm
```diff
@@ -511,6 +511,7 @@ sub _serialize_work
             $gen->attribute({ 
                 type => $_->type->name,
                 "type-id" => $_->type->gid,
+                "value-id" => $_->value_gid,
```
mwiencek: Will that output something if `value_gid` is undef? If so you might need to do something like `$_->type->free_text ? () : ("value-id" => $_->value_gid),`.

gentlecat: Then it doesn't output `value-id`.

lib/MusicBrainz/Server/WebService/XMLSerializer.pm
gentlecat: So something like this?

```perl
my $fallback =
    nsort_by { $fallback_type_order{$_} }
        grep { exists $fallback_type_order{$_} }
            $release_group->all_secondary_types;

if ($fallback) {
    $attr{type} = $fallback->name;
    $attr{"type-id"} = $fallback->gid;
} else {
    $attr{type} = $release_group->primary_type->name;
    $attr{"type-id"} = $release_group->primary_type->gid;
}
```

Though I'm having trouble understanding when `$fallback` is actually used.

mwiencek: Yeah, except in nsort_by/grep you should do `$fallback_type_order{$_->name}` since it's keyed by name.

The fallback stuff is used when the primary type is `Album`. If there's a more specific secondary type, it'll output that instead of `Album`. I'm guessing this is because we used to only have a flat list of release types, so you would just have "Compilation" instead of "Album (primary), Compilation (secondary)", so ouputting "Compilation" is more backwards-compatible.

That's only for the `type` attribute. We still output the primary/secondary types separately below, which people should ideally be using instead.

gentlecat: Thanks. Fixed.

gentlecat:
Tests should be passing now, but I still have some problems with `WikiDoc` ones. Check please?

t/sql/initial.sql
```diff
@@ -785,95 +785,95 @@ INSERT INTO orderable_link_type VALUES (742, 2);
...
+INSERT INTO series_type VALUES (6, 'Event', 'event', NULL, 5, 'A series of events.', '64640a2a-1c31-394d-b7b0-683a32ff9aff');
+INSERT INTO series_type VALUES (7, 'Tour', 'event', 6, 0, 'A series of related concerts by an artist in different locations.', '8ff6df0e-3dce-3bdf-bd57-d386c51b0060');
+INSERT INTO series_type VALUES (8, 'Festival', 'event', 6, 1, 'A recurring festival, usually happening annually in the same location.', '44e9f6b1-34a7-3a17-a5d5-07bb100d8887');
+INSERT INTO series_type VALUES (9, 'Run', 'event', 6, 2, 'A series of performances of the same show at the same venue.', '975e9447-dfa8-3e29-82d6-07908fe36f3e');
+
+INSERT INTO work_alias_type VALUES (1, 'Work name', NULL, 0, NULL, 'a18cab3f-0ae2-3978-8f75-dd9c09702b25');
+INSERT INTO work_alias_type VALUES (2, 'Search hint', NULL, 0, NULL, '02238bc1-dfd8-39a8-bbf8-c697747291ec');
+
+INSERT INTO work_type VALUES (29, 'Musical', NULL, 2, 'Musical theatre is a form of theatrical performance that combines songs, spoken dialogue, acting, and dance.', '9ca5e067-acf7-3cd6-baa4-92bf1975bf24');
```
mwiencek: Can we keep each group sorted so it's easy to see what was actually added? (F5 in sublime, but you prob. knew that.)

mwiencek:
We added a dep. on `Text::Diff3` recently, so it looks like your tests are failing because of that. `cpanm Text::Diff3`.

t/sql/initial.sql
mwiencek: Ignore that, F5 doesn't sort numerically. Forgot I have a plugin to do that. So I'll just use it here.

mwiencek:
I added the necessary changes to the JSON web service and fixed the browse tests. Back to you. :)

t/lib/t/MusicBrainz/Server/Controller/Work/Edit.pm
```diff
@@ -104,12 +104,15 @@ test 'Editing works with attributes' => sub {
...
 DELETE FROM iswc;
 
-INSERT INTO work_attribute_type (id, name, free_text)
+INSERT INTO work_attribute_type (id, gid, name, free_text)
 VALUES
-  (1, 'Attribute', false),
-  (2, 'Free attribute', true);
+  (1, '325c079d-374e-4436-9448-da92dedef3ce', 'Attribute', false),
+  (2, '425c079d-374e-4436-9448-da92dedef3ce', 'Free attribute', true);
 INSERT INTO work_attribute_type_allowed_value (id, work_attribute_type, value)
```
gentlecat: Forgot to add `gid` there.

gentlecat: Fixed.

gentlecat:
*I'm currently a bit stuck with running tests. See http://chatlogs.metabrainz.org/brainzbot/metabrainz/2016-03-26/?msg=3549941&page=4.*

mwiencek:
Fixed some more tests, still a bunch of failures left. But I think it would be easier for us if we merge this and let Jenkins run things (especially since you're having trouble running them on your machine).

gentlecat:
Sure. Is there anything else apart from tests that needs fixing?

mwiencek:
Not that I can think of right now. We might find out something when more tests start failing. :)

gentlecat:
Alright. :)

PR #260, opened 2016-03-21
MBS-8591: up items per page to 100

reosarevok:
Double the number of items per page as per	MBS-8591 - the idea would be to put it on beta and see if it makes it noticeably slower (it doesn't seem to affect my sandbox much, but that's so horribly slow the difference is likely to be hard to notice there in the first place).

### DIFF
--- a/lib/MusicBrainz/Server/Controller.pm
+++ b/lib/MusicBrainz/Server/Controller.pm
@@ -12,7 +12,7 @@ use Try::Tiny;
 
 __PACKAGE__->config(
     form_namespace => 'MusicBrainz::Server::Form',
-    paging_limit => 50,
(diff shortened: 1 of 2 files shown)


uklauer:
Intentional to keep it at 25 where it is 25?

reosarevok:
I'd happily up those to 50 at least, but we didn't discuss changes to edit pagination so not sure if it's desired.

uklauer:
Well, there is [MBS-8412](http://tickets.musicbrainz.org/browse/MBS-8412) about it (but no further discussion).

reosarevok:
I'd rather do that as a separate PR but I'm happy to up those too. Let's discuss that on the ticket :)

PR #261, opened 2016-03-21
MBS-8850: Add events tab to the tag page

reosarevok:


JesseWeinstein:
Looks Good To Me

uklauer:
Is the order of the tabs supposed to make any sense?

reosarevok:
I assumed not, from the current order, so I just threw it at the end - it seems right now they're just "ordered" by when they got added.

PR #262, opened 2016-03-21
MBS-8861: adding text if last login is empty

reosarevok:
I *think* this should work just fine, although I couldn't test it because every user seems to have a last login date in my sandbox.

### DIFF
--- a/root/user/profile.tt
+++ b/root/user/profile.tt
@@ -92,9 +92,13 @@
         [% END %]
 
         [% IF c.user_exists AND (viewing_own_profile OR c.user.is_account_admin);
-             WRAPPER property name=l('Last login:');
-               UserDate.format(user.last_login_date);
-             END;
+            WRAPPER property name=l('Last login:');
+                IF user.last_login_date;
+                    UserDate.format(user.last_login_date);
+                ELSE;
+                    l("Hasn't logged in yet");
+                END;
+            END;
            END %]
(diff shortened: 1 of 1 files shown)


root/user/profile.tt
```diff
@@ -91,11 +91,15 @@
             [% END %]
         [% END %]
 
-        [% IF c.user_exists AND (viewing_own_profile OR c.user.is_account_admin);
-             WRAPPER property name=l('Last login:');
-               UserDate.format(user.last_login_date);
-             END;
-           END %]
+        [% IF c.user_exists AND (viewing_own_profile OR c.user.is_account_admin) %]
+           [% WRAPPER property name=l('Last login:') %]
```
uklauer: Why did you change this whole section to go constantly back and forth between TT instructions and pass-through text, cluttering both the source file (with `[% %] [% %]`) and the output (with lots of spaces)?

reosarevok: Because that is how all the other IF-ELSEs are done in this file, so it seemed more consistent.

uklauer: I’d still prefer it as it was before (not surprising, since I intentionally wrote it that way), but if those `[% %]` are necessary, please make them `[%~ ~%]` at least.

uklauer:
Just for the record, the indentation was messed up …

PR #263, opened 2016-03-26
MBS-8873: Convert and unify artist credit editors to React

mwiencek:
Converts all this crap code to React, which is needed for the alternative tracklists UI. Sorry for the size of the patch, but this kind of stuff is just going to take forever if we don't just do it.

Some parts are terrible because we need to have bridges between knockout <-> React where both are used on the same page. I'll try to explain what's going on...

Can try it at http://bitmap.mbsandbox.org/


root/static/scripts/common/components/DescriptiveLink.js
```diff
@@ -24,7 +25,7 @@ const DescriptiveLink = ({entity, content}) => {
     return l('{entity} by {artist}', {
       __react: 'frag',
       entity: link,
-      artist: <ArtistCreditLink artistCredit={entity.artistCredit} key={1} />,
+      artist: <ArtistCreditLink artistCredit={ko.unwrap(entity.artistCredit)} key={1} />,
```
mwiencek: Some places this is an observable (releases/tracks in the release editor), some places it's not (recordings, etc.). So we need to pass it to ko.unwrap just it case. We do this for a few other variables below.

root/static/scripts/common/entity.js
```diff
@@ -30,6 +38,12 @@ const formatTrackLength = require('./utility/formatTrackLength');
                 toJSON(result, this[key], key);
             }
             return result;
+        },
+
+        renderArtistCredit: function (ac) {
+            return ReactDOMServer.renderToStaticMarkup(
+                <ArtistCreditLink artistCredit={ko.unwrap(ac)} />
```
mwiencek: The knockout stuff used to call https://github.com/metabrainz/musicbrainz-server/pull/263/files#diff-37e659b91af89773d96c6b5faa90d18cL462 to generate an html string for artist credits. Now it calls this, which generates static HTML from the react component we have. So one less instance of our custom knockout rendering crap.

root/static/scripts/common/immutable-entities.js
```diff
@@ -0,0 +1,89 @@
+// This file is part of MusicBrainz, the open internet music database.
```
mwiencek: I expect this file will completely replace root/static/scripts/common/entity.js once we no longer have any need for knockout or observables. Uses simple immutable data structures + functions instead of classes.

root/static/scripts/release-editor/fields.js
```diff
@@ -845,8 +852,118 @@ const validation = require('../edit/validation');
             $(element)
                 .prop("disabled", disabled)
                 .toggleClass("disabled-hint", disabled)
-                .attr("title", disabled ? i18n.l("This medium has one or more discids which prevent this information from being changed.") : "");
+                .attr("title", disabled ? l("This medium has one or more discids which prevent this information from being changed.") : "");
+        }
+    };
+
+
+    ko.bindingHandlers.artistCreditEditor = {
```
mwiencek: This is how the release editor's knockout stuff renders the AC editor for each track. It's initiated like https://github.com/metabrainz/musicbrainz-server/pull/263/files#diff-6edb40ffe5e08874aaa80abd6c4d1045R192

In the `update` handler below (which runs whenever the track's artistCredit observable changes), we use React to render the AC editor into the current element.

root/static/scripts/release-editor/seeding.js
```diff
@@ -79,8 +81,8 @@
         }
 
         if (data.artistCredit) {
-            release.artistCredit.setNames(data.artistCredit);
-            release.artistCredit.saved = fields.ArtistCredit(data.artistCredit);
+            release.artistCredit(artistCreditFromArray(data.artistCredit));
+            release.artistCredit.saved = release.artistCredit.peek();
```
mwiencek: Since the AC objects are immutable we can just pass the same reference around everywhere.

root/static/scripts/tests/release-editor/edits.js
```diff
@@ -786,14 +786,14 @@ test("mediumCreate edits are not given conflicting positions", function (t) {
         "position": 4,
         "name": "foo",
         "release": "f4c552ab-515e-42df-a9ee-a370867d29d1",
-        "hash": "e886dc4907c701cf89e5e7b5fdebcb521fa04e44"
+        "hash": "aca331e8e3448781852995b146feae853acbaa0e"
```
mwiencek: These changed because empty artist credits now get serialized like `[]` instead of `[{name: '', artist: null}]`, or something like that. Not sure why it used to do that but shouldn't really matter.

gentlecat:
I think left part (arrow) of the dialog looks a bit weird without proper shadow underneath. Kind of blending in:
screenshot 2016-03-28 16 11 58
Old version:
screenshot 2016-03-28 16 12 08

mwiencek:
> I think left part (arrow) of the dialog looks a bit weird without proper shadow underneath. Kind of blending in:

Yeah, I noticed that. The problem I had is that the arrow/tail is entirely a CSS border, so if you try to add a box-shadow it looks like this:

Image of bubble tail shadow weirdness

Not sure if there's a good way to do this without using an image. I'll try to find one.

gentlecat:
One inconsistency that I noticed is that when I add new artist credit it shows up as `[removed]` and not empty string.

root/static/scripts/edit/components/ArtistCreditBubble.js
```diff
@@ -0,0 +1,105 @@
...
+  copyArtistCredit,
+  done,
+  entity,
+  extraButtons,
+  extraContent,
+  hide,
+  onNameChange,
+  pasteArtistCredit,
+  removeName,
+}) => (
```
gentlecat: Is that another ES6 thing?

mwiencek: The `({addName, artistCredit, ...})` part is the "destructuring" feature, yeah. It's the same feature as `const {l} = ...` above, except it's implicitly assigned from the function argument. And the `=>` is just arrow functions.

mwiencek: Example from the React blog: https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components

gentlecat: That's cool.

root/static/scripts/edit/components/ArtistCreditNameEditor.js
```diff
@@ -0,0 +1,113 @@
...
+    super(props);
+    this.onArtistChange = this.onArtistChange.bind(this);
+    this.onNameChange = this.onNameChange.bind(this);
+    this.onJoinPhraseChange = this.onJoinPhraseChange.bind(this);
+  }
+
+  onArtistChange(artist) {
+    var update = {artist};
+    var artistName = artist ? artist.name : '';
+    var currentArtistName = _.get(this.props.name, ['artist', 'name'], '');
```
gentlecat: Any particular reason you use `var` here instead of `let`?

mwiencek: Likely absentmindedness/force of habit. Will fix. :+1:

mwiencek:
Fixed the `[removed]`, `var`, and shadow things. The only way to get the drop shadow working (besides using an image) was to use `filter: drop-shadow()` which isn't supported in IE or Opera 12, but oh well.

gentlecat:
Yep, looks better now. :+1:

I found another issue. If you open alias editor, add another value and then close it without changing anything, it's not possible to open it again. I also get these two warnings:
```
EntityLink(...): No `render` method found on the returned component instance: you may have forgotten to define `render`, returned null/false from a stateless component, or tried to render an element whose type is a function that isn't a React component.
Uncaught TypeError: inst.render is not a function
```

mwiencek:
http://chatlogs.metabrainz.org/brainzbot/metabrainz/2016-03-30/?msg=3552940&page=2

gentlecat:
I don't think I have anything else to add. :shipit:, I'd say.

mwiencek:
If you say so :+1:

PR #264, opened 2016-03-26
MBS-8874: Editing an entity sometimes shows it as a possible duplicate of itself

mwiencek:
Gets `sourceEntityGID` from the window location instead of `MB.sourceEntityGID`, in case the relationship editor code isn't running first for some reason (even though the script is outputted first on the page).


gentlecat:
LGTM

PR #265, opened 2016-03-27
Instructions to install bzip2

unreal4u:
bzip2 is not default installed on CentOS systems, so provide general instructions to do this always, even for other systems.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -298,8 +298,12 @@ Creating the database
 
             gpg --recv-keys C777580F
             gpg --verify-files /tmp/dumps/*.asc
+            
+		Before you can actually import the dumps, make sure that bzip2 is installed:
 
-        If this is OK and you wish to continue, you can import them with:
+			apt-get install bzip2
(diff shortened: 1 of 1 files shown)


unreal4u:
If you want, I can also provide CentOS instructions, they aren't that different from the Debian ones, mostly package name changes.

mwiencek:
Thanks. :+1: I think we'd be happy to accept CentOS instructions if the changes are as minimal as they sound, since we've had plenty of people try to set it up there.

INSTALL.md
```diff
@@ -326,7 +330,7 @@ Creating the database
     database level:
 
         ALTER DATABASE musicbrainz_db SET statement_timeout TO 60000;
-
+
```
mwiencek: Can you remove the unrelated whitespace added there?

unreal4u:
Oh sorry, didn't see that one, removed :)

unreal4u:
Thanks!

PR #266, opened 2016-03-30
MBS-8838: Fix tests

gentlecat:


mwiencek:
You can just push to schema-change-2016-q2, it's already broken so not a big deal if you don't fix everything at once :)

PR #267, opened 2016-04-01
Store server.js request params in redis

mwiencek:
This is concerning the data we send to root/server.js in order to render whatever page we're rendering.

Sending a body along with GET requests isn't really up to spec, so I think we ought to be doing something that doesn't rely on the nonconforming behavior of whatever modules we happen to be currently using.

This stores the JSON needed by the page in redis temporarily (keyed by URL + a random token).

This probably negligibly increases the page load time.


lib/MusicBrainz/Server/View/Node.pm
```diff
@@ -38,20 +39,28 @@ sub process {
...
+        },
+    };
 
     my $uri = URI->new;
     $uri->scheme('http');
     $uri->host(DBDefs->RENDERER_HOST || '127.0.0.1');
     $uri->port(DBDefs->RENDERER_PORT);
     $uri->path($c->req->path);
+    $uri->query_param_append('token', generate_token());
+    $uri->query_param_append('user', $c->user->name) if $c->user_exists;
```
mwiencek: The token is really enough, but adding the user too should make it impossible to get a page requested by another user (disaster scenario being that it contains private data).

mwiencek:
Since it depends on this, added in https://github.com/metabrainz/musicbrainz-server/pull/267/commits/b46fbc5e41e00c67843fd89579efa3b1b6603e48 which allows the perl to not have to wait for the requests. Mentioned that before in a previous PR. Doesn't add ETags like the Catalyst stuff does, but we can add those in later when they actually become useful.

See https://github.com/metabrainz/chef-cookbooks/commit/4f2d07f87ce31fdb3de134a7de92b8aa3a7050b2 for the nginx bits.

lib/MusicBrainz/Server/View/Node.pm
uklauer: Umm. At the moment, of course, all this is only used for the 404 page, but in theory, one day, it will be called for every single website request. I see a risk that, at that point, the kernel entropy pool could be depleted rather quickly by this.

mwiencek: Changed to use a single token per server + incrementing ID.

lib/MusicBrainz/Server/View/Node.pm
```diff
@@ -8,12 +8,18 @@ use Encode;
...
 use URI;
+use URI::QueryParam;
+
+use feature 'state';
 
 sub process {
     my ($self, $c) = @_;
 
+    state $server_token = generate_token();
+    state $request_id = 0;
```
gentlecat: Are these two values global (shared between requests)?

mwiencek: Yeah kinda. They're only initialized once, but they're lexically scoped to the `process` subroutine. So not quite global, but are shared between requests.

gentlecat: I don't know much about internals, but is it safe to increment `request_id` below like that?

mwiencek: Should be safe. It definitely gets incremented & persists across requests, if that's what you mean (made sure to test that). & Each process will have it's own state variable here, and a process can only be handling one request at a time.

gentlecat: Ah, cool. I was thinking what would happen if there were multiple simultaneous requests.

gentlecat:
LGTM

mwiencek:
Tweaked slightly so it returns a 500 error instead of 400 at https://github.com/metabrainz/musicbrainz-server/pull/267/commits/b9cb2d2fd6e9b5d21d323e2e0c66550005fd14a7#diff-b5766029be57930fe3ba6cbdb40aab5dR46

PR #268, opened 2016-04-04
Print gulp watch events

mwiencek:
Prints out something when a file is added, changed, or unlinked, and when it's done rebuilding.


gentlecat:
Useful! This works only with scripts, right?

mwiencek:
> Useful! This works only with scripts, right?

Works for styles now too. (It doesn't have a way to only build the .less files that change yet, but rebuilding everything is still really fast there.)

gentlecat:
:ship:

PR #269, opened 2016-04-07
Update IRC and mailing list references.

Freso:
We currently use #metabrainz for MusicBrainz server development discussion and https://community.metabrainz.org/ has replaced the mailing lists.


mwiencek:
:+1:

PR #270, opened 2016-04-08
MBS-8897: add new entity icons and replace old ones

caroline-gschwend:
I'm not sure the background of the icons are transparent, they don't look like it right now. I will see what I can do. 

I haven't implemented the specific artist icons yet, just those that we already use, or that use the blank one. 

I'm not sure if I got all of the places that I need to change.

Also, do I need to specify a height and width anywhere?


caroline-gschwend:
The icon backgrounds were not transparent, but that is now fixed.

Freso:
Just a side-note: I've taken the Area and Instrument icons from this PR and used as icons on Jira: https://tickets.musicbrainz.org/browse/INST and https://tickets.musicbrainz.org/browse/AREQ

root/static/search_plugins/opensearch/musicbrainz_artist.xml
```diff
@@ -6,10 +6,10 @@
   <Description>Indexed search for an artist on MusicBrainz.org</Description>
   <Url type="text/html" method="get" template="http://musicbrainz.org/search/textsearch.html?type=artist&amp;limit=25&amp;handlearguments=1&amp;query={searchTerms}"/>
   <Developer>https://metabrainz.org/contact</Developer>
-  <Image width="16" height="16">http://musicbrainz.org/static/images/entity/artist.png</Image>
+  <Image width="16" height="16">http://musicbrainz.org/static/images/entity/artist.svg</Image>
```
mwiencek: Not sure if that's going to work, to be honest. Reading https://developer.mozilla.org/en-US/Add-ons/Creating_OpenSearch_plugins_for_Firefox I only see ico, jpeg, and png mentioned. Perhaps we should still keep some rendered pngs in the repository.

root/static/styles/entity-icons.less
```diff
@@ -6,13 +6,16 @@
...
-.artistheader    { background-image: data-uri('../images/entity/artist_lg.png') }
-.releaseheader   { background-image: data-uri('../images/entity/release_lg.png') }
-.rgheader        { background-image: data-uri('../images/entity/release_group_lg.png') }
-.recordingheader { background-image: data-uri('../images/entity/recording_lg.png') }
-.placeheader     { background-image: data-uri('../images/entity/place_lg.png') }
-.labelheader     { background-image: data-uri('../images/entity/label_lg.png') }
-.workheader      { background-image: data-uri('../images/entity/work_lg.png') }
-.areaheader, .collectionheader, .seriesheader, .instrumentheader, .eventheader,
- { background-image: data-uri('../images/entity/blank_lg.png') } /* TODO: create remaining icons */
+.blankheader      { background-image: data-uri('../images/entity/blank.svg') }
```
mwiencek: All of these now need `background-size: 32px;`

root/static/styles/entity-icons.less
```diff
@@ -6,13 +6,16 @@
     background: 0 4px no-repeat;
```
mwiencek: This should now be `background: 0 0 no-repeat;`, from what I can see in Chrome.

mwiencek:
The icons look great, thanks so much for making these. It just needs some tweaks to the CSS to make it display properly on the website.

caroline-gschwend:
@mwiencek ,  I changed all of the SVGs so they are now square, added some PNGs to be used in the opensearch parts, and made the changes specified above to the .less files.

mwiencek:
Sorry for the delay again! I tested this locally and only found one small issue in the CSS, which I've fixed. So let's go ahead and merge. The icons look really nice!

PR #271, opened 2016-04-11
MBS-8720: Materialized view of edit note recipients

gentlecat:


gentlecat:
Do we need to remove `edit_note_idx_post_time_edit`? https://github.com/metabrainz/musicbrainz-server/blob/master/admin/sql/updates/20151223-edit-note-index.sql#L4-L5

mwiencek:
I think the `edit_note_idx_post_time_edit` index is still useful for the `ORDER BY post_time DESC, edit DESC` part of the query since it can return things in index order there. We should add that index to the upgrade script, actually. There's currently a broken/unusable `edit_note_idx_post_time_edit` on totoro (took too long to create, had to cancel it), so we might need a `DROP INDEX ... IF EXISTS` before the `CREATE INDEX ...`. :/

lib/MusicBrainz/Server/Data/EditNote.pm
```diff
@@ -120,9 +120,9 @@ sub find_by_recipient {
 
     my $query = <<EOSQL;
         SELECT ${\($self->_columns)}
-          FROM ${\($self->_table)}
+          FROM edit_note_recipient
+          JOIN ${\($self->_table)} ON ${\($self->_table)}.id = edit_note_recipient.edit_note_id
          WHERE editor != \$1
-           AND edit IN (SELECT id FROM edit WHERE editor = \$1)
            AND post_time > now() - interval '3 months'
```
mwiencek: I think we can remove this 3-month limitation now, it should be fast enough if you wanna view edit notes further back than that.

admin/sql/CreateTables.sql
```diff
@@ -382,6 +382,11 @@ CREATE TABLE edit_note
     post_time            TIMESTAMP WITH TIME ZONE DEFAULT NOW()
 );
 
+CREATE TABLE edit_note_recipient (
+    editor_id           INTEGER NOT NULL, -- PK, references editor.id
```
mwiencek: I realized we don't really use this `_id` naming scheme in other tables, we'd just use `editor` and `edit_note` and know it means the primary key, so maybe we should rename them for consistency.

admin/sql/CreateFunctions.sql
```diff
@@ -1079,4 +1079,16 @@ CREATE OR REPLACE FUNCTION track_count_matches_cdtoc(medium, int) RETURNS boolea
...
 
 COMMIT;
+
+-----------------------------------------------------------------------
+-- edit_note triggers
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION a_insert_edit_note() RETURNS trigger AS $$
+BEGIN
+    INSERT INTO edit_note_recipient (editor_id, edit_note_id) VALUES (NEW.editor, NEW.id);
```
mwiencek: `NEW.editor` is the person who left the note, so that actually needs to get the edit author from the `edit` table (join on `NEW.edit = edit.id`).

admin/sql/CreateTables.sql
mwiencek: But `recipient` is still probably more clear than `editor`.

admin/sql/updates/20160516-mbs-8720.sql
```diff
@@ -0,0 +1,35 @@
...
+   REFERENCES editor(id);
+
+ALTER TABLE edit_note_recipient
+   ADD CONSTRAINT edit_note_recipient_fk_edit_note
+   FOREIGN KEY (edit_note_id)
+   REFERENCES edit_note(id);
+
+CREATE INDEX edit_note_recipient_idx_editor_id ON edit_note_recipient (editor_id);
+
+-- Trigger
```
mwiencek: We'll need the function + trigger to go in a separate file, say `20160516-mbs-8720-triggers.sql`, which only runs on standalone servers. Don't want these to be modifying `edit_note_recipient` on slave servers since that table will be replicated. (Then in upgrade.json just add the -triggers file to the standalone list.)

mwiencek: Actually, the table won't be replicated, so ignore this comment!

mwiencek:
Table also needs adding to `@FULL_TABLE_LIST` in lib/MusicBrainz/Server/Constants.pm and `@group_edit` in admin/ExportAllTables

admin/sql/updates/20160516-mbs-8720.sql
gentlecat: 👌

gentlecat:
Done. Also added copying of existing data into the change script.

admin/sql/CreateFunctions.sql
```diff
@@ -1079,4 +1079,21 @@ CREATE OR REPLACE FUNCTION track_count_matches_cdtoc(medium, int) RETURNS boolea
 $$ LANGUAGE SQL IMMUTABLE;
 
 COMMIT;
+
+-----------------------------------------------------------------------
+-- edit_note triggers
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION a_insert_edit_note() RETURNS trigger AS $$
```
mwiencek: Nitpick: our other functions and triggers all use `a_ins_`, not `a_insert_`.

mwiencek:
Looks like you need to run `./admin/GenerateSQLScripts.pl admin/sql/` so the ./admin/sql/Drop* files all get updated.

admin/sql/CreateFunctions.sql
gentlecat: Oops, fixed.

gentlecat:
Generated `./admin/sql/Drop*` files.

admin/sql/CreateFunctions.sql
```diff
@@ -1079,4 +1079,21 @@ CREATE OR REPLACE FUNCTION track_count_matches_cdtoc(medium, int) RETURNS boolea
...
+BEGIN
+    INSERT INTO edit_note_recipient (recipient, edit_note) (
+        SELECT edit.editor, NEW.id
+          FROM edit_note
```
mwiencek: It doesn't look like the edit_note table is used here, should work if you just select from edit where NEW.edit = edit.id?

admin/sql/CreateFunctions.sql
mwiencek: Trigger is fixed but the function name still needs changing.

admin/sql/CreateFunctions.sql
gentlecat: Yeah, that seems to be working.

admin/sql/CreateFunctions.sql
gentlecat: Updated.

lib/MusicBrainz/Server/Data/EditNote.pm
```diff
@@ -120,10 +120,9 @@ sub find_by_recipient {
 
     my $query = <<EOSQL;
         SELECT ${\($self->_columns)}
-          FROM ${\($self->_table)}
+          FROM edit_note_recipient
+          JOIN ${\($self->_table)} ON ${\($self->_table)}.id = edit_note_recipient.edit_note
          WHERE editor != \$1
-           AND edit IN (SELECT id FROM edit WHERE editor = \$1)
```
mwiencek: We need a `AND recipient = \$1` to replace that.

gentlecat: Oh, I was checking with only one user. Done. :)

mwiencek:
LGTM, I think. :) I'll check again tomorrow and we can merge it then. Thanks for working on this.

admin/sql/updates/20160516-mbs-8720.sql
```diff
@@ -0,0 +1,47 @@
...
+      JOIN edit ON edit_note.edit = edit.id
+);
+
+CREATE INDEX edit_note_recipient_idx_editor ON edit_note_recipient (recipient);
+
+-- Trigger
+
+CREATE OR REPLACE FUNCTION a_ins_edit_note() RETURNS trigger AS $$
+BEGIN
+    INSERT INTO edit_note_recipient (recipient, edit_note) (
```
mwiencek: (I lied, found another thing before I closed the tab.) Function definition here needs to be updated with the new one from admin/sql/CreateFunctions.sql.

gentlecat: Late night coding. 🐞

mwiencek:
The test failures are just 'cause the jenkins stuff is using the master-branch job which doesn't point to the new mmd-schema. LGTM otherwise.

PR #272, opened 2016-04-12
Don't ISE if CB doesn't return JSON

mwiencek:


root/release_group/index.tt
```diff
@@ -74,13 +74,13 @@
...
     [%- INCLUDE "components/relationships.tt" source=rg -%]
 
-    <h2>[% l('CritiqueBrainz Reviews') %]</h2>
-
-    [%- PROCESS 'components/critiquebrainz.tt' -%]
-
-    <p>
-      [%~ critiquebrainz_links(rg) ~%]
-    </p>
+    [%~ IF rg.review_count.defined ~%]
```
mwiencek: Probably no point in showing an error here. It shouldn't happen often.

gentlecat:
:shipit:

PR #273, opened 2016-04-16
MBS-8909: Add IMDb and VGMdb favicons to entity sidebar

caroline-gschwend:


root/static/styles/favicons.less
```diff
@@ -34,6 +34,7 @@
 .favicon("genius");
 .favicon("googleplus", 32);
 .favicon("ibdb");
+.favicon("IMDb");
```
Freso: This shouldn't have the `, 32` like googleplus and instagram etc.? (I don't know what that value is, I just noted that the imdb.png is called -32 while the vmdb one is -16…)

Freso: Also, all the others are all lowercase, so this and VGMdb should probably be as well.

gentlecat:
@brainzbot, add to whitelist.

root/static/styles/favicons.less
caroline-gschwend: @Freso , what about "iTunes" and "SMDB"? I'm confused on the capitalizations here.

Freso: I only looked at the ones in the diff. :) I would go with the more prevalent style, which, based on the diffs, seem to be all lower vase.

caroline-gschwend: @Freso, do you want me to change all of them?

root/static/styles/favicons.less
```diff
@@ -74,6 +75,7 @@
 .favicon("tunearch");
 .favicon("twitter");
 .favicon("viaf", 32);
+.favicon("vgmdb");
```
mwiencek: nitpick, but that should go above viaf

mwiencek:
To make the new icons get used, you also have to modify `FAVICON_CLASSES` in lib/MusicBrainz/Server/Entity/URL/Sidebar.pm. Here's a helpful example: https://bitbucket.org/metabrainz/musicbrainz-server/pull-requests/1646/mbs-8560-add-google-play-to-the-sidebar/diff

mwiencek:
Don't worry about the test failure, it's random and unrelated (gotta figure out how to fix that soon).

mwiencek:
Thanks! 👍

PR #274, opened 2016-04-21
MBS-3887: Check for duplicates in alias names

gentlecat:


lib/MusicBrainz/Server/Controller/WS/js/CheckDuplicates.pm
```diff
@@ -37,11 +37,18 @@ sub check_duplicates : Chained('root') PathPart('check_duplicates') Args(0) {
         $c->detach;
     };
 
+    my @duplicates;
+    if (defined $c->model($model)->search_by_names) {
```
mwiencek: That actually calls the method AFAIK (just without any arguments), so a safer way is `$c->model($model)->can('search_by_names')`.

gentlecat: Huh, didn't know. Updated.

mwiencek:
A problem is that if any duplicates are returned, the UI forces you to add a disambiguation comment. But we should only require that where the name duplicates another name, not only an alias. So we may need to modify the query in `search_by_names` to return whether the search term matches a name column, then return items like `{matches_name => 1, entity => $entity}`. It would be way easier if we could just do the unaccenting in the JavaScript, but I don't trust that there's any libraries that work 100% the same.

mwiencek:
Another thing you could do is add an `unaccented_name` attribute to the Entity class, then select `musicbrainz_unaccent(name) AS unaccented_name` and send that to the JavaScript. Not sure why I didn't consider that before.

gentlecat:
Right. Should be better now.

root/static/scripts/edit/check-duplicates.js
```diff
@@ -115,12 +115,17 @@ function isPlaceCommentRequired(duplicates) {
...
 }
 
-function isCommentRequired(type, duplicates) {
+function isCommentRequired(type, name, duplicates) {
   if (type === 'place') {
     return isPlaceCommentRequired(duplicates);
   }
 
-  return duplicates.length > 0;
+  for (let duplicate of duplicates) {
```
mwiencek: I remember `let ... of ...` not working before, it needed `Symbol` to be defined or something. Can we replace with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some?

mwiencek:
LGTM other than JavaScript being dumb.

root/static/scripts/edit/check-duplicates.js
gentlecat: Done. 👍

mwiencek:
Looks like some /ws/js tests broke 'cause of the new property being outputted.

mwiencek:
I removed unused code in ef48045 and fixed a locale-specific issue I ran into while testing in 9334aea (unrelated to your changes, which look good).

PR #276, opened 2016-04-24
Added metabrainz_user_id field containing $c->user->id to /userinfo endpoint response

LordSputnik:
The /userinfo endpoint currently provides no invariant data about the user - the name and email can change. 3rd party applications that store additional information about users need invariant data to reliably match their own information to the user information from MusicBrainz.

This PR introduces an ID field into the /userinfo response, which contains the internal database ID of the user. As this is the user's PK, this is unlikely to change.


mwiencek:
The only thing different I would do is to use something more specific than `id`, so that it's clear what sort of id it is, and so we don't clash with the standard claims or anything else. How about `musicbrainz_user_id`?

LordSputnik:
OK, updated it to `metabrainz_user_id` as discussed in IRC:
http://chatlogs.metabrainz.org/brainzbot/metabrainz/2016-04-24/?msg=3573357&page=3

PR #277, opened 2016-04-25
MBS-8800: Add a geodetic index on place

uklauer:
Create the `earthdistance` extension and use it for a geodetic index on the `place` table, enabling fast queries for places near a given point or within a given region.

### DIFF
--- a/admin/sql/CreateIndexes.sql
+++ b/admin/sql/CreateIndexes.sql
@@ -389,6 +389,7 @@ CREATE UNIQUE INDEX medium_format_idx_gid ON medium_format (gid);
 CREATE UNIQUE INDEX place_idx_gid ON place (gid);
 CREATE INDEX place_idx_name ON place (name);
 CREATE INDEX place_idx_area ON place (area);
+CREATE INDEX place_idx_geo ON place USING gist (ll_to_earth(coordinates[0], coordinates[1])) WHERE coordinates IS NOT NULL;
 
 CREATE INDEX place_alias_idx_place ON place_alias (place);
 CREATE UNIQUE INDEX place_alias_idx_primary ON place_alias (place, locale) WHERE primary_for_locale = TRUE AND locale IS NOT NULL;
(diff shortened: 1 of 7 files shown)


admin/sql/sitemaps/CreateFKConstraints.sql
```diff
@@ -1,7 +1,7 @@
 -- Automatically generated, do not edit.
 \set ON_ERROR_STOP 1
 
-SET search_path = 'sitemaps';
+SET search_path = 'sitemaps', 'musicbrainz', 'public';
```
uklauer: This is necessary because the `earthdistance` extension expects its function names (in the `public` schema) to be always visible. (This can probably be considered a bug: http://www.postgresql.org/message-id/flat/CALy1bpd2rDKk-V246KD79NTSBGYP0j7L4N+yQ33U85uRUUevoQ@mail.gmail.com) Due to the way foreign key constraints are added internally in Postgres, the problem would be triggered below when a FK to `place` is defined.

mwiencek:
@brainzbot, add to whitelist.

admin/sql/CreateIndexes.sql
```diff
@@ -389,6 +389,7 @@ CREATE UNIQUE INDEX medium_format_idx_gid ON medium_format (gid);
 CREATE UNIQUE INDEX place_idx_gid ON place (gid);
 CREATE INDEX place_idx_name ON place (name);
 CREATE INDEX place_idx_area ON place (area);
+CREATE INDEX place_idx_geo ON place USING gist (ll_to_earth(coordinates[0], coordinates[1])) WHERE coordinates IS NOT NULL;
```
uklauer: This is the core addition. `ll_to_earth` is a function that essentially takes a long/lat position and computes a 3D point from it, assuming a perfectly spherical Earth. This 3D point is placed in a GiST index.

mwiencek:
Lgtm. Thanks 👍

PR #278, opened 2016-04-28
MBS-8928: Properly display not-logged-in error

mwiencek:
If the response contains an object with a "message" property, that should be what's displayed, not the object itself.


gentlecat:
LGTM

PR #280, opened 2016-04-30
MBS-4501: Alternative tracklists schema

mwiencek:
Just the schema changes. Hopefully I didn't forget anything (realized I forgot triggers a few minutes before I sent this).


admin/sql/updates/20160310-mbs-4501-alternative-tracklists.sql
```diff
@@ -0,0 +1,134 @@
...
+    name                    VARCHAR
+    CHECK (name != '')
+);
+
+CREATE TABLE alternative_track (
+    id                      SERIAL,
+    name                    VARCHAR,
+    artist_credit           INTEGER,
+    ref_count               INTEGER NOT NULL DEFAULT 0
+    CHECK (name != '' AND (name IS NOT NULL OR artist_credit IS NOT NULL))
```
gentlecat: This checks that either `name` or `artist_credit` is defined, right? And `name != ''` to make sure that no empty strings get inserted.

mwiencek: Yeah, exactly. The intent is that `NULL` means it should fallback to the original track's name or artist. But it doesn't make sense for both to be `NULL`.

upgrade.json
```diff
@@ -26,7 +26,8 @@
   "23": {
     "slave": ["20160516-mbs-8838.sql",
               "20160516-mbs-8720.sql",
-              "20141226-geodesic-index.sql"],
-    "standalone": []
+              "20141226-geodesic-index.sql",
+              "20160310-mbs-4501-alternative-tracklists.sql"],
+    "standalone": ["20160310-mbs-4501-alternative-tracklists-fks.sql"]
```
gentlecat: How do these changes with `slave`, `slave_only`, and `standalone` work? Don't think I asked about that before.

Also why are foreign keys created only in `standalone`?

mwiencek: When we're done with the schema changes, we'll run `./admin/CompileSchemaScripts.pl 23` to generate all these individual scripts concatenated together, but all within a single transaction, so that they're not half-applied.

You can see examples in admin/sql/updates/schema-change/. There are different files for slave, slave_only, and standalone. When people run `upgrade.sh` on their local MBS, it'll execute these files depending on their `REPLICATION_TYPE`.

Foreign keys aren't created on slave servers because they'll usually cause replication packets to fail to apply. The state of the database ends up the same, but it violates those constraints sometimes when it's applying individual changes.

gentlecat:
LGTM

PR #281, opened 2016-05-07
MBS-8669: Primary keys for place tables missing on old slaves

mwiencek:
According to the ticket this can go in slave_only, but it should be harmless to run it everywhere just in case.


gentlecat:
LGTM

PR #282, opened 2016-05-07
MBS-8727: Prevent duplicate votes

mwiencek:


admin/sql/updates/20160507-mbs-8727.sql
```diff
@@ -0,0 +1,17 @@
...
+    1310304, -- edit #752619
+    1381783, -- edit #835828
+    1389065, -- edit #854892 (vote_time is before id=1389064)
+    1429402, -- edit #916266
+    1572699, -- edit #1136298 (vote_time is before id=1572698)
+    1611602  -- edit #1194378
+);
+
+DROP INDEX IF EXISTS vote_idx_editor_edit;
+CREATE UNIQUE INDEX vote_idx_editor_edit ON vote (editor, edit) WHERE superseded = FALSE;
```
mwiencek: Same as the existing `vote_idx_editor_edit` except with `UNIQUE`. If this fails during testing, we'll know we didn't catch all the bad votes above.

admin/sql/updates/20160507-mbs-8727.sql
```diff
@@ -0,0 +1,17 @@
...
+
+BEGIN;
+
+UPDATE vote SET superseded = 't' WHERE id IN (
+    1310304, -- edit #752619
+    1381783, -- edit #835828
+    1389065, -- edit #854892 (vote_time is before id=1389064)
+    1429402, -- edit #916266
+    1572699, -- edit #1136298 (vote_time is before id=1572698)
+    1611602  -- edit #1194378
```
mwiencek: Double-checking these would be appreciated. :)

gentlecat: Edits look right, but I can't check the votes since I don't have them in my local DB.

gentlecat:
LGTM

PR #284, opened 2016-05-09
MBS-8804: Add BRIN indexes for timestamp columns

gentlecat:


mwiencek:
Not 100% sure about dropping the btree ones, but my gut tells me it's probably safe. :) I'll run some tests after I do a proper upgrade, and if it's a disaster we can add them back. :shipit: !

gentlecat:
👍

PR #285, opened 2016-05-09
MBS-8288: Drop existing indexes on *_gid_redirect tables

gentlecat:
*Addition to #283. See [comment in JIRA](http://tickets.musicbrainz.org/browse/MBS-8288?focusedCommentId=35878&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_35878).*

### DIFF
--- a/admin/sql/updates/20160516-mbs-8656.sql
+++ b/admin/sql/updates/20160516-mbs-8656.sql
@@ -1,6 +1,26 @@
 \set ON_ERROR_STOP 1
 BEGIN;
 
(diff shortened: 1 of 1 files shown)


mwiencek:
We can drop the gid ones that exist on totoro, too, i.e. `artist_gid_redirect_idx_gid` and the others that end in `idx_gid`. I'm not sure why those were added in addition to the primary keys.

gentlecat:
Updated.

PR #286, opened 2016-05-11
MBS-6406: allow admins to skip the email verification process

reosarevok:
Tested by hand and it seems to work (inasmuch as I can test it since rika doesn't seem to send emails)


gentlecat:
@brainzbot, add to whitelist.

mwiencek:
LGTM.

PR #291, opened 2016-05-14
MBS-8287: Keep a record of deleted or merged entities

uklauer:
For entity types that can be subscribed to (artist, label, series), deleted entities were already recorded in `artist_deletion` etc. tables, so that subscribers can be notified when one of their subscribed entities has disappeared.

Extend this to all core entity types (except URLs), using a common `deleted_entity` table with a JSON `data` column that can store various information about the former entities. In addition to retaining historical data, this will allow similar notifications about deleted entities that were in a subscribed collection.

In preparation for this, also add a table for keeping track of which collections deleted entities were in. (Only adds the table for later use; no code to access it.)


admin/sql/CreateTables.sql
```diff
@@ -400,6 +392,12 @@ CREATE TABLE country_area ( -- replicate (verbose)
     area                INTEGER -- PK, references area.id
 );
 
+CREATE TABLE deleted_entity (
```
uklauer: I don’t think it makes sense to have separate tables per entity type here, because there is no added type safety (we can’t have foreign keys to the relevant `artist` etc. tables because the entries have been removed from those).

admin/sql/CreateTables.sql
```diff
@@ -2156,6 +2146,11 @@ CREATE TABLE editor_collection_work (
     work INTEGER NOT NULL -- PK, references work.id
 );
 
+CREATE TABLE editor_collection_deleted_entity (
+    collection INTEGER NOT NULL, -- PK, references editor_collection.id
+    gid UUID NOT NULL -- PK, references deleted_entity.gid
+);
```
uklauer: Similar to the existing `editor_subscribe_artist_deleted` etc. tables for direct subscriptions.

admin/sql/updates/20160514-mbs-8287-a.sql
```diff
@@ -0,0 +1,46 @@
...
+    data JSONB NOT NULL,
+    deleted_at timestamptz NOT NULL DEFAULT now()
+);
+
+
+INSERT INTO deleted_entity (gid, deleted_at, data)
+SELECT gid, deleted_at,
+       jsonb_object('{last_known_name, last_known_comment, entity_gid, entity_type}',
+                ARRAY[last_known_name, last_known_comment, gid::text, 'artist']) AS data
+  FROM artist_deletion;
```
uklauer: No need to throw out the existing data about deleted artists, labels, and series.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -247,7 +247,7 @@ sub delete
     $self->rating->delete(@artist_ids);
     $self->subscription->delete(@artist_ids);
     $self->remove_gid_redirects(@artist_ids);
-    $self->delete_returning_gids('artist', @artist_ids);
+    $self->delete_returning_gids(@artist_ids);
```
uklauer: The function name is admittedly a bit strange since we aren’t actually using the returned MBIDs, but there is one place where they are used, and I wasn’t in a mood to change the name.

lib/MusicBrainz/Server/Data/Subscription.pm
```diff
@@ -165,13 +165,14 @@ sub get_subscriptions
...
-               sub.editor, last_known_name, last_known_comment, deleted_by,
+               sub.editor, del.data->>'last_known_name' AS last_known_name,
+               del.data->>'last_known_comment' AS last_known_comment, deleted_by,
                CASE
                  WHEN edit.type = any(?) THEN 'merged'
                  WHEN edit.type = any(?) THEN 'deleted'
                END AS reason
              FROM ${table}_deleted sub
-             JOIN ${column}_deletion del USING (gid)
+             JOIN deleted_entity del USING (gid)
```
uklauer: Here is where the information is used for (direct) subscription emails.

mwiencek:
Thank you!

PR #293, opened 2016-05-16
MBS-8719: Stop materializing of edit and vote counts

uklauer:


admin/sql/updates/20160515-recluster.sql
```diff
@@ -0,0 +1,8 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
+CLUSTER edit USING edit_pkey;
+
+ANALYZE edit;
```
uklauer: The `CLUSTER` will rewrite the table in order of edit IDs. This is important for the performance of the BRIN indexes from MBS-8804 and will ensure that the data from the dropped columns (most importantly, the huge `data` column) is actually removed from the table. (`ALTER TABLE DROP COLUMN` only hides the column, leaving the actual rows unchanged.)

lib/MusicBrainz/Server/Edit.pm
```diff
@@ -135,6 +130,14 @@ has 'votes' => (
     }
 );
 
+sub yes_votes {
+    shift->_grep_votes(sub { $_->vote == $VOTE_YES });
+}
+
+sub no_votes {
+    shift->_grep_votes(sub { $_->vote == $VOTE_NO });
+}
```
uklauer: This is the change we talked about on IRC: Just evaluate the votes instead of separately loading vote counts.

upgrade.json
```diff
@@ -31,11 +31,12 @@
...
               "20160310-mbs-4501-alternative-tracklists.sql",
               "20160516-mbs-8656.sql",
-              "20160516-mbs-8804.sql",
               "20160507-mbs-8727.sql",
               "20160512-edit-data.sql",
               "20160514-mbs-8287-a.sql",
-              "20160514-mbs-8287-b.sql"],
+              "20160514-mbs-8287-b.sql",
+              "20160515-recluster.sql",
+              "20160516-mbs-8804.sql"],
```
uklauer: As I said above, reclustering should happen before creating the BRIN indexes in `20160516-mbs-8804.sql`.

mwiencek:
Thanks again! Good catch with the `CLUSTER` command.

PR #294, opened 2016-05-16
Split MockEdit class from EditQueue tests into two

uklauer:
We need two different mock edit types for the `EditQueue` tests: one that provides access to real votes from the database, and one that also mocks the vote counts. Split `MockEdit` accordingly.


gentlecat:
That fixed *Server::EditQueue* for me. 👍

PR #297, opened 2016-05-23
Multiple issues with read-only mode (MBS-8943)

uklauer:
The read-only phase revealed some issues with the read-only mode. See the individual commits for details.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/CDTOC.pm
+++ b/lib/MusicBrainz/Server/Controller/CDTOC.pm
@@ -139,7 +139,7 @@ sub set_durations : Chained('load') PathPart('set-durations') Edit
     );
 }
 
-sub attach : Local DenyWhenReadOnly
+sub attach : Local DenyWhenReadonly
 {
     my ($self, $c) = @_;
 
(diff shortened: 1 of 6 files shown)


uklauer:
I branched off from `production`, so it should be possible to deploy it even before the schema change if desired.

lib/MusicBrainz/Server/Controller/CDTOC.pm
```diff
@@ -139,7 +139,7 @@ sub set_durations : Chained('load') PathPart('set-durations') Edit
     );
 }
 
-sub attach : Local DenyWhenReadOnly
+sub attach : Local DenyWhenReadonly
```
uklauer: Yes, it’s case-sensitive.

lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
```diff
@@ -578,6 +582,9 @@ sub edit : Chained('/') PathPart('ws/js/edit') CaptureArgs(0) Edit {
     unless ($c->user->has_confirmed_email_address) {
         $c->forward('/ws/js/detach_with_error', ['a confirmed email address is required']);
     }
+    if ($c->user->is_editing_disabled) {
+        $c->forward('/ws/js/detach_with_error', ['you are not allowed to enter edits']);
```
uklauer: This is not related to read-only mode, but was another thing not in sync with the regular edit pages.

mwiencek:
Thanks! :)

PR #298, opened 2016-05-25
Fix the search_path for replication

uklauer:
There was another instance of a `musicbrainz`-only `search_path`. This caused applying replication packets to fail.


uklauer:
Looks like the `-schema-change` tag was made a bit too soon. ;-)

PR #299, opened 2016-05-26
MBS-8950: Fix some favicon class names (in particular for iTunes)

uklauer:
There was some uppercase/lowercase mismatch between favicon class names, meaning some favicons weren’t displayed. Furthermore, there was an entry for a STcollector favicon even though we don’t have one.


uklauer:
The branch is based on production. Only the last two commits are actually part of the PR.

mwiencek:
Thanks 👍

PR #301, opened 2016-05-27
MBS-8957: Update SecondHandSongs icon and add BBC logo in the sidebar

y-van-z:
http://tickets.musicbrainz.org/browse/MBS-8957

PR #300, opened 2016-05-27
MBS-8824: Update the rateyourmusic logo used in the sidebar

y-van-z:
http://tickets.musicbrainz.org/browse/MBS-8824

PR #302, opened 2016-05-28
MBS-8925: Limit BookBrainz relationship to BookBrainz URLs

y-van-z:
- Limit BookBrainz relationship to BookBrainz entity URLs
- Cleanup BookBrainz entity URLs (along with tests)

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -53,6 +53,13 @@ const LINK_TYPES = {
   amazon: {
     release: "4f2e710d-166c-480c-a293-2e2c8d658d87"
   },
+  bookbrainz: {
+    artist: "f82f9342-a08d-46b7-ab7a-d8b6330c805d",
+    label: "b7be2ca4-bdb7-4d87-9619-f2fa50120409",
+    release: "63b84620-ba52-4630-9bfe-8ad3b5504dff",
+    release_group: "27cfc95c-d368-45a9-ae0d-308c274c2017",
+    work: "0ea7cf4e-93dd-4bc4-b748-0f1073cf951c"
+  },
   license: {
     release: "004bd0c3-8a45-4309-ba52-fa99f3aa3d50",
     recording: "f25e301d-b87b-4561-86a0-5d2df6d26c0a"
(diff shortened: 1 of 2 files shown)


gentlecat:
@brainzbot, test this please.

y-van-z:
Sorry about the numerous commits spaghetti in the previous PRs. Got it right this time. (rebase/fixup)
https://github.com/metabrainz/musicbrainz-server/pull/300#issuecomment-222335464
https://github.com/metabrainz/musicbrainz-server/pull/301#issuecomment-222335295

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1530,6 +1551,32 @@ test('Cleanup', function (t) {
             [
                 'http://en.wikipedia.org/wiki/Ramesh_Vinayakam?oldformat=true',
                 'http://en.wikipedia.org/wiki/Ramesh_Vinayakam'
+            ],
+            // BookBrainz
+            [
+                'https://bookbrainz.org/creator/8f3d202f-fa37-4b71-9e81-652db0f8b83d',
```
Freso: Maybe add a `?test` argument or something to the URL or an `#anchor` to make sure those also get filtered out?

y-van-z: Thanks for the review! I added the tests below and fixed the matching pattern.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -821,6 +835,15 @@ validationRules[LINK_TYPES.bbcmusic.artist] = function (url) {
   return /bbc\.co\.uk\/music\/artists\//.test(url);
 };
 
+// allow only BookBrainz entity pages with the BookBrainz rel
+function validateBookBrainz(url) {
+  return /bookbrainz\.org\/[^\/]*\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}/.test(url);
```
mwiencek: I think it'd be worthwhile to make the validation more precise so that `LINK_TYPES.bookbrainz.artist` can only match `/creator/` sub-paths, and etc. for the other entity types. Assuming that's all that's allowed for them.

y-van-z: That was in my first draft, but I removed it as it is out of the scope of the current issue which is: _to prevent using the BookBrainz relationship with unrelated websites such as Amazon_.

I would be happy to make the validation more precise if there was a note on which BB entities should be allowed in relationships for each MB entity (see the test file for non-exhaustive examples). What about artist’s biography for example? Should it be linked to the publication, to the edition, or to the work? Moreover, there are more BookBrainz entity types to come (any soon?). Maybe it is too early to decide on this additional issue. I will try to gather more information about it.

mwiencek: I don't know anything about BookBrainz, so let's just merge as-is then.

mwiencek:
Thanks :shipit:

root/static/scripts/edit/URLCleanup.js
y-van-z: Neither do I. More info for later on: http://chatlogs.metabrainz.org/brainzbot/metabrainz/msg/3618257/

PR #303, opened 2016-05-31
Serialize front cover in release index JSON-LD

mwiencek:
This may fix the image previews Facebook generates when you link to a release page there. Thanks to @zas for reporting and helping test this.


gentlecat:
LGTM

PR #304, opened 2016-05-31
MBS-8959: Generate gids for admin-added types

mwiencek:
I think this should cover all the types which we added gids for.


gentlecat:
LGTM.

Is that happening because of `gid` rows that I added for the schema change?

uklauer:
Yes. :)

mwiencek:
Yeah, I actually forgot we had a UI for adding attributes, since I never used it...

PR #305, opened 2016-05-31
MBS-8941: ACs don't change with release artist

mwiencek:
When all of the ArtistCreditEditor components mount, they each try to take over the (one) AC bubble on the screen. This can become a problem if you open the release AC bubble before mediums have loaded. If that happens, you'll likely end up editing the last track AC instead.


gentlecat:
LGTM

PR #306, opened 2016-05-31
MBS-8942: Keep cursor position in artist field

mwiencek:
Setting the value also resets the cursor position, so don't set it unless the value actually changed.


gentlecat:
LGTM

PR #307, opened 2016-06-01
MBS-7986: make the setlist error message more clear

reosarevok:


uklauer:
I don’t think this is a good change; with it, people will enter something like

    "@ "Foo Bar

and wonder why that doesn’t work.

Furthermore, this is really about MBS-7986. MBS-8960 is just a duplicate (and not the first one).

reosarevok:
You really, honestly think someone will think "oh, I'm sure they mean add the quotes and all"?

uklauer:
Yes, I’d think that would be better, though it’s difficult to predict whether someone would get it wrong or not.

Note that my suggestion went on:

> … or (probably better) accept the input add the missing space on the fly.

Freso:
> You really, honestly think someone will think "oh, I'm sure they mean add the quotes and all"?

I've seen this happen before. Don't overestimate how low the lowest common denominator is…

reosarevok:
Added the quotes and space to the default text, and "followed by a space" instead of the quotes to the error text. That should be safe enough I hope?

@uklauer: I agree that the best fix would be to make setlist editing stop being horrible. But in the meantime, and since I don't have time for that at the moment, I'd rather make it so that people understand what's going on.

uklauer:
OK with the new wording. Can you squash the two commits into one and change the commit message to say MBS-7986 instead?

PR #308, opened 2016-06-01
add new series icon

caroline-gschwend:
The old series icon was not centered, that is fixed with this version.


mwiencek:
Cool, thanks 👍

PR #309, opened 2016-06-01
Two small edit data fixes

mwiencek:
Commits seem self-explanatory.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Edit.pm
+++ b/lib/MusicBrainz/Server/Data/Edit.pm
@@ -7,6 +7,7 @@ use Data::Dumper::Concise;
 use Data::OptList;
 use DateTime;
 use DateTime::Format::Pg;
+use Encode qw( decode );
 use Try::Tiny;
 use List::MoreUtils qw( uniq zip );
 use List::AllUtils qw( any );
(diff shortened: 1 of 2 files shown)


mwiencek:
I didn't investigate where the encoding issue came from. I just assumed it had something to do with text vs. jsonb.

lib/MusicBrainz/Server/Data/Edit.pm
```diff
@@ -76,7 +77,7 @@ sub _new_from_row
         expires_time => $row->{expire_time},
         auto_edit => $row->{autoedit},
         status => $row->{status},
-        raw_data => $row->{data},
+        raw_data => decode('utf-8', $row->{data}),
```
gentlecat: This one is different from `$data` value from above, right?

    my $data = JSON::XS->new->utf8->decode($row->{data});

gentlecat:
LGTM

lib/MusicBrainz/Server/Data/Edit.pm
uklauer: More or less, the `$data` is a Perl hash representation of what is contained in the `$row->{data}` JSON string.

uklauer:
I can only imagine that DBI does some special casing for `text` but not for `jsonb` columns, but not sure either.

PR #311, opened 2016-06-02
Don't populate edit_note_recipient for own notes (related: MBS-8963)

mwiencek:
Stop storing rows in `edit_note_recipient` for edit notes left by the author of the edit. Makes the table a lot smaller and removes the need for a filter in `Data::EditNote::find_by_recipient`.

This means the table no longer maps any arbitrary edit note to the edit author. But we don't need to use it like that.

### DIFF
--- a/admin/sql/CreateFunctions.sql
+++ b/admin/sql/CreateFunctions.sql
@@ -1160,6 +1160,7 @@ BEGIN
         SELECT edit.editor, NEW.id
           FROM edit
          WHERE edit.id = NEW.edit
+           AND edit.editor != NEW.editor
     );
     RETURN NULL;
 END;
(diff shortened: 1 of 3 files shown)


gentlecat:
Yeah, seems like a good idea since editors aren't supposed to see notices about their own notes. LGTM.

uklauer:
As I wrote on the ticket, I think we can do this and it is an improvement, but it won’t actually fix the main issue. So I’d suggest to merge it, but not to associate it with MBS-8963 and leave that ticket open; otherwise we’ll have to reopen it anyway, or make yet another ticket about slow edit notes.

mwiencek:
Yeah, it still took 20-something seconds after removing the filter, so I figured this wouldn't be the end of MBS-8963. I'll amend the commit message and leave that ticket open.

PR #312, opened 2016-06-04
MBS-7963: Prevent >255-char release comments

mwiencek:
Add a maxlength attribute to the `<input />`, and if something somehow gets through, silently truncate it.


gentlecat:
LGTM

PR #313, opened 2016-06-04
MBS-8951: Make recent notes handle NULL post_time

mwiencek:
I guess old edit notes don't have a post time. Sort these last so that /edit/notes-received behaves correctly.

### DIFF
--- a/admin/sql/updates/20160604-mbs-8951.sql
+++ b/admin/sql/updates/20160604-mbs-8951.sql
@@ -1,0 +1,4 @@
+\set ON_ERROR_STOP 1
+
+DROP INDEX CONCURRENTLY edit_note_idx_post_time_edit;
+CREATE INDEX CONCURRENTLY edit_note_idx_post_time_edit ON edit_note (post_time DESC NULLS LAST, edit DESC);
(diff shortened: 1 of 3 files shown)


admin/sql/updates/20160604-mbs-8951.sql
```diff
@@ -0,0 +1,4 @@
+\set ON_ERROR_STOP 1
+
+DROP INDEX CONCURRENTLY edit_note_idx_post_time_edit;
+CREATE INDEX CONCURRENTLY edit_note_idx_post_time_edit ON edit_note (post_time DESC NULLS LAST, edit DESC);
```
mwiencek: `CONCURRENTLY` can't be used in a transaction block.

gentlecat: Will this update go into the next schema change?

gentlecat:
LGTM

admin/sql/updates/20160604-mbs-8951.sql
uklauer: Yeah, we should perhaps make a ticket to collect these things. There are more changes in #311 that need to be carried over, too, and more index changes might be done before the next schema change (whenever it will be).

mwiencek: Good idea. Created http://tickets.musicbrainz.org/browse/MBS-8973 (feel free to update that as necessary).

PR #315, opened 2016-06-07
MBS-8968: Output gender MBIDs in /ws/2 XML

mwiencek:
Already fixed mmd-schema to recognize this.


gentlecat:
Looks like something is wrong with `t::MusicBrainz::Server::Controller::WS::2::LookupArtist`.

mwiencek:
Ah, I think it's because it's building against an outdated mmd-schema. I'll update that and trigger a new build.

mwiencek:
Jenkins PR thing won't build the right commit, but `prove -vl t/tests.t :: --tests LookupArtist` is passing for me locally.

gentlecat:
:shipit:

PR #316, opened 2016-06-10
MBS-8977: Edit searches with relationship type conditions cause an internal server error

mwiencek:
Tested this locally and the search doesn't ISE anymore.


gentlecat:
LGTM

PR #318, opened 2016-06-13
MBS-8985: Fix CPDL URLs matching and clean it up

y-van-z:
- Fix CPDL score `match` regular expression
- Add CPDL score `clean` function
- Update tests accordingly


mwiencek:
LGTM. Thanks.

uklauer:
In principle, the same applies here as for #320. However, there were only 23 CDPL URLs in total, and I updated them manually.

PR #320, opened 2016-06-13
MBS-7111: Improve URL matching of Musik-Sammler.de

y-van-z:
- Normalize to `https://www.musik-sammler.de/(album|artist|media)/(id|clean-name)/`
- Add corresponding tests


uklauer:
I’m not too happy about this. Immediately before the schema change (state of the slave database on rika), there were 9934 musik-sammler URLs not matching the new normalization; meaning that this will produce around 20k add/remove relationship edits over time and confuse a lot of editors (“I’m sure I didn’t do these edits!”).

Such a strict normalization should only be done if the legacy URLs have been converted previously, either via bot or via direct database update (which didn’t work too well in the Wikidata case, though). If that can’t be done, the cleanup should leave at least things such as http/https and the final slash alone.

mwiencek:
I thought we fixed things to not submit changes for URLs you didn't touch...maybe that was only in the release editor. I personally think we should change that if not.

y-van-z:
I filed the editor issue at http://tickets.musicbrainz.org/browse/MBS-8987

As for MBS-7111, this is not part of development priorities, I don't mind to either revert it or use https://github.com/y-van-z/musicbrainz-server/commit/69a50ced1710086d2361f12190dd747ec57df548. Please let me know the best way to fix it for your workflow.

mwiencek:
Assuming MBS-8987 is fixed, I don't see why we shouldn't normalize newly-added URLs, so I think this is fine as it is. But I'd like to include MBS-8987 in the next release.

PR #322, opened 2016-06-14
MBS-8987: Don't run URL cleanup on existing URLs

mwiencek:
This makes it so that the URL cleanup is only run on seeded URLs, not ones that already exist. If you actually modify an existing URL, it'll trigger a cleanup, but as long as you don't touch it, no mysterious edits will be submitted anymore.


gentlecat:
LGTM. Sorry for delay. Had multiple attempts at trying to figure out what's going on there.

mwiencek:
No problem 👍

PR #323, opened 2016-06-18
MBS-8994: Allow &dismax=true in search queries

mwiencek:
See the ticket for details.

### DIFF
--- a/lib/MusicBrainz/Server/Data/WebService.pm
+++ b/lib/MusicBrainz/Server/Data/WebService.pm
@@ -29,6 +29,7 @@ sub xml_search
     my $dur = 0;
     my $offset = 0;
     my $limit = $args->{limit} || 0;
+    my $dismax = 'false'; # MBS-8994
 
     if (defined $args->{offset} && is_positive_integer($args->{offset}))
     {
@@ -46,6 +47,10 @@ sub xml_search
         }
 
         $query = $args->{query};
+
+        if (defined $args->{dismax} && $args->{dismax} eq 'true') {
+            $dismax = 'true';
+        }
     }
     elsif ($resource eq 'artist')
     {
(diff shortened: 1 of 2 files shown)


gentlecat:
What's dismax and what does it change? A bit unclear to me from the description in that ticket.

gentlecat:
LGTM

mwiencek:
> What's dismax and what does it change? A bit unclear to me from the description in that ticket.

My understanding comes from https://cwiki.apache.org/confluence/display/solr/The+DisMax+Query+Parser which says:

> The DisMax query parser is designed to process simple phrases (without complex syntax) entered by users and to search for individual terms across several fields using different weighting (boosts) based on the significance of each field.

You can see that in the code for our search server, the normal artist search (with dismax=false) searches the artist name, aliases, and sort name fields, presumably with equal weightings: https://bitbucket.org/metabrainz/search-server/src/2eb1597bfbf1c9d1dfc52e69988d78e29dc8976f/servlet/src/main/java/org/musicbrainz/search/servlet/ArtistSearch.java?at=master&fileviewer=file-view-default#ArtistSearch.java-24

With dismax=true, it also searches the artist name with accents included (in addition to the above fields), but with different weightings for each field: https://bitbucket.org/metabrainz/search-server/src/2eb1597bfbf1c9d1dfc52e69988d78e29dc8976f/servlet/src/main/java/org/musicbrainz/search/servlet/ArtistDismaxSearch.java?at=master&fileviewer=file-view-default#ArtistDismaxSearch.java-13

The trade-off, as I understand it, is that you can only search for single phrases with dismax=true, not complex expressions with different field terms, boolean logic, etc.

gentlecat:
And weights are specified in a second argument to `DismaxAlias.AliasField`, right?

mwiencek:
That's what I gather from https://bitbucket.org/metabrainz/search-server/src/2eb1597bfbf1c9d1dfc52e69988d78e29dc8976f/servlet/src/main/java/org/musicbrainz/search/servlet/DismaxAlias.java?fileviewer=file-view-default#DismaxAlias.java-63 (although it's called "boost" there).

PR #324, opened 2016-06-19
MBS-8996: Display merge append errors correctly

mwiencek:
`[% form.errors %]` can be an array if there are multiple errors. The only errors that should be possible to trigger by direct user input are for the medium positions, so just display the errors for those fields inline.

### DIFF
--- a/root/release/merge.tt
+++ b/root/release/merge.tt
@@ -53,8 +53,6 @@
         [% USE r = FormRenderer(form) %]
         [% form_row_select(r, 'merge_strategy', l('Merge strategy:')) %]
 
-        [% form.errors %]
-
         <div id="merge-strategy-1" class="merge-strategy">
           <p>
             [% l('Using this merge strategy, all mediums from all releases will be used. You may specify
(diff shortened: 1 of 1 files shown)


root/release/merge.tt
```diff
@@ -63,12 +61,18 @@
...
+                th span.error { margin: 0 12px 0 6px; }
+              </style>
               [% FOR medium=mediums %]
               [% field = form.field('medium_positions').field('map').field(loop.index) %]
               <tr class="subh">
                 <th colspan="4">
                   <label>[% l('New position:') %]</label>
                   [% r.text(field.field('position'), size=2) %]
+                  [%~ IF field.field('position').has_errors;
+                        '<span class="error">' _ html_escape(field.field('position').errors.0) _ '</span>';
```
gentlecat: `.errors.0` is a description?

mwiencek: Yeah, `errors` is an array of error strings for the field: http://search.cpan.org/dist/HTML-FormHandler-0.40058/lib/HTML/FormHandler/Manual/Errors.pod

In TT syntax `.errors.0` means `->errors->[0]`, so this only displays the first error if there are multiple. But that should only happen if they enter a 0 or negative position multiple times...

mwiencek: Here's where those error strings come from: https://github.com/metabrainz/musicbrainz-server/blob/2835d40ce94514d1569e50e2abf3ec6764784c40/lib/MusicBrainz/Server/Form/Merge/Release.pm#L62-L62

gentlecat:
:shipit:

PR #326, opened 2016-06-19
MBS-8999: Track lengths are in the wrong table column in the release editor

mwiencek:
This broke when we released the new AC editor stuff, because now that's all in its own table (and the table is embedded in a single cell). Previously, the release editor had the open-ac button in a separate column.


gentlecat:
:shipit:

PR #327, opened 2016-06-20
MBS-9000: Escape and bidi-isolate entity aliases

uklauer:
An alias was put verbatim into the aliases page, even when it contained special characters. It was a bit better for the sort name, which was escaped, but `<bdi>` tags were missing on both.


uklauer:
The branch is based on `production` because this is a XSS vulnerability: Registered users can create special aliases containing JS code that is then displayed when someone views an `/aliases` page.

PR #328, opened 2016-06-21
MBS-8806: Don’t cache non-existent ACs as empty

uklauer:
When `Data::ArtistCredit::get_by_ids` was called with IDs that didn’t actually exist, it returned empty `ArtistCredit` entities for them. Those were then automatically put into the cache and persisted there. When an artist credit with the relevant ID later actually was created, the empty artist credit was displayed in its place until the old cache entry expired (MBS-8806). It is unclear how `get_by_ids` could be called with a not-yet-existing ID.

Only create and return an `ArtistCredit` entity for IDs actually found in the database. Also emit a stack trace when a non-existent ID is encountered in order to aid further debugging.

### DIFF
--- a/lib/MusicBrainz/Server/Data/ArtistCredit.pm
+++ b/lib/MusicBrainz/Server/Data/ArtistCredit.pm
@@ -3,6 +3,8 @@ use Moose;
 use namespace::autoclean -also => [qw( _clean )];
 
 use Data::Compare;
+use Carp qw( cluck );
+
 use MusicBrainz::Server::Entity::Artist;
 use MusicBrainz::Server::Entity::ArtistCredit;
 use MusicBrainz::Server::Entity::ArtistCreditName;
@@ -25,15 +27,14 @@ sub get_by_ids
                 "JOIN artist ON artist.id=artist_credit_name.artist " .
                 "WHERE artist_credit IN (" . placeholders(@ids) . ") " .
                 "ORDER BY artist_credit, position";
+
     my %result;
     my %counts;
-    foreach my $id (@ids) {
-        my $obj = MusicBrainz::Server::Entity::ArtistCredit->new(id => $id);
-        $result{$id} = $obj;
-        $counts{$id} = 0;
-    }
     for my $row (@{ $self->sql->select_list_of_hashes($query, @ids) }) {
         my $id = $row->{artist_credit};
+        $counts{$id} //= 0;
+        $result{$id} //= MusicBrainz::Server::Entity::ArtistCredit->new(id => $id);
+
         my $acn = MusicBrainz::Server::Entity::ArtistCreditName->new(
                       artist_id => $row->{artist},
                       artist => $self->c->model('Artist')->_new_from_row($row),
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -25,26 +27,31 @@ sub get_by_ids
...
-        $counts{$id} = 0;
-    }
     for my $row (@{ $self->sql->select_list_of_hashes($query, @ids) }) {
-        my %info = (
-            artist_id => $row->{artist},
-            name => $row->{ac_name}
-        );
-        $info{join_phrase} = $row->{join_phrase} // '';
-        my $obj = MusicBrainz::Server::Entity::ArtistCreditName->new(%info);
-        $obj->artist($self->c->model('Artist')->_new_from_row($row));
```
uklauer: Unrelated, but I didn’t like this part with the generic `$obj` name and the three different ways to set properties of the new ACN entity, so I rewrote it. (It’s a separate commit if you want to look at the two changes separately.)

lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -25,26 +27,31 @@ sub get_by_ids
                 "JOIN artist ON artist.id=artist_credit_name.artist " .
                 "WHERE artist_credit IN (" . placeholders(@ids) . ") " .
                 "ORDER BY artist_credit, position";
+
     my %result;
     my %counts;
-    foreach my $id (@ids) {
-        my $obj = MusicBrainz::Server::Entity::ArtistCredit->new(id => $id);
-        $result{$id} = $obj;
```
uklauer: This was the problematic part: For every input ID, an output entity was created that remained empty if the database didn’t return anything about it. The `around` modifier in `Role::EntityCache` then cached these.

mwiencek:
Great, thanks 👍

lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -25,26 +27,31 @@ sub get_by_ids
...
-            artist_id => $row->{artist},
-            name => $row->{ac_name}
-        );
-        $info{join_phrase} = $row->{join_phrase} // '';
-        my $obj = MusicBrainz::Server::Entity::ArtistCreditName->new(%info);
-        $obj->artist($self->c->model('Artist')->_new_from_row($row));
         my $id = $row->{artist_credit};
-        $result{$id}->add_name($obj);
+        $counts{$id} //= 0;
+        $result{$id} //= MusicBrainz::Server::Entity::ArtistCredit->new(id => $id);
```
uklauer: This is now inside the loop processing the database results, so the object is only created if the AC actually exists in the database.

lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -25,26 +27,31 @@ sub get_by_ids
...
+        $result{$id}->add_name($acn);
         $counts{$id} += 1;
     }
     foreach my $id (@ids) {
+        if (!defined $counts{$id}) {
+            # It's unclear how this could happen, but it seems to be the
+            # cause of MBS-8806. Perhaps we'll find out with the call trace
+            # provided by cluck.
+            cluck "Attempted to load non-existent AC $id, please investigate";
+            next;
```
uklauer: `cluck` will not die, it only outputs a trace, so `next` explicitly skips to the next iteration.

I really have no idea how it is possible that the method is called with a soon-to-be-but-not-yet ID. It somehow must leak from the transaction creating a new AC, before it commits, but I don’t see how.

PR #330, opened 2016-06-22
Edit artist icon: thicken outline

caroline-gschwend:
re: https://community.metabrainz.org/t/new-artist-icon-is-hard-to-read/56369 and several other comments

I thickened the outline of the artist icon so it will be easier to read


mwiencek:
Yeah, this looks better. Thanks 👍

PR #331, opened 2016-06-27
Correctly validate Cover Art Archive IDs

uklauer:
IDs in the Cover Art Archive are `BIGINT`s, but they were only accepted as valid if they were in the range of an `INT`. This caused cover art not to be loaded in the normal way for edits, and also caused spurious warnings by `get_by_any_id(s)`.

Add a new `is_valid_id` method to `Data::Entity` and override it for `Data::Artwork` to allow `BIGINT` values.


mwiencek:
:shipit:

PR #332, opened 2016-06-28
MBS-8451: Turn WP/WD URLs into https on cleanup

uklauer:
All Wikimedia projects are https-only now; http requests will be redirected. In light of this and the fact that the external links editor will no longer run cleanup on unchanged pre-existing URLs (MBS-8987), make the URL cleanup code change Wikipedia and Wikidata URLs to https. Also link to Wikimedia Commons using https.

Includes some related updates in tests and documentation.


mwiencek:
Awesome, thanks 👍

PR #334, opened 2016-06-30
Check for Postgresql 9.5 instead of 9.1

SpootDev:
Verify 9.5 instead of 9.1 version.

Sorry, didn't "test" this as my current install is pooched.  And found this while trying to unpooch it.  :)

PR #335, opened 2016-06-30
Postgres version fixes

uklauer:
As it happens, I had recently also stumbled over the outdated Postgres version check now fixed in #334. Here are some related improvements: A typo in the error message, using the machine-readable Pg version instead of parsing the human-readable version (which is guaranteed to keep its format, see https://www.postgresql.org/message-id/4422.1465937807%40sss.pgh.pa.us), and removing an unused version comparison function.


mwiencek:
:shipit:

PR #336, opened 2016-07-03
MBS-3966

Zastai:
Implementation turned out to be very trivial; was just a matter of finding my way in the templates and seeing how the other relationships did it.


mwiencek:
LGTM.

PR #337, opened 2016-07-04
MBS-8961: Relationship table mixes "guest"/"additional" and instrument attributes.

Zastai:
The relationship overview for an artist now avoids rendering attributes as "additional, drums, guest and percussion".
Instrument and vocal attributes are now grouped separately in a comma-separated list at the end, with other attributes space-separated in front of that list; so the case above becomes "additional guest drums and percussion".
If only non-instrument/vocal attributes are present, they still get placed in a comma-separated list (so a producer credit would have "additional, co and executive").

### DIFF
--- a/root/components/relationships-table.tt
+++ b/root/components/relationships-table.tt
@@ -31,11 +31,24 @@
                     [%~ END ~%]
                     <td>
                       [%~ IF rel.link.attributes.size ~%]
-                      [%~ attr_list = [];
+                      [%~ attr_list1 = []; attr_list2 = [];
+                          instrument_root = '0abd7f04-5e28-425b-956f-94789d9bcbe2';
+                          vocals_root     = 'd92884b7-ee0c-46d5-96f3-918196ba8c5b';
                           FOR attr=rel.link.attributes;
-                            attr_list.push(attr.html);
+                            IF attr.type.root.gid == instrument_root OR attr.type.root.gid == vocals_root;
+                              attr_list2.push(attr.html);
+                            ELSE;
+                              attr_list1.push(attr.html);
+                            END;
                           END;
-                          comma_list(attr_list);
+                          IF attr_list1.size AND attr_list2.size;
+                            FOR text = attr_list1; text; ' '; END;
+                            comma_list(attr_list2);
+                          ELSIF attr_list1.size;
+                            comma_list(attr_list1);
+                          ELSIF attr_list2.size;
+                            comma_list(attr_list2);
+                          END;
                        ~%]
                       [%~ END ~%]
                     </td>
--- a/root/components/relationships-table.tt
+++ b/root/components/relationships-table.tt
@@ -33,8 +33,7 @@
                       [%~ IF rel.link.attributes.size ~%]
                       [%~ attr_list1 = []; attr_list2 = [];
                           FOR attr=rel.link.attributes;
-                            IF attr.type.root.id == server_constants.INSTRUMENT_ROOT_ID
-                            OR attr.type.root.id == server_constants.VOCAL_ROOT_ID;
+                            IF attr.type.root.id == 14 OR attr.type.root.id == 3;
                               attr_list2.push(attr.html);
                             ELSE;
                               attr_list1.push(attr.html);
--- a/root/components/relationships-table.tt
+++ b/root/components/relationships-table.tt
@@ -41,14 +41,8 @@
                               attr_list1.push(attr.html);
                             END;
                           END;
-                          IF attr_list1.size AND attr_list2.size;
-                            FOR text = attr_list1; text; ' '; END;
-                            comma_list(attr_list2);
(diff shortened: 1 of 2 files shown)


gentlecat:
@brainzbot, add to whitelist.

Zastai:
Ultimately, this is only a small tweak; I would say it improves things, but I wouldn't consider it a full fix for MBS-8961.

I could look into the bigger rework described in the message for commit 2c1cddd, but since that basically means changing the table's axis from relationship type to relationship source entity type, that probably needs discussion beforehand.

mwiencek:
Cool, seems like a reasonable fix for now.

uklauer:
I’m not sure if it’s actually much of an improvement (but it won’t make things worse, at least): “Co, additional, drums, trumpet and guitar” is still nothing that a human would write.

It seems more natural to me to separate the “normal” attributes and the “instrument” attributes, like one of these:
> co and additional: drums, trumpet and guitar
> drums, trumpet and guitar – co and additional
With a translatable string, the separator could be replaced in a translation, and for French (and similar languages) the order could be inverted, too.

root/components/relationships-table.tt
```diff
@@ -31,10 +31,17 @@
                     [%~ END ~%]
                     <td>
                       [%~ IF rel.link.attributes.size ~%]
-                      [%~ attr_list = [];
+                      [%~ attr_list1 = []; attr_list2 = [];
+                          instrument_root = '0abd7f04-5e28-425b-956f-94789d9bcbe2';
+                          vocals_root     = 'd92884b7-ee0c-46d5-96f3-918196ba8c5b';
```
uklauer: I saw the IRC discussion about this, but since we have `$INSTRUMENT_ROOT_ID` and `$VOCAL_ROOT_ID` constants already that are used all over the place, we should use them here, too. This is internal to the server; stability of the MBIDs doesn’t matter here.

Zastai: I have no issue with changing these to use the other constants; I'd just have to look at how to get at them. Will update this PR when that's done.

As for the other suggestion, my initial version had the colon-separated form, but this immediately drew a remark (from reosarevok I think) on IRC that it looked weird. As for making it translatable, again I'd just have to know how to do that from the template (it looks like those l() calls are enough, so easy enough).

uklauer:
As it is now, it will break the other consumers of `components/relationships-table`.

Zastai:
Do entities other than artist have instrument/vocal attributes? Am not seeing any breakages at the moment for the 3 others (label, place, work); the constants gracefully degrade to undef, so the instrument/vocal test never matches, so you just get a single attribute list just like before.

Zastai:
This is what I'm looking at now: http://imgur.com/3nNYQSc
While using more text, somewhat duplicating the section headers, it's much more readable.
The main issues are the "inverted" relationships (e.g. "contains samples by", or most place relationships like "was mixed at"), where the text expects the page's entity on the right rather than the left. There the verbose phrase has {entity0/1} to help; if I knew how to get the _localized_ source/target entity type name in there, that would be slightly better (e.g. "recording contains samples by artist"). Ideal would be to use the reverse verbose phrase for those cases (so "was the mixing location for"), but that doesn't seem to exist (anymore).

mwiencek:
I don't think it'll break anything currently, but I'd prefer just hardcoding 3 + 14 in the template. Current patch looks good to me (though it still doesn't exactly read like something a human would write, I think it's better).

About the long link phrases...yeah, you have ones where entity1 appears first in the phrase, but even ignoring those, the source entity position isn't consistent because entity0/entity1 are ordered alphabetically by entity type. So I'm not sure how you'd display those in a good way while still keeping the columns consistent (and hopefully sortable in the future). I like the idea though...

Zastai:
As requested, rolled back the constants-via-stash approach, and switched to hardcoded id values.

Still looking at alternatives though. [This version](http://imgur.com/EiYJDOi) drops the table as such, instead having one paragraph per relationship key (so preserving the grouping), and using the full verbose relationship string (but setting it up so that the page's main entity doesn't show its disambig).
This provides the "nicest" string, at the expense of the table form (and possible near-future sortability).

mwiencek:
Looks like you might've rebased incorrectly, since there's an unrelated commit here.

I'm not sure about dropping the table...if we went with that, it'd probably need a way to filter things since it's harder to scan vertically. On the plus side, it's easier to read individual lines (at least for me). Might be worth asking in \#musicbrainz to get some more editor feedback.

Zastai:
Yeah, I'm still new to git. Will see about removing that commit.

I think the nicest result would be obtained by having the long form text, but there is currently only one such text, based on either the forward or reverse text. That makes it less consistent. I will look at making a ticket for adding a long form for both forward and reverse forms of each relationship.

mwiencek:
We're going to close this since it's pretty old and unlikely to be updated, but it can be reopened again if new development happens.

PR #338, opened 2016-07-07
MBS-7903: Only show a matching, not any primary alias

uklauer:
When considering primary aliases for display in the inline search, aliases in the interface language or in English are preferred. However, if none of these exist, any random language was chosen. (E.g., a German user might have been presented with the Japanese or Arabic alias, which they are unlikely to be able to read.) Stop using such random fallback aliases.

In cases where a (matching) alias exists, it is no longer necessary to display the sort name, whose only purpose is to provide a rudimental Latin-script version (cf. MBS-7404).


mwiencek:
LGTM.

PR #339, opened 2016-07-18
MBS-8148: Identify ResidentAdvisor reviews as such

y-van-z:
Guess type of RA links, either `otherdatabases` or `review`, rather than `otherdatabases` only.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -480,7 +480,8 @@ const CLEANUPS = {
   review: {
     match: [
       new RegExp("^(https?://)?(www\\.)?bbc\\.co\\.uk/music/reviews/", "i"),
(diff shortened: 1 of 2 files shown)


uklauer:
Looks good at a quick glance. (This comment does *not* extend to ResidentAdvisor’s URL structure!) I would however prefer the commit message to describe what the change does, not what the changed code used to do; i.e. not “ResidentAdvisor reviews get selected as Other Database” but (e.g.) “Don’t identify ResidentAdvisor review URLs as other database” or even better “Identify ResidentAdvisor reviews as such” or similar. As it is, it sounds like the intention of the change is to mark them _other database_, i.e. exactly backwards.

I would also personally prefer to have the tests added in a separate commit, but that’s a matter of taste.

y-van-z:
Yes, that was because I just copied the title of the issue.
The title of the pull request is now fixed and tests moved apart.

mwiencek:
Thanks!

PR #340, opened 2016-07-18
Docker

mwiencek:


docker/config.all.m4
```diff
@@ -0,0 +1,2 @@
+include(`config.default.m4')dnl
+sinclude(`config.m4')dnl
```
mwiencek: https://www.gnu.org/software/m4/manual/m4-1.4.14/html_node/Include.html
https://www.gnu.org/software/m4/manual/m4-1.4.14/html_node/Dnl.html

docker/musicbrainz-nginx/sites-available/musicbrainz.m4
```diff
@@ -0,0 +1,101 @@
+include(`config.all.m4')dnl
+resolver NGINX_RESOLVER;
```
mwiencek: m4 will replace this with the definition from docker/config.all.m4. If you don't want something to be interpreted as a macro, you have to quote it like `` `NGINX_RESOLVER'``. If you need to use the start quote character in the file (a backtick by default), you need to use `changequote` to change it temporarily. The end quote character can appear without trouble as long as the start quote character doesn't precede it.

docker/musicbrainz-server-base/DBDefs.pm.m4
```diff
@@ -0,0 +1,128 @@
+include(`config.all.m4')dnl
+define(`_boolean_def', `subroutine'(``$1'', `ifelse'($1, ``1'', ``1'', ``0'')))dnl
+define(`_code_def', `subroutine'(``$1'', $1))dnl
+define(`_string_def', `subroutine'(``$1'', `quoted_string'(``$1'')))dnl
+define(`_wrap_def', `ifdef'(``$1'', ``$2''(```$1'''), ``dnl''))dnl
```
mwiencek: Quotes are kind of hard to explain, but they're the way they are because m4 has no escape character, so you can't just escape quotes. https://www.gnu.org/software/m4/manual/m4-1.4.15/html_node/Quoted-strings.html#Quoted-strings

The quotes always have to be balanced, and the actual value has "one level of quotes stripped off". So `` `broken`macro'name `` has the value `brokenmacroname` and ``` ``broken``macro''name'' ``` has the value `` `broken`macro'name' ``.

The quoting rule of thumb mentioned at https://www.gnu.org/software/m4/manual/m4-1.4.15/html_node/Quoting-Arguments.html is useful for making sense of things.

You don't have to quote parentheses in macro definitions as long as they're balanced. Otherwise you might accidentally close the `define(`.

mwiencek: Also, the only thing quoting is for is stopping macro expansion.

docker/musicbrainz-server-base/DBDefs.pm.m4
```diff
@@ -0,0 +1,128 @@
...
+define(`boolean_def', `_wrap_def'(``$1'', ``_boolean_def''))dnl
+define(`code_def', `_wrap_def'(``$1'', ``_code_def''))dnl
+define(`string_def', `_wrap_def'(``$1'', ``_string_def''))dnl
+define(`quoted_string', `changequote'(<VERY UNSAFE TEXT!>)`format'('%s', `patsubst'($1, ['\\], \\\&))`changequote')dnl
```
mwiencek: Consider that a DBDefs configuration value from config.all.m4 might contain a `` ` `` or `'` which could be dangerous when substituted into the file...to work around that, we temporarily change the start quote to `<VERY UNSAFE TEXT!>`.

mwiencek:
The major downside of m4 is its lack of an escape character, which is also what makes it difficult to learn and use safely. It's fun once you know what you're doing, but is not easiest tool for this particular job.

docker/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
+version: '2'
+services:
```
gentlecat: Nitpicking... might be easier to read if you split each service by an empty line.

docker/musicbrainz-server-base/Dockerfile
```diff
@@ -0,0 +1,149 @@
...
+        libxml-parser-perl \
+        libxml-rss-parser-lite-perl \
+        libxml-semanticdiff-perl \
+        libxml-simple-perl \
+        libxml-xpath-perl \
+        libxml2-dev \
+        nodejs \
+        # needed by ./admin/psql. trusty only has 9.3 in the official repos,
+        # but it works fine against a 9.5 server.
+        postgresql-client-9.3 && \
```
gentlecat: I've been adding PG's repo, but I guess this should be fine if works. https://github.com/metabrainz/critiquebrainz/blob/beta/Dockerfile#L6

docker/musicbrainz-server-base/Dockerfile
```diff
@@ -0,0 +1,149 @@
...
+COPY package.json ./
+
+RUN rm -rf node_modules && \
+    npm install --only=production
+
+COPY ./ ./
+COPY docker/local-configs/musicbrainz-server-base/DBDefs.pm lib/
+
+# So we don't need keys if the base repository was using ssh.
+RUN git remote set-url origin https://github.com/mwiencek/musicbrainz-server.git
```
gentlecat: `r/mwiencek/metabrainz`?

docker/musicbrainz-template-renderer/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM musicbrainz-server-base
+MAINTAINER Michael Wiencek <bitmap@metabrainz.org>
+
+ENTRYPOINT ["./script/start_renderer.pl"]
```
gentlecat: This is the new one with React, right?

docker/musicbrainz-server-base/Dockerfile
mwiencek: I used the PG repo in the postgres Dockerfile. But got lazy about copying it here since the 9.3 client worked. This actually might be good to factor out into an m4 file.

docker/musicbrainz-server-base/Dockerfile
mwiencek: Whoops. That was needed before because the docker branch lived there.

docker/musicbrainz-template-renderer/Dockerfile
mwiencek: Correct.

docker/musicbrainz-server-base/Dockerfile
mwiencek: ...or at least I was before. I guess I switched to using an official `postgres:9.5.3` image.

gentlecat:
What would be the difference between live and standby servers?

mwiencek:
> What would be the difference between live and standby servers?

Good question...only the postgres configuration and data volumes should be different. So we can just build different images from the same Dockerfile, I think.

gentlecat:
`make -C docker` doesn't seem to work when there's no `origin` remote:
```
Step 17 : RUN git remote set-url origin https://github.com/metabrainz/musicbrainz-server.git
 ---> Running in 7b707bd98ff6
fatal: No such remote 'origin'
ERROR: Service 'musicbrainz-template-renderer' failed to build: The command '/bin/sh -c git remote set-url origin https://github.com/metabrainz/musicbrainz-server.git' returned a non-zero code: 128
```

gentlecat:
Also, when it starts up there's some issue with template renderer:
```
musicbrainz-template-renderer_1  | module.js:442
musicbrainz-template-renderer_1  |     throw err;
musicbrainz-template-renderer_1  |     ^
musicbrainz-template-renderer_1  |
musicbrainz-template-renderer_1  | Error: Cannot find module '/Users/roman/workspace/musicbrainz-server/root/server.js'
musicbrainz-template-renderer_1  |     at Function.Module._resolveFilename (module.js:440:15)
musicbrainz-template-renderer_1  |     at Function.Module._load (module.js:388:25)
musicbrainz-template-renderer_1  |     at Module.runMain (module.js:575:10)
musicbrainz-template-renderer_1  |     at run (bootstrap_node.js:352:7)
musicbrainz-template-renderer_1  |     at startup (bootstrap_node.js:144:9)
musicbrainz-template-renderer_1  |     at bootstrap_node.js:467:3
```

mwiencek:
Thanks for testing again. :) I removed the `git remote set-url origin` command since it's not really needed. For the template renderer issue, my guess is that because the dev compose file uses a data volume for the code directory, you have to `rm -rf node_modules/; npm install; ./script/compile_resources.sh` on your local machine before starting the container, because something is missing there. (To start the containers again without rebuilding, do `cd docker; docker-compose -f docker-compose.dev.yml -p musicbrainz-server up`.)

At the moment I'm still trying to get email sending with postfix to work.

mwiencek:
Another problem with the code directory being a data volume is that you have to copy the generated DBDefs file like `cp docker/local-configs/musicbrainz-server-base/lib/DBDefs.pm lib/DBDefs.pm` (better backup the previous one too) if they're different...

gentlecat:
Oh, I didn't realise that node modules need to be installed locally before running Docker containers. Though, I still get that the same error even after reinstalling all the modules and compiling resources. Maybe I'm missing something.

```
~/w/musicbrainz-server> ./script/compile_resources.sh
(node:16959) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
(node:16959) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
[11:43:35] Using gulpfile ~/workspace/musicbrainz-server/gulpfile.js
[11:43:35] Starting 'styles'...
[11:43:35] Finished 'styles' after 98 ms
[11:43:35] Starting 'scripts'...
[11:43:47] Finished 'scripts' after 12 s
[11:43:47] Starting 'default'...
[11:43:47] Finished 'default' after 19 μs
~/w/musicbrainz-server> cd docker; docker-compose -f docker-compose.dev.yml -p musicbrainz-server up
Starting musicbrainzserver_postgres-master_1
Starting musicbrainzserver_redis_1
Starting musicbrainzserver_musicbrainz-template-renderer_1
Starting musicbrainzserver_musicbrainz-server_1
Attaching to musicbrainzserver_redis_1, musicbrainzserver_postgres-master_1, musicbrainzserver_musicbrainz-template-renderer_1, musicbrainzserver_musicbrainz-server_1
redis_1                          |                 _._
redis_1                          |            _.-``__ ''-._
postgres-master_1                | 2016-08-10 09:44:10.871 GMT @ 13  LOG:  database system was shut down at 2016-08-10 09:42:36 GMT
redis_1                          |       _.-``    `.  `_.  ''-._           Redis 3.2.1 (00000000/0) 64 bit
redis_1                          |   .-`` .-```.  ```\/    _.,_ ''-._
redis_1                          |  (    '      ,       .-`  | `,    )     Running in standalone mode
redis_1                          |  |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
postgres-master_1                | 2016-08-10 09:44:10.876 GMT @ 13  LOG:  MultiXact member wraparound protections are now enabled
redis_1                          |  |    `-._   `._    /     _.-'    |     PID: 1
redis_1                          |   `-._    `-._  `-./  _.-'    _.-'
redis_1                          |  |`-._`-._    `-.__.-'    _.-'_.-'|
redis_1                          |  |    `-._`-._        _.-'_.-'    |           http://redis.io
redis_1                          |   `-._    `-._`-.__.-'_.-'    _.-'
redis_1                          |  |`-._`-._    `-.__.-'    _.-'_.-'|
redis_1                          |  |    `-._`-._        _.-'_.-'    |
redis_1                          |   `-._    `-._`-.__.-'_.-'    _.-'
redis_1                          |       `-._    `-.__.-'    _.-'
redis_1                          |           `-._        _.-'
redis_1                          |               `-.__.-'
redis_1                          |
redis_1                          | 1:M 10 Aug 09:44:10.758 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
redis_1                          | 1:M 10 Aug 09:44:10.759 # Server started, Redis version 3.2.1
redis_1                          | 1:M 10 Aug 09:44:10.759 # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
redis_1                          | 1:M 10 Aug 09:44:10.759 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
redis_1                          | 1:M 10 Aug 09:44:10.759 * The server is now ready to accept connections on port 6379
postgres-master_1                | 2016-08-10 09:44:10.880 GMT @ 1  LOG:  database system is ready to accept connections
postgres-master_1                | 2016-08-10 09:44:10.881 GMT @ 17  LOG:  autovacuum launcher started
musicbrainz-server_1             | Watching ./lib app.psgi for file updates.
musicbrainz-template-renderer_1  | module.js:442
musicbrainz-template-renderer_1  |     throw err;
musicbrainz-template-renderer_1  |     ^
musicbrainz-template-renderer_1  |
musicbrainz-template-renderer_1  | Error: Cannot find module '/Users/roman/workspace/musicbrainz-server/root/server.js'
musicbrainz-template-renderer_1  |     at Function.Module._resolveFilename (module.js:440:15)
musicbrainz-template-renderer_1  |     at Function.Module._load (module.js:388:25)
musicbrainz-template-renderer_1  |     at Module.runMain (module.js:575:10)
musicbrainz-template-renderer_1  |     at run (bootstrap_node.js:352:7)
musicbrainz-template-renderer_1  |     at startup (bootstrap_node.js:144:9)
musicbrainz-template-renderer_1  |     at bootstrap_node.js:467:3
musicbrainzserver_musicbrainz-template-renderer_1 exited with code 0
```

gentlecat:
> At the moment I'm still trying to get email sending with postfix to work.

Let me know if you manage to figure out a good way to set it up. I'm still stuck on that in https://github.com/metabrainz/metabrainz.org/pull/229.

mwiencek:
Ohh, yes, you found a bug, same one as in https://github.com/metabrainz/musicbrainz-server/pull/340#issuecomment-238771174 except for the template renderer rather than plackup. Just need to tweak the perl library paths so it loads docker/local-configs/musicbrainz-server-base/lib/DBDefs.pm instead of lib/DBDefs.pm.

I'm guessing you have `MB_SERVER_ROOT` explicitly set to `/Users/roman/workspace/musicbrainz-server`, which isn't a path inside the container. If that's defined in lib/DBDefs.pm on your local machine, you can actually just delete it, since the one from lib/DBDefs/Default.pm should work both locally and in the container: https://github.com/metabrainz/musicbrainz-server/blob/25fd6271c1be3bfe2ddec5fa33634834fb3b9099/lib/DBDefs/Default.pm#L56

I didn't see that bug here since I don't have `MB_SERVER_ROOT` defined like that.

mwiencek:
@gentlecat I have a functioning postfix container in https://github.com/mwiencek/musicbrainz-server/tree/docker/docker/postfix that you can try to copy. Let me know if you have trouble getting it working.

gentlecat:
Now it's this:
```
~/w/musicbrainz-server> make -C docker
./config.sh
./config.sh: line 48: realpath: command not found
./config.sh: line 49: realpath: command not found
Warning: backing up  to .1471090101
cp: fts_open: No such file or directory
Warning: copying  to
cp: fts_open: No such file or directory
make: *** [config] Error 1
```

gentlecat:
Ah, `brew install coreutils` seems to have helped with that.

gentlecat:
OK, seems to be working now. Now I just need to figure out how to set up the database.

mwiencek:
Whoops, I thought `realpath` came with OSX. Need to document that then. It was the easiest way I found to get an absolute path on both OSX and Linux, since their `readlink` binaries don't have compatible interfaces.

I'll write a helper script for dropping the initial DB and importing data dumps.

mwiencek:
We should only need one SMTP server for all of MeB. So I'm going to remove the postfix container from here and put it in a standalone repo. These are very difficult to setup properly...if we could use our Google Apps account's SMTP server in production I'd much prefer that, but there's a hard limit to the number of emails you can send with it per day (something like 10,000).

mwiencek:
This should be ready now, I hope.

mwiencek:
There may still be bugs, but this must be merged now.

PR #341, opened 2016-07-27
MBS-9031: Replace all usages of memcached with redis

mwiencek:
I didn't see any point in using `Cache::Redis`, because we'd have to write a wrapper around it anyway (it handles serialization on its own, doesn't have `delete`/`*_multi` methods, etc.), and `Cache::Redis` itself is already [a very small wrapper](http://cpansearch.perl.org/src/SONGMU/Cache-Redis-0.02/lib/Cache/Redis.pm) around the `Redis` module, so instead of writing a wrapper around a wrapper, we can just write a wrapper. But if we were to use `Cache::Redis`, we'd have to add a flag to `MusicBrainz::Server::CacheWrapper` to disable serialization, and add expiration time support to `set_multi`. However, another advantage of having a new wrapper is that we can use Redis-specific features like `mget` in `get_multi` and command pipelining in `set_multi`, which the generic wrapper can't do without breaking the in-memory cache.

Sessions and cached entities are now stored in the same server. Because sessions are in there, we can't enable any new memory limit or LRU eviction, i.e. how memcached operated. So, I've changed `ENTITY_CACHE_TTL` from its default of 0 (no expiration on keys) on standalone servers to a new default of 86400 (1 day).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -49,7 +49,7 @@ Prerequisites
 
         sudo apt-get install git-core
 
-6.  Redis
+5.  Redis
 
     Sessions and cached entities are stored in Redis, so a running Redis server
     is required. Redis can be installed with the following command and will not
@@ -61,7 +61,7 @@ Prerequisites
     in lib/DBDefs.pm.  The defaults should be fine if you don't use
     your redis install for anything else.
 
-7.  Node.js
+6.  Node.js
 
     Node.js is required to build (and optionally minify) our JavaScript and CSS.
     If you plan on accessing musicbrainz-server inside a web browser, you should
@@ -76,7 +76,7 @@ Prerequisites
     This is only needed where it exists, so a warning about the package not being
     found is not a problem.
 
-8.  Standard Development Tools
+7.  Standard Development Tools
 
     In order to install some of the required Perl and Postgresql modules, you'll
     need a C compiler and make. You can install a basic set of development tools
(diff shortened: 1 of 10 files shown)


INSTALL.md
```diff
@@ -49,21 +49,11 @@ Prerequisites
...
-
-    By default the MusicBrainz server requires a Memcached server running on the
-    same server with default settings. To install Memcached, run the following:
-
-        sudo apt-get install memcached
-
-    You can change the memcached server name and port, or configure other datastores
-    in lib/DBDefs.pm.
-
 6.  Redis
```
mineo: This should be renumbered to 5 (and the following entries as well). For some reason, GitHubs rendered versions number these correctly, but https://raw.githubusercontent.com/mwiencek/musicbrainz-server/redis/INSTALL.md looks incorrect.

uklauer: Yes, the Markdown definition of a numbered list is essentially `<any digits>. `. So `1. foo 1. bar 1. baz` is mostly equivalent to HTML `<ol><li>foo</li> <li>bar</li> <li>baz</li>`, generating its own numbering.

Still, the point of MD is to be also readable without rendering, so I agree it should be fixed. :-)

mwiencek: Fixed 👍

mwiencek:
Not sure how we'll test this, since beta and prod must share the same cache.

mwiencek:
Since there's not a great way to test this as part of our normal release process, and I'd also like to base my docker work on top of it, I'm going to close this pull request and do that (rebase my docker branch on top of this). That way, it'll at least get tested as we're deploying the new setup.

PR #342, opened 2016-07-27
Optimize PNGs in root/static/images

scop:
zopflipng -m followed by --lossy_transparent -m, version 1.0.1.

Before: 262010 bytes
After:  202251 bytes


uklauer:
Hmm … Although I see that better compression of the images may be a useful goal, I don’t think this is the right approach. Next week, someone will update one of the favicons, and we’ll forget to re-optimize the result; furthermore, we might want to keep text data (author notices etc.) in the repo even if we don’t deliver them to web users.

The better way to implement this, IMO, would be to do the optimization on the fly when `gulp` generates the CSS.

uklauer:
As an aside, if we pull in `zopfli` as a dependency, we could also consider creating a zopfli-compressed gzip of all the JS and CSS bundles themselves: http://www.cambus.net/serving-precompressed-content-with-nginx-and-zopfli/

mwiencek:
We now use zopfli in production, as suggested above.

PR #343, opened 2016-07-28
Mention rate limiting and the search server in INSTALL.md

mwiencek:
This seems like a useful thing to mention, since so many people ask about it.


gentlecat:
:shipit:

PR #344, opened 2016-07-31
MBS-4864: Add more events to the timeline

mineo:
Unfortunately I don't know where I might find information about the bots mentioned by @uklauer in the ticket, so those are missing.

Fixes [MBS-4864](http://tickets.musicbrainz.org/browse/MBS-4864).


uklauer:
The tenses are all over the place (the site _has got a new design_, a message _will now notify users_, certain edits _are now auto-edits_, Gentlecat _joins the team_). I assume (without checking) that the existing entries are inconsistent already, but we should probably decide on the point of view from which these texts are supposed to be written. A native speaker should have a look over this (and other possible language issues).

uklauer:
Regarding the bots, it seems the relevant date is 2011-12-01, on which day [lukz_bot started running and made 973 edits already](https://musicbrainz.org/search/edits?auto_edit_filter=&order=asc&negation=0&combinator=and&conditions.0.field=editor&conditions.0.operator=%3D&conditions.0.name=lukz_bot&conditions.0.args.0=606116&conditions.1.field=open_time&conditions.1.operator=%3D&conditions.1.args.0=2011-12-01&conditions.1.args.1=), with other bots following in the next days (hrglgrmpf_bot on 2011-12-03, Jokipii_bot on 2011-12-05; murdos_bot a month later).

Unfortunately, I don’t know what to link to; I assume there was some decision around that date to allow bot editing, but there is no blog post about it.

mwiencek:
Not sure what to link to for that either. I couldn't find any discussion about it in #musicbrainz-devel in October or November, or in the 2011 summit notes. Perhaps we should just use "#" or [this comment](https://github.com/metabrainz/musicbrainz-server/pull/344#issuecomment-236459377) as the URL. :)

mwiencek:
I was going to add some events for me and Ian joining the team since you're adding one for Roman, but I guess we never got any announcements. Oh well. Another case where we should decide what to do for events with no URL.

mwiencek:
Thanks 👍

PR #345, opened 2016-07-31
MBS-8301: Fix the cleaning of relationships link phrases in the editor interface

y-van-z:
Remove spurious spaces left from relationships attributes, especially disturbing for tree indentation.

### DIFF
--- a/root/static/scripts/edit/utility/linkPhrase.js
+++ b/root/static/scripts/edit/utility/linkPhrase.js
@@ -23,7 +23,7 @@ exports.clean = _.memoize(function (linkType, backward) {
     // remove {foo} {bar} junk, unless it's for a required attribute.
     var phrase = backward ? typeInfo.reversePhrase : typeInfo.phrase;
 
-    return phrase.replace(attributeRegex, function (match, name, alt) {
+    return clean(phrase.replace(attributeRegex, function (match, name, alt) {
         var id = idsByName[name];
 
         if (typeInfo.attributes[id].min < 1) {
@@ -31,7 +31,7 @@ exports.clean = _.memoize(function (linkType, backward) {
         }
 
         return match;
-    });
+    }));
 }, (a, b) => a + String(b));
 
 exports.interpolate = function (linkType, attributes) {


uklauer:
Please give an explanation of what you are doing (in the commit message, in particular). I can take a wild guess that there are spurious spaces in the relationship description that get removed by `trim`, but I have no idea why that would be.

y-van-z:
Sorry, I did it in JIRA instead, see [this comment](http://tickets.musicbrainz.org/browse/MBS-8301?focusedCommentId=40109&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_40109). I will fix git messages.

y-van-z:
I changed the commit code for a more general fix to the `linkPhrase.clean` function.

y-van-z:
This more general fix impacts the website user interface for non editors.
Using `replace` rather than `trim` for older browsers (IE 8).

uklauer:
I think I understand what’s happening now. (I don’t think the commit message is as detailed as it needs to be. It should explain what the broken code is supposed to do, for context; i.e. removing attribute placeholders from link phrases.) However, as I understand it, there could still remain internal sequences of multiple spaces (`foo {additional|additionally} bared by` → `foo  bared by`), and that should also be fixed. 

Regarding `trim`, a) we don’t care about IE 8 for editing, only for viewing, b) we have lodash (`_`).

y-van-z:
Now cleaning internal sequences of contiguous whitespaces too. Improved message. Thanks!

root/static/scripts/edit/utility/linkPhrase.js
```diff
@@ -31,7 +31,7 @@ exports.clean = _.memoize(function (linkType, backward) {
         }
 
         return match;
-    });
+    }).trim().replace(/ {2,}/g, ' ');
```
mwiencek: That's actually what https://github.com/y-van-z/musicbrainz-server/blob/a486351bec11911b1738d3c09fa251970a4d06a7/root/static/scripts/common/utility/clean.js does and it's already imported in this file. So you can just wrap the return like `return clean(phrase.replace(...));`.

y-van-z:
Cleaner 7-characters patch proposed by [mwiencek’s comment](https://github.com/metabrainz/musicbrainz-server/pull/345/files/a486351bec11911b1738d3c09fa251970a4d06a7#r72913888). Thanks.

mwiencek:
Thank you!

PR #346, opened 2016-08-02
Only allow admins to search for untrusted users

mwiencek:
This flag is only supposed to be set by and visible to account admins, so this can be considered a privacy breach. cc @Freso who alerted this to me.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/EditorFlag.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/EditorFlag.pm
@@ -1,19 +1,44 @@
 package MusicBrainz::Server::EditSearch::Predicate::EditorFlag;
 use Moose;
+use MusicBrainz::Server::Constants qw( $UNTRUSTED_FLAG );
 use namespace::autoclean;
 
 extends 'MusicBrainz::Server::EditSearch::Predicate::Set';
 
+has user => (
+    isa => 'Editor',
+    is => 'ro',
+    required => 1,
+);
+
+sub valid_flags {
+    my ($self) = @_;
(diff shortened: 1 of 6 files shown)


gentlecat:
LGTM

uklauer:
It is supposed? Why?

lib/MusicBrainz/Server/EditSearch/Predicate/EditorFlag.pm
```diff
@@ -1,19 +1,44 @@
...
+    isa => 'Editor',
+    is => 'ro',
+    required => 1,
+);
+
+sub valid_flags {
+    my ($self) = @_;
+
+    my @flags = @{ $self->sql_arguments->[0] };
+    unless ($self->user->is_admin) {
```
uklauer: If it is to be restricted, I don’t think it makes sense to show it to wiki transcluders, relationship editors etc. This should be `is_account_admin` instead (everywhere else in this PR, too).

(And unrelatedly, `is_admin` should probably be extended to test for account admins, too.)

uklauer: If it is to be restricted, I don’t think it makes sense to show it to wiki transcluders, relationship editors etc. This should be `is_account_admin` instead (everywhere else in this PR, too).

(And unrelatedly, `is_admin` should probably be extended to test for account admins, too.)

Freso:
We don't show the "Untrusted" flag on users profiles, so we shouldn't allow non-priv'd users to gain information about who's marked "Untrusted" by way of simply searching for edits by users with this flag.

PR #348, opened 2016-08-03
MBS-9042: Fix misaligned rows in user profile for translated UI

y-van-z:
Each row of the _General Information_ block is now grouped into a `dl`
element, so to avoid misalignment when the translated header is large.
(It uses almost the same approach than `/relationships` pages.)

Update the stylesheet accordingly.

### DIFF
--- a/root/static/styles/extra/user.less
+++ b/root/static/styles/extra/user.less
@@ -1,25 +1,25 @@
-dl.profileinfo {
-    display: block;
+table.profileinfo {
+    margin: 1em 1em 0;
 }
 
-dl.profileinfo dt {
-    clear: left;
-    display: block;
-    font-weight: bold;
-    float: left;
-    width: 125px;
+table.profileinfo th,
+table.profileinfo td {
+    padding: 5px;
+    vertical-align: top;
+}
+
+table.profileinfo th {
     text-align: right;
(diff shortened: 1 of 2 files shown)


mwiencek:
I'm not a huge fan of how this changes the semantics of the markup just for display purposes. I think all the terms related to the editor should be in a single `dl`. But if we're changing the markup, we might as well switch to a `table`.

y-van-z:
Unfortunately, I did not find a way to fix the display through style sheet only.
This is a quick and dirty fix, and yes we might switch to `table` as well.
Either way, I am not a huge fan of the fixed-width first column.

y-van-z:
This second proposition “changes the semantics” even more and get rid of the fixed width as well.

mwiencek:
Right, it's not exactly tabular data unless you want to add "key" and "value" as column headers (please no :)).

We can add `role="presentation"` to the table to indicate it's only used for layout. This stops VoiceOver (only screen reader I have to test with) from announcing it's a table with 5 rows and 2 columns, that you have a table cell focused, etc. The definition list was slightly better since it just announced you've entered a definition list with ten items, then announced when you've left it.

y-van-z:
Added presentation role to the layout table, thanks!

PR #349, opened 2016-08-09
MBS-6314, MBS-7164, MBS-8802 & MBS-9044: Update iTunes URLs’ display, clean-up and validation

y-van-z:
For iTunes *purchase for download* URL relationship
* `geo.itunes.apple.com` URLs (that already matched) are now cleaned up (like other matched URLs). **Standardisation is forced to remove the `geo` prefix from iTunes URLs.** ([MBS-9044](http://tickets.musicbrainz.org/browse/MBS-9044))
* `/audiobook/` and `/podcast/` URLs (that already matched) are now cleaned up (like other matched URLs) ([MBS-8802](http://tickets.musicbrainz.org/browse/MBS-8802)).
* Validation rules are now added so that:
  * iTunes URLs are allowed for the correct entities only ([MBS-7164](http://tickets.musicbrainz.org/browse/MBS-7164)),
  * `linkmaker.itunes.apple.com` URLs are no more accepted ([MBS-9044](http://tickets.musicbrainz.org/browse/MBS-9044)).
* iTunes sidebar links’ titles now feature country code ([MBS-6314](http://tickets.musicbrainz.org/browse/MBS-6314)), default is "iTunes US" per [iTunes doc](http://images.apple.com/itunesaffiliates/US/2009/Document/LinktoiTune.pdf).
* Move the code into a single `itunes` entry of `CLEANUPS`, making use of PR #354 changes.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/URL/iTunes.pm
+++ b/lib/MusicBrainz/Server/Entity/URL/iTunes.pm
@@ -8,7 +8,12 @@ with 'MusicBrainz::Server::Entity::URL::Sidebar';
 sub sidebar_name {
     my $self = shift;
 
-    return "iTunes";
+    if (my ($country) = $self->url->path =~ m{^/([a-z]{2})/}i) {
+        $country =~ tr/a-z/A-Z/;
+        return "iTunes $country";
+    } else {
+        return "iTunes US";
+    }
 }
 
 __PACKAGE__->meta->make_immutable;
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1180,6 +1180,27 @@ validationRules[LINK_TYPES.image.artist] = validateImage;
 validationRules[LINK_TYPES.image.label] = validateImage;
 validationRules[LINK_TYPES.image.place] = validateImage;
 
+// allow iTunes pages only for the correct entities with the downloadpurchase rel
+// block links to linkmaker.itunes.apple.com
+validationRules[LINK_TYPES.downloadpurchase.artist] = function (url) {
+  if (/itunes\.apple\.com/.test(url)) {
+    return /^https:\/\/itunes\.apple\.com\/([a-z]{2}\/)?artist\/id[0-9]+$/.test(url);
+  }
+  return true;
+}
+validationRules[LINK_TYPES.downloadpurchase.recording] = function (url) {
+  if (/itunes\.apple\.com/.test(url)) {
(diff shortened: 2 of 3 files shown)


y-van-z:
`geo` iTunes prefix is new and very uncommon in the current database, thus I removed it. But it was out of the scope of the initial issue. Maybe we should add it whenever it is missing? Or just not standardise?

y-van-z:
Do not merge for now. This will conflict with the bigger oncoming pull request for MBS-6378.

y-van-z:
Validation tests added.

mwiencek:
I'm not 100% sure what the status of this is. Content in iTunes is country-specific. So it seems to me that these links can redirect users to a store where the content is different, or even unavailable. I haven't confirmed that, but I'm not sure how else it would work.

So it seems possible to me that we should actually be disallowing these links, if the content shown is variable. In which case explicitly adding support for cleaning them doesn't make sense to me at this point.

y-van-z:
I don't know how it works. Country-specific releases available from a single `geo` URL are probably identical. When a song or an album is not available from some countries, foreign iTunes stores are probably unavailable from these countries anyway. If any of these assumptions is wrong, then I would agree to strip iTunes URLs of their `geo` prefix and to implement [MBS-6314](http://tickets.musicbrainz.org/browse/MBS-6314) as well.

We did not [decide to either standardise to geo.itunes.apple.com or away from it](https://community.metabrainz.org/t/notes-from-metabrainz-meeting-on-2016-08-15/113988). Meanwhile, it still makes sense to clean them up (without neither adding nor dumping `geo` prefixes). But I don't mind to wait for the final decision before merging any of these changes.

mwiencek:
It definitely doesn't hurt to clean them (strip query params, etc.). But my primary thought was, let's decide if we should even allow them before adding additional code for them.

Here's one example I found:
https://itunes.apple.com/us/album/more-fast-furious-music-from/id962895642
https://itunes.apple.com/nz/album/more-fast-furious-music-from/id498505533

The NZ version is missing track 8. If you create a "geo" link from the NZ id, it'll redirect to the US version, which is incorrect. There may be better examples where tracks are added or the recording versions are different.

y-van-z:
Agreed, this is the example I long-searched for!

mwiencek:
A recent example where the album is unavailable in the US:
https://itunes.apple.com/gb/album/trick/id1132752971
=> https://geo.itunes.apple.com/gb/album/trick/id1132752971?mt=1&app=music
(Second link redirects to the artist page for me.)

y-van-z:
Updated to fit #354 changes.

PR #350, opened 2016-08-11
MBS-9048: Don’t validate untouched existing URLs

y-van-z:
Previously, validation rules were applied to all URL relationships. Errors were triggered for existing URLs that do not validate newly created validation rules.

This branch skips applying validation rules for untouched existing URLs.  Modifying an existing URL or seeding/adding a new one would trigger the rules. Editors won’t be blocked anymore by unexpected errors on existing URLs.


mwiencek:
LGTM. 👍 (And I know I recently fixed this for seeded URLs, but I could've sworn we were already doing this here too. I have a bad case of deja vu right now.)

PR #351, opened 2016-08-12
MBS-8793: Improve Twitter URLs’ clean-up

y-van-z:
Removes query strings (such as `?lang=en-gb` or `?ref_src=example.com`), possible anchors (`#content-main-heading`), and sub-paths (such as `/media` or `/status/47830`) from Twitter URLs.


uklauer:
+1 to the cleanup change.

For the tests, I would prefer _adding_ some tests (possibly similar to existing ones) instead of repurposing old ones. As it currently is in this PR, “normal” URLs without trailing characters aren’t tested anymore; there could be a bug affecting a bog-standard `https://twitter.com/foobar` URL and the tests wouldn’t notice.

y-van-z:
Indeed, thanks. I restored original tests and added mine below.

mwiencek:
LGTM. 👍 Thanks!

PR #352, opened 2016-08-23
MBS-6378 & MBS-5736: Rework URLCleanup tests to support validation

y-van-z:
Previously, only `guessType` and `cleanURL` functions of URLCleanup component were tested, separately.

This patch merges both tests and adds support to  test `validationRules` functions. It also tests that clean URL still matches the same relationship type as input URL.

- First commit message gives extensive implementation details.
- Second commit one-to-one converts all existing tests.
- Further commits merge redundant tests and fix some erroneous ones disclosed by the conversion.
- Last commit finally adds some validation tests.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -165,8 +165,8 @@ test('URL cleanup component: auto-select, clean-up, and validation', {}, functio
         // (Internet) Archive
         {
                              input_url: 'http://web.archive.org/web/20100904165354/i265.photobucket.com/albums/ii229/drsaunde/487015.jpg',
-            // FIXME          input_entity_type: 'release',
-            // FIXME expected_relationship_type: 'NOMATCH',
+                     input_entity_type: 'release',
+            expected_relationship_type: undefined,
                     expected_clean_url: 'http://web.archive.org/web/20100904165354/i265.photobucket.com/albums/ii229/drsaunde/487015.jpg',
         },
         {
@@ -208,8 +208,8 @@ test('URL cleanup component: auto-select, clean-up, and validation', {}, functio
         },
         {
                              input_url: 'https://andrewhuang.bandcamp.com/track/boom-box/?test',
-            // FIXME          input_entity_type: 'recording',
-            // FIXME expected_relationship_type: 'NOMATCH',
+                     input_entity_type: 'recording',
+            expected_relationship_type: undefined,
                     expected_clean_url: 'http://andrewhuang.bandcamp.com/track/boom-box',
         },
         // BBC Music
@@ -239,14 +239,14 @@ test('URL cleanup component: auto-select, clean-up, and validation', {}, functio
         // Blogspot
         {
                              input_url: 'http://49swimmingpools.blogspot.fr/',
-            // FIXME          input_entity_type: 'artist',
-            // FIXME expected_relationship_type: 'NOMATCH',
+                     input_entity_type: 'artist',
+            expected_relationship_type: undefined,
                     expected_clean_url: 'http://49swimmingpools.blogspot.com/',
         },
         {
                              input_url: 'www.afroliciousoriginal.blogspot.pt',
-            // FIXME          input_entity_type: 'artist',
-            // FIXME expected_relationship_type: 'NOMATCH',
+                     input_entity_type: 'artist',
+            expected_relationship_type: undefined,
                     expected_clean_url: 'afroliciousoriginal.blogspot.com/',
         },
         // BookBrainz
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -89,43 +89,43 @@ test('URL cleanup component: auto-select, clean-up, and validation', {}, functio
         },
         {
                              input_url: 'http://amzn.com/B000005SU4',
-                     input_entity_type: 'release', // FIXME unchecked consistency
+                     input_entity_type: 'release',
             expected_relationship_type: 'amazon',
                     expected_clean_url: 'http://www.amazon.com/gp/product/B000005SU4',
         },
         {
                              input_url: 'http://www.amazon.co.jp/dp/tracks/B000Y3JG8U#disc_1',
-                     input_entity_type: 'release', // FIXME unchecked consistency
+                     input_entity_type: 'release',
             expected_relationship_type: 'amazon',
                     expected_clean_url: 'http://www.amazon.co.jp/gp/product/B000Y3JG8U',
         },
         {
                              input_url: 'http://www.amazon.co.uk/IMPOSSIBLE/dp/B00008CQP2/ref=sr_1_1?ie=UTF8&qid=1344584322&sr=8-1',
-                     input_entity_type: 'release', // FIXME unchecked consistency
+                     input_entity_type: 'release',
             expected_relationship_type: 'amazon',
                     expected_clean_url: 'http://www.amazon.co.uk/gp/product/B00008CQP2',
         },
         {
                              input_url: 'http://www.amazon.co.uk/Out-Patients-Vol-3-Various-Artists/dp/B00009W0XE/ref=pd_sim_m_h__1',
-                     input_entity_type: 'release', // FIXME unchecked consistency
+                     input_entity_type: 'release',
             expected_relationship_type: 'amazon',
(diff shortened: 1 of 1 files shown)


y-van-z:
Some stats without validation tests (last commit):
- before: 382 (either `Guess type` or `Cleanup`) test definitions and 382 executed test assertions,
- after: 300 (unified) test definitions and 590 executed test assertions.

y-van-z:
Forced push to clean mismatch of header comments, no real change, see https://ci.metabrainz.org/job/musicbrainz-server_master/561/ for actual test results.

y-van-z:
About TAP tests structure:
- Prior to this pull request, there were two tests and as many assertions as test entries. Assertion label allowed to identify the test entry.
- Now, there is only one test and as many sub-tests as test entries, with up to four assertions (match, cleanup, match cleanup, validation) for each test entry. Sub-test label allows to identify test entry, while assertion label allows to identify the tested functionality only, it is not redundant with either sub-test label or failure message.

Sub-tests are displayed nicely either running `node root/static/scripts/tests/Control/URLCleanup.js` or browsing `/static/scripts/tests/all.html`. Using the current Jenkins setup, sub-test labels are missing and assertion labels alone do not allow to identify test entries, but [support for sub-tests detailed output has been recently added to Jenkins TAP plug-in](https://issues.jenkins-ci.org/browse/JENKINS-33779).

@mwiencek Is it possible to update Jenkins TAP plug-in to version 2.0.1 on [https://ci.metabrainz.org/](https://ci.metabrainz.org/)?

mwiencek:
We're actually not using the TAP plugin at all (it's not installed), we're using JUnit and running prove with `--harness=TAP::Harness::JUnit`. But this comes from before my time, so I'm not sure there's a particular reason for it.

y-van-z:
FWIW, TAP plug-in depends on JUnit plug-in. I added [a commit](https://github.com/metabrainz/musicbrainz-server/commit/2f623f0e8de49e23f2d229e668deaa1ca63f6d6a) into [a separate branch](https://github.com/y-van-z/musicbrainz-server/compare/mbs-6378...y-van-z:mbs-6378-junit) to be merged/rebased with this one just in case that TAP plug-in could not be installed/used.

y-van-z:
Forced push to clean mismatch of headers, to improve failure labels, and to document the above information into commit messages (Use "[View as plain text](https://ci.metabrainz.org/job/musicbrainz-server_master/569/consoleText)" to have fully-detailed test results.)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1787,3 +1776,118 @@ test('Cleanup', function (t) {
...
 });
+
+test('URL cleanup component: auto-select, clean-up, and validation', {}, function (t) {
+    const test_data = [
+        // AllMusic
+        {
+            i_url: 'http://www.allmusic.com/album/here-comes-the-sun-mw0002303439/releases',
+            i_ent: 'release_group',
+            match: 'allmusic',
+            clean: 'http://www.allmusic.com/album/mw0002303439',
```
mwiencek: It's confusing to use a different naming system for inputs and outputs. :( Why not `o_rel` instead of `match` and `o_url` instead of `clean`? Or using 'e_' (for expected). Having a comment in the source code (rather than just in the commit message) that explains what these mean would also be welcome.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1787,3 +1776,118 @@ test('Cleanup', function (t) {
...
+                    var rule = validationRules[rel_uuid];
+                    var is_valid = rule ? rule(clean) || false : true;
+                    results[is_valid].splice(_.sortedIndex(results[is_valid], ent_name), 0, ent_name);
+                    nb_tested_rules += rule ? 1 : 0;
+                    return results;
+                }, {true: [], false: []});
+                if (nb_tested_rules === 0) {
+                    st.fail('Validation test is worthless: No validation rule has been actually tested: ' + JSON.stringify(subtest));
+                } else {
+                    st.deepEqual(validation_results[subtest.valid ? true : false], (subtest.valid || subtest.wrong).sort(),
```
mwiencek: In the [complete file](https://github.com/y-van-z/musicbrainz-server/blob/65c029b18e20d85674f74a76ad2a1aac6d9f1224/root/static/scripts/tests/Control/URLCleanup.js) I only see `wrong` used once and it's an empty list, which I would think is equivalent to just specifying every `valid` entity type. Wouldn't it be a lot simpler to remove support for `wrong` and do that?

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1787,3 +1776,118 @@ test('Cleanup', function (t) {
...
+            if (subtest.valid || subtest.wrong) {
+                var match = subtest.i_rel || subtest.match;
+                var clean = subtest.clean || act_clean;
+                if (!match) {
+                    st.fail('Test is invalid: Validation test can not run without neither "match" nor input relationship type "i_rel": ' + JSON.stringify(subtest));
+                    st.end();
+                    return;
+                }
+                var nb_tested_rules = 0;
+                var validation_results = _.reduce(LINK_TYPES[match], function(results, rel_uuid, ent_name) {
```
mwiencek: We use `entity_type` or `entityType` consistently throughout the codebase, rather than ent*name.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1787,3 +1776,118 @@ test('Cleanup', function (t) {
...
+    const reltypes_by_uuid = _.reduce(LINK_TYPES, function(results, entities, name) {
+        _.each(entities, function(uuid) {
+            (results[uuid] || (results[uuid] = [])).push(name);
+        });
+        return results;
+    }, {});
+
+    var previous_match_tests = [];
+
+    function doMatchSubtest(st, ent, url, label, exp_match) {
```
mwiencek: Here I'd prefer `ent` was replaced with `entity_type` too, so that we're using the same variable name for the same concepts where possible.

root/static/scripts/tests/Control/URLCleanup.js
mwiencek: Alternatively, use longer/more descriptive property names that are self-explanatory (`input_url`, `output_relationship`, etc.). I like that they're all 5 chars but you can always use whitespace to align them. :)

y-van-z: I avoided to use the term _output_ because both `match` and `clean` are actually used as input by the two last tests.
I tried to keep tests conciseness and to add self-explanatory property names (but it looks like I failed that final point).

root/static/scripts/tests/Control/URLCleanup.js
y-van-z: Yes, I will rename also `exp_match` to the new (still undecided) property name for `match`.

root/static/scripts/tests/Control/URLCleanup.js
y-van-z: So I propose the following (lengthy but self-explanatory) (might be less readable though) renaming:
- `i_url` -> `input_url`
- `i_ent` -> `input_entity_type`
- `match` -> `expected_relationship_type`
- `clean` -> `expected_clean_url`
- `i_rel` -> `input_relationship_type`
- `valid` -> `only_valid_entity_types`
- `wrong` -> `only_wrong_entity_types`

mwiencek: Those seem a lot clearer to me. Hopefully they don't hurt readability.

root/static/scripts/tests/Control/URLCleanup.js
y-van-z: I will change it.

root/static/scripts/tests/Control/URLCleanup.js
y-van-z: Yes. I initially implemented this with ResidentAdvisor reviews in mind, the complete file proves this was unneeded.

y-van-z:
One more forced push to fix coding style and remove the _wrong_ feature, thanks to @mwiencek review.

mwiencek:
Thank you. :)

Just a minor note for future PRs, for JavaScript we use camelCase for variable names EXCEPT_FOR_CONSTANTS. It's not a big deal but it's nice to have some consistency across files.

y-van-z:
Thanks for your time at review. :+1: I updated PR #354 to follow this naming convention.

PR #354, opened 2016-09-04
MBS-5733: Allow to validate `otherdatabases` URL relationships

y-van-z:
Previously, it was not handy to define validation rules for `otherdatabases` URLs, because validation rules had to be defined separately for each entity type whereas there were much more other databases.

Changes caused by this improvement:
- `LINK_TYPES` no longer contain fictive duplicate IDs
- `CLEANUPS` are website `match`-related before relationship `type`-related
- New validation rules can be defined in `CLEANUPS.*.validate` and are stricter than old ones:
  - Matched URLs could be blocked for other relationship types if `type` does not match
  - Matched URLs could be blocked for other entity types if `type` does not match
  - Matched URLs could be blocked if unclean

Old validation rules are still supported. Ultimately, all of them should be converted.

Additionally, other restricted URL relationships (`lyrics`, `score`, `youtube`) can be validated the same way.

Two examples are provided: Soundtrack Collector (converted rules) and The Session (created rules).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -736,7 +736,6 @@ const CLEANUPS = {
       new RegExp("^(https?://)?(www\\.)?openlibrary\\.org", "i"),
       new RegExp("^(https?://)?(www\\.)?animenewsnetwork\\.com", "i"),
       new RegExp("^(https?://)?(www\\.)?generasia\\.com/wiki/", "i"),
-      new RegExp("^(https?://)?(www\\.)?soundtrackcollector\\.com", "i"),
       new RegExp("^(https?://)?(www\\.)?rockipedia\\.no", "i"),
       new RegExp("^(https?://)?(www\\.)?whosampled\\.com", "i"),
       new RegExp("^(https?://)?(www\\.)?maniadb\\.com", "i"),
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -232,6 +226,35 @@ const LINK_TYPES = {
   }
 };
 
+const RESTRICTED_LINK_TYPES = _.reduce([
```
y-van-z: It does what it looks like to do. I picked the term _restricted_ from the doc [Style / Relationships / URLs / Restricted relationships](https://musicbrainz.org/doc/Style/Relationships/URLs#Restricted_relationships). Shall I replace it by _whitelisted_ instead?

mwiencek: Name is fine, I think just linking to that page in a comment should be explanatory enough.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -768,19 +788,48 @@ const CLEANUPS = {
...
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?soundtrackcollector\.com\/(composer|title)\/([0-9]+).*$/, "http://soundtrackcollector.com/$1/$2/");
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?soundtrackcollector\.com\/.*\?movieid=([0-9]+).*$/, "http://soundtrackcollector.com/title/$1/");
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?soundtrackcollector\.com\/.*\?composerid=([0-9]+).*$/, "http://soundtrackcollector.com/composer/$1/");
       return url;
+    },
+    validate: function (url, id) {
+      if (id === LINK_TYPES.otherdatabases.artist)
+        return /^http:\/\/soundtrackcollector\.com\/composer\/[0-9]+\/$/.test(url);
```
y-van-z: Done, see 280521ab9a56de2b53c4e7156bd18e2df1dd2f6 and b61b16c086df326c014ea3fd59c34891b40df9d7.

mwiencek:
👍 Thanks. :) Sorry again that it took so long to review.

PR #356, opened 2016-09-29
MBS-9072: Validate the tagger port URL parameter

uklauer:
The contents of the tagger port URL parameter are stored in the user session and later included in various generated pages. Since there used to be no validation, arbitrary text could be injected (a XSS
vulnerability).

Make sure that the tport value is a number (of at most five digits); report an invalid request otherwise.


mwiencek:
:shipit:

PR #357, opened 2016-09-29
MBS-9069: Allow cover art comments to have arbitrary length

uklauer:
The `comment` field for cover art edits used `Field::Comment`, which otherwise is used for entity disambiguation comments and therefore imposes a maximum length on the field contents. Use an unrestricted `Field::Text` instead.


mwiencek:
:shipit:

PR #358, opened 2016-09-29
Discogs URL cleanup/validation improvements (includes MBS-9012)

uklauer:
Some improvements for Discogs URLs:

- Better validation – disallow anything that doesn’t have the proper format.
- Drop everything after the ID (fixes MBS-9012 in particular).
- Harmonize to HTTPS URLs.

This also removes the remaining support for old URLs that didn’t contain a numerical ID.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -254,9 +254,8 @@ const CLEANUPS = {
     type: LINK_TYPES.discogs,
     clean: function (url) {
       url = url.replace(/\/viewimages\?release=([0-9]*)/, "/release/$1");
-      url = url.replace(/^https?:\/\/([^.]+\.)?discogs\.com\/(.*\/(artist|release|master|label))?([^#?]*).*$/, "http://www.discogs.com/$3$4");
+      url = url.replace(/^https?:\/\/(?:[^.]+\.)?discogs\.com\/(?:.*\/)?(user\/[^\/#?]+|(?:artist|release|master(?:\/view)?|label)\/[0-9]+)(?:[\/#?-].*)?$/, "http://www.discogs.com/$1");
       url = url.replace(/^(http:\/\/www\.discogs\.com\/master)\/view\/([0-9]+)$/, "$1/$2");
-      url = url.replace(/^(http:\/\/www\.discogs\.com\/(?:artist|label))\/([0-9]+)-[^+]*$/, "$1/$2"); // URLs containing Discogs IDs
       return url;
     }
   },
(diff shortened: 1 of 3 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -254,13 +254,8 @@ const CLEANUPS = {
...
     clean: function (url) {
       url = url.replace(/\/viewimages\?release=([0-9]*)/, "/release/$1");
-      url = url.replace(/^https?:\/\/([^.]+\.)?discogs\.com\/(.*\/(artist|release|master|label))?([^#?]*).*$/, "http://www.discogs.com/$3$4");
-      url = url.replace(/^(http:\/\/www\.discogs\.com\/master)\/view\/([0-9]+)$/, "$1/$2");
-      url = url.replace(/^(http:\/\/www\.discogs\.com\/(?:artist|label))\/([0-9]+)-[^+]*$/, "$1/$2"); // URLs containing Discogs IDs
-      var m = url.match(/^(http:\/\/www\.discogs\.com\/(?:artist|label))\/(.+)/);
-      if (m) {
-        url = m[1] + "/" + encodeURIComponent(decodeURIComponent(m[2].replace(/\+/g, "%20"))).replace(/%20/g, "+");
-      }
+      url = url.replace(/^https?:\/\/(?:[^.]+\.)?discogs\.com\/(?:.*\/)?(user\/[^\/#?]+|(?:artist|release|master(?:\/view)?|label)\/[0-9]+)(?:[\/-#?].+)?$/, "https://www.discogs.com/$1");
```
mwiencek: There's a parse error in the `[\/-#?]` I think ("Range out of order in character class"). Should the hyphen-minus be escaped?

uklauer: Ah, I guess the `-` needs to be moved to the beginning or end of the range; that’s how it works in Perl at least. Will investigate and fix.

y-van-z: Dash should be escaped in the characters range `[\/\-#?]`. The rest looks good to me. :+1:

uklauer: It appears either way is possible to get a hyphen into a character class. I chose to put it at the edge.

uklauer:
@mwiencek: I still get errors in the release editor tests. I assume the hash values used there need to be regenerated after changing `http` to `https`; but how are they calculated?

y-van-z:
They seem to be calculated from the edit constructor using the [`editHash`](https://github.com/uklauer/musicbrainz-server/blob/4125e8ce58ee2c09db980eaad89c94eb197e5943/root/static/scripts/edit/MB/edit.js#L271-L280) function.

mwiencek:
Right, though the test runner will always tell me what the new hash should be, at least on the command line, and I just copy that if the rest of the fields look correct, since that's what it's calculated from. This should contain the correct hashes: https://gist.github.com/mwiencek/38f3bcee59b2550dd275492520873ab3

uklauer:
Changed the hashes, and tests now complete as successful.

PR #359, opened 2016-09-29
Wikimedia Commons URL cleanup improvements (includes MBS-9053)

uklauer:
Some improvements related to Wikimedia Commons:

- Recognize `commons.m.wikimedia.org` URLs and remove the `m.` on cleanup. (MBS-9053)
- Remove `#fragment` from Wikimedia Commons URLs.
- Add a test for both.
- Fix a comment. (Wikimedia is the organization that, among other projects like Wiktionary and Wikipedia, runs Wikimedia Commons.)

This builds on pull request #354, which should be merged first.


mwiencek:
#354 has been merged, but it looks like this needs to be rebased now.

uklauer:
Should be OK now (without rebase – it seems GitHub no longer notices by itself when some commits have already been merged; I pushed a dummy commit, then removed it again, to make GitHub re-check the branch).

mwiencek:
Huh, that's a pretty confusing regression. Thanks for clearing that up.

PR #361, opened 2016-09-30
Validate Generasia URLs (solves MBS-8744)

y-van-z:
Like most of other databases, Generasia URLs were allowed for all entity types.

This makes use of PR #354 to allow it for Artists, Release Groups, and Works only.

It prevents linking to Generasia from Releases (MBS-8744).


mwiencek:
🚢

PR #362, opened 2016-10-02
Fix auto-generated URL validation rules, part of PR#354

y-van-z:
Fix auto-generated URL validation rules added by commit 8514695c21c1c8acdd0d42ccb5a20ece44b3ae65. The bug was to block URLs for some link types when there is no auto-selected link type for this entity type in the corresponding `CLEANUPS` entry.


mwiencek:
:shipit:

PR #363, opened 2016-10-02
MBS-9078: Cleanup and validation improvements for Wikimedia Commons URLs

uklauer:
See the individual commits. The tests are taken from original URLs found in the database (except for the artificial “special characters” test, for which I copied the corresponding Wikipedia test).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -257,6 +257,14 @@ const RESTRICTED_LINK_TYPES = _.reduce([
   LINK_TYPES.youtube,
 ], function (result, linkType) {return result.concat(_.values(linkType));}, []);
 
+function reencode_mediawiki_localpart(url) {
+  var m = url.match(/^(https?:\/\/[^\/]+\/wiki\/)([^?#]+)(.*)$/);
+  if (m) {
+    url = m[1] + encodeURIComponent(decodeURIComponent(m[2])).replace(/%20/g, "_").replace(/%24/g, "$").replace(/%2C/g, ",").replace(/%2F/g, "/").replace(/%3A/g, ":").replace(/%3B/g, ";").replace(/%40/g, "@") + m[3];
+  }
+  return url;
+}
+
 const CLEANUPS = {
   wikipedia: {
     match: [new RegExp("^(https?://)?(([^/]+\\.)?wikipedia|secure\\.wikimedia)\\.","i")],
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -257,6 +257,14 @@ const RESTRICTED_LINK_TYPES = _.reduce([
   LINK_TYPES.youtube,
 ], function (result, linkType) {return result.concat(_.values(linkType));}, []);
 
+function reencode_mediawiki_localpart(url) {
```
mwiencek: Below you call a function named `cleanup_mediawiki_localpart`.

uklauer: Argh …

Freso:
Thank you! Those hashes have been bugging me for a while.

root/static/scripts/edit/URLCleanup.js
uklauer: Fixed that, but the tests failed again; apparently another syntax error, unfortunately I can’t tell from what MBJenkins tells me.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1713,6 +1713,48 @@ test('URL cleanup component: auto-select, clean-up, and validation', {}, functio
...
+                    expected_clean_url: 'https://commons.wikimedia.org/wiki/File:Michl_M%C3%BCller_Garitz_2013.JPG',
+        },
+        {
+                             input_url: 'https://commons.wikimedia.org/wiki/File%3APolygram_Gold_Disc_Award_Wet_Wet_Wet_Ralph_Ruppert_1987.jpg',
+                    expected_clean_url: 'https://commons.wikimedia.org/wiki/File:Polygram_Gold_Disc_Award_Wet_Wet_Wet_Ralph_Ruppert_1987.jpg',
+        },
+        {
+                             input_url: 'https://commons.wikimedia.org/wiki/Within_Temptation',
+                                        // gallery page
+                     input_entity_type: 'artist'.
```
mwiencek: There it is, a period at the end instead of a comma. :) But I ran the tests again after fixing that locally and there are some more issues:

```
# input URL [291] = https://commons.wikimedia.org/wiki/Within_Temptation
not ok 650 Test is invalid: "input_entity_type" is specified without "expected_relationship_type".
  ---
    operator: fail
  ...
# input URL [292] = https://commons.wikimedia.org/wiki/Category:Ewa_Demarczyk
not ok 651 Test is invalid: "input_entity_type" is specified without "expected_relationship_type".
  ---
    operator: fail
  ...
# input URL [293] = https://commons.wikimedia.org/w/index.php?search=umberto+alongi&title=Special%3ASearch&go=Go
not ok 652 Test is invalid: "input_entity_type" is specified without "expected_relationship_type".
  ---
    operator: fail
  ...
```

uklauer:
So, after some [discussion](https://chatlogs.metabrainz.org/brainzbot/metabrainz/2016-10-03/?msg=3703082&page=4) with @y-van-z, the fix for those image URL tests was to move the global deeplink-blocking rules into a local linktype block, allowing different rules for images to coexist. I hope the tests will pass now.

(While moving those rules, I noticed that Discogs nowadays uses different image domains, so I added those even though it doesn’t strictly belong here.)

PR #365, opened 2016-10-03
Comment and drop some legacy definitions of URL validation rules in URLCleanup

y-van-z:
This follows PR #354 on deprecation and conversion of URL validation rules
- Comment on legacy definitions of URL validation rules
- Drop legacy definitions based on `CLEANUPS.*.match`
- Drop legacy definitions that do no better than `CLEANUPS.*.match`

These validation rules are now auto-generated without weakening them.

The rest of legacy definitions require additional `CLEANUPS.*.validate` functions to replace them.


y-van-z:
“One SecondHandSongs URL validation rules may hide another”… force-pushed new set of commits.

uklauer:
I built on this in PR #366. It’s not complete yet, but I hope I will finish it tomorrow.

PR #366, opened 2016-10-04
[WIP] Replace global URL validation rules

uklauer:
The old global and the new local URL validation rules don’t mix well, so the former need to be replaced. At the same time, I’m tightening some validation rules a bit.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -759,7 +759,7 @@ const CLEANUPS = {
     match: [new RegExp("^(https?://)?vgmdb\\.(net|com)/", "i")],
     type: LINK_TYPES.vgmdb,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?vgmdb\.(?:net|com)\/(album|artist|org)\/([0-9]+).*$/, "http://vgmdb.net/$1/$2");
+      return url.replace(/^(?:https?:\/\/)?vgmdb\.(?:net|com)\/(album|artist|event|org)\/([0-9]+).*$/, "http://vgmdb.net/$1/$2");
     },
     validate: function (url, id) {
       switch (id) {
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1105,12 +1105,6 @@ validationRules[LINK_TYPES.setlistfm.place] = function (url) {
   return /setlist\.fm\/venue\//.test(url);
 };
 
-// avoid wikipedia being added as release-level discography entry
-var isWikipedia = /^(https?:\/\/)?([^.]+\.)?wikipedia\.org\//;
-validationRules[LINK_TYPES.discographyentry.release] = function (url) {
-  return !isWikipedia.test(url);
-};
-
 function validateFacebook(url) {
   if (/facebook.com\/pages\//.test(url)) {
     return /\/pages\/[^\/?#]+\/\d+/.test(url);
@@ -1141,6 +1135,15 @@ _.each(LINK_TYPES, function (linkType) {
     }
   });
 });
+
+// avoid Wikipedia being added as release-level discography entry
+var originalRule = validationRules[LINK_TYPES.discographyentry.release];
+validationRules[LINK_TYPES.discographyentry.release] = function (url) {
+  if (/^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\//.test(url)) {
+    return false;
+  }
+  return originalRule(url);
+};
 
 function guessType(sourceType, currentURL) {
   var cleanup = _.find(CLEANUPS, function (cleanup) {
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -398,6 +398,9 @@ const CLEANUPS = {
     type: LINK_TYPES.bookbrainz,
     clean: function (url) {
       return url.replace(/^(?:https?:\/\/)?(?:[^.]+\.)?bookbrainz\.org\/([^\/]*)\/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})(?:[#\/?].*)?$/, "https://bookbrainz.org/$1/$2");
+    },
+    validate: function (url, id) {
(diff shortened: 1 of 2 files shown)


uklauer:
Still to do: VGMdb, YouTube, Amazon, IMDb, SHS, Bandcamp, Songkick, setlist.fm
(Wikidata is part of PR #364, and deep-linked images are part of PR #363.)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -501,7 +548,11 @@ const CLEANUPS = {
       new RegExp("^(https?://)?(www\\.)?avexnet\\.jp/id/[a-z0-9]{5}/discography/product/[A-Z0-9]{4}-[0-9]{5}\\.html$", "i"),
       new RegExp("^(https?://)?(www\\.)?kingrecords\\.co\\.jp/cs/g/g[A-Z]{4}-[0-9]+/$", "i")
     ],
-    type: LINK_TYPES.discographyentry
+    type: LINK_TYPES.discographyentry,
+    validate: function (url, id) {
+      // avoid wikipedia being added as release-level discography entry
+      return !/^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\//.test(url);
```
uklauer: We have a problem here: For a Wikipedia URL, this function isn’t run because the URL doesn’t match the `match` expressions. I could obviously add a regex for WP there, but that would lead to “discographyentry” being auto-selected when a WP URL is pasted. This would likely confuse the user even more than they already are, trying to make a release–WP link.

uklauer: I wonder if it is good idea to add a `reject:` (or similar) directive, with a list of regexes that should be tested (and if found, rejected), but that wouldn’t cause cleanup or auto-selecting of the link type.

y-van-z: I didn’t read your above message before posting mine, and I still don’t get exactly the idea.

The problem here is closely related to PR #362 that fixed PR #354.

y-van-z: Much better solution: `&& _.includes(CLEANUPS.wikipedia.types, id)` into `validate` function.

y-van-z:
Pursued a bit of migration task on https://github.com/uklauer/musicbrainz-server/compare/validation_rules...y-van-z:validation_rules

y-van-z:
Finished migration task (with a forced push). Note that rules conversion almost always tighten validation rules, because it generally discard all other link types for matched URLs.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -272,6 +275,9 @@ const CLEANUPS = {
         url = m[1] + encodeURIComponent(decodeURIComponent(m[2])).replace(/%20/g, "_").replace(/%24/g, "$").replace(/%2C/g, ",").replace(/%2F/g, "/").replace(/%3A/g, ":").replace(/%3B/g, ";").replace(/%40/g, "@") + m[3];
       }
       return url;
+    },
+    validate: function (url, id) {
+      return /^https:\/\/[a-z]+\.wikipedia\.org\/wiki\//.test(url);
```
y-van-z: This could be tighten by adding the condition `_.includes(CLEANUPS.wikipedia.type, id)` so to discard Wikipedia URLs for other link types, see `amazon` and `youtube` for examples.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -389,6 +452,9 @@ const CLEANUPS = {
     type: LINK_TYPES.bookbrainz,
     clean: function (url) {
       return url.replace(/^(?:https?:\/\/)?(?:[^.]+\.)?bookbrainz\.org\/([^\/]*)\/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})(?:[#\/?].*)?$/, "https://bookbrainz.org/$1/$2");
+    },
+    validate: function (url, id) {
+      return /^https:\/\/bookbrainz\.org\/[^\/]+\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/.test(url);
```
y-van-z: This could be tighten by adding the condition `_.includes(CLEANUPS.bookbrainz.type, id)` so to discard BookBrainz entities for other link types, see `amazon` and `youtube` for examples.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -470,14 +535,32 @@ const CLEANUPS = {
...
   bbcmusic: {
     match: [new RegExp("^(https?://)?(www\\.)?bbc\\.co\\.uk/music/artists/", "i")],
-    type: LINK_TYPES.bbcmusic
+    type: LINK_TYPES.bbcmusic,
+    clean: function (url) {
+      url = url.replace(/\/music\/artists\/([0-9a-f-]+)/, "http://www.bbc.co.uk/music/artists/$1");
+      return url;
+    },
+    validate: function (url, id) {
+      return /^http:\/\/www\.bbc\.co\.uk\/music\/artists\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/.test(url);
```
y-van-z: This could be tighten by adding the condition `_.includes(CLEANUPS.bbcmusic.type, id)` so to discard BBC Music artist pages for other link types, see `amazon` and `youtube` for examples.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -536,15 +619,25 @@ const CLEANUPS = {
...
+    validate: function (url, id) {
+      switch (id) {
+        case LINK_TYPES.secondhandsongs.artist:
+          return /secondhandsongs\.com\/artist\//.test(url);
+        case LINK_TYPES.secondhandsongs.release:
+          return /secondhandsongs\.com\/release\//.test(url);
+        case LINK_TYPES.secondhandsongs.work:
+          return /secondhandsongs\.com\/work\//.test(url);
+      }
+      return true;
```
y-van-z: A `validate` function should never return `true` because it allows to register any matched URL to any restricted link type such as `LINK_TYPE.amazon.release`. It should instead allow unrestricted link type only with `RESTRICTED_LINK_TYPES.indexOf(id) === -1`, or preferably we could add a function `isUnrestricted` that computes it and have `validate` functions return `isUnrestricted(id)` rather than `true`. This applies to `bandcamp`, `songkick`, and `setlistfm` too.

PR #371, opened 2016-10-14
Clean up Amazon and YouTube URLs to HTTPS

uklauer:


lib/MusicBrainz/Server/Report/ASINsWithMultipleReleases.pm
```diff
@@ -20,7 +20,7 @@ sub query {
                 FROM
                     url JOIN l_release_url lru ON lru.entity1 = url.id
                 WHERE
-                    url ~ E'^http://www\\\\.amazon\\\\.(com|ca|co\\\\.uk|fr|de|it|es|co\\\\.jp|cn)'
+                    url ~ E'^https?://www\\\\.amazon\\\\.[^/]+/'
```
uklauer: The list of local Amazon websites had fallen out of sync with the cleanup code and would probably fall out of sync again, so I just removed it. For this report, it’s not important.

PR #373, opened 2016-10-17
MBS-9091: Aliases: Serialize begin/end date and ended flag.

Zastai:
I followed the style used for LifeSpan, i.e. emit null for the dates when they're empty, and always included the ended flag.


Zastai:
Will look at those failing tests. I assume they're just not expecting the alias dates to appear.

mwiencek:
LGTM. Thanks!

PR #375, opened 2016-10-20
Remove the in-memory cache

mwiencek:
This probably isn't critical for performance, and it only causes an issue for us: the cache is per-server (not shared), so we must wait for every cache to expire (within 1 hour) after new entities are added.


mwiencek:
I'm going to merge my pull requests to master after their tests pass, since we're kind of running out of time (keep in mind I'm also going to the GSoC summit) and need to get stuff deployed properly ASAP. Please still comment if you see any problems though.

PR #376, opened 2016-10-20
No need to create PL/pgSQL

mwiencek:
It's already installed by default in PostgreSQL 9.0 and later.

### DIFF
--- a/admin/InitDb.pl
+++ b/admin/InitDb.pl
@@ -171,7 +171,7 @@ sub Create
     my $system_sql;
 
     # Check we can find these programs on the path
-    for my $prog (qw( pg_config createuser createdb createlang ))
+    for my $prog (qw( pg_config ))
     {
         next if `which $prog` and $? == 0;
         die "Can't find '$prog' on your PATH\n";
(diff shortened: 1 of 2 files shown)


admin/InitDb.pl
```diff
@@ -228,14 +228,9 @@ sub Create
...
-    # You can do this via CREATE FUNCTION, CREATE LANGUAGE; but using
-    # "createlang" is simpler :-)
-    my $sys_db = Databases->get($sysname);
-    my $sys_in_thisdb = $sys_db->meta->clone_object($sys_db, database => $dbname);
-    my @opts = $sys_in_thisdb->shell_args;
-    splice(@opts, -1, 0, "-d");
-    $ENV{"PGPASSWORD"} = $sys_db->password;
-    system "createlang", @opts, "plpgsql";
+    $system_sql->auto_commit;
+    $system_sql->do("CREATE OR REPLACE LANGUAGE plpgsql");
```
mwiencek: Though I believe I still need to change the database here.

mwiencek: Wait, is this even needed? My copy of PG 9.5 already has the plpgsql language in the template1 database (it can't be dropped, because "extension plpgsql requires it"), which means the MB database already inherits the language.

mwiencek: Quoting https://www.postgresql.org/docs/9.5/static/plpgsql-overview.html

```
In PostgreSQL 9.0 and later, PL/pgSQL is installed by default.
```

So I guess the answer is "no."

PR #377, opened 2016-10-20
MBS-9096: Upgrade DBD::Pg to version 3.5.3

uklauer:
Increase the minimum supported version for `DBD::Pg` to 3.5.3. The new version handles UTF-8 de-/encoding for all column types (except `bytea`) transparently, whereas the 2.x versions only did this for `text`, `char` and `varchar`, but most notably not for `jsonb`. Therefore, remove the code for explicit de-/encoding that is no longer needed.


mwiencek:
@brainzbot retest this please

mwiencek:
Awesome, thank you.

PR #380, opened 2016-10-21
Improve root/server.js for running inside docker

mwiencek:
The gist is that we're now relying on a DBDefs.js file for our JavaScript instead of environment variables. This can be generated by ./script/dbdefs_to_js.pl in the case of client or server-side JS, or it can be generated by consul-template in the case of server-side JS, and reloaded by root/server.js on HUP.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /admin/apache/startup.pl
 /admin/replication/hooks/post-process
 /root/robots.txt
+/root/static/scripts/common/DBDefs.js
 /blib
 /lib/DBDefs.pm
 /Makefile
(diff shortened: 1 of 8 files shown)


mwiencek:
./script/dbdefs_to_js.pl is useful in combination with ./script/compile_resources.sh. It means we can access DBDefs values in our client JS now.

We'll use consul-template to generate DBDefs.js in the docker container, though, because we don't want to have to install a ton of perl dependencies to execute ./script/dbdefs_to_js.pl. (The template renderer will run in a separate container from the server; it'll mostly just have node installed.)

mwiencek:
This is a bit broken right now since it's using new DBDefs settings (which I might remove) from #381 and that's still WIP.

PR #381, opened 2016-10-21
MBS-9031: Replace memcached with redis

mwiencek:
Re-opening #341

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -204,34 +204,14 @@ Cache
 
 Keys:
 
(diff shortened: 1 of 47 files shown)


mwiencek:
Still WIP. I think we'll want separate redis instances for the entity cache and data store, since we'll want to configure them differently. I'm refactoring things slightly to reduce code duplication and API differences between this new CacheWrapper and DataStore::Redis. (The latter will continue to encode/decode data in the same way for compatibility.)

PR #382, opened 2016-10-21
Support static resource domains (staticbrainz)

mwiencek:
The majority of the changes are to do with images. Images are now copied to root/static/build along with JS and CSS, their filenames appended with a hash, and all references to them looked up in rev-manifest.json.

This makes it trivial to copy all assets to another server and run zopfli on them, for example.

### DIFF
--- a/lib/DBDefs/Default.pm
+++ b/lib/DBDefs/Default.pm
@@ -136,6 +136,9 @@ sub CANONICAL_SERVER          { "https://musicbrainz.org" }
 # The server used to link to CritiqueBrainz users and reviews.
 sub CRITIQUEBRAINZ_SERVER     { "https://critiquebrainz.org" }
 
+# The URL where static resources are located, excluding the trailing slash.
+sub STATIC_RESOURCES_LOCATION { 'http://localhost:5000/static' }
+
 ################################################################################
 # Mail Settings
 ################################################################################
--- a/lib/MusicBrainz/Server/Data/FileCache.pm
+++ b/lib/MusicBrainz/Server/Data/FileCache.pm
@@ -75,6 +75,13 @@ sub template_signature {
     return $instance->file_signatures->{$signature_key};
 }
 
+sub path_to {
+    my ($self, $manifest) = @_;
+
+    my $signature = $self->manifest_signature($manifest);
+    return DBDefs->STATIC_RESOURCES_LOCATION . '/' . $signature;
+}
+
 sub _expand {
     my ($path, $type) = @_;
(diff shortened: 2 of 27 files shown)


lib/DBDefs/Default.pm
```diff
@@ -136,6 +136,9 @@ sub CANONICAL_SERVER          { "https://musicbrainz.org" }
 # The server used to link to CritiqueBrainz users and reviews.
 sub CRITIQUEBRAINZ_SERVER     { "https://critiquebrainz.org" }
 
+# The URL where static resources are located, excluding the trailing slash.
+sub STATIC_RESOURCES_LOCATION { 'http://localhost:5000/static' }
```
uklauer: Since this caused a problem already (IRC), perhaps it should be changed to use the `WEB_SERVER` definition? E.g.

    sub STATIC_RESOURCE_LOCATION { 'http://' . shift->WEB_SERVER . '/static' }

Perhaps even use `'//'` instead og `'http://'` so that it works for someone who has set up SSL, too?

Zastai: Even just mentioning in INSTALL.md that the setting exists and needs to be customized if you start plackup on any port other than 5000 would have helped.

Zastai: Also note that even `'http://' . shift->WEB_SERVER . '/static'` does not work. The resources are compiled to files under root/static/build, so it would need to be `'http://' . shift->WEB_SERVER . '/static/build'` to find those.

uklauer: Hmm, I’m not sure about that. Perhaps it’s just that the images need `build/` added to their path? Adding `build` here in the config will break the JS, in any case, because it will be searched for in `build/build`.

mwiencek: I think https://github.com/metabrainz/musicbrainz-server/commit/6d23419b5703d2df7c4d3d63d3f17931b0efd3fe fixes this, I don't see any immediate issues with JS, CSS, or images.

A sample of rev-manifest.json might look like this:
```
{"/build/jed-nl.js":"jed-nl-84597edbd4.js","/images/image404-125.png":"image404-125-98642f1825.png"}
```

The idea I went with is that the keys are the original paths (relative to static/) and the values are the filenames relative to `STATIC_RESOURCES_LOCATION`. Since everything is contained in build/, `STATIC_RESOURCES_LOCATION` should include build/.

It's a bit weird though, since the JS and CS don't have "original paths," they consist of many original files bundled together.

PR #384, opened 2016-10-23
Convert Makefile.PL to cpanfile

mwiencek:
The cpanfile works with both carton and cpanm and doesn't require all this Module::Install crap (we're not publishing to CPAN).


gentlecat:
No changes to `INSTALL.md` needed?

mwiencek:
No changes needed, `cpanm --installdeps --notest .` still works perfectly fine.

mwiencek:
Well, now it does. :) It seems the 'feature' specifiers aren't installed by default anymore, so I moved some modules to the top where needed. The server starts fine after a clean run of `cpanm --installdeps --notest .` now.

cpanfile
```diff
@@ -8,6 +8,7 @@ requires 'Catalyst::Action::RenderView'               => '0.16';
 requires 'Catalyst::Plugin::Authentication'           => '0.10023';
 requires 'Catalyst::Authentication::Credential::HTTP' => '1.016';
 requires 'Catalyst::Plugin::Cache'                    => '0.12';
+requires 'Catalyst::Plugin::Cache::HTTP'              => '0.001000';
```
mwiencek: `USE_ETAGS` is enabled by default, so the server would break by default without this.

PR #385, opened 2016-10-25
Added output_success() to the JSON serializer.

Zastai:
Makes Accept: application/json work for submissions.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/JSONSerializer.pm
+++ b/lib/MusicBrainz/Server/WebService/JSONSerializer.pm
@@ -151,6 +151,14 @@ sub output_error {
     return encode_json({ error => $err });
 }
 
+sub output_success {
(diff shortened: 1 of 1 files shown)


uklauer:
Can you explain what this does? What submissions? What happened before?

Zastai:
It's the method called by the web service processing for a succesful submission (isrcs, barcodes, etc). Since you post xml for those, it's reasonable to also have the Accept header set to application/xml; but if it's set to application/json, the web service tries to call this method and fails. Adding it is trivial and does no harm, plus it would allow my client library to prefer json all the way.

uklauer:
OK; a bit strange to post XML and receive JSON, but fine with me.

Regarding the rationale: This is exactly what should be in a commit message. Please put it there.

Zastai:
Updated the commit message.

PR #386, opened 2016-10-28
MBS-7795: Display a map in Areas that shows all the places contained in the area

loujine:
http://tickets.musicbrainz.org/browse/MBS-7795


root/components/area-map.tt
```diff
@@ -0,0 +1,40 @@
...
+    [%~ FOR place IN places ~%]
+        [%~ IF place.coordinates.defined ~%]
+            places.push({
+                title: '[% place.name | js %]',
+                type: '[% place.l_type_name %]',
+                lat: [% place.coordinates.latitude %],
+                lon: [% place.coordinates.longitude %]
+            });
+        [%~ END ~%]
+    [%~ END ~%]
```
loujine: I guess there is a cleaner way to transform a "places" data structure server-side into a javascript array, but I didn't find it

root/components/area-map.tt
```diff
@@ -29,10 +29,33 @@
...
+            Other: 'marker',
+            Studio: 'studio-marker',
+            Venue: 'venue-marker',
+            Stadium: 'stadium-marker',
+            'Indoor arena': 'arena-marker',
+            'Religious building': 'religious-marker'
+        };
+        var icons = _.mapValues(iconNames, function (name) {
+            return new LeafIcon(
+                { iconUrl: '/static/lib/leaflet/images/' + name + '-icon.png' }
```
loujine: Icons come from https://github.com/pointhi/leaflet-color-markers where I can't find a license

lib/MusicBrainz/Server/Controller/Area.pm
```diff
@@ -145,6 +145,22 @@ sub releases : Chained('load')
     );
 }
 
+=head2 map
+
+Shows map of places for an area.
+
+=cut
+
+sub map : Chained('load')
```
loujine: Good point, I'll fix it once I find the other places the name needs to be changed :)

gentlecat:
@brainzbot, ok to test.

root/static/lib/leaflet/leaflet.markercluster-src.js
```diff
@@ -0,0 +1,2616 @@
+/*
```
mwiencek: Let's add "leaflet.markercluster" to package.json instead of importing the code (we generally don't do that with JS libs anymore). Then in common.js, just do `require('leaflet.markercluster/dist/leaflet.markercluster-src')`.

root/components/area-map.tt
```diff
@@ -0,0 +1,100 @@
+<div id="largemap"></div>
+<script type="text/javascript">//<![CDATA[
+
+    [%~ DEFAULT places=entities ~%]
+    var map,
```
mwiencek: Wrap the code in an [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression) like this to avoid creating global variables on the page:

```JavaScript
<script type="text/javascript">//<![CDATA[
(function () {
    // code goes here
}());
//]]></script>
```

root/components/area-map.tt
```diff
@@ -0,0 +1,100 @@
...
+        coordinates = [],
+        places = [];
+
+    [%~ FOR place IN places ~%]
+        [%~ IF place.coordinates.defined ~%]
+            places.push({
+                uri: '[% link_entity(place) | js %]',
+                title: '[% place.name | js %]',
+                type: '[% place.l_type_name %]',
```
mwiencek: That needs ` | js` too.

root/static/lib/leaflet/leaflet.markercluster-src.js
mwiencek: BTW, you can add it to package.json automatically like `npm install --save leaflet.markercluster`. But then you have to edit package.json manually and remove the caret from the version, e.g. if it has "^1.0.0" (which matches any 1.x.x release), change it to "1.0.0", because we prefer exact versions.

loujine:
Fixes:
- integrated map in the existing places tab (there was no reason to use another tab except understanding how the tab system works)
- installed leaflet.markercluster through npm
- added IIFE
- fixed missing "... | js"

I realized the leaflet update to 1.0.1 (required by leaflet.markercluster) broke the marker display in the standard Place map (missing "/" in icon path) and fixed it in  	a4abf3a

loujine:
I guess I'll have to rebase my branch on the current master also?

Zastai:
Rebasing shouldn't be needed since there don't seem to be conflicts.
You may need to look at the failing tests though.

gentlecat:
@brainzbot, test this please.

uklauer:
The test failures are unrelated, BTW.

PR #387, opened 2016-10-30
Add some missing calls to localization wrappers

y-van-z:
Fix localization by adding some missing calls to:
- `add_colon` wrapper, in relationship lists for instance
- (new) `l_status_name` wrapper, in summary of failed, rejected, or cancelled edits, and in raw edit data page of all edits.

### DIFF
--- a/root/edit/info.tt
+++ b/root/edit/info.tt
@@ -1,7 +1,7 @@
 [% IF summary %]
     [% IF edit.status != 1 && edit.status != 2 %]
         <div class="edit-status">
-            [%- edit.status_name -%]
+            [%- edit.l_status_name -%]
         </div>
     [% END %]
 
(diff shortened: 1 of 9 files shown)


root/edit/info.tt
```diff
@@ -1,7 +1,7 @@
 [% IF summary %]
     [% IF edit.status != 1 && edit.status != 2 %]
         <div class="edit-status">
-            [%- edit.status_name -%]
+            [%- l(edit.status_name) -%]
```
uklauer: We usually use `l_...` function variants that provide translated object strings; e.g., there is `l_edit_name` (next to untranslated `edit_name`), `l_name` (next to untranslated `name`) in various classes, etc. So, for consistency, it might be better to create an `l_status_name` function and call it here.

y-van-z: Thanks, it is now done in https://github.com/metabrainz/musicbrainz-server/pull/387/commits/5e954125d9d77110b905bdeec7019e90d96c33c7 and https://github.com/metabrainz/musicbrainz-server/pull/387/commits/4c8dac89549dac469a202fa68f486aa1fcf679bc .

uklauer: Seems good then.

(Hmm, I thought you could switch the status to “recommend merge” or something like that now?)

mwiencek:
:shipit:

PR #389, opened 2016-11-03
MBS-9114: Serialize the ordering type for a Series.

Zastai:


Zastai:
The failing test is due to validating against the current mmd-schema (needs https://github.com/metabrainz/mmd-schema/pull/8 to be applied).

mwiencek:
I'm not sure about this one yet. I actually excluded this intentionally from the WS because it's mostly an internal flag that has no useful purpose outside the code.

Zastai:
Well, I only added it because the mmd2 schema included a very similar sounding property that was not getting serialized.
Having said that, it is getting shown in the UI, so I don't see why it would not be in the WS output. If its only mapping in the UI was a checkbox when editing, then I would have only filed a PR for removing it from mmd2-schema.

uklauer:
Perhaps we should rather remove it from the website.

Zastai:
Should I switch this PR to one for the UI removal then (and with a revert commit or a force push dropping the current commits)?
Or close this one and make a new one for the UI removal?

mwiencek:
Well, there are currently 206 series that are manually ordered. We can't remove it from the UI until those are converted to automatically-ordered ones and we see what the use cases are (automatic ordering isn't 100% perfect).

Zastai:
Well, I was only suggesting its removal from the display pages for a series, not from the series editor, where it still has a use.

mwiencek:
Ah...I'd probably just leave it until we can remove it entirely, but if you think it's less confusing to hide it I don't really mind.

Zastai:
No worries; will just close this PR then.

PR #390, opened 2016-11-06
MBS-9117: Serialize correct trackcount for CDStub.

Zastai:
Don't think this will impact any tests.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/CDStub.pm
+++ b/lib/MusicBrainz/Server/Entity/CDStub.pm
@@ -51,6 +51,8 @@ has 'source' => (
     isa => 'Int'
 );
 
+# Note: This property is used for searches ONLY (because the search results do not set up a full CD Stub TOC).
+#       In all other cases, the track_count property of the associated CDStubTOC applies.
 has 'track_count' => (
     is => 'rw',
     isa => 'Int'
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/CDStubTOC.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/CDStubTOC.pm
@@ -25,7 +25,7 @@ sub serialize
                 length => $_->length
             }, $cdstub->all_tracks
         ],
-        "track-count" => $cdstub->track_count
+        "track-count" => $entity->track_count
     };
 };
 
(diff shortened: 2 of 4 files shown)


mwiencek:
Can you explain what's going on here? I'm not sure I understand why there's a fallback, i.e. when is the first part undefined.

Zastai:
It seems like it's _always_ null when doing WS requests (which is understandable in itself, since I don't think a stub can deviate from the TOC it wraps).
I didn't trace through all the code to look for uses, so I assumed it was in use _somewhere_ and added this fallback when serializing, rather than removing the always-null property.

Zastai:
Looks like track_count is only a property on CDStub for the purposes of search results; if I remove it there (and adjust the WS serialization to only ever use the TOC's track count), everything seems to work, except that the search results page for CD stubs no longer shows the track count.
And there's no property to get a toc from the stub object (the link is in the other direction only), so I did not see an obvious way to change the search results to use the TOC's count.

mwiencek:
Okay, I see. I think it's best to ignore the CDStub's track_count and always use the TOC's. I'm a bit confused by the change you saw in the search results page (maybe I'm misunderstanding what you did). The search result pages use JSON directly from the search server, so none of the changes you're making here would affect it.

Zastai:
What I meant was that I removed the track_count property from CDStub altogether and then looked at all uses. The only uses I found were in the serialization (where it was easily fixed to use the TOC's count instead) and in the displaying of search results. The latter seems to instantiate a CDStub without underlying TOC and shows the stub's track count on the results page. I didn't immediately see how those CDStubs are created so couldn't say how easy it would be to adjust that so that a TOC also gets created, just for the count.

Will revise this patch to just use the toc track count, and adding a comment to the stub's track count saying it's used for search results only.

Would you prefer an added commit for the change, or a force push so that this PR contains only this last set of changes?

mwiencek:
Got it, sorry for misunderstanding. For my own pull requests I always force-push so they only include the necessary commits. So I'd personally prefer that. :)

Zastai:
Thinks this does it then.

I do see a
`"track-count" => $cdstub->track_count,`
in t/lib/t/MusicBrainz/Server/WebService/Serializer/JSON/2/CDStubTOC.pm; perhaps that needs to be `"track-count" => $cdstub_toc->track_count,`?

mwiencek:
Yeah, the test is no longer technically correct and should also be changed. Though it will pass either way, because `track_count` is never set on either object.

gentlecat:
@brainzbot, add to whitelist.

mwiencek:
Thanks! Sorry that it took a while to get this merged.

PR #391, opened 2016-11-08
Fix banner overlap with right floating top menu.

jesus2099:
Server banners overlap and hide the top menu’s second row.
Especially when the window is narrow enough to reproduce this.
Confirmed in old browsers: **Internet Explorer 8** and **Opera 12.18**.
“Unwelcome” with un‐dismiss‐able banners.
Top (horizontal) menu is `float: right;` so I have added `clear: both;` to the banner so it goes below the horizontal menu bar.
`clear: right;` would have been enough, but…

### DIFF
--- a/root/static/styles/extra/banner.less
+++ b/root/static/styles/extra/banner.less
@@ -7,6 +7,7 @@
     background: @effectively-white-bg;
     padding: @form-margin;
     position: relative;
+    clear: both;
 
     @media @wide {
         border-radius: 6px;


jesus2099:
Before (Opera 12) : 
mbs-banner-before

After : 
mbs-banner-after

With Firefox and Chromium and IE, the patch does no changes. Before and After are both OK.

jesus2099:
When I dismiss (close) the banner by clicking the red cross icon button, things react as expected.

jesus2099:
I confirm the same problem with another browser : 

Before (Internet Explorer 8)
-----------------------------------

ie8-css-ng

After (same `.banner { clear: both; }` fix)
-----------------------------------

ie8-css-ok

mwiencek:
Thanks! For the screenshots too. :)

jesus2099:
Thanks you two, @uklauer and @mwiencek for your tremendous work !

PR #392, opened 2016-11-13
MBS-9130: Update ReCAPTCHA to version 2

uklauer:
The first version of ReCAPTCHA, as used to protect account creation, is no longer considered sufficient to protect from modern spam bots. Hence, use version 2 (“NoCAPTCHA”, because in many cases users aren’t shown a traditional CAPTCHA anymore).

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -3,7 +3,7 @@ requires 'perl' => '5.18.2';
 # Mandatory modules
 requires 'Algorithm::Diff'                            => '1.1902';
 requires 'Authen::Passphrase'                         => '0.008';
-requires 'Captcha::reCAPTCHA'                         => '0.97';
+requires 'Captcha::reCAPTCHA'                         => '0.99';
 requires 'Catalyst::Action::RenderView'               => '0.16';
 requires 'Catalyst::Authentication::Credential::HTTP' => '1.016';
 requires 'Catalyst::Plugin::Authentication'           => '0.10023';
--- a/lib/MusicBrainz/Server/Controller/Account.pm
+++ b/lib/MusicBrainz/Server/Controller/Account.pm
@@ -424,7 +424,6 @@ sub register : Path('/register') ForbiddenOnSlaves RequireSSL DenyWhenReadonly
     my $form = $c->form(register_form => 'User::Register');
 
     my $captcha = Captcha::reCAPTCHA->new;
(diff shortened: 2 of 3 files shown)


cpanfile
```diff
@@ -3,7 +3,7 @@ requires 'perl' => '5.18.2';
 # Mandatory modules
 requires 'Algorithm::Diff'                            => '1.1902';
 requires 'Authen::Passphrase'                         => '0.008';
-requires 'Captcha::reCAPTCHA'                         => '0.97';
+requires 'Captcha::reCAPTCHA'                         => '0.99';
```
uklauer: ReCAPTCHA 2 isn’t supported before 0.98; as 0.99 fixes some bugs and is in the cpan snapshot file already anyway, why not use that version directly.

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -434,14 +433,13 @@ sub register : Path('/register') ForbiddenOnSlaves RequireSSL DenyWhenReadonly
...
 
-            $captcha_result = $captcha->check_answer(
+            $valid = $captcha->check_answer_v2(
                 DBDefs->RECAPTCHA_PRIVATE_KEY,
-                $c->req->address, $challenge, $response);
-
-            $valid = $captcha_result->{is_valid};
+                $response, $c->req->address
+                )->{is_valid}
+            unless $response eq '';
```
uklauer: No need to bother Google if the user didn’t submit any response at all (likely a non-JS browser or other client).

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -485,8 +483,8 @@ sub register : Path('/register') ForbiddenOnSlaves RequireSSL DenyWhenReadonly
     }
 
     my $captcha_html = "";
-    $captcha_html = $captcha->get_html(
-        DBDefs->RECAPTCHA_PUBLIC_KEY, $captcha_result, $c->req->secure) if $use_captcha;
+    $captcha_html = $captcha->get_html_v2(DBDefs->RECAPTCHA_PUBLIC_KEY)
+        if $use_captcha;
```
uklauer: Uses HTTPS always now.

mwiencek:
LGTM.

PR #393, opened 2016-11-14
Some fixes related to installation and tests

uklauer:
This is what I noticed when setting up a server, except the `npm` version issue is older, and the difficulty to change server types after the fact came up twice on IRC recently.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -43,6 +43,7 @@ requires 'Email::Sender'                              => '1.300010';
 requires 'Email::Valid'                               => '1.192';
 requires 'Encode::Detect'                             => '1.01';
 requires 'Exception::Class'                           => '1.37';
+requires 'File::Slurp';
 requires 'File::Spec'                                 => '3.40';
 requires 'GnuPG'                                      => '0.19';
 requires 'Hash::Merge'                                => '0.200';
@@ -119,6 +120,7 @@ author_requires 'Test::EOL';
 author_requires 'Test::NoTabs';
 
 test_requires 'Cache::Null';
+test_requires 'Catalyst::Plugin::Static::Simple';
 # Broken in Perl >= 5.22; the one test that requires it is disabled for now.
 # test_requires 'Coro';
 test_requires 'HTML::HTML5::Parser';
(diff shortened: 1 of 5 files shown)


cpanfile
```diff
@@ -43,6 +43,7 @@ requires 'Email::Sender'                              => '1.300010';
 requires 'Email::Valid'                               => '1.192';
 requires 'Encode::Detect'                             => '1.01';
 requires 'Exception::Class'                           => '1.37';
+requires 'File::Slurp';
```
uklauer: For production systems, this is installed as a dependency of `Catalyst::Plugin::ErrorCatcher`, but it is required on other systems, too.

cpanfile
```diff
@@ -119,6 +120,7 @@ author_requires 'Test::EOL';
 author_requires 'Test::NoTabs';
 
 test_requires 'Cache::Null';
+test_requires 'Catalyst::Plugin::Static::Simple';
```
uklauer: This is pulled in by `Catalyst::Devel`, but that module is only `author_requires`.

t/lib/t/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -5,7 +5,7 @@ use Test::More;
 
 use MusicBrainz::Server::Data::Recording;
 use MusicBrainz::Server::Data::Search;
-use encoding 'utf8';
+use utf8;
```
uklauer: There isn’t actually any non-ASCII text in this file, but perhaps someone will add UTF-8 test data later … In any case, `use encoding` is deprecated.

lib/MusicBrainz/Server.pm
```diff
@@ -28,8 +28,6 @@ Session::State::Cookie
 
 Cache
 Authentication
-
-Unicode::Encoding
```
uklauer: No longer necessary since Catalyst 5.90040 (see http://wiki.catalystframework.org/wiki/tutorialsandhowtos/using_unicode.view). Only function now is to emit a deprecation warning.

mwiencek:
Thanks. 👍

PR #395, opened 2016-11-21
MBS-9102: Don’t try to calculate an age when there is not enough data

uklauer:
The `Age` role didn’t correctly recognize an ended entity without an end date (just the `ended` flag) nor the case where partial begin and end dates were present, but one of them didn’t include the year. This led to it thinking the entity was still active (age calculation up to the current date) or trying to calculate with `undef` (leading to an ISE, https://beta.musicbrainz.org/artist/5cc78d71-a007-4f2d-a703-cd479fad89ac vs. http://mb.chirlu.de/artist/5cc78d71-a007-4f2d-a703-cd479fad89ac).

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Role/Age.pm
+++ b/lib/MusicBrainz/Server/Entity/Role/Age.pm
@@ -44,9 +44,9 @@ sub has_age
     my @end_comp = $self->end_date->defined_run;
 
     # Shrink @begin_comp and @end_comp to the same size
-    my $shortest_run = min(scalar(@begin_comp) - 1, scalar(@end_comp) - 1);
-    @begin_comp = @begin_comp[0..$shortest_run];
-    @end_comp = @end_comp[0..$shortest_run];
+    my $shortest_run = min(scalar @begin_comp, scalar @end_comp);
+    splice @begin_comp, $shortest_run;
+    splice @end_comp, $shortest_run;
 
     # Compare all elements that are defined in both @begin_comp and
     # @end_comp to determine if @end_comp is greater than @begin_comp.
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Entity/Role/Age.pm
```diff
@@ -38,13 +38,16 @@ sub has_age
             )
         ) == -1;
 
-    # If there is no end date, then the end date is now() (so there is an age).
-    my @end_comp = $self->end_date->defined_run or return 1;
```
uklauer: This incorrectly thought of entities that have ended at an unknown date as non-ended.

lib/MusicBrainz/Server/Entity/Role/Age.pm
```diff
@@ -38,13 +38,16 @@ sub has_age
...
+    return 1 if !$self->ended;
+
+    my @end_comp = $self->end_date->defined_run;
 
     # Shrink @begin_comp and @end_comp to the same size
-    my $shortest_run = min(scalar(@begin_comp) - 1, scalar(@end_comp) - 1);
-    @begin_comp = @begin_comp[0..$shortest_run];
-    @end_comp = @end_comp[0..$shortest_run];
+    my $shortest_run = min(scalar @begin_comp, scalar @end_comp);
+    return 0 if $shortest_run == 0; # no age if there is not even a year
```
uklauer: This line is the fix for the ISE: If there are (partial) dates, but they don’t both have a year, we can’t calculate an age.

lib/MusicBrainz/Server/Entity/Role/Age.pm
```diff
@@ -38,13 +38,16 @@ sub has_age
...
+    splice @begin_comp, $shortest_run;
+    splice @end_comp, $shortest_run;
```
uklauer: The rest of the changes in this block is just for readability. I didn’t like the `-1`s, and the (0-based) index being called “shortest run”.

Arguably, `@begin_comp` and `@end_comp` still aren’t the best possible names, but …

PR #396, opened 2016-11-22
MBS-9136: Fix the TracksWithSequenceIssues report

uklauer:
The `TracksWithSequenceIssues` report had a number of syntax errors in its SQL and therefore failed to run. Even if it had run, the third part of the test would have included all mediums with pre-gap tracks because it did not account for the additional track.

Rewrite the SQL for the WHERE clause to fix these problems. At the same time, remove the additional check that pre-gap tracks can only be used with certain medium types; while having pre-gap tracks on, e.g., vinyl is also wrong, this is a very different type of problem and should have its own report.

(Because the code from commit e002c4d4830502b314e42d178eba8f29ef8e1cd2 never actually worked, this can’t be considered a regression.)

### DIFF
--- a/lib/MusicBrainz/Server/Report/TracksWithSequenceIssues.pm
+++ b/lib/MusicBrainz/Server/Report/TracksWithSequenceIssues.pm
@@ -7,8 +7,7 @@ with 'MusicBrainz::Server::Report::ReleaseReport',
 
 sub query {
     # There are 3 checks going on in this query:
-    # 1. The first track should be '1' or possibly '0' if the medium format
-    #    supports disc IDs.
+    # 1. The first track should be '1' or possibly '0' (pre-gap track).
     # 2. The last track should match the amount of tracks (minus 1 if there's
     #    a pregap track 0).
     # 3. The sum of all tracks should match a standard arithmetic progression.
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Report/TracksWithSequenceIssues.pm
```diff
@@ -35,12 +34,11 @@ FROM
...
    ) s
    JOIN medium ON medium.id = s.medium
-   LEFT JOIN medium_format ON medium_format.id = medium.format
    JOIN release ON release.id = medium.release
    WHERE
-     (first_track != 1 AND NOT (first_track = 0 AND (medium_format.id IS NULL OR medium_format.has_discids))
-     OR (last_track != s.track_count AND NOT (last_track == s.track_count - 1 AND (medium_format.id IS NULL OR medium_format.has_discids))
-     OR (s.track_count * (1 + s.track_count)) / 2 <> track_pos_acc
+     (first_track <> 1 AND first_track <> 0)
+     OR last_track <> s.track_count - (1 - first_track)
```
uklauer: Equivalent to `(first_track = 1 AND last_track = track_count) OR (first_track = 0 AND last_track = track_count - 1)`.

lib/MusicBrainz/Server/Report/TracksWithSequenceIssues.pm
```diff
@@ -35,12 +34,11 @@ FROM
...
+     OR (last_track * (1 + last_track)) <> 2 * track_pos_acc
```
uklauer: Using `last_track` instead of `track_count` because it isn’t influenced by the presence or absence of a pre-gap track. It would be possible to replace `last_track` (twice) with `s.track_count - (1 - first_track)`.

uklauer:
I made [MBS-9137](http://tickets.metabrainz.org/browse/MBS-9137) for the report about mediums with pre-gap tracks on inappropriate medium types.

PR #399, opened 2016-11-24
MBS-7654: Remove “Infer track length from recording” option

uklauer:
The release editor used to have an option to infer missing track lengths from the length of the associated recording. Now, however, this means a circular reasoning because recording lengths are in turn computed from the lengths of the tracks they appear on. This option is therefore removed.


gentlecat:
@brainzbot, add to whitelist.

PR #400, opened 2016-11-26
MBS-9142: Return unmatched variable in link phrase

y-van-z:
Previously, translated (or not) link phrase that contains a variable
which is not an attribute of its relationship/link type would break
the relationship editing dialog in entity editor.

This patch returns the link phrase with the variable name untouched,
rather than crashing the JavaScript interpreter on the next condition.


gentlecat:
@brainzbot, add to whitelist.

PR #404, opened 2016-12-05
MBS-7009: Improve error message for missing replication info

uklauer:
When a server is of type `RT_SLAVE`, but there is no information in the `replication_control` table, it used to crash deep inside Catalyst with a difficult-to-understand error message (because `$c->stash` was called with only one parameter). Catch this situation and die explicitly with a clearer explanation.


mwiencek:
🚢

PR #405, opened 2016-12-07
MBS-1352: Different icon for different artist types on Artist pages

dpmittal:
Display different icons for different artist types on the top of the artist page. Earlier on all artist pages only artist icon was displayed.
Issue Link : https://tickets.metabrainz.org/browse/MBS-1352

### DIFF
--- a/root/artist/header.tt
+++ b/root/artist/header.tt
@@ -1,4 +1,4 @@
-<div class="artistheader">
+<div class="artistheader[% ' ' _ artist.type.name.lower _ '-icon' IF artist.type %]">
     <h1>[% link_entity(artist) %]</h1>
     <p class="subheader">
         <span class="prefix">~</span> [% html_escape(artist.l_type_name) or l('Artist') %]
--- a/root/static/styles/entity-icons.less
+++ b/root/static/styles/entity-icons.less
@@ -8,7 +8,13 @@
 }
 
 .blankheader      { background-image: data-uri('../images/entity/blank.svg') }
-.artistheader     { background-image: data-uri('../images/entity/artist.svg') }
+.artistheader     { background-image: data-uri('../images/entity/artist.svg');
+                    &.choir-icon      { background-image: data-uri('../images/entity/choir.svg') }
+                    &.group-icon      { background-image: data-uri('../images/entity/group.svg') }
+                    &.orchestra-icon  { background-image: data-uri('../images/entity/orchestra.svg') }
+                    &.person-icon     { background-image: data-uri('../images/entity/person.svg') }
+                    &.character-icon  { background-image: data-uri('../images/entity/character.svg') }
+                  }
 .releaseheader    { background-image: data-uri('../images/entity/release.svg') }
 .rgheader         { background-image: data-uri('../images/entity/release_group.svg') }
 .recordingheader  { background-image: data-uri('../images/entity/recording.svg') }


root/artist/header.tt
```diff
@@ -1,4 +1,6 @@
-<div class="artistheader">
+[% header = (html_escape(artist.l_type_name) or l('Artist')) _ 'header' %]
```
y-van-z: Two issues here:
- Localized type name (using either `l_` prefix or `l` function) for class identifier will break for non-English users
- *Other* type is not handled properly
I suggest to combine classes rather than to replace them. Stylesheet will be cleaner (Only need to set `background-image`.) and more robust (Other cases will be auto-handled.).

dpmittal:
@y-van-z  I have changed the following files again trying to implement your suggestions as per my understanding

root/static/styles/entity-icons.less
```diff
@@ -8,7 +8,13 @@
 }
 
 .blankheader      { background-image: data-uri('../images/entity/blank.svg') }
-.artistheader     { background-image: data-uri('../images/entity/artist.svg') }
+.artistheader     { &.artistico     { background-image: data-uri('../images/entity/artist.svg') }
```
y-van-z: `artist.svg` should be available as a fallback, that is, the default `background-image` for the class `artistheader`. The class `artistico` would be unnecessary then and should be removed. That would also be used for artist with either no type or `other` type (or any further unknown type). The rest of changes to this file is fine. 👍

root/artist/header.tt
```diff
@@ -1,10 +1,11 @@
-<div class="artistheader">
+[% header = 'artistheader ' _ (html_escape(artist.l_type_name) or l('Artist')) _ 'ico' %]
```
y-van-z: Remove `or`…`)` as `artistico` becomes unnecessary (see comment to other file).
Do not forget to remove the undue localization prefix `l_` (see previous review).
Possibly move the `artistheader ` class below since it is already lowercase.

mwiencek: Can we append `-icon` instead of `ico`? It's easier to read and there's no reason to abbreviate it.

root/artist/header.tt
```diff
@@ -1,10 +1,11 @@
...
+<div class="[% header.lower() %]">
+
```
mwiencek: Remove this newline.

root/artist/header.tt
```diff
@@ -1,10 +1,11 @@
...
     <h1>[% link_entity(artist) %]</h1>
     <p class="subheader">
         <span class="prefix">~</span> [% html_escape(artist.l_type_name) or l('Artist') %]
     </p>
 </div>
-
```
mwiencek: Add back this newline. :)

dpmittal:
@y-van-z I have made the artistheader attribute default and @mwiencek  I have removed that "div class ..lower" new line and integrated it with default attribute

root/artist/header.tt
```diff
@@ -1,11 +1,11 @@
-<div class="artistheader">
-    <h1>[% link_entity(artist) %]</h1>
+[% header = artist.type_name _ 'icon' %]
+<div class = "artistheader [% header.lower() %]">
+  <h1>[% link_entity(artist) %]</h1>
     <p class="subheader">
-        <span class="prefix">~</span> [% html_escape(artist.l_type_name) or l('Artist') %]
+        <span class="prefix">~</span> [% html_escape(artist.l_type_name) or 'Artist' %]
```
reosarevok: Why remove the l()? That's what makes it translatable.

dpmittal:
@reosarevok I have added back the l() function :) Sorry for that mistake

root/artist/header.tt
```diff
@@ -1,10 +1,10 @@
-<div class="artistheader">
-    <h1>[% link_entity(artist) %]</h1>
+[% header = artist.type_name _ 'icon' %]
+<div class = "artistheader [% header.lower() %]">
```
mwiencek: You don't need the temporary variable, actually. You can just do this:
```
<div class="artistheader [% artist.type_name.lower() _ '-icon' %]">
```

Note: Remove the spaces around the `=` sign, as I did there. Also, in my previous comment I included a hyphen in `-icon`, which is what I meant by easier to read. I think variables consisting of multiple words should usually always use underscores/hyphens, though I know the current CSS isn't consistent about that.

root/artist/header.tt
```diff
@@ -1,10 +1,10 @@
...
+  <h1>[% link_entity(artist) %]</h1>
```
mwiencek: The indentation of this line was changed. :) Add back the two spaces it had (four total).

root/artist/header.tt
mwiencek: This newline still needs to be added back.

root/artist/header.tt
mwiencek: This also doesn't handle artists that have no type. You need to add `IF artist.type_name` at the end of the expression.

dpmittal:
@mwiencek I have rectified all the mistakes and implemented all the suggestions

root/artist/header.tt
```diff
@@ -1,4 +1,4 @@
-<div class="artistheader">
+<div class="artistheader [% artist.type_name.lower() _ '-icon'  IF artist.type.name %]">
```
mwiencek: There's an extra space before the `IF`.

root/static/styles/entity-icons.less
```diff
@@ -8,7 +8,13 @@
 }
 
 .blankheader      { background-image: data-uri('../images/entity/blank.svg') }
-.artistheader     { background-image: data-uri('../images/entity/artist.svg') }
+.artistheader     {  background-image: data-uri('../images/entity/artist.svg');
```
mwiencek: There's an extra space before `background-image` (should be 1 space).

root/static/styles/entity-icons.less
```diff
@@ -8,7 +8,13 @@
...
+                    &.choir-icon      { background-image: data-uri('../images/entity/choir.svg')  }
```
mwiencek: This line and all the other -icon lines except character-icon have an extra space before the closing `  }` (should be 1 space).

dpmittal:
@mwiencek I have removed all the unnecessary whitespaces

root/artist/header.tt
```diff
@@ -1,4 +1,4 @@
-<div class="artistheader">
+<div class="artistheader [% artist.type_name.lower() _ '-icon' IF artist.type.name %]">
```
uklauer: Two aesthetic issues:
- When there is no type, this will result in `class="artistheader "`. While that works fine, the space is a bit annoying. You could change it into `[%~ ' ' _ artist....` – the tilde will cause TT to drop all preceding whitespace, and a space will be output together with the type name if it exists.
- While TT VMethods without parameters optionally can have an empty pair of parentheses, it is more common to omit it. So `.lower()` should preferably be just `.lower`.

y-van-z: In the same idea, note that `.name` at the end of `IF artist.type.name` is unnecessary and thus should preferably be removed. Indeed, `artist.type.name` is undefined if and only if `artist.type` is undefined. See [`type` and `type_name` definitions](https://github.com/dpmittal/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Entity/Role/Type.pm#L22-L30) for reference.

dpmittal:
@uklauer @y-van-z  I have made the necessary changes

PR #407, opened 2016-12-12
MBS-8359: Add 'Guess Case' function for Event names

dpmittal:
Adding the "Guess case" feature to event names while editing .

Issue Link : https://tickets.metabrainz.org/browse/MBS-8359

### DIFF
--- a/root/static/scripts/guess-case/MB/GuessCase/Main.js
+++ b/root/static/scripts/guess-case/MB/GuessCase/Main.js
@@ -119,9 +119,10 @@ const flags = require('../../flags');
         sortname: guess("Work", "guessSortName")
     };
 
-    // Series doesn't have it's own handler, and just uses the work handler
-    // because additional behavior isn't needed.
+    // Series and Event don't have their own handler, and they use the
+    // work handler because additional behavior isn't needed.
     MB.GuessCase.series = MB.GuessCase.work;
+    MB.GuessCase.event = MB.GuessCase.work;
 
     // lol
     MB.GuessCase.instrument = {
--- a/root/event/edit_form.tt
+++ b/root/event/edit_form.tt
@@ -10,7 +10,7 @@
 
       <fieldset>
         <legend>[% l('Event Details') %]</legend>
-        [%- form_row_text_long(r, 'name', l('Name:')) -%]
+        [%- form_row_name_with_guesscase(r) -%]
         [%- form_row_text_long(r, 'comment', l('Disambiguation:')) -%]
         [%- form_row_select(r, 'type_id', l('Type:')) -%]
         [% form_row_checkbox(r, 'cancelled', l('This event was cancelled.')) %]
@@ -53,6 +53,8 @@
   </div>
 
 </form>
+
+[%- guesscase_options() -%]
 
 <script type="text/javascript">//<![CDATA[
   (function () {


root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -1,7 +1,6 @@
 /*
    This file is part of MusicBrainz, the open internet music database.
-   Copyright (c) 2005 Stefan Kestenholz (keschte)
```
Freso: You shouldn't remove keschte's copyright line, unless you've removed all the code he's written in the file.

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -1,7 +1,6 @@
...
-   Copyright (C) 2010 MetaBrainz Foundation
+   Copyright (C) 2016 MetaBrainz Foundation
```
Freso: Append the year instead of replace it. E.g., "2010, 2016" or "2010–2016" (look at how it's done in other files). (Or just don't alter the copyright line at all. Wait and see what @bitmap or @gentlecat say.)

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -123,6 +122,10 @@ const flags = require('../../flags');
     // because additional behavior isn't needed.
     MB.GuessCase.series = MB.GuessCase.work;
 
+    // Event doesn't have it's own handler, and just uses the work handler
+    // because additional behavior isn't needed.
```
Freso: I would probably amend the Series one to say something like "Series and events don't …" and just add…

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -123,6 +122,10 @@ const flags = require('../../flags');
...
+    MB.GuessCase.event = MB.GuessCase.work;
```
Freso: … this below the `MB.GuessCase.series = MB.GuessCase.work;` line.

But again, @bitmap or @gentlecat might know better. :)

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -119,10 +119,12 @@ const flags = require('../../flags');
         sortname: guess("Work", "guessSortName")
     };
 
-    // Series doesn't have it's own handler, and just uses the work handler
+    // Series and Event doesn't have it's own handler, and just uses the work handler
     // because additional behavior isn't needed.
     MB.GuessCase.series = MB.GuessCase.work;
 
+    MB.GuessCase.event = MB.GuessCase.work;
```
uklauer: You shouldn’t have an empty line between these two assignments.

root/static/scripts/guess-case/MB/GuessCase/Main.js
uklauer: Definitely not “2010–2016”, because that would mean that a copyrightable change has been made in each of the years 2010, 2011, 2012, 2013, 2014, 2015, and 2016.

In this case, I agree that the change is too minor to justify extending the copyright statement in whichever way.

dpmittal:
@uklauer I have made the necessary changes :)

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -119,9 +119,10 @@ const flags = require('../../flags');
         sortname: guess("Work", "guessSortName")
     };
 
-    // Series doesn't have it's own handler, and just uses the work handler
+    // Series and Event doesn't have it's own handler, and just uses the work handler
```
Freso: You added another entity type, but the rest of the sentence is still in singular.

Also, you might want to wrap the line. Not sure how long it is now or if MBS has a suggested/hard line length limit. Either way, this is easy to wrap, so let's do it. :)

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -119,9 +119,10 @@ const flags = require('../../flags');
...
     // because additional behavior isn't needed.
     MB.GuessCase.series = MB.GuessCase.work;
+    MB.GuessCase.event = MB.GuessCase.work;
```
Freso: Much better!

root/static/scripts/guess-case/MB/GuessCase/Main.js
uklauer: Yeah, I guess you could move “handler” to the next line, although we don’t really have a hard limit.

While you are touching the comment, you could also correct “it's own handler” into “its own handler”.

dpmittal:
@Freso Made the changes

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -119,9 +119,9 @@ const flags = require('../../flags');
         sortname: guess("Work", "guessSortName")
     };
 
-    // Series doesn't have it's own handler, and just uses the work handler
-    // because additional behavior isn't needed.
+    // Series and Event doesn't have their own handler, and they use the work handler because additional behavior isn't needed.
```
uklauer: So, you missed one singular here (`doesn't`). :)

And it should remain in two lines; the question was only where to break it. You could do it after “use the work handler” (as before) or after “use the”.

dpmittal:
@uklauer Corrected the comment

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -119,9 +119,10 @@ const flags = require('../../flags');
         sortname: guess("Work", "guessSortName")
     };
 
-    // Series doesn't have it's own handler, and just uses the work handler
-    // because additional behavior isn't needed.
+    // Series and Event dont have their own handler, and they use the
```
Freso: `dont`→`don't`

dpmittal: Sorry for that silly mistake , I should have proofread it. I have made the changes

gentlecat:
@brainzbot, add to whitelist.

root/static/scripts/guess-case/MB/GuessCase/Main.js
mwiencek: P.S. My github handle is mwiencek. ;)

PR #408, opened 2016-12-12
MBS-8268: Fix an issue with rating stars display

uklauer:
Short description of the problem:

> Rating links became grey when having focus, such as after a mouse click, partially covering the current rating and the other rating links; in many cases, it appeared to the user as if clicking had not had any effect. The regular display was restored only when the user clicked elsewhere, thereby removing the focus from the rating link.

The details are a bit complicated, please see the individual commits for them. The first commit only does a refactoring of the existing CSS for ratings to make it more LESS, without actually changing anything. The second commit is the actual fix, with a detailed explanation. The third commit is mostly unrelated and removes some CSS statements that were last needed for Firefox 2.

### DIFF
--- a/root/components/rating-macros.tt
+++ b/root/components/rating-macros.tt
@@ -16,7 +16,7 @@ END -%]
 
 [%- MACRO rating_stars(entity, prevent_rating) BLOCK -%]
 <span class="inline-rating">
-    <span class="star-rating">
+    <span class="star-rating" tabindex="-1">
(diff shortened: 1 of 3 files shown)


uklauer:
You can try the fixed version on http://mb.chirlu.de/artist/5cc78d71-a007-4f2d-a703-cd479fad89ac

jesus2099:
Thank you @uklauer, it does work really good on your sandbox ! :clap: :bowing_woman:

PR #409, opened 2016-12-14
MBS-9143: Remove the FreeDB search option

dpmittal:
FreeDB search was dead and it wasn’t coming back due to server changes so it should be removed from the UI as well as the subsequent code was needed to be removed . The following features are removed : - 

- Search from the search bar in the top

- Search by Link

- Import FreeDB Search from adding release tracklist

- <del> Search by FreeDB ID on the search page</del> Added back to functionality after discussion
Issue link : https://tickets.metabrainz.org/browse/MBS-9143


root/otherlookup/form.tt
```diff
@@ -9,7 +9,6 @@
                    , [ 'artist-isni', l('Artist ISNI:') ]
                    , [ 'label-ipi', l('Label IPI:') ]
                    , [ 'label-isni', l('Label ISNI:') ]
-                   , [ 'freedbid', l('FreeDB ID:') ]
```
uklauer: This has nothing to do with the external search.

uklauer:
We may need to decide whether we want to keep the import functionality. It might still make sense to be able to import FreeDB entries for which you know the ID?

y-van-z:
Copy/paste of FreeDB ID (to _FreeDB import_ tab) would not be any good compared to already functional copy/paste of corresponding track list (to _Manual entry_ tab), e.g. from www.freedb.org. IMHO, adding such functionality is beyond the scope of this issue.

root/otherlookup/form.tt
y-van-z: Here is an example: https://musicbrainz.org/otherlookup/freedbid?other-lookup.freedbid=bc0b120d

dpmittal:
@y-van-z @uklauer I have reverted the changes of otherlookups/forms

uklauer:
As I said above, *This leaves a lot behind that should also go*.

dpmittal:
@uklauer @y-van-z  Updated the PR : I have tried to remove all of the unused code excluding the one necessary for FreeDB ID search :)

mwiencek:
The (small) `link_freedb` MACRO in root/components/common-macros.tt will also be unused following these changes. But it looks like everything else is covered.

dpmittal:
@mwiencek I have removed the `link_freedb` MACRO from root/components/common-macros.tt :)

mwiencek:
Thanks, @dpmittal 👍

PR #410, opened 2016-12-20
MBS-1319 : Add Turkish Guess Case

dpmittal:
Adding Turkish Guess Case to Guess Case editor 

Issue Link : https://tickets.metabrainz.org/browse/MBS-1319

### DIFF
--- a/root/static/scripts/guess-case/modes.js
+++ b/root/static/scripts/guess-case/modes.js
@@ -128,6 +128,14 @@ let DefaultMode = {
     );
   },
 
+  toUpperCase(str) {
+    return str.toUpperCase();
+  },
+
+  toLowerCase(str) {
+    return str.toLowerCase();
+  },
+
   /*
    * Pre-process to find any lowercase_bracket word that needs to be put into
    * parentheses. Starts from the back and collects words that belong into the
--- a/root/static/scripts/guess-case/modes.js
+++ b/root/static/scripts/guess-case/modes.js
@@ -8,7 +8,7 @@ const {assign} = require('lodash');
 
 const {l} = require('../common/i18n');
 const getBooleanCookie = require('../common/utility/getBooleanCookie');
-const {isPrepBracketWord, isPrepBracketSingleWord} = require('./utils');
+const {isPrepBracketWord, isPrepBracketSingleWord, turkishUpperCase, turkishLowerCase} = require('./utils');
 
 /*
  * Words which are always written lowercase.
@@ -17,8 +17,9 @@ const {isPrepBracketWord, isPrepBracketSingleWord} = require('./utils');
  * keschte  2005-04-17  added french lowercase characters
  * keschte  2005-06-14  added "tha" to be handled like "the"
  * warp     2011-02-01  added da, de, di, fe, fi, ina, inna
+ * dpmittal 2016-12-20  added Turkish lowercase words
  */
-const LOWER_CASE_WORDS = /^(a|an|and|as|at|but|by|da|de|di|fe|fi|for|in|ina|inna|n|nor|o|of|on|or|tha|the|to)$/;
+const LOWER_CASE_WORDS = /^(a|an|and|as|at|but|by|da|de|di|fe|fi|for|in|ina|inna|n|nor|o|of|on|or|tha|the|to|ve|ile|ya|veya|yahut|ki|mı|mi|mu|mü|mısın|musun|mudur|mıdır|midir|miyim|misin|misiniz|mısınız|müsün|müyüm|müsünüz|mıyım|muyum|musunuz|miyiz|mıyız|muyuz|müyüz|müdür)$/;
 
 /*
  * Words which are always written uppercase.
@@ -289,4 +290,21 @@ exports.Sentence = assign({}, DefaultMode, {
     'language guidelines}.',
     {url: {href: 'https://musicbrainz.org/doc/Style/Language', target: '_blank'}}
   ),
+});
+
+exports.Turkish = assign({}, DefaultMode, {
+  description: l(
+    'This mode handles the Turkish capitalisation of \'i\' (\'İ\') and \'ı\' '+
+    '(\'I\'). Some words may need to be manually corrected '+
+    'according to the {url|Turkish language guidelines}. ',
+    {url: {href: 'https://musicbrainz.org/doc/Style/Language/Turkish', target: '_blank'}}
+  ),
+
+  isSentenceCaps() {
+    return false;
+  },
+
+  toUpperCase:turkishUpperCase,
+
+  toLowerCase:turkishLowerCase,
 });
--- a/root/static/scripts/guess-case/utils.js
+++ b/root/static/scripts/guess-case/utils.js
@@ -87,6 +87,14 @@ var lowerCaseBracketWordsList = [
 
 var lowerCaseBracketWords = new RegExp('^(' + lowerCaseBracketWordsList.join('|') + ')$', 'i');
 
+exports.turkishUpperCase = function(str) {
+  return str.replace(/i/g, 'İ').toUpperCase();
+};
+
+exports.turkishLowerCase = function(str) {
+  return str.replace(/I\u0307/g, 'i').replace(/I/g, 'ı').replace(/İ/g, 'i').toLowerCase();
+};
+
 exports.isLowerCaseBracketWord = function (w) {
     return lowerCaseBracketWords.test(w);
 };
--- a/root/static/scripts/guess-case/utils.js
+++ b/root/static/scripts/guess-case/utils.js
@@ -144,8 +144,8 @@ exports.titleString = function (is, forceCaps) {
     }
 
     var os;
-    var lc = is.toLowerCase();
-    var uc = is.toUpperCase();
+    var lc = gc.mode.toLowerCase(is);
+    var uc = gc.mode.toUpperCase(is);
 
     if (is === uc && is.length > 1 && gc.CFG_UC_UPPERCASED) {
         os = uc;
@@ -174,8 +174,8 @@ exports.titleString = function (is, forceCaps) {
         os = uc;
     } else {
         os = exports.titleStringByMode(lc, forceCaps);
-        lc = os.toLowerCase();
-        uc = os.toUpperCase();
+        lc = gc.mode.toLowerCase(os);
+        uc = gc.mode.toUpperCase(os);
 
         var nextWord = gc.i.getNextWord();
         var followedByPunctuation = nextWord && nextWord.length === 1 && exports.isPunctuationChar(nextWord);
@@ -199,7 +199,7 @@ exports.titleStringByMode = function (is, forceCaps) {
         return '';
     }
 
-    var os = is.toLowerCase();
+    var os = gc.mode.toLowerCase(is);
 
     // See if the word before is a sentence stop character.
     // -- http://bugs.musicbrainz.org/ticket/40
@@ -219,11 +219,11 @@ exports.titleStringByMode = function (is, forceCaps) {
 
     if (doCaps) {
         var chars = os.split('');
-        chars[0] = chars[0].toUpperCase();
+        chars[0] = gc.mode.toUpperCase(chars[0]);
 
         if (is.length > 2 && is.substring(0, 2) === 'mc') {
             // Make it McTitled
-            chars[2] = chars[2].toUpperCase();
+            chars[2] = gc.mode.toUpperCase(chars[2]);
         }
 
         os = chars.join('');
--- a/root/static/scripts/tests/GuessCase.js
+++ b/root/static/scripts/tests/GuessCase.js
@@ -150,7 +150,7 @@ test('Recording', function (t) {
 });
 
 test('Work', function (t) {
-    t.plan(7);
+    t.plan(11);
 
     var tests = [
         {
@@ -187,6 +187,26 @@ test('Work', function (t) {
             input: "Bring It All To Me (f. 50 Cent)",
             expected: "Bring It All to Me (feat. 50 Cent)",
             mode: "English", roman: true, keepuppercase: false
+        },
+        {
+            input: "izarın gül gül olmuş",
+            expected: "İzarın Gül Gül Olmuş",
+            mode: "Turkish", roman: false, keepuppercase: false
+        },
+        {
+            input: "efendim hu nasibim Bu IECELLİ TAKSI\u0307RAT yahu",
+            expected: "Efendim Hu Nasibim Bu Iecelli Taksirat Yahu",
+            mode: "Turkish", roman: false, keepuppercase: false
+        },
+        {
+            input: "ben De YAZDIM",
+            expected: "Ben de YAZDIM",
+            mode: "Turkish", roman: false, keepuppercase: true
+        },
+        {
+            input: "ya devlet başa Ya Kuzgun Leşe",
+            expected: "Ya Devlet Başa ya Kuzgun Leşe",
+            mode: "Turkish", roman: false, keepuppercase: false
         }
     ];
 
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -231,6 +231,7 @@
         <option value="English">[% l('English') %]</option>
         <option value="Sentence">[% l('Sentence') %]</option>
         <option value="French">[% l('French') %]</option>
+        <option value="Turkish">[% l('Turkish') %]</option>
       </select>
       <br />
       <label>
@@ -254,6 +255,7 @@
           <option value="English">[% l('English') %]</option>
           <option value="Sentence">[% l('Sentence') %]</option>
           <option value="French">[% l('French') %]</option>
+          <option value="Turkish">[% l('Turkish') %]</option>
         </select>
         <br />
         <label>


root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +283,27 @@ exports.Sentence = assign({}, DefaultMode, {
...
+  ),
+
+  isSentenceCaps() {
+    return false;
+  },
+
+  runPostProcess(is) {
+    return DefaultMode.runPostProcess(is)
+      .replace(/I/g, 'i') //replace 'I' with 'i' as 'I' is not valid in Turkish
+      //replace of Turkish lowercase words if they have been used at the end of sentence and capitalised
```
y-van-z: `I` is a valid character in Turkish, but it is the upper case of `ı` (_dotless i_) whereas `İ` (_dotted I_) is the upper case of `i`. See https://musicbrainz.org/doc/Style/Language/Turkish

root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +283,27 @@ exports.Sentence = assign({}, DefaultMode, {
...
+      .replace(/Mı/g, 'mı')
+      .replace(/Mi/g, 'mi')
+      .replace(/Mu/g, 'mu')
+      .replace(/Mü/g, 'mü');
```
y-van-z: What is the purpose of these replacements?
Would it be fine with [Mevlevi Ayin Müzikleri](http://musicbrainz.org/recording/d051add7-76fe-42dc-981f-1cacf8570ad0)?

root/static/scripts/guess-case/modes.js
```diff
@@ -17,8 +17,9 @@ const {isPrepBracketWord, isPrepBracketSingleWord} = require('./utils');
  * keschte  2005-04-17  added french lowercase characters
  * keschte  2005-06-14  added "tha" to be handled like "the"
  * warp     2011-02-01  added da, de, di, fe, fi, ina, inna
+ * dpmittal 2016-12-20  added turkish lowercase characters
  */
-const LOWER_CASE_WORDS = /^(a|an|and|as|at|but|by|da|de|di|fe|fi|for|in|ina|inna|n|nor|o|of|on|or|tha|the|to)$/;
+const LOWER_CASE_WORDS = /^(a|an|and|as|at|but|by|da|de|di|fe|fi|for|in|ina|inna|n|nor|o|of|on|or|tha|the|to|ve|ile|ya|veya|yahut|ki|mı|mi|mu|mü|mısın|musun|mudur|mıdır|midir|miyim|misin|misiniz|mısınız|müsün|müyüm|müsünüz)$/;
```
y-van-z: I am not fond of mixing all languages in the same constant, but that probably is a separate issue.

y-van-z:
@dpmittal: Here is a bit of additional guidance because using `runPostProcess` is not in the right way, deeper code changes are required. The root problem is that `'i'.toUpperCase()` returns `'I'` whereas `'İ'` is expected in Turkish. Unfortunately, it cannot be replaced by `toLocaleUpperCase` which relies on browser’s locale only. Instead, a Turkish-friendly alternative to `toUpperCase` (and `toLowerCase`) has to be provided at first, then control code has to be added to make use of this alternative in `Turkish` mode only.

dpmittal:
@y-van-z  I have made a new function for Turkish Guess Case and added tests too :)

dpmittal:
@y-van-z I have made the necessary changes, it's properly functional for all modes :D

root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +287,21 @@ exports.Sentence = assign({}, DefaultMode, {
     {url: {href: 'https://musicbrainz.org/doc/Style/Language', target: '_blank'}}
   ),
 });
+
+exports.Turkish = assign({}, DefaultMode, {
+  description: l(
+    'This mode capitalises almost all words of a sentence, the words ' +
+    'which are restricted to be capitalised are lowercased. Some words, often proper nouns, ' +
```
y-van-z: 1. Please remove `of a sentence` to prevent mix-up with _Sentence_ mode.
2. Moreover, `the words which are restricted to be capitalised are lowercased` is not very informative. Actually, just reuse the wording from _English_ mode, except that `connectives and question-suffixes` (taken from https://musicbrainz.org/doc/Style/Language/Turkish#In_English) should replace `articles and short prepositions`.
3. Finally, please mention that both `İ` and `ı` letters are handled.

root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +287,21 @@ exports.Sentence = assign({}, DefaultMode, {
...
+    'may need to be manually fixed according to the {url|Turkish ' +
+    'language guidelines}.',
+    {url: {href: 'https://musicbrainz.org/doc/Style/Language/Turkish', target: '_blank'}}
+  ),
+
+  isSentenceCaps() {
+    return false;
+  },
+
+  isTurkishMode(){
```
y-van-z: What about defining functions `toLowerCase(s)` and `toUpperCase(s)` instead of `ìsTurkishMode()`?

root/static/scripts/guess-case/utils.js
```diff
@@ -219,11 +257,17 @@ exports.titleStringByMode = function (is, forceCaps) {
...
+        else
+          chars[0] = chars[0].toUpperCase();
 
         if (is.length > 2 && is.substring(0, 2) === 'mc') {
             // Make it McTitled
-            chars[2] = chars[2].toUpperCase();
+            if(gc.mode.isTurkishMode())
+              chars[2] = TurkishUpperCase(chars[2]);
+            else
+              chars[2] = chars[2].toUpperCase();
```
y-van-z: While this works just fine, it has the drawback of cluttering the code up with `if`…`else` conditions.
IMHO, a more elegant way to do it would be to make a single call `gc.mode.toUpperCase(chars[2])` instead.
See related comment.

dpmittal:
I have made the changes :)

root/static/scripts/guess-case/utils.js
```diff
@@ -87,6 +87,30 @@ var lowerCaseBracketWordsList = [
 
 var lowerCaseBracketWords = new RegExp('^(' + lowerCaseBracketWordsList.join('|') + ')$', 'i');
 
+exports.turkicUpperCase = function(str) {
```
mwiencek: This should be equivalent to the following, right?

```JavaScript
exports.turkicUpperCase = function (str) {
    return str.replace(/i/g, 'İ').toUpperCase();
};
```

root/static/scripts/guess-case/utils.js
```diff
@@ -87,6 +87,30 @@ var lowerCaseBracketWordsList = [
...
+  {
+    if (str.charAt(i)=='i')
+      fc += 'İ';
+    else
+      fc += str.charAt(i).toUpperCase();
+  }
+  return fc;
+};
+
+exports.turkicLowerCase = function(str) {
```
mwiencek:
```JavaScript
exports.turkicLowerCase = function (str) {
    return str.replace(/I/g, 'ı').toLowerCase();
};
```

root/static/scripts/guess-case/modes.js
```diff
@@ -8,7 +8,7 @@ const {assign} = require('lodash');
 
 const {l} = require('../common/i18n');
 const getBooleanCookie = require('../common/utility/getBooleanCookie');
-const {isPrepBracketWord, isPrepBracketSingleWord} = require('./utils');
+const {isPrepBracketWord, isPrepBracketSingleWord, turkicUpperCase, turkicLowerCase} = require('./utils');
```
uklauer: The word _Turkic_ doesn’t make any sense here. See https://www.quora.com/Whats-the-difference-between-Turkic-and-Turkish

root/static/scripts/guess-case/utils.js
y-van-z: Absolutely!

root/static/scripts/guess-case/modes.js
uklauer: The following harmonic variants of the question particles are missing: `mıyım`, `muyum`, `musunuz`. The first person plural is missing altogether: `miyiz`, `mıyız`, `muyuz`, `müyüz`.

Regarding `mudur|mıdır|midir`, `müdür` is again missing; however, that might be intentional, because there is a homophone noun (meaning _headmaster_). I can’t judge whether it makes sense to break consistency here, and we should probably ask someone who speaks Turkish properly.

y-van-z:
There seems to be an additional issue with the lower case function, see the test results for `efendim hu nasibim Bu IECELLİ taksirat yahu` in [build #44](https://ci.metabrainz.org/job/musicbrainz-server/44/).

root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +291,27 @@ exports.Sentence = assign({}, DefaultMode, {
...
 });
+
+exports.Turkish = assign({}, DefaultMode, {
+  description: l(
+    'This mode capitalises almost all words, the connectives ' +
+    'and the question suffixes used in a title are lowercased.'+
+    'In Turkish, capital of \'i\' is \'İ\' and of \'ı\' is \'I\' and'+
+    'this mode handles these. Some words, often proper nouns, ' +
+    'may need to be manually fixed according to the {url|Turkish ' +
+    'language guidelines}.',
```
uklauer: This needs language review from a native speaker. I’m heavily doubtful of the use of _capital_ here; also, the first sentence is misleading: You first read “This mode capitalises almost all words, the connectives and the question suffixes used in a title” and it sounds like a list of three items; only then you stumble over the “are lowercased” and see that you need to reparse the sentence.

root/static/scripts/guess-case/modes.js
uklauer: @sertansenturk, perhaps you could weigh in here?

uklauer:
Regarding the test failure pointed out by @y-van-z above, it is due to the input string containing a decomposed `CAPITAL LETTER I WITH DOT ABOVE`, i.e. not the precomposed U+0130, but a regular `I` followed by U+0307 `COMBINING DOT ABOVE`.

I guess this should be handled as a special case in the conversion, and the test string rewritten to make it clear what is going on (probably not `IECELLİ` in the string, but `IECELLI\u0307` or similar).

root/static/scripts/guess-case/modes.js
y-van-z: I proposed it at purpose in https://github.com/metabrainz/musicbrainz-server/pull/410#discussion_r93604082, since dotless/dotted I are also used by _Azerbaijanis and Khazaks_ (when using latin script), see the final note to [MBS-1319](https://tickets.metabrainz.org/browse/MBS-1319).

uklauer: Ah. Still, it isn’t a feature of the Turkic languages, just of some writing systems (Turkish originally, which then spread to neighbouring countries with related languages). I would prefer not to use Turkic here; in the same way as I wouldn’t like English mode to be called “Germanic mode” even in case Norwegian also used it.

root/static/scripts/guess-case/modes.js
y-van-z: Actually, a lot more of lower case words are missing. This list is based on [Turkish style guideline](https://musicbrainz.org/doc/Style/Language/Turkish). It should probably be filed as a STYLE issue.

root/static/scripts/guess-case/modes.js
y-van-z: [Capital](https://en.wikipedia.org/wiki/Letter_case) seems to be correctly used. But the overall phrasing of this description can certainly be improved.

CallerNo6: I tweaked the wording slightly [here](https://gist.github.com/CallerNo6/cea573f90aa56124114c6c0aa2cb5b5e), which feels slightly less awkward to me.

uklauer: Now I notice another thing: “Some words, often proper nouns, …” – that’s not really true anymore, is it? As it is now, the mode is more likely to over-capitalize rarer connectors.

I think I would simply drop the “often proper nouns” part. (In addition to the other changes suggested by @CallerNo6, of course.)

dpmittal:
@y-van-z , @uklauer , @mwiencek I have made the necessary changes :)

uklauer:
What I meant is that the conversion to lowercase should turn both `U+0130` and `U+0049 U+0307` into a regular `i`. With that, the test case would work, too.

sertansenturk:
@uklauer, sorry for late reply, I didn’t have internet connection for the last two days.

When I entered to the issue in Github, I couldn’t find my mention anymore. Is the problem solved? If not, can you re-point me?

If there are any additional doubts about the language, please don’t hesitate to mention me again.

uklauer:
It’s still relevant, @sertansenturk – thank you for replying! You can try this link, but I am not sure if it works: https://github.com/metabrainz/musicbrainz-server/pull/410#discussion_r93687732 If not, please click on the second “Show outdated” button, counting from the bottom. (Or the sixteenth from the top, if I counted right.)

dpmittal:
@uklauer  @y-van-z I have updated the code now and the all the Guess Case tests are passing :D

root/static/scripts/tests/GuessCase.js
```diff
@@ -187,6 +187,26 @@ test('Work', function (t) {
...
             expected: "Bring It All to Me (feat. 50 Cent)",
             mode: "English", roman: true, keepuppercase: false
+        },
+        {
+            input: "izarın gül gül olmuş",
+            expected: "İzarın Gül Gül Olmuş",
+            mode: "Turkish", roman: false, keepuppercase: false
+        },
+        {
+            input: "efendim hu nasibim Bu IECELLİ taksirat yahu",
```
y-van-z: Please make _taksirat_ all-uppercase as well so that decomposed character can be tested as well.

dpmittal:
@y-van-z @uklauer I have changed the code in \utils.js and added the proper test too testing both precomposed as well as decomposed string and the tests are passed :D

root/static/scripts/guess-case/utils.js
```diff
@@ -87,6 +87,14 @@ var lowerCaseBracketWordsList = [
 
 var lowerCaseBracketWords = new RegExp('^(' + lowerCaseBracketWordsList.join('|') + ')$', 'i');
 
+exports.turkishUpperCase = function(str) {
+  return str.normalize().replace(/i/g, 'İ').toUpperCase();
+};
+
+exports.turkishLowerCase = function(str) {
+  return str.normalize().replace(/I/g, 'ı').replace(/İ/g, 'i').toLowerCase();
+};
```
uklauer: Hmm. I guess this is a matter of taste; but NFC normalization is a rather heavy action that may change a number of unrelated things, too. (We will do it on the server later anyway.) I would prefer a simple `.replace(/I\u0307/g, 'i').replace(/İ/g, 'i')`.

And for the uppercasing function, normalization doesn’t serve a purpose.

uklauer:
Two other things about the commit structure. You now have the commit adding the feature after tgr one adding the tests; that’s not a very logical order. Furthermore, there are actually two separate changes here: Adding the infrastructure for mode-specific casing functions, and adding a Turkish mode. Those should be separate commits (so three in total).

dpmittal:
I have updated the PR . Now there are three commits and have made all the requested changes :)

uklauer:
Well, it's three commits, but the split as it is doesn’t make sense. Each commit should contain an independent change, and the code should still work after each one.

Basically, what is now the second commit should be the first, but the definitions of `turkishUpperCase` and `turkishLowerCase` belong into the other one.

root/static/scripts/guess-case/utils.js
```diff
@@ -87,6 +87,14 @@ var lowerCaseBracketWordsList = [
 
 var lowerCaseBracketWords = new RegExp('^(' + lowerCaseBracketWordsList.join('|') + ')$', 'i');
 
+exports.turkishUpperCase = function(str) {
+  return str.replace(/i/g, 'İ').toUpperCase();
+};
+
+exports.turkishLowerCase = function(str) {
+  return str.replace(/I\u0307/g, 'i').replace(/I/g, 'ı').replace(/İ/g, 'i').toLowerCase();
```
uklauer: The third `replace` is still unnecessary, because `toLowerCase` will handle that.

dpmittal: Actually third replace is necessary as 'İ'.toLowerCase() will give 'i̇' i.e. 'i\u0307' i with two dots..

dpmittal:
Updated the PR

root/static/scripts/guess-case/utils.js
uklauer: Which browser did you test that on? In Firefox, `"\u0130". toLowerCase() == "i" ` is true. (`"I\u307". toLowerCase() == "i"` is not, of course, but that’s what the first `replace()` is for.)

uklauer:
The commits still don’t make sense …

root/static/scripts/guess-case/utils.js
uklauer: OK, it seems the test runner (i.e. node.js) indeed behaves that strangely. Leave the third `replace` in, then. :-)

root/static/scripts/tests/GuessCase.js
```diff
@@ -187,6 +187,26 @@ test('Work', function (t) {
             input: "Bring It All To Me (f. 50 Cent)",
             expected: "Bring It All to Me (feat. 50 Cent)",
             mode: "English", roman: true, keepuppercase: false
+        },
```
uklauer: Above, in line 153, the number of tests needs to be updated.

dpmittal: Ohh, just missed it . Sorry for that I'll make the changes :)

root/static/scripts/guess-case/modes.js
sertansenturk: Again, sorry for responding slow. 

Here is the short response to all variants of the question word: 

"mı|mi|mu|mü|mıyım|miyim|muyum|müyüm|mısın|misin|musun|müsün|mıyız|miyiz|muyuz|müyüz|mısınız|misiniz|musunuz|müsünüz|mıdır|midir|mudur|müdür"

"müdür" is a problematic case as @uklauer mentioned. While it is almost always easy to distinguish the usage from the context, it would not be straightforward for a machine without incorporating some natural language processing. 

Below I am adding detailed explanations, in case it's needed. If you need help with other suffixes, conjuctions etc., please don't hesitate to ask. My apologies that I might not reply immediately for the next  days.

----------------------

The base question word in Turkish is "mı" with its conjugations "mi, mu, mü." According to the 	subject (tr. özne) of the sentence, a suffix is added. I'm adding a table of all variants of the question word below:

| Sample Verb         | yapar          | gelir          | uyur          | ölür          |
|---------------------|----------------|----------------|---------------|---------------|
| __1st Person Singular__ | mıyım          | miyim          | muyum         | müyüm         |
| __2nd Person Singular__ | mısın          | misin          | musun         | müsün         |
| __3rd Person Singular__ | mı             | mi             | mu            | mü            |
| __1st Person Plural__   | mıyız          | miyiz          | muyuz         | müyüz         |
| __2nd Person Plural__   | mısınız        | misiniz        | musunuz       | müsünüz       |
| __3rd Person Plural__   | (yapar-lar) mı | (gelir-ler) mi | ~~(uyur-lar) mı~~ | ~~(ölür-ler) mi~~ |

Grammatically the third person plural can only be "mı" or "mi" because of the plural suffix "ler" or "lar" added to the end of the verb.

If there is "no verb," i.e. we use the form which corresponds to "to be" in English, the subject is the third person (singular or plural). Hence, the question word could only be "mı, mi, mu, mü." In this case one can optionally add "dır" (or its harmonic variants "dir, dur, dür") for emphasis.

| Sample Noun         | yalan          | kel          | balon          | düş          |
|---------------------|----------------|----------------|---------------|---------------|
| __* to be__             | mı(dır)          | mi(dir)          | mu(dur)         | mü(dür)         |

The above case also holds for several tenses, e.g. "yapacak mıdır"

root/static/scripts/guess-case/utils.js
```diff
@@ -144,8 +152,8 @@ exports.titleString = function (is, forceCaps) {
     }
 
     var os;
-    var lc = is.toLowerCase();
```
mwiencek: I think what @uklauer meant is that only the changes from here down, plus the `toLowerCase`/`toUpperCase` methods added to `DefaultMode`, should be in the first commit. The message can be something like "Add support for mode-specific casing functions".

Everything else that has to do with the Turkish mode specifically should be in the second commit.

uklauer: Yes. For illustration, I did it in this branch: https://github.com/metabrainz/musicbrainz-server/compare/master...uklauer:pr-410

root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +291,24 @@ exports.Sentence = assign({}, DefaultMode, {
...
+    return false;
+  },
+
+  toUpperCase(str) {
+    return turkishUpperCase(str);
+  },
+
+  toLowerCase(str) {
+    return turkishLowerCase(str);
+  },
```
uklauer: Another thing I noticed is that it isn’t necessary to have these function definitions. Since the signatures match, it should be enough to write `toUpperCase: turkishUpperCase, toLowerCase: turkishLowerCase`.

root/static/scripts/guess-case/modes.js
```diff
@@ -282,3 +291,20 @@ exports.Sentence = assign({}, DefaultMode, {
...
+    {url: {href: 'https://musicbrainz.org/doc/Style/Language/Turkish', target: '_blank'}}
+  ),
+
+  isSentenceCaps() {
+    return false;
+  },
+
+  toUpperCase:turkishUpperCase,
+
+  toLowerCase:turkishLowerCase,
```
uklauer: There is one tiny style issue remaining, I would like to have a space after each of the colons.

But it’s now ripe for merging, I guess.

mwiencek:
I merged this in https://github.com/metabrainz/musicbrainz-server/commit/b1ef56f48e2d48066abf210db8258f5ce48c5f04 with the whitespace fixed. Thanks again, @dpmittal!

PR #415, opened 2016-12-26
Remove some ancient cruft

uklauer:
This removes/replaces some really ancient stuff. (Unbelievable how something like `lib/ModDefs.pm` can hide for years in the codebase.)


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -970,7 +963,7 @@ sub xml_search
                         parameter => 'artist',
                         escape    => 1,
                         split     => 1,
-                        process   => sub { my $term = shift; $term =~ s/\s*(.*?)\s*$/$1/; $term }
+                        process   => sub { shift =~ s/\s*(.*?)\s*$/$1/r }
```
uklauer: Good example of how this simplifies things.

uklauer:
Oh, and where `s///g` was used for simple character replacements, I switched to `tr///`.

lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -195,14 +195,6 @@ sub find_for_artist {
     );
 }
 
-sub _clean {
```
mwiencek: I guess you can remove the `-also` part on `use namespace::autoclean -also => [qw( _clean )];` too.

uklauer: Done. (I also removed the `no Moose` at the end, because `namespace::autoclean` will remove the Moose imports already.)

PR #416, opened 2016-12-26
MBS-9168: Re-enable indentation of certain locales

uklauer:
Since the update to DateTime::Locale 1.10, parts in locale strings are delimited by hyphens instead of underscores. Change the code that detects sub-locales accordingly, so that they can be indented again for visual grouping.


uklauer:
I’m retracting this PR because there are more locale issues that need to be addressed in a concerted way.

chirlu:
Reopening since I haven’t yet got around to doing more, so I guess we should at least fix this display thing. For now, interpret both `-` and `_` as separator.

PR #417, opened 2016-12-27
MBS-9143 (cont.): Remove the rest of the FreeDB search

uklauer:
This removes what was overlooked in PR #409.

Additionally, some visitors are redirected to freedb.org where they might find what they are looking for.


mwiencek:
Well, that's a lot of stuff that was overlooked. Thanks 👍

PR #418, opened 2016-12-27
Make use of non-destructive substitution and transliteration

uklauer:
Simplify some parts by using the `/r` non-destructive flag to the substitution operator (new since Perl 5.14): `$foo =~ s/bar/baz/r` doesn’t modify `$foo` as it would do without the `/r`, but simply returns a value with the substitution applied. Furthermore, replace simple character substitutions with transliterations (`tr///`).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -337,8 +337,7 @@ sub edit_type : Path('/doc/Edit_Types') Args(1) {
     my ($self, $c, $edit_type) = @_;
 
     my $class = EditRegistry->class_from_type($edit_type);
-    my $id = 'Edit Type/$class->edit_name';
-    $id =~ s/ /_/g;
+    my $id = 'Edit Type/$class->edit_name' =~ s/ /_/gr;
 
     my $version = $c->model('WikiDocIndex')->get_page_version($id);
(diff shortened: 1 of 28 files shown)


lib/MusicBrainz/Server/Controller/Edit.pm
```diff
@@ -337,8 +337,7 @@ sub edit_type : Path('/doc/Edit_Types') Args(1) {
     my ($self, $c, $edit_type) = @_;
 
     my $class = EditRegistry->class_from_type($edit_type);
-    my $id = 'Edit Type/$class->edit_name';
-    $id =~ s/ /_/g;
+    my $id = 'Edit Type/$class->edit_name' =~ tr/ /_/r;
```
mwiencek: Huh. Unrelated to your changes, but...single quotes?

mwiencek:
LGTM.

lib/MusicBrainz/Server/Controller/Edit.pm
uklauer: Yeah, good catch. This went in in 2011, so it seems nobody ever tried to add further information on edit types via the wiki. The delayed check whether `$class` actually exists doesn’t look right, either.

(And BTW, I don’t think you can do method calls inside a double-quote string even.)

uklauer: Fixed in PR #419.

PR #419, opened 2016-12-29
Bugfix and validation improvement related to edit documentation

uklauer:


lib/MusicBrainz/Server/Controller/Edit.pm
```diff
@@ -51,7 +51,7 @@ sub base : Chained('/') PathPart('edit') CaptureArgs(0) { }
 sub _load
 {
     my ($self, $c, $edit_id) = @_;
-    return unless is_positive_integer($edit_id);
+    return unless is_database_row_id($edit_id);
```
uklauer: Checks that the number doesn’t exceed the range (2^31).

lib/MusicBrainz/Server/Controller/Edit.pm
```diff
@@ -336,14 +336,16 @@ sub edit_types : Path('/doc/Edit_Types')
 sub edit_type : Path('/doc/Edit_Types') Args(1) {
     my ($self, $c, $edit_type) = @_;
 
-    my $class = EditRegistry->class_from_type($edit_type);
-    my $id = 'Edit Type/$class->edit_name' =~ tr/ /_/r;
+    my $class;
+    $class = EditRegistry->class_from_type($edit_type)
+        if is_database_row_id($edit_type);
+    $class or $c->detach('/error_404');
```
uklauer: This check was a bit late. Didn’t actually matter because `$class->edit_name` wasn’t called from inside the string; but now that it’s fixed, we should check earlier that `$class` really was found.

mwiencek:
:shipit:

PR #421, opened 2016-12-31
Fix some HTML errors/warnings in tests

mwiencek:
See commits...some HTML/CSS has changed, so let me know if something looks wrong or has a better solution.

### DIFF
--- a/root/layout/components/Search.js
+++ b/root/layout/components/Search.js
@@ -44,7 +44,7 @@ const Search = () => (
     {' '}{searchOptions}{' '}
     <input type="hidden" id="headerid-method" name="method" value="indexed" />
     <button type="submit">
-      <img src={manifest.pathTo('/images/icons/search.svg')} />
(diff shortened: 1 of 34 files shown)


mwiencek:
Getting closer to having automated tests work again.

uklauer:
I believe the three PRs together should do it.

root/layout/components/Search.js
```diff
@@ -44,7 +44,7 @@ const Search = () => (
     {' '}{searchOptions}{' '}
     <input type="hidden" id="headerid-method" name="method" value="indexed" />
     <button type="submit">
-      <img src={manifest.pathTo('/images/icons/search.svg')} />
+      <img alt={l('Search')} src={manifest.pathTo('/images/icons/search.svg')} />
```
uklauer: I think I would use an empty `alt=""` here (and in the TT below, of course). We already have “Search“ as the placeholder text for the input field, so it would be unnecessary clutter for a screenreader to repeat it here. That it is the submit button for the search form should be enough.

PR #424, opened 2017-01-02
MBS-7307: Show LinkedIn URLs in the sidebar

anshuman73:


uklauer:
`s/linkedIn/LinkedIn/` in the commit description?

anshuman73:
LinkedIn.
That's there official way of calling it. I'll fix this up in a second.

anshuman73:
@y-van-z 
There were some problems with regex parsing of linked urls with sub-domains (for example - http://musicbrainz.org/artist/5b7c6d07-4847-43f6-817b-8ea0088c5b92).
Fixed the issue, tested it and squashed the commit for a cleaner history.

uklauer:
This is something that should be done in URL cleanup on input. The display code should only deal with cleaned up URLs.

anshuman73:
@uklauer I was talking about the subdomain in the link, _ca_.linkedin.com
But yes, I agree cleanup also needs to be done (unless you're talking about a way to convert subdomain into something like linkedIn.com/_ca_/profile-name

mwiencek:
I think the subdomain should be removed entirely in URL cleanup, since it's not necessary to view the profile (it appears that you can view any profile from any country's subdomain).

As of this writing, there are 305 LinkedIn URLs with two-letter subdomains.

anshuman73:
Created MBS-9188

PR #425, opened 2017-01-05
MBS-9180: Near-identical duplicates of common-xxx.js

mwiencek:
Two changes here. The first is to add `->canonical` to the JSON object, so that the keys aren't outputted in random order. The second is to remove unused settings from the client JS altogether. I think the first fix is still good to have independent of the second fix, because we may have object-like DBDefs settings in the client JS in the future.

### DIFF
--- a/script/compile_resources.sh
+++ b/script/compile_resources.sh
@@ -3,6 +3,8 @@
 MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)
 cd "$MB_SERVER_ROOT"
 
-./script/dbdefs_to_js.pl
+./script/dbdefs_to_js.pl --client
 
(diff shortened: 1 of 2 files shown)


script/compile_resources.sh
```diff
@@ -3,6 +3,8 @@
 MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)
 cd "$MB_SERVER_ROOT"
 
-./script/dbdefs_to_js.pl
+./script/dbdefs_to_js.pl --client
```
mwiencek: Instead of storing the client/server DBDefs settings in different files, I had it generate a file containing only the client settings first, which is bundled in the gulp step, and then a file containing all the settings. The second file is used if the template renderer is run locally.

PR #428, opened 2017-01-08
MBS-9173: Support serializing of instrument and series lists.

Zastai:
The JSON serializer does not currently support serializing lists of series or instruments, breaking browse requests for those entities.
This small patch resolves that, and also fixes a perl warning seen for browse requests (both XML and JSON).

### DIFF
--- a/lib/MusicBrainz/Server/Data/Role/Collection.pm
+++ b/lib/MusicBrainz/Server/Data/Role/Collection.pm
@@ -7,6 +7,8 @@ sub find_by_collection {
     my ($self, $collection_id, $limit, $offset, $order) = @_;
 
     my ($order_by, $extra_join, $also_select) = $self->_order_by($order);
+    $extra_join //= '';
+
     my $table = $self->_table;
     my $type = $self->_type;
 
--- a/lib/MusicBrainz/Server/WebService/JSONSerializer.pm
+++ b/lib/MusicBrainz/Server/WebService/JSONSerializer.pm
@@ -5,6 +5,7 @@ use Moose;
 use JSON;
 use List::MoreUtils qw( any );
 use List::UtilsBy 'sort_by';
+use MusicBrainz::Server::Constants qw( %ENTITIES );
 use MusicBrainz::Server::Data::Utils qw( non_empty ref_to_type type_to_model );
 use MusicBrainz::Server::WebService::WebServiceInc;
 use MusicBrainz::Server::WebService::Serializer::JSON::2::Utils qw( list_of number serializer serialize_entity );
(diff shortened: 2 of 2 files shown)


uklauer:
@brainzbot, add to whitelist.

lib/MusicBrainz/Server/Data/Role/Collection.pm
```diff
@@ -10,6 +10,9 @@ sub find_by_collection {
     my $table = $self->_table;
     my $type = $self->_type;
 
+    # Is mostly unset, causing a warning when interpolating it below.
+    $extra_join = '' unless defined($extra_join);
```
uklauer: This would normally be written `$extra_join //= '';` I would move it immediately below the line where the variable is set and don’t think this is worth a comment.

lib/MusicBrainz/Server/WebService/JSONSerializer.pm
```diff
@@ -54,6 +54,8 @@ sub work_list          { shift->entity_list(@_, "work", "works") };
 sub area_list          { shift->entity_list(@_, "area", "areas") };
 sub place_list         { shift->entity_list(@_, "place", "places") };
 sub event_list         { shift->entity_list(@_, "event", "events") };
+sub instrument_list    { shift->entity_list(@_, "instrument", "instruments") };
+sub series_list        { shift->entity_list(@_, "series", "series") };
```
uklauer: Ideally, this should be generated programmatically. If you don’t want to do that now, please leave a comment like `# FIXME: use entities_with` so that it can be found later.

lib/MusicBrainz/Server/Data/Role/Collection.pm
Zastai: Fair enough, will adjust it.

lib/MusicBrainz/Server/WebService/JSONSerializer.pm
Zastai: Do you mean rewriting serialize() so that it no longer relies on a named sub lookup? I can look into that.
Or do you mean the plural form only? Looking at entities.json, plural_url looks like it has the value I need; is that guaranteed to be the value intended to be used in the JSON serialization?

Zastai: Went for the first approach. It still relies on assuming that foo-bar-list means that the element entity type name (as used in entities.json) is foo_bar though; if that can be avoided (e.g. by having that in entities.json, perhaps as a list_type property), so much the better.

lib/MusicBrainz/Server/WebService/JSONSerializer.pm
```diff
@@ -14,53 +15,33 @@ sub fmt { 'json' }
...
+          $ret->{$singular . '-offset'} = number($entity->{offset});
+          $ret->{$singular . '-count' } = number($entity->{total});
+        }
+        return encode_json($ret);
+      }
+    }
 
-    my ($entity, $inc, $opts) = @data;
+    my $override = $self->meta->find_method_by_name($type);
+    return $override->execute($self, $entity, $inc, $opts) if $override;
```
uklauer: Moving this here means that it is impossible to override, e.g., `label_list` or any of the lists, should that become necessary. So I would prefer to keep the override first.

lib/MusicBrainz/Server/WebService/JSONSerializer.pm
```diff
@@ -14,53 +15,33 @@ sub fmt { 'json' }
...
-    my ($self, $type, @data) = @_;
+    my ($self, $type, $entity, $inc, $opts) = @_;
 
     $type =~ tr/-/_/;
 
-    my $override = $self->meta->find_method_by_name($type);
-    return $override->execute($self, @data) if $override;
+    if ($type =~ m/^(.*)_list$/ && exists($entity->{items})) {
+      if (exists($ENTITIES{$1}) && exists($ENTITIES{$1}->{plural_url})) {
+        $type = $1;
```
uklauer: Style issues:
- MBS uses four spaces (and never tabs) for each indentation level; at least in Perl code.
- The matched entity type deserves a proper (variable) name, IMO. I suggest writing `my ($entity_type) = $type =~ /.../`, then using `$entity_type` in this block.
- Not important: We mostly use plain `/.../` instead of `m/.../` in the existing codebase (by a 114:9 majority).

Zastai:
I know, I fully intend to recreate and force push this branch so that it has only the final approved changes, which is why I included that information in the commit message.

lib/MusicBrainz/Server/WebService/JSONSerializer.pm
Zastai: Well, is this actually intended for actual overriding, or was it only added to handle lists? At the moment, its only use (as far as I could see) was to handle the lists separately, making it obsolete.
I'm not sure I see the use case otherwise; the default behaviour is to use the specific serialization modules for each entity, which seem to be used by the JSON serializer only. So any overrides of behaviour could just be made there (or, as done now for lists, in the serialize method).
The "just add a method with a particular name to override behaviour" idea is fairly icky, as any existing method could be invoked by accident if a new type string comes in to serialize().

lib/MusicBrainz/Server/WebService/JSONSerializer.pm
Zastai: 1. Right, I still need to customize the emacs on my MB devel VM for that; it uses tabs by default.
2. Well until the exists checks succeed, $1 is just a potential entity type, so I kept it as $1 for the if and then assigned it to $type once it was verified to be a valid type. I have no issue with adding an $entity_type or $list_item_type or whatever to hold $1 right away. I'm not a huge fan of assigning the result of =~ (I think it reads better in the if), but if that's the standard in the MB codebase, then I'll use that.
Or I could drop the capture from the regex and use a substr inside the if to get the list item type; that avoids the $1 altogether (not sure if it would give better performance though).
3. No problem, I'll drop the m.

uklauer:
Well, if the overriding isn’t used, I guess we can drop it?

Zastai:
OK, squashed the commits into just two relevant ones. I left in the overriding; it does no harm if unused, and voids breakages if some code somewhere does rely on it.

PR #430, opened 2017-01-10
MBS-9196: Recognize Wiki redirect pages again

uklauer:
The HTML for redirect pages from MediaWiki has changed. Adapt the regex so that redirect pages are again recognized as such.


lib/MusicBrainz/Server/Data/WikiDoc.pm
```diff
@@ -147,7 +147,7 @@ sub _load_page
         return undef;
     }
 
-    if ($content =~ /<span class="redirectText"><a href="(?:https?:)?\/\/.*?\/(.*?)"/) {
+    if ($content =~ m{<ul class="redirectText"><li><a href="(?:https?:)?(?://[^/]+)?/(.*?)"}) {
```
uklauer: There were two changes: 1. the HTML elements, 2. the URL is currently a relative one (but this should work for relative, schema-relative and absolute URLs). I also used `m{}` instead of `//` so that slashes don’t need to be escaped.

legoktm:
It seems like it would be better to use an API request for this? The HTML structure of MediaWiki isn't exactly a stable interface....

e.g. <https://wiki.musicbrainz.org/api.php?action=query&prop=info&inprop=url&titles=Entity&redirects=1&formatversion=2&format=json>

PR #432, opened 2017-01-10
MBS-9198: Show down arrow in a dropdown even in Firefox

uklauer:
The entity type dropdown in the search form in the page header is supposed to show a down arrow. This worked in Chromium, but not in Firefox.

There were three separate issues:
- Firefox doesn’t support the individual CSS properties `background-position-{x,y}` before version 49. Use the combined `background-position` instead.
- When embedding an image as a data URI, it needs to be properly percent-encoded. Move the image to a separate file and let `data-uri()` handle the encoding. (Chromium seems to be less strict here.)
- The SVG file used an invalid SVG version. Set it to version 1.1.


uklauer:
See http://mb.chirlu.de/ (in Firefox).

PR #434, opened 2017-01-10
Fix musicbrainz-tests DBDefs mess

mwiencek:
Removes a file that wasn't supposed to be committed. I'm not sure how it even got there, since the path it's at doesn't make sense.

Copies the HTML_VALIDATOR value to the correct file. It was previously incorrectly changed in the removed file in commit a4f0af5fe1aeb6652f8fa037dfe054fcd6b2b08e.

--

Let's see what the tests say now...


mwiencek:
This fixes the HTML errors. Working on the rest.

PR #439, opened 2017-01-16
MBS-9002: Automatically disable and tick ended date checkbox in editors

y-van-z:
Automatically tick and disable the “ended” date checkbox when the `end_date.year` field is filled in alias, (Area, Artist, Label, and Place) entity and relationship editors.

When the `end_date.year` field is emptied, it restores the state of the “ended” date checkbox.

Additionally factorize date range field set templates, and align the “ended” checkbox with the “end date” text input in the relationship editor, that is, in the same way as in the other editors.

### DIFF
--- a/root/area/edit_form.tt
+++ b/root/area/edit_form.tt
@@ -15,15 +15,7 @@
       [%- form_row_text_list(r, 'iso_3166_3', l('ISO 3166-3:'), l('ISO 3166-3')) -%]
     </fieldset>
 
-    <fieldset>
-      <legend>[% l('Date Period') %]</legend>
-      <p>
-        [%- l('Dates are in the format YYYY-MM-DD. Partial dates such as YYYY-MM or just YYYY are OK, or you can omit the date entirely.') -%]
-      </p>
-      [% form_row_date(r, 'period.begin_date', l('Begin date:')) %]
-      [% form_row_date(r, 'period.end_date', l('End date:')) %]
-      [% form_row_checkbox(r, 'period.ended', l('This area has ended.')) %]
-    </fieldset>
+    [% date_range_fieldset(r, 'area', l('This area has ended.')) %]
 
     [% PROCESS 'forms/relationship-editor.tt' %]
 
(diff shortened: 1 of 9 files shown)


root/area/edit_form.tt
```diff
@@ -42,6 +42,16 @@
 
 <script type="text/javascript">//<![CDATA[
   (function () {
+    const wasEnded = $('#id-edit-area\\\.period\\\.ended').prop('checked');
```
mwiencek: It would definitely be nice to move this to a module, e.g. root/static/scripts/edit/utility/toggleEnded.js or similar. Then require it from root/static/scripts/edit.js and it'll be loaded on all the edit pages.

The only variation seems to be the entity type, and you can detect it from `document.location.pathname` if nothing else.

If you wrap everything in a [jQuery "ready" callback](https://learn.jquery.com/using-jquery-core/document-ready/), the inputs should exist when it in runs, even if the script is included above those elements.

Placing the code in a file has the added benefit that we can safely use `const`, like you did. As an inline script, here it will break in some of our supported browsers. But the bundled files are [transformed using Babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=latest%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=const%20foo%20%3D%201%3B&playground=true), when `compile_resources.sh` is run, so it'll get converted into something compatible in older browsers (a var statement) while still giving us the compile-time benefits (an error will be thrown if Babel sees you tried to re-assign `wasEnded`).

mwiencek: Or if you'd prefer, you can wrap the code in a function which accepts the entity type as a parameter, then put the function on the global `MB` object, and call it here.

y-van-z: @mwiencek: Thank you for the review. By the way, alias and most of entity types are using the same date range form. Shouldn’t that be factorized as well?

mwiencek: Yeah, it looks we can reduce some duplication there as well. Just some of the strings differ.

y-van-z:
Your requested changes are served, bon appétit !

root/static/scripts/edit/utility/toggleEnded.js
```diff
@@ -0,0 +1,24 @@
+// This file is part of MusicBrainz, the open internet music database.
+// Copyright (C) 2017 MetaBrainz Foundation
```
mwiencek: You should replace that with your own name (unless you intended to assign copyright to MeB).

y-van-z: That was intended, thanks for the notice.

PR #347, opened 2016-08-02
MBS-9038 & MBS-9039: Handle new patronage and crowdfunding URL relationships

y-van-z:
For ChangeTip, Flattr, Kickstarter, Indiegogo, Patreon, PayPal.Me and Tipeee:
- auto-detect (new) `patronage` or `crowdfunding` link type and clean URLs,
- display in the sidebar with specific favicon and label.

(Keep patronage pages grouped in the `/relationships` tab.)
(Keep  `crowdfunding` label in the `/relationships` tab.)

PR #440, opened 2017-01-18
MBS-8998: Display CPDL Score links in the sidebar

anshuman73:


lib/MusicBrainz/Server/Entity/URL/CPDL.pm
```diff
@@ -0,0 +1,36 @@
...
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
```
y-van-z: This address is no longer current, see https://github.com/metabrainz/musicbrainz-server/pull/347#r73416211.

lib/MusicBrainz/Server/Entity/URL/CPDL.pm
```diff
@@ -0,0 +1,36 @@
...
+
+use Moose;
+
+extends 'MusicBrainz::Server::Entity::URL';
+with 'MusicBrainz::Server::Entity::URL::Sidebar';
+
+sub sidebar_name {
+    my $self = shift;
+
+    return "Score at CPDL";
```
y-van-z: Maybe that should be labelled more explicitly as "Choral Public Domain Library". @reosarevok: Any opinion?

reosarevok: That seems way too long, and we say "Score at IMSLP"

lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -34,6 +34,7 @@ my %URL_SPECIALIZATIONS = (
     'CDBaby'              => qr{^https?://(?:www.)?cdbaby.com/}i,
     'ChangeTip'           => qr{^https?://(?:www.)?changetip.com/tipme/}i,
     'CiNii'               => qr{^https?://(?:www.)?ci.nii.ac.jp/}i,
+    'CPDL'                => qr{^https?://(?:www.)?cdpl.org/wiki/}i,
     'Commons'             => qr{^https?://commons.wikimedia.org/wiki/File:}i,
```
anshuman73: but I don't think a score link will ever need to match anything out of  http://www2.cpdl.org/wiki/ , would it ?

y-van-z: No it won't, that is exactly why I crossed it out of my review two days ago. 😉 Neither will `www`. And `d` and `p` are inverted, that is, it does not match anything but misspelled URLs.

anshuman73: oh, well, sorry. Missed that.

anshuman73:
> (?:www.)? remains unneeded

Then I guess it's fair to remove that matching group.
And given that the unclean URLs are too being cleaned and the new added links will already be cleaned, it makes no sense for it to be there.

I'll remove them in a sec

PR #444, opened 2017-01-29
MBS-9193: Report instruments without image

y-van-z:
# [MBS-9193](https://tickets.metabrainz.org/browse/MBS-9193)
- Actually report instruments with neither `image` URL relationship nor `wikidata` URL relationship.
- Columns are (linked) instrument name, type, and last updated.
- Rows are sorted by type id, then by instrument name, that is, in the same way as `/instruments` list.

### DIFF
--- a/lib/MusicBrainz/Server/Report/InstrumentReport.pm
+++ b/lib/MusicBrainz/Server/Report/InstrumentReport.pm
@@ -1,0 +1,35 @@
+package MusicBrainz::Server::Report::InstrumentReport;
+use Moose::Role;
+
+with 'MusicBrainz::Server::Report::QueryReport';
+
+around inflate_rows => sub {
+    my $orig = shift;
+    my $self = shift;
+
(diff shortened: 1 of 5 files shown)


root/report/instruments_without_image.tt
```diff
@@ -0,0 +1,42 @@
+[%- WRAPPER 'layout.tt' title=l('Instruments without image') full_width=1 -%]
```
mwiencek: Grammar nitpick: Either "Instruments without an image" (each instrument has no image) or "Instruments without images" (none of them have images) would work. Same in the h1 and in the description below, but I would just make "relationships" plural there, since it's not known whether to use "a" or "an."

y-van-z: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/444/commits/cf109ca175a52f75b7cc016a2381eca2ab8d943f

PR #448, opened 2017-02-02
MBS-8236: Recognise artist and ensemble for Classical Archives URLs

y-van-z:
Fix `otherdatabases` entry for Classical Archives URLs with both `/artist/` and `/ensemble/` paths.


y-van-z:
Actually, all checks have passed in [build #172](https://ci.metabrainz.org/job/musicbrainz-server/172/).

mwiencek:
Yeah, I guess that Redis issue was likely transient—the test that failed does depend on a particular timing of the statements being executed (still strange though).

PR #449, opened 2017-02-02
MBS-9224: The gulpfile sometimes produces a corrupt rev-manifest bundle

mwiencek:
This was almost certainly due to concurrent asynchronous writes to rev-manifest.json, through `writeResource`. The code has been modified to only write rev-manifest.json once at the end of the build, rather than for each bundle, so that there are no longer any concurrent writes. After these changes, I can no longer reproduce the issue after many attempts.

I tested that the new bundles didn't throw any errors on several edit pages, including under different languages. I also checked that the file hashes didn't change between builds.

I adapted the "watch" mode for these changes too, but it seems to have already been broken before this patch (and I didn't fix it).

### DIFF
--- a/docker/musicbrainz-tests/run_tests.sh
+++ b/docker/musicbrainz-tests/run_tests.sh
@@ -15,6 +15,10 @@ while true; do
     fi
 done
 
+# Tests require rev-manifest.json to exist.
+sudo -E -H -u musicbrainz \
+    carton exec -- ./script/compile_resources.sh
+
 exec sudo -E -H -u musicbrainz carton exec -- prove \
     --pgtap-option dbname=musicbrainz_test \
     --pgtap-option host=musicbrainz-test-database \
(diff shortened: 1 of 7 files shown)


mwiencek:
I know the gulpfile is black magic so it'd be nice if others can test this too.

mwiencek:
Not sure what's with the test failures yet, I'm not able to reproduce them.

y-van-z:
I cannot reproduce these failures either, but I encountered two others by running tests in (FF) browser:
1. `too much recursion` error caused by tests nesting, that can be fixed with https://github.com/y-van-z/musicbrainz-server/commit/b074d11ae1f74c4430e68a2494ef36bd97bd33e6
2. Label guess case test failure using Turkish mode, that can be fixed with https://github.com/y-van-z/musicbrainz-server/commit/5c2506372884e376b7095ab3355042e9893eef6a

PR #452, opened 2017-02-06
Fail edits trying to set empty credited-as AC names

chirlu:
Due to a bug (see MBS-8750), there are edits trying to set an empty artist credit name (Artist Foo as `""`). Those edits would never close.

Test for this case when applying an edit and explicitly fail if necessary.


chirlu:
This will kill most of the stuck edits on https://musicbrainz.org/edit/open when it goes to production. The “merge releases” ones, however, will need a separate investigation.

This does not fix the issue itself (being able to submit such edits).

mwiencek:
@brainzbot add to whitelist

mwiencek:
Thanks. Probably should've done this a lot sooner.

PR #455, opened 2017-02-07
Use new exim-relay service

mwiencek:
In production, a new relay service "musicbrainz-org.exim-relay" has been set up to fix mail issues.


mwiencek:
We'll point to that service using the `MBS_SMTP_SERVICE` environment variable in docker-server-configs.

PR #456, opened 2017-02-08
Fix i18n text selection (related to MBS-7795)

loujine:
'… and {place_count} other' should be displayed when markers.length == CLUSTER_POPUP_LIMIT + 1

The bug can be seen e.g. on https://beta.musicbrainz.org/area/74e50e58-5deb-4b99-93a2-decbb365c07f/places when unzooming one step on the map (there's a cluster with 11 entities)


chirlu:
@brainzbot, test this please.

mwiencek:
Oops, I screwed that up. Thanks for the fix. 👍

PR #458, opened 2017-02-11
MBS-9248: Don't show artist name in hover if same as sort name

reosarevok:
Should solve MBS-9248 by checking if the sortname is the same as the name and not printing the name if so.

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -350,8 +350,12 @@ END -%]
     IF namevar AND action == 'show' AND content != ''
             AND (noescape ? html_unescape(content.remove('</?span\b[^>]*>')) : content) != entity.name;
         options.unshift('name_variation');
-        hover = hover != '' ? l('{name} – {additional_info}', { name => entity.name, additional_info => hover })
-                            : entity.name;
+        IF (entity.sort_name && entity.sort_name == entity.name);
+            hoverprint = hover;
+        ELSE;
+            hoverprint = l('{name} – {additional_info}', { name => entity.name, additional_info => hover });
+        END;
+        hover = hover != '' ? hoverprint : entity.name;
     END;
 
     options.unshift('edits_pending') IF entity.edits_pending AND action == 'show';
--- a/root/static/scripts/common/components/EntityLink.js
+++ b/root/static/scripts/common/components/EntityLink.js
@@ -145,7 +145,9 @@ const EntityLink = (props = {}) => {
 
     if (nameVariation) {
       if (hover) {
-        hover = l('{name} – {additional_info}', {name: entity.name, additional_info: hover});
+        if (entity.name !== entity.sortName) {
+          hover = l('{name} – {additional_info}', {name: entity.name, additional_info: hover});	
+        }
       } else {
         hover = ko.unwrap(entity.name);
       }
--- a/t/lib/t/TemplateMacros.pm
+++ b/t/lib/t/TemplateMacros.pm
@@ -178,7 +178,7 @@ test all => sub {
 
             '<span class="mp">' .
                 '<span class="name-variation">' .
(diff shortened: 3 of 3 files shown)


root/components/common-macros.tt
```diff
@@ -350,8 +350,12 @@ END -%]
     IF namevar AND action == 'show' AND content != ''
             AND (noescape ? html_unescape(content.remove('</?span\b[^>]*>')) : content) != entity.name;
         options.unshift('name_variation');
-        hover = hover != '' ? l('{name} – {additional_info}', { name => entity.name, additional_info => hover })
-                            : entity.name;
+        IF (entity.sort_name && entity.sort_name == entity.name);
```
reosarevok: Not 100% sure if entity.sort_name is needed, let me know if I should just get rid?

gentlecat:
@brainzbot, add to whitelist.

chirlu:
I’m against this “improvement”; see the ticket for rationale.

Furthermore, how it is implemented in this PR is broken in several ways, but I would prefer not to waste time on writing review comments unless I actually have to.

Freso:
If you mention it, you have to. Even if the PR is dropped, @reosarevok (or other people watching along) cannot learn and improve from the mistakes they're not told about. So either don't mention that it is flawed, or back it up by arguing why.

root/components/common-macros.tt
chirlu: This is the wrong place to do changes specific for artists, as this is a generic macro that is used for all main entities.

chirlu: Depends on what you want. As it is, it checks whether the sort name is `"0"` or something else.

root/components/common-macros.tt
```diff
@@ -350,8 +350,12 @@ END -%]
...
+            hoverprint = hover;
+        ELSE;
+            hoverprint = l('{name} – {additional_info}', { name => entity.name, additional_info => hover });
+        END;
+        hover = hover != '' ? hoverprint : entity.name;
```
chirlu: I refrain from commenting on this for the moment since it will have to look different when moved to its proper place.

root/components/common-macros.tt
reosarevok: I'm not sure I follow the reasoning there. The main part of this is already in the generic macro, despite only applying to entities with name variations (so, right now, artists, unless I'm forgetting some). How is it any different to also put here the code that applies to entities with name variations which happen to have a sort name equal to their name?

mwiencek: Meaning if `entity.sort_name ` is literally the string `"0"`, it'll be "false" because perl. I think the intent here is to guard against entities that don't have a `sort_name` field at all, so it should be no longer needed if the check is moved as suggested below.

mwiencek: We also use `namevar` when recordings are linked to using a track name. So I think it's correct to do the change in `link_artist` instead, because artists are the only entities with sort names (well, non-alias ones).

root/static/scripts/common/components/EntityLink.js
```diff
@@ -145,7 +145,9 @@ const EntityLink = (props = {}) => {
 
     if (nameVariation) {
       if (hover) {
-        hover = l('{name} – {additional_info}', {name: entity.name, additional_info: hover});
+        if !(entity.name == entity.sortname) {
```
mwiencek: A `!` outside the parens is not valid JavaScript. :P

Try `(entity.name !== entity.sortname)` instead.

mwiencek: Also, `sortname` should be `sortName`.

reosarevok:
We agreed during the meeting to close this and talk about the issue in Jira further.

PR #459, opened 2017-02-12
MBS-9135: Allow admins to modify usernames

reosarevok:
Tested on local sandbox and it seems to work just fine, but let me know if something seems off!

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Account.pm
+++ b/lib/MusicBrainz/Server/Controller/Account.pm
@@ -289,9 +289,30 @@ sub edit : Local RequireAuth DenyWhenReadonly {
     $c->model('Area')->load($editor);
     $c->model('EditorLanguage')->load_for_editor($editor);
 
-    my $form = $c->form( form => 'User::EditProfile', init_object => $editor );
+    my $form = $c->form( 
+        form => 'User::EditProfile', 
+        item => {
+            username          => $editor->name,
+            email             => $editor->email,
+            skip_verification => 0,
+            website           => $editor->website,
+            biography         => $editor->biography,
+            gender_id         => $editor->gender_id,
+            area_id           => $editor->area_id,
+            area              => $editor->area,
+            birth_date        => $editor->birth_date,
+            languages         => $editor->languages,
+        },
+    );
 
     if ($c->form_posted && $form->process( params => $c->req->params )) {
+        my $old_username = $editor->name;
+        my $new_username = $form->field('username')->value;
+              
+        if (defined $new_username && $new_username ne $old_username) {
+            $c->detach('/error_403');            
+        }
+
         $c->model('Editor')->update_profile(
             $editor,
             $form->value
--- a/lib/MusicBrainz/Server/Controller/Admin.pm
+++ b/lib/MusicBrainz/Server/Controller/Admin.pm
@@ -35,6 +35,7 @@ sub edit_user : Path('/admin/user/edit') Args(1) RequireAuth HiddenOnSlaves
     my $form2 = $c->form(
         form => 'User::EditProfile',
         item => {
+            username          => $user->name,
             email            => $user->email,
             skip_verification => 0,
             website            => $user->website,
@@ -43,14 +44,13 @@ sub edit_user : Path('/admin/user/edit') Args(1) RequireAuth HiddenOnSlaves
     );
 
     if ($c->form_posted) {
-        if ($form->submitted_and_valid($c->req->params )) {
+        if ($form->submitted_and_valid($c->req->params) 
+            && $form2->submitted_and_valid($c->req->params )) {
             # When an admin views their own flags page the account admin checkbox will be disabled,
             # thus we need to manually insert a value here to keep the admin's privileges intact.
             $form->values->{account_admin} = 1 if ($c->user->id == $user->id);
             $c->model('Editor')->update_privileges($user, $form->values);
-        }
-
-        if ($form2->submitted_and_valid($c->req->params )) {
+        
             $c->model('Editor')->update_profile(
                 $user,
                 $form2->value
@@ -76,11 +76,11 @@ sub edit_user : Path('/admin/user/edit') Args(1) RequireAuth HiddenOnSlaves
                     $c->forward('/discourse/sync_sso', [$user]);
                 }
             }
+            
+            $c->flash->{message} = l('User successfully edited.');
+            $c->response->redirect($c->uri_for_action('/user/profile', [$form2->field('username')->value]));
+            $c->detach;
         }
-
-        $c->flash->{message} = l('User successfully edited.');
-        $c->response->redirect($c->uri_for_action('/user/profile', [$user->name]));
-        $c->detach;
     }
 
     $c->stash(
--- a/lib/MusicBrainz/Server/Form/User/EditProfile.pm
+++ b/lib/MusicBrainz/Server/Form/User/EditProfile.pm
@@ -2,13 +2,20 @@ package MusicBrainz::Server::Form::User::EditProfile;
 
 use HTML::FormHandler::Moose;
 use List::MoreUtils qw( any all );
-use MusicBrainz::Server::Form::Utils qw( language_options select_options_tree );
+use MusicBrainz::Server::Form::Utils qw( language_options select_options_tree validate_username );
 use MusicBrainz::Server::Translation qw( l ln );
 use MusicBrainz::Server::Validation qw( is_valid_url );
 
 extends 'MusicBrainz::Server::Form';
 
 has '+name' => ( default => 'profile' );
+
+has_field 'username' => (
+    type      => 'Text',
+    required  => 1,
+    maxlength => 64,
+    validate_method => \&validate_username,
+);
 
 has_field 'biography' => (
     type => 'Text',
--- a/lib/MusicBrainz/Server/Form/User/Register.pm
+++ b/lib/MusicBrainz/Server/Form/User/Register.pm
@@ -1,6 +1,7 @@
 package MusicBrainz::Server::Form::User::Register;
 
 use HTML::FormHandler::Moose;
+use MusicBrainz::Server::Form::Utils qw( validate_username );
 use MusicBrainz::Server::Translation qw( l ln );
 
 extends 'MusicBrainz::Server::Form';
@@ -10,7 +11,8 @@ has '+name' => ( default => 'register' );
 has_field 'username' => (
     type      => 'Text',
     required  => 1,
-    maxlength => 64
+    maxlength => 64,
+    validate_method => \&validate_username,
 );
 
 has_field 'password' => (
@@ -34,22 +36,6 @@ has_field 'email' => (
     maxlength => 64,
     required => 1
 );
-
-sub validate_username
-{
-    my ($self, $field) = @_;
-
-    my $username = $field->value;
-    if (defined $username) {
-        if ($username =~ qr{^deleted editor \#\d+$}i) {
-            $field->add_error(l('This username is reserved for internal use'));
-        }
-        my $editor = $self->ctx->model('Editor')->get_by_name($username);
-        if (defined $editor) {
-            $field->add_error(l('Please choose another username, this one is already taken'));
-        }
-    }
-}
 
 1;
 
--- a/lib/MusicBrainz/Server/Form/Utils.pm
+++ b/lib/MusicBrainz/Server/Form/Utils.pm
@@ -22,6 +22,7 @@ use Sub::Exporter -setup => {
                       build_attr_info
                       build_options_tree
                       indentation
+                      validate_username
               )]
 };
 
@@ -229,6 +230,25 @@ sub build_child_info {
 sub indentation {
     my $level = shift;
     return "\N{NO-BREAK SPACE}" x (3 * $level);
+}
+
+sub validate_username {
+    my ($self) = @_;
+
+    my $username = $self->value;
+    my $previous_username = $self->init_value;
+
+    if (defined $username) {
+        unless (defined $previous_username && $previous_username eq $username) { 
+            if ($username =~ qr{^deleted editor \#\d+$}i) {
+                $self->add_error(l('This username is reserved for internal use.'));
+            }
+            my $editor = $self->form->ctx->model('Editor')->get_by_name($username);
+            if (defined $editor) {
+                $self->add_error(l('Please choose another username, this one is already taken.'));
+            }
+        }
+    }
 }
 
 1;
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -268,6 +268,7 @@ sub update_profile
     my $row = hash_to_row(
         $update,
         {
+            name => 'username',
             bio => 'biography',
             area => 'area_id',
             gender => 'gender_id',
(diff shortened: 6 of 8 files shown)


Freso:
As the user would need to log out and back in again for the change to "show up", is there any chance that all their current sessions could get removed/invalidated once the name is changed? (I looked over the code and it doesn't look like this is done currently.)

lib/MusicBrainz/Server/Form/Role/ValidateUsername.pm
```diff
@@ -0,0 +1,22 @@
...
+use MusicBrainz::Server::Translation qw( l );
+
+sub validate_username
+{
+    my ($self, $field) = @_;
+
+    my $username = $field->value;
+    if (defined $username) {
+        if ($username =~ qr{^deleted editor \#\d+$}i) {
+            $field->add_error(l('This username is reserved for internal use'));
```
Freso: Are these the only names reserved for internal use? Also, do we not use proper grammar (ie., punctuation) in other messages?

lib/MusicBrainz/Server/Form/Role/ValidateUsername.pm
```diff
@@ -0,0 +1,22 @@
...
+        }
+        my $editor = $self->ctx->model('Editor')->get_by_name($username);
+        if (defined $editor) {
+            $field->add_error(l('Please choose another username, this one is already taken'));
```
Freso: Punctuation?

lib/MusicBrainz/Server/Form/Role/ValidateUsername.pm
reosarevok: I'd expect they are (this is the validation we were already doing for registering, I've just moved it to a role). Not sure about punctuation, https://github.com/metabrainz/musicbrainz-server/search?q=add_error&type=Code&utf8=%E2%9C%93 suggests the answer is "sometimes"

Freso: Well, if it's "sometimes", let's move towards more consistently using proper grammar than now. :)

lib/MusicBrainz/Server/Form/Role/ValidateUsername.pm
```diff
@@ -0,0 +1,22 @@
+package MusicBrainz::Server::Form::Role::ValidateUsername;
```
chirlu: I don’t think this is appropriate to have as a role. Make it a helper function somewhere.

lib/MusicBrainz/Server/Form/User/EditProfile.pm
```diff
@@ -7,9 +7,15 @@ use MusicBrainz::Server::Translation qw( l ln );
 use MusicBrainz::Server::Validation qw( is_valid_url );
 
 extends 'MusicBrainz::Server::Form';
+with 'MusicBrainz::Server::Form::Role::ValidateUsername';
 
 has '+name' => ( default => 'profile' );
 
+has_field 'username' => (
+    type      => 'Text',
+    maxlength => 64
```
chirlu: Should have a trailing comma, so that adding more entries later doesn’t require changing this line.

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -35,22 +35,22 @@ sub edit_user : Path('/admin/user/edit') Args(1) RequireAuth HiddenOnSlaves
...
         form => 'User::EditProfile',
         item => {
-            email            => $user->email,
+            username          => $user->name,
+            email             => $user->email,
             skip_verification => 0,
-            website            => $user->website,
-            biography        => $user->biography
+            website           => $user->website,
+            biography         => $user->biography
```
chirlu: I wouldn’t bother with this optical alignment stuff. If you do, make it a separate commit please.

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -35,22 +35,22 @@ sub edit_user : Path('/admin/user/edit') Args(1) RequireAuth HiddenOnSlaves
...
         },
     );
 
     if ($c->form_posted) {
-        if ($form->submitted_and_valid($c->req->params )) {
+        if ($form->submitted_and_valid($c->req->params) 
+            && $form2->submitted_and_valid($c->req->params )) {
```
chirlu: Combining the `if`s changes the semantics, unless both forms will always be submitted at the same time. Are you sure that is the case? Why?

lib/MusicBrainz/Server/Form/User/EditProfile.pm
```diff
@@ -7,9 +7,15 @@ use MusicBrainz::Server::Translation qw( l ln );
 use MusicBrainz::Server::Validation qw( is_valid_url );
 
 extends 'MusicBrainz::Server::Form';
+with 'MusicBrainz::Server::Form::Role::ValidateUsername';
 
 has '+name' => ( default => 'profile' );
 
+has_field 'username' => (
```
chirlu: I expect this will let everyone change their own username. (If not, how/where is it prevented?)

lib/MusicBrainz/Server/Controller/Admin.pm
reosarevok: As far as I can tell, both forms *should* always be submitted at the same time. There's only one form element, with one submit button, pressing Enter anywhere submits both forms from what I've tested, and the separation into two forms seems to exist only in the FormHandler part of this. But I'm not super experienced with this, so I might be missing something. Am I?

lib/MusicBrainz/Server/Form/User/EditProfile.pm
reosarevok: It's not, specifically, right now, other than by not having a username field in the template on the non-Admin page. Are you thinking we need to implement an extra check in case someone tries to change anyway it by post request or something to that effect?

lib/MusicBrainz/Server/Form/Role/ValidateUsername.pm
reosarevok: Any suggestions as to where this "somewhere" should be?

mwiencek: I think a role could make sense for username behavior in general, so also adding `has_field 'username'` here at least, assuming the `required` attribute works in both places. But, just moving the function to lib/MusicBrainz/Server/Validation.pm is sane too.

mwiencek: Or maybe lib/MusicBrainz/Server/Form/Utils.pm instead, since the Validation module doesn't really contain anything that operates on FormHandler fields directly right now. Having a generic "Form::Role::Username" still makes sense too IMHO.

lib/MusicBrainz/Server/Form/Role/ValidateUsername.pm
```diff
@@ -0,0 +1,24 @@
...
+use HTML::FormHandler::Moose::Role;
+use MusicBrainz::Server::Translation qw( l );
+
+sub validate_username
+{
+    my ($self, $field) = @_;
+
+    my $username = $field->value;
+    if (defined $username) {
+        unless ($self->ctx->user->name == $username) {
```
mwiencek: The `validate_username` function that was in the Register form didn't have this check...does account creation still work? I expect it would either ISE or not run the validation.

lib/MusicBrainz/Server/Form/User/EditProfile.pm
mwiencek: We'll need a check for that in `Controller::Account`, I think. I would just `$c->detach('/error_403')` if the username changed and the user isn't an account admin.

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -76,11 +76,11 @@ sub edit_user : Path('/admin/user/edit') Args(1) RequireAuth HiddenOnSlaves
                     $c->forward('/discourse/sync_sso', [$user]);
                 }
             }
+            
+	        $c->flash->{message} = l('User successfully edited.');
```
mwiencek: There are some tab characters at the beginning of these 3 lines.

lib/MusicBrainz/Server/Controller/Admin.pm
mwiencek: That sounds correct, i.e. they should both be submitted at the same time. It looks like it was previously the case that you'd get the redirect and "User successfully edited" banner even if one of the two forms contained errors.

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -297,6 +297,13 @@ sub edit : Local RequireAuth DenyWhenReadonly {
             $form->value
         );
 
+        my $old_username = $editor->username;
+        my $new_username = $form->field('username')->value;
+              
+        if (defined $new_username && $new_username ne $old_username) {
+            $c->detach('/error_403');            
+        }
+
```
reosarevok: Can't test this, because the account/edit form seems unrelatedly bugged (about languages) but thought I'd put it up so you could say if it looks reasonable or not at least.

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -297,6 +297,13 @@ sub edit : Local RequireAuth DenyWhenReadonly {
             $form->value
         );
 
+        my $old_username = $editor->username;
```
mwiencek: That should be `->name` (although this code doesn't actually run yet, due to an error below).

lib/MusicBrainz/Server/Form/User/EditProfile.pm
```diff
@@ -2,14 +2,21 @@ package MusicBrainz::Server::Form::User::EditProfile;
...
 use MusicBrainz::Server::Translation qw( l ln );
 use MusicBrainz::Server::Validation qw( is_valid_url );
 
 extends 'MusicBrainz::Server::Form';
 
 has '+name' => ( default => 'profile' );
 
+has_field 'username' => (
+    type      => 'Text',
+    required  => 1,
```
mwiencek: It looks like the `required` field actually breaks the profile page, because since there's no username field on the page, nothing is submitted and an invisible error occurs.

lib/MusicBrainz/Server/Form/User/EditProfile.pm
```diff
@@ -2,14 +2,21 @@ package MusicBrainz::Server::Form::User::EditProfile;
 
 use HTML::FormHandler::Moose;
 use List::MoreUtils qw( any all );
-use MusicBrainz::Server::Form::Utils qw( language_options select_options_tree );
+use MusicBrainz::Server::Form::Utils qw( language_options select_options_tree validate_username);
```
mwiencek: Nitpick: space is missing before the closing `)` like the rest of the imports (same problem in lib/MusicBrainz/Server/Form/User/Register.pm).

lib/MusicBrainz/Server/Form/User/EditProfile.pm
mwiencek: Since `required` is important on admin/edit_user, a simple fix would be to add an `<input type="hidden />` containing the username on the profile page.

lib/MusicBrainz/Server/Form/Utils.pm
```diff
@@ -231,6 +232,25 @@ sub indentation {
     return "\N{NO-BREAK SPACE}" x (3 * $level);
 }
 
+sub validate_username {
+    my ($self) = @_;
+
+    my $username = $self->value;
+    my $previous_username = $self->init_value;
+
+    if (defined $username) {
```
mwiencek: I think `$username` should always be defined, since it's marked as `required` in both forms?

reosarevok: That does sound reasonable, yeah... Wonder why the check was here originally then, given this was only used in the registration page!

mwiencek: Hm, don't trust me too much on that, it's possible it runs all the validation handlers even if it fails the `required` one.

gentlecat:
Not directly related to this PR, but I'm pretty sure that username is used as a reference to a user in all the other projects (ours) that rely on OAuth. So if username changes it will affect ability to access accounts in AB, CB, etc. In the worst case, if a MusicBrainz user gets username of an old account, they'll have access to it in other projects.

Don't remember if there is a user ID or some other permanent reference that can be used as an alternative (and is used in OAuth).

mwiencek:
Yes, we set `sub` (which is supposed to never be reassigned) to the username, which is pretty bad, but we can't change it without breaking every OAuth client. We do also provide the [user ID in the userinfo endpoint](https://github.com/metabrainz/musicbrainz-server/blob/b138e0d1fc5710e14ceee8f8e388f3894411683f/lib/MusicBrainz/Server/Controller/OAuth2.pm#L359-L359), but that's not a standard field. Preventing usernames from being reused in MBS would be a schema change.

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -289,14 +289,35 @@ sub edit : Local RequireAuth DenyWhenReadonly {
...
         $c->model('Editor')->update_profile(
             $editor,
             $form->value
         );
 
+        my $old_username = $editor->name;
+        my $new_username = $form->field('username')->value;
+              
+        if (defined $new_username && $new_username ne $old_username) {
+            $c->detach('/error_403');
```
mwiencek: Needs to come before the `update_profile` call. :)

lib/MusicBrainz/Server/Form/User/Register.pm
```diff
@@ -10,7 +11,8 @@ has '+name' => ( default => 'register' );
 has_field 'username' => (
     type      => 'Text',
     required  => 1,
-    maxlength => 64
+    maxlength => 64,
+    validate_method => \&validate_username
```
chirlu: Trailing comma.

reosarevok: Nothing else in this file has the trailing comma, so I didn't add it because then I'd have to add it everywhere and then it wouldn't belong in this PR anymore I guess...

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -289,14 +289,35 @@ sub edit : Local RequireAuth DenyWhenReadonly {
...
+            username          => $editor->name,
+            email             => $editor->email,
+            skip_verification => 0,
+            website           => $editor->website,
+            biography         => $editor->biography,
+            gender_id         => $editor->gender_id,
+            area_id           => $editor->area_id,
+            area              => $editor->area,
+            birth_date        => $editor->birth_date,
+            languages         => $editor->languages
```
chirlu: Trailing comma.

lib/MusicBrainz/Server/Form/User/EditProfile.pm
chirlu: Works of course, but is very ugly IMO.

mwiencek: My other idea was to split `Form::User::EditProfile`, with common functionality in a role.

lib/MusicBrainz/Server/Form/User/Register.pm
mwiencek: It's def. okay to add the trailing commas in new lines or lines you're changing anyway, even if other parts of the file don't use it. But, it's not like I'll reject a PR just 'cause it didn't have trailing commas*—it's just better to add them.

*If we checked code style in the tests using e.g. Perl::Critic, it'd be much easier to say "here, fix this list of issues if you want the build to pass," than having to make tedious comments on GitHub which usually aren't worth my time (and not because I don't care).

PR #461, opened 2017-02-12
MBS-9166: make guess case ignore "Early" unless already in parens

reosarevok:
This implements the most conservative version of [MBS-9166](https://tickets.metabrainz.org/browse/MBS-9166) (early will still get lowercased if it's inside parenthesis, but not otherwise).

I get the feeling "dialogue" could get the same treatment too, but letting that one be for now.


chirlu:
I would prefer a commit description that properly describes what the change is. Noone will know what “ignore ‘Early’” means: Ignore it by removing it from the text? Ignore it by not changing its casing? Ignore it by treating it like “late”? Ignore it in a different way?

reosarevok:
Better?

PR #462, opened 2017-02-12
MBS-9164: Show edit history (but not notes) when not logged in

reosarevok:
Submitting a WIP PR for [MBS-9164](https://tickets.metabrainz.org/browse/MBS-9164) per bitmap's request. 

Displaying edits seems to work pretty much fine, although I don't discount the possibility of there being some issue I haven't noticed yet. 

Edit search needs to be restricted properly so that it won't actually submit when the non-logged-in options are selected, and it also needs to be restricted so that if someone constructs a URL (or clicks a link) involving the logged-in-only options when not logged in, it will actually reject it and ask for login. 

I haven't looked at tests yet, so I expect this to break a ton of them.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/EditListing.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/EditListing.pm
@@ -6,7 +6,7 @@ use MusicBrainz::Server::Constants qw( :edit_status );
 
 requires '_load_paged';
 
-sub edits : Chained('load') PathPart RequireAuth
+sub edits : Chained('load') PathPart
 {
     my ($self, $c) = @_;
     $self->_list($c, sub {
@@ -26,11 +26,11 @@ sub edits : Chained('load') PathPart RequireAuth
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
               'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              'conditions.0.user_id' => $c->user->id },
+              $c->user_exists ? ('conditions.0.user_id' => $c->user->id) : () },
     );
 }
 
-sub open_edits : Chained('load') PathPart RequireAuth
+sub open_edits : Chained('load') PathPart
 {
     my ($self, $c) = @_;
     $self->_list($c, sub {
@@ -51,7 +51,7 @@ sub open_edits : Chained('load') PathPart RequireAuth
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
               'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              'conditions.0.user_id' => $c->user->id,
+              $c->user_exists ? ('conditions.0.user_id' => $c->user->id) : (),
               'conditions.1.field' => 'status',
               'conditions.1.operator' => '=',
               'conditions.1.args' => $STATUS_OPEN },
--- a/lib/MusicBrainz/Server/Controller/Role/EditListing.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/EditListing.pm
@@ -25,8 +25,7 @@ sub edits : Chained('load') PathPart
               'conditions.0.field' => model_to_type( $self->{model} ),
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
-              'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              $c->user_exists ? ('conditions.0.user_id' => $c->user->id) : () },
+              'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id },
     );
 }
 
@@ -51,7 +50,6 @@ sub open_edits : Chained('load') PathPart
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
               'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              $c->user_exists ? ('conditions.0.user_id' => $c->user->id) : (),
               'conditions.1.field' => 'status',
               'conditions.1.operator' => '=',
               'conditions.1.args' => $STATUS_OPEN },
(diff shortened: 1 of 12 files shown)


lib/MusicBrainz/Server/Controller/Role/EditListing.pm
```diff
@@ -26,11 +26,11 @@ sub edits : Chained('load') PathPart RequireAuth
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
               'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              'conditions.0.user_id' => $c->user->id },
+              defined $c->user ? ('conditions.0.user_id' => $c->user->id) : () },
```
chirlu: There is `$c->user_exists`. (Applies not only here.)

root/edit/index.tt
```diff
@@ -60,18 +62,26 @@
             [% END %]
         [% ELSE %]
             <p>
-                [%- l('You must be logged in to vote on edits') -%]
-                [%- request_login() -%]
+                [%- l('You must be logged in to vote on edits.') %]
+                [% request_login() -%]
```
chirlu: Why are you adding whitespace to the output?

reosarevok: Because I want this to be "You must be logged in to vote on edits. Log in", and not "You must be logged in to vote on editsLog in", as it was before.

chirlu: OK. However, using `=%]` and `[%=` would still be better because it will squash the whitespace into a single space.

Furthermore, since this is unrelated to the main feature being implemented, this would be perfect for a separate commit (which would also give you room to explain why you are doing this in the commit message).

root/edit/edit_header.tt
```diff
@@ -73,8 +78,15 @@
...
-        [% l('Edit by {editor}',
-            { editor => link_entity(edit.editor) }) %]
-        [%- editor_type_info(edit.editor) -%]
+        [%~ IF c.user_exists ~%]
+            [% l('Edit by {editor}',
+                { editor => link_entity(edit.editor) }) %]
+            [%- editor_type_info(edit.editor) -%]
+        [%~ ELSE ~%]
+            [% # We show editor type since knowing if it's, say, a bot edit is useful %]
+            [% l('Edit by a hidden editor') ~%]
```
mwiencek: This looked/sounded kinda weird to me, like everyone hid themselves. Something like "Editor hidden (log in to see who)" seems more straightforward, maybe?

root/edit/components/EditSidebar.js
```diff
@@ -78,9 +79,13 @@ const EditSidebar = ({$c, edit}: Props) => (
...
     <p>
-      <a href={`/edit/${edit.id}/data`}>
-        <bdi>{l('Raw edit data for this edit')}</bdi>
-      </a>
+      {$c.user_exists ? (
+        <a href={`/edit/${edit.id}/data`}>
+          <bdi>{l('Raw edit data for this edit')}</bdi>
+        </a>
+      ) : (
+        <RequestLogin $c={$c} text={l('Log in to see raw edit data')} />
```
mwiencek: We probably shouldn't display anything in this case, since it's just technical data for developers. Anyone who needs this and isn't logged in knows what to do, while someone without an account isn't going to have any idea what this means or is for.

root/edit/edit_header.tt
```diff
@@ -31,7 +31,12 @@
                     </td>
                     <td class="vote-count">
                         [%~ IF edit.latest_vote_for_editor(c.user.id) || c.user.id == edit.editor_id || !edit.is_open ~%]
-                            <div>[%- vote_tally(edit) -%]</div>
+                            <div>
+                                [%~ IF !c.user_exists ~%]
+                                    <strong>[%- l('Vote tally: ') -%]</strong>
```
mwiencek: This adds a new string, since the one in root/edit/index.tt doesn't have a space after the `:`

Actually, we also have a string without the colon in root/edit/search_macros.tt, so perhaps we should replace both this and the one in root/edit/index.tt to use that + add_colon.

lib/MusicBrainz/Server/Controller/Role/EditListing.pm
```diff
@@ -51,7 +51,7 @@ sub open_edits : Chained('load') PathPart RequireAuth
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
               'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              'conditions.0.user_id' => $c->user->id,
+              $c->user_exists ? ('conditions.0.user_id' => $c->user->id) : (),
```
mwiencek: (Same comment about `user_id`)

lib/MusicBrainz/Server/Controller/Role/EditListing.pm
```diff
@@ -26,11 +26,11 @@ sub edits : Chained('load') PathPart RequireAuth
               'conditions.0.operator' => '=',
               'conditions.0.name' => $c->stash->{ $self->{entity_name} }->name,
               'conditions.0.args.0' => $c->stash->{ $self->{entity_name} }->id,
-              'conditions.0.user_id' => $c->user->id },
+              $c->user_exists ? ('conditions.0.user_id' => $c->user->id) : () },
```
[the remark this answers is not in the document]
reosarevok: The line `'conditions.0.user_id' => $c->user->id },` still exists in master, so I imagine that didn't get removed? I'm giving it user->id here, not user_id.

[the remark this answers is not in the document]
mwiencek: I mean, I think the line you're changing here should've been removed with b103fd6004bcf0c8ed888b494e0fc04b949a38c8, i.e. the `conditions.0.user_id` field isn't used by the search at all.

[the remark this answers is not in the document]
reosarevok: Oooh. @yvanzo: since that was your PR, please confirm.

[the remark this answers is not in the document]
yvanzo: Good catch!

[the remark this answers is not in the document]
reosarevok: Ok, removing it! :)

PR #465, opened 2017-02-16
MBS-9250: Places map displays over header menus

loujine:
The menu z-index must be above the one of leaflet map (z-index 400)

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -275,7 +275,7 @@ pre code {
                 ul {
                     position: absolute;
                     left: -10000px;
-                    z-index: 9;
+                    z-index: 500;
                     padding: 1px;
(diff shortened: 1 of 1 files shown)


loujine:
Looking quickly, there are no new CSS conflict in changing this z-index from 9 to 500
- styles/debug.css (z = 10) is never overlapping the opened menu
- the react tooltip component stays above (zIndex = 1000)

I didn't spend too much time on it though, so different eyes are welcome :)

One consequence I noticed is that the header menus were behind relationships modal on the edit page (e.g. opening an instrument relationship on a recording edit page and then the "About us" menu) — with this patch the header menu is above the relationship modal

mwiencek:
> with this patch the header menu is above the relationship modal

That actually makes more sense to me.

PR #466, opened 2017-02-16
MBS-9255: Show recordings length in relationships tables

loujine:
i.e. on

 - Work overview page
 - Artist relationships page
 - Label relationships page
 - Place performances page

### DIFF
--- a/root/components/relationships-table.tt
+++ b/root/components/relationships-table.tt
@@ -8,13 +8,14 @@
                 [% IF show_credits -%]<th>[% l('Credited As') %]</th>[%- END %]
                 <th>[% l('Attributes') %]</th>
                 <th>[% l('Artist') %]</th>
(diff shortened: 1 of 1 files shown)


loujine:
First working draft
Maybe it should be improved by not displaying the column at all if no recordings are present in the relationships table (probably the case on Labels)

mwiencek:
I think we should possibly have separate tables per target entity type, otherwise we're always going to have the problem of useless columns (currently the case with "Artist"). But that seems outside the scope of this PR. It was suggested in https://github.com/metabrainz/musicbrainz-server/pull/337#issuecomment-232444072 to get rid of the table entirely.

If you can find a non-complicated way to hide the column if there are no recordings listed, that would seem like an improvement too.

jesus2099:
:open_mouth: :tada: 
That’s super good news, @loujine!

PR #467, opened 2017-02-16
MBS-9246: Load extra data for Artist Credit field

y-van-z:
Fix [MBS-9246 Tool-tips displaying "undefined" text](https://tickets.metabrainz.org/browse/MBS-9246).

Previously, only artist name, id, gid, and type were loaded for preview in the Artist Credit editor.  Consequently, tool-tips did not contain sort name or disambiguation comment (but _undefined_ instead), and artists with pending edits were not highlighted. It affects Artist Credit field editor of editing forms for recording, release-group, artist credit, and artist split. Only Artist Credit field of release editor is unaffected, because it is initialized in a different way using React.

This patch loads all JSON-serialized artist fields including sort name, disambiguation comment, and status regarding pending edits to the Artist Credit form field. All of the above mentioned bugs are fixed, ensuring consistency with release editor.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
+++ b/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
@@ -122,8 +122,8 @@ sub json {
 
     my $artists = $c->model('Artist')->get_by_ids(map { $_->{artist}->{id} } @$names);
     for my $name (@$names) {
-        $name->{artist}->{entityType} = 'artist';
-        $name->{artist}->{gid} = $artists->{$name->{artist}->{id}}->gid if $artists->{$name->{artist}->{id}};
+        my $artist = $artists->{$name->{artist}->{id}};
+        $name->{artist} = $artist->TO_JSON if $artist;
         $name->{joinPhrase} = delete $name->{join_phrase};
     }
 


lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
```diff
@@ -123,7 +123,14 @@ sub json {
     my $artists = $c->model('Artist')->get_by_ids(map { $_->{artist}->{id} } @$names);
     for my $name (@$names) {
         $name->{artist}->{entityType} = 'artist';
-        $name->{artist}->{gid} = $artists->{$name->{artist}->{id}}->gid if $artists->{$name->{artist}->{id}};
+        my $artist = $artists->{$name->{artist}->{id}};
+        if ($artist)
+        {
```
mwiencek: Nitpick: we prefer opening brackets to always be on the same line as the control statement.

lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
```diff
@@ -123,7 +123,14 @@ sub json {
...
+            $name->{artist}->{gid} = $artist->{gid};
+            $name->{artist}->{sortName} = $artist->{sort_name};
+            $name->{artist}->{comment} = $artist->{comment};
+            $name->{artist}->{editsPending} = $artist->{edits_pending};
```
mwiencek: Looks good, but there's already an `$artist->TO_JSON` method which should encode things in the same format as here—probably with additional fields that aren't needed, but using it could prevent future issues if the format changes.

As an aside, for blessed Moose objects we always use the accessor methods (`$artist->gid` vs. `$artist->{gid}`). Both will work, but the accessors make it easier to see it's a blessed instance, not a hash reference (and are also easier to type).

lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
y-van-z: Meta Nitpick: “Curly braces are always on their own line” according to [Server Coding Style](https://musicbrainz.org/doc/Development/Server_Coding_Style) 😆

lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
y-van-z: Ok. At first glance Artist JSON seemed to output a lot more details than what is useful to display Artists Credit, but I will look into it again. Thanks for the blessed tip!

lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
mwiencek: Whooops...at least the heading on that page is accurate: "Status: This page is work in progress and has very questionable content." :)

I'll fix it to be less questionable.

lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
y-van-z: It is indeed reasonable to use full JSON export. Done in https://github.com/metabrainz/musicbrainz-server/pull/467/commits/3cd70df0a8552d592c5e0d15f3f9e892d91ce475.

PR #468, opened 2017-02-16
MBS-9237: Fix complex artist credit detection

y-van-z:
Fix [MBS-9237: Don't immediately fill back the default AC while editor is typing](https://tickets.metabrainz.org/browse/MBS-9237).

Previously, empty Artist Credits were mistakenly considered as simple, so triggering the Artist field Autocomplete to fill back user emptied Artist Credit in the Artist Credit editor bubble. This is because empty artist credit is reduced to artist name.

This patch fixes the `isComplexArtistCredit` to consider empty Artist Credit as complex.  This prevents the automatic fill-back of Artist Credit field in the Artist Credit editor to mistakenly happen before the field looses focus. See `onNameBlur` for proper fill-back.

### DIFF
--- a/root/static/scripts/common/immutable-entities.js
+++ b/root/static/scripts/common/immutable-entities.js
@@ -37,7 +37,7 @@ const reduceArtistCredit = ac => ac.names.reduce(reduceName, '');
 const isComplexArtistCredit = function (ac) {
   const firstName = ac.names.get(0);
   if (firstName && hasArtist(firstName)) {
-     return firstName.artist.name !== reduceArtistCredit(ac);
+     return !nonEmpty(firstName.name) || firstName.artist.name !== reduceArtistCredit(ac);
   }
   return false;
 };
--- a/root/static/scripts/tests/common/immutable-entities.js
+++ b/root/static/scripts/tests/common/immutable-entities.js
@@ -5,7 +5,11 @@
 
 const test = require('tape');
 
+const Immutable = require('immutable');
+
 const {
+    ArtistCredit,
+    ArtistCreditName,
     artistCreditFromArray,
     artistCreditsAreEqual,
     isComplexArtistCredit,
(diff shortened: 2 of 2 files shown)


root/static/scripts/common/immutable-entities.js
```diff
@@ -37,7 +37,7 @@ const reduceArtistCredit = ac => ac.names.reduce(reduceName, '');
 const isComplexArtistCredit = function (ac) {
   const firstName = ac.names.get(0);
   if (firstName && hasArtist(firstName)) {
-     return firstName.artist.name !== reduceArtistCredit(ac);
+     return firstName.artist.name !== reduceArtistCredit(ac) || !nonEmpty(firstName.name);
```
mwiencek: Yeah, I think this makes sense. My only recommendation would be to take advantage of short-circuit evaluation and put the `!nonEmpty(firstName.name)` check before the first one, since `reduceArtistCredit` is more expensive.

mwiencek:
Test failures look unrelated, despite being for ACs. Those are the same ones that happened in one of my PRs and then disappeared.

root/static/scripts/common/immutable-entities.js
y-van-z: Done in https://github.com/metabrainz/musicbrainz-server/pull/468/commits/37dd5949ef47d7606c78f2b3b959d0b8d31c272f

mwiencek:
One more request (sorry :)). We have tests for `isComplexArtistCredit` in root/static/scripts/tests/common/immutable-entities.js and it would be great to see a test case added for this new behavior.

y-van-z:
Thanks for spotting this. Done in https://github.com/metabrainz/musicbrainz-server/pull/468/commits/bd989199afcd2142575d6fb472399412ee47353f.

PR #470, opened 2017-02-17
MBS-9129: Adjust JSON serialization for RG ratings on release request.

Zastai:
The first commit is straightforward - a release has no ratings so should not try to serialize them.

The second part aims to replicate the XML serializer's behaviour - a request for ratings on a release is "forwarded" to the release's RG (if it's also requested).
However, the JSON serializer is made up of a bunch of disjoint sources, and the bit that handles ratings has no way of knowing it's handling a release group that is being serialized as part of a release request.
If I change MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Rating to ignore the toplevel flag, all release groups (including those in (requested) relationships) in all requests will serialize their ratings; instead I (ab)use $inc to pass a special flag for it to pick up.

I'm not very happy with that, but the only other way I see is to change the basic serialize() signature and either add an extra parameter, or change the semantics of the current $toplevel flag (e.g. by calling it $flags, and changing tests of $toplevel to checks for $flags->{TOPLEVEL}). Other approaches/suggestions welcome.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
@@ -7,7 +7,6 @@ extends 'MusicBrainz::Server::WebService::Serializer::JSON::2';
 with 'MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Aliases';
 with 'MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Annotation';
 with 'MusicBrainz::Server::WebService::Serializer::JSON::2::Role::GID';
-with 'MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Rating';
 with 'MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Relationships';
 with 'MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Tags';
 
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
@@ -81,8 +81,15 @@ sub serialize
     $body{collections} = list_of($entity, $inc, $stash, "collections")
         if $inc && ($inc->collections || $inc->user_collections);
 
-    $body{"release-group"} = serialize_entity($entity->release_group, $inc, $stash)
-        if $inc && $inc->release_groups;
+    if ($inc && $inc->release_groups)
+    {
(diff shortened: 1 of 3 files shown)


mwiencek:
The better method would be to use `local` within a block to [temporarily override a package-level variable](https://github.com/metabrainz/musicbrainz-server/blob/ab4e11cf88f80014d97a3629b6dbc0901b5bffd5/lib/MusicBrainz/Server/WebService/XMLSerializer.pm#L676-L676).

I guess the only place to put the variable would be JSON/2/Role/Rating.pm. So you'd need something like this in JSON/2/Release.pm:

```perl
{
    local $MusicBrainz::Server::WebService::Serializer::JSON::2::Role::Rating::show_ratings = 0;
    # $show_ratings is only 0 within this block.
};
```

Zastai:
Good idea - will force-push a fixup commit using that approach.

Zastai:
That should do it. Tests don't seem to affected.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
```diff
@@ -82,8 +81,14 @@ sub serialize
     $body{collections} = list_of($entity, $inc, $stash, "collections")
         if $inc && ($inc->collections || $inc->user_collections);
 
-    $body{"release-group"} = serialize_entity($entity->release_group, $inc, $stash)
-        if $inc && $inc->release_groups;
+    if ($inc && $inc->release_groups)
+    {
+        # MBS-9129: If ratings are requested (even though a release doesn't have any), those of the release group should be
```
mwiencek: I prefer always limiting comments to <80 chars per line, since it makes them much easier to read. (Usually code too, but I don't strictly follow that.)

mwiencek:
BTW, is adding a test something you'd be willing to look into?

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
Zastai: I'm used to 132 as 80 can get very restrictive (not so much for perl, but for C++ it is), so my emacs is set up for that.
I can reformat those lines, no problem.

Zastai:
Sure, I can look at making a test, depending a little on what it all entails.
Will have a look at the .t files and see where I get.

Zastai:
Tests added.
Once everything else is fine, I'll squash 53e8036 into 5cee0a9 and force push.

mwiencek:
LGTM.

Zastai:
Comment fixup squashed. Probably ready for merge now.

PR #471, opened 2017-02-19
MBS-9215: Add entity type to attribute editing form

y-van-z:
Collection and series types admin editing form missed entity type.

This adds support for entering entity type without specific check.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Admin/Attributes.pm
+++ b/lib/MusicBrainz/Server/Form/Admin/Attributes.pm
@@ -1,12 +1,13 @@
 package MusicBrainz::Server::Form::Admin::Attributes;
 
 use HTML::FormHandler::Moose;
-use MusicBrainz::Server::Form::Utils qw( select_options_tree );
+use MusicBrainz::Server::Form::Utils qw( select_options select_options_tree );
+use MusicBrainz::Server::Constants qw( entities_with );
 
 extends 'MusicBrainz::Server::Form';
 with 'MusicBrainz::Server::Form::Role::Edit';
 
-sub edit_field_names { qw( parent_id child_order name description year has_discids free_text ) }
+sub edit_field_names { qw( parent_id child_order name description year has_discids free_text entity_type ) }
 
 has '+name' => ( default => 'attr' );
 
@@ -41,9 +42,18 @@ has_field 'free_text' => (
     type => 'Boolean',
 );
 
+has_field 'entity_type' => (
+    type => 'Select',
+);
+
 sub options_parent_id {
     my ($self, $model) = @_;
     return select_options_tree($self->ctx, $self->ctx->stash->{model}, accessor => 'name');
+}
+
+sub options_entity_type {
+    my ($self) = @_;
+    return map { $_ => $_ } sort { $a cmp $b } entities_with('collections');
 }
 
 1;
--- a/root/admin/attributes/form.tt
+++ b/root/admin/attributes/form.tt
@@ -36,6 +36,16 @@
     [% ELSE %]
         [%~ IF model == "CollectionType" || model == "SeriesType" ~%]
             [% form_row_select(r, 'entity_type', add_colon(l('Entity type'))) %]
+            [%~ IF c.action.name == "edit" ~%]
+                <script>//<![CDATA[
+                    $(function () {
+                        $('#id-attr\\.entity_type').prop('disabled', true);
+                        $('form[action="[% c.req.uri %]"]').bind('submit', function () {
+                            $('#id-attr\\.entity_type').prop('disabled', false);
+                        });
+                    });
+                //]]></script>
+            [%~ END ~%]
         [%~ END ~%]
 
         [% form_row_select(r, 'parent_id', add_colon(l('Parent'))) %]
--- a/root/admin/attributes/form.tt
+++ b/root/admin/attributes/form.tt
@@ -34,6 +34,10 @@
         </p>
 
     [% ELSE %]
+        [%~ IF model == "CollectionType" || model == "SeriesType" ~%]
+            [% form_row_select(r, 'entity_type', add_colon(l('Entity type'))) %]
+        [%~ END ~%]
+
         [% form_row_select(r, 'parent_id', add_colon(l('Parent'))) %]
 
         [% WRAPPER form_row %]


lib/MusicBrainz/Server/Form/Admin/Attributes.pm
```diff
@@ -41,11 +42,20 @@ has_field 'free_text' => (
...
+);
+
 sub options_parent_id {
     my ($self, $model) = @_;
     return select_options_tree($self->ctx, $self->ctx->stash->{model}, accessor => 'name');
 }
 
+sub options_entity_type {
+    my ($self) = @_;
+    return map { $_ => $_ } sort { $a cmp $b } entities_with('collections');
```
mwiencek: Apparently [this works with lists too](http://search.cpan.org/dist/HTML-FormHandler-0.40057/lib/HTML/FormHandler/Field/Select.pm) (elsewhere we use array refs).

mwiencek:
Ideally, it would validate that the parent attribute has the same entity type, but hopefully admins won't make that mistake.

y-van-z:
I would prefer to simply forbid modifying entity type field for any existing attribute type. This could be checked in [Controller::Admin::Attributes](https://github.com/y-van-z/musicbrainz-server/blob/3db840307ea4b5501b7cc773308c380ff35880af/lib/MusicBrainz/Server/Controller/Admin/Attributes.pm).

On a side note, there are many other ways for admins to create flawed attribute types:

- change an entity type into another one different from the one of its parent/children
- create a new attribute type with an entity type different from its parent’s one
- change a parent id to an attribute type with a different entity type
- loop with parent ids

Hopefully they can be fixed by the same admins using the same form and, more importantly, they don’t corrupt existing entities.

mwiencek:
Agreed on preventing modifying the entity type altogether, since that is most likely to cause issues, and doesn't really have a good use case. Good point about the others. I think the parent/child issues mentioned are probably not worth validating, since they wouldn't cause anything terrible if they happened.

y-van-z:
Done in https://github.com/metabrainz/musicbrainz-server/pull/471/commits/d7a3993fdbdce48f08bf39a006ba6414e6bbef19

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -93,12 +95,22 @@ sub edit : Chained('attribute_base') Args(1) RequireAuth(account_admin) {
...
-        });
-
-        $c->response->redirect($c->uri_for('/admin/attributes', $model));
-        $c->detach;
+        my $old_entity_type = $attr->entity_type;
+        my $new_entity_type = $form->field("entity_type")->value;
+
+        if ($old_entity_type ne $new_entity_type) {
+            $form->field('entity_type')->add_error(
+                l('Entity type \'{entity_type}\' cannot be changed; please delete and readd attribute instead.',
```
mwiencek: I think the `<select>` should be grayed/disabled on the editing form, and we should simply filter out `entity_type` from `$form->edit_fields` before calling `update` below. We're not worried about admins abusing the form, so no need to give translators a new message nobody will ever see. :)

y-van-z: About this, messages from admin interface should probably not be localized. Right?

mwiencek: Hmm, I think it's probably right for them to be localized. Even though all the admins know English, they might have a different language selected in the UI and it would be weird to make an exception here. But I'd avoid adding strings to the admin UI where not necessary, or I'd make them simple if required.

y-van-z: Done in https://github.com/metabrainz/musicbrainz-server/pull/471/commits/73daae629f043dae0ee38d7e12867402c91cb093.
On a side note: I disabled the `<select>` element using JS because TT seems to ignore `disabled` attribute.

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -93,6 +93,8 @@ sub edit : Chained('attribute_base') Args(1) RequireAuth(account_admin) {
     my $form = $c->form( form => $form_name, init_object => $attr );
 
     if ($c->form_posted && $form->process( params => $c->req->params )) {
+        $form->field("entity_type")->value( $attr->entity_type );
```
mwiencek: There's a small problem: disabled fields aren't submitted by the browser, so if an error occurs in another field (try entering non-numeric characters in child order), the page will reload and the `<select>` will have no selection. To fix that, I think this should go above `$form->process`.

root/admin/attributes/form.tt
```diff
@@ -34,6 +34,17 @@
         </p>
 
     [% ELSE %]
+        [%~ IF model == "CollectionType" || model == "SeriesType" ~%]
+            [% form_row_select(r, 'entity_type', add_colon(l('Entity type'))) %]
+            [%~ IF r.form.field('entity_type').value.size ~%]
```
mwiencek: Also related to errors happening in other fields: this will disable the field even if you're on the "create" form. But that's a minor issue we can probably ignore.

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
y-van-z: I removed it in https://github.com/metabrainz/musicbrainz-server/pull/471/commits/cde0df93294647929c713687922ce70b55df3d01 (where field `entity_type` input is enabled just before form submission) since it is unnecessary anyway, unless we want proper validation which should then take place in a separate package for attributes with `entity_type` field.

root/admin/attributes/form.tt
y-van-z: Thank you for having spotted this issue.
I replaced the faulty condition by a cleaner one on `c.action.name` in https://github.com/metabrainz/musicbrainz-server/pull/471/commits/cde0df93294647929c713687922ce70b55df3d01.

y-van-z:
Failed tests are unrelated to this change.

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
mwiencek: I think it was fine to keep it, but that's good enough for me. :)

PR #472, opened 2017-02-19
MBS-9219: Fix editing known languages field in user profile form

y-van-z:
Fix bug [MBS-9219:[Multiple languages] (Basic) is automatically added when editing/saving an editor profile](https://tickets.metabrainz.org/browse/MBS-9219) and clean template code up a bit.

### DIFF
--- a/root/account/edit.tt
+++ b/root/account/edit.tt
@@ -77,6 +77,8 @@
 
       <script>//<![CDATA[
       $(function () {
+          $(document.body).append( $('#add-language-template') );
+
           var languageCount = $('li.language').size();
 
           $('#edit-profile-form').on('click', 'button.remove', function (event) {
--- a/root/account/edit.tt
+++ b/root/account/edit.tt
@@ -65,8 +65,8 @@
             </span>
             [% extra_field = form.field('languages').field(form.field('languages').add_extra(1) - 1) %]
             <span id="add-language-template" style="display: none">
-              [% r.select(extra_field.field('language_id'), id => 'id-language-id-template', class => 'language_id', name => 'language-id-template') %]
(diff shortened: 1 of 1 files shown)


root/account/edit.tt
```diff
@@ -65,8 +65,8 @@
             </span>
             [% extra_field = form.field('languages').field(form.field('languages').add_extra(1) - 1) %]
             <span id="add-language-template" style="display: none">
-              [% r.select(extra_field.field('language_id'), id => 'fluency-template', class => 'language_id',) %]
-              [% r.select(extra_field.field('fluency'), class => 'fluency') %]
+              [% r.select(extra_field.field('language_id'), id => 'id-language-id-template', class => 'language_id', name => 'language-id-template') %]
+              [% r.select(extra_field.field('fluency'), id => 'id-fluency-template', class => 'fluency', name => 'fluency-template') %]
```
mwiencek: Something I'm wondering is, are these id and name attributes actually needed? It seems they aren't used anywhere and are just overridden in the script below, so we can just remove them.

y-van-z: Indeed `id`s are of no use. They had to be unique, hence the override, but removal is even better.
Only `name`s are needed for form submission. And yes, template’s `name` is worthless.

Cleaned up in https://github.com/metabrainz/musicbrainz-server/pull/472/commits/547b27e529989a708b1e61616f40fde7bc6736b3

y-van-z:
Cleaned up in https://github.com/metabrainz/musicbrainz-server/pull/472/commits/547b27e529989a708b1e61616f40fde7bc6736b3 (to be squashed with previous commit)

PR #474, opened 2017-02-20
MBS-9230: Hook up MusicBrainz to Sentry

mwiencek:
This seems to basically work for both client and server errors, but marking as WIP since I'd like to test it further, especially in the release editor.


lib/Catalyst/Plugin/Sentry.pm
```diff
@@ -0,0 +1,167 @@
...
+use DBDefs;
+use Devel::StackTrace;
+use File::Slurp qw( read_file );
+use IO::File;
+use Moose;
+use Sentry::Raven;
+
+our $suppress = 0;
+
+sub execute {
```
mwiencek: A lot of what this does should in theory be [handled by Sentry::Raven already](https://github.com/getsentry/perl-raven/blob/291fc8444258c2ea6112206051550ab036812fa7/lib/Sentry/Raven.pm#L229). Unfortunately, I haven't yet found a way to get that to work with Catalyst, but I'll keep trying.

mwiencek: Why it's a pain to use Sentry::Raven::capture_errors with Catalyst:

* It doesn't expose the captured error, so we can't e.g. check its class to
suppress timeouts.
* Using `next::method` within a callback doesn't do what is intended. While we
could ditch plugins and do error handling within `dispatch`, we still have
the first problem.

PR #475, opened 2017-02-21
MBS-9258: Map "Accept: application/json" to "fmt=json" for search server WS requests.

Zastai:
Only question is whether or not it's safe to modify the args entry in the stash, or whether I should make a copy before adding the fmt value.
Alternatively, maybe it's Format.pm that should handle this by guaranteeing that fmt will be set appropriately (it already looks at the Accept header).

### DIFF
--- a/t/lib/t/MusicBrainz/Server/WebService/Format.pm
+++ b/t/lib/t/MusicBrainz/Server/WebService/Format.pm
@@ -1,0 +1,130 @@
+package t::MusicBrainz::Server::WebService::Format;
+
+use JSON qw( decode_json encode_json );
+use Test::JSON import => [qw( is_valid_json is_json )];
+use Test::More;
+use Test::Routine;
+use Test::XML::SemanticCompare;
+
+with 't::Mechanize', 't::Context';
+
+test 'webservice request format handling (XML)' => sub {
+
+    my $test = shift;
+    MusicBrainz::Server::Test->prepare_test_database($test->c, '+webservice');
+
+    my $mech = $test->mech;
+
+    my $Test = Test::Builder->new();
+
+    my $expected = '<?xml version="1.0" encoding="UTF-8"?>
+<metadata xmlns="http://musicbrainz.org/ns/mmd-2.0#">
+    <artist type="Person" type-id="b6e035f4-3ce9-331c-97df-83397230b0df" id="472bc127-8861-45e8-bc9e-31e8dd32de7a">
+        <name>Distance</name><sort-name>Distance</sort-name>
+        <disambiguation>UK dubstep artist Greg Sanders</disambiguation>
+    </artist>
+</metadata>';
+
+    $Test->note('Accept: <blank>');
+    $mech->default_header('Accept' => '');
+    $mech->get('/ws/2/artist/472bc127-8861-45e8-bc9e-31e8dd32de7a');
+    ok($mech->success, 'request successful');
+    is_xml_same($mech->content, $expected);
+
+    $Test->note('Accept: */*');
+    $mech->default_header('Accept' => '*/*');
+    $mech->get('/ws/2/artist/472bc127-8861-45e8-bc9e-31e8dd32de7a');
+    ok($mech->success, 'request successful');
+    is_xml_same($mech->content, $expected);
+
+    $Test->note('Accept: application/xml');
+    $mech->default_header('Accept' => 'application/xml');
+    $mech->get('/ws/2/artist/472bc127-8861-45e8-bc9e-31e8dd32de7a');
+    ok($mech->success, 'request successful');
+    is_xml_same($mech->content, $expected);
+
+    $Test->note('fmt=xml');
+    $mech->default_header('Accept' => 'application/something-else');
+    $mech->get('/ws/2/artist/472bc127-8861-45e8-bc9e-31e8dd32de7a?fmt=xml');
+    ok($mech->success, 'request successful');
+    is_xml_same($mech->content, $expected);
+
+};
+
+test 'webservice request format handling (JSON)' => sub {
+
+    my $test = shift;
+    MusicBrainz::Server::Test->prepare_test_database($test->c, '+webservice');
+
+    my $mech = $test->mech;
+
+    my $Test = Test::Builder->new();
+
+    my $expected = {
+        id => '472bc127-8861-45e8-bc9e-31e8dd32de7a',
+        name => 'Distance',
+        'sort-name' => 'Distance',
+        country => JSON::null,
+        area => JSON::null,
+        begin_area => JSON::null,
+        end_area => JSON::null,
+        disambiguation => 'UK dubstep artist Greg Sanders',
+        'life-span' => {
+            begin => JSON::null,
+            end => JSON::null,
+            ended => JSON::false,
+        },
+        type => 'Person',
+        'type-id' => 'b6e035f4-3ce9-331c-97df-83397230b0df',
+        ipis => [],
+        isnis => [],
+        gender => JSON::null,
+        'gender-id' => JSON::null,
+    };
+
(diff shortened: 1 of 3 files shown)


mwiencek:
Yes, this looks like the job of `MusicBrainz::Server::WebService::Format`. I'm actually not sure why it doesn't handle this with the existing code.

Zastai:
Looks like Format.pm happily instantiates the JSON serializer, but the search requests only pass args to xml_search, so the serializer in the stash is ignored.
I could just add a third arg to xml_search and pass $c->stash->{serializer}->fmt to that, or adjust the current patch to set $args->{fmt} from $c->stash->{serializer}->fmt (to avoid duplicating the accept header check).

mwiencek:
Okay, I see. I think the best solution is to modify the args validator [here](https://github.com/metabrainz/musicbrainz-server/blob/cde0df93294647929c713687922ce70b55df3d01/lib/MusicBrainz/Server/WebService/Validator.pm#L330-L330), since that's what's responsible for setting up `$c->stash->{args}`. We can simply always force the fmt there.

```Perl
$c->stash->{args}{fmt} = $c->stash->{serializer}->fmt;
```

It would be great to have a small test that only sets the Accept header.

Zastai:
Code changed to do it in the validator (although I only set the fmt arg if it wasn't set yet, just in case that ever matters).
Will look at adding a test.

Zastai:
Added a test for all Accept header processing. Validating the "bad format specified" error message did require making it deterministic (by sorting hash keys).

t/lib/t/MusicBrainz/Server/Controller/WS/2/DefaultFormat.pm
```diff
@@ -0,0 +1,91 @@
...
+use JSON qw( decode_json encode_json );
+use Test::Deep qw( cmp_deeply );
+use Test::JSON import => [qw( is_valid_json is_json )];
+use Test::More;
+use Test::Routine;
+use Test::XML::SemanticCompare;
+
+with 't::Mechanize', 't::Context';
+
+test 'MBS-9258: deduction of fmt from Accept header' => sub {
```
mwiencek: These tests look useful, but I don't think they test MBS-9258 at all, which was an issue with search requests, not lookups...right?

But I also don't think we have any tests related to search requests right now, and those would be more difficult to add, since you'd need to mock the parts that return data from the search server.

I'm okay with merging this if those aren't easy to add, but we should remove "MBS-9258" from this line if it doesn't test it.

mwiencek: Actually, there's at least one indexed search test in t/lib/t/MusicBrainz/Server/Controller/Search/Indexed.pm

t/lib/t/MusicBrainz/Server/Controller/WS/2/DefaultFormat.pm
```diff
@@ -0,0 +1,91 @@
+package t::MusicBrainz::Server::Controller::WS::2::DefaultFormat;
```
mwiencek: Pretty sure most of the test packages map directly to the package name they're testing, so `t/lib/t/MusicBrainz/Server/WebService/Format.pm` seems like a better location.

t/lib/t/MusicBrainz/Server/Controller/WS/2/DefaultFormat.pm
Zastai: You're right - it tests the code that was changed for MBS-9258, but not the MBS-9258 use case.
I'll look at adding JSON cases (with and without explicit fmt) to that indexed search test.

t/lib/t/MusicBrainz/Server/Controller/WS/2/DefaultFormat.pm
Zastai: I was unaware of that 1-to-1 mapping; I'll move it there and remove the MBS-9528 label.

t/lib/t/MusicBrainz/Server/Controller/WS/2/DefaultFormat.pm
Zastai: I'm adding a SearchArtists.pm analogous to BrowseArtists.pm.
However, I think I do see why this is currently not tested (and perhaps might not be testable before mbssss is fully in place): while the other tests can and do use a specifically set up test database, any search request returns data directly from the search server index. So for tests to be reproducible, that would need to be mocked too, using indexes build from the test db.
Will see how it plays out on Jenkins; I may need to comment out the actual tests for later reinstatement.

Zastai:
Some test cases added for artist search. Not specific to MBS-9258, just tests for both XML and JSON results (the latter of which will happen to fail if MBS-9258 is not fixed).

Will see what Jenkins makes of them; they may need to be commented out, since it's basically testing data from the search server's indexes rather than the static test database contents. Not sure if that would apply to a mbssss backend too.

Depending on outcomes, will squash some of these commits and force push for a cleaner merge.

mwiencek:
Yeah, I'm not sure I like the idea of adding tests against the live search server. :( The data in our tests is bound to get out of sync with the data in the production indexes.

These will be useful when we have a test container with the Solr stuff running, so we can point the tests to that. But I'd prefer to merge those tests until then (the direct lookup ones are fine).

Zastai:
(and now also cleaned up sneaky trailing whitespace)

mwiencek:
> If you would prefer removing the file entirely, that's fine too - I'll keep it around on a local branch.

mwiencek:
> If you would prefer removing the file entirely, that's fine too - I'll keep it around on a local branch.

I would prefer this. I'm still unsure about search tests overall—e.g. testing the output itself might better belong in the search server's own tests, whereas in MBS it'd be useful to only test the parts where the Perl interacts with the search server, like what's involved in MBS-9258.

Zastai:
OK, dropped the file.

Zastai:
Once it's possible to avoid hitting a "real" search server, I suppose a test could/should be added for this by running some search guaranteed not to return any matches (e.g. by looking for an impossible MBID?), and simply verifying that the response is valid xml/json. But it's hardly the most important thing to check.

t/lib/t/MusicBrainz/Server/WebService/Format.pm
```diff
@@ -0,0 +1,136 @@
+package t::MusicBrainz::Server::WebService::Format;
+
+use JSON qw( decode_json encode_json );
+use Test::JSON import => [qw( is_valid_json is_json )];
+use Test::More;
+use Test::Routine;
+use Test::XML;
```
mwiencek: Hm, `Test::XML` isn't in the cpanfile, so the tests are failing. We do have `Test::XML::SemanticCompare`. Is this module needed?

mwiencek:
> Once it's possible to avoid hitting a "real" search server, I suppose a test could/should be added for this by running some search guaranteed not to return any matches (e.g. by looking for an impossible MBID?), and simply verifying that the response is valid xml/json. But it's hardly the most important thing to check.

Yeah, certainly we'll have to check *some* output here to see if the Perl is doing its job. But hopefully only minimal responses that serve that purpose.

Zastai:
Hmm, had missed that. It's certainly not _required_; I'll try and see why I used one over the other (I _think_ the one I use reports differences is a more useful way, but I'd have to check to be sure). It should not be an issue to use `Test::XML::SemanticCompare` instead.

Zastai:
Switch the test to `Test::XML::SemanticCompare`.
Looks like the main reason I used Test::XML was that it provides a separate well-formed test, making the test sequences more alike between XML and JSON.
Hopefully with this change everything passes.

mwiencek:
LGTM. Thanks!

PR #476, opened 2017-02-24
MBS-9261: Handle URL search requests.

Zastai:
This adds the same search interface the other entities have.


mwiencek:
This seems fine to expose. It looks like mbsssss already supports the same interface, so it should remain stable.

But, can you update the docs too? :)

https://wiki.musicbrainz.org/Development/XML_Web_Service/Version_2
https://wiki.musicbrainz.org/Development/XML_Web_Service/Version_2/Search

Zastai:
I'll look at the docs this weekend, also ensuring the json examples are up to date. Need to do that anyway in order to make a ticket for the inconsistent json data models between search and browse requests :)

PR #477, opened 2017-02-27
Fix running tests in browser

y-van-z:
Fix two small bugs while running tests in browser, as reported in https://github.com/metabrainz/musicbrainz-server/pull/449#issuecomment-277520087:

1. `too much recursion` error caused by tests nesting, resolved by splitting URL Cleanup test (currently composed of subtests) into a myriad of tests
2. Label guess case test failure using Turkish mode, resolved by initializing user settings as necessary

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -9,7 +9,6 @@ const _ = require('lodash');
 
 const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/URLCleanup');
 
-test('URL cleanup component: auto-select, clean-up, and validation', {}, function (t) {
     const test_data = [
         // 45cat
         {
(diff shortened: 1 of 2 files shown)


mwiencek:
Ah, I see the guess case one is caused by changing the settings in the UI outside of the tests (because they share cookies, of course).

I guess Chrome and my version of Firefox both have higher recursion limits, though. :) Does it say in the console where in the code it's looping? Because if these simple/singly-nested tests are blowing up, I'd like to know if we're using the tape library incorrectly somehow.

y-van-z:
Indeed, tests fail with Firefox/Archlinux whereas they pass with any Blink-based browser.

It is not _looping_. It can either stop on _too much recursion_ error or _Stack overflow_ error.
_Where in the code_ depends on compiled scripts and browser usage.

IMHO, it is still an issue of JavaScript implementation of tape for **nested** tests. I cannot find the link again, but I recall a similar issue for a large number of (not-nested) tests which has been fixed a long time ago.

mwiencek:
Sorry, looping was a poor word choice. I'm able to reproduce something with `node --stack-size=500 root/static/scripts/tests/node-runner.js` but that seems like a very low limit for a web browser and I can't make sense of it right away.

PR #478, opened 2017-02-27
MBS-9263: Use HTTPS for WikiDoc transclusion table

legoktm:
The wiki is now HTTPS-only, so don't use HTTP.


mwiencek:
@brainzbot, add to whitelist

PR #479, opened 2017-03-01
MBS-8905, MBS-8927, MBS-9120: Allow for more auto-edits

y-van-z:
Allow for auto-edits in the three following cases:
- Editing release in a 1 hour delay, see [MBS-8905](https://tickets.metabrainz.org/browse/MBS-8905)
- Lowering release data quality in a 1 hour delay, see [MBS-8927](https://tickets.metabrainz.org/browse/MBS-8927)
- Adding time and setlist to an event that doesn’t have one, see [MBS-9120](https://tickets.metabrainz.org/browse/MBS-9120)

Although there is no test for amending release for now, this change doesn’t conceal other tests.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Release/ChangeQuality.pm
+++ b/lib/MusicBrainz/Server/Edit/Release/ChangeQuality.pm
@@ -3,13 +3,14 @@ use Moose;
 use Method::Signatures::Simple;
 use MooseX::Types::Moose qw( Int Str );
 use MooseX::Types::Structured qw( Dict );
-use MusicBrainz::Server::Constants qw( $EDIT_RELEASE_CHANGE_QUALITY );
+use MusicBrainz::Server::Constants qw( $EDIT_RELEASE_CHANGE_QUALITY $QUALITY_HIGH );
 use MusicBrainz::Server::Edit::Exceptions;
 use MusicBrainz::Server::Translation qw( N_l );
 
 extends 'MusicBrainz::Server::Edit';
 with 'MusicBrainz::Server::Edit::Release::RelatedEntities';
 with 'MusicBrainz::Server::Edit::Release';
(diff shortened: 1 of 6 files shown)


y-van-z:
Note: Technically `AllowAmendingRelease` should not be a `Role` anymore, but I want to add parameters (that specify edit table name and create edit type) to it for [MBS-8134](https://tickets.metabrainz.org/browse/MBS-8134).

lib/MusicBrainz/Server/Edit/Release/ChangeQuality.pm
```diff
@@ -88,5 +89,11 @@ method accept
     );
 }
 
+sub allow_auto_edit
+{
+    my $self = shift;
+    return $self->can_amend && $self->data->{new}{quality} != $QUALITY_HIGH;
```
mwiencek: Reversing these will save a query against the edit table if the quality is high.

mwiencek:
I'm surprised that the calling order is undefined for multiple `around` modifiers. [For superclasses it seems to be defined](http://search.cpan.org/~ether/Moose-2.2004/lib/Moose/Manual/MethodModifiers.pod#Execution_order_of_method_modifiers_and_inheritance), and I assumed for roles it depended on where the `with` statement is inserted, but guess not.

But your solution is much more straightforward, so that's good.

y-van-z:
You are most probably correct about this, although the documentation is incomplete.
Anyway, it would not be very handy and we need it as a method to change release quality.

lib/MusicBrainz/Server/Edit/Release/ChangeQuality.pm
y-van-z: Done in https://github.com/metabrainz/musicbrainz-server/pull/479/commits/3b899da260a9de1e83a6f9257f4008615f813cb3.

PR #480, opened 2017-03-02
MBS-9260: Make changing ellipsis version an auto-edit

y-van-z:
[MBS-9260](https://tickets.metabrainz.org/browse/MBS-9260): Make changing between version (visually similar to three dots) of an ellipsis an auto-edit.

### DIFF
--- a/lib/MusicBrainz/Server/Validation.pm
+++ b/lib/MusicBrainz/Server/Validation.pm
@@ -355,6 +355,11 @@ sub normalise_strings
         # 05BE Hebrew maqaf, 2010 hyphen, 2012 figure dash, 2013 en-dash, 2212 minus
         $t =~ tr/\x{05BE}\x{2010}\x{2012}\x{2013}\x{2212}/-/;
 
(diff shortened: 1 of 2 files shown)


y-van-z:
Two latest commits are _to be squashed_ depending on the outcome of ongoing discussion about this ticket.

mwiencek:
If my opinion matters at all, I agree with the current proposition ("only horizontal three-dots ellipses"). :)

I'm a bit confused, since I couldn't figure out where else this is implemented, from the ticket description: "Currently, changing 3 full stops "..." to a horizontal ellipsis is an autoedit, but not changing between any others." But I didn't dig very deeply.

y-van-z:
It happens to be done by [Text::Unaccent](http://search.cpan.org/~pjacklam/Text-Unaccent-PurePerl-0.05/lib/Text/Unaccent/PurePerl.pm#Other_conversions).

Thanks for your input, I figure that this change is now consensual enough to get merged.

PR #481, opened 2017-03-02
MBS-8748: Wrap medium name in release tracklist

y-van-z:
[MBS-8748](https://tickets.metabrainz.org/browse/MBS-8748): Long medium names caused horizontal scrolling on release pages.

This patch wraps medium name and visually split the header line into (borderless) cells so that medium format and medium number are clearly distinguished from wrapped medium name.

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -828,15 +828,14 @@ END -%]
 [%- END -%]
 
 [%~ MACRO medium_description(medium) BLOCK;
-    args = {
+    format_position = l('{medium_format} {position}', {
         medium_format => medium_format_name(medium),
         position => medium.position,
-    };
+    });
     IF medium.name != '';
-        args.medium_name = isolate_text(medium.name);
-        l('{medium_format} {position}: {medium_name}', args);
+        '<span>' _ add_colon(format_position) _ '</span> <span class="medium-name">' _ isolate_text(medium.name) _ '</span>';
     ELSE;
-        l('{medium_format} {position}', args);
+        '<span>' _ format_position _ '</span>';
     END;
 END ~%]
 
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -590,6 +590,18 @@ table.tbl {
 
         a {
             color: @text-black;
+
+            > * {
+                display: table-cell;
+            }
+
+            > :not(:last-child) {
+                padding-right: 0.3em;
+            }
+
+            .medium-name {
+                white-space: normal;
+            }
         }
 
         &:first-child {


root/components/common-macros.tt
```diff
@@ -834,9 +834,9 @@ END -%]
     };
     IF medium.name != '';
         args.medium_name = isolate_text(medium.name);
-        l('{medium_format} {position}: {medium_name}', args);
+        l('<span>{medium_format} {position}:</span> {medium_name}', args);
```
mwiencek: I think it'd be good to add something like `class="medium-format-position"` here and refer to that in the CSS, so we don't have to worry about ambiguity later. Also easier to grep where a style is coming from that way.

y-van-z: There is no specific rule for this element in the style sheet as `> span` rules include `span.expand-triangle` also. Instead, I could wrap `{medium_name}` into a `span` with `medium-name` class to refer to in the style sheet in place of not-semantic `bdi`. Or do both?

mwiencek: Right, though I think it's good to avoid CSS selectors that depend too much on the structure of the HTML. Unless the style would break if it wasn't a direct child of `<a>`.

About ambiguity, I mean if we change the surrounding HTML later and the generic `<span>` no longer works, we'll have to change the string again (unless it has a class name we could use).

Feel free to change it to whatever you think is best (or we can merge it as-is).

y-van-z: I cleaned CSS rules and reduced changes made to localized messages in updated  https://github.com/metabrainz/musicbrainz-server/pull/481/commits/83e0fc444446f75bd816d9d6a2d251c3663cc5aa
I added `medium-name` class only which is useful to apply _normal_ _white-space_ wrapping style .
Other classes can be added later on if necessary without changing localized messages.

root/components/common-macros.tt
```diff
@@ -828,15 +828,17 @@ END -%]
...
+    format_position = l('{medium_format} {position}', {
         medium_format => medium_format_name(medium),
         position => medium.position,
-    };
+    });
     IF medium.name != '';
-        args.medium_name = isolate_text(medium.name);
-        l('{medium_format} {position}: {medium_name}', args);
+        l('{medium_format_position} {medium_name}', {
+            medium_format_position => '<span>' _ add_colon(format_position) _ '</span>',
```
mwiencek: I like how the HTML was moved out of the string. 👍

I guess this'll allow translators to change the order of `{medium_format_position}` and `{medium_name}`, and maybe add additional words for context. Does that make sense if these are supposed to be in separate cells?

There seems to be a contradiction between having these cells visually separated from each other, but allowing translators to edit the string as if it's all in one cell. So maybe we shouldn't have this "combined" string anymore?

y-van-z: It makes sense for right-to-left locales such as Arabic, although Arabic translation is currently 0% complete and most of UI (except medium header) is not ready for right-to-left locales yet. 😆 

But I agree, left-to-right ordering of cells would have to be done by template, not by message.

Updated in https://github.com/metabrainz/musicbrainz-server/pull/481/commits/bc878ed9bd7255bec05dee633c88aa11a3a13698

PR #484, opened 2017-03-03
MBS-9268: Hide empty columns in relationships table

y-van-z:
Implement [MBS-9268](https://tickets.metabrainz.org/browse/MBS-9268), branched from PR #466.

### DIFF
--- a/root/components/relationships-table.tt
+++ b/root/components/relationships-table.tt
@@ -18,58 +18,65 @@
             <tr>
                 <th>[% l('Date') %]</th>
                 <th>[% l('Title') %]</th>
-                [% IF has.credit -%]<th>[% l('Credited As') %]</th>[%- END %]
-                [% IF has.attribute -%]<th>[% l('Attributes') %]</th>[%- END %]
-                [% IF has.artist -%]<th>[% l('Artist') %]</th>[%- END %]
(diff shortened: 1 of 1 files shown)


root/components/relationships-table.tt
```diff
@@ -1,50 +1,82 @@
...
+            <tr class="[% loop.parity %]">
+                <td>[% rel.link.formatted_date %]</td>
+                <td>
+                    [%~ '<span class="mp mp-rel">' IF rel.edits_pending ~%]
+                    [% link_entity(rel.target, 'show', '', rel.target_credit) %]
+                    [%~ '</span>' IF rel.edits_pending ~%]
+                </td>
+                [%- IF has.credit %]
+                <td>
+                    [%~ rel.source_credit IF rel.source_credit ~%]
```
mwiencek: Perhaps this should be fixed to use the same check as above, `rel.source_credit != ''`. Seems like it'll break with "0", as usual.

y-van-z: Silent bug fix added in https://github.com/metabrainz/musicbrainz-server/pull/484/commits/204956e1207907a4f81a89c29560658520589a73

mwiencek:
:shipit:

Thanks @y-van-z & @loujine.

PR #487, opened 2017-03-09
Hook Perl::Critic into the build

mwiencek:
I made the .perlcriticrc file (which nobody really uses) useful by turning it into a whitelist of non-broken policies, and checking those in the tests. I fixed a few additional policies which were easy to fix. Stuff like `CodeLayout::RequireTrailingCommas` probably isn't worth ever fixing, based on the number of errors I saw.

I tried using `Test::Perl::Critic` at first, but that took over 10 minutes to test everything for some reason, so we are definitely not adding that to the build. I wrote a shell script which invokes the `perlcritic` command and converts its output to TAP format instead, and that finishes in under a minute for me.

There's a list of policies at http://search.cpan.org/~thaljef/Perl-Critic-1.126/lib/Perl/Critic/PolicySummary.pod but we should obviously only enable the ones we like.

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -6,5 +6,7 @@ severity=1
 
 [ControlStructures::ProhibitUnreachableCode]
 
+[Subroutines::ProhibitReturnSort]
+
 [TestingAndDebugging::ProhibitNoStrict]
 allow=refs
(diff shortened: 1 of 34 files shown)


y-van-z:
Works for me. Maybe `HACKING.md` could be updated too. I looked at other policies but got hungry after reading [RequireEndWithOne](http://search.cpan.org/~thaljef/Perl-Critic-1.126/lib/Perl/Critic/Policy/Modules/RequireEndWithOne.pm#DESCRIPTION).

PR #488, opened 2017-03-10
MBS-8465, MBS-9288: Improve changing other relationship credits

yvanzo:
Previously, relationship editor had an option to change artist credits for all relationships on the page at the same time. 

Improvement [MBS-8465](https://tickets.metabrainz.org/browse/MBS-8465) involves allowing editor to change a subset only of these relationships, filtered either by type or by entity types.

To that end, this patch adds three radio buttons to the current checkbox input which is now labelled “Change other” rather than “Change all”.  Radio buttons are visible if and only if input is checked.  Label lines are wrapped if translation is too long.

Additionally, it fixes bug [MBS-9288](https://tickets.metabrainz.org/browse/MBS-9288) which prevented to change other relationship credits for source Artist from an Artist-Artist relationship.

### DIFF
--- a/root/components/relationship-editor.tt
+++ b/root/components/relationship-editor.tt
@@ -32,7 +32,7 @@
     <td class="section" data-bind="text: MB.i18n.addColon(MB.i18n.strings.entityName[source.entityType])"></td>
     <td>
       <span data-bind="html: source.html({ 'target': '_blank', creditedAs: relationship().creditField(source)() })"></span>
-      <!-- ko template: {name: "template.dialog-entity-credit", data: {entity: source, relationship: relationship()}} --><!-- /ko -->
+      <!-- ko template: {name: "template.dialog-entity-credit", data: {entity: source, relationship: relationship(), role: "source"}} --><!-- /ko -->
     </td>
   </tr>
 </script>
@@ -47,21 +47,21 @@
     </div>
     <div>
       <label>
-        <input type="checkbox" data-bind="checked: $parent.changeOtherRelationshipCredits" />
+        <input type="checkbox" data-bind="checked: $parent.changeOtherRelationshipCredits[role]" />
         <span data-bind="html: MB.i18n.l('Change credits for other {entity} relationships on the page.', {entity: entity.html()})">
         </span>
       </label>
     </div>
-    <div data-bind="visible: $parent.changeOtherRelationshipCredits">
+    <div data-bind="visible: $parent.changeOtherRelationshipCredits[role]">
       <div>
         <label>
-          <input type="radio" value="all" data-bind="checked: $parent.selectedRelationshipCredits" />
+          <input type="radio" value="all" data-bind="checked: $parent.selectedRelationshipCredits[role]" />
           [% l('All of these relationships.'); %]
         </label>
       </div>
       <div>
         <label>
-          <input type="radio" value="same-entity-types" data-bind="checked: $parent.selectedRelationshipCredits" />
+          <input type="radio" value="same-entity-types" data-bind="checked: $parent.selectedRelationshipCredits[role]" />
           <span data-bind="html: MB.i18n.l('Only relationships to {entity_type} entities.',
                                            {entity_type: MB.i18n.strings.entityName[relationship.target(entity).entityType].toLocaleLowerCase()})">
           </span>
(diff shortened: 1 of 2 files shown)


root/components/relationship-editor.tt
```diff
@@ -47,11 +47,31 @@
     </div>
     <div>
       <label>
-        <input type="checkbox" data-bind="checked: $parent.changeAllRelationshipCredits" />
-        <span data-bind="html: MB.i18n.l('Change all relationship credits for {entity} on the page.', {entity: entity.html({target: '_blank'})})">
+        <input type="checkbox" data-bind="checked: $parent.changeOtherRelationshipCredits" />
+        <span data-bind="html: MB.i18n.l('Change credits to {entity} for other relationships on the page.', {entity: entity.html({target: '_blank'})})">
```
mwiencek: The `entity` argument you're passing in doesn't have the new credit, so it looks like "Change credits to [the same name]."

I think there's a way to fix this which also makes the string easier to understand:

```JavaScript
MB.i18n.l('Change credits to “{credit}” for other {entity} relationships on the page.', {credit: relationship().creditField(source)(), entity: entity.html({target: '_blank'})})
```

Hopefully changing "other relationships" to "other {entity} relationships" makes it clear that even the options below, like "all relationships," are still restricted to relationships to that entity, not literally all relationships on the page.

root/components/relationship-editor.tt
```diff
@@ -47,11 +47,31 @@
...
+      <div>
+        <label>
+          <input type="radio" value="same-entity-types" data-bind="checked: $parent.selectedRelationshipCredits" />
+          [% l('Only relationships with the same entity types.'); %]
+        </label>
+      </div>
+      <div>
+        <label>
+          <input type="radio" value="same-relationship-type" data-bind="checked: $parent.selectedRelationshipCredits" />
+          [% l('Only relationships of the same type.'); %]
```
mwiencek: It might benefit users to make these messages a bit more specific too, like "Only relationships to {entity_type} entities" and "Only relationships of type {relationship_type}" (especially since the user can change the relationship type, so it's not clear which one it uses).

yvanzo:
Strings have been clarified, see [screenshot](https://tickets.metabrainz.org/secure/attachment/11584/screenshot-mbs-8465.png).

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -326,6 +335,11 @@ const PART_OF_SERIES_LINK_TYPE_GIDS = _.values(PART_OF_SERIES_LINK_TYPES);
             this.showLinkTypeHelp(!this.showLinkTypeHelp.peek());
         },
 
+        linkTypeName: function () {
```
mwiencek: In your screenshot, the link type name shown in the string ("composed") is different from the one in the `<select>` ("composer"), which reads oddly. I guess it's something to do with the `backwards` argument to `linkPhrase.clean`. (Backwards = the target is entity0.) I'd see how the select options handles it.

root/components/relationship-editor.tt
```diff
@@ -47,11 +47,36 @@
...
+                                           {entity_type: MB.i18n.strings.entityName[relationship.target(entity).entityType]})">
+          </span>
+        </label>
+      </div>
+      <div>
+        <label>
+          <input type="radio" value="same-relationship-type" data-bind="checked: $parent.selectedRelationshipCredits[role]" />
+          <span data-bind="html: MB.i18n.l('Only “{relationship_type}” relationships to {entity_type} entities.',
+                                           {relationship_type: $parent.linkTypeName(),
+                                            entity_type: MB.i18n.strings.entityName[relationship.target(entity).entityType]})">
```
mwiencek: Ideally the entity type wouldn't be capitalized in the resulting string, but it's hard to fix:

1. I don't think we want to make people translate lowercase versions of every string.
2. `String.prototype.toLocaleLowerCase()` uses the locale of the browser, not of MusicBrainz. Maybe that's better than nothing.

I guess this is something we can figure out later, after this is merged.

root/static/scripts/relationship-editor/common/dialog.js
yvanzo: Used `backwards` argument to select the same link phrase in https://github.com/metabrainz/musicbrainz-server/pull/488/commits/8c853861912e9eed2f589c70ff157b63d7993219.

root/components/relationship-editor.tt
yvanzo: Called `toLocaleLowerCase()` for entity type in https://github.com/metabrainz/musicbrainz-server/pull/488/commits/8c853861912e9eed2f589c70ff157b63d7993219#diff-dbe28ef3e31e909f5576de5c67298782R339.

PR #489, opened 2017-03-15
Stringify blessed /ws/js errors for encode_json

mwiencek:
Stops it from throwing another error while it's trying to return the original error.

https://sentry.metabrainz.org/metabrainz/musicbrainz-server/issues/441/

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js.pm
@@ -12,6 +12,7 @@ use MusicBrainz::Server::Data::Search qw( escape_query alias_query );
 use MusicBrainz::Server::Constants qw( entities_with );
 use MusicBrainz::Server::Validation qw( is_guid );
 use Readonly;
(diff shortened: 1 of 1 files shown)


mwiencek:
I don't think there's any point in enabling `convert_blessed`, since the exceptions probably don't have `TO_JSON` methods, so we'd just output `undef`. They usually overload the quote operation to format a message, though.

lib/MusicBrainz/Server/Controller/WS/js.pm
```diff
@@ -297,7 +298,9 @@ sub detach_with_error : Private {
     my ($self, $c, $error, $status) = @_;
 
     $c->res->content_type('application/json; charset=utf-8');
-    $c->res->body(encode_json({ error => $error }));
+    $c->res->body(encode_json({
+        error => (blessed($error) ? "$error" : $error),
```
mwiencek: For some errors we output a hash ref, which is why we don't always quote it.

PR #490, opened 2017-03-16
MBS-8578: Handle NoChanges edits better in the release editor

mwiencek:
Prior to this patch, if an edit threw a NoChanges exception it'd cause the request to 400, and cause it to stop processing edits (a submission could contain multiple edits). This was a problem, because the release editor batches recording and medium edits together in one request; at the same time, the JavaScript ignored 400 errors caused by NoChanges exceptions (which were indicated by a specific code in the response), and continued submitting other edits. The result was usually a stub release with no mediums.

The fix is simply that NoChanges exceptions will no longer cause a request to /ws/js/edit to fail (though they are still indicated in the response).

### DIFF
--- a/root/static/scripts/release-editor/edits.js
+++ b/root/static/scripts/release-editor/edits.js
@@ -10,7 +10,7 @@ const debounce = require('../common/utility/debounce');
 const isPositiveInteger = require('../edit/utility/isPositiveInteger');
 const validation = require('../edit/validation');
 
-const ERROR_NO_CHANGES = 3;
+const WS_EDIT_RESPONSE_OK = 1;
 
 (function (releaseEditor) {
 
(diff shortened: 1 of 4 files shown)


root/static/scripts/release-editor/edits.js
```diff
@@ -496,20 +496,20 @@ const ERROR_NO_CHANGES = 3;
 
             let submissionDone = function (data) {
                 if (data && current.callback) {
-                    current.callback(release, data.edits);
+                    current.callback(
+                        release,
+                        data.edits.filter(
+                            x => x.response === WS_EDIT_RESPONSE_OK
```
mwiencek: The callbacks in `releaseEditor.orderedEditSubmissions` are only written to handle edits that actually went through. But, this technically isn't needed, because "create" edits (which are mostly what we need to handle in those callbacks) can't trigger NoChanges exceptions.

PR #495, opened 2017-03-28
MBS-9273, MBS-9274: Carry over intermediary schema changes

mwiencek:


upgrade.json
```diff
@@ -48,7 +48,8 @@
   },
 
   "24": {
-    "slave": [],
+    "slave": ["20160602-mbs-8963.sql",
```
mwiencek: Scripts under `slave` are run on all nodes, not just slaves. (For that, you'd want `slave_only`.)

These scripts apply only to edit tables which aren't replicated, but those tables, functions, and indexes obviously still exist and may even contain old data on slave instances.

PR #497, opened 2017-03-28
MBS-390/MBS-9088: Fix regression/bug in edit search

yvanzo:
First, it fixes a small bug in _editor_ fields of edit search form ([MBS-9088](https://tickets.metabrainz.org/browse/MBS-9088)) due to incomplete ID validation.

Then it addresses the following NGS regressions of edit search features ([MBS-390](https://tickets.metabrainz.org/browse/MBS-390)):

1. [MBS-3362](https://tickets.metabrainz.org/browse/MBS-3362): Add “is me”/“is not me” operator to _editor_ filters
2. [MBS-3914](https://tickets.metabrainz.org/browse/MBS-3914): Add “is in my subscriptions” operator to _editor_ filters
3. [MBS-3665](https://tickets.metabrainz.org/browse/MBS-3665): Add a link to wiki documentation page for format in _date_ fields
4. [MBS-5681](https://tickets.metabrainz.org/browse/MBS-5681): Add “Abstain”/“Approve” votes to the _vote count_ filter
5. [MBS-2673](https://tickets.metabrainz.org/browse/MBS-2673): Extend filter for current user _vote_ filter it for any _voter_

Besides, there is a bit of code refactoring/cleaning of edit search packages/hideous hidden fields.


lib/MusicBrainz/Server/EditSearch/Predicate/EditNoteAuthor.pm
```diff
@@ -1,47 +1,24 @@
...
 
-sub combine_with_query {
-    my ($self, $query) = @_;
-
-    my $sql = 'EXISTS (
+with 'MusicBrainz::Server::EditSearch::Predicate::Role::User' => {
+    template_clause => 'EXISTS (
         SELECT TRUE FROM edit_note
-         WHERE edit_note.editor = ?
+         WHERE ROLE_CLAUSE(edit_note.editor)
```
mwiencek: This `ROLE_CLAUSE` stuff made my head spin a bit, but I think I understand it now. Maybe it's a bit backwards though? I wonder if we should instead be passing a subroutine here which accepts whatever arguments the role wants to give it, and constructing/returning the SQL based on those arguments. I could be missing something.

lib/MusicBrainz/Server/EditSearch/Predicate/VoteCount.pm
```diff
@@ -0,0 +1,56 @@
...
+use MusicBrainz::Server::Constants qw( :vote );
+use MusicBrainz::Server::Types qw( VoteOption );
+use MusicBrainz::Server::Validation qw( is_integer );
+
+extends 'MusicBrainz::Server::EditSearch::Predicate::ID';
+
+around operator_cardinality_map => sub {
+    my ($orig, $self) = splice(@_, 0, 2);
+    my %orig_map = $self->$orig;
+    return map { $_ => 1 + $orig_map{$_} } keys %orig_map;
```
mwiencek: Is that because the vote type is now `args.0`? Can't we just make that `args.2` and leave the rest alone? Shouldn't matter where it appears in the actual HTML. (Maybe I'm completely misunderstanding why this is needed.)

yvanzo: It is because there are three arguments for vote count instead of two for other ID predicates.

lib/MusicBrainz/Server/EditSearch/Predicate/EditNoteAuthor.pm
yvanzo: I tried to create a subroutine at first, finally found this approach to be cleaner, but it definitely **has to** be comprehensible. Maybe a bit of renaming would make things more clear:

- `template_clause` could be renamed `context_clause`,
- `ROLE_CLAUSE` could be renamed `ROLE_OPERATOR` or `ROLE_TEST`

Or any other more convincing names… 🙄

root/edit/search_macros.tt
```diff
@@ -253,6 +253,35 @@
...
 
+[% MACRO predicate_vote_count(field, field_contents) WRAPPER wfield predicate="vote_count" %]
+  <select name="args.0" class="arg">
+    [% FOR item=[ [  2,   lp('Approve', 'vote') ]
+                , [  1,   lp('Yes', 'vote') ]
+                , [  0,   lp('No', 'vote') ]
+                , [ -1,   lp('Abstain', 'vote') ]
+                ] %]
+    <option value="[% html_escape(item.0) %]"
+            [% 'selected="selected"' IF field_contents.find_argument(item.0).defined %]
```
yvanzo: Cleaned all predicates in https://github.com/metabrainz/musicbrainz-server/pull/497/commits/ec4553cd4de9ef1e7900f0128c1de67ad16c1b2b.

lib/MusicBrainz/Server/EditSearch/Predicate/EditNoteAuthor.pm
mwiencek: Hmm. Those renames don't make things clearer for me, so I'm fine with how it is. But I might have more questions.

root/edit/search_macros.tt
```diff
@@ -253,6 +253,35 @@
...
+            >[% html_escape(item.1) %]</option>
+    [% END %]
+  </select>
+
+  [% operators([ [ '=', l('is') ],
```
yvanzo: Moving vote type to a named parameter allowed to have the exact same HTML code and to deduplicate it in https://github.com/metabrainz/musicbrainz-server/pull/497/commits/3afe77594eb5359c98feb72bc8e329f66cf3a095.

lib/MusicBrainz/Server/EditSearch/Predicate/VoteCount.pm
yvanzo: What matters though is the order arguments are filed into the SQL query. Instead of making that `args.2`, I moved `args.0` to `vote` and dropped this method modifier in https://github.com/metabrainz/musicbrainz-server/pull/497/commits/3afe77594eb5359c98feb72bc8e329f66cf3a095.

yvanzo:
Maybe https://github.com/metabrainz/musicbrainz-server/pull/497/commits/3afe77594eb5359c98feb72bc8e329f66cf3a095 fixed your issue with multiple “Vote tally” conditions. At least, I cannot reproduce it, but I have not been able to reproduce your issue with multiple “Voter” conditions either…

mwiencek:
> Maybe 3afe775 fixed your issue with multiple “Vote tally” conditions. At least, I cannot reproduce it, but I have not been able to reproduce your issue with multiple “Voter” conditions either…

Yeah, that fixes the "Vote tally" issue. 👍 I'm still seeing it for "Voter":

untitled

yvanzo:
Oh, I see, I can now reproduce your issue. 👍  Actually, I always tested `=` operator before `me` operator, but then `voter_id` was always provided. The bug is that `voter_id` should not be required for operators `me`, `not_me` and `subscribed`.

yvanzo:
Issue fixed in https://github.com/metabrainz/musicbrainz-server/pull/497/commits/71845c4f6da18614a01c1ed8ba0b8080250cd039. It should be alright now. I made a forced update to wrap previous commit message. Build failures seem are random and unrelated to this change.

mwiencek:
Working better now. Some more issues I found:

* The vote tally has "is after" / "is before" options, but they should be "is more than" / "is less than" (I guess `predicate_int` should be used there instead).
* I tried doing a search for edits I've approved that had at least one No vote, using the following conditions: "Voter is me and voted approve," and "Voter is not me and voted No." That doesn't work, and just returns all edits I've approved, without any No votes. If I change the second condition to "Vote tally No is after 0" it works, but I expect the former should be equivalent, right?

yvanzo:
- Fixed vote tally operators wording in a series of fixes, some of which will have to be squashed
  See https://github.com/metabrainz/musicbrainz-server/pull/497/commits/0a5485c5ee662de9abedb0ba7de6aa7507d4e87c https://github.com/metabrainz/musicbrainz-server/pull/497/commits/621aaf992117b627ae562271ed07762b1647a241 https://github.com/metabrainz/musicbrainz-server/pull/497/commits/9a3627525e71f29878d932c38d4a305683c298e9
- Fixed voter negative operators which in your example computed “Voter is me and did not vote No”
  See https://github.com/metabrainz/musicbrainz-server/pull/497/commits/60add1336586a0ec795e140142cf4209594a3dbb

Additionally added voter “is in my subscriptions” operator in https://github.com/metabrainz/musicbrainz-server/pull/497/commits/05dbc3739ae7bf8f2129cfcb5ee942ed9a64aa3e

yvanzo:
The _Edit Note Author_ condition is debatable too. It currently behaves in the same way _Voter_ condition did: “Edit Note Author is not me” actually means “No edit note author is me” rather than “At least one edit note author is not me”. Current behaviour is probably the most useful one, but maybe both should be made available. In this case, there will probably be less code shared with _Editor_ condition in _User_ role.

mwiencek:
Sorry for the delay. I tested things again and I think it looks good now. ("Vote tally" timed out for most of my queries, but it does on the main server too.)

I guess we could add a note count field to "Edit Note Author," so you could do something like "Edit Note Author is me and left 0 notes" to get the "No edit note author is me" behavior. But there may not be much interest in that, and not in this PR anyway. :)

yvanzo:
Ok, I squashed fix commits as expected and pushed a forced update.

I will let you continue merging branches into master for now… 🌵

PR #501, opened 2017-04-02
MBS-8640: Make adding work attribute auto-edit

yvanzo:
[MBS-8640](https://tickets.metabrainz.org/browse/MBS-8640): Make adding work attribute auto-edit

Previously, adding work attribute was not an auto-edit, both because of a potential deliberate choice and a bug in implementation (`old` `attributes` array is always defined when adding work attributes).

This patch allows adding work attributes to be an auto-edit, only if there is no existing attribute of the same type for this work.

It doesn't allow modifying/deleting work attributes to be auto-edit.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Edit/Work/Edit.pm
+++ b/t/lib/t/MusicBrainz/Server/Edit/Work/Edit.pm
@@ -161,6 +161,153 @@ test 'Changing work language is not an auto-edit for non-auto-editors' => sub {
     accept_edit($c, $edit);
 };
 
+test 'Adding first work attributes is an auto-edit for non-auto-editors' => sub {
+    my $test = shift;
+    my $c = $test->c;
+
+    MusicBrainz::Server::Test->prepare_test_database($c, '+edit_work_attributes');
+
+    my $work = $c->model('Work')->get_by_id(1);
+    $c->model('WorkAttribute')->load_for_works($work);
+
+    my $edit = create_edit(
+        $c,
+        $work,
+        attributes => [
+            {
+                attribute_type_id => 1,
+                attribute_value_id => 10,
+                attribute_text => undef,
+            },
+            {
+                attribute_type_id => 2,
+                attribute_text => 'Attr value',
+                attribute_value_id => undef
+            }
+        ]
+    );
+
+    ok(!$edit->is_open);
+};
+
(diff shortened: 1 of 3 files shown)


yvanzo:
Weird thing happens: I tested this change successfully on my server but tests which fail for non-auto-edits.

mwiencek:
It fails locally for me, too.

yvanzo:
Sorry, I meant I cannot reproduce these failures through the UI.

Since then, I found that edit `data old attributes` is not set during tests, whereas `data old language_id` is. Mechanizing edit creation during the test might solve this issue. But is this expected?

mwiencek:
Sorry for the delay. :(

Not sure if it's really "expected," but I see the reason why. You need to call `$c->model('WorkAttribute')->load_for_works($work)` first on the work you pass in to `create_edit_with_attributes`.

They're [loaded in the controller](https://github.com/metabrainz/musicbrainz-server/blob/2d17cd4d7dddf107b5b35d64ab52bfa0a436d258/lib/MusicBrainz/Server/Controller/Work.pm#L114-L114) for the edit page, of course, which is why it works there.

They're also [loaded in current_instance](https://github.com/metabrainz/musicbrainz-server/blob/2d17cd4d7dddf107b5b35d64ab52bfa0a436d258/lib/MusicBrainz/Server/Edit/Work/Edit.pm#L192-L192), but that method is not used by the `Edit::WithDifferences` role. It just uses the object you pass in for `to_edit`. Perhaps that role should use `current_instance`, but I haven't looked close enough to say for sure. :)

t/lib/t/MusicBrainz/Server/Edit/Work/Edit.pm
```diff
@@ -252,18 +358,23 @@ test 'Check conflicts (conflicting edits)' => sub {
     is($work->name, 'A', 'work renamed');
 };
 
-sub create_edit {
+sub create_edit_with_attributes {
     my $c = shift;
     my $work = shift;
     return $c->model('Edit')->create(
```
mwiencek: Maybe having a default value is preferable to separate methods?

```Perl
my %opts = @_;
$opts{attributes} //= [];
```

yvanzo:
Updated this pull request with proposed changes and solved conflicts.

PR #502, opened 2017-04-03
Add a plugin to embed React components within TT

mwiencek:
Based on https://github.com/metabrainz/musicbrainz-server/pull/500 so it includes those changes for now.

Does what the title says, which I'm pretty sure is an idea @chirlu at least has suggested over a year ago.

It also does the following, broadly:

 * Makes the template renderer communicate with the Perl over a UNIX socket, which is vastly more performant than the current HTTP + Redis combo. Kind of important since we're going to be communicating with it possibly many times per page load, rather than just once.
 * Renders the `<head>` element on every page using React.
 * Fixes any other related things that affected the development of the above two items.


mwiencek:
I fixed some bugs related to the new UNIX socket communication, and made the external links/favicons in the sidebar also get rendered by React. Not because I want to stuff more things in this PR (I don't), but I wanted to reduce the amount of data needed to be sent over the socket for each request (which includes the Catalyst stash), and doing that removed `FAVICON_CLASSES` from the stash, which was a large chunk of it.

mwiencek:
Ok, I was hoping to merge this sooner, but let's try to release it with the schema change.

PR #503, opened 2017-04-04
MBS-8736 prerequisites

mwiencek:
I haven't finished MBS-8736 yet, but that branch contains other useful refactoring which I'm just going to get out of the way and submit now. The "Factor out role to traverse FKs" commit in particular would now also be useful for creating a script that can generate a list of recently updated entities in our main web service (rather than for JSON-LD).

There's also a test for admin/ExportAllTables now, which is good to have.

### DIFF
--- a/admin/ExportAllTables
+++ b/admin/ExportAllTables
@@ -394,7 +394,7 @@ my $fCallback;
 my @tablelist;
 
 GetOptions(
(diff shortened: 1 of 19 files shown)


mwiencek:
I need this for further work on the JSON dumps and there were no comments after a month, so merging.

PR #505, opened 2017-05-01
MBS-9323: Fix long URLs distorting page layout.

Sophist-UK:
Long URLs do not wrap and cause page distortion e.g. in https://musicbrainz.org/artist/f2b1690b-7b7a-4364-9764-098ecf87f794 a Wikipedia url with a very long #id caused the sidebar to take c. 80% of page width.

This proposed fix is applied to all `<a>` tags on every page. Further selectors can be added to e.g. restrict this to the sidebar if that is prefered. e.g. 
```
&#sidebar {
    word-break: break-all;
}
```
Resolves https://tickets.metabrainz.org/browse/MBS-9323

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -18,7 +18,7 @@ body {
 a {
     color:@link-default;
     text-decoration: none;
-    word-break: keep-all;
+    word-break: break-all;
 
     &:visited {
         color:@link-visited;
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -18,6 +18,7 @@ body {
 a {
     color:@link-default;
     text-decoration: none;
+    word-break: keep-all;
 
     &:visited {
         color:@link-visited;
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -19,10 +19,6 @@ a {
     color:@link-default;
     text-decoration: none;
     
-    &#sidebar {
-        word-break: break-all;
-    }
-
     &:visited {
         color:@link-visited;
     }
@@ -388,6 +384,10 @@ div.warning img.warning {
 
 #sidebar p {
     margin-top: 0;
+}
+
+#sidebar a {
+    word-break: break-all;
 }
 
 #sidebar ul.links {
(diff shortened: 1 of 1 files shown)


root/static/styles/layout.less
```diff
@@ -18,6 +18,7 @@ body {
 a {
     color:@link-default;
     text-decoration: none;
+    word-break: keep-all;
```
d4rkie: `word-break: break-all;` should be a better choice here, IMO. works in most browsers, including webkit-based ones

d4rkie:
`keep-all` will distort the UI in the opposite direction (making it overscroll instead). `break-all` will keep the characters in and not break the layout

Sophist-UK:
Agreed.

mwiencek:
Unfortunately this overrides our other `word-break` rules and causes every link to break mid-word (which is ugly) where they'd otherwise break between words. I think this rule should be specific to the sidebar links, which either don't have words or are already very short if they do.

Sophist-UK:
@mwiencek I will make it specific - as offered in the original description.

root/static/styles/layout.less
```diff
@@ -18,6 +18,10 @@ body {
 a {
     color:@link-default;
     text-decoration: none;
+    
+    &#sidebar {
```
mwiencek: That matches `a#sidebar`, i.e. a link with `id="sidebar"`. :) You need an `a` rule inside the `#sidebar` block instead.

Sophist-UK: Oh - yes - you are right. Apologies.

root/static/styles/layout.less
```diff
@@ -18,7 +18,7 @@ body {
 a {
     color:@link-default;
     text-decoration: none;
-
+
```
Freso: Random whitespace? :(

Sophist-UK: :-(

Sophist-UK:
@yvanzo: Just tested this with Firefox and you are indeed right.

root/static/styles/layout.less
```diff
@@ -386,6 +386,10 @@ div.warning img.warning {
     margin-top: 0;
 }
 
+#sidebar ul.external_links {
```
mwiencek: We have a block for `#sidebar ul.external_links` right below this on line 398, so it'd be good to put this rule inside the existing block.

PR #506, opened 2017-05-02
MBS-5452: Support multiple lyric language values for works

mwiencek:
Todo: ability to select >1 language in the work creation dialog in the release relationship editor (might have to wait).

Since users might expect the first language in the UI to be the primary language, and since the first language is now returned for the old single-language field in the web service (so, we'll want that value to remain consistent at the very least), I made the order of the languages in the editing UI be preserved on submission. That includes the ability to change the order via edit. The alternative would be to just always order them by language ID.


yvanzo:
Other options are to add a `primary` flag (x)or a `weight` percentage within the `work_language` table.

mwiencek:
> Other options are to add a primary flag (x)or a weight percentage within the work_language table.

Agreed, there are much better ways than abusing the `created` column.

I couldn't find any discussion about this though, so I didn't want to make any other additions to the schema on my own. Especially since a lot of code is shared with IPIs and ISNIs. I wanted the tables to keep the same structure to avoid special-cases.

My feeling was, the order might be important so let's preserve it as submitted (seems intuitive enough), but not set anything in stone before there's a discussion about it.

mwiencek:
* Tests fixed
* mmd-schema updated in https://github.com/metabrainz/mmd-schema/tree/schema-change-2017-q2
* search-server updated in https://github.com/metabrainz/search-server/pull/4

Freso:
It seems like the web service now returns the "first" language as its single language, but shouldn't it rather return "mul" if there are >1 languages given (isn't that what it's doing now?)?

mwiencek:
> It seems like the web service now returns the "first" language as its single language [...], but shouldn't it rather return "mul" if there are >1 languages given (isn't that what it's doing now?)?

That's a good point. For some reason I was only thinking of the cases where it wasn't already set to mul (I guessed in some cases people would've chosen the primary language or left it blank), and wanted to not change the data already being returned. But what I have now is changing it where it was already set to mul, and returning mul now is arguably more correct.

mwiencek:
Changed in https://github.com/metabrainz/musicbrainz-server/commit/6a6b9d1c80fafe688ccdd71fb89584ebdb15a9ae

PR #507, opened 2017-05-03
MBS-9329: Populate event_meta

mwiencek:


mwiencek:
I noticed this when I was writing a script to generate sample data for the JSON dump, using rating/tag information to order entities by "popularity."

mwiencek:
Ran these changes on bowie so I can use the stats for this script. There were only 140 ratings.

PR #511, opened 2017-05-17
MBS-9347: Fix regression regarding page titles

yvanzo:
Fix [MBS-9347](https://tickets.metabrainz.org/browse/MBS-9347): append “- MusicBrainz” to the title of all pages but the homepage.

### DIFF
--- a/root/layout/components/Head.js
+++ b/root/layout/components/Head.js
@@ -18,7 +18,7 @@ function canonicalize(url) {
 function getTitle(props) {
   let {title, pager} = props;
 
-  if (props.homepage) {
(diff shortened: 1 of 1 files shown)


yvanzo:
Initially reported by @jesus2099 in https://chatlogs.metabrainz.org/brainzbot/metabrainz/msg/3909309/
@mwiencek: Should it be made against production branch to hot-fix the current server?

mwiencek:
It doesn't seem like it causes any major breakage (I could be wrong), so master seems fine unless there's a reason it can't wait.

yvanzo:
Agreed.

PR #518, opened 2017-06-03
Fix URL::BigCartel module name

yvanzo:


yvanzo:
@brainzbot retest this please

PR #515, opened 2017-05-23
Reduce spammer motivation (MBS-9353, MBS-9354 MBS-9355)

yvanzo:
- [MBS-9353](https://tickets.metabrainz.org/browse/MBS-9353): Do not show collections or user profiles to people not logged in
- <s>[MBS-9354](https://tickets.metabrainz.org/browse/MBS-9354): Block `/collection` as per `robots.txt`</s> (see https://github.com/metabrainz/musicbrainz-server/pull/519#discussion_r120005672 instead)
- [MBS-9355](https://tickets.metabrainz.org/browse/MBS-9355): Do not allow limited user to set bio/link

PR #519, opened 2017-06-04
Disallow more stuff in robots.txt and use Crawl-delay option

zas:


yvanzo:
@brainzbot, add to whitelist

root/robots.txt.production
```diff
@@ -5,12 +5,14 @@ Disallow: /ajax/
 Disallow: /area/
 Disallow: /cdstub/
 Disallow: /cdtoc/attach
+Disallow: /collection/create
```
yvanzo: This will be covered by https://github.com/metabrainz/musicbrainz-server/pull/515/files#diff-d18cb9e7c5f98fd00839f1bfa79034e9R8

root/robots.txt.production
```diff
@@ -19,15 +21,36 @@ Disallow: /register
 Disallow: /report/
 Disallow: /reports
 Disallow: /search
+Disallow: /set-beta-preference
+Disallow: /statistics/timeline
```
mwiencek: How come? We don't really link it anywhere except the menu and it seems useful to show up in searches for "musicbrainz timeline" etc.

mwiencek:
I think we should block `/set-language/` too.

zas:
> How come? We don't really link it anywhere except the menu and it seems useful to show up in searches for "musicbrainz timeline" etc.

Oops, i added this one by error, fixed

PR #521, opened 2017-06-05
MBS-9309: Ignore swapped recordings in medium merges

mwiencek:
If two mediums are being merged, where recording 1 on medium 1 is being merged into recording 2 on medium 2, and recording 2 on medium 2 is being merged into recording 1 on medium 1 (because the positions are swapped), then the current merge logic will fail, because after performing the first merge, it proceeds to try and merge the same recording into itself.

This situation is now ignored, so no merge is performed, which I think is safer (it certainly wouldn't have been intended in edit #40241956). I'm not even sure it makes sense to support the scenario in MBS-8614 anymore; that should probably fail the edit and at the very least display a warning.

I think we should show a warning in this case, too, but we have separate pieces of logic which should probably be unified first (calculate_recording_merges, determine_recording_merges, this), and the edit referenced in the ticket has already been open for 10 months, so that can probably wait.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Medium.pm
+++ b/lib/MusicBrainz/Server/Data/Medium.pm
@@ -230,7 +230,7 @@ sub merge
 {
     my ($self, $new_medium_id, $old_medium_id) = @_;
     my @recording_merges = @{
-        $self->sql->select_list_of_lists(
+        $self->sql->select_list_of_hashes(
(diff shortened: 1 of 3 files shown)


mwiencek:
@yvanzo I'd say this is probably fine to merge, even if improvements can be made later. Agreed? :)

mwiencek:
@yvanzo gave this a a merge approval on IRC

PR #523, opened 2017-06-07
MBS-9369: Add docker image creating database & importing data dumps

ferbncode:
Added template `Dockerfile.import-database.m4` for creating MB database and importing dumps for development environment of other projects. Also, while running the image, if host directory containing dumps is not mounted as a volume, latest dumps (`mbdump.tar.bz2` and `mbdump-derived.tar.bz2`) are downloaded and the import is then carried out.


mwiencek:
Can you modify the existing test database image instead, and just control the imports via an environment variable?

docker/templates/Dockerfile.import-database.m4
```diff
@@ -0,0 +1,42 @@
...
+    sudo_mb(`carton install --deployment') && \
+    apt_purge(`test_db_build_deps')
+
+COPY admin/ admin/
+COPY lib/ lib/
+COPY script/ script/
+COPY t/sql/initial.sql t/sql/
+COPY entities.json entities.json
+RUN mkdir dumps && \
+    chmod 777 dumps
```
mwiencek: I think this directory should be in the home directory, not the source checkout. Someone might want to mount a volume in either place. Replace both lines with ``chown_mb(`/home/musicbrainz/dumps')``.

script/import_db.sh
```diff
@@ -0,0 +1,28 @@
...
+
+DATABASE=READWRITE
+DIRECTORY=./dumps
+
+source ./admin/functions.sh
+
+if ! script/database_exists $DATABASE; then
+    count=`ls -l $DIRECTORY/mbdump*.tar.bz2 2>/dev/null | wc -l`
+    if [ $count -ne 0 ]; then
+        echo "Creating Database and Importing Dump"
```
mwiencek: I'd remove all these `echo`s, since they don't say anything additional, or otherwise fix the capitalization. :)

script/import_db.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
mwiencek: This file should go under `docker/scripts/`.

script/import_db.sh
```diff
@@ -0,0 +1,28 @@
...
+
+set -o errexit
+cd `dirname $0`/..
+
+DATABASE=READWRITE
+DIRECTORY=./dumps
+
+source ./admin/functions.sh
```
mwiencek: This line can be removed.

alastair:
There are some additional perl dependencies required to do the import - is it OK to add them to the test db image?

script/import_db.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if ! script/database_exists $DATABASE; then
+    count=`ls -l $DIRECTORY/mbdump*.tar.bz2 2>/dev/null | wc -l`
+    if [ $count -ne 0 ]; then
```
alastair: You could reverse this check so that you download the files if they don't exist, then you only need to have the `InitDb.pl` line once

script/import_db.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o errexit
+cd `dirname $0`/..
+
+DATABASE=READWRITE
+DIRECTORY=./dumps
+
+source ./admin/functions.sh
+
+if ! script/database_exists $DATABASE; then
+    count=`ls -l $DIRECTORY/mbdump*.tar.bz2 2>/dev/null | wc -l`
```
alastair: There are a few other ways of doing this without running 2 extra commands, you could try one of the options here: https://stackoverflow.com/questions/2937407/test-whether-a-glob-has-any-matches-in-bash

If you stay with the current solution, you don't need the `-l` flag to ls (ls will automatically print one item per line if its output is a pipe). Also, you use quotes here, but `$()` in the command below to get the value of `LATEST`. These are equivalent and you should be consistent.

mwiencek:
> There are some additional perl dependencies required to do the import - is it OK to add them to the test db image?

Yeah, I'd greatly prefer that than duplicating the files with only minimal changes.

docker/templates/Dockerfile.import-database.m4
ferbncode: I tried running `chown` but I get an error stating `Permission Denied` when files are downloaded in the directory.

script/import_db.sh
ferbncode: I have moved `create_test_db.sh`. Is it preferable to move `script/create_test_db.sh` to `docker/scripts/`?

mwiencek: No, sorry, I didn't mean to suggest you modify or move create_test_db.sh at all. I meant keep the existing import_db script you had, and call it when the container starts if an environment variable is set. Currently we have docker/musicbrainz-test-database/create_test_db.sh run *when the image is built*, but you of course shouldn't import dumps when building the image, only after the container starts. There will already be an (empty) database at that point, but you can just use `dropdb` first.

docker/templates/Dockerfile.test-database.m4
```diff
@@ -22,6 +22,7 @@ COPY \
     ./
 
 ENV PERL_CPANM_OPT --notest --no-interactive
+ENV test=true
```
mwiencek: This should be removed from the Dockerfile (it should default to not set), and you should use a more descriptive name elsewhere, like `MB_IMPORT_DUMPS`.

docker/templates/Dockerfile.test-database.m4
```diff
@@ -6,7 +6,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 # install_extensions.sh removes certain build dependencies that we need, so we
 # can't install everything here.
 # Note: curl is also a dependency of carton.
-RUN apt_install(`ca-certificates curl sudo')
+RUN apt_install(`ca-certificates curl sudo wget')
```
mwiencek: `wget` should already be installed, but if it's not, is there a reason you can't use `curl`?

docker/templates/Dockerfile.test-database.m4
```diff
@@ -31,6 +32,10 @@ COPY admin/ admin/
 COPY lib/ lib/
 COPY script/ script/
 COPY t/sql/initial.sql t/sql/
+COPY entities.json entities.json
```
mwiencek: Why is that needed?

docker/templates/Dockerfile.import-database.m4
mwiencek: Ah, I guess that's because the script is run as the `postgres` user. So this should be correct:

```
RUN mkdir -p /home/musicbrainz/dumps && \
    chown -R postgres:postgres /home/musicbrainz/dumps
```

script/create_test_db.sh
```diff
@@ -1,89 +0,0 @@
-#!/bin/bash
```
mwiencek: As mentioned in https://github.com/metabrainz/musicbrainz-server/pull/523#discussion_r120760834, this file should be left alone, and you should decide which script to call in docker/musicbrainz-test-database/create_test_db.sh based on the value of `$MB_IMPORT_DUMPS` (which is set on the container via `docker run --env`).

mwiencek: It's just that script/create_test_db.sh should stay where it is because too many things depend on its location.

And changing the file like you did works, but I'd rather not put everything under a giant `if` block and just keep the scripts separate, since they're for different things. :)

docker/templates/Dockerfile.test-database.m4
ferbncode: In `admin/MBImport.pl`, `lib/MusicBrainz/Server/Constants.pm` is [used](https://github.com/metabrainz/musicbrainz-server/blob/master/admin/MBImport.pl#L36), which further makes use of `entities.json` [here](https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Constants.pm#L380).

docker/docker-compose.tests.yml
```diff
@@ -9,6 +9,8 @@ services:
     build:
       context: ../
       dockerfile: docker/Dockerfile.test-database
+    environment:
+      MB_IMPORT_DUMPS: "false"
```
mwiencek: In shell world, "false" is true. ;) (Any non-empty string is truthy.) This should be removed, since an undefined variable will do what you intend.

docker/scripts/import_db.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -o errexit
+cd `dirname $0`/../..
```
mwiencek: Since this is only intended to run inside a container (it has hardcoded paths), let's just use `cd /home/musicbrainz/musicbrainz-server` here for clarity.

docker/docker-compose.tests.yml
ferbncode: This variable `MB_IMPORT_DUMPS` being set to `false` works as it actually runs the bash command `false`. However, as you suggested, that an undefined variable should do what is expected and the [issues](https://stackoverflow.com/questions/2953646/how-to-declare-and-use-boolean-variables-in-shell-script/21210966) with the my current syntax, I have changed the syntax to `if ["$MB_IMPORT_DUMPS" = "true"]`. :sweat_smile:

mwiencek: Yeah, I missed that you didn't have the brackets before. Thanks for changing it though. 👍

PR #524, opened 2017-06-11
MBS-9372: Allow notes from voting-disabled editor

yvanzo:
Fix bug [MBS-9372](https://tickets.metabrainz.org/browse/MBS-9372): Revoking editing/voting editor rights also renders them unable to write edit notes.

Prior to this patch, if admin revoked editing/voting privileges of any editor, this editor would not be able to submit edit notes, since it would be incidentally disallowed at the same time as votes.

This patch refines permissions to submit edit notes and votes so to allow submitting edit notes when editor is editing/voting-disabled and no vote is submitted.  If any vote is submitted at the same time, edit notes are rejected as well and an error page is returned.

### DIFF
--- a/root/edit/cannot_vote.tt
+++ b/root/edit/cannot_vote.tt
@@ -1,0 +1,4 @@
+[% WRAPPER 'layout.tt' full_width=1 title=l('Error Voting on Edits') %]
+    <h1>[% l('Error Voting on Edits') %]</h1>
+    <p>[% l('You’re not currently allowed to vote, please read the banner.') %]</p>
+[% END %]
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -81,16 +81,21 @@ sub data : Chained('load') RequireAuth
                template => 'edit/data.tt' );
 }
 
-sub enter_votes : Local RequireAuth(editing_enabled) DenyWhenReadonly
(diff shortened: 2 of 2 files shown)


root/edit/cannot_vote.tt
```diff
@@ -0,0 +1,8 @@
+[% WRAPPER 'layout.tt' full_width=1 title=l('Error Voting Edit') %]
+    <h1>[% l('Error Voting Edit') %]</h1>
```
mwiencek: "Error Voting on Edits" sounds more correct to me.

root/edit/cannot_vote.tt
```diff
@@ -0,0 +1,8 @@
...
+    <p>
+        [% l('You’re currently not allowed to vote because an admin has revoked your privileges.
```
mwiencek: Could we use the same message that's in root/layout.tt? I guess the only difference is `s/vote/edit or vote/`. But that seems perfectly fine and then won't have to be translated again.

mwiencek: A macro for this `<p>` might be even better.

yvanzo: Oops, the message is actually displayed twice. I simplified the message instead in https://github.com/metabrainz/musicbrainz-server/pull/524/commits/03f2172b2abcaed0661ce723dede01c363d59217

yvanzo:
Fixed up.

PR #526, opened 2017-06-26
Update RemoveSpamAccount admin script for MBS-9357 (part 1)

yvanzo:
First part of [MBS-9357](https://tickets.metabrainz.org/browse/MBS-9357):
- Allow filtering on **email**, **website**, and **bio** (additionally to name)
- Support POSIX regular expression (in place of prefix)
- Use case-sensitive operator by default (instead of insensitive)
- Preserve accounts with edits/votes/OAuth tokens by default
- Add options for operator case-sensitivity and verbosity level
- Show help message (documentation) in case of incorrect usage

### DIFF
--- a/admin/RemoveSpamAccounts.pl
+++ b/admin/RemoveSpamAccounts.pl
@@ -5,7 +5,7 @@ use warnings;
 #
 #   MusicBrainz -- the open internet music database
 #
-#   Copyright (C) 2011 MetaBrainz Foundation
+#   Copyright (C) 2011-2017 MetaBrainz Foundation
 #
 #   This program is free software; you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


admin/RemoveSpamAccounts.pl
```diff
@@ -28,46 +28,130 @@
...
+my $pattern = undef;
+my $column = undef;
+
 GetOptions(
+    "column|c=s"  => \$column,
+    "pattern|p=s"  => \$pattern,
     "dryrun|d"    => \$dry_run,
     "force|f"     => \$force,
-) or return 2;
+    "caseinsensitive|i"    => \$case_insensitive,
```
mwiencek: `ignore-case|i` for matching `i`s?

admin/RemoveSpamAccounts.pl
```diff
@@ -28,46 +28,130 @@
...
+my $c = MusicBrainz::Server::Context->create_script_context();
+my $sql = Sql->new($c->conn);
+my $dbh = $c->dbh;
+
+my $regexp_operator = $case_insensitive ? '~*' : '~';
+my $editors = $c->sql->select_list_of_hashes("SELECT id, name, $column FROM editor WHERE $column $regexp_operator ?", $pattern);
 foreach my $ed (@{$editors}) {
+    my $details = "'" . $ed->{name} . "'";
+    if ($verbose && $column ne 'name') {
+        $details .=  " [${column}='" . $ed->{$column} . "']";
```
mwiencek: Since you're intentionally adding surrounding `''`, I assume you want that to be a valid SQL literal, so I might suggest `$dbh->quote($ed->{$column})` instead. I suppose it doesn't matter if this is just for logging, but outputting unescaped quotes could potentially be irritating in either case. :P

PR #530, opened 2017-07-19
MBS-9409: Disallow merging mediums with & without pregaps

mwiencek:
This does what the title says (because I believe allowing it would be annoying and error-prone), and also fixes the UI to allow moving track 1 to track 0, so that such releases can be made mergeable if it's correct to. That can be accomplished by enabling the pregap track and (1) moving track 1 up and deleting the now-empty track 1, or (2) removing the empty track 0 (which now moves track 1 into its place).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -548,12 +548,13 @@ around _merge_submit => sub {
         };
     }
 
-    if ($c->model('Release')->can_merge(%merge_opts)) {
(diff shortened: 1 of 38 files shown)


mwiencek:
Jenkins returned some unrelated test failures related to our JSON-LD, because the script contents weren't valid XML. I removed the XHTML checks because they're pointless nowadays.

mwiencek:
Unrelated JS failures should be fixed by https://github.com/metabrainz/musicbrainz-server/pull/531

PR #531, opened 2017-08-01
Run UI tests in a headless chrome environment

mwiencek:
Our Node/jsdom setup is too finicky and has hard-to-debug, transient failures in the ArtistCreditEditor tests. We can't really test for element visibility properly without a real DOM.

This also updates some modules and adds a ton of missing `require()` statements, getting us closer to being able to remove global variable hacks.

Most of the changes under root/static/scripts/release-editor/ are just whitespace/indentation.


mwiencek:
Unrelated JSON-LD failures should be fixed by https://github.com/metabrainz/musicbrainz-server/pull/530

mwiencek:
Merging so I can start writing integration tests that depend on this work.

PR #532, opened 2017-08-03
Add TO_JSON method to PartialDate.

mkubenka:
Changes introduced in 23e11bd break *Add release* page because `PartialDate`do not have `TO_JSON` method:

```
Caught exception in MusicBrainz::Server::Controller::ReleaseEditor->add "Can't locate object method "TO_JSON" via package "MusicBrainz::Server::Entity::PartialDate" at lib/MusicBrainz/Server/Entity/Role/DatePeriod.pm line 76."
```

### DIFF
--- a/lib/MusicBrainz/Server/Entity/PartialDate.pm
+++ b/lib/MusicBrainz/Server/Entity/PartialDate.pm
@@ -1,6 +1,8 @@
 package MusicBrainz::Server::Entity::PartialDate;
 use Moose;
 
+extends 'MusicBrainz::Server::Entity';
+
 use Date::Calc;
 use List::AllUtils qw( any first_index );
 use MusicBrainz::Server::Data::Utils qw( take_while );


mwiencek:
Hi, thanks for the pull request. Can you give some steps to reproduce the error? I can load /release/add and even submit a release okay, so I'm wondering where it's trying to serialize a `PartialDate` to JSON.

lib/MusicBrainz/Server/Entity/PartialDate.pm
```diff
@@ -1,6 +1,8 @@
 package MusicBrainz::Server::Entity::PartialDate;
 use Moose;
 
+extends 'MusicBrainz::Server::Entity';
```
mwiencek: I don't think it's correct to extend `MusicBrainz::Server::Entity`, since it's not exactly an entity in the way that's used, and has other effects like adding `entityType` to the JSON.

I think we should just define a custom `TO_JSON` method in this file that serializes the year/month/day.

But I'm still wondering how to trigger this error. :)

yvanzo: How to reproduce: Checkout the commit preceding 23e11bdd65d9d5c92328af47496bb256daaa8255, compile resources (against outdated scripts), checkout master, run the server, try to edit any entity with either date period or partial date set.

This bug cannot be triggered on production/beta/master with up-to-date compilation of resources.

Reported in https://tickets.metabrainz.org/browse/MBS-9446

Other solution with custom `TO_JSON` method proposed in PR #541.

PR #533, opened 2017-08-09
Kill aclass

mwiencek:
Replaces aclass with standard ES6 classes. The major change involved in doing this was adding `new` to all the instantiations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,11 +50,6 @@
         }
       }
     },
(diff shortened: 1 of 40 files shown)


JesseWeinstein:
Fixed version of Paste-a-date! that was broken by this change, here: https://gist.github.com/JesseWeinstein/368bddf748f4faf44b5775addf7f3db0

PR #536, opened 2017-08-21
MBS-9353: Fix RequireAuth attributes for user profile and collection

yvanzo:
[MBS-9353](https://tickets.metabrainz.org/browse/MBS-9353): Do not show collections or user profiles to people not logged in

Set `RequireAuth` attribute to every subroutine of Collection & User rather than `base` subroutine.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User.pm
+++ b/lib/MusicBrainz/Server/Controller/User.pm
@@ -303,7 +303,7 @@ sub contact : Chained('load') RequireAuth HiddenOnSlaves
     }
 }
 
-sub collections : Chained('load') PathPart('collections')
+sub collections : Chained('load') PathPart('collections') RequireAuth
 {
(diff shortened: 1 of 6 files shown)


lib/MusicBrainz/Server/Controller/User.pm
```diff
@@ -465,7 +465,7 @@ sub tag : Chained('load') PathPart('tag') Args(1)
     );
 }
 
-sub privileged : Path('/privileged')
```
mwiencek: I don't think we need/ought to `RequireAuth` for `/privileged`.

yvanzo: That was completely unnecessary and is reverted in https://github.com/metabrainz/musicbrainz-server/pull/536/commits/69d3c67b20df402de75ae1e0a9ae73436555e2e1

PR #540, opened 2017-08-31
JSON dumps

mwiencek:
This has been running in production since...April? So the actual dumps work, but this contains a lot of other changes too, mostly to speed up the web service significantly.


mwiencek:
I'm still going through and restructuring the commits to be smaller and more logical, but the general changes will remain the same.

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

lib/MusicBrainz/Script/Role/IncrementalDump.pm
```diff
@@ -18,9 +18,7 @@ use MusicBrainz::Server::Replication::Packet qw(
     decompress_packet
     retrieve_remote_file
 );
-use Parallel::ForkManager 0.7.6;
```
yvanzo: ForkManager `0.7.6` was 2010, any chance that the forking errors would vanish using the latest `1.1.9`?

mwiencek: We've definitely been using 1.19 in production. The `0.7.6` just indicates the minimum version the code will work with and croaks otherwise (IIRC there were API changes).

I'm just not going to attempt writing any concurrent code in Perl anymore. Even JavaScript is a far better language for that nowadays. :(

PR #543, opened 2017-09-06
MBS-9450: Import DateTime module to UserDate

yvanzo:
Fix the most plausible cause of ISEs in MBS-9450.

### DIFF
--- a/lib/MusicBrainz/Server/Plugin/UserDate.pm
+++ b/lib/MusicBrainz/Server/Plugin/UserDate.pm
@@ -6,6 +6,8 @@ use Text::Trim qw( trim );
 
 use base 'Template::Plugin';
 
+use DateTime;
+
 sub preferences { shift->{preferences}; }
 sub locale { shift->{locale}; }
 


mwiencek:
The issue is actually something different:

```
Can't locate object method "clone" via package "2017-09-06 13:31:00.46611+00"
```

This means that `$dt` is literally the string `"2017-09-06 13:31:00.46611+00"`, so it's trying to call `->clone` on a string.

I figured out where these strings are coming from (my fault of course): I deployed a new version of the sitemaps builder last night which contains commits from #540, including https://github.com/metabrainz/musicbrainz-server/pull/540/commits/b51116cd0dbe46289ffbf465a04d4e281538312b. Unfortunately I didn't realize that the sitemaps container shares a Redis cache with the production containers.

Sorry about that. :( I've separated the caches now and cleared the production cache to remove the bad strings.

yvanzo:
@mwiencek: Still, shouldn't `DateTime` be imported in this module (although it is already imported where it is used)?

mwiencek:
No, AFAIK you don't have to import packages that your subroutines operate on, unless they need to create objects of that class. So `DateTime` only has to be imported where `$dt` was created.

PR #544, opened 2017-09-09
MBS-9455: Show orderable relationship attributes absent grouping

mwiencek:
For orderable link types, we "cleaned" non-required attributes from the link phrases in `interpolate` so that they are grouped together in the relationships display and editor UIs. However, the release relationship editor as well as recording-work relationships on the release page don't require any grouping, so this had the effect of hiding those attributes there. This patch makes that behavior optional, so that the cleaned phrases are only used where grouping is intended.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Relationship.pm
+++ b/lib/MusicBrainz/Server/Entity/Relationship.pm
@@ -237,11 +237,7 @@ sub _interpolate {
     foreach my $attr (@attrs) {
         my $name = lc $attr->type->root->name;
 
-        if (exists $attrs{$name}) {
-            push @{$attrs{$name}}, $attr->html;
-        } else {
-            $attrs{$name} = [ $attr->html ];
-        }
+        push @{$attrs{$name}}, $attr->html;
     }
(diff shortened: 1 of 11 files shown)


mwiencek:
@brainzbot test this please

lib/MusicBrainz/Server/Entity/Relationship.pm
```diff
@@ -224,20 +244,19 @@ sub _interpolate {
         }
     }
     my %extra_attrs = %attrs;
-    my $type_is_orderable = $self->link->type->orderable_direction > 0;
 
     # Ordered relationships in a series should all share the same link phrase,
     # even if their attributes differ, so that they remain grouped together
     # in the relationships display.
-    %attrs = () if $type_is_orderable;
+    %attrs = () if $for_grouping;
```
yvanzo: I would move this comment to the above `_build_grouping_phrase` subroutine and make a more general comment about dropping unnecessary attributes for grouping here.

mwiencek: Should be better now :)

PR #548, opened 2017-09-10
Replace babel es2015 preset with env, add babel-runtime

mwiencek:
The env preset is intended to replace es2015, and is better for us since we can simply specify which browsers we're targeting in .babelrc, and it'll perform the correct transformations automatically (based on the features supported in those browsers).

babel-runtime increases the size of our JS a bit (see the commit for details), but allows us to use async + await, which are essential to sanely work with promises.


mwiencek:
This patch is a no-op as far as the output goes, as expected (hashes of the JS bundles in root/static/build/ were identical before and after).

PR #550, opened 2017-09-11
First flow types; port external links editor tests to Selenium

mwiencek:
This begins adding static type checking to our JavaScript with Flow: https://flow.org/

Only files beginning with a `// @flow` comment are checked, so this can be done incrementally.

The types aren't yet checked during builds. I'll do that in a future PR, but for now devs are expected to run `flow` on their own.

The types are stripped via a babel plugin.

This PR also ports the external links editor tests to Selenium, because I had trouble getting the finicky JS ones to pass consistently.


mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

.flowconfig
```diff
@@ -1,11 +1,13 @@
 [ignore]
 .*/lib/.*
+.*/node_modules/jsx-control-statements/jsx-control-statements.flow.js
 .*/root/static/build/.*
 .*/t/.*
 
 [include]
 
 [libs]
+./node_modules/jsx-control-statements/jsx-control-statements.latest.flow.js
```
yvanzo: Using `npm` `3.10.10`, I have `./node_modules/jsx-control-statements/jsx-control-statements.flow.js` instead, resulting into the following message in `flow` logs:

> Skipping ./node_modules/jsx-control-statements/jsx-control-statements.latest.flow.js: No such file or directory

mwiencek: Huh, guess I didn't `git add` the package.json changes. It should be fixed now if you run `npm install` again.

mwiencek: (I'd also recommend doing `npm install -g npm@latest` if you can though, the latest versions of `npm` seem much less flaky/buggy for me.)

root/static/scripts/edit/externalLinks.js
```diff
@@ -91,15 +108,15 @@ class ExternalLinksEditor extends React.Component {
   }
 
   getOldLinksHash() {
```
yvanzo: Shall we check return type (similarly to parameters type)?

mwiencek: Good point actually. Flow has type inference, so normally we don't have to specify types where it can figure them out on its own. But for external modules, [it says this](https://flow.org/en/docs/libdefs/#toc-general-best-practices):

> If a third-party library that has no type information is used by your project, Flow will treat it like any other untyped dependency and mark all of its exports as `any`. Interestingly, this is the only place that Flow will implicitly inject `any` into your program.

Since we're returning the result of a lodash function, I believe the inferred return type is `any`, which disables type checking wherever the function is used.

So adding return types is a good idea wherever we're calling an external library function and returning its result. A better idea, if possible, is pulling in types for the library via `flow-typed` (mentioned in the link above), and letting flow infer the types from that.

mwiencek: flow-typed only has types for lodash v4, so I'd prefer to do that in a later PR.

yvanzo: I started to tickle updating to lodash v4 alredy.

PR #554, opened 2017-09-19
MBS-9471: Do not map undefined rel attribute

yvanzo:
### Fix [MBS-9471](https://tickets.metabrainz.org/browse/MBS-9471): TypeError: Cannot read property 'name' of undefined
Should hotfix beta/production, since editing form is currently unusable for most works with relationships.

### DIFF
--- a/root/static/scripts/edit/utility/linkPhrase.js
+++ b/root/static/scripts/edit/utility/linkPhrase.js
@@ -13,7 +13,9 @@ const clean = require('../../common/utility/clean');
 var attributeRegex = /\{(.*?)(?::(.*?))?\}/g;
 
 function mapNameToID(result, info, id) {
-    result[info.attribute.name] = id;
+    if (info.attribute !== undefined) {
+        result[info.attribute.name] = id;
+    }
 }
 
 exports.clean = _.memoize(function (linkType, backward) {


root/static/scripts/edit/utility/linkPhrase.js
```diff
@@ -13,7 +13,9 @@ const clean = require('../../common/utility/clean');
 var attributeRegex = /\{(.*?)(?::(.*?))?\}/g;
 
 function mapNameToID(result, info, id) {
-    result[info.attribute.name] = id;
+    if (info.attribute !== undefined) {
```
mwiencek: It doesn't make sense for this to be undefined, and will cause things to break elsewhere, so the bug is actually: what is causing this to be undefined?

It looks like the problem is that @reosarevok added a new link attribute type named "number," but there's already one named "number," so it overwrites the original one here: https://github.com/metabrainz/musicbrainz-server/blob/cb20c520de4f197d3b83a8a7624d91b6f88a3b2c/lib/MusicBrainz/Server/Form/Utils.pm#L221

We'll have to see if it's necessary for things to be keyed by name there, or for names to be unique (I believe they've always been up until this point, perhaps 'cause we only expose the name in the web service, and not for any other reason).

mwiencek:
Closing in favor of https://github.com/metabrainz/musicbrainz-server/pull/555

PR #557, opened 2017-09-30
MBS-9488: Changing "Limited editor" to "Beginner"

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9488 - I think I caught it all but I can't test stuff locally because no local DB.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Vote.pm
+++ b/lib/MusicBrainz/Server/Data/Vote.pm
@@ -52,7 +52,7 @@ sub enter_votes
     Sql::run_in_transaction(sub {
         $self->sql->do('LOCK vote IN SHARE ROW EXCLUSIVE MODE');
 
-        # Deal with votes on closed or own edits, by limited users, etc.
+        # Deal with votes on closed or own edits, by beginners/limited users, etc.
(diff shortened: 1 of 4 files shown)


mwiencek:
After seeing it I may have changed my mind about renaming it in the code, since it's confusing there (new implies only recent to me, and limited is already well understood in the code). Of course I'm fine with changing the user visible text. Sorry. :/ Unless @yvanzo disagrees.

reosarevok:
Changed this to only make visual changes.

yvanzo:
Actually “new implies only recent to me” too, and probably to many people as well, whereas this status is based on time and editing experience (less than 10 accepted edits). Would “beginner” be a best fit?

reosarevok:
I thought about that but considered that new can be seen in both time and experience (even if I created my account ten years ago, if I haven't edited anything, I'm still a new editor). But I'm ok with both options.

yvanzo:
That would be the same wording as in “[Beginners Guide](https://musicbrainz.org/doc/Beginners_Guide)”.

mwiencek:
I'm ok with either, I like "Beginner" slightly more.

reosarevok:
Changed it to "Beginner"

PR #558, opened 2017-10-02
MBS-9496: Modbot "AC being changed" edit note is confusing

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9496

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -469,6 +469,7 @@ with 'MusicBrainz::Server::Controller::Role::Edit' => {
                 my ($edit, $form) = @_;
 
                 my $editid = $edit->id;
+                my $artistname = $artist->name;
                 my $name = $form->field('name')->value;
                 if ($name ne $artist->name) {
                     my %rename = %{ $form->rename_artist_credit_set };
@@ -485,7 +486,7 @@ with 'MusicBrainz::Server::Controller::Role::Edit' => {
                         $c->model('EditNote')->add_note(
                             $ac_edit->id,
                             {
-                                text => "The artist name has been changed in edit #$editid.",
+                                text => "This credit is being changed because the main name for the artist “$artistname” is being modified by edit #$editid.",
                                 editor_id => $EDITOR_MODBOT
                             }
                         );


reosarevok:
Changed!

lib/MusicBrainz/Server/Controller/Artist.pm
```diff
@@ -485,7 +486,7 @@ with 'MusicBrainz::Server::Controller::Role::Edit' => {
                         $c->model('EditNote')->add_note(
                             $ac_edit->id,
                             {
-                                text => "The artist name has been changed in edit #$editid.",
+                                text => "This credit is being changed because the main name for the artist qq{$artistname} is being modified by edit #$editid.",
```
mwiencek: `qq{$artistname}` should just be `$artistname` (you're already inside quotes so that would literally output `qq{U2}`).

Also, should we add display quotes around `“$artistname”`? e.g. `the artist “is” is` reads a bit more clearly than `the artist is is`.

mwiencek:
@yvanzo These notes aren't translated at present, unless I misunderstood you. I'm not sure they could be easily right now, since we only have the interpolated values when the note is created.

lib/MusicBrainz/Server/Controller/Artist.pm
reosarevok: Adding display quotes is exactly what I was trying to achieve here. Do I just use “ ” then?

mwiencek: Yes, you can use the unicode ones directly in the string.

Otherwise the ASCII ones would just need to be escaped with a backslash: `artist \"$artistname\" is`.

`qq` only makes sense on the outside as an alternative quote operator, so you can do e.g. `qq("a'bunch"of'nonsense")` without having to escape those normal quotes.

yvanzo: For the record, it requires both `utf8` and a backslash for a proper encoding of unicode quotes.
Fixed in https://github.com/metabrainz/musicbrainz-server/commit/9fc9a4cd99193729c720cf556937c3115f10ae1e

mwiencek: Now I'm curious, what happens without the backslash?

yvanzo: Fancy characters are displayed instead: https://musicbrainz.org/edit/49349519

It is only necessary for not-localized messages since encoding is done by [`l` function](https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Plugin/Translation.pm#L27).

mwiencek: Sorry, I meant I saw the edit in IRC, but I don't understand why the backslash is needed. I never had to backslash Unicode quotes before and couldn't find what this does in any documentation, so I want to learn what it does. :)

yvanzo: Sorry, I just meant it has the same weird output. I guess it is probably the same as:
`"This ... artist" . encode('utf-8', '“') . $artistname . encode('utf-8', '”') . "is ..."`

Note that this message is not passed directly to template. It is stored into the database first. That is probably why it requires special attention compared to other messages.

PR #559, opened 2017-10-09
SOLR-47 Fix DropTriggers creation function

samj1912:
Allow the regex to parse the schema when specified.
See https://github.com/metabrainz/sir/blob/master/sir/trigger_generation/sql_generator.py#L47
Fixes https://tickets.metabrainz.org/browse/SOLR-47


yvanzo:
@brainzbot, add to whitelist.

PR #561, opened 2017-10-14
MBS-9313, MBS-9505: Auto-editing release event, amending within one day

yvanzo:
- Fix bug [MBS-9313](https://tickets.metabrainz.org/browse/MBS-9313): Adding first release event should be an auto-edit.
- Implement [MBS-9505](https://tickets.metabrainz.org/browse/MBS-9505): Extend amending delay from one hour to one day.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Release/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Release/Edit.pm
@@ -340,7 +340,7 @@ around allow_auto_edit => sub {
     return 0 if defined $self->data->{old}{language_id};
     return 0 if defined $self->data->{old}{script_id};
 
-    return 0 if defined $self->data->{old}{events};
+    return 0 if defined $self->data->{old}{events} && scalar @{ $self->data->{old}{events} } > 0;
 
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Edit/Release/Edit.pm
```diff
@@ -340,7 +340,7 @@ around allow_auto_edit => sub {
     return 0 if defined $self->data->{old}{language_id};
     return 0 if defined $self->data->{old}{script_id};
 
-    return 0 if defined $self->data->{old}{events};
+    return 0 if defined $self->data->{old}{events} && scalar $self->data->{old}{events} > 0;
```
mwiencek: That should be `scalar @{ $self->data->{old}{events} }` if events is an array ref.

yvanzo: Oops, fixed in https://github.com/metabrainz/musicbrainz-server/pull/561/commits/45fe51bbecf78561ba5832318ee61464be95f403

PR #562, opened 2017-10-17
MBS-9515: Block anchors in new Wikipedia URLs

yvanzo:
## [MBS-9515](https://tickets.metabrainz.org/browse/MBS-9515): Block adding Wikipedia relationships to anchors

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -196,6 +196,13 @@ class ExternalLinksEditor extends React.Component<LinksEditorProps, LinksEditorS
               error = l('This relationship type is deprecated and should not be used.');
             } else if ((!isPositiveInteger(link.relationship) || (oldLink && link.url !== oldLink.url)) && checker && !checker(link.url)) {
               error = l('This URL is not allowed for the selected link type, or is incorrectly formatted.');
+            } else if ((!isPositiveInteger(link.relationship) || (oldLink && link.url !== oldLink.url)) && /^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\/.*#/.test(link.url)) {
+              // Kludge for MBS-9515 to be replaced with the more general MBS-9516
+              error = l('Links to specific sections of Wikipedia articles are not allowed. Please remove “{fragment}” if still appropriate. See the {url|guidelines}.', {
+                __react: true,
+                fragment: <span className='url-quote'>{link.url.replace(/^(?:https?:\/\/)?(?:[^.\/]+\.)?wikipedia\.org\/[^#]*#(.*)$/, '#$1')}</span>,
+                url: { href: '/relationship/' + typeInfo.gid, target: '_blank' }
+              });
             } else if ((linksByTypeAndUrl[linkTypeAndUrlString(link)] || []).length > 1) {
               error = l('This relationship already exists.');
             }
--- a/root/static/styles/forms.less
+++ b/root/static/styles/forms.less
@@ -620,6 +620,12 @@ table.artist-credit-editor {
     margin-right: 4px;
 }
 
+#external-links-editor span.url-quote {
+    display: inline-block;
+    font-style: italic;
+    word-break: break-all;
+}
+
 /* Relationship editor */
 
 .remove-button { cursor: pointer; color: @button-text; padding: 0 0.5em; }


root/static/scripts/edit/externalLinks.js
```diff
@@ -196,6 +196,11 @@ class ExternalLinksEditor extends React.Component<LinksEditorProps, LinksEditorS
               error = l('This relationship type is deprecated and should not be used.');
             } else if ((!isPositiveInteger(link.relationship) || (oldLink && link.url !== oldLink.url)) && checker && !checker(link.url)) {
               error = l('This URL is not allowed for the selected link type, or is incorrectly formatted.');
+            } else if ((!isPositiveInteger(link.relationship) || (oldLink && link.url !== oldLink.url)) && /^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\/.*#/.test(link.url)) {
+              // Kludge for MBS-9515 to be replaced with the more general MBS-9516
+              error = l('Anchor (#Example_section) Wikipedia links are not allowed, see the {url|guidelines for URL}.', {
+                url: { href: '/doc/Style/Relationships/URLs', target: '_blank' }
```
reosarevok: This should probably link to something like https://musicbrainz.org/relationship/29651736-fa6d-48e4-aadc-a557c6add1cb 

It's kind of annoying that we don't have a central "Wikipedia page" rel to link to, but only one per entity type :/ But linking to a specific one, like artist, might be better than nothing.

root/static/scripts/edit/externalLinks.js
```diff
@@ -196,6 +196,12 @@ class ExternalLinksEditor extends React.Component<LinksEditorProps, LinksEditorS
               error = l('This relationship type is deprecated and should not be used.');
             } else if ((!isPositiveInteger(link.relationship) || (oldLink && link.url !== oldLink.url)) && checker && !checker(link.url)) {
               error = l('This URL is not allowed for the selected link type, or is incorrectly formatted.');
+            } else if ((!isPositiveInteger(link.relationship) || (oldLink && link.url !== oldLink.url)) && /^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\/.*#/.test(link.url)) {
+              // Kludge for MBS-9515 to be replaced with the more general MBS-9516
+              error = l('Anchors Wikipedia links are not allowed, please remove {fragment} if still appropriate, see the {url|guidelines}.', {
```
mwiencek: Will some users know what an anchor is? I guess it's fine, but I wonder if something like "Wikipedia links to specific sections of the page are not allowed, ..." would make more sense.

The comma splice at the end sounds slightly unnatural so I'd replace "appropriate, see" with "appropriate. See" too.

root/static/scripts/edit/externalLinks.js
```diff
@@ -196,6 +196,12 @@ class ExternalLinksEditor extends React.Component<LinksEditorProps, LinksEditorS
...
+                fragment: '<i>' + _.escape(link.url.replace(/^(?:https?:\/\/)?(?:[^.\/]+\.)?wikipedia\.org\/[^#]*#(.*)$/, '#$1')) + '</i>',
```
mwiencek: It's much safer to pass `__react: true` here and pass in a React element directly, like `fragment: <i>{link.url.replace(...)}</li>`. Then you can revert the `dangerouslySetInnerHTML` below.

mwiencek: Note that for the URL syntax you don't need to pass in an `<a>`, the i18n module will do that for you with `__react: true`. So the `url: {...}` part can stay as it is.

root/static/scripts/edit/externalLinks.js
yvanzo: https://github.com/metabrainz/musicbrainz-server/pull/562/commits/46feadc2646625daf81abd03ce12adc529f33f23 links to the specific relationship page for the current URL.
Besides, I reported the lack of central page for similar relationship types in [MBS-9517](https://tickets.metabrainz.org/browse/MBS-9517).

root/static/scripts/edit/externalLinks.js
yvanzo: Almost applied in https://github.com/metabrainz/musicbrainz-server/pull/562/commits/46feadc2646625daf81abd03ce12adc529f33f23

root/static/scripts/edit/externalLinks.js
yvanzo: Applied in https://github.com/metabrainz/musicbrainz-server/pull/562/commits/46feadc2646625daf81abd03ce12adc529f33f23, with extra style in https://github.com/metabrainz/musicbrainz-server/pull/562/commits/4007afdc64f65f392f1a769eaca344d18773fe37 that wraps the quoted URL fragment in order to break it if longer than line.

yvanzo:
Selenium tests seem to deadlock [build #631](https://ci.metabrainz.org/job/musicbrainz-server/631/).

PR #563, opened 2017-10-19
 Handle now-unexpected alerts in selenium tests

mwiencek:
The `confirm()` dialog in the release editor that tells you you're going to lose all your changes is now breaking the tests, because it prevents subsequent commands from running.

It's unclear why it worked previously, but an update to chrome broke it.

Also included are two unrelated improvements to selenium.js: checking if errors occurred on the page after every command, and logging every command (so when it fails we can see where it was at).


mwiencek:
Okay, looks like this fixes the build, so I'll merge.

PR #565, opened 2017-10-26
MBS-9525: retry fullexport rsync if it fails

mwiencek:
This has failed twice recently: https://community.metabrainz.org/t/mbdump-tar-bz2-missing-from-latest-full-export/333175/5

This just retries it forever, sleeping for 30 seconds between each attempt.

### DIFF
--- a/admin/functions.sh
+++ b/admin/functions.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 make_temp_dir()
 {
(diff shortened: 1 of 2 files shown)


yvanzo:
I am not comfortable with this infinite loop.
Would it be reasonable to limit the number of retries?
Or to double the pause duration after each retry?

mwiencek:
I think it's a good idea. And I figured out how to create a `retry` function in bash so we don't have to repeat the logic in both places now.

admin/functions.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
mwiencek: AFAICT this file is not sourced from any script that doesn't use `bash`.

mwiencek:
Oops, I accidentally committed with the delay I was testing with. I changed the initial delay back to 15 seconds (so the delay before the final attempt should be 4 minutes, I think).

PR #566, opened 2017-10-30
Reject edits with empty name to mitigate MBS-9512

yvanzo:
Mitigate bug [MBS-9512](https://tickets.metabrainz.org/browse/MBS-9512) _(Changing recording name to empty string should not be allowed)_ by rejecting edits which violate `only_non_empty` and `only_non_empty_sort_name` constraints.


mwiencek:
Discovered that it's really easy to reproduce this: just enter some spaces for one of the track names. 😆 

I pushed a new commit which fixes things on the client & server in the appropriate spots (resolving my previous review).

PR #567, opened 2017-11-06
MBS-9540: Add ratings for non top level entities

samj1912:
This is used in picard to get track ratings.


lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Release.pm
```diff
@@ -87,8 +87,8 @@ sub serialize
         # any), those of the release group should be serialized (to match the
         # behaviour of the XML serializer). So we call `serialize_rating`
```
mwiencek: The comment here ought to be updated too.

samj1912: hmm yes, and we can skip the intermediate $output below

samj1912:
@brainzbot retest this please

PR #574, opened 2017-12-01
Add flow types to the tag editor

mwiencek:
immutable-js had to be updated because the old version can't type Records properly. Not sure when it'll go out of RC, it's been months already.

### DIFF
--- a/root/static/scripts/common/components/TagEditor.js
+++ b/root/static/scripts/common/components/TagEditor.js
@@ -1,7 +1,9 @@
-// This file is part of MusicBrainz, the open internet music database.
+// @flow
 // Copyright (C) 2015 MetaBrainz Foundation
-// Licensed under the GPL version 2, or (at your option) any later version:
-// http://www.gnu.org/licenses/gpl-2.0.txt
+//
+// This file is part of MusicBrainz, the open internet music database,
+// and is licensed under the GPL version 2, or (at your option) any
+// later version: http://www.gnu.org/licenses/gpl-2.0.txt
 
 const Immutable = require('immutable');
 const $ = require('jquery');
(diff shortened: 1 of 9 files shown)


yvanzo:
Is `flow` latest version `0.60.1` supposed to parse 30K+ files?
Its server repeatedly crashes with a 300M+ sized core dump.

root/static/scripts/common/components/TagEditor.js
```diff
@@ -360,20 +426,20 @@ function init_tag_editor(Component, mountPoint) {
...
-        combined.push(Tag(t));
+        combined.push(newTag(t));
       }
     });
 
     ReactDOM.render(
       <Component entity={entity} more={more}
                  initialState={{tags: sortedTags(Immutable.List(combined))}} />,
-      document.getElementById(mountPoint)
+      (document.getElementById(mountPoint): any)
```
mwiencek: `getElementById` can return `null`, but it shouldn't ever be `null` here, so I used `any` to silence the error.

root/static/scripts/common/components/TagEditor.js
```diff
@@ -10,9 +12,15 @@ const React = require('react');
 const ReactDOM = require('react-dom');
 
 const {l, lp} = require('../i18n');
+const MB = require('../MB');
 const request = require('../utility/request');
+const TagLink = require('./TagLink');
 
-var Tag = Immutable.Record({tag: '', count: 0, vote: 0});
+const newTag: RecordFactory<UserTagT> = Immutable.Record({
```
mwiencek: The docs for typing Records I used are here: https://github.com/facebook/immutable-js/blob/d23746f44ffc025d7bdcee7d7bb79c37d3dc3ea5/type-definitions/Immutable.d.ts#L2239

PR #575, opened 2017-12-05
MBS-9570: Fix encoding of email header fields

yvanzo:
## Fix [MBS-9570](https://tickets.metabrainz.org/browse/MBS-9570): Email header fields with non-ASCII characters are not correctly encoded
Encode email header fields `From` and `Subject` in quoted-printable if a non-ASCII character is detected.

### DIFF
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -35,6 +35,16 @@ has 'c' => (
 
 Readonly our $url_prefix => 'https://' . DBDefs->WEB_SERVER_USED_IN_EMAIL;
 
+sub _encode_header {
+    my $header = shift;
+
+    if ($header =~ /[^\x20-\x7E]/) {
+        return encode("MIME-Q", $header);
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Email.pm
```diff
@@ -35,17 +35,29 @@ has 'c' => (
 
 Readonly our $url_prefix => 'https://' . DBDefs->WEB_SERVER_USED_IN_EMAIL;
 
+sub _encode_header {
+    my $header = @_;
+
+    if ($header =~ /[^\x20-\x7E]/) {
+        return encode("MIME-Q", $header);
```
mwiencek: Would it be incorrect, or cause any problems, if we always performed this encoding?

yvanzo: Not really, it would just add a few characters `=?UTF-8?Q?…?=`, unnecessarily in most cases.
I also noticed a small side effect: space is replaced with underscore.

PR #577, opened 2017-12-06
Convert the commons image components to React

mwiencek:


lib/MusicBrainz/Server/Controller/Role/CommonsImage.pm
```diff
@@ -13,16 +13,22 @@ Readonly my $WIKIDATA_PROP_LOCATOR_MAP_IMAGE => "P242";
 after load => sub {
     my ($self, $c) = @_;
 
-    $self->_get_commons_image($c, 1);
+    my $entity = $c->stash->{entity};
+    my $endpoint = $c->uri_for_action($self->action_for('commons_image'), [$entity->gid]);
+    $c->stash->{image} = $self->_get_commons_image($c, 1);
```
mwiencek: This runs on every page load. The `1` passed as the second argument to `_get_commons_image` indicates that it should only try loading the image info from the cache. If it's not cached, it's loaded asynchronously after the page loads. (This is not new behavior.)

lib/MusicBrainz/Server/Controller/Role/CommonsImage.pm
```diff
@@ -13,16 +13,22 @@ Readonly my $WIKIDATA_PROP_LOCATOR_MAP_IMAGE => "P242";
...
 sub commons_image : Chained('load') PathPart('commons-image') {
     my ($self, $c) = @_;
 
-    $self->_get_commons_image($c, 0);
+    my $image = $self->_get_commons_image($c, 0);
 
     $c->res->headers->header('X-Robots-Tag' => 'noindex');
-    $c->stash->{template} = 'components/commons-image.tt';
+    $c->res->content_type('application/json; charset=utf-8');
+    $c->res->{body} = $c->json->encode({image => $image});
```
mwiencek: This is the endpoint used to fetch the image info asynchronously. Previously it returned fully rendered HTML. Now it returns JSON to be used as props for the `CommonsImage` component, which is rendered client-side by React in that case.

root/components/common-macros.tt
```diff
@@ -895,18 +895,13 @@ END ~%]
...
-        <script type="text/javascript">
-            $.get('[% image_url %]',
-                  function (data) {
-                      $('#image-insertion-point').replaceWith(data);
-                  },
-                  'html');
-        </script>
-    [%- END -%]
+    <div id="commons-image">
+        [%~ React.embed(c, 'root/components/CommonsImage', { image => image }) ~%]
```
mwiencek: If `image` was loaded from the cache and is defined here, this will render completely on the server. We don't embed the `commons-image.js` bundle in that case either, due to the `UNLESS image` below.

root/static/gulpfile.js
```diff
@@ -278,6 +278,10 @@ runYarb('work.js', function (b) {
   b.external(editBundle).external(guessCaseBundle);
 });
 
+runYarb('commons-image.js', function (b) {
+  b.external(commonBundle);
+});
```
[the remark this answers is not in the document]
mwiencek: Thanks, I increased it to 16 in 69a739f1901b98602bf94e32bdd3952b2a024da0 so we don't have to update it as often, if that sounds okay.

PR #578, opened 2017-12-07
Convert sidebar-rating.tt to React

mwiencek:
The `RatingStars` component can be used for the ratings in the entity lists too, but I think there are too many of those to use `React.embed` for each row, so I'll wait to convert those lists to React first.


lib/MusicBrainz/Server/Entity/Role/Rating.pm
```diff
@@ -16,6 +16,18 @@ has 'rating_count' => (
     isa => 'Int'
 );
 
+around TO_JSON => sub {
+    my ($orig, $self) = @_;
+
+    my $json = $self->$orig;
+
+    $json->{rating} = $self->rating;
+    $json->{rating_count} = $self->rating_count;
```
yvanzo: I would expect it to default to `0`.

mwiencek: Fixed; also fixed the Flow types to account for `rating` or `user_rating` possibly being null.

PR #581, opened 2017-12-12
MBS-9578: Hide private tabs from editor profile

yvanzo:
### Fix [MBS-9578](https://tickets.metabrainz.org/browse/MBS-9578): Editor private subscriptions/tags/ratings pages should not be linked to

Subscriptions, tags, and ratings are tabs in editor profile. Accessing their content may be forbidden by privacy preferences.

This patch hides private tabs from editor profile, rather than linking to a page that returns a 403 Forbidden Request error.

### DIFF
--- a/root/user/profile/layout.tt
+++ b/root/user/profile/layout.tt
@@ -9,15 +9,30 @@
 
 [%- info_links = [
     ['index', simple_link(c.uri_for_action("/user/profile", [ user.name ]), l('Profile'))],
-    ['subscriptions', simple_link(c.uri_for_action("/user/subscriptions/artist", [ user.name ]), l('Subscriptions'))],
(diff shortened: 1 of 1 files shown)


root/user/profile/layout.tt
```diff
@@ -9,15 +9,30 @@
 
 [%- info_links = [
     ['index', simple_link(c.uri_for_action("/user/profile", [ user.name ]), l('Profile'))],
+] -%]
+
+[%- IF c.user.is_account_admin || viewing_own_profile || user.preferences.public_subscriptions -%]
+[%- info_links = [
```
mwiencek: Shouldn't this (and below) be using `info_links.push` so that they don't overwrite the previous items?

yvanzo: Oops! Fixed in https://github.com/metabrainz/musicbrainz-server/pull/581/commits/2a608a2afd7fe42fa82b5f099607a95950e87e77

PR #583, opened 2017-12-20
MBS-8576: Show work type descriptions in work creation form

naiveaiguy:
Show the description of the selected type next to the type dropdown. Somewhat inelegant solution with a bunch of preloaded `<span>`s, but works well.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Work.pm
+++ b/lib/MusicBrainz/Server/Controller/Work.pm
@@ -186,6 +186,12 @@ after create => sub {
     stash_work_form_json($c);
 };
 
+before qw( create edit ) => sub {
+    my ($self, $c) = @_;
+    my %work_types = map {$_->id => $_} $c->model('WorkType')->get_all();
+    $c->stash->{work_types} = \%work_types;
+};
+
 1;
 
 =head1 COPYRIGHT
--- a/lib/MusicBrainz/Server/Data/WorkType.pm
+++ b/lib/MusicBrainz/Server/Data/WorkType.pm
@@ -38,6 +38,14 @@ sub in_use {
         $id);
 }
 
+sub get_description
+{
+    my ($self, $id) = @_;
+    return $self->sql->select_single_value(
+        'SELECT description FROM work_type WHERE id = ? LIMIT 1',
+        $id);
+}
+
 __PACKAGE__->meta->make_immutable;
 no Moose;
 1;
--- a/lib/MusicBrainz/Server/Data/WorkType.pm
+++ b/lib/MusicBrainz/Server/Data/WorkType.pm
@@ -38,14 +38,6 @@ sub in_use {
         $id);
 }
 
-sub get_description
-{
-    my ($self, $id) = @_;
-    return $self->sql->select_single_value(
-        'SELECT description FROM work_type WHERE id = ? LIMIT 1',
-        $id);
-}
-
 __PACKAGE__->meta->make_immutable;
 no Moose;
 1;
--- a/root/static/scripts/work.js
+++ b/root/static/scripts/work.js
@@ -189,3 +189,16 @@ store.subscribe(renderWorkLanguages);
 renderWorkLanguages();
 
 MB.Control.initializeBubble('#iswcs-bubble', 'input[name=edit-work\\.iswcs\\.0]');
+
+let typeIdField = 'select[name=edit-work\\.type_id]';
+MB.Control.initializeBubble('#type-bubble', typeIdField);
+$(typeIdField).on('change', function() {
+  if (!this.value.match(/\S/g)) {
+    $('.type-bubble-description').hide();
+    $('#type-bubble-default').show();
+  } else {
+    $('#type-bubble-default').hide();
+    $('.type-bubble-description').hide();
+    $(`#type-bubble-description-${this.value}`).show();
+  }
+});
--- a/root/work/edit_form.tt
+++ b/root/work/edit_form.tt
@@ -99,6 +99,18 @@
 
   <div class="documentation">
     [%- iswc_bubble(link_entity(work)) -%]
+
+    <div class="bubble" id="type-bubble">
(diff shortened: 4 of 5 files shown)


yvanzo:
@brainzbot, test this please.

root/static/scripts/work.js
```diff
@@ -189,3 +189,19 @@ store.subscribe(renderWorkLanguages);
 renderWorkLanguages();
 
 MB.Control.initializeBubble('#iswcs-bubble', 'input[name=edit-work\\.iswcs\\.0]');
+
+let $typeBubble = $('#type-bubble')[0];
```
yvanzo: This variable is unused. Please remove it.

root/static/scripts/work.js
```diff
@@ -189,3 +189,19 @@ store.subscribe(renderWorkLanguages);
...
+let typeIdField = 'select[name=edit-work\\.type_id]';
```
yvanzo: This variable is used only once. Please move its value directly to the function call two lines below.

root/static/scripts/work.js
```diff
@@ -189,3 +189,19 @@ store.subscribe(renderWorkLanguages);
...
+
+$('.type-bubble-description').hide();
```
yvanzo: It seems to be more simple to use `style="display:none"` in the template instead.

root/work/edit_form.tt
```diff
@@ -99,6 +99,18 @@
 
   <div class="documentation">
     [%- iswc_bubble(link_entity(work)) -%]
+
+    <div class="bubble" id="type-bubble">
+      <span id="type-bubble-default">
+        [%- l('Select any type from the list to see its description. If the work
+               doesn\'t seem to match any type, just leave this blank.') -%]
```
yvanzo: Should be `doesn’t` as [typographically-correct punctuation is preferred](https://musicbrainz.org/doc/Style/Miscellaneous).

root/work/edit_form.tt
```diff
@@ -99,6 +99,18 @@
...
+      </span>
+      [% FOREACH id IN form.field('type_id').options %]
+         <span class="type-bubble-description" id="type-bubble-description-[%- id.value -%]">
+           <strong>Description: </strong>[%- c.model('WorkType').get_description(id.value) -%]
```
yvanzo: I am not sure the label `Description:` is really necessary. If it is, it should use the macro `add_colon` for i18n.

root/static/scripts/work.js
naiveaiguy: Actually, it's used again after that to bind to its `change` event

root/work/edit_form.tt
naiveaiguy: Personally, I think it makes it clear for someone who didn't fully read the initial message what the bubble is.

naiveaiguy:
Amended commit with changes.

naiveaiguy:
:shipit:

lib/MusicBrainz/Server/Data/WorkType.pm
```diff
@@ -38,6 +38,14 @@ sub in_use {
         $id);
 }
 
+sub get_description
```
mwiencek: You should delete this, then use `$c->model('WorkType')->get_all` in `Controller::Work` to create a hash of work type descriptions by ID. Then place that hash in `$c->stash` to access it from the template.

root/work/edit_form.tt
```diff
@@ -99,6 +99,18 @@
...
     [%- iswc_bubble(link_entity(work)) -%]
+
+    <div class="bubble" id="type-bubble">
+      <span id="type-bubble-default">
+        [%- l('Select any type from the list to see its description. If the work
+               doesn’t seem to match any type, just leave this blank.') -%]
+      </span>
+      [% FOREACH id IN form.field('type_id').options %]
+         <span style="display:none" class="type-bubble-description" id="type-bubble-description-[%- id.value -%]">
+           <strong>[%- add_colon(l('Description')) -%] </strong>[%- c.model('WorkType').get_description(id.value) -%]
```
mwiencek: Calling `get_description` inside a loop in the template isn't the best idea (it will cause N additional database queries per page load, none of which are cached, where N = the number of work types). Calling `get_all` once in the controller as described above will fix that.

There's another problem here which is that the descriptions are not marked for translation. You'll need to create an `l_description` method in `Entity::WorkType`, similar to here: https://github.com/metabrainz/musicbrainz-server/blob/2dc039d04f49166edf20dcdd29ffe9d983cda26d/lib/MusicBrainz/Server/Entity/EventType.pm#L17-L20

lib/MusicBrainz/Server/Data/WorkType.pm
```diff
@@ -4,6 +4,7 @@ use Moose;
 use namespace::autoclean;
 use MusicBrainz::Server::Entity::WorkType;
 use MusicBrainz::Server::Data::Utils qw( load_subobjects );
+use MusicBrainz::Server::Translation::Attributes qw( lp );
```
mwiencek: That can be removed now too

lib/MusicBrainz/Server/Controller/Work.pm
```diff
@@ -186,6 +186,12 @@ after create => sub {
     stash_work_form_json($c);
 };
 
+before qw( create edit ) => sub {
+    my ($self, $c) = @_;
+    my %work_types = map {$_->id => $_} $c->model('WorkType')->get_all();
+    $c->stash->{work_types} = \%work_types
```
mwiencek: Nitpick: add a semicolon at the end of the line? :)

naiveaiguy: Darn.

naiveaiguy:
*Now* :shipit:.

naiveaiguy:
OFFICIALLY IN THE BETA! I am so happy about this for some reason!!

PR #584, opened 2017-12-21
MBS-8206: Add Disc ID report

naiveaiguy:
This report shows all the releases that have at least one medium that has a format with the `has_discids` property, but that don't actually have a Disc ID attached to it.

### DIFF
--- a/lib/MusicBrainz/Server/Report/ReleasesMissingDiscIds.pm
+++ b/lib/MusicBrainz/Server/Report/ReleasesMissingDiscIds.pm
@@ -1,0 +1,36 @@
+package MusicBrainz::Server::Report::ReleasesMissingDiscIds;
+use Moose;
+
+with 'MusicBrainz::Server::Report::ReleaseReport',
+     'MusicBrainz::Server::Report::FilterForEditor::ReleaseID';
+
+sub query {
+    q{
+        SELECT DISTINCT ON (r.id)
+            r.id AS release_id,
+            row_number() OVER (ORDER BY r.artist_credit, r.name)
+        FROM
+        ( SELECT r.id, r.artist_credit, r.name
+          FROM release r
+          WHERE id IN (
+              SELECT release FROM medium
+              WHERE (SELECT has_discids FROM medium_format WHERE medium_format.id = medium.format)
+              AND id NOT IN (SELECT medium FROM medium_cdtoc)
+          )
+        ) r
+        WHERE r.status NOT IN (3, 4) -- ignore psuedo and bootleg releases
+    }
+}
+
+__PACKAGE__->meta->make_immutable;
+1;
+
+=head1 COPYRIGHT
+
+Copyright (C) 2017 MetaBrainz Foundation
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
--- a/root/report/index.tt
+++ b/root/report/index.tt
@@ -108,6 +108,7 @@
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithNoMediums') %]">[% l('Releases with no mediums') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVACredit') %]">[% l('Releases not credited to "Various Artists" but linked to VA') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVALink') %]">[% l('Releases credited to "Various Artists" but not linked to VA') %]</a></li>
+            <li><a href="[% c.uri_for_action('/report/show', 'ReleasesMissingDiscIds') %]">[% l('Releases missing Disc IDs') %]</a></li>
         </ul>
 
         <h2>[% l('Recordings') %]</h2>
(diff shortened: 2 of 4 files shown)


naiveaiguy:
I've just noticed that [MBS-8206](https://tickets.metabrainz.org/browse/MBS-8206) specifies that we should ignore psedo and bootleg releases, but the GCI task I am coming from specifies to only list pseudo-releases. The latter feels like a mistake to me, so I'm going to amend the commit using the former specification.

lib/MusicBrainz/Server/Report/ReleasesWithoutDiscIds.pm
```diff
@@ -0,0 +1,45 @@
...
+    q{
+        SELECT DISTINCT ON (r.id)
+            r.id AS release_id,
+            row_number() OVER (ORDER BY r.artist_credit, r.name)
+        FROM
+        ( SELECT r.id, r.artist_credit, r.name
+          FROM release r
+          WHERE id IN (
+              SELECT release FROM medium
+              WHERE (SELECT has_discids FROM medium_format WHERE id = format)
```
yvanzo: I would use `medium_format.id = medium.format` to make it less ambiguous.

lib/MusicBrainz/Server/Report/ReleasesWithoutDiscIds.pm
```diff
@@ -0,0 +1,45 @@
...
+              AND status NOT IN (3, 4) -- ignore psuedo and bootleg releases
+          )
+        ) r
+    }
+}
+
+__PACKAGE__->meta->make_immutable;
+1;
+
+=head1 COPYRIGHT
```
yvanzo: Please use the lighter and up-to-date section [`COPYRIGHT AND LICENSE`](https://musicbrainz.org/doc/Development/Server_Coding_Style#Perl_Module_Template_and_Documentation_Guidelines) instead.

root/report/index.tt
```diff
@@ -108,6 +108,7 @@
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithNoMediums') %]">[% l('Releases with no mediums') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVACredit') %]">[% l('Releases not credited to "Various Artists" but linked to VA') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVALink') %]">[% l('Releases credited to "Various Artists" but not linked to VA') %]</a></li>
+            <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutDiscIds') %]">[% l('Releases without Disc IDs') %]</a></li>
```
yvanzo: The URL is fine, but the title is a bit confusing. It is legitimate to have (e.g. Vinyl) releases without Disc IDs. The title should explain the weirdness of reported releases. Replacing _without_ by _missing_ will probably do.

root/report/releases_without_disc_ids.tt
```diff
@@ -0,0 +1,16 @@
+[%- WRAPPER 'layout.tt' title=l('Releases Without Disc IDs') full_width=1 -%]
+
+<h1>[% l('Releases Without Disc IDs') %]</h1>
+
+<ul>
+  <li>[% l('This report shows releases (official and promotional only) that have at least
+            one medium without a Disc ID. For instructions on how to add one, see the {add_discids|documentation page}.',
```
yvanzo: The same remark as for the title. Maybe a bit more detailed explanation would clarify it.

naiveaiguy:
Am here. On it.

naiveaiguy:
Amended commit with changes. Note, when I added the line to ignore pseudo and bootleg releases earlier, I made the grave mistake of not testing it. Now, when I was, I saw that the query takes absolutely **forever** with that line. For now, I have removed it.

lib/MusicBrainz/Server/Report/ReleasesMissingDiscIds.pm
```diff
@@ -0,0 +1,35 @@
...
+        ) r
+    }
+}
+
+__PACKAGE__->meta->make_immutable;
+1;
+
+=head1 COPYRIGHT
+
+Copyright (C) 2017 MetaBrainz Foundation
```
Freso: You should use your own name here. MetaBrainz is not paying for your work, so you retain your copyright. :)

naiveaiguy: Well I'd like to think I'm part of the team 😸

yvanzo:
Sorry, I just noticed you removed the status filter. It should not impair performances that much.  Here is a comparison: https://gist.github.com/yvanzo/53c381842775f6cd053a68e42efd8a73 Can you put it back?

naiveaiguy:
@yvanzo Amended. Problem with my test before was that I added `AND status NOT IN (3, 4)` in the *subquery* rather than after the `FROM` block, where it works equally well.

lib/MusicBrainz/Server/Report/ReleasesMissingDiscIds.pm
yvanzo: Actually, I did the same as a free contributor in the past. Giving the copyright to the MetaBrainz Foundation is mostly about [_procedural advantages_ [...] _under US copyright law_](https://www.gnu.org/licenses/why-assign.html). Copyright in your own name would be accepted too.

naiveaiguy: Alright. Will do in #588 as well.

PR #587, opened 2017-12-24
Renderer socket fixes

mwiencek:
This seems to fix cases where the renderer socket file doesn't get removed after server.js exits. I then removed the explicit `unlink` calls I added in places, which were causing their own problems (sometimes the socket file would get removed, so the unlink would fail).

No idea if it fixes issues like [MBS-9370](https://tickets.metabrainz.org/browse/MBS-9370).

### DIFF
--- a/app.psgi
+++ b/app.psgi
@@ -43,7 +43,6 @@ BEGIN {
             use POSIX;
             my $action = POSIX::SigAction->new(sub {
                 kill 'TERM', $child;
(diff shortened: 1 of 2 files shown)


yvanzo:
Yay! I have been able to reproduce the issue and this patch fixes it, at least in my case.

Note that if a socket file has not been closed (by an instance of a previous server release), it still has to be removed by hand.

PR #588, opened 2017-12-25
MBS-9425: Add 'recordings with possibly wrong artists' report

naiveaiguy:
Hope I haven't made as many stupid mistakes here 😅

Merry Christmas everybody ☃️ and have a happy new year!


root/report/recordings_possibly_assigned_to_the_wrong_artist.tt
```diff
@@ -0,0 +1,17 @@
+[%- WRAPPER 'layout.tt' title=l('Recordings possibly assigned to the wrong artist(s)') full_width=1 -%]
```
yvanzo: Although the title is _OK_ for Rochus, I still think it is not specific enough. There are many reasons for recordings to be _possibly assigned to the wrong artist_. The previous ticket title was more specific. Was it unclear or unreadable? Input from a native English speaker such as @mwiencek would be welcome.

naiveaiguy: There are many reasons, you're right, which is precisely the reason I feel we should use this title. In the description I clarified exactly what it means to be 'possibly wrong', but we could change the way this works and add more conditions to evaluate it better and the title would still hold. However, a precise title also makes sense. @mwiencek is the tie breaker here then.

lib/MusicBrainz/Server/Report/RecordingsPossiblyAssignedToTheWrongArtist.pm
```diff
@@ -0,0 +1,39 @@
...
+
+sub query {
+    # Find all recordings with the same name that have artists with the same name.
+    "
+    SELECT
+        recording_id,
+        row_number() OVER (ORDER BY musicbrainz_collate(rname), recording_id)
+    FROM (
+        SELECT
+            DISTINCT r1.id AS recording_id, r1.NAME AS rname, a1.NAME AS aname
```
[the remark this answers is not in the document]
naiveaiguy: Amended.

root/report/recordings_possibly_assigned_to_the_wrong_artist.tt
mwiencek: I suppose I prefer the precise title ("Recordings with the same name by different artists with the same name").

And actually, I'm not sure it's a good idea to make this report any more complicated (e.g. combining the track AC idea into here) because it would make the report harder to step through for volunteers. They'd have to figure out which of the two cases the problem is each time and switch workflows while fixing them. Having a separate report just makes much more sense to me.

naiveaiguy: Done.

lib/MusicBrainz/Server/Report/RecordingsWithTheSameNameByDifferentArtistsWithTheSameName
```diff
@@ -0,0 +1,37 @@
...
+        SELECT
+            DISTINCT r1.id AS recording_id, r1.NAME AS rname, a1.NAME AS aname
+        FROM
+            recording r1
+            INNER JOIN recording r2 ON (r1.NAME = r2.NAME AND r1.id != r2.id)
+            INNER JOIN artist_credit_name acn1 ON r1.artist_credit = acn1.artist_credit
+            INNER JOIN artist_credit_name acn2 ON r2.artist_credit = acn2.artist_credit
+            INNER JOIN artist a1 ON acn1.artist = a1.id
+            INNER JOIN artist a2 ON (acn2.artist = a2.id AND a1.NAME = a2.NAME AND a1.id != a2.id)
+    ) r
```
mwiencek: This query only works for artist credits with 1 artist, which I think is fine because it'd be really complicated to make it work for more, but I think it's better to be explicit about that in the query rather than it being an accident:

```SQL
SELECT
    recording_id,
    row_number() OVER (ORDER BY musicbrainz_collate(rname), recording_id)
FROM (
    SELECT
        DISTINCT r1.id AS recording_id, r1.name AS rname, a1.name AS aname
    FROM
        recording r1
        JOIN recording r2 ON (r1.name = r2.name AND r1.id != r2.id)
        JOIN artist_credit ac1 ON (r1.artist_credit = ac1.id AND ac1.artist_count = 1)
        JOIN artist_credit ac2 ON (r2.artist_credit = ac2.id AND ac2.artist_count = 1)
        JOIN artist_credit_name acn1 ON ac1.id = acn1.artist_credit
        JOIN artist_credit_name acn2 ON ac2.id = acn2.artist_credit
        JOIN artist a1 ON acn1.artist = a1.id
        JOIN artist a2 ON (acn2.artist = a2.id AND a1.id != a2.id)
        WHERE (acn1.name = acn2.name OR a1.name = a2.name)
) r
```

Note, I made some style changes and changed the condition: `(acn1.name = acn2.name OR a1.name = a2.name)` will catch either case where the artist names are the same, or just the credited names.

lib/MusicBrainz/Server/Report/RecordingsWithTheSameNameByDifferentArtistsWithTheSameName
```diff
@@ -0,0 +1,37 @@
+package MusicBrainz::Server::Report::RecordingsWithTheSameNameByDifferentArtistsWithTheSameName
```
mwiencek: This needs to be shorted (here and elsewhere) like how we do with the `ArtistsDisambiguationSameName` report, to something like `RecordingsSameNameDifferentArtistsSameName`, or shorter if @yvanzo can think of something.

root/report/recordings_with_the_same_name_by_different_artists_with_the_same_name.tt
```diff
@@ -0,0 +1,17 @@
+[%- WRAPPER 'layout.tt' title=l('Recordings with the same name by different artists with the same name') full_width=1 -%]
+
+<h1>[% l('Recordigns with the sae name by different artists with the same name') %]</h1>
```
mwiencek: s/Recordigns/Recordings/
s/sae/same/

root/report/recordings_with_the_same_name_by_different_artists_with_the_same_name.tt
```diff
@@ -0,0 +1,17 @@
...
+
+<ul>
+    <li>[% l('This report shows all recordings with the same name that have different artist MBIDs with the same name.') %]</li>
+    <li>[% l('These are most likely cases where the artist is incorrect for at least one
+              one of the instances. Make sure to check that the {artist_credit|Artist Credit} is correct as well.',
```
mwiencek: I would say: "These are most likely cases where the {ac|artist credit} is incorrect for at least one of the recordings." (Note: no line breaks.)

root/report/recordings_with_the_same_name_by_different_artists_with_the_same_name.tt
```diff
@@ -0,0 +1,17 @@
+[%- WRAPPER 'layout.tt' title=l('Recordings with the same name by different artists with the same name') full_width=1 -%]
+
+<h1>[% l('Recordigns with the sae name by different artists with the same name') %]</h1>
+
+<ul>
+    <li>[% l('This report shows all recordings with the same name that have different artist MBIDs with the same name.') %]</li>
```
mwiencek: I would say: "This report lists recordings with the same name that are credited to different artists with the same name."

root/report/recordings_with_the_same_name_by_different_artists_with_the_same_name.tt
```diff
@@ -0,0 +1,17 @@
...
+    <li>[% l('These are most likely cases where the artist is incorrect for at least one
+              one of the instances. Make sure to check that the {artist_credit|Artist Credit} is correct as well.',
+              { artist_credit => doc_link('Artist_Credits') }) %]</li>
+    <li>[% l('Total releases found: {count}', { count => pager.total_entries }) %]</li>
```
mwiencek: s/releases/recordings/

mwiencek:
To make the report useful, I think we should perhaps display the artist MBID next to the artist in parentheses.

lib/MusicBrainz/Server/Report/RecordingsWithTheSameNameByDifferentArtistsWithTheSameName
yvanzo: Doesn’t the original query work correctly for any number of artists?

I thought it correctly reported the two following examples (where `'` stands for _distinct homonym_):

1. Recording Foo credited to X and Y
2. Recording Foo' credited to Y and X'

and

3. Recording Bar credited to X feat. Y
4. Recording Bar' credited to X'

naiveaiguy: It doesn't appear to, but it's all very confusing. I'm attempting to get some clear test data up so I can do this methodically.

naiveaiguy:
Will open a separate PR for multiple artists fixing.

root/report/recordings_same_name_different_artists_same_name.tt
```diff
@@ -0,0 +1,19 @@
+[%- WRAPPER 'layout.tt' title=l('Recordings with the same name by different artists with the same name') full_width=1 -%]
+
+<h1>[% l('Recordings with the same name by different artists with the same name') %]</h1>
+
+<ul>
+    <li>[% l('This report shows all recordings with the same name that have different artist MBIDs with the same name.') %]</li>
```
mwiencek: Grammar nitpick: "different artist MBIDs with the same name" means the names of the MBIDs themselves, which doesn't make much sense. I would put the explanation in parentheses like "different artists (having different MBIDs) with the same name."

root/report/recordings_same_name_different_artists_same_name.tt
```diff
@@ -0,0 +1,19 @@
...
+    <li>[% l('Currently, this only works with recordings that have one artist.') %]</li>
```
mwiencek: Nitpick: I would add slightly more context to this string by doing s/this/this report/. Also, move this below the `These are most likely ...` item because it's less important than that.

root/report/recordings_same_name_different_artists_same_name.tt
```diff
@@ -0,0 +1,19 @@
+[%- WRAPPER 'layout.tt' title=l('Recordings with the same name by different artists with the same name') full_width=1 -%]
+
+<h1>[% l('Recordings with the same name by different artists with the same name') %]</h1>
+
+<ul>
+    <li>[% l('This report shows all recordings with the same name that have different artist (having different MBIDs) with the same name.') %]</li>
```
mwiencek: s/artist/artists/

lib/MusicBrainz/Server/Report/RecordingsSameNameDifferentArtistsSameName.pm
```diff
@@ -0,0 +1,41 @@
...
+use Moose;
+
+with 'MusicBrainz::Server::Report::RecordingReport',
+     'MusicBrainz::Server::Report::FilterForEditor::RecordingID';
+
+sub query {
+    "
+    SELECT
+        recording_id,
+        row_number() OVER (ORDER BY musicbrainz_collate(rname), recording_id)
```
mwiencek: It would be much better for the report to be ordered by recording name, then artist *id*. Right now, while the MBIDs make it easy to spot which artists are probably wrong, the wrong MBIDs are not sorted next to each other at all.

PR #592, opened 2018-01-05
Migrate to Solr search

samj1912:


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -402,12 +402,15 @@ sub schema_fixup
                 my $area = delete $data->{$json_prop};
                 $area->{gid} = $area->{id};
                 $area->{id} = 1;
+                for my $area_prop (qw( type type-id )) {
+                    delete $area->{$area_prop};
```
mwiencek: Unless there's a reason the old search didn't output these, why not use `schema_fixup_type` instead of deleting them?

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -402,12 +402,15 @@ sub schema_fixup
...
+                }
                 $data->{$prop} = MusicBrainz::Server::Entity::Area->new($area);
             }
         }
     }
     if ($type eq 'artist' && defined $data->{gender}) {
-        $data->{gender} = MusicBrainz::Server::Entity::Gender->new( name => ucfirst($data->{gender}) );
+        $data->{gender} = MusicBrainz::Server::Entity::Gender->new( name => ucfirst($data->{gender}->{name}) );
```
mwiencek: I'm not sure I understand all these `->{name}` additions because it suggests to me the output is incompatible, and the search server should be fixed instead.

lib/DBDefs/Default.pm
```diff
@@ -103,6 +103,7 @@ sub WEB_SERVER                { "localhost:5000" }
 # Relevant only if SSL redirects are enabled
 sub WEB_SERVER_SSL            { "localhost" }
 sub LUCENE_SERVER             { "search.musicbrainz.org" }
+sub SOLR_SERVER               { "solr.musicbrainz.org/solr" }
```
zas: What will happen to x-accel-redirect with this change ?
See https://github.com/metabrainz/musicbrainz-server/blob/4ecff4126780c938008b5ff767d7166071d8a1f8/lib/MusicBrainz/Server/Data/WebService.pm#L211

samj1912:
Fixed the schema changes in https://github.com/metabrainz/mb-solr/pull/15

lib/MusicBrainz/Server/Data/Search.pm
samj1912: The old search server didn't output them. Although I did end up using schema_fixup_type anyways.

samj1912:
Did a force push to remove a the commits, dont forget to use reset --hard while reviewing.

lib/DBDefs/Default.pm
samj1912: done

lib/DBDefs/Default.pm
```diff
@@ -103,6 +103,7 @@ sub WEB_SERVER                { "localhost:5000" }
 # Relevant only if SSL redirects are enabled
 sub WEB_SERVER_SSL            { "localhost" }
 sub LUCENE_SERVER             { "search.musicbrainz.org" }
```
mwiencek: Can you remove this and grep for other locations where it's used and removed them too?

lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -202,16 +202,16 @@ sub xml_search
         };
     }
 
-    my $format = ($args->{fmt} // "") eq "json" ? "jsonnew" : "xml";
-
-    my $url_ext = "/ws/2/$resource/?" .
-        "max=$limit&type=$resource&fmt=$format&offset=$offset" .
-        "&query=" . uri_escape_utf8($query) . "&dismax=$dismax";
+    my $format = ($args->{fmt} // "") eq "json" ? "mbjson" : "mbxml";
+    my $defType = $dismax ? 'lucene' : 'dismax';
```
mwiencek: `$defType` is set to `'dismax'` if `$dismax` is false?

samj1912: fixed.

lib/DBDefs/Default.pm
samj1912: done

lib/MusicBrainz/Server/Data/WebService.pm
samj1912: fixed.

lib/DBDefs.pm.sample
```diff
@@ -153,7 +153,7 @@ sub DB_SCHEMA_SEQUENCE { 24 }
 sub WEB_SERVER                { "www.musicbrainz.example.com" }
 # Relevant only if SSL redirects are enabled
 # sub WEB_SERVER_SSL            { "localhost" }
-# sub LUCENE_SERVER             { "search.musicbrainz.org" }
+# sub SOLR_SERVR                { "solr.musicbrainz.org/solr" }
```
mwiencek: s/SOLR_SERVR/SOLR_SERVER/

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -422,9 +422,8 @@ sub find_recent_by_artists
 {
     my ($self, $artist_ids) = @_;
 
-    my $search_url = sprintf("http://%s/ws/2/recording/?query=title:\"\"",
-                              DBDefs->LUCENE_SERVER);
-
+    my $search_url = sprintf("http://%s/recording/select?q=recording:*&wt=mbxml",
+                                 DBDefs->SOLR_SERVER,);
```
mwiencek: Not that perl cares, but you added a trailing comma there which looks weird.

lib/DBDefs.pm.sample
samj1912: fixed.

lib/MusicBrainz/Server/Data/Recording.pm
samj1912: old habit from python dicts. easier to add to it. Will fix.

mwiencek: It would definitely make sense if `);` were on its own line, since then additions or removals at the end would only be a one-line change instead of two. Sorry, I was just being picky.

lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -202,16 +202,16 @@ sub xml_search
...
+    my $url_ext = "/$resource/select?" .
+        "rows=$limit&wt=$format&start=$offset" .
+        "&q=" . uri_escape_utf8($query) . "&defType=$defType";
 
     if (DBDefs->LUCENE_X_ACCEL_REDIRECT) {
-        return { redirect_url => '/internal/search/' . DBDefs->LUCENE_SERVER . $url_ext }
+        return { redirect_url => '/internal/search/' . DBDefs->SOLR_SERVER . $url_ext }
```
zas: What about using `SEARCH_SERVER`, `SEARCH_X_ACCEL_REDIRECT`, `SEARCH_ENGINE` and create wrappers so we can switch between search engines easily ?

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -422,9 +422,16 @@ sub find_recent_by_artists
 {
     my ($self, $artist_ids) = @_;
 
-    my $search_url = sprintf("http://%s/ws/2/recording/?query=title:\"\"",
-                              DBDefs->LUCENE_SERVER);
-
+    if (DBDefs->SEARCH_ENGINE == 'LUCENE')
```
mwiencek: s/==/eq/

In perl, numeric comparisons use `==` and string comparisons use `eq`.

This will always evaluate to true because it coerces both strings to `0` and then compares them numerically.

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -422,9 +422,16 @@ sub find_recent_by_artists
...
+    {
```
mwiencek: Can you fix the brace style to follow the majority of our codebase? :)

```
if (foo) {
    bar;
} else {
    baz;
}
```

samj1912: Sure, didn't know which one to use since I saw usages of both in Search.pm

docker/templates/DBDefs.pm.ctmpl
```diff
@@ -188,7 +188,7 @@ sub GIT_SHA { {{executeTemplate "QUOTED_STRING" (env "GIT_SHA")}} }
 {{executeTemplate "BOOLEAN_DEF" "DEVELOPMENT_SERVER"}}
 {{executeTemplate "BOOLEAN_DEF" "FORK_RENDERER"}}
 {{executeTemplate "BOOLEAN_DEF" "IS_BETA"}}
-{{executeTemplate "BOOLEAN_DEF" "LUCENE_X_ACCEL_REDIRECT"}}
+{{executeTemplate "BOOLEAN_DEF" "SEARCH_X_ACCEL_REDIRECT"}}
```
zas: I think we need compatibility for old configs here: `LUCENE_X_ACCEL_REDIRECT` or `SEARCH_X_ACCEL_REDIRECT`
@mwiencek : ^^ ?

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -422,9 +422,13 @@ sub find_recent_by_artists
 {
     my ($self, $artist_ids) = @_;
 
-    my $search_url = sprintf("http://%s/ws/2/recording/?query=title:\"\"",
-                              DBDefs->LUCENE_SERVER);
-
+    if (DBDefs->SEARCH_ENGINE eq 'LUCENE') {
+        my $search_url = sprintf("http://%s/ws/2/recording/?query=title:\"\"",
```
zas: Shouldn't `$search_url` declared before the `if`/`else` since it is used outside the scope of each branch ?

```perl
$my search_url;
if (...) {
   $search_url = 'A....';
} else {
   $search_url = 'B....';
}

do_something($search_url);
```

docker/templates/DBDefs.pm.ctmpl
mwiencek: It might not be necessary, since the .ctmpl is only used by our production servers and we'll have to fix our consul config anyway to add `SEARCH_ENGINE` etc. But we'd have to make sure to update the config files correctly (leaving the old keys in place until after the new code is deployed).

lib/MusicBrainz/Server/Data/Recording.pm
mwiencek: Yes, unlike python perl has lexical scope, so `$search_url` needs to be declared outside the `if`.

samj1912: Ah, let me fix that.

samj1912:
Sorry for all the perl/python mistakes. Still learning perl.

zas:
> Still learning perl.

Me too. Good luck ;)

samj1912:

PR #593, opened 2018-01-05
MBS-9374: Display multiple lyrics languages in work search result

yvanzo:
## Fix [MBS-9374](https://tickets.metabrainz.org/browse/MBS-9374): Langcode not displayed when searching works

Since MBS-5452 has been implemented, lyrics language was not displayed anymore in work search result.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -666,9 +666,7 @@ sub schema_fixup
         if (@languages) {
             $data->{languages} = [map {
                 MusicBrainz::Server::Entity::WorkLanguage->new({
-                    language => MusicBrainz::Server::Entity::Language->new({
-                        iso_code_3 => $_,
-                    }),
+                    language => $self->c->model('Language')->find_by_code($_),
                 })
             } @languages];
         }
--- a/root/search/results-work.tt
+++ b/root/search/results-work.tt
@@ -10,7 +10,7 @@
                             <th>[%- l('Artists') -%]</th>
                             <th>[%- l('ISWC') -%]</th>
                             <th>[%- l('Type') -%]</th>
-                            <th>[%- l('Language') -%]</th>
+                            <th>[%- l('Languages') -%]</th>
                         </tr>
                     </thead>
                     <tbody>
@@ -45,9 +45,15 @@
                                 [% result.entity.type.l_name %]
                             </td>
                             <td>
-                              <abbr title="[% result.entity.language.l_name %]">
-                                [% result.entity.language.iso_code_3 %]
-                              </abbr>
+                              <ul>
+                              [%- FOR work_language=result.entity.languages %]
+                                <li>
+                                  <abbr title="[% work_language.language.l_name %]">
+                                    [% work_language.language.iso_code_3 %]
+                                  </abbr>
+                                </li>
+                              [%- END %]
+                              </ul>
                             </td>
                         </tr>
                         [%- END -%]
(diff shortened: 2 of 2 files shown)


yvanzo:
It works fine for direct database search, but language name (on abbreviation rollover) is still missing for indexed search. This might be an issue of the search server.

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -666,9 +666,7 @@ sub schema_fixup
         if (@languages) {
             $data->{languages} = [map {
                 MusicBrainz::Server::Entity::WorkLanguage->new({
-                    language => MusicBrainz::Server::Entity::Language->new({
-                        iso_code_3 => $_,
-                    }),
+                    language => $self->c->model('Language')->find_by_code($_),
```
mwiencek: What do you think about removing this entire section and calling `$self->c->model('Language')->load_for_works(@entities);` [in external_search](https://github.com/yvanzo/musicbrainz-server/blob/521dab3/lib/MusicBrainz/Server/Data/Search.pm#L823-L828) so that we're only performing one query per search?

A comment explaining why we're doing this (the search doesn't include the name) seems appropriate, since we normally don't load information from the DB here.

root/search/results-work.tt
```diff
@@ -10,7 +10,7 @@
                             <th>[%- l('Artists') -%]</th>
                             <th>[%- l('ISWC') -%]</th>
                             <th>[%- l('Type') -%]</th>
-                            <th>[%- l('Language') -%]</th>
+                            <th>[%- l('Languages') -%]</th>
```
mwiencek: I think we use "Lyrics Languages" as the label everywhere else.

lib/MusicBrainz/Server/Data/Search.pm
yvanzo: Although it performs more queries, most of these should be cached because of the small number of languages and their distribution for lyrics.

Removing this section doesn’t seem possible, as each lyrics `Language` still has to be linked to the related `Work` through `WorkLanguage`.

Calling `load_for_works` for indexed search opens to inconsistency between query and results display. It makes impossible to know which lyrics languages are currently indexed for a given work.

root/search/results-work.tt
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/593/commits/4e37e35ffeb3f3d792f4d5840a47c8172406280a (to be squashed on merge)

lib/MusicBrainz/Server/Data/Search.pm
mwiencek: I agree with your last point, but I don't believe `find_by_code` uses the cache currently. It calls `_get_by_key` which appears to always fetch from the database - I couldn't find any role that modifies its behavior to add caching either.

yvanzo: Right, I still don't see how to _remove this entire section_, but I can make `_get_by_key` use cache. Would that be satisfactory? Or should we make search server indexing languages?

mwiencek: Well, I believe `$self->c->model('Language')->load_for_works` does handle creating the `WorkLanguage` links, so I think you'd be able to remove the entire section in that case. But I agree with you that it creates inconsistencies. Changing `_get_by_key` to use the catch would be great!

mwiencek: Exposing the language names in the WS might be a good idea since I'm not sure how people are supposed to get that information otherwise, if they don't have a DB. But it's a more involved change for sure, so just making the `find_by_code` queries cached sounds good to me.

mwiencek:
I'm fine with merging this now and doing the cache changes later if needed. The `find_by_code` lookups should be very fast. Sorry for holding things up.

PR #595, opened 2018-01-07
MBS-9590: Rewrite the instrument list in React/JSX

spellew:
Replaced the current template root/instrument/list.tt with an equivalent template written in React/JSX. Accepted on Google Code-In by yvanzo.

https://tickets.metabrainz.org/browse/MBS-9590


mwiencek:
@brainzbot add to whitelist

root/instrument/List.js
```diff
@@ -0,0 +1,47 @@
...
+const EntityLink = require('../static/scripts/common/components/EntityLink');
+const {l, lp} = require('../static/scripts/common/i18n');
+
+const instrumentDescription = (description) => ({ __html: " &mdash; " + description });
+const Instrument = (props) => {
+  const instrument = props.instrument;
+  return (
+    <li>
+      <EntityLink entity={instrument} />
+      { instrument.description ? <span dangerouslySetInnerHTML={instrumentDescription(instrument.description)} /> : null }
```
mwiencek: I'd remove the `instrumentDescription` function and just inline it:

```
={instrumentDescription(instrument.description)}
vs.
={{__html: ' — ' + instrument.description}}
```

You can write the em-dash directly instead of using the `&mdash;` character entity reference.

Also remove the spaces `{ around this }` so it's `{like this}` to match our existing style.

root/instrument/List.js
```diff
@@ -0,0 +1,47 @@
...
+    </li>
+  );
+};
+
+const InstrumentList = () => (
+  <Layout>
```
mwiencek: You need to set `title={l('Instrument List')}` and `fullWidth={true}` props here.

root/instrument/List.js
```diff
@@ -0,0 +1,47 @@
+const React = require('react');
```
mwiencek: You need a copyright header atop this file. Copy it from here, excluding the `@flow` line: https://github.com/metabrainz/musicbrainz-server/blob/master/root/types.js#L2-L6

Replace "MetaBrainz Foundation" with your own real name. Keep one blank line after the header.

root/instrument/List.js
```diff
@@ -0,0 +1,47 @@
...
+              instrument => <Instrument key={instrument.id} instrument={instrument} />
+            )}
+          </div>
+        </ul>
+      </div>
+    </div>
+  </Layout>
+);
+
+module.exports = InstrumentList;
```
mwiencek: Can you add a single newline at the end of the file?

mwiencek:
Also, you can remove root/instrument/list.tt now too.

root/instrument/List.js
mwiencek: How about this, so the HTML makes a bit more sense (import `Frag` from root/components/Frag.js):
```JSX
{instrument.description
  ? <Frag>
      {' — '}
      <span className="description"
            dangerouslySetInnerHTML={{__html: instrument.description}} />
    </Frag>
  : null}
```

spellew:
They commits have all been squashed, a copyright has been added, and the other minor changes were made.

root/instrument/List.js
```diff
@@ -0,0 +1,59 @@
...
+const Instrument = (props) => {
+  const instrument = props.instrument;
+  return (
+    <li>
+      <EntityLink entity={instrument} />
+      {instrument.description
+        ? <Frag>
+          {' — '}
+          <span className="description"
+            dangerouslySetInnerHTML={{__html: instrument.description}} />
```
yvanzo: Instrument description should be localized.

spellew:
Instrument description now localized

root/instrument/List.js
```diff
@@ -0,0 +1,59 @@
...
+const Instrument = (props) => {
+  const instrument = props.instrument;
+  return (
+    <li>
+      <EntityLink entity={instrument} />
+      {instrument.description
+        ? <Frag>
+          {' — '}
+          <span className="description"
+            dangerouslySetInnerHTML={{__html: l(instrument.description)}} />
```
mwiencek: I don't think it will work, because i18n.js doesn't support domains other than `mb_server` right now. (The instrument descriptions are in the `instrument_descriptions` domain.) Also, it appears node-gettext doesn't handle domains correctly until version 2 (we're on version 1). I'll submit a PR to correct these issues.

mwiencek: https://github.com/metabrainz/musicbrainz-server/pull/599

mwiencek:
I fixed the i18n issues, and the HTML to remove some unnecessary `<div>`s (including an illegal one inside the `<ul>`). Please make sure when converting things that you match the HTML *exactly* as the TT has it, wherever technically possibly. :)

I also added some Flow types.

PR #597, opened 2018-01-12
MBS-9368: Ask for confirmation when leaving format empty

naiveaiguy:
After some discussion with @chhavijustme and @reosarevok:

https://chatlogs.metabrainz.org/brainzbot/metabrainz/2018-01-12/?msg=4096763&page=1
https://chatlogs.metabrainz.org/brainzbot/metabrainz/2018-01-12/?msg=4096935&page=2

We decided that a checkbox affirming that the user doesn't know what the format would result in a better UX.

https://tickets.metabrainz.org/browse/MBS-9368

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -280,17 +280,24 @@
         <td class="format">
           <label data-bind="attr: { for: 'disc-format-' + uniqueID }">[% l('Format:') %]</label>
           <select data-bind="value: formatID, attr: { id: 'disc-format-' + uniqueID }">
-            <option value="" selected="selected">&#160;</option>
+            <option value="" selected="selected">—</option>
             [%- FOR format=formats %]
             <option value="[% format.value %]">[% format.label %]</option>
             [%- END %]
           </select>
+          <label>
+            <input type="checkbox" data-bind="checked: formatUnknownToUser, enable: formatID() == ''" />
+            [% l('I don’t know') %]
+          </label>
           (<a href="[% c.uri_for('/doc/Release/Format') %]" target="_blank">[% l('help') %]</a>)
           <!-- ko if: release.mediums().length > 1 || name() -->
             <label data-bind="attr: { for: 'disc-title-' + uniqueID }">[% l('Disc title:') %]</label>
             <input type="text" data-bind="value: name, attr: { id: 'disc-title-' + uniqueID }" />
             <button type="button" class="icon guesscase-title" title="[% l('Guess case disc title') %]" data-click="guessCaseMediumName"></button>
           <!-- /ko -->
+          <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: needsFormat">
+            [%~ l('A format is required. If you don’t know it, tick the “I don’t know” checkbox next to the format dropdown.') %]
+          </p>
         </td>
 
         <td class="icon" style="text-align: right">
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -298,6 +298,7 @@ class Medium {
         this.name = ko.observable(data.name);
         this.position = ko.observable(data.position || 1);
         this.formatID = ko.observable(data.formatID);
+        this.formatUnknownToUser = ko.observable(Boolean(data.id && !data.formatID));
 
         var tracks = data.tracks;
         this.tracks = ko.observableArray(utils.mapChild(this, tracks, Track));
@@ -395,6 +396,16 @@ class Medium {
 
         this.needsTracks = ko.computed(function () {
             return self.loaded() && self.tracks().length === 0;
+        });
+
+        this.needsFormat = validation.errorField(function() {
+            return !(self.formatID() || self.formatUnknownToUser());
+        });
+
+        this.formatID.subscribe(function (value) {
+            if (value) {
+                self.formatUnknownToUser(false);
+            }
         });
     }
 


root/release/edit/tracklist.tt
```diff
@@ -286,6 +286,8 @@
             [%- END %]
           </select>
           (<a href="[% c.uri_for('/doc/Release/Format') %]" target="_blank">[% l('help') %]</a>)
+          <input type="checkbox" data-bind="checked: formatUnknownToUser, enable: formatID() == ''" />
```
naiveaiguy: NB: `disable: formatID` is shorter and also works here, but I felt it was better to be more explicit.

yvanzo:
@brainzbot, add to whitelist.

root/release/edit/tracklist.tt
```diff
@@ -286,6 +286,8 @@
...
+          [% l('I don’t know what the format is') %]
```
yvanzo: Please group the two above lines within a `label` tag.

yvanzo: Maybe the _help_ link should be put after that label?

chhavijustme:
@naiveaiguy Added the review on the ticket, please check.

naiveaiguy:
Addressed all concerns in amended commit.

root/release/edit/tracklist.tt
```diff
@@ -280,11 +280,15 @@
         <td class="format">
           <label data-bind="attr: { for: 'disc-format-' + uniqueID }">[% l('Format:') %]</label>
           <select data-bind="value: formatID, attr: { id: 'disc-format-' + uniqueID }">
-            <option value="" selected="selected">&#160;</option>
+            <option value="" selected="selected">--</option>
```
mwiencek: I have no opinion on the "no value" text but would like to point out that every other input on the website uses a blank option (see the information tab for example :)) so there is an inconsistency. Though it's out of the scope of this PR to change all of them, and I don't know what the benefit would be.

root/release/edit/tracklist.tt
```diff
@@ -324,6 +328,10 @@
     <!-- /ko -->
 
     <div data-bind="visible: loaded() && !collapsed()">
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsFormat">
+        [%~ l('A format is required. Select the checkbox next to the input if you don’t know it. Make sure to check the top of the tracklist.') %]
```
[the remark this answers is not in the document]
naiveaiguy: That seems to imply that the checkbox is at the top of the tracklist?

root/static/scripts/release-editor/fields.js
```diff
@@ -298,6 +298,7 @@ class Medium {
         this.name = ko.observable(data.name);
         this.position = ko.observable(data.position || 1);
         this.formatID = ko.observable(data.formatID);
+        this.formatUnknownToUser = ko.observable(data.formatUnknownToUser);
```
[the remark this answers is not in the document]
naiveaiguy: Oh yeah, will do.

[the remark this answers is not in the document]
naiveaiguy: Fixed.

root/release/edit/tracklist.tt
naiveaiguy: I do agree with that, but since @chhavijustme asked for it, I did it. It's a good start, and it's not too terrible of an inconsistency so as to confuse users.

root/release/edit/tracklist.tt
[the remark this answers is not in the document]
mwiencek: Hm? It is at the top of the tracklist (and the previous wording implies the same thing).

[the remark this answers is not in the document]
naiveaiguy: Oh, that may be confusing. What's actually meant to be said here is to check at the top of the tracklist for the *release format*. Maybe should make that more explicit?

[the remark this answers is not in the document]
yvanzo: I agree, referring to labels (format/unknown) seems more user-friendly than just tag names (input/checkbox).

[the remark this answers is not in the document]
mwiencek: I don't understand the distinction being made. The medium format ("input") is at the top of tracklist. The "I don't know" checkbox is next to the medium format, so it's also at the top of the tracklist.

I agree that referring to specific labels rather than inputs is better.

naiveaiguy:
@yvanzo Changed error type from `showErrorWhenTabIsSwitched` to `showErrorRightAway`.

naiveaiguy:
Need final review from @chhavijustme

yvanzo:
In the tab “Edit Note”, the button “Enter edit” should be disabled and the following banner should be displayed:
> _Some errors were detected in the data you’ve entered. Click on the tabs shaded red and correct any visible errors._

root/release/edit/tracklist.tt
chhavijustme: We are disabling checkbox if some option is selected. In case you don't know the format, one has to unselect the option, to select the checkbox. In which case, you are to choose the "no value text", which is not very evident.  I expect an "*It's not letting me unselect an option to select the checkbox*" case. Hence the suggestion to use "--" .

@mwiencek  @yvanzo You can take a call on if we want to start using this now or later to keep up with the consistency. I feel we will have to address this someday for sure though :)

mwiencek: I'm okay with updating them incrementally if you think it's a good idea, I don't think anyone will be confused.

Not to bikeshed here, but would it make sense to use an em dash (—) instead of two hyphens (--) since we prefer Unicode punctuation?

chhavijustme: Yes. em dash should work equally fine.

naiveaiguy: Updated.

naiveaiguy:
@mwiencek Fixed this, is it good to merge?

yvanzo:
Thank you for your changes, they address the above mentioned issues.

Final <s>boss</s> issue: Editing an existing release abnormally requires to tick the format checkbox for each medium, even the ones not modified by the editor. To avoid that, the checkbox should be ticked by default for existing medium. However, it should not be ticked by default for new medium.

naiveaiguy:
Not sure how to do above - determining whether a medium is existing or new seems impossible. @mwiencek?

mwiencek:
If `medium.id` is truthy then it's an existing medium.

naiveaiguy:
Made `formatUnknownToUser` initialise to `Boolean(data.id)`. `this.id` is set further down, but only if the id isn't null - which is not what we want in this case.

root/static/scripts/release-editor/fields.js
```diff
@@ -298,6 +298,7 @@ class Medium {
         this.name = ko.observable(data.name);
         this.position = ko.observable(data.position || 1);
         this.formatID = ko.observable(data.formatID);
+        this.formatUnknownToUser = ko.observable(Boolean(data.id));
```
mwiencek: But only if there's no format ID selected. So it should be `Boolean(data.id && ! data.formatID)`, right?

naiveaiguy: I'm not sure about Knockout's architecture here - is the `Medium` object constructed before or after the UI is made? It seems to be that it would be before, so this wouldn't be necessary, but I'm not sure.

yvanzo:
Thank you, it perfectly addresses the behaviour issue for new/existing medium.

From further tests, there is just a tiny layout issue when there are more than 1 medium for the release: Error message is displayed in between medium format inputs and medium name, breaking the line into three parts. It should be put after medium name rather than before it.

PR #598, opened 2018-01-13
MBS-9582: Add UtaTen to the lyrics whitelist

TehTotalPwnage:
As part of Google Code-in, here's code that adds UtaTen to the lyrics whitelist. It also performs cleaning of the input URL and validates to make sure that the URL conforms to the entity type being entered.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1656,6 +1656,26 @@ const CLEANUPS = {
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?kickstarter\.com\/projects\/(\d+)\/([\w\-]+)(?:[\/?#].*)?$/, "https://www.kickstarter.com/projects/$1/$2");
       return url;
     }
+  },
+  utaten: {
+    match: [new RegExp("^(https?://)?([^/]+\\.)?utaten\\.com/", "i")],
+    type: LINK_TYPES.lyrics,
+    clean: function (url) {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1656,6 +1656,26 @@ const CLEANUPS = {
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?kickstarter\.com\/projects\/(\d+)\/([\w\-]+)(?:[\/?#].*)?$/, "https://www.kickstarter.com/projects/$1/$2");
       return url;
     }
+  },
+  utaten: {
+    match: [new RegExp("^(https?://)?([^/]+\\.)?utaten\\.com/", "i")],
+    type: LINK_TYPES.lyrics,
+    clean: function (url) {
+      return url.replace(/^(?:https?:\/\/)?utaten\.com\/(artist|lyric\/.+)\/(.+?(?=\/|#|\?|$)).*$/, "https://utaten.com/$1/$2");
```
yvanzo: I cannot understand the meaning of combined quantifiers at the end of the regexp.
For example, what is the meaning of `.+?` and why is the `?` necessary here?

Instead, I suggest to replace `\/(.+?(?=\/|#|\?|$)).*$`  with the much simpler `\/([^\/?#]+).*$`

TehTotalPwnage: The original way I wrote the regular expression was to use a lookahead. `.+?` means to match any character one to unlimited times, lazily instead of greedily. It would stop at the lookahead which would match either of the four characters `\/|#|\?|$`.

I do agree that this seems to be a lot cleaner however, and the tests seem to pass unlike the original non-capturing group that you proposed. I'll make these changes when I find the time.

yvanzo:
@brainzbot, test this please.

PR #599, opened 2018-01-15
MBS-9599, MBS-9600, MBS-9603: Support other gettext domains in the JS

mwiencek:
Replaces node-gettext with Jed (which we use on the client) and performs some mild refactoring so we can reuse some of the i18n module for domain-specific variants.

I tested this as it relates to the referenced tickets, and switched languages on various pages and didn't see any obvious problems, but more testing would be appreciated.

Fixes [MBS-9599](https://tickets.metabrainz.org/browse/MBS-9599), [MBS-9600](https://tickets.metabrainz.org/browse/MBS-9600), [MBS-9603](https://tickets.metabrainz.org/browse/MBS-9603)


root/static/scripts/series.js
```diff
@@ -37,7 +37,10 @@ $(function () {
   });
 
   series.orderingTypeDescription = ko.computed(function () {
-    return lp(MB.orderingTypesByID[series.orderingTypeID()].description, 'series_ordering_type');
+    return lp_attributes(
```
mwiencek: I thought about whether we should just return the `l_name` and other translated props in `/ws/js`, so we wouldn't need to bundle strings from other domains in the JavaScript.

A downside is that if we decide to cache the `/ws/js` results, we'd have to do it per language. We'd also use slightly more bandwidth, though that probably isn't a real concern. But if they're in JS files, the client should keep them cached.

On the other hand, more JavaScript to load/parse/execute will make the page slightly slower, though probably not in a noticeable way.

mwiencek: In db210f3 we now have separate bundles for each domain. So the attributes bundle is only included where it's used (entity edit pages).

yvanzo: Caching translated props would be very convenient to convert templates effortlessly, but it seems to be issue-prone and further in-place localization might require to have domains available from JS anyway.

mwiencek:
> Just one possible change: provisioning BUNDLE_DOMAINS with all domains.

Amended the commit to build all text domains.

PR #600, opened 2018-01-15
Fix SEARCH_X_ACCEL_REDIRECT to work with solr

samj1912:
A fix by changing the redirect URL so that regex parsing is valid
at
https://github.com/metabrainz/openresty-gateways/blob/master/files/nginx/includes/musicbrainz-website_common.conf#L1

### DIFF
--- a/lib/MusicBrainz/Server/Data/WebService.pm
+++ b/lib/MusicBrainz/Server/Data/WebService.pm
@@ -218,7 +218,8 @@ sub xml_search
     }
 
     if (DBDefs->SEARCH_X_ACCEL_REDIRECT) {
-        return { redirect_url => '/internal/search/' . DBDefs->SEARCH_SERVER . $url_ext }
+        my $redirect_url = (DBDefs->SEARCH_ENGINE eq 'LUCENE') ? '/internal/search/' : '/internal/solr/';
+        return { redirect_url => $redirect_url . DBDefs->SEARCH_SERVER . $url_ext };
     } else {
         my $url = 'http://' . DBDefs->SEARCH_SERVER . $url_ext;
         my $response = $self->c->lwp->get($url);


mwiencek:
Can you explain? There's no `/internal/solr/` rule in the linked nginx config (and I don't understand why you would need a separate one in any case).

lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -204,21 +204,24 @@ sub xml_search
...
     } else {
         my $format = ($args->{fmt} // "") eq "json" ? "mbjson" : "mbxml";
         my $def_type = $dismax eq 'true' ? 'dismax' : 'lucene';
+        $redirect_url = '/internal/solr/'
         $url_ext = "/$resource/select?" .
             "rows=$limit&wt=$format&start=$offset" .
             "&q=" . uri_escape_utf8($query) . "&defType=$def_type";
     }
 
     if (DBDefs->SEARCH_X_ACCEL_REDIRECT) {
```
zas: Since the trick is only related to X_ACCEL thing, i would keep the code together:
```perl
my $redirect_url = (DBDefs->SEARCH_ENGINE eq 'LUCENE') ? '/internal/search/' : '/internal/solr/';
return { redirect_url => $redirect_url . DBDefs->SEARCH_SERVER . $url_ext }
```

zas:
> Can you explain? There's no /internal/solr/ rule in the linked nginx config (and I don't understand why you would need a separate one in any case).

I added one, not yet committed, as we need to ensure it works first

```nginx
location ~* ^/internal/solr/(.*?)/solr/(.*) {
  internal;

  # Extract url from the request
  set $search_uri $2;
  set $search_host $1;

  # Compose url
  set $search_url http://$search_host/solr/$search_uri?$args;

  proxy_pass $search_url;

  add_header Access-Control-Allow-Origin *;
}
```

lib/MusicBrainz/Server/Data/WebService.pm
mwiencek: It also unbreaks the code (missing semicolons :)).

PR #603, opened 2018-01-18
MBS-9606: Rewrite MusicBrainz ISRC index page in React/JSX

spellew:
Replaced all the isrc/index.tt file with an equivalent template written in React/JSX. Accepted on Google Code-In by yvanzo.


yvanzo:
Depends on PR https://github.com/metabrainz/musicbrainz-server/pull/599.

mwiencek:
The page couldn't load for me (it would throw an exception right away) and was broken in various ways. I pushed some fixes and added some Flow types. I think it's in a mergeable state now.

The types create conflicts with #608 but I intend to fix those myself

lib/MusicBrainz/Server/Entity/CoreEntity.pm
```diff
@@ -35,7 +35,11 @@ has 'unaccented_name' => (
 around TO_JSON => sub {
     my ($orig, $self) = @_;
 
-    return {%{ $self->$orig }, gid => $self->gid};
+    my $json = $self->$orig;
+    $json->{gid} = $self->gid;
+    $json->{name} = $self->name;
+    $json->{unaccented_name} = $self->unaccented_name;
+    return $json;
```
samj1912: Can we make this and the other subs like so, just to maintain consistency?
```perl
around TO_JSON => sub {
    my ($orig, $self) = @_;

    return {
        %{$self->$orig},
        $json->{gid} = $self->gid;
        $json->{name} = $self->name;
        $json->{unaccented_name} = $self->unaccented_name;
    };
};
```

mwiencek: Okay. I only used the current method because it's faster (about 2.7x with perl 5.26 on my macbook; it doesn't have to allocate a new hash and then merge the hashes) and these `TO_JSON` methods have to be called hundreds or thousands of times on each page load. But I didn't do a proper benchmark so it may be negligible once you take the slow Moose method calls into account.

root/static/scripts/common/constants.js
```diff
@@ -1,111 +1,120 @@
+/*
+ * @flow
+ * Copyright (C) 2015 MetaBrainz Foundation
```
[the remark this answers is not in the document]
mwiencek: It's supposed to be the year of first publication, so I'm not sure it makes sense to update it unless significant changes are made.

lib/MusicBrainz/Server/Entity/CoreEntity.pm
samj1912: I don't mind either actually. I just saw a lot of the code wrapping the TO_JSON as I quoted, so thought I'd point it out. If the PR version is indeed faster then maybe we'd want to keep that and switch newer code to it too.

mwiencek: Well, I guess I'll keep what I had since I'm lazy. :) But it was probably premature optimization. Moose is slow, and the slowness of the `around` modifiers probably dwarfs the hash allocations.

mwiencek:
Force-pushed some tweaks to root/types.js to avoid conflicts with #608

PR #604, opened 2018-01-18
MBS-9608: Update URL cleanup to handle new Bandsintown’s URL format

yvanzo:
## [MBS-9608](https://tickets.metabrainz.org/browse/MBS-9608): Update Bandsintown URL cleanup to the new format

Bandsintown recently changed its default URL pattern:  It uses one-letter prefix (instead of no prefix for artist and type name for `event` and `venue`) and numeric ID everywhere (instead of artist name).
While Bandinstown correctly redirect URLs using the old format, MusicBrainz doesn’t accept entering URLs in the new format yet.
    
This patch updates Bandsintown URL cleanup to handle both URL formats from user input, to convert input URLs in old format to the new format, and to validate URLs in the new format only.

Additionally, it normalizes Bandsintown URLs to use the new default subdomain `ẁww` of `bandsintown.com`. It should prevent potential further reopening of [MBS-9559](https://tickets.metabrainz.org/browse/MBS-9559).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1053,18 +1053,18 @@ const CLEANUPS = {
       if (m) {
         var prefix = m[1];
         var number = m[2];
-        url = "https://bandsintown.com/" + prefix + "/" + number;
+        url = "https://www.bandsintown.com/" + prefix + "/" + number;
       } else {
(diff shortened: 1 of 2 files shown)


yvanzo:
Failing `tests from t/script/NotifyOfNewReleases.t` in [build #806](https://ci.metabrainz.org/job/musicbrainz-server/806/) are unrelated to this change.

yvanzo:
@brainzbot, retest this please.

PR #606, opened 2018-01-29
Add a selenium test for /work/create, add more flow types, and remove Immutable.js

mwiencek:
This PR wraps together several changes that are somewhat related. I hope it's not too large, and that the individual commits are easier to digest. I can split the commits up further if necessary.

The files added under the flow-typed/ directory were installed using the `flow-typed` tool.

I sadly look forward to Immutable.js being gone, since it's been giving me headaches for a while as I've been trying to reconcile components and functions used in and outside of editing UIs. Here I propose using plain JS objects and arrays everywhere, and using lenses for immutable updates in the editing UIs (see the relevant commits).

### DIFF
--- a/root/static/scripts/work.js
+++ b/root/static/scripts/work.js
@@ -52,10 +52,22 @@ const store = createStore(function (state = formFromHash(form), action) {
   return state;
 });
 
+function subfieldErrors(field, accum = []) {
+  if (field.errors) {
+    accum = accum.concat(field.errors);
+  }
+  if (field.field) {
+    _.each(field.field, function (subfield) {
+      accum = subfieldErrors(subfield, accum);
+    });
+  }
+  return accum;
+}
+
 class WorkAttribute {
   constructor(data, parent) {
     this.attributeValue = ko.observable(_.get(data, ['field', 'value', 'value']));
-    this.errors = ko.observableArray(data.errors);
+    this.errors = ko.observableArray(subfieldErrors(data));
     this.parent = parent;
     this.typeHasFocus = ko.observable(false);
     this.typeID = ko.observable(_.get(data, ['field', 'type_id', 'value']));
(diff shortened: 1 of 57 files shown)


mwiencek:
I further split the "Remove Immutable.js" commit into three separate commits. I think each commit is reviewable in isolation. The enormous files added under flow-typed/ artificially inflate the size of the diff and can be ignored. Pinging @yvanzo :)

root/static/scripts/work.js
```diff
@@ -1,6 +1,14 @@
+// @flow
+// Copyright (C) 2015-2018 MetaBrainz Foundation
+//
+// This file is part of MusicBrainz, the open internet music database,
+// and is licensed under the GPL version 2, or (at your option) any
+// later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+const $ = require('jquery');
```
samj1912: Why do we need jQuery?

root/static/scripts/work.js
```diff
@@ -1,6 +1,14 @@
+// @flow
```
samj1912: We can do an eslint run here? It picked up quite a few inconsistencies.

mwiencek: Opened this PR before I setup eslint :) I will go back and run it on things.

root/static/scripts/work.js
mwiencek: It's used down here https://github.com/mwiencek/musicbrainz-server/blob/588282c88ea40f945071485367545b0eda2dcfd0/root/static/scripts/work.js#L249-L258 so this just adds the import for it so that Flow doesn't complain. (It's also a global variable, hence why it previously worked.)

The edit pages haven't really been ported to React yet, so you'll see a lot of jQuery in places.

mwiencek:
Nice, the added selenium test caught a bug in a previously merged commit: 3dfdf7edf2ae4cfe097b100ca7b4330419e9bedf

PR #608, opened 2018-02-08
MBS-9626: Convert entity headers to React

samj1912:
Converts the following templates to React - 
* header
  * url
  * series
  * work
  * release-group
  * recording
  * label
  * event
  * place
  * area
  * artist
* entity-tabs
* tabs
* tagger-icon


samj1912:
I have yet to figure out flow types. Will add that soon. Also, I am *extremely* new to JS. So code style might be all over the place and the code might not be the best ever. So reviews appreciated.

The component seems to be working correctly from what I tested.

samj1912:
Also, any reason we don't use es6 import and export?

root/url/Header.js
```diff
@@ -0,0 +1,43 @@
...
+  if (url.edits_pending)
+    urlObj = (
+        <span className="mp">
+          {urlEntity}
+        </span>
+      );
+
+  let subHeading = l('URL');
+  if (url.type_name)
+    subHeading = url.type_name;
```
samj1912: I am not sure where this is defined. Can anyone help me with this?

yvanzo: `type_name` is not defined for URL entities, it is defined for Artist, Event, Series, and Work.
There is no plan to specify `type` for URLs since their roles are actually specified with relationship types.
Just drop that!

samj1912:
I added the flow types but I am not sure how correct they are. Please verify.

root/components/Tabs.js
```diff
@@ -0,0 +1,32 @@
...
+type LinkT = [string, React.Node];
+
+type Props = {
+  list: LinkT[];
+  page: string;
+};
+
+const Tabs = ({list, page}: Props) => {
+  const tabs =  list.map((link) => {
+    return <li className={link[0] == page ? "sel" : null} key={link[0]}>{link[1]}</li>;
```
mwiencek: I think I would use a different, simpler approach that doesn't involve building lists and mapping them here.

1. Make the `Tabs` component accept no props. Just embed `{children}` and expect consumers to pass in the `<li>`s.
2. Replace `buildLinkObject` with an `EntityTabLink` component:
   ```JavaScript
   const EntityTabLink = ({selected, ...linkProps}) => (
     <li className={selected ? 'sel' : null}>
       <EntityLink {...linkProps} />
     </li>
   );
   ```
3. In the `EntityTabs` component, pass `page` to `buildLinks`. You can use this to set the `selected` prop on each `EntityTabLink` component.

root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -10,8 +10,7 @@ const EntityLink = require('./EntityLink');
 const ArtistCreditLink = ({artistCredit, showDeleted = true, ...props}) => {
   const names = artistCredit.names;
   let parts = [];
-  for (let i = 0; i < names.size; i++) {
```
mwiencek: `.size` should never be `undefined` if you're passing in an `Immutable.List` object, which is what `artistCreditFromArray` creates, so you should use that and revert this.

(Aside 1: `.map` creates a new list and then throws it away here; you wanted `forEach`, since you only wanted to iterate the list for side effects, not create a new list from the old one.)

(Aside 2: Immutable.js is being removed in https://github.com/metabrainz/musicbrainz-server/pull/606, so in the future we won't have to worry about converting plain arrays to Immutable lists.)

root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -23,13 +22,13 @@ const ArtistCreditLink = ({artistCredit, showDeleted = true, ...props}) => {
         <EntityLink
           content={credit.name}
           entity={artist}
-          key={i}
+          key={artist.gid}
```
mwiencek: Using the artist MBID as a key is incorrect, because an artist credit can link to the same artist multiple times (i.e. it doesn't uniquely identify each artist credit name). Using the index (position) as the key was correct in this case.

root/types.js
```diff
@@ -57,6 +57,15 @@ declare type OptionTreeT = {|
   +description: string,
 |};
 
+declare type WorkT = EntityT & {
```
mwiencek: All the types you've added here should be in alphabetical order, as noted in the comment at the top of the file. ;)

root/types.js
```diff
@@ -57,6 +57,15 @@ declare type OptionTreeT = {|
...
+    l_type_name?: string;
+};
+
+declare type URLT = EntityT & {
+    decoded: string;
+    edits_pending: bool;
```
mwiencek: In another branch I created an `EditsPendingT` to shared between multiple types. Seems like a good idea to add it now to avoid conflicts later:
```JavaScript
declare type EditsPendingT = {|
  +editsPending: boolean,
|};

[...]

declare type URLT =
  & EditsPendingT
  & EntityT;
```

root/url/Header.js
```diff
@@ -0,0 +1,45 @@
...
+const EntityTabs = require('../components/EntityTabs');
+const EntityLink = require('../static/scripts/common/components/EntityLink');
+
+type Props = {
+  url: URLT;
+  page: string;
+  entity_properties: EntityPropsT;
+};
+
+const URLHeader = ({url, page, entity_properties}: Props) => {
```
mwiencek: `entity_properties` is a constant that should exist in `$c.stash` after you add it to this list: https://github.com/metabrainz/musicbrainz-server/blob/612cc7d/lib/MusicBrainz/Server.pm#L438

Constants don't need to be passed as props (since they never change, stateless functional components that reference them will remain pure). So every place that you either pass or accept `entity_properties` as a prop can be removed, and all places that read from `entity_properties` can instead read from `$c.stash.entity_properties`.

root/url/Header.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const urlEntity = <EntityLink content={url.decoded} entity={url} subPath={'show'} />;
+
+  let urlObj = urlEntity;
+  if (url.editsPending)
+    urlObj = (
+        <span className="mp">
```
mwiencek: Wrapping links in `<span className="mp" />` is already a job performed by the `EntityLink` component: https://github.com/metabrainz/musicbrainz-server/blob/612cc7d/root/static/scripts/common/components/EntityLink.js#L165-L165

I'd remove this and figure out why the existing behavior isn't being triggered.

root/release_group/Header.js
```diff
@@ -0,0 +1,35 @@
...
+  entity_properties: EntityPropsT;
+};
+
+const ReleaseGroupHeader = ({rg, page, entity_properties}: Props) => {
+  const artistCredit = <ArtistCreditLink artistCredit={artistCreditFromArray(rg.artistCredit)} />;
+  const header = (
+    <div className="rgheader">
+      <h1><EntityLink entity={rg} /></h1>
+      <p className="subheader">
+        <span className="prefix">~</span> {l(`Release group by {artistCredit}`, {__react: true, artistCredit: artistCredit})}
```
mwiencek: Make sure to use the same strings as the original header files, so they don't have to be retranslated: `l('Release group by {artist}', ...)`.

root/url/Header.js
mwiencek: Sorry, I forgot how our code works.

Don't change that line I linked, and don't access `$c.stash.entity_properties`. Just import `ENTITIES` from root/static/scripts/common/constants.js. :)

mwiencek:
I fixed our eslint setup in the master branch to help you adjust the code style, so rebase to get that and do an `npm install`. Then run `./node_modules/.bin/eslint` on individual files. Also, very useful: pass `--fix` to have it automatically fix certain issues for you (make sure you have no unrelated, unstaged changes, because it'll modify the file directly).

If you see it suggesting or changing something weird, ping me, because the rules might need adjusting still.

samj1912:

root/components/Header.js
```diff
@@ -0,0 +1,46 @@
...
+      {$c.session.tport ? <TaggerIcon entity={entity} /> : null}
+      <h1>{heading}</h1>
+      <SubHeader subHeading={subHeading} />
+    </div>
+  );
+
+  const tabs = <EntityTabs entity={entity} {...props} />;
+
+  return [header, tabs];
+};
```
mwiencek: This whole component can be more simply written:

```JavaScript
const Header = ({entity, headerClass, heading, subHeading, ...props}: Props) => (
  <React.Fragment>
    <div className={headerClass}>
      {$c.session.tport ? <TaggerIcon entity={entity} /> : null}
      <h1>
        {heading || <EntityLink entity={entity} />}
      </h1>
      <SubHeader subHeading={subHeading} />
    </div>
    <EntityTabs entity={entity} {...props} />
  </React.Fragment>
);
```

Right now you're allocating an `<EntityLink />` even if it's unused.

`React.Fragment` is similar to returning an array, but doesn't require keys and allows for a consistent syntax.

root/components/EntityTabs.js
```diff
@@ -0,0 +1,104 @@
...
+  map: l('Map'),
+  performances: l('Performances'),
+  places: l('Places'),
+  recordings: l('Recordings'),
+  releases: l('Releases'),
+  users: l('Users'),
+  works: l('Works'),
+};
+
+const buildLinkObject = (content, entity, subPath, page, pageName) => {
```
mwiencek: s/buildLinkObject/buildLink/ maybe? It's shorter and I'm not sure `Object` adds any information. :) (You don't use `buildLinkObjects` below.)

root/types.js
```diff
@@ -1,20 +1,40 @@
...
+ * TO_JSON methods under lib/MusicBrainz/Server/Entity/; those are precisely
+ * how data is serialized for us.
+ */
+
+declare type AreaT = CoreEntityT & {
+    typeName: string;
+    containment?: AreaT[];
+};
+
+declare type ArtistCreditT = {
```
mwiencek: Rename this to `ArtistCreditNameT` and define it as:

```
declare type ArtistCreditNameT = {
  artist: ArtistT;
  joinPhrase: string;
  name: string;
};
```

Add an `ArtistCreditT` defined as:

```
declare type ArtistCreditT = Array<ArtistCreditNameT>;
```

root/types.js
```diff
@@ -1,20 +1,40 @@
...
+ * Types are in alphabetical order.
+ *
+ * The definitions in this file are intended to model the output of the
+ * TO_JSON methods under lib/MusicBrainz/Server/Entity/; those are precisely
+ * how data is serialized for us.
+ */
+
+declare type AreaT = CoreEntityT & {
+    typeName: string;
+    containment?: AreaT[];
```
mwiencek: These should be indented two spaces. (We don't support Flow in our eslint setup yet.)

root/types.js
```diff
@@ -1,20 +1,40 @@
...
+declare type ArtistT = CoreEntityT & {
+    typeName: string;
+};
 
 type CatalystContextT = {
-  stash: CatalystStashT,
-  user_exists: bool,
+  stash: CatalystStashT;
+  user_exists: bool;
+  session: {};
```
mwiencek: More accurately:
```
  session: {[string]: mixed} | null,
  sessionid: string | null,
```

root/components/Header.js
```diff
@@ -0,0 +1,46 @@
...
+const Header = ({entity, headerClass, subHeading, ...props}: Props) => {
+  let heading = <EntityLink entity={entity} />;
+
+  if (props.heading) {
+    heading = props.heading;
+  }
+
+  const header = (
+    <div className={headerClass}>
+      {$c.session.tport ? <TaggerIcon entity={entity} /> : null}
```
mwiencek: I think `$c.session` is sometimes null (perhaps if you're not logged in). So you'd have to do `{($c.sessionid && $c.session.tport) ? ...}`.

root/label/Header.js
```diff
@@ -0,0 +1,31 @@
...
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+const React = require('react');
+const {l} = require('../static/scripts/common/i18n');
+const Header = require('../components/Header');
+
+type Props = {
+  label: EntityT;
```
mwiencek: `CoreEntityT` would be more accurate. Even better, add a `LabelT` (can just be an alias for `CoreEntityT`). Then we won't have to fix this later.

root/series/Header.js
```diff
@@ -0,0 +1,31 @@
...
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+const React = require('react');
+const {lp} = require('../static/scripts/common/i18n');
+const Header = require('../components/Header');
+
+type Props = {
+  series: EntityT;
```
mwiencek: `SeriesT`?

root/static/scripts/common/utility/isSpecialPurposeArtist.js
```diff
@@ -0,0 +1,8 @@
+const {DARTIST_ID, VARTIST_ID, VARTIST_GID} = require('../constants');
```
mwiencek: Add a license header here.

root/components/EntityTabs.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return propLinks;
+};
+
+const buildLinks = (editTab, entity, page) => {
+  let infoLinks = [buildLinkObject(l('Overview'), entity, 'show', page, 'index')];
+  infoLinks = infoLinks.concat(buildPropertyLinks(entity, page));
+
+  if (typeof editTab === 'string' && editTab === 'default') {
```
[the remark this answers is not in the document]
samj1912: I went with a slightly different approach. I added a hideEditTab (as opposed to showEditTab, in order to make the hiding more explicit since it was only in special cases that we need to hide the edit tab and I pass in an editTab arg. If editTab is not supplied, it uses the default edit tab and hideEditTab hides the edit tab regardless of the args.)

root/components/Header.js
```diff
@@ -0,0 +1,40 @@
...
+  headerClass: string;
+  heading?: string | React.Node;
+  page: string;
+  subHeading: string;
+};
+
+const Header = ({entity, headerClass, heading, subHeading, ...props}: Props) => (
+  <Frag>
+    <div className={headerClass}>
+      {$c.session && $c.session.tport ? <TaggerIcon entity={entity} /> : null}
```
mwiencek: Hmm, it's not clear why the `tport` is checked outside the component, but the entity types are checked inside the component. Maybe just move the entity checks here too? But in either case, move the entity array constant (`['recording', 'release']`) outside the render definition so that it isn't reallocated. (We actually have a babel plugin enabled that does [similar magic for constant react elements](https://babeljs.io/docs/plugins/transform-react-constant-elements/).)

root/types.js
```diff
@@ -1,20 +1,49 @@
...
+/*
+ * Types are in alphabetical order.
+ *
+ * The definitions in this file are intended to model the output of the
+ * TO_JSON methods under lib/MusicBrainz/Server/Entity/; those are precisely
+ * how data is serialized for us.
+ */
+
+declare type AreaT = CoreEntityT & {
+  typeName: string;
```
mwiencek: I realize the other object types mostly used semi-colons to separate the fields, but that's actually a [deprecated syntax](https://flow.org/en/docs/types/objects/) (they should use commas).

root/types.js
```diff
@@ -43,30 +76,67 @@ declare type EntityT = {
 declare type InstrumentT =
   & EntityT
   & {
-    +description: string,
```
mwiencek: The `+` there marked the property as covariant, which basically makes it read-only (you can't do `instrument.description = blah;`): https://flow.org/blog/2016/10/04/Property-Variance/

I think it makes sense to keep, and it should be added to other properties that should be read-only (if not all, nearly all).

root/types.js
```diff
@@ -43,30 +76,67 @@ declare type EntityT = {
...
 declare type RatableT = EntityT & {
   rating: number | null;
   rating_count: number;
   user_rating: number | null;
 };
 
+declare type ReleaseGroupT =
+  & CoreEntityT
+  & {
+    artistCredit: Array<ArtistCreditT>;
```
mwiencek: This should be just `ArtistCreditT` now.

root/types.js
```diff
@@ -43,30 +76,67 @@ declare type EntityT = {
...
+};
+
+declare type RecordingT =
+  & CoreEntityT
+  & {
+    artistCredit: Array<ArtistCreditT>;
```
mwiencek: Same here, just `ArtistCreditT`.

root/types.js
mwiencek: From what I see in lib/MusicBrainz/Server/Entity/Role/Type.pm, *all* of the `typeName` properties should be typed as `typeName?: string` (meaning they could be missing).

It's important the types exactly model what the `TO_JSON` gives us, otherwise the type checking won't be sound and we could introduce logical errors.

root/types.js
```diff
@@ -1,20 +1,49 @@
...
+declare type ArtistCreditNameT = {
+  artist: ArtistT;
+  joinPhrase: string;
+  name: string;
+};
+
+declare type ArtistCreditT = ArtistCreditNameT[];
+
+declare type ArtistT = CoreEntityT & {
+    typeName: string;
```
mwiencek: Nit: there's still some types in here using 4 spaces.

samj1912:
@brainzbot retest this please

root/components/Header.js
```diff
@@ -0,0 +1,45 @@
...
+type Props = {
+  entity: EntityT;
+  editTab?: string | React.Node;
+  headerClass: string;
+  heading?: string | React.Node;
+  page: string;
+  subHeading: string;
+};
+
+const TaggableEntities = ['recording', 'reelase'];
```
mwiencek: lowercase 't' in `taggableEntities` (only classes should use PascalCase) and s/reelase/release/

mwiencek:
If it's easy to squash things into one commit per entity type, that would be preferable, but if not you can squash the whole thing.

root/components/Header.js
```diff
@@ -0,0 +1,45 @@
...
+type Props = {
+  editTab?: string | React.Node;
+  entity: EntityT;
+  headerClass: string;
+  heading?: string | React.Node;
+  page: string;
+  subHeading: string;
+};
+
+const taggableEntities = ['recording', 'relase'];
```
mwiencek: release is still spelled wrong

samj1912: Late night rebasing :P

mwiencek:
@brainzbot retest this please

mwiencek:
It looked excellent overall, I just fixed a few issues and style nits. If you curious about what I changed, here's the diff: https://gist.github.com/mwiencek/322590ade949874d814a73ca96bb5371 (let me know if something looks wrong).

Not sure why tests are failing...

mwiencek:
@brainzbot retest this please

root/components/EntityHeader.js
```diff
@@ -0,0 +1,56 @@
...
+  heading,
+  hideEditTab = false,
+  page,
+  // $FlowFixMe
+  preHeader = null,
+  subHeading,
+}: Props) => (
+  <Frag>
+    <div className={headerClass}>
+      {preHeader}
```
mwiencek: I added a `preHeader` prop and moved the `TaggerLink` inclusion outside, so that the component has no entity-specific code at all.

root/components/EntityTabs.js
```diff
@@ -0,0 +1,110 @@
...
+  <EntityTabLink
+    content={content}
+    entity={entity}
+    key={subPath}
+    selected={pageName === page}
+    subPath={subPath}
+  />
+);
+
+function buildLinks(
```
mwiencek: I combined `buildPropertyLinks` into `buildLinks` because I think it was slightly over-abstracted (it's clearer to me to see all the links being built in one place, and the arrays don't have to be concatenated).

root/types.js
```diff
@@ -1,24 +1,72 @@
...
+ * Types are in alphabetical order.
+ *
+ * The definitions in this file are intended to model the output of the
+ * TO_JSON methods under lib/MusicBrainz/Server/Entity/, those are precisely
+ * how data is serialized for us.
+ */
+
+declare type AreaT =
+  & CoreEntityT
+  & TypeRoleT<AreaTypeT>
```
mwiencek: Added a `TypeRoleT` to model `MusicBrainz::Server::Entity::Role::Type`.

It doesn't currently use the type parameter, but I'm thinking we'll add a `type` prop there (to replace `typeName`, i.e. we'd access `entity.type.name` instead, and have access to the full type).

root/types.js
```diff
@@ -1,24 +1,72 @@
...
+  & {|
+      +containment?: $ReadOnlyArray<AreaT>,
+      +entityType: 'area',
+    |};
+
+export opaque type AreaTypeT: OptionTreeT = OptionTreeT;
```
mwiencek: Opaque types are documented at https://flow.org/en/docs/types/opaque-types/

They need `opaque` because otherwise e.g. you could pass an `AreaTypeT` where an `ArtistTypeT` is expected, since they are both aliases to `OptionTreeT`. The `opaque` hides the underlying type from code outside the file.

root/artist/ArtistHeader.js
```diff
@@ -0,0 +1,37 @@
...
+  if (artist.typeName) {
+    headerClass += ` ${artist.typeName.toLowerCase()}-icon`;
+  }
+  return (
+    <EntityHeader
+      entity={artist}
+      headerClass={headerClass}
+      hideEditTab={isSpecialPurposeArtist(artist)}
+      page={page}
+      subHeading={artist.typeName ? lp_attributes(artist.typeName, 'artist_type') : l('Artist')}
```
mwiencek: Fixed a couple of the type name fallbacks which were lowercase.

root/recording/RecordingHeader.js
```diff
@@ -0,0 +1,41 @@
...
+  const lArgs = {
+    __react: true,
+    artist: artistCredit,
+  };
+  <EntityHeader
+    entity={recording}
+    headerClass="recordingheader"
+    page={page}
+    preHeader={<TaggerIcon entity={recording} />}
+    subHeading={recording.video ? l('Video by {artist}', lArgs) : l('Recording by {artist}', lArgs)}
```
mwiencek: Previously this was:

```JavaScript
l(`${recording.video ? 'Video' : 'Recording'} by {artist}`, {__react: true, artist: artistCredit})
```
That's improper because `xgettext` can't extract strings from that - it can only extract string literals. It can't execute code (and ignores template literals entirely).

mwiencek:
Fixed one more issue. The strings in `tabLinkNames` were being translated *when the module was loaded*, not when the page was being rendered (which is important because `l()` depends on the current language). I added `N_*` functions to i18n.js to solve this, similar to what we have in the Perl.

PR #610, opened 2018-02-14
MBS-9636: Load types and other work attributes on opening side-bar links

samj1912:


lib/MusicBrainz/Server/Controller/Work.pm
```diff
@@ -102,7 +102,7 @@ with 'MusicBrainz::Server::Controller::Role::Merge' => {
     edit_type => $EDIT_WORK_MERGE,
 };
 
-before qw( show aliases tags details edit ) => sub {
+before qw( show aliases tags details edit edit_annotation open_edits edits ) => sub {
```
samj1912: I ordered this top to bottom on sidebar order since the others seem to be order LTR according to tab order.

mwiencek: Are the languages and work attributes needed on these new pages? If not, probably better to add a separate `before` that just loads the type.

samj1912: Is this better?

lib/MusicBrainz/Server/Controller/Work.pm
```diff
@@ -102,10 +102,15 @@ with 'MusicBrainz::Server::Controller::Role::Merge' => {
...
     my ($self, $c) = @_;
     my $work = $c->stash->{work};
     $c->model('WorkType')->load($work);
+};
+
+before qw( show aliases tags details edit ) => sub {
```
samj1912: Is this better?

PR #613, opened 2018-02-16
MBS-3643: Add Musixmatch to the lyrics whitelist

yvanzo:
## [MBS-3643](https://tickets.metabrainz.org/browse/MBS-3643): Add Musixmatch to the lyrics whitelist

Add auto-select, cleanup and validation for Musixmatch _artist_ and _lyrics_ URLs to be linked to MusicBrainz _artist_ and _work_, respectively.

It doesn’t allow linking Musixmatch _album_ which matches neither MusicBrainz _release group_ nor _release_.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1571,6 +1571,28 @@ const CLEANUPS = {
       return false;
     }
   },
+  musixmatch: {
+    match: [new RegExp("^(https?://)?([^/]+\\.)?musixmatch\\.com/", "i")],
+    type: LINK_TYPES.lyrics,
+    clean: function (url) {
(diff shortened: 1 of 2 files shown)


mwiencek:
The second-to-last comment on MBS-3643 is from 2015 and says "[t]his was reverted because of lack of style approval." I'm assuming it has since been re-approved by reosarevok, but it would be nice to link to the context, otherwise the ticket history is a bit confusing.

yvanzo:
That comment was added as he reopened the coding task [https://MBS-3643](https://tickets.metabrainz.org/browse/MBS-3643) just after he closed the request [STYLE-518](https://tickets.metabrainz.org/browse/STYLE-518) as _Fixed_ which means to me it has been approved as a lyrics source. Both tickets are already linked together. I agree it still is a bit confusing, that probably explains why it wasn’t addressed since 2015. I don’t know if more context (I don’t know about) can be added, maybe @reosarevok knows?

reosarevok:
This has my full Style BDFL approval and all

PR #614, opened 2018-02-18
 Port components/alias.tt to React

sambhav:


root/components/Aliases.js
```diff
@@ -0,0 +1,105 @@
...
+          <a href={`/${entity.entityType}/${entity.gid}/alias/${alias.id}/delete`}>
+            {l('Remove')}
+          </a>
+        </Frag>
+      );
+    }
+    const row = (
+      <tr className={i % 2 ? 'even' : 'odd'} key={i}>
+        <td colSpan={showSortName ? 1 : 2}>
+          {alias.edits_pending ? <span className="mp">{isolateText(alias.name)}</span> : isolateText(alias.name)}
```
mwiencek: `editsPending`. You need to run Flow on your existing code. And you'll need to define an `AliasT` in root/types.js.

root/components/Aliases.js
```diff
@@ -0,0 +1,105 @@
...
+        </td>
+        {showSortName ? <td>{isolateText(alias.sort_name)}</td> : null}
+        <td>{formatDate(alias.begin_date)}</td>
+        <td>{alias.ended ? alias.end_date ? formatDate(alias.end_date) : l('Unknown') : null}</td>
+        <td>{alias.typeName}</td>
+        <td>
+          {alias.locale ? getLanguageName(alias.locale) : null}
+          {alias.primary_for_locale ? l(' ({primary})', {__react: true, primary: <span className="comment">{l('primary')}</span>}) : null }
```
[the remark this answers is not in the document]
samj1912: Not sure how to do this with bracketed? It does not support react, it felt it was easier to do it this way.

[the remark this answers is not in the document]
mwiencek: You should be able to tweak it to allow passing additional args to `l`:

```JavaScript
function bracketed(text, args) {
  if (text) {
    return l(' ({text})', {text, ...args});
  }
  return '';
}

bracketed(<span className="comment">{l('primary')}</span>, {__react: true});
```

It's just nice if all the usages are consistent in case we want to change something later.

root/components/Aliases/index.js
```diff
@@ -0,0 +1,41 @@
+/*
```
mwiencek: Hrm, in https://github.com/metabrainz/musicbrainz-server/pull/603 we put the index page for the isrc page alongside index.tt, at root/isrc/Index.js. In either case, I'd like the name of the file and the name of the exported component to be the same. What do you think?

If we go with your layout I think I would lowercase 'aliases' in the path and capitalize 'Index.js'.

root/components/Aliases/AliasTableBody.js
```diff
@@ -0,0 +1,39 @@
...
+          {' | '}
+          <a href={entityHREF(entity.entityType, entity.gid, `/alias/${alias.id}/delete`)}>
+            {l('Remove')}
+          </a>
+        </Frag>
+      );
+    }
+    aliasRows.push(<AliasTableRow actions={actions} alias={alias} row={i % 2 ? 'even' : 'odd'} />);
+  }
+  return aliasRows;
```
mwiencek: Nit: Since the name of the component is `AliasTableBody`, return a `<tbody>`?

root/components/Aliases/AliasTableRow.js
```diff
@@ -0,0 +1,35 @@
...
+      ? null
+      : <td>{isolateText(alias.sort_name)}</td>}
+    <td>{formatDate(alias.begin_date)}</td>
+    <td>{alias.ended ? alias.end_date ? formatDate(alias.end_date) : l('Unknown') : null}</td>
+    <td>{alias.typeName}</td>
+    <td>
+      {alias.locale ? locales[alias.locale] : null}
+      {alias.primary_for_locale ? l(' ({primary})', {__react: true, primary: <span className="comment">{l('primary')}</span>}) : null }
+    </td>
+    <td>{actions}</td>
```
mwiencek: Seems clearer if you pass `allowEditing` to `AliasTableRow` and move the actions into here. Unless there is a case where you need to customize the actions for specific rows.

root/components/Aliases/index.js
samj1912: Actually, I found various usages online of this way of organizing components when the components are made of various files. For eg - “How to better organize your React applications?” @alexmngn https://medium.com/@alexmngn/how-to-better-organize-your-react-applications-2fd3ea1920f1

This allows you to reference the Aliases folder as a component importing Aliases imports from the index file. What we can do however is move the component from index.js to its own file and simply import export from index.js. This allows us to keep the react embed names consistent. 'components/Aliases' as opposed to lowercase version of the same.

root/components/Aliases/AliasTableBody.js
samj1912: Yeah, I noticed that later, but I was too sleepy at the time. Will fix.

root/components/Aliases/index.js
mwiencek: Hmm, makes sense for "scenes" (I don't like that word, would much prefer "entry" or "pages") and components made up of various files. But I think I would not create a Component/index.js structure (as opposed to just Component.js) in cases where it's clearly a leaf component that has no dependencies.

I also generally don't like how quickly the folders become deeply with these very long paths for just a few  components. For example, according to that article your directory structure should be this, I think:
```
root/components/Aliases/index.js
root/components/Aliases/components/AliasTable/index.js
root/components/Aliases/components/AliasTable/components/AliasTableBody/index.js
root/components/Aliases/components/AliasTable/components/AliasTableBody/index.js
```

mwiencek: I can see the logic behind it though. One problem I have now with the webpack conversion is finding a way to bundle every entry point (scene/page) for the server without listing them all explicitly. If we used a simplified structure like this:

```
root/pages/Artist/index.js
root/pages/Artist/SomeComponentWhichIsntAPage.js
root/pages/Artist/someUtil.js
root/pages/Artist/Aliases/index.js
```

then it's not clear how to determine which files are the actual pages, and which are the supporting components. If we followed the article more closely, it would be clear (pages would always appear directly under a folder named pages/). What do you think?

samj1912: What's the final verdict on this then?

root/components/Aliases/AliasTableBody.js
```diff
@@ -0,0 +1,30 @@
...
+const {l} = require('../../static/scripts/common/i18n');
+
+type Props = {
+  +aliases: $ReadOnlyArray<AliasT>,
+  +allowEditing: boolean,
+  +entity: $Subtype<EntityT>,
+};
+
+
+const AliasTableBody = ({aliases, allowEditing, entity}: Props) => {
```
samj1912: I originally wanted to write this as {aliases, ...props}
but I don't know how to make flow work with that.

root/components/Aliases/index.js
mwiencek: I don't know, but let's keep what you have for now.

root/components/Aliases/AliasTableBody.js
mwiencek: Seems to only work if the props for `AliasTableRow` are inexact: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzACUBTAQw0IF4wAnEgRwFcBLegCgHJ6L9OBKANzp8ATwAOJMAAVaccbgCMYGgG8APqjBgA1GQBcYArRbYA5gBotugEaHjpy9Z0Z7+E+avqAvsNTBgGTkFACYwAFsmAjAbKVMSAA9eMHZsODANMAA6HLAffizUMUkg+VwwtWc7I3dHK20XNw8nX3QsPEIAYThw8RwSbGoU8UNZMpD+FQA+FOtxLJtdMHmMVCE2nGiE0eClFQyDME5OCxjDY9PXI85W9ujVMlOcrPoCb32EvzvCKDh0mgAPN1ev1Bhlnq98O9gFNhEA

root/components/Aliases/AliasTableRow.js
```diff
@@ -0,0 +1,59 @@
...
+    </td>
+    {alias.name === alias.sort_name
+      ? null
+      : <td>{isolateText(alias.sort_name)}</td>}
+    <td>{formatDate(alias.begin_date)}</td>
+    <td>{alias.ended ? alias.end_date ? formatDate(alias.end_date) : l('Unknown') : null}</td>
+    <td>{alias.typeName}</td>
+    <td>
+      {alias.locale ? locales[alias.locale] : null}
+      {alias.primary_for_locale ? l(' ({primary})', {__react: true, primary: <span className="comment">{l('primary')}</span>}) : null }
```
mwiencek: I still think it should use `bracketed`, but otherwise can we change the string to `{text}` so they share the same translation?

samj1912: Ah yes, forgot about this. Let me fix it.

samj1912: Done :)

root/components/Aliases/AliasTableRow.js
```diff
@@ -0,0 +1,60 @@
...
+const AliasTableRow = ({alias, allowEditing, entity, row}: Props) => (
+  <tr className={row} key={alias.id}>
+    <td colSpan={alias.name === alias.sort_name ? 2 : 1}>
+      {alias.editsPending ? <span className="mp">{isolateText(alias.name)}</span> : isolateText(alias.name)}
+    </td>
+    {alias.name === alias.sort_name
+      ? null
+      : <td>{isolateText(alias.sort_name)}</td>}
+    <td>{formatDate(alias.begin_date)}</td>
+    <td>{alias.ended ? alias.end_date ? formatDate(alias.end_date) : l('Unknown') : null}</td>
```
mwiencek: s/Unknown/[unknown]/ as the TT has it

root/components/Aliases/AliasTableRow.js
```diff
@@ -0,0 +1,60 @@
...
+    <td>{alias.typeName}</td>
```
mwiencek:
```JavaScript
lp_attributes(alias.typeName, 'alias_type')
```

mwiencek:
I noticed the date period types were wrong (they were as strings instead of nullable objects with year/month/day fields), but I already wrote some in https://github.com/metabrainz/musicbrainz-server/pull/624 so I cherry-picked the commits here and fixed things. I also added some line breaks to some very long lines.

mwiencek:
Fixed one more issue: AliasT.locale should be string | null.

sambhav:
@mwiencek can we merge this if it is okay? This will prolly causes some conflicts with some other already pending PRs.

PR #615, opened 2018-02-20
Use yarn instead of npm

samj1912:
npm5 has a bug that makes it remove git deps. on installing any new
packages. The only fix right now seems to be switching to yarn.


docker/templates/macros.m4
```diff
@@ -20,13 +20,13 @@ m4_define(`NODEJS_DEB', `nodejs_7.9.0-1nodesource1~xenial1_amd64.deb')
...
     `install_javascript',
     `m4_dnl
-COPY package.json package-lock.json ./
 RUN apt_install(``git'') && \
     cd /tmp && \
     curl -sLO https://deb.nodesource.com/node_7.x/pool/main/n/nodejs/NODEJS_DEB && \
     dpkg -i NODEJS_DEB && \
     cd - && \
-    sudo_mb(``npm install$1'')
+    sudo_mb(``npm install yarn'') && \
```
mwiencek: Needs `-g`

PR #618, opened 2018-02-21
Remove jsx-control-statements

mwiencek:
It can't be typed properly with Flow, so it just makes development harder and gets in our way of refactoring things.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -235,8 +235,8 @@ rules:
                                       assignment: parens-new-line,
                                       return: parens-new-line,
                                       arrow: parens-new-line,
-                                      condition: parens-new-line,
-                                      logical: parens-new-line,
+                                      condition: ignore,
+                                      logical: ignore,
                                       prop: ignore}]
 
   ##########################
--- a/root/components/forms.js
+++ b/root/components/forms.js
@@ -99,6 +99,10 @@ function subfieldErrors(field, accum = Immutable.List()) {
   return accum;
 }
 
+const buildErrorListItem = (error, index) => (
+  <li key={index}>{error}</li>
+);
+
 const FieldErrors = ({field}) => {
   if (!field) {
     return null;
@@ -110,9 +114,7 @@ const FieldErrors = ({field}) => {
   if (errors.size) {
     return (
       <ul className="errors">
-        <For each="error" of={errors.toArray()}>
-          <li>{error}</li>
-        </For>
+        {errors.toArray().map(buildErrorListItem)}
       </ul>
     );
   }
@@ -124,16 +126,6 @@ const FormRow = ({children, ...props}) => (
     {children}
   </div>
 );
-
-function selectOptions(options) {
-  return (
-    <For each="option" index="index" of={options}>
-      <option key={index} value={option.value}>
-        {option.label}
-      </option>
-    </For>
-  );
-}
 
 function getValue(field, options, allowEmpty) {
   if (field.value != null) {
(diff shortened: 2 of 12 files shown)


root/components/forms.js
```diff
@@ -99,6 +99,10 @@ function subfieldErrors(field, accum = Immutable.List()) {
   return accum;
 }
 
+const makeErrorListItem = (error, index) => (
```
mwiencek: In cases where the `map` callback depends only the array items, I factored out the callback function to avoid creating a new function/closure on each render.

root/components/forms.js
```diff
@@ -161,18 +165,11 @@ const SelectField = ({
...
-    <If condition={options[0].optgroup}>
-      <For each="optgroup" index="index" of={options}>
-        <optgroup key={index} label={optgroup.optgroup}>
-          {selectOptions(optgroup.options)}
-        </optgroup>
-      </For>
-    <Else />
-      {selectOptions(options)}
-    </If>
+    {allowEmpty &&
```
mwiencek: TBH, this is also much more readable to me with the plugin removed.

root/components/forms.js
```diff
@@ -191,7 +188,7 @@ const FormRowSelectList = ({
   <div className="row">
     <label>{addColon(label)}</label>
     <div className="form-row-select-list">
-      <For each="subfield" index="index" of={repeatable.field.toArray()}>
+      {repeatable.field.toArray().map((subfield, index) => (
```
mwiencek: I didn't factor out the `map` callback here, as done above, because it references other props in the closure.

root/components/forms.js
samj1912: Just a note -  we used buildFoo in the headers PR.

root/static/scripts/common/components/EntityLink.js
```diff
@@ -45,12 +45,9 @@ class EventDisambiguation extends React.Component {
...
     return (
       <Frag>
-        <If condition={dates}>
-          {bracketed(dates)}
-        </If>
-        <If condition={event.cancelled}>
-          <Comment className="cancelled" comment={l('cancelled')} />
-        </If>
+        {dates && bracketed(dates)}
+        {event.cancelled &&
```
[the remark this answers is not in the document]
mwiencek: I'll change them all to ternary expressions to be consistent.

When I first started this branch I was actually looking to replace everything with https://github.com/tc39/proposal-do-expressions

But the problem with do expressions is the same, no Flow support. :\

root/static/scripts/common/components/EntityLink.js
```diff
@@ -155,47 +152,50 @@ const EntityLink = (props = {}) => {
...
-        {' '}
-        [<a href={infoLink}>{l('info')}</a>]
-      </If>
-    </Frag>
-  );
+  const parts = [content];
+
+  if (showDisambiguation) {
+    if (entityType === 'event') {
+      parts.push(<EventDisambiguation event={entity} key={1} />);
```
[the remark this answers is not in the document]
mwiencek: Good idea, though I realized using an incremented number isn't a good idea to begin with. We should assign a unique key to each part, like "event_disambiguation" here, so that the keys are stable if some individual parts are added/removed.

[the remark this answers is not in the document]
mwiencek: (That's why I hardcoded the numbers, but strings would make the intent more clear.)

.eslintrc.yaml
```diff
@@ -235,8 +235,8 @@ rules:
                                       assignment: parens-new-line,
                                       return: parens-new-line,
                                       arrow: parens-new-line,
-                                      condition: parens-new-line,
```
mwiencek: Changed these because it would only allow this:
```
{foo
  ? (
      <div>
        <p></p>
      </div>
    )
  : (
      <div>
        <p></p>
      </div>
    )}
```

Instead of this:

```
{foo
  ? <div>
      <p></p>
    </div>

  : <div>
      <p></p>
    </div>}
```

I'm used the the latter style but can revert this if anyone strongly prefers the former.

PR #619, opened 2018-02-25
Add an Object.assign polyfill for IE11

mwiencek:
Fixes [MBS-9628](https://tickets.metabrainz.org/browse/MBS-9628) and probably many other issues in IE11.

Note: core-js is what babel-polyfill uses, but by using it directly we can include just the parts we need.

### DIFF
--- a/root/static/scripts/common.js
+++ b/root/static/scripts/common.js
@@ -1,3 +1,6 @@
+// IE 11 doesn't support Object.assign.
+require('core-js/modules/es6.object.assign');
+
 const DBDefs = require('./common/DBDefs');
 const MB = require('./common/MB');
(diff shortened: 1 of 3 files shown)


root/static/scripts/common.js
```diff
@@ -1,3 +1,6 @@
+// IE 11 doesn't support Object.assign.
+require('core-js/modules/es6.object.assign');
```
mwiencek: All it does is create a polyfill for `Object.assign` if it doesn't already exist. If it exists, it falls back to the native one. You can see the source code here: https://github.com/zloirock/core-js/blob/master/modules/es6.object.assign.js

PR #620, opened 2018-02-25
Remove TT header, embed React one

mwiencek:
Does what the title says, and also adds Flow types and performs some cleanup on the existing header components.


root/layout/components/BottomMenu.js
```diff
@@ -10,14 +14,16 @@ const {VARTIST_GID} = require('../../static/scripts/common/constants');
 const {l, lp} = require('../../static/scripts/common/i18n');
 
 function languageName(language, selected) {
-  let {id, native_language, native_territory} = language[1];
+  const id = language.id;
```
sambhav: Why not aliased destructuring?
```
{id:id , native_language: nativeLanguage} 
```

root/layout/components/BottomMenu.js
```diff
@@ -28,29 +34,38 @@ function languageName(language, selected) {
...
-      {$c.stash.server_languages.map(function (l, index) {
-        let inner = languageLink(l);
-
-        if (l[0] === $c.stash.current_language) {
-          inner = <strong>{inner}</strong>;
-        }
-
-        return <li key={index}>{inner}</li>;
-      })}
+      {$c.stash.server_languages && (
```
sambhav: `x ? y : null` ?

root/layout/components/BottomMenu.js
```diff
@@ -54,15 +55,17 @@ const LanguageMenu = () => (
       )}
     </span>
     <ul>
-      {$c.stash.server_languages.map(function (l, index) {
```
sambhav: Can we decouple this from $c.stash and instead pass server_languages as a prop?

mwiencek: We could, but root/layout/components/Head.js is still used with `React.embed`, so we have to keep `server_languages` in the stash either way.

It would also be slightly trickier than usual, since it's used by *every single page*, so every Catalyst controller action will need it added to `component_props`. We can do that by setting a default `component_props` in the "root" controller, but all the downstream actions will need to be changed to not overwrite `component_props` but to merge values into it.

mwiencek: See https://github.com/metabrainz/musicbrainz-server/pull/620/commits/cbca98e3309e0818f04d7a51f92e9bc4d7692020

PR #621, opened 2018-02-26
Webpack

mwiencek:


.babelrc
```diff
@@ -21,6 +20,9 @@
     "@babel/plugin-transform-react-jsx",
     "@babel/plugin-transform-react-constant-elements",
     "@babel/plugin-transform-runtime",
-    "@babel/plugin-proposal-class-properties"
-  ]
+    "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-syntax-dynamic-import",
+    "react-hot-loader/babel"
+  ],
+  "sourceType": "unambiguous"
```
mwiencek: `@babel/plugin-syntax-dynamic-import` is needed to allow the [dynamic import](https://webpack.js.org/guides/code-splitting/#dynamic-imports) syntax used for code splitting.

`react-hot-loader/babel` is https://github.com/gaearon/react-hot-loader - basically it lets us edit a component file and have the changes automagically reload in the browser. I only enabled it for the external links editor for now, just to test it out. (You'll need to be running `./script/compile_resources.sh --watch`.)

`"sourceType": "unambiguous"` is explained in e90bd3fa297b41292548b2bc1dd7dad69a2398bc.

docker/musicbrainz-website/consul-template-template-renderer.conf
```diff
@@ -7,7 +7,7 @@ template {
 exec {
     command = "start_template_renderer.sh"
     splay = "10s"
-    reload_signal = "SIGHUP"
+    reload_signal = "SIGTERM"
```
mwiencek: So, part of the necessary changes is that the template renderer is now built with Webpack. Everything it needs is bundled into a single file, including itself and all of the React components it can render (those have to be specified in root/server/components.js, so that Webpack can pick them up).

`SIGHUP` only spawns new workers. That's not really useful if the configuration (DBDefs) changes, because the configuration is compiled *into the bundle* that is executing. We have to give a `SIGTERM` now to rebundle the whole thing.

root/layout/components/HeaderLogo.js
```diff
@@ -0,0 +1,20 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+const HeaderLogo = () => (
+  <img
+    alt="MusicBrainz"
+    className="logo"
+    src={require('../../static/images/layout/header-logo.svg')}
```
mwiencek: Using `require` on an image file may look funny, since we weren't able to do this before. But Webpack has the concept of [loaders](https://webpack.js.org/concepts/loaders/) to deal with imports. We use the [file-loader](https://www.npmjs.com/package/file-loader) to handle image and CSS imports. So what this actually does is copy the image or CSS to our build dir, and return the *public URI* here. Quite nifty. You can find all of the necessary `file-loader` config is in webpack/moduleConfig.js.

For LESS files, we actually pipe them through `less-loader` (to convert them to CSS) first, then `file-loader` (to copy them and get the URI). You can see that in moduleConfig.js too.

mwiencek:
A `--watch` flag has been added to `./script/compile_resources.sh` to have Webpack automatically recompile client and server JS whenever a file is changed. Note that this starts a Webpack "dev server" to serve resources from in-memory. You'll need to make sure `STATIC_RESOURCES_LOCATION` is set to `//localhost:8080/static/build` in DBDefs to use this. (Port 8080 isn't configurable yet.)

root/static/scripts/common.js
```diff
@@ -1,3 +1,8 @@
+// Needed by root/release/cover_art_uploader.tt, which uses the
+// css_manifest TT macro that requires common.less to exist in
+// rev-manifest.json.
+require('../styles/common.less');
```
mwiencek: The `css_manifest` TT macro uses `Data::FileCache` under the hood, which tries to look things up in rev-manifest.json. This is to make sure it gets added there for the cover art uploader.

There's a `require('../../static/styles/common.less')` already present in root/layout/components/Head.js, but that does *not* add it to the manifest, because that file is only included in the server build, and the manifest is generated from the client build.

mwiencek: So note that for React-rendered pages, we don't need CSS or image files in rev-manifest.json at all - the URIs for those resources are got via `file-loader`, as in root/layout/components/Head.js. This is just a hack for an old TT page which expects CSS in the manifest.

mwiencek:
@brainzbot test this please

.babelrc
mwiencek: I've removed `react-hot-loader` from this PR to reduce the amount of things needing testing. (`--watch` mode still exists, but you'll have to reload the page manually.) We can always add it later.

docker/musicbrainz-website/consul-template-template-renderer.conf
mwiencek: This comment is no longer accurate, since I changed it to only bundle root/server/components.js and not the entire server. This is a more conservative change with less chance of causing deployment issues. (But it means we can't use ES modules or non-comment-Flow-types in `root/server*` or files required through there.)

root/server/components.js
```diff
@@ -0,0 +1,198 @@
...
+  'entity/Subscribers': require('../entity/Subscribers'),
+  'entity/Tags': require('../entity/Tags'),
+  'elections/Index': require('../elections/Index'),
+  'elections/Nominate': require('../elections/Nominate'),
+  'elections/NotFound': require('../elections/NotFound'),
+  'elections/Show': require('../elections/Show'),
+  'event/NotFound': require('../event/NotFound'),
+  'genre/List': require('../genre/List'),
+  'instrument/List': require('../instrument/List'),
+  'instrument/List': require('../instrument/List'),
```
yvanzo: There are some duplicates in these exports.

root/context.js
```diff
@@ -7,8 +7,10 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-import React from 'react';
-import type {ComponentType} from 'react';
+// NOTE: Don't convert to an ES module; this is used by root/server.js.
```
yvanzo: Additionally, ESLint rules should be changed somehow (via either local comment or configuration file) to https://eslint.org/docs/rules/#nodejs-and-commonjs

mwiencek: I don't think we use any of those ATM, but I added some `eslint-disable import/no-commonjs` comments to disable the current errors.

root/server/components.js
mwiencek: Whoops, fixed

mwiencek:
> `./script/compile_resources.sh --watch` works for me even without setting `STATIC_RESOURCES_LOCATION`. It doesn’t use port 8080 either.

Yeah, I made an edit to the comment at https://github.com/metabrainz/musicbrainz-server/pull/621#issuecomment-443474541 where I mentioned the `webpack-dev-server` was removed for now. So no fancy in-memory resource server at 8080 anymore, but `--watch` works well otherwise. `webpack-dev-server` is necessary if we want to use the React hot reload stuff later, where the components literally update on the page by themselves after you save a file.

> `root/server/components.js` is meh, is there any hope to get rid of it later on?

It'd definitely be possible to simplify things if we moved all "entry" components to specific subdir of root/, since there are easy ways to require all files under a specific dir (or matching a specific pattern) shown at https://webpack.js.org/guides/dependency-management/

PR #622, opened 2018-03-01
Replace hacky TT timeline code

mwiencek:
We were using some very wacky TT to generate JavaScript code for the client. This replaces all of that with actual JavaScript. The dynamic bits are fetched via ajax and constructed on the client.


mwiencek:
@yvanzo which version of FF? They are displayed for me in 52.7.2 ESR and 60.0b4. 🤔

PR #623, opened 2018-03-01
Convert entity sidebars to React

mwiencek:


mwiencek:
No longer WIP. All of the core entities are done now and I've split up the commits. We can do the cdstub, collection, and edit sidebars in a later PR.

root/layout/components/sidebar/SidebarBeginDate.js
```diff
@@ -0,0 +1,38 @@
...
+  +age?: [number, number, number] | null,
+  +entity: {...DatePeriodRoleT},
+  +label: string,
+|};
+
+const SidebarBeginDate = ({age, entity, label}: Props) => (
+  isDateEmpty(entity.begin_date) ? (
+    null
+  ) : (
+    <SidebarProperty className="begin-date" label={label}>
```
reosarevok: This does not add a space: "1995(23 years ago)"

root/static/scripts/common/components/TagEditor.js
```diff
@@ -380,17 +387,21 @@ class SidebarTagEditor extends TagEditor {
...
+            {this.props.more
+              // createTagRows uses tags as keys; \0 is simply used here to
+              // guarantee that there isn't any conflict.
+              ? <li key="\0:more">
+                  <a href={getTagsPath(this.props.entity)}>{l('more...')}</a>
+                </li>
+              : null}
+          </ul>
+        ) : (
+          <p>{lp('(none)', 'tag')}</p>
```
reosarevok: There's no vertical space between Tags and (none), unlike in production. There probably should be.

root/layout/components/sidebar/InstrumentSidebar.js
```diff
@@ -0,0 +1,89 @@
...
+            <MergeLink entity={instrument} />
+
+            <RemoveLink entity={instrument} />
+
+            <li className="separator" role="separator" />
+          </>
+        ) : null}
+      </EditLinks>
+
+      <SubscriptionLinks entity={instrument} />
```
reosarevok: Instruments can't be subscribed to.

root/layout/components/sidebar/InstrumentSidebar.js
```diff
@@ -0,0 +1,89 @@
...
+
+type Props = {|
+  +$c: CatalystContextT,
+  +instrument: InstrumentT,
+|};
+
+const InstrumentSidebar = ({$c, instrument}: Props) => {
+  return (
+    <div id="sidebar">
+      <CommonsImage
```
reosarevok: This does load the Commons images - not sure if that's blocked elsewhere in prod, but we're supposed not to show them for now :)

root/components/CritiqueBrainzLinks.js
```diff
@@ -0,0 +1,53 @@
...
+);
+
+type Props = {|
+  +releaseGroup: ReleaseGroupT,
+|};
+
+const CritiqueBrainzLinks = ({releaseGroup}: Props) => {
+  const reviewCount = releaseGroup.review_count;
+
+  if (reviewCount === 0) {
```
reosarevok: The way this is written looks like we want it to be displayed always, but with the "write a review" link if no reviews exist. Yet on the release sidebar, we only show it if releaseGroup.review_count is true. Which seemingly means it does not get shown unless there are already some reviews, and this section is apparently pointless. We should probably either remove this or always display it (I'd go for always display it).

This is already the case, so it's not introducing a regression or anything, but we might as well change it while we're at it?

root/layout/components/sidebar/InstrumentSidebar.js
mwiencek: Yeah, that's controlled by a DBDefs flag set in prod. You should still see them locally if you haven't set that flag.

mwiencek:
This also now converts the collection, edit, and cdstub sidebars, which were the only ones left.

root/edit/components/EditSidebar.js
```diff
@@ -0,0 +1,101 @@
...
+          <div className="edit-expiration">
+            <ExpirationTime date={edit.expires_time} user={$c.user} />
+          </div>
+        </SidebarProperty>
+      ) : (
+        <SidebarProperty className="" label={l('Closed:')}>
+          <div className="edit-expiration">
+            {edit.status === EDIT_STATUS_TOBEDELETED
+              ? l('<em>Cancelling</em>')
+              : formatUserDate(edit.close_time)}
```
reosarevok: This date is displaying in bold and without the time zone - possibly because you didn't add $c.user to it like on line 48 above?

mwiencek:
@brainzbot test this please

PR #624, opened 2018-03-03
Convert the area search results page to React

mwiencek:


root/types.js
```diff
@@ -108,12 +113,11 @@ declare type CommonsImageT = {|
...
-declare type CompoundFieldT<F: {+[string]: mixed}> = {|
-  +errors: $ReadOnlyArray<string>,
-  +field: F,
-  +has_errors: boolean,
-  +id: number,
-|};
+declare type CompoundFieldT<F: {+[string]: AnyFieldT}> =
+  & FieldRoleT
+  & {|
+      +field: F,
```
sambhav: Shouldn't this also be changed according to PR #629 ? Same for the other types declared below.

root/components/FieldErrors.js
```diff
@@ -0,0 +1,36 @@
...
+    return (
+      <ul className="errors">
+        {errors.map(buildErrorListItem)}
+      </ul>
+    );
+  }
+  return null;
+};
+
+export default FieldErrors;
```
sambhav: Why not export it with the definition?

root/components/FormRowSelectList.js
```diff
@@ -0,0 +1,81 @@
...
+  options,
+  removeClassName,
+  removeLabel,
+  repeatable,
+}: Props) => (
+  <div className="row">
+    <label>{addColon(label)}</label>
+    <div className="form-row-select-list">
+      {repeatable.field.map((subfield, index) => (
+        <div className="select-list-row" key={subfield.id}>
```
sambhav: Maybe refactor this out to a build* function too?

root/isrc/Index.js
```diff
@@ -56,7 +57,7 @@ const Index = ({isrcs, recordings}: PropsT) => {
           <tbody>
             {recordings.map((recording, index) => (
               <tr
-                className={(index + 1) % 2 ? 'odd' : 'even'}
+                className={loopParity(index)}
```
sambhav: Much needed! :tada:

root/components/WithPager.js
```diff
@@ -0,0 +1,52 @@
...
+
+type Props = {|
+  +children: ReactNode,
+  +pager: PagerT,
+  +query: string,
+  +search?: boolean,
+  +total?: boolean,
+|};
+
+const WithPager = ({
```
sambhav: `WithPager` doesn't really match the rest of our component names. Simply Pager should do, no?

root/types.js
```diff
@@ -268,6 +270,19 @@ declare type OptionTreeT =
...
 
+/*
+ * See http://search.cpan.org/~lbrocard/Data-Page-2.02/lib/Data/Page.pm
+ * Serialized in MusicBrainz::Server::TO_JSON.
+ */
+declare type PagerT = {|
+  +current_page: number,
+  +first_page: 1,
+  +last_page: number,
+  +next_page: number | null,
```
sambhav: Can we decide on some consistent repr. for nullable fields? either `?type` or `type | null`, I personally prefer the former

root/types.js
mwiencek: Yeah, will fix

root/components/FieldErrors.js
mwiencek: `export default {var,let,const}` isn't valid syntax, though I can change it a named function and do that.

root/components/WithPager.js
mwiencek: You may be right, I was just copying root/components/with-pager.tt

I guess the idea was that it's not really a pager, it's wrapping content and *adding* pagers above/below it? But I think pager is a fake word in either case. :)

root/types.js
mwiencek: Well, Perl's JSON module will always serialize `undef` as `null` (JSON doesn't have `undefined` like JavaScript does). And I haven't come across a case yet where I wanted to allow a property to be either null or undefined, so I tend to avoid `?type`.

root/components/FieldErrors.js
mwiencek: But not for components using an implicit return (can't do that with named functions). I think I'll keep them consistent.

root/components/FormRowSelectList.js
mwiencek: I tried but it got really messy because it uses of ton of variables in the outer scope, not just `subfield` and `index`.

lib/MusicBrainz/Server/ControllerUtils/JSON.pm
```diff
@@ -0,0 +1,32 @@
...
+
+our @EXPORT_OK = qw(
+    serialize_pager
+);
+
+sub serialize_pager {
+    my ($pager) = @_;
+
+    return {
+        current_page => $pager->current_page + 0,
```
mwiencek: The `+0` is needed so that Perl sees it as a number, and the JSON encoder outputs e.g. `3` instead of `"3"`.

lib/MusicBrainz/Server/ControllerUtils/JSON.pm
```diff
@@ -0,0 +1,32 @@
+package MusicBrainz::Server::ControllerUtils::JSON;
+
+use base 'Exporter';
+
+our @EXPORT_OK = qw(
+    serialize_pager
+);
+
+sub serialize_pager {
```
mwiencek: Wasn't sure of a good place to put this, so I created a new package for functions that convert external modules to JSON.

root/components/WithPager.js
mwiencek: Changed it to `PaginatedResults` if that sounds better.

mwiencek:
There were some bugs in the paginator and type name display with indexed search that I've fixed now.

root/search/components/SearchForm.js
```diff
@@ -0,0 +1,115 @@
...
+            field={form.field.query}
+            label={l('Query:')}
+            required
+          />
+          <FormRowSelect
+            field={form.field.type}
+            label={l('Type:')}
+            onChange={noop}
+            options={typeOptions}
+          />
```
yvanzo: Type field is actually required from https://github.com/mwiencek/musicbrainz-server/blob/46badf4b4df5c2aaf49dc2567759ae847e63532c/lib/MusicBrainz/Server/Form/Search/Search.pm#L16

root/components/FormRowSelect.js
```diff
@@ -0,0 +1,43 @@
...
+  options,
+  required = false,
+}: Props) => (
+  <FormRow>
+    <FormLabel forField={field} label={label} required={required} />
+    <SelectField
+      field={field}
+      onChange={onChange}
+      options={options}
+    />
```
yvanzo: Either add `allowEmpty={!required}` here or optionally propagate `allowEmpty` from caller (and set it for search form fields `type` and `limit`).

yvanzo: Indeed, maybe `type` should not be `required` in the first place, and have a default value in Moose, even though it will be present in the POST anyway.

mwiencek:
@brainzbot test this please

PR #625, opened 2018-03-04
Convert election pages to react

sambhav:


root/components/ElectionTable/ElectionTableRows.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+  return false;
+};
+
+const ElectionTableRows = ({elections}) => (
+  elections.map(election => (
+    <tr className="odd" key={election.id}>
+      <td><EditorLink editor={election.candidate} /></td>
+      <td>{lp(election.shortStatusName, 'autoeditor election status (short)')}</td>
+      <td>{formatUserDate($c.user, election.proposedTime)}</td>
```
sambhav: I am having a bit of a problem with this function. Not sure if I am using this correctly. @mwiencek any help?

sambhav:
@brainzbot  ok to test.

sambhav:
@brainzbot  ok to test.

root/components/ElectionTable/ElectionTableRows.js
sambhav: NVM, fixed it.

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -279,6 +279,7 @@ around TO_JSON => sub {
         is_location_editor => boolean_to_json($self->is_location_editor),
         is_relationship_editor => boolean_to_json($self->is_relationship_editor),
         is_wiki_transcluder => boolean_to_json($self->is_wiki_transcluder),
+        isAutoEditor => boolean_to_json($self->is_auto_editor),
```
sambhav: Not sure which convention we are following. Snake or camel case?

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -612,6 +613,17 @@ sub boolean_from_json {
     return $bool ? 1 : 0;
 }
 
+sub datetime_to_iso8601 {
+    my $date = shift;
+
+    if ($date) {
```
mwiencek: I would add this at the top instead:
```Perl
return unless defined $date;
```
Assuming it makes sense to allow passing undef into here, which maybe it does. But if someone were to pass 0 or '', we should probably let it throw an exception and not return the incorrect data back.

root/components/ElectionTable/ElectionTableRows.js
```diff
@@ -0,0 +1,44 @@
...
+
+const React = require('react');
+const {l, lp} = require('../../static/scripts/common/i18n');
+const EditorLink = require('../../static/scripts/common/components/EditorLink');
+const formatUserDate = require('../../utility/formatUserDate');
+const {votesVisible} = require('../../utility/voting');
+
+const ElectionTableRows = ({elections}) => (
+  elections.map(election => (
+    <tr className="odd" key={election.id}>
```
mwiencek: `className` should probably alternate between "even" and "odd".

root/utility/votesVisible.js
```diff
@@ -0,0 +1,23 @@
...
+
+const votesVisible = (election, user) => {
+  const involvedUser = user && (
+    election.proposer.id === user.id ||
+    election.candidate.id === user.id ||
+    (election.seconder1 && election.seconder1.id === user.id) ||
+    (election.seconder2 && election.seconder2.id === user.id)
+  );
+
+  if (election.isClosed || involvedUser) {
```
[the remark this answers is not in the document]
sambhav: Isn't that what is being checked first?

root/utility/ExpirationDate.js
```diff
@@ -0,0 +1,45 @@
+/*
```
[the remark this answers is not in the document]
sambhav: Yeah, will fix that. I wrote most of the code half asleep hoping to refactor it once I got everything working.

root/utility/votesVisible.js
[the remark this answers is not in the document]
mwiencek: I mean in order to calculate `involvedUser` you have to perform a bunch of comparisons and property lookups, but it isn't necessary to calculate `involvedUser` at all if `election.isClosed` is true - you can exit immediately.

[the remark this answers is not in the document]
sambhav: Oh sorry. Not sure how this sneaked in. I replaced this with the voting util file below. This isn't used anywhere.

lib/MusicBrainz/Server/Data/Utils.pm
sambhav: This seems to mess up with the output somehow.

mwiencek: That would be strange...did you add it below the `my $date = shift;`?

sambhav: Yup. It somehow messes up the `TO_JSON` sub and causes keys to become values. and messes with the output.

mwiencek: Ah, that's because when you call a function inside a hash definition, like `{ foo => datetime_to_iso8601($self->foo), bar => 1 }`, it calls `datetime_to_iso8601` in *list* context as opposed to *scalar* context. The bare `return` inside `datetime_to_iso8601` returns an empty list in list context, resulting in `{ foo => (), bar => 1 }` which is the same as `{ 'foo', 'bar', 1 }` - that's why the keys become values. (Tip: A `=>` is the same as a comma, except it automatically quotes the key for you.)

There are two solutions:

1. Change the return to `return undef unless defined $date;` (note the added `undef`), so that it returns `(undef,)` in list context instead.
2. Alternatively, call the function in scalar context outside of the hash definition. This can be done by assigning the result of the function to a scalar variable.

lib/MusicBrainz/Server/Entity/AutoEditorElection.pm
```diff
@@ -185,6 +186,30 @@ sub current_expiration_time
...
+        expiration_time => datetime_to_iso8601($self->current_expiration_time),
+        is_closed => boolean_to_json($self->is_closed),
+        is_open => boolean_to_json($self->is_open),
+        is_pending => boolean_to_json($self->is_pending),
+        no_votes => $self->no_votes,
+        proposer => $self->proposer,
+        propose_time => datetime_to_iso8601($self->propose_time),
+        open_time => datetime_to_iso8601($self->open_time),
+        seconder1 => $self->seconder_1,
+        seconder2 => $self->seconder_2,
```
mwiencek: s/seconder1/seconder_1/
s/seconder2/seconder_2/

lib/MusicBrainz/Server/Entity/AutoEditorElection.pm
```diff
@@ -185,6 +186,30 @@ sub current_expiration_time
...
 }
 
+around TO_JSON => sub {
+    my ($orig, $self) = @_;
+
+    return {
+        %{ $self->$orig },
+        candidate => $self->candidate,
+        close_time => datetime_to_iso8601($self->close_time),
+        expiration_time => datetime_to_iso8601($self->current_expiration_time),
```
mwiencek: s/expiration_time/current_expiration_time/

root/elections/ElectionVoting.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const ElectionVoting = ({election, user}: PropsT) => {
+  let message = l('To find out if you can vote for this candidate, please {url|log in}.',
+    {__react: true, url: '/login'});
+  if (user) {
+    if (election.is_closed) {
+      message = l('Voting is closed.');
+    } else if (election.is_pending) {
+      message = l('Voting is not yet open. If you would like to support this candidate, you can second their nomination. If you do not support this candidate, please note that you cannot cast a "No" vote (or abstain) until two seconders have been found.');
```
mwiencek: You changed the order of the elsifs here compared to the original template, so now it will show the user this message (suggesting they can second the nomination) even if they aren't an auto-editor.

root/elections/Index.js
```diff
@@ -0,0 +1,23 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+const React = require('react');
+const Layout = require('../layout');
+const {l} = require('../static/scripts/common/i18n');
+const ElectionTable = require('./ElectionTable');
+
+
+const Index = () => (
```
mwiencek: (In the interest of not adding more uses of the global `$c`) can we:
* make this component accept just the `elections` as a prop
* send the elections to this component by setting the `component_props` stash item in the controller
* remove `elections` from the stash so it isn't serialized twice

root/elections/Show.js
```diff
@@ -0,0 +1,39 @@
...
+
+const React = require('react');
+const Layout = require('../layout');
+const {l} = require('../static/scripts/common/i18n');
+const ElectionDetails = require('./ElectionDetails');
+const ElectionVotes = require('./ElectionVotes');
+const ElectionVoting = require('./ElectionVoting');
+
+const Show = () => {
+  const election = $c.stash.election;
```
mwiencek: Same comment as I left on root/elections/Index.js about sending this as a prop instead (but don't worry about `$c.user`).

root/utility/voting.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export function isInvolved(election: AutoEditorElectionT, user: ?EditorT) {
+  return !!user && (
+    election.proposer.id === user.id ||
+    election.candidate.id === user.id ||
+    (election.seconder1 && election.seconder1.id === user.id) ||
+    (election.seconder2 && election.seconder2.id === user.id)
+  );
+};
```
mwiencek: Added https://github.com/metabrainz/musicbrainz-server/commit/ea051f8c39fb08679f6d1cede75403107d8160d1 ;)

(There's no case where a semicolon is ever needed after a function declaration, only function expressions.)

lib/MusicBrainz/Server/Data/Utils.pm
sambhav: Wouldn't it be equivalent to `return $date` in case $date in undefined?

root/components/ExpirationDate.js
```diff
@@ -0,0 +1,52 @@
...
+  +date: string,
+  +user?: EditorT,
+};
+
+const ExpirationDate = ({date, user}: PropsT) => {
+  const dateMoment = moment(date);
+  const userDate = formatUserDate(user, date);
+  let durationString = null;
+  let result = l('Already expired');
+  if (dateMoment.isAfter()) {
```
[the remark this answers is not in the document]
sambhav: https://momentjs.com/docs/#/query/is-after/ If nothing is specified it takes the current time.

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: Yes, but `undef` is not the only falsy value that `$date` can be. Anyway it doesn't matter that much.

root/components/ExpirationDate.js
```diff
@@ -0,0 +1,52 @@
...
+        {num: duration.hours()});
+    } else {
+      durationString = ln('{num} minute',
+        '{num} minutes',
+        duration.minutes(),
+        {num: duration.minutes()});
+    }
+  }
+  if (durationString) {
+    result = l('Expires in {time}', {__react: true, time: <span className="tooltip" title={userDate}>{durationString}</span>});
```
mwiencek: For some reason, the `title` here is now just `%c` instead of the actual date.

root/utility/voting.js
```diff
@@ -0,0 +1,36 @@
...
+export function isInvolved(election: AutoEditorElectionT, user: ?EditorT) {
+  return !!user && (
+    election.proposer.id === user.id ||
+    election.candidate.id === user.id ||
+    (election.seconder_1 && election.seconder_1.id === user.id) ||
+    (election.seconder_2 && election.seconder_2.id === user.id)
+  );
+}
+
+export function votesVisible(election: AutoEditorElectionT, user: ?EditorT) {
```
mwiencek: This isn't equivalent to the old `can_see_vote_count`, because it returns `true` if the election is pending (still awaiting seconders). So the templates now show a vote tally even if the vote didn't start yet.

root/elections/ElectionTable/ElectionTableRows.js
```diff
@@ -0,0 +1,49 @@
...
+  +elections: $ReadOnlyArray<AutoEditorElectionT>,
+};
+
+const ElectionTableRows = ({elections}: PropsT) => (
+  elections.map((election, index) => (
+    <tr className={index % 2 ? 'even' : 'odd'} key={election.id}>
+      <td><EditorLink editor={election.candidate} /></td>
+      <td>{lp(election.status_name_short, 'autoeditor election status (short)')}</td>
+      <td>{formatUserDate($c.user, election.propose_time)}</td>
+      <td>{election.close_time ? formatUserDate($c.user, election.close_time) : '-'}</td>
```
mwiencek: Same here, just displays `%c` for me. (I have a custom date format set in /account/preferences if that's related.)

sambhav: This is because moment-strftime does not support %c or %x/%X. How do you want to deal with it?

The easiest way I can see is to add a json comversion sub to convert these formats to moment compatible formats.

mwiencek: Not sure I see a reason moment-strftime can't support those, since moment has localized format options. If so I'd fork moment-strftime and submit a patch upstream, then just use the fork in package.json for now.

mwiencek: I actually ran into the same issue in the area search results PR (displays "%c" for the last updated time). So if you'd rather I look into it, I can tomorrow, but otherwise feel free. :) If I'm wrong and moment-strftime can't be easily patched, then a conversion method sounds fine, though I'd probably do the conversion in the JavaScript instead.

mwiencek:
LGTM. I fixed some bugs related to the "No elections found" message, switch to ES imports and made the new types use object spreads.

sambhav:
@mwiencek maybe we should consider rendering the profile preferences for date/time on the front end too - just passing the format in the backend to maintain consistency with moment-strftime

PR #626, opened 2018-03-11
 Move component props from stash to stash.component_props

sambhav:


lib/MusicBrainz/Server/Controller/ISRC.pm
```diff
@@ -35,9 +35,12 @@ sub show : Chained('load') PathPart('')
     my @recordings = sort_by { $_->name } $c->model('Recording')->load(@$isrcs);
     $c->model('ArtistCredit')->load(@recordings);
     $c->stash(
-        recordings => \@recordings,
         current_view => 'Node',
         component_path => 'isrc/Index.js',
+        component_props => {
+            isrcs => $isrcs,
```
mwiencek: Ok, they're already in the stash so normally I would say not to serialize them twice, but the isrcs should be a small list.

sambhav: should I remove them from the stash earlier on then?

lib/MusicBrainz/Server/Controller/Instrument.pm
```diff
@@ -157,10 +157,12 @@ sub list : Path('/instruments') Args(0) {
     }
 
     $c->stash(
-        instruments_by_type => $entities,
-        instrument_types => \@types,
         current_view => 'Node',
         component_path => 'instrument/List',
+        component_props => {
```
mwiencek: It's a bit unsatisfactory to create broken commits like this - if someone bisecting lands on this commit, the page will be broken for them because the component isn't changed to understand this until b137a99a2a11c486bd1050f9c8f2f9a45925760e from what I can tell.

lib/MusicBrainz/Server/Controller/ISRC.pm
mwiencek: Ah, perhaps. I thought other actions might be using them since they are set in `_load`, but it looks like this is the only action in this controller.

lib/MusicBrainz/Server/Controller/Instrument.pm
sambhav: Fixed.

lib/MusicBrainz/Server/Controller/ISRC.pm
sambhav: Yup, which is why I didn't touch it.

root/instrument/List.js
```diff
@@ -15,59 +16,62 @@ const {l} = require('../static/scripts/common/i18n');
 const {lp_attributes} = require('../static/scripts/common/i18n/attributes');
 const {l_instrument_descriptions} = require('../static/scripts/common/i18n/instrument_descriptions');
 
+type PropsT = {
+  instrument_types: $ReadOnlyArray<InstrumentTypeT>,
+  instruments_by_type: {|
+    +[number]: $ReadOnlyArray<InstrumentT>,
+    unknown: $ReadOnlyArray<InstrumentT>,
```
mwiencek: This is now missing the `+` on all the props expect `[number]` (did you say some eslint rule was removing those? we should disable it)

Making `PropsT` an exact object type would be nice too: `{| ... |}`

PR #628, opened 2018-03-11
Add guidelines and instructions for porting to React

sambhav:


HACKING.md
```diff
@@ -226,6 +226,60 @@ specify its name in an argument:
...
+        my ($orig, $self) = @_;
+        return {
+            %{ $self->$orig },
+            prop_name => covert_to_json(prop_val)
+        };
+   };
+   ```
+   where the `convert_to_json` is a function that converts `prop_val` to its appropriate
+   JSON representation. If `prop_val` is an entity that has a `TO_JSON` sub defined, it will
+   most probably be called by the Server and you can simply pass `prop_val`.
```
mwiencek: We should probably remove "most probably" since it should be called, and that makes it sounds more mysterious than it is. :P

HACKING.md
```diff
@@ -226,6 +226,60 @@ specify its name in an argument:
...
+           component_path => 'relative/path/to/component/from/root',
+           component_props => {prop_name => prop_value}
+       );
+       ```
+    2. In case you are embedding a React component in a TT page use
+        ```perl
+        [%~ React.embed(c, 'relative/path/to/component/from/root', { prop_name => prop_val }) ~%]
+        ```
+3. You can access the CatalystContext via the variable `$c`.
+4. To communicate between Perl and Node Server(which renders React components for us), you need to
```
mwiencek: "between the Perl and Node servers (which" (+space after servers)

HACKING.md
```diff
@@ -226,6 +226,60 @@ specify its name in an argument:
...
+
+ * Ported server-side common react components will reside in `root/components/`.
+   This will generally include ported subs from `root/common-macros.tt` and components
+   common across multiple pages.
+ * Similarly any client side components will reside in `root/static/scripts/common/components`.
+ * Any server-side utility file will go in `root/utility`.
+ * All components must be type-annotated. We use flow for static type checks. You can find more
+   about it at [flow docs](https://flow.org/en/docs/).
+ * Global types are defined in `root/types.js`. They can be used without imports.
+ * Make sure your JS files conform to our ESlint setup.
```
mwiencek: Not sure we explain how to run eslint anywhere yet either.

mwiencek:
This looks very good. 👍

PR #629, opened 2018-03-17
Object type spreads

mwiencek:
The previous code is wrong because an object can't simultaneously satisfy two distinct, *exact* object types.

The correct way is using object type spreads as described here: facebook/flow#2626

I'm not sure why things currently type-check.


mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

PR #630, opened 2018-03-17
EntityLink cleanup

mwiencek:
The EntityLink component is used by a ton of pages and components, so it'd be very beneficial to have types for it to ensure it's being used correctly (and is implemented correctly).

I added types and performed some misc. fixes/cleanups.


root/static/scripts/common/components/EntityLink.js
```diff
@@ -32,96 +36,109 @@ const DeletedLink = ({name, allowNew}) => {
...
   }
 
-  if (entityType === 'artist' && !nonEmpty(hover)) {
+  if (entity.entityType === 'artist' && !nonEmpty(hover)) {
     hover = entity.sort_name + bracketed(comment);
   }
 
-  if (entityType === 'artist' || entityType === 'instrument') {
-    content = content || entity.l_name;
+  if (entity.entityType === 'area') {
```
mwiencek: See 75252bf82a090f255f1ad3fb7bbbf36add5302ad. I only discovered this bug because of the added types (Flow complained about `l_name` not existing, which caused me to dig deeper and see that the condition was also wrong). So they're already helping. :)

root/static/scripts/common/components/EntityLink.js
```diff
@@ -32,96 +36,109 @@ const DeletedLink = ({name, allowNew}) => {
...
+  +showDeleted?: boolean,
+  +showDisambiguation?: boolean,
+  +subPath?: string,
+
+  // ...anchorProps
+  href?: string,
+  title?: string,
+  +target?: '_blank',
+};
+/* eslint-enable sort-keys, flowtype/sort-keys */
```
mwiencek: I wanted the keys that are only supposed to exist in the `anchorProps` spread to be grouped together, so I disabled those eslint rules for this block.

root/static/scripts/common/components/EntityLink.js
```diff
@@ -32,96 +36,109 @@ const DeletedLink = ({name, allowNew}) => {
...
-  const entityType = entity.entityType;
-  const comment = ko.unwrap(entity.comment);
+  const comment = entity.comment ? ko.unwrap(entity.comment) : '';
 
   if (showDisambiguation === undefined) {
     showDisambiguation = !hasCustomContent;
   }
 
-  if (entityType === 'artist' && !nonEmpty(hover)) {
+  if (entity.entityType === 'artist' && !nonEmpty(hover)) {
```
mwiencek: Since `CoreEntityT` is now a [disjoint union](https://flow.org/en/docs/types/unions/#toc-disjoint-unions), the `entityType` check allows Flow to determine that the entity must be an `ArtistT` inside this block, allowing the `entity.sort_name` access.

But for some reason this feature only works if you access `entity.entityType` directly in the conditional, rather than caching it in a variable as we were doing before.

root/static/scripts/common/components/EntityLink.js
```diff
@@ -35,7 +37,11 @@ const DeletedLink = ({name, allowNew}) => {
 const Comment = ({className, comment}) => (
   <Frag>
     {' '}
-    <span className={className}>({isolateText(comment)})</span>
+    <span className={className}>
+      {'('}
```
sambhav: Why not `bracketed` ?

root/static/scripts/common/components/EntityLink.js
```diff
@@ -72,16 +78,17 @@ const AreaDisambiguation = ({area}: {|+area: AreaT|}) => {
...
   return <Comment className="historical" comment={comment} />;
-}
+};
 
 const NoInfoURL = ({url, allowNew}) => (
   <Frag>
     <a href={url}>{url}</a>
     {' '}
-    <DeletedLink name={'[' + l('info') + ']'} allowNew={allowNew} />
+    <DeletedLink allowNew={allowNew} name={'[' + l('info') + ']'} />
```
sambhav: Maybe we can modify bracketed to do all 3 kinds of brackets depending on flags passed and default to round?

root/static/scripts/common/components/EntityLink.js
```diff
@@ -209,7 +224,7 @@ const EntityLink = ({
   }
 
   if (infoLink) {
-    parts.push(' [', <a href={infoLink} key="info">{l('info')}</a>, ']')
+    parts.push(' [', <a href={infoLink} key="info">{l('info')}</a>, ']');
```
sambhav: this can also be replaced by the refactored bracketed util?

mwiencek:
@sambhav Agreed about using bracketed, though the result isn't 100% pretty: ab7c48ac97fdf03ed9f5e8fb6737d1255534d701

About removing the space from the strings as mentioned in the commit: I'm not sure if it's necessary for some languages to control the presence of the space, and if that's the reason it was included, or just because it was convenient to have because most callers want the space.

yvanzo:
Yes, this space has to be moved out of this function eventually.

It happens that bracketed text always comes after another text, thus there always is a space before. IMHO, it is overfactorization. I don’t think it is necessary to any language, at least it is not useful to any of currently supported language, not to mention it does not support LTR languages.

mwiencek:
@yvanzo Agreed, I removed the leading space from the JS version of `bracketed` in 484ae622cb769b532f772351a75f3662a7f1daa7

root/static/scripts/common/components/EntityLink.js
yvanzo: Is that limitation documented or reported?

mwiencek: I found https://github.com/facebook/flow/issues/3932 and https://github.com/facebook/flow/issues/4828 (though they're related to destructuring the values more specifically, I guess the same logic would apply).

It might also be related to [refinement invalidation](https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations) in that Flow can't know if `entity.entityType` has changed value since it was stored in the constant...so it may be a way to enforce soundness.

mwiencek:
Yes, good call. Thanks!

PR #631, opened 2018-03-19
MBS-9658: /instruments page breaks if a new instrument type is added but not used

mwiencek:
Handles cases where `instrumentsByType` returns undefined for unused types.

I changed the Flow type to indicate the key can be unset. You can't do `[number]?` as I thought for indexer props, but `?Array<InstrumentT>` should be equivalent for our needs.

### DIFF
--- a/root/types.js
+++ b/root/types.js
@@ -19,8 +19,8 @@ type CatalystContextT = {
 
 type CatalystStashT = {
   instruments_by_type?: {
-    [number]: Array<InstrumentT>,
-    unknown: Array<InstrumentT>,
+    [number]: ?Array<InstrumentT>,
+    unknown: ?Array<InstrumentT>,
   },
   instrument_types?: Array<InstrumentTypeT>,
 };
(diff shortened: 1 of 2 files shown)


root/types.js
```diff
@@ -19,7 +19,7 @@ type CatalystContextT = {
 
 type CatalystStashT = {
   instruments_by_type?: {
-    [number]: Array<InstrumentT>,
+    [number]: ?Array<InstrumentT>,
     unknown: Array<InstrumentT>,
```
yvanzo: The above change should apply to `unknown` as well, to cover the idyllic case where the type of every instrument is known.

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
Tests were failing because of the outdated chromedriver in the production branch, I think. Cherry-picked https://github.com/metabrainz/musicbrainz-server/commit/cd4388fa42d1a640fbcc7e0b56d2f539f310d8ac from master, so we'll see if they pass now.

PR #635, opened 2018-03-28
Remove CATALYST_DEBUG views during Selenium tests

mwiencek:


mwiencek:
Tested the changes with `CATALYST_DEBUG` set to 1, and the Selenium tests then pass.

PR #639, opened 2018-04-10
MBS-9672: Prevent alphabetical sorting of search options with ESLint & Vertically align values

yvanzo:
[MBS-9672](https://tickets.metabrainz.org/browse/MBS-9672) is already fixed in `beta` with https://github.com/metabrainz/musicbrainz-server/commit/caec7c46a6a3fdadd2ab00e4b2d4250c17920b41 (cherry-picked to `master` as https://github.com/metabrainz/musicbrainz-server/commit/54579586754c51c92614c86488a8c87739a50c37).

This display bug has been inadvertently introduced by running `eslint`. This patch:

1. disables ESLint `sort-keys` rule for search type options,
2. sets ESLint `key-spacing` rule to vertically align values in object literal props.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -85,7 +85,7 @@ rules:
   implicit-arrow-linebreak: [warn, beside]
   indent: [warn, 2, {SwitchCase: 1}]
   jsx-quotes: [warn, prefer-double]
-  key-spacing: [warn, {beforeColon: false, afterColon: true}]
+  key-spacing: [warn, {mode: minimum}]
   keyword-spacing: [warn, {before: true, after: true}]
   linebreak-style: [warn, unix]
   lines-between-class-members: [warn, always]
--- a/root/layout/components/Search.js
+++ b/root/layout/components/Search.js
@@ -11,48 +11,57 @@ import React from 'react';
 
 import * as manifest from '../../static/manifest';
 import * as DBDefs from '../../static/scripts/common/DBDefs';
-import {compare, l, lp} from '../../static/scripts/common/i18n';
+import {compare, l, lp, N_l, N_lp} from '../../static/scripts/common/i18n';
 
 const TYPE_OPTION_GROUPS = [
   {
-    artist:        l('Artist'),
+    artist:        N_l('Artist'),
   },
   { // musical production
-    event:         l('Event'),
-    recording:     l('Recording'),
-    release:       l('Release'),
-    release_group: l('Release Group'),
-    series:        lp('Series', 'singular'),
-    work:          l('Work'),
+    event:         N_l('Event'),
+    recording:     N_l('Recording'),
+    release:       N_l('Release'),
+    release_group: N_l('Release Group'),
+    series:        N_lp('Series', 'singular'),
+    work:          N_l('Work'),
   },
   { // other core entities
-    area:          l('Area'),
-    instrument:    l('Instrument'),
-    label:         l('Label'),
-    place:         l('Place'),
+    area:          N_l('Area'),
+    instrument:    N_l('Instrument'),
+    label:         N_l('Label'),
+    place:         N_l('Place'),
   },
   { // derived data
-    annotation:    l('Annotation'),
-    tag:           lp('Tag', 'noun'),
+    annotation:    N_l('Annotation'),
+    tag:           N_lp('Tag', 'noun'),
   },
   {
-    cdstub:        l('CD Stub'),
+    cdstub:        N_l('CD Stub'),
   },
   {
-    editor:        l('Editor'),
+    editor:        N_l('Editor'),
   },
   {
-    doc:           DBDefs.GOOGLE_CUSTOM_SEARCH ? l('Documentation') : null,
+    doc:           DBDefs.GOOGLE_CUSTOM_SEARCH ? N_l('Documentation') : null,
   },
 ];
+
+function localizedTypeOption(group, key) {
+  return (key === 'series' || key === 'tag') ? lp(group[key])
+    : (key === 'doc' && group[key] === null) ? null
+      : l(group[key]);
+}
 
 const searchOptions = (
   <select id="headerid-type" name="type">
     {TYPE_OPTION_GROUPS.map(<TogT: {}>(group: TogT, groupIndex) => (
       Object.keys(group).sort(function (a, b) {
-        return compare(group[a], group[b]);
+        return compare(
+          localizedTypeOption(group, a),
+          localizedTypeOption(group, b),
+        );
       }).map(function (key, index) {
-        const text = group[key];
+        const text = localizedTypeOption(group, key);
         if (!text) {
           return null;
         }
--- a/root/layout/components/Search.js
+++ b/root/layout/components/Search.js
@@ -13,24 +13,26 @@ import * as manifest from '../../static/manifest';
 import * as DBDefs from '../../static/scripts/common/DBDefs';
 import {l, lp} from '../../static/scripts/common/i18n';
 
+/* eslint-disable sort-keys, flowtype/sort-keys */
 const TYPE_OPTIONS = {
-  artist:         l('Artist'),
-  release_group:  l('Release Group'),
-  release:        l('Release'),
-  recording:      l('Recording'),
-  work:           l('Work'),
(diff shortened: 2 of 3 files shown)


.eslintrc.yaml
```diff
@@ -85,7 +85,7 @@ rules:
   implicit-arrow-linebreak: [warn, beside]
   indent: [warn, 2, {SwitchCase: 1}]
   jsx-quotes: [warn, prefer-double]
-  key-spacing: [warn, {beforeColon: false, afterColon: true}]
+  key-spacing: [warn, {align: value}]
```
mwiencek: Does it require keys to be aligned everywhere, or just allow it? I wouldn't very much like enforcing alignment everywhere, since you often have to adjust the spacing for an entire object when you add/remove a key, and that increases diff size/noise for not much readability gain (though it's nice for large objects like `TYPE_OPTIONS`).

root/layout/components/Search.js
```diff
@@ -13,24 +13,26 @@ import * as manifest from '../../static/manifest';
...
-  place:          l('Place'),
-  annotation:     l('Annotation'),
-  cdstub:         l('CD Stub'),
-  editor:         l('Editor'),
-  tag:            lp('Tag', 'noun'),
-  instrument:     l('Instrument'),
-  series:         lp('Series', 'singular'),
-  event:          l('Event'),
-  doc:            DBDefs.GOOGLE_CUSTOM_SEARCH ? l('Documentation') : null,
+  artist:        l('Artist'),
```
mwiencek: The iteration order for an object is technically arbitrary and undefined, it just so happens that most engines respect the insertion order. I wouldn't rely on it though (as an anecdote, a lot of our Perl code broke when we updated to 5.18 for the same reason). We should be using a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

mwiencek:
While we're on the topic of the ordering of this list, it seems non-controversial to me that all of the core entities should be next to each other at least (instrument, event, and series should not be below annotation, cdstub, editor, ...).

.eslintrc.yaml
yvanzo: Agreed, it shouldn’t enforce it. I relaxed `key-spacing` rule with [minimum mode](https://eslint.org/docs/rules/key-spacing#mode) in https://github.com/metabrainz/musicbrainz-server/pull/639/commits/847229904fb91677849eb0906b853a231551e4a7.

root/layout/components/Search.js
yvanzo: Agreed, I replaced it with an array of option groups, and, within each group, sorted options with localized `compare` on titles in https://github.com/metabrainz/musicbrainz-server/pull/639/commits/bb2532d239eb845befa8eca1527ec340f9013619.

yvanzo:
Agreed, https://github.com/metabrainz/musicbrainz-server/pull/639/commits/bb2532d239eb845befa8eca1527ec340f9013619 splits core entity types into three groups: artist, their musical productions, and others.

root/layout/components/Search.js
```diff
@@ -9,38 +9,59 @@
...
-  editor:         l('Editor'),
-  tag:            lp('Tag', 'noun'),
-  instrument:     l('Instrument'),
-  series:         lp('Series', 'singular'),
-  event:          l('Event'),
-  doc:            DBDefs.GOOGLE_CUSTOM_SEARCH ? l('Documentation') : null,
-};
+const TYPE_OPTION_GROUPS = [
+  {
+    artist:        l('Artist'),
```
mwiencek: Oops, I noticed one more issue which isn't strictly related to MBS-9672, but maybe you'd like to fix it as part of this PR. :) https://tickets.metabrainz.org/browse/MBS-9673

We have `N_l` and `N_lp` functions in i18n.js that you can import, but they'll be slightly annoying to use due to series...you'd need `{key === 'series' ? lp(text) : l(text)}` in `searchOptions` below since the output of `N_l` has to be fed into `l`, and likewise for `N_lp` -> `lp`.

Another alternative is just turning all of the strings in `TYPE_OPTION_GROUPS` into functions like `artist: () => l('Artist')` so we can delay their evaluation.

mwiencek: A cool idea might be to have the `N_*` functions return objects with a `.translate()` method that calls the appropriate `l*` function. Sometime later maybe.

yvanzo: I still don’t get the exact causes of [MBS-9673](https://tickets.metabrainz.org/browse/MBS-9673) as I cannot reproduce it with my local setup, although I guess it is some kind of caching issue.
1. Why German translations are currently sometimes returned (see [comment](https://tickets.metabrainz.org/browse/MBS-9673?focusedCommentId=45971#comment-45971))?
2. Why entity tabs are not affected by this bug whereas they use `l` instead of `N_l` too?

Apart from this bug, we should probably be using `i18n.strings.entityName` for core entity type search options. That would require to change `i18n.strings.*` implementation too.

mwiencek: So basically, since `TYPE_OPTION_GROUPS` is defined at the top level, all of these `l()` calls execute *when the module loads*. In Node.js, once a module is `require()`d, it's executed once and cached - if you `require()` it again (anywhere, at any point in the same process), it just returns the cached exports from the first require.

So if the entity type options are sometimes in another language, it's because the *very first request* which `require()`d that module was from a user with that language enabled.

Note that if you have `DEVELOPMENT_SERVER` set to `1`, this won't be reproducible, because we clear `require.cache` (which holds the cached modules) after each request. So on every request, it will execute every module again. `git grep clearRequireCache` will show you how that works.

Not sure which part of the entity tabs you're referring to - in root/components/EntityTabs.js I see `tabLinkNames` using `N_l`.

mwiencek: Steps to reproduce the different language being displayed:

1. Disable `DEVELOPMENT_SERVER`.
2. Switch to a language other than English.
3. Restart root/server.js.
4. Reload the page. The entity types will now be in the language you selected.
5. Switch the language back to English. Note the dropdown is still in the previous language, because the module is still cached from the first request.
6. Restart root/server.js again.
7. Reload the page. The types are in English now, as expected.

yvanzo: Thanks! I was referring to _Overview_ and all, but they are translated from within the `buildLinks` function.

yvanzo:
Looks like a Flow issue to me, I made a [minimal example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArAQQCdKBDATzAF4wBtVMMAb3Y7AAcAjAC4wAIgBqNGAFcApmAGiANDw58ATCIlS5Ydcp4BfFR269+AZi2SZ8iwY7HUAXQDc6LHkJQAlpQIA8mRMRGRUtHQsAAxuHjgEYL7++EGkNnK4IQAUqmCpshj4AHQA1rJ0uFlJgWQAlEUAtjR8VdLYhT44YFlldLVcuRyUsvjSlNiJfjWkLL1uYMDAYADqcJQluACEqoa1qLXumPGEUjDpspnMORwkFNT0jc2t7fidE1k02HSpSmA+2AATWQAD36Zl4+UKpXKlU+3zqjxaUDaHS6PXKYMGYGGo3GYDhqVm5XmizAAFFqGttuZdkY9gcgA), it is very close to https://github.com/facebook/flow/issues/4294.

mwiencek:
Not sure why this works, but looks like [annotating anyObj helps](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArAQQCdKBDATzAF4wBtVMMAb3Y7AAcAjAC4wAIgBqNGAFcApmAGiANDw58ATCIlS5Ydcp4BfFR269+AZi2SZ8iwY7HUAXQDc6LHkJSYNubiYwAAoeEgpqegA6AFsaPiCoaWwMfABLHDAAHgAVEU5DAD4gmmw6AHkyEWylMFTsABNZAA8ASi5VMArSWRTIgGtZOlxi0q6WmLiEpJT07GCBujazczBKWXxpSjmS8rIWBbcOwxajE5bXIA) in your example.

PR #641, opened 2018-04-10
MBS-9548: Filter external links by unique target

yvanzo:
### Fix [MBS-9548](https://tickets.metabrainz.org/browse/MBS-9548): Same link showing twice on the sidebar

Restore `unique` filtering on external links from late [root/layout/sidebar/sidebar-favicons.tt#L31](https://github.com/mwiencek/musicbrainz-server/blob/3205207c3c4f524e62119489957dc05531db162f/root/layout/sidebar/sidebar-favicons.tt#L31).

### DIFF
--- a/root/layout/components/ExternalLinks.js
+++ b/root/layout/components/ExternalLinks.js
@@ -6,6 +6,7 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+const _ = require('lodash');
 const React = require('react');
 const URL = require('url');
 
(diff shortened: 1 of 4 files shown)


yvanzo:
@mwiencek: What’s the most appropriate to test this regression? Selenium?

root/layout/components/ExternalLinks.js
```diff
@@ -87,10 +87,18 @@ const ExternalLinks = ({entity, empty, heading}) => {
...
-    return compare(a.target.sidebar_name, b.target.sidebar_name);
+    return compare(a.target.sidebar_name, b.target.sidebar_name) ||
+      compare(a.target.href_url, b.target.href_url);
   });
 
-  links.push.apply(links, otherLinks.map(function (relationship) {
+  const uniqueOtherLinks = otherLinks.filter(
+    function (link, index, sortedLinks) {
+      return index === 0 ||
+        sortedLinks[index - 1].target.href_url !== link.target.href_url;
```
mwiencek: Would `const uniqueOtherLinks = _.uniqBy(otherLinks, x => x. target.href_url)` (using `uniqBy` from lodash) also do the trick? I think that's what the sort-and-compare-pairwise logic does, but hard to be sure :)

mwiencek:
Yes, I think I would add some test links to t/sql/selenium.sql and load the page using Selenium.

root/layout/components/ExternalLinks.js
mwiencek: Ah true. Maybe it can be sorted after the uniq instead of before, but probably not worth futzing with.

yvanzo: I just found that Lodash provides [`sortedUniqBy`](https://lodash.com/docs#sortedUniqBy) too! I used it with your first suggestion in https://github.com/metabrainz/musicbrainz-server/pull/641/commits/4ad24ac319205b1df009fc62ad721922b961195.

PR #642, opened 2018-04-11
MBS-9669: Fix UI language selector menu header for locale with country code

yvanzo:
### Fix [MBS-9669](https://tickets.metabrainz.org/browse/MBS-9669): UI language menu not available in Spanish and Greek

Locale code is displayed in BCP 47 format in URL, cookie, and UI language selector menu, but it is used in POSIX format internally for Jed and PO files. The bug was to compare current language code in internal POSIX format with available UI languages code in BCP 47 format.

This patch calls component `LanguageMenu` with current language in BCP 47 format.

### DIFF
--- a/root/layout/components/BottomMenu.js
+++ b/root/layout/components/BottomMenu.js
@@ -48,18 +48,18 @@ const LanguageLink = ({language}: {|+language: ServerLanguageT|}) => (
 );
 
 type LanguageMenuProps = {|
-  +currentLanguage: string,
+  +currentBCP47Language: string,
   +serverLanguages: $ReadOnlyArray<ServerLanguageT>,
 |};
 
 const LanguageMenu = ({
-  currentLanguage,
+  currentBCP47Language,
   serverLanguages,
 }: LanguageMenuProps) => (
   <li className="language-selector" tabIndex="-1">
     <span className="menu-header">
       {languageName(
-        _.find(serverLanguages, x => x.name === currentLanguage),
+        _.find(serverLanguages, x => x.name === currentBCP47Language),
         true,
       )}
     </span>
@@ -67,7 +67,7 @@ const LanguageMenu = ({
       {serverLanguages.map(function (language, index) {
         let inner = <LanguageLink language={language} />;
 
-        if (language.name === currentLanguage) {
+        if (language.name === currentBCP47Language) {
           inner = <strong>{inner}</strong>;
         }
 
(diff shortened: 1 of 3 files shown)


yvanzo:
Or testbed needs improving first. Tests are run against MB UI languages in production only, not those in beta. I am a bit reluctant to add languages from [beta](https://github.com/metabrainz/docker-server-configs/blob/master/consul/services/beta.musicbrainz-server.json#L28) to [docker/config/services/dev.musicbrainz-server.json](https://github.com/metabrainz/musicbrainz-server/blob/master/docker/config/services/dev.musicbrainz-server.json#L8) as it would imply that production setup won’t be tested anymore.

Using distinct `dev.musicbrainz-server.json` files for each `BUILD_TAG` in Jenkins won’t suffice as the language menu tests won’t pass with production setup anyway. It has to be somehow skipped depending on either `BUILD_TAG` or `MB_LANGUAGES`.

mwiencek:
Hmm, adding the extra languages for now seems fine to me, to be honest. Do you think it will hide issues we'd only see with the production list?

But about adding them, I think docker/config/services/dev.musicbrainz-server.json is not actually used by anything. The tests use a static DBDefs configuration at docker/musicbrainz-tests/DBDefs.pm.

mwiencek:
(Coming back to this since some of the necessary test changes were merged), the solution still makes sense to me, but I'm wondering about an alternative: any reason we can't change the Jed files/po symlinks/DBDefs values to all use BCP47 instead?

yvanzo:
Both Transifex and Perl (error messages) handle locale code in POSIX format.

mwiencek:
> Both Transifex and Perl (error messages) handle locale code in POSIX format.

If we change `current_language` in the stash from POSIX to BCP47, I'm not sure I understand how it affects Transifex or any error messages (though it may have other issues).

mwiencek:
BTW, other than the conflicts, I've approved this - I didn't mean to throw a wrench into things with my last two comments. :)

I do think it's unfortunate we have to have these `s/_/-/g` and vice-versa throughout the code (I can list the other places we do this, if it's useful), so standardizing to BCP47 outside of Transifex would be nice, if it's possible.

One more note, since this PR was opened the tests image should have languages installed/configured already, making the last two commits obsolete.

yvanzo:
Just squashed fixes and rebased onto master.

PR #644, opened 2018-04-12
Release editor Selenium tests

mwiencek:
* Adds a large selenium test for the release editor, and converts the existing seeding tests to selenium too. In a later PR I'd like to port everything under root/static/scripts/tests/release-editor/. 
* Makes the selenium tests isolated/idempotent. You can run individual tests now, in any order, and they won't conflict with each other. To just run the release editor tests: `./t/selenium.js t/selenium/release-editor/*.html`.
* Adds some command line flags to selenium.js that are useful for development and debugging. For example, this runs the seeding tests in non-headless mode, and keeps the page open so you can inspect things: `./t/selenium.js --headless=false --stay-open=true t/selenium/Release_Editor_Seeding.html`.
* Automatically creates/drops a `musicbrainz_selenium` DB while running the tests, and instructs MBS to use it if allowed (see HACKING.md). No need to switch the databases in DBDefs.pm anymore.


t/selenium.js
```diff
@@ -270,50 +292,173 @@ async function handleCommand(command, target, value, baseURL, t) {
...
+(async function runTests() {
+  const TEST_TIMEOUT = 60000; // 60 seconds
+
+  const cartonPrefix = process.env.PERL_CARTON_PATH
+    ? 'carton exec -- '
+    : '';
+
+  function pgPasswordEnv(db) {
+    if (db.password) {
+      return {env: {PGPASSWORD: db.password}};
```
yvanzo: Shouldn’t it return `{PGPASSWORD: db.password}` only?

t/selenium.js
```diff
@@ -270,50 +292,173 @@ async function handleCommand(command, target, value, baseURL, t) {
...
+  const seleniumDbCheck = await execFile(
+    'psql', [...hostPort, '-U', testDb.user, '-c', 'SELECT 1', 'musicbrainz_selenium'],
+    pgPasswordEnv(testDb),
+  ).catch(x => x);
+
+  if (seleniumDbCheck.code === 0) {
+    await dropSeleniumDb();
+  } else if (seleniumDbCheck.code !== 2) {
+    // An error other than the database not existing occurred.
+    throw error;
```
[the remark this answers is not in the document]
mwiencek: Yep, fixed

t/selenium.js
mwiencek: I think not since it ultimately gets passed as the `options` argument to https://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback

yvanzo: Alright, returning broken options set misleadingly masked some errors for me. The actual issue is that some variables in the process environment are necessary to run shell commands on Linux. Cloning `process.env` allows to work around it.

yvanzo:
Test _Create Account_ doesn’t pass if `WEB_SERVER` is not set to `localhost:5000` in `lib/DBDefs.pm`.

mwiencek:
> Test Create Account doesn’t pass if WEB_SERVER is not set to localhost:5000 in lib/DBDefs.pm.

Fixed in https://github.com/metabrainz/musicbrainz-server/pull/644/commits/cc26c99b1c9f8d1d1c99096c313f8544cb9f9a8f

yvanzo:
Invariably fails in _The Downward Spiral_ with
```
# click target="css=.tabs a[href$=\"/edit\"]" value=""
# click target="xpath=//a[@href='#tracklist']" value=""
# pause target="1000" value=""
# click target="css=button.reset-track-numbers" value=""
# click target="xpath=//a[@href='#edit-note']" value=""
# click target="id=enter-edit" value=""
not ok 97 caught exception
  ---
    operator: fail
    at: process._tickCallback (internal/process/next_tick.js:182:7)
    stack: |-
      Error: caught exception
          at Test.assert [as _assert] (/musicbrainz-server/node_modules/tape/lib/test.js:212:54)
          at Test.bound [as _assert] (/musicbrainz-server/node_modules/tape/lib/test.js:64:32)
          at Test.fail (/musicbrainz-server/node_modules/tape/lib/test.js:277:10)
          at Test.bound [as fail] (/musicbrainz-server/node_modules/tape/lib/test.js:64:32)
          at /musicbrainz-server/t/selenium.js:563:15
          at <anonymous>
          at process._tickCallback (internal/process/next_tick.js:182:7)
  ...
{ WebDriverError: unknown error: Element <button type="button" id="enter-edit" class="positive" data-click="submitEdits" data-bind="visible: activeTabID() === '#edit-note', enable: allowsSubmission()" style="">...</button> is not clickable at point (206, 604). Other element would receive the click: <div id="edit-note" aria-labelledby="ui-id-4" class="ui-tabs-panel ui-widget-content ui-corner-bottom" role="tabpanel" aria-expanded="true" aria-hidden="false" style="display: block;">...</div>
  (Session info: headless chrome=66.0.3359.139)
  (Driver info: chromedriver=2.37.544315 (730aa6a5fdba159ac9f4c1e8cbc59bf1b5ce12b7),platform=Linux x86_64)
    at Object.checkLegacyResponse (/musicbrainz-server/node_modules/selenium-webdriver/lib/error.js:529:15)
    at parseHttpResponse (/musicbrainz-server/node_modules/selenium-webdriver/lib/http.js:509:13)
    at doSend.then.response (/musicbrainz-server/node_modules/selenium-webdriver/lib/http.js:441:30)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7)
From: Task: WebElement.click()
    at thenableWebDriverProxy.schedule (/musicbrainz-server/node_modules/selenium-webdriver/lib/webdriver.js:807:17)
    at WebElement.schedule_ (/musicbrainz-server/node_modules/selenium-webdriver/lib/webdriver.js:2015:25)
    at WebElement.click (/musicbrainz-server/node_modules/selenium-webdriver/lib/webdriver.js:2096:17)
    at handleCommand (/musicbrainz-server/t/selenium.js:337:22)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7) name: 'WebDriverError', remoteStacktrace: '' }
```

yvanzo:
It turns out to be just time passing much faster on my machine!

It completed once I added a `0` to the timeout for _The Downward Spiral_ in in `t/selenium.js`. How is it computed?

It now fails intermittently on asserts right after `clickAndWait` items with:
> TimeoutError: Waiting element to become stale
> Wait timed out after 15002ms

yvanzo:
Some tests don’t contain any `assert*` and thus count as 0 test.

mwiencek:
@yvanzo Hmm, I've increased all of the timeouts now so they hopefully pass on more machines. The old timeouts were mainly just calculated so that things passed on my machine + Jenkins.

> Some tests don’t contain any assert* and thus count as 0 test.

The MBS-7456 one doesn't, at least, since it's testing that an exception isn't thrown. With zero assertions it should pass by default unless it hits https://github.com/mwiencek/musicbrainz-server/blob/cea163603b0ee9e34f820c7642e5a173cb8f2a91/t/selenium.js#L244

mwiencek:
The only downside to increasing the timeouts is that it'll take longer for the tests to fail if there's an actual problem, so I don't want to increase them by a crazy amount.

mwiencek:
@yvanzo It might have been flaky because we weren't using `clickAndWait` when clicking the submit buttons, which I changed in 6d590f2dc8eca226be47c4d9aeb167224b2de623

mwiencek:
@yvanzo Does it fail if you only run the seeding tests? Can you see where it's stuck with `./t/selenium.js --headless=false --stay-open=true t/selenium/release-editor/Seeding.html`? I'm wondering if it's still on the `0_based_indexes.html` page when it encounters that error, or if it's in the release editor but no seed errors are displayed.

PR #646, opened 2018-04-19
MBS-9664: Add database constraints to disallow loop relationship

yvanzo:
### Implement [MBS-9664](https://tickets.metabrainz.org/browse/MBS-9664): Add database constraints to disallow loop relationship

### DIFF
--- a/admin/sql/updates/20180331-mbs-9664-non-loop-checks.sql
+++ b/admin/sql/updates/20180331-mbs-9664-non-loop-checks.sql
@@ -1,0 +1,17 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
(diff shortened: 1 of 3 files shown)


admin/sql/updates/20180331-mbs-9664-non-loop-checks.sql
```diff
@@ -0,0 +1,17 @@
+\set ON_ERROR_STOP 1
```
mwiencek: Just needs adding to upgrade.json too. Shall we run it in production now (if you haven't already)?

yvanzo: Did both, thanks!

PR #651, opened 2018-04-24
MBS-9691: Fix aliases editing links

yvanzo:
### Fix [MBS-9691](https://tickets.metabrainz.org/browse/MBS-9691): Beta: Alias editing links are gone

The issue was about computing permissions to display edit links:

1. Default editing permission has been given to location editor only
2. There were no input Prop to override the default permission

This patch fixes both by adding `allowEditing` as a Prop and setting
default value to whether user is a logged-in not-disabled editor.

`allowEditing` values for area and instrument were correctly set.

### DIFF
--- a/root/components/Aliases/index.js
+++ b/root/components/Aliases/index.js
@@ -18,11 +18,11 @@ const AliasTable = require('./AliasTable');
 type Props = {
   +$c: CatalystContextT,
   +aliases: $ReadOnlyArray<AliasT>,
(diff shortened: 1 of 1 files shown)


mwiencek:
Sorry, that was my fault. I'm not sure what I was thinking there https://github.com/metabrainz/musicbrainz-server/commit/8099a6ac893560734daae4ec49d83c7898e27223#diff-a4e968adac3304fa174d28e2cdc1c779

PR #655, opened 2018-04-25
Silence some gulp/uglifyify warnings

yvanzo:
Resources compilation logs were full of useless warnings, this patch removes two of those:
* _MaxListenersExceededWarning_ which was due to `MaxListeners` not being proportionate to the number of MB languages. (This patch does not set it to `Infinity` either.)
* _Output exceed 256 characters_ (x432) which cluttered the output and is just dropped here.

### DIFF
--- a/root/static/gulpfile.js
+++ b/root/static/gulpfile.js
@@ -18,9 +18,19 @@ const yarb = require('yarb');
 const poFile = require('../server/gettext/poFile');
 const DBDefs = require('./scripts/common/DBDefs');
 
-// This may need to be increased when a new bundle is added, to silence
-// warnings of the form "Possible EventEmitter memory leak detected."
-EventEmitter.defaultMaxListeners = 32;
+const POSIX_LANGUAGES = _(DBDefs.MB_LANGUAGES || '')
+  .split(/\s+/)
+  .compact()
+  .without('en')
+  .map(langToPosix);
+
+{
+  let langCount = POSIX_LANGUAGES.size();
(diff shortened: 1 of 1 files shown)


root/static/gulpfile.js
```diff
@@ -138,14 +150,22 @@ function runYarb(resourceName, vinyl, callback) {
...
+
     bundle.transform('uglifyify', {
       // See https://github.com/substack/node-browserify#btransformtr-opts
       global: true,
 
       // Uglify options
       output: {
         comments: /@preserve|@license/,
-        max_line_len: 256
+        max_line_len: BUNDLE_MAX_LINE_LENGTH
```
mwiencek: I think we should perhaps remove this setting. It was to make line numbers in stack traces more usable, but I can't think of a case where it's helped that much (judging by the warnings, it's not even respecting the limit in many cases?), and this problem would be better solved by [providing source maps to Sentry](https://docs.sentry.io/clients/javascript/sourcemaps/).

yvanzo: As stated in the commit message, there is no other way to _fix the `max_line_len` warning_ except by removing the `max_line_len` option. Do we remove this option now? Or do we merge this work around until we provide source maps to Sentry?

mwiencek: I'd personally remove `max_line_len` now; it's broken and not useful. Providing source maps will have to wait for Webpack, since it's too difficult with our current setup.

PR #657, opened 2018-04-25
MBS-9687: Fix editing forms field error message localization

yvanzo:
### Fix [MBS-9687](https://tickets.metabrainz.org/browse/MBS-9687): Field error messages are not always translated

Move localization of these error messages from Moose code to templates.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Artist.pm
+++ b/lib/MusicBrainz/Server/Form/Artist.pm
@@ -69,21 +69,21 @@ sub validate {
     if ($self->field('type_id')->value &&
         $self->field('type_id')->value == 2) {
         if ($self->field('gender_id')->value) {
-            $self->field('gender_id')->add_error(l('Group artists cannot have a gender'));
+            $self->field('gender_id')->add_error(l('Group artists cannot have a gender.'));
         }
     }
 
     if ($self->field('type_id')->value &&
         $self->field('type_id')->value == 5) {
         if ($self->field('gender_id')->value) {
-            $self->field('gender_id')->add_error(l('Orchestras cannot have a gender'));
+            $self->field('gender_id')->add_error(l('Orchestras cannot have a gender.'));
         }
     }
 
     if ($self->field('type_id')->value &&
         $self->field('type_id')->value == 6) {
         if ($self->field('gender_id')->value) {
-            $self->field('gender_id')->add_error(l('Choirs cannot have a gender'));
+            $self->field('gender_id')->add_error(l('Choirs cannot have a gender.'));
         }
     }
 }
--- a/lib/MusicBrainz/Server/Form/Artist.pm
+++ b/lib/MusicBrainz/Server/Form/Artist.pm
@@ -1,6 +1,7 @@
 package MusicBrainz::Server::Form::Artist;
 use HTML::FormHandler::Moose;
 use MusicBrainz::Server::Form::Utils qw( select_options_tree );
+use MusicBrainz::Server::Translation qw( l );
 
 extends 'MusicBrainz::Server::Form';
 with 'MusicBrainz::Server::Form::Role::Edit';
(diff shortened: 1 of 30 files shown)


yvanzo:
@mwiencek: Error messages about text length for Barcode, (disambiguation) Comment, and ISRC are currently not translated. It would require to have custom messages for `minlength` and `maxlength` checks. But there is probably more things to improve with `HTML::FormHandler::Field::Text` and `MusicBrainz::Server::Form::Field::Text`.

yvanzo:
For some reason, changes to TT `field_errors` macro were not pushed. Now fixed.

mwiencek:
Looking at `git grep '\->add_error\W' -- lib/MusicBrainz/Server/Form`, I think there may have been a few spots missed. For example, in lib/MusicBrainz/Server/Form/Alias.pm there's a string in `validate_primary_for_locale` which I guess should be using `N_l` instead of `l`, and in the `validate` subroutine it's using an (at present) untranslated string. I guess the other untranslated strings we can fix later.

yvanzo:
My bad, I did not check modules outside of `MusicBrainz::Server::Form::Field::` at all, just found 8 other `l` to convert and many more `add_error` calls missing localization.

yvanzo:
Before that, there is an additional issue with parametrized error message, for example in [Field::ArtistCredit at line 48](https://github.com/yvanzo/musicbrainz-server/blob/05c080af430a1b69082358f4b88e057c3f42d401/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm#L48). Two ways out: either carry error message as a compound instead of a string, or just stop moving everything out of there and use the right language setting for each call to Moose classes.

yvanzo:
I opted for the first way out: _carry error message as a compound_ (encoded into a JSON string) _instead of a_ (localized) _string_. It may slightly increase overhead. Most of field errors have been moved to `J_l`.

@mwiencek: Can you please make an intermediate review of this to check whether the approach seems to be acceptable and whether it should be extended to other messages such as controller errors?

mwiencek:
Sorry it's been a while here. :) Can you remind me why we don't want to translate messages on the server? My first impression was "huh, this is complicated," but I don't fully understand it.

82b0c277c9482b6ae62c8beb8dd4ac8e7cc300a7 says it's to allow parameterized error messages to be translated and expanded later on in React, but not why the translation/expansion has to be done by the client.

For the issue behind MBS-9687, I thought there were simple ways around it, like passing a subroutine to the `message =>` bits in form modules.

yvanzo:
> Can you remind me why we don't want to translate messages on the server?

That would be _the second way out_, if it exists, but I did not found it.

> huh, this is complicated

True.

> passing a subroutine to the `message =>` bits in form modules.

This is already the case, see [ISRC form field module](https://github.com/metabrainz/musicbrainz-server/blob/v-2018-05-30/lib/MusicBrainz/Server/Form/Field/ISRC.pm#L18)
(for `message` and [ArtistCredit form field module](https://github.com/metabrainz/musicbrainz-server/blob/v-2018-05-30/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm#L55) for `add_error`).

mwiencek:
> This is already the case, see ISRC form field module (for message and ArtistCredit form field module for add_error).

I mean a code reference, which would presumably be evaluated during the request:

```Perl
message => sub { l('This is not a valid ISRC') },
```

mwiencek:
The artist credit example you linked is strange, because the `validate_field` *is* evaluated during the request. In fact, all of the other errors there display as correctly translated when I trigger them. I wonder if the actual bug is the whitespace in the message causing it to not be looked up properly.

yvanzo:
Thanks, this :cloud_with_lightning: was indeed unneeded to adjust these small localization issues. And the artist credit example was indeed due to _the extraneous whitespace/newlines_ (fixed in https://github.com/metabrainz/musicbrainz-server/pull/657/commits/d9c8d932373384029e6415b98894beabfbffdbfe).

One last glitch: Shall we normalize error message to end with a full stop or add bullet to `li`?
`HTML::FormHandler` messages have no period, but there is no separator in field errors `ul`.

lib/MusicBrainz/Server/Form/Artist.pm
```diff
@@ -69,21 +69,21 @@ sub validate {
     if ($self->field('type_id')->value &&
         $self->field('type_id')->value == 2) {
         if ($self->field('gender_id')->value) {
-            $self->field('gender_id')->add_error(l('Group artists cannot have a gender'));
```
mwiencek: It sounds a bit odd to me without the article (I only ever see/hear gender identity as a 
countable noun).

yvanzo: Ok, I reverted this change.

PR #658, opened 2018-04-26
MBS-9696: Replace @users.musicbrainz.org with noreply@musicbrainz.org in hidden email From field

yvanzo:
The purpose of this change is to provide DKIM signature.
See https://tickets.metabrainz.org/browse/MBS-9696

### DIFF
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -50,7 +50,7 @@ sub _user_address
     my ($user, $hidden) = @_;
 
     my $quoted_name = _encode_header($user->name);
-    my $email = $hidden ? 'noreply@musicbrainz.org' : $user->email;
+    my $email = $hidden ? $EMAIL_NOREPLY_ADDR_SPEC : $user->email;
 
     return Email::Address->new($quoted_name, $email)->format;
 }
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -50,14 +50,9 @@ sub _user_address
     my ($user, $hidden) = @_;
 
     my $quoted_name = _encode_header($user->name);
+    my $email = $hidden ? 'noreply@musicbrainz.org' : $user->email;
 
-    if ($hidden) {
-        # Hide the real address
-        my $email = sprintf '"%s"@users.musicbrainz.org', $quoted_name;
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Email.pm
```diff
@@ -50,14 +50,9 @@ sub _user_address
     my ($user, $hidden) = @_;
 
     my $quoted_name = _encode_header($user->name);
+    my $email = $hidden ? 'noreply@musicbrainz.org' : $user->email;
```
zas: Instead of using hardcoded noreply address it'd be more consistent to modify/re-use https://github.com/metabrainz/musicbrainz-server/blob/f48a16b81f8ffecdef3552f41b48092c4adc3199/lib/MusicBrainz/Server/Constants.pm#L293
 (it'd need some changes though as format is different)

yvanzo: @zas: Moved around in https://github.com/metabrainz/musicbrainz-server/pull/658/commits/119f904c20d7dbb73755144772e25cb22700936c. Technically it is still _hardcoded_, but at least these related constants are defined in a single place.

t/lib/t/MusicBrainz/Server/Email.pm
```diff
@@ -47,7 +47,7 @@ test all => sub {
     is($delivery->{envelope}->{from}, 'noreply@musicbrainz.org', "Envelope from is noreply@...");
     my $e = $delivery->{email};
     $email->transport->clear_deliveries;
-    is($e->get_header('From'), '"Editor 1" <"Editor 1"@users.musicbrainz.org>', 'Header from is Editor 1 @users.musicbrainz.org');
+    is($e->get_header('From'), '"Editor 1" <noreply@musicbrainz.org>', 'Header from is "Editor 1" <noreply@musicbrainz.org>');
```
mwiencek: I'm not sure why we quote the editor name there, but.

yvanzo: This is done by the function `format` of the module `Email::Address` to follow RFC 2822.

mwiencek: Right, I meant 'cause it doesn't appear to contain any characters that need to be quoted (?), but I guess it just always quotes it.

PR #661, opened 2018-04-28
MBS-9705: Hide /show subpath from Overview tab

yvanzo:
### Fix [MBS-9705](https://tickets.metabrainz.org/browse/MBS-9705): Overview tab link is now appended with /show

Both URLs are correct, but `/show` subpath is unneeded and was not returned by `uri_for`.

### DIFF
--- a/root/components/EntityTabs.js
+++ b/root/components/EntityTabs.js
@@ -70,7 +70,7 @@ function buildLinks(
   page: string,
   editTab: React.Node,
 ): React.Node {
-  const links = [buildLink(l('Overview'), entity, 'show', page, 'index')];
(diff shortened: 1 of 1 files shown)


root/components/EntityTabs.js
```diff
@@ -70,7 +70,7 @@ function buildLinks(
   page: string,
   editTab: React.Node,
 ): React.Node {
-  const links = [buildLink(l('Overview'), entity, 'show', page, 'index')];
+  const links = [buildLink(l('Overview'), entity, undefined, page, 'index')];
```
mwiencek: Passing an empty string should also work and fix the Flow error.

PR #663, opened 2018-05-03
MBS-9708: Speed up area containment queries

mwiencek:
The `area_containment` view has a major flaw, which is that (at least in PostgreSQL 9.5) the recursive CTE it uses requires calculating containments for all areas, even though we only need them for the areas passed to `load_containment` (usually just one area).

We have a cache for the containments set up, but it isn't very effective because it must be invalidated often (for all areas).

The `area_containment` view was apparently added for SEARCH-359, but it's not used anywhere in the current or future search indexers.

In this diff, I drop the `area_containment` view and inline it into the appropriate places, so we can parameterize it and ensure it only calculates containments for the areas we need.


lib/MusicBrainz/Server/Data/Area.pm
```diff
@@ -66,100 +70,57 @@ sub load
...
+sub get_containment_query {
+    my ($self, $link_type_param, $descendant_param) = @_;
+
+    return (qq{
+        WITH RECURSIVE area_descendants AS (
+            SELECT entity0 AS parent, entity1 AS descendant, 1 AS depth
+              FROM l_area_area laa
+              JOIN link ON laa.link = link.id
+             WHERE link.link_type = $link_type_param
+               AND entity1 = $descendant_param
```
mwiencek: This `AND entity1 = ?` bit is the critical part which was added and makes the recursive query much faster.

admin/sql/CreateViews.sql
```diff
@@ -13,36 +13,6 @@ CREATE OR REPLACE VIEW release_event AS
...
 
--- First, construct a table (recursively) of parent -> descendant connections
--- for areas, including an array of the path (the 'descendants' array).
-
--- Then, find the shortest path to each type of parent by joining to area,
--- distinct on descendant, type, and order by the length of the array of descendants.
-
--- link type de7cc874-8b1b-3a05-8272-f3834c968fb7 is the area-area 'parent of' relation
-CREATE OR REPLACE VIEW area_containment AS
-        WITH RECURSIVE area_descendants AS (
```
mwiencek: See https://www.postgresql.org/docs/9.5/static/queries-with.html for some insight on why this was slow:

> However, the other side of this coin is that the optimizer is less able to push restrictions from the parent query down into a WITH query than an ordinary subquery. The WITH query will generally be evaluated as written, without suppression of rows that the parent query might discard afterwards.

t/lib/t/MusicBrainz/Server/Data/Area.pm
```diff
@@ -58,33 +58,35 @@ EOSQL
...
     is($area->name, 'parent city', 'correct descendant is loaded');
 
     $test->c->model('Area')->load_containment($area);
-    is($area->parent_city, undef, 'no parent city is loaded');
-    is($area->parent_subdivision->name, 'parent subdivision', 'correct parent subdivision is loaded');
-    is($area->parent_country->name, 'parent country', 'correct parent country is loaded');
+    is(scalar @{$area->containment}, 3);
+    is($area->containment->[0]->name, 'parent subdivision', 'correct parent subdivision is loaded');
+    is($area->containment->[1]->name, 'parent country', 'correct parent country is loaded');
+    is($area->containment->[2]->name, 'parent meta-country', 'parent meta-country is loaded');
```
mwiencek: This is the one change in behavior I can see: the containment array holds the entire hierarchy of areas with type in (1, 2, 3), so if there's a parent and parent-parent country, they'll both get included in the containment.

mwiencek:
@zas Here's some perf info: https://gist.github.com/mwiencek/39632dbace05bfa26741c7af776f7939

Gist of the gist: execution time is decreased by about 99.251% on average.

zas:
> Gist of the gist: execution time is decreased by about 99.251% on average.

Whooa ! Good job, i'm eager to see this patch in production, it should have a visible impact on bowie's load (and number of active connections).

mwiencek:
@yvanzo That sounds like what https://github.com/mwiencek/musicbrainz-server/blob/fda5318/lib/MusicBrainz/Server/Data/Editor.pm#L166 is doing too (to find all editors under a particular area). With the `get_containment_query` method I added, it's still easy to embed the containment subquery where you need it. You can join against subqueries, too. The view looked cleaner, but this patch shouldn't hurt code reuse overall.

Re the discussion about loop relationships on IRC, rest assured they're still [handled here](https://github.com/mwiencek/musicbrainz-server/blob/fda5318/lib/MusicBrainz/Server/Data/Area.pm#L89), so the recursive query can't go into an infinite loop (I made sure to test that case locally).

PR #668, opened 2018-05-14
Convert the homepage to React

mwiencek:
The `ArtworkImage` commit is cherry-picked from the entity sidebar PR.


yvanzo:
Maybe we can use a Markdown renderer to support emphasis? This will be needed for [OTHER-77](https://tickets.metabrainz.org/browse/OTHER-77).

mwiencek:
@yvanzo Maybe a limited subset for emphasis only? I'm even thinking we'd want to support either asterisks or underscores for emphasis but not both, to decrease the syntax burden for translators. A full markdown renderer would create a pretty big change in semantics for all of our source strings, so I'm not sure about it yet. I could see how it would unlock some possibilities for translating larger pieces of text.

For now, we could also change the React interpolation code to recognize `<strong>` tags in source strings, and output the right thing. Then I can leave the strings unchanged. What do you think is a good course of action for this PR?

yvanzo:
@mwiencek: For this PR, recognizing `<strong>` tags in source strings, definitely. :+1: 

For the record, source strings currently contain:
* 40 `<strong>` (and `<b>`)
* 21 `<code>`
* 17 `<span>` (with specific class names)
* 8 `<em>`
* 7 `<a>` (mostly from relationships)
* 3 `<li>` and 1 `<ul>` (from relationships)
* (And 2 `<abbr>` (with title) that can probably be moved out of strings)

mwiencek:
Okay, this took a century but there's a new i18n string parser here which can convert HTML into React elements. The old `expand` code is still there and used, particularly by our knockout stuff - the new one can primarily only be used with React.

I reverted all of the changes I made to the strings on the homepage.

root/static/scripts/tests/i18n/expand2.js
```diff
@@ -0,0 +1,127 @@
...
+  expandTest(
+    '<a href="javascript:alert(\'HAx0r\')"></a>',
+    null,
+    '&lt;a href=&quot;javascript:alert(&#x27;HAx0r&#x27;)&quot;&gt;&lt;/a&gt;',
+  );
+  expandTest(
+    '<script>alert("HAx0r")</script>',
+    null,
+    '&lt;script&gt;alert(&quot;HAx0r&quot;)&lt;/script&gt;',
+  );
```
yvanzo: That would be nice to capture and test thrown error for the two latest samples.

That should not prevent merging this pull request anyway.

yvanzo: Some test suggestions for attributes filtering:
```html
<br size="&{alert('HAx0r')}">
<li style="behavior: url(http://hax.0r);">
<span style="background-image: url(javascript:alert('HAx0r'))">
```

yvanzo: I don't know which browser is still vulnerable to the first example which comes from [OWASP](https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet).

Looking at it again, testing for javascript protocol scheme might not be that safe. Other examples on the same page defeat the current [`jsProtocol`](https://github.com/metabrainz/musicbrainz-server/pull/668/files#diff-24e22d15e660de887ecc6765c640286cR31) pattern: `javacript&colon;`, `&NewLine;javascript:`. Moreover, some other protocol schemes may be used for XSS: `applescript`, `vbscript`…  It is difficult to assume that there is no/won't be any other way to execute code in there. It might be safer to whitelist internal paths (starting with `/`) and protocols `http` and `https`, which is all we need for now.

mwiencek: Good point! I'll switch it to a whitelist instead. Also, thankfully, that bizarre `&{}` syntax seems to be a [non-standard proprietary remnant of Netscape <= 4](https://security.stackexchange.com/questions/64925/xss-technique-javascript-include/64926#64926) and not usable in any other browser.

PR #674, opened 2018-05-27
Fix the SOLR search basic search end point to use the appropriate boosting config

samj1912:


lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -218,16 +218,21 @@ sub xml_search
...
-                "rows=$limit&wt=$format&start=$offset" .
-                "&q=" . uri_escape_utf8($query) . "&defType=dismax";
-        } else {
-            $url_ext = "/$resource/edismax?" .
-                "rows=$limit&wt=$format&start=$offset" .
-                "&q=" . uri_escape_utf8($query);
+            # Solr has a bug where the dismax end point behaves differently
+            # from edismax when the query size is 1. This is a fix for that
+            # See https://issues.apache.org/jira/browse/SOLR-12409
+            if (split(" ", $query) == 1) {
```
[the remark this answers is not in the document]
yvanzo: Splitting on the first word breaks:
```perl
split(/\b{wb}/, $query, 1) == 1
```
Notes:
* Word breaks are [_tailored to Perl expectations_](http://perldoc.perl.org/perlrebackslash.html), that is, for example, underscore `_` is not considered as a word separator.
* Separators are counted as tokens, that is:
** `Query!` has two tokens `Query` and `!`,
** ` query   ` has three tokens ` `, `query` and `   `.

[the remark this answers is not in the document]
yvanzo: Simpler test to match at most one word (or one separator group) without underscore:
```perl
$query =~ /^[^_](?:\B{wb}[^_])*$/
```

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -427,7 +427,7 @@ sub find_recent_by_artists
         $search_url = sprintf('http://%s/ws/2/recording/?query=title:""',
                               DBDefs->SEARCH_SERVER);
     } else {
-        $search_url = sprintf('http://%s/recording/select?q=recording:*&wt=mbxml',
+        $search_url = sprintf('http://%s/recording/edismax?q=recording:*&wt=mbxml',
```
samj1912: Should this be edismax or dismax?

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -1035,7 +1043,7 @@ sub xml_search
     if (DBDefs->SEARCH_ENGINE eq 'LUCENE') {
         $search_url_string = "http://%s/ws/$version/%s/?query=%s&offset=%s&max=%s&fmt=xml";
     } else {
-        $search_url_string = "http://%s/%s/select?q=%s&start=%s&rows=%s&wt=mbxml&fl=score";
+        $search_url_string = "http://%s/%s/edismax?q=%s&start=%s&rows=%s&wt=mbxml";
```
samj1912: same doubt here, although earlier it was using the lucene syntax

lib/MusicBrainz/Server/Data/WebService.pm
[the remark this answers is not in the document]
samj1912: Used a much simpler regex that works for all the cases.

lib/MusicBrainz/Server/Data/Recording.pm
mwiencek: This was only added to get the time that recordings were last indexed by the search server: 6cdbe908d137d7de9924f1aba2ebad33940631d1 (because recordings aren't indexed every three hours like the code previously expected). So I guess it doesn't matter what query it uses as long as there's a `created` field in the response.

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -748,12 +748,20 @@ sub external_search
...
-            $search_url_string = "http://%s/%s/edismax?q=%s&start=%s&rows=%s&wt=mbjson&fl=score";
-        } else {
-            $search_url_string = "http://%s/%s/select?q=%s&start=%s&rows=%s&wt=mbjson&fl=score&defType=dismax";
+        my $endpoint = "advanced";
+        if (!$adv)
+        {
+            # Solr has a bug where the dismax end point behaves differently
+            # from edismax when the query size is 1. This is a fix for that
+            # See https://issues.apache.org/jira/browse/SOLR-12409
+            if (split(/[\P{Word}_]+/, $query) == 1) {
```
mwiencek: I suggest adding `, 2` as the last argument to `split` so that it stops right after it finds the first split (since you only care that it finds one, not how many).

lib/MusicBrainz/Server/Data/Search.pm
mwiencek: `edismax` seems correct given the `$query` it constructs above.

lib/MusicBrainz/Server/Data/Recording.pm
mwiencek: Of course, later we can delete this plus lib/MusicBrainz/Server/Controller/WS/js/LastUpdatedRecordings.pm and the code that uses it in root/static/scripts/release-editor/recordingAssociation.js once the new search is live, since it updates near-instantly.

lib/MusicBrainz/Server/Data/Search.pm
samj1912: Done.

PR #679, opened 2018-06-06
MBS-9408: Add Juno Download links to the sidebar

Kebabpizza:
https://tickets.metabrainz.org/browse/MBS-9408


mwiencek:
@brainzbot add to whitelist

lib/MusicBrainz/Server/Entity/URL/JunoDownload.pm
```diff
@@ -0,0 +1,26 @@
...
+    return "Juno Download";
+}
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2018 MetaBrainz Foundation
```
yvanzo: _You should replace that with your own name (unless you intended to assign copyright to MeB)._

Kebabpizza: Good catch, updated.

PR #682, opened 2018-06-15
Remove the note, now that Solr keeps the indexes updated every minute

samj1912:


root/search/components/SearchForm.js
```diff
@@ -103,10 +103,6 @@ const SearchForm = ({form}: Props) => {
           </div>
         </form>
       </div>
-      <div className="searchinfo">
-        <strong>{l('Please note:')}</strong>
-        <p>{l('Search indexes are updated every 3 hours, use the direct database search if you require up to the minute correct results.')}</p>
-      </div>
```
yvanzo: I would drop the strong `Please note:` but replace the paragraph with documentation:

``` javascript
        <p>
          {l('For more information, check the {doc_doc|documentation}.', {
            __react: true,
            doc_doc: '/doc/Search',
          })}
        </p>
```

It is already used in some entity edit forms, it is not an additional localization string.

samj1912:
@yvanzo updated

mwiencek:
You need to update root/search/form.tt too (the React form is only used for areas and artists).

yvanzo:
Fixed in https://github.com/metabrainz/musicbrainz-server/commit/c1d414a6e42f4fd0626ad56b6c43db5999ac7686.

PR #683, opened 2018-06-18
MBS-9734: Fix invalid format for ISRCs while parsing WS/2 json search…

samj1912:
… output

Currently JSON search outputs ISRCs as

`"isrcs": [{"id": "XYZ"}]`

This changes it to be consistent with WS/2 lookup
`"isrcs": ["XYZ"]`

Also - this is currently dependent on Solr search, so only deploy it once Solr is deployed.


mwiencek:
We could also conditionally change how it's parsed depending on whether `DBDefs->SEARCH_ENGINE` is set to `LUCENE`.

samj1912:
@yvanzo done already

PR #685, opened 2018-06-20
Fix consul template definition for DB_STAGING_SERVER_DESCRIPTION

samj1912:
It is currently defined as CODE_DEF. It should be STRING_DEF.


mwiencek:
We used to set this to something like `shift->DB_STAGING_SERVER_DESCRIPTION_BETA` (which returned a translated string) which is why it was `CODE_REF` before. But we don't use this anymore and shouldn't need to, so I'd probably just delete the line instead. :)

PR #686, opened 2018-06-20
MBS-9744: OAuth userinfo endpoint returns latin1 encoded JSON

mwiencek:
This also fixes MBS-8334, because I ran into that issue while updating tests. I ended up reverting the test changes I was making, but since the fix was already done and pretty small, I left it in the PR.


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -70,13 +70,30 @@ sub _entity_class
     return 'MusicBrainz::Server::Entity::Editor';
 }
 
+=item lc_name
+
+Return an editor name with ASCII characters lower-cased.
```
mwiencek: Another issue I ran into while adding tests...

I named the new test editor `æditorⅣ`, which broke `get_by_name` here because `lc` lowercases the roman numeral while `lower(name)` doesn't (in the C locale). So the editor couldn't be found. I wasn't actively looking for more things to break. :)

lib/MusicBrainz/Server/View/JSON.pm
```diff
@@ -1,46 +0,0 @@
...
-__PACKAGE__->mk_accessors(qw( encoder ));
-
-sub new
-{
-    my ($class, $c, $arguments) = @_;
-
-    my $self = $class->next::method($c);
-    # we are not using properly encoded unicode strings and the JSON
-    # module expects them, so we need to cheat here a little
-    $self->encoder(JSON->new->latin1);
```
mwiencek: (I don't understand why this was being done to begin with.)

PR #688, opened 2018-06-21
Add code to keep WS1 alive on old search when Solr is being used

samj1912:


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -1042,14 +1042,14 @@ sub xml_search
...
-    if (DBDefs->SEARCH_ENGINE eq 'LUCENE') {
+    if ($version eq "1" or DBDefs->SEARCH_ENGINE eq 'LUCENE') {
         $search_url_string = "http://%s/ws/$version/%s/?query=%s&offset=%s&max=%s&fmt=xml";
     } else {
         $search_url_string = "http://%s/%s/edismax?q=%s&start=%s&rows=%s&wt=mbxml";
     }
 
     my $search_url = sprintf($search_url_string,
-                                 DBDefs->SEARCH_SERVER,
+                                 $version eq "1" ? DBDefs->OLD_SEARCH_SERVER : DBDefs->SEARCH_SERVER,
```
mwiencek: Apparently the `xml_search` subroutine in here is *only* used by WS1, so the version check isn't necessary, but it's good for clarity. (WS2 uses the `xml_search` in `Data::WebService`. Not sure why they're separate.)

PR #689, opened 2018-07-01
SOLR-90: Fix tag lookups for artist

samj1912:
Tag lookups for artist was using a search modifier to search for sortnames and aliases with a particular boost. This is not needed or supported anymore (since it uses a weird syntax) with Solr since these fields are searched by default with a particular set of boosts


samj1912:
@mwiencek updated.

samj1912:
@brainzbot retest this

samj1912:

goodenep:
Just tested and Artist only tag lookup still does not work.

yvanzo:
@goodenep : It is now fixed for every MusicBrainz server, check [artist only tag lookup](https://musicbrainz.org/taglookup?tag-lookup.artist=beatles).

PR #692, opened 2018-07-08
MBS-9757: Search page form is sometimes in the wrong language

mwiencek:
Fixes the same bug as MBS-9673 (see ccdb147). The idea is to delay the localization of the strings from when the module is loaded until the request lifecycle.

### DIFF
--- a/root/components/FormRowRadio.js
+++ b/root/components/FormRowRadio.js
@@ -41,7 +41,7 @@ const FormRowRadio = ({
               value={option.value}
             />
             {' '}
-            {option.label}
+            {option.label.toLocaleString()}
           </label>
           {index < options.length - 1 ? <br /> : null}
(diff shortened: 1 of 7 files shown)


mwiencek:
> Maybe rename toString to toLocaleString or autoExpand or any more explicit name.

I forgot to mention that I wanted to continue supporting strings for `option.label` in `FormRowRadio` and `SelectField`, since it likely will be a string in most cases, and `toString` also works on strings (it just returns the same string back). The `toString` method of an object is also called automatically when coerced to a string like `String(obj)` or `'' + obj`, sort of like `__str__` in Python.

Luckily `String.prototype.toLocaleString` also exists, and the default implementation inherited from `Object` just calls `toString()`, so we can use that for free.

PR #693, opened 2018-07-09
MBS-9760: Convert the event search results page to React

mwiencek:
Note: The "Artists" and "Location" columns in the results will be blank for indexed search, but that's not a new bug.


root/search/components/EventResults.js
```diff
@@ -0,0 +1,115 @@
...
+  const event = result.entity;
+  const score = result.score;
+
+  return (
+    <tr className={loopParity(index)} key={event.id}>
+      <td>{score}</td>
+      <td>
+        <DescriptiveLink entity={event} />
+      </td>
+      <td>{event.typeName ? l_attributes(event.typeName) : null}</td>
```
yvanzo: Context is missing for localization, should be:
```js
lp_attributes(event.typeName, 'event_type')
```
I just found the same bug for area results and artist results components.

mwiencek: Thanks for the fix. :)

PR #695, opened 2018-07-12
MBS-9767: Hide search scores from the UI

yvanzo:
# Resolve [MBS-9767](https://tickets.metabrainz.org/browse/MBS-9767):

Hide search scores from the UI by moving it to the added attribute `data-score` for each result.

### DIFF
--- a/root/search/components/AreaResults.js
+++ b/root/search/components/AreaResults.js
@@ -27,8 +27,7 @@ function buildResult(result, index) {
   const score = result.score;
 
   return (
-    <tr className={loopParity(index)} key={area.id}>
-      <td>{score}</td>
+    <tr className={loopParity(index)} data-score={score} key={area.id}>
       <td>
         <DescriptiveLink entity={area} />
       </td>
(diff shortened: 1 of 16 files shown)


root/search/components/AreaResults.js
```diff
@@ -27,8 +27,7 @@ function buildResult(result, index) {
   const score = result.score;
 
   return (
-    <tr className={loopParity(index)} key={area.id}>
-      <td>{score}</td>
+    <tr className={loopParity(index)} dataScore={score} key={area.id}>
```
mwiencek: I think the data attributes aren't camelCased in React for some reason, so you have to use `data-score` still.

yvanzo:
@brainzbot test this please

root/search/components/AreaResults.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/695/commits/da66c002d6d50d9abb77fc4feefd5dc5efade0f9

PR #699, opened 2018-08-12
MBS-9699 (part 1/2): Convert user account pages to React/JSX (but forms)

yvanzo:
# [MBS-9699](https://tickets.metabrainz.org/browse/MBS-9699) (Part 1/2): Convert user account pages to React/JSX that don’t contain any form.

Additionally extend the component `PaginatedResults`, convert `*NotFound` components to ES6 syntax, and factorize the component `NotFound` into `StatusPage`. The idea is to extend `StatusPage` later on with additional properties (about status: success, error, warning, and so on) that could be used to decide on page decoration.


root/account/applications/Index.js
```diff
@@ -0,0 +1,134 @@
...
+        {l('Revoke Access')}
+      </a>
+    </td>
+  </tr>
+);
+
+function formatScopes(token: EditorOAuthTokenT) {
+  const lScopes = [];
+
+  token.permissions.forEach((perm) => {
```
mwiencek: You can initialize the list with a map like `const lScopes = token.permissions.map(perm => l(perm));` to be slightly more expressive

root/account/applications/Index.js
```diff
@@ -0,0 +1,134 @@
...
+        </PaginatedResults>
+      ) : (
+        <p>{l('You have not authorized any applications.')}</p>
+      )}
+
+    <h2>{l('Developer Applications')}</h2>
+
+    <p>
+      {l('Do you want to develop an application that uses the {ws|MusicBrainz web service}? {register|Register an application} to generate OAuth tokens. See our {oauth2|OAuth documentation} for more details.', {
+        __react: 'true',
```
mwiencek: Small nit, it should be a boolean without quotes (here and elsewhere). It doesn't really matter except that it might not type-check later, and I think uglify can minify `true` to `!1`.

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -145,7 +146,13 @@ sub lost_password : Path('/lost-password') ForbiddenOnSlaves
     my ($self, $c) = @_;
 
     if (exists $c->request->params->{sent}) {
-        $c->stash(template => 'account/lost_password_sent.tt');
+        $c->stash(
+            current_view => 'Node',
+            component_path => 'account/LostPasswordSent',
+            component_props => {
+                contactURL => $CONTACT_URL,
```
mwiencek: Should we define it in root/static/scripts/common/constants.js instead of passing it as a prop?

root/account/Donation.js
```diff
@@ -0,0 +1,55 @@
...
+  <StatusPage title={l('Donation Check')}>
+    {nag
+      ? (
+        <Frag>
+          <p>
+            {l('We have not received a donation from you recently. If you have just made a PayPal donation, then we have not received a notification from PayPal yet. Please wait a few minutes and reload this page to check again.')}
+          </p>
+          <p>
+            {l('If you would like to make a donation, {donate|you can do that here}. If you have donated, but you are still being nagged, please {contact|contact us}.',
+              {__react: 'true', contact: contactURL, donate: 'https://metabrainz.org/donate'})}
```
mwiencek: Maybe we should add `DONATE_URL` to constants.js too? It's also used in root/layout/components/Footer.js

root/account/applications/Index.js
yvanzo: Done at https://github.com/metabrainz/musicbrainz-server/pull/699/commits/79ae598a8cec954a67da07209377d94c74ae508a#diff-4c3d028cef4917d20675baed1fb0ce28R54

root/account/applications/Index.js
yvanzo: Fixed at https://github.com/metabrainz/musicbrainz-server/pull/699/commits/79ae598a8cec954a67da07209377d94c74ae508a#diff-4c3d028cef4917d20675baed1fb0ce28R99 and everywhere else

lib/MusicBrainz/Server/Controller/Account.pm
yvanzo: So be it with https://github.com/metabrainz/musicbrainz-server/pull/699/commits/8b5059909982e2bbf8bde1493b2c2547495c6a29.
That’s definitely better than passing it as a prop.
Should we define it outside of `root/static` as it is not useful to client-side scripts?

root/account/Donation.js
yvanzo: Done with https://github.com/metabrainz/musicbrainz-server/pull/699/commits/0d4d014f6ab9fccef6852d9ffd4f3e16a09b8260. Same question as for `CONTACT_URL`.

lib/MusicBrainz/Server/Controller/Account.pm
mwiencek: Hmm, sure, it sounds useful to have a root/constants that's not bundled for the client. (Haven't checked if anything else could be moved there.)

yvanzo: Done in https://github.com/metabrainz/musicbrainz-server/pull/699/commits/5b775265bd5497e9ed5bf3cec03dcf69d7ac5844 and https://github.com/metabrainz/musicbrainz-server/pull/699/commits/e0f9d92c4f5e34bfde9fe66f9564c42e73d46e39.

PR #700, opened 2018-08-12
Add React URL Form Field

yvanzo:


root/components/FormRow.js
```diff
@@ -10,12 +10,21 @@
...
+
 type Props = {
   +children: ReactNode,
 };
 
+function hasFormLabel (children: ReactNode) {
+  return typeof child.type !== 'string'
+    && React.Component.prototype.isPrototypeOf((child.type as any).prototype))
+    && ((child.type as any).prototype instanceof FormLabel
+      || children.props.children.some(hasFormLabel));
```
mwiencek: There are some cases where this might break, e.g. we use `no-label` in cases where there's a label with an `inline` class (for checkboxes I believe).

I think it would be nicer to add a `noLabel` or `className` prop to extend the class name, since we always know where there's no label and can avoid recursing the html at runtime to find it.

yvanzo: Agreed, that was a pointless attempt. Prop added in https://github.com/metabrainz/musicbrainz-server/pull/700/commits/b876fd601ebfcbd8420c6057ac18f3f5e33f6533, used in https://github.com/metabrainz/musicbrainz-server/pull/700/commits/56b9330558e6b0bd4249ad302fa4ecedb83e4263.

PR #703, opened 2018-08-14
Render entity links in release/relationship editors with React

mwiencek:
Removes the lodash `_.template` code which was used to render links, and plugs in the React components which we've had for a while. It's a bit ugly, but not as bad as before.

### DIFF
--- a/root/static/scripts/common/components/EditorLink.js
+++ b/root/static/scripts/common/components/EditorLink.js
@@ -18,11 +18,16 @@ const EditorLink = ({editor, content, avatarSize, subPath}) => {
     avatarSize = 12;
   }
 
-  const gravatar = editor.gravatar + '&s=' + (avatarSize * 2);
+  let gravatar;
+  if (editor.gravatar) {
+    gravatar = editor.gravatar + '&s=' + (avatarSize * 2);
+  }
 
   return (
     <a href={entityHref(editor, subPath)}>
-      <img src={gravatar} height={avatarSize} width={avatarSize} className="gravatar" alt="" />
+      {gravatar ? (
+        <img src={gravatar} height={avatarSize} width={avatarSize} className="gravatar" alt="" />
+      ) : null}
       {isolateText(content)}
     </a>
   );
--- a/root/static/scripts/common/components/EntityLink.js
+++ b/root/static/scripts/common/components/EntityLink.js
@@ -203,6 +203,15 @@ const EntityLink = ({
     }
   }
 
+  if (!subPath && entity.entityType === 'recording' && entity.video) {
+    content = (
+      <Frag>
+        <span className="video" title={l('This recording is a video')} />
+        {content}
+      </Frag>
+    );
+  }
+
   if (!showDisambiguation && !infoLink) {
     return content;
   }
--- a/root/static/scripts/common/entity.js
+++ b/root/static/scripts/common/entity.js
@@ -9,6 +9,8 @@ const React = require('react');
 const ReactDOMServer = require('react-dom/server');
 
 const ArtistCreditLink = require('./components/ArtistCreditLink');
+const EditorLink = require('./components/EditorLink');
+const EntityLink = require('./components/EntityLink');
 const {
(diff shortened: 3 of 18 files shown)


root/static/scripts/common/components/EditorLink.js
```diff
@@ -18,11 +18,16 @@ const EditorLink = ({editor, content, avatarSize, subPath}) => {
     avatarSize = 12;
   }
 
-  const gravatar = editor.gravatar + '&s=' + (avatarSize * 2);
+  let gravatar;
+  if (editor.gravatar) {
+    gravatar = editor.gravatar + '&s=' + (avatarSize * 2);
```
mwiencek: The editor classs knockout uses doesn't store gravatars, so I'm making that prop optional here.

root/static/scripts/common/components/EntityLink.js
```diff
@@ -218,7 +218,14 @@ const EntityLink = ({
         <Comment className="comment" comment={comment} key="comment" />,
       );
     }
-    if (entity.entityType === 'area') {
+    if (entity.entityType === 'recording') {
+      parts.push(
+        ' ',
+        <span className="comment" key="video">
+          {bracketed(l('video'), {__react: true})}
```
mwiencek: I think it should be fine to always show `(video)` now, but let me know if that would be out of place somewhere.

root/static/scripts/common/entity.js
```diff
@@ -177,32 +197,16 @@ const formatTrackLength = require('./utility/formatTrackLength');
...
-        "<% if (data.video) { %> <span class=\"comment\">" +
-        "(<%- data.videoString %>)</span><% } %>" +
-        "<% if (data.nameVariation) { %></span><% } %>" +
-        "<% if (data.editsPending) { %></span><% } %>",
-        {variable: "data"}
-    );
-
-    class Editor extends CoreEntity {}
+    class Editor extends CoreEntity {
+        html() {
```
mwiencek: I'm not positive, but I don't think this `html` method is ever called anywhere. I believe this was just a stub class to allow the editor search autocompletes to work (and that doesn't render editor links either). I've ported it over just in case.

root/static/scripts/common/components/EntityLink.js
yvanzo: Is it about [MBS-9306](https://tickets.metabrainz.org/browse/MBS-9306)?

On one hand, it would be redundant with the video cell/icon in recordings tab for artist page, but on the other, it would add missing info to the recordings (rels) table in work page.

Btw, I spotted another bug about video cell/icon ruining layout in recording search results, e.g. https://musicbrainz.org/search?query=Last+Angel&type=recording

It would be nice to display the video icon everywhere for consistency.

mwiencek: I think it's an unrelated issue with the inline search according to the [linked thread](https://community.metabrainz.org/t/track-recording-titles-for-bonus-dvds/244281/15) (I'm also having trouble reproducing it).

Good point about the video icons...should we remove the `(video)` disambiguation everywhere, but display the icon everywhere in its place? Or we can keep `(video)` in the inline search (in addition to the icon), but only show the icon next to links.

mwiencek: I pushed some changes to standardize the video recording display, and fix the recording search results.

The changes are described in 231a8e2acb0bf7f063a31e6d173315814dd73b2f. Let me know how that looks.

yvanzo: Better! That many bugs fixed at once! :fireworks: 

Assuming the `title` attribute of the video icon `span` tag is used by screen readers, maybe the video icon can replace the (video) comment? That would realign titles and remove redundancy.

mwiencek: Removed the (video) comment (agreed it seems redundant now) but I left the video icon on the left side of the links since that's how we display country flags and editor avatars too. Though there might be a good reason to make them different.

PR #704, opened 2018-08-15
Remove MB.i18n references from knockout bindings

mwiencek:
This is a cleanup which improves a few things:

1. Removes the `i18n` reference from the global `MB` object.

2. Removes some ugly hacks where we had to pass strings translated by TT to a JavaScript function (`expand`). Now we can just use `l` directly in the JS.

3. Interpolating raw HTML into a translated string will no longer be possible soon: all raw strings will be HTML-escaped, so you'll have to use JSX to interpolate markup. Moving these instances out of the knockout templates will allow us to easily fix them later.

### DIFF
--- a/root/static/scripts/common/MB.js
+++ b/root/static/scripts/common/MB.js
@@ -35,9 +35,6 @@ const MB: {[string]: *} = {
 
     // Hold constants for knockout templates that depend on globals.
     constants,
-
-    // Deprecated reference needed by knockout templates
-    i18n: require('./i18n')
 };
 
 global.MB = MB;
--- a/root/static/scripts/release-editor/dialogs.js
+++ b/root/static/scripts/release-editor/dialogs.js
@@ -6,9 +6,13 @@
 const $ = require('jquery');
 const ko = require('knockout');
 const _ = require('lodash');
+const React = require('react');
+const ReactDOMServer = require('react-dom/server');
 
+const Frag = require('../../../components/Frag');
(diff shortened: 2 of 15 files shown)


root/static/scripts/release-editor/dialogs.js
```diff
@@ -124,6 +128,29 @@ class SearchResult {
             track.artistCredit = [{ name: track.artist }];
         }
     }
+
+    link() {
```
mwiencek: Here's one of the strings that changed. See https://github.com/metabrainz/musicbrainz-server/pull/704/files#diff-6edb40ffe5e08874aaa80abd6c4d1045L123

I think it's an improvement since I removed the `by {artist}` string (concatenated to the release name) and reused an existing `{release} by {artist}` string. Not sure if the way I handle the format after the release name is the best.

root/static/scripts/common/MB.js
```diff
@@ -35,9 +35,6 @@ const MB: {[string]: *} = {
 
     // Hold constants for knockout templates that depend on globals.
     constants,
-
-    // Deprecated reference needed by knockout templates
-    i18n: require('./i18n')
```
mwiencek: The problem with globals like this is that technologies like Webpack can only look at the imports to create a dependency tree and decide how to bundle things. It would have no idea where this global variable is used, thus we have to include i18n.js on every page (which maybe is necessary in this example, but isn't in others - it can't tell either way).

PR #705, opened 2018-08-17
Solr edismax endpoint was renamed to advanced

mwiencek:
According to https://github.com/metabrainz/mb-solr-cloud/pull/2#issuecomment-413768128

### DIFF
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -741,8 +741,8 @@ sub external_search
         if (!$adv)
         {
             # Solr has a bug where the dismax end point behaves differently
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -741,8 +741,8 @@ sub external_search
         if (!$adv)
         {
             # Solr has a bug where the dismax end point behaves differently
-            # from edismax when the query size is 1. This is a fix for that
-            # See https://issues.apache.org/jira/browse/SOLR-12409
+            # from edismax (advanced) when the query size is 1. This is a fix
```
mwiencek: I assume this comment is still relevant (that the advanced endpoint still uses the edismax query parser), but correct me if I'm wrong.

PR #707, opened 2018-08-21
Update Babel and use builtin <> syntax

mwiencek:
### Bump Babel to v7.0.0-rc.1

The stage presets were deprecated in favor of specifying the plugins you actually use.

### Replace <Frag> component with builtin <> syntax

Babel 7 supports `<>`, so we can use that now. It's one less thing we have to import to use a basic feature.

In cases where we need to specify a key prop on the fragment, I used `React.Fragment` directly.


yvanzo:
LGTM but the build failure.

mwiencek:
Fixed, there was a small issue in the order that the babel plugins were running.

mwiencek:
And 7.0.0-rc.2 was released right after I opened this, so updated to use that now.

PR #708, opened 2018-08-23
MBS-9699 (1⅛/2): Convert the Register Application page to React/JSX

yvanzo:
Convert the user account page that allows registering application to React/JSX, part of [MBS-9699](https://tickets.metabrainz.org/browse/MBS-9699).


root/account/applications/Register.js
```diff
@@ -0,0 +1,93 @@
...
+  constructor(props: Props) {
+    super(props);
+    this.state = {withURL: (props.form.field.oauth_type.value || oauthTypeOptions.options[0].value) === 'web'};
+    //this.handleOauthTypeChange = this.handleOauthTypeChange.bind(this);
+  }
+
+  //handleOauthTypeChange: (event: SyntheticEvent<HTMLSelectElement>) => void;
+
+  handleOauthTypeChange(e: SyntheticEvent<HTMLSelectElement>) {
+    (this: any).setState({withURL: e.currentTarget.value === 'web'});
```
yvanzo: Cast to `any` to please `flow`. `handleOauthTypeChange` is never called anyway.

root/account/applications/Register.js
```diff
@@ -0,0 +1,93 @@
...
+          <form action="/account/applications/register" method="post">
+            <FormRowText
+              field={this.props.form.field.name}
+              label={l('Name:')}
+              required
+            />
+            <FormRowSelect
+              field={this.props.form.field.oauth_type}
+              label={l('Type:')}
+              onChange={this.handleOauthTypeChange}
```
yvanzo: That is supposed to update the component state and thus to show/hide the URL field.

root/account/applications/Register.js
```diff
@@ -71,10 +71,11 @@ class RegisterApplication extends React.Component<Props, State> {
               label={l('Type:')}
               onChange={this.handleOauthTypeChange}
               options={oauthTypeOptions}
+              value={this.state.oauthType}
```
yvanzo: Additional vain attempt to sync the selected value with an elaborate component state.

root/account/applications/Register.js
```diff
@@ -0,0 +1,93 @@
...
+    {label: N_l('Web Application'), value: 'web'},
+    {label: N_l('Installed Application'), value: 'installed'},
+  ],
+};
+
+class RegisterApplication extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {withURL: (props.form.field.oauth_type.value || oauthTypeOptions.options[0].value) === 'web'};
+    //this.handleOauthTypeChange = this.handleOauthTypeChange.bind(this);
```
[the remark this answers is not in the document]
mwiencek: It looks correct to me. What issue are you seeing?

root/account/applications/Register.js
mwiencek: It looks like the cause is the `+oauthType` in `State`. You need to remove the `+` so that `oauthType` is writable. Though it still thinks `string` is incompatible with `OauthTypeT` so you'll still need to cast `(e.currentTarget.value: any)` (not sure of another workaround).

root/account/applications/Register.js
mwiencek: My mental concept of the React form stuff is that the `form` should be stored in state (unless you're using redux, like in root/static/scripts/work.js), and you should update the form object so that the state (values, errors) stay in one place. One way to do that while keeping the state immutable is using the lens utility:

```JavaScript
const {Lens, prop, set, compose3} = require('../../static/scripts/common/utility/lens');

const oauthTypeFieldLens: Lens<ApplicationRegisterForm, OauthTypeT> =
  compose3(prop('field'), prop('oauth_type'), prop('value'));

// ...
  handleOauthTypeChange(e: SyntheticEvent<HTMLSelectElement>) {
    this.setState({
      form: set(oauthTypeFieldLens, (e.currentTarget.value: any), this.state.form),
    });
  }
// ...
```

The advantage is that we keep all of the form state in a single place, and the form components only have to look at the field they receive and not any overrides. The form/fields also store errors in addition to values, so it will make more sense once we're displaying dynamic field errors too.

root/components/SelectField.js
```diff
@@ -28,20 +28,22 @@ type Props = {|
...
   options,
+  value,
 }: Props) => (
   <select
     className="with-button"
     id={'id-' + field.html_name}
     name={field.html_name}
     onChange={onChange}
-    value={getSelectValue(field, options, allowEmpty)}
+    value={value || getSelectValue(field, options, allowEmpty)}
```
mwiencek: There might be an issue here for blank options with `value=""`. Though I would prefer to keep all the "current" state inside `field` as mentioned in https://github.com/metabrainz/musicbrainz-server/pull/708#discussion_r212449001 and not allow any overrides.

root/account/applications/Register.js
mwiencek: Let me know if root/utility/hydrate.js makes sense. Basically you'll need to factor out a component just to hold the `<form />` in a separate file, say `RegisterApplicationForm`, since the rest of the page is static. Then `export default hydrate('register-application', RegisterApplicationForm);` from there. The first argument `'register-application'` is the class name it'll use for the wrapper div that acts as the mount point for `RegisterApplicationForm`.

root/components/SelectField.js
yvanzo: The following message can be dismissed by initializing text values to `''` in Perl:
> Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
Not sure it is the best way to address it though.

yvanzo:
This is now working, except the URL field is emptied when it is hidden/shown.

root/account/applications/RegisterForm.js
```diff
@@ -0,0 +1,109 @@
...
+const oauthTypeFieldLens: Lens<RegisterApplicationFormT, OauthTypeT> =
+  compose3(prop('field'), prop('oauth_type'), prop('value'));
+
+class RegisterApplicationForm extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      form: set(
+        oauthTypeFieldLens,
+        (getSelectValue(
```
mwiencek: Not sure I understand this bit. Shouldn't the correct value be in `props.form` already so you can just set `this.state = {form: props.form}`?

root/account/applications/Register.js
```diff
@@ -0,0 +1,27 @@
...
+import {l} from '../../static/scripts/common/i18n';
+
+import RegisterApplicationForm from './RegisterForm';
+import type {RegisterApplicationFormPropsT} from './RegisterForm';
+
+const RegisterApplication = (props: RegisterApplicationFormPropsT) => (
+  <Layout fullWidth title={l('Applications')}>
+    <h1>{l('Register Application')}</h1>
+    <RegisterApplicationForm {...props} />
+    <script async="async" src={manifest.pathTo('/scripts/account/applications/register.js')} />
```
mwiencek: Hm, that's always a 404 for me. You'll need to add something like this to the gulpfile so that it's bundled properly:

```JavaScript
runYarb('account/applications/register.js', function (b) {
  b.external(commonBundle);
});
```

There's also a `manifest.js` utility you can use to replace the `<script/>`, to get the correct path:

```JavaScript
{manifest.js('account/applications/register', {async: 'async'})}
```

mwiencek:
About the URL field issue, I got a warning in the console which linked to https://reactjs.org/docs/events.html#event-pooling

The problem is that in `handleOauthTypeChange`, `this.setState` is asynchronous, so you have to store `e.currentTarget.value` in a variable before calling it.

root/account/applications/RegisterForm.js
yvanzo: It should, but it wasn’t.
It is now set from https://github.com/metabrainz/musicbrainz-server/pull/708/commits/5220b4f3c742e02924f2594db2fa497d638a6609#diff-5053a22c07db9b62eceff446a9b99731R710
Note that other fields are set with default value `''` as React is unhappy with `null`.

root/account/applications/Register.js
yvanzo: Oops, I got it right first, then removed `runYarb` and it continued to work, probably due to some cache. That explained why the handler for `oauth_redirect_uri` didn’t work then at that time.

yvanzo:
Good catch, surprisingly I did not have this warning…

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -705,6 +705,18 @@ sub register_application : Path('/account/applications/register') RequireAuth Re
         });
         $c->response->redirect($c->uri_for_action('/account/applications'));
         $c->detach;
+    } else {
+        $form->field('oauth_redirect_uri')->value('');
+        $form->field('oauth_type')->value('web');
+        $form->field('name')->value('');
```
mwiencek: These will clear/change the fields if there's an error in the submission, though.

After I remove these lines, I can't seem to reproduce the warning you're getting in https://github.com/metabrainz/musicbrainz-server/pull/708#discussion_r212809463 (which I assumed this is meant to resolve). How can I trigger that?

yvanzo:

lib/MusicBrainz/Server/Controller/Account.pm
mwiencek: Thanks! Now I see the horrible error.

I think I agree with your last sentence (setting the `default` in the form class), so that the component doesn't have to deal with weird data. There doesn't seem to be any purpose in keeping the default value to `undef` here (if anywhere; not sure why FormHandler does that, but perhaps to distinguish between unset fields and fields explicitly set to `''`). At least now we know about the issue.

yvanzo: Applied with https://github.com/metabrainz/musicbrainz-server/pull/708/commits/c76f6eb4461f9b03162201d60689bb6b7338b8ad.

Note: In either case, unset fields should be set to JS `undefined` not JS `null` for React.

PR #712, opened 2018-09-03
MBS-8062: query for performers and locations when rendering external event search

reosarevok:
Event search was not loading additional (relationship-based) info for external search, only for direct search. Changing it to do so, following the example of works.

load_ids is basically the same method as for works - not sure if we have some way of generalising this that might be good here.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Event.pm
+++ b/lib/MusicBrainz/Server/Data/Event.pm
@@ -318,6 +318,31 @@ sub find_related_entities
     return %map;
 }
 
+=method load_ids
+
+Load internal IDs for event objects that only have GIDs.
+
+=cut
+
+sub load_ids
+{
+    my ($self, @events) = @_;
+
+    my @gids = map { $_->gid } @events;
+    return () unless @gids;
+
+    my $query = "
+        SELECT gid, id FROM event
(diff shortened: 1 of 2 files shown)


yvanzo:

reosarevok:
Changed the commit message. Are these things written somewhere where I can check and get used to them? :)

yvanzo:
Thanks. These are common practices. Some links:
- https://git-scm.com/docs/git-commit#_discussion
- https://duckduckgo.com/?q=how+to+write+commit+message

yvanzo:

yvanzo:
@reosarevok: Jenkins seems to get stuck on it. Can you `git rebase master` and `git push --force` it?

lib/MusicBrainz/Server/Data/Event.pm
```diff
@@ -318,6 +318,31 @@ sub find_related_entities
     return %map;
 }
 
+=method load_ids
+
+Load internal IDs for work objects that only have GIDs.
```
mwiencek: `s/work/event/` here.

Looks good otherwise except there's still some unrelated commits here.

reosarevok:
Think it should be better now on both counts!

PR #713, opened 2018-09-03
MBS-9763: clean up setlist.fm to https

reosarevok:
This was requested to ensure duplicate http and https can't be added. They seem to redirect to https anyway nowadays, so we might as well.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1170,6 +1170,9 @@ const CLEANUPS = {
   setlistfm: {
     match: [new RegExp("^(https?://)?([^/]+\\.)?setlist\\.fm","i")],
     type: LINK_TYPES.setlistfm,
+    clean: function (url) {
+      return url.replace(/^http:\/\//, "https://");
+    },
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1460,7 +1463,7 @@ const CLEANUPS = {
           case LINK_TYPES.otherdatabases.work:
             return prefix == 'works/work';
           case LINK_TYPES.otherdatabases.artist:
-            return prefix !== 'works/work';	
+            return prefix !== 'works/work';
```
reosarevok: Not sure how this happened (my IDE or me) but it seems it makes it more consistent so I didn't revert anyway.

yvanzo: It was inconsistent, thus it is fine to remove it, but it should preferably be done in a separate commit. Anyway, https://github.com/metabrainz/musicbrainz-server/commit/d247b5c219ae231b9da2e24b4de3d70583601624#diff-87814c26df342be7e4c2a85531bbd47cL1463 has already been merged.

PR #715, opened 2018-09-04
MBS-9788: Indicate "series" in series edit lists

reosarevok:
This fixes an issue where, say, a release series just says "Release" in edit lists, making it look like the edit list for a release.

Alternatively, we could keep the code as it is, but rename all the existing types to include "series" when appropriate, as with instruments (e.g. "string instrument"). I'm fine with this if it is the preferred solution.

### DIFF
--- a/root/series/SeriesHeader.js
+++ b/root/series/SeriesHeader.js
@@ -9,6 +9,7 @@
 
 const React = require('react');
 const {lp} = require('../static/scripts/common/i18n');
+const {lp_attributes} = require('../static/scripts/common/i18n/attributes');
 const EntityHeader = require('../components/EntityHeader');
 
 type Props = {|
@@ -21,7 +22,7 @@ const SeriesHeader = ({series, page}: Props) => (
     entity={series}
     headerClass="seriesheader"
     page={page}
-    subHeading={lp('Series', 'singular')}
+    subHeading={series.typeName ? lp_attributes(series.typeName, 'series_type') : lp('Series', 'singular')}
   />
 );
 


root/entity/edits.tt
```diff
@@ -12,6 +12,7 @@
     SET layout_title = all_edits ? l('Edits for {name}', { name => name_for_title }) : l('Open Edits for {name}', { name => name_for_title });
     SET heading = all_edits ? l('Edits for {name}', { name => link_for_heading }) : l('Open Edits for {name}', { name => link_for_heading });
     SET subheading = html_escape(entity.l_type_name or format_entity_type_name(ent_type));
+    SET subheading = l('{series_type} series', { series_type => subheading }) IF ent_type == 'series';
```
yvanzo: It is not consistent with subheading for other series pages at https://github.com/metabrainz/musicbrainz-server/blob/7d2a56fe1bd8a4f6b691aab72a177768db4e7439/root/series/SeriesHeader.js#L24.

reosarevok: It's not, but it wasn't before either - series pages show just Series with the type on the sidebar, while the edit list does not have a sidebar to show the type. Do you feel we should just show "series" on all? Or show the type on all?

yvanzo: Either way, just make sure it can be localized correctly. `lp('Series', 'singular')` is already localized while `'{series_type} series'` is probably not localizable. It is composed of the singular `{series_type}` whereas plural term is required in many/most languages.

mwiencek: Specifying the series type in the subheader seems like a nice change to me, though I'd show it on all (change the series header yvanzo linked to, too).

For proper localization you may need a macro that specifies each string separately, something like this:

```
[%~ MACRO series_subheading(type) BLOCK;
    IF type = 'artist';
        lp('Artist series', 'singular');
    ELSIF type = 'event':
        lp('Event series', 'singular');
    # ... etc.
    END;
END -%]
```

reosarevok: My only problem with that is needing to update it every time we add a new series type, but I guess we can live with that :) Will do.

mwiencek: You're right, though there are other spots we use this technique and we really only have to update them every $x years, so it's manageable. :P

yvanzo: Alternatively, you may directly use the name of series type if it can be changed as follows:
- _Catalogue_ :ok:,
- _Event_ → _Event series_,
- _Festival_ :ok:,
- _Recording_ → _Recording series_,
- …

Series type names would be more consistent if they always designate the same thing: here the series type rather than the series item type. It would be consistent with other entity type attributes too.

However, such change probably has larger implications.

reosarevok:
Changed this to just show the specific type everywhere - fixed the original issue by improving the type names in the DB.

yvanzo:
Sorry, I missed another consistency issue: We do not do that for other entity types. You even [mentioned](https://github.com/metabrainz/musicbrainz-server/pull/715#discussion_r214902703) motivation behind this: Type attribute is displayed in the sidebar already. Plus, entity base type is displayed at least once (but in the `/edits` page).

reosarevok:
Sure, but it actually feels to me that it makes sense to make the type more visible - "X: Catalogue" looks much more obvious than "X: Series" where you need to look on the side to figure out what can be linked to it. So I think I'd still like to keep it this way. We do it this way for Place, which also has the type on the sidebar.

mwiencek:
Don't we display the type name in the subheader for most entity types? The only other ones I can see where we don't are release groups and labels (or recordings and URLs, which don't have types, though we display "Video" for recordings where applicable, which is sort of a type).

PR #717, opened 2018-09-04
MBS-9699 (1⅐/2): Convert the rest of application forms to React/JSX

yvanzo:
Following pull requests #699 and #708, it converts lasting forms for editing/removing/revoking user applications to React/JSX as part of [MBS-9699](https://tickets.metabrainz.org/browse/MBS-9699).


root/layout/components/Footer.js
```diff
@@ -36,7 +36,7 @@ const Footer = ({$c, ...props}) => {
         ]}
 
         {!!DBDefs.GIT_BRANCH && [
```
mwiencek: Alternatively, if we remove the array here and use a fragment (I guess this was written before we had those), we wouldn't need keys.

root/account/applications/Edit.js
```diff
@@ -0,0 +1,56 @@
...
+import type {OauthTypeT, ApplicationFormT} from './types';
+
+type Props = {|
+  +form: ApplicationFormT,
+|};
+
+const EditApplication = ({form}: Props) => (
+  <Layout fullWidth title={l('Edit Application')}>
+    <h1>{l('Edit Application')}</h1>
+    <form method="post">
```
mwiencek: Any way you can reuse root/account/applications/RegisterForm.js here? Might need some additional props to change some of the labels, but at least the `onChange` props will be a no-op on the server, as well as the `hydrate` stuff (as long as you don't include client JS, that'll be a no-op too). So it could possibly be reused without too many changes.

I guess it would be good to figure out how to do this since we probably won't want to duplicate all our entity forms between `/create` and `/edit` either, for example.

root/layout/components/Footer.js
yvanzo: Done in https://github.com/metabrainz/musicbrainz-server/pull/717/commits/993ed78551d2d25d056139f4c59ec8e4f4d2e0b3

root/account/applications/Edit.js
yvanzo: Done in https://github.com/metabrainz/musicbrainz-server/pull/717/commits/cc0bc8af84772df2256e6bdf265e57d77421eac7 by adding `action` and `submitLabel` to the props.

root/account/components/ApplicationForm.js
```diff
@@ -73,25 +75,26 @@ class RegisterApplicationForm extends React.Component<Props, State> {
...
     }));
   }
 
   render() {
     return (
-      <form action="/account/applications/register" method="post">
+      <form method="post">
         <FormRowText
           field={this.state.form.field.name}
           label={addColon(l('Name'))}
```
mwiencek: So since this file contains translated strings, it has to be moved to somewhere under root/static/scripts/ per https://github.com/metabrainz/musicbrainz-server/pull/723

mwiencek:
I get "Oauth type field is required" when trying to edit an application now (I assume because browsers don't send disabled fields in form submissions).

mwiencek:
When I click on any remove link, I get this error:
```
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.
    at invariant (/Users/michael/code/musicbrainz-server/node_modules/fbjs/lib/invariant.js:42:15)
    at ReactDOMServerRenderer.render (/Users/michael/code/musicbrainz-server/node_modules/react-dom/cjs/react-dom-server.node.development.js:2477:7)
    at ReactDOMServerRenderer.read (/Users/michael/code/musicbrainz-server/node_modules/react-dom/cjs/react-dom-server.node.development.js:2307:19)
    at Object.renderToString (/Users/michael/code/musicbrainz-server/node_modules/react-dom/cjs/react-dom-server.node.development.js:2679:25)
    at getResponse (/Users/michael/code/musicbrainz-server/root/server/response.js:76:31)
    at Socket.<anonymous> (/Users/michael/code/musicbrainz-server/root/server/createServer.js:71:31)
    at bound (domain.js:396:14)
    at Socket.runBound (domain.js:409:12)
    at Socket.emit (events.js:182:13)
    at Socket.EventEmitter.emit (domain.js:442:20)
    at addChunk (_stream_readable.js:283:12)
    at readableAddChunk (_stream_readable.js:264:11)
    at Socket.Readable.push (_stream_readable.js:219:10)
    at Pipe.onread (net.js:639:20)
```

yvanzo:
> I get "Oauth type field is required" when trying to edit an application now (I assume because browsers don't send disabled fields in form submissions).

Indeed, fixed with https://github.com/metabrainz/musicbrainz-server/pull/717/commits/ea9c97547d51a9014039c8e424113776fbf199e7 by adding an hidden input (https://github.com/metabrainz/musicbrainz-server/pull/717/commits/379e45f97136a2cd048fc0f9ce8a60e67ccfdd33) after select tag (https://github.com/metabrainz/musicbrainz-server/pull/717/commits/985f16a99208395515d0d68c44accdaba4df6616).

> When I click on any remove link, I get this error: `Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.`

There was a syntactic error in both Remove and RevokeAccess components, fixed in https://github.com/metabrainz/musicbrainz-server/pull/717/commits/d7833291b3092129a2e944e4d32ff3bfe147a59d and https://github.com/metabrainz/musicbrainz-server/pull/717/commits/a45c0b3a413360f2c0cbe5ba44f6d80bfd86444b. (ConfirmLayout and Nominate components are fine.)

yvanzo:
Additionally forward the parameter `required` from form row components to input/select tag in https://github.com/metabrainz/musicbrainz-server/pull/717/commits/7ee57e4c0ad97f4f2d0b0d0483639d816f26500d.

mwiencek:
Better :) New issue: when I go to remove an application and click "Cancel" when it asks me if I'm sure, it removes it anyway.

yvanzo:
This issue was unreported rather than _new_. It is now tracked at [MBS-9844](https://tickets.metabrainz.org/browse/MBS-9844) and fixed with https://github.com/metabrainz/musicbrainz-server/pull/717/commits/591f7b6adfd686dca99c0a5bc61daacd0c948cd5.

PR #718, opened 2018-09-05
MBS-9699 (1⅙/2): Convert the account preferences form to React/JSX

yvanzo:
Convert the user account page that allow editing preferences to React/JSX as part of [MBS-9699](https://tickets.metabrainz.org/browse/MBS-9699).

Additionally:
* Fix a link in the page for saved preferences (converted in pull request #699);
* Convert the tabs component and the layout template for user/account pages to React/JSX;
* Improve regional setting fields with:
  * date/time format options built according to selected timezone;
  * helper button to guess user timezone;
* Remove unused checkbox (_Notify me by email on new releases to my watched artists_) for not yet implemented feature [MBS-3285](https://tickets.metabrainz.org/browse/MBS-3285) which should be go to branch `metabrainz/mbs-3285` instead.


yvanzo:
Work-in-progress is about solving two issues:
- date/time format selection cannot be changed;
- preferences form is not properly localized.

root/components/UserAccountLayout.js
```diff
@@ -0,0 +1,49 @@
...
+ * @flow
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+import type {Node as ReactNode} from 'react';
```
mwiencek: Apparently the line above this one imports all the types too, according to https://flow.org/en/docs/react/types/, so you can reference `React.Node` without this line.

root/static/scripts/account/components/PreferencesForm.js
```diff
@@ -0,0 +1,256 @@
...
+  +form: PreferencesFormT,
+|};
+
+type State = {|
+  form: PreferencesFormT,
+|};
+
+const timezoneOptions = {
+  grouped: false,
+  options: moment.tz.names().map(v => ({
```
mwiencek: These options aren't in sync with what `DateTime::TimeZone` has. :(

https://gist.github.com/mwiencek/3261f94f4e4c6bb63eeb28348bd86535

So, many of those options (e.g. Africa/Addis_Ababa) will be rejected by the form. I guess `MusicBrainz::Server::Plugin::UserDate` won't work with them either since it uses `DateTime::TimeZone` under the hood.

Maybe updating both modules will solve this if the latest versions have the latest zoneinfo.

root/components/UserAccountLayout.js
```diff
@@ -0,0 +1,49 @@
...
+  page,
+  user,
+  title,
+  ...layoutProps
+}: Props) => (
+  <Layout
+    fullWidth
+    title={
+      l('Editor “{user}”', {user: user.name}) +
+      (title ? ' - ' + title : '')}
```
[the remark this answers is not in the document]
yvanzo: Yes, it is! Done with https://github.com/metabrainz/musicbrainz-server/pull/718/commits/1131789cdb8d27f0f37f0c86209264a7fc09dfa7.

root/components/UserAccountLayout.js
yvanzo: Indeed. Removed with https://github.com/metabrainz/musicbrainz-server/pull/718/commits/862762256e00e7266b46284a41c9a4425cf216e8.

root/static/scripts/account/components/PreferencesForm.js
yvanzo: I updated both Javascript `moment-timezone` and Perl `DateTime::TimeZone` to use the latest IANA TimeZone Database (2018e) with https://github.com/metabrainz/musicbrainz-server/pull/718/commits/30ae79cfe23699aa2dcbf1c3a89fc7683544567a. How to update `cpanfile.snapshot` from there?

It still doesn’t solve the issue as JS module includes timezone aliases while Perl module doesn’t. Forking either module is not desirable (even though it is probably possible for moment-timezone). The best path seems to be serializing options for the field timezone from Perl to JSON. This is exactly the same issue I addressed at https://github.com/metabrainz/musicbrainz-server/pull/730/commits/6bf052f9e07bc0125130410a409dcff7924a0638#diff-5053a22c07db9b62eceff446a9b99731R414. Shouldn’t options (or at least some set of valid values) be serialized for most of Select fields? Would it be fine to implement it from [Form::Role::ToJSON#L46](https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Form/Role/ToJSON.pm#L46)?

mwiencek: Yeah, I agree we should serialize the options for the form somewhere. It might be tricky to do that in `Form::Role::ToJSON` in general though, because some select options require special customizations or have optgroups added (see `MusicBrainz::Server::Form::Utils`), so we'd need a way to customize the serialization for certain fields. Also, for repeatable fields the options are associated with the subfield, but we only want to serialize them once for the entire form. Just putting the options on the parent repeatable may not work if there are nested repeatables.

For now, maybe we should just serialize the options in the controller, using one of the subroutines from `MusicBrainz::Server::Form::Utils`, then pass them as a prop to the top-level form component. What do you think?

yvanzo: I agree with serializing options in the controller and using the convention `<fieldname>_options`  for now, that is, on the model of `language_options` in the above mentioned example https://github.com/metabrainz/musicbrainz-server/commit/6bf052f9e07bc0125130410a409dcff7924a0638#diff-5053a22c07db9b62eceff446a9b99731R414

Timezone options are serialized the same way at https://github.com/metabrainz/musicbrainz-server/pull/718/commits/e43cd919655f2876c267196503ac28ba2f557214#diff-5053a22c07db9b62eceff446a9b99731R484

yvanzo:
Just rebased on master and fixed new conflicts.

yvanzo:
For some reason, build test fails again whereas it passed just before rebasing on master. I suspect a potential incompatibility with `WWW::Mechanize` even though `root/user/login.tt` did not change since 2015. Would that be alright to drop [the only occurrence of `WWW::Mechanize::tick`](https://github.com/metabrainz/musicbrainz-server/blob/master/t/lib/t/MusicBrainz/Server/Controller/Admin/DeleteEditor.pm#L18)?

root/components/FormRowSelect.js
```diff
@@ -52,7 +54,7 @@ const FormRowSelect = ({
         options={options}
         required={required}
       />
-      {frozen ? <HiddenField field={field} /> : null}
+      {frozen ? <HiddenField field={field} /> : helpers}
```
mwiencek: It's a bit non-intuitive to me that you can only show `helpers` when `frozen` is `false` - they seem unrelated (e.g. I can imagine having a ? button explaining why the field is disabled or what it is).

root/components/UserAccountTabs.js
```diff
@@ -0,0 +1,91 @@
...
+    <a href={path}>{title}</a>
+  </li>
+);
+
+function buildTabs(
+  $c: CatalystContextT,
+  user: EditorT,
+  page: string,
+): React.Node {
+  const viewOwnProfile = $c.user && $c.user.id === user.id;
```
mwiencek: Small nit, the old template had s/view/viewing/ and I think it makes more grammatical sense.

mwiencek:
Looks great! Thanks for the related improvement to generate_cpanfile_snapshot.sh.

root/static/scripts/account/components/PreferencesForm.js
```diff
@@ -0,0 +1,257 @@
...
+    this.handleSubscriptionsEmailPeriodChange =
+      this.handleSubscriptionsEmailPeriodChange.bind(this);
+  }
+
+  handleTimezoneChange: (e: SyntheticEvent<HTMLSelectElement>) => void;
+
+  handleTimezoneChange(e: SyntheticEvent<HTMLSelectElement>) {
+    const selectedTimezone = e.currentTarget.value;
+    this.setState(prevState => ({
+      form: set(
```
mwiencek: As we convert more forms to React maybe we'll want to (later) write some helper utility to generate all these lenses/event handlers for us based on a concise description of the fields.

root/components/UserAccountTabs.js
yvanzo: fixed in https://github.com/metabrainz/musicbrainz-server/pull/718/commits/32cdf4950da69520edb47ffaa776e2383b35c7cb

root/components/FormRowSelect.js
yvanzo: fixed in https://github.com/metabrainz/musicbrainz-server/pull/718/commits/558f1358f6ea589e42ac275f592aa6fffd6e09b6

PR #721, opened 2018-09-07
MBS-9208, MBS-9210: Increase bcrypt cost parameter to 12, for all active accounts

yvanzo:
Address two related issues:
- [MBS-9210](https://tickets.metabrainz.org/browse/MBS-9210): Update password hash on login if bcrypt cost parameter increased, so as to gradually rehash password for all active accounts.
- [MBS-9208](https://tickets.metabrainz.org/browse/MBS-9208): Increase bcrypt cost parameter to 12.

### DIFF
--- a/lib/MusicBrainz/Server/Constants.pm
+++ b/lib/MusicBrainz/Server/Constants.pm
@@ -81,6 +81,7 @@ our @EXPORT_OK = (
         %ENTITIES_WITH_RELATIONSHIP_CREDITS
         %ENTITIES entities_with @RELATABLE_ENTITIES
         $EDITOR_SANITISED_COLUMNS
+        $PASSPHRASE_BCRYPT_COST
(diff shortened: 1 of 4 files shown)


mwiencek:
I had trouble getting this to work and realized it was because I was logging in as `bitmap` but my username is stored as `Bitmap`. It allows me to login, but doesn't upgrade the password in that case.

The solution would be to change the query in `Data::Editor::update_password` to use `WHERE lower(name) = lower(?)` instead of the current `WHERE name = ?`.

yvanzo:
Reported as [MBS-9816](https://tickets.metabrainz.org/browse/MBS-9816) and fixed in https://github.com/metabrainz/musicbrainz-server/pull/721/commits/af2f7d3c7b18373d84495ff8bebb7224bbb3bfb2.

PR #716, opened 2018-09-04
Convert search results to React

reosarevok:
This includes [MBS-9800](https://tickets.metabrainz.org/browse/MBS-9800), [MBS-9802](https://tickets.metabrainz.org/browse/MBS-9802), [MBS-9803](https://tickets.metabrainz.org/browse/MBS-9803), [MBS-9805](https://tickets.metabrainz.org/browse/MBS-9805), [MBS-9806](https://tickets.metabrainz.org/browse/MBS-9806), [MBS-9807](https://tickets.metabrainz.org/browse/MBS-9807), [MBS-9808](https://tickets.metabrainz.org/browse/MBS-9808), [MBS-9809](https://tickets.metabrainz.org/browse/MBS-9809), [MBS-9810](https://tickets.metabrainz.org/browse/MBS-9810), [MBS-9818](https://tickets.metabrainz.org/browse/MBS-9818), [MBS-9819](https://tickets.metabrainz.org/browse/MBS-9819), [MBS-9823](https://tickets.metabrainz.org/browse/MBS-9823)

PR #722, opened 2018-09-07
MBS-9604: Convert ISWC index to React

reosarevok:
A large part of this duplicates https://github.com/metabrainz/musicbrainz-server/pull/716/commits/8cecce8119acbacbb852f6d0c8535ac675092d31 but it should be a good enough duplication that no conflicts happen anyway.


root/iswc/Index.js
```diff
@@ -0,0 +1,130 @@
...
+                  </ul>
+                </td>
+                <td>{work.typeName ? lp_attributes(work.typeName, 'work_type') : null}</td>
+                <td>
+                  <ul>
+                    {work.languages.map(r => (
+                      <li key={r.language_id}>
+                        <abbr title={l_languages(r.language.name)}>
+                          {r.language.iso_code_3}
+                        </abbr>
```
yvanzo: There is enough place in this page to display language name rather than just its code.

root/iswc/Index.js
```diff
@@ -0,0 +1,130 @@
...
+                      name="add-to-merge"
+                      type="checkbox"
+                      value={work.id}
+                    />
+                  </td>
+                ) : null}
+                <td><EntityLink entity={work} /></td>
+                <td>
+                  <ul>
+                    {work.writers.map(r =>(
```
yvanzo: Please use meaningful variable name, here and below. :writing_hand:

yvanzo: Yes, for example. The same for arguments of other `map`. :+1:

root/iswc/Index.js
yvanzo: Small fixes/improvements are welcome too as long as they are not doubtful.
As you stated it is currently broken, parity is not even wished here. :wink:

reosarevok: I agree with that bit, but just thought it might make sense to have consistency in all search results (so, have all language/script shown normally or with the code). At the very least, I'd want this one to match the Work search results (where we currently also show code only). Would you (and @mwiencek) suggest to change both to use work name instead?

lib/MusicBrainz/Server/Entity/Language.pm
```diff
@@ -63,6 +63,7 @@ around TO_JSON => sub {
 
     my $json = $self->$orig;
     $json->{name} = $self->name;
+    $json->{iso_code_3} = $self->iso_code_3;
```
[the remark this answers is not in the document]
reosarevok: I imagine it is because it's more consistent (BCP47 seems to use code 1 when available and if not code 3, which means a combination of 2 and 3 letter codes and seems to give unneeded priority to larger languages).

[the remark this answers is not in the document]
yvanzo: About _consistency_, we provide BCP47 in JSON-LD property [inLanguage](https://schema.org/inLanguage).

lib/MusicBrainz/Server/Entity/WorkLanguage.pm
```diff
@@ -20,6 +20,15 @@ has language => (
     isa => 'Language',
 );
 
+sub TO_JSON {
+    my ($self) = @_;
+
+    return {
+        language_id => $self->language_id,
+        language => $self->language->TO_JSON,
+    };
```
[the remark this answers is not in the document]
mwiencek: One potential issue with that is `WorkLanguage` has an `edits_pending` property to indicate if a particular entry is being edited. We don't appear to using it, actually, but we probably should later.

root/iswc/Index.js
mwiencek: I'm okay with changing both if it's not difficult.

reosarevok: It's trivial (after all, we already load the name for the abbr hover). Should we still display the codes in any way, if we change this?

Ideally we would also change the release results, but language and script *do* compete with a lot of other things in that table. I'm just thinking that if that's the only place we use the abbreviations, it might feel kinda strange.

mwiencek: Yeah, not sure about the release results, and it would be strange if they were inconsistent. Let's just keep parity in this PR?

lib/MusicBrainz/Server/Entity/WorkLanguage.pm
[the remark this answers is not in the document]
yvanzo: (property from Role::Editable through Role::ValueSet)

`language.id` would still be a better fit than `language_id` as it will be needed for release editor.

root/iswc/Index.js
yvanzo: Honestly I just feel like the whole `abbr` is buggish, but it ain't the main purpose of this PR.

lib/MusicBrainz/Server/Entity/Work.pm
```diff
@@ -103,21 +105,16 @@ no Moose;
 1;
 
 =head1 COPYRIGHT
-
```
mwiencek: Any reason these line breaks are being removed?

lib/MusicBrainz/Server/Entity/WorkLanguage.pm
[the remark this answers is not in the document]
mwiencek: Agreed that we can remove `language_id` at least.

lib/MusicBrainz/Server/Entity/Work.pm
reosarevok: Huh. No idea, so I guess I'll blame my IDE. Will revert.

root/iswc/Index.js
```diff
@@ -0,0 +1,130 @@
...
+              <th>{l('Title')}</th>
+              <th>{l('Writers')}</th>
+              <th>{l('Artists')}</th>
+              <th>{l('Type')}</th>
+              <th>{l('Language')}</th>
+            </tr>
+          </thead>
+          <tbody>
+            {works.map((work, index) => (
+              <tr
```
mwiencek: This duplicates 95% of `buildResult` in root/search/components/WorkResults.js so it would be nice to create a reusable component for both pages to share.

You'll just need to add props to hide the iswcs column and possibly the merge checkboxes.

mwiencek: To clarify, the [WorkResults.js in your search PR](https://github.com/metabrainz/musicbrainz-server/pull/716/files#diff-b63869045ac7691d7b0edeedbfed8784R29) (forgot it's not visible here).

PR #727, opened 2018-09-13
MBS-9470: Expose UUID of relationship attribute type

reosarevok:
Goes with https://github.com/metabrainz/mmd-schema/pull/16

For JSON, created new attribute-ids section. This is not ideal - we *already* have two "number" attributes with different meanings and IDs which will be indistinguishable from one another when using name as key - but it's probably the only non-breaking option.


mwiencek:
Merged the mmd-schema changes. Can you look into the test failures? :)

reosarevok:
This seems to be passing tests locally... so I'm not sure why it's failing here.

mwiencek:
Updating `MMD_SCHEMA_COMMIT` seems to fix things.

lib/MusicBrainz/Server/WebService/XMLSerializer.pm
```diff
@@ -1061,11 +1061,11 @@ sub _serialize_relation
     push @list, $gen->attribute_list(
         map {
             if (non_empty($_->text_value)) {
-                $gen->attribute({ value => $_->text_value }, $_->type->name);
+                $gen->attribute({ value => $_->text_value, id => $_->type->gid }, $_->type->name);
```
mwiencek: I think `type-id` is more consistent with the rest of our XML than `id`.

mwiencek:
Re-fixed mmd-schema, LGTM now.

PR #728, opened 2018-09-13
MBS-7192: Show work types in inline search

reosarevok:
Chose to use the same way as we do with artists and writers because the more consistent / usual way (parentheses after the name) was blocked by the way we do work languages. Alternatively, we could do the usual with types *and* move languages to a "Languages: X, Y" line like writers / artists.

### DIFF
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -12,6 +12,7 @@ const MB_entity = require('../../entity');
 const i18n = require('../../i18n');
 const commaOnlyList = require('../../i18n/commaOnlyList');
 import {l_languages} from '../../i18n/languages';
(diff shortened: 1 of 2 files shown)


root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -762,6 +762,11 @@ MB.Control.autocomplete_formatters = {
                      _.escape(commaOnlyList(comment)) + ')</span>');
         }
 
+        if (item.typeName)
+        {
+            a.append('<br /><span class="autocomplete-comment"> Type: ' + _.escape(item.typeName) + '</span>');
```
mwiencek: About your suggestion to "do the usual with types and move languages to a 'Languages: X, Y' line like writers / artists," I guess the question is whether languages are more important than types for works?

Anyway, if we keep the display you have here, the "Type:" part needs to be fixed so that it's translated.

PR #733, opened 2018-10-07
MBS-9846: Fix Beatport URL cleanup

yvanzo:
## Fix bug [MBS-9846](https://tickets.metabrainz.org/browse/MBS-9846) about broken Beatport URL cleanup

Additionally improve cleanup code to better handle legacy format (documented in [MBS-6130](https://tickets.metabrainz.org/browse/MBS-6130)) and stop handling URL without slug (which may come from Beatport redirect).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -597,15 +597,29 @@ const CLEANUPS = {
     type: LINK_TYPES.downloadpurchase,
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(?:(?:classic|pro|www)\.)?beatport\.com\//, "https://www.beatport.com/");
-      url = url.replace(/^(https:\/\/www\.beatport\.com)\/[\w-]+\/html\/content\/([\w-]+)\/0*([0-9]+)\/([\w-]+).*$/, "$1/$2/$4/$3");
-      url = url.replace(/^(https:\/\/www\.beatport\.com)\/[\w-]+\/html\/content\/([\w-]+)\/0*([0-9]+)(?:[\/?#].*)?$/, "$1/$2/-/$3");
-      url = url.replace(/^(https:\/\/www\.beatport\.com)\/([\w-]+)\/([\w-]+)\/0*([0-9]+).*$/, "$1/$2/$3/$4");
-      url = url.replace(/^(https:\/\/www\.beatport\.com)\/([\w-]+)\/\/+0*([0-9]+)(?:[\/?#].*)?$/, "$1/$2/-/$3");
-      url = url.replace(/^(https:\/\/www\.beatport\.com)\/([\w-]+)\/0*([0-9]+)(?:[\/?#].*)?$/, "$1/$2/$3");
+      var m = url.match(/^(https:\/\/www\.beatport\.com)\/[\w-]+\/html\/content\/([\w-]+)\/detail\/0*([0-9]+)\/([^\/?&#]*).*$/);
+      if (m) {
+        const slug = m[4].toLowerCase()
+          .replace(/%21/g, '!')
+          .replace(/%23/g, '-pound-')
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -498,6 +498,27 @@ const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/U
...
+        },
+        {
+                             input_url: 'https://www.beatport.com/en-US/html/content/release/detail/161035/Back%20To%20The%20Future',
+                     input_entity_type: 'release',
+            expected_relationship_type: 'downloadpurchase',
+                    expected_clean_url: 'https://www.beatport.com/release/back-to-the-future/161035',
+               only_valid_entity_types: ['release']
+        },
+        {
+                             input_url: 'https://www.beatport.com/en-US/html/content/release/detail/06130/22%%21%20&tracks#',
```
mwiencek: This URL is a 400 error (rather than just a 404, assuming it's made up) - is the URL malformed in some way?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -597,15 +597,28 @@ const CLEANUPS = {
...
+          .replace(/%24/g, '-money-').replace(/\$/g, '-money-')
+          .replace(/%25/g, '-percent-')
+          .replace(/%26/g, '-and-')
+          .replace(/%40/g, '-at-').replace(/@/g, '-at-')
+          .replace(/%[0-9a-f]{2}/g, '-')
+          .replace(/%/g, '-percent-')
+          .replace(/[^a-z0-9!]/g, '-')
+          .replace(/-+/g, '-')
+          .replace(/^-|-$/g, '')
+          .replace(/^-?$/g, '---');
```
mwiencek: Not sure what these last three replacements are doing, but I'll hope they're covered by some test already. :)

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: Good catch, the error comes from the first `%` sign which is not encoded properly.

root/static/scripts/edit/URLCleanup.js
yvanzo: It strip extraneous hyphens and replace empty slug with 3 hyphens (see track [?????? ??????](https://www.beatport.com/release/---/1990510) for example). They are fully covered by the new made up example.

yvanzo:
Made small changes (keep handling URL without slug because they may come from Beatport redirect), improved commit message, and commented tests.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -498,6 +498,41 @@ const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/U
...
+        },
+        {                               // Legacy URL format (real example)
+                             input_url: 'https://www.beatport.com/en-US/html/content/release/detail/161035/Back%20To%20The%20Future',
+                     input_entity_type: 'release',
+            expected_relationship_type: 'downloadpurchase',
+                    expected_clean_url: 'https://www.beatport.com/release/back-to-the-future/161035',
+               only_valid_entity_types: ['release']
+        },
+        {                               // Legacy URL format (made up to test slug conversion)
+                             input_url: 'https://www.beatport.com/en-US/html/content/release/detail/06130/%40@%26%25%24$%23%22%21!-&tracks#',
```
mwiencek: Still a 400 error? (Not sure if intentional.)

yvanzo: Not intentional but it doesn’t matter as it is a made up example that spares about 10 real examples (about 1 per character, that can be found on Beatport). It is only supposed to cover the conversion of all reserved characters at once (as well as removing leading/repeated/trailing hyphens), not necessarily to resolve on Beatport.

PR #736, opened 2018-11-02
MOVED TO #742: MBS-9827: Redirect Merged Entities

Ge0rg3:
As discussed in tickets [MBS-9827](https://tickets.metabrainz.org/browse/MBS-9827) and [MBS-4937](https://tickets.metabrainz.org/browse/MBS-4937), it was difficult to tell whether you were on the page of a merged ~~artist~~ entity. This PR will cause a 301 redirect to the "correct" page if this is not the case.  
Currently, the only way to tell is through the **link rel="canonical"** tag.

i.e:
[f21a407e-3af9-4539-ab3d-c92a5230dff6](https://musicbrainz.org/artist/f21a407e-3af9-4539-ab3d-c92a5230dff6) will redirect you to [1155431a-d35e-4863-9ae0-e3c24eb61aa9](https://musicbrainz.org/artist/1155431a-d35e-4863-9ae0-e3c24eb61aa9).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Load.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Load.pm
@@ -108,7 +108,7 @@ role
 
         if ($entity) {
             if ($id_is_guid && $entity->gid ne $id) {
-                $c->res->redirect($c->uri_for($c->action, [$entity->gid]), 301);
+                $c->res->redirect($c->uri_for($c->action, [$entity->gid], $c->req->params), 301);
             }
             $c->model($model)->load_gid_redirects($entity) if exists $entity_properties->{mbid} && $entity_properties->{mbid}{multiple};
             return $entity;


yvanzo:
@brainzbot test this please.

lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -107,6 +107,9 @@ role
         }
 
         if ($entity) {
+          if ($entity->gid != $id) {
```
mwiencek: Just a few changes here:

```suggestion
            if ($id_is_guid && $entity->gid ne $id) {
```

(1) You need to include `$id_is_guid` in the check since `$entity` won't always have a `gid`, and (2) use `ne` since `!=` is for numeric comparisons in Perl.

Indentation was also misaligned, it should be 4 spaces per level rather than 2 (same for below).

lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -107,6 +107,9 @@ role
...
+            $c->res->redirect($c->uri_for($c->action,[$entity->gid]), 301);
```
mwiencek:
```suggestion
                $c->res->redirect($c->uri_for($c->action, [$entity->gid]), 301);
```

lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -107,6 +107,9 @@ role
...
+          }
```
mwiencek:
```suggestion
            }        
```

Freso:
If anything, I'd say it would be good to have some tests for this as well, maybe even before it gets merged into MBS, but for the GCI side of this, it looks to me like it's good. If @mwiencek or @yvanzo can't find any flaws in the code as-is, I'm happy to approve the GCI task, and maybe we can make a follow-up task for you to write tests for this new redirect. ;)

lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -107,6 +107,9 @@ role
         }
 
         if ($entity) {
+            if ($id_is_guid && $entity->gid ne $id) {
+                $c->res->redirect($c->uri_for($c->action, [$entity->gid]), 301);
```
mwiencek: Ah, another thing: you need to include the old query parameters in the new URL (this is especially important for the web service).

```suggestion
                $c->res->redirect($c->uri_for($c->action, [$entity->gid], $c->req->params), 301);
```

mwiencek:
> I'm happy to approve the GCI task, and maybe we can make a follow-up task for you to write tests for this new redirect. ;)

That sounds great!

Ge0rg3:
@mwiencek I've tested the recommended addition, and have now committed it 👍

mwiencek:
@brainzbot test this please

PR #741, opened 2018-11-08
MBS-9765: Do not display event dates twice in search results

Ge0rg3:
### [MBS-9765](https://tickets.metabrainz.org/browse/MBS-9765):

> Event dates are displayed:
>1. in parenthesis right after the name in the column “Name”,
>2. in the column “Date”.
>
> Example: https://musicbrainz.org/search?query=test&type=event
> It is redundant and unnecessarily widens the event results page.

I also merged the date & time columns, since otherwise they were on the other side of the screen (or taking too much space next to eachother as seperate columns).
<br/>
Before:
image

After:
image

### DIFF
--- a/root/search/components/EventResults.js
+++ b/root/search/components/EventResults.js
@@ -14,6 +14,7 @@ import {l} from '../../static/scripts/common/i18n';
 import {lp_attributes} from '../../static/scripts/common/i18n/attributes';
 import ArtistRoles from '../../static/scripts/common/components/ArtistRoles';
 import DescriptiveLink from '../../static/scripts/common/components/DescriptiveLink';
+import EntityLink from '../../static/scripts/common/components/EntityLink';
 import formatDatePeriod from '../../static/scripts/common/utility/formatDatePeriod';
 import loopParity from '../../utility/loopParity';
 import type {ResultsPropsT} from '../types';
@@ -28,8 +29,10 @@ function buildResult(result, index) {
   return (
     <tr className={loopParity(index)} data-score={score} key={event.id}>
       <td>
-        <DescriptiveLink entity={event} />
+        <EntityLink entity={event} showDisambiguation showEventDate={false} />
       </td>
+      <td>{formatDatePeriod(event)}</td>
+      <td>{event.time}</td>
       <td>{event.typeName ? lp_attributes(event.typeName, 'event_type') : null}</td>
       <td>
         <ArtistRoles relations={event.performers} />
@@ -48,8 +51,6 @@ function buildResult(result, index) {
           ))}
         </ul>
       </td>
-      <td>{formatDatePeriod(event)}</td>
-      <td>{event.time}</td>
     </tr>
   );
 }
@@ -68,11 +69,11 @@ const EventResults = ({
       columns={
         <>
           <th>{l('Name')}</th>
+          <th>{l('Date')}</th>
+          <th>{l('Time')}</th>
           <th>{l('Type')}</th>
           <th>{l('Artists')}</th>
           <th>{l('Location')}</th>
-          <th>{l('Date')}</th>
-          <th>{l('Time')}</th>
         </>
       }
       pager={pager}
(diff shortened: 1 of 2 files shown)


yvanzo:
@brainzbot, add to whitelist.

root/search/components/EventResults.js
```diff
@@ -28,8 +28,9 @@ function buildResult(result, index) {
   return (
     <tr className={loopParity(index)} data-score={score} key={event.id}>
       <td>
-        <DescriptiveLink entity={event} />
+        <a href={'/event/'+event.id}>{event.name}</a>
```
yvanzo: Disambiguation comment and cancellation are missing here, whereas only date should be removed.

Another way to do it is to add options `DescriptiveLink` and `EntityLink` to skip adding date for event.

root/search/components/EventResults.js
```diff
@@ -28,8 +28,9 @@ function buildResult(result, index) {
...
       </td>
+      <td>{formatDatePeriod(event)}{event.time ? ", "+event.time : null}</td>
```
yvanzo: The comma separator is not localizable. The string concatenation doesn’t support right-to-left locales either. I suggest adding a function instead, following the example of `i18n.addColon`.

root/search/components/EventResults.js
mwiencek: I agree with adding a prop like `showEventDate` to `EntityLink`. Not sure there's a reason to be using `DescriptiveLink` here since it's just a wrapper around `EntityLink` that doesn't appear to add anything for events.

root/search/components/EventResults.js
mwiencek: I like the columns separate, personally, since it's easier for me to read.

Ge0rg3: So should I put them together or keep them seperate?

mwiencek: Merging the columns seems unrelated to the ticket, so I'd probably leave it out of the PR unless there's a better rationale ("they were on the other side of the screen" doesn't make sense to me since you can move them without merging them, and "taking too much space" doesn't appear to be true to me).

There are some other places we show time in a separate column so it introduces another inconsistency is all.

Ge0rg3: Okay, no problem--I was just acting on the comments on the Jira ticket :)

root/search/components/EventResults.js
Ge0rg3: Sure, that seems like a far cleaner approach.

Ge0rg3: @mwiencek why would we want to use showEventDate? Surely that shows the event date?

mwiencek: Well, as @yvanzo pointed out we still want to show the disambiguation comment and "canceled" status, so you'd need:

```JSX
<EntityLink entity={event} showDisambiguation />
```

But `showDisambiguation` also shows the dates. So by adding a `showEventDate` flag (defaulting to true) you can set it to false there:

```JSX
<EntityLink entity={event} showDisambiguation showEventDate={false} />
```

Ge0rg3: Ah right, I think I understand. I'd wrongly assumed that you could only add flags like this to add to a component. and hadn't thought about setting anything to false 😄 

Will commit this in a sec

root/search/components/EventResults.js
yvanzo: Thank you for having tried at least!

It was suggested as _optional_ because it might be a hassle for little gain. In the case of dates range, it would be more clear to have start time just after start date rather than after end date.

root/static/scripts/common/components/EntityLink.js
```diff
@@ -41,9 +41,9 @@ const Comment = ({className, comment}: {|+className: string, +comment: string|})
   </>
 );
 
-const EventDisambiguation = ({event}: {|+event: EventT|}) => {
+const EventDisambiguation = ({event, showDate}: {|+event: EventT, +showDate: boolean|}) => {
   const dates = formatDatePeriod(event);
-  if (!dates && !event.cancelled) {
+  if (showDate === false || (!dates && !event.cancelled)) {
```
mwiencek: That'll cause it to return null even if `event.cancelled` is true. The check should be `((!dates || !showDate) && !event.cancelled)`.

Then you need to use `showDate` to exclude the date below:

```JSX
{dates && showDate ? ' ' + bracketed(dates) : null}
```

PR #742, opened 2018-11-10
MBS-9827: Redirect Merged Entities

Ge0rg3:
### Description
As discussed in tickets [MBS-9827](https://tickets.metabrainz.org/browse/MBS-9827) and [MBS-4937](https://tickets.metabrainz.org/browse/MBS-4937), it was difficult to tell whether you were on the page of a merged ~~artist~~ entity. This PR will cause a 301 redirect to the "correct" page if this is not the case.  
Currently, the only way to tell is through the **link rel="canonical"** tag.

i.e:
[f21a407e-3af9-4539-ab3d-c92a5230dff6](https://musicbrainz.org/artist/f21a407e-3af9-4539-ab3d-c92a5230dff6) will redirect you to [1155431a-d35e-4863-9ae0-e3c24eb61aa9](https://musicbrainz.org/artist/1155431a-d35e-4863-9ae0-e3c24eb61aa9).

### Test cases
Selenium test cases for #736.
This tests:
1. Redirection of `/artist/gid`.
2. Redirection of `/artist/gid/releases` (actions).
3. Redirection of `/artist/recordings?filter.name=do` (GET parameters).

The test database has also been updated to add a merged entity.

The order of additions in the SQL file may seem random, but this is because I kept the additions alphabetical where possible (similarly to how it looked before).

### Note:
This PR was originally at #736, but has been moved here due to the original fork being deleted.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Load.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Load.pm
@@ -107,6 +107,11 @@ role
         }
 
         if ($entity) {
+            if ($id_is_guid && $entity->gid ne $id) {
+                my @captures = @{ $c->req->captures };
+                $captures[0] = $entity->gid;
+                $c->res->redirect($c->uri_for($c->action, \@captures, $c->req->params), 301);
+            }
             $c->model($model)->load_gid_redirects($entity) if exists $entity_properties->{mbid} && $entity_properties->{mbid}{multiple};
             return $entity;
         }
--- a/t/selenium.js
+++ b/t/selenium.js
@@ -397,6 +397,7 @@ const seleniumTests = [
   {name: 'Artist_Credit_Editor.html', login: true},
   {name: 'External_Links_Editor.html', login: true, timeout: 90000},
   {name: 'Work_Editor.html', login: true},
+  {name: 'Redirect_Merged_Entities.html', login: true},
   {name: 'release-editor/The_Downward_Spiral.html', login: true, timeout: 120000},
   {name: 'release-editor/Seeding.html', login: true, sql: 'vision_creation_newsun.sql'},
 ];
--- a/t/selenium/Redirect_Merged_Entities.html
+++ b/t/selenium/Redirect_Merged_Entities.html
@@ -1,0 +1,178 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+<head profile="http://selenium-ide.openqa.org/profiles/test-case">
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+<link rel="selenium.base" href="http://localhost:5000" />
+<title>Redirect_Merged_Entities</title>
+</head>
+<body>
+<table cellpadding="1" cellspacing="1" border="1">
+<thead>
+<tr><td rowspan="1" colspan="3">Redirect_Merged_Entities</td></tr>
+</thead><tbody>
+<!--Artist tests (default redirects, actions, GET parameters and two actions)-->
+<tr>
+	<td>open</td>
+	<td>/artist/f21a407e-3af9-4539-ab3d-c92a5230dff6</td>
+	<td></td>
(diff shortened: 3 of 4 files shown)


Ge0rg3:
Do the Jenkins tests use the updated `selenium.sql` file?

Freso:
AFAIK, it *should* use the data available in the source tree that it's testing, so I'm pretty sure it should be using your updated `.sql` file.

Also, I would probably add at least one redirect test for each entity, to make sure nothing pops into the code later that blocks it only for URL or Area or whatever entities, which does not get caught because the tests are only looking at `/artist/`.

Freso:
Also, it seems https://github.com/metabrainz/musicbrainz-server/pull/736 hasn't been merged, so maybe just make this as a commit on top of that one branch and continue the PR there?

Ge0rg3:
> Also, it seems #736 hasn't been merged, so maybe just make this as a commit on top of that one branch and continue the PR there? (Is the test failing because it's relying on redirect code that isn't in `master` yet?)

Ohh of course, that would be it! So I should close this PR, and add the tests to #736?

Ge0rg3:
@Freso turns out that I deleted the fork from `#736` after it was approved.. Would it be best to close both `#736` and this PR, and then open one that includes both the testing and the redirects?

Freso:
> Would it be best to close both #736 and this PR, and then open one that includes both the testing and the redirects?

@mwiencek or @yvanzo, what would be the best approach here?

mwiencek:
At this point I would just keep them as separate PRs - we can review/merge them easily in their current state so no need to complicate things if you already deleted the first branch.

lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -107,6 +107,9 @@ role
         }
 
         if ($entity) {
+            if ($id_is_guid && $entity->gid ne $id) {
+                $c->res->redirect($c->uri_for($c->action, [$entity->gid], $c->req->params), 301);
```
mwiencek: There's a small bug here involving actions with multiple captures. For example, in the path `/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/alias/15504/edit` there are two captures: the gid and the alias id (15504). But when calculating the redirect here we only pass one capture arg (`[$entity->gid]`). Since for core entities the gid will always be the first capture arg, I believe we can get away with just making a copy of the old args and changing the first element:
```suggestion
                my @captures = @{ $c->req->captures };
                $captures[0] = $entity->gid;
                $c->res->redirect($c->uri_for($c->action, \@captures, $c->req->params), 301);
```
Mind fixing that and adding one more test case for /alias/id/edit?

mwiencek:
Other than the multiple-captures issue, the selenium test you added looks good and passes locally. 👍  Very impressive work!

t/selenium/Redirect_Merged_Entities.html
```diff
@@ -0,0 +1,178 @@
...
+	<td></td>
+</tr>
+<tr>
+	<td>open</td>
+	<td>/artist/f21a407e-3af9-4539-ab3d-c92a5230dff6/alias/37382/edit</td>
+	<td></td>
+</tr>
+<tr>
+	<td>assertLocationMatches</td>
+	<td>\/artist\/f21a407e-3af9-4539-ab3d-c92a5230dff6\/alias\/37382\/edit</td>
```
mwiencek: Should be:

```suggestion
	<td>\/artist\/1155431a-d35e-4863-9ae0-e3c24eb61aa9\/alias\/37382\/edit</td>
```

Otherwise it's testing that it doesn't redirect. The reason it passes now is that it's an edit page and you need to be logged-in to view those, so it immediately hits a login screen. You can specify that the test should run while logged-in in t/selenium.js.

t/selenium.js
```diff
@@ -397,6 +397,7 @@ const seleniumTests = [
   {name: 'Artist_Credit_Editor.html', login: true},
   {name: 'External_Links_Editor.html', login: true, timeout: 90000},
   {name: 'Work_Editor.html', login: true},
+  {name: 'Redirect_Merged_Entities.html'},
```
mwiencek: Add `login: true` to run the test logged-in.

t/selenium/Redirect_Merged_Entities.html
Ge0rg3: Whoops...

Ge0rg3:
@mwiencek I've fixed that mistake :)

PR #746, opened 2018-11-15
Fix link to profile from preferences saved dialog

reosarevok:
"profile/" here led to a non-working link, that also caused a situation where someone registering the "profile" username would get seen every time someone changed their prefs. I've also removed the malicious "profile" user, but this fixes it so that the link will actually go to the appropriate user profile.

### DIFF
--- a/root/account/PreferencesSaved.js
+++ b/root/account/PreferencesSaved.js
@@ -21,7 +21,7 @@ const PreferencesSaved = ({$c}: Props) => (
   <StatusPage title={l('Preferences')}>
     <p>
       {l('Your preferences have been saved. Click {link|here} to continue to your user page.',
-        {__react: true, link: $c.user ? '/user/profile/' + $c.user.name : '/register'})}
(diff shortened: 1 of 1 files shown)


root/account/PreferencesSaved.js
```diff
@@ -21,7 +21,7 @@ const PreferencesSaved = ({$c}: Props) => (
   <StatusPage title={l('Preferences')}>
     <p>
       {l('Your preferences have been saved. Click {link|here} to continue to your user page.',
-        {__react: true, link: $c.user ? '/user/profile/' + $c.user.name : '/register'})}
+        {__react: true, link: $c.user ? '/user/' + $c.user.name : '/register'})}
```
mwiencek: Technically the issue is still there in that anyone can create a user named `profile/whatever` or `reosarevok/whatever` and have it link to any user they want. The name has to be URI-encoded:
```suggestion
        {__react: true, link: $c.user ? '/user/' + encodeURIComponent($c.user.name) : '/register'})}
```

PR #748, opened 2018-11-15
MBS-9899: parity for tag/genre display in XML/JSON WS

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9899

Changed this from "do not display tags/genres unless $toplevel" to "do not display tags/genres if loaded as part of a relationship", to mirror https://github.com/reosarevok/musicbrainz-server/blob/master/lib/MusicBrainz/Server/WebService/XMLSerializer.pm#L1222

I see no reason not to make these available for recordings and artists without requiring additional calls (e.g. for Picard), so doing it like the XML WS does seems reasonable.


mwiencek:
I think it's the same issue as MBS-9400 (?)

Not sure 100% parity is desirable here. My suggestion in the comments in MBS-9400 was to only return tags for recording/track artists which differ from the release artist.

reosarevok:
I guess I could change that - it introduces further complexity that I'm not sure makes a significant difference, but I guess it might?

reosarevok:
It's not exactly the same issue in that, unless I'm misremembering, they're also currently missing from recordings in the release (where genres at least are useful).

mwiencek:
Ah yeah, recordings too. I suggest the following for `/ws/2/release/$mbid?inc=tags+genres&fmt=json` then:

 * Include tags/genres in the release AC.
 * Include tags/genres in the release group AC for artists not in the release AC.
 * Include recording tags/genres.
 * Include track/recording tags/genres in the track/recording ACs for artists not in the release AC.

It should help reduce response sizes (even after gzip) for most releases, since most releases are single-artist or single-AC. For example, https://musicbrainz.org/ws/2/release/8067f190-dc3e-362a-8117-8a13df522b2c?inc=aliases%2Bartist-credits%2Bdiscids%2Blabels%2Brecordings+release-groups+tags+genres is 51,824 bytes or 3,573 bytes gzipped, and removing the duplicated tags/genres in the track ACs (which are identical to the release artist) reduces that to 48,088 bytes or 4,077 bytes gzipped. The numbers would be different for JSON, but aggregated across a hundred-thousand requests I believe it'd add up. And that's just about response size, not the CPU time it takes to encode all the additional JSON.

reosarevok:
What's your suggestion for how to do that? The release serializer calls the medium serializer which calls the recording serializer which calls the recording artist serializer - I could pass the release ac all the way down and then compare at the recording serializer level but it seems kinda weird to add some optional argument for the recording serializer. Or would you suggest adding it to the stash maybe and passing it that way?

reosarevok:
Changed it to show track artist tags/genres only if they don't match. They still show aliases even when repeated - would it make sense to change that too? (that said, that'd break something that is currently there so maybe we need to keep it as-is anyway).

I'm not getting any tags/genres for RG artists yet, will look into that later (seems $opts->{tags} is just empty for RG artist for some reason).

reosarevok:
@brainzbot retest this please

mwiencek:
It should be working now. The previous patch didn't accomplish quite what I hoped, which was to compare artists rather than artist credits (the way it was, tags/genres would only be excluded if the artist credits were exactly identical). I added a test which demonstrates the new behavior. Release group tags and recording artist tags (which were also broken) are now working, too. For recording artists, it excludes the tags for artists which exist in the release or associated track AC.

PR #749, opened 2018-11-17
MBS-9901: Change Dailymotion cleanup to https

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9901

Dailymotion itself redirects to https nowadays.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -956,19 +956,19 @@ const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/U
                              input_url: 'https://dailymotion.com/who-knows#uploads',
                      input_entity_type: 'artist',
             expected_relationship_type: 'videochannel',
-                    expected_clean_url: 'http://www.dailymotion.com/who-knows',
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -953,22 +953,22 @@ const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/U
         },
         // Dailymotion
         {
-                             input_url: 'https://dailymotion.com/who-knows#uploads',
+                             input_url: 'http://dailymotion.com/who-knows#uploads',
```
yvanzo: There is no more test with `https` input Dailymotion URL.

reosarevok: Should there be? It's not like https is going to get cleaned up to anything but https. But I can change one back if desired :)

yvanzo: Just making sure the it is still/will still be recognized by the system. This kind of regression happened in the past.

reosarevok: Ok, changed one for this :)

PR #751, opened 2018-11-18
MBS-9873: Add autoselect for Discogs compositions

Ge0rg3:
## [MBS-9873](https://tickets.metabrainz.org/browse/MBS-9873)
Discogs now has support for works (or as they're called on Discogs, 'compositions'). This PR allows these to be recognized.
Unlike all other Discogs entities (apart from users), the IDs aren't numerical, i.e:
* https://www.discogs.com/composition/1b383ed3-7b58-45ff-8f1b-1135c41988d4-In-Verbundenheit
* https://www.discogs.com/composition/d6652f48-35ad-4900-9c42-d0b0f2f84674-Mr-Henri-Rousseaus-Dream
* https://www.discogs.com/composition/27b17569-3e40-40b5-9819-409794c2d5d9-In-The-Hospital

Because of this, I also implemented a new replace statement just for compositions.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -24,7 +24,8 @@ const LINK_TYPES = {
     place: "1c140ac8-8dc2-449e-92cb-52c90d525640",
     release: "4a78823c-1c53-4176-a5f3-58026c76f2bc",
     release_group: "99e550f3-5ab4-3110-b5b9-fe01d970b126",
-    series: "338811ef-b1a9-449d-954e-115846f33a44"
+    series: "338811ef-b1a9-449d-954e-115846f33a44",
+    work: "d78b7280-eb9e-4a57-86c3-cedaa1aa2175"
   },
   imdb: {
     artist: "94c8b0cc-4477-4106-932c-da60e63de61c",
@@ -321,14 +322,14 @@ const CLEANUPS = {
     type: LINK_TYPES.discogs,
     clean: function (url) {
       url = url.replace(/\/viewimages\?release=([0-9]*)/, "/release/$1");
-      url = url.replace(/^https?:\/\/(?:[^.]+\.)?discogs\.com\/(?:.*\/)?(user\/[^\/#?]+|(?:artist|release|master(?:\/view)?|label)\/[0-9]+)(?:[\/#?-].*)?$/, "https://www.discogs.com/$1");
+      url = url.replace(/^https?:\/\/(?:[^.]+\.)?discogs\.com\/(?:.*\/)?(user\/[^\/#?]+|(?:composition\/[^-]+-[^-]+-[^-]+-[^-]+-[^-]+)|(?:artist|release|master(?:\/view)?|label)\/[0-9]+)(?:[\/#?-].*)?$/, "https://www.discogs.com/$1");
       url = url.replace(/^(https:\/\/www\.discogs\.com\/master)\/view\/([0-9]+)$/, "$1/$2");
       return url;
     },
     validate: function (url, id) {
-      var m = /^https:\/\/www\.discogs\.com\/(?:(artist|label|master|release)\/[1-9][0-9]*|(user)\/.+)$/.exec(url);
+      var m = /^https:\/\/www\.discogs\.com\/(?:(artist|label|master|release)\/[1-9][0-9]*|(user)\/.+|(composition)\/(?:[^-]*-){4}[^-]*)$/.exec(url);
       if (m) {
-        var prefix = m[1] || m[2];
+        var prefix = m[1] || m[2] || m[3];
         switch (id) {
           case LINK_TYPES.discogs.artist:
             return prefix === 'artist' || prefix === 'user';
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -321,12 +322,13 @@ const CLEANUPS = {
     type: LINK_TYPES.discogs,
     clean: function (url) {
       url = url.replace(/\/viewimages\?release=([0-9]*)/, "/release/$1");
+      url = url.search(/\/composition\//) != -1 ? url.replace(/\/composition.*/, /\/composition([^-])+-[^-]+-[^-]+-[^-]+-[^-]+/.exec(url)[0]) : url;
```
yvanzo: The calls to `search`, `exec`, and the ternary operator are not needed. KISS!

Ge0rg3: When I don't use the ternary operator to check whether the URL is that of a composition, I get an error with accessing the element at index 0.
Also, without using ".exec", the URL gets replaced with the actual expression itself.
Is there a way that I can avoid these?

yvanzo: Yes, use a single `replace` on the model of the line above.

yvanzo: Additional hint: unlike your regexp, the above line is using _capture group_.

Ge0rg3: Okay, I think I've sorted it out 👍

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1126,6 +1126,18 @@ const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/U
...
                     expected_clean_url: 'https://www.discogs.com/release/5578',
         },
+        {
+                             input_url: 'http://www.discogs.com/composition/27b17569-3e40-40b5-9819-409794c2d5d9-In-The-Hospital',
+                     input_entity_type: 'work',
+            expected_relationship_type: 'discogs',
+                    expected_clean_url: 'https://www.discogs.com/composition/27b17569-3e40-40b5-9819-409794c2d5d9',
+        },
+        {
+                             input_url: 'http://www.discogs.com/composition/1b383ed3-7b58-45ff-8f1b-1135c41988d4-In-Verbundenheit',
```
yvanzo: This test should be dropped as it covers the same case as in the previous test.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -321,12 +322,13 @@ const CLEANUPS = {
     type: LINK_TYPES.discogs,
     clean: function (url) {
       url = url.replace(/\/viewimages\?release=([0-9]*)/, "/release/$1");
-      url = url.replace(/^https?:\/\/(?:[^.]+\.)?discogs\.com\/(?:.*\/)?(user\/[^\/#?]+|(?:artist|release|master(?:\/view)?|label)\/[0-9]+)(?:[\/#?-].*)?$/, "https://www.discogs.com/$1");
+      url = url.replace(/\/composition\/([^-]+-[^-]+-[^-]+-[^-]+-[^-]+).*/, "/composition/$1");
+      url = url.replace(/^https?:\/\/(?:[^.]+\.)?discogs\.com\/(?:.*\/)?(user|composition\/[^\/#?]+|(?:artist|release|master(?:\/view)?|label)\/[0-9]+)(?:[\/#?-].*)?$/, "https://www.discogs.com/$1");
```
yvanzo: Main issue is you broke the pattern `user\/[^\/#?]+`. Plus, you could merge these two added lines by moving the _capture group_ for UUID from the first line into the RegExp on the second line.

Indeed, the purpose of this line is twofold: first, normalizing the heading scheme to `https` and the subdomain to `www`; then, trimming extraneous trailing characters. The preceding line you added is somehow duplicating the second purpose of this line.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -321,12 +322,13 @@ const CLEANUPS = {
...
       url = url.replace(/^(https:\/\/www\.discogs\.com\/master)\/view\/([0-9]+)$/, "$1/$2");
       return url;
     },
     validate: function (url, id) {
-      var m = /^https:\/\/www\.discogs\.com\/(?:(artist|label|master|release)\/[1-9][0-9]*|(user)\/.+)$/.exec(url);
+      var m = /^https:\/\/www\.discogs\.com\/(?:(artist|label|master|release)\/[1-9][0-9]*|(user|composition)\/.+)$/.exec(url);
```
yvanzo: Validation can be improved by checking that the expected UUID after composition is complete.

Ge0rg3: Okay--my new solution checks for 4 seperate "chunks" after /composition/, which means that it's a valid UUID. This *does* mean that I also need to check for m[3], which is unlike any other test case.

yvanzo:
Thank you very much for this contribution and your patience and thoroughness with RegExp!

PR #752, opened 2018-11-18
    MBS-9900:  Add information on requesting instruments/genres to their lists

PROTechThor:
Documentation on how to request a new instrument or a new genre is missing from https://musicbrainz.org/instruments and https://musicbrainz.org/genres. 
This PR has the addition of such documentation. 

https://tickets.metabrainz.org/browse/MBS-9900

### DIFF
--- a/root/genre/List.js
+++ b/root/genre/List.js
@@ -33,11 +33,11 @@ const GenreList = ({genres}: PropsT) => (
         ))}
       </ul>
       <p>
-        {l('Is a genre missing from the list? Request it by')} + 
-         <a href="https://tickets.metabrainz.org/secure/CreateIssueDetails!init.jspa?pid=10032&issuetype=2&summary=Enter%20the%20genre%20name%20here!">
-         {l(' adding a style ticket ')}
-      </a> + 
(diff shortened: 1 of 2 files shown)


yvanzo:
@brainzbot test this please.

root/genre/List.js
```diff
@@ -32,6 +32,13 @@ const GenreList = ({genres}: PropsT) => (
           </li>
         ))}
       </ul>
+      <p>
+        {l('Is a genre missing from the list? Request it by')} +
```
yvanzo: The paragraph should not be split into several strings, as you can assert the localization in other languages will follow the same schema. Please use only one call to `l` instead. There are several examples in the code for embedding links into a localizable string.

PROTechThor: @yvanzo  I was able to find some examples which I used to emulate the syntax. :)

yvanzo:

yvanzo:
Thanks for your contribution!

PROTechThor:
@yvanzo Pleased. :) Thanks for merging. That makes me so proud of myself.

PR #753, opened 2018-11-19
MBS-9748: Add Ko-fi to the Patronage relationship

Ge0rg3:
As described on the [Jira ticket](https://tickets.metabrainz.org/browse/MBS-9748):

> Ko-fi patronage links should be added to the patronage relationship and autoselect/cleanup.
> https://ko-fi.com/35MJZ8OL4IO
> All usernames on the site seem to just be a random string and don't seem to include any parameter lists.

This adds support for them, and a corresponding test case.


Ge0rg3:
Would you like me to also create an entity in lib/MusicBrainz/Server/Entity/URL/ for ko-fi?

yvanzo:
Yes, please do so! Also add favicon and style too. Last, test for `URLCleanup` is missing.

Ge0rg3:
@yvanzo I've used "**Ko-fi**" whenever Ko-fi is just used as a string, and "**Kofi**" when as a filename or package name (as is the case with some of the other URL entities).

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1736,6 +1736,13 @@ const {LINK_TYPES, cleanURL, guessType, validationRules} = require('../../edit/U
                      input_entity_type: 'release',
             expected_relationship_type: 'discographyentry',
         },
+        // Ko-fi
+        {
+                             input_url: 'http://ko-fi.com/35MJZ8OL4IO',
+                     input_entity_type: 'artist',
+            expected_relationship_type: 'patronage',
+                    expected_clean_url: 'https://www.ko-fi.com/35MJZ8OL4IO',
```
yvanzo: You should have removed `ẁww` from the output/expected clean URL rather than from the input URL.

Ge0rg3: Ahh I think I forgot to compile my tests when checking.

Ge0rg3:
@yvanzo tests should actually make sense now!

yvanzo:
Thank you again!

PR #754, opened 2018-11-19
MBS-9905: Tipeee Broken URL validation

Ge0rg3:
## [MBS-9905](https://tickets.metabrainz.org/browse/MBS-9905)
When inputting a Tipeee URL that begins with a language-related subdomain, i.e...

https://en.tipeee.com/borisdalstein

...the URL is not recognised as tipeee. This is the case with all 5 languages that the Tipeee website supports: **en**, **it**, **fr**, **de** & **es**.

An example of a _working_ URL would be:
https://www.tipeee.com/borisdalstein

----

image

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2025,7 +2025,7 @@ const CLEANUPS = {
       new RegExp("^(https?://)?(www\\.)?flattr\\.com/profile/[^/?#]", "i"),
       new RegExp("^(https?://)?(www\\.)?patreon\\.com/[^/?#]", "i"),
       new RegExp("^(https?://)?(www\\.)?paypal\\.me/[^/?#]", "i"),
-      new RegExp("^(https?://)?(www\\.)?tipeee\\.com/[^/?#]", "i"),
+      new RegExp("^(https?://)?(?:[^/]+\\.)?tipeee\\.com/[^/?#]", "i"),
       new RegExp("^(https?://)?(www\\.)?d\\.rip/[^/?#]", "i"),
       new RegExp("^(https?://)?(www\\.)?drip\\.kickstarter.com/[^/?#]", "i"),
     ],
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2025,7 +2025,7 @@ const CLEANUPS = {
       new RegExp("^(https?://)?(www\\.)?flattr\\.com/profile/[^/?#]", "i"),
       new RegExp("^(https?://)?(www\\.)?patreon\\.com/[^/?#]", "i"),
       new RegExp("^(https?://)?(www\\.)?paypal\\.me/[^/?#]", "i"),
-      new RegExp("^(https?://)?(www\\.)?tipeee\\.com/[^/?#]", "i"),
+      new RegExp("^(https?://)?(www\.)?(?:(?:fr|en|de|es|it)\.)?tipeee\.com/[^/?#]", "i"),
```
yvanzo: It is probably not needed to match `www.en.tipeee` but it's still better than the previous version of this RegExp.

Ge0rg3: Okay, would it be better as:
[...] https?://?`((?:www|fr|en|de|es|it)\.)`?tipeee\.com [...]

yvanzo: Yes, but if you want it to keep working for any additional language in the future, just use `(?:[^/]+\\.)?tipeee\\.com` instead, that will match any subdomain.

Note that dot should be escaped twice `\\.` here.

Ge0rg3: That sounds best. I'll commit that this evening :)

yvanzo:
I altered https://github.com/metabrainz/musicbrainz-server/commit/1c3e68855b0c4bd1b5c0662eafac862c3b722f9a commit message so as to:
 * reliably refer to the ticket key rather than to refer to the obsolete commit id;
 * follow formatting standards (first line by 50 characters, other lines wrapped to 68 characters).

yvanzo:
Thank you for this patch!

PR #758, opened 2018-11-28
MBS-9909: Convert reports to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9909

Chose to have different templates for annotation and relationship based reports, rather than trying to replicate the TT macros to insert new columns, because anyway the Flow types are different and I feel it makes it much easier to make sense of the whole thing.


mwiencek:
Closing, since @reosarevok is splitting this into separate PRs.

PR #759, opened 2018-11-28
MBS-9930: Disable CAA images on the homepage for Chrome-based browsers

yvanzo:
## [MBS-9930](https://tickets.metabrainz.org/browse/MBS-9930)

Replace each cover art with a message explaining that issue [CAA-116](https://tickets.metabrainz.org/browse/CAA-116) is pending, for every browser with "Chrome" in the user agent string.

See the ticket for motivations.

### DIFF
--- a/root/main/index.js
+++ b/root/main/index.js
@@ -7,6 +7,7 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import {omit} from 'lodash';
 import he from 'he';
 import React from 'react';
 
@@ -20,11 +21,13 @@ import entityHref from '../static/scripts/common/utility/entityHref';
 type Props = {|
   +blogEntries: $ReadOnlyArray<BlogEntryT> | null,
   +newestReleases: $ReadOnlyArray<ArtworkT>,
+  +withSafeBrowsing: boolean,
 |};
 
 const Homepage = ({
   blogEntries,
   newestReleases,
+  withSafeBrowsing,
 }: Props) => (
   <Layout fullWidth homepage title={l('MusicBrainz - The Open Music Encyclopedia')}>
     <div id="maincontent">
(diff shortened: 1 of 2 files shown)


root/main/index.js
```diff
@@ -192,10 +204,19 @@ const ReleaseArtwork = ({artwork}: {|+artwork: ArtworkT|}) => {
             entity: release.name,
           })}
         >
-          <ArtworkImage
-            artwork={artwork}
-            fallback={release.cover_art_url || ''}
-          />
+          {withSafeBrowsing ? (
+            <span className="cover-art-image">
+              {l('Image temporarily disabled until {link|issue CAA-116} gets solved', {
```
reosarevok: I'd suggest replacing the entire artwork div (instead than each specific artwork box) with a text such as "Recent artwork has been disabled for Chrome users due to overzealous Google Safe Browsing phishing alerts. Our cover art is stored in Internet Archive servers, and while those servers might also have been used to host dubious content (leading to a server-wide Safe Browsing alert), the images themselves are perfectly safe. Images have only been disabled in the front page; if you get a "Deceptive site ahead" alert elsewhere while browsing MusicBrainz, you can ignore it by clicking "Details" and then "visit this unsafe site". You can follow {link|issue CAA-116} to be informed about any changes to this situation."

yvanzo: I still like the idea to show spots missing images. Could your longer text be put in a dedicated banner (at the top of the page) that shows up with Chrome-based browsers only?

reosarevok: What do we win by showing the spots? If you want, we can do both, but I really don't see the point tbh :/

mwiencek: I have a third take. :) I'd rather just show some clickable img alt text for Chrome users, like "Release by Artist," as I don't see a reason to remove or cripple the feature entirely. I realize the main purpose is to showcase some nice artwork on the front page, but it's still somewhat interesting as a live feed of sorts showing activity on the site. And we don't know how long we'll need it like this, so the less obnoxious the better.

Then I'd put reosarevok's explanation above it, or copy it to a doc page and link it as "Why aren't images showing here?"

reosarevok: As long as the explanation is clearly visible, so that when they find a red screen of bullcrap elsewhere on the site they can maybe understand what's going on, I like that idea.

root/main/index.js
```diff
@@ -167,34 +170,60 @@ const Homepage = ({
 
     <div className="feature-column" style={{clear: 'both', paddingTop: '1%'}}>
       <h2>{l('Recent Additions')}</h2>
+      {withSafeBrowsing ? (
+        <div className="banner">
+          {l('Cover art on the homepage is disabled for Chrome-based browsers which are known to issue incorrect phishing warning since the 11th of November 2018, see {ticket_link|ticket CAA-116} for follow-up.<br/>Phishing warning may show up on any other page with cover art. You are encouraged to {report_link|report errors to Google Safe Browsing}. Sorry for the inconvenience!', {
```
mwiencek: A few small grammar tweaks:
```suggestion
          {l('Cover art on the homepage is disabled for Chrome-based browsers, which are known to issue an incorrect phishing warning since the 11th of November 2018; see {ticket_link|ticket CAA-116} for follow-up.<br/>A phishing warning may show up on any other page with cover art. You are encouraged to {report_link|report errors to Google Safe Browsing}. Sorry for the inconvenience!', {
```

yvanzo:
Fixed, tested, squashed and rebased on master.

PR #760, opened 2018-11-29
MBS-3685: Report for releases where label name = artist name

Ge0rg3:
## [MBS-3685](https://tickets.metabrainz.org/browse/MBS-3685): Report: releases with label name = artist name
This commit adds support for a new daily report, which shows all releases which have label names that are identical to their artist names.

If "ReleaseLabelSameArtist" is too ambiguous, please don't hesitate to suggest an alternative name.

https://tickets.metabrainz.org/browse/MBS-3685

### DIFF
--- a/root/report/index.tt
+++ b/root/report/index.tt
@@ -109,6 +109,7 @@
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVACredit') %]">[% l('Releases not credited to "Various Artists" but linked to VA') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVALink') %]">[% l('Releases credited to "Various Artists" but not linked to VA') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesMissingDiscIDs') %]">[% l('Releases missing disc IDs') %]</a></li>
+            <li><a href="[% c.uri_for_action('/report/show', 'ReleaseLabelSameArtist') %]">[% l('Releases with label the same as the artist') %]</a></li>
         </ul>
 
         <h2>[% l('Recordings') %]</h2>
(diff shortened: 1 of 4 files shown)


root/report/index.tt
```diff
@@ -109,6 +109,7 @@
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVACredit') %]">[% l('Releases not credited to "Various Artists" but linked to VA') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesWithoutVALink') %]">[% l('Releases credited to "Various Artists" but not linked to VA') %]</a></li>
             <li><a href="[% c.uri_for_action('/report/show', 'ReleasesMissingDiscIDs') %]">[% l('Releases missing disc IDs') %]</a></li>
+            <li><a href="[% c.uri_for_action('/report/show', 'ReleaseLabelSameArtist') %]">[% l('Releases with label the same as the artist') %]</a></li>
```
reosarevok: "Releases where artist name and label name are the same" maybe?

root/report/release_label_same_artist.tt
```diff
@@ -0,0 +1,15 @@
+[%- WRAPPER 'layout.tt' title=l('Releases with label the same as the artist') full_width=1 -%]
+
+<h1>[% l('Releases with label the same as the artist') %]</h1>
```
reosarevok: Same as with the previous file, "Releases where artist name and label name are the same" might be more clear.

root/report/release_label_same_artist.tt
```diff
@@ -0,0 +1,15 @@
...
+
+<ul>
+    <li>[% l('This report lists releases that have their label set to be the same
+              as their artist.') %]</li>
```
reosarevok: This is not exactly true - a label can't be the same as an artist. It's the names that are the same. So I would rephrase this as something like 'This report lists releases where the label name is the same as the artist name. Often this means the release is self-released, and the label {SpecialPurposeLabel|should be "[no label]" instead}.', {SpecialPurposeLabel: '/doc/Style/Unknown_and_untitled/Special_purpose_label'}

It's important that on seeing the report a user will understand why this report exists and what they are supposed to do with it :) Otherwise they'll think "ok, this report is here, but what's the point?".

yvanzo:
Thank you for this contribution. Reported releases make this report worth it already. It can be refined later on.

PR #762, opened 2018-12-03
MBS-9907: Report for places without coordinates

Ge0rg3:
## [MBS-9907](https://tickets.metabrainz.org/browse/MBS-9907): New report: Places without coordinates
This commit adds support for a new daily report, which shows all the places which haven't been assigned coordinates.

https://tickets.metabrainz.org/browse/MBS-9907

I originally did this with an "annotation" column, but seeing as the number of entries with annotations is relatively low, it seemed more of an inconvenience.

The final result looks as follows:

image

### DIFF
--- a/lib/MusicBrainz/Server/Report/PlaceReport.pm
+++ b/lib/MusicBrainz/Server/Report/PlaceReport.pm
@@ -13,6 +13,8 @@ around inflate_rows => sub {
         map { $_->{place_id} } @$items
     );
 
+    $self->c->model('Area')->load(values %$places);
+
     return [
         map +{
             %$_,
--- a/lib/MusicBrainz/Server/Report/PlacesWithoutCoordinates.pm
+++ b/lib/MusicBrainz/Server/Report/PlacesWithoutCoordinates.pm
@@ -1,0 +1,28 @@
+package MusicBrainz::Server::Report::PlacesWithoutCoordinates;
+use Moose;
+
+with 'MusicBrainz::Server::Report::PlaceReport';
+
+sub query {
+    q{
+      SELECT DISTINCT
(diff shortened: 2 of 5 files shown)


Ge0rg3:
@reosarevok changes made :)

Ge0rg3:
The [new commit](https://github.com/metabrainz/musicbrainz-server/pull/762/commits/72ab4a16347d47b624cad5e4fd2a2b72bac18a4e) introduces a search button, which will take the user to a Google maps search page with all of the data in the row filled in automatically:

image

yvanzo:
I’m not sure linking to Google Maps is appropriate here as label’s map is displayed from OpenStreetMap. @reosarevok?

Ge0rg3:
Should I remove this for now? @reosarevok @yvanzo

Freso:
Maybe make a "Search" button/link for OpenStreetMap instead? That way it'd still have the convenience of a one-click help to find the coordinates, but we'd also be supporting another open data project (whose data we're also consuming/using in the form of their maps). But yeah, having more text than just "Search" would probably be good. "Search OpenStreetMap" is maybe too long? It does seem like there's a lot of empty space in the columns though.

Ge0rg3:
@Freso Cool, how about having "Search OpenStreetMap" as the column header, and then just "Search" on the buttons?

Ge0rg3:
@yvanzo Okay, I've now committed that.

yvanzo:
Nit-picking: you did not update latest commit message accordingly.

Ge0rg3:
> Nit-picking: you did not update latest commit message accordingly.

Good spot, fixed.

yvanzo:
Thank you for this additional contribution!

PR #763, opened 2018-12-03
Add JS xgettext replacement

mwiencek:
The new utility handles both multiline and concatenated strings:

```JavaScript
// These all extract to the same msgid

l('Hello World');

l('Hello  World');

l('Hello \
   World');

l('Hello ' +
  'World');
```

I updated the .pot files in 03d157d (using the old xgettext extraction), then again in e0e4e79 (with the new utility), so that you can see the changes. I'm not sure if the header changes matter (or why they occur). It does appear that the new utility found some strings that the GNU xgettext failed to parse.


yvanzo:
Note for later, would be nice to handle ES6 template string too.

mwiencek:
> Note for later, would be nice to handle ES6 template string too.

Just for the multiline support, right? I'm not sure how we would guard against `${}` existing in translations (where any JS expression could be embedded and executed). Not in the `expand` implementation, since that wouldn't receive the raw template string, it'd be executed first.

mwiencek:
> Note for later, would be nice to handle ES6 template string too.

Done so in https://github.com/metabrainz/musicbrainz-server/pull/778 (ignore my previous comment which didn't make sense).

PR #765, opened 2018-12-04
Improve Label Report from #760

Ge0rg3:
This continues on the work from #760.

In [MBS-3685](https://tickets.metabrainz.org/browse/MBS-3685), a new report page was created for artists which have the same label name as artist name. However, this did not include featured artists. This PR compares the artist names from **artist_credit_name** instead of just **artist_credit**,  which means that all featured artists will also be checked.
For example, a release like...
https://musicbrainz.org/release/14bfe5a4-ae36-4642-9ed3-fdadfe83f2e5 
... is now identified.

Before, there were 2964 entities found in the report, but now there are 7026.

### DIFF
--- a/lib/MusicBrainz/Server/Report/ReleaseLabelSameArtist.pm
+++ b/lib/MusicBrainz/Server/Report/ReleaseLabelSameArtist.pm
@@ -8,6 +8,8 @@ sub query {
     q{
       SELECT DISTINCT
         r.id AS release_id,
+        l.name AS label_name,
+        l.gid AS label_gid,
         row_number() OVER (ORDER BY r.artist_credit, r.name)
       FROM
         release AS r
--- a/lib/MusicBrainz/Server/Report/ReleaseLabelSameArtist.pm
+++ b/lib/MusicBrainz/Server/Report/ReleaseLabelSameArtist.pm
@@ -7,15 +7,18 @@ with 'MusicBrainz::Server::Report::ReleaseReport',
 sub query {
     q{
       SELECT DISTINCT
-        release.id AS release_id,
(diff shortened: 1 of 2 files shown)


Freso:
Maybe that report could also show the Label name (and link to the Label)?

Ge0rg3:
There's no entry in common-macros.tt for doing links on labels (without the entity), and I guess that loading the entire Label entity into ReleaseReport.pm seems extra just for the one report. So I'll just manually form the URL inside the release_label_same_artist.tt file, unless anyone wants it done differently?

Ge0rg3:
New column implemented:
image

lib/MusicBrainz/Server/Report/ReleaseLabelSameArtist.pm
```diff
@@ -7,15 +7,17 @@ with 'MusicBrainz::Server::Report::ReleaseReport',
...
-        INNER JOIN artist ON release.artist_credit=artist.id
-        INNER JOIN release_label ON release_label.release=release.id
-        INNER JOIN label ON release_label.label=label.id
+        release AS r
+        INNER JOIN artist_credit_name AS acn ON acn.artist_credit = r.artist_credit
+        INNER JOIN release_label AS rl ON rl.release=r.id
+        INNER JOIN label AS l ON rl.label=l.id
       WHERE
-        label.name=artist.name
+        l.name = acn.name
```
yvanzo: Why not comparing both names rather than replacing one by the other? Those are not necessarily the same string. IMHO, any match is worth being reported.

Ge0rg3: Doing this resulted in another 15 entities being found 👍

PR #766, opened 2018-12-05
MBS-9941: Load more gettext domains on edit pages

mwiencek:
This fixes the inline search for works in the relationship editor, and the area selection bubbles, when using a non-English translation.

The attributes domain isn't currently needed, AFAICT, but it will be once we fix all the untranslated strings in Autocomplete.js.


mwiencek:
@yvanzo This should be fixed now, with a couple more changes cherry-picked from your PR #642.

PR #767, opened 2018-12-05
MBS-8178: Create report for sequential events not in series

Ge0rg3:
## [MBS-9178](https://tickets.metabrainz.org/browse/MBS-8178): Report for Events with sequential numbering that are not part of a series

This PR creates a new report page (EventSequenceNotInSeries) which uses regex to identify events which may be part of an actual series, but aren't in a series on MusicBrainz.

The regex used was taken from the ticket (and slightly adjusted).

The report looks as follows:
image


Ge0rg3:
^Btw, that force push was to replace the text in the report page, but since the change was small I just rebased the first commit instead of creating a second.

Ge0rg3:
@yvanzo changed description 👍

root/report/event_sequence_not_in_series.tt
```diff
@@ -0,0 +1,14 @@
+[%- WRAPPER 'layout.tt' title=l('Events which should be part of series or larger event') full_width=1 -%]
+
+<h1>[% l('Events which should be part of series or larger event') %]</h1>
+
+<ul>
+    <li>[% l('This report lists events where the event name indicates that it may have to be part of a series or a larger event.') %]</li>
+    <li>[% l('Total releases found: {count}', { count => pager.total_entries }) %]</li>
```
yvanzo: Those are "events", not "releases".

Ge0rg3: That's what I get for copy-pasting...

PR #768, opened 2018-12-06
MBS-9402: fix multilang display on work merge/sidebar

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9402

Work search was fixed while moving it to React, but work merge (which is still TT) was still broken, and the work merge edit wasn't loading works in the first place. 

Additionally fixed the sidebar to show "English, Japanese" rather than "EnglishJapanese".

### DIFF
--- a/root/work/merge.tt
+++ b/root/work/merge.tt
@@ -52,7 +52,13 @@
                       </ul>
                     </td>
                     <td>[% entity.l_type_name %]</td>
-                    <td><abbr title="[% entity.language.l_name %]">[% entity.language.iso_code_3 %]</abbr></td>
+                    <td>
+                      <ul>
+                        [% FOR language=entity.languages %]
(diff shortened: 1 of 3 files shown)


root/work/merge.tt
```diff
@@ -52,7 +52,13 @@
                       </ul>
                     </td>
                     <td>[% entity.l_type_name %]</td>
-                    <td><abbr title="[% entity.language.l_name %]">[% entity.language.iso_code_3 %]</abbr></td>
+                    <td>
+                      <ul>
+                        [% FOR language=entity.languages %]
+                        <li><abbr title="[% language.language.l_name %]">[% language.language.iso_code_3 %]</abbr></li>
```
mwiencek: Just need to escape the title there:
```suggestion
                        <li><abbr title="[% language.language.l_name | html %]">[% language.language.iso_code_3 %]</abbr></li>
```

PR #770, opened 2018-12-08
MBS-9904: Convert inline list of users from TT to React

TheFaR8:
https://tickets.metabrainz.org/browse/MBS-9904

Only a part of them (not fully converted yet)


root/user/privileged.tt
```diff
@@ -1,15 +1,4 @@
...
-    [%- FOR editor=editors -%]
-      [%- link_entity(editor) -%]
-      [%- IF loop.count < editors.size -%], [%- END -%]
-    [%- END -%]
-  [% ELSE %]
-    [% l('No users found') %]
-  [% END %]
-</p>
-[% END %]
+[%- React.embed(c, 'components/UserInlineList', {editors => editors}) -%]
```
yvanzo: The path must be relative from `root/`, see point 2. ii. at [Porting TT to React](https://github.com/metabrainz/musicbrainz-server/blob/master/HACKING.md#porting-tt-to-react), that is `user/components/UserInlineList` here.

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import * as React from 'react';
+
+import {l} from '../../static/scripts/common/i18n';
+import EntityLink from '../../static/scripts/common/components/EntityLink';
+import commaOnlyList from '../../static/scripts/common/i18n/commaOnlyList';
+
+type Props = {|
+  +editors: EditorT,
```
yvanzo: `editors` is an array which is accessed through the `map` function on line 24.

Should be typed as:
```javascript
  +editors: $ReadOnlyArray<EditorT>,
```

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
...
+|};
+
+const UserInlineList = ({editors}: Props) => {
+  <p>
+    {editors.size ? (
```
yvanzo: `size` doesn’t exist for array, you mean `length`

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
...
+      {commaOnList(
```
yvanzo: Typo, should be `commaOnlyList`

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
...
+        editors.map((editor) => (
+          <EntityLink entity={editor} />
```
yvanzo: Should use `EditorLink` instead, with `editor` property

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
...
+
+import {l} from '../../static/scripts/common/i18n';
+import EntityLink from '../../static/scripts/common/components/EntityLink';
+import commaOnlyList from '../../static/scripts/common/i18n/commaOnlyList';
+
+type Props = {|
+  +editors: EditorT,
+|};
+
+const UserInlineList = ({editors}: Props) => {
```
yvanzo: Opening bracket is not appropriate here as it should be used to enclose JavaScript expression whereas `<p>` is a JSX element. You can replace it with opening parenthesis if you like. (And update closing bracket on line 32 accordingly.)

root/user/components/UserInlineList.js
yvanzo: Opening bracket is not appropriate here either, as you appropriately opened a bracket for JavaScript expressions as children on line 22 already. Just remove it (from line 23). (And remove closing bracket on line 28 accordingly.)

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
...
+  <p>
+    {editors.size ? (
+      {commaOnList(
+        editors.map((editor) => (
+          <EntityLink entity={editor} />
+        )),
```
yvanzo: There is an extra closing parenthesis, as the following line should be the second parameter to call the `map` function. Just move the extra parenthesis to line 28.

root/user/components/UserInlineList.js
```diff
@@ -0,0 +1,32 @@
+/*
+ * @flow
+ * Copyright (C) 2018 MetaBrainz Foundation
```
yvanzo: You might want to add your own copyright notice here, or replace the current one, or not change anything. Note that this notice will be public and indexed by search engines.

root/user/components/UserInlineList.js
TheFaR8: Change `{` with `(` , right?

root/user/components/UserInlineList.js
yvanzo: Sorry, I was wrong here, it was correct. The first `)` closed the mapped anonymous function while the second `)` closes the `map` function.

TheFaR8:
Thank you too for the suggested revisions :)

PR #772, opened 2018-12-10
MBS-9340: Improve lyrics language selector

Ge0rg3:
## [MBS-9340](https://tickets.metabrainz.org/browse/MBS-9340): Don't allow more languages if [No lyrics] is selected
At the moment, the user can select both "No lyrics", and other languages when editing works' lyrics. This PR aims to prevent this.

### DIFF
--- a/root/components/FormRowSelectList.js
+++ b/root/components/FormRowSelectList.js
@@ -19,6 +19,7 @@ type Props<S> = {|
   +addId: string,
   +addLabel: string,
   +getSelectField: (S) => FieldT<number | string>,
+  +hideAddButton?: boolean,
   +label: string,
   +onAdd: (event: SyntheticEvent<HTMLButtonElement>) => void,
   +onEdit: (index: number, value: string) => void,
@@ -33,6 +34,7 @@ const FormRowSelectList = <F, S: AnyFieldT<F>>({
   addId,
   addLabel,
   getSelectField,
+  hideAddButton,
   label,
   onAdd,
   onEdit,
@@ -62,16 +64,18 @@ const FormRowSelectList = <F, S: AnyFieldT<F>>({
           <FieldErrors field={getSelectField(subfield)} />
         </div>
       ))}
-      <div className="form-row-add">
-        <button
-          className="with-label add-item"
-          id={addId}
-          onClick={onAdd}
-          type="button"
-        >
-          {addLabel}
-        </button>
-      </div>
+      {hideAddButton ? null : (
+        <div className="form-row-add">
+          <button
+            className="with-label add-item"
+            id={addId}
+            onClick={onAdd}
+            type="button"
+          >
+            {addLabel}
+          </button>
+        </div>
+      )}
     </div>
   </FormRow>
 );
--- a/root/static/scripts/work.js
+++ b/root/static/scripts/work.js
@@ -286,6 +286,7 @@ function renderWorkLanguages() {
       addId="add-language"
       addLabel={l('Add Language')}
       getSelectField={_.identity}
+      hideAddButton={_.intersection(form.field.languages.field.map(lang => String(lang.value)), ["486", "284"]).length > 0}
       label={l('Lyrics Languages')}
       onAdd={addLanguage}
       onEdit={editLanguage}
--- a/t/selenium/Work_Editor.html
+++ b/t/selenium/Work_Editor.html
@@ -44,6 +44,16 @@
 	<td>label=[No lyrics]</td>
 </tr>
 <tr>
+	<td>assertEval</td>
+	<td>!!window.document.getElementById('add-language')</td>
+	<td>false</td>
+</tr>
+<tr>
+	<td>select</td>
+	<td>name=edit-work.languages.0</td>
+	<td>label=French</td>
+</tr>
+<tr>
 	<td>click</td>
 	<td>css=#add-language</td>
 	<td></td>


root/static/scripts/work.js
```diff
@@ -70,6 +70,11 @@ const store = createStore(function (state: WorkForm = form, action) {
         action.languageId,
         state,
       );
+      if (_.intersection(state.field.languages.field.map(lang => lang.value), ["486", "284"]).length > 0) {
+        state.field.languages.errors = 1;
+      } else {
+        state.field.languages.errors = 0;
+      }
```
yvanzo: It is a bit weird to set a number to `errors` which it is supposed to hold a string array. I suggest to move the condition (on line 73) into a function to be passed to `FormRowSelectList` as a property (see the latest suggested type in previous comment).

Ge0rg3:
@yvanzo Wrt putting this into a function, should I add the event call into:
```
            onChange={event => onEdit(index, event.currentTarget.value)}
```

root/static/scripts/work.js
```diff
@@ -70,6 +70,7 @@ const store = createStore(function (state: WorkForm = form, action) {
         action.languageId,
         state,
       );
+      state.field.languages.errors = (_.intersection(state.field.languages.field.map(lang => lang.value), ["486", "284"]).length > 0);
```
mwiencek: We try to avoid mutating `state` here. I like the idea of the callback function or just passing the `length > 0` check as a boolean.

Note that the type of `lang.value` is `number`, so either `["486", "284"]` doesn't work or they're actually strings and we're not storing it correctly. You can do `String(lang.value)` in the map callback to be sure, though.

Ge0rg3: Great, made relevant changes in new commit.

root/static/scripts/work.js
```diff
@@ -295,7 +294,7 @@ function renderWorkLanguages() {
       removeClassName="remove-language"
       removeLabel={l('Remove Language')}
       repeatable={form.field.languages}
-      noLyrics={form.field.languages.errors}
+      noLyrics={_.intersection(form.field.languages.field.map(lang => String(lang.value)), ["486", "284"]).length > 0}
```
yvanzo: * I guess `String(lang.value)` was just a test _to be sure_. Since it is known to be `number`, it would be simpler to just compare it with number.
* This function call returns a boolean, but you changed the type of `noLyrics`.
* `noLyrics` should rather be named something general-purpose like `hideAddButton`.

Ge0rg3: Okay, will make these changes.
However, it does seem like lang.value is stored as a string.
When I originally load the page, the value is stored as an int when viewed in the React chrome extension, but after editing the value, it's turned into a string. So it seems like it needs to be converted at some stage?

yvanzo: Oh right, sorry, keep it as string then.

Ge0rg3:
@yvanzo Changes made 😄

Ge0rg3:
Looks like I'll also have to update the test case.

root/components/FormRowSelectList.js
```diff
@@ -72,6 +75,7 @@ const FormRowSelectList = <F, S: AnyFieldT<F>>({
           {addLabel}
         </button>
       </div>
+      }
```
mwiencek: Just a small eslint issue:

```JSX
// Change this
{foo ? null :
<div />
}

// To this
{foo ? null : (
  <div /> // indent +2 spaces
)}
```

Ge0rg3:
@mwiencek I've updated the style in latest commit, and @yvanzo I've updated the test case so that it now checks that "Add Language" does not appear (the failing test is unrelated to these commits) 😄

root/components/FormRowSelectList.js
Ge0rg3: Done1

root/components/FormRowSelectList.js
```diff
@@ -19,6 +19,7 @@ type Props<S> = {|
   +addId: string,
   +addLabel: string,
   +getSelectField: (S) => FieldT<number | string>,
+  +hideAddButton: boolean,
```
yvanzo: Hiding add button should be optional.

```suggestion
  +hideAddButton?: boolean,
```

t/selenium/Work_Editor.html
```diff
@@ -43,6 +43,16 @@
 	<td>name=edit-work.languages.0</td>
 	<td>label=[No lyrics]</td>
 </tr>
+<tr>
+	<td>assertEval</td>
+	<td>!!window.document.getElementById('add-language')</td>
```
yvanzo: This is the only unusual addition to Selenium tests. 
I would prefer using `assertElementNotPresent` instead.
This command has to be added to `t/selenium.js` though.

root/components/FormRowSelectList.js
mwiencek: Fixed.

t/selenium/Work_Editor.html
mwiencek: I tried adding this (implemented in the same way `assertElementPresent` is), but the web driver kept returning an element despite it not existing on the page, and I couldn't figure out why. This method works fine, at least.

mwiencek:
@brainzbot test this please

mwiencek:
This has a small flaw in that you can work around it by just adding multiple empty languages before selecting any (and it needs to be validated in the Form server-side, too), but is a good initial improvement.

PR #774, opened 2018-12-12
MBS-8915: Allow editors to choose delimiter in track parser

Ge0rg3:
### [MBS-8915](https://tickets.metabrainz.org/browse/MBS-8915)
This PR allows users to choose a custom delimiter, which can also be text like "\t". This also required a change to the CSS to fit the extra option, and a change to the optionCookie function to allow for strings as well as just boolean values.

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -77,6 +77,8 @@
         <label data-bind="visible: useCustomDelimiter">
           [%= l('Custom delimiter') %]
           <input type="text" name="custom-delimiter" data-bind="value: customDelimiter" />
+          (<a href="#" data-bind="click: toggleDelimiterHelp">[% l('help') %]</a>)
+          <div data-bind="visible: delimiterHelpVisible"><i>Enter the delimiter that separates the track name from the artist name (i.e. "," or "\t").</i></div>
         </label>
       </td>
     <tr>
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -24,6 +24,12 @@
           [%= l('Use track numbers') %]
         </label>
       </td>
+      <td>
+        <label>
+          [%= l('Custom time delimiter') %]
+          <input type="text" name="time-delimiter" data-bind="value: timeDelimiter" />
+        </label>
+      </td>
     </tr>
     <tr>
       <td style="padding-left: 2em">
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -24,12 +24,6 @@
           [%= l('Use track numbers') %]
         </label>
       </td>
-      <td>
-        <label>
-          [%= l('Custom time delimiter') %]
-          <input type="text" name="time-delimiter" data-bind="value: timeDelimiter" />
-        </label>
-      </td>
     </tr>
     <tr>
       <td style="padding-left: 2em">
@@ -72,6 +66,20 @@
         </label>
       </td>
     </tr>
+    <tr>
+    <td>
+      <label>
+      <input type="checkbox" name="use-custom-delimiter" data-bind="checked: useCustomDelimiter" />
+        [%= l('Use custom artist delimeter') %]
+      </label>
+    </td>
+      <td>
+        <label data-bind="visible: useCustomDelimiter">
+          [%= l('Custom delimiter') %]
+          <input type="text" name="custom-delimiter" data-bind="value: customDelimiter" />
+        </label>
+      </td>
+    <tr>
   </table>
 </script>
 
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -78,7 +78,7 @@
           [%= l('Custom delimiter') %]
           <input type="text" name="custom-delimiter" data-bind="value: customDelimiter" />
           (<a href="#" data-bind="click: toggleDelimiterHelp">[% l('help') %]</a>)
-          <div data-bind="visible: delimiterHelpVisible"><i>Enter the delimiter that separates the track name from the artist name (i.e. "," or "\t").</i></div>
+          <div data-bind="visible: delimiterHelpVisible"><i>{l('Enter the delimiter that separates the track name from the artist name (i.e. "," or "\t").')}</i></div>
         </label>
       </td>
     <tr>
--- a/root/static/scripts/release-editor/trackParser.js
+++ b/root/static/scripts/release-editor/trackParser.js
@@ -39,7 +39,8 @@ releaseEditor.trackParser = {
         hasTrackNumbers: optionCookie("trackparser_tracknumbers", true),
         hasTrackArtists: optionCookie("trackparser_trackartists", true),
         hasVinylNumbers: optionCookie("trackparser_vinylnumbers", false),
-        timeDelimiter: optionCookie("trackparser_timeDelimiter", "", false),
+        customDelimiter: optionCookie("trackparser_customdelimiter", "", false),
+        useCustomDelimiter: optionCookie("trackparser_usecustomdelimiter", false),
         useTrackNumbers: optionCookie("trackparser_usetracknumbers", true),
         useTrackNames: optionCookie("trackparser_usetracknames", true),
         useTrackArtists: optionCookie("trackparser_usetrackartists", true),
@@ -286,12 +287,6 @@ releaseEditor.trackParser = {
         // numbers if the line only contains a time.
 
         // Assume the track time is at the end.
-        if (options.timeDelimiter != "") {
-          //Escape most regex characters
-          options.timeDelimiter = options.timeDelimiter.replace(/[.*?+^$(){}|[\]]/g, '\\$&');
-          this.trackTime = new RegExp("\\(?((?:[0-9０-９]+"+options.timeDelimiter+")?[0-9０-９\\?]+"+options.timeDelimiter+"[0-5０-５\\?][0-9０-９\\?])\\)?$");
-        }
-
         var match = line.match(this.trackTime);
 
         if (match !== null) {
@@ -328,6 +323,13 @@ releaseEditor.trackParser = {
                 data.name = clean(line);
             }
             return data;
+        }
+
+        //Use custom delimeter as separator.
+        if (options.useCustomDelimiter && options.customDelimiter != "") {
+          //Escape most regex characters.
+          options.customDelimiter = options.customDelimiter.replace(/[.*?+^$(){}|[\]]/g, '\\$&');
+          this.separators = new RegExp("("+options.customDelimiter+")");
         }
 
         // Split the string into parts, if there are any.
--- a/root/static/scripts/release-editor/trackParser.js
+++ b/root/static/scripts/release-editor/trackParser.js
@@ -44,7 +44,11 @@ releaseEditor.trackParser = {
         useTrackNumbers: optionCookie("trackparser_usetracknumbers", true),
         useTrackNames: optionCookie("trackparser_usetracknames", true),
         useTrackArtists: optionCookie("trackparser_usetrackartists", true),
-        useTrackLengths: optionCookie("trackparser_tracktimes", true)
+        useTrackLengths: optionCookie("trackparser_tracktimes", true),
+        delimiterHelpVisible: optionCookie("trackparser_delimiterhelpvisible", false),
+        toggleDelimiterHelp: function() {
+          this.delimiterHelpVisible(!this.delimiterHelpVisible());
+        }
     },
 
     parse: function (str, medium) {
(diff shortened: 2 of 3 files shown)


root/static/scripts/release-editor/trackParser.js
```diff
@@ -33,12 +33,13 @@ releaseEditor.trackParser = {
     trackNumber: /^(?:M[\.\-])?([０-９0-9]+(?:-[０-９0-9]+)?)(?:\.|．|\s?-|:|：|;|,|，|$)?/,
     vinylNumber: /^([０-９0-9a-z]+)(?:\/[０-９0-9a-z]+)?(?:\.|．|\s?-|:|：|;|,|，|$)?/i,
 
-    trackTime: /\(?((?:[0-9０-９]+[：，．':,.])?[0-9０-９\?]+[：，．':,.][0-5０-５\?][0-9０-９\?])\)?$/,
+    trackTime: new RegExp("\\(?((?:[0-9０-９]+[：，．':,.])?[0-9０-９\\?]+[：，．'\\+:,.][0-5０-５\\?][0-9０-９\\?])\\)?$"),
```
yvanzo: It is not about track time, it is about superseding `separators` (defined on [line 30](https://github.com/metabrainz/musicbrainz-server/pull/774/files#diff-95bd94154407c3fafde7a3dd2dfcd828R30)) with a custom delimiter.

root/release/edit/tracklist.tt
```diff
@@ -24,6 +24,12 @@
           [%= l('Use track numbers') %]
         </label>
       </td>
+      <td>
+        <label>
+          [%= l('Custom time delimiter') %]
+          <input type="text" name="time-delimiter" data-bind="value: timeDelimiter" />
```
yvanzo: It should be optional, that is, probably hidden behind an unchecked checkbox labelled with _Use custom delimiter_.

Ge0rg3:
@yvanzo Not too sure why I assumed the delimiter was for the time... Anyway, that's fixed now. I've chosen to hide the input box if the "Use custom artist delimiter" option isn't selected, instead of disabling it since it didn't seem clear that the box was disabled.
The form now looks like this:
image

root/release/edit/tracklist.tt
```diff
@@ -66,6 +66,22 @@
...
+      <input type="checkbox" name="use-custom-delimiter" data-bind="checked: useCustomDelimiter" />
+        [%= l('Use custom artist delimeter') %]
+      </label>
+    </td>
+      <td>
+        <label data-bind="visible: useCustomDelimiter">
+          [%= l('Custom delimiter') %]
+          <input type="text" name="custom-delimiter" data-bind="value: customDelimiter" />
+          (<a href="#" data-bind="click: toggleDelimiterHelp">[% l('help') %]</a>)
+          <div data-bind="visible: delimiterHelpVisible"><i>Enter the delimiter that separates the track name from the artist name (i.e. "," or "\t").</i></div>
```
yvanzo: The help message should be localized.
```suggestion
          <div data-bind="visible: delimiterHelpVisible"><i>{l('Enter the delimiter that separates the track name from the artist name (i.e. "," or "\t").')}</i></div>
```

root/static/scripts/release-editor/trackParser.js
```diff
@@ -323,6 +329,13 @@ releaseEditor.trackParser = {
             return data;
         }
 
+        //Use custom delimeter as separator.
+        if (options.useCustomDelimiter && options.customDelimiter != "") {
+          //Escape most regex characters.
```
yvanzo: Nit-picking:
* space is missing after comment mark `//`
* indentation is inconsistent with the 4-space indent used in the rest of this file
```suggestion
            // Escape most regex characters.
```

root/release/edit/tracklist.tt
```diff
@@ -66,6 +66,22 @@
         </label>
       </td>
     </tr>
+    <tr>
+    <td>
+      <label>
+      <input type="checkbox" name="use-custom-delimiter" data-bind="checked: useCustomDelimiter" />
+        [%= l('Use custom artist delimeter') %]
```
mwiencek: s/delimeter/delimiter/

root/static/scripts/release-editor/trackParser.js
```diff
@@ -323,6 +329,13 @@ releaseEditor.trackParser = {
             return data;
         }
 
+        //Use custom delimeter as separator.
```
mwiencek: s/delimeter/delimiter/

root/release/edit/tracklist.tt
```diff
@@ -66,6 +66,22 @@
...
+      </label>
+    </td>
+      <td>
+        <label data-bind="visible: useCustomDelimiter">
+          [%= l('Custom delimiter') %]
+          <input type="text" name="custom-delimiter" data-bind="value: customDelimiter" />
+          (<a href="#" data-bind="click: toggleDelimiterHelp">[% l('help') %]</a>)
+          <div data-bind="visible: delimiterHelpVisible"><i>{l('Enter the delimiter that separates the track name from the artist name (i.e. "," or "\t").')}</i></div>
```
mwiencek: JSX syntax (`{}`) in TT?

root/release/edit/tracklist.tt
```diff
@@ -66,6 +66,22 @@
...
+    <tr>
+    <td>
+      <label>
+      <input type="checkbox" name="use-custom-delimiter" data-bind="checked: useCustomDelimiter" />
+        [%= l('Use custom artist delimeter') %]
+      </label>
+    </td>
+      <td>
+        <label data-bind="visible: useCustomDelimiter">
+          [%= l('Custom delimiter') %]
```
mwiencek: `add_colon(l('Custom delimiter'))` to separate the text from the input with some punctuation would be nice.

root/static/scripts/release-editor/trackParser.js
```diff
@@ -33,16 +33,22 @@ releaseEditor.trackParser = {
     trackNumber: /^(?:M[\.\-])?([０-９0-9]+(?:-[０-９0-9]+)?)(?:\.|．|\s?-|:|：|;|,|，|$)?/,
     vinylNumber: /^([０-９0-9a-z]+)(?:\/[０-９0-9a-z]+)?(?:\.|．|\s?-|:|：|;|,|，|$)?/i,
 
-    trackTime: /\(?((?:[0-9０-９]+[：，．':,.])?[0-9０-９\?]+[：，．':,.][0-5０-５\?][0-9０-９\?])\)?$/,
+    trackTime: new RegExp("\\(?((?:[0-9０-９]+[：，．':,.])?[0-9０-９\\?]+[：，．'\\+:,.][0-5０-５\\?][0-9０-９\\?])\\)?$"),
```
mwiencek: Not sure why this was changed to `new RegExp`, since the pattern still looks to be static.

root/static/scripts/release-editor/trackParser.js
```diff
@@ -33,16 +33,22 @@ releaseEditor.trackParser = {
...
         hasTrackArtists: optionCookie("trackparser_trackartists", true),
         hasVinylNumbers: optionCookie("trackparser_vinylnumbers", false),
+        customDelimiter: optionCookie("trackparser_customdelimiter", "", false),
+        useCustomDelimiter: optionCookie("trackparser_usecustomdelimiter", false),
         useTrackNumbers: optionCookie("trackparser_usetracknumbers", true),
         useTrackNames: optionCookie("trackparser_usetracknames", true),
         useTrackArtists: optionCookie("trackparser_usetrackartists", true),
-        useTrackLengths: optionCookie("trackparser_tracktimes", true)
+        useTrackLengths: optionCookie("trackparser_tracktimes", true),
+        delimiterHelpVisible: optionCookie("trackparser_delimiterhelpvisible", false),
```
mwiencek: I don't think we need a cookie for this. I doubt anyone will need the help visible by default once they learn it once.

root/static/scripts/release-editor/trackParser.js
```diff
@@ -323,6 +329,13 @@ releaseEditor.trackParser = {
             return data;
         }
 
+        //Use custom delimeter as separator.
+        if (options.useCustomDelimiter && options.customDelimiter != "") {
+            // Escape most regex characters.
+          options.customDelimiter = options.customDelimiter.replace(/[.*?+^$(){}|[\]]/g, '\\$&');
```
mwiencek: Can you use root/static/scripts/common/utility/escapeRegExp.js here?

root/static/scripts/release-editor/trackParser.js
```diff
@@ -323,6 +329,13 @@ releaseEditor.trackParser = {
...
+          this.separators = new RegExp("("+options.customDelimiter+")");
```
mwiencek:
```suggestion
          this.separators = new RegExp("(" + options.customDelimiter + ")");
```

root/static/scripts/release-editor/trackParser.js
mwiencek: I guess not, since you don't want to escape backslashes. Did you consider allowing full RegExp syntax? Maybe there's a good reason not to, but a non-programmer probably won't know what `\t` is in either case.

root/static/scripts/release-editor/trackParser.js
Ge0rg3: I changed this when this originally was for a tracktime delimiter, but forgot to change it back D:

root/static/scripts/release-editor/trackParser.js
```diff
@@ -323,6 +329,13 @@ releaseEditor.trackParser = {
             return data;
         }
 
+        //Use custom delimeter as separator.
+        if (options.useCustomDelimiter && options.customDelimiter != "") {
```
mwiencek: `options.customDelimiter` is an observable, so this probably doesn't work as intended. You need to use `options.customDelimiter()` to unwrap the value.

There is a similar issue when setting the value below; it should be `options.customDelimiter(value)`, not `options.customDelimiter = value;`. However, this will change what the user entered in the field, which may be confusing. (Not an issue if you allow any RegExp syntax.)

root/static/scripts/release-editor/trackParser.js
mwiencek: And `options.customDelimiter()` instead of `options.customDelimiter`.

Ge0rg3:
Still fixing this up, sorry for the delay!

root/release/edit/tracklist.tt
Ge0rg3: What do you mean?

Ge0rg3:
@mwiencek:
> Glad there's a help message, though it would be even better if it explained what the syntax is or what is supported.
> 
> @reosarevok @yvanzo do you think we should allow the user to enter any RegExp, given that this is a power-user feature, or just decide on a list of backslash-char sequences like \t we want to support and document them somewhere?

@reosarevok @yvanzo Please could you take a look?

root/release/edit/tracklist.tt
mwiencek: Delimiter is spelled wrong (sorry for being oblique, s// performs a substitution in vim or perl)

Ge0rg3: Oh right 👍👍

root/release/edit/tracklist.tt
yvanzo: I don't see any reason not to allow for any RegExp here.

yvanzo:
Sorry for the delay, replied to it.

root/release/edit/tracklist.tt
reosarevok: Any regexp seems good to me too.

mwiencek: We can probably just link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp then.
```suggestion
          <div data-bind="visible: delimiterHelpVisible"><i>[% l('Enter the delimiter that separates the track name from the artist name (any valid {url|regular expression} is accepted).', { url => 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp' }) %]</i></div>
```

Ge0rg3: Great, thanks guys.

mwiencek: Or https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Writing_a_regular_expression_pattern might be friendlier

yvanzo: Or https://developer.mozilla.org/docs/Web/JavaScript/Guide/Regular_Expressions for localized version.

root/static/scripts/release-editor/trackParser.js
reosarevok: It actually seems to work as intended and not work if I use () in fact.

reosarevok:
Since this can't be edited anymore, replaced with https://github.com/metabrainz/musicbrainz-server/pull/1034

PR #777, opened 2018-12-17
MBS-9954: Convert the tag cloud and tag entity list pages to React

mwiencek:


root/tag/TagIndex.js
```diff
@@ -0,0 +1,94 @@
...
+    <React.Fragment key={entityType}>
+      <h2>{title}</h2>
+      <ul>
+        {tags.tags.map(tag => (
+          <li key={tag.entity_id}>
+            <DescriptiveLink entity={tag.entity} />
+          </li>
+        ))}
+        {tags.count > tags.tags.length ? (
+          <li>
```
yvanzo: Key attribute would probably be welcome although not absolutely needed here:
```suggestion
           <li key="count">
```

PR #783, opened 2018-12-18
 Generate test typeInfo.js on the fly

mwiencek:
This ensures the data doesn't become out of sync with what's in t/sql/initial.sql, and means we don't have to manually fix it whenever the data serialization format changes.

A call to `compile_resources.sh web-tests` has been moved from the tests' Dockerfile, to run_tests.sh (meaning it now compiles the web-tests bundle after the container is started, rather than baking it into the image). This is necessary because dump_js_type_info.pl requires database access.

This is cherry-picked from PR #737.


mwiencek:
I neglected to document that you had to run ./script/dump_js_type_info.pl, but it makes more sense to run that in compile_resources, so I added it [there](https://github.com/metabrainz/musicbrainz-server/compare/bbf5dd4d1fc15794444db30d05702883b03bb61e..3e169ee4f085d925f1f67bd0fe36505b3e345d4f). The reason I didn't before is that compile_resources runs in the Dockerfile (no database access), but we can easily check if the database exists.

PR #785, opened 2018-12-20
Unbreak plackup after running Selenium tests

mwiencek:
Running Selenium tests against a development server (with `USE_SELENIUM_HEADER` enabled) required a restart of plackup after they were done, due to the database handle still pointing to the `SELENIUM` DB. This sets/resets the handle properly during the request lifecycle (there's now no issue performing `READWRITE` requests concurrently, even).

There were related issues with the DBDefs settings not being reset, and the Redis cache being shared (which could clobber login sessions on the development server). To resolve the DBDefs issue, we save CODE references to the original DBDefs methods and reset them after dispatch. To resolve the cache issue, we introduce a `CACHE_NAMESPACE` setting and modify it to add a 'Selenium:' prefix during Selenium test requests.

### DIFF
--- a/lib/DBDefs/Default.pm
+++ b/lib/DBDefs/Default.pm
@@ -201,6 +201,13 @@ sub GOOGLE_CUSTOM_SEARCH { '' }
 # Cache Settings
 ################################################################################
 
+# A namespace prefix to be applied to all items in all caches, whether for
+# entities or user login sessions. Note that this is used by the default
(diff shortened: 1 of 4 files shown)


mwiencek:
This is kinda icky and I'd prefer we just spawn separate plack/node servers for Selenium, but I'll have to investigate if it'd cause any issues to be sharing JS with the development server.

PR #786, opened 2018-12-20
Remove last-updated-recordings endpoint

mwiencek:
This is no longer needed now that search updates near-instantaneously.

PR #787, opened 2018-12-20
Replace direct SOLR access with /ws/2 rewrite

mwiencek:
NOTE: This is not mergeable until search.mb.org/wstest/2 becomes live at search.mb.org/ws/2, but it needs some more work from Zas first. (The commit message below is written as if it's already been deployed.)

The relevant openresty config is at https://github.com/metabrainz/openresty-gateways/blob/master/files/nginx/includes/search-server_server_http_05.conf

---

We want people with MB mirror servers to continue to be able to use our search servers, as they've always been able to do. But allowing public access to our SOLR cluster is problematic from a security standpoint. We could set up a proxy [1] in front of SOLR to block bad requests, but a better option is to just use the /ws/2 compatibility layer already in place in our openresty config. This layer rewrites "old" API requests understood by our Lucene-based search server into a format expected by SOLR, and performs an internal redirect. There's nothing outdated or impractical about this API, so no reason it needs to be "old."

[1] https://github.com/dergachev/solr-security-proxy

### DIFF
--- a/lib/MusicBrainz/Server/Data/Recording.pm
+++ b/lib/MusicBrainz/Server/Data/Recording.pm
@@ -427,7 +427,7 @@ sub find_recent_by_artists
         $search_url = sprintf('http://%s/ws/2/recording/?query=title:""',
                               DBDefs->SEARCH_SERVER);
     } else {
-        $search_url = sprintf('http://%s/recording/advanced?q=recording:*&wt=mbxml',
+        $search_url = sprintf('http://%s/ws/2/recording/?query=recording:*&fmt=xml&dismax=false',
                               DBDefs->SEARCH_SERVER);
     }
 
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -732,26 +732,8 @@ sub external_search
     $query = uri_escape_utf8($query);
     $type =~ s/release_group/release-group/;
 
-    my $search_url_string;
-    if (DBDefs->SEARCH_ENGINE eq 'LUCENE') {
-        my $dismax = $adv ? 'false' : 'true';
-        $search_url_string = "http://%s/ws/2/%s/?query=%s&offset=%s&max=%s&fmt=jsonnew&dismax=$dismax&web=1";
-    } else {
-        my $endpoint = "advanced";
-        if (!$adv)
-        {
-            # Solr has a bug where the dismax end point behaves differently
-            # from edismax (advanced) when the query size is 1. This is a fix
-            # for that. See https://issues.apache.org/jira/browse/SOLR-12409
-            if (split(/[\P{Word}_]+/, $query, 2) == 1) {
-                $endpoint = "basic";
-            } else {
-                $endpoint = "select";
-            }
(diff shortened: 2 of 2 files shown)


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -732,26 +732,8 @@ sub external_search
...
-        my $dismax = $adv ? 'false' : 'true';
-        $search_url_string = "http://%s/ws/2/%s/?query=%s&offset=%s&max=%s&fmt=jsonnew&dismax=$dismax&web=1";
-    } else {
-        my $endpoint = "advanced";
-        if (!$adv)
-        {
-            # Solr has a bug where the dismax end point behaves differently
-            # from edismax (advanced) when the query size is 1. This is a fix
-            # for that. See https://issues.apache.org/jira/browse/SOLR-12409
-            if (split(/[\P{Word}_]+/, $query, 2) == 1) {
```
mwiencek: Zas is looking into moving this logic into lua, but we need to switch to a deb package or something for openresty to get PCRE/Unicode support.

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -1036,13 +1018,7 @@ sub xml_search
...
     $query = uri_escape_utf8($query);
 
-    my $search_url_string;
-    if ($version eq "1" or DBDefs->SEARCH_ENGINE eq 'LUCENE') {
-        $search_url_string = "http://%s/ws/$version/%s/?query=%s&offset=%s&max=%s&fmt=xml";
-    } else {
-        $search_url_string = "http://%s/%s/advanced?q=%s&start=%s&rows=%s&wt=mbxml";
-    }
-
+    my $search_url_string = "http://%s/ws/$version/%s/?query=%s&offset=%s&max=%s&fmt=xml&dismax=false";
```
mwiencek: Added `dismax=false` to get the /advanced endpoint internally.

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -427,7 +427,7 @@ sub find_recent_by_artists
         $search_url = sprintf('http://%s/ws/2/recording/?query=title:""',
                               DBDefs->SEARCH_SERVER);
     } else {
-        $search_url = sprintf('http://%s/recording/advanced?q=recording:*&wt=mbxml',
+        $search_url = sprintf('http://%s/ws/2/recording/?query=recording:*&fmt=xml&dismax=false',
```
mwiencek: Added `dismax=false` to get the /advanced endpoint internally.

mwiencek:
I think the best solution is to keep the logic for direct access (we'll want it in production anyway), keep `SEARCH_ENGINE` set to `LUCENE` by default for people running mirrors (this will use the compatibility API), tweak the "old" queries so that they work as intended, and possibly rename `SEARCH_ENGINE` to `SEARCH_API`.

mwiencek:
Will open a new PR that combines removing /ws/1, #786, and the /wstest/2 changes suggested here, since they touch the same code and will likely all be deployed at once.

PR #788, opened 2018-12-20
Prepare for SOLR access via search.mb.org (MBH-502)

mwiencek:
NOTE: This is not mergeable until search.mb.org/wstest/2 becomes live at search.mb.org/ws/2, but it needs some more work from Zas first.

The relevant openresty config is at https://github.com/metabrainz/openresty-gateways/blob/master/files/nginx/includes/search-server_server_http_05.conf

Edit: this change is now in production, as 2020-05-26

---

 * WS1 is removed.
 * The default `SEARCH_ENGINE` setting is changed to 'SOLR'. If 'SOLR' is in use and `SEARCH_SERVER` is set to the default public search server (search.musicbrainz.org), it uses the compatibility API described in MBH-502. This is the very same code used if `SEARCH_ENGINE` is set to 'LUCENE' (which I assume the VM still needs?). Otherwise it assumes direct SOLR access (so, production or someone setting up their own instance).
 * The endpoint /ws/js/last-updated-recordings is removed (it was only a workaround for search indexing delays).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -345,7 +345,8 @@ The server by itself doesn't rate limit any request it handles. If you're
 receiving 503s, then you're likely performing
 [search queries](https://musicbrainz.org/doc/Search_Server) without having set
 up a local instance of the
-[search server](https://github.com/metabrainz/search-server). By default,
+[search server](https://github.com/metabrainz/mb-solr) along with the
+[search index rebuilder](https://github.com/metabrainz/sir). By default,
 search queries are sent to search.musicbrainz.org and are rate limited.
 
 Once you set up your own instance, change `SEARCH_SERVER` in lib/DBDefs.pm to
--- a/lib/DBDefs.pm.sample
+++ b/lib/DBDefs.pm.sample
@@ -170,7 +170,7 @@ sub WEB_SERVER                { "www.musicbrainz.example.com" }
 # Relevant only if SSL redirects are enabled
 # sub WEB_SERVER_SSL            { "localhost" }
 # sub SEARCH_SERVER             { "search.musicbrainz.org" }
-# sub SEARCH_ENGINE             { "LUCENE" }
+# sub SEARCH_ENGINE             { "SOLR" }
 # Used, for example, to have emails sent from the beta server list the
 # main server
 # sub WEB_SERVER_USED_IN_EMAIL  { my $self = shift; $self->WEB_SERVER }
--- a/lib/DBDefs/Default.pm
+++ b/lib/DBDefs/Default.pm
@@ -103,7 +103,7 @@ sub WEB_SERVER                { "localhost:5000" }
 # Relevant only if SSL redirects are enabled
 sub WEB_SERVER_SSL            { "localhost" }
 sub SEARCH_SERVER             { "search.musicbrainz.org" }
-sub SEARCH_ENGINE             { "LUCENE" }
+sub SEARCH_ENGINE             { "SOLR" }
 # Whether to use x-accel-redirect for webservice searches,
 # using /internal/search as the internal redirect
 sub SEARCH_X_ACCEL_REDIRECT   { 0 }
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -794,7 +794,7 @@ sub external_search
     $type =~ s/release_group/release-group/;
 
     my $search_url_string;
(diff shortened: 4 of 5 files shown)


lib/MusicBrainz/Server/Controller.pm
```diff
@@ -299,4 +299,12 @@ sub error {
     $c->detach;
 }
 
+sub ws1_gone : Chained('/') PathPart('ws/1') {
+    my ($self, $c) = @_;
+
+    $c->res->content_type('text/plain; charset=utf-8');
+    $c->res->body('https://blog.metabrainz.org/2018/02/01/web-service-ver-1-0-ws-1-will-be-removed-in-6-months/');
```
mwiencek: This could probably be done on the gateways, and maybe a different response is more appropriate. (Note that /ws/1 did return errors as text/plain in general, for some reason.)

zas: A simple "/ws/1 is dead since ..." message with proper error code (503) should do. And yes, it should be done on gateways.

zas: Yup, 410 Gone is even better of course

mwiencek:
AFAICT we can merge commit https://github.com/metabrainz/musicbrainz-server/pull/788/commits/1ccd25d00d4c51894dcc9b696d2583b4e6994569 now since nothing depends on /ws/1 existing. The search changes in commit https://github.com/metabrainz/musicbrainz-server/pull/788/commits/b76be6296505342ad9a89f6dfe7aa46719d3fa53 are unrelated. @yvanzo does that make sense?

yvanzo:
LGTM. About public access to search server, what would you recommend for mirror search servers?

mwiencek:
> About public access to search server, what would you recommend for mirror search servers?

I don't have any good recommendations for people creating their own mirror, other than try to use https://github.com/metabrainz/docker-solr :)

Or do you mean how would they connect it to their MBS mirror? After this patch, they would simply change `SEARCH_SERVER`.

yvanzo:
This was about https://github.com/metabrainz/musicbrainz-server/commit/b76be6296505342ad9a89f6dfe7aa46719d3fa53 being "unrelated". Isn’t i allowing public access to SOLR "problematic from a security standpoint"?

I am working on embedding SOLR into the next MB VM which is likely to be used for mirror servers and expose port 8983 to other containers only, not to the outside world.

mwiencek:
> Isn’t allowing public access to SOLR "problematic from a security standpoint"?

Yep, that's exactly why Zas is working on the linked OpenResty config to filter insecure query parameters.

> I am working on embedding SOLR into the next MB VM which is likely to be used for mirror servers and expose port 8983 to other containers only, not to the outside world.

If the the port is only exposed to other containers, I don't see the security problem there.

yvanzo:
IIUYC, the best solution is to disallow public access to (local) search server and to use (local) /ws/2 instead.

mwiencek:
> IIUYC, the best solution is to disallow public access to (local) search server and to use (local) /ws/2 instead.

Hmm, I think I better understand the problem first. :) In the VM, all the containers would be running on the same docker network and can communicate freely. I'm not sure what you mean by disallow public access here: it's up to the user what ports they want to expose to the outside world, since they control their own machine. If they choose to expose their VM on the internet, they'd probably want to put a SOLR proxy in front of it to filter out unsafe requests. About accessing `/ws/2`: I assume they would want to do this anyway, but since they have complete control over their VM, they can also access SOLR directly if they'd like (from outside the VM, even without making it public on the internet).

This is a bit different from the problem described in b76be6296505342ad9a89f6dfe7aa46719d3fa53, which is about people accessing our production SOLR instance from their local mirror.

yvanzo:
Agreed. To clarify, this is a change from old search server which used to be accessible _publicly_, I mean, from outside the VM. I will suggest using https://github.com/dergachev/solr-security-proxy in VM notes.

mwiencek:
Got it. :) The SOLR proxy definitely makes sense if they're exposing SOLR on the internet for some reason, though I don't know why they would want to do that.

mwiencek:
Merged commit 1ccd25d00d4c51894dcc9b696d2583b4e6994569 into master and rebased, so this PR only concerns MBH-502 now.

mwiencek:
You just mean fixing the link to https://github.com/metabrainz/search-server, right? That I will do.

yvanzo:
Right! :100: 

Just found out that the introduction of [doc / Search Server](https://musicbrainz.org/doc/Search_Server) will need to be updated too, afterwards.

PR #789, opened 2018-12-21
Remove old i18n.expand

mwiencek:
Also removes the `__react: true` flag, which was used to switch between implementations. We no longer have to specify `__react: true` anywhere. `l`, `ln`, `lp`, etc. will only return a React fragment if the string contains HTML or if you interpolate a React element into it. Otherwise, it should always return a string and is suitable for use wherever the old `expand` was used - but see 5264a5b991bc47b6167e70520b347eb639e42009 for the fixes that had to be made.


root/static/scripts/common/i18n/expand2.js
```diff
@@ -386,7 +386,7 @@ export default function expand(source: string, args?: ?VarArgs): React.Node {
     result.length > 1
       ? React.createElement(React.Fragment, null, ...result)
       : result[0]
-  ) : null;
+  ) : '';
```
yvanzo: Is there an example use case that requires backward compatibility?

mwiencek: I think it makes no difference in practice. Some tests in root/static/scripts/tests/i18n/expand2.js fail if I revert this because there are strings like `'{x:%|}'` which can actually produce an empty result, but you wouldn't see those in actual code. I think it still makes sense to return an empty string in those cases, but I'll reword the commit message to make that more clear.

PR #790, opened 2018-12-25
Fix some warnings from search result pages

yvanzo:
Those warnings were harmless but to logs size.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Util/JSON.pm
+++ b/lib/MusicBrainz/Server/Entity/Util/JSON.pm
@@ -20,7 +20,8 @@ sub add_linked_entity {
     my ($entity_type, $id, $entity) = @_;
 
     my $entities = ($linked_entities->{$entity_type} //= {});
-    unless (defined $entities->{$id}) {
+    # schema fixup creates type instances without id
+    unless (!defined $id || defined $entities->{$id}) {
         $entities->{$id} = $entity;
     }
     return;
--- a/root/components/Paginator.js
+++ b/root/components/Paginator.js
@@ -43,21 +43,21 @@ const Paginator = ({
     <nav>
       <ul className="pagination">
         {pager.previous_page ? (
-          <li>
+          <li key="previous">
             <a href={uriWith(reqUri, {page: pager.previous_page})}>
               {l('Previous')}
             </a>
           </li>
         ) : (
-          <li>
+          <li key="no-previous">
             <span>{l('Previous')}</span>
           </li>
         )}
 
-        <li className="separator" />
+        <li className="separator" key="separate-previous" />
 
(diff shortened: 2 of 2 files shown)


root/layout/components/Footer.js
```diff
@@ -42,7 +42,7 @@ const Footer = ({$c, ...props}) => {
 
         {DBDefs.GIT_BRANCH ? (
           <>
-            <br />
+            <br key="git_details_br" />
```
mwiencek: Do these need keys? They don't appear in an array and I don't get any errors logged for them.

root/components/Paginator.js
```diff
@@ -43,74 +43,74 @@ const Paginator = ({
...
-          <li>
+          <li key="after-first">
             <span>{l('…')}</span>
           </li>
         ) : null}
 
         {range(start, end + 1).map(page => (
           (pager.current_page === page) ? (
-            <li>
+            <li key={"current-" + page}>
```
mwiencek: I would use `"number-" + page` here too since it can patch the existing element if it exists.

lib/MusicBrainz/Server/Entity/Util/JSON.pm
```diff
@@ -20,7 +20,7 @@ sub add_linked_entity {
     my ($entity_type, $id, $entity) = @_;
 
     my $entities = ($linked_entities->{$entity_type} //= {});
-    unless (defined $entities->{$id}) {
+    unless (!defined $id || defined $entities->{$id}) {
```
mwiencek: (For the record, happens because we don't have the ids when creating the type instances in the schema fixup methods.)

root/layout/components/Footer.js
yvanzo: Oops, that was before before https://github.com/metabrainz/musicbrainz-server/commit/9312de9a07d8b7aab8c1fa539c66987ba7768598. Removed from https://github.com/metabrainz/musicbrainz-server/pull/790/commits/9c8a90a688a2fda95ed746bc4e7b36d0137e6c0b.

root/components/Paginator.js
yvanzo: Replaced in https://github.com/metabrainz/musicbrainz-server/pull/790/commits/9c8a90a688a2fda95ed746bc4e7b36d0137e6c0b.

lib/MusicBrainz/Server/Entity/Util/JSON.pm
yvanzo: Commented in https://github.com/metabrainz/musicbrainz-server/pull/790/commits/ecb46fa25d8ff1d6e7aca86de74f810a8b6eb9b0.

PR #791, opened 2018-12-25
MBS-9957: Reindex release cover art on cover art type change

yvanzo:
### Fix [MBS-9957](https://tickets.metabrainz.org/browse/MBS-9957): Cover art image types not updated properly

by triggering a reindex from `cover_art_archive.cover_art_type` table change.


mwiencek:
LGTM, though it might be a good idea to add the upgrade script for this now in case we forget when the next schema change comes around. :)

mwiencek:
Oops, never mind, I guess we don't create the RabbitMQ triggers by default.

PR #792, opened 2018-12-26
MBS-9867: Fix display when editor data is missing

yvanzo:
## Fix [MBS-9867](https://tickets.metabrainz.org/browse/MBS-9867): Slave server cannot display page with annotation

- Display deleted tooltip `[missing editor]` on the model of deleted tooltip `[removed]` for deleted entity link. (Note that deleted editor are linked as `Deleted Editor #id`.)

- Return made up `[missing editor]` on sanitizing null editor.


root/utility/sanitizedEditor.js
```diff
@@ -7,7 +7,20 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-export default function sanitizedEditor(editor: EditorT): SanitizedEditorT {
+export default function sanitizedEditor(editor: EditorT | SanitizedEditorT | null): SanitizedEditorT {
+  if (!editor) {
+    editor = {
```
mwiencek: It does feel a bit weird to add this here, since this function is for removing private data. Can we just avoid calling `sanitizedEditor` to begin with by modifying the check at https://github.com/metabrainz/musicbrainz-server/blob/6643637/root/static/scripts/common/components/Annotation.js#L126 to `if (props.annotation && props.annotation.editor)` with a comment explaining when the editor is null?

root/static/scripts/common/components/EditorLink.js
```diff
@@ -6,10 +6,23 @@
...
+const MissingEditorLink = () => {
+  return (
+    <span className="deleted tooltip" title={l('This editor is missing from this server, and cannot be displayed correctly.')}>
+      {isolateText(l('[missing editor]'))}
+    </span>
+  );
+};
+
 const EditorLink = ({editor, content, avatarSize, subPath}) => {
+  if (!editor || editor.id == 1 && editor.name == '[missing editor]') {
```
mwiencek: I think the `!editor` check is sufficient if you remove the fake editor and let it stay `null`.

root/utility/sanitizedEditor.js
yvanzo: Agreed, done in https://github.com/metabrainz/musicbrainz-server/pull/792/commits/da3428846427526719bf706eaa3af9c2bc6afefe, I also set `editor` property in `AnnotationT` type to be potentially `null`.

root/static/scripts/common/components/EditorLink.js
yvanzo: Done in https://github.com/metabrainz/musicbrainz-server/pull/792/commits/da3428846427526719bf706eaa3af9c2bc6afefe

thomasboehm:
I'm getting this error on a page without annotation too.
https://musicbrainz.org/artist/183d6ef6-e161-47ff-9085-063c8b897e97

On my slave Musicbrainz VM I get the "Oops, something went wrong!" error.

PR #795, opened 2018-12-30
Workaround for MBS-9865

mwiencek:
I suggest trying to test/break this locally to see if it works and ignoring my useless commit message, since I got frustrated trying to fix this and could barely remember why things worked afterward. 😆

---

This works around the issue while we wait for the release editor to be converted to React, and for the artist credit editor/bubble code to be implemented properly.

The main fix is in setting the `target` and `componentInst` data values on the bubble div before rendering it, *not* after. I don't remember why that works, exactly, after half an hour in the Chrome debugger, but it's not important or interesting since this code is going in the scrap heap.

Fixing this then revealed that the "change matching artists" option was broken, because someone (me) thought mixing knockout and React was a good idea to begin with. The solution there is to (handwavy) bring some codey bits closer to React.

### DIFF
--- a/root/static/scripts/release-editor/bindingHandlers.js
+++ b/root/static/scripts/release-editor/bindingHandlers.js
@@ -42,11 +42,18 @@ ko.bindingHandlers.artistCreditEditor = {
         return $('#artist-credit-bubble').data('target');
     },
 
+    uncheckChangeMatchingArtists: function () {
+        const input = document.getElementById('change-matching-artists');
+        if (input) {
+            input.checked = false;
+        }
+    },
+
     previousTrack: function () {
         const entity = this.currentTarget();
         const prev = entity.medium.tracks()[entity.position() - 2];
         if (prev) {
-            prev.artistCreditEditorInst.updateBubble(true);
+            prev.artistCreditEditorInst.updateBubble(true, this.uncheckChangeMatchingArtists);
         }
     },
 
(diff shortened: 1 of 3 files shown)


root/static/scripts/release-editor/bindingHandlers.js
```diff
@@ -79,54 +58,43 @@ ko.bindingHandlers.artistCreditEditor = {
...
-
-        this.initialArtistText('');
     },
 
     update: function (element, valueAccessor) {
         const bindingHandler = ko.bindingHandlers.artistCreditEditor;
         const entity = valueAccessor();
+        // Subscribe to the artistCredit observable so that we
+        // re-render the ArtistCreditEditor when it changes.
+        entity.artistCredit();
```
mwiencek: In case you're not familiar with how knockout works, this update function runs inside a [computed observable](https://knockoutjs.com/documentation/computedObservables.html). Simply calling `entity.artistCredit()` (even ignoring the return value) registers it as a dependency, so this binding handler will run again when it changes.

yvanzo:
I cannot reproduce the issue, even without this patch.

mwiencek:
The steps as written in MBS-9865 reproduce the issue for me in Chrome or Firefox - can you summarize what you're trying?

mwiencek:
Never mind, found a fix for that too. @yvanzo, can you check if those two issues are resolved for you now?

yvanzo:
They are! :+1:

PR #801, opened 2019-01-07
i18n Flow types

mwiencek:
Edit: This description became outdated, so see the individual commits.


root/static/scripts/common/i18n/expand2text.js
```diff
@@ -0,0 +1,70 @@
...
+ * as var args. The result is always a string.
+ *
+ * This is useful in cases where strings are /required/, such as HTML
+ * title or aria attributes and select option values. It can also be
+ * used where plain text is expected (if not required), simply for the
+ * stricter types.
+ */
+export default function expand2text(
+  source: string,
+  args?: ?{+[string]: StrOrNum},
```
yvanzo: Shouldn’t `args` object be mandatory?

mwiencek: Yeah, for `expand2text` it should. (Not `expand2react`, since there are places like [here](https://github.com/metabrainz/musicbrainz-server/blob/eb98965/root/edit/components/EditSidebar.js#L61) where it's only used to convert HTML.)

One issue you've made me notice: `expand2text` still decodes HTML entities. So `expand2text('&lt;')` would become `'<'` (meaning it isn't entirely a no-op if no args are provided). I think this should perhaps be changed, since it doesn't handle any other HTML. Though it will require translators to be aware that they can probably only use HTML entities if the original string uses them. (I'm not sure why they would need to otherwise.)

reosarevok:
What's the specific benefit of having a separate AddColon etc for strings?

mwiencek:
> What's the specific benefit of having a separate AddColon etc for strings?

They have different return types, i.e. you otherwise can't use `addColon` where a string is expected, because it has a return type of `React.Element<any> | string`.

They also have different input types, e.g. passing a non-string to `addColonText` is a type error. It's also guaranteed to return a string at run-time, so it's even safe to use if receiving `any` typed data.

mwiencek:
@reosarevok @yvanzo sorry this took a while to fix/rebase, but I managed to split up the commits a bit more and reduce the PR size from +1,395 −914 to +977 −538. It's still larger than I'd like, but adding types to `l` affects a lot of code.

I'd love to get this in before I update #737 since it conflicts with that.

One thing I changed to reduce code churn is kept the `l(string, args)` signature for expansion (instead of adding `expand.l` and `expand.text.l`). This uses `expand2react` as before, so keeps compatibility with what we have. In order to force text now, you have to call it like `l(string, args, {text: true})`. I'm not sure how much I like this (perhaps `expand.l` and `expand.text.l` were better even though they required changing more code), so let me know what you think.

reosarevok:
So, what was the alternative to adding ", TEXT"? If it still involves adding something to every call, then ", TEXT" seems fine to me - if it can be made smarter in some way so we don't have to remember to add extra things to l() every time, I'd prefer that.

mwiencek:
> So, what was the alternative to adding ", TEXT"? If it still involves adding something to every call, then ", TEXT" seems fine to me - if it can be made smarter in some way so we don't have to remember to add extra things to l() every time, I'd prefer that.

Well, the alternative I previously had was adding separate methods for React/text, `expand.l()` and `expand.text.l()`. You'd have to import `expand` from the i18n module anytime you wanted to expand/interpolate something, and call one of those.

Unfortunately there's a limit to how smart we can make things, because Flow needs some way to determine what result you'll get. So we either need separate methods, or need to change the signature of `l()` to differentiate between React/text calls. I chose the latter (by [abusing function overloading](https://github.com/metabrainz/musicbrainz-server/blob/1f45848c5f66c7f7bd559d976b86ae28eaf08343/root/static/scripts/common/i18n/wrapGettext.js#L71-L73)) since it required changing less code, but we could do separate methods if that's preferred.

mwiencek:
Added 1ea773bf2a96f23d7e2f526c4e34ed5726251181 for comparison, which goes the route of having separate methods, though it increases the size of the PR by quite a bit. I chose the names `exp.l` (for React) and `texp.l` (for text) to minimize the increase in verbosity.

That commit also makes use of the Webpack [provide-plugin](https://webpack.js.org/plugins/provide-plugin/) to auto-import i18n functions. Which should limit any pain associated with the change by making it so you don't have to remember to import something new just to display a string.

mwiencek:
This is now running on https://test.musicbrainz.org

root/static/scripts/common/i18n/expand2.js
```diff
@@ -332,7 +329,13 @@ const parseLinkSubst = saveMatch(function (args) {
   return state.match;
 });
 
-const parseCondSubst = saveMatch(function (args) {
+const condSubstStart = /^\{([0-9A-z_]+):/;
```
yvanzo: `[0-9A-z_]` can be replaced with `\w`

mwiencek:
> Works for me! 🎊
> 
> Just hoping we won’t have to twist the code that much again to make it fit with flow.
> (Examples of made up tweaks: `NO_MATCH_VALUE`, `*2react` and `*2text`)

Yeah, I'm not happy about the `NO_MATCH` stuff at all.

The `expand2react` and `expand2text` I'm pretty happy with, since it's something we'd need with any type system, not just Flow. They have different behavior at run-time, too, so aren't strictly to please the type system.

PR #805, opened 2019-01-09
MBS-9975: Add release/track sortname guess-case (for aliases)

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9975

Release (and RG) and Track (so, Recording) have no sort name methods, because they don't have sort names. Which is great... but their aliases have, so those break. This just adds the simplest possible sort name method (copied from Work) which is, IMO, all they need.


reosarevok:
@brainzbot, test this please

PR #808, opened 2019-01-15
MBS-9849 (III): Port entity Tags tab to React and use it for RG

reosarevok:
Change-Id: I4b76974e58016da89db5a0eb652fd6c0ad7c082f


root/static/scripts/common/components/TagEditor.js
```diff
@@ -477,7 +477,7 @@ class MainTagEditor extends TagEditor {
         </p>
         <form id="tag-form" onSubmit={this.addTags}>
           <p>
-            <textarea row="5" cols="50" ref={this.setTagsInput}></textarea>
+            <textarea rows="5" cols="50" ref={this.setTagsInput}></textarea>
```
reosarevok: Not valid HTML5 otherwise

root/components/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+              )}
+            </div>
+          </>
+        ) : null
+      }
+    </div>
+  );
+};
+
+export default TagsComponent;
```
mwiencek: Do you remember what the issue was with just modifying `MainTagEditor` to add `$c.user_exist` to hide the login-only parts? It doesn't seem too difficult at a glance, so I'd prefer we did that instead of duplicating the component.

reosarevok: IIRC it was that the entire logic of the component was based on users existing and on voting, so it was hard to just hide the login-only parts.

root/entity/Tags.js
```diff
@@ -0,0 +1,53 @@
...
+  entity,
+  lastUpdated,
+  moreTags,
+  userTags,
+}: Props) => {
+  const entityType = entity.entityType;
+  const LayoutComponent = chooseLayoutComponent(entityType);
+
+  return (
+    <LayoutComponent page="tags" releaseGroup={entity} title={l('Tags')}>
```
mwiencek: `releaseGroup={entity}` will be incorrect for other entity types, though. I suggest that we just modify all the entity-specific layout components to accept an `entity` prop. You can still alias it as `releaseGroup` etc. in the props destructuring:

```JavaScript
const ReleaseGroupLayout = ({
  entity: releaseGroup, // prop "entity" is assigned to "releaseGroup" in the function body
  ...otherProps
}) => { /* ... */ };
```

root/static/scripts/common/components/TagEditor.js
```diff
@@ -575,9 +619,10 @@ function createInitialTagState(
 }
 
 function init_tag_editor(Component, mountPoint) {
-  return function (entity, aggregatedTags, userTags, more) {
+  return function ($c, entity, aggregatedTags, userTags, more) {
```
mwiencek: Ah, the call to `MB.init_main_tag_editor` in root/entity/tags.tt will need to be passed a `$c` now though. For now, we can probably do this:

```
          MB.init_main_tag_editor(
            {user_exists: true},
            {entityType: '[% entity_type %]', gid: '[% entity.gid %]'},
            [% closing_tag_escape(tags_json) %],
            [% closing_tag_escape(user_tags_json) %],
            [% more_tags ? 'true' : 'false' %]
          );
```
i.e. pass `{user_exists: true}` for `$c`. It's not a real context, but should work fine because `user_exists` is the only property accessed by the component (and JS inside TT is not checked by Flow). We'll be removing `init_tag_editor` soon once all entities are converted, anyway.

root/entity/Tags.js
mwiencek: (Also, I was wondering why Flow didn't complain about that and realized `chooseLayoutComponent` can literally only return `ReleaseGroupLayout` at present, so it out-smarted me.)

lib/MusicBrainz/Server/Controller/Role/Tag.pm
```diff
@@ -33,13 +33,33 @@ sub tags : Chained('load') PathPart('tags') {
...
-        display_tags => \@display_tags,
-        display_genres => \@display_genres,
-        tags_json => $c->json->encode(\@tags),
-        template => 'entity/tags.tt',
-    );
+    my $tags_model = $c->model($self->{model})->tags;
+    my @user_tags = $tags_model->find_user_tags($c->user->id, $entity->id)
+        if $c->user_exists;
+    my $count = $tags_model->find_tag_count($entity->id);
+    my @toptags = $tags_model->find_top_tags($entity->id, $TOP_TAGS_COUNT);
```
mwiencek: (Sorry, missed this before.) All(?) the lines you added here are already done in `load` above - you can just reference `$c->stash->{top_tags}`, `$c->stash->{user_tags}`, etc. to get the same data.

reosarevok: Should be done now :)

PR #809, opened 2019-01-15
MBS-9849 (IV): Port RG Alias tab to React

reosarevok:
Change-Id: I413df4ab3d022ebd2d9d0dba1f4bd0836cb6f136

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Alias.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Alias.pm
@@ -1,6 +1,7 @@
 package MusicBrainz::Server::Controller::Role::Alias;
 use Moose::Role -traits => 'MooseX::MethodAttributes::Role::Meta::Role';
 use MusicBrainz::Server::ControllerUtils::Delete qw( cancel_or_action );
+use MusicBrainz::Server::Data::Utils qw( type_to_model );
 
 requires 'load';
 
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/Controller/Role/Alias.pm
```diff
@@ -42,9 +44,26 @@ sub aliases : Chained('load') PathPart('aliases')
...
     my $aliases = $m->alias->find_by_entity_id($entity->id);
     $m->alias_type->load(@$aliases);
-    $c->stash(
-        aliases => $aliases,
-    );
+
+    if ($entity->entity_type eq 'release_group') {
+        my %props = (
+            aliases => $aliases,
+            canonicalURL => MusicBrainz::Server::Plugin::Canonicalize::replace_gid($self, $c, $entity->gid),
```
mwiencek: I don't think this is needed anymore now that we redirect. The MBID in the URL will already be the same as `$entity->gid`.

lib/MusicBrainz/Server/Controller/Role/Alias.pm
```diff
@@ -42,9 +44,26 @@ sub aliases : Chained('load') PathPart('aliases')
     my $m = $c->model($self->{model});
     my $aliases = $m->alias->find_by_entity_id($entity->id);
     $m->alias_type->load(@$aliases);
-    $c->stash(
```
mwiencek: You can have as many `$c->stash` calls as you want, so can just leave this here and not have a separate `else` below.

lib/MusicBrainz/Server/Controller/Role/Alias.pm
mwiencek: Opened https://github.com/metabrainz/musicbrainz-server/pull/814 to clean that up.

lib/MusicBrainz/Server/Controller/Role/Alias.pm
```diff
@@ -1,6 +1,8 @@
 package MusicBrainz::Server::Controller::Role::Alias;
 use Moose::Role -traits => 'MooseX::MethodAttributes::Role::Meta::Role';
 use MusicBrainz::Server::ControllerUtils::Delete qw( cancel_or_action );
+use MusicBrainz::Server::Data::Utils qw( type_to_model );
+use MusicBrainz::Server::Plugin::Canonicalize qw ( replace_gid );
```
mwiencek: This import can be removed too, but LGTM otherwise

PR #810, opened 2019-01-15
MBS-9926: Change subscriptions to use React (1 - per-entity pages)

reosarevok:
Change-Id: I3cf9190bfc4a2ce70d22dd88d407dd92b62cd031


root/artist/ArtistSubscribers.js
```diff
@@ -0,0 +1,99 @@
...
+  subscribed,
+}: Props) => (
+  <ArtistLayout artist={entity} canonicalURL={canonicalURL} page="subscribers" title={l('Subscribers')}>
+    <h2>{l('Subscribers')}</h2>
+
+    {(publicEditors.length || (privateEditors > 0)) ? (
+      <>
+        <p>{ln(
+          'There is currently {num} user subscribed to {artist}:',
+          'There are currently {num} users subscribed to {artist}:',
```
mwiencek: Would it be possible to have a single, generic `root/entity/Subscribers.js` component using the `chooseLayoutComponent` utility you added?

From what I can tell, only some of the strings differ, and that's no issue if we use `N_ln` etc. and store them as constants at the top of the file:

```JavaScript
const subscriberCountMessages = {
  artist: N_ln(
    'There is currently {num} user subscribed to {artist}:',
    'There are currently {num} users subscribed to {artist}:',
  ),
  label: N_ln(/* ... */),
};

// below, in render()...
<p>
  {subscriberCountMessages[entityType].toLocaleString(
    count,
    {[entityType]: entity.name, num: count},
  )}
</p>
```

root/artist/ArtistSubscribers.js
```diff
@@ -0,0 +1,99 @@
...
+        </p>
+
+        <ul>
+          {publicEditors.map(editor => (
+            <li key={editor.id}>
+              <EditorLink editor={editor} />
+            </li>
+          ))}
+          {publicEditors.length && (privateEditors > 0) ? (
+            <li>{ln(
```
mwiencek: Nitpick, `<li>` should be on its own line unless the entire thing fits on one line. I committed 916c4362d798048f42ee1d69cbdd37f7e8d2a392 so that you can just run `eslint --fix` to have it correct automatically.

root/artist/ArtistSubscribers.js
mwiencek: The URL should always be the same except for the entity type, so `` `/account/subscriptions/${entityType}/remove?id=${entity.id}` `` should work without needing a constant. AFAICT there are only two translated strings that would need to be moved to use `N_ln` — for the rest we could just use JS template strings and computed property names (e.g. `{[entityType]: foo}` is the same as `{artist: foo}` if `entityType` is "artist").

PR #813, opened 2019-01-16
MBS-9985: Edit LastUpdated to deal with null "dates"

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9985

Change-Id: Icb6fb511010adc094a1b38b65debd25ca4af3915

### DIFF
--- a/root/layout/components/sidebar/LastUpdated.js
+++ b/root/layout/components/sidebar/LastUpdated.js
@@ -18,12 +18,19 @@ type Props = {|
   +entity: CoreEntityT,
 |};
 
-const LastUpdated = ({$c, entity}: Props) => (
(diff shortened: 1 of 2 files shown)


mwiencek:
P.S. You can `rm .git/hooks/commit-msg` or move it now since it was only really useful with Gerrit.

reosarevok:
IMO it's useful to show something because it's much more consistent than not showing anything in some cases (that might lead to a second bug report "last updated does not appear on some releases" ;) )

mwiencek:
One thing I forgot here: can we change the type of `last_updated` in `LastUpdateRoleT` (root/types.js) to `string | null`?

PR #814, opened 2019-01-16
Remove replace_gid

mwiencek:
This doesn't seem to be needed anymore now that we perform a 301 redirect for merged MBIDs. The MBID in the URL should always be the same as the main entity gid in that case.


mwiencek:
@brainzbot test this please

PR #816, opened 2019-01-16
 Shell-quote database_configuration output

mwiencek:
The password on my development DB has some non-friendly shell characters in it.


mwiencek:
Selenium test failure is unrelated.

PR #817, opened 2019-01-16
Move image_type inserts to t/sql/initial.sql

mwiencek:
Just some cleanup.


reosarevok:
I understand this is deduplicating inserts and always running them while creating a test DB? If so, it seems sensible.

mwiencek:
You are correct. t/sql/initial.sql contains all the initial data for the test database, as you guessed, so basically anything constant (like types) that don't change between tests.

PR #819, opened 2019-01-16
Disable t/critic.t

mwiencek:
We don't add a ton of new Perl code these days, so I think this is of limited utility to have run on *every* build (even ones where no Perl is touched). I don't remember the last time it reported an issue, either. It takes around a minute to run this locally, so I think it could save a good chunk of time if we disable it. If someone does add a lot of Perl code in some PR, it can still easily be run locally.

### DIFF
--- a/docker/musicbrainz-tests/run_tests.sh
+++ b/docker/musicbrainz-tests/run_tests.sh
@@ -47,7 +47,6 @@ exec sudo -E -H -u musicbrainz carton exec -- prove \
     --source pgTAP \
     -I lib \
     -j 9 \
-    t/critic.t \
     t/flow.sh \
     t/js.t \
     t/web.js \


reosarevok:
I understand that t/critic is basically eslint for Perl? In which case, it's probably not too critic even if Perl files are added, so it's probably ok to remove this since it won't cause a huge mess even if we forget?

mwiencek:
Yeah, we use it for linting Perl just like eslint is used for JS (though the number of rules it checks is [much smaller](https://github.com/metabrainz/musicbrainz-server/blob/master/.perlcriticrc)), and I agree there is only a very small chance we'll create a huge mess right after this is removed from the tests. :) Most of the checks are obvious things you'd see in review, other than whitespace.

PR #822, opened 2019-01-17
Per-database lock file in LoadReplicationChanges

mwiencek:
This is needed because both DumpJSON.t and ExportAllTables.t use LoadReplicationChanges, and they may run concurrently during tests.


mwiencek:
Nope, the lock file is a hidden implementation detail. And 99.99% of slave server admins will be using `READWRITE`, so this doesn't even change the location of the lock file for them.

mwiencek:
Updated to just specify `--lockfile` in the tests, which achieves the same result without modifying LoadReplicationChanges at all.

PR #823, opened 2019-01-17
Add missing imports

mwiencek:
This fixes all of the undef variables eslint found in our own code (but not in external libraries, e.g. the ones under root/static/lib/).

The command I used to find them is:

```sh
./node_modules/.bin/eslint \
  --no-eslintrc \
  --env 'browser,es6,node' \
  --parser babel-eslint \
  --plugin react \
  --plugin flowtype \
  --rule 'no-undef: [error]' \
  --rule 'flowtype/define-flow-type: warn' \
  root/static/scripts/
```

This is a prerequisite to Webpack.

### DIFF
--- a/root/static/scripts/common/MB/Control/EditList.js
+++ b/root/static/scripts/common/MB/Control/EditList.js
@@ -3,7 +3,10 @@
 // Licensed under the GPL version 2, or (at your option) any later version:
 // http://www.gnu.org/licenses/gpl-2.0.txt
 
+const $ = require('jquery');
+
 const {l} = require('../../i18n');
+const MB = require('../../MB');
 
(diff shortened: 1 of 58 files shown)


reosarevok:
Probably a silly question but... how did all this stuff actually work before, then?

mwiencek:
> Probably a silly question but... how did all this stuff actually work before, then?

We add them to `window` at https://github.com/metabrainz/musicbrainz-server/blob/e738fc80b8b36e1e3985147ef2803dc71c857192/root/static/scripts/common.js#L25-L27 which makes them global variables (which is bad and should be removed, though some jQuery/Knockout plugins need those to be global).

PR #824, opened 2019-01-17
Bump React to 16.7.0

mwiencek:
Just trying to not lag too far behind before it becomes a pain to update.

There should be no breaking changes for us. There are a ton of bug fixes, a few new features, and a couple fixed vulnerabilities.

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -803,7 +803,7 @@ END ~%]
 -%]
     [%- React.embed(c, 'static/scripts/common/components/WikipediaExtract', {
         entity => {entityType => entity.entity_type, gid => entity.gid},
-        wikipediaExtract => wikipedia_extract,
+        cachedWikipediaExtract => wikipedia_extract,
     }) -%]
 [%- END -%]
 
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "querystring": "0.2.0",
     "raven": "2.0.0",
     "raven-js": "3.17.0",
-    "react": "16.3.2",
+    "react": "16.7.0",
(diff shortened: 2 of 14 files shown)


reosarevok:
What does the new scheduler do? (since it's not mentioned in the desc)

mwiencek:
> What does the new scheduler do? (since it's not mentioned in the desc)

Not entirely sure, but it has something to do with https://reactjs.org/blog/2018/11/27/react-16-roadmap.html#react-16x-q2-2019-the-one-with-concurrent-mode which should be nice for improving responsiveness, though doesn't appear to be stable yet.

mwiencek:
There was a bug fix to `getDerivedStateFromProps` in 16.4 which apparently broke a small number of peoples' code. I don't think we'd have been affected, but it turned out all our uses of this method were anti-patterns, anyway, so I removed all of them. See https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

root/components/common-macros.tt
```diff
@@ -803,7 +803,7 @@ END ~%]
 -%]
     [%- React.embed(c, 'static/scripts/common/components/WikipediaExtract', {
         entity => {entityType => entity.entity_type, gid => entity.gid},
-        wikipediaExtract => wikipedia_extract,
+        cachedWikipediaExtract => wikipedia_extract,
```
mwiencek: This should help make clear what the prop's for, and that it won't affect the displayed extract if the prop changes. (The prop couldn't change here anyway.)

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

PR #830, opened 2019-01-23
MBS-9994: Sitemaps are not rsynced to gateways

mwiencek:
This failed because `$STATICBRAINZ_SERVERS` (set on the container via `--env`) is unset when bin/rsync-staticbrainz-files runs.

I've made three changes to address this:

 1. Run rsync-staticbrainz-files with `bash -u` so that it errors on unset variables.
 2. Source /etc/container_environment.sh in the crontab so that rsync-staticbrainz-files has access to `$STATICBRAINZ_SERVERS`.
 3. Give the musicbrainz user access to /etc/container_environment.sh by changing permissions in the Dockerfile. This is described in https://github.com/phusion/baseimage-docker#security

### DIFF
--- a/bin/rsync-staticbrainz-files
+++ b/bin/rsync-staticbrainz-files
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -u
 
 KEY="$1"
 SOURCE="$2"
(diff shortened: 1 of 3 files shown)


mwiencek:
Selenium test failure is unrelated (probably due to JSON dumps running on williams and slowing stuff down).

PR #833, opened 2019-01-23
MBS-9909: Change reports to use React (3 - events)

reosarevok:


root/report/components/EventList.js
```diff
@@ -0,0 +1,77 @@
...
+        {items.map((item, index) => (
+          <tr className={loopParity(index)} key={item.event.id}>
+            <td>
+              <EntityLink
+                entity={item.event}
+                showDisambiguation
+                showEventDate={false}
+              />
+            </td>
+            <td>{item.event.typeName ? lp_attributes(item.event.typeName, 'item.event_type') : null}</td>
```
mwiencek: Small context error:
```suggestion
            <td>{item.event.typeName ? lp_attributes(item.event.typeName, 'event_type') : null}</td>
```

root/report/EventSequenceNotInSeries.js
```diff
@@ -0,0 +1,48 @@
...
+  items,
+  pager,
+}: ReportDataT<ReportEventT>) => (
+  <Layout fullWidth title={l('Events which should be part of series or larger event')}>
+    <h1>{l('Events which should be part of series or larger event')}</h1>
+
+    <ul>
+      <li>
+        {l(`This report lists events where the event name indicates that it may
+        have to be part of a series or a larger event.`)}
```
mwiencek:
```suggestion
            have to be part of a series or a larger event.`)}
```
(so that it starts at the same column as "This" in the first line)

root/report/DuplicateEvents.js
```diff
@@ -0,0 +1,49 @@
...
+  items,
+  pager,
+}: ReportDataT<ReportEventT>) => (
+  <Layout fullWidth title={l('Possible duplicate events')}>
+    <h1>{l('Possible duplicate events')}</h1>
+
+    <ul>
+      <li>
+        {l(`This report lists events happening at the same place on the same date.
+            If they are duplicates (for example, if there are separate events for headliner
```
mwiencek: \<annoying\>I was gonna complain about changing "they're" to "they are," but actually I'd suggest using neither and switch to "there are" here, because "they" refers to the entire list to me, which doesn't make as much sense. And we already correctly use "there are" in the following parentheses.\</annoying\>

root/report/EventSequenceNotInSeries.js
reosarevok: I'm pretty sure that's how I had it at first and eslint complained?

reosarevok: Or maybe that was something else :D Changed now.

root/report/DuplicateEvents.js
```diff
@@ -0,0 +1,49 @@
...
+  items,
+  pager,
+}: ReportDataT<ReportEventT>) => (
+  <Layout fullWidth title={l('Possible duplicate events')}>
+    <h1>{l('Possible duplicate events')}</h1>
+
+    <ul>
+      <li>
+        {l(`This report lists events happening at the same place on the same date.
+            If there are duplicates (for example, if there are separate events for headliner
```
yvanzo: Message should be changed in a separate commit, so as to allow noticing and reviewing such change. (Just noticed changes from previous PRs.)

PR #834, opened 2019-01-23
MBS-9632: Convert Stats to React (III)

reosarevok:
Index


root/statistics/Index.js
```diff
@@ -0,0 +1,1078 @@
...
+        <tbody>
+          <tr className="thead">
+            <th colSpan="5">{l_statistics('Release Status')}</th>
+          </tr>
+          <tr>
+            <th colSpan="3">{l_statistics('Releases:')}</th>
+            <td>{formatCount(stats.data['count.release'], $c)}</td>
+            <td />
+          </tr>
+          {((Object.values(statuses): any): $ReadOnlyArray<ReleaseStatusT>).map(status => (
```
reosarevok: These "any" are really not something that eslint is happy with. How could they be replaced? (these came from the original spellew work, and I'm honestly not fully sure what they do).

root/statistics/Index.js
```diff
@@ -0,0 +1,1078 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+import {range} from 'lodash';
+
+import manifest from '../static/manifest';
+import {addColon} from '../static/scripts/common/i18n';
+import {lp_attributes} from '../static/scripts/common/i18n/attributes';
+import {l_statistics, ln_statistics, lp_statistics} from '../static/scripts/common/i18n/statistics';
```
mwiencek: These should be `l_statistics as l, ln_statistics as ln, lp_statistics as lp`. This is only needed for the statistics domain - attributes for example are extracted from the database, not source files, so `lp_attributes` is fine.

root/statistics/Index.js
```diff
@@ -0,0 +1,1078 @@
...
+  placeTypes,
+  primaryTypes,
+  secondaryTypes,
+  seriesTypes,
+  stats,
+  statuses,
+  workAttributeTypes,
+  workTypes,
+}: MainStatsT) => {
+  const nonGroupCount = stats.data['count.artist.type.null'] + stats.data['count.artist.type.person'] + stats.data['count.artist.type.character'] + stats.data['count.artist.type.other'];
```
mwiencek: Very Long Line!

root/statistics/types.js
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export type StatsT = {|
+  +data: {[string]: number},
+  +date_collected: string,
```
mwiencek: You already pass a `dateCollected` prop, so I'm not sure why this needs to be in this object. Or rather, why `StatsT` isn't just `{[string]: number}`, so that instead of having to do `stats.data['foo']` everywhere, you can just do `stats['foo']`.

mwiencek: And if it's just `{[string]: number}` I would delete this type and this file, since you can inline `{[string]: number}` in the one place it's used.

mwiencek:
LGTM, but please amend the commit message to credit Shamroy Pellew/spellew for his work which this was based on.

reosarevok:
How should that be done? :)

mwiencek:
Nothing fancy, maybe something like this:

```
Author: Shamroy Pellew (spellew), with additional work by me
Original PR: https://github.com/metabrainz/musicbrainz-server/pull/609
```

reosarevok:
Ok, done

PR #835, opened 2019-01-24
Set CACHE_NAMESPACE during JSONDump/Sitemap tests

mwiencek:
This is the logical follow-up to d26c649.


mwiencek:
Done so in https://github.com/metabrainz/musicbrainz-server/pull/835/commits/b159916a28ef4ba92e7a7014aa229b49325b64ac

PR #836, opened 2019-01-24
Move i18n.strings strings

mwiencek:
`i18n.strings` isn't a great idea, since it encourages hoarding unrelated strings in a single file (perhaps causing them to be loaded even if they aren't needed).

Move these constants to the files where they're actually used, or to constants.js if they're actually needed by many pages.

This is cherry-picked from #801.


root/static/scripts/common/dialogs.js
```diff
@@ -91,7 +91,6 @@ const MB = require('./MB');
             }
 
             var entity = this.options.entity;
-            this.options.title = strings.addANewEntity[entity];
```
reosarevok: This is being removed but not substituted by anything - is that intentional?

yvanzo: It is substituted at [line no. 454 of `MB/Control/Autocomplete.js`](https://github.com/metabrainz/musicbrainz-server/pull/836/files#diff-a122aa5e611cd820d2a4644b76d96dccR454).

PR #837, opened 2019-01-24
React.embed icons

mwiencek:
These need to be moved into JavaScript in order to take advantage of Webpack's file-loader: https://webpack.js.org/loaders/file-loader/

This'll allow us to `require()` image files in JavaScript and have them be copied to our build directory (root/static/build/) and versioned automatically. At runtime, the require call will return the URL of the image.


root/layout/components/sidebar/SidebarLicenses.js
```diff
@@ -20,23 +20,59 @@ import {l} from '../../../static/scripts/common/i18n';
...
-  CCBYNCND: /creativecommons\.org\/licenses\/by-nc-nd\//,
-  CCBYNCSA: /creativecommons\.org\/licenses\/by-nc-sa\//,
-  CCBYND: /creativecommons\.org\/licenses\/by-nd\//,
-  CCBYSA: /creativecommons\.org\/licenses\/by-sa\//,
-  CCNCSamplingPlus: /creativecommons\.org\/licenses\/nc-sampling\+\//,
-  CCPD: /creativecommons\.org\/licenses\/publicdomain\//,
-  CCSampling: /creativecommons\.org\/licenses\/sampling\//,
-  CCSamplingPlus: /creativecommons\.org\/licenses\/sampling\+\//,
+  ArtLibre: {
+    icon: manifest.pathTo('/images/licenses/ArtLibre.png'),
```
mwiencek: There's a way to do dynamic imports like ``require(`../images/licenses/${className}.png`)`` with Webpack, though I haven't tested it and it's easier to just specify them all separately.

root/static/scripts/common/components/SearchIcon.js
```diff
@@ -0,0 +1,23 @@
...
+import * as React from 'react';
+
+import * as manifest from '../../../manifest';
+import {l} from '../i18n';
+
+const SearchIcon = () => (
+  <img
+    alt={l('Search')}
+    className="search"
+    src={manifest.pathTo('/images/icons/search.svg')}
```
mwiencek: This uses the svg icon everywhere now.

PR #840, opened 2019-01-24
MBS-9742 (from comment):  Allow LC Authority File URL for Label

reosarevok:
as requested in comment to MBS-9742, https://tickets.metabrainz.org/browse/MBS-9742

PR #841, opened 2019-01-24
Run eslint --fix on URLCleanup

reosarevok:
Step one of many of making this less messy

Includes the tiny one line from https://github.com/metabrainz/musicbrainz-server/pull/840 since it's the same file and I fully expect that to be merged way before this :)

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1,251 +1,253 @@
-// This file is part of MusicBrainz, the open internet music database.
-// Copyright (C) 2010 MetaBrainz Foundation
-// Licensed under the GPL version 2, or (at your option) any later version:
-// http://www.gnu.org/licenses/gpl-2.0.txt
+/*
+ * This file is part of MusicBrainz, the open internet music database.
+ * Copyright (C) 2010 MetaBrainz Foundation
+ * Licensed under the GPL version 2, or (at your option) any later version:
+ * http://www.gnu.org/licenses/gpl-2.0.txt
+ */
 
 const _ = require('lodash');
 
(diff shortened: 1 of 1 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -299,37 +308,38 @@ function disallow(url, id) {
  * - clean: Function to clean up/normalize matched URL.
  * - validate: Function to validate matched (clean) URL
  *             for an auto-selected relationship type.
- **/
+ *
```
yvanzo: Unneeded line? Probably unneeded extra `*` on line 301 too.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -299,37 +308,38 @@ function disallow(url, id) {
...
+ */
 const CLEANUPS = {
-  wikipedia: {
-    match: [new RegExp("^(https?://)?(([^/]+\\.)?wikipedia|secure\\.wikimedia)\\.","i")],
+  'wikipedia': {
```
yvanzo: What makes eslint add single quotes to `CLEANUPS` keys?

reosarevok: Some keys were already quoted because they start with numbers, so I assume it's for consistency inside the same object

yvanzo:
For some reason, character diff stops to display correctly on GitHub after line 1130, I did not got to the end yet.

Can you please be more specific about rules being applied when changing about 900 lines at once?

reosarevok:
Yeah, the test failure is because of that, I just fixed it in the original PR and plan to rebase this when that gets merged

yvanzo:
Merged #840 (with direct link to the comment)

PR #842, opened 2019-01-24
Use separate DB for sitemaps tests

mwiencek:
Similar to what was done for the DumpJSON and ExportAllTables tests, create a separate musicbrainz_test_sitemaps DB to allow running it concurrently in a safe manner.

In order to use this DB for requests made during the incremental sitemaps tests, I've adapted the `selenium` header into a more generic `mb-set-database` header to allow selecting any database for the duration of the request.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -183,7 +183,7 @@ These are a bit more involved to set up:
    Just make sure the connection options match what you're using for the
    `TEST` database.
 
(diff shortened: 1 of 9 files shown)


mwiencek:
Added to DBDefs.pm.sample with a shorter, more relevant comment given its purpose.

PR #844, opened 2019-01-24
Eslint fixes for URLCleanup test

reosarevok:
Includes #840 since it touches the same file and I fully expect that to be merged before this :)

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -1,3517 +1,3523 @@
-// This file is part of MusicBrainz, the open internet music database.
-// Copyright (C) 2014 MetaBrainz Foundation
-// Licensed under the GPL version 2, or (at your option) any later version:
-// http://www.gnu.org/licenses/gpl-2.0.txt
+/*
+ * This file is part of MusicBrainz, the open internet music database.
+ * Copyright (C) 2014 MetaBrainz Foundation
+ * Licensed under the GPL version 2, or (at your option) any later version:
+ * http://www.gnu.org/licenses/gpl-2.0.txt
+ */
 
 import test from 'tape';
 import _ from 'lodash';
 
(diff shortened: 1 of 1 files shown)


reosarevok:
I can maybe see the key order, although I don't particularly think having it nice and alphabetical makes it worse, but the indents just look and feel weird to me and I don't feel they provide anything of use - the file is equally readable after a6afd69f8c57b10a29457344af605b609406fb6e

yvanzo:
Values alignment helps to compare these. For example, it is particularly helpful with comparing differences between input URL and expected output URL. If _the file is equally readable after_ (which I don't agree with), then maybe this commit is not worth it.

reosarevok:
I feel following the same rules everywhere and the consistency that gives is enough of a reason to do it, and dislike the idea of disabling rules on a per-file basis, tbh. 

Re: indent, at the very least I'd expect to see everything unindented to the left of the file except for the key-value pairs themselves.

yvanzo:
It doesn’t sound very rational. Ability to [disable specific eslint rules on demand](https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments) is there for a reason.

mwiencek:
So let's revert the changes to `testData` and add `eslint-disable indent, sort-keys`/`eslint-enable indent, sort-keys` comments around it? Since @yvanzo is the primary maintainer of this file I don't see an advantage in making it less readable for him.

reosarevok:
What about now? :) 

I changed the 4 space indents to 2 spaces, but left the visual indentation untouched and the keys unsorted.

I moved a few comments that were after the first key to be on top of the first key (so https://github.com/metabrainz/musicbrainz-server/pull/844/files#diff-eb68718856b483c030b05815375ce614L3292 to https://github.com/metabrainz/musicbrainz-server/pull/844/files#diff-eb68718856b483c030b05815375ce614R3291) because that feels much more clear to me, but I can undo if desired.

PR #845, opened 2019-01-24
MBS-9993: /oauth2/ CORS

mwiencek:
Allow * origins and add support for preflight requests to the following
endpoints:

  /oauth2/token
  /oauth2/tokeninfo
  /oauth2/userinfo

The browser should normally only send OPTIONS to /oauth2/userinfo since
it makes use of a non-safelisted header (Authorization).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/OAuth2.pm
+++ b/lib/MusicBrainz/Server/Controller/OAuth2.pm
@@ -119,7 +119,12 @@ sub token : Local Args(0)
 {
     my ($self, $c) = @_;
 
+    $c->res->header('Access-Control-Allow-Origin' => '*');
+
     $self->_enforce_tls($c);
+
+    $self->_send_options_response($c, 'POST')
+        if $c->request->method eq 'OPTIONS';
 
     $self->_send_error($c, 'invalid_request', 'Only POST requests are allowed')
         if $c->request->method ne 'POST';
@@ -228,6 +233,13 @@ sub _send_error
     });
 }
 
+sub _send_options_response {
+    my ($self, $c, $allow) = @_;
+
+    $c->res->headers->header('Allow' => "$allow, OPTIONS");
+    $self->_send_response($c, {message => 'OK'});
+}
+
 sub _send_response
 {
     my ($self, $c, $response) = @_;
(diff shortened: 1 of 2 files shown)


mwiencek:
I have this on test.mb and it appears to work with @kepstin's magicisrc beta.

lib/MusicBrainz/Server/Controller/OAuth2.pm
```diff
@@ -119,6 +119,10 @@ sub token : Local Args(0)
 {
     my ($self, $c) = @_;
 
+    $c->res->header('Access-Control-Allow-Origin' => '*');
+    $self->_send_options_response($c, 'POST')
+        if $c->request->method eq 'OPTIONS';
+
     $self->_enforce_tls($c);
```
yvanzo: Should this be checked first?

lib/MusicBrainz/Server/Controller/OAuth2.pm
```diff
@@ -312,6 +323,10 @@ sub tokeninfo : Local
 {
     my ($self, $c) = @_;
 
+    $c->res->header('Access-Control-Allow-Origin' => '*');
+    $self->_send_options_response($c, 'GET')
+        if $c->request->method eq 'OPTIONS';
+
     $self->_enforce_tls($c);
```
yvanzo: Same as above

lib/MusicBrainz/Server/Controller/OAuth2.pm
```diff
@@ -344,6 +359,13 @@ sub userinfo : Local
...
     my ($self, $c) = @_;
 
+    $c->res->header('Access-Control-Allow-Origin' => '*');
+
+    if ($c->request->method eq 'OPTIONS') {
+        $c->res->headers->header('Access-Control-Allow-Headers' => 'authorization');
+        $self->_send_options_response($c, 'GET');
+    }
+
     $self->_enforce_tls($c);
```
yvanzo: Same as above

lib/MusicBrainz/Server/Controller/OAuth2.pm
mwiencek: Yeah, well, it should go after the `$c->res->header` line and before the `$self->_send_options_response` line. Fixed now.

PR #846, opened 2019-01-24
MBS-9909: Change reports to use React (4 - instruments)

reosarevok:


root/report/InstrumentsWithoutAnImage.js
```diff
@@ -0,0 +1,46 @@
...
+  items,
+  pager,
+}: ReportDataT<ReportInstrumentT>) => (
+  <Layout fullWidth title={l('Instruments without an image')}>
+    <h1>{l('Instruments without an image')}</h1>
+
+    <ul>
+      <li>
+        {l(`This report shows instruments without image
+              relationships nor Wikidata relationships.`)}
```
mwiencek: The string change here should go in a follow-up commit as @yvanzo suggested in a previous PR.

root/report/instruments_without_an_image.tt
```diff
@@ -1,42 +0,0 @@
-[%- WRAPPER 'layout.tt' title=l('Instruments without an image') full_width=1 -%]
-
-<h1>[% l('Instruments without an image') %]</h1>
-
-<ul>
-    <li>[% l('This report shows instruments without {image_rel}
-              relationships or {wikidata_rel} relationships.', {
-              image_rel => 'image', wikidata_rel => 'wikidata' }) %]</li>
```
yvanzo: Oops, that was unfinished code, it should rather link to:
* https://musicbrainz.org/relationship/f64eacbd-1ea1-381e-9886-2cfb552b7d90
* https://musicbrainz.org/relationship/1486fccd-cf59-35e4-9399-b50e2b255877
and use `l_relationships` for labels.

reosarevok: Well, it doesn't really matter because I'm removing the links (which are useless, since this is meant for instrument editors anyway - nobody else can edit the instruments in the first place) with this PR :)

mwiencek: That makes sense. This seems resolved?

yvanzo: Well, my intent was to use localization of "image" and "wikidata" relationship types with `l_relationships` rather than requiring translators to keep it in sync. But that would require to map relationship type names with MBIDs in JavaScript, that is, `LINK_TYPES` from `edit/URLCleanup.js`. So, that’s not possible for now. :(

PR #849, opened 2019-01-25
Add Flow types to formatDatePeriod

mwiencek:
This is cherry-picked from #801.

### DIFF
--- a/root/static/scripts/common/utility/formatDatePeriod.js
+++ b/root/static/scripts/common/utility/formatDatePeriod.js
@@ -1,3 +1,4 @@
+// @flow
 // This file is part of MusicBrainz, the open internet music database.
 // Copyright (C) 2015–2016 MetaBrainz Foundation
 // Licensed under the GPL version 2, or (at your option) any later version:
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/utility/formatDatePeriod.js
```diff
@@ -8,12 +9,12 @@ const ko = require('knockout');
 const {l} = require('../i18n');
 import formatDate from './formatDate';
 
-function formatDatePeriod(entity) {
+function formatDatePeriod<+T: {...DatePeriodRoleT}>(entity: T) {
```
mwiencek: `DatePeriodRoleT` is an exact object type, so `{...DatePeriodRoleT}` is done to make it inexact (meaning `T` can have other properties outside of the ones in `DatePeriodRoleT`, which of course entities do).

In the future we may need to use a new syntax and do `{...DatePeriodRoleT, ...}`, since the Flow team plan to make objects exact by default: https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf

PR #850, opened 2019-01-25
URLCleanup eslint fixes: fix key sort for second level keys

reosarevok:
LINK_TYPES should definitely have keys sorted. If we don't want CLEANUP keys sorted because of them matching the way the code works or whatever I can undo that part, let me know


yvanzo:
Exactly, that’s sorted to match execution flow, see also the general comment to `CLEANUPS`:
https://github.com/metabrainz/musicbrainz-server/blob/032510ffadc376af9ce41752f83d132bc8cb5405/root/static/scripts/edit/URLCleanup.js#L302-L310

reosarevok:
Ok then, switched that back

PR #851, opened 2019-01-25
MBS-9936: Move from CJS style to ES6 (imports, exports) (V)

reosarevok:
Relationship editor


reosarevok:

reosarevok:

reosarevok:
I'm having an unrelated issue with tests I haven't been able to figure out (I get the following error even on master locally for some reason) and it means the test that is breaking here breaks for me even earlier, so I can't really run it locally. Any ideas on how to fix either my issue or the issue here? :) 

`Uncaught TypeError: Unable to process binding "template: function(){return {
                name:"template.edit-attribute",data:_.assign({ relationship:$parent.relationship()},$data)} }"
Message: $parent.relationship is not a function`

mwiencek:
This PR breaks some imports in root/static/scripts/tests/relationship-editor.js (that's the source of the cryptic "Super expression must either [...]" error).

If you didn't know, you can open root/static/scripts/tests/web.html in your browser after running `./script/compile_resources.sh web-tests`. Have the dev tools open while it's running and it'll pause on uncaught exceptions. Does this `$parent.relationship is not a function` error occur even for the web-tests?

reosarevok:
That does not happen there, so I fixed the error I saw there. Then I got two more. One of those I also get on master (https://github.com/metabrainz/musicbrainz-server/pull/859) and the other is https://github.com/metabrainz/musicbrainz-server/blob/2e01c49330310281abba4ce6897665ce1604f808/root/static/scripts/tests/autocomplete.js#L98

mwiencek:
Yeah, that `menu is hidden after blurring the autocomplete` one can happen when you run the tests via web.html. It has to do with the document not being focused, e.g. if you have the dev tools open and focused while the tests are running, it might fail. If you click anywhere on the document before the tests run (or before reloading the page), it should pass. This is only an issue when running them in the browser.

reosarevok:

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -328,7 +328,7 @@ const PART_OF_SERIES_LINK_TYPE_GIDS = _.values(PART_OF_SERIES_LINK_TYPES);
 
         linkTypeName() {
             var linkTypeID = this.relationship().linkTypeID();
-            return linkTypeID ? linkPhrase.clean(linkTypeID, this.backward()) : "";
+            return linkTypeID ? linkPhrase.stripAttributes(linkTypeID, this.backward()) : "";
```
mwiencek: This throws an exception: `linkPhrase.stripAttributes is not a function`

reosarevok: Oh, duh, yeah, this is changed as a separate part of this I haven't yet split and PRd. Will revert for now.

PR #852, opened 2019-01-27
MBS-9656: Change release quality → Change data quality

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9656 (small change but hopefully makes it much more clear)


reosarevok:
We could, but it seems a bit long for the sidebar?

mwiencek:
It's about the same length as "Release group external links," but I always thought that string was too long for the sidebar. So I'm fine with the shorter version here.

PR #857, opened 2019-01-28
 URLCleanup eslint fixes: key sort for LINK_TYPES + camelcasing

reosarevok:


root/static/scripts/edit/URLCleanup.js
```diff
@@ -281,7 +281,7 @@ const RESTRICTED_LINK_TYPES = _.reduce([
   return result.concat(_.values(linkType));
 }, []);
 
-function reencode_mediawiki_localpart(url) {
+function reencodeMediawikiLocalpart(url) {
```
yvanzo: Nitpicking the nitpicker: should be `reecondeMediawikiLocalPart` :stuck_out_tongue:

reosarevok: Nitpicking the nitpicker of the nitpicker, you mean `reencodeMediawikiLocalPart` ;) I can change it to that if desired, I just thought because it was in one word "localpart" was supposed to be one word for some reason!

PR #858, opened 2019-01-29
URLCleanup eslint fixes: fix key sort for CLEANUPS

reosarevok:
The diff is a mess, but basically: I just collapsed every object in CLEANUPS in my IDE, then put them in the right order (while collapsed so that they moved as a block).


reosarevok:
Wonder how that happened, but fixed now :)

PR #859, opened 2019-01-29
Relationship editor test: "guitars" has been renamed

reosarevok:
Was getting this failure even on master. Not entirely sure how this hasn't been an issue before?

### DIFF
--- a/root/static/scripts/tests/relationship-editor.js
+++ b/root/static/scripts/tests/relationship-editor.js
@@ -198,7 +198,7 @@ relationshipEditorTest("link phrase interpolation", function (t) {
             linkTypeID: 154,
             attributes: ids2attrs([1, 69, 75, 109, 302]),
             expected: "contains additional samples by",
-            expectedExtra: "strings, guitars, lyre, plucked string instruments"
+            expectedExtra: "strings, guitar family, lyre, plucked string instruments"
         },
         // MBS-6129
         {


mwiencek:
It hasn't been renamed in t/sql/initial.sql, which is where the test data comes from. I think this may be a caching issue.

Here's what should happen:

1. `./script/create_test_db.sh` will populate `musicbrainz_test` with data from t/sql/initial.sql. This SQL file [has "guitars"](https://github.com/metabrainz/musicbrainz-server/blob/0c0344240bc8e8f2f8e7f6d5198df273bd46d3e3/t/sql/initial.sql#L123).
2. `./script/compile_resources.sh web-tests` will call `./script/dump_js_type_info.pl`, which dumps the attribute data from `musicbrainz_test` into root/static/scripts/tests/typeInfo.js (it won't overwrite the file if it already exists). So this should also have "guitars" from step 1.

The cache issue can arise in `./script/dump_js_type_info.pl`: this calls out to `$c->model('LinkAttributeType')->get_tree` which uses the main Redis cache. So if you have "guitar family" in your main DB, this is likely where it's getting it from.

Running `redis-cli flushall && rm root/static/scripts/tests/typeInfo.js && ./script/compile_resources.sh web-tests` should confirm if that's the issue. We should fix dump_js_type_info to not use the cache.

mwiencek:
If you change "guitars" in t/sql/initial.sql too we can merge this. Though it might cause other tests to fail. Can fix dump_js_type_info to not use Redis in a separate PR.

PR #861, opened 2019-01-29
Fix QUALITY_NAMES to use N_l

mwiencek:
These should not be cached at the module-level, otherwise they won't change when the language changes. (`DEVELOPMENT_SERVER` must be disabled to observe this.)

I added a `SidebarDataQuality` component to hold code used by both the release and edit sidebars.

### DIFF
--- a/root/layout/components/sidebar/SidebarDataQuality.js
+++ b/root/layout/components/sidebar/SidebarDataQuality.js
@@ -1,0 +1,30 @@
+/*
+ * @flow
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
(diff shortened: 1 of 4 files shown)


root/layout/components/sidebar/SidebarDataQuality.js
```diff
@@ -0,0 +1,30 @@
...
+import {SidebarProperty} from './SidebarProperties';
+
+type Props = {|
+  +quality: QualityT,
+|};
+
+const SidebarDataQuality = ({quality}: Props) => {
+  const name = QUALITY_NAMES.get(quality);
+  return name ? (
+    <SidebarProperty className="data-quality" label={addColon(l('Data Quality'))}>
```
yvanzo: That changed message, for the best.

mwiencek: Yes, I saw we already had a `Data Quality` string, though I now see it's from the statistics domain. (Not sure if Transifex suggest translations from different domains.)

PR #862, opened 2019-01-29
Replace NO_MATCH symbol with null

mwiencek:
The `NO_MATCH` Symbol is returned by our expand2 parsers whenever they don't find a match for whatever they're trying to parse.

Flow doesn't support Symbols at all, so using a Symbol for this makes the code and associated types uglier.

This commit switches to just using `null` to indicate no match. Why wasn't `null` used to begin with? Well, I thought it might be useful if a parser could actually produce `null` as a match result, but that isn't used at all and I don't have a current use case in mind.

### DIFF
--- a/root/static/scripts/common/i18n/expand2.js
+++ b/root/static/scripts/common/i18n/expand2.js
@@ -12,8 +12,6 @@ import Raven from 'raven-js';
 import * as React from 'react';
 import ReactDOMServer from 'react-dom/server';
 
-const NO_MATCH = Symbol();
-
 const textContent = /^[^<>{}]+/;
 const varSubst = /^\{([0-9A-z_]+)\}/;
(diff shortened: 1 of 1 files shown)


mwiencek:
Never mind, closing this. Upon reading my commit message again I realized there's a case where it makes a difference:

```JavaScript
expand2('{foo}bar', {foo: null});
```

With this patch, this produces a parse error, because the match result of `{foo}` is `null`, which causes it to think it couldn't parse `{foo}` at all.

I also see there's a PR on the Flow repo to add basic Symbol support, so my original motivation probably won't be valid for long.

mwiencek:
Committed https://github.com/metabrainz/musicbrainz-server/commit/59433400ccdad2083d09588d355cd3cf15955c6e to add tests for that.

PR #863, opened 2019-01-29
Allow % replacements to be React nodes

mwiencek:
This is needed for instrument links in relationship link phrases. For example, in 'performed {instrument:%|instruments} on', we will want to substitute the '%' with a link to the actual instrument page, as our Perl-based expand code currently does.


root/static/scripts/common/i18n/expand2.js
```diff
@@ -172,11 +173,22 @@ function parseTextContent() {
   if (typeof text !== 'string') {
     return NO_MATCH;
   }
-  if (state.replacement !== NO_MATCH) {
-    /* flow-include if (state.replacement instanceof Symbol) throw 'impossible' */
-    text = text.replace(/%/g, he.encode(state.replacement));
```
mwiencek: Obviously `text.replace` can only replace text with text, so we need to change this to split on '%' and replace the '%' in the list with our replacement.

PR #864, opened 2019-01-29
 MBS-9909: Change reports to use React (5 - labels)

reosarevok:


root/report/DiscogsLinksWithMultipleLabels.js
```diff
@@ -0,0 +1,47 @@
...
+const DiscogsLinksWithMultipleLabels = ({
+  $c,
+  canBeFiltered,
+  filtered,
+  generated,
+  items,
+  pager,
+}: ReportDataT<ReportLabelURLT>) => (
+  <Layout fullWidth title={l('Discogs URLs linked to multiple labels')}>
+    <h1>{l('Discogs URLs linked to multiple labels')}</h1>
```
reosarevok: These two lines have been changed to lowercase them (if I wasn't reusing an old commit I would have commited them separately, but reverting to then re-change seemed a bit much).

PR #865, opened 2019-01-30
MBS-10005 (I): Use React Details and Tags tabs for area

reosarevok:

### DIFF
--- a/root/area/AreaLayout.js
+++ b/root/area/AreaLayout.js
@@ -13,6 +13,7 @@ import type {Node as ReactNode} from 'react';
 import Layout from '../layout';
 import AreaSidebar from '../layout/components/sidebar/AreaSidebar';
 import {hyphenateTitle, l} from '../static/scripts/common/i18n';
+import localizeAreaName from '../static/scripts/common/i18n/localizeAreaName';
 
 import AreaHeader from './AreaHeader';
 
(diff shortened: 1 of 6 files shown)


root/area/AreaLayout.js
```diff
@@ -0,0 +1,46 @@
...
+
+const AreaLayout = ({
+  children,
+  entity: area,
+  fullWidth,
+  page,
+  title,
+}: Props) => (
+  <Layout
+    title={title ? hyphenateTitle(l(area.name), title) : l(area.name)}
```
mwiencek: The TT file calls out to `area.l_name`, which isn't as simple: https://github.com/metabrainz/musicbrainz-server/blob/8ce0b73c6e6378542e39f0fac84b4f4ce8ca03ca/lib/MusicBrainz/Server/Entity/Area.pm#L22-L30

I suppose we should add a `localizeAreaName` utility to implement that logic. There's an `AREA_TYPE_COUNTRY` constant already in root/static/scripts/common/constants.js. Note that it should also use `l_countries`.

reosarevok:
I never claimed it addressed the whole of MBS-9914, but having the layout and two tabs converted is certainly part of MBS-9914 - I thought the entire point of sending small PRs was that we could review and convert stuff in parts. Yes, this only does these two tabs, but that's two less tabs that are using slower TT.

I can add a different ticket for this, but that would still be one ticket divided in n PRs (for each new layout creation, for reviewing them easily) so would that really be an improvement?

yvanzo:
Yes, because these PRs are likely to be merged at the same time and released within the same server update. The improvement is that users know what changed.

reosarevok:
Ok, I can make a ticket - should I also edit every commit name?

yvanzo:
Ideally yes, PRs too, thanks! :gift_heart:

PR #867, opened 2019-01-30
MBS-9926: Change subscriptions to use React (2 - collection pages)

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9926


root/components/SubHeader.js
```diff
@@ -7,9 +7,9 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-import React from 'react';
+import * as React from 'react';
 
-const SubHeader = ({subHeading}: {subHeading: string}) => (
+const SubHeader = ({subHeading}: {subHeading: string | React.Node}) => (
```
mwiencek: Nitpick, `string` is a subtype of `React.Node` so you can just put `React.Node` here. (There's the definition if you're curious: https://github.com/facebook/flow/blob/27630518e31ccecfac1466883a5736a70309b116/lib/react.js#L13)

root/collection/CollectionLayout.js
```diff
@@ -0,0 +1,51 @@
...
+  +$c: CatalystContextT,
+  +children: ReactNode,
+  +entity: CollectionT,
+  +fullWidth?: boolean,
+  +page: string,
+  +title?: string,
+|};
+
+const CollectionLayout = ({
+  $c,
```
mwiencek: It looks like `$c` isn't used in this component

root/collection/CollectionHeader.js
```diff
@@ -0,0 +1,102 @@
...
+const CollectionHeader = ({$c, collection, page}: Props) => {
+  const subHeading = (
+    <>
+      {collection.public ? (
+        l('Public collection by {owner}',
+          {owner: <EditorLink editor={collection.editor} />})
+      ) : l('Private collection')}
+      {collection.editor ? (
+        <span className="small">
+          {[
```
mwiencek: I was gonna ask why not use a fragment, but actually, why not use `bracketed` instead. :P

root/collection/CollectionHeader.js
```diff
@@ -0,0 +1,102 @@
...
+            ' (',
+            <a
+              href={'/user/' + collection.editor.name + '/collections'}
+              key={collection.editor.name}
```
mwiencek: The `key` can be removed too once it's no longer inside a list.

root/collection/CollectionHeader.js
```diff
@@ -0,0 +1,102 @@
...
+      {collection.public ? (
+        l('Public collection by {owner}',
+          {owner: <EditorLink editor={collection.editor} />})
+      ) : l('Private collection')}
+      {collection.editor ? (
+        <span className="small">
+          {[
+            ' (',
+            <a
+              href={'/user/' + collection.editor.name + '/collections'}
```
mwiencek:
```suggestion
              href={'/user/' + encodeURIComponent(collection.editor.name) + '/collections'}
```

root/collection/CollectionHeader.js
reosarevok: Changed. I've been wondering - we have many cases of having to add {' '} before a parens or whatever - should we have a spaceBefore() function instead?

mwiencek: Dunno, `{spaceBefore(...)}` doesn't seem like a real improvement over `{' '}{...}`, but that's just me. :)

PR #869, opened 2019-01-30
MBS-10005 (IV): Use React Details and Tags tabs for instrument

reosarevok:


root/static/scripts/common/i18n/localizeInstrumentName.js
```diff
@@ -0,0 +1,19 @@
...
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import {l_instruments, lp_instruments} from './instruments';
+
+function localizeInstrumentName(instrument: InstrumentT) {
+  if (instrument.comment) {
+    return lp_instruments(instrument.name, instrument.comment);
```
yvanzo: I cannot understand what it does or even what it is intended to do.

reosarevok: It is used for translators to be able to distinguish between different instruments of the same name (in English) that might have different names in their language.

yvanzo: So you actually fixed a bug.

PR #871, opened 2019-01-30
MBS-10003: Generalize entity Aliases tabs to one file

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10003


reosarevok:
@yvanzo mentioned that artists have more content on their page - which is true. When we reach that point we could either have a separate template for that or add a conditional to the shared file.

root/entity/Aliases.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+import chooseLayoutComponent from '../utility/chooseLayoutComponent';
+import AliasesComponent from '../components/Aliases';
```
yvanzo: `AliasesComponent` doesn’t exist

reosarevok: AliasesComponents is an alias for the default export of components/Aliases because this one is also Aliases.

mwiencek: Well, the only other way is to do `import {default as AliasesComponent}` but I'm not sure that's an improvement.

PR #873, opened 2019-01-30
MBS-10005 (VI): Use React Details and Tags tabs for recording

reosarevok:


root/recording/RecordingLayout.js
```diff
@@ -0,0 +1,61 @@
...
+  +children: ReactNode,
+  +entity: RecordingT,
+  +fullWidth?: boolean,
+  +page: string,
+  +title?: string,
+|};
+
+const RecordingLayout = ({
+  children,
+  entity: recording,
```
yvanzo: Just noticed this, used in other PRs too. Is it a good practice? Should we generalize using it?

reosarevok: It was a suggestion from @mwiencek when we decided to generalize... something, I forget what exactly :) I think it makes it more clear than using entity everywhere, but dunno :)

PR #874, opened 2019-01-30
MBS-10005 (VII): Use React Details and Tags tabs for release

reosarevok:


reosarevok:
This has (known) flow errors because release.releaseGroup is apparently optional. It should always exist where the header is loaded, but should I still do "release.releaseGroup ? whatever"?

PR #875, opened 2019-01-31
MBS-9923 (I): Convert URL Layout to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9923

Index probably can't be converted until the relationship components are merged, meaning this Layout file is currently unused, but it shouldn't hurt to have it already. Once the relationships component is merged it should be very easy to finish MBS-9923

UrlHeader renamed for consistency on @mwiencek's suggestion.

### DIFF
--- a/root/url/UrlHeader.js
+++ b/root/url/UrlHeader.js
@@ -18,7 +18,7 @@ type Props = {|
   url: UrlT,
 |};
 
-const URLHeader = ({url, page}: Props) => (
+const UrlHeader = ({url, page}: Props) => (
   <EntityHeader
     entity={url}
     headerClass="workheader"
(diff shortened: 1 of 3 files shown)


root/url/UrlHeader.js
```diff
@@ -18,7 +18,7 @@ type Props = {|
   url: UrlT,
 |};
 
-const URLHeader = ({url, page}: Props) => (
+const UrlHeader = ({url, page}: Props) => (
```
yvanzo: Should ideally be renamed in a separate commit. `root/url/layout.tt` be updated too.

reosarevok: Changed

yvanzo:
Oh OK, so this is dead code... Probably not up for merge right now.

PR #876, opened 2019-01-31
MBS-10005 (VIII): Use React Details and Tags tabs for work

reosarevok:

### DIFF
--- a/root/work/WorkLayout.js
+++ b/root/work/WorkLayout.js
@@ -1,0 +1,53 @@
+/*
+ * @flow
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+import type {Node as ReactNode} from 'react';
+
+import Layout from '../layout';
+import WorkSidebar from '../layout/components/sidebar/WorkSidebar';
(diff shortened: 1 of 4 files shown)


root/work/WorkLayout.js
```diff
@@ -0,0 +1,52 @@
...
+
+const WorkLayout = ({
+  children,
+  entity: work,
+  fullWidth,
+  page,
+  title,
+}: Props) => {
+  const mainTitle = l('{type} “{work}”', {
+    type: work.typeName ? l(work.typeName) : l('Work'),
```
mwiencek:
```suggestion
    type: work.typeName ? lp(work.typeName, 'work_type') : l('Work'),
```

reosarevok: When to choose lp vs lp_attributes? (both seem to include the work type name for some reason, since lp works here and lp_attributes works on the sidebar/header?)

mwiencek: Crap, sorry. I messed up, it should be `lp_attributes`. Does `lp` actually work? I tried in the node REPL and it didn't:

```bash
$ babel-node
> const {lp} = require('./root/static/scripts/common/i18n');
undefined
> const gettext = require('./root/server/gettext');
undefined
> gettext.setLocale('de');
undefined
> lp('Song', 'work_type');
'Song'
> const {lp_attributes} = require('./root/static/scripts/common/i18n/attributes');
undefined
> lp_attributes('Song', 'work_type');
'Lied'
```

reosarevok: I tried and it seemed to work but I guess I just checked the wrong place or something, lol. Fixing.

PR #877, opened 2019-01-31
URLCleanup eslint fixes (long lines, eq) + regex simplification

reosarevok:


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1470,7 +1475,7 @@ const CLEANUPS = {
       new RegExp('^(https?://)?(www\\.)?metal-archives\\.com/(bands?|albums|artists|labels)', 'i'),
       new RegExp('^(https?://)?(www\\.)?spirit-of-metal\\.com/', 'i'),
       new RegExp('^(https?://)?(www\\.)?ibdb\\.com/', 'i'),
-      new RegExp('^(https?://)?(www\\.)?lortel.\\org/', 'i'),
+      new RegExp('^(https?://)?(www\\.)?lortel\\.org/', 'i'),
```
reosarevok: This is the only thing here that should be a bug fix rather than a readability/speed improvement, AFAICT

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1290,7 +1295,7 @@ const CLEANUPS = {
       return url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?lyric\.evesta\.jp\/([al]\w+\.html).*$/, 'http://lyric.evesta.jp/$1');
     },
     validate: function (url, id) {
-      const m = /^http:\/\/lyric\.evesta\.jp\/(a|l)\w+\.html$/.exec(url);
+      const m = /^http:\/\/lyric\.evesta\.jp\/[al]\w+\.html$/.exec(url);
       if (m) {
         const prefix = m[1];
```
mwiencek: Won't this break, since `(a|l)` was `m[1]` here? (A test would be a good idea, if so.)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -536,7 +541,7 @@ const CLEANUPS = {
       return url;
     },
     validate: function (url, id) {
-      const m = /^https:\/\/www.bandsintown\.com\/(?:(a|e|v)\/)?([^\/?#]+)$/.exec(url);
+      const m = /^https:\/\/www.bandsintown\.com\/(?:[aev]\/)?([^\/?#]+)$/.exec(url);
```
mwiencek: This looks like it might break something too.

yvanzo: Because this removes a capturing group, thus modifies match results `m[X]`, see https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/RegExp#grouping-back-references

yvanzo: This is most probably what caused test failure at https://circleci.com/gh/metabrainz/musicbrainz-server/87. Note: you can run these tests locally with the command:
```shell
node root/static/scripts/tests/Control/URLCleanup.js
```

reosarevok: Oooh, that's what m[x] meant. Sorry, thought it would be match 1 or something :) ok, will check.

root/static/scripts/edit/URLCleanup.js
reosarevok: Got it now, should be fixed.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1245,7 +1245,7 @@ const CLEANUPS = {
     type: LINK_TYPES.license,
     clean: function (url) {
       url = url.replace(/^(https?:\/\/)?([^\/]+\.)?creativecommons\.org\//, 'https://creativecommons.org/');
-      url = url.replace(/^https:\/\/creativecommons\.org\/(licenses|publicdomain)\/(.+)\/((legalcode|deed)((\.|-)[A-Za-z_]+)?)?/, 'https://creativecommons.org/$1/$2/');
+      url = url.replace(/^https:\/\/creativecommons\.org\/(licenses|publicdomain)\/(.+)\/((legalcode|deed)([.-][A-Za-z_]+)?)?/, 'https://creativecommons.org/$1/$2/');
```
yvanzo: Incidentally removed capturing group too.

reosarevok:
The failure here *now* seems unrelated, I imagine?

root/static/scripts/edit/URLCleanup.js
yvanzo: Adding adhoc test would be neat still.

root/static/scripts/edit/URLCleanup.js
yvanzo: Oh, actually, this one was not used, thus it is fine. The three latest groups could probably even be turned into non-capturing groups with `?:` as in:

```javascript
       url = url.replace(/^https:\/\/creativecommons\.org\/(licenses|publicdomain)\/(.+)\/(?:(?:legalcode|deed)(?:[.-][A-Za-z_]+)?)?/, 'https://creativecommons.org/$1/$2/');
```

PR #879, opened 2019-01-31
MBS-10007: Convert Add Annotation edit to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10007


root/edit/details/AddAnnotation.js
```diff
@@ -0,0 +1,67 @@
...
+import formatEntityTypeName from '../../static/scripts/common/utility/formatEntityTypeName';
+
+const AddAnnotation = ({edit}: {edit: EditT}) => (
+  <table
+    className={`details add-${edit.display_data.entity_type}-annotation`}
+  >
+    <table
+      className={`details add-${edit.display_data.entity_type}-annotation`}
+    >
+      {edit.display_data[edit.display_data.entity_type] || !edit.preview ? (
```
reosarevok: edit.preview is a problem with Flow, but tbh I'm not sure where this is even coming from, so I'm not sure what type to add :/

mwiencek: It comes from `Edit::Role::Preview`, thought not all edit types consume this role. So in `Edit::TO_JSON` you'll need a conditional check:

```Perl
my $can_preview = $self->does('MusicBrainz::Server::Edit::Role::Preview');
return {
    # ...
    $can_preview ? (preview => boolean_to_json($self->preview)) : (),
    # ...
};
```

And the Flow type should be `preview?: boolean`.

lib/MusicBrainz/Server/Edit/Annotation/Edit.pm
```diff
@@ -29,6 +30,7 @@ role {
         isa => Dict[
             editor_id => Int,
             text      => Nullable[Str],
+            html      => Nullable[Str],
```
mwiencek: We shouldn't need to store the `html` in the database, from what I can see - you're already calling `format_wikitext($self->data->{text})` (so, using `text`) for the display data generation as it's needed.

root/types.js
```diff
@@ -416,10 +416,12 @@ declare type EditT = {|
   |},
   +created_time: string,
   +data: Object,
+  +display_data: Object,
```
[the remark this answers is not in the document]
reosarevok: Hmm. This specific one has $data->{$entity_type} = $loaded->{$model}->{$self->$entity_id} so it has work: WorkT if it's a work, artist: ArtistT for an artist, etc. How can we indicate this?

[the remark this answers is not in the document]
mwiencek: Good question...there's an `AnnotatedEntityT` in root/static/scripts/common/components/Annotation.js so maybe we can move that to root/types.js and make use of it.

This might work. Technically it would allow things like `artist: WorkT` but it doesn't really matter since it's a generic component and the props come from the server anyway (meaning, if the server screws up the data, better Flow types don't help us).

```JavaScript
type AnnotatedEntityTypeT = $ElementType<AnnotatedEntityT, 'entityType'>;

// ...
+display_data: {|
  +entity_type: AnnotatedEntityTypeT,
  [AnnotatedEntityTypeT]: AnnotatedEntityT,
|},
```

lib/MusicBrainz/Server/Edit.pm
```diff
@@ -336,6 +338,7 @@ sub TO_JSON {
             auto_edit => boolean_to_json($conditions->{auto_edit}),
         },
         created_time => datetime_to_iso8601($self->created_time),
+        display_data => $self->display_data,
```
mwiencek: This broke the display of some Edit::Relationship::Delete edits (which our Selenium tests caught), so I fixed that in 1db864e1cc1dbca11d5e6d588186fe9e9a906c8c, but we'll need to test this against more edit types to be sure (especially historical ones).

root/types.js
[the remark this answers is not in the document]
reosarevok: https://github.com/metabrainz/musicbrainz-server/pull/881

PR #880, opened 2019-02-02
CircleCI

mwiencek:


mwiencek:
All checks completed in 7 minutes 33 seconds.  😱

CircleCI is awesome! Being able to ssh into a running test container after it fails is an incredible feature.

docker/Dockerfile.tests
```diff
@@ -0,0 +1,135 @@
+FROM phusion/baseimage:0.11
```
mwiencek: This image was built locally and pushed to docker hub as `metabrainz/musicbrainz-tests:v-2019-02`.

It's intended only as a base image: we only have to build it (and push a new version) when we want to update software that the tests use, like node, chrome, postgres, etc. It doesn't include any code in the image; all code is checked out/built/cached on CircleCI.

.circleci/config.yml
```diff
@@ -0,0 +1,193 @@
...
+      - js:
+          requires:
+            - build
+      - perl-and-pgtap:
+          requires:
+            - build
+      - selenium:
+          requires:
+            - build
+      - build
```
mwiencek: These jobs all run in separate containers, but have a shared "workspace." When the `build` job finishes, it calls `persist_to_workspace` to upload the build result, and other jobs use `attach_workspace` to download a cached copy of it. This is surprisingly quite fast.

.circleci/config.yml
```diff
@@ -0,0 +1,193 @@
...
+      - run: sudo -E -H -u musicbrainz yarn
+      - save_cache:
+          key: v2-node-{{ checksum "yarn.lock" }}
+          paths:
+            - "node_modules"
+
+      - restore_cache:
+          keys:
+            - v3-carton-{{ checksum "cpanfile.snapshot" }}
+            - v3-carton-
```
mwiencek: Another critical feature here to speed up our builds is caching the build data using [save_cache](https://circleci.com/docs/2.0/configuration-reference/#save_cache) and [restore_cache](https://circleci.com/docs/2.0/configuration-reference/#restore_cache). Here it uses the checksum of `cpanfile.snapshot` as the cache key, but also accepts any partial match on `v3-carton-` as a fallback. (`carton` would then fix any missing or outdated modules.)

docker/Dockerfile.tests
mwiencek: Also, having postgres, redis, and musicbrainz-server in the same container vastly simplified the setup and configuration. Obviously we wouldn't do that in production, but I think it's perfectly fine for the tests.

.circleci/config.yml
mwiencek: That we can push a new `musicbrainz-tests` docker image without breaking our build caches is very nice, too. But it's important to be aware that this can cause issues in rare cases where we update a library in the image while our build cache references the old version. The `v3-` prefix should be incremented whenever we update build libs in that way.

PR #882, opened 2019-02-04
MBS-10009: Fix untranslated sidebar types

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10009

### DIFF
--- a/root/layout/components/sidebar/ReleaseSidebar.js
+++ b/root/layout/components/sidebar/ReleaseSidebar.js
@@ -61,7 +61,7 @@ const ReleaseSidebar = ({$c, release}: Props) => {
     ? new URL(releaseCoverUrl).host
     : null;
   const barcode = formatBarcode(release.barcode);
-  const typeName = releaseGroup.typeName;
+  const typeName = releaseGroup.l_type_name;
   const language = release.languageID
     ? $c.linked_entities.language[release.languageID]
     : null;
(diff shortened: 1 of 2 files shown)


root/layout/components/sidebar/ReleaseSidebar.js
```diff
@@ -61,7 +61,7 @@ const ReleaseSidebar = ({$c, release}: Props) => {
     ? new URL(releaseCoverUrl).host
     : null;
   const barcode = formatBarcode(release.barcode);
-  const typeName = releaseGroup.typeName;
+  const typeName = releaseGroup.l_type_name;
```
reosarevok: This is a combined type name, which we don't translate. Elsewhere, we're just making the server serve an l_type_name - I think that's the most reasonable option here too.

root/layout/components/sidebar/SidebarType.js
```diff
@@ -24,7 +24,10 @@ type Props = {|
 const SidebarType = ({$c, entity, typeType}: Props) => (
   entity.typeID ? (
     <SidebarProperty className="type" label={l('Type:')}>
-      {l_attributes($c.linked_entities[typeType][entity.typeID].name)}
+      {lp_attributes(
+        $c.linked_entities[typeType][entity.typeID].name,
+        typeType,
+      )}
```
reosarevok: Kept it as similar as possible to the original, but wouldn't it make sense to just change from typeID to typeName and use entity.typeName instead of looking into CatContext?

mwiencek: I believe it saves a bit of memory to avoid duplicating the types on pages with tons of entities, though I haven't measured it to be honest. The main issue is that `$c.linked_entities[typeType][entity.typeID].name` is ugly and verbose. There are ways I think we can hide that, like defining a `localizeTypeName($c, entity)` utility.

PR #885, opened 2019-02-04
Add ProhibitInterpolationOfLiterals to Critic

reosarevok:
https://docs.activestate.com/activeperl/5.26/perl/lib/Perl/Critic/Policy/ValuesAndExpressions/ProhibitInterpolationOfLiterals.html seems like a sensible rule to try to follow, so @yvanzo and I were thinking it might make sense to add it to Critic :)


mwiencek:
Agreed, I already follow this (I generally don't like magic things happening with strings unless I request it).

PR #886, opened 2019-02-04
MBS-10011: Congratulate/thank users on their birthdays

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10011

Just a nice touch.

This is currently only implemented in React. Given only a bunch of pages use React layout at the moment, it might make sense to add it to layout.tt too, but given it's not at all a critical feature, I'm not sure it's worth the effort. You tell me! :)


root/layout/index.js
```diff
@@ -88,6 +89,28 @@ const Layout = ({$c, ...props}) => (
...
 
+      {$c.user.birth_date &&
+      moment($c.user.birth_date).dayOfYear() === moment().dayOfYear() &&
+      !getRequestCookie($c.req, 'birthday_message_dismissed_mtime') &&
+      <div className="banner birthday-message">
+        <p>
+          <span aria-label="cakes" role="img">
+            {String.fromCodePoint(0x1F382)}
+            {String.fromCodePoint(0x1F382)}
+            {String.fromCodePoint(0x1F382)}
```
reosarevok: Those are birthday cake emojis.

root/layout/index.js
```diff
@@ -88,6 +89,28 @@ const Layout = ({$c, ...props}) => (
           </p>
         </div>}
 
+      {$c.user.birth_date &&
+      moment($c.user.birth_date).dayOfYear() === moment().dayOfYear() &&
```
yvanzo: This does not work equally with leap years and even years.

root/layout/index.js
```diff
@@ -88,6 +89,28 @@ const Layout = ({$c, ...props}) => (
...
+      !getRequestCookie($c.req, 'birthday_message_dismissed_mtime') &&
+      <div className="banner birthday-message">
+        <p>
+          <span aria-label="cakes" role="img">
+            {String.fromCodePoint(0x1F382)}
+            {String.fromCodePoint(0x1F382)}
+            {String.fromCodePoint(0x1F382)}
+          </span>
+          {' '}
+          {l('Happy birthday, and thanks for contributing to MusicBrainz!')}
```
yvanzo: Not everyone appreciates automated birthday message, I would suggest an opt-in instead. Plus, birthday is quite unrelated to contributing to MusicBrainz, user registration anniversary or day after the 10..0th edit would probably make more sense.

mwiencek: Heh. Now I'm not sure which is stranger: someone upset because we didn't wish them a happy birthday, or someone upset because we did?

I'd really prefer not adding an option for this, and just deciding whether to use their real birthday or account birthday. I see no problem with either.

root/layout/index.js
```diff
@@ -88,6 +89,28 @@ const Layout = ({$c, ...props}) => (
           </p>
         </div>}
 
+      {$c.user.birth_date &&
```
mwiencek: The test failures aren't exactly unexpected: this throws an exception because `$c.user` can be undefined.

root/layout/index.js
reosarevok: I specifically did a banner, and not an email or something, because I fully understand people don't like to get pointless emails. But I feel a small banner makes sense. I would *also* add one for the account birthday, since thanking people twice a year is nicer than thanking them once! :P

root/layout/index.js
```diff
@@ -88,6 +89,30 @@ const Layout = ({$c, ...props}) => (
           </p>
         </div>}
 
+      {$c.user &&
+      $c.user.birth_date &&
+      moment($c.user.birth_date).month() === moment().month() &&
+      moment($c.user.birth_date).date() === moment().date() &&
+      !getRequestCookie($c.req, 'birthday_message_dismissed_mtime') &&
```
reosarevok: How often do our cookies expire? I'm assuming this would have expired before their next birthday, but I might be wrong.

yvanzo: This one should probably expire after 24h?

reosarevok: Should. Made changes for that.

reosarevok:
I mentioned that on the original comment :) 

Re: client side, dunno? I just did it like other banners but if we think it's a problem and you can give me a hint how to change it to client side, I can look into it.

yvanzo:
By hydrating a `BirthdayBanner` component that set its state via `componentDidMount` method. See `CommonsImage` for example. But the cure might be worse than the disease here... :pill:

root/layout/index.js
```diff
@@ -88,6 +89,30 @@ const Layout = ({$c, ...props}) => (
...
+      <div className="banner birthday-message">
+        <p>
+          <span aria-label="cakes" role="img">
```
mwiencek: Cool! But if the label is going to be read by screen readers, we should probably translate that and be slightly more descriptive, right?
```suggestion
          <span aria-label={l('Birthday cakes')} role="img">
```

root/layout/index.js
```diff
@@ -88,6 +89,30 @@ const Layout = ({$c, ...props}) => (
           </p>
         </div>}
 
+      {$c.user &&
+      $c.user.birth_date &&
```
mwiencek: Nitpick, these aren't on the same indentation level as the first condition.

mwiencek:
It should only have a minimal effect on page load, since the React elements aren't even created if the date doesn't match. The slowest part is likely creating the moment date objects.

Doing everything on the client would require us to serialize the user's birthdate into the markup of every page, from what I can tell, which isn't a good idea. Unless, you perform the check of whether to include the client JS on the server - but then you're doing the likely-slowest-part on the server again.

root/layout/index.js
reosarevok: I'm never sure what indentation these should have. Eslint seems to like all of them. Changed it, better now?

root/layout/index.js
reosarevok: Did that, and moved it into a separate component instead of repeating it twice.

root/layout/index.js
mwiencek: Yeah, that's a bit better. These are a bit annoying so I usually try to move long conditions into a separate variable outside the JSX.

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -293,6 +293,7 @@ around TO_JSON => sub {
     return {
         %{$self->$orig},
         biography               => $self->biography,
+        birth_date              => datetime_to_iso8601($self->birth_date),
```
mwiencek: If we just do `$self->birth_date->TO_JSON` here, we'll get a `{year, month, day}` object, so we won't have to parse anything in the JS.

root/layout/index.js
```diff
@@ -88,6 +97,22 @@ const Layout = ({$c, ...props}) => (
           </p>
         </div>}
 
+      {$c.user &&
+       $c.user.birth_date &&
+       moment($c.user.birth_date).month() === moment().month() &&
```
mwiencek: If you change the `birth_date` serialization above, we can do `$c.user.birth_date.month === new Date().getMonth() + 1` here (+ 1 because the months are 0-indexed), and `$c.user.birth_date.day === new Date().getDate()` (no + 1 because the day is 1-indexed, because JavaScript), which is much faster than using moment.

mwiencek:
I rebased and changed `showBirthdayBanner` slightly to only use one Date object, in case the day/month changes between CPU cycles, lol.

PR #888, opened 2019-02-05
MBS-9936: Move from CJS style to ES6 (imports, exports) (VI)

reosarevok:
scripts/edit

I would love to eslint --fix all of these too, but let's do one thing at a time so it's reviewable.


root/static/scripts/edit/MB/Control/Area.js
```diff
@@ -29,4 +29,4 @@ MB.Control.Area = function () {
     });
 };
 
-module.exports = MB.Control;
+export default MB.Control;
```
mwiencek: There's an import in root/static/scripts/place.js which this breaks.

Perhaps replace this line with something like `export const initializeArea = MB.Control.Area;` and change the import to `import {initializeArea} from './edit/MB/Control/Area';`.

root/static/scripts/edit/MB/Control/Bubble.js
```diff
@@ -332,4 +332,4 @@ MB.Control.initializeBubble = function (bubble, control, vm, canBeShown) {
     return bubbleDoc;
 };
 
-module.exports = MB.Control;
+export default MB.Control;
```
mwiencek: This breaks an import in root/static/scripts/place.js. I suggest `export const initializeBubble = MB.Control.initializeBubble;` and no default export.

root/static/scripts/work.js
```diff
@@ -19,8 +19,8 @@ const {lp_attributes} = require('./common/i18n/attributes');
 const MB = require('./common/MB');
 const scriptArgs = require('./common/utility/getScriptArgs')();
 const {Lens, prop, index, set, compose3} = require('./common/utility/lens');
-const {buildOptionsTree} = require('./edit/forms');
-const {initializeBubble} = require('./edit/MB/Control/Bubble');
+import forms from './edit/forms';
+import bubble from './edit/MB/Control/Bubble';
```
mwiencek: Per comment above,
```suggestion
import {initializeBubble} from './edit/MB/Control/Bubble';
```

root/static/scripts/edit/forms.js
```diff
@@ -279,4 +279,4 @@ ko.bindingHandlers.withLabel = {
     }
 };
 
-module.exports = MB.forms;
+export default MB.forms;
```
mwiencek: I'd prefer no default export here too, and adding named exports for `buildOptionsTree`, `linkTypeOptions`, etc., which is how this module was previously consumed.

`MB.forms` is a bogus namespace from before we had actual modules - given that this module holds several utility functions and doesn't have/need a primary export, it makes much more sense to use named exports.

root/static/scripts/work.js
```diff
@@ -19,8 +19,8 @@ const {lp_attributes} = require('./common/i18n/attributes');
 const MB = require('./common/MB');
 const scriptArgs = require('./common/utility/getScriptArgs')();
 const {Lens, prop, index, set, compose3} = require('./common/utility/lens');
-const {buildOptionsTree} = require('./edit/forms');
-const {initializeBubble} = require('./edit/MB/Control/Bubble');
+import forms from './edit/forms';
```
mwiencek: Per comment above,
```suggestion
import {buildOptionsTree} from './edit/forms';
```

root/static/scripts/edit/forms.js
reosarevok: Do I just *remove* MB.forms completely then, leaving only separate functions?

mwiencek: It's used in root/components/relationship-editor.tt by knockout so I wouldn't remove the reference, just the export for it.

root/static/scripts/place.js
```diff
@@ -10,7 +10,7 @@ const _ = require('lodash');
 
 const isBlank = require('./common/utility/isBlank');
 const initializeDuplicateChecker = require('./edit/check-duplicates');
```
mwiencek:
```suggestion
const initializeDuplicateChecker = require('./edit/check-duplicates').default;
```
There's a similar issue in root/static/scripts/series.js

PR #889, opened 2019-02-06
MBS-10014: Convert Add Area edit to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10014


root/edit/details/AddArea.js
```diff
@@ -0,0 +1,123 @@
...
+          <th>{addColon(l('Name'))}</th>
+          <td>{display.name}</td>
+        </tr>
+
+        {display.sort_name ? (
+          <tr>
+            <th>{addColon(l('Sort name'))}</th>
+            <td>{display.sort_name}</td>
+          </tr>
+        ) : null}
```
reosarevok: Is there any reason we have this anyway? AFAICT areas never have a sort name?

root/edit/details/AddArea.js
```diff
@@ -0,0 +1,122 @@
...
+import commaOnlyList from '../../static/scripts/common/i18n/commaOnlyList';
+import formatDate from '../../static/scripts/common/utility/formatDate';
+import isDateEmpty from '../../static/scripts/common/utility/isDateEmpty';
+import DescriptiveLink from '../../static/scripts/common/components/DescriptiveLink';
+import yesNo from '../../static/scripts/common/utility/yesNo';
+
+type AddAreaEditT = {|
+  ...EditT,
+  +display_data: {|
+    ...CommentRoleT,
```
mwiencek: Nitpick, though it wouldn't have an effect on the current code, `CommentRoleT` has a slightly different type. The type here should be `+comment: string | null`.

reosarevok: Changed. Ideally at some point we might try to have more consistent types (and actual serializations, obvs) for stuff that should really be the same though :/

PR #890, opened 2019-02-06
MBS-10015: Convert Add Artist edit to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10015


root/edit/details/AddArtist.js
```diff
@@ -0,0 +1,168 @@
...
+            <th>{addColon(l('Type'))}</th>
+            <td>{lp_attributes(display.type.name, 'artist_type')}</td>
+          </tr>
+        ) : null}
+
+        {display.gender && display.gender.name ? (
+          <tr>
+            <th>{addColon(l('Gender'))}</th>
+            {/* $FlowFixMe */}
+            <td>{lp_attributes(display.gender.name, 'gender')}</td>
```
reosarevok: Flow refuses to understand if this is inside && display.gender.name then it doesn't matter there's no .name in gender = null.

Although actually, gender is not really null, it's either GenderT or `''`. Not sure if that should be changed or this should be changed or we should leave it be since it works anyway.

t/lib/t/MusicBrainz/Server/Edit/Artist/Create.pm
```diff
@@ -68,7 +69,7 @@ SQL
     is($edit->display_data->{gender}->{name}, 'Male', "Gender is correct in display data.");
     is($edit->display_data->{comment}, 'Canadian electronica duo', "Comment is correct in display data.");
     is($edit->display_data->{begin_date}->format, "1981-05", "Begin date is correct in display data." );
-    is($edit->display_data->{ended}, 1, "Has 'ended' set in display data.");
+    is($edit->display_data->{ended}, boolean_to_json(1), "Has 'ended' set in display data.");
```
reosarevok: I wasn't sure what to compare with here, so I decided we might as well just compare by using boolean_to_json here as well. But we might want to put exactly the expected output of boolean_to_json here in case boolean_to_json itself breaks... which would be what exactly? :)

t/lib/t/MusicBrainz/Server/Controller/Edit/Show.pm
```diff
@@ -12,7 +12,7 @@ my $hard_mock_edit_class = $mock_edit_class + 1;
     package t::Controller::Edit::Show::MockEdit;
     use Moose;
     extends 'MusicBrainz::Server::Edit';
-    sub edit_name { 'Add artist' } # Just so we use an edit template
+    sub edit_name { 'Edit artist' } # Just so we use an edit template
```
reosarevok: At some point we'll need to change this further, but for now I'm changing it to a template with diffs since I know I won't be changing those soon.

root/edit/details/AddArtist.js
mwiencek: Set `const gender = display.gender;` above and reference `gender` here, then the error should go away. Flow is pretty conservative with refinements, so if it sees a function call like `addColon` that it can't prove doesn't modify `display`, it'll invalidate the refinement. Saving it to the variable works around that, since it can't be modified.

I'd greatly prefer accurate types even if we don't think it makes a difference, since I trust the computer more than I to know if it does. :)

t/lib/t/MusicBrainz/Server/Edit/Artist/Create.pm
mwiencek: This seems fine but to answer your question, `JSON::true` or `\1` both work. :)

root/edit/details/AddArtist.js
```diff
@@ -0,0 +1,169 @@
...
+  +display_data: {|
+    ...CommentRoleT,
+    ...DatePeriodRoleT,
+    +area: AreaT | null,
+    +artist: AreaT,
+    +begin_area: AreaT | null,
+    +end_area: AreaT | null,
+    +gender: GenderT | null,
+    +ipi_codes: $ReadOnlyArray<string>,
+    +isni_codes: $ReadOnlyArray<string>,
```
mwiencek: `ipi_codes` and `isni_codes` can both be `null` (old edits currently break due to the `.length` checks below).

PR #892, opened 2019-02-06
MBS-10017: Convert artist/cannot_split.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10017

### DIFF
--- a/root/artist/CannotSplit.js
+++ b/root/artist/CannotSplit.js
@@ -1,0 +1,28 @@
+/*
+ * @flow
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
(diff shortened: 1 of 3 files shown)


root/artist/CannotSplit.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import React from 'react';
+
+import {l} from '../static/scripts/common/i18n';
+
+import ArtistLayout from './ArtistLayout';
+
+const CannotSplit = ({artist}: {artist: ArtistT}) => (
+  <ArtistLayout entity={artist} page="cannot_split">
```
reosarevok: ArtistLayout requires page. We could just make that optional instead, if preferred (tt didn't pass a page here)

reosarevok: In fact, it goes all the way up to EntityHeader requiring page. Could still be changed, but there might be a specific reason this was done that way :)

mwiencek: Looks like it could technically be made optional, though it doesn't hurt much to specify it either.

PR #893, opened 2019-02-06
MBS-10018: Convert artist/special_purpose.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10018

The current implementation does not display any info about the artist being (not) edited. I think that's confusing and it's much better to have the header present (both to show the user *which* artist they're trying to edit, and to allow them more navigation options away from the error) so changed this from the original layout.tt to use ArtistLayout.

### DIFF
--- a/root/artist/SpecialPurpose.js
+++ b/root/artist/SpecialPurpose.js
@@ -1,0 +1,31 @@
+/*
+ * @flow
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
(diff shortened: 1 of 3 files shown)


root/artist/SpecialPurpose.js
```diff
@@ -0,0 +1,31 @@
...
+
+import {l} from '../static/scripts/common/i18n';
+
+import ArtistLayout from './ArtistLayout';
+
+const SpecialPurpose = ({artist}: {artist: ArtistT}) => (
+  <ArtistLayout
+    entity={artist}
+    fullWidth
+    page="special_purpose"
```
reosarevok: As with https://github.com/metabrainz/musicbrainz-server/pull/892 - we might decide we don't want to send any page here.

PR #894, opened 2019-02-07
MBS-10019: Convert the artist and place event pages to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10019


root/components/Paginator.js
```diff
@@ -27,7 +27,7 @@ const Paginator = ({
   pager,
   pageVar = 'page',
 }: Props) => {
-  if (pager.last_page <= 1) {
+  if (pager.last_page <= 1 || !pager.last_page) {
```
reosarevok: The pager JSON from these pages does *not* include last_page at all, so the paginator was always appearing regardless of whether there were only a couple events.

If this is unexpected, then we should instead try and figure out why last_page isn't always being added to the JSON.

root/components/events-list.tt
```diff
@@ -1,100 +0,0 @@
...
-      [%~ IF c.user_exists AND checkboxes ~%]
-        <th style="width: 1em">
-          <input type="checkbox" />
-        </th>
-      [%~ END ~%]
-
-      [%~ IF series_item_numbers ~%]
-        <th style="width: 1em">[%~ l('#') ~%]</th>
-      [%~ END ~%]
-      <th>[%~ sortable ? sortable_table_header('name', l('Event')) : l('Event') ~%]</th>
```
reosarevok: This component accepted a "sortable" boolean. This was (unless I've missed some use I haven't touched at all, which I don't think so) completely unused, so I didn't implement it on the React component. We might eventually want these tables to be sortable, but then I don't see any reason why this would be done with a preference anyway - we'd want *all* tables to be sortable, I'd expect.

root/static/scripts/common/components/EventLocations.js
```diff
@@ -0,0 +1,29 @@
...
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import DescriptiveLink from './DescriptiveLink';
+
+const EventLocations = ({event}: {event: EventT}) => (
```
reosarevok: Split this out because the same code would be copied in three different places otherwise.

root/series/index.tt
```diff
@@ -15,8 +15,7 @@
       IF series.type.item_entity_type == 'event';
         '<h2>' _ l('Event') _ '</h2>';
         IF entities.size;
-          INCLUDE 'components/events-list.tt'
-                  events=entities checkboxes='' no_type=1
+          React.embed(c, 'components/EventsList', { events => entities, noType => 1, seriesItemNumbers = series_item_numbers})
```
reosarevok: Substituted checkboxes='' for an optional checkboxes on the component and nothing here, since the desired effect here was to have checkboxes be false anyway.

root/edit/details/merge_events.tt
```diff
@@ -2,13 +2,13 @@
...
     <td>
-      [%~ INCLUDE 'components/events-list.tt' events=edit.display_data.old checkboxes='' no_ratings=1 ~%]
+      [%~ React.embed(c, 'components/EventsList', { events => edit.display_data.old, noRatings => 1}) ~%]
     </td>
   </tr>
   <tr>
     <th>[% l('Into:') %]</th>
     <td>
-      [%~ INCLUDE 'components/events-list.tt' events=edit.display_data.new checkboxes='' no_ratings=1 ~%]
+      [%~ React.embed(c, 'components/EventsList', { events => [edit.display_data.new], noRatings => 1}) ~%]
```
reosarevok: Since EventsList expects an array, events here is made into one. If a different solution would be preferred, let me know!

root/components/Paginator.js
mwiencek: I think because you're not using `serialize_pager` (from lib/MusicBrainz/Server/ControllerUtils/JSON.pm) whereas other pages do.

reosarevok: Oh. Duh. Will do.

root/components/events-list.tt
mwiencek: We set `sortable=1` in root/collection/index.tt, so it does get used on collection pages from what I can see. I definitely agree we'd want all tables to be sortable, though at the moment it's not supported everywhere (since these lists are paginated, it has to be done on the server rather than the client - that's what the `?order=` query parameters on collection pages allow for).

reosarevok: Oh, damn, I entirely missed the collection page (since it didn't come up with a search for the component filename). I guess we don't test collections work too well, given nothing complained :p 

So wait, how do I port the sorting, if it's currently done on the Perl level?

root/edit/details/merge_events.tt
mwiencek: I think it's what the TT is doing too, so sure. :) (The TT is probably compiled into something like `my @events = edit->display_data->{new}` which creates a list of one item).

root/components/events-list.tt
mwiencek: Well firstly you gotta port the `sortable_table_header` macro and add back the `sortable` prop along with it. Then you don't need to touch anything on the server side, just set `sortable={true}` on collection pages.

mwiencek: If you mean "how do I make the other pages sortable too," it'd need some changes to the controller/data layers which are prob outside the scope of this PR. :) For collections, we take an `order` query parameter and pass it to `find_by_collection` [here](https://github.com/reosarevok/musicbrainz-server/blob/b0a09f5/lib/MusicBrainz/Server/Controller/Collection.pm#L91:L96), which constructs an appropriate `ORDER BY` entry into the query. So we'd need to do something similar in the relevant Data:: methods these other pages use.

root/components/SortableTableHeader.js
```diff
@@ -0,0 +1,40 @@
...
+import {withCatalystContext} from '../context';
+import uriWith from '../utility/uriWith';
+
+function printSortArrows(name, order) {
+  if (order === name) {
+    return ' ▴';
+  } else if (order === '-' + name) {
+    return ' ▾';
+  }
+  return ' ▴/▾';
```
reosarevok: We currently don't show anything if there's no order set, but that means the fact these are orderable is basically a hidden feature. So this is my attempt at unhiding that.

mwiencek: Sounds like a good idea, though it looks a bit busy now. Maybe we should set 50% opacity when it's ▴/▾? That's what a few other table sort plugins I've looked at do.

lib/MusicBrainz/Server/Controller/Place.pm
```diff
@@ -104,8 +105,18 @@ sub events : Chained('load')
...
+        events      => $events,
+        place       => $c->stash->{place},
+        pager        => serialize_pager($c->stash->{pager}),
+    );
+
+    $c->stash(
+        component_path  => 'place/PlaceEvents.js',
+        component_props => \%props,
+        current_view    => 'Node',
+    );}
```
mwiencek: The } should go on the next line.

root/components/EventsList.js
```diff
@@ -0,0 +1,167 @@
...
+              <SortableTableHeader
+                label={l('Event')}
+                name="name"
+                order={order ? order : null}
+              />
+            )
+            : l('Event')}
+        </th>
+        {noType
+          ? null : (
```
mwiencek: I'd suggest the same style used for `noType` below, so `{noType ? null : (` on one line, de-indenting the code inside the `()`, and closing with `)}`. It takes up less space and is a bit more readable, I think

root/components/EventsList.js
```diff
@@ -0,0 +1,167 @@
...
+          )}
+          <td>{formatDatePeriod(event)}</td>
+          <td>{event.time}</td>
+          {noRatings ? null : (
+            <td>
+              <RatingStars entity={event} />
+            </td>
+          )}
+        </tr>
+
```
mwiencek: Extra blank line here

root/components/EventsList.js
```diff
@@ -0,0 +1,167 @@
...
+  +artist?: ArtistT,
+  +artistRoles?: boolean,
+  +checkboxes?: string,
+  +events: $ReadOnlyArray<EventT>,
+  +noArtists?: boolean,
+  +noLocation?: boolean,
+  +noRatings?: boolean,
+  +noType?: boolean,
+  +order?: string,
+  +seriesItemNumbers?: $ReadOnlyArray<string>,
```
mwiencek: It's actually an object, so the type here should be `{+[number]: string}`.

root/components/EventsList.js
```diff
@@ -0,0 +1,167 @@
...
+          {noArtists ? null : (
+            <td>
+              <ArtistRoles relations={event.performers} />
+            </td>
+          )}
+          {artist && artistRoles ? (
+            <td>
+              {event.performers.map(performer => (
+                performer.entity.id === artist.id ? (
+                  performer.roles.join(', ')
```
mwiencek: Can we use `commaOnlyList` here?

root/components/SortableTableHeader.js
reosarevok: 0.5 seemed a bit too strong still, 0.4 seemed about right to me, so did that.

PR #897, opened 2019-02-11
URLCleanup fixes: split grouped cleanups for clarity

reosarevok:
This just separates things that have their own match + cleanup but were before grouped into a huge ball of per-relationship-type cleanup. I feel it's much easier now to follow stuff. I left together under otherdatabases the ones that are only matched, but not cleaned up in any other way - I'd suggest to split those only if we start also cleaning them up.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -437,6 +437,14 @@ const CLEANUPS = {
       return id === LINK_TYPES.lyrics.work && /^http:\/\/animationsong\.com\/archives\/\d+\.html$/.test(url);
     },
   },
+  'animenewsnetwork': {
+    match: [new RegExp('^(https?://)?(www\\.)?animenewsnetwork\\.com', 'i')],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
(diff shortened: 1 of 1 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1515,65 +1603,25 @@ const CLEANUPS = {
...
     clean: function (url) {
-      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?changetip\.com\/tipme\/([^\/?#]+)(?:.*)?$/, 'https://www.changetip.com/tipme/$1');
-      url = url.replace(/^(?:https?:\/\/)?([^\/?#]+)\.tip\.me(?:[\/?#].*)?$/, 'https://www.changetip.com/tipme/$1');
-      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?flattr\.com\/profile\/([^\/?#]+)(?:.*)?$/, 'https://flattr.com/profile/$1');
       url = url.replace(/^((?:https?:\/\/)?(?:www\.)?patreon\.com\/user)\/(?:community|posts)(\?u=.*)$/, '$1$2');
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?patreon\.com\/((?:user\?u=)?[^\/?&#]+)(?:.*)?$/, 'https://www.patreon.com/$1');
+      return url;
+    },
+  },
+  'paypalme': {
```
yvanzo: May use the same identifier `paypal` as in [root/static/scripts/common/constants.js](https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/common/constants.js#L122)?

reosarevok: Sure, done :)

PR #898, opened 2019-02-11
MBS-9632: Convert Stats to React (VI)

reosarevok:
Cover art


root/statistics/CoverArt.js
```diff
@@ -0,0 +1,320 @@
...
+  +stat_name: string,
+  +type: string,
+|};
+
+const nameOrNull = (name: string, defaultName: string) => {
+  if (name === 'null') {
+    return defaultName;
+  }
+
+  return l(name);
```
yvanzo: This is inconsistent with every call to `nameOrNull` but one.

reosarevok:
Good catches, fixed!

PR #900, opened 2019-02-11
MBS-9936: Move from CJS style to ES6 (imports, exports) (VII)

reosarevok:
utility

### DIFF
--- a/root/utility/ratingTooltip.js
+++ b/root/utility/ratingTooltip.js
@@ -5,7 +5,7 @@
 // and is licensed under the GPL version 2, or (at your option) any
 // later version: http://www.gnu.org/licenses/gpl-2.0.txt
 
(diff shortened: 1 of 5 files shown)


root/utility/ratingTooltip.js
```diff
@@ -5,12 +5,12 @@
...
+import {l, ln} from '../static/scripts/common/i18n';
 
 const ratingTooltip = (rating: number) => (
   rating === 0
     ? l('Remove your rating')
     : ln('Rate: {rating} star', 'Rate: {rating} stars', rating, {rating})
 );
 
-module.exports = ratingTooltip;
+export default ratingTooltip;
```
mwiencek: Looks like this breaks an import in root/static/scripts/common/ratings.js

PR #902, opened 2019-02-11
MBS-10022: Event tab for areas

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10022

I don't see any real reason not to add this, and in fact it should at least help lower the amount of stuff cluttering the Overview page of areas. Given some take a minute to load, this should be a slight improvement. Plus, places have this tab and we use place and area as effectively equivalent everywhere else for events.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Area.pm
+++ b/lib/MusicBrainz/Server/Controller/Area.pm
@@ -123,6 +123,9 @@ sub events : Chained('load')
         $c->model('Event')->find_by_area($c->stash->{area}->id, shift, shift);
     });
     $c->model('Event')->load_related_info(@$events);
+    $c->model('Area')->load(map { $_->{entity} } map { $_->all_places } @$events);
+    $c->model('Area')->load_containment(map { (map { $_->{entity} } $_->all_areas),
+                                              (map { $_->{entity}->area } $_->all_places) } @$events);
     $c->model('Event')->rating->load_user_ratings($c->user->id, @$events) if $c->user_exists;
 
     my %props = (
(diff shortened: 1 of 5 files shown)


mwiencek:
This is cool, but I think it would be more useful and less confusing if it took into account containing areas. What do you think about modifying `Data::Event::find_by_area` to do that? `Data::Editor::find_by_area` can be used as a guide: https://github.com/metabrainz/musicbrainz-server/blob/af2f7d3c7b18373d84495ff8bebb7224bbb3bfb2/lib/MusicBrainz/Server/Data/Editor.pm#L167:L184

reosarevok:
Did that. Since we're doing this we also need to actually show the areas and places the event is linked to, but given that helps find problems that might not be too bad.

mwiencek:
> Works for me, except that inclusion of contained areas should be somehow documented.

The location column shows the specific sub-area that the event is in, which seems pretty straightforward to me. (I'd more expect actual text documentation if it didn't include sub-areas, since it'd be very unexpected to click on Illinois and not see all events in Illinois :))

yvanzo:
Improving overall documention is certainly beyond the scope of this issue as every other tab of area page works the same. At some point, that should be still documented somewhere, at least for the many areas with no event in any subdivision or no event at all.

lib/MusicBrainz/Server/Controller/Area.pm
```diff
@@ -105,6 +110,37 @@ sub artists : Chained('load')
...
+=cut
+
+sub events : Chained('load')
+{
+    my ($self, $c) = @_;
+    my $events = $self->_load_paged($c, sub {
+        $c->model('Event')->find_by_area($c->stash->{area}->id, shift, shift);
+    });
+    $c->model('Event')->load_related_info(@$events);
+    $c->model('Area')->load(map { map { $_->{entity} } $_->all_places } @$events);
```
mwiencek: The code works, but maps within maps are confusing (it can be hard to make sense of the brackets and which scope `$_` belongs to). This can be written more simply as:
```suggestion
    $c->model('Area')->load(map { $_->{entity} } map { $_->all_places } @$events);
```

PR #906, opened 2019-02-14
MBS-10029: Include editor on page title of Edits by Editor

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10029

I'm not happy with the capitalization (we use "Release group X by Y" for example, so I'd want "Open edits" rather than "Open Edits") so changing that on a second commit but I can drop that one if it's a problem.

### DIFF
--- a/root/user/edits.tt
+++ b/root/user/edits.tt
@@ -5,23 +5,23 @@
             page_title = l('Votes by {name}', title_param);
             page_heading = l('Votes by {name}', heading_param);
         CASE 'open';
-            page_title = l('Open Edits by {name}', title_param);
-            page_heading = l('Open Edits by {name}', heading_param);
+            page_title = l('Open edits by {name}', title_param);
+            page_heading = l('Open edits by {name}', heading_param);
         CASE 'cancelled';
-            page_title = l('Cancelled Edits by {name}', title_param);
-            page_heading = l('Cancelled Edits by {name}', heading_param);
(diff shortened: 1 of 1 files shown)


root/user/edits.tt
```diff
@@ -1,29 +1,29 @@
...
         CASE 'votes';
-            page_title = l('Votes');
-            page_heading = l('Votes by {name}', param);
+            page_title = l('Votes by {username}', param);
+            page_heading = l('Votes by {editor}', param);
         CASE 'open';
-            page_title = l('Open Edits');
-            page_heading = l('Open Edits by {name}', param);
+            page_title = l('Open edits by {username}', param);
+            page_heading = l('Open edits by {editor}', param);
```
mwiencek: How come for all of these, one string uses `username` for the var arg and the other `editor`? I'd think `editor` makes perfect sense for both (doesn't really make a difference that one's clickable and one's not).

Otherwise, I support eradicating bizarre capitalization.

reosarevok: Because how do I pass a different argument to each then? (If I pass editor to the title, it shows it as an HTML a element)

mwiencek: By having two vars like
```
[%- title_param = { editor => user.name };
    heading_param = { editor => link_editor(user) };
    SWITCH c.action.name;
        CASE 'votes';
            page_title = l('Votes by {editor}', title_param);
            page_heading = l('Votes by {editor}', heading_param);
```

PR #907, opened 2019-02-14
MBS-9632: Convert Stats to React (VII)

reosarevok:
Edits


reosarevok:
We can choose to have 
`import {l as l_mb_server} from '../static/scripts/common/i18n';` (more consistent, but makes eslint cry) or `import {l as lMbServer} from '../static/scripts/common/i18n';` (camel case, but these functions usually aren't). Which is preferred?

root/statistics/Edits.js
```diff
@@ -0,0 +1,91 @@
...
+
+type EditsStatsT = {|
+  +$c: CatalystContextT,
+  +dateCollected: string,
+  +stats: {[string]: number},
+  +statsByCategory: {[string]: $ReadOnlyArray<EditCategoryT>},
+|};
+
+const Edits = ({$c, dateCollected, stats, statsByCategory}: EditsStatsT) => {
+  console.log(stats);
```
mwiencek: ❗️

root/statistics/Edits.js
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <StatisticsLayout fullWidth page="edits" title={l('Edits')}>
+      {manifest.css('statistics')}
```
mwiencek: `StatisticsLayout` already includes the CSS, so you can remove this plus the `manifest` import at the top (I'll have to do it in my Webpack PR in any case)

root/statistics/Edits.js
```diff
@@ -0,0 +1,91 @@
...
+      <h2>{l('Edits')}</h2>
+      {Object.keys(statsByCategory).length === 0 ? (
+        <p>
+          {l('No edit statistics available.')}
+        </p>
+      ) : (
+        <table className="database-statistics">
+          <tbody>
+            <tr>
+              <th colSpan="2">{l('Edits:')}</th>
```
mwiencek: `addColon` here?

root/statistics/Edits.js
reosarevok: Did that for every file in a second commit. One of the files (root/statistics/LanguagesScripts.js) also includes {manifest.js('statistics')}, which is not in the Layout, so I left that in.

PR #909, opened 2019-02-15
MBS-10035: Add text to "report user" page asking for links

Freso:
<https://tickets.metabrainz.org/browse/MBS-10035>

I'm not sure about any (coding) style I need to adhere to for this, so please let me know of any changes I may need to make.

### DIFF
--- a/root/user/report.tt
+++ b/root/user/report.tt
@@ -9,6 +9,20 @@
 
     <p>[%- l('If you’d like us to be able to easily respond to your report, please check “Reveal my email address” below.') -%]</p>
 
+    <p>
+        <strong>[%- add_colon(l('Note')) -%]</strong>
+        [%- l('Be sure to provide direct links to examples of the behaviour
+               you’re reporting (for example, use
+               “https://musicbrainz.org/edit/23” instead of “edit #23”, and
+               use “https://musicbrainz.org/edit/42 and
+               https://musicbrainz.org/edit/43” instead of
+               “https://musicbrainz.org/user/SomeUser/edits”).
+               Providing links makes it much easier for the recipients of the
+               report to look into the issues; a report without links is
+               unlikely to be acted on fast, since it will require a lot of
+               additional research.') -%]
+    </p>
+
     [%- USE r = FormRenderer(form) -%]
 
     <form action="[% c.req.uri %]" method="post" class="report-form">


root/user/report.tt
```diff
@@ -9,6 +9,8 @@
 
     <p>[%- l('If you’d like us to be able to easily respond to your report, please check “Reveal my email address” below.') -%]</p>
 
+    <p>[%- l('<strong>Note:</strong> Be sure to provide direct links to examples of the behaviour you're reporting (e.g., https://musicbrainz.org/edit/23 instead of "edit #23", and "https://musicbrainz.org/edit/42" and "https://musicbrainz.org/edit/43" instead of "https://musicbrainz.org/user/<em>SomeUser</em>/edits"). And please always do provide example links so the recipients of the report can look into it without first spending an hour to research what you're actually referring to.') -%]</p>
```
reosarevok: I would suggest two changes here: one is to turn `<strong>Note:</strong>` into a separate `<strong>{%- l('Note:') -%]</strong>` or even `<strong>{%- add_colon(l('Note')) -%]</strong>` - this means we can reuse what might well be an existing translation for "Note", while getting the HTML out of the translation string and ensuring it's always bold. I think this should work, anyway!

The other is to make the string multiline, for readability. So something like: 
```
[%- l('Be sure to provide direct links to examples of the behaviour you're reporting 
      (e.g., https://musicbrainz.org/edit/23 instead of "edit #23", and') -%]
```
(make it so each new line starts just under "Be")

reosarevok: For the text itself, I'd write it a bit more like:

"Be sure to provide direct links to examples of the behaviour you're reporting (e.g., https://musicbrainz.org/edit/23 instead of "edit #23", and "https://musicbrainz.org/edit/42" and "https://musicbrainz.org/edit/43" instead of "https://musicbrainz.org/user/SomeUser/edits"). Providing links makes it much easier for the recipients of the report to look into the issues; a report without links is unlikely to be acted on fast, since it will require a lot of additional research.') -%]</p>

Freso: Thanks! What's the suggested line-length limit?

reosarevok: It's 78 for Javascript, something like that might make sense here too? But at the very least not more than the 131 for the previous line :p

Freso: I pushed too fast. :) Does it look okay now? Should I make another push with a 78 limit?

reosarevok: I would do that, plus indent them to all start under the start of the string ("Be"), not the template itself. See https://github.com/metabrainz/musicbrainz-server/blob/master/root/report/multiple_discogs_links.tt for example

Freso: Alright, pushed. Not sure I agree that this is better/more readable though, but if that's what is wanted. :+1:

PR #910, opened 2019-02-15
MBS-10037: Send editor reports to user admin mailing list

Freso:
Also sets the Reply-To header to said mailing list.

Again, not sure of code style, so let me know what needs changing for this to be acceptable!


lib/MusicBrainz/Server/Email.pm
```diff
@@ -524,20 +524,20 @@ $message
 EOF
     }
 
-    my $admin_addresses = join ', ', map { _user_address($_) } @{ $opts{admins} };
+    my $admin_address = 'MusicBrainz Account Admins <musicbrainz-account-admin@metabrainz.org>'
```
Freso: Should this maybe be moved to a constant or `DBDefs`?

Freso:
I don't know how to read the `perl-and-pgtap` error messages.

lib/MusicBrainz/Server/Email.pm
mwiencek: Yeah, I would move it to lib/MusicBrainz/Server/Constants.pm together with the other EMAIL_* addresses.

lib/MusicBrainz/Server/Email.pm
```diff
@@ -524,20 +524,20 @@ $message
 EOF
     }
 
-    my $admin_addresses = join ', ', map { _user_address($_) } @{ $opts{admins} };
```
mwiencek: `$opts{admins}` is no longer used, so you can remove the two lines from https://github.com/metabrainz/musicbrainz-server/blob/3caba434099a225efcac614bda92ecf6709a1a20/lib/MusicBrainz/Server/Controller/User.pm#L524-L530 that fetch them and pass them in.

lib/MusicBrainz/Server/Email.pm
Freso: Should be fixed as of https://github.com/metabrainz/musicbrainz-server/commit/0eb2f810dfcf71840c6cbcfa898d49054e42335a

lib/MusicBrainz/Server/Email.pm
Freso: Should be fixed as of https://github.com/metabrainz/musicbrainz-server/commit/a2931c07aaf105db13e5d0c290e32873a4f669e8

PR #912, opened 2019-02-15
Fix APT package installation failure by refreshing keys

yvanzo:
**Problem:** Yarn Packaging subkey often changed during the latest months, breaking our build process.
**Solution:** Refresh APT keys when installing yarn package or on package installation failure.

### DIFF
--- a/docker/templates/macros.m4
+++ b/docker/templates/macros.m4
@@ -3,8 +3,10 @@ m4_divert(-1)
 m4_define(
     `apt_install',
     `m4_dnl
-apt-get update && \
-    apt-get install --no-install-suggests --no-install-recommends -y $1 && \
+apt-get update && ( \
+    apt-get install --no-install-suggests --no-install-recommends -y $1 || ( \
+        apt-key adv --keyserver hkps.pool.sks-keyservers.net --refresh-keys && \
+        apt-get install --no-install-suggests --no-install-recommends -y $1 ) ) && \
     rm -rf /var/lib/apt/lists/*')
 
 m4_define(`apt_purge', `apt-get purge --auto-remove -y $1')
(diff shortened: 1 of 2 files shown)


docker/templates/macros.m4
```diff
@@ -3,8 +3,10 @@ m4_divert(-1)
 m4_define(
     `apt_install',
     `m4_dnl
-apt-get update && \
-    apt-get install --no-install-suggests --no-install-recommends -y $1 && \
+apt-get update && ( \
+    apt-get install --no-install-suggests --no-install-recommends -y $1 || ( \
+        apt-key adv --keyserver hkps.pool.sks-keyservers.net --refresh-keys && \
```
mwiencek: Any reason we don't always run this first, instead of only inside an `|| ()`? (If the first `apt-get install` fails, it's not necessarily the keys that were an issue either.)

yvanzo: Cause it is unneeded in most cases. Depending on the key server drawn from the pool, refreshing keys might take some time. Retrying after refreshing keys won’t hurt even though keys might not be the issue. 

The only case of failure we have is with yarn package. Would you prefer systematically refreshing yarn keys only? See https://github.com/metabrainz/musicbrainz-docker/commit/2906242e6c5063022b0c6737d8462c109de57b47 for example, which could apply to https://github.com/metabrainz/musicbrainz-server/blob/c6ddafd6cf4e9b6f7d5ea0e1bafc6de809ad8e3f/docker/templates/macros.m4#L25

mwiencek: If refreshing only the yarn keys is fast, I don't see an issue with doing it unconditionally, otherwise what you have works too.

yvanzo: It is not really faster than refreshing all keys, as the delay is usually due to server latency/load. I stick with what I have.

PR #913, opened 2019-02-15
MBS-10031: Fix schema mapping of entities in relationships

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10031

Reason seemed to be the sort-name served by the search result wasn't being turned into sort_name since the fixup didn't run on the relationship entity. This does now load it properly, but I'm not 100% sure this is the best possible way of doing it and won't have any unintended consequences.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -591,10 +591,11 @@ sub schema_fixup
 
             my %entity = %{ $rel->{$entity_type} };
 
+            $self->schema_fixup(\%entity, $entity_type);
+
             # The search server returns the MBID in the 'id' attribute, so we
-            # need to rename that.
-            $entity{gid} = delete $entity{id};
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -594,6 +600,7 @@ sub schema_fixup
             # The search server returns the MBID in the 'id' attribute, so we
             # need to rename that.
             $entity{gid} = delete $entity{id};
+            %entity = %{ $self->schema_fixup_mapping(\%entity) };
             %entity = %{ $self->schema_fixup_type(\%entity, $entity_type) };
```
mwiencek: I *think* you can just replace the `schema_fixup_type` line here with a call to `schema_fixup`, and revert everything else. That should both fix the mapping and call `schema_fixup_type` (and perhaps do other things we've forgotten, so it seems somewhat preferable).

PR #914, opened 2019-02-16
MBS-10021: Make artist roles translatable

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10021

### DIFF
--- a/lib/MusicBrainz/Server/Data/Event.pm
+++ b/lib/MusicBrainz/Server/Data/Event.pm
@@ -392,7 +392,7 @@ sub _find_performers
         $map->{$event_id} ||= [];
         push @{ $map->{$event_id} }, {
             entity => $artists->{$artist_id},
-            roles => $roles
+            roles => [ uniq @{ $roles } ]
         }
     }
 }
--- a/root/components/EventsList.js
+++ b/root/components/EventsList.js
@@ -12,6 +12,7 @@ import React from 'react';
 import {l} from '../static/scripts/common/i18n';
 import {lp_attributes} from '../static/scripts/common/i18n/attributes';
 import commaOnlyList from '../static/scripts/common/i18n/commaOnlyList';
+import localizeArtistRoles from '../static/scripts/common/i18n/localizeArtistRoles';
 import {withCatalystContext} from '../context';
 import loopParity from '../utility/loopParity';
 import ArtistRoles from '../static/scripts/common/components/ArtistRoles';
@@ -138,7 +139,7 @@ const EventsList = ({
             <td>
               {event.performers.map(performer => (
                 performer.entity.id === artist.id ? (
-                  commaOnlyList(performer.roles)
+                  commaOnlyList(localizeArtistRoles(performer.roles))
                 ) : null
               ))}
             </td>
--- a/root/static/scripts/common/components/ArtistRoles.js
+++ b/root/static/scripts/common/components/ArtistRoles.js
@@ -29,7 +29,7 @@ const ArtistRoles = ({relations}: Props) => (
       <li key={r.entity.id}>
         {l('{artist} ({roles})', {
           artist: <EntityLink entity={r.entity} />,
-          roles: commaOnlyList(uniq(localizeArtistRoles(r.roles))),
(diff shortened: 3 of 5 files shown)


root/static/scripts/common/components/ArtistRoles.js
```diff
@@ -24,14 +25,11 @@ type Props = {|
 
 const ArtistRoles = ({relations}: Props) => (
   <ul>
-    {relations.map(r =>(
+    {relations.map(r => (
```
reosarevok: This was an accident but actually it seems to be more consistent with how we usually do it, so I left it in!

root/static/scripts/common/components/ArtistRoles.js
```diff
@@ -24,14 +25,11 @@ type Props = {|
...
       <li key={r.entity.id}>
         {l('{artist} ({roles})', {
           artist: <EntityLink entity={r.entity} />,
-          roles: r.roles.length > 1
-            // $FlowFixMe
-            ? commaOnlyList(uniq(r.roles))
-            : r.roles[0],
+          roles: commaOnlyList(uniq(localizeArtistRoles(r.roles))),
```
reosarevok: Got rid of the if here because AFAICT commaOnlyList will still give a no comma one element list if there's just one, so I don't see what we're winning by having this here.

root/components/EventsList.js
```diff
@@ -138,7 +139,7 @@ const EventsList = ({
             <td>
               {event.performers.map(performer => (
                 performer.entity.id === artist.id ? (
-                  commaOnlyList(performer.roles)
+                  commaOnlyList(localizeArtistRoles(performer.roles))
```
reosarevok: We do uniq() in ArtistRoles.js but not here - should we?

reosarevok:
Does this also fix https://tickets.metabrainz.org/browse/MBS-5782 ? (that page isn't in React yet, but once it is it will use ArtistRoles). Or should I link a "Turn artist work page into React" ticket as "resolves this"?

root/static/scripts/common/i18n/localizeArtistRoles.js
```diff
@@ -0,0 +1,21 @@
...
+
+import {l_relationships} from './relationships';
+
+function localizeArtistRoles(roles: $ReadOnlyArray<string>) {
+  const translatedRoles = [];
+  roles.map(role => {
+    translatedRoles.push(l_relationships(role));
+  });
+  return translatedRoles;
+}
```
mwiencek: You can simplify this function to just:
```JavaScript
function localizeArtistRoles(roles: $ReadOnlyArray<string>) {
  return roles.map(role => l_relationships(role));
}
```

(Map returns a new list with each item mapped through the given function - you're using it more like `forEach` which isn't necessary.)

root/components/EventsList.js
mwiencek: Not sure - I'd think it should be done on the server in both cases.

root/static/scripts/common/components/ArtistRoles.js
mwiencek: That's correct, I just figured the majority would only have a single role, so optimized for the common case.

reosarevok: I'm not sure I understand whether you think we should keep the conditional or not :)

mwiencek: It was premature optimization, so no, it's fine to remove. :)

lib/MusicBrainz/Server/Data/Event.pm
```diff
@@ -392,7 +392,7 @@ sub _find_performers
         $map->{$event_id} ||= [];
         push @{ $map->{$event_id} }, {
             entity => $artists->{$artist_id},
-            roles => $roles
+            roles => [ uniq map { $_ } @{ $roles } ]
```
mwiencek: The `map` doesn't appear to be doing anything here. `uniq @{ $roles }` should be sufficient?

reosarevok: True! (tbh I just fought the stuff for quite a while until I found something that worked and then I didn't dare to touch it again without supervision :) )

PR #916, opened 2019-02-18
MBS-9999: Allow custom scheme for redirect URI

yvanzo:
### Resolve [MBS-9999](https://tickets.metabrainz.org/browse/MBS-9999) by adding optional redirect URI for “installed applications”

Scheme of such URI must be a reverse-DNS string, e.g. `org.example.app`.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Application.pm
+++ b/lib/MusicBrainz/Server/Form/Application.pm
@@ -1,6 +1,7 @@
 package MusicBrainz::Server::Form::Application;
 
 use HTML::FormHandler::Moose;
+use MusicBrainz::Server::Constants qw( :oauth_redirect_uri_re );
 use MusicBrainz::Server::Translation qw( l );
 
 extends 'MusicBrainz::Server::Form';
@@ -40,7 +41,21 @@ sub validate
 
     if ($self->field('oauth_type')->value eq 'web') {
         if (!$self->field('oauth_redirect_uri')->value) {
-            $self->field('oauth_redirect_uri')->add_error(l('Redirect URL must be entered for web applications.'));
+            $self->field('oauth_redirect_uri')->add_error(
+                l('Redirect URL must be entered for web applications.')
+            );
+        } elsif ($self->field('oauth_redirect_uri')->value !~ $OAUTH_WEB_APP_REDIRECT_URI_RE) {
+            $self->field('oauth_redirect_uri')->add_error(
+                l('Redirect URL scheme must be either <code>http</code> or ' .
+                  '<code>https</code> for web applications.')
+            );
+        }
+    } elsif ($self->field('oauth_redirect_uri')->value) {
+        if ($self->field('oauth_redirect_uri')->value !~ $OAUTH_INSTALLED_APP_REDIRECT_URI_RE) {
+            $self->field('oauth_redirect_uri')->add_error(
+                l('Redirect URL scheme must be a reverse-DNS string, as in ' .
+                  '<code>org.example.app://auth</code>, for installed applications.')
+            );
         }
     }
 }
--- a/root/static/scripts/account/components/ApplicationForm.js
+++ b/root/static/scripts/account/components/ApplicationForm.js
@@ -8,7 +8,6 @@
  */
 
 import * as React from 'react';
-import noop from 'lodash/noop';
 
 import FormRow from '../../../../components/FormRow';
 import FormRowSelect from '../../../../components/FormRowSelect';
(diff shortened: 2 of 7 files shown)


root/static/scripts/account/components/ApplicationForm.js
```diff
@@ -101,14 +101,21 @@ class ApplicationForm extends React.Component<Props, State> {
...
         />
-        {this.state.form.field.oauth_type.value === 'web' ? (
-          <FormRowURLLong
-            field={this.state.form.field.oauth_redirect_uri}
-            label={addColon(l('Callback URL'))}
-            onChange={this.handleOauthRedirectURIChange}
-            required
-          />
-        ) : null}
+        <FormRowURLLong
```
reosarevok: This has type="url" so it always auto-validates, meaning the example ("org.example.app") is not accepted.

yvanzo: It was actually invalidated by `MusicBrainz::Server::Form::Field::OAuthRedirectURI`, now fixed with https://github.com/metabrainz/musicbrainz-server/pull/916/commits/3f84f7718c1f2c322b1fc2bb4b77bd0f871c164f#diff-e6382dac47170663e5de065f3af86848

root/static/scripts/account/components/ApplicationForm.js
```diff
@@ -101,14 +101,21 @@ class ApplicationForm extends React.Component<Props, State> {
...
+          label={addColon(l('Callback URL'))}
+          onChange={this.handleOauthRedirectURIChange}
+          required={this.state.form.field.oauth_type.value === 'web'}
+        />
+        {this.state.form.field.oauth_type.value === 'web' ? null : (
+          <FormRow hasNoLabel>
+            <span className="input-note">
+              {l(`Callback URI is optional for installed applications.
+                  If set, its scheme must be a custom reverse-DNS
+                  string, e.g. <code>org.example.app</code>`)}
```
mwiencek:
```suggestion
                  string, e.g. <code>org.example.app</code>.`)}
```

lib/MusicBrainz/Server/Entity/Util/Application.pm
```diff
@@ -0,0 +1,29 @@
+package MusicBrainz::Server::Entity::Util::Application;
```
mwiencek: The `Entity::Util` namespace seems a bit weird since it just contains a couple constants. Maybe they'd fit in lib/MusicBrainz/Server/Constants.pm with `OAUTH_*` prefixes?

root/static/scripts/account/components/ApplicationForm.js
yvanzo: Fixed at https://github.com/metabrainz/musicbrainz-server/pull/916/commits/8514b3e73781dbe9fde1f964f22f96ba53414a25#diff-640c8996ba9f667164580a89f0cdf4c8R115

lib/MusicBrainz/Server/Entity/Util/Application.pm
yvanzo: Yep, done in https://github.com/metabrainz/musicbrainz-server/pull/916/commits/8514b3e73781dbe9fde1f964f22f96ba53414a25

lib/MusicBrainz/Server/Form/Application.pm
```diff
@@ -40,7 +41,21 @@ sub validate
...
+            $self->field('oauth_redirect_uri')->add_error(
+                l('Redirect URL scheme must be either <code>http</code> or ' .
+                  '<code>https</code> for web applications.')
+            );
+        }
+    } elsif ($self->field('oauth_redirect_uri')->value) {
+        if ($self->field('oauth_redirect_uri')->value !~ $OAUTH_INSTALLED_APP_REDIRECT_URI_RE) {
+            $self->field('oauth_redirect_uri')->add_error(
+                l('Redirect URL scheme must be a reverse-DNS string, e.g. ' .
+                  '<code>org.example.app</code>, for installed applications.')
```
mwiencek: Can we specify `://` there too? Apparently Android requires that at least and this message confused me at first.
```suggestion
                  '<code>org.example.app://</code>, for installed applications.')
```

yvanzo: As `://` is technically not part of the scheme, I made a [full example](https://github.com/metabrainz/musicbrainz-server/pull/916/commits/f6d353a1a8b7384be4af93b6c9f0c02b6ffa8e33#diff-4cf7869253939843a3ad96313ecbce3dR56) instead.

root/static/scripts/account/components/ApplicationForm.js
reosarevok: If I enter the example ("org.example.app") I still get "Please enter a URL."

It does work if I enter "org.example.app://" (with the "://" end) but we really shouldn't have a situation where the example we give is actually not allowed for entry.

mwiencek: Oops, I meant to leave the comment at https://github.com/metabrainz/musicbrainz-server/pull/916#discussion_r259857876 for both strings. Can we just copy that here?

yvanzo: I replaced the misleading URL sheme example in the input note with the full URL example I already fixed in the error message.

PR #917, opened 2019-02-18
MBS-10032: detect and cleanup iTunes "author" URLs for artists

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10032

Since every artist seems to have an author page automatically (https://itunes.apple.com/us/author/yung-beef/id1090884617 is a rapper with zero books, so it's not even connected to having multiple types of media), and the IDs are the same, I feel it makes sense to just store the links as "artist" (which is specifically about music AFAICT).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1205,7 +1205,10 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?itunes\\.apple\\.com/', 'i')],
     type: LINK_TYPES.downloadpurchase,
     clean: function (url) {
-      return url.replace(/^https?:\/\/(?:geo\.)?itunes\.apple\.com\/([a-z]{2}\/)?(artist|album|audiobook|music-video|podcast|preorder)\/(?:[^?#\/]+\/)?(?:id)?([0-9]+)(?:\?.*)?$/, 'https://itunes.apple.com/$1$2/id$3');
+      url = url.replace(/^https?:\/\/(?:geo\.)?itunes\.apple\.com\/([a-z]{2}\/)?(artist|album|audiobook|author|music-video|podcast|preorder)\/(?:[^?#\/]+\/)?(?:id)?([0-9]+)(?:\?.*)?$/, 'https://itunes.apple.com/$1$2/id$3');
+      // Author seems to be a different interface for artist with the same ID
+      url = url.replace(/^(https:\/\/itunes\.apple\.com(?:\/[a-z]{2})?)\/author\//, '$1/artist/');
+      return url;
     },
     validate: function (url, id) {
       const m = /^https:\/\/itunes\.apple\.com\/(?:[a-z]{2}\/)?([a-z-]{3,})\/id[0-9]+$/.exec(url);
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1205,7 +1205,10 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?itunes\\.apple\\.com/', 'i')],
     type: LINK_TYPES.downloadpurchase,
     clean: function (url) {
-      return url.replace(/^https?:\/\/(?:geo\.)?itunes\.apple\.com\/([a-z]{2}\/)?(artist|album|audiobook|music-video|podcast|preorder)\/(?:[^?#\/]+\/)?(?:id)?([0-9]+)(?:\?.*)?$/, 'https://itunes.apple.com/$1$2/id$3');
+      url = url.replace(/^https?:\/\/(?:geo\.)?itunes\.apple\.com\/([a-z]{2}\/)?(artist|album|audiobook|author|music-video|podcast|preorder)\/(?:[^?#\/]+\/)?(?:id)?([0-9]+)(?:\?.*)?$/, 'https://itunes.apple.com/$1$2/id$3');
+      // Author seems to be a different interface for artist with the same ID
+      url = url.replace(/itunes\.apple\.com\/([a-z]{2}\/)?author/, 'itunes.apple.com/$1artist');
```
yvanzo: This defeats cleanup for every other URL part: scheme, host, `id` prefix, query, fragment.
* The first `replace` statement should not have `author` removed.
* The second `replace`’s pattern should preferably affect clean URLs only and can be slightly simplified by capturing a larger group, as follows:
```suggestion
      url = url.replace(/^(https:\/\/itunes\.apple\.com(?:\/[a-z]{2})?)\/author\//, '$1/artist/');
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1609,6 +1609,13 @@ const testData = [
             expected_clean_url: 'https://itunes.apple.com/artist/id444923726',
        only_valid_entity_types: ['artist'],
   },
+  {
+                     input_url: 'https://itunes.apple.com/us/author/paige-lewis/id348965238',
```
yvanzo: Adding a fragment would allow to detect the above mentioned issue:
```suggestion
                     input_url: 'https://itunes.apple.com/us/author/paige-lewis/id348965238#',
```

root/static/scripts/edit/URLCleanup.js
reosarevok: Wait, isn't that what it does? The first is the same as before, but allowing author, and the second takes the clean url and just switches author for artist. I guess it might even be safe to literally do "if author then say artist" since there shouldn't be any other "author" in the URL pattern, but.

yvanzo: like with `http://screwed.itunes.apple.com/ko/author/dj-author/7`? :clown_face:

reosarevok: But aren't we already dropping that earlier when doing https://itunes.apple.com/$1$2/id$3 ?

yvanzo: No, because of `screwed`.

reosarevok: Huh, I see now, but why do we even accept those urls in the first place? :/

yvanzo: It is denied by the function `validate` which is called after the function `clean`.

PR #918, opened 2019-02-18
MBS-9909: Change reports to use React (8 - ISRCs)

reosarevok:
Also changed ISRC to Isrc in the same way we did with ISWCs.


root/report/IsrcsWithManyRecordings.js
```diff
@@ -0,0 +1,109 @@
...
+              <th>{l('Length')}</th>
+            </tr>
+          </thead>
+          <tbody>
+            {items.map((item, index) => {
+              lastIsrc = currentIsrc;
+              currentIsrc = item.isrc;
+
+              return (
+                <>
```
yvanzo: Should be `<React.Fragment key={…}>` here. Found the same issue with already converted `CollaborationRelationships`, `DuplicateArtists`, and `IswcsWithManyWorks`.

reosarevok: Done (I think?)

yvanzo: Almost, except that key should probably be a concatenation of children’s keys rather than `index`.

reosarevok: Better?

yvanzo: Yep, thank you! I just edited your commit to restore `key` attributes on `tr` elements.

reosarevok: Aren't the keys only supposed to be on the top element of the map? That's what I understood from @mwiencek, that those tr keys were unneeded.

yvanzo: Maybe I have been a bit hasty indeed, but keys don’t seem to hurt.

root/report/IsrcsWithManyRecordings.js
```diff
@@ -0,0 +1,109 @@
...
+              lastIsrc = currentIsrc;
+              currentIsrc = item.isrc;
+
+              return (
+                <React.Fragment key={item.isrc + '-' + item.recording.gid}>
+                  {lastIsrc === item.isrc ? null : (
+                    <tr className="even" key={item.isrc}>
+                      <td>
+                        <a href={'/isrc/' + item.isrc}>{item.isrc}</a>
+                        <span>{' (' + item.recordingcount + ')'}</span>
```
mwiencek: `bracketed`?

reosarevok:
Added a few eslint fixes and changed old URL files to Url too in the reports folder.

root/report/IsrcsWithManyRecordings.js
```diff
@@ -0,0 +1,110 @@
...
+              lastIsrc = currentIsrc;
+              currentIsrc = item.isrc;
+
+              return (
+                <React.Fragment key={item.isrc + '-' + item.recording.gid}>
+                  {lastIsrc === item.isrc ? null : (
+                    <tr className="even">
+                      <td>
+                        <a href={'/isrc/' + item.isrc}>{item.isrc}</a>
+                        <span>{bracketed(item.recordingcount)}</span>
```
mwiencek: Pretty sure you need a `{' '}` between the a and span too since bracketed doesn't include one (sorry)

reosarevok: Done (also for ISWCs)

PR #919, opened 2019-02-18
MBS-9936: Move from CJS style to ES6 (imports, exports) (IX)

reosarevok:
scripts/components

### DIFF
--- a/root/static/scripts/area/places-map.js
+++ b/root/static/scripts/area/places-map.js
@@ -9,10 +9,12 @@ const _ = require('lodash');
 const React = require('react');
 const ReactDOMServer = require('react-dom/server');
 
-const EntityLink = require('../common/components/EntityLink');
+const EntityLink = require('../common/components/EntityLink').default;
 const {l, ln} = require('../common/i18n');
 const {createMap, L} = require('../common/leaflet');
-const {places} = require('../common/utility/getScriptArgs')();
+import getScriptArgs from '../common/utility/getScriptArgs';
+
+const {places} = getScriptArgs();
 
 const CLUSTER_POPUP_LIMIT = 10;
 
--- a/root/static/scripts/common/MB.js
+++ b/root/static/scripts/common/MB.js
@@ -18,9 +18,9 @@
 
 'use strict';
 
-const constants = require('./constants');
+import global from '../global';
 
-const global = require('../global');
+import * as constants from './constants';
 
 // Namespaces
 const MB: {[string]: *} = {
@@ -39,4 +39,4 @@ const MB: {[string]: *} = {
 
 global.MB = MB;
 
-module.exports = MB;
+export default MB;
--- a/root/static/scripts/common/components/TagEditor.js
+++ b/root/static/scripts/common/components/TagEditor.js
@@ -5,21 +5,21 @@
 // and is licensed under the GPL version 2, or (at your option) any
 // later version: http://www.gnu.org/licenses/gpl-2.0.txt
 
-const _ = require('lodash');
-const React = require('react');
-const ReactDOM = require('react-dom');
+import _ from 'lodash';
+import React from 'react';
+import ReactDOM from 'react-dom';
 import keyBy from 'terable/keyBy';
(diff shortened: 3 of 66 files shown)


root/static/scripts/common/components/TagEditor.js
```diff
@@ -436,158 +436,166 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
   }
 }
 
-class MainTagEditor extends TagEditor {
```
reosarevok: The diff here is absolutely nuts for some reason, but: this moves the hydrate from the export at the end to here, `export const MainTagEditor = hydrate<TagEditorProps>(whatever)` etc. Every change inside the class extends is indentation.

root/static/scripts/common/entity.js
```diff
@@ -3,31 +3,28 @@
...
-const ko = require('knockout');
-const _ = require('lodash');
-const React = require('react');
-const ReactDOMServer = require('react-dom/server');
-
-const ArtistCreditLink = require('./components/ArtistCreditLink');
-const EditorLink = require('./components/EditorLink');
-const EntityLink = require('./components/EntityLink');
-const {
-    ENTITY_NAMES,
```
yvanzo: ENTITY_NAMES seems to be still needed, see https://github.com/metabrainz/musicbrainz-server/pull/919/files#diff-37e659b91af89773d96c6b5faa90d18cR71

reosarevok: Yeah, my mistake, this was probably added to the file after I had written the code changing this and I missed it. Fixed!

root/static/scripts/place/map.js
```diff
@@ -4,7 +4,9 @@
 // http://www.gnu.org/licenses/gpl-2.0.txt
 
 const {createMap, L} = require('../common/leaflet');
-const {draggable, place, title} = require('../common/utility/getScriptArgs')();
+import getScriptArgs from '../common/utility/getScriptArgs';
```
mwiencek: The presence of an import statement here causes Babel to determine that it's an ES module, which breaks the `exports.foo` assignments below (since the `exports` object is only defined in CommonJS environments). If you load the places map page you'll see those throw an exception.

The solution is to revert this or convert all the exports to ES export statements.

root/static/scripts/common/MB.js
```diff
@@ -39,4 +39,4 @@ const MB: {[string]: *} = {
 
 global.MB = MB;
 
-module.exports = MB;
+export default MB;
```
mwiencek: There's a `require` in root/static/scripts/common/MB/edit_search.js still which doesn't have `.default` added.

root/static/scripts/area/places-map.js
```diff
@@ -12,7 +12,9 @@ const ReactDOMServer = require('react-dom/server');
 const EntityLink = require('../common/components/EntityLink');
```
mwiencek: Needs `.default`

root/static/scripts/common/utility/debounce.js
```diff
@@ -14,4 +14,4 @@ function debounce(value, delay) {
     });
 }
 
-module.exports = debounce;
+export default debounce;
```
mwiencek: debounce require in root/static/scripts/timeline.js needs `.default`

root/static/scripts/common/utility/isBlank.js
```diff
@@ -1,3 +1,3 @@
-module.exports = function isBlank(str){
+export default function isBlank(str) {
```
mwiencek: isBlank require in root/static/scripts/place.js needs `.default`

root/static/scripts/common/utility/request.js
```diff
@@ -68,4 +68,4 @@ function request(args, context) {
     return promise;
 }
 
-module.exports = request;
+export default request;
```
mwiencek: request require in root/static/scripts/common/components/TagEditor.js needs `.default`

PR #920, opened 2019-02-18
Run yarn in compile_resources.sh

mwiencek:
This seems like a good idea, in case people forget to do it when updating their git checkout or switching branches. I thought about this after merging the Webpack PR.

If node_modules/ is up-to-date and this command is a no-op, it only takes about ~0.5s to run on my machine.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -184,21 +184,20 @@ Below outlines how to setup MusicBrainz server with local::lib.
 
     (Do not overlook the dot at the end of that command.)
 
-Installing Node.js dependencies
+Building static web resources
 -------------------------------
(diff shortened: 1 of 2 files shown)


mwiencek:
> MB VM was used to `--production` flag for `yarn install`.

Would it be possible to set `NODE_ENV` there? https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production-true-false

mwiencek:
> Yes, thanks, might be worth a note about that in `INSTALL.md`.

Updated INSTALL.md with a suggestion to use `NODE_ENV=production`.

PR #922, opened 2019-02-18
Bump some JS dependencies

yvanzo:
* shell-quote ([History](https://github.com/substack/node-shell-quote/compare/1.4.3...1.6.1)): from `1.4.3` to `1.6.1`
* Lodash ([Changelog](https://github.com/lodash/lodash/wiki/Changelog)): from `4.17.4` to `4.17.11`
* React ([Releases](https://github.com/facebook/react/releases)): from `16.7.0` to `16.8.2`
* React DOM ([Releases](https://github.com/facebook/react/releases)): from `16.3.2` to `16.8.2`

Seems to have no breaking change so far.


mwiencek:
Thanks (I noticed I forgot to update react-dom recently).

We should probably just update both react & react-dom to 16.8.2 while we're at it - there's no breaking changes compared to 16.7 and it has [Hooks](https://reactjs.org/blog/2019/02/06/react-v16.8.0.html) which are pretty cool and we'll likely want to use for new stateful components.

PR #923, opened 2019-02-19
MBS-9904: Convert list of privileged users to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9904


reosarevok:
It's just weirdly inconsistent though, might make translators wonder whether there's supposed to be a difference and they just can't tell. But yeah, it's the least ideal of them, so I can drop that if you guys want me to :)

mwiencek:
I agree it's weirdly inconsistent, though I doubt anyone will be confused very much by it, and I think breaking the translation bothers me slightly more than the inconsistency. :)

root/server/components.js
```diff
@@ -191,6 +192,5 @@ module.exports = {
   'static/scripts/edit/components/InformationIcon': require('../static/scripts/edit/components/InformationIcon'),
   'static/scripts/edit/components/WarningIcon': require('../static/scripts/edit/components/WarningIcon'),
   'url/UrlHeader': require('../url/UrlHeader'),
-  'user/components/UserInlineList': require('../user/components/UserInlineList'),
```
mwiencek: Thanks for remembering to remove that :)

PR #925, opened 2019-02-20
Add @flow to ArtistCreditLink and fix accordingly

reosarevok:


root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -1,38 +1,53 @@
...
   for (let i = 0; i < artistCredit.length; i++) {
     const credit = artistCredit[i];
     if (props.plain) {
       parts.push(credit.name);
     } else {
-      let artist = credit.artist;
-      if (!artist) {
-        artist = {entityType: 'artist', name: credit.name};
+      const artist = credit.artist;
+      if (artist) {
```
reosarevok: Just switched around because of no-negative-conds.

root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -1,38 +1,53 @@
...
+            content={credit.name}
+            entity={artist}
+            key={i}
+            showDeleted={showDeleted}
+            target={props.target}
+          />,
+        );
+      } else {
+        parts.push(
+          <DeletedLink allowNew={false} name={credit.name} />,
```
reosarevok: Instead of sending a "fake artist" to EntityLink (which was a flow mess), and since we just get a DeletedLink in this case anyway (because gid was missing), we might as well just export DeletedLink and use it here.

reosarevok:
This fails but seems to work fine if rebased with https://github.com/metabrainz/musicbrainz-server/pull/919 so we can look at this further once that's merged (probably no point fiddling with imports etc here when that PR is already open).

PR #926, opened 2019-02-20
MBS-10042: Port all Aliases tabs to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10042


root/components/Aliases/index.js
```diff
@@ -15,14 +15,30 @@ import EntityLink from '../../static/scripts/common/components/EntityLink';
...
+        return !$c.user.is_editing_disabled;
+    }
+  }
+  return false;
+}
+
 type Props = {
   +$c: CatalystContextT,
   +aliases: $ReadOnlyArray<AliasT>,
   +allowEditing?: boolean,
```
mwiencek: We can remove the `allowEditing` prop from the type too.

root/components/Aliases/index.js
```diff
@@ -15,14 +15,30 @@ import EntityLink from '../../static/scripts/common/components/EntityLink';
...
 
+function canEdit($c: CatalystContextT, entityType: string) {
+  if ($c.user) {
+    switch (entityType) {
+      case 'area':
+        return $c.user.is_location_editor;
+      case 'instrument':
+        return $c.user.is_relationship_editor;
+      default:
+        return !$c.user.is_editing_disabled;
```
mwiencek: I think it'd be more logically correct to `return true;` here and change the `if` to `if ($c.user && !$c.user.is_editing_disabled) {`.

reosarevok:
Changes made, plus added a small extra commit fixing too long lines in the alias components

reosarevok:
Updated - take a quick look to confirm before I merge maybe? :)

yvanzo:
Confirmed!

PR #895, opened 2019-02-09
MBS-10026: Convert the Labels page for Area to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10026

PR #927, opened 2019-02-20
MBS-10026 fixup

reosarevok:
https://github.com/metabrainz/musicbrainz-server/pull/895 got merged before I could add the things I found out during https://github.com/metabrainz/musicbrainz-server/pull/894 - so doing it here. The sortable bit isn't too important, but as it is label collections are just broken in master, so let's get this merged before we release any of this? :)

### DIFF
--- a/root/components/LabelsList.js
+++ b/root/components/LabelsList.js
@@ -14,7 +14,8 @@ import {lp_attributes} from '../static/scripts/common/i18n/attributes';
 import {withCatalystContext} from '../context';
 import formatLabelCode from '../utility/formatLabelCode';
 import loopParity from '../utility/loopParity';
-import DescriptiveLink from '../static/scripts/common/components/DescriptiveLink';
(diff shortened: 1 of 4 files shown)


root/components/LabelsList.js
```diff
@@ -33,19 +38,71 @@ const LabelsList = ({
...
-        <th>{l('Label')}</th>
-        <th>{l('Type')}</th>
-        <th>{l('Code')}</th>
+        <th>
+          {sortable
+            ? (
+              <SortableTableHeader
+                label={l('Label')}
+                name="name"
+                order={order ? order : null}
```
mwiencek: `order || null` is a bit more to-the-point though we should prob just be able to pass `order` and have SortableTableHeader handle undefined

reosarevok: Added an extra commit for that

PR #931, opened 2019-02-21
MBS-9909: Change reports to use React (12 - places)

reosarevok:
To be reviewed and merged after https://github.com/metabrainz/musicbrainz-server/pull/930

PR #932, opened 2019-02-21
MBS-9909: Change reports to use React (13 - recordings)

reosarevok:
To be reviewed and merged after https://github.com/metabrainz/musicbrainz-server/pull/931


mwiencek:
Was it intentional to remove the Artist MBID column from the "Recordings with the same name by different artists with the same name" report? It at least seems somewhat useful where none of the entities have disambiguation comments (so you can see where the artist changes in the list).

root/report/RecordingsSameNameDifferentArtistsSameName.js
```diff
@@ -0,0 +1,93 @@
...
+
+    <ul>
+      <li>
+        {l(`This report shows all recordings with the same name that have
+            different artists (having different MBIDs) with the same name.`)}
+      </li>
+      <li>
+        {exp.l(`These are most likely cases where the {ac|artist credit} is
+            incorrect for at least one of the recordings.`,
+        {ac: '/doc/Artist_Credits'})}
```
mwiencek: Indentation looks a bit wonky on these two lines.

reosarevok: Indentation is all fixed in one go in PR 15

reosarevok:
Only one artist doesn't have a disambiguation though (unless something has gone wrong, anyway) since we force entering one after that. So there's no actual benefit to that column.

mwiencek:
Oh, duh. The old page didn't even display the comments.

PR #933, opened 2019-02-21
MBS-9909: Change reports to use React (14 - releases)

reosarevok:
To be reviewed and merged after https://github.com/metabrainz/musicbrainz-server/pull/932


yvanzo:
Seriously, please split it up, one template one commit, more or less. :scissors:

root/report/AsinsWithMultipleReleases.js
```diff
@@ -0,0 +1,55 @@
...
+  items,
+  pager,
+}: ReportDataT<ReportReleaseUrlT>) => (
+  <Layout fullWidth title={l('Amazon URLs linked to multiple releases')}>
+    <h1>{l('Amazon URLs linked to multiple releases')}</h1>
+
+    <ul>
+      <li>
+        {exp.l(`This report shows Amazon URLs which are linked to multiple
+            releases. In most cases Amazon ASINs should map to MusicBrainz
```
yvanzo: `releases` should be vertically aligned with `This`, and so on for the following lines, happens in some other files where `exp.l` is used too.

root/report/AsinsWithMultipleReleases.js
```diff
@@ -0,0 +1,55 @@
...
+            releases 1:1, so only one of the links will be correct. Just
+            check which MusicBrainz release fits the release in Amazon (look
+            at the format, tracklist, etc). If the release has a barcode,
+            you can also search Amazon for it and see which ASIN matches.
+            You might also find some ASINs linked to several discs
+            of a multi-disc release: just merge those (see
+            {how_to_merge_releases|How to Merge Releases}).`,
+        {how_to_merge_releases: '/doc/How_to_Merge_Releases'})}
```
yvanzo: Unexpected indentation, happens in some other files too, checked with `eslint`.

reosarevok: As in the previous PR, these are eslint changes from after this code was written and affect all reports code, so they're being fixed in the PR 15 :)

yvanzo: Still feels weird to commit odd code and then to merge another commit to get it right. :p

mwiencek:
Sorry for the delay here too! Took a bit longer to check than the other ones.

Are we squashing this then? (I'd feel bad to since it looks like a lot of work was put into splitting it, but I don't know how hard it'd be to further fix the commits.)

reosarevok:
It was only split for ease of reviewing anyway - IMO it makes sense to squash it so it's the same as the other stages of the ticket now (except MBS-9909: Consistent casing of initials which is separate)

PR #934, opened 2019-02-21
MBS-9909: Change reports to use React (15 - final improvements)

reosarevok:
To be reviewed and merged after https://github.com/metabrainz/musicbrainz-server/pull/933

After this, https://tickets.metabrainz.org/browse/MBS-9909 should be ready for closing.

Interestingly, eslint didn't detect these lines, even though many were clearly too long. Is there a second rule we can apply for this that will check specifically for too long jsx lines?

### DIFF
--- a/root/report/DuplicateArtists.js
+++ b/root/report/DuplicateArtists.js
@@ -52,8 +52,14 @@ const DuplicateArtists = ({
             },
           )}
         </li>
-        <li>{texp.l('Total duplicate groups: {count}', {count: pager.total_entries})}</li>
-        <li>{texp.l('Generated on {date}', {date: formatUserDate($c.user, generated)})}</li>
+        <li>
+          {texp.l('Total duplicate groups: {count}',
+                  {count: pager.total_entries})}
+        </li>
+        <li>
+          {texp.l('Generated on {date}',
+                  {date: formatUserDate($c.user, generated)})}
+        </li>
 
         {canBeFiltered ? <FilterLink filtered={filtered} /> : null}
       </ul>
--- a/root/report/components/UrlRelationshipList.js
+++ b/root/report/components/UrlRelationshipList.js
@@ -32,7 +32,9 @@ const UrlRelationshipList = ({
         {items.map((item, index) => (
           <tr className={loopParity(index)} key={item.url.gid}>
             <td>
-              <a href={'/relationship/' + item.link_gid}>{l_relationships(item.link_name)}</a>
+              <a href={'/relationship/' + encodeURIComponent(item.link_gid)}>
+                {l_relationships(item.link_name)}
+              </a>
             </td>
             <td>
               <EntityLink entity={item.url} />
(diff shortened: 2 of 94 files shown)


root/report/components/UrlRelationshipList.js
```diff
@@ -32,7 +32,9 @@ const UrlRelationshipList = ({
         {items.map((item, index) => (
           <tr className={loopParity(index)} key={item.url.gid}>
             <td>
-              <a href={'/relationship/' + item.link_gid}>{l_relationships(item.link_name)}</a>
+              <a href={'/relationship/' + encodeURIComponent(item.link_gid)}>
```
reosarevok: This was missing so I added it here :)

root/report/DuplicateArtists.js
```diff
@@ -38,22 +38,26 @@ const DuplicateArtists = ({
...
+                  names. If two artists are actually the same, please merge
+                  them (remember to {how_to_write_edit_notes|write an edit
+                  note} and give your proof). If they\'re different, add
+                  {disambiguation_comment|disambiguation comments} to them
+                  (and once a group of similarly named artists have
+                  disambiguation comments, they will stop appearing here).`,
+                 {
+                   disambiguation_comment: '/doc/Disambiguation_Comment',
+                   how_to_write_edit_notes: '/doc/How_to_Write_Edit_Notes',
+                 })}
```
mwiencek: There are a few changes like this, but I'm not sure I'd call them improvements. I think both of these styles are good depending on the circumstances, e.g. the changes to the two strings below look good. But here, we'd just be adding 5 extra spaces of indentation, giving these very long lines even less space. The style on the left works well for avoiding too much indentation when you have really long arguments to fit.

I also like how the one on the left keeps substitutions like `{how_to_write_edit_notes|write an edit note}` on a single line. It's a really small thing, but makes it easier to read the text and parse the syntax.

reosarevok: @yvanzo: do you want to revisit this (since it's your commit) or should I? :)

yvanzo: @reosarevok: I just pushed three more commits to address this, normalizing indentation the suggested way instead, should probably be squashed within https://github.com/metabrainz/musicbrainz-server/pull/934/commits/ae2c3a78c0475ff75bfd2633ecf7b7df715967c1 once accepted.

I agree with _avoiding too much indentation when you have really long arguments to fit_. Should starting multi-line message on a new line be added to MBS coding style?

@mwiencek: Does breaking line in the middle of substitution defeat the current parser? (I don’t really find it any different to read.) Could that be fixed in the parser or should messages be automatically tested?

mwiencek: It doesn't break the parser if the line breaks come after the `{var|` part, no. As for whether it's easier to read, I guess it's just me, so never mind. :)

reosarevok: I've generally tried to keep those in one line, although sometimes they're so long they won't even fit anyway, and then I've just split them because it's not that important IMO.

yvanzo: Ok, no worry, I did not break any line within substitution here anyway.

PR #935, opened 2019-02-21
Call tryLoadDomain on every string

mwiencek:
If the same renderer process handles two requests in two different languages, the `wrapGettext` instance won't call `tryLoadDomain` for one of the languages. The call must be moved so that it's invoked for every string, as it was before 77a1286.


mwiencek:
This fixes some ``Domain `foo` was not found`` exceptions on some pages @reosarevok was converting

PR #936, opened 2019-02-21
MBS-6574: Recursive area tabs

reosarevok:
https://tickets.metabrainz.org/browse/MBS-6574


lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -101,14 +101,21 @@ sub find_by_subscribed_editor
...
     my ($self, $area_id, $limit, $offset) = @_;
+    my (
+        $containment_query,
+        @containment_query_args,
+    ) = $self->c->model('Area')->get_containment_query('$2', 'any(array[area, begin_area, end_area])');
     my $query = "SELECT " . $self->_columns . "
                  FROM " . $self->_table . "
-                    LEFT JOIN area ON artist.area = area.id
-                    LEFT JOIN area begin_area ON artist.begin_area = begin_area.id
-                    LEFT JOIN area end_area ON artist.end_area = end_area.id
```
reosarevok: All these joins seem completely unnecessary - the query does nothing with the area tables after this.

reosarevok:
Since the label table was already converted, I added the areas to it here already. Made it sortable since it makes sense IMO to be able to see all labels from the same area together (even if the sort is kinda stupid since it's not by name but by id). While doing that I also realized the containment wasn't being loaded for collections, so I added that in a separate commit.

PR #937, opened 2019-02-22
Replace _.assign with Object.assign/spreads

mwiencek:
Removing unnecessary uses of lodash. This doesn't help with our bundle sizes yet (we'll have to kill lodash chaining first).

`_.extend` differs in that it enumerates over inherited properties, but there was no case where we needed or wanted that.

Note: We already have an `Object.assign` polyfill for IE.

### DIFF
--- a/root/components/relationship-editor.tt
+++ b/root/components/relationship-editor.tt
@@ -140,7 +140,7 @@
     <td data-bind="foreach: relationship().linkTypeAttributes()">
       <!-- ko template: {
                 name: "template.edit-attribute",
-                data: _.assign({ relationship: $parent.relationship() }, $data)
+                data: Object.assign({ relationship: $parent.relationship() }, $data)
               }
         --><!-- /ko -->
     </td>
--- a/root/static/scripts/common/entity.js
+++ b/root/static/scripts/common/entity.js
@@ -34,7 +34,7 @@ import formatTrackLength from './utility/formatTrackLength';
     class Entity {
 
         constructor(data) {
-            _.assign(this, data);
+            Object.assign(this, data);
             this.name = this.name || "";
         }
 
@@ -283,7 +283,7 @@ import formatTrackLength from './utility/formatTrackLength';
         }
 
         toJSON() {
-            return _.assign(super.toJSON(), { isrcs: this.isrcs, appearsOn: this.appearsOn });
+            return Object.assign(super.toJSON(), { isrcs: this.isrcs, appearsOn: this.appearsOn });
         }
     }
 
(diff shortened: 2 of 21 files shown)


root/static/scripts/release-editor/edits.js
```diff
@@ -78,7 +78,7 @@ releaseEditor.edits = {
             edits.push(MB.edit.releaseCreate(newData));
         }
         else if (!_.isEqual(newData, oldData)) {
-            newData = _.extend(_.clone(newData), { to_edit: release.gid() });
+            newData = Object.assign({}, newData, { to_edit: release.gid() });
```
mwiencek: Instead of using `_.clone`, I made it assign to an empty object, which has the same effect (`_.clone` creates a shallow clone).

reosarevok:
Is "kill lodash chaining" the same as "import only what we specifically need from lodash instead of _" ?

mwiencek:
> Is "kill lodash chaining" the same as "import only what we specifically need from lodash instead of _" ?

It's related to that.

By chaining I mean lodash's method-chaining feature, where you can invoke things like `_([1, 2, 3]).map(x => x + 1).filter(x => x % 2).value()`. Since that allows you to invoke (almost all) lodash methods without importing them, it necessitates importing *all* of lodash. We'd fix that by doing `_.filter(_.map([1, 2, 3], x => x + 1), x => x % 2)` instead.

Re: "import only what we specifically need," that's the more general idea, though I've found ways we can achieve that without changing our imports:

1. We can use [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash), which will split the imports for us.
2. We can use [lodash-es](https://www.npmjs.com/package/lodash-es) which apparently works with [tree shaking](https://webpack.js.org/guides/tree-shaking/) in Webpack. Meaning, Webpack would be able to detect lodash code we don't use, and eliminate it.

However, neither of those options will work while we use method chaining.

mwiencek:
Rebased and edited the title to indicate I'm also using object spreads where appropriate. If you're wondering about any technical differences between the two, see http://2ality.com/2016/10/rest-spread-properties.html#spreading-objects-versus-objectassign (read: there is none for us, because we don't use setters or read-only properties anywhere).

root/components/relationship-editor.tt
```diff
@@ -140,7 +140,7 @@
     <td data-bind="foreach: relationship().linkTypeAttributes()">
       <!-- ko template: {
                 name: "template.edit-attribute",
-                data: _.assign({ relationship: $parent.relationship() }, $data)
+                data: Object.assign({ relationship: $parent.relationship() }, $data)
```
mwiencek: (I didn't use an object spread here because knockout code isn't transpiled, of course.)

root/static/scripts/common/entity.js
```diff
@@ -34,7 +34,7 @@ import formatTrackLength from './utility/formatTrackLength';
     class Entity {
 
         constructor(data) {
-            _.assign(this, data);
+            Object.assign(this, data);
```
mwiencek: As here, I should mainly only be using `Object.assign` where we want to assign into an existing object (though I may have missed some, it's not a super important distinction).

PR #938, opened 2019-02-22
Remove explicit calls to _.chain

mwiencek:
Method chaining is bad because it prevents use from being able to remove unused lodash code.

This commit doesn't remove implicit chaining using `_()`, which is much more common.

### DIFF
--- a/root/static/scripts/edit/ExampleRelationships.js
+++ b/root/static/scripts/edit/ExampleRelationships.js
@@ -35,9 +35,9 @@ ERE.init = function (config) {
     });
     ERE.viewModel.selectedEntityType.subscribe(autocomplete.changeEntity);
     ERE.viewModel.availableEntityTypes(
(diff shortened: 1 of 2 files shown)


yvanzo:
@mwiencek: An alternative would be to use [lodash/fp](https://github.com/lodash/lodash/wiki/FP-Guide), seems worth it.

mwiencek:
> @mwiencek: An alternative would be to use lodash/fp, seems worth it.

Yeah I definitely agree lodash/fp or terable (for iterable support) are good for complex chains and utilities like groupBy/union/etc. Though I think native JS methods often make more sense for simple maps/filters/folds since they require no imports or added code and accomplish the same thing.

Sometimes these long method chains are not exactly ideal: for example, a `.filter(...).map(...).flatten()` chain iterates the result 3 separate times creating multiple intermediary arrays, but you could accomplish the same thing with a single `reduce`. And JS has a built-in `reduce` function which works great.

PR #940, opened 2019-02-23
MBS-10047: Convert /oauth2/ pages to React

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10047


root/oauth2/Oob.js
```diff
@@ -0,0 +1,37 @@
...
+        `You have granted access to {app}. Next, return to {app} and
+         copy this token to complete the authorization process:`,
+        {app: application.name},
+      )}
+    </p>
+    <p><code>{code}</code></p>
+  </Layout>
+);
+
+export default OAuth2Oob;
```
reosarevok: Why not call the file OAuth2Oob.js? Same for the others, especially Oauth2Error

root/oauth2/Authorize.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+type Props = {|
+  +$c: CatalystContextT,
+  +application: ApplicationT,
+  +offline: boolean,
+  +permissions: $ReadOnlyArray<number>,
+|};
+
+const Oauth2Authorize = withCatalystContext(({
```
reosarevok: Is there any benefit to using withCatalystContext here instead in the export of as we usually do?

root/oauth2/Authorize.js
```diff
@@ -0,0 +1,75 @@
...
+import {l, N_l} from '../static/scripts/common/i18n';
+
+const ACCESS_SCOPE_PERMISSIONS = {
+  [ACCESS_SCOPE_COLLECTION]: N_l('View and modify your private collections'),
+  [ACCESS_SCOPE_EMAIL]: N_l('View your email address'),
+  [ACCESS_SCOPE_PROFILE]: N_l('View your public account information'),
+  [ACCESS_SCOPE_RATING]: N_l('View and modify your private ratings'),
+  [ACCESS_SCOPE_SUBMIT_BARCODE]: N_l('Submit new barcodes to the database'),
+  [ACCESS_SCOPE_SUBMIT_ISRC]: N_l('Submit new ISRCs to the database'),
+  [ACCESS_SCOPE_TAG]: N_l('View and modify your private tags'),
```
reosarevok: Probably stupid question, but: why can't this just be l()?

root/oauth2/Oob.js
mwiencek: Hm, I guess we're not very consistent about that, e.g. all the files named `/Index`, `/List`, `/Results`, `/NotFound` etc. that don't repeat the entity/subpath name. Not sure which I prefer. I guess it's nice if the file name matches the component name, but shorter/non-repetitive paths are nice too and it doesn't change the name you import by

root/oauth2/Authorize.js
mwiencek: Nope, since that way is more common I'll change it for consistency

root/oauth2/Authorize.js
mwiencek: Because these statements are executed when the file is required and loaded, not when the component renders. `l()` return a translated string for the current locale, but the locale when the worker process starts is probably English, so `l()` here would just always return an English string.

(In case it's unclear, modules aren't reloaded every time you require() them within the same process - they're executed once and then cached, so if you require() the same file again it return the same reference back.)

root/oauth2/Oob.js
reosarevok: I think I would rather have unique names when possible, but maybe that's just me.

root/oauth2/Authorize.js
mwiencek: I'll point out though that if you have `DEVELOPMENT_SERVER` enabled, we [explicitly clear the module cache](https://github.com/metabrainz/musicbrainz-server/blob/a6d9d1ed9aaca7999a8fc61b62cb52917deb9c37/root/server/createServer.js#L68) after every request, so the module *is* reloaded and evaluated every time you load the page in that case. :) (So you wouldn't observe any translation bug, but it would happen in production.)

root/oauth2/Oob.js
mwiencek: Fixed to use the full component name in the file names.

mwiencek:
@reosarevok @yvanzo ping! (I think this is the oldest PR not blocked by anything.)

root/constants.js
```diff
@@ -7,6 +7,24 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+export const ACCESS_SCOPE_PROFILE: 1 = 1;
+export const ACCESS_SCOPE_EMAIL: 2 = 2;
+export const ACCESS_SCOPE_TAG: 4 = 4;
+export const ACCESS_SCOPE_RATING: 8 = 8;
+export const ACCESS_SCOPE_COLLECTION: 16 = 16;
+export const ACCESS_SCOPE_SUBMIT_ISRC: 64 = 64;
+export const ACCESS_SCOPE_SUBMIT_BARCODE: 128 = 128;
```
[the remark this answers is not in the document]
mwiencek: (Not stupid at all.) If you do `const x = 1` Flow will see the type of x as `number`, whereas in `const x: 1 = 1` it'll see it as the number literal `1`.

Here's an example where it could conceivably make a difference: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzAEEBhEgUQGVKB9SkgeQAVyaAlIgFQEkA5AcQBcYABxgAvKIDcmHAWJkqtes1aMAMuvIkeDXsICMANglhjM1PgCeABwCmCitTqMWlCajBgAPmGv24KEclF1V2Lj5+Tx8-WztA4OcVFhoNLR1uPWiLLDxCKGxTAAoAQ1xhUidlVyoASgkAPjAAV2wAEzsoAEtsOzaLAqKDACYAZlqZIA (if you remove the `: 8` or `: 16` bits, the type of `ACCESS_SCOPES` is just `number` and `fn` accepts anything.)

root/oauth2/OAuth2Error.js
```diff
@@ -0,0 +1,30 @@
...
+  errorMessage,
+}: Props) => (
+  <Layout fullWidth title={l('OAuth Authorization Error')}>
+    <h1>{texp.l('Error: {error}', {error: errorMessage})}</h1>
+    <p>{errorDescription}</p>
+    <p>{exp.l('{doc|Learn more}', {doc: '/doc/Development/OAuth2'})}</p>
+  </Layout>
+);
+
+export default Oauth2Error;
```
yvanzo: File name `OAuth2Error` and default component name `Oauth2Error` are inconsistent about capitalization. `OAuth2` seems to be more frequently used.

root/oauth2/OAuth2Authorize.js
```diff
@@ -0,0 +1,61 @@
...
+        </button>
+        <button name="confirm.submit" type="submit" value="1">
+          {l('Allow access')}
+        </button>
+      </span>
+    </form>
+  </Layout>
+);
+
+export default withCatalystContext(Oauth2Authorize);
```
yvanzo: Not the same capitalization as in the file name.

root/server/components.js
```diff
@@ -76,6 +76,9 @@ module.exports = {
   'main/404': require('../main/404'),
   'main/index': require('../main/index'),
   'otherlookup/NotFound': require('../otherlookup/NotFound'),
+  'oauth2/OAuth2Authorize': require('../oauth2/OAuth2Authorize'),
+  'oauth2/OAuth2Error': require('../oauth2/OAuth2Error'),
+  'oauth2/OAuth2Oob': require('../oauth2/OAuth2Oob'),
```
yvanzo: Should come before the above `'otherlookup/NotFound'` export.

root/account/applications/Index.js
```diff
@@ -51,7 +52,9 @@ const buildTokenRow = (token: EditorOAuthTokenT, index: number) => (
 );
 
 function formatScopes(token: EditorOAuthTokenT) {
-  const lScopes = token.permissions.map(perm => l(perm));
+  const lScopes = token.permissions.map(
+    perm => ACCESS_SCOPE_PERMISSIONS[perm]()
```
yvanzo:
```suggestion
    perm => ACCESS_SCOPE_PERMISSIONS[perm](),
```
makes eslint happy.

root/constants.js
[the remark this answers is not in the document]
yvanzo: I still don’t get why Flow needs annotation here, can it not deduce that `x` will be `1` forever?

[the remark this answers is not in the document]
mwiencek: Not currently. There are some open issues for it, e.g. https://github.com/facebook/flow/issues/2639 and https://github.com/facebook/flow/issues/3439.

PR #944, opened 2019-02-23
MBS-7463: Add redirect from /tag to /tags

reosarevok:
https://tickets.metabrainz.org/browse/MBS-7463

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Tag.pm
+++ b/lib/MusicBrainz/Server/Controller/Tag.pm
@@ -21,6 +21,16 @@ sub _load
     return $c->model('Tag')->get_by_name($name);
 }
 
+sub no_tag_provided : Path('/tag') Args(0)
+{
+    my ($self, $c, $name) = @_;
+
+    # If no tag is passed, redirect to the tag cloud
+    $c->response->redirect(
+        $c->uri_for_action('/tag/cloud'));
+    $c->detach;
+}
+
 sub cloud : Path('/tags')
 {
     my ($self, $c, $name) = @_;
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/Tag.pm
```diff
@@ -21,6 +21,16 @@ sub _load
...
 }
 
+sub no_tag_provided : Path('/tag')
+{
+    my ($self, $c, $name) = @_;
+
+    # If no tag is passed, redirect to the tag cloud
+    $c->response->redirect(
+        $c->uri_for_action('/tag/cloud'));
+    $c->detach;
```
mwiencek: You should be able to just do `$c->detach('/tag/cloud');` here and remove the above line.

lib/MusicBrainz/Server/Controller/Tag.pm
```diff
@@ -21,6 +21,16 @@ sub _load
     return $c->model('Tag')->get_by_name($name);
 }
 
+sub no_tag_provided : Path('/tag')
```
mwiencek: I think `Args(0)` is implied there (I guess it must be if it works, but can't find any relevant docs), but just for clarity it'd be nice to add `Args(0)` after the `Path` part to indicate it's the end of the path.

lib/MusicBrainz/Server/Controller/Tag.pm
mwiencek: Never mind, it's definitely more useful if it performs a redirect.

lib/MusicBrainz/Server/Controller/Tag.pm
reosarevok: I think Path only works for the full path, otherwise it'd be PathPart?

reosarevok: Or actually maybe not, since /tags/whatever leads to the tag cloud (wonder if it should actually redirect to /tag/whatever, heh). Ok, I'll change it.

PR #945, opened 2019-02-23
MBS-9013: Removing specific ordering of relationships doesn't create an edit

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9013

### DIFF
--- a/root/static/scripts/relationship-editor/generic.js
+++ b/root/static/scripts/relationship-editor/generic.js
@@ -10,6 +10,7 @@ import _ from 'lodash';
 import {SERIES_ORDERING_TYPE_AUTOMATIC} from '../common/constants';
 import MB from '../common/MB';
 import formatDate from '../common/utility/formatDate';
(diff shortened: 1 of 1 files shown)


root/static/scripts/relationship-editor/generic.js
```diff
@@ -203,7 +203,10 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
             pushInput(prefix, "link_type_id", editData.linkTypeID || "");
 
             if (relationship.getLinkType().orderable_direction !== 0) {
-                if (relationship.added() || changeData.linkOrder) {
+                if (relationship.added() ||
+                    changeData.linkOrder ||
+                    // Otherwise 0 is falsy and order can't be removed
+                    changeData.linkOrder === 0) {
```
mwiencek: Nice catch. I've run into this a few times and generally use root/static/scripts/common/utility/nonEmpty.js (it simply checks that the value isn't null, undefined, or an empty string), so that might be ideal here.

PR #946, opened 2019-02-23
MBS-9990: Make formatUserDate check it's being passed a valid date

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9990

https://momentjs.com/guides/#/parsing/strict-mode/ (moment.defaultFormat is YYYY-MM-DDTHH:mm:ssZ, https://momentjs.com/docs/#/displaying/format/)

If not being passed a valid date, it will just say Invalid date. Is there anywhere where we're passing this function anything else than a YYYY-MM-DDTHH:mm:ssZ format? (multiple formats can be allowed, if needed)

### DIFF
--- a/root/elections/ElectionDetails.js
+++ b/root/elections/ElectionDetails.js
@@ -77,7 +77,11 @@ const ElectionDetails = ({election, user}: PropsT) => (
           {election.is_open
             ? (
               lp(election.status_name, 'autoeditor election status', {
-                date: formatUserDate(user, election.open_time),
+                date: (
+                  election.open_time
+                    ? formatUserDate(user, election.open_time)
+                    : null
+                ),
               })
             ) : null}
 
--- a/root/utility/formatUserDate.js
+++ b/root/utility/formatUserDate.js
@@ -11,7 +11,7 @@ import 'moment-timezone';
 
 function formatUserDate(user, dateString, options) {
   let preferences = user ? user.preferences : null;
-  let result = moment(dateString);
+  let result = moment(dateString, moment.defaultFormat, true);
   let format = '%Y-%m-%d %H:%M %Z';
 
   if (preferences) {
--- a/root/utility/formatUserDate.js
+++ b/root/utility/formatUserDate.js
@@ -1,7 +1,10 @@
-// This file is part of MusicBrainz, the open internet music database.
-// Copyright (C) 2015 MetaBrainz Foundation
-// Licensed under the GPL version 2, or (at your option) any later version:
-// http://www.gnu.org/licenses/gpl-2.0.txt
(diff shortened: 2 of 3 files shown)


root/elections/ElectionDetails.js
```diff
@@ -77,7 +77,11 @@ const ElectionDetails = ({election, user}: PropsT) => (
           {election.is_open
             ? (
               lp(election.status_name, 'autoeditor election status', {
-                date: formatUserDate(user, election.open_time),
+                date: (
+                  election.open_time
+                    ? formatUserDate(user, election.open_time)
+                    : l('an unknown date')
+                ),
```
reosarevok: I'm not sure there's ever a case where election.open_time doesn't exist if there's election.is_open, but I guess it doesn't hurt to make sure anyway? (it's optional in types.js)

mwiencek: Yeah, it'll always be defined if `is_open` is true (it's only undefined if still waiting for seconders). I think there's a way to express that to Flow using intersection/union types, but having the check seems fine for now. I'd probably output `null` in the else case instead of making people translate a string that'll never be displayed though. :)

reosarevok: I just feel it's weird to even have the chance to display "Voting open since " (and nothing else), but I guess it if never happens that's ok. Just strange :)

mwiencek: That's the thing, there is no chance. :P It's easy to prove that [on the server](https://github.com/metabrainz/musicbrainz-server/blob/962c5c01e64988b8197eb6ea1325bec6af69ff7f/lib/MusicBrainz/Server/Data/AutoEditorElection.pm#L102-L104) but we just need to convince Flow of it somehow later.

reosarevok: Ok then, updated.

root/utility/formatUserDate.js
```diff
@@ -1,17 +1,24 @@
...
 import 'moment-strftime';
 import 'moment-timezone';
 
-function formatUserDate(user, dateString, options) {
-  let preferences = user ? user.preferences : null;
-  let result = moment(dateString);
+function formatUserDate(
+  user: ?(CatalystUserT | SanitizedEditorT),
+  dateString: string,
+  options?: {[string]: boolean},
```
mwiencek: Can we replace the indexer with the specific possible options for a little extra type safety? e.g. `{dateOnly?: boolean}` would catch an error if you accidentally passed `daetOnly`.

reosarevok: Done - dateOnly seems to be the only one we have at the moment anyway, so :)

PR #948, opened 2019-02-25
MBS-10048: Add Musixmatch album pages to lyrics whitelist

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10048

I'm a bit confused about why we didn't allow them originally, given we even had a test to ensure they were *not* allowed so we clearly knew they were there! But anyway :)


reosarevok:
It is, but I don't think that's necessarily a reason to block it, especially since it seems to be a very popular service for some reason.

mwiencek:
Reading that community thread, the conclusion seems to be that we probably shouldn't link them. Can this be closed, then?

reosarevok:
I guess so. Oh well.

PR #950, opened 2019-02-26
Eslint fixes for search/

reosarevok:
While checking my old code for examples of form implementation I figured out there were a bunch of unused vars and stuff. This fixes that and makes eslint happy.

### DIFF
--- a/root/search/components/RecordingResults.js
+++ b/root/search/components/RecordingResults.js
@@ -15,7 +15,7 @@ import EntityLink from '../../static/scripts/common/components/EntityLink';
 import TaggerIcon from '../../static/scripts/common/components/TaggerIcon';
 import formatTrackLength from '../../static/scripts/common/utility/formatTrackLength';
 import loopParity from '../../utility/loopParity';
-import type {InlineResultsPropsT, ResultsPropsT} from '../types';
+import type {InlineResultsPropsT, ResultsPropsWithContextT} from '../types';
 import ArtistCreditLink
   from '../../static/scripts/common/components/ArtistCreditLink';
 import CodeLink from '../../static/scripts/common/components/CodeLink';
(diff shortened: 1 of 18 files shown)


root/search/components/WorkResults.js
```diff
@@ -58,7 +58,7 @@ const WorkResults = ({
       query={query}
       results={results}
     />
-    {$c.user && !$c.user.is_editing_disabled ? (
+    {$c && $c.user && !$c.user.is_editing_disabled ? (
```
mwiencek: That's a bit unfortunate since we know it always exists here. Can you add `+$c: CatalystContextT,` within the definition for `ResultsPropsT` after the `...InlineResultsPropsT<T>,` line to indicate it's always defined in `ResultsPropsT`?

reosarevok: It's not though - it's for example not defined for InstrumentResults, since that doesn't load $c.

reosarevok: Another option would be to have ResultsPropsT and ResultsPropsWithContextT and use the appropriate type each time.

reosarevok: Changed to use ResultsPropsWithContextT

root/search/components/RecordingResults.js
```diff
@@ -124,7 +124,7 @@ export const RecordingResultsInline = ({
         <th>{l('Artist')}</th>
         <th>{l('ISRCs')}</th>
         <th>{l('Release')}</th>
-        {$c.session && $c.session.tport ? <th>{l('Tagger')}</th> : null}
+        {$c && $c.session && $c.session.tport ? <th>{l('Tagger')}</th> : null}
```
reosarevok: This can possibly also be substituted by something else if desired, but appears only once.

PR #951, opened 2019-02-26
MBS-10049: Convert user/report to React (+ MBS-8922)

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10049
https://tickets.metabrainz.org/browse/MBS-8922

Opted to leave the checkbox for MBS-8922 but make it on by default. If a user specifically does not want anyone to contact them, they can uncheck that (and then we should honor it, even if it involves not acting on the report because we can't ask for more info).


reosarevok:
I tested this manually and it seems to do everything right at least to the point where it should send the email out - I don't have anything set up for email sending locally so I couldn't test that I actually received the email though, but given the backend part hasn't changed, it should be fine.

root/user/ReportUser.js
```diff
@@ -0,0 +1,158 @@
...
+import UserAccountLayout from '../components/UserAccountLayout';
+import FormRow from '../components/FormRow';
+import FormRowCheckbox from '../components/FormRowCheckbox';
+import FormRowSelect from '../components/FormRowSelect';
+import FormRowTextArea from '../components/FormRowTextArea';
+import FormSubmit from '../components/FormSubmit';
+import {withCatalystContext} from '../context';
+
+type ReportReasonT = 'enforcing_guidelines' | 'ignoring_guidelines'
+                      | 'other' | 'spam' | 'unresponsiveness' | 'voting';
```
mwiencek: We normally follow this style for long union types:

```JavaScript
type ReportReasonT =
  | 'enforcing_guidelines'
  | 'ignoring_guidelines'
  | 'other'
  | 'spam'
  | 'unresponsiveness'
  | 'voting'
  ;
```

root/user/ReportUser.js
```diff
@@ -0,0 +1,158 @@
...
+          required
+          rows={10}
+        />
+
+        <FormRow hasNoLabel>
+          <p>
+            {l(`If you don't want our admins to contact you further regarding
+                this report, you can uncheck the checkbox below.`)}
+            <br />
+            {l(`We recommend leaving it checked, so that you can be contacted
```
mwiencek: Would it make sense to combine these two strings + the `<br />` into one? The second kinda follows directly from the first.

reosarevok: I was trying to avoid having the <br /> in the string, but I guess that'd make sense? Done

root/user/ReportUser.js
```diff
@@ -0,0 +1,164 @@
...
+    },
+  ],
+};
+
+const ReportUser = ({
+  $c,
+  form,
+  user,
+}: Props) => {
+  form.field.reveal_address.value = true;
```
mwiencek: A default value should go in lib/MusicBrainz/Server/Form/User/Report.pm instead (you can pass a `default` key to the `reveal_address` field). This would reset it to true even if they previously unchecked it but the page reloaded due to an unrelated error.

root/user/ReportUser.js
```diff
@@ -0,0 +1,164 @@
...
+            uri: {
+              href: '/doc/Code_of_Conduct',
+              target: '_blank',
+            },
+          })}
+      </p>
+
+      <p>
+        {l(`Your report will be sent to our {uri|account administrators},
+           who will decide what action to take.`,
```
mwiencek:
```suggestion
            who will decide what action to take.`,
```

root/user/ReportUser.js
```diff
@@ -93,23 +93,18 @@ const ReportUser = ({
...
       {l(`Be sure to provide direct links to examples of the behaviour you’re
-          reporting (for example, use “https://musicbrainz.org/edit/23”
-          instead of “edit #23”, and use “https://musicbrainz.org/edit/42 and
-          https://musicbrainz.org/edit/43” instead of
-          “https://musicbrainz.org/user/SomeUser/edits”).
-          Providing links makes it much easier for the recipients of the
-          report to look into the issues; a report without links is
-          unlikely to be acted on fast, since it will require a lot of
-          additional research.`)}
+        reporting (for example, use “https://musicbrainz.org/edit/23”
```
yvanzo: Unexpected and erroneous indent change in https://github.com/metabrainz/musicbrainz-server/pull/951/commits/87f85b374f4ad97d2b28cb03fdb626f7f8480c50

root/user/ReportUser.js
```diff
@@ -62,7 +62,7 @@ const reportReasonOptions = {
     },
     {
       label: N_l(
-        'Editor has violated some other part of our Code of Conduct'
+        'Editor has violated some other part of our Code of Conduct',
```
yvanzo: Coding style change in https://github.com/metabrainz/musicbrainz-server/pull/951/commits/87f85b374f4ad97d2b28cb03fdb626f7f8480c50 is unrelated to MBS-8922.

root/user/ReportUser.js
```diff
@@ -128,6 +123,17 @@ const ReportUser = ({
         rows={10}
       />
 
+      <FormRow hasNoLabel>
+        <p>
+          {exp.l(`If you don't want our admins to contact you further
```
yvanzo: Should use typographic apostrophe as in `If you don’t want`.

PR #952, opened 2019-02-26
Change user to entity on UserAccountLayout for use with chooseLayoutComponent

reosarevok:
CatCat caught this one on test - we weren't actually using UserAccountLayout in chooseLayoutComponent, but we were using it for entity/Subscribers, so the user subscriber page broke. In order to use it on cLC, it needs to take "entity" rather than "user", like all other entity layouts, so changed that.

### DIFF
--- a/root/account/Donation.js
+++ b/root/account/Donation.js
@@ -24,11 +24,16 @@ const Donation = ({days, nag}: Props) => (
       ? (
         <>
           <p>
-            {l('We have not received a donation from you recently. If you have just made a PayPal donation, then we have not received a notification from PayPal yet. Please wait a few minutes and reload this page to check again.')}
+            {l(`We have not received a donation from you recently. If you have
+                just made a PayPal donation, then we have not received a
+                notification from PayPal yet. Please wait a few minutes and
+                reload this page to check again.`)}
           </p>
           <p>
-            {l('If you would like to make a donation, {donate|you can do that here}. If you have donated, but you are still being nagged, please {contact|contact us}.',
-              {contact: CONTACT_URL, donate: DONATE_URL})}
+            {l(`If you would like to make a donation, {donate|you can do that
+                here}. If you have donated, but you are still being nagged,
+                please {contact|contact us}.`,
+               {contact: CONTACT_URL, donate: DONATE_URL})}
           </p>
         </>
       ) : (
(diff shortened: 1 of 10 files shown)


reosarevok:
While at it, fixed a few long lines and a case where we weren't using the defined Props.

root/account/Preferences.js
```diff
@@ -21,11 +23,12 @@ type Props = {|
   ...PreferencesFormPropsT,
 |};
 
-const Preferences = withCatalystContext(({$c, ...props}) => (
+const Preferences = withCatalystContext(({$c, ...props}: Props) => (
   <UserAccountLayout
+    // $FlowFixMe We won't ever reach this template if $c.user does not exist
```
mwiencek: You can tell Flow it always exists by changing the prop to `+$c: {user: EditorT} & CatalystContextT,`.

root/account/Donation.js
```diff
@@ -24,11 +24,16 @@ const Donation = ({days, nag}: Props) => (
...
+                notification from PayPal yet. Please wait a few minutes and
+                reload this page to check again.`)}
           </p>
           <p>
-            {l('If you would like to make a donation, {donate|you can do that here}. If you have donated, but you are still being nagged, please {contact|contact us}.',
-              {contact: CONTACT_URL, donate: DONATE_URL})}
+            {l(`If you would like to make a donation, {donate|you can do that
+                here}. If you have donated, but you are still being nagged,
+                please {contact|contact us}.`,
+            {contact: CONTACT_URL, donate: DONATE_URL})}
```
mwiencek: I'm honestly not sure why eslint wants this object indented outside of the enclosing `{}` - that seems like a bug to me. For example, you wouldn't write:

```JSX
const x = (
  <div>
    {hey('whats',
    'up')}
  </div>
);
```

You'd indent `'up'` in like you would in any other case. In fact, eslint complains about that in this example...it's only buggy if you make `'whats'` a multiline string.

reosarevok: It's always asked for this in multiline strings, yeah. I assumed there was a reason, but we could always change it - should we report upstream and ask first?

mwiencek: I figured out how to get it working with https://github.com/metabrainz/musicbrainz-server/pull/953

With that it'll want all arguments aligned with each other, as in the opening `` ` `` and opening `{` here in each argument:

```JSX
{l(`multiple
    lines are {cool}`,
   {cool: true})}
```

reosarevok: Neat! Let's get that one merged, then I can use it to fix this and a bunch of other cases like this one too.

reosarevok: Updated these

PR #956, opened 2019-02-28
MBS-9632: Convert Stats to React (VIII)

reosarevok:
Editors

@yvanzo mentioned having issues here where editor was null and it crashed on the "key={editorStat.editor.id}" of line 59 (https://gist.github.com/yvanzo/78a7a66cd6c181e8ae7034b7aeef0cf6#file-statistics-editors). I haven't seen a case where editor is null, yet dataPoints.length > 0, but since that suggests it can happen, we might want to turn EditorStatT.entity into EditorT | null. I'm still confused whether that should actually ever happen or whether it's a bug somewhere though, since even deleted editors are supposed to exist and just be named Deleted Editor Whatever


root/statistics/Editors.js
```diff
@@ -0,0 +1,128 @@
...
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+import {l_statistics as l} from '../static/scripts/common/i18n/statistics';
```
reosarevok: I understand in this case we still need to keep the import for the aliasing?

mwiencek: Correct.

root/statistics/Editors.js
```diff
@@ -0,0 +1,128 @@
...
+  topRecentlyActiveVoters,
+  topVoters,
+}: EditorsStatsT) => (
+  <StatisticsLayout fullWidth page="editors" title={l('Editors')}>
+    <p>
+      {l('Last updated: {date}',
+         {date: dateCollected})}
+    </p>
+    <p>
+      {l('For the vote statistics, only yes or no votes are counted, abstain \
```
yvanzo: Would be neat to use `` instead to wrap message.

yvanzo:
> I'm still confused whether that should actually ever happen or whether it's a bug somewhere though

This may happen with slave server.  Better have graceful degradation (or an explanatory error message, but this is more work) rather than this raw error message.

mwiencek:
> This may happen with slave server. Better have graceful degradation (or an explanatory error message, but this is more work) rather than this raw error message.

I can see how that could happen. The editor table isn't replicated, but the statistic table is, so it may reference editor IDs that you don't have?

Anyway, the `EditorLink` component should already do the correct thing by displaying `[missing editor]` if it's passed a null/undefined editor - I believe the only issue is where it tries to use `editorStat.editor.id` as a key. We can just pass something like `` `missing-${index}` `` for the key if `editorStat.editor` is falsy.

root/statistics/Editors.js
```diff
@@ -0,0 +1,128 @@
...
+          <th>
+            {countLabel}
+            <div className="arrow" />
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        {dataPoints.length > 0 ? (
+          dataPoints.map((editorStat, index) => (
+            <tr className={loopParity(index)} key={editorStat.editor.id}>
```
mwiencek: Am I missing something? `EditorStatT` has an `entity` prop, but here and below you're referencing an `editor` prop.

mwiencek: Also, ``key={editorStat.editor ? editorStat.editor.id : `missing-${index}`}`` (or s/editor/entity/?) seems like it'd resolve @yvanzo's error.

reosarevok: Duh. EditorStatsTable was missing types. It works, so I assume the EditorStatT type should have editor not entity. Will check.

yvanzo: Thanks, page is not crashing anymore for me!

root/statistics/Editors.js
```diff
@@ -0,0 +1,143 @@
...
+));
+
+const Editors = ({
+  dateCollected,
+  topEditors,
+  topRecentlyActiveEditors,
+  topRecentlyActiveVoters,
+  topVoters,
+}: EditorsStatsT) => (
+  <StatisticsLayout fullWidth page="editors" title={l('Editors')}>
```
yvanzo: Tab labels are not localized for this page and some others.

reosarevok: I know. I have a fix for that, I think it was in the next PR though

PR #959, opened 2019-03-06
MBS-10057: Convert instruments-list.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10057


root/components/InstrumentsList.js
```diff
@@ -0,0 +1,98 @@
...
+          <td>
+            <DescriptiveLink entity={instrument} />
+          </td>
+          <td>
+            {instrument.typeName
+              ? lp_attributes(instrument.typeName, 'instrument_type')
+              : null}
+          </td>
+          <td>
+            {instrument.description
```
mwiencek: Apparently the description can contain HTML, so we need to use `dangerouslySetInnerHTML` on the `<td>`. I see we have the same issue in root/search/components/InstrumentResults.js and root/instrument/List.js.

root/search/components/InstrumentResults.js
```diff
@@ -21,17 +21,12 @@ function buildResult(result, index) {
   const score = result.score;
 
   return (
-    <tr className={loopParity(index)} data-score={score} key={instrument.id}>
-      <td>
-        <EntityLink entity={instrument} />
```
reosarevok: InstrumentListEntry uses DescriptiveLink instead, but AFAICT that is exactly equivalent for instruments.

root/instrument/List.js
```diff
@@ -27,7 +27,12 @@ const Instrument = ({instrument}) => (
       ? (
         <>
           {' — '}
-          {l_instrument_descriptions(instrument.description)}
+          <span
+            dangerouslySetInnerHTML={{
+              __html:
+                l_instrument_descriptions(instrument.description),
```
mwiencek: Sorry, I was dumb for suggesting `dangerouslySetInnerHTML`. We should actually be using:

```JSX
{expand2react(l_instrument_descriptions(instrument.description))}
```
(`expand2react` has to be imported, and a `<span>` wrapper is no longer needed.) This is safer since it only parses a safe subset of HTML.

reosarevok: Done :)

mwiencek:
The description display changes should resolve https://tickets.metabrainz.org/browse/MBS-10112

PR #960, opened 2019-03-06
MBS-10059: Convert series-list.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10059


lib/MusicBrainz/Server/Entity/Series.pm
```diff
@@ -56,6 +56,7 @@ around TO_JSON => sub {
 
     my $json = $self->$orig;
     $json->{orderingTypeID} = $self->ordering_type_id;
+    $json->{orderingType} = $self->ordering_type;
```
mwiencek: It's added in the linked entities above, so in the JS you can already access it via

```JS
const orderingType =
  linkedEntities.series_ordering_type[series.orderingTypeID];
```

reosarevok: Unless I'm mistaken, we don't have linkedEntities until https://github.com/metabrainz/musicbrainz-server/pull/737 is merged? :) (although this can wait until that is merged and then be changed, obvs).

mwiencek: Oops, yes. :) You can access it through `$c.linked_entities` instead but I guess it can wait too.

reosarevok: Made the change, now that's merged.

PR #962, opened 2019-03-06
MBS-9926 (III): Also use Subscribers.js for user subscriptions

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9926

I think it would make sense to publish this after this is merged and create a new ticket for root/user/subscriptions/ since that seems self-contained, but I'll let you decide about that :)


reosarevok:
For the second commit: we were already differentiating between "edits you make" and "edits {user} makes" for the case where there's no subscribers (on the TT template) so there seemed to be no reason not to do the same for the case where there *are* some subscribers.

reosarevok:
Also, and unrelated to this PR other than because I found it by accident while working on it: there seems to be nothing stopping a user from subscribing to their own edits, other than the link not being shown on their profile. Should this be enforced at the DB level?

yvanzo:
> I think it would make sense to publish this after this is merged

Do both “this” and “this” refer to this pull request?

> there seems to be nothing stopping a user from subscribing to their own edits

Is there any issue with this?

reosarevok:
The first this is the commits here plus the two already merged MBS-9926 related PRs, https://github.com/metabrainz/musicbrainz-server/pull/810 and https://github.com/metabrainz/musicbrainz-server/pull/867 (AFAICT they're not yet released).

There's probably technically no issue with someone being able to subscribe to themselves (I guess they'll just see their own edits in their subscription lists?), but it's clearly not supposed to happen.

yvanzo:
> there seems to be nothing stopping a user from subscribing to their own edits

Is there any issue with this? Actually, I subscribed to my own edits.

reosarevok:
I mean, maybe that's ok - it's just since we actively choose not to display that link on the editors' own profile, I assumed we just want to make sure they can't do it :)

root/entity/Subscribers.js
```diff
@@ -76,32 +101,68 @@ const Subscribers = ({
           </ul>
         </>
       ) : (
-        <p>
-          {texp.l('There are currently no users subscribed to {entity}.',
-            {entity: entity.name})}
-        </p>
+        entityType === 'editor' ? (
+          viewingOwnProfile ? (
+            <p>
```
yvanzo: Nitpicking: `p` element could actually precede the those two above conditions, thus needing less lines of code. The same change could be made with most other `p` elements of this PR, in a separate commit if needed for clarity.

yvanzo:
On the self-subscription thing: I think we should probably add a profile setting to explicitly allow having editor’s own  changes along with subscribed editors’ changes in notification emails.

reosarevok:
Wanna add a ticket? ;)

root/entity/Subscribers.js
reosarevok: Done

yvanzo:
Added https://tickets.metabrainz.org/browse/MBS-10083

root/entity/Subscribers.js
```diff
@@ -76,32 +116,68 @@ const Subscribers = ({
...
+        subscribed ? (
           <p>
-            {exp.l('Be the first! {sub|Subscribe}?',
-              {sub: '/account/subscriptions/' + entityType + '/add?id=' + entity.id})}
+            {exp.l('You are currently subscribed. {unsub|Unsubscribe}?',
+                   {
+                     unsub: '/account/subscriptions/' +
+                            entityType +
+                            '/remove?id=' +
+                            entity.id,
```
mwiencek: Might it make sense to build these links above instead? e.g.

```JavaScript
const unsubLink = '/account/subscriptions/' + entityType +
  '/remove?id=' + entity.id;

// ...

{exp.l('You are currently subscribed. {unsub|Unsubscribe}?',
       {unsub: unsubLink})}
```

('Cause while the old version slightly exceeded 78 chars, I think blowing it up like this inline made it look a bit busier than it needs to.)

root/entity/Subscribers.js
reosarevok: And I forgot to push the commit, lol. Well, I'll push it after looking at @mwiencek's comments.

yvanzo: Thank you for https://github.com/metabrainz/musicbrainz-server/pull/962/commits/dba82eab57d6ce1afbdb96c26f3abda10a4fc4db

PR #963, opened 2019-03-06
MBS-9936: Move from CJS style to ES6 (imports, exports) (X)

reosarevok:
Final missing bits

https://tickets.metabrainz.org/browse/MBS-9936

As far as I can tell, this should be the last part of this. Anything I've missed can be pointed out for me to fix now with this PR, or we can just fix it as we see it in the future :)

### DIFF
--- a/root/static/scripts/global.js
+++ b/root/static/scripts/global.js
@@ -1,5 +1,9 @@
+let defaultExport;
+
 if (typeof global === 'undefined') {
-  module.exports = window;
+  defaultExport = window;
 } else {
-  module.exports = global;
+  defaultExport = global;
 }
+
+export default defaultExport;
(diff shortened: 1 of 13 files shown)


root/static/scripts/global.js
```diff
@@ -1,5 +1,9 @@
...
+
 if (typeof global === 'undefined') {
-  module.exports = window;
+  defaultExport = window;
 } else {
-  module.exports = global;
+  defaultExport = global;
 }
+
+export default defaultExport;
```
mwiencek: There's an import in root/static/scripts/tests/react-macros.js which needs `.default` added

reosarevok: Done! (BTW, eslint thinks most variables in that file are unused - should we remove them, should we disable that eslint check on this file, or should we do some third thing?)

mwiencek: Ah, we should disable the eslint check on that file (the imports are used in `eval`'d code sent from t/lib/t/TemplateMacros.pm, basically).

PR #943, opened 2019-02-23
MBS-10043: Convert place-list.tt to React + MBS-10044

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10043
https://tickets.metabrainz.org/browse/MBS-10044

Areas, instruments, places, series. Was getting too big for review, so submitting this part already.

PR #967, opened 2019-03-07
MBS-10064: Convert artists-list.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10064

Also includes changes to show areas on the Artists page for area (as per https://github.com/metabrainz/musicbrainz-server/pull/936)


root/collection/index.tt
```diff
@@ -19,6 +19,9 @@
       [%- IF collection.type.item_entity_type == 'event' -%]
         [% React.embed(c, 'components/EventsList', { events => entities, checkboxes => my_collection ? 'remove' : '', order => order, sortable => 1})
             WRAPPER 'components/with-pager.tt' %]
+      [%- ELSIF collection.type.item_entity_type == 'artist' -%]
+        [% React.embed(c, 'components/ArtistsList', { artists => entities, checkboxes => my_collection ? 'remove' : '', noAreas => 1, order => order, sortable => 1})
```
reosarevok: Hiding areas on collection page since the three area columns might be a bit overwhelming and nobody has asked for those so far :)

root/components/ArtistsList.js
```diff
@@ -0,0 +1,141 @@
...
+          ) : null}
+          <td>
+            <DescriptiveLink entity={artist} />
+          </td>
+          <td>
+            {artist.typeName
+              ? lp_attributes(artist.typeName, 'artist_type')
+              : null}
+          </td>
+          <td>{artist.gender ? l(artist.gender.name) : null}</td>
```
mwiencek:
```suggestion
          <td>{artist.gender ? lp_attributes(artist.gender.name, 'gender') : null}</td>
```

reosarevok:
Extracted the info into ArtistListEntry now. While changing that I made it so area is always shown, and begin and end areas and dates can be turned on or off. Begin and end dates make the Area artist tab a bit crowded, but I can't see anything there that does not seem useful, except maybe ratings, but those are shown on all the other tabs so we probably should not drop them.

reosarevok:
We need the areas because it needs to show *which* area it is (remember that it will also count for areas which are part of the one the user is in. We don't need the dates as such, it just seems very odd to not display them.

yvanzo:
Artists list for a selected area should display that much redundant information. Most entries will have one, two or three times the full name of the selected area. It is very different from artist search for which areas vary among artists. I would prefer MBS-6574 changes to be submitted in a separate PR if possible, so that we can move on to schema change first.

reosarevok:
It doesn't matter how many times it has it, it's never redundant. It's 100% relevant whether the matching area is the main area, the begin area, the end area, or all of them, and whether it matches the area the user is on or an area lower in the containment.

yvanzo:
Which is why I suggested having a “Role” column in the first place. Alternatively or complementarily, some columns could be hidden when not needed, similarly to the credited “Artist” column in release page.

If the above doesn’t seem stupid, I would prefer this issue to be discussed later on, after merging MBS-10064 related changes only.

reosarevok:
I really wouldn't want to start splitting that from the commits now. If we really must change something (I feel it's still fairly readable, but it is a bit on the edge), what about dropping the dates? As a user I'd still want to see all this info, though...

yvanzo:
A third alternative: adding filters, on the model of artist’s discography.

yvanzo:
I just found out that some other tabs for area unconditionally display area info too: Events, Labels, Places with unreleased PR #943. Fortunately, these have only one area field with dates, not three.

[Here is an example](https://musicbrainz.org/area/e68879f9-bd95-41ff-96ba-c082ff37cc74/labels
) of what I try to avoid: 190 links to the current area from the Area column holding the exact same value for every listed entry, thickening rows height for little added info.

reosarevok:
Your example is absolutely intentional though, and needed - because of [what happens one level higher](https://musicbrainz.org/area/f9caf2d8-9638-4b96-bc49-8462339d4b2e/labels). I'm still not sure if you don't understand why this change was made or you understand it but just disagree with it anyway :)

yvanzo:
Not needed in my example, needed in yours. It can be useful in some pages and not in some other pages, just like the above-mentioned “Artist” credits column of tracklist in release pages.

mwiencek:
Let's discuss a way to improve this post-schema change, please? Especially since the same minor issue is already present on other pages, and I don't see a quick resolution here. This needs to be merged if we want any chance of having collection UI stuff ready. Ideally we would be freezing the schema change branch on Monday save for bug fixes, since we need a lot of time for testing and preparing upgrade notes.

PR #968, opened 2019-03-07
MBS-10068: Convert works-list.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10068

Since we already have WorkListEntry.js which did most of what we need, I just added a few optional sections to it to support this. We might eventually want to do similar things for other lists which reappear in mostly similar ways around the codebase.


root/static/scripts/common/components/DynamicAttributesList.js
```diff
@@ -0,0 +1,37 @@
...
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+import bracketed from '../utility/bracketed';
+
+const DynamicAttributesList = ({entity}: {entity: WorkT}) => {
```
reosarevok: Named it like this to already make it a bit more general in preparation for @yvanzo's work on attributes, which I imagine will lead to this component being used more widely :)

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -9,42 +9,61 @@
...
+  hasAttributesColumn,
   hasIswcColumn,
-  hasMergeColumn,
+  hasRatingsColumn,
+  seriesItemNumbers,
   work,
 }: WorkListRowProps) => (
   <>
-    {hasMergeColumn ? (
+    {$c.user_exists && checkboxes ? (
```
reosarevok: This generalization also fixes a bug where a logged-out user would be shown the checkbox on ISWC pages.

root/static/scripts/common/components/DynamicAttributesList.js
```diff
@@ -0,0 +1,37 @@
...
+  console.log(entity.attributes);
```
mwiencek:
```suggestion
```

root/static/scripts/common/components/DynamicAttributesList.js
```diff
@@ -0,0 +1,37 @@
...
+    entity.attributes ? (
+      <ul>
+        {entity.attributes.map(attribute => (
+          <li key={attribute.id}>
+            {lp_attributes(
+              attribute.value,
+              'work_attribute_type_allowed_value',
+            )}
+            {' '}
+            {bracketed(lp_attributes(
```
mwiencek: You could use `bracketedText` here since it's expected to be a string (but `bracketed` will work fine too).

root/components/WorksList.js
```diff
@@ -0,0 +1,92 @@
...
+
+import {withCatalystContext} from '../context';
+import WorkListEntry from '../static/scripts/common/components/WorkListEntry';
+
+import SortableTableHeader from './SortableTableHeader';
+
+type Props = {|
+  +$c: CatalystContextT,
+  +checkboxes?: string,
+  +noRatings?: boolean,
```
mwiencek: Just having looked at your artists-list PR, it'd be nice to make the prop names consistent between them, so inverting this and making it `showRatings`.

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -9,42 +9,61 @@
...
 import EntityLink from './EntityLink';
 
 type WorkListRowProps = {|
-  +hasIswcColumn: boolean,
-  +hasMergeColumn: boolean,
+  +$c: CatalystContextT,
+  +checkboxes?: string,
+  +hasAttributesColumn?: boolean,
+  +hasIswcColumn?: boolean,
+  +hasRatingsColumn?: boolean,
```
mwiencek: Replacing `has` with `show` in these three props would also seem more consistent

root/static/scripts/common/components/DynamicAttributesList.js
mwiencek: Makes sense. The `lp_attributes` contexts are hardcoded for works, but I guess it can be figured out how to generalize that further when more entities are added.

root/static/scripts/common/components/WorkListEntry.js
mwiencek: And not sure `Column` is needed since you don't have `showRatingsColumn`?

root/components/WorksList.js
reosarevok: I've sent extra commits changing the other existing lists to use positive rather than negative conditions (which seems reasonable, in the vein of doing X ? bar : foo rather than !X ? foo : bar)

root/components/EventsList.js
```diff
@@ -46,12 +46,12 @@ const EventsList = ({
...
-  noArtists,
-  noLocation,
-  noRatings,
-  noType,
   order,
   seriesItemNumbers,
+  showArtists,
+  showLocation,
+  showRatings,
+  showType,
```
mwiencek: It probably would've made sense if you kept the previous defaults though (i.e. with `showArtists = true,` etc. here), so that you don't have to specify these everywhere unless they're false. (Not sure if that's true for all of them, but in general it seems better for the default to reflect what's most common.)

reosarevok: IMO that's kinda confusing. If we're going to make it default to true and have showArtists={false} or whatever, then we're back to specifically negating parts rather than requesting them, and then having it as noArtists seems to make more sense.

mwiencek: Is it? The problem I see with `noArtists` is that double-negatives like `noArtists={false}` are confusing, but I don't think negating a positive statement like `showArtists={false}` is confusing.

If you think it makes sense for all the positive props to default to false then let's do that, though I wonder if that doesn't make the component more annoying to use when you need `showArtists={true}` in 4 out of 5 places.

reosarevok: Well, it did make it a bit annoying to convert. But I feel in the future, when using it, it'll be much easier to remember "if I want it to show artists, I should add showArtists here" :)

root/components/WorksList.js
```diff
@@ -0,0 +1,92 @@
...
+  order,
+  seriesItemNumbers,
+  sortable,
+  works,
+}: Props) => (
+  <table className="tbl">
+    <thead>
+      <tr>
+        {$c.user_exists && checkboxes ? (
+          <th style={{width: '1em'}}>
```
yvanzo: Unneeded style attribute.

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -9,42 +9,64 @@
...
 type WorkListRowProps = {|
-  +hasIswcColumn: boolean,
-  +hasMergeColumn: boolean,
+  +$c: CatalystContextT,
+  +checkboxes?: string,
+  +hasAttributesColumn?: boolean,
+  +hasIswcColumn?: boolean,
+  +hasMergeColumn?: boolean,
+  +hasRatingsColumn?: boolean,
+  +seriesItemNumbers?: {+[number]: string},
```
yvanzo: New type available!

root/static/scripts/common/components/DynamicAttributesList.js
yvanzo: Simply name it `AttributesList` then, thanks!

root/components/WorksList.js
reosarevok: Why is it unneeded? It's in all other EntityList.js files, so I can remove it all as an extra commit, just not fully sure why that's the case :) Is it because checkboxes are always 1em and the table autoadjusts to them?

yvanzo: > Why is it unneeded?

It was not in the corresponding TT file at least and I don’t see any need for it either.

> It's in all other EntityList.js files, so I can remove it all as an extra commit

Sure, we just missed that.

reosarevok: It was in several TT files so I added to all lists for consistency, but it might be it's unneeded in all of them. I'll submit an extra commit here then.

yvanzo: This column only contain a checkbox per row. IMHO, there is no need to set the column width. It would be find out if style has been set or unset at purpose for some templates, or if these were just inconsistent.

PR #970, opened 2019-03-12
MBS-10066: Convert release_groups-list.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10066


root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,263 @@
...
+      <td>
+        <ArtistCreditLink
+          artistCredit={releaseGroup.artistCredit}
+        />
+      </td>
+    ) : null}
+    {noGrouping ? (
+      <td>
+        {releaseGroup.typeName
+          // TODO: replace with releaseGroupType when it works here
```
reosarevok: utility/releaseGroupType currently uses $c.linked_entities. That does *not* contain RG type info for RG collections. @yvanzo said that $c.linked_entities is mostly intended for release pages - if that's the case we might want to consider switching how releaseGroupType works.

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,263 @@
...
+  checkboxes,
+  noGrouping,
+  noRatings,
+  order,
+  releaseGroups,
+  seriesItemNumbers,
+  showArtists,
+  sortable,
+}: ReleaseGroupsListProps) => {
+  const groupedReleaseGroups = groupBy(releaseGroups, 'typeName');
```
reosarevok: This actually ensures proper grouping, although it doesn't necessarily ensure proper sorting (say, in RG collection pages). That said, current collection pages don't even group RGs properly, so this already improves on that by making sure there's only one "Album+Compilation" grouping instead of N. We might want to ensure the returned RGs are sorted in the same way they are for artist overview pages, but that's not part of this PR unless you want it to be :) 

BTW, sortable actually sorts globally here, not each mini-table, meaning while the contents of each mini-table will be correctly sorted, the tables *themselves* will also be sorted, so that if the RG with the earliest year of all is "Single" then "Single" will be the first table in the collection if you sort it by year. Not sure if intended.

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,263 @@
...
+
+      Object.keys(groupedReleaseGroups).map((type) => {
+        const releaseGroupsOfType = groupedReleaseGroups[type];
+        return (
+          <React.Fragment key={type}>
+            <h3>
+              {type === 'null'
+                ? l('Unspecified type')
+                // TODO: replace with releaseGroupType when it works here
+                : releaseGroupsOfType[0].l_type_name
```
reosarevok: Same as line 160, plus a bit of a hack to get the actual translated type (the releaseGroupType alternative would still use releaseGroupsOfType[0], since that function takes an RG, not a type string).

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,263 @@
...
+                showArtists={showArtists}
+                sortable={sortable}
+              />
+              <tbody>
+                {releaseGroupsOfType.map((releaseGroup, index) => (
+                  <ReleaseGroupsListEntry
+                    $c={$c}
+                    checkboxes={checkboxes}
+                    index={index}
+                    key={releaseGroup.id}
```
reosarevok: I'm not quite sure this key is required / useful.

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,263 @@
...
+      </td>
+    ) : null}
+    {seriesItemNumbers ? (
+      <td style={{width: '1em'}}>
+        {seriesItemNumbers[releaseGroup.id]}
+      </td>
+    ) : null}
+    <td className="c">
+      {releaseGroup.firstReleaseDate
+        ? new Date(releaseGroup.firstReleaseDate).getFullYear()
```
reosarevok: Since we get date as a string, we can't just do .year like in TT. I think this is the easiest alternative?

reosarevok:
Now includes the commit from https://github.com/metabrainz/musicbrainz-server/pull/971 so that it can use releaseGroupType

root/components/ReleaseGroupsList.js
mwiencek: My preference would be to change the `firstReleaseDate` serialization to use `$self->first_release_date->TO_JSON` rather than `->format`, so that we can use `.year` here. (The places that need it as a string can use the `formatDate` utility.)

Technically the `Date` constructor wouldn't be able to parse our `PartialDate` string format if the dates are actually partial, e.g. `1999-??-01`.

root/components/ReleaseGroupsList.js
mwiencek: Ah, the previous collection page kept the items sorted globally, so it sorted everything as a single list and then only grouped adjacent items of the same type (so there could be duplicate groups if they weren't adjacent).

The new sorting doesn't seem easier to understand, though grouping them correctly is probably better overall. The UI is just confusing in general (I was expecting the column headers to only sort the specific table they're in, I guess).

reosarevok: I guess we could make all sorts of sort sort by type first?

root/components/ReleaseGroupsList.js
reosarevok: The problem is firstReleaseDate is currently used in Autocomplete.js, where it is taken from TO_JSON for direct search, but taken from the WS (as a string) for indexed.

I guess one option would be to serialize both firstReleaseDate and firstReleasePartialDate, and use the latter here but the former for Autocomplete.js?

mwiencek: Ah, I forgot about indexed. Then what you have works except I'd import root/static/scripts/common/utility/parseDate.js and do `parseDate(releaseGroup.firstReleaseDate).year` to get the year.

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,248 @@
...
+  order,
+  releaseGroups,
+  seriesItemNumbers,
+  showRatings,
+  sortable,
+}: ReleaseGroupsListProps) => {
+  const groupedReleaseGroups = groupBy(releaseGroups, 'typeName');
+  return (
+    groupByType ? (
+      Object.keys(groupedReleaseGroups).map((type) => {
```
mwiencek: I'd probably make the grouped listing a separate component from the basic list one (and the grouped one could use the basic list one somehow, as the release-group-list seems mostly repeated between them). But it shouldn't block merging this.

reosarevok: Started the split, added a TODO to finish the split once we don't need to embed the component from TT.

root/components/ReleaseGroupsList.js
yvanzo: Sorting globally is kinda broken, but it is probably less broken than having duplicate groups.

Ideally, we should have a single table with type as a column, and the ability to sort and filter on more than one column. This can wait for website redesign.

root/components/ReleaseGroupsList.js
yvanzo: See https://reactjs.org/docs/lists-and-keys.html#keys

root/components/ReleaseGroupsList.js
yvanzo: I would rather create a task ticket and refer to it from component calls in TT files.

reosarevok: Done (if that's what you meant, anyway)

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,257 @@
...
+  +sortable?: boolean,
+|};
+
+type ReleaseGroupsListEntryProps = {|
+  +$c: CatalystContextT,
+  +checkboxes?: string,
+  +groupByType?: boolean,
+  +index: number,
+  +releaseGroup: ReleaseGroupT,
+  +seriesItemNumbers?: {+[number]: string},
```
yvanzo: Should use the new type instead.

root/components/ReleaseGroupsList.js
```diff
@@ -0,0 +1,257 @@
...
+  +showRatings?: boolean,
+|};
+
+type ReleaseGroupsListProps = {|
+  +$c: CatalystContextT,
+  +checkboxes?: string,
+  +groupByType?: boolean,
+  +order?: string,
+  +releaseGroups: $ReadOnlyArray<ReleaseGroupT>,
+  +seriesItemNumbers?: {+[number]: string},
```
yvanzo: Same as above.

root/artist/index.tt
```diff
@@ -35,7 +35,9 @@ END %]
 
     [%- IF release_groups.size -%]
       <form action="[% c.uri_for_action('/release_group/merge_queue') %]" method="post">
-        [% INCLUDE 'components/release_groups-list.tt' release_groups=release_groups checkboxes='add-to-merge' WRAPPER 'components/with-pager.tt' %]
+        [% # TODO: MBS-10155 %]
+        [% React.embed(c, 'components/ReleaseGroupsList', { releaseGroups => release_groups, checkboxes => 'add-to-merge', showRatings => 1, groupByType => 1 })
```
yvanzo: I suggest to comment in line (so as not to output an extra line) and to detail when to check it:
```suggestion
        [% React.embed(c, 'components/ReleaseGroupsList', { releaseGroups => release_groups, checkboxes => 'add-to-merge', showRatings => 1, groupByType => 1 }) # TODO: On conversion to React, please check MBS-10155.
```

root/edit/details/merge_release_groups.tt
```diff
@@ -2,15 +2,13 @@
     <tr>
         <th>[% l('Merge:') %]</th>
         <td>
-            [% INCLUDE 'components/release_groups-list.tt' release_groups=edit.display_data.old
-                checkboxes='' no_ratings=1 no_grouping=1 show_artists=1 %]
+            [% React.embed(c, 'components/ReleaseGroupsList', { releaseGroups => edit.display_data.old }) % # TODO: On conversion to React, please check MBS-10155.]
```
yvanzo: Ending marker `%]` is broken here (and at some other places).
```suggestion
            [% React.embed(c, 'components/ReleaseGroupsList', { releaseGroups => edit.display_data.old }) # TODO: On conversion to React, please check MBS-10155. %]
```

PR #972, opened 2019-03-13
MBS-10072: JSON WS: hyphenate begin_area, end_area

mwiencek:
# Fix [MBS-10072](https://tickets.metabrainz.org/browse/MBS-10072)

As noted in the ticket, SOLR already uses `begin-area` and `end-area`, and this difference is probably causing more harm than fixing it will (see the linked Picard ticket).

I don't feel comfortable making breaking changes to the JSON WS, though technically we never announced that it was out of beta. One thing we can possibly do is announce the change, output all four properties for 6 months or so, then remove the underscored ones. The problem is I'm not sure where to announce this besides the blog, and if that would even reach the intended audience (see: the people still using /ws/1).


reosarevok:
Output all four seems ok, since this will otherwise no doubt break someone's tools. We can then add a ticket to remove the underscored ones with the next ws version.

mwiencek:
I've updated this to output all four properties. We can announce it later in the release blog post, or write a dedicated post for it, and in the community forums.

PR #973, opened 2019-03-13
MBS-10074: '-' on release search format/tracks if no medium exists

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10074

### DIFF
--- a/root/search/components/ReleaseResults.js
+++ b/root/search/components/ReleaseResults.js
@@ -37,8 +37,12 @@ function buildResult(result, index) {
       <td>
         <ArtistCreditLink artistCredit={release.artistCredit} />
       </td>
-      <td>{release.combined_format_name}</td>
-      <td>{release.combined_track_count}</td>
+      <td>
+        {release.combined_format_name || l('[missing media]')}
+      </td>
+      <td>
+        {release.combined_track_count || '-'}
+      </td>
       <td>
         <ReleaseDates events={release.events} />
       </td>


root/search/components/ReleaseResults.js
```diff
@@ -37,8 +37,16 @@ function buildResult(result, index) {
...
-      <td>{release.combined_track_count}</td>
+      <td>
+        {release.combined_format_name
+          ? release.combined_format_name
+          : '-'}
+      </td>
+      <td>
+        {release.combined_track_count
+          ? release.combined_track_count
+          : '-'}
```
mwiencek: (Nitpick: `{release.combined_track_count || '-'}` is shorter and does the same thing.)

reosarevok:
@yvanzo suggested using "n/a" on both lists (because it conveys meaning much more clearly). What do you think about that alternative?

mwiencek:
Either works for me, though I guess people might wonder why it's not applicable. `[no media]` or something might be more to the point if we want to be descriptive? (Also, I wonder if we're not being consistent with brackets, since we show `(unknown)` for unset formats where elsewhere we use `[unknown]`, `[removed]`, `[none]`, `[missing editor]`, ...)

reosarevok:
What about having [missing media] on the format column, but stick to - in the track number one? Since both will always be displayed together AFAICT, and I feel two [missing media] columns by each other might be a bit much. But maybe I'm overthinking it.

mwiencek:
> What about having [missing media] on the format column, but stick to - in the track number one?

Sounds good to me

PR #979, opened 2019-03-15
MBS-10086: Convert attributes admin lists to React

anirudhjain75:
https://tickets.metabrainz.org/browse/MBS-10086


lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -31,32 +31,39 @@ my @models = qw(
...
+);
 }
 
 sub attribute_base : Chained('/') PathPart('admin/attributes') CaptureArgs(1) RequireAuth(account_admin) {
     my ($self, $c, $model) = @_;
 
     $c->detach('/error_404') unless $model ~~ @models;
 
-    $c->stash->{model} = $model;
+    $c->stash(model => $model);
```
reosarevok: No real reason to change this. Same below on line 117

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,100 @@
...
+type attributeT = {
+  id: number,
+  name: string,
+  description: string,
+  childOrder: number,
+  parentID: number,
+  year: number,
+  hasDiscids: string,
+  entityType: string,
+  freeText: string
```
reosarevok: Should end with a comma.

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type Props = {
+  attributes: Array<attributeT> | null,
+  model: string,
+  models: Array<string>,
```
reosarevok: Clearly models can either be undefined or null, otherwise "models ? this : that" on line 35 would be pointless :)

anirudhjain75: It gave me flow errors when I allowed it to be null, let me see what I can do

anirudhjain75: the type of model cant be null but models can be so I am changing just that.

reosarevok:
One thing I've just realized is you haven't removed the .tt files you replaced. Unless they're still being used somewhere, it'd make sense to remove them :)

anirudhjain75:
> One thing I've just realized is you haven't removed the .tt files you replaced. Unless they're still being used somewhere, it'd make sense to remove them :)

I just kept them in case they may be required for reference. I'll remove it in sometime, test and push the changes :)

reosarevok:
The GitHub diff will still show the files being removed though, so there's no problem with just removing them now, and it avoids the need to send a separate PR later just for that (plus it's generally better if the porting + removal is one and the same commit, because if we want to check it in a couple years or something it will all be clearly connected)

yvanzo:
@anirudhjain75 please ([doc](https://github.com/metabrainz/musicbrainz-server/blob/master/HACKING.md#porting-tt-to-react)) “make sure your JS files conform to our ESlint rules by running_”

    ./node_modules/.bin/eslint $(git diff master.. --name-only '*.js')

anirudhjain75:
> @anirudhjain75 please ([doc](https://github.com/metabrainz/musicbrainz-server/blob/master/HACKING.md#porting-tt-to-react)) “make sure your JS files conform to our ESlint rules by running”
> 
> ```
> ./node_modules/.bin/eslint $(git diff master.. --name-only '*.js')
> ```

Hi yvanzo I've fixed all the lint errors except the $ReadOnlyArray because the arrays we are using are not read only. there are operations being performed on them

yvanzo:
As for ESLint warnings, maybe we should just disable `flow/no-mutable-array`, this rule seems unsafe.

reosarevok:
Would it make sense to make attributes $ReadOnlyArray and then create an array sortedAttributes that is initially equal to attributes but which does get changed?

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,100 @@
...
+        <h1>{l('Attributes')}</h1>
+        <ul>
+          {models ? models.sort().map((item) => (<li key={item}><a href={'/admin/attributes/' + item}>{item}</a></li>)) : null}
+        </ul>
+      </div>
+    ) : (
+      <div>
+        <h1>
+          <a href="/admin/attributes">{l('Attributes')}</a>
+          <text>{' / ' + model}</text>
```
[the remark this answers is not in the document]
anirudhjain75: Yea changing that

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,100 @@
...
+            <tr>
+              <th>{l('ID')}</th>
+              <th>{l('Name')}</th>
+              <th>{l('Description')}</th>
+              <th>{l('Child order')}</th>
+              <th>{l('Parent ID')}</th>
+              {model === 'MediumFormat' ? <th>{l('Year')}</th> : null}
+              {model === 'MediumFormat' ? <th>{l('Disc IDs allowed')}</th> : null}
+              {model === 'SeriesType' || model === 'CollectionType' ? <th>{l('Entity type')}</th> : null}
+              {model === 'WorkAttributeType' ? <th>{l('Free text')}</th> : null}
```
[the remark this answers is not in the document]
anirudhjain75: I think this is the best approach, also only expressions can be put inside the render block, if you add switch instead you get error stating that expressions are only allowed.

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,100 @@
...
+              <th>{l('Actions')}</th>
+            </tr>
+          </thead>
+          <tbody>
+            {attributes ? attributes.sort((a, b) => {
```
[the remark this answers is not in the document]
anirudhjain75: Will check and make the changes. Thanks :)

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,100 @@
+/*
+ * @flow
+ * Copyright (C) 2019 Anirudh Jain
+ * Copyright (C) 2019 MetaBrainz Foundation
```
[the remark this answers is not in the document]
anirudhjain75: Hmm but its being converted in 2019 🤔

root/admin/attributes/Index.js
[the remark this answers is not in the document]
anirudhjain75: @reosarevok What do you think about this ?

root/admin/attributes/Index.js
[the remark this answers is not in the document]
yvanzo: Rewriting is a derivative work. :bulb:

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,92 @@
+/*
+ * @flow
+ * eslint-disable flowtype/no-mutable-array
+ * Copyright (C) 2014 Anirudh Jain
```
yvanzo: But your contribution is 2019 :1234: 
Check other files too :writing_hand:

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,92 @@
...
+              <th>{l('Parent ID')}</th>
+              {model === 'MediumFormat' ? <th>{l('Year')}</th> : null}
+              {model === 'MediumFormat' ? <th>{l('Disc IDs allowed')}</th> : null}
+              {model === 'SeriesType' || model === 'CollectionType' ? <th>{l('Entity type')}</th> : null}
+              {model === 'WorkAttributeType' ? <th>{l('Free text')}</th> : null}
+              <th>{l('Actions')}</th>
+            </tr>
+          </thead>
+          <tbody>
+            {attributes ? attributes.sort((a, b) => compare(a.name.toLowerCase(), b.name.toLowerCase())).map((attribute) => (
```
yvanzo: `.toLowerCase()` calls are unneeded here.

root/admin/attributes/Language.js
```diff
@@ -0,0 +1,71 @@
...
+            <th>{l('ISO 639-3')}</th>
+            <th>{l('Frequency')}</th>
+            <th>{l('Actions')}</th>
+          </tr>
+        </thead>
+        {attributes
+          .sort((a, b) => compare(a.name.toLowerCase(), b.name.toLowerCase()))
+          .reverse()
+          .sort((a, b) => compare(str(a.frequency), str(b.frequency)))
+          .reverse()
```
yvanzo: `.toLowerCase` calls are unneeded here too.

Using `i18n.compare` (for localized string array) to sort attributes by frequency (of type integer) is a bit expensive (and `str` is not defined), I would suggest using `terable/sortBy` instead, see [SidebarLicenses.js](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-01-22/root/layout/components/sidebar/SidebarLicenses.js) for example.

yvanzo: `terable/sortBy` might be tricky to work with here.
I push 3 commits to use `lodash/orderBy` instead.
Feel free to reuse it.

mwiencek: You can already sort by multiple criteria in a `.sort()` callback. e.g.

```JavaScript
attributes.sort((a, b) => (b.frequency - a.frequency) || compare(a.name, b.name))
```

Will sort by frequency numerically (descending), then by name (ascending - if you want descending there simply swap the argument order).

The `.sort()` callback only needs to return a negative number if `a < b`, or a positive number if `a > b` (0 if they're equal). If `(b.frequency - a.frequency)` is 0 then they have the same frequency, and it'll short-circuit to the `compare` since 0 is falsey.

anirudhjain75: I used lodash on recommendation by yvanzo in the irc, I'd love to discuss about this

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -46,17 +50,17 @@ sub attribute_index : Chained('attribute_base') PathPart('') RequireAuth(account
     my ($self, $c) = @_;
     my $model = $c->stash->{model};
     my @attr = $c->model($model)->get_all();
-
```
yvanzo: Unneeded removal of empty line.

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -46,17 +50,17 @@ sub attribute_index : Chained('attribute_base') PathPart('') RequireAuth(account
...
+    my $component_path = $component_paths{$model} // "admin/attributes/Index.js";
 
     $c->stash(
-        template => $template,
-        attributes => \@attr,
-    );
+    current_view => 'Node',
+    component_path => $component_path,
+    component_props => {attributes => \@attr, model => $model}
+);
```
yvanzo: The four above lines are missing proper indentation too.

root/admin/attributes/Index.js
```diff
@@ -0,0 +1,133 @@
+/*
+ * @flow
+ * eslint-disable flowtype/no-mutable-array
+ * Copyright (C) 2019 Anirudh Jain
+ * Copyright (C) 2019 MetaBrainz Foundation
```
yvanzo: Sorry for not having been clear, it should be:
 * Copyright (C) 2014 MetaBrainz Foundation
 * Copyright (C) 2019 Anirudh Jain

root/types.js
```diff
@@ -575,7 +575,11 @@ export opaque type LabelTypeT: OptionTreeT = OptionTreeT;
 
 declare type LanguageT = {|
   +id: number,
+  +iso_code_1: string | null,
+  +iso_code_2b: string | null,
+  +iso_code_2t: string | null,
   +iso_code_3: string | null,
+  +frequency: number,
```
yvanzo: It should be above `id` to follow coding style.

root/types.js
```diff
@@ -788,7 +792,10 @@ declare type SanitizedEditorT = {|
 |};
 
 declare type ScriptT = {|
+  +id: number,
   +iso_code: string,
+  +iso_number: string | null,
+  +frequency: number,
```
yvanzo: It should be above `id` to follow coding style.

anirudhjain75:
@yvanzo I've made the minor fixes you've recommended.

yvanzo:
@anirudhjain75 I just pushed removing `index.tt` and cosmetic changes, feel free to squash these.

anirudhjain75:
Thanks will make the changes :)

yvanzo:
Thank you @anirudhjain75, I just changed pull request’s title after ticket’s new title, and will use description for merge commit’s message.

PR #981, opened 2019-03-18
MBS-9491/MBS-10062: Add SQL tables for genre

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9491 / https://tickets.metabrainz.org/browse/MBS-10062

Generated the gids with uuid_4 in Perl since we will probably want to do that in the future.


admin/sql/updates/20190317-mbs-9941-mbs-10062.sql
```diff
@@ -0,0 +1,454 @@
...
+ALTER TABLE genre_alias ADD CONSTRAINT genre_alias_pkey PRIMARY KEY (id);
+
+CREATE UNIQUE INDEX genre_idx_gid ON genre (gid);
+CREATE UNIQUE INDEX genre_idx_name ON genre (LOWER(name));
+
+CREATE INDEX genre_alias_idx_genre ON genre_alias (genre);
+CREATE UNIQUE INDEX genre_alias_idx_primary ON genre_alias (genre, locale) WHERE primary_for_locale = TRUE AND locale IS NOT NULL;
+
+INSERT INTO genre (gid, name)
+VALUES  ('54c01942-22fd-4184-9877-1db0089b18f1', 'acid house'),
```
mwiencek: I'll admit pre-generating the gids seems weird, but that's probably because we've never added a new entity that was seeded like this before. I think v4 UUIDs do make sense though, since genres come from the outside world and have no unique name guarantee.

admin/sql/CreateFKConstraints.sql
```diff
@@ -957,6 +957,11 @@ ALTER TABLE gender
    FOREIGN KEY (parent)
    REFERENCES gender(id);
 
+ALTER TABLE genre_alias
```
mwiencek: You'll need an additional fks script (run on standalone servers only) that adds this.

admin/sql/CreateTables.sql
```diff
@@ -852,6 +852,24 @@ CREATE TABLE gender ( -- replicate
...
 );
 
+CREATE TABLE genre ( -- replicate (verbose)
+    id                  SERIAL, -- PK
+    name                VARCHAR NOT NULL,
+    comment             VARCHAR(255) NOT NULL DEFAULT '',
+    gid                 uuid NOT NULL
+);
+
+CREATE TABLE genre_alias ( -- replicate (verbose)
```
mwiencek: Shouldn't we have a `unique_primary_genre_alias` function (and associated trigger) too? (See the other `unique_primary_*_alias` functions in admin/sql/CreateFunctions.sql.)

admin/sql/CreateTables.sql
```diff
@@ -852,6 +852,24 @@ CREATE TABLE gender ( -- replicate
     gid                 uuid NOT NULL
 );
 
+CREATE TABLE genre ( -- replicate (verbose)
+    id                  SERIAL, -- PK
+    gid                 uuid NOT NULL,
+    name                VARCHAR NOT NULL,
+    comment             VARCHAR(255) NOT NULL DEFAULT ''
```
mwiencek: Will genre aliases be managed through the edit system? (I assume so at some point, since the `genre_alias` table has an `edits_pending` column.) If so, we'll need an `edits_pending` column here too, since that's expected to exist by `Edit::Alias::Add` at the very least.

admin/sql/CreateFunctions.sql
```diff
@@ -1100,6 +1100,18 @@ BEGIN
 END;
 $$ LANGUAGE 'plpgsql';
 
+CREATE OR REPLACE FUNCTION unique_primary_genre_alias()
```
mwiencek: Now you need a corresponding `unique_primary_for_locale` trigger that calls this function.

admin/sql/CreateTables.sql
reosarevok: Oh. I guess it should be the edit system? It's a bit weird in the sense that it will all probably be under the admin section at least for now, but if we hope to make them entities eventually (and I think we do) then having the *possibility* at least to do it via the edit system seems sensible.

admin/sql/CreateFunctions.sql
reosarevok: Added. To -fks only, since it seems to have been done like that before, but do confirm.

reosarevok:
> * Still using `tag` and `<entity type>_tag` tables I guess. Would it make sense to have constraints ensuring each `genre` entry has `name in `tag.name`?

I thought about this, but then decided that we should be able to add a genre even if nobody has used it as a tag yet. Seems unlikely, but shouldn't be forbidden.

> * For definition-oriented links, such as `{genre} is associated to {area} from {begin} to {end}`, will we have to wait for another schema change?

Probably. I would like to be able to add URLs to genres at least, but we didn't really have the time to do that this time, so let's improve it one thing at a time :) 

> Nit-picking: reserved word `FALSE` and data type `UUID` should probably be written in uppercase.

I used lowercase uuid because that's what the table before this one had, but now I see a lot of them have UUID and a lot of them have uuid. I guess changing it all to UUID is a bad idea, tempting as it is? :) Same seems true of false vs FALSE. I'll uppercase these, anyway.

yvanzo:
> > How do you plan to link it to other entity types? Still using `tag` and `<entity type>_tag` tables I guess. Would it make sense to have constraints ensuring each genre entry has name in tag.name`?
> 
> I thought about this, but then decided that we should be able to add a genre even if nobody has used it as a tag yet. Seems unlikely, but shouldn't be forbidden.

Getting back to this. If nobody uses it and it does not appear in `tag` list, how will you link to it then?

reosarevok:
That is a fair (implementation) question! It would still appear on the genre list, so it could just appear as a name without a link until it gets used.

PR #982, opened 2019-03-19
MBS-10081: formatUserDate needs datetime_to_iso8601(last_updated)

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10081

### DIFF
--- a/root/entity/Details.js
+++ b/root/entity/Details.js
@@ -19,7 +19,6 @@ import formatUserDate from '../utility/formatUserDate';
 type Props = {|
   +$c: CatalystContextT,
   +entity: CoreEntityT,
-  +lastUpdated: string,
 |};
 
 const XMLLink = ({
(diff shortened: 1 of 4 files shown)


root/entity/Details.js
```diff
@@ -75,7 +73,7 @@ const Details = ({
         <tr>
           <th>{l('Last updated:')}</th>
           <td>
-            {lastUpdated ? formatUserDate($c.user, lastUpdated) : l('(unknown)')}
+            {entity.last_updated ? formatUserDate($c.user, entity.last_updated) : l('(unknown)')}
```
yvanzo: For the record, this works thanks to: https://github.com/metabrainz/musicbrainz-server/blob/cf9270aeea263d200289280a836084c824e63fc4/root/types.js#L321-L323

PR #986, opened 2019-03-28
MBS-10086 tweaks

mwiencek:
I took too long to review #979, so submitting my suggested changes as a PR.

As luck would have it, when I was doing b95f79700d1d980f5faffff1a304ecbb8790ce57 I ran into a crazy Flow issue which made some unrelated and previously-invisible errors visible. I fixed them all in a084116e999c86f6edac0d179ec1ac7bab47add6 and 9a5bad2538d494cb4e2edfe05fa50614a97df74f.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -304,7 +304,7 @@ rules:
   flowtype/generic-spacing: [warn, never]
   flowtype/no-dupe-keys: error
   flowtype/no-flow-fix-me-comments: off
-  flowtype/no-mutable-array: warn
+  flowtype/no-mutable-array: off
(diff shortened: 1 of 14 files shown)


yvanzo:
By the way, you might want to disable eslint’s rule `flowtype/no-mutable-array` in `root/static/scripts/common/i18n/expand2react.js` just like it has been done there:
https://github.com/metabrainz/musicbrainz-server/blob/5911563dc19bcdd3c6955cba8a5b82de0682d26a/root/admin/attributes/Index.js#L3

anirudhjain75:
Thanks... I will keep these changes in mind next time while converting :)

PR #987, opened 2019-03-28
MBS-10087: Convert doc pages to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10087

Talking with @mwiencek we decided to get rid of the bare_ pages, because they didn't seem to be requested ever and we can't find any info of why they were added.


lib/MusicBrainz/Server/Controller/Doc.pm
```diff
@@ -26,24 +27,31 @@ sub show : Path('')
...
-        google_custom_search => DBDefs->GOOGLE_CUSTOM_SEARCH,
-    );
-
     if ($id =~ /^[^:]+:/i && $id !~ /^Category:/i) {
         $c->response->redirect(sprintf('http://%s/%s', DBDefs->WIKITRANS_SERVER, $id));
         $c->detach;
     }
 
+    my %props = (
+        useGoogleCustomSearch   => boolean_to_json(DBDefs->GOOGLE_CUSTOM_SEARCH),
```
reosarevok: We don't care about the Google Custom Search token here, just about whether it exists. As such, it seems better IMO to pass a boolean to JSON rather than the actual token.

lib/MusicBrainz/Server/Controller/Doc.pm
```diff
@@ -26,24 +27,31 @@ sub show : Path('')
...
+            component_path  => 'doc/DocPage',
+            component_props => \%props,
+            current_view    => 'Node',
+        );
+    }    else {
         $c->response->status(404);
-        $c->stash->{template} = $bare ? 'doc/bare_error.tt' : 'doc/error.tt';
+        $c->stash(
+            component_path  => 'doc/DocError',
+            component_props => \%props,
```
reosarevok: "page" and "wikiServer" aren't used here - in fact page will always be undefined. I would like to avoid having two sets of %props though, so hopefully this works fine?

root/doc/DocError.js
```diff
@@ -0,0 +1,67 @@
...
+    <Layout fullWidth title={l('Page Not Found')}>
+      <div className="wikicontent" id="content">
+        {useGoogleCustomSearch ? <DocSearchBox /> : null}
+
+        <h1>{l('Page Not Found')}</h1>
+
+        <p>
+          <strong>
+            {texp.l('Sorry, “{id}” is not a valid documentation page.',
+                    {id: id.replace(/_/g, ' ')})}
```
reosarevok: Need to use a regex because for some reason Javascript will otherwise only replace the first appearance (which seems stupid, but eh).

root/doc/DocPage.js
```diff
@@ -0,0 +1,110 @@
...
+
+const DocPage = ({
+  id,
+  page,
+  useGoogleCustomSearch,
+  wikiServer,
+}: Props) => {
+  let doc = '';
+  let lastDoc = '';
+  const wikiPageUrl = `//${wikiServer}/${id}`;
```
reosarevok: I was going to use encodeURIComponent() here, but it messes up with the slashes, so hopefully we don't need to.

root/doc/DocPage.js
```diff
@@ -0,0 +1,110 @@
...
+      <div className="wikicontent" id="content">
+        {useGoogleCustomSearch ? <DocSearchBox /> : null}
+
+        <h1 className="hierarchy-links">
+          {page.hierarchy.map((link, index, array) => {
+            lastDoc = doc;
+            doc = index === 0 ? link : lastDoc + '/' + link;
+            const isLast = array.length - 1 === index;
+            return (
+              <React.Fragment key={index + '-' + link}>
```
reosarevok: Added index to the key just in case we have something like /Style/Style at some point.

root/doc/DocPage.js
```diff
@@ -0,0 +1,110 @@
...
+
+        {page.version ? null : (
+          <div className="wikidocs-header">
+            {exp.l(`This page has not been reviewed by our documentation team
+                    ({more_info|more info}).`,
+                   {more_info: '/doc/WikiDocs'})}
+          </div>
+        )}
+
+        <div dangerouslySetInnerHTML={{__html: page.content}} />
```
reosarevok: This div didn't exist on TT, but is needed in React - shouldn't be a problem hopefully.

root/main/404.js
```diff
@@ -26,14 +30,14 @@ const _404 = ({$c, ...props}) => (
...
-           {doc: '/doc/MusicBrainz_Documentation', faq: '/doc/FAQ'})}
+               {doc: '/doc/MusicBrainz_Documentation', faq: '/doc/FAQ'})}
       </p>
       <p>
         {exp.l('Found a broken link on our site? Please {report|report a bug} and include any error message that is shown above.', {
           report: bugTrackerURL(
             'Nonexistent page: ' + $c.req.uri + '\n' +
-            'Referrer: ' + ($c.req.headers.referer || '')
-          )
+            'Referrer: ' + ($c.req.headers.Referer || ''),
```
reosarevok: Our headers have uppercased Referer - this doesn't work otherwise.

mwiencek: Hrm, that seems flaky, especially seeing as we reference `req.headers.host` in another file. I think I'd prefer if we explicitly lowercased them on the server during serialization.

lib/MusicBrainz/Server/Controller/Doc.pm
```diff
@@ -26,24 +27,31 @@ sub show : Path('')
...
     if ($id =~ /^[^:]+:/i && $id !~ /^Category:/i) {
         $c->response->redirect(sprintf('http://%s/%s', DBDefs->WIKITRANS_SERVER, $id));
         $c->detach;
     }
 
+    my %props = (
+        useGoogleCustomSearch   => boolean_to_json(DBDefs->GOOGLE_CUSTOM_SEARCH),
+        id                      => $id,
+        page                    => $page,
+        wikiServer              => DBDefs->WIKITRANS_SERVER,
```
mwiencek: Moving this to DBDefs.js rather than passing it as a prop would make sense to me

root/main/404.js
reosarevok: Would that involve         $headers{lc($name)} = $req->headers->header($name); in https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server.pm#L550 ?

lib/MusicBrainz/Server/Controller/Doc.pm
reosarevok: Made that change. Not sure if this should use client or standard DBDefs, went for client since it doesn't need anything that isn't there.

root/main/404.js
mwiencek: Yup, that should work. The header is not case-sensitive in the `->header` call.

lib/MusicBrainz/Server/Controller/Doc.pm
mwiencek: DBDefs-client is mostly only needed if the component is ever used in the browser, so you'd be fine to use the "server" DBDefs here. DBDefs-client isn't wrong either, but I suppose it might be confusing if someone expects it to be the server one since it's a server-only component.

root/main/404.js
reosarevok: Ok, done (I think!)

root/doc/DocError.js
```diff
@@ -0,0 +1,67 @@
...
+                 })}
+        </p>
+
+        <p>
+          {exp.l(`Found a broken link on our site? Please let us know by
+                  {report|reporting a bug}.`,
+                 {
+                   report: bugTrackerURL(
+                     'Broken link: ' + $c.req.uri + '\n' +
+                     'Referrer: ' + $c.req.headers.Referer,
```
mwiencek: So this should be lowercase now that you have the "Lowercase headers during serialization" commit.

lib/MusicBrainz/Server/Controller/Doc.pm
mwiencek: That makes sense. Though `GOOGLE_CUSTOM_SEARCH` is already in DBDefs.js - I'd think you can just import and perform this check in JS rather than passing an extra prop.

root/doc/bare_error.tt
```diff
@@ -1,20 +0,0 @@
-<h1>[%- l('Page Not Found') -%]</h1>
```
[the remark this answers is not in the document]
reosarevok: Can't, because doc_error actually uses it ([% PROCESS 'doc/bare_error.tt' %])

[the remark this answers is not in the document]
reosarevok: https://github.com/metabrainz/musicbrainz-server/blob/d371cc3752a0ff2949b8f778d8d143b0cabbcc25/root/doc/error.tt#L7

PR #990, opened 2019-03-28
MBS-10088: Correctly check for cardinality on l_x_x rels

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10088

The current query used "${source_id}_cardinality = 0" if $use_cardinality; to exclude relationships affected by cardinality.

That works great, except on l_x_x relationships where *both* sides of the relationship need to be checked for cardinality. In that case, only entity1 (the final value of $source_id) was being checked, meaning if entity_1 had cardinality=1 it would always return no results, even if the entity we were looking for was entity0 with cardinality=0.

This changes it so that the condition strings are actually fully built inside the $type checks, which ensures we'll get "WHERE (entity0 IN (?) AND entity0_cardinality = 0) OR (entity1 IN (?) AND entity1_cardinality = 0)" rather than "WHERE (entity0 IN (?) OR entity1 IN (?)) AND entity1_cardinality = 0".

### DIFF
--- a/lib/MusicBrainz/Server/Data/Relationship.pm
+++ b/lib/MusicBrainz/Server/Data/Relationship.pm
@@ -135,14 +135,22 @@ sub _load
         my (@cond, @params, $target, $target_id, $source_id, $query);
 
         if ($type eq $type0) {
-            push @cond, "entity0 IN (" . placeholders(@ids) . ")";
+            my $condstring = "entity0 IN (" . placeholders(@ids) . ")";
+            if ($use_cardinality) {
+                $condstring = "($condstring AND entity0_cardinality = 0)";
+            }
+            push @cond, $condstring;
             push @params, @ids;
             $target = $type1;
             $target_id = 'entity1';
             $source_id = 'entity0';
         }
         if ($type eq $type1) {
-            push @cond, "entity1 IN (" . placeholders(@ids) . ")";
+            my $condstring = "entity1 IN (" . placeholders(@ids) . ")";
+            if ($use_cardinality) {
+                $condstring = "($condstring AND entity1_cardinality = 0)";
+            }
+            push @cond, $condstring;
             push @params, @ids;
             $target = $type0;
             $target_id = 'entity0';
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -135,14 +135,22 @@ sub _load
         my (@cond, @params, $target, $target_id, $source_id, $query);
 
         if ($type eq $type0) {
-            push @cond, "entity0 IN (" . placeholders(@ids) . ")";
+            my $condstring = "entity0 IN (" . placeholders(@ids) . ")";
+            if ($use_cardinality) {
+                $condstring = '(' . $condstring . ' AND entity0_cardinality = 0)'
```
mwiencek: There's a semicolon missing at the end of the line (and I think I'd just use double quotes here instead of concatenation, though it doesn't particularly matter other than saving characters).

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -135,14 +135,22 @@ sub _load
...
             push @params, @ids;
             $target = $type1;
             $target_id = 'entity1';
             $source_id = 'entity0';
         }
         if ($type eq $type1) {
-            push @cond, "entity1 IN (" . placeholders(@ids) . ")";
+            my $condstring = "entity1 IN (" . placeholders(@ids) . ")";
+            if ($use_cardinality) {
+                $condstring = '(' . $condstring . ' AND entity1_cardinality = 0)'
```
mwiencek: Same here

reosarevok:
Well, in that case `$target` will just be `$type0`, but since both types have the same value, it doesn't really matter - the important thing is that it will be "artist" or whatever and will work. Unless I'm missing something.

mwiencek:
Oops, I meant more `$target_id` than `$target`, since it uses that to join to the entity table for sorting (by name). `$target_id` may not correspond to the correct entity when cardinality is in use.

reosarevok:
Oh. I thought about that, and at first was like "oh, I should join both if they're both the same entity type!" - but that just meant the query gave the rels twice. So I decided not to touch it. Open to ideas on how to improve it though. That wouldn't be connected to cardinality though, would it? If there's an issue, there's an issue with every l_x_x table query.

mwiencek:
Yeah, you are right it isn't connected to cardinality, I suppose it just made it more obvious to me. I think the only real way to fix it is, we'd have to join the target table twice (for both entity0 and entity1) if the types are the same, then use some `CASE` statement in the `ORDER BY` to figure out which name to use (but that might be crazy and not work at all). Anyway, this can be merged on its own.

PR #991, opened 2019-03-29
Fixing all cases of 'occured' to 'occurred'

reosarevok:
It was annoying me.


MonkeyDo:
I want to leave a trace to say "I concurred".

PR #992, opened 2019-04-01
MBS-10095: Meaningful error for editing nonexisting user

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10095

400 seemed appropriate, since the username being requested is incorrect. 404 would work too but it doesn't seem to accept any sort of extra explanatory message, so it seems less ideal. Of course, we could always add an optional extra message to our 404 pages :)

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Root.pm
+++ b/lib/MusicBrainz/Server/Controller/Root.pm
@@ -150,10 +150,11 @@ sub error_403 : Private
 }
 
 sub error_404 : Private {
-    my ($self, $c) = @_;
+    my ($self, $c, $message) = @_;
 
     $c->response->status(404);
     $c->stash->{current_view} = 'Node';
+    $c->stash->{component_props}{message} = $message;
 }
 
 sub error_500 : Private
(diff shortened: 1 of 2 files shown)


reosarevok:
Seems so, although we weren't actually passing it anywhere. Changed that.

reosarevok:
That seems like it'd make sense, made the change. We'll need to specifically pass the right parameter after merging https://github.com/metabrainz/musicbrainz-server/pull/1009 but that should be doable :)

lib/MusicBrainz/Server/Controller/Root.pm
```diff
@@ -154,6 +154,7 @@ sub error_404 : Private {
 
     $c->response->status(404);
     $c->stash->{current_view} = 'Node';
+    $c->stash->{component_props}->{message} = $c->stash->{message} if $c->stash->{message};
```
mwiencek: Instead of having to add the message to the stash, we can accept a `$message` argument here. You can pass the message like `$c->detach('/error_404', ['message'])`. That seems easier to follow to me (because if you add something to the stash, it's hard to tell what's using it or where it's being used later).

reosarevok: Done!

PR #993, opened 2019-04-02
MBS-10094: Right-align barcodes in lists

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10094

IMO it makes no sense to do it like that only on one kind of release list - since I feel it makes sense, I'm making the change in all lists of releases (of which we clearly have way too many).

If you'd rather have a class "barcode-row" or something and then the text-align on CSS, that works for me too.

### DIFF
--- a/root/cdtoc/attach_list.tt
+++ b/root/cdtoc/attach_list.tt
@@ -10,7 +10,7 @@
     <td>[% release_countries(release.events) %]</td>
     <td>[% release_label_list(release.labels) %]</td>
     <td>[% release_catno_list(release.labels) %]</td>
-    <td>[% release.barcode.format %]</td>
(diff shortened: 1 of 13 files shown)


reosarevok:
If this gets merged I'll change https://github.com/metabrainz/musicbrainz-server/pull/975 accordingly :)

yvanzo:
> On IRC I suggested a generic `.align-right` might be useful, but I don't have any real objections to an inline style.

I cannot find this in chat logs, but I found 8 barcode cells being styled in this PR.

* With generic class name, it would be tedious to change style for every barcode cell later on.
* With content specific class name, it just require to change style once.

Thus I would still prefer content specific class name.

PR #996, opened 2019-04-05
MBS-10099: Use backwards link phrase when needed on ExternalLinks (+ eslint fixes)

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10099

It also includes a ton of eslint fixes on the file, moving the amount of failures from like a hundred to 11 warnings. The rest would require a more significant rewrite and/or me understanding refs and callbacks and state stuff which I'm still working on. But it should already be quite an improvement.

Edit: around half the remaining warnings are for react/jsx-no-bind which @mwiencek suggested we might want to stop applying.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -105,9 +105,11 @@ export class ExternalLinksEditor
   }
 
   removeLink(index: number) {
-    const newLinks = this.state.links.concat();
-    newLinks.splice(index, 1);
-    this.setState({links: newLinks}, () => {
+    this.setState(prevState => {
+      const newLinks = prevState.links.concat();
+      newLinks.splice(index, 1);
+      return {links: newLinks};
+    }, () => {
       $(ReactDOM.findDOMNode(this))
         .find('tr:gt(' + (index - 1) + ') button.remove:first, ' +
               'tr:lt(' + (index + 1) + ') button.remove:last')
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -584,7 +584,7 @@ MB.createExternalLinksEditor = function (options: InitialOptionsT) {
     const urls = {};
     let match;
 
-    while (match = seededLinkRegex.exec(window.location.search)) {
+    while ((match = seededLinkRegex.exec(window.location.search))) {
       const [/* unused */, index, key, value] = match;
       (urls[index] = urls[index] || {})[key] = decodeURIComponent(value);
     }
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -278,6 +278,7 @@ export class ExternalLink extends React.Component<LinkProps> {
     var linkType = props.type ? linkedEntities.link_type[props.type] : null;
     var typeDescription = '';
     var faviconClass: string | void;
+    var backward = linkType && linkType.type1 > 'url';
 
     if (linkType && linkType.description) {
       typeDescription = exp.l('{description} ({url|more documentation})', {
@@ -313,7 +314,11 @@ export class ExternalLink extends React.Component<LinkProps> {
               </LinkTypeSelect>
             : <label>
                 {faviconClass && <span className={'favicon ' + faviconClass + '-favicon'}></span>}
-                {(linkType ? l_relationships(linkType.link_phrase) : null) ||
+                {(linkType ? (
+                  backward
+                    ? l_relationships(linkType.reverse_link_phrase)
+                    : l_relationships(linkType.link_phrase)
+                ) : null) ||
                   (props.isOnlyLink ? l('Add link:') : l('Add another link:'))}
               </label>}
         </td>
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -582,8 +582,8 @@ MB.createExternalLinksEditor = function (options: InitialOptionsT) {
     let match;
 
     while (match = seededLinkRegex.exec(window.location.search)) {
-      (urls[match[1]] = urls[match[1]] || {})[match[2]] =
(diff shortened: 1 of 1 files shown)


root/static/scripts/edit/externalLinks.js
```diff
@@ -577,8 +577,8 @@ MB.createExternalLinksEditor = function (options: InitialOptionsT) {
     let match;
 
     while (match = seededLinkRegex.exec(window.location.search)) {
-      (urls[match[1]] = urls[match[1]] || {})[match[2]] =
-        decodeURIComponent(match[3]);
+      const [, index, key, value] = match;
```
yvanzo: An inline comment within `[,` would clarify that `match[0]` is not used at purpose.

root/static/scripts/edit/externalLinks.js
```diff
@@ -579,7 +579,7 @@ MB.createExternalLinksEditor = function (options: InitialOptionsT) {
     const urls = {};
     let match;
 
-    while (match = seededLinkRegex.exec(window.location.search)) {
+    while ((match = seededLinkRegex.exec(window.location.search))) {
```
yvanzo: I don’t get how it works around `no-cond-assign`.

root/static/scripts/edit/externalLinks.js
```diff
@@ -310,17 +315,19 @@ class LinkTypeSelect extends React.Component<LinkTypeSelectProps> {
...
   type: number | null,
   typeChangeCallback: (number, SyntheticEvent<HTMLSelectElement>) => void,
   typeOptions: Array<React.Element<'option'>>,
   url: string,
-  urlBlurCallback: (number, SyntheticEvent<HTMLInputElement>) => void,
-  urlChangeCallback: (number, SyntheticEvent<HTMLInputElement>) => void,
   urlMatchesType: boolean,
   video: boolean,
-  videoChangeCallback: (number, SyntheticEvent<HTMLInputElement>) => void,
+
```
yvanzo: Unneeded blank line?

root/static/scripts/edit/externalLinks.js
reosarevok: https://github.com/eslint/eslint/blob/master/docs/rules/no-cond-assign.md - "except-parens" (default) allows assignments in test conditions only if they are enclosed in parentheses (for example, to allow reassigning a variable in the test of a while or do...while loop)

root/static/scripts/edit/externalLinks.js
reosarevok: Changed, let me know if it looks ok

yvanzo: it looks ok, thanks

root/static/scripts/edit/externalLinks.js
```diff
@@ -209,24 +211,19 @@ export class ExternalLinksEditor
               error = l(`Please select a link type for the URL
                          you’ve entered.`);
             } else if (
-              linkType.deprecated &&
-                (!isPositiveInteger(link.relationship) ||
-                  (oldLink && +link.type !== +oldLink.type))
+              linkType.deprecated && (isNewLink || linkChanged)
```
mwiencek: The previous code checked that the type changed, not the URL.

root/static/scripts/edit/externalLinks.js
```diff
@@ -118,32 +124,32 @@ export class ExternalLinksEditor extends React.Component<LinksEditorProps, Links
...
   }
 
   getFormData(startingPrefix: string,
-              startingIndex: number,
-              pushInput: (string, string, string) => void) {
-    var index = 0;
-    var backward = this.props.sourceType > 'url';
-    var {oldLinks, newLinks, allLinks} = this.getEditData();
+    startingIndex: number,
+    pushInput: (string, string, string) => void) {
```
mwiencek: The indentation made more sense before, I think? Now only `startingIndex` is on its own line and `startingPrefix` isn't lined up with the ones below. I guess they should all have their own line if eslint wants that.

```JS
  getFormData(
    startingPrefix: string,
    startingIndex: number,
    pushInput: (string, string, string) => void,
  ) {
```

root/static/scripts/edit/externalLinks.js
```diff
@@ -28,40 +30,41 @@ import * as URLCleanup from './URLCleanup';
...
   constructor(props: LinksEditorProps) {
     super(props);
     this.state = {links: withOneEmptyLink(props.initialLinks)};
   }
 
   setLinkState(index: number,
-               state: $Shape<LinkStateT>,
-               callback?: () => void) {
+    state: $Shape<LinkStateT>,
+    callback?: () => void) {
```
mwiencek: Heh, this looks weird now too, it was more readable before. Putting them all on their own line as suggested for `getFormData` would make sense to me.

root/static/scripts/edit/externalLinks.js
```diff
@@ -259,6 +259,15 @@ export class ExternalLinksEditor
             return (
               <ExternalLink
                 errorMessage={error || ''}
+                handleUrlBlurCallback={
```
mwiencek: I'd also remove `Callback` from all the names, since `handle` already indicates it's a handler, and callback doesn't really make sense combined with it (sounds like the handler is supposed to call it or something, but it's the handler itself).

reosarevok: I think I fixed them all now. It does have handleUrlBlur binding this.handleUrlBlur, but that's probably fine?

root/static/scripts/edit/externalLinks.js
reosarevok: Whoops. Created a different constant for this then. What's with the + here though? I just realized those were there. Are they needed and why?

root/static/scripts/edit/externalLinks.js
mwiencek: Yeah, looks better now.

PR #997, opened 2019-04-05
MBS-10100: Don't require non-required fields for language

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10100

This might not be the right way to do this (it works here, but it might break something else). If so, please do let me know what would be the right way - Perl forms still confuse me.

Also includes https://tickets.metabrainz.org/browse/MBS-10210 and https://tickets.metabrainz.org/browse/MBS-10215

### DIFF
--- a/lib/MusicBrainz/Server/Form/Admin/Attributes/Language.pm
+++ b/lib/MusicBrainz/Server/Form/Admin/Attributes/Language.pm
@@ -2,6 +2,9 @@ package MusicBrainz::Server::Form::Admin::Attributes::Language;
 
 use HTML::FormHandler::Moose;
 
+use MusicBrainz::Server::Constants qw( :language_frequency );
+use MusicBrainz::Server::Translation qw( lp );
+
 extends 'MusicBrainz::Server::Form';
 with 'MusicBrainz::Server::Form::Role::Edit';
 
@@ -37,8 +40,17 @@ has_field 'iso_code_3' => (
 );
 
 has_field 'frequency' => (
-    type => '+MusicBrainz::Server::Form::Field::Integer',
+    type => 'Select',
+    required => 1
 );
+
+sub options_frequency {
+    return [
+        $LANGUAGE_FREQUENCY_HIDDEN, lp('Hidden', 'language optgroup'),
+        $LANGUAGE_FREQUENCY_OTHER, lp('Other', 'language optgroup'),
+        $LANGUAGE_FREQUENCY_FREQUENT, lp('Frequently used', 'language optgroup'),
+    ]
+}
 
 1;
 
--- a/lib/MusicBrainz/Server/Form/Admin/Attributes/Script.pm
+++ b/lib/MusicBrainz/Server/Form/Admin/Attributes/Script.pm
@@ -26,6 +26,7 @@ has_field 'iso_code' => (
 
 has_field 'iso_number' => (
     type => 'Text',
+    required  => 1,
     maxlength => 3,
 );
 
--- a/lib/MusicBrainz/Server/Form/Admin/Attributes/Script.pm
+++ b/lib/MusicBrainz/Server/Form/Admin/Attributes/Script.pm
@@ -2,6 +2,9 @@ package MusicBrainz::Server::Form::Admin::Attributes::Script;
 
 use HTML::FormHandler::Moose;
 
+use MusicBrainz::Server::Constants qw( :script_frequency );
+use MusicBrainz::Server::Translation qw( lp );
+
 extends 'MusicBrainz::Server::Form';
 with 'MusicBrainz::Server::Form::Role::Edit';
 
@@ -27,8 +30,18 @@ has_field 'iso_number' => (
 );
 
 has_field 'frequency' => (
-    type => '+MusicBrainz::Server::Form::Field::Integer',
+    type => 'Select',
+    required => 1
 );
+
+sub options_frequency {
+    return [
+        $SCRIPT_FREQUENCY_HIDDEN, lp('Hidden', 'script frequency'),
+        $SCRIPT_FREQUENCY_UNCOMMON, lp('Other (Uncommon)', 'script frequency'),
+        $SCRIPT_FREQUENCY_OTHER, lp('Other', 'script frequency'),
+        $SCRIPT_FREQUENCY_FREQUENT, lp('Frequently used', 'script frequency'),
+    ]
+}
 
 1;
 
--- a/root/admin/attributes/form.tt
+++ b/root/admin/attributes/form.tt
@@ -50,7 +50,7 @@
         [% WRAPPER form_row %]
             [% r.label('child_order', add_colon(l('Child order'))) %]
             [% r.text('child_order', { size => 5 }) %]
-            [% form_field_errors(form, 'child_order') %]
+            [% field_errors(form, 'child_order') %]
         [% END %]
 
         [% form_row_text(r, 'name', add_colon(l('Name'))) %]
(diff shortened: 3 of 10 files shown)


root/admin/attributes/form.tt
```diff
@@ -7,7 +7,11 @@
         [% form_row_text(r, 'iso_code_2b', add_colon(l('ISO 639-2/B'))) %]
         [% form_row_text(r, 'iso_code_2t', add_colon(l('ISO 639-2/T'))) %]
         [% form_row_text(r, 'iso_code_3', add_colon(l('ISO 639-3'))) %]
-        [% form_row_text(r, 'frequency', add_colon(l('Frequency'))) %]
+        [% WRAPPER form_row %]
+            [% r.label('frequency', add_colon(l('Frequency'))) %]
+            [% r.number('frequency') %]
+            [% field_errors(form, 'frequency') %]
+        [% END %]
```
reosarevok: It's an int, so this seems preferable? (also not sure why we don't have form_row_number or form_row_int as a macro, it seems to be used in quite a few places)

lib/MusicBrainz/Server/Form/Admin/Attributes/Language.pm
```diff
@@ -38,8 +39,23 @@ has_field 'iso_code_3' => (
 
 has_field 'frequency' => (
     type => '+MusicBrainz::Server::Form::Field::Integer',
+    default => 0,
 );
 
+after validate => sub {
+    my ($self) = @_;
+
+    my @allowed_frequencies = [0, 1, 2];
```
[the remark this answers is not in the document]
reosarevok: Removed it and added range_start and range_end.

root/admin/attributes/form.tt
yvanzo: Seems unrelated to MBS-10100, and should probably be wrapped in `form_row_select` with options labelled after https://github.com/metabrainz/musicbrainz-server/blob/v-2019-05-13-schema-change/root/admin/attributes/form.tt#L12-L18

lib/MusicBrainz/Server/Form/Admin/Attributes/Language.pm
```diff
@@ -1,6 +1,7 @@
 package MusicBrainz::Server::Form::Admin::Attributes::Language;
 
 use HTML::FormHandler::Moose;
+use MusicBrainz::Server::Translation qw( l );
```
yvanzo: Not sure for what reason this line has been added and how is it related to MBS-10100.

reosarevok: It's no longer needed since we changed validation, will remove.

root/admin/attributes/form.tt
reosarevok: Split it into its own commit now. Not sure about the select, what's the big problem with having it as an int field just as it was?

yvanzo: `select` prevents entering invalid values and replaces integer with human-readable label.

The same reasons we don’t have an int field for language, vote, work type, and so on.

reosarevok: Yes, but we auto-generate the select options from the DB for those. I guess we could give this a static 0, 1, 2 set of options, if you want, I can try to figure it out. I still feel the current text description for frequency under the field is better than a shortened select label though, and adding the entire lines as a select would look weird.

reosarevok: (the range we added here already should prevent adding invalid values)

yvanzo: Not for all of those, vote options are defined [in constants](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-06-03/lib/MusicBrainz/Server/Constants.pm#L304-L307), not from the database.

Labels could be “Frequent”, “Other”, “Hidden”. The rest of the explanation for 0 could be kept apart. Then, the field should probably renamed “Display Group” or something.

reosarevok: Oh! That might make sense then. I'll look into it!

reosarevok: Changed it to use a select, please review :)

lib/MusicBrainz/Server/Form/Admin/Attributes/Language.pm
```diff
@@ -37,9 +40,18 @@ has_field 'iso_code_3' => (
...
-    type => '+MusicBrainz::Server::Form::Field::Integer',
+    type => 'Select',
+    required => 1
 );
 
+sub options_frequency {
+    return [
+        $LANGUAGE_FREQUENCY_HIDDEN, lp('Hidden', 'language_script_frequency'),
+        $LANGUAGE_FREQUENCY_OTHER, lp('Other', 'language_script_frequency'),
+        $LANGUAGE_FREQUENCY_FREQUENT, lp('Frequent', 'language_script_frequency'),
```
yvanzo: 1. There is no need to put underscores in message context, it supports spaces;
2. Terms `HIDDEN` and `OTHER` refer to “display group” but are used for “frequency”, it might be more appropriate to replace these with `UNUSED` and `RARE` respectively;
3. Label 'Frequent' could probably be replaced with 'Frequently used' as in [Form::Utils](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-06-03/lib/MusicBrainz/Server/Form/Utils.pm#L60).

lib/MusicBrainz/Server/Form/Admin/Attributes/Script.pm
```diff
@@ -27,9 +30,19 @@ has_field 'iso_number' => (
...
-    type => '+MusicBrainz::Server::Form::Field::Integer',
+    type => 'Select',
+    required => 1
 );
 
+sub options_frequency {
+    return [
+        $SCRIPT_FREQUENCY_HIDDEN, lp('Hidden', 'language_script_frequency'),
+        $SCRIPT_FREQUENCY_UNCOMMON, lp('Uncommon', 'language_script_frequency'),
+        $SCRIPT_FREQUENCY_OTHER, lp('Other', 'language_script_frequency'),
```
yvanzo: `OTHER` should come before `UNCOMMON` to follow ascending order (as in constants).

root/admin/attributes/form.tt
```diff
@@ -22,14 +21,12 @@
...
-        [% form_row_text(r, 'frequency', add_colon(l('Frequency'))) %]
+        [% form_row_select(r, 'frequency', add_colon(l('Frequency'))) %]
         <p>
             [% l('Frequency notes:') %]
             <ul>
-                <li>[% l('4: Shown in the commonly used section') %]</li>
-                <li>[% l('3: Shown in the "other" section, used for scripts which are likely to get some usage') %]</li>
-                <li>[% l('2: Shown in the "other" section, used for scripts in Unicode which are unlikely to be used') %]</li>
-                <li>[% l('1: Hidden, used for scripts not in Unicode') %]</li>
+                <li>[% l('Both Other and Uncommon are shown in the "other" section, but Uncommon should be used for scripts in Unicode which are unlikely to be used.') %]</li>
```
yvanzo: You set `OTHER` to 2 in constants but here document to use `UNCOMMON` (set to 3 in constants) instead for scripts unlikely to be used.

(Scripts of frequency 2 and 3 are mixed together in “Other” option group anyway.)

lib/MusicBrainz/Server/Form/Admin/Attributes/Language.pm
reosarevok: I replaced frequent but I'd leave other, specifically so we can also reuse the other 'language optgroup' string (instead of adding a new context for this).

root/admin/attributes/form.tt
reosarevok: Whoops. Put it upside down in constants. Fixed.

lib/MusicBrainz/Server/Form/Admin/Attributes/Script.pm
reosarevok: It was wrong in constants :) Fixed.

reosarevok:
I guess it could - these changes have already messed up https://github.com/metabrainz/musicbrainz-server/pull/1049 so I guess it doesn't hurt anymore to poke into this stuff more :)

PR #999, opened 2019-04-07
 MBS-10101 / MBS-9082: SecondHandSongs cleanup and autoselect additions

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10101
https://tickets.metabrainz.org/browse/MBS-9082

I still suck at regexes, so please take a good look at those :)

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1818,8 +1818,13 @@ const CLEANUPS = {
   'secondhandsongs': {
     match: [new RegExp('^(https?://)?([^/]+\\.)?secondhandsongs\\.com/', 'i')],
     type: LINK_TYPES.secondhandsongs,
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:[^/]+\.)?secondhandsongs\.com/, 'https://secondhandsongs.com');
+      url = url.replace(/^(https:\/\/secondhandsongs\.com\/\w+\/[\d+]+)[\/#?-].*$/, '$1');
+      return url;
+    },
     validate: function (url, id) {
-      const m = /secondhandsongs\.com\/([a-z]+)\//.exec(url);
+      const m = /^https:\/\/secondhandsongs\.com\/(\w+)\/[\d+]+$/.exec(url);
       if (m) {
         const prefix = m[1];
         switch (id) {
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -165,6 +165,8 @@ export const LINK_TYPES = {
   },
   secondhandsongs: {
     artist: '79c5b84d-a206-4f4c-9832-78c028c312c3',
+    label: 'e46c1166-2aae-4623-ade9-34bd067dfe02',
+    recording: 'a98fb02f-f289-4778-b34e-2625d922e28f',
     release: '0e555925-1b7d-475c-9b25-b9c349dcc3f3',
     work: 'b80dff64-9560-445a-b824-c8b432d77a52',
   },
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1818,6 +1818,11 @@ const CLEANUPS = {
   'secondhandsongs': {
     match: [new RegExp('^(https?://)?([^/]+\\.)?secondhandsongs\\.com/', 'i')],
     type: LINK_TYPES.secondhandsongs,
+    clean: function (url) {
+      url = url.replace(/^(https?:\/\/)?([^/]+\.)?secondhandsongs\.com/, 'https://secondhandsongs.com');
```
yvanzo: Capturing groups `(…)` in this regular expression are not used in the replacement string, thus these should be replaced with non-capturing groups `(?:…)`.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1818,6 +1818,11 @@ const CLEANUPS = {
...
+      url = url.replace(/secondhandsongs\.com\/(\w+)\/([\d+]+)(?:[\/#?-].*)$/, 'secondhandsongs.com/$1/$2');
```
yvanzo: * URL should probably match from the beginning using `^https:\/\/`
* A single capturing group could be used for the whole not trimmed part and just have `'$1'` for the replacement string

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2603,33 +2603,38 @@ const testData = [
...
-                     input_url: 'http://www.secondhandsongs.com/release/888',
+                     input_url: 'http://secondhandsongs.com/release/888',
              input_entity_type: 'release',
     expected_relationship_type: 'secondhandsongs',
+            expected_clean_url: 'https://secondhandsongs.com/release/888',
        only_valid_entity_types: ['release'],
   },
   {
-                     input_url: 'http://www.secondhandsongs.com/work/1409',
+                     input_url: 'https://secondhandsongs.com/work/140348/adaptations#nav-entity',
```
yvanzo: Changing scheme `http` and authority `www.…` parts is weakening test here.
If more test variations are needed, `label` and `performance` tests can be changed instead.

reosarevok: I thought if this was tested once (on the artist one) we didn't need to test it more, but I guess it can't hurt. Will change.

yvanzo: Not changing it should allow to test cleaning a combination of unclean scheme/authority part(s) and unclean path/query/fragment part(s).

reosarevok:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1816,8 +1816,13 @@ const CLEANUPS = {
     },
   },
   'secondhandsongs': {
-    match: [new RegExp('^(https?://)?([^/]+\\.)?secondhandsongs\\.com/', 'i')],
+    match: [new RegExp('^(https?://)?(?:[^/]+\\.)?secondhandsongs\\.com/', 'i')],
```
yvanzo: If you go this direction, none of the groups in `match` rules needs to be capturing, but that should probably be done in a separate commit.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1816,8 +1816,13 @@ const CLEANUPS = {
...
     type: LINK_TYPES.secondhandsongs,
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:[^/]+\.)?secondhandsongs\.com/, 'https://secondhandsongs.com');
+      url = url.replace(/^https:\/\/secondhandsongs\.com\/(\w+\/[\d+]+)[\/#?-].*$/, 'https://secondhandsongs.com/$1');
```
yvanzo: Capturing group could start with `https`… so that replacement string just needs to be `'$1'`.

PR #1000, opened 2019-04-07
MBS-9732: Clarify conditions for edits and voting

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9732

This also removes a bunch of code related to data quality and edits which should now be irrelevant: https://tickets.metabrainz.org/browse/MBS-10197


root/utility/edit.js
```diff
@@ -55,19 +54,36 @@ export function getEditStatusName(edit: EditT) {
...
+      return l(
+        'This edit is open and awaiting votes before it can be applied.',
+      );
     case EDIT_STATUS_APPLIED:
       return l('This edit has been successfully applied.');
     case EDIT_STATUS_FAILEDVOTE:
-      return l('This edit failed because there were insufficient "yes" votes.');
+      return l(
+        'This edit failed because there were insufficient "yes" votes.',
+      );
```
Freso: FWIW, formatting changes like these make it a lot harder to review wording changes, as it makes it a lot harder to tell when it’s just a formatting change and when wording has actually changed as well.

reosarevok: That's why it's on a different commit :) d3a8d3f4bc48e9be77382961879efb683db4dd64 and bcc306703d43645af5ade7c9a2a9486afc23c999 are the only ones which affect display, unless I messed up somewhere.

root/edit/components/EditSidebar.js
```diff
@@ -75,7 +75,7 @@ const EditSidebar = ({$c, edit}: Props) => (
         )}
       </SidebarProperty>
 
-      <SidebarProperty className="" label={l('Conditions:')}>
+      <SidebarProperty className="" label={l('If no votes cast:')}>
```
mwiencek: Isn't "If not enough votes cast" more accurate to say?

root/utility/edit.js
```diff
@@ -119,7 +135,7 @@ export function editorMayApprove(edit: EditT, editor: ?EditorT): boolean {
     case EDIT_SERIES_EDIT:
       const oldOrderingType = get(edit, 'data.old.ordering_type_id', 0);
       const newOrderingType = get(edit, 'data.new.ordering_type_id', 0);
-      if (oldOrderingType != newOrderingType) {
+      if (oldOrderingType !== newOrderingType) {
```
mwiencek: I used != there intentionally since I've seen some edit data store numbers as strings for no reason (though I don't remember if this particular edit type does).

Though if you want to do `String(oldOrderingType) !== String(newOrderingType)` then that's fine too, if more verbose.

reosarevok: Oh, ok. Can revert :) (eventually we could fix those and the old data, but that sounds like an awful lot of effort for something small).

root/edit/components/EditSidebar.js
reosarevok: I guess. It's just long, but we can probably live with long.

reosarevok: Wait, actually, is it? I thought getEditExpireAction only mattered if no votes at all were cast, right? (well, I guess nowadays it doesn't really matter anywhere, but for old edits it did).

mwiencek: Checked the code and you're right. Technically it'll also perform the expire action if only abstain votes are cast, but that's basically "no votes" rather than "not enough." https://github.com/metabrainz/musicbrainz-server/blob/9fe476c5f6bc0ae4b80c5e532a99bce0dae7a069/lib/MusicBrainz/Server/EditQueue.pm#L211

root/edit/components/EditSidebar.js
```diff
@@ -64,9 +63,7 @@ const EditSidebar = ({$c, edit}: Props) => (
         </SidebarProperty>
       )}
 
-      <SidebarDataQuality quality={edit.quality} />
-
-      <SidebarProperty className="" label={l('Requires:')}>
+      <SidebarProperty className="" label={l('For quicker closing:')}>
```
yvanzo: New messages should not end with colon:
```suggestion
      <SidebarProperty className="" label={addColonText(l('For quicker closing'))}>
```

root/edit/components/EditSidebar.js
```diff
@@ -75,7 +72,7 @@ const EditSidebar = ({$c, edit}: Props) => (
         )}
       </SidebarProperty>
 
-      <SidebarProperty className="" label={l('Conditions:')}>
+      <SidebarProperty className="" label={l('If no votes cast:')}>
```
yvanzo: New messages should not end with colon:
```suggestion
      <SidebarProperty className="" label={addColonText(l('If no votes cast'))}>
```

reosarevok:
Made the changes, let me know how it looks :)

PR #1003, opened 2019-04-09
MBS-9849: Convert RG index page to React

reosarevok:
If I'm not mistaken, this is the last bit needed for https://tickets.metabrainz.org/browse/MBS-9849

A bunch of changes here were made by @mwiencek to get rid of flow errors in files that were otherwise unchanged, but started erroring when used here.


root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,214 @@
...
+  return (
+    <>
+      <tr className="subh">
+        <CatalystContext.Consumer>
+          {($c: CatalystContextT) => (
+            $c.user_exists
+              ? <th />
+              : null
+          )}
+        </CatalystContext.Consumer>
```
reosarevok: If there is a way to either use withCatalystContext or pass $c here, we might prefer that to calling CatalystContext.Consumer, which is less readable. Let me know!

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,213 @@
...
+                ) : null
+            )}
+          </CatalystContext.Consumer>
+          <td>
+            <EntityLink entity={release} />
+          </td>
+          <td>{release.combined_format_name || l('[missing media]')}</td>
+          <td>{release.combined_track_count || '-'}</td>
+          <td>
+            <ReleaseDates events={release.events} />
```
mwiencek: Not sure why, but the ul inside the `ReleaseDates` component is missing the "nowrap" class (which the TT macro had), so the dates get squished in the table.

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,213 @@
...
+                  <th>{l('Country')}</th>
+                  <th>{l('Label')}</th>
+                  <th>{l('Catalog#')}</th>
+                  <th>{l('Barcode')}</th>
+                  {$c.session && $c.session.tport
+                    ? <th>{l('Tagger')}</th> : null}
+                </tr>
+              </thead>
+              <tbody>
+                {releases.map(buildReleaseStatusTable)}
```
mwiencek: I would do
```suggestion
                {releases.map(r => buildReleaseStatusTable($c, r))}
```
here and remove all the `CatalystContext` stuff from `buildReleaseStatusTable`.

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,213 @@
...
+      entity={releaseGroup}
+      numberOfRevisions={numberOfRevisions}
+    />
+    <WikipediaExtract
+      cachedWikipediaExtract={wikipediaExtract || null}
+      entity={releaseGroup}
+    />
+    {releases.length ? (
+      <>
+        {releaseGroup.typeName ? <h2>{l(releaseGroup.typeName)}</h2> : null}
```
mwiencek:
```suggestion
        {releaseGroup.typeName ? (
          <h2>{lp_attributes(releaseGroup.typeName, 'release_group_primary_type')}</h2>
        ) : null}
```

reosarevok: Actually that wouldn't work, but neither would my version. Changed to releaseGroupType(releaseGroup) since this has to support type combos.

root/release_group/ReleaseGroupIndex.js
reosarevok: Added the class

root/release_group/ReleaseGroupIndex.js
mwiencek: Thanks, I forgot l_type_name did that

mwiencek:
Synced the Flow types with what's in https://github.com/metabrainz/musicbrainz-server/pull/1005 so they don't conflict.

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,196 @@
...
+      <tr className="subh">
+        {$c.user_exists ? <th /> : null}
+        <th colSpan={$c.session && $c.session.tport ? 9 : 8}>
+          {status && status.name
+            ? lp_attributes(status.name, 'release_status')
+            : l('(unknown)')}
+        </th>
+      </tr>
+      {releaseStatusGroup.map((release, index) => (
+        <tr className={loopParity(index)} key={release.id}>
```
yvanzo: Just notice now that results converted to React have their first entry shaded whereas templates did shade the second entry. That is probably because first entry is of index 0 and thus marked as “even”. This was less noticeable with search results as these usually count more than one entry. Not sure it such an issue, anyway can be fixed globally in a later pull request if necessary.

reosarevok: Yeah, I noticed this when I converted one of my first tables - it's always been like this though for React, so by now it probably doesn't matter in any way :)

lib/MusicBrainz/Server/Controller/ReleaseGroup.pm
```diff
@@ -10,6 +10,8 @@ use MusicBrainz::Server::Constants qw(
     $EDIT_RELEASEGROUP_SET_COVER_ART
     %ENTITIES
 );
+
```
yvanzo: extraneous blank line

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,196 @@
...
+    {eligibleForCleanup ? (
+      <p className="cleanup">
+        {l(
+          `This release group has no relationships or releases associated,
+           and will be removed automatically in the next few days. If this
+           is not intended, please add more data to this release group.`,
+        )}
+      </p>
+    ) : null
+    }
```
yvanzo: should be on the previous line

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,196 @@
...
+            ) : null}
+        </div>
+      </>
+    )}
+    {manifest.js('release-group/index.js', {async: 'async'})}
+  </ReleaseGroupLayout>
+);
+
+export default withCatalystContext(ReleaseGroupIndex);
+
```
yvanzo: extraneous blank line (eslint complains about)

root/static/scripts/common/components/Annotation.js
```diff
@@ -19,17 +19,24 @@ import entityHref from '../utility/entityHref';
 import Collapsible from './Collapsible';
 import EditorLink from './EditorLink';
 
+type EntityProp = $ReadOnly<MinimalCoreEntityT & {
```
yvanzo: It’s not really a component property set, it looks more like an entity type, could be named `MinimalAnnotativeEntityT` or something.

root/static/scripts/common/components/Annotation.js
```diff
@@ -19,17 +19,24 @@ import entityHref from '../utility/entityHref';
...
+  +latest_annotation?: AnnotationT,
+}>;
+
 type Props = {|
   +$c: CatalystContextT | SanitizedCatalystContextT,
-  annotation: ?{...AnnotationT, editor: SanitizedEditorT},
+  +annotation: ?({+editor: EditorT | SanitizedEditorT | null} & AnnotationT),
```
yvanzo: Used twice already and likely to be used for other entity types index, could be its own type.

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,196 @@
...
+type Props = {|
+  +$c: CatalystContextT,
+  +eligibleForCleanup: boolean,
+  +mostPopularReview: CritiqueBrainzReviewT,
+  +mostRecentReview: CritiqueBrainzReviewT,
+  +numberOfRevisions: number,
+  +pager: PagerT,
+  +releaseGroup: ReleaseGroupT,
+  +releases: $ReadOnlyArray<$ReadOnlyArray<ReleaseT>>,
+  +wikipediaExtract: WikipediaExtractT,
```
yvanzo: One last thing: there isn’t necessary a Wikipedia extract for each release group.
```suggestion
  +wikipediaExtract: WikipediaExtractT | null,
```

root/release_group/ReleaseGroupIndex.js
```diff
@@ -0,0 +1,196 @@
...
+    ) : null
+    }
+    <Annotation
+      annotation={releaseGroup.latest_annotation}
+      collapse
+      entity={releaseGroup}
+      numberOfRevisions={numberOfRevisions}
+    />
+    <WikipediaExtract
+      cachedWikipediaExtract={wikipediaExtract || null}
```
yvanzo: The above type change would also make `|| null` unnecessary here:
```suggestion
      cachedWikipediaExtract={wikipediaExtract}
```

root/static/scripts/common/components/Annotation.js
mwiencek: My intent was for the singular Prop to indicate it's for one prop, not a set. :) But `MinimalAnnotatedEntityT` sounds pretty good (with Annotated instead of Annotative)

root/static/scripts/common/components/Annotation.js
mwiencek: This is local to Annotation.js (doesn't need to appear in any index file). There is also a slight difference in the types in that `editor` is (intentionally) read-only here and not below, though you could make `MaybeSanitizedEditorT` as an alias for `EditorT | SanitizedEditorT | null` since it's also used in EditorLink.js.

yvanzo: Oh right, thought it was in some RG file.

root/static/scripts/common/components/EditorLink.js
```diff
@@ -23,7 +23,7 @@ const MissingEditorLink = () => {
 type Props = {|
   +avatarSize?: number,
   +content?: string,
-  +editor: EditorT | SanitizedEditorT | null,
+  +editor: ?(EditorT | SanitizedEditorT),
```
mwiencek: I think you accidentally reverted some of the type changes I synced here, this and the `AnnotationT` change in root/types.js

PR #1004, opened 2019-04-10
Fix messages changed since v-2019-01-22 and update localization files

yvanzo:
Updating localization files is usually done by pushing commits directly to `master` or `beta`, but this time I reviewed differences with v-2019-01-22 to `po/*.pot` files and fixed a lot of messages in `root/` before.

### DIFF
--- a/root/static/scripts/account/components/EditProfileForm.js
+++ b/root/static/scripts/account/components/EditProfileForm.js
@@ -23,6 +23,7 @@ import SelectField from '../../../../components/SelectField';
 import DBDefs from '../../common/DBDefs-client';
 import Autocomplete from '../../common/components/Autocomplete';
 import Warning from '../../common/components/Warning';
+import {N_lp_attributes} from '../../common/i18n/attributes';
 import {pushCompoundField} from '../../edit/utility/pushField';
 import hydrate from '../../../../utility/hydrate';
 
@@ -65,9 +66,9 @@ type State = {|
 const genderOptions = {
   grouped: false,
   options: [
-    {label: N_l('Male'), value: 1},
-    {label: N_l('Female'), value: 2},
-    {label: N_l('Other'), value: 3},
+    {label: N_lp_attributes('Male', 'gender'), value: 1},
+    {label: N_lp_attributes('Female', 'gender'), value: 2},
+    {label: N_lp_attributes('Other', 'gender'), value: 3},
   ],
 };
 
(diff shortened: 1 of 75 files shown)


yvanzo:
Just updated commit https://github.com/metabrainz/musicbrainz-server/pull/1004/commits/fc7bfce4d93d8a0daf23e7b85d9fa88a3297e87e

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -47,15 +47,19 @@ export const WorkListRow = ({
     ) : null}
     <td><EntityLink entity={work} /></td>
     <td>
-      <ArtistRoles relations={work.writers} />
+      {work.writers.length
+        ? <ArtistRoles relations={work.writers} />
+        : l('-')}
```
reosarevok: The only bit where I'm not quite sure. This hyphen was only here for reports, it seems, but it wasn't there for searches and other work lists. IMO just an empty cell is a better choice here (to represent "nothing entered yet") than hyphen. If we're missing a type, gender, or whatnot for an artist we don't use a hyphen, we just leave it blank (https://musicbrainz.org/search?query=a&type=artist&limit=25&method=indexed) - I don't see why this should be different.

yvanzo: @reosarevok:  I removed that commit from this pull request and made it available to you at https://github.com/yvanzo/musicbrainz-server/commit/ee3c531495daecb44f361049c79a2354d14fe0e3.

root/static/scripts/account/components/EditProfileForm.js
```diff
@@ -67,7 +67,7 @@ const genderOptions = {
   options: [
     {label: N_l('Male'), value: 1},
     {label: N_l('Female'), value: 2},
-    {label: N_l('Other'), value: 3},
+    {label: N_lp('Other', 'gender option'), value: 3},
```
mwiencek: I'd expect this should be `() => lp_attributes('Other', 'gender')` instead (and similarly for the two above).

yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1004/commits/eda986d60a635c0fc7d1b9deb06e03b1f6dcee21, re-updated pot files in https://github.com/metabrainz/musicbrainz-server/pull/1004/commits/a6a90c291a71b0ab275918f034638b88a0400c23 (no more gender in server resource).

PR #1009, opened 2019-04-10
MBS-9855: Generalize NotFound templates to one file

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9855


root/entity/NotFound.js
```diff
@@ -0,0 +1,163 @@
...
+      ) : (
+        expand2react(
+          notFoundStringPicker[namespace](),
+          namespace === 'genre' ? (
+            {
+              genre_list: '/genres',
+            }
+          ) : (
+            {
+              search_url: namespace === 'edit' ? '/search/edits' : '/search',
```
mwiencek: I like the idea of this PR, but might do things slightly differently. I think one of the goals should be making adding a new not-found page as easy as possible (as simple as adding a new configuration block above), not necessarily making the code as compact as possible (though reducing the amount of code is nice too).

Adding a new page is not so easy now though, since you have to add some configuration above to two objects, then possibly `namespacesWithoutSearch`, and also possibly tweak the mess of ternary conditions down here if you somehow create a new exception.

My idea would be to have a single object like:

```JS
const notFoundPages = {
  'namespace': {
    title: '',
    message: '',
    args: {},     // expand2react args
    footer: null, // for e.g. elections
  },
};
```
and then change `NotFound` to simply look up the namespace in this object and display the relevant properties. So to add a new page, you'd only ever need to update the `notFoundPages` object by adding a single entry.

To help with any repetition that causes, we could have constants for args like `{search_url: '/search'}` which are shared among the objects that need it.

reosarevok: That actually sounds much clearer than my version. Made the changes.

PR #1012, opened 2019-04-11
Support React templates in /ws/js/edit/preview

mwiencek:
For edit templates which have been converted to React, support these in /ws/js/edit/preview by checking `edit_template_react` and going through the Node renderer instead of TT in that case.

Fixes [MBS-10109](https://tickets.metabrainz.org/browse/MBS-10109)


mwiencek:
Eventually the Reactified edits should just be rendered directly on the client, but this is much easier for now.

PR #1014, opened 2019-04-12
MBS-10116: Convert the recording index page to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10116

Please do review the flow types specifically, since those still confuse me.


lib/MusicBrainz/Server/Entity/Track.pm
```diff
@@ -68,6 +68,7 @@ around TO_JSON => sub {
         %{ $self->$orig },
         isDataTrack     => boolean_to_json($self->is_data_track),
         length          => $self->length,
+        medium          => $self->medium->TO_JSON;
```
mwiencek: (Haven't tested the PR yet, but: syntax error here with `;` at the end.)

root/recording/RecordingIndex.js
```diff
@@ -0,0 +1,137 @@
...
+  +numberOfRevisions: number,
+  +pager: PagerT,
+  +recording: RecordingT,
+  +tracks: $ReadOnlyArray<$ReadOnlyArray<{...TrackT, +medium: MediumT}>>,
+|};
+
+const RecordingAppearancesTable = ({
+  tracks,
+}: {
+  tracks: $ReadOnlyArray<$ReadOnlyArray<{...TrackT, +medium: MediumT}>>,
```
mwiencek: You could use `$ElementType<Props, 'tracks'>` there to simplify things a bit.

mwiencek:
Can we also remove root/recording/index.tt now?

reosarevok:
Feel free to change loopParity to work that way, if you think it's better! :)

mwiencek:
Done so in https://github.com/metabrainz/musicbrainz-server/commit/e8cb66244bd3869f23e24bc0bdefcf798af27b74

PR #1015, opened 2019-04-15
MBS-9028: Skip redirects when loading Wikipedia extracts

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9028

Loading redirects as extracts is generally unwanted, since in most cases they end up returning the extract for a more general page that isn't really a 1:1 mapping to the entity on our side. This is especially unwanted if it means an English redirect is being preferred over an actual proper mapping in a different language.

For testing, https://www.wikidata.org/wiki/Q514294 is a good option (redirect en but valid it and de).

I didn't move all the requests to formatversion=2 because it seems to break some unicode formatting. For example, version 1 returns "(stylized as <b>NI\u0418</b>)" while version 2 returns "(stylized as <b>NI\x{d0}\x{98}</b>)" in Nine Inch Nails' extract (the latter doesn't display properly).


lib/MusicBrainz/Server/Data/WikipediaExtract.pm
```diff
@@ -124,6 +83,59 @@ sub get_available_languages
     }
     return (undef, undef);
 }
+sub _get_language_to_use {
```
mwiencek: Can we add a newline above this subroutine for consistency?

reosarevok:
I think the recursion is unlikely to be much of a problem tbh - for it to have more than a couple levels of recursion it would require multiple preferred languages to have a redirect, which seems uncommon enough. In most cases here there will be no recursion at all, since the first found result won't be a redirect.

mwiencek:
I had an idea of how to remove the recursion and make things slightly clearer, so I tried that and it seems to work (see the latest commit). It calculates the entire list of languages it can use *first* (ordered by preference), then returns the first one that doesn't redirect. Which probably does slightly more initial work, but much less if it needs to choose a fallback. And I optimized it slightly to use a `%languages_by_code` hash instead of looping over `@$languages` in each fallback scenario.

yvanzo:
Implementation in https://github.com/metabrainz/musicbrainz-server/pull/1015/commits/41c96c7717043398bf637081031b523767bd079c neither is recursive nor increases initial work, but it has a bit of code duplication. Array/hash usage can probably be optimized.

Also had to fix a weird issue with format version 2 not returning an array, see https://github.com/metabrainz/musicbrainz-server/pull/1015/commits/f8bee3180c47354a1de416591318d55372e1b0d9.

PR #1018, opened 2019-04-17
MBS-9923: Convert the url index page to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9923


reosarevok:
I think I already fixed this to master at some point but didn't rebase this in the meantime. Otherwise I'll fix it :)

PR #1020, opened 2019-04-20
MBS-10127: Convert relationships-table.tt to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10127

Also includes https://tickets.metabrainz.org/browse/MBS-10189


root/components/RelationshipsTable.js
```diff
@@ -0,0 +1,158 @@
...
+  +showCredits: boolean,
+|};
+
+const RelationshipsTable = ({
+  entity,
+  heading,
+  showCredits,
+}: Props) => {
+  const appearances = generateRelationshipAppearancesList(entity);
+  if (!appearances || appearances.length === 0) {
```
mwiencek: Since appearances is an object, `appearances.length` will be undefined (unless we add a relationship type named "length" :)).

`Object.keys(appearances).length` should work. Maybe do `const relationshipTypes = Object.keys(appearances);` first since you also need the keys below where it loops over them.

(This should fix a bug where an empty table is displayed.)

lib/MusicBrainz/Server/Entity/Role/Linkable.pm
```diff
@@ -74,6 +75,7 @@ sub relationships_by_link_type_names
     } $self->all_relationships ];
 }
 
+# Converted to JavaScript at root/utility/generateRelationshipAppearancesList.js
 sub appearances {
```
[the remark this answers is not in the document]
reosarevok: Seems sensible, fixed (I think!)

root/types.js
```diff
@@ -726,7 +727,9 @@ declare type RelationshipT = {|
   +id: number,
   +linkOrder: number,
   +linkTypeID: number,
+  +sourceCredit?: string,
   +target: CoreEntityT,
+  +targetCredit?: string,
```
mwiencek: You can remove these additions (including html on LinkAttrT above)

root/utility/generateRelationshipAppearancesList.js
```diff
@@ -0,0 +1,49 @@
...
+
+const pickAppearancesTypes = (entityType) => {
+  switch (entityType) {
+    case 'artist':
+    case 'label':
+    case 'place': {
+      return ['recording', 'release', 'release_group', 'work'];
+    }
+    case 'work': {
+      return ['recording'];
```
mwiencek: Minor nitpick: can we make these two arrays constants outside of `pickAppearancesTypes`, and return the constants here? That'll avoid at least one unnecessary allocation each time the template renders. :)

root/components/RelationshipsTable.js
```diff
@@ -0,0 +1,177 @@
...
+  const relationshipTypes = Object.keys(appearances);
+  if (!appearances || relationshipTypes.length === 0) {
+    return <p>{fallbackMessage}</p> || null;
+  }
+  let hasCreditColumn = 0;
+  let hasAttributeColumn = 0;
+  let hasArtistColumn = 0;
+  let hasLengthColumn = 0;
+  for (const relationshipType of relationshipTypes) {
+    for (const relationship of appearances[relationshipType]) {
```
mwiencek: Since `appearances` is just `entity.relationships` grouped by type, you should be able to remove the outer loop and just use `entity.relationships` in place of `appearances[relationshipType]` here.

root/components/RelationshipsTable.js
```diff
@@ -0,0 +1,177 @@
...
+            <th>{l('Date')}</th>
+            <th>{l('Title')}</th>
+            {hasCreditColumn ? <th>{l('Credited As')}</th> : null}
+            {hasAttributeColumn ? <th>{l('Attributes')}</th> : null}
+            {hasArtistColumn ? <th>{l('Artist')}</th> : null}
+            {hasLengthColumn ? <th>{l('Length')}</th> : null}
+          </tr>
+        </thead>
+        <tbody>
+          {Object.keys(appearances).sort().map(relationshipType => (
```
mwiencek:
```suggestion
          {relationshipTypes.sort().map(relationshipType => (
```

root/components/RelationshipsTable.js
```diff
@@ -0,0 +1,177 @@
...
+                        {sourceCredit || null}
+                      </td>
+                    ) : null}
+                    {hasAttributeColumn ? (
+                      <td>
+                        {attributes && attributes.length > 0 ? (
+                          attributes.forEach(attribute => {
+                            attributeList.push(displayLinkAttribute(attribute));
+                          }),
+                          commaList(attributeList)
```
mwiencek: You can replace the `forEach` with `map` here, and pass it directly to `commaList`. Then `attributeList` can be removed, too

root/components/RelationshipsTable.js
mwiencek: Ignore, I forgot most of the pages other than for place load other "non-appearance" relationships for display.

PR #1021, opened 2019-04-21
MBS-10131: Change LyricWiki domain

reosarevok:
[MBS-10131](https://tickets.metabrainz.org/browse/MBS-10131): LyricWiki moved from wikia to fandom (like most wikia wikis).

After conversation, edited to include [MBS-10135](https://tickets.metabrainz.org/browse/MBS-10135): Support WikiaParoles as its own site rather than LyricWiki


lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -99,7 +99,7 @@ my %URL_SPECIALIZATIONS = (
     'Loudr'               => qr{^https?://(?:www\.)?loudr\.fm/}i,
     'LyricEvesta'         => qr{^https?://lyric\.evesta\.jp/}i,
     'LyricsNMusic'        => qr{^https?://lyricsnmusic\.com/}i,
-    'LyricWiki'           => qr{^https?://lyrics\.wikia\.com/}i,
+    'LyricWiki'           => qr{^https?://lyrics\.fandom\.com/}i,
```
reosarevok: Should this regex also match wikia links since we haven't changed them yet?

yvanzo: Yes! It would be nice also to replace `wikia` with `fandom` in URL display by overriding `href_url` in [LyricWiki.pm](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-01-22/lib/MusicBrainz/Server/Entity/URL/LyricWiki.pm), see [VIAF.pm](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-01-22/lib/MusicBrainz/Server/Entity/URL/VIAF.pm) for example.

reosarevok: Made the updates, tested locally and they seem to work, but let me know :)

reosarevok:
Given each language subdomain seems to be a different wiki, and that we're only matching the main domain for display on the sidebar, I'm not even sure it's *intended* to allow the subdomains.

What I have now clearly needs changing, but should we change it towards allowing all subdomains (and display them all as "LyricsWiki", even if that's not 100% true), allowing specific subdomains (and display them under their actual name, but block the others) or allowing only the main domain?

Looking at the actual usage with subdomains, we seem to have literally 1515 links to fr. (and one link to http://vocaloidlyrics.wikia.com/wiki/Category:Vagenda_songs_list which seems unrelated but might be the last result in @yvanzo's 1516). So I'd suggest allowing fr. with its own sidebar name, URL file etc, and blocking other subdomains for the time being.

lib/MusicBrainz/Server/Entity/URL/WikiaParoles.pm
```diff
@@ -0,0 +1,30 @@
...
+
+sub url_is_scheme_independent { 1 }
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2010 MetaBrainz Foundation
```
yvanzo: So 2010 :wink:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1435,6 +1435,8 @@ const CLEANUPS = {
     type: LINK_TYPES.lyrics,
     clean: function (url) {
       url = url.replace(/^https?:\/\/([^/]+\.)?genius\.com/, 'http://$1genius.com');
+      url = url.replace(/^https?:\/\/lyrics.(?:wikia|fandom)\.com/, 'https://lyrics.fandom.com');
```
yvanzo: Escaping dot:
```suggestion
      url = url.replace(/^https?:\/\/lyrics\.(?:wikia|fandom)\.com/, 'https://lyrics.fandom.com');
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1435,6 +1435,8 @@ const CLEANUPS = {
...
+      url = url.replace(/^https?:\/\/fr.lyrics.(?:wikia|fandom)\.com/, 'https://lyrics.fandom.com/fr');
```
yvanzo: Escaping dots:
```suggestion
      url = url.replace(/^https?:\/\/fr\.lyrics\.(?:wikia|fandom)\.com/, 'https://lyrics.fandom.com/fr');
```

lib/MusicBrainz/Server/Entity/URL/LyricWiki.pm
```diff
@@ -5,6 +5,13 @@ use Moose;
...
+override href_url => sub {
+    # Turn the official permalink into what LyricWiki currently redirects to.
+    # See https://community.fandom.com/wiki/Help:Fandom_domain_migration for more info
+    shift->url->as_string =~
+        s{^https?://lyrics\.wikia\.com/}{https://lyrics.fandom.com/}r;
+};
+
 sub sidebar_name { 'LyricWiki' }
 
 sub url_is_scheme_independent { 1 }
```
yvanzo: That should be removed as `href_url` is overridden for LyricWiki too, see https://github.com/metabrainz/musicbrainz-server/pull/1021#discussion_r278022541

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1058,6 +1058,15 @@ const CLEANUPS = {
       return true;
     },
   },
+  'fandomlyrics': {
+    match: [new RegExp('^(https?://)?(fr.)?lyrics\\.(wikia|fandom)\\.com', 'i')],
```
yvanzo: Escaping dot
```suggestion
    match: [new RegExp('^(https?://)?(fr\\.)?lyrics\\.(wikia|fandom)\\.com', 'i')],
```

PR #1023, opened 2019-04-22
Fix relationship display issues (inc. MBS-10111)

mwiencek:
This PR fixes [MBS-10111](https://tickets.metabrainz.org/browse/MBS-10111) and various other display issues I ran into while trying to fix that.


mwiencek:
Fixed the typo @yvanzo found and another I found in the added code comment.

PR #1024, opened 2019-04-22
MBS-7480: Add filesize columns for cover art

mwiencek:
This allows for storing file sizes for cover images and their thumbnails. This'll allow us to better detect when a thumbnail isn’t available, and also allow us to display file sizes to the user before they download an image.

MBS hasn't been modified to actually use these fields here (they'll have to be populated first to be useful).


mwiencek:
> 2GB ought to be enough for anybody. 😉

LOL, let's hope so.

> Don’t forget to regenerate schema 25 scripts before merging! 😛

Thanks for the reminder. :)

PR #1025, opened 2019-04-23
Fix for updating track ACs with the release AC

mwiencek:
This is a hack, but fixes an issue where the track artist credits would not appear to be updated when the matching release artist credit was changed. It achieves this by setting the component state directly, similar to how we do inside `doneCallback` in bindingHandlers.js.

I don't know if it's related to [MBS-10115](https://tickets.metabrainz.org/browse/MBS-10115) without specific steps to reproduce that.

Also, I saw some talk on IRC about the "Change all artists on this release that match ..." feature not working, but I can't reproduce any issue with that (and don't see any tickets related to it). It was recently fixed with https://github.com/metabrainz/musicbrainz-server/pull/795 so if anyone has more details about how to break it (again), please open a ticket with steps! :)

### DIFF
--- a/root/static/scripts/release-editor/init.js
+++ b/root/static/scripts/release-editor/init.js
@@ -157,6 +157,9 @@ releaseEditor.init = function (options) {
                     _.each(medium.tracks(), function (track) {
                         if (reduceArtistCredit(track.artistCredit()) === reduceArtistCredit(savedReleaseAC)) {
                             track.artistCredit(artistCreditFromArray(releaseAC));
+                            track.artistCreditEditorInst.setState({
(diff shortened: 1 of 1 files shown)


reosarevok:
Why is it a hack, and is there a non-hacky way of doing this? (we can still merge this, just wondering)

mwiencek:
> Why is it a hack, and is there a non-hacky way of doing this? (we can still merge this, just wondering)

 Well, you're not really meant to save references to React component instances and then call `setState` on them from outside the component, though it technically works. It's just not a sensical way to use React. :)

The non-hacky way would be for the release editor to be converted to React, heh.

PR #1026, opened 2019-04-23
MBS-10133: Clarify "empty query" bad request error

reosarevok:
Fix [MBS-10133](https://tickets.metabrainz.org/browse/MBS-10133)
----

Collections aren't searched through the search server at all, if https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2/Search is correct. As such, this should just say that the blank query was blank.

Also generalize what is seen as a blank query: given we can't have empty string or whitespaces as names, `""`, `''` or `"    "` should also be rejected as blank queries. *Escaped* quote marks, such as `query=\"`, should be allowed in order to allow searching for fields containing quote marks.

### DIFF
--- a/lib/MusicBrainz/Server/Data/WebService.pm
+++ b/lib/MusicBrainz/Server/Data/WebService.pm
@@ -205,7 +205,7 @@ sub xml_search
     if ($query =~ /^\s*$/)
     {
         return {
-            error => "Must specify a least one parameter (other than 'limit', 'offset' or empty 'query') for collections query.",
+            error => "You submitted a blank search query. You must include a non-blank 'query=' parameter with your search.",
             code  => HTTP_BAD_REQUEST
         };
     }
--- a/lib/MusicBrainz/Server/Data/WebService.pm
+++ b/lib/MusicBrainz/Server/Data/WebService.pm
@@ -201,8 +201,8 @@ sub xml_search
     }
 
     $query =~ s/^ AND //;
-    # In case we have a blank query
-    if ($query =~ /^\s*$/)
+    # In case we have a blank query (only whitespace, or whitespace surrounded by unescaped quotes)
+    if ($query =~ /^(?:\s*|"\s*")$/)
     {
         return {
             error => "You submitted a blank search query. You must include a non-blank 'query=' parameter with your search.",
--- a/lib/MusicBrainz/Server/Test/WS.pm
+++ b/lib/MusicBrainz/Server/Test/WS.pm
@@ -31,7 +31,7 @@ Readonly our $FORBIDDEN_XML_RESPONSE => <<'EOXML';
 <?xml version="1.0" encoding="UTF-8"?>
 <error>
   <text>You are not authorized to access this resource.</text>
-  <text>For usage, please see: http://musicbrainz.org/development/mmd</text>
+  <text>For usage, please see: https://musicbrainz.org/development/mmd</text>
 </error>
 EOXML
 
@@ -39,7 +39,7 @@ Readonly our $UNAUTHORIZED_XML_RESPONSE => <<'EOXML';
 <?xml version="1.0" encoding="UTF-8"?>
 <error>
   <text>Your credentials could not be verified. Either you supplied the wrong credentials (e.g., bad password), or your client doesn't understand how to supply the credentials required.</text>
-  <text>For usage, please see: http://musicbrainz.org/development/mmd</text>
+  <text>For usage, please see: https://musicbrainz.org/development/mmd</text>
 </error>
 EOXML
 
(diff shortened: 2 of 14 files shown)


lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -201,11 +201,11 @@ sub xml_search
     }
 
     $query =~ s/^ AND //;
-    # In case we have a blank query
-    if ($query =~ /^\s*$/)
+    # In case we have a blank query (only whitespace, or whitespace surrounded by unescaped quotes)
+    if ($query =~ /^['"]?\s*['"]?$/)
```
mwiencek: Lucene doesn't support single quotes for quoting things, so I'd remove those. The double quotes also have to be balanced, so I'd do something like this instead (untested):
```suggestion
    if ($query =~ /^(?:\s*|"\s*")$/)
```

lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -201,11 +201,11 @@ sub xml_search
...
     {
         return {
-            error => "Must specify a least one parameter (other than 'limit', 'offset' or empty 'query') for collections query.",
+            error => "You submitted a blank query. You must include a non-blank 'query' parameter with your query.",
```
mwiencek: Nit: the word query at the end sounds a bit funny/repetitive with what comes before. Maybe "with your search" instead?

lib/MusicBrainz/Server/Data/WebService.pm
```diff
@@ -205,7 +205,7 @@ sub xml_search
     if ($query =~ /^\s*$/)
     {
         return {
-            error => "Must specify a least one parameter (other than 'limit', 'offset' or empty 'query') for collections query.",
+            error => "You submitted a blank query. You must include a non-blank 'query' parameter with your search.",
```
yvanzo: It would be nice to additionally link to https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2/Search

reosarevok: I guess since this is a WS error, "link to" literally means including something like "See https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2/Search for more info." at the end of the error line?

yvanzo: It would probably be more convenient if the URL ended the line as 
in [XML WS error message](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-06-03/lib/MusicBrainz/Server/WebService/XMLSerializer.pm#L1329): `For more info, please see: https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2/Search`

I just noticed that JSON WS documentation doesn’t even mention search.

yvanzo: And now I realize that the assumption “You submitted a blank query.” might be wrong because `fmt=json` may have been submitted as query string.

mwiencek: Not sure I understand, can you given an example of `fmt=json` being submitted as a query string?

yvanzo: Here is an example: https://musicbrainz.org/ws/2/artist?fmt=json

mwiencek: Terminology conflict! I thought you meant the literal `query` parameter.

But that's what the message is supposed to mean by "blank query," that no search `query` was given, not that no query parameters were given.

yvanzo: @mwiencek: So the message is still unclear. Another issue is it wrongly assumes the request is a search request missing its `query` query parameter (because other valid cases have been checked beforehand) whereas it could also be a browse request missing its MBID subpath or a lookup request missing its entity query parameter. Would a more general error message do?
`Your request is missing either an MBID or a query parameter.`

@reosarevok: I pushed a commit to add doc link to every JSON WS error, not just this one, on the model of XML WS error (which is updated to https in another commit). Can you please review it?

reosarevok: @yvanzo Your commits seem reasonable and work locally :) 

It assumes it's a search request because this is xml_search. If it's not supposed to be a search, then it shouldn't hit xml_search, should it? :)

mwiencek: That's https://github.com/metabrainz/musicbrainz-server/pull/1027 - once that PR is updated, browse requests won't end up here (that and anything else, including broken lookups, should return `The given parameters do not match any available query type for the artist resource.`). So this'll be specific to searches.

Not sure how to make the message more clear. `You submitted a blank search query. You must include a non-blank 'query=' parameter with your search.`?

yvanzo: Alright, let’s leave that other issue to the other PR. Your latest suggestion is fine by me.

reosarevok: Updated wording to @mwiencek's suggestion, which is nice :)

PR #1027, opened 2019-04-23
MBS-10134: Don't send bad browse queries to the search server

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10134

Any browse request with an unsupported attribute will be understood as a search request. Since there's no good way of distinguishing those from an actual search request with "query=" (an empty query attribute), we need an error message that fits both options. 

It isn't useful to send either of these to the search service anyway, so we can just give a bad request error before we send them.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/Area.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/Area.pm
@@ -26,6 +26,7 @@ my $ws_defs = Data::OptList::mkopt([
                          optional => [ qw(fmt limit offset) ],
      },
      area => {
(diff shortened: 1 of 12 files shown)


reosarevok:
So specify an action for every WS/2/Whatever method section? :) (seems sensible to me, just making sure before I do it!)

mwiencek:
I think you'd only have to do it for the lookup ones, because those are the only defs with no `required` or `linked` properties, AFAIK (meaning they'll otherwise match any GET request to any action).

reosarevok:
I *think* I've now updated this as requested :)

PR #1028, opened 2019-04-23
[WS/3 - non-mergeable] mbid -> MBID in text strings

reosarevok:
Back with another very important, hard-hitting PR, but this was annoying me. It's not mbid, it's MBID, dammit.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/Area.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/Area.pm
@@ -74,7 +74,7 @@ sub area_browse : Private
 
     if (!is_guid($id))
     {
-        $c->stash->{error} = "Invalid mbid.";
+        $c->stash->{error} = "Invalid MBID.";
         $c->detach('bad_req');
     }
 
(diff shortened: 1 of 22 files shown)


reosarevok:
As far as I can tell, these are all error messages and tests and it won't affect any translations.

mwiencek:
We probably shouldn't change these until WS3, since people can be using the error messages to check what kind of error occurred. :/ (To be fair, we don't provide any other sort of code they can check.)

Just looking at results on github, I found these pieces of code where people are doing an exact string check on "Invalid mbid.":
1. https://github.com/bexi/ArtistAlbumAPI/blob/54231b73824a3a7332fecec91cc71dce45305fb5/musicbrainzConnection.js#L26
2. https://github.com/gjavolce/viap_demo_api/blob/71816510639d4f367a7d2e041a5ed7f40da8f16c/src/main/resources/application.properties#L29
3. https://github.com/thomas-p-wilson/aula-radio/blob/b7f6c26ce400ed4bbfcd2f7edab66068bc0a55fc/test/seed.js#L122
4. https://github.com/alastair/python-musicbrainzngs/blob/9573a4e3b5f14ea60112f28364ea593b176d7e5d/test/test_mbxml.py#L26

PR #1030, opened 2019-04-23
MBS-1658 (SQL): Add free comment field to collection entries

reosarevok:
SQL for https://tickets.metabrainz.org/browse/MBS-1658

### DIFF
--- a/admin/sql/CreateTables.sql
+++ b/admin/sql/CreateTables.sql
@@ -2266,62 +2266,74 @@ CREATE TABLE editor_collection_collaborator (
 
 CREATE TABLE editor_collection_area (
     collection INTEGER NOT NULL, -- PK, references editor_collection.id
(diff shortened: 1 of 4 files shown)


reosarevok:
Either generate them once, or generate them each time, but actually generate them for the existing branch *first* so that only related stuff is there. IMO it's clearer if you can see the actual changes also on those files, but if you guys prefer just generating them once at the end, that works for me too.

mwiencek:
TBH I don't mind generating them with the PRs because it lets me see right away that everything is ending up in the right place. There was also a case I recall where we forgot to generate them afterward, and it wasn't noticed that they were incomplete until well into the testing phase. :P So that's in the back of my mind and probably makes me eager to always see them generated correctly.

But I agree we should generate them for the existing changes first so they aren't mixed up in this PR (and definitely again after merging into master just to be sure).

PR #1032, opened 2019-04-23
MBS-10052: Add event art archive schema

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10052

### DIFF
--- a/admin/InitDb.pl
+++ b/admin/InitDb.pl
@@ -277,11 +277,11 @@ sub CreateRelations
     RunSQLScript($SYSMB, "Extensions.sql", "Installing extensions");
 
     RunSQLScript($DB, "CreateTables.sql", "Creating tables ...");
-    RunSQLScript($DB, "caa/CreateTables.sql", "Creating tables ...");
-    RunSQLScript($DB, "eaa/CreateTables.sql", "Creating tables ...");
+    RunSQLScript($DB, "caa/CreateTables.sql", "Creating CAA tables ...");
+    RunSQLScript($DB, "eaa/CreateTables.sql", "Creating EAA tables ...");
     RunSQLScript($DB, "documentation/CreateTables.sql", "Creating documentation tables ...");
     RunSQLScript($DB, "json_dump/CreateTables.sql", "Creating json_dump tables ...");
-    RunSQLScript($DB, "report/CreateTables.sql", "Creating tables ...");
+    RunSQLScript($DB, "report/CreateTables.sql", "Creating report tables ...");
     RunSQLScript($DB, "sitemaps/CreateTables.sql", "Creating sitemaps tables ...");
     RunSQLScript($DB, "statistics/CreateTables.sql", "Creating statistics tables ...");
     RunSQLScript($DB, "wikidocs/CreateTables.sql", "Creating wikidocs tables ...");
--- a/admin/InitDb.pl
+++ b/admin/InitDb.pl
@@ -265,6 +265,7 @@ sub CreateRelations
             musicbrainz
             cover_art_archive
             documentation
+            event_art_archive
             json_dump
             report
             sitemaps
@@ -277,6 +278,7 @@ sub CreateRelations
 
     RunSQLScript($DB, "CreateTables.sql", "Creating tables ...");
     RunSQLScript($DB, "caa/CreateTables.sql", "Creating tables ...");
+    RunSQLScript($DB, "eaa/CreateTables.sql", "Creating tables ...");
     RunSQLScript($DB, "documentation/CreateTables.sql", "Creating documentation tables ...");
     RunSQLScript($DB, "json_dump/CreateTables.sql", "Creating json_dump tables ...");
     RunSQLScript($DB, "report/CreateTables.sql", "Creating tables ...");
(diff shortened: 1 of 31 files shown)


yvanzo:
Also, the project name is _Event **Art** Archive_ (EAA) - as it is expected to hold posters, flyers and more - rather than _Event **Poster** Archive_ (would be EPA) this PR and the related tickets are named after.

mwiencek:
> LGTM but admin/sql/update/* scripts should probably include ticket reference to follow naming convention from the three latest years, that is, be named after mbs-10052-eaa rather than just eaa.

Fixed that and the Environmental Protection Agency mixup.

> Depending on cover_art_archive for image_type is not ideal, but it has the merit to avoid duplication and to avoid making breaking change for now. Could it be moved to a shared schema in a further SC?

I was thinking something similar, and it could be, but might be the only item in the schema (unless we just add it to the musicbrainz one).

> Also just wondering, is there any plan to add image/thumbs filesize to CAA?

That's one of the tickets in the schema change! :) There's an open PR for it at #1024.

> DROP TRIGGER statements are missing for EAA Queue triggers eaa_reindex eaa_move and delete_event, see admin/sql/caa/DropPGQ.sql.

Oh, that file is unused and broken (we haven't used PGQ in ages). I could fix (and rename) it to DropMQTriggers and add a corresponding file for the EAA as a development tool, if desired, though I'll have to check how to port the pgq API calls to amqp.

admin/sql/eaa/CreateMQTriggers.sql
```diff
@@ -0,0 +1,173 @@
...
+
+        RETURN NULL;
+    END;
+$$ LANGUAGE 'plpgsql';
+
+CREATE TRIGGER eaa_reindex AFTER UPDATE
+ON musicbrainz.artist FOR EACH ROW
+EXECUTE PROCEDURE reindex_artist();
+
+CREATE OR REPLACE FUNCTION reindex_l_artist_event() RETURNS trigger AS $$
```
mwiencek: Forgot to explain these. In the IA slack, for the *_mb_metadata.xml files we settled on sending the place XML with artist-rels and place-rels included (since at a minimum it'd probably be useful for the IA to show "who" and "where" on their details pages). So this will have it reindex whenever those relationships change. It doesn't reindex if, say, an attribute on one of the relationships changes. We *could* add that in the future, if they actually use attributes somehow, but it seems overkill for now.

yvanzo: Did you consider joining event-rels (9K) and series-rels (18K) which are mostly about “parts”?

Also, it would be nice to allow event art for event series (tour, festival, run).

mwiencek: > Did you consider joining event-rels (9K) and series-rels (18K) which are mostly about “parts”?

Not really. But I think we should wait and see what the IA does with their details pages, 'cause we don't want to trigger on changes for data they don't use. We could add new triggers/functions (or even modify the existing ones) after the schema change, since these have to be created manually anyway, and aren't part of the core schema - nobody will have these in their DB unless they're hacking on CAA stuff.

> Also, it would be nice to allow event art for event series (tour, festival, run).

Yeah, probably for a future schema change. :P Or we can have an event art type for it and put it on the first event in the series for now.

admin/InitDb.pl
```diff
@@ -277,6 +278,7 @@ sub CreateRelations
 
     RunSQLScript($DB, "CreateTables.sql", "Creating tables ...");
     RunSQLScript($DB, "caa/CreateTables.sql", "Creating tables ...");
+    RunSQLScript($DB, "eaa/CreateTables.sql", "Creating tables ...");
```
reosarevok: Wouldn't it make sense for this (and caa and report) to specify which tables, since the others do?

admin/sql/eaa/CreateFunctions.sql
```diff
@@ -0,0 +1,72 @@
...
+            CASE (
+                SELECT event_art_presence FROM musicbrainz.event_meta
+                WHERE id = NEW.event
+            )
+                WHEN 'absent' THEN
+                    UPDATE musicbrainz.event_meta
+                    SET event_art_presence = 'present'
+                    WHERE id = NEW.event;
+                WHEN 'darkened' THEN
+                    RAISE EXCEPTION 'This event has been darkened and cannot have new event art';
```
reosarevok: This is less obviously true for events than releases (if a poster has been taken down, that doesn't necessarily mean a ticket wouldn't be ok to upload) but I guess darkening is rare enough it shouldn't matter anyway.

admin/sql/eaa/CreateMQTriggers.sql
reosarevok: > Also, it would be nice to allow event art for event series (tour, festival, run).

Keep in mind that for festivals at least, each year's festival is its own main event (so a festival poster can be uploaded there). Agreed that a tour poster would be nice to have at the series level, but it doesn't seem super important just yet :)

yvanzo: By the way, this feature is also missing for CAA and release (group) series.

reosarevok: What would be the use of that? The series logo or something? That's already more like https://tickets.metabrainz.org/browse/MBS-4683

admin/InitDb.pl
mwiencek: Added in ac64e7bdf65c6de41bf24d36daae8a27570dd75f

admin/sql/eaa/CreateFunctions.sql
mwiencek: That's true...though from what I understand, if it's darkened on the IA side, uploading any artwork to there wouldn't work, so the exception would also be technically true.

admin/sql/updates/schema-change/25.slave.sql
```diff
@@ -9,4 +10,147 @@ SELECT '20180503-mbs-9708-drop-area-containment-view.sql';
...
+    id                  SERIAL NOT NULL, -- PK
+    name                TEXT NOT NULL,
+    parent              INTEGER, -- references event_art_archive.art_type.id
+    child_order         INTEGER NOT NULL DEFAULT 0,
+    description         TEXT,
+    gid                 uuid NOT NULL
+);
+
+INSERT INTO art_type VALUES
+    (1, 'Poster', NULL, 0, NULL, generate_uuid_v3('6ba7b8119dad11d180b400c04fd430c8', 'even_art_archive.art_type' || 1));
```
mwiencek: Let me know if there's other types we want to add. I added poster at least because the index listing query relies on `type_id = 1` for the `is_front` calculation.

yvanzo: Most probably “Flyer” (or “Leaflet”) and “Ticket”, but it seems reasonable to ask the community about their needs.

reosarevok: Agreed with just letting the community decide on others. Do we need to tell the IA specifically if we add more?

mwiencek: I don't think so. They were mainly concerned about how to select a representative image, since they use a type=Front check for the CAA.

mwiencek:
@yvanzo did I address all your feedback?

mwiencek:
Fixed the test failures, and added the relevant fixes to https://github.com/metabrainz/musicbrainz-server/pull/983. Please check that, so we can merge those fixes into master (and then master into the schema change branch) before merging this. :)

PR #1034, opened 2019-04-24
MBS-8915: Allow editors to choose delimiter in track parser

reosarevok:
https://tickets.metabrainz.org/browse/MBS-8915

Had to resubmit this since I couldn't push to https://github.com/metabrainz/musicbrainz-server/pull/774 - so it's still mostly @Ge0rg3's work.

Made a few changes as requested on the comments on the previous PR, and a change to move both components of the delimiter to the left column in a way that seems to make more sense to me.

Commit text: 
"This commit allows users to choose a custom delimiter, which can be any regular expression. This also required a change to the CSS to fit on, and a change to the optionCookie function to allow for strings as well as just boolean values."

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -49,8 +49,13 @@
       </td>
     </tr>
     <tr>
-      <td></td>
       <td>
+        <label>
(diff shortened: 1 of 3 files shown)


reosarevok:
Merged then. Thanks @Ge0rg3! :)

PR #1035, opened 2019-04-25
MBS-5387: Show edits pending on artist credits

mwiencek:
One thing I need help with: if an AC has only one artist, how do we *visually* distinguish between the AC having pending edits and the artist having pending edits? Note, an "Edit Artist Credit" edit does increment the pending edits counter of each artist in the AC. So the only real way to tell if the AC has pending edits is seeing if a join phrase is highlighted, and those aren't always present.


reosarevok:
We could have a separate icon instead of going all orange on it. Alternatively, on the Aliases tab at least, we could highlight the Edit link orange.

root/components/Aliases/ArtistCreditList.js
```diff
@@ -17,7 +17,7 @@ import loopParity from '../../utility/loopParity';
 
 type Props = {
   +$c: CatalystContextT,
-  +artistCredits: $ReadOnlyArray<ArtistCreditT>,
+  +artistCredits: $ReadOnlyArray<{+id: number} & ArtistCreditT>,
```
reosarevok: Wouldn't it make sense for this to be its own type?

reosarevok: Or is this just saying "in this case, id is not optional"?

reosarevok:
This seems generally fine - I assume you've tested the AC changes locally a bit to make sure everything still seems to work? Not approving yet because we might want to talk more about how to indicate it in the UI, but further changes to that can also be made as separate PRs later on.

root/components/Aliases/ArtistCreditList.js
mwiencek: Yeah, it's to say the `id` is always present here (otherwise we run into errors below that `id` can be `undefined`, which is obviously not true on this page).

Right now the `ArtistCreditT` definition is a bit weird since `id`, `entityType`, and `editsPending` are optional, even though they're not for ACs from the server. So we should probably have a separate `WritableArtistCreditT` (not sure of a good name) for ones which just have `names` (for use on edit pages).

mwiencek:
> This seems generally fine - I assume you've tested the AC changes locally a bit to make sure everything still seems to work? Not approving yet because we might want to talk more about how to indicate it in the UI, but further changes to that can also be made as separate PRs later on.

I tested as much as I could on edit pages, but will definitely spend more time testing things.

Both the icon idea and highlighting the Edit link sound good to me (maybe we should just do both?). Not sure what icon to use - of the ones we have now, warning.png shrunk down might work to indicate an important notice, or information.png for something less scary looking. The icon could also open a tooltip with a link to an edit search for the open "Edit Artist Credit" edits, if desired.

mwiencek:
Updated with some potential UI changes. Please check and let me know if it's tolerable. :)

Freso:
I don’t much like the idea of introducing "warning" icons there (if so, we should also "warn" in general when edits are pending, which would in turn defeat the distinction desired here), but a "hover" and/or highlighting the Edit link seem fine to me.

mwiencek:
Thanks for the feedback! We could remove the icon and just show the informational tooltip if they hover anywhere over the AC. That might conflict slightly with the native tooltips we have on artist links, but maybe it'll be fine...

mwiencek:
So, I tried that and having both tooltips display at once was pretty weird. I think having no visible indication and requiring the user to hover over the AC still isn't ideal because the point is to display some kind of indication. I changed it to use https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/images/icons/information.png instead, but we can come up with a better icon or indicator in the future.

mwiencek:
As in #1032, the last seven commits here fix the test failures and have been added to https://github.com/metabrainz/musicbrainz-server/pull/983 (so let's merge that first).

root/components/Aliases/ArtistCreditList.js
yvanzo: More of not sure good names: `ArtistCreditStubT`, `CoreArtistCreditT`.

mwiencek: `ArtistCreditStubT` sounds pretty good. I was also thinking about `EditableArtistCreditT` since it's, well, intended to be editable and used on edit pages.

I wasn't planning on making these changes in this PR though.

PR #1037, opened 2019-04-26
MBS-10139: Indicate why recording length can't be edited

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10139

### DIFF
--- a/root/recording/edit_form.tt
+++ b/root/recording/edit_form.tt
@@ -24,9 +24,16 @@
       [%- form_row_name_with_guesscase(r, { guessfeat => 1 }) -%]
       <div id="artist-credit-editor"></div>
       [%- form_row_text_long(r, 'comment', add_colon(l('Disambiguation'))) -%]
-      [%- IF !form.used_by_tracks || form.field('length').has_errors;
-            form_row_text_long(r, 'length', l('Length:'));
-          END -%]
+      [%- IF !form.used_by_tracks || form.field('length').has_errors -%]
+        [%- form_row_text_long(r, 'length', l('Length:')) -%]
+      [%- ELSE -%]
+        [% WRAPPER form_row %]
+          <label>[% l('Length:') %]</label>
+          [% l('{recording_length} ({length_info|derived} from the associated track lengths)',
+                recording_length => format_length(form.field('length').value),
+                length_info => doc_link('Recording') ) %]
+        [%- END -%]
+      [%- END -%]
       [%- form_row_checkbox(r, 'video', l('Video')) -%]
       [%- form_row_text_list(r, 'isrcs', l('ISRCs:'), l('ISRC')) -%]
     </fieldset>


root/recording/edit_form.tt
```diff
@@ -24,9 +24,15 @@
...
       [%- form_row_text_long(r, 'comment', add_colon(l('Disambiguation'))) -%]
-      [%- IF !form.used_by_tracks || form.field('length').has_errors;
-            form_row_text_long(r, 'length', l('Length:'));
-          END -%]
+      [%- IF !form.used_by_tracks || form.field('length').has_errors -%]
+        [%- form_row_text_long(r, 'length', l('Length:')) -%]
+      [%- ELSE -%]
+        [% WRAPPER form_row %]
+          <label>[% l('Length:') %]</label>
+          [% l('{length_info|Automatically derived} from the associated track lengths.',
```
mwiencek: The only thing weird to me is that it doesn't show the length that it derived anywhere, even though we know what it is and now have this "Length:" row. So I think we should display it, possibly with this message in parentheses next to it (though that causes things to wrap on my screen, so I had to remove "Automatically" to fit it). You might have a better idea.

```suggestion
          [% l('{recording_length} ({length_info|derived} from the associated track lengths)',
                recording_length => format_length(form.field('length').value),
                length_info => doc_link('Recording') ) %]
```

reosarevok: Sounds like a good idea, done.

root/recording/edit_form.tt
```diff
@@ -24,9 +24,16 @@
...
-            form_row_text_long(r, 'length', l('Length:'));
-          END -%]
+      [%- IF !form.used_by_tracks || form.field('length').has_errors -%]
+        [%- form_row_text_long(r, 'length', l('Length:')) -%]
+      [%- ELSE -%]
+        [% WRAPPER form_row %]
+          <label>[% l('Length:') %]</label>
+          [% l('{recording_length} ({length_info|derived} from the associated track lengths)',
+                recording_length => format_length(form.field('length').value),
+                length_info => doc_link('Recording') ) %]
```
yvanzo: Link can probably jumps to the corresponding section:
```suggestion
                length_info => doc_link('Recording#Length') ) %]
```

reosarevok: I'm going to merge it as-is for now, because this doesn't actually jump to the section: it loads Recording%23Length which doesn't work.

yvanzo: That’s weird because it works if you credit recording to various artists:

* https://github.com/metabrainz/musicbrainz-server/pull/1037/files/e4688a174cbe63f1d4deac5fff203669a935c2e2#diff-9fe5af5c827d1214721636da8d36ededR7
* https://github.com/metabrainz/musicbrainz-server/blob/v-2019-06-03/root/components/common-macros.tt#L27-L29

Creating a specific macro for this link might work?

reosarevok: That link doesn't work for me either. Sure, it does show "Special purpose artist#List of official SPAs" - but it still goes to the top of the page.

yvanzo: Indeed, it actually points to [`Special_purpose_artist%23List_of_official_SPAs`](https://musicbrainz.org/doc/Style/Unknown_and_untitled/Special_purpose_artist%23List_of_official_SPAs).

Also just found that the transcluded wiki page [`Special_purpose_artist%23List_of_official_SPAs`](https://wiki.musicbrainz.org/Style/Unknown_and_untitled/Special_purpose_artist%23List_of_official_SPAs) actually redirects to [`Special_purpose_artist#List_of_official_SPAs`](https://wiki.musicbrainz.org/Style/Unknown_and_untitled/Special_purpose_artist%23List_of_official_SPAs).

PR #1038, opened 2019-04-28
MBS-10143: Allow querying for tags/genres on discid ws requests

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10143

# Action

- [ ] Update https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2#discid

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/DiscID.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/DiscID.pm
@@ -13,7 +13,8 @@ use constant DURATION_LOOKUP_RANGE => 10000;
 my $ws_defs = Data::OptList::mkopt([
      discid => {
                          method   => 'GET',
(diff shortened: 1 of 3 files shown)


mwiencek:
This looks good, but can we add/update a corresponding test for the JSON WS in t/lib/t/MusicBrainz/Server/Controller/WS/2/JSON/LookupDiscID.pm?

reosarevok:
Sure! Done :)

reosarevok:
Does it actually require any documentation update? The docs say "the 'inc=' arguments supported are identical to a lookup request for a release".

yvanzo:
Oh ok, so the doc is currently wrong about `inc=` arguments supported for `discid` lookup, and this patch makes the web service behaves as expected/already documented?

This is a bit beyond this patch, but why the doc refers to `release` here and where can one find this list of supported arguments?

PR #1039, opened 2019-04-29
MBS-10146: Load number of releases in RG for header

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10146


yvanzo:
Tested, works for me. It could either target the branch `production` or be cherry-picked from `master`.

PR #1040, opened 2019-04-29
MBS-5818 (SQL): Add integer position column to collection entries

yvanzo:
# Change schema to prepare for [MBS-5818](https://tickets.metabrainz.org/browse/MBS-5818): ordered collections

Add `position` column of `INTEGER` type to every `editor_collection_<entity_type>` table.

It will allow collection owner to manually order collection entries.

Default is still unordered collection, with positions set to `NULL`.

### DIFF
--- a/admin/sql/CreateTables.sql
+++ b/admin/sql/CreateTables.sql
@@ -2267,72 +2267,84 @@ CREATE TABLE editor_collection_collaborator (
 CREATE TABLE editor_collection_area (
     collection INTEGER NOT NULL, -- PK, references editor_collection.id
     area INTEGER NOT NULL, -- PK, references area.id
+    position INTEGER NOT NULL DEFAULT 0 CHECK (position >= 0),
     comment TEXT DEFAULT '' NOT NULL
 );
 
 CREATE TABLE editor_collection_artist (
     collection INTEGER NOT NULL, -- PK, references editor_collection.id
     artist INTEGER NOT NULL, -- PK, references artist.id
+    position INTEGER NOT NULL DEFAULT 0 CHECK (position >= 0),
     comment TEXT DEFAULT '' NOT NULL
 );
 
 CREATE TABLE editor_collection_event (
     collection INTEGER NOT NULL, -- PK, references editor_collection.id
     event INTEGER NOT NULL, -- PK, references event.id
+    position INTEGER NOT NULL DEFAULT 0 CHECK (position >= 0),
     comment TEXT DEFAULT '' NOT NULL
 );
 
 CREATE TABLE editor_collection_instrument (
     collection INTEGER NOT NULL, -- PK, references editor_collection.id
     instrument INTEGER NOT NULL, -- PK, references instrument.id
+    position INTEGER NOT NULL DEFAULT 0 CHECK (position >= 0),
     comment TEXT DEFAULT '' NOT NULL
 );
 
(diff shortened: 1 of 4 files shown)


admin/sql/CreateTables.sql
```diff
@@ -2267,72 +2267,84 @@ CREATE TABLE editor_collection_collaborator (
 CREATE TABLE editor_collection_area (
     collection INTEGER NOT NULL, -- PK, references editor_collection.id
     area INTEGER NOT NULL, -- PK, references area.id
+    position INTEGER,
```
mwiencek: If NULL means it's unordered, does position start at 0 or 1?

Starting at 1 (which we do for medium.position and link_order) would have the nice properties that a falsey position always means it's unordered (you don't have both NULL and 0) and that we can use the position for display without having to do position + 1 everywhere. In that case, a value of 0 would mean it's unordered.

reosarevok: So you would suggest NOT NULL DEFAULT 0 as better than just nullable then? :) (I'm not @yvanzo  but just to make sure).

mwiencek: I guess `INTEGER NOT NULL DEFAULT 0 CHECK (position >= 0)` (the check is to prevent negative values) like we have for link_order. But starting at 1 just seems consistent with the other places we use the position for display, and I'm not aware of any position/ordering columns that allow both NULL and 0.

admin/sql/updates/schema-change/25.slave.sql
```diff
@@ -30,6 +31,22 @@ CREATE TABLE editor_collection_collaborator (
 
 ALTER TABLE editor_collection_collaborator ADD CONSTRAINT editor_collection_collaborator_pkey PRIMARY KEY (collection, editor);
 
+--------------------------------------------------------------------------------
+SELECT '20190422-mbs-5818-collection-position.sql';
+
+ALTER TABLE editor_collection_area ADD COLUMN position INTEGER NOT NULL DEFAULT 0;
```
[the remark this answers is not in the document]
yvanzo: Oops, regenerated schema change 25 file.

admin/sql/CreateTables.sql
yvanzo: Where do we not have both NULL and 0? If it seems safer not to use both, should we add a constraint to disallow 0 for this column in the database?

reosarevok: Why not just use `INTEGER NOT NULL DEFAULT 0 CHECK (position >= 0)` if we already do that elsewhere?

mwiencek: We don't allow NULL for positions in general, from what I can see (did I miss any?):

* medium.position has NOT NULL (always ordered, positions start at 1)
* track.position has NOT NULL
* artist_credit_name.position has NOT NULL
* relationships' link_order has NOT NULL (0 indicates unordered, ordering starts at 1)
* types' child_order has NOT NULL (0 indicates unordered, ordering starts at 1)

yvanzo: No worry, I just did not had your two comments before I submitted mine and actually replied to _(you don't have both NULL and 0)_. The proposed check is perfectly fine, added into https://github.com/metabrainz/musicbrainz-server/pull/1040/commits/cb3d0c70b737275a3a4c1576f3595fa90091af95.

PR #1042, opened 2019-04-29
MBS-10148: Allow changing empty collection to any type

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10148

### DIFF
--- a/lib/MusicBrainz/Server/Form/Collection.pm
+++ b/lib/MusicBrainz/Server/Form/Collection.pm
@@ -40,10 +40,12 @@ sub options_type_id {
 
     if ($collection && blessed $collection) {
         my $entity_type = $collection->type->item_entity_type;
-        my %valid_types =
-            map { $_->id => 1 }
-            $self->ctx->model('CollectionType')->find_by_entity_type($entity_type);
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Form/Collection.pm
```diff
@@ -40,10 +40,12 @@ sub options_type_id {
 
     if ($collection && blessed $collection) {
         my $entity_type = $collection->type->item_entity_type;
-        my %valid_types =
-            map { $_->id => 1 }
-            $self->ctx->model('CollectionType')->find_by_entity_type($entity_type);
-        $types = [grep { $valid_types{$_->{value}} } @$types];
+        unless ($self->ctx->model('Collection')->is_empty($entity_type, $collection->{id})) {
+            my %valid_types =
+                map {$_->id => 1}
```
mwiencek: Nitpick: you removed the spaces inside the map block but added them inside the array ref below. If anything I'd prefer the previous styles since those are more prevalent.

yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/commit/0324a59e21978ec77bd0d477b02ced84987dc478

PR #1046, opened 2019-04-30
MBS-9973 (SQL): Add timestamp column to collection entries

yvanzo:
# Change schema to prepare for [MBS-9973](https://tickets.metabrainz.org/browse/MBS-9973): collections sortable by date

Add `added` column of `TIMESTAMP WITH TIME ZONE` type to every `editor_collection_<entity_type>` table.

Set `added` column to `NULL` for all existing entries as it is unknown. 

It will allow to sort collection entries by added date.


mwiencek:
I think we should set all existing columns to NULL in the upgrade script.

reosarevok:
Is this supposed to be set by the user (e.g. "indicate the date it was bought" or whatevs), or set automatically (just "date it was added to collection")?

admin/sql/updates/20190422-mbs-9973-collection-added-date.sql
```diff
@@ -0,0 +1,17 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
+ALTER TABLE editor_collection_area ADD COLUMN added TIMESTAMP WITH TIMESTAMP NOT NULL DEFAULT NOW();
```
mwiencek: These have NOT NULL but the definitions in admin/sql/CreateTables.sql don't. I think we should allow NULL because we have no other value for existing items.

mwiencek:
> Is this supposed to be set by the user (e.g. "indicate the date it was bought" or whatevs), or set automatically (just "date it was added to collection")?

It's set automatically as "date it was added to collection," but allowing it to be set by the user in addition might be a good idea? Then they can fill in the exact date they acquired it if desired.

admin/sql/updates/20190422-mbs-9973-collection-added-date.sql
yvanzo: Oops, yet another glitch, fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/1046/commits/5a55f6d2c2fd92e87aa39b0ef5f78f02803be6e7.

reosarevok:
I'd like that (would also be good for the ones who do have all the info saved about when they bought the ones in their collection already, which I'm sure exist) but also doesn't sound like something that even needs to be added at first, could be a future improvement :)

yvanzo:
It seems doable as further improvement, not necessarily part of the schema change release though.

majkinetor:
I am the author of original Jira request. The main intent is to set here date when it was added to collection. This is the natural way people remember music they posses.

Ability to change it afterwards might not be bad idea too, but its less relevant. Perhaps better to have some arbitrary user notes.

Some other media managers have the option to change this date - Calibre, ebook manager for example, makes this by default as date added to collection but you can change it afterwards.

yvanzo:
Thank you @majkinetor for your input, allowing _arbitrary user notes_ is work in progress, see [MBS-1658](https://tickets.metabrainz.org/browse/MBS-1658).

jesus2099:
> The main intent is to set here date when it was added to collection. This is the natural way people remember music they posses.

I agree.

> Ability to change it afterwards might not be bad idea too, but its less relevant. 

But this would be required as hundreds of releases added before that pull request was merged, have a date added being naturally not set, `NULL`.

PR #1047, opened 2019-04-30
MBS-9925: Convert collection pages to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9925


reosarevok:
Not fully ready since the -list.tt PRs need merging first, but probably ready enough to be reviewed and checked so I can improve any obvious issues.

lib/MusicBrainz/Server/Entity/Collection.pm
```diff
@@ -44,6 +45,7 @@ around TO_JSON => sub {
     $json->{editor} = $self->editor ? $self->editor->TO_JSON : undef;
     $json->{public} = boolean_to_json($self->public);
     $json->{description} = $self->description;
+    $json->{descriptionHtml} = format_wikitext($self->description);
```
mwiencek: Would mildly prefer `description_html` just for consistency with Moose attributes being snake case.

root/collection/CollectionEditForm.js
```diff
@@ -0,0 +1,71 @@
...
+import type {CollectionFormT} from './types';
+
+type Props = {|
+  +$c: CatalystContextT,
+  +collectionTypes: SelectOptionsT,
+  +form: CollectionFormT,
+|};
+
+const CollectionEditForm = ({$c, collectionTypes, form}: Props) => {
+  console.log(collectionTypes);
```
mwiencek:
```suggestion
```

root/collection/CollectionEditForm.js
```diff
@@ -0,0 +1,71 @@
...
+    options: collectionTypes,
+  };
+
+  return (
+    <form action={$c.req.uri} method="post">
+      <fieldset>
+        <legend>{l('Collection details')}</legend>
+        <FormRowTextLong
+          field={form.field.name}
+          label={l('Name:')}
```
mwiencek: `addColon` here and below would be nice since I'm sure we have translations for these without the colon already (probably with too, but)

root/collection/CollectionIndex.js
```diff
@@ -0,0 +1,157 @@
...
+        {collection.descriptionHtml ? (
+          <>
+            <h2>{l('Description')}</h2>
+            {$c.user_exists ||
+              (collection.editor && !collection.editor.is_limited) ? (
+                expand2react(collection.descriptionHtml)
+              ) : (
+                <p className="deleted">
+                  {exp.l(`This content is hidden to prevent spam. 
+                        To view it, please {url|log in}.`,
```
mwiencek:
```suggestion
                          To view it, please {url|log in}.`,
```

root/static/scripts/common/i18n/expand2react.js
```diff
@@ -46,7 +46,7 @@ const condSubstThenTextContent = /^[^<>{}|]+/;
 const percentSign = /(%)/;
 const linkSubstStart = /^\{([0-9A-z_]+)\|/;
 const htmlTagStart = /^<(?=[a-z])/;
-const htmlTagName = /^(a|abbr|b|br|code|em|li|span|strong|ul)(?=[\s\/>])/;
+const htmlTagName = /^(a|abbr|b|br|code|em|li|p|span|strong|ul)(?=[\s\/>])/;
```
mwiencek: Did you end up using `<p>` anywhere? Didn't see any places at a glance.

reosarevok: The collection description is wrapped in <p> by default, apparently.

mwiencek: Oh right, from format_wikitext

reosarevok:
Includes https://github.com/metabrainz/musicbrainz-server/pull/1055 (which should ideally be merged before this)

root/collection/CollectionIndex.js
```diff
@@ -0,0 +1,224 @@
...
+
+const listPicker = (props: Props, ownCollection: boolean) => {
+  switch (props.collectionEntityType) {
+    case 'area':
+      return (
+        <AreaList
+          areas={props.entities}
+          checkboxes={ownCollection ? 'remove' : ''}
+          order={props.order}
+          sortable
```
reosarevok: Is there a way to just specify `checkboxes={ownCollection ? 'remove' : ''}`, `order={props.order}
` and `sortable` once and then pass it to all? Possibly also `entities={props.entities}`, if we choose to generalize that on the lists.

mwiencek: If you put them in an object fist, you can use JSX spread syntax:

```JSX
const sharedProps = {
  checkboxes: ownCollection ? 'remove' : '',
  order: props.order,
  sortable: true,
};

// ...

<AreaList
  areas={props.entities}
  {...sharedProps}
/>
```

For `areas={props.entities}` you can either generalize all the prop names to `entities` as you mentioned, or look up the plural name in entities.json.

reosarevok:
Now includes https://github.com/metabrainz/musicbrainz-server/pull/1056 as well

mwiencek:
One issue I noticed is that the paginator is now shown even if there's only one page.

Another (not from this PR) is that `FormRowCheckbox` doesn't seem to add a space between the checkbox and the label.

root/collection/CollectionCreate.js
```diff
@@ -22,10 +22,10 @@ type Props = {|
 const CollectionCreate = ({collectionTypes, form}: Props) => (
   <Layout
     fullWidth
-    title={l('Add a New Collection')}
+    title={l('Add a new collection')}
```
yvanzo: It can be replaced with already localized message `Create a new collection`

root/collection/CollectionCreate.js
```diff
@@ -22,10 +22,10 @@ type Props = {|
...
   >
     <div id="content">
-      <h1>{l('Add a New Collection')}</h1>
+      <h1>{l('Add a new collection')}</h1>
```
yvanzo: Same as above

root/collection/CollectionDelete.js
```diff
@@ -25,7 +25,7 @@ const CollectionDelete = ({$c, collection}: Props) => (
     entity={collection}
     fullWidth
     page="delete"
-    title={l('Remove')}
+    title={l('Remove collection')}
```
yvanzo: Not sure this is necessary, as it comes just after `Collection “collection”` anyway.
It is consistent with entity delete pages.

root/collection/CollectionEdit.js
```diff
@@ -24,7 +24,7 @@ const CollectionEdit = ({collection, collectionTypes, form}: Props) => (
     entity={collection}
     fullWidth
     page="edit"
-    title={l('Edit Collection')}
+    title={l('Edit collection')}
```
yvanzo: I would be fine with just `Edit`, for the same reasons as above (about `Remove`).

root/collection/CollectionDelete.js
reosarevok: Hmm. Remove instrument says "Remove Instrument" on both title and heading. I'd want to keep "Remove collection" at least on the heading here, and by then I might as well leave it on the title - although you have a point that for example edit instrument just says "Edit". So yeah, will leave title as "Remove" but have heading as "Remove collection", if that's ok.

yvanzo: Heading did not change, my concern was only about duplicating _collection_ term, thanks.

To be fair, I checked artist edit page and recording edit/delete pages only (artist doesn’t have a delete page). Too bad for instrument pages, it should probably use the shorter version instead, as it follows instrument type which makes it clear it is an instrument page already, e.g. _Wind instrument “accordina”_. I just checked release edit/delete pages and realized titles don’t even follow the same pattern. There is probably no consistency here across entity types.

reosarevok: Add a ticket for later? Making them consistent would be reasonable :)

PR #1048, opened 2019-05-01
MBS-9491: Update genre list to use database genres

reosarevok:
First step of genres-in-DB, creates Data::Genre and Entity::Genre.

Also: I haven't added a new entity nor data page since I think 2013, so comments on those would be very appreciated.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Genre.pm
+++ b/lib/MusicBrainz/Server/Data/Genre.pm
@@ -1,0 +1,78 @@
+package MusicBrainz::Server::Data::Genre;
+
+use Moose;
+use MusicBrainz::Server::Data::Utils qw(
+    hash_to_row
+    load_subobjects
+);
+use MusicBrainz::Server::Entity::Genre;
+
+extends 'MusicBrainz::Server::Data::CoreEntity';
+with 'MusicBrainz::Server::Data::Role::Alias' => { type => 'genre' };
+with 'MusicBrainz::Server::Data::Role::CoreEntityCache';
+with 'MusicBrainz::Server::Data::Role::SelectAll';
+with 'MusicBrainz::Server::Data::Role::DeleteAndLog' => { type => 'genre' };
+
+sub _type { 'genre' }
+
+sub _columns {
+    return 'genre.id, genre.gid, genre.name,
+            genre.comment, genre.edits_pending, genre.last_updated';
+}
+
+sub _column_mapping {
+    return {
+        id => 'id',
+        gid => 'gid',
+        name => 'name',
+        comment => 'comment',
+        edits_pending => 'edits_pending',
+        last_updated => 'last_updated',
+    };
+}
+
+sub _id_column {
+    return 'genre.id';
+}
+
+sub load {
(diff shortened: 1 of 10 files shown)


lib/MusicBrainz/Server/Entity/Genre.pm
```diff
@@ -0,0 +1,42 @@
...
+with 'MusicBrainz::Server::Entity::Role::Comment';
+with 'MusicBrainz::Server::Entity::Role::LastUpdate';
+
+sub entity_type { 'genre' }
+
+around TO_JSON => sub {
+    my ($orig, $self) = @_;
+
+    return {
+        %{ $self->$orig },
```
mwiencek: We can probably remove this method since it doesn't do anything and isn't likely to have any additions in the near future (but maybe I'm forgetting something).

lib/MusicBrainz/Server/Data/Genre.pm
```diff
@@ -0,0 +1,115 @@
...
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
```
mwiencek: Can we use the shorter copyright notice we started using recently?

lib/MusicBrainz/Server/Data/Genre.pm
```diff
@@ -0,0 +1,81 @@
+package MusicBrainz::Server::Data::Genre;
+
+use Moose;
+use MusicBrainz::Server::Data::Utils qw(
+    hash_to_row
+    load_subobjects
+    merge_string_attributes
+    merge_table_attributes
+    order_by
```
mwiencek: merge_string_attributes, merge_table_attributes, and order_by are unused in this file

reosarevok:
I want to use sentence case everywhere we can. My reasoning is mostly similar to https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98 but in general, I'm not really sure I would call a tab name a title, and I certainly wouldn't call an h2 a title. Let's use title case where we have titles (so, for songs and artists and all that stuff) but not in random places of the interface :) 

I like the suggestion for the component name, will change!

yvanzo:
Here it is an h1 (not h2), a full page title (not tab name) and a menu entry, let's keep title case :)

reosarevok:
But title case looks absolutely awful and feels bad! :(

reosarevok:
Also FWIW: with "tab name" I mean "the name of a tab in my browser" - which I assume is what you mean with page title :) I don't think those are a real title - for me that's like the title of a work of art or something.

reosarevok:
But whatever, if changing those back means I don't have to have an argument about it and can spend time usefully, let's revert. Let me know if it's ready to merge otherwise.

PR #1049, opened 2019-05-02
[WIP] Convert form.tt and edit.tt in admin attributes to React 

anirudhjain75:


root/admin/attributes/Form.js
```diff
@@ -0,0 +1,120 @@
...
+  frequency: FieldT<number>,
+  iso_code: FieldT<string | null>,
+  iso_number: FieldT<string | null>,
+  name: FieldT<string>,
+};
+
+type Props = {
+  form: FormT<LanguageFieldT> | FormT<ScriptFieldT>,
+  id: number,
+  model: string,
```
mwiencek: About the Flow issues you mentioned on IRC:

You need a way to have Flow associate the model here with the type of form. The primary way to do that is by using a union type for the props:

```JS
type BaseProps = {id: number, parentOptions: ...};

type Props =
  | {form: FormT<LanguageFieldT>, model: 'Language', ...BaseProps}
  | {form: FormT<ScriptFieldT>, model: 'Script', ...BaseProps}
  ;
```

You can then use a [disjoin union refinement](https://flow.org/en/docs/types/unions/#toc-disjoint-unions) below. A limitation of Flow is that this only works if you reference `props.model` in the switch and `props.form` for the form (don't destructure them in the parameter list), though you can save the latter to a variable after the refinement is made.

anirudhjain75: @mwiencek I was trying something like this 

> `type Props =
>   | {
>     form: FormT<LanguageFieldT>,
>     id: number,
>     model: 'Language',
>     parentOptions: $ReadOnlyArray<{label: string, value: number}>}
>   | {
>     form: FormT<ScriptFieldT>,
>     id: number,
>     model: 'Script',
>     parentOptions: $ReadOnlyArray<{label: string, value: number}>};`

But still getting errors and also getting error for model now. 
Screenshot 2019-05-11 at 4 13 06 PM
And below is the previous type issue I had 
Screenshot 2019-05-11 at 4 14 18 PM

PR #1050, opened 2019-05-02
MBS-8867: Don't uppercase "C'mon"

reosarevok:
https://tickets.metabrainz.org/browse/MBS-8867

All eslint fixes except for the first commit (+7|-1) and the last one (which just fixes a couple typos etc in comment strings)

### DIFF
--- a/root/static/scripts/guess-case/utils.js
+++ b/root/static/scripts/guess-case/utils.js
@@ -114,7 +114,7 @@ export function isLowerCaseBracketWord(w) {
 
 /*
  * Words which the pre-processor looks for and puts them into brackets
- * if they arent yet.
+ * if they aren't yet.
  */
 const prepBracketWords = /^(?:cd|disk|12["”]|7["”]|a_cappella|re_edit)$/i;
 
@@ -147,7 +147,7 @@ export function trim(is) {
 }
 
 /*
- * Upper case first letter of word unless it's one of the words
+ * Uppercase first letter of word unless it's one of the words
  * in the lowercase words array.
  */
 export function titleString(gc, is, forceCaps) {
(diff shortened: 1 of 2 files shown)


root/static/scripts/guess-case/utils.js
```diff
@@ -173,7 +173,8 @@ export function titleString(gc, is, forceCaps) {
     // we got an 'til (Until = 'til), lowercase.
     // we got an 'way (Away = 'way), lowercase.
     // we got an 'round (Around = 'round), lowercased
-    } else if (isApostrophe(gc.i.getPreviousWord()) && lc.match(/^(s|round|em|ve|ll|d|cha|re|til|way|all)$/i)) {
+    // we got a 'mon (Come on = C'mon), lowercase
```
mwiencek: Should we uppercase the C below, too (in the "Make it O'Titled, Y'All" part)? (If so an extra small test for that would be nice.)

reosarevok: Good idea. Just changed the new test a bit for both.

mwiencek:
I think the "Regex improvements" commit somehow got squashed with a bunch of eslint changes. It'd be nice to split those because I can't see what any of the actual regex changes are.

reosarevok:
ARGH. Will fix.

reosarevok:
Split that. There were two different eslint commits there too, but that's harder to split so hopefully it's ok.

mwiencek:
Much better now. :) Though there's one other small issue: 44121aeb6d3376cf265f0cdc13cb83c6789f67e9 changes the Make it O'Titled, Y'All regexp from `/^(c|o|y)$/i` to `/^(coy)$/i` which is broken and not an eslint change. It'd be good to revert that so that the intermediary commits don't introduce any unexpected breaking changes (which can be a PITA when using git bisect).

reosarevok:
Good catch! Fixed :)

PR #1051, opened 2019-05-02
MBS-9491: Add last_updated column to genre

reosarevok:
We forgot this, but it is on every entity and there's no reason not to have it for genre.


admin/sql/CreateTables.sql
```diff
@@ -857,7 +857,8 @@ CREATE TABLE genre ( -- replicate (verbose)
     gid                 UUID NOT NULL,
     name                VARCHAR NOT NULL,
     comment             VARCHAR(255) NOT NULL DEFAULT '',
-    edits_pending       INTEGER NOT NULL DEFAULT 0 CHECK (edits_pending >=0)
+    edits_pending       INTEGER NOT NULL DEFAULT 0 CHECK (edits_pending >=0),
+    last_updated        TIMESTAMP WITH TIME ZONE DEFAULT NOW()
```
mwiencek: You also need a trigger to update it
```SQL
CREATE TRIGGER b_upd_genre BEFORE UPDATE ON genre
    FOR EACH ROW EXECUTE PROCEDURE b_upd_last_updated_table();
```

PR #1053, opened 2019-05-02
MBS-9491: Basic page for genre

reosarevok:
To be reviewed after https://github.com/metabrainz/musicbrainz-server/pull/1048

Adds a basic entity page for genre (display only)

PR #1054, opened 2019-05-02
MBS-9491: Allow adding, editing and removing genres

reosarevok:
To be reviewed after https://github.com/metabrainz/musicbrainz-server/pull/1053


root/layout/components/sidebar/GenreSidebar.js
```diff
@@ -9,14 +9,30 @@
...
 import React from 'react';
 
+import {withCatalystContext} from '../../../context';
+
 import LastUpdated from './LastUpdated';
+import RemoveLink from './RemoveLink';
+
+type Props = {|
+  +$c: CatalystContextT,
+  +genre: GenreT,
```
yvanzo: FWIW, `genre: GenreT` was fine too, it is not necessary to accept subtypes with `+` for genre.

root/genre/GenreEditForm.js
```diff
@@ -0,0 +1,49 @@
...
+          label={addColonText(l('Disambiguation'))}
+        />
+      </fieldset>
+    </div>
+    <div className="row no-label">
+      {$c.action.name === 'create' ? (
+        <FormSubmit label={l('Create genre')} />
+      ) : (
+        <FormSubmit label={l('Update genre')} />
+      )}
```
yvanzo: These labels are temporary and could be replaced with already replaced label `'Submit'`.
When genres have edit types, this will likely be replaced with label `'Enter edit'`.

root/server/components.js
```diff
@@ -54,6 +54,9 @@ module.exports = {
   'elections/NotFound': require('../elections/NotFound'),
   'elections/Show': require('../elections/Show'),
   'event/NotFound': require('../event/NotFound'),
+  'genre/GenreCreate': require('../genre/GenreCreate'),
+  'genre/GenreDelete': require('../genre/GenreDelete'),
+  'genre/GenreEdit': require('../genre/GenreEdit'),
```
yvanzo: Verb should probably come first in component name.

reosarevok:
> Entering an homonymous genre with a different comment results in an ISE, shouldn’t it be supported?
> Adding tests in t/lib/t would probably help here.

We cannot support that for now, since we still depend on genres matching tags and tags are just text. If we eventually disconnect genres from tags, *then* we can properly support multiple genres of the same name. But I guess for now I should then just disable adding duplicate names completely.

reosarevok:
> Deleting a deleted genre results in a 404 error, should redirect to the genre list and/or throw an error message instead

Added a NotFound for genre

> Entering a duplicate genre results in an ISE, should throw an error message instead

Will this be automatically solved if and when we start using the edit system for genres? If so, I would just leave it be for now, because it will only be allowed for relationship editors anyway.

root/genre/GenreEditForm.js
reosarevok: Done, and added a TODO linking to https://tickets.metabrainz.org/browse/MBS-10165

root/server/components.js
reosarevok: Doesn't it make sense to keep the same GenreX format as all others? I mean, I do see the point that CreateGenre etc is a bit more acceptable English, but it would feel odd to me to have all other pages as GenreX and these as XGenre.

root/layout/components/sidebar/GenreSidebar.js
reosarevok: We seem to use + pretty much everywhere, so I feel it'd be a bit weird to not use it here even if it's ok not to.

mwiencek: `+` makes the field read-only, so you can't modify the props (because that's bad).

It also allows subtypes of `GenreT` for the property because of that, i.e. *because* the field isn't writable, but here it's only useful for making it read-only. https://flow.org/blog/2016/10/04/Property-Variance/

yvanzo: Did not really mean it should changed, but anyway I got it wrong as I thought default was read-only but it is just type invariant.

root/server/components.js
yvanzo: > Doesn't it make sense to keep the same GenreX format as all others?

No, it doesn’t. :wink:

> I mean, I do see the point that CreateGenre etc is a bit more acceptable English,

Indeed, it is a name not a path.

> but it would feel odd to me to have all other pages as GenreX and these as XGenre.

On the contrary, these are editing actions whereas the others are just displaying data. This is the naming convention already used in practice even though it has not been formalized.

root/genre/NotFound.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import React from 'react';
+
+import NotFound from '../components/NotFound';
+
+const GenreNotFound = () => (
+  <NotFound title={l('Genre Not Found')}>
+    <p>
+      {l('Sorry, we could not find a genre with that MusicBrainz ID')}
```
yvanzo: Final stop `.` is missing.

Maybe linking to the genre list page and/or tag search would be helpful?

root/genre/GenreEditForm.js
yvanzo: Thanks!

root/server/components.js
reosarevok: Fair enough, changed!

root/genre/NotFound.js
reosarevok: Done :)

yvanzo: Perfect!

root/server/components.js
yvanzo: Thanks.

Now it feels odd to me to have `GenreEditForm` instead of `EditGenreForm`, but most probably that is just me.

lib/MusicBrainz/Server/Form/Genre.pm
```diff
@@ -0,0 +1,19 @@
...
+    required => 1,
+);
+
+has_field 'comment' => (
+    type => '+MusicBrainz::Server::Form::Field::Comment',
+);
+
+sub edit_field_names { qw( name comment ) }
+
+1;
```
mwiencek: Can we add the license trailer here?

root/genre/types.js
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export type GenreFormT = FormT<{|
+  +comment: FieldT<string>,
+  +name: FieldT<string>,
```
mwiencek: Nitpick: `ReadOnlyFieldT` for these

lib/MusicBrainz/Server/Form/Genre.pm
reosarevok: Sure thing! We should add it to a lot of other places, but not forgetting it in the new files is probably a good start :D

PR #1056, opened 2019-05-03
Move entity lists to their own folder

reosarevok:
Use entities everywhere, for easy of reuse in components that pick in between several lists


reosarevok:
Dropped the entities generalising, since we couldn't find any way for it to work in a way that didn't make Flow angry anyway.

`ReleaseCatnoList` and `ReleaseLabelList` seem like they're more associated with things like `ReleaseDates`. They could all be in a folder like /release but probably not worth moving them right now since we have enough to do as it is. @yvanzo: maybe add a refactoring task to organize root/components? (all the Form components certainly could moved to a folder /form).

PR #1057, opened 2019-05-05
MBS-10141: Look at all area levels for Area tabs

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10141

get_area_containment_query only looks at area types:
1 (Country)
2 (Subdivision)
3 (City)

While this might make sense for listing containment while linking to the entity, it's quite confusing to skip levels when loading the contained entities in the different tabs of the Area page.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -105,7 +105,7 @@ sub find_by_area {
     my (
         $containment_query,
         @containment_query_args,
-    ) = get_area_containment_query('$2', 'any(array[area, begin_area, end_area])');
+    ) = get_area_containment_query('$2', 'any(array[area, begin_area, end_area])', check_all_levels => 1);
     my $query = "SELECT " . $self->_columns . "
                  FROM " . $self->_table . "
                  WHERE \$1 IN (area, begin_area, end_area) OR EXISTS (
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -105,7 +105,7 @@ sub find_by_area {
     my (
         $containment_query,
         @containment_query_args,
-    ) = get_area_containment_query('$2', 'any(array[area, begin_area, end_area])');
+    ) = get_area_containment_query('$2', 'any(array[area, begin_area, end_area])', 1);
```
mwiencek: Changes look good, but it might improve clarity slightly if we made this a keyword arg like `, check_all_levels => 1)`.

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -256,9 +258,9 @@ sub get_area_containment_query {
         SELECT ad.descendant, ad.parent, ad.depth
           FROM area_descendants ad
           JOIN area a ON a.id = ad.parent
```
mwiencek: I think we could also move the JOIN itself to the `$levels_condition` (don't seem to use `a` otherwise)?

reosarevok: Sure, seems reasonable. Done!

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -398,23 +398,13 @@ __PACKAGE__->meta->make_immutable;
 no Moose;
 1;
 
-=head1 COPYRIGHT
+=head1 COPYRIGHT AND LICENSE
 
 Copyright (C) 2009 Lukas Lalinsky
-Copyright (C) 2011 MetaBrainz Foundation
+Copyright (C) 2019 MetaBrainz Foundation
```
mwiencek: Thanks for fixing these to use the shorter version. :) I would generally leave the publication dates untouched unless the file is rewritten, though (or if substantial modifications are made, just append the current date).

reosarevok: Oh, I was wondering about that. Mostly I've seen the date just been changed (but not sure *when* to change it) and some have 2013-15 or whatever (whatever *that* means)...

mwiencek:
@yvanzo can you give an example area MBID where you don't see any area containments? I compared the pages against production (connecting my local server to queen) and didn't see any differences, but may have missed something.

yvanzo:
@mwiencek, you did not, I did miss that sample data dump may lacks some relationships even though it contains area `a420c2cb-d753-4916-bf71-1713421b13bf`.

PR #1058, opened 2019-05-05
MBS-10142: Weibo autoselect improvements + sidebar

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10142


lib/MusicBrainz/Server/Entity/URL/Weibo.pm
```diff
@@ -0,0 +1,26 @@
...
+use Moose;
+
+extends 'MusicBrainz::Server::Entity::URL';
+with 'MusicBrainz::Server::Entity::URL::Sidebar';
+
+override href_url => sub {
+    shift->url->as_string =~ s{^http:}{https:}r;
+};
+
+sub sidebar_name { 'Sina Weibo' }
```
yvanzo: Is _Sina Weibo_ the title of the website displayed when accessed from China?
Here it seems to be _微博国际_ (which translates to _Weibo International_).
If so, I would suggest to use just _Weibo_ instead.
It should probably be localized too.

reosarevok: Changed!

yvanzo: Commit message excepted!

PR #1061, opened 2019-05-08
MBS-9428: Allow multiple users to share one collection

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9428

Also solves https://tickets.metabrainz.org/browse/MBS-9381 and https://tickets.metabrainz.org/browse/MBS-10246


root/user/collections.tt
```diff
@@ -15,19 +15,22 @@
...
-    <h2>[% l("Collections") %]</h2>
     [% IF viewing_own_profile %]
         <p>[% l('You have no collections.') -%]</p>
     [%- ELSE -%]
         <p>[% l('{user} has no public collections.', { user => html_escape(user.name) }) -%]</p>
     [%- END -%]
 [%- END -%]
 
+<h2>[% l("Collaborative collections") %]</h2>
+[% INCLUDE 'collaborator-collections' %]
```
mwiencek: Maybe I don't have the terminology mapped out in my head correctly yet, but it seems slightly strange how these two lines use different terms for what I think are the same thing (collaborative vs. collaborator collections).

reosarevok:
I thought it would be a bit weird to have collections which are private yet shared, but maybe that makes sense as well for like a group of friends or something :) Will look into it (and other things) tomorrow.

reosarevok:
This now allows people to collaborate on private collections. 

While doing this I also fixed https://tickets.metabrainz.org/browse/MBS-9381 which turned out to be even more broken than the original reporter thought. It should now do the intended thing (see commit desc).

lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -264,6 +268,23 @@ sub delete_editor {
     );
 }
 
+sub set_collaborators {
+    my ($self, $collection_id, $collaborators) = @_;
+
+    $self->sql->begin;
+
+    $self->sql->do('DELETE FROM editor_collection_collaborator WHERE collection = ?', $collection_id);
+    $self->sql->insert_many(
```
yvanzo: [`insert_many`](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-04-26/lib/Sql.pm#L159-L183) returns immediately if no array is given, but it seems to execute useless statements if empty array is given, or am I missing something? Ideally `insert_many` should be updated to skip empty array too, but if it is intended its input be checked here before calling it.

reosarevok: Did that (I think, anyway!)

reosarevok:
I figured out that the select on CreateCollection doesn't let you select any other options than the default. Looking into that, so this isn't really mergeable just yet.

reosarevok:
Form fixed, this can be reviewed again.

lib/MusicBrainz/Server/Controller/Role/Collection.pm
```diff
@@ -45,27 +56,39 @@ role
...
             }
         }
 
         $c->stash
-          (collections => $collections,
+          (own_collections => $own_collections,
+           collaborator_collections => $collaborator_collections,
            containment => \%containment,
-           all_collections => $entity_collections,
+           number_of_collections => $number_of_visible_collections + $number_of_non_visible_collections,
```
mwiencek: Something isn't right here, because if I'm on an entity which isn't in any collection, the collection section in the sidebar isn't displayed at all, so I can't add it to anything.

root/layout/components/sidebar/CollectionLinks.js
```diff
@@ -20,33 +20,37 @@ type Props = {|
 |};
 
 const CollectionLinks = ({$c, entity}: Props) => {
-  const allCollections = $c.stash.all_collections;
-  if (!$c.user_exists || !allCollections) {
+  const numberOfCollections = $c.stash.number_of_collections || 0;
+  if (!$c.user_exists || numberOfCollections < 1) {
```
mwiencek: Perhaps we should remove the `numberOfCollections` check? Then at least an "Add to a new collection" link would be shown, which still seems useful.

mwiencek:
Re: "A couple more issues," those are if the collection is private.

lib/MusicBrainz/Server/Data/Collection.pm
yvanzo: Sorry, ignore my above comment, I confused array with array ref: `insert_many` takes an array rather than an array ref - it is  given a dereferenced array ref here - thus it makes no difference.

reosarevok: Should I drop 0fe4da2ad47154e893fcac1c70b0635c29e7b1eb then?

root/layout/components/sidebar/CollectionLinks.js
reosarevok: Sure, seems reasonable. Done.

reosarevok:
Made changes that should help with those issues.

lib/MusicBrainz/Server/Data/Collection.pm
yvanzo: Sure

root/layout/components/sidebar/AttendanceLinks.js
```diff
@@ -25,28 +25,32 @@ const AttendanceLinks = ({$c, event}: Props) => {
...
+      addCollectionText={l('Add to a new list')}
+      collaborativeCollections={$c.stash.collaborator_collections}
+      collaborativeCollectionsHeader={l('Collaborative lists')}
+      collaborativeCollectionsNoneText={
+        l('Not collaborating in any attendance lists!')
+      }
+      entity={event}
+      header={l('Attendance')}
+      ownCollections={$c.stash.own_collections}
+      ownCollectionsHeader={l('My lists')}
```
yvanzo: It is very vague, please either rename as 'My attendance lists' or add context 'attendance'.

root/layout/components/sidebar/AttendanceLinks.js
```diff
@@ -25,28 +25,32 @@ const AttendanceLinks = ({$c, event}: Props) => {
...
-          />
-        }
-      />
-    </>
+    <CollectionList
+      addCollectionText={l('Add to a new list')}
+      collaborativeCollections={$c.stash.collaborator_collections}
+      collaborativeCollectionsHeader={l('Collaborative lists')}
+      collaborativeCollectionsNoneText={
+        l('Not collaborating in any attendance lists!')
```
yvanzo: What does collaborating in an attendance list mean? Isn’t that something personal?

root/user/collections.tt
```diff
@@ -72,3 +75,47 @@
...
+                            <td>[% collection.public ? l('Public') : l('Private') %]</td>
+                        </tr>
+                    [%- END -%]
+                </tbody>
+            </table>
+        [% END %]
+    [% END %]
+[% ELSE %]
+    [% IF viewing_own_profile %]
+        <p>[% l('You aren\'t collaborating in any collections.') -%]</p>
```
yvanzo: Should use the typographic apostrophe and collaborate **on** collections.
```perl
         <p>[% l('You aren’t collaborating on any collections.') -%]</p>
```

root/user/collections.tt
```diff
@@ -72,3 +75,47 @@
...
+    [%- ELSE -%]
+        <p>[% l('{user} isn\'t collaborating in any collections.', { user => html_escape(user.name) }) -%]</p>
```
yvanzo: Same as above

root/layout/components/sidebar/AttendanceLinks.js
reosarevok: Whatever the users want. I assume it's not common, but you could have a bunch of friends who pick concerts to go to as a group or something.

reosarevok:
> It seems easy to spam through collaborative collections list, no invitation/application?

Yes, because we don't have any good way of doing that without resorting to sending emails and making people click links, that I can see. So the idea is we trust our users to only add people who are ok with it, and if that seems not to be the case we can change the way it works later and make it more complicated.

If you're worried about actual spammers, we can only show the collaborator form to non-beginner users or something.

reosarevok:
> I cannot distinguish my collections with collaborators from my other collections in collections list.
> I cannot see which collections I’m owning/collaborating on when visiting collections list of other editors.

Both fixed with the latest commit.

root/layout/components/sidebar/CollectionLinks.js
```diff
@@ -25,28 +25,32 @@ const CollectionLinks = ({$c, entity}: Props) => {
...
+      addCollectionText={l('Add to a new collection')}
+      collaborativeCollections={$c.stash.collaborator_collections}
+      collaborativeCollectionsHeader={l('Collaborative collections')}
+      collaborativeCollectionsNoneText={
+        l('Not collaborating in any collections!')
+      }
+      entity={entity}
+      header={l('Collections')}
+      ownCollections={$c.stash.own_collections}
+      ownCollectionsHeader={l('My collections')}
```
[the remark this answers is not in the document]
reosarevok: That doesn't work at all with the style of all the other headings here though.

root/user/collections.tt
mwiencek: So what was the resolution here? :) I'm still not sure what the difference is but they're both still used about equally.

root/user/collections.tt
```diff
@@ -72,3 +88,47 @@
...
+        [%- IF collabentity.value.size -%]
+            <h3>[% format_entity_collection_title(collabentity.key) %]</h3>
+
+            <table class="tbl">
+                <thead>
+                    <tr>
+                        <th>[% l('Collection') %]</th>
+                        <th>[% l('Type') %]</th>
+                        <th>[% format_plural_entity_type_name(collabentity.key) %]</th>
+                        [% IF viewing_own_profile %]
```
mwiencek: Any reason we don't have a column for the collaborator count down here too?

mwiencek:
Shouldn't `inc=user-collections` on /ws/2/collection and /ws/2/release include private collections you're collaborating on? Otherwise e.g. Picard won't list those as collections you can add/remove releases from. We include `<editor>` elements in those to indicate the owners of the collections, so I don't see any real issue there.

root/user/collections.tt
reosarevok: Collaborative sounds more English so let's use that.

root/user/collections.tt
reosarevok: Not really, actually. Also made a change so that it specifies if it's your collection in the privacy column (there because then it's easier to understand why you're getting shown a private collection).

reosarevok:
Collection::find_by still works, AFAICT, but it's becoming messy. We might want to have a refactoring ticket to make it less spaghetiish sometime?

root/layout/components/sidebar/AttendanceLinks.js
yvanzo: Until now the number of attendance lists used to match the number of MusicBrainz editors to attend this event. Maybe users want to know the number of editors attending to the event rather than the number of attendance lists?

reosarevok: There's no real way of knowing though - that a list has 3 collaborators does not guarantee us that all 3 are really planning to attend.

yvanzo: It is a bit confusing that attendance lists don’t have the same meaning for everyone.

By the way, I just found another (preexisting) issue: all event collections are listed as attendance lists whereas “Attending” and “Maybe attending” are just subtypes of event collections.

(Collections were not even listed at all before [React conversion of event sidebar](https://github.com/metabrainz/musicbrainz-server/commit/87d726642ca812b6712b68c6c0394fd49af264af).)

mwiencek:
`inc=user-collections` works for collaborators on /ws/2/collection browse requests now. 👍 

It doesn't seem to work for /ws/2/release/, though.

A top-level request to /ws/2/collection should also include collaborative collections, I think.

reosarevok:
I think those should work now? :)

root/layout/components/sidebar/CollectionLinks.js
```diff
@@ -20,33 +20,37 @@ type Props = {|
...
-          />
-        }
-      />
-    </>
+    <CollectionList
+      addCollectionText={l('Add to a new collection')}
+      collaborativeCollections={$c.stash.collaborative_collections}
+      collaborativeCollectionsHeader={l('Collaborative collections')}
+      collaborativeCollectionsNoneText={
+        l('Not collaborating on any collections!')
```
mwiencek: Should we be showing a "Collaborative collections" section in the sidebar if there are none? Most users won't be participating in any and telling them that on every page doesn't seem super useful, since it's not like they can take any action about it (you have to be added to one by someone else). (But the only downside is it taking up space and adding more work for translators.)

root/layout/components/sidebar/AttendanceLinks.js
```diff
@@ -20,33 +20,37 @@ type Props = {|
...
+      addCollectionText={l('Add to a new list')}
+      collaborativeCollections={$c.stash.collaborative_collections}
+      collaborativeCollectionsHeader={l('Collaborative lists')}
+      collaborativeCollectionsNoneText={
+        l('Not collaborating on any attendance lists!')
+      }
+      entity={event}
+      header={l('Attendance')}
+      ownCollections={$c.stash.own_collections}
+      ownCollectionsHeader={l('My attendance lists')}
```
mwiencek: When I think about it, it's weird that we assume every event collection should be an attendance list. I think having the default "Attending," "Maybe attending" collections is enough to satisfy that use case, and doesn't mean we should label all uses as "Attendance lists." Especially with collaborative collections, I can imagine some like "events where band X performed with lineup Y" or "highest grossing events" or "highest attended events" (granted that last one kind of is an attendance list of sorts :)). Or any collection where the events share some attribute about the artists/performers that might be interesting and either hard to store in MB or hard to search for.

root/layout/components/sidebar/CollectionLinks.js
reosarevok: I thought it might be nice as a way of showing that they are a thing, but you might have a point that the section would soon become more annoying than useful. Will look into it :)

root/layout/components/sidebar/AttendanceLinks.js
reosarevok: I've always thought it's a bit weird, so I'm more than happy to consolidate all of these as just collections :)

root/layout/components/sidebar/CollectionLinks.js
reosarevok: Hidden. I also added a "Global usage" heading to the "in X user collections" the bottom, because otherwise it didn't look very good. Not completely sure about the wording though!

mwiencek: Hmm, might be clearer if it mirrors the wording of the other heading ("My collections"), so maybe "Other user collections" or similar if it's not too long.

reosarevok: Changed it to "Other collections" (since it's under "My collections" and "Collaborative collections", it should be clear it's other users')

mwiencek:
I think this has sat in review for long enough and would be much better served if people could test it on beta. We can always make further fixes/modifications to the code later (though it looks good overall to me, as I said in my last review :)).

PR #1062, opened 2019-05-09
 MBS-9491: Use DB genres in more places

reosarevok:
This replaces every place (AFAICT) where the current JS layer uses the hardcoded genres.

The autoselect is worse than before though, since the ws/js direct search only matches on full words.


reosarevok:
One possibility for the autoselect would be to just implement https://tickets.metabrainz.org/browse/MBS-9880 as a list of all genre strings, and use that instead of a ws/js search.

root/types.js
```diff
@@ -16,7 +16,7 @@
  */
 
 declare type AggregatedTagT = {|
-  +tag: string,
+  +tag: TagT,
   +count: number,
```
yvanzo: Should _last updated_ be moved here instead of deleted?

root/tag/TagIndex.js
```diff
@@ -65,28 +73,40 @@ function buildSection<T>(
...
-    {buildSection(props, 'label', l('Labels'), N_ln('See all {num} labels', 'See all {num} labels'))}
-    {buildSection(props, 'place', l('Places'), N_ln('See all {num} places', 'See all {num} places'))}
-    {buildSection(props, 'release_group', l('Release Groups'), N_ln('See all {num} release groups', 'See all {num} release groups'))}
-    {buildSection(props, 'release', l('Releases'), N_ln('See all {num} releases', 'See all {num} releases'))}
-    {buildSection(props, 'recording', l('Recordings'), N_ln('See all {num} recordings', 'See all {num} recordings'))}
-    {buildSection(props, 'series', l('Series'), N_ln('See all {num} series', 'See all {num} series'))}
-    {buildSection(props, 'work', l('Works'), N_ln('See all {num} works', 'See all {num} works'))}
-  </TagLayout>
-);
+      {buildSection(props, 'area', l('Areas'), N_ln('See all {num} areas', 'See all {num} areas'))}
```
yvanzo: Would be nice to commit large reindent apart, plus above comment is not correctly indented.

lib/MusicBrainz/Server/Entity/Tag.pm
```diff
@@ -10,16 +10,30 @@ has 'name' => (
...
-    return shift->name;
-}
+has 'genre_id' => (
+    is => 'rw',
+    isa => 'Int'
+);
+
+has 'genre' => (
+    is => 'rw',
+    isa => 'Genre'
```
yvanzo: Shouldn’t that be optional/maybe?

reosarevok: I thought about that, but Entity::Artist for example has 

`has 'begin_area_id' => (
    is => 'rw',
    isa => 'Int'
);

has 'begin_area' => (
    is => 'rw',
    isa => 'Area'
);`

mwiencek: You could say `Maybe[Genre]` or `Maybe[Area]` are more correct since not all tags have a genre and not all artists have a begin area, but it wouldn't cause any issue without `Maybe` unless you pass `genre => undef` or `begin_area => undef` to the constructor, which you wouldn't normally do. (Moose differentiates between an attribute being unset, and being set to undef. Since these are `rw`, you don't need to set them.)

lib/MusicBrainz/Server/Entity/AggregatedTag.pm
```diff
@@ -34,7 +34,7 @@ sub TO_JSON {
     my ($self) = @_;
 
     return {
-        tag => $self->tag->TO_JSON,
+            tag => $self->tag,
```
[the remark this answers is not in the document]
mwiencek: Also, how come the `->TO_JSON` was removed?

root/types.js
reosarevok: Why? It AggregatedTag doesn't seem to have a last_updated, or am I missing something?

root/tag/TagIndex.js
reosarevok: But if I commit the reindent separately then I'll have a commit with the wrong style, isn't that bad?

lib/MusicBrainz/Server/Controller/User.pm
```diff
@@ -430,8 +430,8 @@ sub tags : Chained('load') PathPart('tags')
     }
 
     my $tags = $c->model('Editor')->get_tags($user);
-    my @display_tags = grep { !$_->{tag}->is_genre_tag } @{ $tags->{tags} };
-    my @display_genres = grep { $_->{tag}->is_genre_tag } @{ $tags->{tags} };
+    my @display_tags = grep { !$_->{tag}->{genre_id} } @{ $tags->{tags} };
+    my @display_genres = grep { $_->{tag}->{genre_id} } @{ $tags->{tags} };
```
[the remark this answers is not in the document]
mwiencek: Looks like it has an object like `{tag => Entity::Tag, count => Int}`. So no, this outer object is unblessed/not a Moose object, only the tag is.

root/static/scripts/common/components/TagEditor.js
```diff
@@ -362,32 +357,41 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
...
+            },
+            vote: t.vote,
+          };
+
+          if (index >= 0) {
+            newTags[index] = tag;
+          } else {
+            newTags.push(tag);
+          }
+          this.setState({tags: sortedTags(newTags)});
```
mwiencek: These changes look a bit off since it only needs to make a request if the tag wasn't deleted, and it should only call `setState` once at the end. (It doesn't appear to call `setState` at all for deleted tags. It's also not impossible that a callback for one `$.getJSON` call might run before a previous one, or not complete at all.)

But this is kind of tricky to do cleanly, and given that the autocomplete is also made worse by having to use `/ws/js`, I think we should maybe just seed the `TagEditor` with a list of all genre names in the props and mostly stick to the previous implementation. :/ What do you think?

reosarevok: We were talking about making it so that there was a way to ask the WS for all genre strings (and deal with https://tickets.metabrainz.org/browse/MBS-9880 at the same time). Not sure what's the best endpoint for that.

mwiencek: We should probably have a browse endpoint in /ws/2 that includes the gid, comment, etc. Not certain if we'll have a /ws/2 endpoint that just returns a giant list of names, though, which is basically what we need here. I think we could just add the latter to /ws/js for now (with the list cached in redis).

mwiencek: But I think we'll still want to pass the list as a prop, rather than fetching them from /ws/js in the component, because the component should still display something if the user isn't logged in or has JavaScript disabled.

root/tag/TagIndex.js
yvanzo: Oh right, that is just due to new block level, not fixing old indent.

root/types.js
yvanzo: I probably confused it with something else, cannot recall what.

root/static/scripts/common/components/TagEditor.js
reosarevok: Updated this (passing it through the stash since it seems easier that way than having it as a prop on every place that has genres on the sidebar). Quick test seems to work, but it can certainly use more eyes.

root/types.js
```diff
@@ -846,6 +847,9 @@ declare type SanitizedCatalystContextT = {|
   +req: {|
     +uri: string,
   |},
+  +stash?: {
```
mwiencek: Looks like the change below always includes an empty object, so no `?`

root/utility/sanitizedContext.js
```diff
@@ -20,13 +20,17 @@ function sanitizedContext(
...
   return {
     action: {
       name: $c.action.name,
     },
     req: {
       uri: $c.req.uri,
     },
+    stash: genre_list ? ({
+      genre_list: $c.stash.genre_list,
+    }) : {},
```
mwiencek: `$c.stash.genre_list` is stored in a variable already so you can simplify that a bit to `genre_list ? {genre_list} : {}`

root/static/scripts/common/components/TagEditor.js
```diff
@@ -420,21 +425,27 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
...
-          _.sortBy(
-            ($.ui.autocomplete.filter(
-              _.without(GENRE_TAGS_ARRAY, ...terms),
-              last,
-            ): $ReadOnlyArray<string>),
-            [x => x.startsWith(last) ? 0 : 1, _.identity],
-          ),
-        );
+        const genreTerms = genreList.filter(function (genre) {
+          return genre.name;
```
mwiencek: When would `genre` be defined but not `genre.name`?

root/utility/sanitizedContext.js
mwiencek: Though eslint might complain about the `genre_list` assignment above too, should probably name it `genreList` and do `{genre_list: genreList}` there

root/static/scripts/common/components/TagEditor.js
```diff
@@ -362,16 +362,21 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
 
   updateTags(updatedUserTags: $ReadOnlyArray<TagUpdateT>) {
     const newTags = this.state.tags.slice(0);
-
```
mwiencek: Sometimes blank lines are removed like here, but otherwise added like in https://github.com/metabrainz/musicbrainz-server/pull/1062/files#diff-0f30970e52953a6ba4194b9157fe8b4b and I'm not sure I see an improvement (or un-improvement) either way, so I think it'd be nicer to keep the diff clean, but doesn't matter much at this point anyway :)

root/static/scripts/common/components/TagEditor.js
```diff
@@ -284,21 +281,20 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
       };
 
       if (!this.state.positiveTagsOnly || isAlwaysVisible(t)) {
-        const genre = isGenre(t);
-
+        const isGenre = this.genreList.find(genre => (genre.name === t.tag.name));
```
mwiencek: I noticed the primary use of `genreList` is searching the entire list for a specific tag name. It might be better then if it were `genreMap`, so you could simply lookup genres by name directly.

There is a case below that does `genreList.map(genre => genre.name)` to get a list of names, but that could be replaced with a call to `Object.keys(genreMap)`.

lib/MusicBrainz/Server/Data/EntityTag.pm
```diff
@@ -108,13 +107,71 @@ sub find_user_tags_for_entities
...
+        my ($model, $row) = @_;
+        return MusicBrainz::Server::Entity::UserTag->new(
+            tag_id    => $row->{tag_id},
+            editor_id => $user_id,
+            entity_id => $row->{entity},
+            is_upvote => $row->{is_upvote},
+        );
+    });
+
+    $self->c->model('Tag')->load(@tags);
```
mwiencek: You actually have all the info needed to construct each tag in the query, so could just add a `tag => MusicBrainz::Server::Entity::Tag->new(...)` key/value pair to the `UserTag` construction above.

lib/MusicBrainz/Server/Data/EntityTag.pm
```diff
@@ -108,13 +107,71 @@ sub find_user_tags_for_entities
...
 }
 
+sub find_genres_for_entities
+{
+    my ($self, @ids) = @_;
+
+    return unless scalar @ids;
+
+    my $query = "SELECT tag.id as tag_id, tag.name, entity_tag.count,
+                        entity_tag.".$self->type." AS entity, genre.id as genre_id
```
mwiencek: just for readability
```suggestion
                        entity_tag." . $self->type . " AS entity, genre.id as genre_id
```

lib/MusicBrainz/Server/Controller/WS/js/Genre.pm
```diff
@@ -0,0 +1,29 @@
...
+
+sub type { 'genre' }
+
+sub search : Chained('root') PathPart('genre') {
+    my ($self, $c) = @_;
+    $c->stash->{args}{direct} = 'true';
+    $self->dispatch_search($c);
+}
+
+1;
```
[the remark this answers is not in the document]
reosarevok: Heh. That's because it's missing on all (or most, anyway) of the ws/js controllers. But I'll add it here at least.

root/static/scripts/common/components/TagEditor.js
reosarevok: I removed the new line on that file at least, since that was the only change to that file apparently in the end, and readded it here, since it seems sensible to have.

lib/MusicBrainz/Server/Data/EntityTag.pm
reosarevok: Did that as a separate commit.

mwiencek:
Rebased and made some small fixes/tweaks (removed an unneeded `$c->model('Tag')->load`, moved some sorts to `ORDER BY`, changed some indentation).

root/utility/sanitizedContext.js
```diff
@@ -21,6 +21,10 @@ function sanitizedContext(
 ) /*: SanitizedCatalystContextT */ {
   const stash = $c.stash;
   const user = $c.user;
+  const genreMap = stash.genre_map;
+  const genreMapStashObject = genreMap
+    ? ({genre_map: genreMap})
+    : {};
```
reosarevok: This feels a bit weird, so if there's a better way to combine this with current_language, let me know!

mwiencek:
The changes I made in https://github.com/metabrainz/musicbrainz-server/pull/1062#issuecomment-532993850 were reverted so I re-applied those.

PR #1063, opened 2019-05-10
MBS-9973 (part 2): Load and render collection items added date

yvanzo:
## [MBS-9973](https://tickets.metabrainz.org/browse/MBS-9973): Date Added as sortable column in collection

* Create a proxy entity CollectionItem for entities that contains `added` date. (Other columns could be loaded on the same model.)
* JSON export of CollectionItem is the base entity with an extra `collection_item` attribute object that has an `added` attribute.
* New function `find_items_by_collection` is used for display only. (Web service endpoint keeps using `find_by_collection` for now.)
* Collection-specific columns display is implemented using callback functions so as to provision for comment display and for conditional hiding of empty extra columns.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Collection.pm
+++ b/lib/MusicBrainz/Server/Controller/Collection.pm
@@ -92,9 +92,10 @@ sub show : Chained('load') PathPart('') {
     my $order = $c->req->params->{order};
 
     my $model = $c->model(type_to_model($entity_type));
-    my $entities = $self->_load_paged($c, sub {
-        $model->find_by_collection($collection->id, shift, shift, $order);
+    my $collection_items = $self->_load_paged($c, sub {
+        $model->find_items_by_collection($collection->id, shift, shift, $order);
     });
+    my $entities = map { $_->entity } @$collection_items;
 
     if ($model->can('load_related_info')) {
         $model->load_related_info(@$entities);
@@ -154,7 +155,7 @@ sub show : Chained('load') PathPart('') {
     my %props = (
         collection           => $collection,
         collectionEntityType => $entity_type,
-        entities             => $entities,
+        entities             => $collection_items,
         order                => $order,
         pager                => serialize_pager($c->stash->{pager}),
     );
--- a/lib/MusicBrainz/Server/Data/Role/Collection.pm
+++ b/lib/MusicBrainz/Server/Data/Role/Collection.pm
@@ -29,6 +29,32 @@ sub find_by_collection {
     $self->query_to_list_limited($query, [$collection_id], $limit, $offset);
 }
 
+sub find_items_by_collection {
+    my ($self, $collection_id, $limit, $offset, $order) = @_;
+
+    my ($order_by, $extra_join, $also_select) = $self->_order_by($order);
+    $extra_join //= '';
+
+    my $table = $self->_table;
+    my $type = $self->_type;
+
+    my $query = "
+      SELECT *
+      FROM (
+      SELECT DISTINCT ON (" . $self->_id_column . ")
+        " . $self->_columns .
+          ($also_select ? ", $also_select" : "") . "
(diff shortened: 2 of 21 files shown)


lib/MusicBrainz/Server/Data/Role/Collection.pm
```diff
@@ -29,6 +32,69 @@ sub find_by_collection {
     $self->query_to_list_limited($query, [$collection_id], $limit, $offset);
 }
 
+sub find_items_by_collection {
```
mwiencek: Wouldn't it make sense to just return the same order in the web service?

Since this duplicates `find_by_collection` other than the ordering and entity creation, it'd seem better if we can just do that, and fix the one spot in the web service to handle `CollectionItems`.

lib/MusicBrainz/Server/Controller/Collection.pm
```diff
@@ -92,9 +92,10 @@ sub show : Chained('load') PathPart('') {
     my $order = $c->req->params->{order};
 
     my $model = $c->model(type_to_model($entity_type));
-    my $entities = $self->_load_paged($c, sub {
-        $model->find_by_collection($collection->id, shift, shift, $order);
+    my $collection_items = $self->_load_paged($c, sub {
+        $model->find_items_by_collection($collection->id, shift, shift, $order);
     });
+    my $entities = map { $_->entity } @$collection_items;
```
mwiencek: That'll assign the size of the array, so you instead need:
```suggestion
    my $entities = [map { $_->entity } @$collection_items];
```
I would've suggested just `my @entities = ...` but it uses `@$entities` like a hundred times below.

lib/MusicBrainz/Server/Entity/CollectionItem.pm
```diff
@@ -0,0 +1,51 @@
...
+
+has 'entity' => (
+    is => 'rw',
+    isa => 'Entity',
+    required => 1,
+);
+
+has 'added' => (
+    is => 'rw',
+    isa => 'PgDateStr',
```
mwiencek:
```suggestion
    isa => 'Maybe[Str]',
```

(Moose doesn't know how to deal with `Maybe[PgDateStr]` for some reason, but it's fine to use `Str` in any case since we're not coercing from `DateTime` objects.)

lib/MusicBrainz/Server/Data/Role/Collection.pm
```diff
@@ -29,6 +32,69 @@ sub find_by_collection {
...
+        ORDER BY id
+      ) $type
+      ORDER BY $order_by";
+
+    $self->query_to_list_limited(
+        $query,
+        [$collection_id],
+        $limit,
+        $offset,
+        $self->_new_collection_item_from_row);
```
mwiencek: That's a method call, so you instead need:
```suggestion
        \&_new_collection_item_from_row);
```

lib/MusicBrainz/Server/Entity/CollectionItem.pm
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    my $added = $self->added;
+    if (defined $added) {
+        $added = DateTime::Format::Pg->parse_datetime($self->added);
+        $added->set_time_zone('UTC');
+        $json->{collection_item}->{added} = $added->iso8601 . 'Z';
+    }
+
+    return $json;
```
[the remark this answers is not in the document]
yvanzo: Right, I modelled website JSON types with web service in mind, but it does not have to be the same.

lib/MusicBrainz/Server/Data/Role/Collection.pm
yvanzo: The web service should indeed use the same function in the end, I duplicated it on purpose so as to focus on the website first.

PR #1064, opened 2019-05-10
MBS-9491: List genre on tag search results

reosarevok:
Goes after https://github.com/metabrainz/musicbrainz-server/pull/1062

It makes sense to show genres for tags - it will make even more sense once we have aliases and the genre associated to a tag can have a different name than the tag itself.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Search.pm
+++ b/lib/MusicBrainz/Server/Controller/Search.pm
@@ -178,6 +178,10 @@ sub direct : Private
             $c->model('Event')->load_related_info(@entities);
             $c->model('Event')->load_areas(@entities);
         }
+        when ('tag') {
+            #TODO: add support for genre aliases when finishing MBS-10062
+            $c->model('Genre')->load(@entities);
+        }
     }
 
     if ($type =~ /(recording|release|release_group)/)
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Controller/Search.pm
```diff
@@ -178,6 +178,9 @@ sub direct : Private
             $c->model('Event')->load_related_info(@entities);
             $c->model('Event')->load_areas(@entities);
         }
+        when ('tag') {
+            $c->model('Genre')->load(@entities);
+        }
```
yvanzo: Won’t it fail on non primary tags?

reosarevok: You mean on tags that are not genres? If so, it seemed to work fine when I tested it.

yvanzo: On tags that are associated with genre but not as primary tag?

reosarevok: Oh. We don't actually have those yet, so not sure :D

reosarevok: Added a comment to remind us to support aliases once we add them

PR #1067, opened 2019-05-13
MBS-9762: Standardize Songkick URLs

yvanzo:
# Resolves [MBS-9762](https://tickets.metabrainz.org/browse/MBS-9762): Standardize Songkick URLs

1. Replace songkick.com with www.songkick.com and deny other domains

2. Remove slug, subpath, query, fragment, and deny unclean URLs
   * It helps with preventing duplicates (with/without slug)
   * Only festivals have mandatory `/id/<edition id>` subpath
   * Songkick accepts URLs without slug and auto-add slug to it

Update tests accordingly, and add artist test with unneeded subpath.

Edit: Additionally disallow linking Songkick festivals to places (which was an unexploited bug).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1937,7 +1937,7 @@ const CLEANUPS = {
           case LINK_TYPES.songkick.event:
             return prefix === 'concerts' || prefix === 'festivals';
           case LINK_TYPES.songkick.place:
-            return prefix === 'venues' || prefix === 'festivals';
+            return prefix === 'venues';
         }
       }
       return false;
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1920,12 +1920,17 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?songkick\\.com', 'i')],
     type: LINK_TYPES.songkick,
     clean: function (url) {
-      return url.replace(/^http:\/\//, 'https://');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?songkick\.com\//, 'https://www.songkick.com/');
+      url = url.replace(/^(https:\/\/www\.songkick\.com\/[a-z]+\/[0-9]+)(?:-[\w-]*)?(\/id\/[0-9]+)?(?:[-\/?#].*)?$/, '$1$2');
+      return url;
     },
     validate: function (url, id) {
-      const m = /songkick\.com\/([a-z]+)\//.exec(url);
+      const m = /^https:\/\/www\.songkick\.com\/([a-z]+)\/[0-9]+(?:\/(id)\/[0-9]+)?$/.exec(url);
       if (m) {
         const prefix = m[1];
+        if ((m[2] === 'id') !== (prefix === 'festivals')) {
+          return false;
+        }
         switch (id) {
           case LINK_TYPES.songkick.artist:
             return prefix === 'artists';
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2699,18 +2699,25 @@ const testData = [
...
+            expected_clean_url: 'https://www.songkick.com/artists/3909026',
+       only_valid_entity_types: ['artist'],
+  },
   {
                      input_url: 'http://www.songkick.com/festivals/74586-ruhrpott-rodeo/id/19803209-ruhrpott-rodeo-festival-2014',
              input_entity_type: 'event',
     expected_relationship_type: 'songkick',
-            expected_clean_url: 'https://www.songkick.com/festivals/74586-ruhrpott-rodeo/id/19803209-ruhrpott-rodeo-festival-2014',
+            expected_clean_url: 'https://www.songkick.com/festivals/74586/id/19803209',
        only_valid_entity_types: ['event', 'place'],
```
reosarevok: Why would festivals be valid for place?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1920,12 +1920,17 @@ const CLEANUPS = {
...
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?songkick\.com\//, 'https://www.songkick.com/');
+      url = url.replace(/^(https:\/\/www\.songkick\.com\/[a-z]+\/[0-9]+)(?:-[\w-]*)?(\/id\/[0-9]+)?(?:[-\/?#].*)?$/, '$1$2');
+      return url;
     },
     validate: function (url, id) {
-      const m = /songkick\.com\/([a-z]+)\//.exec(url);
+      const m = /^https:\/\/www\.songkick\.com\/([a-z]+)\/[0-9]+(?:\/(id)\/[0-9]+)?$/.exec(url);
       if (m) {
         const prefix = m[1];
+        if ((m[2] === 'id') !== (prefix === 'festivals')) {
```
reosarevok: I'm not sure I'm following this one. Is this supposed to be "if m[2] is id then prefix should be festivals"? If so, isn't there a more legible way of writing that? :/

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: After digging the code base for [initial commit](https://github.com/metabrainz/musicbrainz-server/commit/790ffea4522ab5526927806bafd370852c9d4b60#diff-3e811e588fddba5fbac605a42e3e13d9R888) and the database which has no festival linked to place, there seems to be no reason, it looks more like a bug, fixed in https://github.com/metabrainz/musicbrainz-server/pull/1067/commits/e7c9b6529c3d6f77b8923ba7061bd30c741487b4.

root/static/scripts/edit/URLCleanup.js
yvanzo: It is more than that: _"if m[2] is id then prefix should be festivals"_ and "if prefix is festivals then m[2] should be id", that is "m[2] is id **if and only if** prefix is festivals", which is negated here with central `!`.

reosarevok: If this is the easiest way to say this, then let it be :)

mwiencek: It's basically an XOR, which JavaScript doesn't have an operator for, so it's at least the most compact way to write it. The alternative is `if ((m[2] === 'id' && prefix !== 'festivals') || (m[2] !== 'id' && prefix === 'festivals')) return false` but that's not much clearer to me.

PR #1068, opened 2019-05-13
MBS-10167: Display the right medium position when reusing existing media

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10167

Also fixed https://tickets.metabrainz.org/browse/MBS-10171 which I found when testing, let me know if that looks right.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js.pm
@@ -49,6 +49,12 @@ sub medium : Chained('root') PathPart Args(1) {
     my ($self, $c, $id) = @_;
 
     my $medium = $c->model('Medium')->get_by_id($id);
+
+    unless ($medium) {
+        $c->stash->{error} = 'No medium found with this ID.';
+        $c->detach('bad_req');
+    }
+
     $c->model('MediumFormat')->load($medium);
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Controller/WS/js.pm
```diff
@@ -49,6 +49,13 @@ sub medium : Chained('root') PathPart Args(1) {
     my ($self, $c, $id) = @_;
 
     my $medium = $c->model('Medium')->get_by_id($id);
+
+    unless ($medium) {
+        $c->stash->{error} = "No medium found with this ID.";
+        $c->detach('bad_req');
+        return;
```
mwiencek: Technically the return is redundant since detach dies (i.e. never returns to the current action).

lib/MusicBrainz/Server/Controller/WS/js.pm
```diff
@@ -118,12 +125,10 @@ sub tracklist_results {
         my $count = 0;
```
yvanzo: Not needed anymore.

reosarevok:
Probably because it was badly written. We checked with @mwiencek and couldn't find any reasoning for it.

PR #1069, opened 2019-05-13
MBS-10160: Display username on sidebar for youtube/c/ links

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10160

It's unclear to me ATM if we should just clean those up to remove /c/ (the one case I tried worked without it), but this shouldn't hurt in the meantime.


reosarevok:
My point is https://www.youtube.com/MetaBrainz is the same as https://www.youtube.com/c/MetaBrainz :)

yvanzo:
Oh, we should probably keep only one format then. I would be surprised if there is no doc about this.

reosarevok:
https://support.google.com/youtube/answer/2657968?hl=en "This is called a custom URL and looks like youtube.com/yourcustomname and youtube.com/c/yourcustomname" - sounds like they are equivalent, yes.

reosarevok:
@yvanzo: added a cleanup for removing /c/, let me know if it seems ok.

PR #1070, opened 2019-05-13
MBS-10163: Regenerate locale list for DateTime::Locale 1.23

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10163


mwiencek:
cpanfile.snapshot looks correct, I think (1.23 is already the version referenced in that file, and running in the containers).

docker/musicbrainz-test-database/cpanfile.snapshot should possibly be updated at some point, but we don't use that image for MB (it's used by CB and possibly some other Python projects for their tests). Nor does it run any MBS code other than for DB imports (nothing server related). So it doesn't necessarily need to be updated as it can import dumps properly for those projects' tests.

PR #1072, opened 2019-05-13
MBS-6087: Show relationship edits for recordings on release edit list

reosarevok:
https://tickets.metabrainz.org/browse/MBS-6087


reosarevok:
Unrelated and tricky (since we'd want to only select the specific relationship somehow - is this a thing we can do in any reasonable way at the moment? do we have edits_pending for specific relationships? If we do, then we need a ticket and I can look into it at some point too)

mwiencek:
There should be a way by overriding `alter_edit_pending`, but it's not something we do for relationships in general, so I'm not sure it makes sense to create an exception here unless we do it for all. For example, if you're to currently add or edit or remove a relationship, we don't even increment `edits_pending` for the entities *on* the relationship, let alone indirect entities (like releases using a particular recording). Perhaps that's something we want, but it's a separate issue at least.

PR #1075, opened 2019-05-15
MBS-9491 / MBS-10062: Dump genre and genre_alias tables

reosarevok:


reosarevok:
We don't seem to generally add aliases tables fully to SampleDump but this one is small enough that it might make sense, so I added it. Let me know if you think it shouldn't be there.

PR #1077, opened 2019-05-17
WS container fixes

mwiencek:
Fixes https://tickets.metabrainz.org/browse/MBS-8826 plus an issue related to the git_info script failing.

### DIFF
--- a/lib/MusicBrainz/Server/ControllerBase/WS/2.pm
+++ b/lib/MusicBrainz/Server/ControllerBase/WS/2.pm
@@ -103,7 +103,12 @@ sub method_not_allowed : Private {
     ));
 }
 
-sub begin : Private { }
+sub begin : Private {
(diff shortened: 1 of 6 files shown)


lib/MusicBrainz/Server/ControllerBase/WS/2.pm
```diff
@@ -103,7 +103,12 @@ sub method_not_allowed : Private {
     ));
 }
 
-sub begin : Private { }
+sub begin : Private {
+    my ($self, $c) = @_;
+
+    $c->stash->{serializer} = $self->get_serialization($c);
```
mwiencek: This previously happened in `validate`, but that's only run if we already have a valid path. We need a serializer even for unknown paths, so it has to be initialized in `begin` now.

PR #1078, opened 2019-05-21
Remove some unused code

mwiencek:
Just a couple small things I noticed over the last couple days that I figured I'd open a PR for.


mwiencek:
> sortable_table_header is obvious to me, but why the other ones? Did we use to test more things than we do now? 

We used to test the external links editor in JS using ReactTestUtils, but those were converted to Selenium back in 21235a805c0936f6aa91d1cc88d5db799dac9eaf. Which is testing the same things more thoroughly.

PR #1079, opened 2019-05-21
Update babel-eslint to 11.0.0-beta.0

mwiencek:
This resolves an issue with parsing Flow's inexact object type syntax (`{...}`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
   },
   "devDependencies": {
     "@babel/cli": "7.1.0",
-    "babel-eslint": "10.0.1",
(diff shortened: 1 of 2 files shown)


mwiencek:
We generally don't, but I don't see a problem with it if it works (especially for dev dependencies - this is only used by us developers running eslint :)).

reosarevok:
It also adds semver but that's not a beta anyway. Sure, why not.

mwiencek:
Right, semver is a new dependency of babel-eslint, so we don't have any control over that.

PR #1080, opened 2019-05-22
MBS-10108: Allow linking tweets to recordings only

yvanzo:
[MBS-10108](https://tickets.metabrainz.org/browse/MBS-10108): Allow linking tweets to recordings only
----

Handle Twitter update permalinks such as `https://twitter.com/user/status/123`:

* auto-select “stream music for free” relationship type,
* same cleanup as for linking to Twitter accounts,
* validate linking Twitter updates to recordings only.

See https://blog.twitter.com/en_us/a/2006/twitter-has-permalinks-and-rss-feeds.html

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2094,7 +2094,15 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?twitter\\.com/', 'i')],
     type: LINK_TYPES.socialnetwork,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:(?:www|mobile)\.)?twitter\.com(?:\/#!)?\/@?([^\/?#]+)(?:[\/?#].*)?$/, 'https://twitter.com/$1');
+      url = url.replace(
+        /^(?:https?:\/\/)?(?:(?:www|mobile)\.)?twitter\.com(?:\/#!)?\//,
+        'https://twitter.com/'
+      );
+      url = url.replace(
+        /^(https:\/\/twitter\.com)\/@?([^\/?#]+)(?:[\/?#].*)?$/,
+        '$1/$2'
+      );
+      return url;
     },
   },
   'unwelcomeimages': { // Block images from sites that don't allow deeplinking
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2092,17 +2092,32 @@ const CLEANUPS = {
   },
   'twitter': {
     match: [new RegExp('^(https?://)?([^/]+\\.)?twitter\\.com/', 'i')],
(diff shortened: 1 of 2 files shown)


reosarevok:
I guess having the cleanup be shared would be tricky if at all possible and it's not that big an extra amount of code, so that should be fine.

yvanzo:
@reosarevok Found the trick, let me know which version you prefer.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2106,7 +2107,10 @@ const CLEANUPS = {
       const m = /^https:\/\/twitter\.com\/[^\/?#]+(\/status\/\d+)?$/.exec(url);
       if (m) {
         const isATweet = !!m[1];
-        return isATweet === (id === LINK_TYPES.streamingmusic.recording);
+        if (_.includes(LINK_TYPES.streamingmusic, id)) {
+          return isATweet && (id === LINK_TYPES.streamingmusic.recording);
+        }
+        return !isATweet;
```
reosarevok: Can you please explain what this one does to me? :)

yvanzo: There was no validation for Twitter accounts in the previous version.

This one auto-selects `streamingmusic` for any Twitter link added to recording/release, validates tweets only for recording, and invalidates any Twitter link for release. It keeps auto-selecting `socialnetwork` for other entity types (in `socialnetwork`) and validates Twitter accounts only for these.

* v1 has :-1: two CLEANUPS entries with :-1: duplicated code (`match` and `cleanup`), :+1: simple `type`s and :+1: only one simple `validate` for tweets:
https://github.com/metabrainz/musicbrainz-server/blob/2c1b2a250568744e185ee0a557b7346ec054f190/root/static/scripts/edit/URLCleanup.js#L2088-L2124

* v2 has :+1: only one CLEANUPS entry with :+1: no duplicated code, :grey_question: aggregated `type` and :grey_question: slightly more complicated `validate` for both tweets and accounts:
  https://github.com/metabrainz/musicbrainz-server/blob/36e7b43b58fa8dc4a460a225f0a5b29b2f716138/root/static/scripts/edit/URLCleanup.js#L2088-L2117

----

P.S. I can actually simplify type aggregation by removing
```javascript
      {recording: LINK_TYPES.streamingmusic.recording},
```
from `type` as `socialnetwork` has no `recording` entry.

reosarevok: I meant from the previous version in the previous commit where it only has the isATweet check, but I understand that's to block them completely from release? I guess that's fine then :)

yvanzo: OK, the last commit adds relationship type auto-select for Twitter URLs on releases and invalidates such link. Linking to Twitter on releases used to be unchecked and thus free for linking with any unrestricted relationship type.

Both versions do prevent linking tweets to releases, differences are mostly about coding style.

PR #1087, opened 2019-05-27
MBS-9914: Convert the area public pages to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9914


lib/MusicBrainz/Server/Controller/Area.pm
```diff
@@ -256,6 +256,29 @@ sub places : Chained('load')
...
+                map {
+                    my $json = $_->TO_JSON;
+                    # These arguments aren't needed at all to render the map,
+                    # and only increase the page size.
+                    delete @{$json}{qw(annotation unaccented_name)};
+                    $json;
+                } grep { $_->coordinates } @$places
+            ],
+        }),
+        places      => $places,
```
yvanzo: Is it still necessary to stash `places` (and `map_data_args`) above?

reosarevok: Oh. Duh. No, it really isn't.

PR #1088, opened 2019-05-29
Convert add_remove_alias.tt in /root/edit/details/ to React

anirudhjain75:


root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
+// @flow
```
mwiencek: Copy the license header from here: https://github.com/metabrainz/musicbrainz-server/blob/de7d769042e189d4b64236da28146b4235aa903c/root/edit/details/AddArea.js#L1-L8 except replace "MetaBrainz Foundation" with your name. :)

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
...
+/* eslint-disable no-negated-condition */
```
mwiencek: I'd remove this and fix the negated conditions, unless there's a good reason to have them.

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
...
+};
+
+const AddRemoveAlias = ({edit}: {edit: Props}) => {
+  const display = edit.display_data;
+  const entityType = display.entity_type;
+  const entity = display[entityType];
+  return (
+    <table className={`details ${edit.edit_kind}-${entityType}-alias`}>
+      <tr>
+        <th>{addColon(formatEntityTypeName(entityType))}</th>
```
mwiencek:
```suggestion
        <th>{addColonText(formatEntityTypeName(entityType))}</th>
```

Same for the uses of `addColon` below.

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
...
+        <th>{addColon(l('Alias'))}</th>
+        <td>{isolateText(display.alias)}</td>
+      </tr>
+      {(display.sort_name !== '') ? (
+        <tr>
+          <th>{addColon(l('Sort name'))}</th>
+          <td>{display.sort_name}</td>
+        </tr>
+      ) : null}
+      {(display.locale) ? (
```
mwiencek: I'd remove the unnecessary parentheses around all the conditions for consistency with our other React code.

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
...
+        <>
+          <tr>
+            <th>{addColon(l('Locale'))}</th>
+            <td>{display.locale}</td>
```
mwiencek: Small issue here: this displays the code ("de") instead of the name ("German"). You'll need to import root/static/scripts/common/constants/locales.json and do `locales[display.locale]` here to get the name. (Unfortunately it won't be translated, but neither was the Perl.)

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
...
+          </tr>
+        </>
+      ) : null}
+      {(display.type) ? (
+        <tr>
+          <th>{addColon(l('Type'))}</th>
+          <td>{display.type.name}</td>
+        </tr>
+      ) : null}
+      {(!isDateEmpty(display.begin_date)) ? (
```
mwiencek: To resolve those negated conditions, for example, just do `{isDateEmpty(display.begin_date) ? null : (...)}` here.

root/types.js
```diff
@@ -419,6 +419,7 @@ declare type EditT = {|
   +created_time: string,
   +data: Object,
   +edit_type: number,
+  +edit_kind: string,
```
mwiencek: Nitpick: should be sorted above `edit_type`.

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,95 @@
...
+          <tr>
+            <th>{addColon(l('Primary for locale'))}</th>
+            <td>{yesNo(display.primary_for_locale)}</td>
+          </tr>
+        </>
+      ) : null}
+      {(display.type) ? (
+        <tr>
+          <th>{addColon(l('Type'))}</th>
+          <td>{display.type.name}</td>
```
mwiencek:
```suggestion
          <td>{lp_attributes(display.type.name, 'alias_type')}</td>
```
to translate the type name.

root/edit/details/AddRemoveAlias.js
anirudhjain75: I dont understand why there is an error, if possible please give me a hint on why the error is caused

mwiencek: I gave an example here: https://github.com/metabrainz/musicbrainz-server/pull/1088/files/de7d769042e189d4b64236da28146b4235aa903c#discussion_r290369636

Instead of `!foo ? y : x`, do `foo ? x : y`.

root/edit/details/AddRemoveAlias.js
anirudhjain75: addColonText was used so suppress flow errors, then return of formatEntityTypeName is Str so I used addColon, I dont think there is much difference in both other than PropTypes

root/edit/details/AddRemoveAlias.js
anirudhjain75: Yea Ill look into this

root/edit/details/AddRemoveAlias.js
anirudhjain75: I'll look into this... thanks

root/edit/details/AddRemoveAlias.js
mwiencek: `addColonText` is guaranteed (both at run time and by the type system) to return a string, so it should be used where a string is expected.

root/edit/details/AddRemoveAlias.js
anirudhjain75: I'll look into this too and make the necessary changes. :)

root/edit/details/AddRemoveAlias.js
anirudhjain75: why is this done ?

mwiencek: `display.type.name` is the type name in English. `lp_attributes(display.type.name, 'alias_type')` passes it to gettext (which we use to translate strings). So if someone is browsing MusicBrainz in another language, it'll display the type name in their language. :)

root/edit/details/AddRemoveAlias.js
mwiencek: To be clear, it wouldn't cause any particular issue at runtime if `addColon` returned a React element here, since it would display fine inside a `<th>` element. But it would be unexpected, since our expectation is for a plain string to be displayed. :)

(The parser is also (slightly) more efficient if you use `addColonText` since it doesn't need to look for HTML, but that's a negligible difference.)

anirudhjain75:
@mwiencek I've made the changes you requested and pushed the code.

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,104 @@
...
+import isDateEmpty from '../../static/scripts/common/utility/isDateEmpty';
+import DescriptiveLink from '../../static/scripts/common/components/DescriptiveLink';
+import EntityLink from '../../static/scripts/common/components/EntityLink';
+import locales from '../../static/scripts/common/constants/locales';
+import isolateText from '../../static/scripts/common/utility/isolateText';
+import yesNo from '../../static/scripts/common/utility/yesNo';
+
+type AliasEntityTypeT = $ElementType<AliasEntityTypeT, 'entityType'>;
+
+type DisplayDataT = {
```
mwiencek: I'd prefer creating an `AddRemoveAliasEditT` type using the definition from `Props` below, and inlining the `display_data` definition there. Since we may want to refer to these edits outside of props. See root/edit/details/AddAnnotation.js for example.

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,104 @@
...
+  const entityType = display.entity_type;
+  const entity = display[entityType];
+  return (
+    <table className={`details ${edit.edit_kind}-${entityType}-alias`}>
+      <tr>
+        <th>{addColonText(formatEntityTypeName(entityType))}</th>
+        <td>
+          <DescriptiveLink entity={entity} />
+          {' '}
+          {(entity && entity.gid) ? bracketed(<EntityLink content={l('view all aliases')} entity={entity} subPath="aliases" />) : null}
```
mwiencek: This line is too long

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,104 @@
...
+        </td>
+      </tr>
+      <tr>
+        <th>{addColonText(l('Alias'))}</th>
+        <td>{isolateText(display.alias)}</td>
+      </tr>
+      {(display.sort_name === '') ? null : (
```
mwiencek: There's a small bug here, since `sort_name` can be `null`. Just use `display.sort_name ?` (and invert the order of the expressions).

root/edit/details/AddRemoveAlias.js
```diff
@@ -0,0 +1,104 @@
...
+      )}
+      {display.locale ? (
+        <>
+          <tr>
+            <th>{addColonText(l('Locale'))}</th>
+            <td>{locales[display.locale]}</td>
+          </tr>
+          <tr>
+            <th>{addColonText(l('Primary for locale'))}</th>
+            <td>{yesNo(display.primary_for_locale)}</td>
```
mwiencek: There's another bug here in that `primary_for_locale` can be stored as `"0"` in the edit data (which evaluates as truthy here). The fix is to use `boolean_to_json` on the server.

lib/MusicBrainz/Server/Edit/Alias/Add.pm
```diff
@@ -3,6 +3,7 @@ use MooseX::Role::Parameterized;
 use MooseX::Types::Moose qw( Bool Int Str );
 use MooseX::Types::Structured qw( Dict Optional );
 use MusicBrainz::Server::Data::Utils qw( model_to_type non_empty );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
```
mwiencek: `Data::Utils` is already imported right above this line, so just add `boolean_to_json` inside the `qw()` above. (Perl tip: `qw` stands for "quote word" and turns words separated by whitespace into a list of words. so `qw( a b )` is the same as `('a', 'b',)`.)

lib/MusicBrainz/Server/Edit/Alias/Delete.pm
```diff
@@ -4,6 +4,7 @@ use MooseX::ABC;
 
 use MooseX::Types::Moose qw( Bool Int Str );
 use MooseX::Types::Structured qw( Dict Optional );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
```
mwiencek: Same comment, `Data::Utils` is already imported below.

lib/MusicBrainz/Server/Edit/Alias/Edit.pm
```diff
@@ -17,6 +17,7 @@ use MusicBrainz::Server::Edit::Utils qw(
     date_closure
     merge_partial_date
 );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
```
mwiencek: Same comment

mwiencek:
Did you see the following text from my previous two reviews?

> I think we can also remove root/edit/details/add_remove_alias.tt in this PR?

> root/edit/details/add_remove_alias.tt still needs to be removed as mentioned in my previous review

PR #1089, opened 2019-05-31
MBS-10206: Convert user profile to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10206


root/types.js
```diff
@@ -364,28 +364,40 @@ declare type EditorPreferencesT = {|
...
   +is_wiki_transcluder: boolean,
+  +languages: $ReadOnlyArray<EditorLanguageT> | null,
+  +last_login_date: string | null,
   +name: string,
   +preferences: EditorPreferencesT,
   +registration_date: string,
   +website: string | null,
 |};
 
+declare type EditorLanguageT = {|
```
mwiencek: It might be a good idea to define a `TO_JSON` subroutine in lib/MusicBrainz/Server/Entity/EditorLanguage.pm (rather than rely on the auto-conversion) so there aren't any surprises. :)

root/user/UserProfile.js
```diff
@@ -0,0 +1,515 @@
...
+            )}
+          </UserProfileProperty>
+        ) : null}
+
+        {user.languages && user.languages.length > 0 ? (
+          <UserProfileProperty name={l('Languages:')}>
+            <ul className="inline">
+              {user.languages.map(language => (
+                <li key={language.language.id}>
+                  {`${l_languages(language.language.name)} (${l(language.fluency)})`}
```
mwiencek: I don't think the `l()` call works there, because the fluency is lowercase while the translated strings in lib/MusicBrainz/Server/Form/User/EditProfile.pm have their first letter capitalized. That's probably why `l_fluency` in Perl uses `ucfirst`, first. You could import `lodash/capitalize` to do the same.

It'd also be nice if `bracketedText` was used for the parentheses. (If that's done, it'd not be necessary to embed any of it inside a template string.)

root/user/UserProfile.js
```diff
@@ -0,0 +1,515 @@
...
+            <>
+              {viewingOwnProfile ? user.email : l('(hidden)')}
+              {' '}
+              {user.email_confirmation_date ? (
+                exp.l('(verified at {date})',
+                      {
+                        date: formatUserDate(
+                          $c.user, user.email_confirmation_date,
+                        ),
+                      })
```
mwiencek: The whitespace looks a little excessive compared to something like
```JavaScript
                exp.l('(verified at {date})', {
                  date: formatUserDate($c.user, user.email_confirmation_date),
                })
```

root/user/UserProfile.js
```diff
@@ -0,0 +1,515 @@
...
+      <h2>{l('General Information')}</h2>
+
+      <table className="profileinfo" role="presentation">
+        <UserProfileProperty name={l('Email:')}>
+          {user.email ? (
+            <>
+              {viewingOwnProfile ? user.email : l('(hidden)')}
+              {' '}
+              {user.email_confirmation_date ? (
+                exp.l('(verified at {date})',
```
mwiencek: Some of these should probably use texp.l/bracketedText (though it's not a huge deal).

root/user/UserProfile.js
```diff
@@ -0,0 +1,515 @@
...
+                ),
+              )}
+            </>
+          ) : null}
+        </UserProfileProperty>
+
+        {user.biography ? (
+          <UserProfileProperty className="biography" name={l('Bio:')}>
+            {showBioAndURL ? (
+              <div dangerouslySetInnerHTML={{__html: user.biography}} />
```
mwiencek: Would it make sense to use `expand2react` instead?

root/user/UserProfile.js
```diff
@@ -0,0 +1,515 @@
...
+            </th>
+          </tr>
+        </thead>
+
+        <tbody>
+          <UserProfileProperty name={lp('Accepted', 'edit descriptor')}>
+            {exp.l(
+              '{count} ({view_url|view})',
+              {
+                count: Number(editStats.accepted_count).toLocaleString(),
```
mwiencek: Using `formatCount` from root/statistics/utilities.js seems like a good idea since it passes the current UI language to `toLocaleString` (rather than defaulting, I assume, to the browser's language).

root/user/UserProfile.js
reosarevok: I thought we wanted date to be at the same indentation as the string. But ok, made the change.

root/user/UserProfile.js
reosarevok: They should, but isn't it better not to break the translation?

root/user/UserProfile.js
reosarevok: Done, but we should probably consider moving the function out of statistics/utilities and using it in PaginatedResults and TagCloud too then, since they also use toLocaleString().

Wondering also whether to use formatPercentage (from the same file) for percentages here which are currently l('{percentage}%').

root/user/UserProfile.js
mwiencek: Yeah, that's true, If every argument is on a separate line they should line up. Just not sure we need to force them onto separate lines in all cases.

root/user/UserProfile.js
mwiencek: Heh, I didn't realize I was commenting on a line that had brackets in the string (didn't mean to imply changing it). I just meant changing `exp.l` to `texp.l` and `bracketed` to `bracketedText` where possible elsewhere in the file (unrelated to this line).

reosarevok: I did that in a few places :)

root/user/UserProfile.js
```diff
@@ -0,0 +1,520 @@
...
+    <>
+      <h2>{l('Statistics')}</h2>
+
+      <table className="statistics">
+        <thead>
+          <tr>
+            <th colSpan="2">
+              {exp.l(
+                'Edits ({view_url|view})',
+                {view_url: `/user/${encodeURIComponent(user.name)}/edits/all`},
```
mwiencek: Can we store encodedName in a var since it's used so many times?

root/user/UserProfile.js
```diff
@@ -0,0 +1,520 @@
...
+        </tbody>
+        <tfoot>
+          <tr>
+            <th headers="table_vote_summary_vote">
+              {voteTotals.name}
+            </th>
+            <th className="totals" headers="table_vote_summary_recent">
+              {formatCount($c, Number(voteTotals.recent.count))}
+              {' '}
+              {bracketedText(l('100%'))}
```
mwiencek: I agree we should use `formatPercentage` here as wondered in https://github.com/metabrainz/musicbrainz-server/pull/1089#discussion_r290687724

root/user/UserProfile.js
```diff
@@ -0,0 +1,520 @@
...
+            </th>
+          </tr>
+        </thead>
+
+        <tbody>
+          <UserProfileProperty name={lp('Accepted', 'edit descriptor')}>
+            {exp.l(
+              '{count} ({view_url|view})',
+              {
+                count: formatCount($c, Number(editStats.accepted_count)),
```
mwiencek: How come these all need to be wrapped in `Number`? (The Flow type is already `number`, so either that's wrong or this isn't needed.)

root/user/UserProfile.js
```diff
@@ -0,0 +1,537 @@
...
+        ) : null}
+
+        {user.languages && user.languages.length > 0 ? (
+          <UserProfileProperty name={l('Languages:')}>
+            <ul className="inline">
+              {user.languages.map(language => (
+                <li key={language.language.id}>
+                  {l_languages(language.language.name)}
+                  {' '}
+                  {bracketedText(l(capitalize(language.fluency)))}
```
yvanzo: This is a faithful rewrite of the original Perl subroutine `l_fluency` that was calling `ucfirst` but it was/is still hackish. It only works because fluency’s labels can be programmatically deduced from associated text values. Maybe that trick should be documented in https://github.com/metabrainz/musicbrainz-server/blob/v-2019-08-22/lib/MusicBrainz/Server/Form/User/EditProfile.pm#L70-L77 so as to avoid breaking user profile when renaming one of these options.

root/user/UserProfile.js
```diff
@@ -0,0 +1,537 @@
...
+import formatUserDate from '../utility/formatUserDate';
+import {canNominate} from '../utility/voting';
+
+function generateUserTypesList(user: EditorT) {
+  const typesList = [];
+  if (user.deleted) {
+    typesList.push(l('Deleted User'));
+  }
+  if (user.is_auto_editor) {
+    typesList.push(exp.l('{doc|Auto-Editor}', {doc: '/doc/Editor'}));
```
yvanzo: Bonus: While you are at it, would it be ok to link to the specific anchor for every link to this doc page?

t/lib/t/MusicBrainz/Server/Controller/User/Edit.pm
```diff
@@ -86,9 +86,9 @@ test 'Hide biography and website/homepage of beginners/limited users from not-lo
     $mech->get('/user/new_editor');
     html_ok($mech->content);
     $tx = test_xpath_html($mech->content);
-    $tx->ok('//tr[@class="biography"]/td[count(*)=1]/p[@class="deleted" and starts-with(text(), "This content is hidden to prevent spam.")]',
+    $tx->ok('//tr[@class="biography"]/td[count(*)=1]/div[@class="deleted" and starts-with(text(), "This content is hidden to prevent spam.")]',
         'biography field of beginner/limited user is hidden from not-logged-in user');
-    $tx->ok('//tr/td[preceding-sibling::th[1][normalize-space(text())="Homepage:"]]/p[@class="deleted" and starts-with(text(), "This content is hidden to prevent spam.")]',
+    $tx->ok('//tr/td[preceding-sibling::th[1][normalize-space(text())="Homepage:"]]/div[@class="deleted" and starts-with(text(), "This content is hidden to prevent spam.")]',
```
yvanzo: I don’t notice any difference with vertical alignment here. Which browser was impacted?

By the way, commit message suggests biography used to be inside a `<p>` tag but it doesn’t seem to have ever been the case. The one change seems to actually be: The message about website/biography of beginner being hidden from visitor is displayed in a `<div>` tag instead of a `<p>` tag previously.

yvanzo: I can reproduce this bug with Chrome and Firefox too, but for _Homepage_ field only.

_Biography_ field has a style workaround in [`root/static/styles/extra/user.less` on line 20](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-10-17-hotfixes/root/static/styles/extra/user.less#L20-L22)

root/user/UserProfile.js
yvanzo: It duplicates [`options_languages_fluency`](https://github.com/metabrainz/musicbrainz-server/blob/62c2a79bc1db08fe0a40240e75d84c42ddc94bac/lib/MusicBrainz/Server/Form/User/EditProfile.pm#L70-L77) now.

reosarevok: Hmm. Can we reuse fluencyOptions from EditProfileForm somehow for this instead? (Guess that'd be easier than passing it from an unrelated Perl file, anyway).

t/lib/t/MusicBrainz/Server/Controller/User/Edit.pm
reosarevok: Would you prefer to have the same style workaround for the homepage field then?

root/user/UserProfile.js
yvanzo: It is potentially used by `FormHandler`, otherwise probably yes.

t/lib/t/MusicBrainz/Server/Controller/User/Edit.pm
yvanzo: No preference. I just wanted to mention it affects homepage only, not biography.

PR #1090, opened 2019-06-01
React autocomplete component

mwiencek:
This is a basic implementation that just shows a list of names; the entity-specific formatters which display more details will be converted later.

To test this, run `./script/compile_resources.sh tests` and navigate to http://localhost:5000/static/scripts/tests/autocomplete2.html.

I'd like if we could check that the basic functionality and controls work as we'd like here (I tried to keep them almost identical to the existing autocomplete), and if they do, merge this mostly as-is plus any requested changes. We could then convert the entity formatters in a followup-PR, and soon after open a PR to replace the jQuery UI one with this.

### DIFF
--- a/flow-typed/npm/lodash_v4.x.x.js
+++ b/flow-typed/npm/lodash_v4.x.x.js
@@ -1,5 +1,5 @@
-// flow-typed signature: a9b75804169260d49cda34b56dcfabe1
-// flow-typed version: e9dac1347c/lodash_v4.x.x/flow_>=v0.63.x
+// flow-typed signature: 978ec408ef4dc26808325571d235d3ca
+// flow-typed version: 9f33da9d84/lodash_v4.x.x/flow_>=v0.63.x
 
 declare module "lodash" {
   declare type Path = $ReadOnlyArray<string | number> | string | number;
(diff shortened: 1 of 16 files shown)


mwiencek:
Updated this to show how it could be used as a vocal autocomplete (or any kind of autocomplete for a predefined list of items) on the autocomplete2.html page.

So I'd intend to replace not just the entity autocomplete, but also root/static/scripts/relationship-editor/common/multiselect.js with this (which is used for attribute types like vocals in the relationship editor dialog).

mwiencek:
That happens in master too (since a long time ago, after we switched to webpack). I didn't find a nice way to fix it yet, but it doesn't prevent the tests from running and can be safely ignored.

yvanzo:
[HACKING.md#javascript](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-06-30/HACKING.md#javascript) should probably mention `autocomplete2.html` and be updated to link to `web.html` instead of `all.html`, along with a notice about ignoring these errors.

I noticed that “Show more…” often returns less than 10 entities, even though more can be shown.

mwiencek:
> HACKING.md#javascript should probably mention autocomplete2.html

I plan to remove this file once the new component is actually used on our edit pages, so don't intend to keep or document it - the purpose was just to give reviewers somewhere to test it for this PR.

> HACKING.md#javascript should [...] be updated to link to web.html instead of all.html, along with a notice about ignoring these errors.

We can resolve those issues in another PR (I'll look into the canvas one again), but they aren't related to this PR so I'll avoid adding more stuff here. :)

> I noticed that “Show more…” often returns less than 10 entities, even though more can be shown.

Can you give an example search query which demonstrates this, and which search method you're using (direct or indexed)? Can you also try the same query (and search method) in the old autocomplete and see if it behaves the same way? From your description I'm not sure if that's a problem with the component itself, or the queries returning duplicate results (in which case the old autocomplete would behave the same way).

PR #1091, opened 2019-06-02
Bump Flow to 0.100.0

mwiencek:
The main items from the changelog (https://github.com/facebook/flow/releases) that required fixing were:

 * "The statics of function types used to be any but are now typed as an empty object."
 * "Fix an issue where Flow would not catch certain errors involving React function components with unannotated props."
 * "Destructuring patterns could previously include missing properties if the resulting binding was unused. This is now an error even when unused."

### DIFF
--- a/root/components/FormRowRadio.js
+++ b/root/components/FormRowRadio.js
@@ -30,7 +30,6 @@ type Props = {|
 const FormRowRadio = ({
   field,
   label,
-  onChange,
   options,
   required = false,
 }: Props) => (
--- a/root/types.js
+++ b/root/types.js
@@ -547,6 +547,19 @@ declare type IswcT = {|
   +work_id: number,
 |};
 
+declare type KnockoutObservable<T> = {
+  [[call]]: (() => T) & ((T) => empty),
+  peek: () => T,
+  subscribe: ((T) => void) => {dispose: () => empty},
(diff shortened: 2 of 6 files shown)


root/types.js
```diff
@@ -547,6 +547,19 @@ declare type IswcT = {|
   +work_id: number,
 |};
 
+declare type KnockoutObservable<T> = {
+  [[call]]: (() => T) & ((T) => empty),
+  peek: () => T,
+  subscribe: ((T) => void) => empty,
```
mwiencek: This is to fix the issue with function statics being an empty object now instead of `any`. Knockout observables are functions, hence the `[[call]]` type, but they also expose "methods" like peek, subscribe, push, remove, etc.

root/components/FormRowRadio.js
```diff
@@ -30,7 +30,6 @@ type Props = {|
 const FormRowRadio = ({
   field,
   label,
-  onChange,
```
mwiencek: Unused prop (which doesn't appear in the `Props` type, either), which Flow now doesn't like.

root/types.js
mwiencek: I used `empty` to indicate that the return value shouldn't be used in those cases.

PR #1094, opened 2019-06-04
 MBS-10204: Sum accepted/rejected edits in stats

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10204

Depends on https://github.com/metabrainz/musicbrainz-server/pull/1089

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User/Edits.pm
+++ b/lib/MusicBrainz/Server/Controller/User/Edits.pm
@@ -158,6 +158,28 @@ sub autoedits : Chained('/user/load') PathPart('edits/autoedits') RequireAuth Hi
     );
 }
 
+sub applied : Chained('/user/load') PathPart('edits/applied') RequireAuth HiddenOnSlaves {
+    my ($self, $c) = @_;
+    $self->_edits($c, sub {
+        return $c->model('Edit')->find({
+            editor => $c->stash->{user}->id,
+            status => $STATUS_APPLIED,
+        }, shift, shift);
+    });
+    $c->stash(
+        refine_url_args =>
+            { auto_edit_filter => '', order=> 'desc', negation=> 0,
+              combinator=>'and',
+              'conditions.0.field' => 'editor',
+              'conditions.0.operator' => '=',
+              'conditions.0.name' => $c->stash->{user}->name,
+              'conditions.0.args.0' => $c->stash->{user}->id,
+              'conditions.1.field' => 'status',
+              'conditions.1.operator' => '=',
+              'conditions.1.args' => $STATUS_APPLIED },
+    );
+}
+
 sub all : Chained('/user/load') PathPart('edits') RequireAuth HiddenOnSlaves {
     my ($self, $c) = @_;
     $self->_edits($c, sub {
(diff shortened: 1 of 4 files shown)


mwiencek:
I like the idea of displaying these totals, just not 100% sure about the labels, 'cause I always thought of accepted/rejected as voting outcomes rather than specific edit statuses. You excluded cancelled edits from the rejected count, but I'm not sure I'd say an edit that fails due to an error is rejected either (maybe by the system, but not by voters, and it seems weird to fuzz those meanings together). Or that an auto-edit is accepted (rather than applied automatically). Maybe I'm the only one who thinks like that...

"All applied"/"All not applied" (including cancelled in the latter) could work, I guess, but maybe we already use accepted/rejected like this elsewhere in the UI.

reosarevok:
Well, the edit is accepted into the database or rejected from the database, is my line of thought. I don't think "cancelled" fits with either. "Cancelled" has been used by people to specifically avoid having their "failed/rejected" numbers go higher by cancelling the edits before they get rejected, and as such it doesn't feel right to me to add them to the big bad negative number and undo that.

reosarevok:
For an example (I edit too much):

Screenshot from 2019-11-06 22-49-19

reosarevok:
I have had users confused because their numbers weren't almost going up anymore (because most of their edits such as add release became autoedits). 
I'd also be fine with removing "all not applied" and marking only "voted down" in red, I guess, since "all applied" is the general "good" number you care about when seeing a profile and "voted down" is really the main "bad" number you care about (since failed isn't really the user's fault). Would that seem clearer?

yvanzo:
Yes. I would also suggest to make clear the count of applied edits is the sum of the two above, maybe by labelling it “Total applied”?

reosarevok:
Made the changes mentioned, and added a (view all) link / page for applied edits (which I was surprised to see didn't exist yet).
Screenshot from 2019-11-18 22-13-59

lib/MusicBrainz/Server/Controller/User/Edits.pm
```diff
@@ -158,6 +158,28 @@ sub autoedits : Chained('/user/load') PathPart('edits/autoedits') RequireAuth Hi
...
+    $self->_edits($c, sub {
+        return $c->model('Edit')->find({
+            editor => $c->stash->{user}->id,
+            status => $STATUS_APPLIED,
+        }, shift, shift);
+    });
+    $c->stash(
+        refine_url_args =>
+            { auto_edit_filter => '', order=> 'desc', negation=> 0,
+              combinator=>'and',
```
mwiencek: I know this was copied from above but it'd be nice to make the spacing around `=>` consistent and put each property on its own line.

root/user/UserProfile.js
```diff
@@ -341,6 +341,8 @@ const UserProfileStatistics = withCatalystContext(({
 }: UserProfileStatisticsProps) => {
   const voteTotals = votes.pop();
   const encodedName = encodeURIComponent(user.name);
+  const allAppliedCount = Number(editStats.accepted_count) +
+                           Number(editStats.accepted_auto_count);
```
mwiencek: These are defined as `number` in `EditStatsT`, so the type must be wrong (which isn't ideal).

Does it help if we change https://github.com/metabrainz/musicbrainz-server/blob/e1886bf/lib/MusicBrainz/Server/Data/Editor.pm#L582-L582 to do `$count + 0` instead? That should cause the JSON serializer to produce actual numbers, hopefully.

reosarevok: I'm not sure why I was doing this in the first place. I checked, and even looking at a just-created user with 0 edits, everything seems to just work.

reosarevok: I'm not sure why I was doing this in the first place. I checked, and even looking at a just-created user with 0 edits, everything seems to just work.

PR #1096, opened 2019-06-05
MBS-10217: Explain AC renaming in artist/edit_form

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10217

### DIFF
--- a/root/artist/edit_form.tt
+++ b/root/artist/edit_form.tt
@@ -73,7 +73,13 @@
       <fieldset>
         <legend>[% l('Artist Credits') %]</legend>
         <p>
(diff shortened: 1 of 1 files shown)


root/artist/edit_form.tt
```diff
@@ -73,7 +73,13 @@
       <fieldset>
         <legend>[% l('Artist Credits') %]</legend>
         <p>
-            [%- l('Please select the artist credits that you want to rename to follow the new artist name.') -%]
+            [%- l('Please select the {doc|artist credits} that you want to rename to follow the new artist name.', {doc => doc_link('Artist Credits')}) -%]
```
reosarevok: Didn't want to break the existing translation, but it seems like this really should link to what ACs actually *are*.

PR #1097, opened 2019-06-05
MBS-6048: Allow edit notes when submitting ISRCs

reosarevok:
### Implement MBS-6048

Companion: https://github.com/metabrainz/mmd-schema/pull/24

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/Recording.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/Recording.pm
@@ -8,7 +8,7 @@ use MusicBrainz::Server::Constants qw(
     $EDIT_RECORDING_ADD_ISRCS
     $ACCESS_SCOPE_SUBMIT_ISRC
 );
-
+use MusicBrainz::Server::Data::Utils qw( non_empty );
 use MusicBrainz::Server::Validation qw( is_valid_isrc is_guid );
 use MusicBrainz::Server::WebService::XML::XPath;
 use Readonly;
@@ -201,6 +201,8 @@ sub recording_submit : Private
 
     my $xp = MusicBrainz::Server::WebService::XML::XPath->new( xml => $c->request->body );
 
+    my $edit_note = $xp->find('/mb:metadata/mb:edit-note')->string_value;
+
     my (%submit_isrc);
     for my $node ($xp->find('/mb:metadata/mb:recording-list/mb:recording')->get_nodelist)
     {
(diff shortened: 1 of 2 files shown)


phw:
I very much appreciate having the ability to include an edit note for ISRC submission. But I think having this as a query parameter is a bit of a hack.

The clean way would be to include this in the request body. Maybe even per recording. Yes, adding a `edit-note` tag is a change to the XML schema, but it still stays backward compatible for existing clients if it is optional.

jesus2099:
Absolutely not per recording ! As valuable ISRC edits are per release. :)

phw:
You are partially right. Only a single edit note makes sense since it is a single edit. But it is not one edit per release, is one edit per submission.

> Isn’t the client ID also a query parameter?

Yes, but the ient ID is neither really part of the core payload nor is it supposed to contain a longer text. Submitting a potentially long text with line breaks and stuff in the query is a bit dirty.

mwiencek:
Let's change this to fit the edit note into the XML, then. I agree that seems cleaner.

reosarevok:
Changed this to use XML, added a test (the results of which suggested I should be trimming the received note before submission just in case, so I did that too).

t/lib/t/MusicBrainz/Server/Controller/WS/2/SubmitRecording.pm
```diff
@@ -62,6 +62,9 @@ my $content = '<?xml version="1.0" encoding="UTF-8"?>
       </isrc-list>
     </recording>
   </recording-list>
+  <edit-note>
+    Testy test!
+  </edit-note>
```
mwiencek, line 67: Could we add some newlines here to simulate paragraphs, and some XML entity references (e.g. `&quot;`, `&amp;`) to make sure those are preserved/converted properly?

lib/MusicBrainz/Server/Controller/WS/2/Recording.pm
```diff
@@ -199,6 +199,10 @@ sub recording_submit : Private
 
     my $xp = MusicBrainz::Server::WebService::XML::XPath->new( xml => $c->request->body );
 
+    my $edit_note = MusicBrainz::Server::Data::Utils::trim(
```
mwiencek, line 202: Probably just add `trim` to the import list with non_empty
```suggestion
    my $edit_note = trim(
```

lib/MusicBrainz/Server/Controller/WS/2/Recording.pm
```diff
@@ -242,12 +246,22 @@ sub recording_submit : Private
             on_full => sub {
                 my $contents = shift;
                 try {
-                    $c->model('Edit')->create(
+                    my $rem = $c->model('Edit')->create(
```
mwiencek, line 249: (curious, what does `$rem` stand for? 🤔)

reosarevok: :D I think it was meant as "removal" where I copied it from, lol. Will amend to $edit or whatever.

mwiencek (approved):

yvanzo (approved):

PR #1099, opened 2019-06-06
MBS-10219: muziekweb.eu integration

SothoTalKer:
https://tickets.metabrainz.org/browse/MBS-10219


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1540,6 +1540,13 @@ const CLEANUPS = {
       return /^https:\/\/musopen\.org\/music\/\d+\/$/.test(url);
     },
   },
+  'muziekweb': {
+    match: [new RegExp('^(https?://)?www\\.muziekweb\\.eu/Link/', 'i')]
```
yvanzo: Any reason to auto-select relationship type for URLs with path starting with `/Link/` only?
That might be an issue to enter `https://www.muziekweb.eu/en/Link/JK239095/Alientronic`.
```suggestion
    match: [new RegExp('^(https?://)?www\\.muziekweb\\.eu/', 'i')]
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1540,6 +1540,13 @@ const CLEANUPS = {
...
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      return url.replace(/^(?:https?:\/\/)?(?:www\.)?muziekweb\.(eu|nl)\/(en\/)?Link\/([0-9A-Z]+).*$/, 'https://www.muziekweb.eu/Link/$3/');
```
yvanzo: It works but some improvements can be made:
* Skip capturing groups which are not used in the replacement pattern by adding `?:`
* Optionally match any language code in case the website gets more translation languages.
```suggestion
      return url.replace(/^(?:https?:\/\/)?(?:www\.)?muziekweb\.(?:eu|nl)\/(?:[a-z]{2}\/)?Link\/([0-9A-Z]+).*$/, 'https://www.muziekweb.eu/Link/$1/');
```

lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -109,6 +109,7 @@ my %URL_SPECIALIZATIONS = (
     'MusikSammler'        => qr{^https?://(?:www\.)?musik-sammler\.de/}i,
     'Musixmatch'          => qr{^https?://(?:www\.)?musixmatch\.com/}i,
     'Musopen'             => qr{^https?://(?:www\.)?musopen\.org/}i,
+    'Muziekweb'           => qr{^https?://www\.muziekweb\.eu/Link/}i,
```
yvanzo: Matching `/Link/` is probably not necessary here either.
```suggestion
    'Muziekweb'           => qr{^https?://www\.muziekweb\.eu/}i,
```

root/static/scripts/edit/URLCleanup.js
reosarevok: Missing trailing comma (which breaks parsing, so it's a proper bug!)

lib/MusicBrainz/Server/Entity/URL/Muziekweb.pm
```diff
@@ -0,0 +1,22 @@
...
+
+sub sidebar_name { 'Muziekweb' }
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2019 MetaBrainz Foundation
```
yvanzo: You may want to put your own name or pseudo instead, that’s your code after all.
The project just require contributions to be under the same license terms, thanks.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2136,6 +2136,21 @@ const testData = [
...
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://www.muziekweb.eu/Link/M00000052618/',
+       only_valid_entity_types: ['artist'],
+  },
+  {
+                     input_url: 'https://www.muziekweb.eu/en/Link/JK95205/The-slim-shady-LP',
+             input_entity_type: 'release',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://www.muziekweb.eu/Link/JK95205/',
+       only_valid_entity_types: ['release'],
```
yvanzo: You will need a `validate` function for this test to pass.

root/static/scripts/edit/URLCleanup.js
SothoTalKer: This should be fixed now.

lib/MusicBrainz/Server/Entity/URL/Muziekweb.pm
SothoTalKer: Nah, I waive all my personal rights for this code. I 'donate' it to the Metabrainz Foundation.

It's not very much my code imho, as I only changed the name, no creativity at all was needed.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2136,6 +2136,21 @@ const testData = [
...
+  // muziekweb.eu (National Dutch music library)
+  {
+                     input_url: 'https://www.muziekweb.eu/en/Link/M00000052618/POPULAR/Eminem',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://www.muziekweb.eu/Link/M00000052618/',
+       only_valid_entity_types: ['artist'],
+  },
+  {
+                     input_url: 'https://www.muziekweb.eu/en/Link/JK95205/The-slim-shady-LP',
```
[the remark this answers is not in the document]
SothoTalKer: I have added seperate examples for the NL domain

PR #1101, opened 2019-06-07
MBS-10223: Use relationship credits in ArtistRoles

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10223

This makes it so that if a relationship used in ArtistRoles (work writers or event performers) has a relationship credit for the artist, the credit is also displayed in the ArtistRoles display.

The search results for works don't include credits for work relationships, so in search results the credits will be missing until SOLR-110 is fixed.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -711,9 +711,11 @@ sub schema_fixup
                 map {
                     my @relationships = @{ $relationship_map{$_} };
                     {
+                        # TODO: Pass the actual credit when SEARCH-585 is fixed
+                        credit => '',
                         entity => $relationships[0]->entity1,
-                            roles  => [ map { $_->link->type->name } grep { $_->link->type->entity1_type eq 'artist' } @relationships ]
-                        }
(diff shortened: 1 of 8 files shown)


yvanzo:
@reosarevok: If I understand correctly, this patch addresses a much more general issue than the ticket… Can you please update the ticket (bug report?) to reflect that? Screenshot examples would be welcome to illustrate where data was/is displayed, and where displayed data came/comes from. Does it change anything to search results? Should there be a separate MBS ticket linked to SOLR-110? :thinking:  :kissing:

reosarevok:
I don't think it's more general than the ticket, AFAICT this changes literally that "ArtistRoles lists don't use relationship credits" :)

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -711,9 +711,11 @@ sub schema_fixup
                 map {
                     my @relationships = @{ $relationship_map{$_} };
                     {
+                        # TODO: Pass the actual credit when SOLR-110 is fixed
```
yvanzo: FWIW, SOLR-110 moved to SEARCH-585.

reosarevok: Changed the number.

PR #1103, opened 2019-06-13
MBS-8725: Allow mediums to have an unknown tracklist

reosarevok:
https://tickets.metabrainz.org/browse/MBS-8725

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Medium/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Medium/Edit.pm
@@ -173,7 +173,7 @@ sub initialize
             $self->_changes($entity, %opts)
         };
 
-        if ($tracklist) {
+        if ($tracklist && @$tracklist) {
             $self->c->model('Track')->load_for_mediums($entity);
             $self->c->model('ArtistCredit')->load($entity->all_tracks);
 
@@ -193,6 +193,13 @@ sub initialize
                 $data->{old}{tracklist} = $old;
                 $data->{new}{tracklist} = $new;
             }
+        } else {
+            $self->c->model('Track')->load_for_mediums($entity);
+            $self->c->model('ArtistCredit')->load($entity->all_tracks);
+
+            $data->{old}{tracklist} = tracks_to_hash($entity->tracks);
+            $data->{new}{tracklist} = [];
+
         }
 
         MusicBrainz::Server::Edit::Exceptions::NoChanges->throw
--- a/root/release/edit/recordings.tt
+++ b/root/release/edit/recordings.tt
@@ -16,7 +16,14 @@
       <legend data-bind="text: formattedName()"></legend>
 
       <table id="track-recording-assignation" data-bind="if: loaded">
-        <thead>
+        <tbody data-bind="if: tracks().length === 0">
+          <tr>
+            <td></td>
+            <td class="name" style="padding: 1em">[% l('The tracklist for this medium is unknown.') %]</td>
+          <tr>
+        </tbody>
+
+        <thead data-bind="if: tracks().length > 0">
           <tr>
             <th></th>
             <th colspan="2" style="text-align: left">[% l('track') %]</th>
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -264,7 +264,7 @@
     <p>
       <label>
         <input type="checkbox" data-bind="checked: hasUnknownTracklist" />
-        [% l('The tracklist for this release is currently unknown.') %]
+        [% l('I have no information about this release’s media and tracklist.') %]
       </label>
     </p>
   <!-- /ko -->
@@ -352,8 +352,17 @@
     <!-- /ko -->
 
     <div data-bind="visible: loaded() && !collapsed()">
+      <!-- ko if: tracks().length === 0 -->
+      <p style="margin: 1em">
+        <label>
+          <input id="tracks-unknown" type="checkbox" data-bind="checked: tracksUnknownToUser" />
+          [% l('I don’t know the tracklist for this medium.') %]
+        </label>
+      </p>
+      <!-- /ko -->
+
       <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTracks">
-        [%~ l('A tracklist is required.') ~%]
+        [%~ l('Tracks are required. If you don’t know them, but do know some info (such as the format or number of media), tick the “I don’t know the tracklist for this medium” checkbox above and enter the available info. If you have no information about this release’s media, remove all media.') ~%]
       </p>
 
       <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackInfo">
--- a/root/release/index.tt
+++ b/root/release/index.tt
@@ -6,7 +6,7 @@
 
     [% IF release.mediums.size == 0 %]
       <p>
-        [% l('The tracklist for this release is currently unknown.') %]
+        [% l('We have no information about this release’s media and tracklist.') %]
       </p>
     [% ELSE %]
       <button type="button" id="toggle-credits" class="btn-link"></button>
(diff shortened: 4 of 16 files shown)


root/release/edit/tracklist.tt
```diff
@@ -341,8 +341,16 @@
     <!-- /ko -->
 
     <div data-bind="visible: loaded() && !collapsed()">
+
+      <p>
+        <label>
+          <input id="tracks-unknown" type="checkbox" data-bind="checked: tracksUnknownToUser, enable: tracks() == ''" />
```
mwiencek: I think the `enable` binding should be something like `tracks().length === 0` instead? Not sure I understand the string comparison.

lib/MusicBrainz/Server/Data/Medium.pm
```diff
@@ -321,6 +321,11 @@ sub load_related_info {
     $self->c->model('Track')->load_for_mediums(@mediums);
 
     my @tracks = map { $_->all_tracks } @mediums;
+
+    for my $medium (@mediums) {
+        $medium->has_loaded_tracks(1);
```
mwiencek: It might make more sense to move this to `$c->model('Track')->load_for_mediums`, since that method is called in a ton of places throughout the codebase and seems to be what actually loads the tracks.

root/static/scripts/release-editor/fields.js
```diff
@@ -309,6 +309,7 @@ class Medium {
         this.formatUnknownToUser = ko.observable(Boolean(data.id && !data.formatID));
 
         var tracks = data.tracks;
+        this.tracksUnknownToUser = ko.observable(Boolean(data.id && !tracks));
```
mwiencek: Will `tracks` be null/undefined there, or an empty array? 'Cause in JS `![]` is false. (Since these files aren't Flow typed, maybe `data.id && !(tracks && tracks.length)` would be the safest.)

root/static/scripts/release-editor/fields.js
```diff
@@ -366,9 +367,9 @@ class Medium {
             }
         });
 
-        this.needsRecordings = this.tracks.any("needsRecording");
-        this.hasTrackInfo = this.tracks.all("hasNameAndArtist");
-        this.hasVariousArtistTracks = this.tracks.any("hasVariousArtists");
+        this.needsRecordings = !self.tracksUnknownToUser() && this.tracks.any("needsRecording");
```
mwiencek: The problem here and below is that these were observables, but are being changed to either `false` (a literal value that will never change), or the `this.tracks.any` observable as it was (which will never be affected by `tracksUnknownToUser` if it changes), since `x && y` just returns either the first falsey value or the last truthy value.

To clarify, this constructor only runs once when the medium is created, not whenever the observables change. If you want a "live" observable that updates when other observables change, you must use a computed observable:

```JS
this.needsRecordings = ko.computed(function () {
  return !self.tracksUnknownToUser() && self.tracks().any(t => t.needsRecording());
});
```

Note that changing `this.tracks.any("needsRecording")` (which actually returns a computed observable) to calling the builtin `any` directly on the unwrapped array is also important.

root/release/edit/tracklist.tt
```diff
@@ -349,8 +349,16 @@
     <!-- /ko -->
 
     <div data-bind="visible: loaded() && !collapsed()">
+
+      <p>
```
mwiencek, line 353: I feel like this needs `margin: 1em` 'cause it looks smooshed against the left border.

root/release/edit/tracklist.tt
```diff
@@ -349,8 +349,16 @@
...
+        <label>
+          <input id="tracks-unknown" type="checkbox" data-bind="checked: tracksUnknownToUser, enable: tracks().length === 0" />
```
mwiencek, line 355: Would it make sense to hide the checkbox if there are tracks? 'Cause you can't check it unless you remove them all anyway. (If you were able to remove all the existing tracks by checking it, hiding the tracklist editor entirely, that'd make a bit more sense to show it to me, but.)

root/static/scripts/release-editor/fields.js
```diff
@@ -391,12 +392,18 @@ class Medium {
...
+        });
+        this.hasTrackInfo = ko.computed(function () {
+            return self.tracksUnknownToUser() || self.tracks().every(t => t.hasNameAndArtist());
+        });
+        this.hasVariousArtistTracks = ko.computed(function () {
+            return !self.tracksUnknownToUser() && self.tracks().some(t => t.hasVariousArtists());
+        });
         this.confirmedVariousArtists = ko.observable(this.hasVariousArtistTracks());
         this.needsTrackInfo = ko.computed(function () {
-            return !self.hasTrackInfo();
```
mwiencek, line 399: Removing the semi-colon here seems unintentional

root/release/index.tt
```diff
@@ -6,7 +6,7 @@
 
     [% IF release.mediums.size == 0 %]
       <p>
-        [% l('The tracklist for this release is currently unknown.') %]
+        [% l('We have no information about this release’s media and tracklist.') %]
```
mwiencek, line 9: This was reworded to "We have no information about ..."  while the new medium specific string still uses the previous wording ("The tracklist for this medium is unknown"). Should we make those consistent?

root/release/edit/recordings.tt
```diff
@@ -16,7 +16,14 @@
       <legend data-bind="text: formattedName()"></legend>
 
       <table id="track-recording-assignation" data-bind="if: loaded">
-        <thead>
+        <tbody data-bind="if: tracks().length === 0">
+          <tr>
+            <td></td>
+            <td class="name">[% l('The tracklist for this medium is unknown.') %]</td>
```
mwiencek, line 22: This needs padding too, looks smooshed.

lib/MusicBrainz/Server/Data/Medium.pm
```diff
@@ -271,6 +271,7 @@ sub load_related_info {
     $self->c->model('Track')->load_for_mediums(@mediums);
 
     my @tracks = map { $_->all_tracks } @mediums;
+
```
mwiencek, line 274: Nitpick: this is the only change in this file, so seems unnecessary

lib/MusicBrainz/Server/Edit/Medium/Edit.pm
```diff
@@ -192,6 +192,16 @@ sub initialize
...
+        } else {
+            $self->c->model('Track')->load_for_mediums($entity);
+            $self->c->model('ArtistCredit')->load($entity->all_tracks);
+
+            my $old = tracks_to_hash($entity->tracks);
+            my $new = [];
+
+            $data->{old}{tracklist} = $old;
+            $data->{new}{tracklist} = $new;
+
```
mwiencek, line 204:
```suggestion
```

You could probably also assign the `$old` / `$new` values directly (unless you feel the variables make it clearer)

root/release/index.tt
reosarevok: I reworded it to make it more general, so I'm not quite sure how to make it consistent ("The media for this release are unknown" sounds off to me). But open to suggestions :)

root/release/edit/tracklist.tt
reosarevok: That's actually a very good point. Made the change.

root/release/edit/tracklist.tt
```diff
@@ -349,8 +349,17 @@
...
+        <label>
+          <input id="tracks-unknown" type="checkbox" data-bind="checked: tracksUnknownToUser" />
+          [% l('I don’t know the tracklist for this medium.') %]
+        </label>
+      </p>
+      <!-- /ko -->
+
       <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTracks">
-        [%~ l('A tracklist is required.') ~%]
+        [%~ l('Tracks are required. If you don’t know the tracks on this medium, but do know some info (such as the format or number of media), tick the “I don’t know” checkbox above and enter the available info. If you don’t know that either, remove all mediums.') ~%]
```
mwiencek, line 362: I noticed it said to tick the "I don't know" checkbox above, but there's a checkbox labeled *exactly* that for the format which isn't what we mean, so might be less confusing to use the full label. That made the string sound slightly repetitive though, so I changed the first "the tracks on this medium" to "them" to directly follow from the previous sentence. Also replaced "mediums" with "media" at the end since we use media for the plural elsewhere in this PR. But feel free to only apply the parts you think make sense.

```suggestion
        [%~ l('Tracks are required. If you don’t know them, but do know some info (such as the format or number of media), tick the “I don’t know the tracklist for this medium” checkbox above and enter the available info. If you have no information about this release’s media, remove all media.') ~%]
```

mwiencek:
> It'd be nice to have at least two small WS tests (one XML, one JSON) for requesting a release with an empty medium.

If you could look into this too, please do :) There's no coverage for this case at the moment.

mwiencek:
> In the release relationship editor, if the tracklist is unknown I think we should display a message there too instead of an empty table.

I think this wasn't resolved either.

reosarevok:
Added tests, changed the relationship editor display for the medium to match the release overview one.

PR #1106, opened 2019-06-17
MBS-10229: Block more shortened/“smart” URLs

yvanzo:
Do [MBS-10229](https://tickets.metabrainz.org/browse/MBS-10229) block 12 more URL shorteners
----

Throw error message “Please don't use shortened URLs.”

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -212,7 +212,8 @@ export class ExternalLinksEditor
             } else if (!isValidURL(link.url)) {
               error = l('Enter a valid url e.g. "http://google.com/"');
             } else if (isShortened(link.url)) {
-              error = l("Please don't use shortened URLs.");
+              error = l(`Please don’t enter bundled/shortened URLs,
+                         enter the destination URL(s) instead.`);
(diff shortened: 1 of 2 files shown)


yvanzo:
The boundaries between URL shorteners and aggregators are too thin to make a difference, since recent services can do both. So I prefer to keep one list and make a more general error message, see https://github.com/metabrainz/musicbrainz-server/pull/1106/commits/1692f9aabe2119f99aa0df61227a8fb08f6d07cc.

root/static/scripts/edit/externalLinks.js
```diff
@@ -212,7 +212,8 @@ export class ExternalLinksEditor
             } else if (!isValidURL(link.url)) {
               error = l('Enter a valid url e.g. "http://google.com/"');
             } else if (isShortened(link.url)) {
-              error = l("Please don't use shortened URLs.");
+              error = l(`Please don’t enter bundled/shortened URL,
```
mwiencek: Small grammar fix: URL should be plural here.
```suggestion
              error = l(`Please don’t enter bundled/shortened URLs,
```

PR #1107, opened 2019-06-19
MBS-10235: Drop format when reusing mediums

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10235

The medium format being kept when reusing a medium causes a lot of problems (when editors forget to unset it) yet it rarely is of any use: a tracklist being the same doesn't imply the format will be and in fact it's often the opposite, with the same release having CD, vinyl and digital versions, for example. This drops the format when reusing a medium, so that the user will have to either select a format or select the "I don't know" checkbox for it.

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -108,6 +108,7 @@
 
   <!-- ko if: searchResults() && searchResults().length > 0 -->
   <p>[% l('Select a medium from the search results below and then click "Add Medium".') %]</p>
+  <p>[% l('The tracklist for the medium (including the track/recording associations, but not other data such as medium format) will be loaded as a medium on this release.') %]</p>
   <!-- /ko -->
 
   <!-- ko if: totalPages() > 1 -->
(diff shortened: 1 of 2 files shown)


reosarevok:
Where would you add these instructions? The release duplicates tab? (and does it really matter? the user will anyway see what gets copied soon after :) )

yvanzo:
> Where would you add these instructions? The release duplicates tab?

On the line 93 of `tracklist.tt` I guess?

> (and does it really matter? the user will anyway see what gets copied soon after :) )

We could also relabel the button “Add Medium” as “Button” and let user see what happens ;)

reosarevok:
Ok, added something, how does it look?

root/release/edit/tracklist.tt
```diff
@@ -108,6 +108,7 @@
 
   <!-- ko if: searchResults() && searchResults().length > 0 -->
   <p>[% l('Select a medium from the search results below and then click "Add Medium".') %]</p>
+  <p>[% l('The tracklist for the medium (including the track/recording associations) will be loaded as a medium on this release.') %]</p>
```
yvanzo: It should be made more clear that other medium’s attributes won’t be copied:
```suggestion
  <p>[% l('Only the tracklist for the medium (including the track/recording associations) will be loaded as a medium on this release.') %]</p>
```

reosarevok: Modified the comment somewhat, let me know how it sounds.

yvanzo: SGTM

PR #1108, opened 2019-06-19
MBS-975: 301 redirect from track/ to recording/

reosarevok:
https://tickets.metabrainz.org/browse/MBS-975

When we redirect an (old, pre-NGS) track link to a recording link, we should 301, not 303, since this is not a "see other" situation (such as with tracks in release tracklists) but a "this track permanently turned into this recording" situation. 

This was decided in a dev meeting in 2014, but never implemented (see jira comments)

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Track.pm
+++ b/lib/MusicBrainz/Server/Controller/Track.pm
@@ -37,6 +37,8 @@ sub show : Chained('load') PathPart('')
     if (defined ($c->stash->{recording}))
     {
         $uri = $c->uri_for_action('/recording/show', [ $c->stash->{recording}->gid ]);
+        # The track link is now a recording link: this should be considered a permanent move
+        $c->response->redirect($uri, 301);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Controller/Track.pm
```diff
@@ -37,6 +37,8 @@ sub show : Chained('load') PathPart('')
     if (defined ($c->stash->{recording}))
     {
         $uri = $c->uri_for_action('/recording/show', [ $c->stash->{recording}->gid ]);
+        # The track link is now a recording link: this should be considered a permanent move
```
yvanzo: This is unclear to me, can you please provide an example URL?

reosarevok: https://musicbrainz.org/track/56b85b37-d64b-4be6-ae3d-04d730b6dfcf

yvanzo: Thanks, I thought it was about some track links, it is just about any track link.

PR #1109, opened 2019-06-19
 MBS-10228 /  MBS-10243 /  MBS-10244: Various fixes for no-media releases

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10228
https://tickets.metabrainz.org/browse/MBS-10243
https://tickets.metabrainz.org/browse/MBS-10244

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Release/Merge.pm
+++ b/lib/MusicBrainz/Server/Edit/Release/Merge.pm
@@ -219,6 +219,11 @@ override build_display_data => sub
                 mediums => $_->{mediums}
             }, @{ $self->data->{medium_changes} }
         ];
+        $data->{empty_releases} = [
+            map +{
+                release => $loaded->{Release}{ $_->{id} } // Release->new(name => $_->{name}),
+            }, grep { defined $_->{mediums} && scalar @{ $_->{mediums} } == 0 } @{ $self->data->{old_entities} }
+        ];
     } elsif ($self->data->{merge_strategy} == $MusicBrainz::Server::Data::Release::MERGE_MERGE) {
         my $recording_merges = [];
         if ($self->data->{recording_merges}) {
--- a/root/edit/details/merge_releases.tt
+++ b/root/edit/details/merge_releases.tt
@@ -2,8 +2,8 @@
   <tr>
     <td>[% link_entity(release) %]</td>
     <td>[% artist_credit(release.artist_credit) %]</td>
-    <td>[% html_escape(release.combined_format_name) %]</td>
-    <td>[% release.combined_track_count %]</td>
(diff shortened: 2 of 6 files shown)


lib/MusicBrainz/Server/Edit/Release/Merge.pm
```diff
@@ -219,6 +219,11 @@ override build_display_data => sub
                 mediums => $_->{mediums}
             }, @{ $self->data->{medium_changes} }
         ];
+        $data->{empty_releases} = [
+            map +{
+                release => $loaded->{Release}{ $_->{id} }
```
mwiencek: I'd add at least
```suggestion
                release => $loaded->{Release}{ $_->{id} } // Release->new(name => $_->{name}),
```
here so that it displays the old name rather than `[removed]` once the merge succeeds. (The edit data contains more information that just the name we could display, though we'd have to build all the Moose entities here ourselves, and it already only uses the name for the non-empty releases, so I guess it's more of a future improvement.)

root/edit/details/merge_releases.tt
```diff
@@ -2,8 +2,8 @@
   <tr>
     <td>[% link_entity(release) %]</td>
     <td>[% artist_credit(release.artist_credit) %]</td>
-    <td>[% html_escape(release.combined_format_name) %]</td>
-    <td>[% release.combined_track_count %]</td>
+    <td>[% html_escape(release.combined_format_name) || l('[missing media]') %]</td>
+    <td>[% release.combined_track_count || l('-') %]</td>
```
mwiencek: In #1242 I used `lp('-', 'missing data')` in some places instead. Maybe we can do the same here and below?

lib/MusicBrainz/Server/Edit/Release/Merge.pm
```diff
@@ -219,6 +219,11 @@ override build_display_data => sub
                 mediums => $_->{mediums}
             }, @{ $self->data->{medium_changes} }
         ];
+        $data->{empty_releases} = [
+            map +{
+                release => $loaded->{Release}{ $_->{id} } // Release->new(name => $_->{name}),
+            }, grep { scalar @{ $_->{mediums} } == 0 } @{ $self->data->{old_entities} }
```
mwiencek: The `mediums` array was added in https://github.com/metabrainz/musicbrainz-server/commit/b630292550f7b44af64e48a4ec27ea7922b749a7 (marked as `Optional`), so not all edits will have it.

`@{ $_->{mediums} }` will throw an error where `$_->{mediums}` is undef, so we'll have to check that it's defined first:
```suggestion
            }, grep { defined $_->{mediums} && scalar @{ $_->{mediums} } == 0 } @{ $self->data->{old_entities} }
```

PR #1111, opened 2019-06-20
MBS-1736: Block setting format on too early releases

reosarevok:
https://tickets.metabrainz.org/browse/MBS-1736

Look at the earliest release year of all release events. If the release year is earlier than the first year of use stored for the format, block the user from selecting that format for any mediums.

This currently has the issue that we don't have a format year for, for example. 7" vinyl, only for vinyl as a whole. Since I don't think we use year externally anyway and it's just used for things like this, I'd suggest we can just use the same year for all vinyls (and the CD year for specific types of CD if we can't find the year), to at least block *some* incorrect usage. No idea what would be a sensible year for digital media though! :)


root/static/scripts/release-editor/fields.js
```diff
@@ -800,6 +804,15 @@ class Release extends MB_entity.Release {
             });
         });
 
+        this.earliestYear = ko.computed(function () {
+            let years = [];
+            _(self.events()).each(function (event) {
```
mwiencek: Instead of lodash I'd prefer using the native `.forEach` array method here, or better `.map`:

```JavaScript
return Math.min(
  ...self.events().map(e => e.unwrapDate().year),
);
```

root/static/scripts/release-editor/fields.js
```diff
@@ -394,6 +394,10 @@ class Medium {
             return !self.canHaveDiscID() && (self.hasExistingTocs() || hasPregap() || hasDataTracks());
         });
 
+        this.hasTooEarlyFormat = ko.computed(function () {
+            return !!self.formatID() && !!self.release && !!self.release.earliestYear() && (self.release.earliestYear() < MB.formatDates[self.formatID()]);
```
mwiencek: To fix `self.release` being undefined, you may just need to pass `release` to the `fields.Medium` constructor in whichever test was failing. Here it might be better to remove the check and have it fail immediately, since the computed observable won't work correctly if it only has a subscription on `formatID`.

lib/MusicBrainz/Server/Controller/ReleaseEditor.pm
```diff
@@ -43,6 +43,7 @@ sub _init_release_editor
 
     my @medium_formats = $c->model('MediumFormat')->get_all;
     my $discid_formats = [ grep { $_ } map { $_->has_discids ? ($_->id) : () } @medium_formats ];
+    my %format_dates = map { $_->id => $_->year } @medium_formats;
```
mwiencek: Would it make sense to fall back to the parent/root format's year if a specific one isn't set?

root/release/edit/tracklist.tt
```diff
@@ -309,6 +309,9 @@
           <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: needsFormat">
             [%~ l('A format is required. If you don’t know it, tick the “I don’t know” checkbox next to the format dropdown.') %]
           </p>
+          <p class="error field-error" style="padding-left: 1em" data-bind="showErrorRightAway: hasTooEarlyFormat">
+            [%~ l('This medium is set to a format that did not exist on the selected release date for the release.') %]
```
mwiencek: Just to make it sound slightly less wordy, I think "release date" implies it's for the release well enough

```suggestion
            [%~ l('This medium is set to a format that did not exist on the selected release date.') %]
```

root/static/scripts/release-editor/fields.js
```diff
@@ -424,6 +424,10 @@ class Medium {
             return !self.canHaveDiscID() && (self.hasExistingTocs() || hasPregap() || hasDataTracks());
         });
 
+        this.hasTooEarlyFormat = ko.computed(function () {
+            return !!self.formatID() && !!self.release.earliestYear() && (self.release.earliestYear() < MB.formatDates[self.formatID()]);
```
mwiencek: Assigning `MB.formatDates[self.formatID()]` to something and checking if it's undefined before doing the comparison would be ideal - although `any_number < undefined` will be false, Flow won't like that sort of thing once we add types.

root/static/scripts/release-editor/fields.js
```diff
@@ -876,6 +880,12 @@ class Release extends mbEntity.Release {
                 const isDuplicate = _.filter(events, nonEmptyEvent).length > 1;
                 events.forEach(e => e.isDuplicate(isDuplicate));
             });
+        }); 
+
+        this.earliestYear = ko.computed(function () {
+            return Math.min(
+            ...self.events().map(e => e.unwrapDate().year),
```
mwiencek: This should be indented more (or perhaps it can all fit in the previous line)
```suggestion
                ...self.events().map(e => e.unwrapDate().year),
```

root/static/scripts/tests/release-editor/common.js
```diff
@@ -26,6 +26,7 @@ export function setupReleaseAdd(data) {
 export function setupReleaseEdit() {
   releaseEditor.action = "edit";
   var release = new fields.Release(testRelease);
+  //release.mediums()[0].release = testRelease;
```
mwiencek:
```suggestion
```

lib/MusicBrainz/Server/Controller/ReleaseEditor.pm
reosarevok: I'm not sure I'd dare to block by inheritance - I'm sure there'll be some obscure case where that doesn't work. Honestly, the main use case for this is just blocking CDs before the 1980s, so either way would work.

mwiencek: It seems weird that the error basically never triggers for any of the child formats even when it obviously should, but they're probably rare enough that it doesn't matter.

root/static/scripts/release-editor/fields.js
```diff
@@ -424,6 +424,11 @@ class Medium {
             return !self.canHaveDiscID() && (self.hasExistingTocs() || hasPregap() || hasDataTracks());
         });
 
+        this.hasTooEarlyFormat = ko.computed(function () {
+            const formatDate = MB.formatDates[self.formatID()];
+            return !!self.formatID() && !!self.release.earliestYear() && formatDate && (self.release.earliestYear() < formatDate);
```
mwiencek, line 429: You can avoid checking `self.formatID()` there again since `formatDate` will definitely be undefined if the format was, and perhaps wrap the whole thing in `!!` for simplicity (one of the terms, `formatDate`, was missing a `!!`). `showErrorRightAway` works fine if the return value isn't boolean anyway, but.

```suggestion
            const formatDate = MB.formatDates[self.formatID()];
            return !!(formatDate && self.release.earliestYear() && self.release.earliestYear() < formatDate);
```

PR #1112, opened 2019-06-21
MBS-1870 / MBS-8477 / MBS-10690 / MBS-10691: User tag page improvements

reosarevok:
https://tickets.metabrainz.org/browse/MBS-1870
https://tickets.metabrainz.org/browse/MBS-8477

I know, I know, these should be two (or three) different commits, but I worked on both things at once while converting to React and then it was quite hard to just split them in chunks. Hopefully it's not too hard to review as it is, but I can try to change it if needed (it'll just take a fair amount of time).


yvanzo:
I just created ticket [MBS-10341](https://tickets.metabrainz.org/browse/MBS-10341) “Convert user tag entity list page to React” which actually covers the main changes of this PR (if I skimmed through it correctly). It seems impossible to split MBS-10341 from [MBS-1870](https://tickets.metabrainz.org/browse/MBS-1870) as component is paginated. But [MBS-8477](https://tickets.metabrainz.org/browse/MBS-8477) should be easy to commit separately.

reosarevok:
@yvanzo: While I agree that ideally I should have split those two, they touch the same files and interact a lot so I really don't feel the (fairly large) amount of time that would go into splitting those would be worth it to be honest. I'll try to avoid overloading commits in the future (well, I'm already trying, but this will help keep it in mind).

yvanzo:
Ok but please add a reference to MBS-10341. (That won't make the subject line go over 72 chars :stuck_out_tongue:)

reosarevok:
For some test of how this works (better to use the RO Pink DB rather than the sample): /user/reosarevok/tag/to%20clean%20up (per-entity shorter lists + pagination), /user/reosarevok/tag/classical (upvoted vs downvoted).

root/user/components/DownvotedSwitch.js
```diff
@@ -0,0 +1,41 @@
...
+        {upvotes_link|Show their votes for the tag instead}.`,
+        {upvotes_link: uriWith($c.req.uri, {show_downvoted: 0})},
+      )
+    ) : (
+      exp.l(
+        `Showing entities where the user voted for this tag.
+        {downvotes_link|Show their votes against the tag instead}.`,
+        {downvotes_link: uriWith($c.req.uri, {show_downvoted: 1})},
+      )
+    )}
```
mwiencek: I wonder if we should change the page headings from just `Tag “foo”` to `Votes for tag “foo”` and `Votes against tag “foo”` depending on what's being viewed, instead of explaining it down here. It'd then be nice to make the link text a bit more descriptive here by including the user/tag in the link text, e.g. `"Show {upvotes_link|entities {user} has tagged “{tag}”}.` and `"Show {downvotes_link|entities where {user} has voted against “{tag}”}.`. What do you think? :)

root/tag/utility/buildTagListSection.js
```diff
@@ -0,0 +1,85 @@
...
+
+import * as React from 'react';
+
+import ENTITIES from '../../../entities';
+import DescriptiveLink
+  from '../../static/scripts/common/components/DescriptiveLink';
+import expand2text from '../../static/scripts/common/i18n/expand2text';
+
+
+type Props = {
```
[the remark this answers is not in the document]
reosarevok: It's not quite the same - this has optional booleans (because they don't exist for TagIndex) but UserTag has mandatory ones. Might not be worth the distinction though, I dunno.

root/user/components/DownvotedSwitch.js
reosarevok: Seems sensible. Done!

yvanzo:
> Hopefully it's not too hard to review as it is

:rofl:

root/components/TagList.js
```diff
@@ -0,0 +1,142 @@
...
+      * The below use N_ln so languages with non-Germanic pluralization
+      * rules (i.e., any that make number distinctions above the
+      * threshold where we'll actually show the string) can translate
+      * properly. However, the strings are the same in English because
+      * we do not make a distinction other than for 1, which will never
+      * show in this case.
+    */}
+    {buildTagListSection(props, 'area', l('Areas'), N_ln(
+      'See all {num} areas',
+      'See all {num} areas',
```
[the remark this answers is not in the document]
reosarevok: These aren't new strings, should I still make a change?

reosarevok:
Added two more commits here, one converting the list of tags by a user to React, another allowing seeing downvoted tags (which wasn't a new issue, FWIW - that was always filtered by is_upvote - but I agree is useful!).

Adding pagination to the list of tags would require very significant changes to how it works, I think - right now it makes one query and then splits it into tags and genres, but that wouldn't work with pagination. We'd need to paginate each list separately, probably requiring moving to two queries. I'd think the number of users who could benefit from pagination on that page is relatively small, so I wouldn't bother for now and would wait until an actual user asks :)

root/components/TagList.js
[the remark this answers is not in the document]
yvanzo: Oops, I missed that, never mind.

root/user/UserTagEntity.js
```diff
@@ -0,0 +1,68 @@
...
+        entityType={entityType}
+        pager={pager}
+      />
+    ) : (
+      <UserHasNotUsedTag
+        entityType={entityType}
+        showDownvoted={showDownvoted}
+        tag={tag}
+        user={user}
+      />
```
mwiencek, line 63: Is saying "user has not tagged anything with foo" or "user has not voted against any foo tags" here when they haven't with this particular entity type but *have* with others a bit misleading? At least it appears under e.g. "0 artists found" so it has some context, but dunno if that applies as much for other translations. It might be clearer with only the heading, actually ("0 artists found") and no explanatory text.

root/components/TagList.js
```diff
@@ -0,0 +1,142 @@
...
+        {tags.count > tags.tags.length ? (
+          <li key="see-all">
+            <em>
+              {user ? (
+                <a
+                  href={'/user/' + encodeURIComponent(user.name) + '/tag/' +
+                        encodeURIComponent(props.tag.name) + '/' + url}
+                >
+                  {expand2text(seeAllMessage(tags.count), {num: tags.count})}
+                </a>
```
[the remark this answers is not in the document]
reosarevok: Did that on the commit that introduces UserTagLink (also did the same for TagLink below)

root/user/UserTagEntity.js
reosarevok: I think the explanatory text makes sense, but I made it so that it changes based on entity.

root/user/components/DownvotedSwitch.js
```diff
@@ -0,0 +1,87 @@
...
+const DownvotedSwitch = ({
+  $c,
+  showDownvoted,
+  tag,
+  user,
+}: DownvotedSwitchProps) => (
+  <p>
+    {showDownvoted ? (
+      exp.l(
+        `Show {upvotes_link|entities {user}
```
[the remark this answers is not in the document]
reosarevok: I made it use the same text as AllDownvotedSwitch: "Showing tags {user} <strong>has voted against</strong>. Show {upvotes_link|entities {user} has tagged with “{tag}”} instead." and whatnot. I think that makes the order seem sensible again :)

reosarevok:
Rebased this, fixed a couple small issues I found when testing. Tested by visiting http://localhost:5000/user/reosarevok/tags, http://localhost:5000/user/reosarevok/tag/classical (for up/downvoted switch) and http://localhost:5000/user/reosarevok/tag/to%20clean%20up (for many tagged entities collapsing into a list and see more link, plus per-entity-type lists).

root/user/components/UserHasNotUsedTag.js
```diff
@@ -0,0 +1,106 @@
...
+  label: N_l('{user} has not tagged any labels with “{tag}”.'),
+  place: N_l('{user} has not tagged any places with “{tag}”.'),
+  recording: N_l('{user} has not tagged any recordings with “{tag}”.'),
+  release: N_l('{user} has not tagged any releases with “{tag}”.'),
+  release_group: N_l(
+    '{user} has not tagged any release groups with “{tag}”.',
+  ),
+  series: N_l('{user} has not tagged any series with “{tag}”.'),
+  work: N_l('{user} has not tagged any works with “{tag}”.'),
+};
```
yvanzo, line 60: Please provide a URL example using these strings, if any.

yvanzo (changes requested):
Please write user-friendly descriptions of linked tickets so they can find affected pages for them.

Additionally tested with:
* http://localhost:5000/user/reosarevok/tags?show_downvoted=1
* http://localhost:5000/user/reosarevok/tag/classical?show_downvoted=1
* http://localhost:5000/user/reosarevok/tag/to%20clean%20up?show_downvoted=1
* http://localhost:5000/user/reosarevok/tag/calypso (for tag not used by an editor)
* http://localhost:5000/user/reosarevok/tag/calypso?show_downvoted=1 (for tag not voted against by an editor)

I did not find any use of entity type-specific messages yet. Are they really used anywhere?

root/user/components/UserHasNotUsedTag.js
reosarevok: http://localhost:5000/user/reosarevok/tag/classical/release /

reosarevok:
Updated the tickets a bit with some examples - do check if you think they look good! :)

root/user/components/UserHasNotUsedTag.js
yvanzo: Thanks, how do you reach these pages?

reosarevok: I edited the comment in the meantime to specify that, sorry! :)

yvanzo: Found it: You click on “See all {num} works” under the short list of 10 first works available from https://musicbrainz.org/user/jesus2099/tag/instrumental

So there seems to be two issues:
* Navigation: There should be a way to go back to `/tag/instrumental` from `/tag/instrumental/work`.
* Description: Why are these strings entity type-specific and neither the header `Vote for/against “{tag}”` nor its following paragraph? If one of these strings should be entity type-specific that should probably be the header, other strings can probably be entity type-neutral, like what we have for subscriptions already.

reosarevok: Both should be, because just saying "jesus2099 has not voted against any “instrumental” tags" would be a lie :) But that's fair, other lines here should also be specific, at least the "Showing entities where" bit.

mwiencek:
I reviewed the recent additions here and decided to play around with some UI changes, because I wasn't too happy with the number of strings being added. The UI also isn't very easy to use despite all these strings being added. I'm no designer, but can think of a few reasons...

  * I have to read several complete sentences to decide what I want to do or where I want to navigate. (As opposed to having a breadcrumb or simple dropdown.)
  * The strings repeat each other, or are too specific without adding any new information. I think the "no results" page illustrates that best: 
    Example of repetitive strings on the no results page
    The header and first sentence here basically repeat each other. So do the last two sentences. I'm not sure we need a string like "x has not tagged any areas with y" when "0 areas found" suffices. It's obvious from the context of the page what "0 areas found" means: that we found 0 areas the user tagged with y.
  * The terminology feels inconsistent. Across all of these pages, I see: "tagged," "uses of tag," "has used," "voted for," "voted against."

Since I don't want you to have to rework this PR again, I pushed a "WIP" commit here that I hope we could work off of. It adds a "vote filter" select, breadcrumbs, and consolidates a lot of the strings. It's not perfect, but closer to what I think things should be. Can you have a look and see if the changes make sense?

Example of the new no results page
Example of the user tagged entities overview page

reosarevok:
Having filter *and* an "Upvoted tags" link such as below is very confusing IMO.
Screenshot from 2021-03-10 12-54-01

I understand it's meant as a breadcrumb for "All upvoted tags", but maybe that could be shown either on top or as part of the header? The other page we use breadcrumbs on that I can think of is admin/attributes (https://musicbrainz.org/admin/attributes/WorkAttributeType for example). There we use the header, but I guess the main point of a separate header here is that it allows us to link to the tag. So maybe on top is better.

I'd also consider using "Show:" instead of "Vote filter:", and ideally having that under the heading.

For an example: 
Screenshot from 2021-03-10 13-13-45

yvanzo (approved):
Tested with the above URLs again. The breadcrumb improves navigation a lot. Plus it skips adding too many strings! :rocket:

PR #1114, opened 2019-06-24
MBS-10256: Unescape description HTML in rels editor

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10256

React's renderToStaticMarkup seems to be sanitizing the HTML in a way that means what Knockout actually gets is not directly usable in a HTML binding. So I'm parsing it back into HTML before displaying it.


reosarevok:
Oh, duh. Made the change!

PR #1115, opened 2019-06-25
MBS-1921: Display edit link under annotations

reosarevok:
https://tickets.metabrainz.org/browse/MBS-1921

This adds an extra "Edit annotation" link under annotations.
This link will make it clearer to users that they can edit the current annotation, which isn't always that clear at the moment since the only edit link is on the sidebar, very far away from the text.

Because the "View annotation history" link is already there sometimes, I'm separating them with a slash when that's the case. Seems cleaner than using brackets or periods.

### DIFF
--- a/root/static/scripts/common/components/Annotation.js
+++ b/root/static/scripts/common/components/Annotation.js
@@ -92,14 +92,18 @@ const Annotation = ({
                 date: formatUserDate($c, annotation.creation_date),
                 user: <EditorLink editor={annotation.editor} />,
               })}
+              {' '}
               {numberOfRevisions && numberOfRevisions > 1 ? (
                 <>
-                  {' '}
                   <a href={entityHref(entity, '/annotations')}>
                     {l('View annotation history')}
                   </a>
+                  {' | '}
                 </>
               ) : null}
+              <a href={entityHref(entity, 'edit_annotation')}>
+                {l('Edit annotation')}
+              </a>
             </>
           ) : (
             exp.l(
--- a/root/static/scripts/common/components/Annotation.js
+++ b/root/static/scripts/common/components/Annotation.js
@@ -29,7 +29,7 @@ type Props = {
   +annotation: ?$ReadOnly<{
     ...AnnotationT,
     +editor: EditorT | SanitizedEditorT | null,
-    ...
+    ...,
   }>,
   +collapse?: boolean,
   +entity: AnnotatedEntityT | MinimalAnnotatedEntityT,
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/components/Annotation.js
```diff
@@ -82,14 +84,18 @@ const Annotation = ({
...
                 user: <EditorLink editor={annotation.editor} />,
               })}
+              {' '}
               {numberOfRevisions && numberOfRevisions > 1 ? (
                 <>
-                  {' '}
                   <a href={entityHref(entity, '/annotations')}>
                     {l('View annotation history')}
                   </a>
+                  {' / '}
```
mwiencek: It appears we typically use ` | ` for separating edit links though, e.g. see `git grep -F "{' | '}" -- '*.js'`.

root/static/scripts/common/components/Annotation.js
```diff
@@ -82,14 +84,18 @@ const Annotation = ({
...
                 </>
               ) : null}
+              <a href={returnUri($c, entityHref(entity, 'edit_annotation'), 'returnto', $c.req.uri)}>
```
mwiencek: Where does it take you without setting `returnto`?

reosarevok: Seems to go back to the same spot - should I just remove the returnto then?

mwiencek: If the Annotation component is only used on the index pages (which I think is the case), I'd just remove it, since that's where it'll return by default.

root/static/scripts/common/components/Annotation.js
```diff
@@ -92,14 +94,18 @@ const Annotation = ({
...
               {numberOfRevisions && numberOfRevisions > 1 ? (
                 <>
-                  {' '}
                   <a href={entityHref(entity, '/annotations')}>
                     {l('View annotation history')}
                   </a>
+                  {' | '}
                 </>
               ) : null}
+              <a href={returnUri($c, entityHref(entity, 'edit_annotation'))}>
```
mwiencek: Sorry for the confusion, but with the previous review comment I was more wondering if using `returnUri` is necessary at all here. I think just specifying  `href={entityHref(entity, 'edit_annotation')}` here is enough, unless I'm missing something.

reosarevok: You were right, it's unneeded. Got rid!

root/types.js
```diff
@@ -926,9 +926,8 @@ declare type SanitizedCatalystContextT = {
   +action: {
     +name: string,
   },
-  +req: {
```
mwiencek: All the changes outside of root/static/scripts/common/components/Annotation.js should probably be removed now BTW.

PR #1117, opened 2019-06-25
MBS-2176: Add "Finish" button to go to last tab

reosarevok:
https://tickets.metabrainz.org/browse/MBS-2176

This adds a "Finish" button that goes straight to the Edit Note tab, to make it more clear that a user can go to the Edit Note tab and enter the edit from any tab, without having to go through every single one.


reosarevok:
What about showing it everywhere except the first tab for that case?

reosarevok:
Changed the condition to `(activeTabIndex() > 0 || action === 'edit') && (activeTabIndex() < tabCount - 1)` (so it will not be shown on Information if this is a release add)

PR #1119, opened 2019-06-26
MBS-2663: Search edits by non-subscribed entity

reosarevok:
https://tickets.metabrainz.org/browse/MBS-2663

I don't quite understand how the whole edit search code works yet, so let me know if something feels weird - but I tested it and it seems to work fine.


mwiencek:
I think `'not_subscribed:' 0` also needs adding to `cardinalityMap.subscription` in root/static/scripts/common/MB/edit_search.js, but looks good otherwise. (The cardinality is how many parameters the operator takes, and the JS uses that to show/hide the appropriate number of fields, so if it works without that I guess it's by accident.)

PR #1122, opened 2019-06-28
Remove moment, use Intl.DateTimeFormat (fixes MBS-10264)

mwiencek:
Fixes [MBS-10264](https://tickets.metabrainz.org/browse/MBS-10264)

moment is a rather bloated library which we barely make use of. And in fact the places we do use it could easily be using the native `Date` and `Intl.DateTimeFormat` APIs instead.

(To illustrate how large it is, with `DEVELOPMENT_SERVER` enabled, i.e. without minification, the size of root/static/build/common-chunks.js is reduced on my system from 5.0M to 4.3M, just by removing moment and its extensions.)

`Intl.DateTimeFormat` works perfectly fine on the server and in modern browsers. IE <= 11, Edge <= 18, and Safari <= 10 will have issues where we use `formatUserDate` in the browser. We fall back to using `Date.prototype.toString()` in those rare cases, which I think is sufficient.

Note: `formatUserDate` had to be changed to accept `$c` rather than `$c.user` as the first argument, as it now needs access to the current server language (in the same way that the TT `UserDate` plugin does).


root/report/AnnotationsArtists.js
```diff
@@ -40,7 +40,7 @@ const AnnotationsArtists = ({
       </li>
       <li>
         {texp.l('Generated on {date}',
-                {date: formatUserDate($c.user, generated)})}
+                {date: formatUserDate($c, generated)})}
       </li>
```
reosarevok: This kinda seems like it could be a shared component.

root/static/scripts/account/components/PreferencesForm.js
```diff
@@ -62,12 +62,11 @@ const allowedDateTimeFormats = [
   '%m.%d.%Y %H:%M',
 ];
 
-function buildDateTimeFormatOptions(timezone) {
-  const hereAndNow = moment.tz(timezone);
+function buildDateTimeFormatOptions($c, hereAndNow, timezone) {
```
reosarevok: Is there a reason why we can't calculate hereAndNow inside the function?

root/utility/formatUserDate.js
```diff
@@ -6,35 +6,152 @@
...
+try {
+  const now = new Date();
+  formatDateUsingPattern(now, '%a', 'America/Chicago', 'en-US');
+  formatDateUsingPattern(now, '%c', 'America/Chicago', 'en-US');
+} catch (e) {
+  HAS_INTL_DATETIMEFORMAT = false;
+}
+
+function formatDateUsingPattern(
+  date: Date,
```
reosarevok: Minor af, but shouldn't this have a space before the paren? Or is it different if all arguments are on the next lines?

root/utility/parseIsoDate.js
```diff
@@ -0,0 +1,46 @@
...
+  const match = isoDate.match(isoFormat);
+
+  if (!match) {
+    return null;
+  }
+
+  const [year, month, day, hour, minute, second] =
+    match.slice(1, 7).map(parseIntBase10);
+
+  return new Date(Date.UTC(year, month - 1, day, hour, minute, second));
```
reosarevok: Why - 1? Does month start at 0?

root/report/AnnotationsArtists.js
mwiencek: Agreed, not sure I want to change it all in this PR though :)

root/utility/parseIsoDate.js
mwiencek: Added a comment

root/utility/formatUserDate.js
mwiencek: There's a space before the opening parenthese for anonymous functions (`function ()`), but only because there's always a space after `function`, so it's consistent in both cases. There's never a space after the function name, though (compare `git grep 'function [0-9A-z] \+('` with `git grep 'function [0-9A-z]\+('`).

root/static/scripts/account/components/PreferencesForm.js
mwiencek: Don't think so - can't remember why I moved that, so reverted.

root/report/AnnotationsArtists.js
reosarevok: fI'll send a second PR on top of this for it :)

mwiencek:
Rebased & fixed conflicts again.

PR #1123, opened 2019-06-28
MBS-10137: Fix guess feat. on recording pages

mwiencek:
Fix [MBS-10137](https://tickets.metabrainz.org/browse/MBS-10137): Guess feat button doesn't change AC on recording pages
----

The `guessFeat` function expects an entity's `artistCredit` to be an observable, which was previously only true on release pages. This is now made true on the other entity forms.

We also need a way to update the `ArtistCreditEditor` component when the observable changes. This is done by storing the component instance on the table node, and later calling `setState` on it. (Unfortunately, we can't just render `FormRowArtistCredit` again using a different key, because `ArtistCreditEditor` does some extremely bad things with `ArtistCreditBubble` that leave behind stale state. These components should hopefully be rewritten and changed to be fully controlled in the future.)

### DIFF
--- a/root/static/scripts/edit/components/ArtistCreditEditor.js
+++ b/root/static/scripts/edit/components/ArtistCreditEditor.js
@@ -360,7 +360,15 @@ class ArtistCreditEditor extends React.Component {
 
     return (
       <>
-        <table key="artist-credit-editor" className="artist-credit-editor">
+        <table
+          key="artist-credit-editor"
+          className="artist-credit-editor"
+          ref={node => {
+            if (node) {
+              $(node).data('componentInst', this);
+            }
+          }}
+        >
           <tbody>
             <tr>
               <td>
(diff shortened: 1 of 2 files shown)


mwiencek:
The bug is that the button doesn't work at all on recording pages, so there's not much to reproduce besides "see that nothing happens." :) The error in the ticket gets logged to the console. You have to test it on a recording that has "feat." in the title, of course, otherwise the button would do nothing regardless.

mwiencek:
What gets dropped exactly? I tried it on /recording/be3d34a4-8541-4d71-bd35-c25b5b2c095b/edit and it adds `; Cecil Aronowitz` to the AC. (I guess it should be changed to use a comma instead of a semicolon per the the current CSG. Or do you mean the "feat. viola" is dropped and should be used in the join phrase?)

reosarevok:
I tried it there and did nothing, but I assumed it was because the artist was already on the AC. But I copied the same to a random recording title, pressed guess, and it got removed but not added to the AC.

mwiencek:
It sounds like the JS wasn't recompiled (or it's using the old JS), since that's what happens without the patch applied.

reosarevok:
Yeah, but it worked fine with a normal feat! Will test again, maybe I messed something up :)

PR #1124, opened 2019-06-28
MBS-10263: Don't insert annotation if edit is open

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10263


mwiencek:
A downside of this is the edit data no longer stores the annotation IDs. I think we should update the edit data in `accept` to include it.

reosarevok:
I removed them because they weren't being used at all. You think they're useful for debugging and history then or?

mwiencek:
Right, it was fine to remove them from the display data, but I'd keep them in the raw edit data for those reasons.

reosarevok:
Ok, made the changes - let me know if I missed an easier way of doing it :)

reosarevok:
Changes not being visible is intentional. In most cases annotations are an autoedit anyway, unless the user explicitly asks to make it open for a vote, so in that case it follows they aren't sure whether the annotation should look like that or not.

Having a way to see if an annotation has open edits does sound like it could be useful though, although I suspect because of how uncommon this is it shouldn't be a big deal either way.

PR #1125, opened 2019-06-28
MBS-10209: Finetune instagram regex

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10209

The current instagram regex means that if a user starts typing the URL by hand, it will add the slash at the end and let the user keep typing, eventually leading to something like instagram.com/s/o/m/e/t/h/i/n/g/

This changes the regex so that it doesn't allow typing anything after the first slash in the user name section. While this means the user can't just type the whole URL in one go, it does avoid it turning into a mess, and the user can just move back to before the slash and finish typing.

It also redirects stories links to the main user page (those can easily be accessed from the user page anyway), and blocks posts (/p/) and post location (/explore/) pages, since those aren't correctly selected as "social network" and there's no particular reason to let them be added as anything else at the moment.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -1601,12 +1601,35 @@ const testData = [
              input_entity_type: 'artist',
     expected_relationship_type: 'socialnetwork',
             expected_clean_url: 'https://www.instagram.com/deadmau5/',
+       only_valid_entity_types: ['artist', 'event', 'label', 'place', 'series'],
   },
   {
                      input_url: 'https://i.instagram.com/yorickvannorden/?ref=badge',
              input_entity_type: 'artist',
     expected_relationship_type: 'socialnetwork',
             expected_clean_url: 'https://www.instagram.com/yorickvannorden/',
+       only_valid_entity_types: ['artist', 'event', 'label', 'place', 'series'],
+  },
+  {
+                     input_url: 'https://www.instagram.com/stories/nathanwpylestrangeplanet/',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.instagram.com/nathanwpylestrangeplanet/',
+       only_valid_entity_types: ['artist', 'event', 'label', 'place', 'series'],
+  },
+  {
+                     input_url: 'https://www.instagram.com/p/B3Mew-Cl2Z9/',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.instagram.com/p/B3Mew-Cl2Z9/',
+       only_valid_entity_types: [],
+  },
+  {
+                     input_url: 'https://www.instagram.com/explore/locations/277133756/pacha-club-ibiza/',
+             input_entity_type: 'place',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.instagram.com/explore/locations/277133756/pacha-club-ibiza/',
+       only_valid_entity_types: [],
   },
   // Irish Traditional Music Tune Index (Alan Ng's Tunography)
   {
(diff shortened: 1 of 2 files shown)


reosarevok:
Do we intentionally allow for those? I wouldn't have expected us to link to those at all.

yvanzo:
I think it is intentional, even though there are only 13 Instagram location URLs in the MBDB for now.

Also, there are 60 Instagram photo URLs such as:
  `https://www.instagram.com/p/BmQ-tfpFBpV/` [[info](https://musicbrainz.org/url/5914b31f-575e-4a89-bac0-0a0480e3786b)]

These might be misfits though (any Style Leader around?).

Finally, there are 4 Instagram stories URLs of the form `https://www.instagram.com/stories/<username>/` that should probably be replaced with `https://www.instagram.com/<username>/`.

reosarevok:
I certainly wouldn't want to link to photos, and stories disappear anyway. The location... I'm not sure. It's just a bunch of pics tagged as being in that place, so if we wanted to link it it shouldn't be with the social media rel (it's not a social media page for the place).

yvanzo:
Whatever style decision is made, URLs should not be silently converted to not real Instagram usernames `explore`, `p`, or `stories`. These should be either blocked or converted to related username.

reosarevok:
Added stuff to deal with these (and updated the commit / PR message). /stories/ should redirect IMO, /p/ and /explore/ be blocked, so I implemented that.

reosarevok:
Because it's not a social media page ran by the place. It's an aggregating page for photos taken in the place, which is something different - I don't know if we want to link to it, but if we do, we should probably do it with a different sidebar name at the very least, and possibly a different relationship. In general, we use the Instagram links for the social media feeds of the entity - the profile for a concert venue and a list of photos taken by everyone at the venue are not interchangeable.

yvanzo:
> the profile for a concert venue and a list of photos taken by everyone at the venue are not interchangeable.

These are complementary.

Would an “[online community](https://musicbrainz.org/relationship/35b3a50f-bf0e-4309-a3b4-58eeed8cee6a)” relationship type (but applied to place) be legitimate?

reosarevok:
The only relationship I could see making sense for this is a specific "images from this place" one, that could also be used by some other pages if they do the same. But I'm not really too happy with allowing these at all, honestly. There's also zero control that the photos actually *are* from the place.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1579,6 +1579,26 @@ const testData = [
     expected_relationship_type: 'socialnetwork',
             expected_clean_url: 'https://www.instagram.com/yorickvannorden/',
   },
+  {
+                     input_url: 'https://www.instagram.com/stories/nathanwpylestrangeplanet/',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.instagram.com/nathanwpylestrangeplanet/',
+  },
```
yvanzo: Since your patch adds validation, please test it by adding a property, to every Instagram test entry, as follows:

```suggestion
       only_valid_entity_types: ['artist'],
  },
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1579,6 +1579,26 @@ const testData = [
...
+             input_entity_type: 'artist',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.instagram.com/p/B3Mew-Cl2Z9/',
+       only_valid_entity_types: [],
+  },
+  {
+                     input_url: 'https://www.instagram.com/explore/locations/51473250/tartu-estonia/',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.instagram.com/explore/locations/51473250/tartu-estonia/',
```
yvanzo: Please replace with the real example discussed above, because linking an artist to Tartu isn’t credible. ;)

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: [ 'artist', 'event', 'label', 'place', 'series' ] in fact, but added :)

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Replaced with a bigger place instead, and set input type to place while at it.

PR #1126, opened 2019-06-28
MBS-10268: Use DescriptiveLink for the RG bubble

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10268

Knowing the release group artist(s) is quite useful, so this changes the release group bubble in the release editor to use DescriptiveLink, which shows both title and AC rather than just title as before.

### DIFF
--- a/root/static/scripts/common/entity.js
+++ b/root/static/scripts/common/entity.js
@@ -11,6 +11,7 @@ import ReactDOMServer from 'react-dom/server';
 import ArtistCreditLink from './components/ArtistCreditLink';
 import EditorLink from './components/EditorLink';
 import EntityLink from './components/EntityLink';
+import DescriptiveLink from './components/DescriptiveLink';
 import {
   ENTITY_NAMES,
   PART_OF_SERIES_LINK_TYPES,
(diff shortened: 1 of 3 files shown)


root/static/scripts/common/entity.js
```diff
@@ -317,7 +318,7 @@ import formatTrackLength from './utility/formatTrackLength';
         selectionMessage() {
             return ReactDOMServer.renderToStaticMarkup(
                 exp.l('You selected {releasegroup}.', {
-                    releasegroup: this.reactElement({target: '_blank'}),
+                    releasegroup: <DescriptiveLink entity={this} target="_blank" />,
```
mwiencek: It doesn't look like `DescriptiveLink` allows passing through `target="_blank"` (and didn't complain since it's not Flow-typed).

reosarevok: Hmm. Seems EntityLink defaults to target blank anyway, which is why this works. So removing target.

mwiencek: I don't think it does, but we do have some JS that adds it to all links inside a bubble after it opens, so that's probably what's happening:
https://github.com/metabrainz/musicbrainz-server/blob/1039c43d60a99a140cca350c71bb3b22779334cb/root/static/scripts/edit/MB/Control/Bubble.js#L46-L48

reosarevok: Would you suggest allowing passing it to DescriptiveLink instead?

mwiencek: It might be needed by https://github.com/metabrainz/musicbrainz-server/pull/1128 (even though that's also inside a bubble, it doesn't seem to share the same bubble code linked above). So if you do add it for that, it wouldn't hurt to be explicit here too.

PR #1127, opened 2019-06-28
MBS-3112: Switch order of release editor buttons

reosarevok:
https://tickets.metabrainz.org/browse/MBS-3112

Other menus (e.g. merge menus) have Enter first, Cancel second. This doesn't matter too much for normal clicking but it makes a lot of difference in the tab order (when navigating with the keyboard).

In the release editor, we generally want to submit the edits if we're in the edit note page, and otherwise navigate either forwards or backwards, much more often than we want to cancel the whole thing (which in my experience is easier done by closing the window anyway).

### DIFF
--- a/root/release/edit/layout.tt
+++ b/root/release/edit/layout.tt
@@ -55,7 +55,7 @@
     </div>
 
     <div class="buttons">
-      <button type="button" class="negative" data-click="cancelPage">[% l('Cancel') | html_entity %]</button>
+      <button type="button" id="enter-edit" class="positive" data-click="submitEdits" data-bind="visible: activeTabID() === '#edit-note', enable: allowsSubmission()">[% l('Enter edit') %]</button>
 
       <button type="button" data-bind="visible: activeTabIndex() > 0" data-click="previousTab">[% l('« Previous') | html_entity %]</button>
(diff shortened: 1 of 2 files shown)


root/release/edit/layout.tt
```diff
@@ -64,7 +64,7 @@
         <button type="button" data-click="open">[% l('Add Medium') %]</button>
```
yvanzo: “Add Medium” should probably be put on the left of “Previous” for consistency with “Enter edit” button.

mwiencek:
Heh, I'm used to cancel always being first since that's how it is on macOS (and GNOME, IIRC), but the important thing is that it's consistent, so sure. Looks like root/components/ConfirmLayout.js still has cancel first though - ideally we can make them all consistent in on PR.

root/release/edit/layout.tt
reosarevok: Not sure I agree, because while you pretty much always want to enter the edit on the last tab, you don't always (or even often) want to add a new medium on the tracklist tab. That said, I'm not opposed to ordering it like that if you prefer either :)

reosarevok:
I guess. Pushed a commit for that one.

PR #1128, opened 2019-06-29
MBS-10269: Open AC bubble artists in new window

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10269

Some changes made in https://github.com/metabrainz/musicbrainz-server/pull/1126 also affected this, so rebased it on top of that.


root/static/scripts/edit/components/ArtistCreditBubble.js
```diff
@@ -71,11 +71,13 @@ const ArtistCreditBubble = ({
                       Object.create(entity), {artistCredit: artistCredit},
                     )}
                     showDeletedArtists={false}
+                    target="_blank"
```
mwiencek: Same comment as https://github.com/metabrainz/musicbrainz-server/pull/1126#discussion_r299671510

PR #1129, opened 2019-07-01
MBS-10272: Convert the header / navbar to Bootstrap

mwiencek:
How Bootstrap is set up:

dist/bootstrap.css from the design-system repo has been copied to root/static/styles/bootstrap/bootstrap.css. We should be depending on the design-system repo directly and import this from node_modules/ instead, but I haven't done that yet because design-system currently pulls in a bunch of unnecessary dependencies and needs to be cleaned up.

For the actual components, I chose not to use react-bootstrap for a number of reasons:

 * It does weird things with context, causing different markup to be rendered on the client and server when hydrating.
 * Its components obscure what the HTML will actually look like, making the official Bootstrap docs hard to use.
 * It outputs large component trees (in terms of component hierarchy, not markup), which has performance implications for server-rendered pages.
 * It doesn't skip pointless client-only work (like creating event handlers) on the server, which also degrades performance.

A rough performance comparison of a react-bootstrap based navbar vs. a plain React one (both passed to `ReactDOMServer.renderToString`) showed the react-bootstrap one to be 5X slower. Even though it was a contrived test for a specific component rather than a whole page, we do plan to make heavy use of Bootstrap on our pages, and I don't think making our pages slower on the server is a good idea if we can avoid it.

The only issue with avoiding react-bootstrap was a lack of JS for interactive components like menus. Fortunately, basic implementations that are "good enough" for our purposes are easy to write in React: I added root/static/scripts/common/hooks/useDropdown.js for the menus here.


reosarevok:
Uglify commit released to production as part of a hotfix for MBS-10427

mwiencek:
Fixed the conflicts here and merged in the recent menu changes that were made (IIRC just replacing /edits/all with /edits, and the new /votes link).

root/layout/components/BottomMenu.js
```diff
@@ -56,273 +47,267 @@ const LanguageMenu = ({
...
-      </ul>
-    </div>
-  );
+type Props = {
+  +$c: CatalystContextT,
+  +currentLanguage: string,
+  +serverLanguages: ?$ReadOnlyArray<ServerLanguageT>,
 };
 
-export default withCatalystContext(BottomMenu);
```
reosarevok: Why did you decide to pass $c as a prop rather than use withCatalystContext?

root/layout/components/Header.js
```diff
@@ -9,20 +9,41 @@
...
-    <a className="logo" href="/" title="MusicBrainz">
+const HeaderContents = ({$c}) => (
+  <>
+    <a className="navbar-brand my-auto" href="/" title="MusicBrainz">
       <HeaderLogo />
     </a>
-    <div className="right">
-      <TopMenu />
-      <BottomMenu />
+    <div className="collapse navbar-collapse flex-md-column">
```
reosarevok: I was expecting the menus to collapse into one of those clickable bootstrap menus when on phone version, but it does seem to just put them one after another in different lines and nothing else:

Screenshot from 2019-11-16 01-32-10

(FWIW I'd probably expect to collapse all into one menu when on small screens, and leave the search shown on its own)

root/layout/components/Search.js
```diff
@@ -42,7 +41,7 @@ const TYPE_OPTION_GROUPS = [
     editor:        N_l('Editor'),
   },
   {
-    doc:           DBDefs.GOOGLE_CUSTOM_SEARCH ? N_l('Documentation') : null,
+    doc:           N_l('Documentation'),
```
reosarevok: Do we just assume we have GOOGLE_CUSTOM_SEARCH or why this change?

root/layout/components/BottomMenu.js
```diff
@@ -56,273 +47,267 @@ const LanguageMenu = ({
...
+    </a>
+    <a className="dropdown-item" href="/doc/XML_Web_Service">
+      {l('XML Web Service')}
+    </a>
+    <a className="dropdown-item" href="/doc/Live_Data_Feed">
+      {l('Live Data Feed')}
+    </a>
+    <div className="dropdown-divider" />
+    <a className="dropdown-item" href="/doc/FreeDB_Gateway">
+      {l('FreeDB Gateway')}
```
reosarevok: This no longer exists, we should get rid.

root/layout/index.js
```diff
@@ -103,7 +103,7 @@ const Layout = ({$c, ...props}) => (
     <Head {...props} />
 
     <body>
-      <Header {...props} />
+      <Header $c={$c} />
```
reosarevok: Aren't you using withCatalystContext on Header anwyay?

mwiencek:
> Not sure which of these changes are intentional:
> 
>     1. Orange is missing on sub-menu rollover and for unfolded menu border.
> 
>     2. Font size changed for both menu and sub-menu.
> 
>     3. Font weight changed for menu.

With these Bootstrap conversions, it's expected in general for the style to look different in some ways, since it's using the Bootstrap theme developed in the design-system repo. However, I don't think these changes you listed were intentionally designed that way: a full style for the Navbar wasn't developed during the GSoC project. So I'm happy to fix some of these to match the current style (but I would prefer to keep it consistent with the design-system theme where I can, too).

root/layout/components/BottomMenu.js
mwiencek: From what I can see, because the `Header` already uses `withCatalystContext`, so `BottomMenu` isn't referenced from anywhere we can't easily pass `$c` through already.

PR #1130, opened 2019-07-01
MBS-10274: Make sure to run postgres to listen on localhost

ferbncode:
Run `postgres` and listen on localhost before trying to create database and setup schema. I think after changes due to https://github.com/docker-library/postgres/pull/440, during internal initialization, postgres only accepts connections over a unix domain socket. Please steer in the right direction if this issue should be handled differently :) (Ticket link: https://tickets.metabrainz.org/browse/MBS-10274)

### DIFF
--- a/docker/musicbrainz-test-database/create_test_db.sh
+++ b/docker/musicbrainz-test-database/create_test_db.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+pg_ctl -o "-c listen_addresses='localhost'" -w restart
 cd /home/musicbrainz/musicbrainz-server
 if [ "$MB_IMPORT_DUMPS" = "true" ]; then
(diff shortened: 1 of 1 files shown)


ferbncode:
@mwiencek Awesome! :) I tested it on my machine and this does seem to solve the issue. Thanks! :)

yvanzo:
Thank you @ferbncode!

yvanzo:
Thanks for your input @afarber. However, our Dockerfile is currently based on `postgres:16`. If that is still pertaining, the instruction should be added to [`docker/templates/Dockerfile.test-database.m4`](https://github.com/metabrainz/musicbrainz-server/blob/master/docker/templates/Dockerfile.test-database.m4), please feel free to open a pull request.

PR #1132, opened 2019-07-02
MBS-10273: Readd missing "pos" class to ReleaseGroupList

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10273

This was accidentally dropped when converting to React.


mwiencek:
Do the other list components not need it? e.g. looking at `git grep -2 -F '<input type="checkbox" />'` none of the others have it, but unclear why they wouldn't.

reosarevok:
Added it to all with a separate class name.

PR #1134, opened 2019-07-02
MBS-10248: Show relationship types in instrument lists

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10248

Sometimes it can be useful to distinguish whether an artist played an instrument in a recording, or programmed it, or arranged it.
This changes the credit column to a roles column, with the credit also shown when it exists.

### DIFF
--- a/root/components/list/RecordingList.js
+++ b/root/components/list/RecordingList.js
@@ -11,7 +11,6 @@ import React from 'react';
 
 import {withCatalystContext} from '../../context';
 import loopParity from '../../utility/loopParity';
-import {commaListText} from '../../static/scripts/common/i18n/commaList';
 import ArtistCreditLink
   from '../../static/scripts/common/components/ArtistCreditLink';
 import CodeLink from '../../static/scripts/common/components/CodeLink';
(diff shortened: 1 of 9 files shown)


mwiencek:
Hrm. This seems pretty weird to me, because the name of the link type on which the instrument appears doesn't typically make sense as the "role" (function) of the instrument.

Here are all the recording link types that have instrument attributes (so these would be all the possible "role" names):
 * samples from artist
 * balance
 * instrument arranger
 * samples material
 * instrument
 * programming

Short of having custom names for each role, maybe just changing the column name is better, but I'm not sure to what. "Relationship Appearances"?

reosarevok:
You're right - changed it to use "Relationship Types", which is actually what it is and probably makes it more clear.

mwiencek:
That's better. :) It's still weird how this basically just adds a column that says "instrument" in every cell on 90% of pages, but I'm not sure how to improve that as long as we're trying to condense relationship info into a contextless column. (By contextless I mean without the entities and link phrase to show how the instrument actually functions in the relationship.)

So maybe the solution to that is displaying the full relationships inline in the list, exactly how we do on release pages, with a button to hide them. But I'm also fine with merging this as an improvement.

root/components/list/RecordingList.js
```diff
@@ -143,11 +145,22 @@ const RecordingList = ({
...
+                      const relType = JSON.parse(json);
+                      let listElement = l_relationships(relType.name);
+                      if (relType.credit) {
+                        listElement = listElement + ' ' +
+                          bracketedText(texp.l('as “{credit}”', {credit: relType.credit}));
+                      }
+                      return listElement;
+                    }),
+                  )
+                ) : null}
```
mwiencek: It'd be nice to factor this out into a function that can be shared with root/components/list/ReleaseList.js, since it is kinda complex enough that we might want to make changes/fixes in both places at some point.

PR #1136, opened 2019-07-08
MBS-3873: Report: non-matching RG/release titles

reosarevok:
https://tickets.metabrainz.org/browse/MBS-3873

To avoid false positives (which would be quite likely otherwise), only release groups with only one release inside of them are checked for this report.

### DIFF
--- a/lib/MusicBrainz/Server/Report/ReleaseRGDifferentName.pm
+++ b/lib/MusicBrainz/Server/Report/ReleaseRGDifferentName.pm
@@ -1,0 +1,40 @@
+package MusicBrainz::Server::Report::ReleaseRGDifferentName;
+use Moose;
+
+with 'MusicBrainz::Server::Report::ReleaseReport',
+     'MusicBrainz::Server::Report::ReleaseGroupReport',
+     'MusicBrainz::Server::Report::FilterForEditor::ReleaseID';
+
+sub table { 'release_rg_different_name' }
+sub component_name { 'ReleaseRgDifferentName' }
+
+sub query {
+    "
+        SELECT
+            r.id AS release_id, rg.id AS release_group_id,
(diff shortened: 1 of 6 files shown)


lib/MusicBrainz/Server/Report/ReleaseRGDifferentName.pm
```diff
@@ -0,0 +1,37 @@
...
+}
+
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT
+
+This file is part of MusicBrainz, the open internet music database.
```
yvanzo: Should follow [coding style](https://musicbrainz.org/doc/Development/Server_Coding_Style#Perl_Module_Template_and_Documentation_Guidelines) instead.

lib/MusicBrainz/Server/Report/ReleaseRGDifferentName.pm
```diff
@@ -0,0 +1,38 @@
...
+sub query {
+    "
+        SELECT
+            r.id AS release_id, rg.id AS release_group_id,
+            row_number() OVER (ORDER BY musicbrainz_collate(r.name), musicbrainz_collate(rg.name))
+        FROM
+            release r
+            JOIN release_group rg 
+            ON r.release_group = rg.id
+        WHERE (SELECT COUNT(*) FROM release WHERE release_group = rg.id) = 1
```
mwiencek: I'd recommend joining `release_group_meta` here and checking `rgm.release_count = 1` instead of doing the subquery. Although we have an index on `release.release_group`, I ran both variants on pink and using `release_group_meta` was about 3 seconds faster (~8s vs ~5s).

PR #1137, opened 2019-07-08
MBS-9981 / MBS-10278: PossibleCollaborations improvements

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9981
https://tickets.metabrainz.org/browse/MBS-10278

### DIFF
--- a/lib/MusicBrainz/Server/Report/PossibleCollaborations.pm
+++ b/lib/MusicBrainz/Server/Report/PossibleCollaborations.pm
@@ -9,12 +9,19 @@ sub query {
         SELECT artist.id AS artist_id, row_number() OVER ( ORDER BY musicbrainz_collate(artist.name) )
         FROM
             artist
-            LEFT JOIN l_artist_artist ON l_artist_artist.entity1=artist.id
-            LEFT JOIN link ON link.id=l_artist_artist.link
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Report/PossibleCollaborations.pm
```diff
@@ -20,8 +20,7 @@ sub query {
                 WHERE 
                     l_artist_artist.entity1=artist.id
                     AND link_type.name IN
-                      ('collaboration', 'member of band')
-            )
+                      ('collaboration', 'conductor position', 'founder', 'member of band', 'subgroup')            )
```
yvanzo: Report’s description should be updated accordingly.
By the way, I cannot understand why it mentions URL relationships.

reosarevok: Fair enough (but it doesn't mention URL relationships?)

yvanzo: Oops, I was looking at CollaborationRelationships instead of PossibleCollaborations!
Please link to the related MB report in further PRs about MB reports.

yvanzo:
@reosarevok, I pushed a trivial commit to move to typographic characters, please check!

PR #1139, opened 2019-07-08
MBS-8967: Events aren't dupes if all have comment

reosarevok:
https://tickets.metabrainz.org/browse/MBS-8967

Same as we do with artists, we should assume that if all the events in a possible duplicate set have disambiguation comments, they are not actually duplicates.


reosarevok:
@yvanzo: In general, since we don't display the time in all places as we do with the date, I'd still prefer to get something like "early concert" vs "late concert" or whatever as disambiguation to make it more clear in these cases (if everything else would be the same). Arguably, a better thing would be to check whether they have artists linked and none match, maybe, but that could be added as a separate ticket :)

PR #1140, opened 2019-07-08
MBS-7097: Avoid listing releases multiple times

reosarevok:
https://tickets.metabrainz.org/browse/MBS-7097

The old query had one line for each non-digital medium, and one line for each download relationship. We only need one line per release, so the two conditions are moved to subqueries. We check that at least one medium is non-digital, since if the whole thing isn't digital then it follows you can't just download the whole thing.

### DIFF
--- a/lib/MusicBrainz/Server/Report/ReleasesWithDownloadRelationships.pm
+++ b/lib/MusicBrainz/Server/Report/ReleasesWithDownloadRelationships.pm
@@ -6,17 +6,23 @@ with 'MusicBrainz::Server::Report::ReleaseReport',
 
 sub query {
     "
-        SELECT
-            DISTINCT r.id AS release_id,
-            row_number() OVER (ORDER BY musicbrainz_collate(ac.name), musicbrainz_collate(r.name))
-        FROM
-            release r
-            JOIN artist_credit ac ON r.artist_credit = ac.id
-            JOIN l_release_url lru ON lru.entity0 = r.id
-            JOIN link ON lru.link = link.id
-            JOIN medium ON r.id = medium.release
-        WHERE link.link_type IN (74, 75)
-          AND medium.format != 12
+        SELECT DISTINCT r.id AS release_id,
+        row_number() OVER (ORDER BY musicbrainz_collate(ac.name), musicbrainz_collate(r.name))
+          FROM release r
+          JOIN artist_credit ac ON r.artist_credit = ac.id
+        WHERE EXISTS (
+            SELECT TRUE
+                FROM medium
+            WHERE medium.release = r.id
+                AND medium.format != 12
+        )
+        AND EXISTS (
+            SELECT TRUE
+                FROM l_release_url lru
+                JOIN link ON lru.link = link.id
+            WHERE lru.entity0 = r.id 
+                AND link.link_type IN (74, 75)
+        )
     ";
 }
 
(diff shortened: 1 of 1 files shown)


reosarevok:
> Isn’t download relationship legitimate when it links to bonus tracks available online?

Not sure, it seems wrong to claim it's the download link for the album.

> Shouldn’t streaming relationships (link_type = 85) be covered too?

Maybe. I'm still unconvinced about a streaming link being wrong here. The description of the guideline is "This relationship type is used to link a release to a site where the tracks can be legally streamed for free, e.g. Spotify." (which is true here). I guess the question is whether we see streaming as "real release" (which would be digital), or as "you can listen to the music in this release here" (for which the actual format would be irrelevant).

yvanzo:
> > 1. Isn’t download relationship legitimate when it links to bonus tracks available online?
> 
> Not sure, it seems wrong to claim it's the download link for the album.

It’s the download link for bonus tracks of the album. Should it be put into annotation instead?
Seems worth to document that in report’s description at least.

> > 2. Shouldn’t streaming relationships (link_type = 85) be covered too?
> 
> Maybe. I'm still unconvinced about a streaming link being wrong here. The description of the guideline is "This relationship type is used to link a release to a site where the tracks can be legally streamed for free, e.g. Spotify." (which is true here). I guess the question is whether we see streaming as "real release" (which would be digital), or as "you can listen to the music in this release here" (for which the actual format would be irrelevant).

Relationship’s description refers to the latter but this isn’t clear for everyone, including me, since some tracks can be listened to through streaming exclusively, e.g. Spotify has a few ones. Streaming is a way to release music while preventing listeners to own it. How should _streaming_ music be entered if not a release? That should probably be documented in [Style / Release](https://musicbrainz.org/doc/Style/Release) and/or [Style / Relationships / URLs](https://musicbrainz.org/doc/Style/Relationships/URLs).

reosarevok:
Exclusive streaming-only releases should certainly be added separately, I'm just not sure if I would claim it's wrong to link a CD to Spotify :)

lib/MusicBrainz/Server/Report/ReleasesWithDownloadRelationships.pm
```diff
@@ -12,11 +12,19 @@ sub query {
...
+              FROM medium
+              WHERE medium.release = r.id
+              AND medium.format != 12
+          )
+          AND EXISTS (
+              SELECT TRUE
+              FROM l_release_url lru
+                JOIN link ON lru.link = link.id
+              WHERE lru.entity0 = r.id 
+                AND link.link_type IN (74, 75)
```
mwiencek: The indentation here is wonky

yvanzo: @reosarevok: Made a suggestion with commit https://github.com/metabrainz/musicbrainz-server/pull/1140/commits/f3e00691883bbd9acf4f263f1bbd81c2b67899a7 which sticks back to initial indent style.

reosarevok: @mwiencek: are you okay with the style @yvanzo chose?

mwiencek: Sure. I prefer right-aligning the leftmost keyword, but it's a personal preference. I can read the query fine as it is.

PR #1141, opened 2019-07-08
MBS-5755: Improved unlikely language/script pairs

reosarevok:
https://tickets.metabrainz.org/browse/MBS-5755

All Western languages can also be transliterated into Katakana, and since a combination of Latin and Katakana is possible, [multiple scripts] should be accepted.
Additionally, Japanese is usually written in a specific number of scripts as well, so limiting it to Japanese scripts, Latin (for transliterations) and [multiple scripts].


reosarevok:
Sure, but it's called unlikely, not impossible :) I think it's ok to have a few false positives, MBS-2919 will eventually help with that.

PR #1142, opened 2019-07-09
MBS-10280: Cancel open edits from newest to oldest

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10280

When deleting a user, we cancel all their open edits. This can cause issues if, for example, an "Add Release Group" edit is cancelled before its associated "Add Release" edit is (the cancelling
the RG edit will fail), or if an "Add Release" edit is cancelled before edits associated with that release's recordings (the recordings won't get autoremoved and will remain orphaned).
This change makes sure we always delete the edits in reverse chronological order, so that no edits depend on already cancelled ones.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -506,15 +506,16 @@ sub delete {
         for (entities_with('ratings', take => 'model'));
 
     # Cancel any open edits the editor still has
-    my @edits = values %{ $self->c->model('Edit')->get_by_ids(
-        @{ $self->sql->select_single_column_array(
-            'SELECT id FROM edit WHERE editor = ? AND status = ?',
-            $editor_id, $STATUS_OPEN)
-       }
-    ) };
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -506,15 +506,16 @@ sub delete {
...
-       }
-    ) };
-
-    for my $edit (@edits) {
-        $self->c->model('Edit')->cancel($edit);
+    # We want to cancel the latest edits first, to make sure
+    # no conflicts happen that make some cancelling fail and all
+    # entities that should be autoremoved do get removed
+    my $ids = $self->sql->select_single_column_array(
+            'SELECT id FROM edit WHERE editor = ? AND status = ? ORDER BY open_time DESC',
```
yvanzo: It should work since `open_time` [timestamp](https://www.postgresql.org/docs/9.6/datatype-datetime.html) is written down to the microsecond, but ordering by descending `id` would probably have worked too.

PR #1144, opened 2019-07-10
MBS-3920: Add more matching to Featuring reports

reosarevok:
https://tickets.metabrainz.org/browse/MBS-3920

This extends the Featuring reports with more options that also mostly mean the artist needs to be fixed: "f/", "featuring", "ft.", "w/" and "duet with". It does not add other similar options (such as "con", "avec", "with") because they seem to cause too many false positives (mostly but not only related to classical music).

### DIFF
--- a/root/report/FeaturingReleases.js
+++ b/root/report/FeaturingReleases.js
@@ -34,11 +34,15 @@ const FeaturingReleases = ({
     <ul>
       <li>
         {exp.l(
-          `This report shows releases with (feat. Artist) in the title. For
-          classical releases, consult the {CSG|classical style guidelines}.
-          For non-classical releases, this is inherited from an older
-          version of MusicBrainz and should be fixed. Consult the
-          {featured_artists|page about featured artists} to know more.`,
+          `This report shows releases with “(feat. Artist)”
+           (or similar) in the title. For classical releases, 
+           consult the {CSG|classical style guidelines}. For 
+           non-classical releases, this is usually inherited from an
+           older version of MusicBrainz and should be fixed. Consult the
+           {featured_artists|page about featured artists} to know more.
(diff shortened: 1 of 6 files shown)


root/report/FeaturingReleases.js
```diff
@@ -34,11 +34,12 @@ const FeaturingReleases = ({
...
-          `This report shows releases with (feat. Artist) in the title. For
-          classical releases, consult the {CSG|classical style guidelines}.
-          For non-classical releases, this is inherited from an older
-          version of MusicBrainz and should be fixed. Consult the
-          {featured_artists|page about featured artists} to know more.`,
+          `This report shows releases with “(feat. Artist)”
+           (or similar) in the title. For classical releases, 
+           consult the {CSG|classical style guidelines}. For 
+           non-classical releases, this is usually inherited from an
+           older version of MusicBrainz and should be fixed. Consult the
```
yvanzo: Maybe add _(both on releases and on the tracks!)_ since it is even easier to edit tracks from release than from recordings.

reosarevok: Hmmm, why not, I can mention that :)

reosarevok:
Made the changes. I don't think we should list these specific words elsewhere in the documentation, simply because these are only some of the words that can apply (it's just I can't add many others because they cause too many false positives).

root/report/FeaturingReleases.js
```diff
@@ -34,11 +34,15 @@ const FeaturingReleases = ({
...
+           {featured_artists|page about featured artists} to know more.
+           Don't forget that the same generally applies to tracks, so if
```
yvanzo: typographical nitpicking
```suggestion
           Don’t forget that the same generally applies to tracks, so if
```

PR #1146, opened 2019-07-15
MBS-4538: Prevent years with less than 4 digits

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4538

Cases where years before 1000 are correct are rare. This requires the user to enter padding zeroes for those cases, which is not too problematic yet avoids all the cases where a too short year is added accidentally (for example because the user thought this was the day field).

After a chat with @mwiencek we decided not to try to implement this also on the Perl side.

### DIFF
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -187,6 +187,36 @@
     [% END %]
 [%- END -%]
 
+[%- MACRO too_short_year_js(type) BLOCK -%]
+  function blockTooShortBeginYear() {
+    var beginYear = $('#id-edit-[% type %]\\\.period\\\.begin_date\\\.year').val();
+    var allowed = (!beginYear || beginYear.trim().length === 4);
+    $('.submit').prop('disabled', !allowed);
+    $('#too_short_begin_year').toggle(!allowed);
+  }
+
+  function blockTooShortEndYear() {
+    var endYear = $('#id-edit-[% type %]\\\.period\\\.end_date\\\.year').val();
+    var allowed = (endYear === null || endYear === '' || endYear.length === 4);
+    $('.submit').prop('disabled', !allowed);
+    $('#too_short_end_year').toggle(!allowed);
+  }
+
+  $('#id-edit-[% type %]\\\.period\\\.begin_date\\\.year').keyup(_.debounce(blockTooShortBeginYear, 500)).change(_.debounce(blockTooShortBeginYear, 500));
+  blockTooShortBeginYear();
+
+  $('#id-edit-[% type %]\\\.period\\\.end_date\\\.year').keyup(_.debounce(blockTooShortEndYear, 500)).change(_.debounce(blockTooShortEndYear, 500));
+  blockTooShortEndYear();
+[%- END -%]
+
+[%- MACRO too_short_year_error(id) BLOCK -%]
+  <div class="error field-error" id="[% id %]" style="display: none">
+    <div class="row no-label"> 
+      [%- l('The year should have four digits. If you want to enter a year earlier than 1000 CE, please pad with zeros, such as “0123”.') -%]
+    </div>
+  </div>
+[%- END -%]
+
 [%- MACRO date_range_fieldset(r, type, ended_label) BLOCK -%]
     <fieldset>
       <legend>[% l('Date Period') %]</legend>
@@ -194,11 +224,14 @@
         [%- l('Dates are in the format YYYY-MM-DD. Partial dates such as YYYY-MM or just YYYY are OK, or you can omit the date entirely.') -%]
       </p>
       [% form_row_date(r, 'period.begin_date', l('Begin date:')) %]
+      [% too_short_year_error('too_short_begin_year') %]
       [% form_row_date(r, 'period.end_date', l('End date:')) %]
+      [% too_short_year_error('too_short_end_year') %]
       [% form_row_checkbox(r, 'period.ended', ended_label) %]
     </fieldset>
     <script type="text/javascript">
       MB.initializeToggleEnded('id-edit-[% type %]');
+      [% too_short_year_js(type) %] 
     </script>
 [%- END -%]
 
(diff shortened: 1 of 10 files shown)


root/components/forms.tt
```diff
@@ -187,18 +187,48 @@
     [% END %]
 [%- END -%]
 
+[%- MACRO too_short_year_error(id) BLOCK -%]
+  <div class="error field-error" id="[% id %]" style="display: none">
+    [% WRAPPER form_row %]
+      [%- l('The year should have four digits (if before year 1000, pad with zeros, such as 0123)') -%]
```
yvanzo: Maybe it’s just me but it sounds ambiguous whether “if” starts a new sentence or applies to the previous one. I would suggest to replace parenthesis with full sentence, and to be gentle. (Applies to both occurrences of this message.)
```suggestion
      [%- l('The year should have four digits. (If before year 1000, please pad with zeros, such as 0123.)') -%]
```

reosarevok: Changed the comment a bit following your suggestions but with a wording that seems clearer to me, let me know what you think :)

root/components/forms.tt
```diff
@@ -187,18 +187,48 @@
     [% END %]
 [%- END -%]
 
+[%- MACRO too_short_year_error(id) BLOCK -%]
+  <div class="error field-error" id="[% id %]" style="display: none">
+    [% WRAPPER form_row %]
```
mwiencek: The row div here needs the `no-label` class added so that it's aligned properly, but you'll have to add the `<div class="row no-label">` here directly to do that, since the `form_row` macro doesn't allow it.

root/components/forms.tt
```diff
@@ -187,18 +187,48 @@
...
       [% form_row_date(r, 'period.end_date', l('End date:')) %]
+      [% too_short_year_error('too_short_end_year') %]
       [% form_row_checkbox(r, 'period.ended', ended_label) %]
     </fieldset>
     <script type="text/javascript">
       MB.initializeToggleEnded('id-edit-[% type %]');
+
+      function blockTooShortBeginYear() {
+        var beginYear = $('#id-edit-[% type %]\\\.period\\\.begin_date\\\.year').val();
+        var allowed = (beginYear === null || beginYear === '' || beginYear.length === 4);
```
mwiencek: I don't think the value can be `null` there, only `undefined` if the input doesn't exist (but it should exist). A slightly more robust check might be `!beginYear || beginYear.trim().length === 4`, which also removes whitespace before checking the length.

root/event/edit_form.tt
```diff
@@ -58,5 +60,25 @@
...
+      var allowed = (endYear === null || endYear === '' || endYear.length === 4);
+      $('.submit').prop('disabled', !allowed);
+      $('#too_short_end_year').toggle(!allowed);
+    }
+
+    $('#id-edit-event\\\.period\\\.begin_date\\\.year').keyup(blockTooShortBeginYear).change(blockTooShortBeginYear);
+    blockTooShortBeginYear();
+
+    $('#id-edit-event\\\.period\\\.end_date\\\.year').keyup(blockTooShortEndYear).change(blockTooShortEndYear);
+    blockTooShortEndYear();
```
mwiencek: A macro which takes the entity `type` as an argument and outputs this script would be nicer than duplicating it 3x (here, root/artist/edit_form.tt, and root/components/forms.tt).

root/static/scripts/release-editor/fields.js
```diff
@@ -625,6 +630,11 @@ class ReleaseEvent {
             var date = self.unwrapDate();
             return !dates.isDateValid(date.year, date.month, date.day);
         });
+
+        this.hasTooShortYear = ko.computed(function () {
```
mwiencek: One slightly annoying thing is how it now flashes the error message while you're typing in any year. We could work around that by importing root/static/scripts/common/utility/debounce.js and replacing `ko.computed` with `debounce`, which will only update the observable when it stops changing for 500ms (or any configured timeout).

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -195,6 +196,27 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
 
             this.changeOtherRelationshipCredits = {source: ko.observable(false), target: ko.observable(false)};
             this.selectedRelationshipCredits = {source: ko.observable('all'), target: ko.observable('all')};
+
+            this.tooShortBeginYearError = debounce(function () {
+                const relationship = self.relationship();
+                if (!relationship.begin_date) {
```
mwiencek: This *should* always be defined: https://github.com/metabrainz/musicbrainz-server/blob/dbfa033383754f3632aa53421b776ad1f5074adf/root/static/scripts/relationship-editor/common/fields.js#L65-L65

Do you remember what issue occurs or which test fails if the check is removed?

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -195,6 +196,27 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
...
+                    return '';
+                }
+                const date = relationship.begin_date;
+                const valid = dates.isYearFourDigits(date.year());
+                return valid ? '' : l('The year should have four digits. If you want to enter a year earlier than 1000 CE, please pad with zeros, such as “0123”.');
+            });
+
+            this.tooShortEndYearError = debounce(function () {
```
mwiencek: It'd be nice to merge these definitions though. One way is we can have a `tooShortYear(date)` function right above these and call it inside both debounces, passing `self.relationship().begin_date` or `self.relationship().end_date`.

root/static/scripts/relationship-editor/common/dialog.js
reosarevok: No, but I removed it and it seems to work so I guess there wasn't one!

PR #1147, opened 2019-07-15
MBS-10255: Explanations for "copy all" checkboxes

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10255

### DIFF
--- a/root/release/edit/recordings.tt
+++ b/root/release/edit/recordings.tt
@@ -115,6 +115,12 @@
 
   <!-- ko if: hasTracks() -->
     <fieldset>
+      <p>
+        [% l('If the {recording_doc|recording titles/artists} listed above are incorrect, you can check the boxes below to replace them with the track titles/artists for this release.', {recording_doc => doc_link('Recording')}) %]
+      </p>
+      <p>
+        [% l('<strong>Note:</strong> In classical music, {csg_recording_artist|recording artists} should almost always be different from {csg_track_artist|track artists}, so don’t use the “copy artist credits” option while editing a classical release unless you’re very sure it’s a good idea!', {csg_recording_artist => doc_link('Style/Classical/Recording/Artist'), csg_track_artist => doc_link('Style/Classical/Track/Artist')}) %]
+      </p>
       <legend>[% l('Options') %]</legend>
       <p>
         <label>


root/release/edit/recordings.tt
```diff
@@ -128,6 +128,12 @@
           [% l('Copy all track artist credits to associated recordings.') %]
         </label>
       </p>
+      <p>
+        [% l('Only check these boxes if the recording titles/artists listed above are incorrect and you want to replace them with the track titles/artists) for this release.') %]
+      </p>
+      <p>
+        [% l('💡<strong>Note:</strong> In classical music, {csg_recording_artist|recording artists} should almost always be different from {csg_track_artist|track artists}, so don’t use the “copy artist credits” option while editing a classical release unless you’re very sure it’s a good idea!', {csg_recording_artist => doc_link('Style/Classical/Recording/Artist'), csg_track_artist => doc_link('Style/Classical/Track/Artist')}) %]
```
yvanzo: 1. 💡 should belong to style, not to message;
2. If we plan to add more notes like this, a dedicated TT macro/React component using a message '{alert}: {content}' would probably be helpful.

root/release/edit/recordings.tt
```diff
@@ -128,6 +128,12 @@
           [% l('Copy all track artist credits to associated recordings.') %]
         </label>
       </p>
+      <p>
+        [% l('Only check these boxes if the recording titles/artists listed above are incorrect and you want to replace them with the track titles/artists) for this release.') %]
```
yvanzo: Please link “recording” to [`/doc/Recording`](https://musicbrainz.org/doc/Recording) as it shows the difference between recordings and tracks.

mwiencek: There's an errant ")" in there, and I feel like a comma would help since the clauses are pretty long
```suggestion
        [% l('Only check these boxes if the recording titles/artists listed above are incorrect, and you want to replace them with the track titles/artists for this release.') %]
```

root/release/edit/recordings.tt
mwiencek: I'm not sure what "{alert}: {content}" would give the translator to be honest. It'd allow you to change the order of the alert and content messages, or change the colon, but that's already possible here. There's also less context if you detach the messages from each other.

mwiencek:
> If this paragraph is intended to help with understanding when editor may need options, then it should probably be placed ahead of check boxes and be phrased as an introduction text, e.g., “If the recording titles/artists listed above are incorrect, the two options below allow you to replace them with the track titles/artists) for this release.” It delivers the same information but it empowers editors instead of hinding them.

I like this suggestion (phrasing things like "If X, you can do Y" rather than "Don't do Y unless X")

root/release/edit/recordings.tt
reosarevok: So the suggestion would be putting💡out of the string and into a note_icon() macro or something like that? Or literally add it via CSS?

yvanzo: @mwiencek: The benefit of having `<strong>Note:</strong>` apart is that it doesn’t have to be translated 20 times when there are 20 notes. Context can be added with `lp`. But there is probably not 20 notes right now, so this might be more pain for little help at the moment.

@reosarevok: Probably both.

mwiencek: I understand that, but I meant the "{alert}: {content}" string specifically. :) If you have separate strings already, it seems to me that you can just display them next to each other rather than interpolate them together like that.

yvanzo: Good catch, that string is erroneous anyway.

reosarevok:
Updated this by moving the help above the checkboxes as suggested, and removing the icon from the note. The `<strong>Note:</strong>` structure is already used in two other strings in the code, so I'd just keep it inside the string here too, unless we want to change those as well (and one is a reltype description so probably can't really be changed).

Screenshot from 2019-12-04 21-31-06

PR #1148, opened 2019-07-15
Fix release merge issues MBS-10188 and MBS-10279

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10188
https://tickets.metabrainz.org/browse/MBS-10279

### DIFF
--- a/lib/MusicBrainz/Server/Data/Release.pm
+++ b/lib/MusicBrainz/Server/Data/Release.pm
@@ -1,13 +1,16 @@
 package MusicBrainz::Server::Data::Release;
 
+use 5.18.2; # enables the state feature
+
 use Moose;
 use namespace::autoclean -also => [qw( _where_status_in _where_type_in )];
 
 use Carp 'confess';
 use DBDefs;
+use JSON::XS;
 use List::AllUtils qw( all );
 use List::MoreUtils qw( part );
-use List::UtilsBy qw( partition_by );
+use List::UtilsBy qw( nsort_by partition_by );
 use MusicBrainz::Server::Constants qw( :quality $EDIT_RELEASE_CREATE $STATUS_APPLIED );
 use MusicBrainz::Server::Entity::Barcode;
 use MusicBrainz::Server::Entity::PartialDate;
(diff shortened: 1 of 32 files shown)


lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -880,54 +883,155 @@ sub can_merge {
...
+        my ($new_recording, $old_recordings) = @{$possible_merge}{qw(
+            new_recording
+            old_recordings
+        )};
+
+        # As mentioned above, remove old (source) recordings that have
+        # multiple targets.
+        #
+        # Perl fact: this assignment modifies the array in $possible_merge.
+        @{$old_recordings} = grep { $target_count{$_->{id}} == 1 } @{$old_recordings};
```
reosarevok: I'm confused by this. Wouldn't we want to just block the release merge if it leads to a situation like this one? What is the alternative? Just leaving the old recording behind, orphaned?

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -880,54 +883,155 @@ sub can_merge {
...
+        $new_recording = $merge_targets{$new_recording->{id}} // $new_recording;
+        my $new_id = $new_recording->{id};
+
+        for my $old_recording (@{$old_recordings}) {
+            my $old_id = $old_recording->{id};
+
+            # If two recordings' positions are swapped (e.g. recording 1 is being
+            # merged into recording 2, and recording 2 is being merged into
+            # recording 1), then we don't merge them in that case, because it's
+            # probably not intentional.
```
reosarevok: I'd expect this to also cancel the merge.

reosarevok: (I'd also suggest to not even allow entering a merge edit if this is the case when the edit is entered)

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -1,13 +1,16 @@
 package MusicBrainz::Server::Data::Release;
 
+use 5.18.2;
```
reosarevok: Why? (might make sense to specify in the commit message?)

root/edit/details/merge_releases.tt
```diff
@@ -121,7 +121,7 @@
         [%~ END; END %]
     [%~ END ~%]
     </tbody></table>
-[%~ ELSIF edit.display_data.merges_are_calculated && !edit.is_open ~%]
+[%~ ELSIF !edit.is_open ~%]
     <p><strong>[% l('This edit does not store recording merge information and is closed, so no recording merge information can be shown.') %]</strong></p>
 [%~ ELSE  ~%]
```
reosarevok: Could fix the double space while at it :)

root/edit/details/merge_releases.tt
```diff
@@ -121,7 +121,7 @@
         [%~ END; END %]
     [%~ END ~%]
     </tbody></table>
-[%~ ELSIF edit.display_data.merges_are_calculated && !edit.is_open ~%]
+[%~ ELSIF !edit.is_open ~%]
     <p><strong>[% l('This edit does not store recording merge information and is closed, so no recording merge information can be shown.') %]</strong></p>
```
reosarevok: Do we have a way of knowing whether no recordings needed merged in the first place? Otherwise all the stuff covered by ELSE below will become this as soon as the edit closes, no?

reosarevok:
For the ModBot bit, I'd suggest also adding a separate ticket to convert previous notes to the new format with a SQL script at some point (a lot should be trivial to improve).

lib/MusicBrainz/Server/Data/Release.pm
reosarevok: Oh, I see that a following commit does exactly that, IIUC. Nevermind then, I guess!

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -810,7 +816,7 @@ sub can_merge {
             @old_ids, $new_id);
 
         if ($mediums_differ) {
-            $opts->{_cannot_merge_reason} = N_l('The track counts on at least one set of corresponding mediums do not match.');
+            $opts->{_cannot_merge_reason} = $RELEASE_MERGE_ERRORS{medium_track_counts};
             return 0;
         }
```
reosarevok: This should probably be `if ($medium_track_counts_differ)` then? `$mediums_differ` could mean anything :)

lib/MusicBrainz/Server/Data/Release.pm
mwiencek: Yes, that's just how it always worked since 7eaca53cbc2e75f83237098868cbb6d8182203a8, but I changed it to fail the edit in one of the commits.

lib/MusicBrainz/Server/Data/Release.pm
mwiencek: That's handled by commit "Reject recording merge cycles"

lib/MusicBrainz/Server/Data/Release.pm
mwiencek: To use the [state](https://perldoc.perl.org/functions/state.html) function (I'll add a comment)

root/edit/details/merge_releases.tt
mwiencek: That's a good point. I changed the conditions here to check if `recording_merges` is defined in the display data. Now, if it's not defined in the edit data (the edit is old), it's left undefined in the display data too so we can detect that.

But the answer is no, for old edits we can't easily know whether no recordings needed merging, since it doesn't store anything either way.

lib/MusicBrainz/Server/Data/Release.pm
mwiencek: Technically this error is also triggered if the target release has less mediums than the source release (but it lets you merge in the other direction, I don't recall why). So the message is also wrong in that case. Should I add a separate message for that case, or should we allow that type of merge if possible?

root/edit/details/merge_releases.tt
mwiencek: But for edits newer than 0e41a26a69ab8f5d45b81f10ba321022791a13f0 yes, if the edit data has as an empty `recording_merges` array, then we know the recordings were already all merged.

lib/MusicBrainz/Server/Data/Release.pm
reosarevok: The other direction is needed so you can merge an orphaned CD 1 out of a set for example into the set :) I'd make it a different message for the big into small merge, but still block it - it'd be hard to understand the user's expected result in that case IMO.

mwiencek: Right, I mean I knew why it was needed but didn't recall why it only allowed one direction and not the other. But if it keeps the intent clear I'll leave it blocked & and come up with a suitable message for that.

mwiencek: This should be fixed by e17884f :)

mwiencek:
Tweaked the message, since it might also show if the medium counts are the same, but a medium position is missing from the target.

PR #1150, opened 2019-07-16
MBS-2663 (II): Search edits by non-subscribed editor

reosarevok:
https://tickets.metabrainz.org/browse/MBS-2663

As a follow-up to https://github.com/metabrainz/musicbrainz-server/pull/1119 - I just realized that I didn't implement it for editors. Same general idea.

### DIFF
--- a/root/static/scripts/common/MB/edit_search.js
+++ b/root/static/scripts/common/MB/edit_search.js
@@ -15,7 +15,7 @@ $(function () {
             '=': 1, '!=': 1 // Not directly true, but it here it means "show one argument control"
         },
         'voter': {
-            '=': 1, '!=': 1, 'me': 0, 'not_me': 0, 'subscribed': 0, 'not_subscribed': 0 // Not directly true, but it here it means "show one argument control"
+            '=': 1, '!=': 1, 'me': 0, 'not_me': 0, 'subscribed': 0, 'not_subscribed': 0
         },
         'subscription': {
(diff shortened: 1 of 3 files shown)


root/static/scripts/common/MB/edit_search.js
```diff
@@ -15,7 +15,7 @@ $(function () {
             '=': 1, '!=': 1 // Not directly true, but it here it means "show one argument control"
         },
         'voter': {
-            '=': 1, '!=': 1, 'me': 0, 'not_me': 0, 'subscribed': 0 // Not directly true, but it here it means "show one argument control"
+            '=': 1, '!=': 1, 'me': 0, 'not_me': 0, 'subscribed': 0, 'not_subscribed': 0 // Not directly true, but it here it means "show one argument control"
```
mwiencek: Sorry to request something unrelated to your issue, but can we remove the "Not directly true" comment here in an additional commit? :) It confused me during review and it doesn't seem to apply to this line at all. It makes sense for the "set" predicate above, and that's where it was originally added, so I'm guessing it was just copied here by accident. (It makes sense to me for "set" because they can have multiple items, so it's weird to say the cardinality is "1," but it's only for showing 1 input as mentioned.)

mwiencek: Sorry (again) for the noise, I forgot I can just push a commit to this PR directly instead of bothering you to do it. Done that.

PR #1151, opened 2019-07-17
MBS-6791: Search for edits made by beginners

reosarevok:
https://tickets.metabrainz.org/browse/MBS-6791

The query checks for all the conditions of a beginner editor, except whether they have a confirmed email address, since you can't edit without a confirmed email address anyway.

Works great on my sample DB, hope it scales fine to a full DB :)

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Editor.pm
+++ b/lib/MusicBrainz/Server/Entity/Editor.pm
@@ -178,6 +178,7 @@ has 'preferences' => (
 
 sub is_limited
 {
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -178,7 +178,8 @@ has 'preferences' => (
 
 sub is_limited
 {
-    my $self = shift;
+   # Please keep the logic in sync with Report::LimitedEditors and EditSearch::Predicate::Role::User
+   my $self = shift;
```
mwiencek: The indentation was screwed up slightly here

PR #1153, opened 2019-07-18
MBS-4161: List blogs under external links

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4161

This lists blogs under external links. They get listed below homepages, but above other links. The used icon is from the same set as the homepage icon, both for consistency and for ease of attribution.


root/layout/components/ExternalLinks.js
```diff
@@ -54,6 +54,7 @@ const ExternalLink = ({className, relationship, text}) => {
 const ExternalLinks = ({entity, empty, heading}) => {
   const relationships = entity.relationships;
   const links = [];
+  const blogLinks = [];
```
reosarevok: Added a separate blogLinks because it seems like the easiest way to ensure they always sort under the official website, whatever the language.

root/layout/components/ExternalLinks.js
```diff
@@ -75,12 +76,21 @@ const ExternalLinks = ({entity, empty, heading}) => {
           text={l('Official homepage')}
         />,
       );
+    } else if (/^blog$/.test(linkType.name)) {
```
yvanzo: Would be nice to add a comment to allow finding such hard-coded database value.
```suggestion
    // Hard-coded condition on data: musicbrainz.link_type.name = 'blog'
    } else if (/^blog$/.test(linkType.name)) {
```
That can be discussed/done outside of this PR though.

mwiencek: Or just compare against `linkType.gid` here? And add a `BLOG_LINK_TYPE` constant to root/static/scripts/common/constants.js to make it nicer.

mwiencek: (Though that's a bit more annoying since you have to factor in the entity types, and it's unlikely the name should change, so a comment seems fine.)

PR #1156, opened 2019-07-18
MBS-4090: Report: non-matching recording/track titles

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4090

To avoid false positives (which would be quite likely otherwise), only recordings linked to just one track are checked for this report.

The original request also asked for differences in artist credits, but that would make almost all classical recordings false positives, and titles-only will still probably catch most cases that need fixing.

### DIFF
--- a/root/report/RecordingTrackDifferentName.js
+++ b/root/report/RecordingTrackDifferentName.js
@@ -1,0 +1,95 @@
+/*
+ * @flow
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import {withCatalystContext} from '../context';
+import Layout from '../layout';
+import formatUserDate from '../utility/formatUserDate';
+import PaginatedResults from '../components/PaginatedResults';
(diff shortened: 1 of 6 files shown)


root/report/RecordingTrackDifferentName.js
```diff
@@ -0,0 +1,95 @@
...
+      {canBeFiltered ? <FilterLink filtered={filtered} /> : null}
+    </ul>
+
+    <PaginatedResults pager={pager}>
+      <table className="tbl">
+        <thead>
+          <tr>
+            <th>{l('Artist')}</th>
+            <th>{l('Recording')}</th>
+            <th>{l('Track')}</th>
```
reosarevok: I'm starting to think I should have kept the option we had in TT to pass a list template extra columns that will be displayed before or after the default ones. But I'm not sure how that'd be doable without angering Flow.

reosarevok:
I wouldn't call those cases false positives at all - if we have a release, I'd expect them to follow the title there :) I understand the point, though, in that it maybe had a different name when standalone or whatnot, but I think we do generally consider properly-released superior to standalone data. Merged this, if we see that becomes a problem we can always add an extra filter.

yvanzo:
> I wouldn't call those cases false positives at all

Those are false postives in the sense the title for standalone recording matches an external source. Maybe it should not be the same recording to start with.

> I think we do generally consider properly-released superior

It’s a play on semantics, standalone recordings most often are proper online releases. Maybe we should think about making easier to add digital single from a URL to get the music.

PR #1157, opened 2019-07-18
MBS-5641: Show language/script in relevant reports

reosarevok:
https://tickets.metabrainz.org/browse/MBS-5641

Mostly follows how it's done on release search results, but I chose to show "-" rather than nothing when language or script are missing, since I feel it makes *what* is missing more understandable.
Arguably could have been done so that the NoScript report only shows language and NoLanguage only script, but by showing both it also serves to show the editor whether some have already been fixed (since the language and script are loaded from the release entry and not from the saved report data).

### DIFF
--- a/root/report/components/ReleaseList.js
+++ b/root/report/components/ReleaseList.js
@@ -16,42 +16,73 @@ import type {ReportReleaseT} from '../types';
 import ArtistCreditLink
   from '../../static/scripts/common/components/ArtistCreditLink';
 
+type Props = {|
+  +items: $ReadOnlyArray<ReportReleaseT>,
+  +pager: PagerT,
+  +showLanguageAndScript?: boolean,
+|};
+
 const ReleaseList = ({
   items,
   pager,
-}: {items: $ReadOnlyArray<ReportReleaseT>, pager: PagerT}) => (
-  <PaginatedResults pager={pager}>
-    <table className="tbl">
-      <thead>
-        <tr>
-          <th>{l('Release')}</th>
-          <th>{l('Artist')}</th>
-        </tr>
-      </thead>
-      <tbody>
-        {items.map((item, index) => (
-          <tr className={loopParity(index)} key={item.release_id}>
-            {item.release ? (
-              <>
-                <td>
-                  <EntityLink entity={item.release} />
-                </td>
-                <td>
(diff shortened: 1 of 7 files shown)


root/report/components/ReleaseList.js
```diff
@@ -16,42 +16,69 @@ import type {ReportReleaseT} from '../types';
...
+                      artistCredit={item.release.artistCredit}
+                    />
+                  </td>
+                  {showLanguageAndScript ? (
+                    <td>
+                      {item.release.language ? (
+                        <abbr title={l_languages(item.release.language.name)}>
+                          {item.release.language.iso_code_3}
+                        </abbr>
+                      ) : lp('?', 'unset_language_script')}
```
reosarevok: I thought about using '[unset]' at first but it feels too long.

root/report/components/ReleaseList.js
```diff
@@ -16,42 +16,73 @@ import type {ReportReleaseT} from '../types';
...
+                        {script ? (
+                          <abbr title={l_scripts(script.name)}>
+                            {script.iso_code}
+                          </abbr>
+                        ) : lp('?', 'unset_language_script')}
+                      </td>
+                    ) : null}
+                  </>
+                ) : (
+                  <td colSpan={colspan}>
```
yvanzo: Nitpicking:
```suggestion
                  <td colSpan={colSpan}>
```

root/report/components/ReleaseList.js
```diff
@@ -16,42 +16,73 @@ import type {ReportReleaseT} from '../types';
...
-                </td>
-              </>
-            ) : (
-              <td colSpan="2">
-                {l('This release no longer exists.')}
-              </td>
-           )}
+  showLanguageAndScript,
+}: Props) => {
+  const colspan = showLanguageAndScript ? 3 : 2;
```
yvanzo: Following other comment:
```suggestion
  const colSpan = showLanguageAndScript ? 3 : 2;
```

reosarevok:
I won't change it to match that because this is releases and for now we haven't changed release display. Maybe we should, eventually, but :)

reosarevok:
Changed ? to - to match other tables where we display the fact we're missing data :)

PR #1158, opened 2019-07-18
MBS-3609: Show language/script on more release lists 

reosarevok:
https://tickets.metabrainz.org/browse/MBS-3609

There's certainly cases where the only thing differentiating two releases is language/script. As such, it makes sense to have language/script in all release lists, not just the search results.

Yes, this makes an already huge table even more crowded. Eventually we should try and figure out a better way to show this data than a table, possibly. But I think it still makes sense.


yvanzo:
> this makes an already huge table even more crowded

It might be even more concerning once languages and scripts are formatted as in https://github.com/metabrainz/musicbrainz-server/pull/1420.

reosarevok:
We haven't decided whether we want to do that for releases. We might decide for releases it does make sense to shorten them since they're two different fields, not one, and release tables are already crowded.

root/components/ReleaseLanguageScript.js
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+const ReleaseLanguageScript = ({release}: {release: ReleaseT}) => {
```
mwiencek, line 12: There's another place we can use this in root/report/components/ReleaseList.js, I think, though instead of `null` it uses `lp('-', 'missing data')`. Perhaps that's something we can do everywhere though?

reosarevok: Sounds like a good plan, made that change!

reosarevok:
Made a small change to make the component shared between search results and release reports. If the first commit seems good, we could cherry-pick it as a start, some consistency seems useful.

Aerozol:
Reo asked me to have a look at this :)

I agree that it already a pretty crammed table, and this does not seem _critical_ to identifying a releases in the vast majority of cases.

The rest of this is looking at other display solutions, ignore if out of your scope:

**Option A.** This would allow us to do way more with columns. There is a wealth of info that we could let users choose to surface (length, external links, language, ratings, LB global listens, LB your listens, etc).

I have this as part of the redesign mockup - we keep talking about "waiting for the redesign" but perhaps this would be something that could be done now? A lot of sites have customizable columns:

image

That would be a drop down with a list of columns that you can check and uncheck. I imagine the default would be the same or very similar to what it is now.

**Option B.** Quicker, but not nearly as cool :)
Would be nullified by the alternate tracklistings update?

We change the display for transliterated tracklistings only, as @yvanzo mentions, which seems fine to me, as they serve a very specific purpose, and most of the columns are explicitly not meant to be set anyway.

B1: Add the language to the existing title, for instance:

image

B2: Add a different column for the Pseudo-release subsection:

image

B3: Pretty much the same as B2, but separate pseudo-release out a little into its own table, which gives a less confusing display:

image


P.S. Do we no longer show it on hover? I tried hovering everything, no luck

PR #1159, opened 2019-07-19
MBS-10753: Use artist sort names for artist collection ordering

reosarevok:
### Implement MBS-10753

The reason we have a sort name is to sort by it, so when sorting collections by artist we should use the artist's sort names, not their names.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -226,13 +226,13 @@ sub _order_by {
     my ($self, $order) = @_;
     my $order_by = order_by($order, "name", {
         "name" => sub {
-            return "musicbrainz_collate(name)"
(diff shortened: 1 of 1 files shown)


mwiencek:
Sorry for the long silence here. The issue with computing an "AC sort name" on the fly is that it's very, very slow. I noticed this locally in a collection with just 4 releases (the page took nearly 2 additional seconds to load), so I went and measured the queries in production against a slightly larger collection (72 releases). While sorting the 72-release collection by artist with the old query took around 3.5ms on pink, this new query takes over 13.5s (so close to 4000x slower).

I think we really will need a schema change to materialize these sort names on the artist_credit table in order to do this. But we can still merge 14b1dedf70f0e341685746b3a26e14846bfe958f as-is if you want to update the PR to just do that.

reosarevok:
Ok, I split this into two tickets then, the artist collection bit is now MBS-10753

PR #1160, opened 2019-07-20
MBS-4091: Avoid breaking lists while trimming annotations

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4091

From annotation formatting:
Lists: tab or 4 spaces and:
* bullets or 1., a., A., i., I. numbered items (rendered with 1.)

We are trimming all whitespace at the start and end of the annotation though (FormHandler's default), meaning if the annotation starts with a list the formatting for the first element will be broken. This overrides the default trim method with one that will leave one tab or four spaces if there's that or more than that at the start of the annotation.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Annotation.pm
+++ b/lib/MusicBrainz/Server/Form/Annotation.pm
@@ -7,6 +7,13 @@ has '+name' => (default => 'edit-annotation');
 
 has_field 'text' => (
     type     => 'Text',
+    trim => { transform => sub {
+        my $string = shift;
+        # Not trimming starting spaces to avoid breaking list formatting,
+        # consider trimming again once this uses Markdown 
+        $string =~ s/\s+$//;
+        return $string;
+    } }
 );
 
 has_field 'changelog' => (


reosarevok:
Alternatively, we might want to assume too much spacing is always an error and trim it unless it's *exactly* 4 spaces or one tab.

lib/MusicBrainz/Server/Form/Annotation.pm
```diff
@@ -7,6 +7,17 @@ has '+name' => (default => 'edit-annotation');
...
 has_field 'text' => (
     type     => 'Text',
+    trim => { transform => sub {
+        my $string = shift;
+        # We don't want to break list formatting at the start of the annotation
+        if ($string =~ m/^\s*(?:(?: {4})|\t)[^\s]/) {
+            $string =~ s/^\s*((?:(?: {4})|\t)[^\s])/$1/;
+        } else {
+            $string =~ s/^\s+//;
+        }
```
mwiencek: TBH this makes me uneasy looking at it, especially since the list formatting syntax depends on the wikitext parser being used.

Maybe we should just trim whitespace at the end, and leave the beginning of the string untouched.

reosarevok:
Changed to trim ending whitespace only.

PR #1161, opened 2019-07-20
MBS-9111: Stricter validation for empty edit notes

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9111
Also https://tickets.metabrainz.org/browse/MBS-10373 as an addendum

We keep having issues with people adding edit notes with no content.
This blocks edit notes with only whitespace and (common) punctuation combinations, and asks the user to provide an edit note that will be useful to fellow editors.

It also blocks one-character ASCII notes, but not one-character notes in wider Unicode, since glyph languages (such as Japanese) might have a full meaningful word that is one character long.

People can obviously still enter just "á" or something, but at least it might block some of the bad edit notes while reminding users why a note is important.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Role/EditNote.pm
+++ b/lib/MusicBrainz/Server/Form/Role/EditNote.pm
@@ -2,6 +2,7 @@ package MusicBrainz::Server::Form::Role::EditNote;
 use HTML::FormHandler::Moose::Role;
 
 use MusicBrainz::Server::Translation qw( l );
+use MusicBrainz::Server::Validation qw( is_valid_edit_note );
 
 has 'requires_edit_note' => ( is => 'ro', default => 0 );
 
@@ -18,8 +19,12 @@ has_field 'make_votable' => (
 after validate => sub {
     my $self = shift;
 
-    if ($self->requires_edit_note && (!defined $self->field('edit_note')->value || $self->field('edit_note')->value eq '')) {
+    if ($self->requires_edit_note && (!defined $self->field('edit_note')->value)) {
         $self->field('edit_note')->add_error(l('You must provide an edit note'));
+    }
+
+    if ($self->requires_edit_note && (!is_valid_edit_note($self->field('edit_note')->value))) {
+        $self->field('edit_note')->add_error(l('Your edit note seems to have no actual content. Please provide a note that will be helpful to your fellow editors!'));
     }
 };
 
--- a/lib/MusicBrainz/Server/Validation.pm
+++ b/lib/MusicBrainz/Server/Validation.pm
@@ -56,6 +56,7 @@ require Exporter;
         is_valid_iso_3166_2
         is_valid_iso_3166_3
         is_valid_partial_date
+        is_valid_edit_note
         encode_entities
         normalise_strings
         is_nat
@@ -306,6 +307,20 @@ sub is_valid_partial_date
         # partial dates with year <= 0 are OK, but complete dates are not (don't ask)
         return 0 unless $year > 0;
     }
+
+    return 1;
+}
+
+sub is_valid_edit_note
+{
+    my $edit_note = shift;
+
+    # An edit note with only spaces and / or punctuation is useless
+    return 0 if $edit_note =~ /^[[:space:][:punct:]]+$/;
+
+    # An edit note with just one ASCII character is useless
+    # A one-character Japanese note (for example) might be useful, so limited to ASCII 
+    return 0 if $edit_note =~ /^[[:ascii:]]$/;
 
     return 1;
 }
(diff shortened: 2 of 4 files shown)


lib/MusicBrainz/Server/Form/Role/EditNote.pm
```diff
@@ -18,9 +19,13 @@ has_field 'make_votable' => (
 after validate => sub {
     my $self = shift;
 
-    if ($self->requires_edit_note && (!defined $self->field('edit_note')->value || $self->field('edit_note')->value eq '')) {
+    if ($self->requires_edit_note && (!defined $self->field('edit_note')->value)) {
         $self->field('edit_note')->add_error(l('You must provide an edit note'));
     }
+
+    if ($self->requires_edit_note && (!is_valid_edit_note($self->field('edit_note')->value))) {
+        $self->field('edit_note')->add_error(l('Your edit note seems to have no actual content. Please provide a note that will be helpful to your fellow editors!'));
```
yvanzo: It would probably be helpful to link it to [`/doc/How_to_Write_Edit_Notes`](https://musicbrainz.org/doc/How_to_Write_Edit_Notes)

reosarevok: This doesn't seem to work on error texts at the moment (I got `You must provide <a href="/doc/How_to_Write_Edit_Notes">an edit note</a>`) but in any case it seems like it would be relevant always, not just if an error appears. So I added [MBS-10373](https://tickets.metabrainz.org/browse/MBS-10373) for that and added it to the general edit note text.

lib/MusicBrainz/Server/Validation.pm
```diff
@@ -310,6 +311,24 @@ sub is_valid_partial_date
...
+    my $edit_note = shift;
+
+    # An edit note with only spaces and / or punctuation is useless
+    if ($edit_note =~ /^[\s.,+\-\\\/'"´`@#~()\[\]{}]+$/) {
+        return 0
+    }
+
+    # An edit note with just one ASCII character is useless
+    # A one-character Japanese note (for example) might be useful, so limited to ASCII 
+    if ($edit_note =~ /^[ -~]{1}$/) {
```
mwiencek:
```suggestion
    if ($edit_note =~ /^[[:ascii:]]$/) {
```
would be slightly clearer

lib/MusicBrainz/Server/Validation.pm
```diff
@@ -310,6 +311,24 @@ sub is_valid_partial_date
     return 1;
 }
 
+sub is_valid_edit_note
+{
+    my $edit_note = shift;
+
+    # An edit note with only spaces and / or punctuation is useless
+    if ($edit_note =~ /^[\s.,+\-\\\/'"´`@#~()\[\]{}]+$/) {
```
mwiencek:
```suggestion
    if ($edit_note =~ /^[[:space:][:punct:]]+$/) {
```

lib/MusicBrainz/Server/Form/Role/EditNote.pm
```diff
@@ -18,9 +19,13 @@ has_field 'make_votable' => (
 after validate => sub {
     my $self = shift;
 
-    if ($self->requires_edit_note && (!defined $self->field('edit_note')->value || $self->field('edit_note')->value eq '')) {
```
mwiencek: The `''` check was removed, but it doesn't look like `is_valid_edit_note` checks for an empty string either. (Or is it just always undefined when empty? Unclear why the check was removed anyway.)

lib/MusicBrainz/Server/Validation.pm
reosarevok: I didn't know that was possible! Does seem better, will switch.

lib/MusicBrainz/Server/Form/Role/EditNote.pm
reosarevok: AFAICT it's just undefined if empty. I tried to enter a string, then remove it, then submit again, and I still get the "You must provide an edit note" message. (I'd love to add a period to that string but breaking the translation for it feels silly).

mwiencek: I see, seems to be documented at https://metacpan.org/pod/HTML::FormHandler::Field#not_nullable

lib/MusicBrainz/Server/Validation.pm
```diff
@@ -310,6 +311,24 @@ sub is_valid_partial_date
     return 1;
 }
 
+sub is_valid_edit_note
+{
+    my $edit_note = shift;
+
+    # An edit note with only spaces and / or punctuation is useless
+    if ($edit_note =~ /^[[:space:][:punct:]]+$/) {
+        return 0
```
mwiencek: Nitpick: a lack of semicolons after the returns, though in-lining them like `return 0 if ...` seems more appropriate anyway.

reosarevok: Seems reasonable. Done!

PR #1165, opened 2019-08-10
MBS-6532: Show work edits in recording / release history

reosarevok:
https://tickets.metabrainz.org/browse/MBS-6532

Most work edits are being made based on data from releases, so it seems reasonable for those edits to be shown on the release edit list. Since the connection between release and work is through a recording, it makes sense to show the edits on the recording list as well.


lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
```diff
@@ -17,15 +17,25 @@ around _build_related_entities => sub {
...
+        my ($recordings, $hits) = $self->c->model('Recording')->find_by_work(@work_ids);
+        my @work_recording_ids = map { $_->id } @$recordings;
+        push @{ $direct->{recording} }, @work_recording_ids;
+
 
         # For recordings, we want relationship edits to also appear on the releases they're on
         my @recording_ids = map { $_->id } grep { $_->isa('MusicBrainz::Server::Entity::Recording') } @entities;
-        my ($releases, $hits) = $self->c->model('Release')->find_by_recording(@recording_ids);
+        push @recording_ids, @work_recording_ids;
+        my ($releases, $hits) = $self->c->model('Release')->find_by_recording(\@recording_ids);
```
reosarevok: I'm honestly not sure why this seems to not work unless I use \@ now. Ideas? :)

mwiencek: The old code was broken, from what I can see. It used the first recording id in the list, but then used the second id as `$limit`, the third id as `$offset`, and the rest as `%args`. So it only found releases for the first recording and had random numbers for the limit/offset if there were more than one. The new code looks fine (apparently `find_by_recording` already accepts multiple recording ids in an array ref, despite its name).

reosarevok:
Made the changes, seems to work locally at least, but please make sure I didn't do a dumb again (since the last time the missing \ got through review).

lib/MusicBrainz/Server/Edit/Work/RelatedEntities.pm
```diff
@@ -13,12 +13,29 @@ around '_build_related_entities' => sub
         $self->c->model('Work')->get_by_ids($self->work_ids)
     };
 
+    my ($recordings, $recording_hits) = $self->c->model('Recording')->find_by_works($self->work_ids);
```
reosarevok: Do I need to do anything to avoid taking over offset etc here? If so, what?

lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
```diff
@@ -17,15 +17,25 @@ around _build_related_entities => sub {
         my $model = type_to_model($type);
         my @ids = @{ $direct->{$type} };
         my @entities = values %{ $self->c->model($model)->get_by_any_ids(@ids) };
-        $self->c->model('ArtistCredit')->load(@entities);
-        $direct->{artist} ||= [];
-        push @{ $direct->{artist} }, map { $_->artist_id }
-            map { $_->artist_credit->all_names }
-                @entities;
+        if ($type ~~ qw( recording release release_group )) {
```
mwiencek: Might be a good use for `entities_with`?

lib/MusicBrainz/Server/Edit/Work/RelatedEntities.pm
mwiencek: (This currently works since `work_ids` just returns one ID, and `find_by_works` accepts that, but assuming the above change)
```suggestion
    my ($recordings, $recording_hits) = $self->c->model('Recording')->find_by_works([$self->work_ids]);
```

lib/MusicBrainz/Server/Edit/Work/RelatedEntities.pm
```diff
@@ -13,12 +13,29 @@ around '_build_related_entities' => sub
...
+    my @recordings = @$recordings;
+
+    my @recording_ids = map { $_->id } @recordings;
```
mwiencek: You can remove the `my @recordings` declaration above and just do
```suggestion
    my @recording_ids = map { $_->id } @$recordings;
```
since this is the only place it needs to be dereferenced.

lib/MusicBrainz/Server/Edit/Work/RelatedEntities.pm
```diff
@@ -13,12 +13,29 @@ around '_build_related_entities' => sub
...
+
+    my @releases;
+    if (@recording_ids) {
+        my ($releases, $release_hits) = $self->c->model('Release')->find_by_recording(\@recording_ids);
+        @releases = @$releases;
+    }
```
mwiencek: Just the `my ($releases, $release_hits) = ...` line seems necessary, since `find_by_recording` already returns early if the list is empty. Avoiding the function call by checking `if (@recording_ids)` is nice, but on the other hand it'd be simpler if we didn't have to declare `my @releases` and copy the result.

lib/MusicBrainz/Server/Edit/Work/RelatedEntities.pm
```diff
@@ -13,12 +13,29 @@ around '_build_related_entities' => sub
...
 
     return {
         artist => [
             map { $_->entity0_id } map { $_->all_relationships } @works
         ],
+        recording => [
+            @recording_ids
+        ],
+        release => [
+            map { $_->id } @releases
```
mwiencek: (Assuming the above change)
```suggestion
            map { $_->id } @$releases
```

lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
```diff
@@ -17,15 +18,25 @@ around _build_related_entities => sub {
         my $model = type_to_model($type);
         my @ids = @{ $direct->{$type} };
         my @entities = values %{ $self->c->model($model)->get_by_any_ids(@ids) };
-        $self->c->model('ArtistCredit')->load(@entities);
-        $direct->{artist} ||= [];
-        push @{ $direct->{artist} }, map { $_->artist_id }
-            map { $_->artist_credit->all_names }
-                @entities;
+        if ($type ~~ entities_with('artist_credits')) {
```
mwiencek, line 21: This doesn't seem to work. I'd honestly recommend to avoid ever using `~~`, since it's experimental and has complicated semantics. :( But luckily we don't need anything fancy for this check:
```suggestion
        if ($ENTITIES{$type}{artist_credits}) {
```

lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
```diff
@@ -2,11 +2,12 @@ package MusicBrainz::Server::Edit::Relationship::RelatedEntities;
 use Moose::Role;
 use namespace::autoclean;
 
+use MusicBrainz::Server::Constants qw( entities_with );
```
mwiencek, line 5:
```suggestion
use MusicBrainz::Server::Constants qw( %ENTITIES );
```

PR #1167, opened 2019-08-12
MBS-4478: Improve messages for edits that create entities

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4478

When editing a medium and using a new recording, the current tooltip claims that the entity will be created once the edit is entered (both in the preview and... once the edit is entered). This is not really true: it is created once the edit is *applied*. Tooltip changed accordingly.

Additionally, it doesn't really make a difference when a recording in the new column has been removed or is being created, but we do have that information (the former has a recording_id but the latter does not), so I'm changing the code to actually check what is the case.

There's a third issue where the "will be created" message is never replaced by a link to the actual recording (MBS-7814), but that's a much less straightforward fix, so let's start with this.

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -264,7 +264,7 @@ END -%]
       SET text = html_escape(entity.name) IF text == '';
       SET text = html_escape(l('[removed]')) IF text == '';
       caption = allow_new
-        ? l("This entity will be created when edits are entered.")
+        ? l("This entity will be created by this edit.")
         : l("This entity has been removed, and cannot be displayed correctly.");
       '<span class="' _ (!allow_new && ('deleted' _ ' ')) _ 'tooltip" title="' _ html_escape(caption) _ '">' _
         '<bdi>' _ text _ '</bdi>' _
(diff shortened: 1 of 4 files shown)


root/components/common-macros.tt
```diff
@@ -264,7 +264,7 @@ END -%]
       SET text = html_escape(entity.name) IF text == '';
       SET text = html_escape(l('[removed]')) IF text == '';
       caption = allow_new
-        ? l("This entity will be created when edits are entered.")
```
mwiencek: It appears this message does make sense if you're adding a new medium with new recordings: even if you make the add-medium edit votable, the recordings are created immediately when the edit is entered.

But if you're *editing* a medium to use a new recording, it's not created until the edit is applied, so the message below makes more sense.

reosarevok: Hmm. "This entity will be created during the editing process" is probably more precise, then, but it still suggests "when entering the edit" to me. Any ideas?

mwiencek: "This entity is pending creation with the edit?" Or we can have two separate messages controlled with `allow_new_before_entered` and `allow_new_before_applied` if we want to keep the distinction.

reosarevok: Updated the wording to be more general.

PR #1168, opened 2019-08-15
MBS-10316: Add "Artists" tab for instruments

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10316

On top of (and to be checked after) https://github.com/metabrainz/musicbrainz-server/pull/1134

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -119,6 +119,46 @@ sub find_by_area {
     );
 }
 
+sub find_by_instrument {
+    my ($self, $instrument_id, $limit, $offset) = @_;
+
+    my $query =
+        "SELECT " . $self->_columns . ",
+                array_agg(json_build_object('typeName', link_type.name, 'credit', lac.credited_as)) AS instrument_credits_and_rel_types
+            FROM " . $self->_table . "
+                JOIN (
+                    SELECT * FROM l_artist_artist
+                    UNION ALL
+                    SELECT * FROM l_artist_recording
+                    UNION ALL
+                    SELECT * FROM l_artist_release
+                ) rels ON rels.entity0 = artist.id
+                JOIN link ON link.id = rels.link
+                JOIN link_type ON link_type.id = link.link_type
+                JOIN link_attribute ON link_attribute.link = link.id
+                JOIN link_attribute_type ON link_attribute_type.id = link_attribute.attribute_type
+                JOIN instrument ON instrument.gid = link_attribute_type.gid
+                LEFT JOIN link_attribute_credit lac ON (
+                    lac.link = link_attribute.link AND
+                    lac.attribute_type = link_attribute.attribute_type
+                )
+            WHERE instrument.id = ?
+            GROUP BY artist.id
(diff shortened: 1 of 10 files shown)


root/components/InstrumentRelTypes.js
```diff
@@ -30,7 +30,10 @@ const InstrumentRelTypes = ({
             let listElement = l_relationships(relType.name);
             if (relType.credit) {
               listElement = listElement + ' ' +
-                bracketedText(texp.l('as “{credit}”', {credit: relType.credit}));
+                bracketedText(texp.l(
+                  'as “{credit}”',
+                  {credit: relType.credit},
+                ));
```
yvanzo: Related to this silent eslint change, `/artists` gives me:
```
Couldn't render template "undef error - Error: No translation key found.
    at Jed.dcnpgettext (/musicbrainz-server/node_modules/jed/jed.js:169:13)
    at Jed.dgettext (/musicbrainz-server/node_modules/jed/jed.js:85:16)
    at eval (webpack:///./root/static/scripts/common/i18n/wrapGettext.js?:56:20)
    at eval (webpack:///./root/components/InstrumentRelTypes.js?:23:23)
    at Array.map (<anonymous>)
    at InstrumentRelTypes (webpack:///./root/components/InstrumentRelTypes.js?:21:290)
    ...
```
even after rebuilding `po/mb_server.pot` and recompiling resources.

reosarevok: Not related to eslint, just to me being an idiot (the line breaking was the one above, because I was trying to check relType.name but the actual object had relType.role - fixed).

yvanzo: Thank you, it works for `/artists` now but all other tabs started to give that error.

reosarevok: Sigh, I'm dumb. Ok, should work now.

yvanzo: Thank you, still silent coding style change isn’t a good thing, please avoid these.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -119,6 +119,46 @@ sub find_by_area {
     );
 }
 
+sub find_by_instrument {
+    my ($self, $instrument_id, $limit, $offset) = @_;
+
+    # NOTE: if more tables than l_artist_recording are added here, check admin/BuildSitemaps.pl
```
mwiencek: This comment looks wrong/irrelevant here, so should be removed (we select from more than l_artist_recording below, and since this is a new page/has no JSON-LD markup, it wouldn't be in the sitemaps listing anyway).

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -119,6 +119,46 @@ sub find_by_area {
...
+    my $query = "SELECT " . $self->_columns . ",
+                     array_agg(json_build_object('typeName', link_type.name, 'credit', lac.credited_as)) AS instrument_credits_and_rel_types
+                 FROM " . $self->_table . "
```
mwiencek: Just a small thing, but to decrease the amount of indentation here we could start the `"SELECT` on the next line at 8 spaces. Might help with the long lines a bit.

PR #1170, opened 2019-08-16
MBS-4486: Remove parentheses around disambiguation

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4486

Disambiguation is already presented in parentheses everywhere, but people sometimes add it *in* parentheses,leading to stuff being shown as ((disambiguation)). This drops parentheses if the disambiguation comment both starts and ends with them.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -70,6 +70,7 @@ our @EXPORT_OK = qw(
     sanitize
     take_while
     trim
+    trim_comment
     type_to_model
     split_relationship_by_attributes
 );
@@ -339,6 +340,14 @@ sub trim {
     $t = Text::Trim::trim($t);
 
     return $t;
+}
+
+sub trim_comment {
+    my $t = shift;
+
+    $t =~ s/^\s*\(([^()]+)\)\s*$/$1/;
+
+    return trim($t);
 }
 
 sub remove_direction_marks {
--- a/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
@@ -37,6 +37,7 @@ use MusicBrainz::Server::Data::Utils qw(
     split_relationship_by_attributes
     sanitize
     trim
+    trim_comment
     non_empty
 );
 use MusicBrainz::Server::Renderer qw( render_component );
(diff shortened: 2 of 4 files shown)


reosarevok:
This works, but it works silently, which might be confusing. An option would be to use JS to remove parentheses on blur, although that would still not show anything if the editor doesn't click on the field at all, but the parentheses already there will be dropped. Open to other suggestions as well.

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -341,6 +342,14 @@ sub trim {
     return $t;
 }
 
+sub trim_comment {
+    my $t = shift;
+
+    $t =~ s/^\s*\((.+)\)\s*$/$1/;
```
yvanzo: It should probably trim spaces within parentheses, and unpaired parentheses too:
```suggestion
    $t =~ s/^[\s\(]*(.+)[\)\s]*$/$1/;
```

reosarevok: If what you mean with unpaired parentheses is stuff like "whatever)", then it shouldn't because that'd also break a valid disambiguation like "some stuff (more stuff)".

reosarevok: Once the parentheses are removed, then this does trim($t), so any newly-trimmable whitespace will still get trimmed.

reosarevok:
I added a line about this to the disambiguation bubble in the release editor. We really should extend the bubbles to the rest of the forms, but I guess for now it's fine to leave the others drop the parens silently and hope that they all get the bubble eventually.

yvanzo:
I’m not sure the added line is really necessary, it looks to be more helpful to developer than user.

(About using bubble everywhere and such, we should really move to React and Bootstrap first.)

mwiencek:
> I’m not sure the added line is really necessary

I feel the same. I can't imagine anyone being confused or even noticing it being dropped silently, but I could imagine someone being confused by the added text (even though it makes sense to us, it makes the system appear more complicated to the user than is necessary IMO).

reosarevok:
My worry is in a case where the user has, say, "(this stuff) (that stuff)" and it becomes "this stuff) (that stuff". Of course, maybe the solution is just to not drop the parens if there's more than a set (if you think that'd be better, maybe you can help with the regex? :) )

mwiencek:
I suppose you can replace the `(.+)` in the regexp with `([^()]+)`. That won't detect if they're balanced, but it should work fine in the general case.

reosarevok:
Changed the regex

PR #1171, opened 2019-08-16
MBS-8328 / MBS-13917: Limit work authors to authorship rels

reosarevok:
https://tickets.metabrainz.org/browse/MBS-8328

Work writers are currently calculated as "every single artist-work rel". This worked when we came up with it, but doesn't anymore, since we've added several relationships that aren't really connected to writing (such as "publisher", "commissioned by" or "dedicated to"). These shouldn't appear on the writers list, so I've hardcoded the relationships that should be considered writing-related in Constants.


reosarevok:
That plus "previous attribution", as the comment on https://github.com/metabrainz/musicbrainz-server/pull/1171/files#diff-e11edc95d90abe8b020956c678bb56afR390 says :) I thought about that, but then thought we might add more non-composition ones that we might still want here, or add some composition ones we don't want here.

reosarevok:
> would show less unrelated relationships in the column “writers” but would still show many

What am I missing? :) Every relationship there is directly related to writing (except previous attribution, which is indirectly so, as in "it was believed they wrote it").

yvanzo:
1. Not “every relationship there is directly related to writing”, e.g. [arranger](https://musicbrainz.org/relationship/d3fd781c-5894-47e2-8c12-86cc0e2c8d08), a sibling of [writer](https://musicbrainz.org/relationship/a255bca1-b157-4518-9108-7b147dc3fc68), not a descendent, is still being added: https://github.com/metabrainz/musicbrainz-server/pull/1171/commits/533917499ba61c800977b0c2116f1ce5ab7a8672#diff-e11edc95d90abe8b020956c678bb56afR399
2. Your previous comment suggested it would not be always match writing:
   > we might add more non-composition ones that we might still want here, or add some composition ones we don't want here

reosarevok:
But if I'm looking at a list of writers in a search, arranger is important. It's certainly connected to the information I need to identify this (arrangement) work :) Any new ones that might be added should also be stuff you would expect to find under "writers"

yvanzo:
> But if I'm looking at a list of writers in a search, arranger is important. It's certainly connected to the information I need to identify this (arrangement) work :)

I’m sure it can be important but the column should **not** be named “writers” then.

> Any new ones that might be added should also be stuff you would expect to find under "writers"

I don’t expect to find anything under “writers” but [writer](https://musicbrainz.org/relationship/a255bca1-b157-4518-9108-7b147dc3fc68) relationships and children (in [rels type tree](https://musicbrainz.org/relationships/artist-work)).

Please let’s continue this conversation in comment to the ticket if you agree this is a design discussion beyond code review and not just my misunderstanding.

reosarevok:
How we organise the rels in the tree has nothing to do with this. If you're so bothered I can move arrangements under writer so that it fits your definition, but I'm not sure where this obsession with only those few rels being writers comes from.

reosarevok:
I don't really agree this is a design discussion beyond code review: IMO all writing-related relationships are under the "composition" header, not the "writer" one (writer is a specific type of writing-related role). The only thing I could see is moving "previous attribution" under "composition" and automatically taking all the composition rels - I guess it wouldn't be wrong to say that since previous attribution is meant as "was previously thought to have something to do with writing this", it could be put under "composition" :) If we decided to do this, we could I guess just ask for the list of composition rels by querying the DB for all descendants of "composition". Although I'm not sure running a many-level recursive query is better than hardcoding in this case, but it might be :)

mwiencek:
There might be some confusion over which pages this affects (see my comment on the ticket).

This appears to fix a bug to me. The writers list previously displayed non-writing relationships, now it only displays ones related to writing.

I'm not very knowledgeable about composition, but it didn't seem odd to me to list arrangers under writers, since they can change the composition. But I guess that's not always strictly true so I don't have an opinion on where it belongs in the tree.

If just the name of the column is at issue, I can't personally think of a better one, so I'm not sure how useful I'll be there. Some fuzziness with the terminology might be unavoidable? e.g. If you move everything under "composition" in the tree and get them that way, it might seem logical to name the column "Composers" instead, but we have a more specific relationship named "composer," etc.

reosarevok:
I changed this in the last commit to have three columns as we talked about on IRC at some point, FWIW. @yvanzo, do let me know if you're happy with the way it works now.

mwiencek:
This still looks good to me, and seems to match what we previously agreed upon in our discussion.  The only possible improvement I can see might be to hide the "Other artists" column if it's empty for every row.

 We have at least one other PR blocked on this (#1277), and another (#1708) which I suggested would benefit from the use of `@WRITER_RELATIONSHIP_GIDS`.

reosarevok:
Updated this to not use placeholders() and to follow our draft SQL guidelines.

reosarevok:
Added a set of tests here while at it, for all three subsets of work artists.

reosarevok:
I renamed the "composition" holding rel to "authorship" (on the site) and changed our uses of "writer" to "author" when meant in the more general concept of "did the writing, arranging and whatnot for a work". Since "previous attribution" is now under "authorship", I just load the list of authorship rels from the DB by checking reltype roots rather than hardcoding it.

I considered dropping the misc artists column from searches, but I'm not sure if it's worth adding extra complication in `WorkListEntry` for that - and sometimes these rels can be of use in the search tables. 

I also extended the filter option in the artist works page (added in the meantime) to support the new misc option.

lib/MusicBrainz/Server/Form/Filter/Work.pm
```diff
@@ -37,7 +37,8 @@ sub filter_field_names {
 sub options_role_type {
     return [
         { value => 1, label => l('As performer') },
-        { value => 2, label => l('As writer') },
+        { value => 2, label => l('As author') },
+        { value => 3, label => l('As something else') },
```
mwiencek, line 41: Wouldn't it make sense for these to match the column names if that's what they're filtering by? Right now only "author" matches.

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -405,46 +416,78 @@ sub find_artists
...
+    $self->_find_authors_or_misc_artists(\@ids, \%map);
     for my $work (@works) {
-        $work->add_writer(@{ $map{$work->id} })
+        $work->add_author(@{ $map{$work->id} })
             if exists $map{$work->id};
     }
 }
 
-sub _find_writers
+sub load_misc_artists
```
mwiencek, line 442: Same here, it seems we call them "other artists" in the UI and "misc. artists" in the code (or maybe I'm conflating two different things).

mwiencek:
Tested this locally and it seems to work well for me, as well as addresses the previous feedback from @yvanzo. I did push one commit to simplify `get_authorship_relationship_gids` (let me know if that looks OK), and pointed out what look like a couple inconsistencies in the labels we use.

lib/MusicBrainz/Server/Data/Work.pm
reosarevok: Hmm. I agree it would make sense to have everything use the same name. I just don't quite have a good one - but actually I don't like misc. artists because on second thought it doesn't make it clear at all it is not just "whatever" but "things other than authors and recording artists". So I might change the code to other_artists as well.

reosarevok:
Your commit looks and works fine, thanks!

mwiencek (approved):
Thanks!

PR #1172, opened 2019-08-19
MBS-10320: Only nag user if in prod or beta site

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10320

There's little reason to nag users in slave servers, test, and other non-core MB installs. The users are either already supporters, just testing for bugs (which is also a way of supporting us), or at the very least know enough to set up their own server, which also should mean they know we'd like them to donate without nagging them.
In many of these servers, there's no way to know if the user has donated or not, so we would risk annoying a user who has already donated by nagging them again locally.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/TagLookup.pm
+++ b/lib/MusicBrainz/Server/Controller/TagLookup.pm
@@ -83,7 +83,7 @@ sub nag_check
     my ($self, $c) = @_;
 
     # Never nag users in non-core MB servers
-    return 0 unless DBDefs->WEB_SERVER =~ /^(beta\.)?musicbrainz\.org$/;
+    return 0 unless DBDefs->WEB_SERVER =~ /^(?:beta\.)?musicbrainz\.org$/;
 
     # Always nag users who are not logged in
     return 1 unless $c->user_exists;
--- a/lib/MusicBrainz/Server/Controller/TagLookup.pm
+++ b/lib/MusicBrainz/Server/Controller/TagLookup.pm
@@ -82,6 +82,9 @@ sub nag_check
 {
     my ($self, $c) = @_;
 
+    # Never nag users in non-core MB servers
+    return 0 unless DBDefs->WEB_SERVER =~ /^(beta\.)?musicbrainz\.org$/;
+
     # Always nag users who are not logged in
     return 1 unless $c->user_exists;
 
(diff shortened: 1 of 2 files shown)


root/taglookup/Index.js
```diff
@@ -10,19 +10,24 @@
...
-  </Layout>
-);
+const TagLookup = (props: TagLookupPropsT) => {
+  const musicbrainzRegex = /^(beta\.)?musicbrainz\.org$/;
+  const isMusicBrainzServer = WEB_SERVER.match(musicbrainzRegex);
+  return (
+    <Layout fullWidth title={l('Tag Lookup')}>
+      <div className="content">
+        <h1>{l('Tag Lookup')}</h1>
+        {props.nag && isMusicBrainzServer ? <TagLookupNagSection /> : null}
```
mwiencek: Hmm, since `props.nag` is calculated on the server, perhaps we should just change the calculation there instead of adding additional logic here.

i.e. changing `sub nag_check` in lib/MusicBrainz/Server/Controller/TagLookup.pm to `return 0` early if `DBDefs->WEB_SERVER` matches? Then the component here wouldn't need modification at all.

reosarevok: That seems like a good idea. Changed.

reosarevok:
So, the test is failing because the test environment doesn't have the right WEB_SERVER in DBDefs, so the nag is not shown. Options:

1. Also check $ENV{MUSICBRAINZ_RUNNING_TESTS} in the sub nag_check. Kinda gross, plus it kinda only tests that it works during testing in that case.
2. Just remove the nag test (probably the easiest, but we might want to still test for it).
3. Try and find a way to temporarily override the value of WEB_SERVER for the test (would be ideal, since it would let us test both with and without the specific value to make sure the nag only appears when it should, but I don't know if it's even doable). 

What do you think? :)

mwiencek:
There's a way to do 3 (just wasn't sure it'd work at first, 'cause I forgot if the requests are handled in the same process).

So add this before it fetches the page in the test, maybe with a comment explaining why:

```perl
    no warnings qw( redefine );
    use DBDefs;
    local *DBDefs::WEB_SERVER = sub { 'musicbrainz.org' };
```

reosarevok:
Thanks! Did that. I guess there's no good way to check the nag is *not* shown with other WEB_SERVER, other than making another full request, but that's probably less important.

lib/MusicBrainz/Server/Controller/TagLookup.pm
```diff
@@ -82,6 +82,9 @@ sub nag_check
 {
     my ($self, $c) = @_;
 
+    # Never nag users in non-core MB servers
+    return 0 unless DBDefs->WEB_SERVER =~ /^(beta\.)?musicbrainz\.org$/;
```
yvanzo: Micro-optimization: `(beta\.)` group doesn’t need to be captured for later use:

```suggestion
    return 0 unless DBDefs->WEB_SERVER =~ /^(?:beta\.)?musicbrainz\.org$/;
```

PR #1173, opened 2019-08-20
MBS-10299: Add page with suggested edit searches

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10299

This currently includes three sets of links: one for destructive edits, one for unreviewed / controversial edits, and one for edits by beginners.

Once [MBS-7912](https://tickets.metabrainz.org/browse/MBS-7912) is available, it should be possible to also add a filter for these (when relevant, at least) to let the user see edits related to their suscriptions only.

### DIFF
--- a/root/layout/components/BottomMenu.js
+++ b/root/layout/components/BottomMenu.js
@@ -252,7 +252,7 @@ const EditingMenu = () => (
         <a href="/event/create">{lp('Add Event', 'button/menu')}</a>
       </li>
       <li className="separator">
-        <a href="/edit/open">{l('Vote on Edits')}</a>
+        <a href="/vote">{l('Vote on Edits')}</a>
       </li>
       <li>
         <a href="/reports">{l('Reports')}</a>
--- a/root/vote/VotingIndex.js
+++ b/root/vote/VotingIndex.js
@@ -1,0 +1,381 @@
+/*
+ * @flow
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import {withCatalystContext} from '../context';
+import Layout from '../layout';
+
+type VotingGuideRowProps = {
+  +guideName: string,
+  +mainUrl: string,
+  +showSubscribedArtistsUrl?: boolean,
+};
+
+const VotingGuideRow = ({
+  guideName,
+  mainUrl,
+  showSubscribedArtistsUrl,
+}: VotingGuideRowProps) => {
+  const subscribedArtistsCondition =
+    '&conditions.9.field=artist&conditions.9.operator=subscribed';
+  return (
+    <li>
+      <a href={mainUrl}>
+        {guideName}
+      </a>
+      {showSubscribedArtistsUrl ? (
+        <ul>
+          <li>
+            <a href={mainUrl + subscribedArtistsCondition}>
+              {l('…related to artists in my subscriptions')}
+            </a>
+          </li>
+        </ul>
+      ) : null}
+    </li>
+  );
+};
+
+const VotingIndex = () => (
+  <Layout fullWidth title={l('Voting suggestions')}>
+    <div id="content">
+      <h1>{l('Voting suggestions')}</h1>
+
+      <p>
+        {exp.l(
+          `If you’d like to help ensuring the changes made to MusicBrainz data
+           are correct, but do not know where to start, the following
+           suggestions should be useful. These are pre-defined
+           {edit_search|edit searches}; once you’re comfortable with the edit
+           search you can refine them further or just create your own personal
+           searches and bookmark them for your own use!`,
+          {edit_search: '/search/edits'},
+        )}
+      </p>
+      <p>
+        {exp.l(
+          `While reviewing the work of your fellow editors, always keep the
+           {coc|Code of Conduct} in mind. Almost all editors want to help, so
+           your goal as a voter is to help them help better. This applies even
+           more for beginners: always try to be helpful and patient with them,
+           even if they are making mistakes, so that they’ll hopefully grow
+           into better editors! That said, if you find an editor that seems
+           to be vandalizing the data, you can always report them
+           from their profile.`,
+          {coc: '/doc/Code_of_Conduct'},
+        )}
+      </p>
+      <p>
+        {l(
+          `By default, these searches skip your own edits and edits you have
+           already voted on (when relevant). To change that, load the search
+           and then remove the conditions “Editor is not me” and “Voter is me
+           and voted No vote”, respectively.`,
+        )}
+      </p>
+
+      <h2>{l('Destructive edits')}</h2>
+      <p>
+        {l(
+          `Destructive edits (removals and merges) are often very hard or even
+           impossible to revert. As such, an incorrect destructive edit that
+           applies unnoticed can cause quite a big mess! Most are guaranteed
+           to remain open for at least two full days even if they get three
+           “Yes” votes, to avoid them closing too quickly, but it’s always
+           good to get more eyes on them. Below you can find four different
+           searches: one for all destructive edits (which might be
+           overwhelming sometimes), one for entity merges and removals only
+           (the edits more likely to cause a mess if they incorrectly go
+           through), one for relationship removals only, and one for
+           destructive changes to releases (track, medium and release label
+           removals).`,
+        )}
+      </p>
+      <ul>
+        {/* eslint-disable max-len */}
+        <VotingGuideRow
+          guideName={l('All open destructive edits')}
+          mainUrl={'/search/edits?' +
+            'conditions.0.field=type&conditions.0.operator=%3D&conditions.0.args=9&conditions.0.args=84&conditions.0.args=4&conditions.0.args=153&conditions.0.args=134&conditions.0.args=14&conditions.0.args=64&conditions.0.args=74&conditions.0.args=24&conditions.0.args=225%2C223%2C311&conditions.0.args=143&conditions.0.args=44&conditions.0.args=83&conditions.0.args=3&conditions.0.args=315&conditions.0.args=152&conditions.0.args=133&conditions.0.args=78&conditions.0.args=410&conditions.0.args=13&conditions.0.args=53&conditions.0.args=63&conditions.0.args=73&conditions.0.args=23&conditions.0.args=36&conditions.0.args=224&conditions.0.args=142&conditions.0.args=211&conditions.0.args=43&conditions.0.args=47&' +
+            'conditions.1.field=status&conditions.1.operator=%3D&conditions.1.args=1&' +
+            'conditions.2.field=editor&conditions.2.operator=not_me&conditions.2.name=&conditions.2.args.0=&' +
+            'conditions.3.field=voter&conditions.3.operator=me&conditions.3.name=&conditions.3.voter_id=&conditions.3.args=no'}
+          showSubscribedArtistsUrl
+        />
+        <VotingGuideRow
+          guideName={l('All open entity merges and removals')}
+          mainUrl={'/search/edits?' +
+            'conditions.0.field=type&conditions.0.operator=%3D&conditions.0.args=84&conditions.0.args=4&conditions.0.args=153&conditions.0.args=134&conditions.0.args=14&conditions.0.args=64&conditions.0.args=74&conditions.0.args=24&conditions.0.args=143&conditions.0.args=44&conditions.0.args=83&conditions.0.args=3&conditions.0.args=152&conditions.0.args=133&conditions.0.args=13&conditions.0.args=63&conditions.0.args=73&conditions.0.args=310%2C212&conditions.0.args=23&conditions.0.args=142&conditions.0.args=43&' +
+            'conditions.1.field=status&conditions.1.operator=%3D&conditions.1.args=1&' +
+            'conditions.2.field=editor&conditions.2.operator=not_me&conditions.2.name=&conditions.2.args.0=&' +
+            'conditions.3.field=voter&conditions.3.operator=me&conditions.3.name=&conditions.3.voter_id=&conditions.3.args=no'}
+          showSubscribedArtistsUrl
+        />
+        <VotingGuideRow
+          guideName={l('All open relationship removals')}
+          mainUrl={'/search/edits?' +
+            'conditions.0.field=type&conditions.0.operator=%3D&conditions.0.args=92%2C235&' +
+            'conditions.1.field=status&conditions.1.operator=%3D&conditions.1.args=1&' +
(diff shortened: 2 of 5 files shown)


root/report/VotingReportsIndex.js
```diff
@@ -0,0 +1,363 @@
...
+           removals only, and one for destructive changes to releases (track,
+           medium and release label removals).`,
+        )}
+      </p>
+      <ul>
+        <VotingReportRow
+          noOwnEditsUrl="/search/edits?auto_edit_filter=&order=asc&negation=0&combinator=and&conditions.0.field=type&conditions.0.operator=%3D&conditions.0.args=9&conditions.0.args=84&conditions.0.args=4&conditions.0.args=153&conditions.0.args=134&conditions.0.args=14&conditions.0.args=64&conditions.0.args=74&conditions.0.args=24&conditions.0.args=225%2C223%2C311&conditions.0.args=143&conditions.0.args=44&conditions.0.args=83&conditions.0.args=3&conditions.0.args=315&conditions.0.args=152&conditions.0.args=133&conditions.0.args=78&conditions.0.args=410&conditions.0.args=13&conditions.0.args=53&conditions.0.args=63&conditions.0.args=73&conditions.0.args=23&conditions.0.args=36&conditions.0.args=224&conditions.0.args=142&conditions.0.args=211&conditions.0.args=43&conditions.0.args=47&conditions.1.field=status&conditions.1.operator=%3D&conditions.1.args=1&conditions.2.field=editor&conditions.2.operator=not_me&conditions.2.name=&conditions.2.args.0="
```
mwiencek: I'd mainly just like to see something done about these URLs if possible. I guess splitting them all across multiple lines would also be horrible and make the file 10,000 lines long, but...maybe there's an alternative. For example, all the `noOwnEditsUrl` URLs probably have a common set of conditions and other parameters to filter out own edits. Perhaps you could extract the common parts of these different URLs into constants, and concatenate them where needed? (So any common conditions included in a constant would have to start with `conditions.0` and end at `conditions.n` for some `n`, and whatever strings use the constant would have to have their other conditions all start at `n + 1`.) This also sounds really tedious though and perhaps not a good use of dev time. So if you don't think any of that is worth the effort I guess we can include it as-is.

reosarevok: I hate the URLs here as much as you do, but tbh I think this should be fixed by fixing our stupid edit search URLs. If you have any idea how to make them less absurdly long, I'd be interested :p Your suggestion would make a lot of sense to me if it wasn't because of the condition numbering, but once it starts requiring modifying the constant to allow for changing condition numberings, it stops sounding like it saves effort / helps readability...

root/report/VotingReportsIndex.js
```diff
@@ -0,0 +1,363 @@
...
+            {l('…by editors other than me')}
+          </a>
+          {noVoteYetUrl ? (
+            <ul>
+              <li>
+                <a href={noVoteYetUrl}>
+                  {l('…that I haven’t voted on yet')}
+                </a>
+              </li>
+            </ul>
```
mwiencek: This `ul` should be merged with the one below, since they're at the same level and part of the same semantic list. It wouldn't look visually different, but def. creates a confusing page structure for screen readers in particular.

There appear to be similar cases below.

root/report/VotingReportsIndex.js
mwiencek: Yeah, perhaps we should add some shortcut parameters for common conditions, like "me_no_vote=1" (expect a better name and example, lol). But something we can improve on later.

mwiencek:
Maybe we should default to having all links exclude edits you've voted on, to reduce the number of URL permutations? I expect that'd be the most common use case, and it's quite simple to remove that condition if you want to see ones you already voted on (we could even mention that on the page).

reosarevok:
That seems sensible. Made the changes, which make for a much less scary-looking page, and added a note about it at the end of the introduction.

mwiencek:
This looks great. :) One small thing I noticed in the text is that we use the phrase "No votes" a lot, but this might be confused to mean either "has votes against" ("No" votes) or "has zero votes" (no votes). Maybe we can keep the "No" in quotes when referencing it as a type of vote?

reosarevok:
Did that (for “Yes”, “No” and “Abstain”, since it'd look super weird otherwise).

root/report/VotingReportsIndex.js
yvanzo: Maybe it would be more readable to split by conditions at least, for this example:
```suggestion
          noOwnEditsUrl="/search/edits?" +
          "conditions.0.field=type&conditions.0.operator=%3D&conditions.0.args=9&conditions.0.args=84&conditions.0.args=4&conditions.0.args=153&conditions.0.args=134&conditions.0.args=14&conditions.0.args=64&conditions.0.args=74&conditions.0.args=24&conditions.0.args=225%2C223%2C311&conditions.0.args=143&conditions.0.args=44&conditions.0.args=83&conditions.0.args=3&conditions.0.args=315&conditions.0.args=152&conditions.0.args=133&conditions.0.args=78&conditions.0.args=410&conditions.0.args=13&conditions.0.args=53&conditions.0.args=63&conditions.0.args=73&conditions.0.args=23&conditions.0.args=36&conditions.0.args=224&conditions.0.args=142&conditions.0.args=211&conditions.0.args=43&conditions.0.args=47&" +
          "conditions.1.field=status&conditions.1.operator=%3D&conditions.1.args=1&" +
          "conditions.2.field=editor&conditions.2.operator=not_me"
```

I also removed unneeded defaults.

root/layout/components/BottomMenu.js
```diff
@@ -251,6 +251,9 @@ const EditingMenu = () => (
       <li>
         <a href="/reports">{l('Reports')}</a>
       </li>
+      <li>
+        <a href="/voting-reports">{l('Voting Reports')}</a>
```
yvanzo: Maybe it would be simpler to just replace the above _Vote on Edits_? Both have the same purpose. The new page is just more sophisticated.

yvanzo: Also, _Voting Reports_ might be misleading as these are searches rather than daily built reports like (data) _Reports_.

reosarevok: I guess it might make sense - at the very least we should rename "Vote on Edits" to "All Open Edits" or something, because right now it's actually kind of weird. But I'd be fine with replacing the link too - we might just have to add an "All open edits" link to this page.

reosarevok:
I changed this to not use the word "Report". I also replaced the open edits link in the menu, and added both "all open edits" and "all edits" to the bottom of the page (I used an edit search for open edits rather than edit/open because that's easier to refine later).

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,465 @@
...
+
+import * as React from 'react';
+
+import {withCatalystContext} from '../context';
+import Layout from '../layout';
+
+type VotingGuideRowProps = {
+  +guideName: string,
+  +mainUrl: string,
+  +subscribedArtistsUrl?: string,
```
yvanzo: That might be replaced with a flag, as the URL for subscribed artists can be generated by appending `&conditions.9.field=artist&conditions.9.operator=subscribed` to the main URL.

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+};
+
+const VotingIndex = () => (
+  <Layout fullWidth title={l('Voting reports')}>
+    <div id="content">
+      <h1>{l('Voting reports')}</h1>
+
+      <p>
+        {exp.l(
+          `If you'd like to help ensuring the changes made to MusicBrainz data
```
yvanzo:
```suggestion
          `If you’d like to help ensuring the changes made to MusicBrainz data
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+      </p>
+
+      <h2>{l('Destructive edits')}</h2>
+      <p>
+        {l(
+          `Destructive edits (removals and merges) are often very hard or even
+           impossible to revert. As such, an incorrect destructive edit that
+           applies unnoticed can cause quite a big mess! Most are guaranteed
+           to remain open for at least two full days even if they get three
+           “Yes” votes, to avoid them closing too quickly, but it's always
```
yvanzo:
```suggestion
           “Yes” votes, to avoid them closing too quickly, but it’s always
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+  <Layout fullWidth title={l('Voting reports')}>
+    <div id="content">
+      <h1>{l('Voting reports')}</h1>
+
+      <p>
+        {exp.l(
+          `If you'd like to help ensuring the changes made to MusicBrainz data
+           are correct, but do not know where to start, the following reports
+           should be useful. These are pre-defined
+           {edit_search|edit searches}; once you're comfortable with the edit
```
yvanzo:
```suggestion
           {edit_search|edit searches}; once you’re comfortable with the edit
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+          {edit_search: '/search/edits'},
+        )}
+      </p>
+      <p>
+        {exp.l(
+          `While reviewing the work of your fellow editors, always keep the
+           {coc|Code of Conduct} in mind. Almost all editors want to help, so
+           your goal as a voter is to help them help better. This applies even
+           more for beginners: always try to be helpful and patient with them,
+           even if they are making mistakes, so that they'll hopefully grow
```
yvanzo:
```suggestion
           even if they are making mistakes, so that they’ll hopefully grow
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+      <p>
+        {exp.l(
+          `Beginner editors are the ones most likely to need a friendly eye on
+           their edits. When reviewing these, look for cases where the editor
+           is making edits that go against the MusicBrainz guidelines, or seem
+           otherwise wrong, and leave notes guiding the editor so that they
+           can do better in the future. Make sure to link to the appropriate
+           guidelines, or to relevant examples of well-entered data, and above
+           all remember to be nice. Chances are if these users are making a
+           terrible mess, they're not doing it on purpose but out of
```
yvanzo:
```suggestion
           terrible mess, they’re not doing it on purpose but out of
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           confusion: MusicBrainz can be daunting for newcomers! If at all
+           possible, fix the errors (and let the editor know that you've done
```
yvanzo:
```suggestion
           possible, fix the errors (and let the editor know that you’ve done
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           that and that they can check the edits you made to see how it
+           should look like) rather than voting “No” on edits, since “No”
+           votes can be quite discouraging, especially as a new editor.
+           If something is just so bad that there's no fixing it and it is
```
yvanzo:
```suggestion
           If something is just so bad that there’s no fixing it and it is
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           making the existing data worse, do vote against the edit, but
+           make sure to explain nicely why that is needed, rather than just
+           silently “No”-voting. For a longer overview of the attitude we're
```
yvanzo:
```suggestion
           silently “No”-voting. For a longer overview of the attitude we’re
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           hoping for, see {voting_blog|this blog post about voting}.
+           Also, if a beginner is making especially good edits,
+           you might want to let them know so they'll feel good about it!`,
```
yvanzo:
```suggestion
           you might want to let them know so they’ll feel good about it!`,
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+          {
+            voting_blog: 'https://blog.metabrainz.org/2015/01/09/editing-making-musicbrainz-better/',
+          },
+        )}
+      </p>
+      <p>
+        {l(
+          `The filters for edits you haven't voted on yet below will only show
```
yvanzo:
```suggestion
          `The filters for edits you haven’t voted on yet below will only show
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           open edits (since you can't vote on closed edits anyway).
```
yvanzo:
```suggestion
           open edits (since you can’t vote on closed edits anyway).
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+        {/* eslint-enable max-len */}
+      </ul>
+
+      <h2>{l('All edits')}</h2>
+      <p>
+        {l(
+          `Sometimes you might want to just check all edits for some reason.
+           If you feel like being overwhelmed by a very long list of edits
+           is just what the doctor ordered, just check the searches below!
+           Don't forget you can always experiment with different edit search
```
yvanzo:
```suggestion
           Don’t forget you can always experiment with different edit search
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+      <h1>{l('Voting reports')}</h1>
+
+      <p>
+        {exp.l(
+          `If you’d like to help ensuring the changes made to MusicBrainz data
+           are correct, but do not know where to start, the following reports
+           should be useful. These are pre-defined
+           {edit_search|edit searches}; once you’re comfortable with the edit
+           search you can refine them further or just create your own personal
+           reports for your own use!`,
```
mwiencek: Are we sticking with using the "reports" terminology here?

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+          showSubscribedArtistsUrl
+        />
+        {/* eslint-enable max-len */}
+      </ul>
+
+      <h2>{l('Unreviewed and potentially problematic edits')}</h2>
+      <p>
+        {l(
+          `Edits that nobody have seen can always benefit from a quick check:
+           even if you're not familiar with the music in question, you might
```
yvanzo:
```suggestion
           even if you’re not familiar with the music in question, you might
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           be able to notice that something seems wrong. Don't forget you
+           don't need to vote on every edit: it's perfectly fine to just
+           abstain if you feel something doesn't seem wrong but it's also not
```
yvanzo:
```suggestion
           abstain if you feel something doesn’t seem wrong but it’s also not
```

root/vote/VotingIndex.js
```diff
@@ -0,0 +1,381 @@
...
+           100% obvious that it is right without further checks.`,
```
yvanzo: definitely not “right without further checks” :wink:

root/vote/VotingIndex.js
reosarevok: Whoops. Nope! Fixed.

PR #1175, opened 2019-08-21
MBS-4913: Indicate the number of CAA pieces on tab

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4913

In order to easily see whether we have only one or two pieces of cover art or an entire scanned booklet, it's useful to have the number of images listed as part of the cover art tab name.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artwork.pm
+++ b/lib/MusicBrainz/Server/Data/Artwork.pm
@@ -145,6 +145,18 @@ sub find_front_cover_by_release
     return \@artwork;
 }
 
+sub find_count_by_release
+{
+    my ($self, $release_id) = @_;
+
+    return unless $release_id; # nothing to do
+    my $query = "SELECT count(*)
+        FROM cover_art_archive.index_listing
+        WHERE cover_art_archive.index_listing.release = ?";
+
+    return $self->sql->select_single_value($query, $release_id);
+}
+
 sub load_for_release_groups
 {
     my ($self, @release_groups) = @_;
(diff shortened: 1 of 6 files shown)


lib/MusicBrainz/Server/Data/Artwork.pm
```diff
@@ -145,6 +145,20 @@ sub find_front_cover_by_release
...
 }
 
+sub find_count_by_release
+{
+    my ($self, $release_id) = @_;
+
+    return unless $release_id; # nothing to do
+    my $query = "SELECT count(*)
+        FROM cover_art_archive.index_listing
+        JOIN cover_art_archive.cover_art
```
mwiencek: Is that `JOIN` needed?

root/components/EntityTabs.js
```diff
@@ -17,19 +17,18 @@ import Tabs from './Tabs';
...
-  'fingerprints': N_l('Fingerprints'),
-  'labels': N_l('Labels'),
-  'map': N_l('Map'),
-  'performances': N_l('Performances'),
-  'places': N_l('Places'),
-  'recordings': N_l('Recordings'),
-  'releases': N_l('Releases'),
-  'users': N_l('Users'),
-  'works': N_l('Works'),
+  artists: N_l('Artists'),
```
mwiencek: Does eslint actually complain if none of the keys need quotes? (Otherwise I'd probably keep them, 'cause it doesn't seem unlikely that we'll have others with hyphens.)

reosarevok: Yup. "Properties shouldn't be quoted as all quotes are redundant." (quote-props)

lib/MusicBrainz/Server/Data/Artwork.pm
reosarevok: Guess not, actually, now that I properly looked into index_listing!

PR #1176, opened 2019-08-21
MBS-4644: Show releases that have CAA images in lists

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4644

For the artist, label and release group release lists, it is useful to know which releases have cover art (both if you want to find which ones to improve and if you just want to look at cover art). This adds an icon after the release name, similarly to the video recording icon.

Took the icon from the AndroidIcons set, since we've been using that one before and it's easier for crediting.


root/components/list/EventList.js
```diff
@@ -67,9 +67,10 @@ const EventList = ({
       const nameColumn =
         defineNameColumn<EventT>(
           l('Event'),
+          // to use EntityLink without dates (separate column for that)
+          {descriptive: false},
           order,
           sortable,
-          false, // to use EntityLink without dates (separate column for that)
```
yvanzo: Such change should probably be made and explained in a separate commit.

root/utility/tableColumns.js
```diff
@@ -239,9 +239,12 @@ export function defineInstrumentUsageColumn(
 
 export function defineNameColumn<T: CoreEntityT | CollectionT>(
   title: string,
+  {descriptive = true, showCaaPresence = false}: {
+    descriptive?: boolean,
+    showCaaPresence?: boolean,
+  },
```
yvanzo: IMHO, all parameters to `define*Column` should be named, just like for React components already.
That would also avoid adding meaningless empty `{}` all over the code.

mwiencek: @reosarevok when you have time, you can update this PR now that the named parameter stuff got merged. should resolve this comment

root/static/styles/layout.less
```diff
@@ -592,6 +592,16 @@ span.ac-mp {
     margin-left: -5px;
 }
 
+span.caa-icon {
+    background-image: data-uri('../images/icons/picture.png');
+    background-repeat: no-repeat;
+    display: inline-block;
+    height: 18px;
+    margin-left: 5px;
+    vertical-align: middle;
```
mwiencek, line 601: `vertical-align: bottom;` actually looks more middle-aligned to me with respect to the release link, I guess because the `<a>` for the link doesn't also have `vertical-align: middle;`.

PR #1179, opened 2019-08-24
Link from guess case options to guess case documentation page

yvanzo:
Thanks to @reosarevok, https://musicbrainz.org/doc/Guess_Case is currently up-to-date.
This patch links to this page from guess case options dialog and tracklist guess case form.
It additionally slightly improves display of guess case options and removes unused style.

### DIFF
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -281,6 +281,7 @@
         <option value="French">[% l('French') %]</option>
         <option value="Turkish">[% l('Turkish') %]</option>
       </select>
+      (<a href="/doc/Guess_Case" target="_blank">[% l('help') %]</a>)
       <br />
       <label>
         <input type="checkbox" data-bind="checked: keepUpperCase" />
@@ -305,6 +306,7 @@
           <option value="French">[% l('French') %]</option>
           <option value="Turkish">[% l('Turkish') %]</option>
         </select>
+        (<a href="/doc/Guess_Case" target="_blank">[% l('help') %]</a>)
         <br />
         <label>
           <input type="checkbox" data-bind="checked: keepUpperCase" />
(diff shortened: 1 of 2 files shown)


root/components/forms.tt
```diff
@@ -281,6 +281,7 @@
         <option value="French">[% l('French') %]</option>
         <option value="Turkish">[% l('Turkish') %]</option>
       </select>
+      (<a href="/doc/Guess_Case" target="_blank" class="open">[% l('help') %]</a>)
```
mwiencek: I don't see any `open` class in our CSS that'd apply here.

root/components/forms.tt
```diff
@@ -306,6 +307,8 @@
           <option value="Turkish">[% l('Turkish') %]</option>
         </select>
         <br />
```
mwiencek: This `<br />` seems like it should be removed now, so that the link is on the same line as in the case above.

root/components/forms.tt
yvanzo: I don’t recall why I used this class, removed from https://github.com/metabrainz/musicbrainz-server/pull/1179/commits/ee0f533c7ecfb8e664a23ef076835f1adbd050be.

root/components/forms.tt
yvanzo: Removed from https://github.com/metabrainz/musicbrainz-server/pull/1179/commits/ee0f533c7ecfb8e664a23ef076835f1adbd050be as well.

yvanzo:
The Guess Case Options popup looks ugly but this isn’t new to this PR either. IMHO, the _(help)_ link displayed below the mode select doesn’t look as bad as as the _Uppercase_ displayed below its checkbox.

Bubbles are built using jQuery UI Dialog, and the only sane thing to do about that is to rewrite it using React. This is far beyond the scope of this PR, but definitely something that has to be done.

PR #1181, opened 2019-08-26
MBS-4787: Redirect WS namespace to WS docs

reosarevok:
# Implements [MBS-4787](https://tickets.metabrainz.org/browse/MBS-4787)

While the namespace URI doesn't have to point to anything to be valid, it seems sensible to point it to the documentation to give users a quick link to the docs rather than a kinda frustrating 404.
This was apparently also the pre-NGS behaviour.


admin/nginx/mbserver-rewrites.conf
```diff
@@ -83,6 +83,7 @@ rewrite ^/list\.html$                            $mb_server/doc/Communication/Ma
 rewrite ^/mod_faq\.html$                         $mb_server/doc/Editing_FAQ                  permanent;
 rewrite ^/mod_intro\.html$                       $mb_server/doc/How_Editing_Works            permanent;
 rewrite ^/news/licenses\.html$                   $mb_server/doc/About/Data_License           permanent;
+rewrite ^/ns/mmd-2.0$                            $mb_server/doc/Development/XML_Web_Service  permanent;
```
zas: You need to escape the dot:
`rewrite ^/ns/mmd-2\.0$`

reosarevok: Whoops. Done!

PR #1182, opened 2019-08-26
MBS-10240: Allow seeding "release has no barcode"

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10240

Needs updating the barcode section of https://musicbrainz.org/doc/Development/Release_Editor_Seeding to "The barcode of the release. May be any valid barcode without whitespace. To indicate there is no barcode, seed "none"" if it gets merged.

### DIFF
--- a/root/static/scripts/release-editor/seeding.js
+++ b/root/static/scripts/release-editor/seeding.js
@@ -75,7 +75,11 @@ releaseEditor.seedRelease = function (release, data) {
     }
 
     if (data.barcode) {
-        release.barcode.value(data.barcode);
+        if (data.barcode === 'none') {
+            release.barcode.none(true);
+        } else {
+            release.barcode.value(data.barcode);
+        }
     }
 
     if (data.artistCredit) {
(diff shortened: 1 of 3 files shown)


root/static/scripts/release-editor/seeding.js
```diff
@@ -75,7 +75,11 @@ releaseEditor.seedRelease = function (release, data) {
     }
 
     if (data.barcode) {
-        release.barcode.value(data.barcode);
```
mwiencek: I kinda think we should handle this on the server by doing something like `$result->{barcode} = '' if $result->{barcode} eq 'none';` below https://github.com/metabrainz/musicbrainz-server/blob/3702485a813fa82ce894a555c7b6a9b9a40447e8/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm#L148-L148
since that's where the other seeded data processing is done at present. Though that requires changing the `data.barcode` check in the JS above to `!= null`.

mwiencek: (Decided in IRC to do something different, since it wasn't as simple as that.)

reosarevok:
I updated the docs already because we will certainly forget otherwise, hopefully nobody is too misled in the next week or so.

PR #1183, opened 2019-08-27
Add support for expanding tag <i> to React

yvanzo:


yvanzo:
I don’t find the reason why `<b>` is supported. I just recall this has been discussed when rewriting localization to React. That might be because of existing strings with `<b>` in it. Honestly, I wouldn’t mind to remove `<b>` instead of adding `<i>`. These tags doesn’t seem to be needed at all.

reosarevok:
I guess that works for me too, but if we're going to disallow some of the tags everyone knows about, we should also add some info to the fields that support HTML letting people know. That's at least instrument descriptions, but probably some others too?

PR #1184, opened 2019-08-27
MBS-5049: Link to edit note syntax from edit pages

reosarevok:
https://tickets.metabrainz.org/browse/MBS-5049

We currently only show some info about edit note syntax when entering an edit, but not when voting on it. The info is equally useful there.

While at it, added a reminder to follow the CoC while voting: if we're going to have text, we might as well remind people to be nice!

### DIFF
--- a/root/edit/notes.tt
+++ b/root/edit/notes.tt
@@ -23,6 +23,11 @@
                   <textarea class="edit-note" rows="5" placeholder="[% l('Add an edit note') %]"
                       name="enter-vote.vote.[% index %].edit_note"></textarea>
               </div>
+            <p>
+                  [% l('Edit notes support {doc_formatting|a limited set of wiki formatting options}. Please do always keep the {doc_coc|Code of Conduct} in mind when writing edit notes!',
+                     { doc_coc => doc_link('Code_of_Conduct'), doc_formatting => doc_link('Edit_Note') }) %]
+            </p>
+
           </div>
       [% ELSE %]
           <p>


root/edit/notes.tt
```diff
@@ -23,6 +23,11 @@
                   <textarea class="edit-note" rows="5" placeholder="[% l('Add an edit note') %]"
                       name="enter-vote.vote.[% index %].edit_note"></textarea>
               </div>
+            <p>
+                  [% l('HTML is not supported, but {url|some basic syntax} is available. Please do always keep the {coc|Code of Conduct} in mind when writing edit notes!',
```
yvanzo: It would be nice to use the same wording as for [annotation formatting](https://github.com/metabrainz/musicbrainz-server/blob/v-2019-08-22/root/annotation/edit.tt#L38) for the first sentence, e.g. “Edit notes support a limited set of wiki formatting options”.

root/edit/notes.tt
```diff
@@ -23,6 +23,11 @@
...
+                     { coc => doc_link('Code_of_Conduct'), url => doc_link('Edit_Note') }) %]
```
yvanzo: Please use consistent keys, e.g.
* `doc_syntax` instead of `url`
* `doc_coc` instead of `coc`

There already are some `doc_*` keys around. It is probably more readable to translators.

reosarevok: Good point! :)

root/edit/notes.tt
reosarevok: Oh, that does sound good. I struggled to find some decent wording here so I had gone with what the wiki had.

PR #1186, opened 2019-08-30
MBS-4826: Show RGs for recordings in artist list

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4826
https://tickets.metabrainz.org/browse/MBS-4881

A popular artist's recording list is usually a lot of the same name with only a length to differentiate them. This makes it so that the release groups each recording appears in are also shown, making it easier to figure out if they're different recordings (e.g. album vs live bootleg) or whether maybe they should be looked into for merging.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -426,6 +426,7 @@ sub recordings : Chained('load')
 
     $c->model('Recording')->load_meta(@$recordings);
 
+    my %release_group_appearances = $c->model('Recording')->appears_on($recordings, 10, 1);
     if ($c->user_exists) {
         $c->model('Recording')->rating->load_user_ratings($c->user->id, @$recordings);
     }
@@ -447,6 +448,7 @@ sub recordings : Chained('load')
             hasVideo => boolean_to_json($has_video),
             pager => serialize_pager($c->stash->{pager}),
             recordings => to_json_array($recordings),
+            releaseGroupAppearances => \%release_group_appearances,
             standaloneOnly => boolean_to_json($standalone_only),
             videoOnly => boolean_to_json($video_only),
         },
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -426,7 +426,7 @@ sub recordings : Chained('load')
 
     $c->model('Recording')->load_meta(@$recordings);
 
-    my %release_group_appearances = $c->model('Recording')->appears_on($recordings, 10, 1);
+    my $release_group_appearances = $c->model('Recording')->appears_on($recordings, 10, 1);
     if ($c->user_exists) {
         $c->model('Recording')->rating->load_user_ratings($c->user->id, @$recordings);
     }
@@ -448,7 +448,7 @@ sub recordings : Chained('load')
             hasVideo => boolean_to_json($has_video),
             pager => serialize_pager($c->stash->{pager}),
             recordings => to_json_array($recordings),
-            releaseGroupAppearances => \%release_group_appearances,
+            releaseGroupAppearances => $release_group_appearances,
             standaloneOnly => boolean_to_json($standalone_only),
             videoOnly => boolean_to_json($video_only),
         },
--- a/lib/MusicBrainz/Server/Data/Recording.pm
+++ b/lib/MusicBrainz/Server/Data/Recording.pm
@@ -3,7 +3,7 @@ package MusicBrainz::Server::Data::Recording;
 use Moose;
 use namespace::autoclean;
 use List::MoreUtils qw( uniq );
-use List::UtilsBy qw( rev_nsort_by sort_by uniq_by );
+use List::UtilsBy qw( nsort_by sort_by uniq_by );
 use MusicBrainz::Server::Constants qw(
     $EDIT_RECORDING_CREATE
     $EDIT_HISTORIC_ADD_TRACK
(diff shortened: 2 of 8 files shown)


reosarevok:
One is to show the number of releases (that is still the case). The other is indeed to show releases, but showing RGs seems to give as good an overview of the usage without causing a huge list in cases where an album for example has been reissued a million times.

mwiencek:
Rebased this so I could test easier, and fixed a few tiny Flow/eslint issues.

This looks good, but we already have a method in `Data::Recording` that gets release group appearances: `sub appears_on`. Only difference is it doesn't include the release counts (total and with recording).

So I think we could combine those methods, but do we also even need the release counts? They look kind of weird and don't seem very useful (almost all of them are `(1/1)` or `(2/2)`). They also don't seem necessary to resolve MBS-4881, since simply seeing how many release groups are listed looks like a good way to gauge number of appearances. (Though perhaps we should display the number of collapsed items like we do for release events, too.)

reosarevok:
I guess? I feel it's mostly useful for cases where it's not x/x but a smaller amount (maybe it's just a bonus track in a few releases?) but it might be that even that is not worth the effort. @yvanzo: what do you think?

Good point about showing the amount, will do.

reosarevok:
Changed this to use the same collapsible style as ReleaseEvents and WorkArtists. By now, we should probably be refactoring this out into its own component though...

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -175,6 +175,39 @@ sub find_by_works
...
+                 WHERE rec.id IN (" . placeholders(@ids) . ")
+                 GROUP BY rec.gid, rg.id, rg.gid, rg.name, rgm.release_count
+                 ORDER BY rec.gid, count(release.gid) DESC, rgm.release_count";
+
+    my %map;
+    for my $row (@{ $self->sql->select_list_of_hashes($query, @ids) }) {
+        $map{ $row->{recording_gid} } ||= [];
+        push @{ $map{ $row->{recording_gid} } }, {
+            countTotal         => $row->{total_release_count},
+            countWithRecording => $row->{recording_release_count},
```
mwiencek, line 203: The longer names seem less confusing (countTotal of what). Although see my general review comment, 'cause I'm not sure what these even mean. :P

mwiencek (changes requested):
Testing http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/recordings?page=1 with pink data and I see weird things like "Broken (910/17)" (very apt name). Can you look into that?

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -175,6 +175,39 @@ sub find_by_works
...
+            releaseGroup       => $self->c->model('ReleaseGroup')->_new_from_row($row),
+        }
+    }
+
+    return %map;
```
mwiencek, line 208: I'd return `\%map` here, otherwise we make a copy during assignment

reosarevok:
> Broken (910/17)

Heh. That recording is used many times per release, so I guess 910 is the amount of times it appears in total...

mwiencek:
I'm partial to dropping them, unless you think the counts can be more useful than not. If we do, I think we can then re-use the existing `appears_on` method from `Data::Recording`.

reosarevok:
I updated this to drop that info and use appears_on. Seems to work, but could certainly use an eye again :)

lib/MusicBrainz/Server/Controller/Artist.pm
```diff
@@ -426,6 +426,7 @@ sub recordings : Chained('load')
 
     $c->model('Recording')->load_meta(@$recordings);
 
+    my %release_group_appearances = $c->model('Recording')->appears_on($recordings, undef, 1);
```
mwiencek, line 429: Might be a good idea to still set a limit (like 100 or something)? Looks like quite a few recordings have 300+ RGs and the most has 700 (though I didn't take into account duplicate name+type pairs). We could modify the `appears_on` query to join `release_group_meta` and sort by first_release_date columns if we do that.

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -387,7 +388,7 @@ then singles, etc..) and then by name.
 
 sub appears_on
```
mwiencek, line 389: I know it was like this before this PR, but it'd be a good idea to return a hash ref from here so we don't have to make a copy every time it's called.

Also, if you change the query to add the date sorting as suggested, you can probably remove the `recording` join and just use `track.recording` in place of `recording.id`.

root/static/scripts/common/components/Collapsible.js
```diff
@@ -61,11 +62,20 @@ class Collapsible extends React.Component<Props, State> {
 
     return (
       <>
-        <div
-          className={_className}
-          dangerouslySetInnerHTML={{__html: html}}
-          ref={this.containerRef}
-        />
+        {html !== '' && html != null ? (
```
mwiencek, line 65:
```suggestion
        {nonEmpty(html) ? (
```

mwiencek:
> Changed this to use the same collapsible style as ReleaseEvents and WorkArtists. By now, we should probably be refactoring this out into its own component though...

Agreed. You're also adding another use in https://github.com/metabrainz/musicbrainz-server/pull/2024. :) Maybe we can do the refactoring in that PR once this is merged.

mwiencek:
I pushed a version of the query that sorts by first release date instead of type and limits the amount of rows returned from PG. Let me know if it seems better or worse...

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -430,7 +430,7 @@ sub appears_on
         }
     }
 
-    return %map;
+    return \%map;
```
reosarevok, line 463: Oh, sorry, with all the other things we talked about I forgot this one :) Thanks!

reosarevok:
The new query seems like a good improvement, if the speed is fine (which I guess it is, compared with "loading all into Perl, then slicing"). Thanks!

PR #1189, opened 2019-09-02
MBS-10331: Autoselect and sidebar for Amazon Music

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10331

This should also fix https://tickets.metabrainz.org/browse/MBS-10329 for obvious reasons.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -202,6 +202,11 @@ export const LINK_TYPES = {
     place: '1cd2eb89-2997-4901-87e9-838ac9a68da9',
     series: '4789521b-57b9-4689-9644-46de63190f66',
   },
+  streaming: {
```
yvanzo: That should probably be committed apart and refer to [STYLE-1287](https://tickets.metabrainz.org/browse/STYLE-1287).

reosarevok: It is commited apart, that's why we have two commits (https://github.com/metabrainz/musicbrainz-server/pull/1189/commits/55320838837b68b65e2d1a3834a1ef33874ffc61)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -83,6 +83,11 @@ export const LINK_TYPES = {
     recording: '92777657-504c-4acb-bd33-51a201bd57e1',
     release: '98e08c20-8402-4163-8970-53504bb6a1e4',
   },
+  freestreaming: {
```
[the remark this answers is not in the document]
reosarevok: You don't purchase a stream, but streamingpaid makes sense. Made the change.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -429,6 +434,53 @@ const CLEANUPS = {
...
   },
+  'amazonmusic': {
+    match: [new RegExp('^(https?://)?music\\.amazon\\.(com|ca|co\\.uk|fr|at|de|it|co\\.jp|jp|cn|es|in|com\\.br|com\\.mx|com\\.au)/(albums|artists)', 'i')],
+    type: LINK_TYPES.streaming,
+    clean: function (url) {
+      /*
+       * determine tld, asin from url, and build standard format [1],
+       * if both were found.
+       *
+       * [1] "https://www.amazon.<tld>/(albums|artists)/<ASIN>"
```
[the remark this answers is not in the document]
reosarevok: Probably yes for the tld bit. Not sure for the rest, since I'm also taking the type this time unless on the ASIN version. Still worth it? :)

root/static/scripts/edit/URLCleanup.js
yvanzo: It isn’t, it is in the same commit as changes for Amazon Music. :)

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: Sure, please deduplicate what can be. A constant for Amazon public suffixes/effective TLDs would probably help with.

root/static/scripts/edit/URLCleanup.js
reosarevok: Oh, you mean you'd add the rels first and then the Amazon Music bit. But why? They aren't needed up until Amazon Music is added.

yvanzo: To improve readability? :) This is the only change related to STYLE-1287.

reosarevok: Ok, I think I split it now (I was annoyed because it's hard to split only some of the changes to a file off, but I just discovered git add -p :) )

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
reosarevok: Extracted a findAmazonTld function.

The tld constant is not so trivial because we need to escape dots with \ in some places and \\ in others, unless there's something I'm missing that makes that trivial too :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -388,7 +408,7 @@ const CLEANUPS = {
     },
   },
   'amazon': {
-    match: [new RegExp('^(https?://)?([^/]+\\.)?(amazon\\.(com|ca|co\\.uk|fr|at|de|it|co\\.jp|jp|cn|es|in|com\\.br|com\\.mx|com\\.au)|amzn\\.com)', 'i')],
+    match: [new RegExp('^(https?://)?([^/]+(?<!music)\.)?(amazon\\.(com|ca|co\\.uk|fr|at|de|it|co\\.jp|jp|cn|es|in|com\\.br|com\\.mx|com\\.au)|amzn\\.com)', 'i')],
```
mwiencek: What's the `(?<!` syntax? I get `SyntaxError: invalid regexp group` if I paste this in the console, so it seems suspect, but then I'm not sure how the tests are passing.

lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -21,6 +21,7 @@ my %URL_SPECIALIZATIONS = (
     '45cat'               => qr{^https?://(?:www\.)?45cat\.com/}i,
     '45worlds'            => qr{^https?://(?:www\.)?45worlds\.com/}i,
     'Allmusic'            => qr{^https?://(?:www\.)?allmusic\.com/}i,
+    'AmazonMusic'         => qr{^https?://music.amazon(.*?)(?:\:[0-9]+)?/.*/([0-9B][0-9A-Z]{9})(?:[^0-9A-Z]|$)}i,
```
mwiencek: This pattern seems strange, since it doesn't care if it matches against something like `music.amazonnotreally.com`, but cares about port numbers which would never appear in the URL (they're not accepted by the URL cleanup in the first place).

(Guess it was mostly copied from the ASIN one below, but judging by 9f9ba4898348b528b2d1e502de3e2319e17a0fb8 that was probably based on a general URL pattern with support for ASINs added at the end.)

It'd seem reasonable to me to just reuse the RegExp added for amazonmusic in the URLCleanup file.

lib/MusicBrainz/Server/Entity/URL/AmazonMusic.pm
```diff
@@ -0,0 +1,39 @@
...
+extends 'MusicBrainz::Server::Entity::URL';
+with 'MusicBrainz::Server::Entity::URL::Sidebar';
+
+use DBDefs;
+
+sub pretty_name
+{
+    my $self = shift;
+
+    if ($self->url =~ m{^(?:https?:)?//music.(.*?\.)([a-z]+)(?:\:[0-9]+)?/.*/(?:[0-9B][0-9A-Z]{9})(?:[^0-9A-Z]|$)}i) {
```
mwiencek: Same re: using a pattern closer to the one from the URLCleanup file. But I don't think it needs to match anything after the TLD here, since that's all you need for the country. And regarding the country, would this convert TLDs like `.co.uk` and `.co.jp` correctly?

root/static/scripts/edit/URLCleanup.js
yvanzo: Oops, it is [negated lookbehind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-negative-lookbehind) _(Added in ES2018, not yet supported in Firefox)_.

reosarevok: Any suggestions what to use that does work in FF?

reosarevok: Moving it to be a lookahead before [^/]+ seems to to do the trick, but let me know if that seems like a mistake :)

lib/MusicBrainz/Server/Entity/URL/AmazonMusic.pm
reosarevok: I changed this to use the other pattern and changed how it converts the TLD to be safe.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -388,7 +408,7 @@ const CLEANUPS = {
     },
   },
   'amazon': {
-    match: [new RegExp('^(https?://)?([^/]+\\.)?(amazon\\.(com|ca|co\\.uk|fr|at|de|it|co\\.jp|jp|cn|es|in|com\\.br|com\\.mx|com\\.au)|amzn\\.com)', 'i')],
+    match: [new RegExp('^(https?://)?((?!music)[^/]+\.)?(amazon\\.(com|ca|co\\.uk|fr|at|de|it|co\\.jp|jp|cn|es|in|com\\.br|com\\.mx|com\\.au)|amzn\\.com)', 'i')],
```
yvanzo: As a follow-up to https://github.com/metabrainz/musicbrainz-server/pull/1189#discussion_r331579261, here is an alternative that works with `anysubdomainto.music.amazon.co.jp` but also dismisses `anysubdomainwithmusicinside.amazon.co.jp`:
```suggestion
    match: [new RegExp('^(https?://)?(((?!music)[^/])+\.)?(amazon\\.(com|ca|co\\.uk|fr|at|de|it|co\\.jp|jp|cn|es|in|com\\.br|com\\.mx|com\\.au)|amzn\\.com)', 'i')],
```

PR #1190, opened 2019-09-03
MBS-10304: Reallow special codes as editor language

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10304

Reverting commit b645483e87b0a08f150a59d813fe807245680d08 as requested
by several editors, including our community manager.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Utils.pm
+++ b/lib/MusicBrainz/Server/Form/Utils.pm
@@ -38,7 +38,7 @@ sub language_options {
     my @languages = $c->model('Language')->get_all;
     if ($context eq "editor") {
         for my $language (@languages) {
-            if ($language->iso_code_3 && $language->iso_code_3 =~ /mis|mul|qaa|und|zxx/) {
+            if ($language->iso_code_3 && $language->iso_code_3 eq 'mul') {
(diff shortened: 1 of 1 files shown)


yvanzo:
Is that a joke?

reosarevok:
Why would it be? If the people involved with community stuff ask for this to be brought back for the community, why shouldn't we do it?

yvanzo:
I genuinely thought that it was a joke, ticket included.

reosarevok:
Changed this to still disallow `mul`, as per comments on the ticket :)

mwiencek:
@yvanzo thoughts on this now?

yvanzo:
Per yesterday’s discussion, it would be to avoid heterogeneous styling. Maybe having a style similar to `ar-desc` class, like in pull request https://github.com/metabrainz/musicbrainz-server/pull/1217, would do?

mwiencek:
Don't think it's related - at least I can't reproduce the issue locally or by connecting to the pink or williams DBs.

yvanzo:
For the record, that crash was due to a language with frequency set to 3 in the test database. Even though restored values still make no sense to me, these cannot be mistakenly selected by editors at least.

PR #1194, opened 2019-09-03
Eslint fixes for dialogs.js

reosarevok:
To be reviewed after https://github.com/metabrainz/musicbrainz-server/pull/1193

### DIFF
--- a/root/static/scripts/release-editor/dialogs.js
+++ b/root/static/scripts/release-editor/dialogs.js
@@ -216,13 +216,12 @@ class SearchTab {
     if (event.keyCode === 13) { // Enter
       this.search(data, event);
       return false;
-    } else {
-      /*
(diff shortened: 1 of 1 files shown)


root/static/scripts/release-editor/dialogs.js
```diff
@@ -20,341 +22,359 @@ import trackParser from './trackParser';
...
-            });
+  }
+
+  buttonClicked() {
+    this.searching() ? this.cancelSearch() : this.search();
+  }
+
+  keydownEvent(data, event) {
+    if (event.keyCode === 13) { // Enter
+      this.search(data, event);
```
reosarevok: consistent-return wants this to also return something. Should we, or should we just disable consistent-return for this bit?

reosarevok:
I think most of this if not all is covered by the other opened Eslint PRs.

PR #1196, opened 2019-09-05
MBS-2436: Link to created recordings on EditMedium

reosarevok:
https://tickets.metabrainz.org/browse/MBS-2436

Once an Edit Medium edit applies, creating new recordings, we should update the edit data so that the new recordings get displayed in the history. This updates the edit data to use the final tracklist while applying the edit.

Same general idea as https://github.com/metabrainz/musicbrainz-server/pull/1124

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Medium/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Medium/Edit.pm
@@ -32,6 +32,7 @@ use MusicBrainz::Server::Log qw( log_assertion log_debug );
 use MusicBrainz::Server::Validation 'normalise_strings';
 use MusicBrainz::Server::Translation qw( N_l );
 use MusicBrainz::Server::Track qw( format_track_length );
+use JSON::XS;
 use Try::Tiny;
 
 extends 'MusicBrainz::Server::Edit::WithDifferences';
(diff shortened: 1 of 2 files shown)


reosarevok:
I'm confused about the test failure. It seems the test expects to see unchanged `position` after the reordering: https://github.com/metabrainz/musicbrainz-server/blob/962c5c01e64988b8197eb6ea1325bec6af69ff7f/t/lib/t/MusicBrainz/Server/Edit/Medium/Edit.pm#L584 specifically shows it expects the same hash on {old} and {new}, just differently placed in the array. That seems wrong though? I checked an existing edit on test, and the `position` does get changed in the data. So I'm confused why this even passes in prod.

mwiencek:
From what I can see, it passes just because that's how the test creates the edit data:
https://github.com/metabrainz/musicbrainz-server/blob/962c5c01e64988b8197eb6ea1325bec6af69ff7f/t/lib/t/MusicBrainz/Server/Edit/Medium/Edit.pm#L570-L578
Whereas the release editor wouldn't submit the edit like that, it'd renumber the positions first. The edit just stores whatever it's given. Since you're now updating what's stored (and `@final_tracklist` has things renumbered correctly), it now fails.

So the test wasn't totally wrong (it still checked that the tracks were reordered fine), but it stored the positions wrong, which was unrealistic and now breaks. We should just fix the test to store the correct positions.

PR #1197, opened 2019-09-06
MBS-3839: Allow removing from queue during merge

reosarevok:
https://tickets.metabrainz.org/browse/MBS-3839

While working on it I also implemented https://tickets.metabrainz.org/browse/MBS-10456

### DIFF
--- a/root/area/merge.tt
+++ b/root/area/merge.tt
@@ -4,16 +4,29 @@
     <p>[% l('You are about to merge the following areas into a single area. Please
              select the area which you would like other areas to be merged into:') %]</p>
     <form action="[% c.req.uri %]" method="post">
-        <ul>
+        <table class="tbl">
+            <thead>
+                <tr>
+                    <th style="width: 1em">
+                    </th>
+                    <th>[%- l('Area') -%]</th>
+                    [%- remove_from_merge_header(to_merge) -%]
+                </tr>
+            </thead>
+            <tbody>
             [% FOR entity=to_merge %]
-            <li>
-                <input type="hidden" name="merge.merging.[% loop.index %]" value="[% entity.id %]" />
-                <input type="radio" name="merge.target" value="[% entity.id %]"
-                   [% 'checked="checked"' IF entity.id == form.field('target').value %] />
-                [% descriptive_link(entity) %]
-            </li>
+                <tr class="[% loop.parity %]">
+                    <td>
+                        <input type="hidden" name="merge.merging.[% loop.index %]" value="[% entity.id %]" />
+                        <input type="radio" name="merge.target" value="[% entity.id %]"
+                           [% 'checked="checked"' IF entity.id == form.field('target').value %] />
+                    </td>
+                    <td>[% descriptive_link(entity) %]</td>
+                    [%- remove_from_merge('area', to_merge) -%]
+                </tr>
             [% END %]
-        </ul>
+            </tbody>
+        </table>
         [% field_errors(form, 'target') %]
 
         [% INCLUDE "forms/edit-note.tt" %]
--- a/root/release/merge.tt
+++ b/root/release/merge.tt
@@ -19,30 +19,32 @@
                     <th>[% l('Label') %]</th>
                     <th>[% l('Catalog#') %]</th>
                     <th>[% l('Barcode') %]</th>
+                    [%- remove_from_merge_header(to_merge) -%]
                 </tr>
             </thead>
             <tbody>
-            [%- FOR release=to_merge -%]
+            [%- FOR entity=to_merge -%]
               <tr class="[% loop.parity %]">
                 <td>
-                  <input type="hidden" name="merge.merging.[% loop.index %]" value="[% release.id %]" />
-                  <input type="radio" name="merge.target" value="[% release.id %]" />
+                  <input type="hidden" name="merge.merging.[% loop.index %]" value="[% entity.id %]" />
+                  <input type="radio" name="merge.target" value="[% entity.id %]" />
                 </td>
                   <td>
-                    [%~ link_entity(release) ~%]
+                    [%~ link_entity(entity) ~%]
                   </td>
-                  <td>[% artist_credit(release.artist_credit) %]</td>
-                  <td>[% html_escape(release.combined_format_name) or "-" %]</td>
-                  <td>[% release.combined_track_count or "-" %]</td>
-                  <td>[% release_dates(release.events) %]</td>
-                  <td>[% release_countries(release.events) %]</td>
+                  <td>[% artist_credit(entity.artist_credit) %]</td>
+                  <td>[% html_escape(entity.combined_format_name) or "-" %]</td>
+                  <td>[% entity.combined_track_count or "-" %]</td>
+                  <td>[% release_dates(entity.events) %]</td>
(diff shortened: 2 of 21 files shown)


root/release_group/merge.tt
```diff
@@ -22,7 +23,8 @@
                     <td>
                         <input type="hidden" name="merge.merging.[% loop.index %]" value="[% entity.id %]" />
                         <input type="radio" name="merge.target" value="[% entity.id %]"
-                            [% 'checked="checked"' IF entity.id == form.field('target').value %] />
+                            [% 'checked="checked"' IF entity.id == form.field('target').value %]
+                            [% 'disabled="disabled"' IF to_merge.size <= 1 %] />
```
yvanzo: It would be nice to avoid showing a non actionable merge page with only one selected entity, either by redirecting to the still selected entity, or by only allowing to cancel the merge when two entities are selected for merge.

root/release_group/merge.tt
```diff
@@ -31,14 +33,27 @@
                     <td>[% entity.l_type_name %]</td>
                     <td class="c">[% entity.first_release_date.year or '&#x2014;' %]</td>
                     <td>[% entity.release_count %]</td>
+                    <td >
+                      <a href="/release_group/merge?remove=[% entity.id %]&submit=remove">
+                        <button type="button" class="remove-item icon" title="[% l('Remove from merge') %]"></button>
```
yvanzo: It would be more consistent to use the same full-text link (and column name) as in [aliases list](https://test.musicbrainz.org/artist/72c536dc-7137-4477-a521-567eeb840fa8/aliases) (for example).

reosarevok: Hmm. That takes a lot more space, which is why I went with this, but I guess it'd work too. It just might be a bit confusing to just say "Remove" - that kinda suggests you remove the entity itself, to me?

root/release_group/merge.tt
reosarevok: Made that change.

root/release_group/merge.tt
yvanzo: Maybe “Deselect”?

reosarevok:
Rebased on top of https://github.com/metabrainz/musicbrainz-server/pull/1200 to use the same "actions" CSS here.

root/release_group/merge.tt
mwiencek: TBH, I liked the icon better, and the title text "Remove from merge" is clearer and more accessible than either shorter string. "Deselect" can be confusing since the page itself is for selecting the merge target, so it sounds like it's undoing the selection.

So if consistency is an issue, I'd rather change the aliases list to use an icon too. :)

mwiencek: IIRC, chhavi's mockups also heavily used icons for editing actions, e.g. small pencils. This seems smarter to me - people can process images much faster than text, and they allow adding more detailed and accessible descriptions as title text or aria attributes.

reosarevok: @yvanzo: what do you think?

mwiencek: > It is displayed at least 3 times, one above the other. It’s clearly undoing the selection of one element, not the whole selection.

I actually agree with you, and don't think "Deselect" would be confusing to many users — nor would "Remove." Just that neither is particularly helpful to someone using a screen reader who can't view the whole page at once.

At least if a user thinks they're purging the entity from the database by clicking the icon, they'll soon find out they didn't (and maybe we should be glad :P).

> Full-text actions are not just in the aliases list

We should make them consistent at some point, IMO, but I also think:
1. We shouldn't make a new thing worse for the sake of consistency
2. We shouldn't have to change every old thing for the sake of consistency before adding a better new thing

(i.e. I'd prefer incremental progress over having to fix everything at once, unless it's easy to do so of course :))

root/release_group/merge.tt
```diff
@@ -14,11 +14,14 @@
...
                     <th class="count c">[%- l('Releases') -%]</th>
+                    [% IF to_merge.size > 2 %]
+                        <th style="width: 1em" />
+                    [% END %]
                 </tr>
             </thead>
             <tbody>
             [% FOR entity=to_merge %]
-                 <tr class="[% loop.parity %]">
+                <tr class="[% loop.parity %]">
```
reosarevok: This was seemingly off by one.

root/release_group/merge.tt
reosarevok: Made the change back.

root/release_group/merge.tt
```diff
@@ -31,6 +34,12 @@
                     <td>[% entity.l_type_name %]</td>
                     <td class="c">[% entity.first_release_date.year or '&#x2014;' %]</td>
                     <td>[% entity.release_count %]</td>
+                    [% IF to_merge.size > 2 %]
+                        <td >
```
mwiencek:
```suggestion
                        <td>
```

root/release_group/merge.tt
```diff
@@ -31,6 +34,12 @@
...
+                            <a href="/release_group/merge?remove=[% entity.id %]&submit=remove">
+                              <button type="button" class="remove-item icon" title="[% l('Remove from merge') %]"></button>
+                        </td>
```
mwiencek: Looks like there's no closing `</a>` here

root/release_group/merge.tt
```diff
@@ -14,11 +14,14 @@
                     <th>[%- l('Type') -%]</th>
                     <th class="year c">[%- l('Year') -%]</th>
                     <th class="count c">[%- l('Releases') -%]</th>
+                    [% IF to_merge.size > 2 %]
+                        <th style="width: 1em" />
```
mwiencek: Can we add `aria-label="[% l('Remove from merge') %]"` to the th here since it has no label otherwise?

root/work/merge.tt
```diff
@@ -62,8 +63,8 @@
                     <td>
                       [% work_attributes_list(entity) %]
                     </td>
-                  </tr>
```
reosarevok: This seemed to be a completely unmatched `</tr>` (just TT things) but let me know if I'm being stupid.

root/release/merge.tt
```diff
@@ -19,30 +19,32 @@
                     <th>[% l('Label') %]</th>
                     <th>[% l('Catalog#') %]</th>
                     <th>[% l('Barcode') %]</th>
+                    [%- remove_from_merge_header(to_merge) -%]
                 </tr>
             </thead>
             <tbody>
-            [%- FOR release=to_merge -%]
+            [%- FOR entity=to_merge -%]
```
reosarevok: Renaming release to entity for consistency (and for ease of reusing the current macro).

root/area/merge.tt
```diff
@@ -4,16 +4,29 @@
     <p>[% l('You are about to merge the following areas into a single area. Please
              select the area which you would like other areas to be merged into:') %]</p>
     <form action="[% c.req.uri %]" method="post">
-        <ul>
```
reosarevok: Area and artist are the only merges currently using an `<ul>` for some reason. Changed this to a table for consistency (and to reuse the same macro).

reosarevok:
This should now work for all entities.

PR #1200, opened 2019-09-09
MBS-10347: Make action columns take only needed space

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10347

Actions columns are never the main focus of a table, so they should take as little space as they need for "Remove", "Edit", or whatever else goes in them. There's already a min-width of 10em, so they won't look too tiny anyway.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -1015,6 +1015,11 @@ a.tagger-icon {
     min-width: 10em;
 }
 
+.actions {
+    white-space: nowrap;
+    width: 1px;
+}
+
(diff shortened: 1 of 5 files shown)


root/static/styles/layout.less
```diff
@@ -1015,6 +1015,11 @@ a.tagger-icon {
     min-width: 10em;
```
mwiencek: I'd think you can remove this rule now, and the `actions-header` class with it, since the column will never wrap now (it was added for [MBS-5395](https://tickets.metabrainz.org/browse/MBS-5395), but having `white-space: nowrap` resolves that just as well).

PR #1201, opened 2019-09-09
More WS improvements

mwiencek:
This fixes an issue where the WS would *still* try to contact the template renderer (timing out).

It fixes another type of timeout caused by requests that return a ridiculous number of tracks; release browse-type requests are now limited to returning at most 500 tracks. (Many that included box sets would try to return thousands, invariably timing out, and due to a bug in the discid endpoint, potentially pass an array of millions to `load_subobjects`.)

(Note: Release lists will always include at least one release, so >500 tracks can still be returned if the first release has that many.)

The PR also makes a bunch of performance improvements I found while profiling (using `Devel::NYTProf`), which improves the latency of large requests further. The measured improvement depends on the type of request, but the difference can be quite large, in many cases 1.5x to 2x as fast on my development machine.

I'd like to put this on test.mb for further inspection and measurements.


lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
```diff
@@ -261,7 +261,7 @@ sub serialize_relationships {
             join("\t",
                  $_->link->type->name,
                  (sprintf "%09d", $_->link_order // 0),
-                 $_->target_key)
+                 $_->target->id)
```
yvanzo: That doesn’t seem to be equivalent to the previous `target_key`.

mwiencek: Yeah, that's true, though the order is mostly arbitrary to ensure it's stable, so I picked the smallest working key.

Actually I'm not sure this code is needed after b90e1e7f3c1d96a7882196f734a857d7fe718fc3, we can probably just return them in the order they're loaded.

mwiencek: There's also an old issue where XML element attributes are serialized in random order, so `Controller::Role::ETags` isn't very effective and makes most of the sorting we do for naught. I haven't checked if the `XML::LibXML` switch fixes that.

mwiencek:
I put a test WS container with this branch on queen, and ran `ab -n 5` against it and the prod container with the URLs below.

```
/ws/2/release/8d46fe71-e6b4-4fdf-b5f9-ba8876ccf881?inc=release-groups+media+discids+recordings+artist-credits+artists+aliases+labels+isrcs+collections+artist-rels+release-rels+url-rels+recording-rels+work-rels+recording-level-rels+work-level-rels
Time per request (prod):       4064.384 [ms] (mean)
Time per request (ws-fast):       2579.475 [ms] (mean)

/ws/2/release/?artist=164f0d73-1234-4e2c-8743-d77bf2191051&limit=100&offset=10&inc=recordings+artist-credits
Time per request (prod):       776.168 [ms] (mean)
Time per request (ws-fast):       275.342 [ms] (mean)

/ws/2/release/e9ebce87-ed2f-4c7f-a778-25151efe52d4?inc=release-groups+media+discids+recordings+artist-credits+artists+aliases+labels+isrcs+collections+artist-rels+release-rels+url-rels+recording-rels+work-rels+recording-level-rels+work-level-rels
Time per request (prod):       8167.081 [ms] (mean)
Time per request (ws-fast):       4144.386 [ms] (mean)

/ws/2/discid/a1tDhxPTisO0mJlKqK2tYslCwG8-?inc=recordings+artist-credits&cdstubs=no&toc=1+99+332850+0+544+2244+5935+9475+13073+16892+20483+24133+28030+31035+34861+38953+42431+46438+49589+53537+57109+60818+64590+68272+71958+75622+79249+82764+86433+90211+93829+97577+101255+104882+108665+112392+116173+119864+123445+127116+130823+134439+138142+141705+145341+149014+152758+156399+160067+163319+167523+171173+174967+178580+182139+185818+189375+193018+196780+200645+204201+207069+211549+215270+218737+222744+226246+229864+233537+237024+240846+244274+248310+251924+255832+259107+262732+266545+270288+273957+277488+280939+284973+288683+292397+295987+299781+301707+303374+305151+307088+308746+310789+312646+314112+318121+320154+321965+323932+326862+329951+331733
Time per request (prod):       N/A (cannot complete)
Time per request (ws-fast):       1461.052 [ms] (mean)
```

reosarevok:
> release browse-type requests are now limited to returning at most 500 tracks

Is there an offset option added to get the other tracks?

lib/MusicBrainz/Server/Entity/ArtistCredit.pm
```diff
@@ -35,10 +36,15 @@ has 'artist_count' => (
...
     my ($a, $b) = @_;
+
     return 0 unless
         (defined($a) && defined($b)) &&
-        (ref($a) eq ref($b)) &&
-        ($a->name_count == $b->name_count);
+        (ref($a) eq ref($b));
+
+    # refaddr is needed since == is overloaded
+    return 1 if refaddr($a) == refaddr($b);
```
yvanzo: How does it differ from the above `(ref($a) eq ref($b))`?

mwiencek: `ref` returns a string with the type of reference or name of the class (so `'MusicBrainz::Server::Entity::ArtistCredit'`, for example). That's checking that they're both artist credits. https://perldoc.perl.org/functions/ref.html

`refaddr` returns the memory address, a number. That line checks if they're the same instance in memory. Normally `==` would accomplish that, but it's overloaded to point to `is_equal`, so it'd cause infinite recursion. https://metacpan.org/pod/Scalar::Util#refaddr

yvanzo:
@reosarevok: There is the basic offset option to get further releases, since it always returns at least one release, for example: [https://musicbrainz.org/ws/2/release?area=…&**limit=1**&**offset=42**](https://musicbrainz.org/ws/2/release?area=8a754a16-0027-3a29-b6d7-2b40ea0481ed&limit=1&offset=42).
@mwiencek: It’d be neat to somehow allow client to detect when the load limit has been reached.

mwiencek:
Yeah, limit/offset work as usual for browse requests, though the fuzzy TOC lookup doesn't have those parameters. I'll push a commit to add them.

> @mwiencek: It’d be neat to somehow allow client to detect when the load limit has been reached.

You should be able to detect it by checking if the number of releases in the response is less than min((count - offset), limit).

lib/MusicBrainz/Server/ControllerBase/WS/2.pm
```diff
@@ -310,6 +311,33 @@ sub make_list
...
+sub limit_releases_by_tracks {
+    my ($self, $c, $releases) = @_;
+
+    my $track_count = 0;
+    my $release_count = 0;
+
+    for my $release (@{$releases}) {
+        $c->model('Medium')->load_for_releases($release);
+        $track_count += sum map { $_->track_count } $release->all_mediums;
+        last if $track_count > 500 && $release_count > 0;
```
yvanzo: Just curious, how did you get to pick this particular number for the limit?
Would that be alright to define it as a constant or even as a DBDefs entry?

yvanzo:
@mwiencek: I copied your remark on best match in fuzzy TOC lookups to [MBS-10354](https://tickets.metabrainz.org/browse/MBS-10354) for the record. These WS improvements are quite impressive! Maybe these are worth an Improvement ticket too so as to notify editors of behavior changes in release notes.

lib/MusicBrainz/Server/Controller/WS/2/Release.pm
```diff
@@ -190,6 +188,18 @@ sub release_toplevel {
         }
     }
 
+    if (@ac_entities) {
+        $c->model('ArtistCredit')->load(@ac_entities);
+
+        my @acns = map { $_->all_names } uniq_by { refaddr($_) }
```
reosarevok: What's the benefit of uniq_by refaddr rather than just uniq here?

mwiencek: Don't think there was a reason there. Changed it.

mwiencek:
Added these tickets:
https://tickets.metabrainz.org/browse/MBS-10404
https://tickets.metabrainz.org/browse/MBS-10405
https://tickets.metabrainz.org/browse/MBS-10406
https://tickets.metabrainz.org/browse/MBS-10407

Additionally, I've marked https://tickets.metabrainz.org/browse/MBS-5322 as being resolved by this.

But I goofed slightly and merged this before I could add the tickets numbers to the commits.

PR #1204, opened 2019-09-10
MBS-5083: Display recording artist on release view

reosarevok:
https://tickets.metabrainz.org/browse/MBS-5083

There's currently no good way of telling if the  recording artist is different from the track artist from a release page without actually opening the recording page for each recording.

This patch makes it so that if the track and recording ACs differ, the recording AC is displayed under the track name (since the track artist column, which might seem like a more obvious choice, doesn't appear at all if all track artists match the release artist).

This is particularly useful in two cases:
a) when a user has fixed a track artist but forgotten to fix the recording artist, or vice versa.
b) for classical music, where the recording artist should usually be different from the track artist.

### DIFF
--- a/root/components/medium.tt
+++ b/root/components/medium.tt
@@ -48,7 +48,13 @@
     </td>
     <td>
       [%- link_recording(track.recording, 'show', track.name) -%]
-      <br/>
+      [% # Show recording artist only to logged in users to avoid confusing visitors with recordings -%]
+      [% IF c.user_exists && track.artist_credit_id != track.recording.artist_credit_id %]
+          <div class="small">
+          [% l('Recording artist:') %]
+          [% artist_credit(track.recording.artist_credit) %]
+        </div>
+      [% END %]
       <div class="ars" style="display: none;">
         [% grouped_relationships(track.recording) %]
       </div>
--- a/lib/MusicBrainz/Server/Data/Medium.pm
+++ b/lib/MusicBrainz/Server/Data/Medium.pm
@@ -329,6 +329,7 @@ sub load_related_info {
     $self->c->model('ArtistCredit')->load(@tracks);
 
     my @recordings = $self->c->model('Recording')->load(@tracks);
+    $self->c->model('ArtistCredit')->load(@recordings);
     $self->c->model('Recording')->load_meta(@recordings);
     $self->c->model('Recording')->load_gid_redirects(@recordings);
     $self->c->model('Recording')->rating->load_user_ratings($user_id, @recordings) if $user_id;
--- a/t/selenium/release-editor/The_Downward_Spiral.html
+++ b/t/selenium/release-editor/The_Downward_Spiral.html
@@ -2543,6 +2543,7 @@ Nothing Records</td>
 	</td>
 </tr>
 <!-- Reset the track numbers. -->
+<tr>
 	<td>clickAndWait</td>
 	<td>css=.tabs a[href$="/edit"]</td>
 	<td></td>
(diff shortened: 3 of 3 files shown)


mwiencek:
> (since the track artist column, which might seem like a more obvious choice, doesn't appear at all if all track artists match the release artist).

Can we also make it appear if any of the recording artists differ, then?

Showing this under the title doesn't seem too bad, but the artist column still seems more fitting overall.

reosarevok:
We could, but it feels like it might be odd, if people are already used to it appearing only if track artists differ. I would probably start looking for the broken track artist, at least at first...

root/components/medium.tt
```diff
@@ -48,7 +48,12 @@
     </td>
     <td>
       [%- link_recording(track.recording, 'show', track.name) -%]
-      <br/>
+      [% IF track.artist_credit_id != track.recording.artist_credit_id %]
+          <div class="small">
+          [% l('Recording artist:') %]
```
mwiencek: What about reusing the `Recording by {artist}` string we have in the recording header?

reosarevok: Both would probably work, but I feel (since it's not under a recording but a track) "Recording artist: " is slightly more clear. I can also see the usefulness of reusing a string though. @yvanzo: what do you think?
Both options: 
Screenshot from 2019-10-03 20-33-19
Screenshot from 2019-10-03 20-37-14

mwiencek: I can see the argument that "Recording by ... " right under the track could (wrongly) hint that the above name is the recording name, though it seems unlikely to cause confusion in practice since it's in a track listing (which can be expected to lists tracks). But I'm fine with "Recording artist:" if you think it's more clear.

reosarevok:
Selenium issues fixed (thanks @mwiencek!!).

This is not really addressed with #1207 - they complement each other. In most cases, people fix recording artists from the release page, so that's where they should see the issues. If the recording artist isn't incorrect, this will only be seen in cases where it *should* be different, like classical music (as per the earlier screenshots). Below are examples of when it should be useful in non-classical editing: 

Recording artist differs for no clear reason (editors should think whether both should be Diamond Head or both should be [no artist], probably).
Screenshot from 2019-09-10 13-54-16

Someone fixed the track artist, but forgot to fix the recording artist, and now they don't match:
Screenshot from 2019-10-08 10-03-52

yvanzo:
There are many legitimate cases for having a track artist credit different from recording artist credit, and it is not limited to classical style, it can just be small differences in the join phrases or different artist aliases. I’m still not convinced that benefits for advanced editors are worth the confusion for visitors (people visiting the website, not necessarily having an account, not necessarily having visited MusicBrainz beforehand). There should be a way to content both.

reosarevok:
This seems like a terrible idea to me tbh. This is not a relationship, that'd be like if clicking this button also moved track artists down under "credits"

mwiencek:
I agree it'd seem a bit odd to display the artist credits for recordings down there, but not for other entities. (It'd also require a lot of code changes to something I'm currently rewriting in JavaScript, which I'd like to avoid. :))

I initially suggested the recording artist should be displayed in the track artist column (only problem being we don't always display that column). And I wouldn't expect the artists there to be moved, so moving them to the title column doesn't really change that (despite now being displayed next to relationships).

mwiencek:
Sounds good! I'll change the base branch of the PR to beta so I can merge it there.

didierboudon:
My feedback from beta is that this fix is awesome!

PR #1205, opened 2019-09-10
Allow showing tracklists everywhere when attaching/viewing discIDs

reosarevok:
### Implement MBS-5225

When importing a discID normally, you get given a nice table with a lot of data, including tracklist (see cdtoc/attach_list.tt). 
When importing it from a CDStub, a list of "Possible Mediums" was given, with only release title, artist, and medium number, forcing users to choose semi-blindly.

This patch changes the table to be more similar to the one in attach_list, with both more info and collapsible tracklists. attach_list itself can't directly be reused without changing the way we look for the possible mediums, but this reuses as many components as possible.

The second commit allows seeing tracklists both on the list of "Matching CDs" shown in Lookup CD, and on the list of mediums attached to a discid on cdtoc/index.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/CDTOC.pm
+++ b/lib/MusicBrainz/Server/Controller/CDTOC.pm
@@ -42,6 +42,9 @@ sub _load_releases
     my ($self, $c, $cdtoc) = @_;
     my @medium_cdtocs = $c->model('MediumCDTOC')->find_by_discid($cdtoc->discid);
     my @mediums = $c->model('Medium')->load(@medium_cdtocs);
+    $c->model('Track')->load_for_mediums(@mediums);
+    my @tracks = map { $_->all_tracks } @mediums;
+    $c->model('Recording')->load(@tracks);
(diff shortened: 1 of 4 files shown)


reosarevok:
Added a second commit to this that makes it show in even more places.

For the first commit, with a full DB at least, see
http://localhost:5000/cdtoc/attach?toc=1%2011%20261300%20183%2023690%2047995%2078523%2093365%20111708%20147658%20166640%20184125%20221213%20243485

For the second, with a full DB at least, see 
http://localhost:5000/cdtoc/pYouMn_9fJlpZuMPtRCRajKsdfc-
http://localhost:5000/cdtoc/attach?toc=1+13+277777+150+24715+43800+75377+96035+114327+136650+159417+177980+192980+217122+232110+254127

yvanzo (approved):
Thanks, the first link doesn’t find _any CDs with the TOC_ anymore, but other links have been helpful.

yvanzo:

mwiencek (approved):

PR #1207, opened 2019-09-11
MBS-10355: Display track artist on recording pages

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10355

This patch adds a "Track artist" column to recording pages. While in some cases that will be the same as the release and recording artist, IMO it makes sense to always show it for consistency rather than hiding it only when every single track in the page happens to have the same artist as its release or as the recording.

### DIFF
--- a/root/recording/RecordingIndex.js
+++ b/root/recording/RecordingIndex.js
@@ -35,8 +35,10 @@ type Props = {|
 |};
 
 const RecordingAppearancesTable = ({
+  recording,
   tracks,
 }: {
+  recording: RecordingT,
   tracks: $ElementType<Props, 'tracks'>,
 }) => (
   <table className="tbl">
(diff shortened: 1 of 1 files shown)


root/recording/RecordingIndex.js
```diff
@@ -91,6 +94,14 @@ const RecordingAppearancesTable = ({
                   </td>
                   <td>{isolateText(track.name)}</td>
                   <td>{formatTrackLength(track.length)}</td>
+                  {/* The class being added is for usage with userscripts */}
+                  <td className={
+                    track.artistCredit.id === recording.artistCredit.id
+                      ? null
+                      : 'track-recording-clash'}
```
mwiencek: Something like `artist-variation` might be more consistent with the `name-variation` class we already have, but it looks fine otherwise.

reosarevok: By comparison with name-variation, that would kinda sound to me like "the name is the same but the artists aren't", but maybe not :) @yvanzo, what do you think?

mwiencek: Okay, `artist-credit-variation`? :) Mentioning artists somehow would be good though, since it's only indicating if the ACs clash.

yvanzo: Perfecto!

PR #1208, opened 2019-09-12
MBS-10356: Replace <b> with <strong>

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10356

This replaces all of our `<b>` messages, including the ones not yet ported to JS, since we will eventually need them to be `<strong>` anyway. Ideally we should also fix the translations (which this will break, but can be reused by simply changing `<b>` to `<strong>` there as well). @yvanzo: any suggestions on how to do that? :) 

I added some eslint fixes to files I was touching anyway. I fixed all except no-weak-types, since I'm not sure what's the right fix for that one. We can either merge it as-is, or I can fix those too if you let me know how :)

### DIFF
--- a/root/instrument/edit_form.tt
+++ b/root/instrument/edit_form.tt
@@ -14,6 +14,16 @@
           [%- r.textarea('description', { rows => 5 }) -%]
           [%- field_errors(form, 'description') -%]
       [%- END -%]
+      [% WRAPPER form_row %]
+        <p>
+          [% # When converting to React, please move the list below to expand2react.js and import from there # %]
+          [% l('HTML tags allowed in the description: {tag_list}.', 
+            {tag_list => comma_only_list(
+              ['a', 'abbr', 'br', 'code', 'em', 'li', 'p', 'span', 'strong', 'ul'],
+            )}) %]
+        </p>
+      [%- END -%]
+
     </fieldset>
 
     [% PROCESS 'forms/relationship-editor.tt' %]
--- a/root/static/scripts/tests/i18n/expand2.js
+++ b/root/static/scripts/tests/i18n/expand2.js
@@ -1,5 +1,6 @@
 import test from 'tape';
 import React from 'react';
+
 import {VarArgs} from '../../common/i18n/expand2';
 import expand2html from '../../common/i18n/expand2html';
 import expand2text from '../../common/i18n/expand2text';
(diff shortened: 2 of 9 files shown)


root/static/scripts/tests/i18n/expand2.js
```diff
@@ -131,15 +131,14 @@ test('expand2', function (t) {
...
+    'A {apple_fruit|<strong>{prefix} APPLE!</strong>}',
     {apple_fruit: 'http://www.apple.com', prefix: 'dang'},
-    'A <a href="http://www.apple.com"><b><strong>dang APPLE!</strong></b></a>',
+    'A <a href="http://www.apple.com"><strong>dang APPLE!</strong></a>',
   );
 
   expandText('{x:y|}', null, '{x:y|}');
   expandText('{x:y|}', {x: true}, 'y');
   expandText('{x:y|}', {x: false}, '');
-  expandHtml('{x:<b>|</b>|}', {x: true}, '<b>|</b>');
```
mwiencek: Can you replace those with `<strong>` too? (unclear why the test was removed)

reosarevok: I removed it because it used <b> but I guess it's a more general test, so sure. Will do.

mwiencek:
If it's not trivial (or is annoying) to fix all the strings on Transifex, a better idea might be to keep the existing `<b>` tags but have the `expand2` parser output `<strong>` for them.

yvanzo:
There are only two strings with `<b>` so far, it is okay to break the corresponding translations, and trivial to fix it on Transifex since it doesn’t require any linguistic skill: you can just reuse the previous translation and replace `b` tags with `strong` tags.

root/instrument/edit_form.tt
```diff
@@ -14,6 +14,15 @@
           [%- r.textarea('description', { rows => 5 }) -%]
           [%- field_errors(form, 'description') -%]
       [%- END -%]
+      [% WRAPPER form_row %]
+        <p>
+          [% l('HTML tags allowed on description: {tag_list}.',
```
mwiencek:
```suggestion
          [% l('HTML tags allowed in the description: {tag_list}.', 
```
sounds more englishy to me

mwiencek:
This looks pretty much ready other than conflicts.

root/instrument/edit_form.tt
```diff
@@ -14,6 +14,15 @@
           [%- r.textarea('description', { rows => 5 }) -%]
           [%- field_errors(form, 'description') -%]
       [%- END -%]
+      [% WRAPPER form_row %]
+        <p>
+          [% l('HTML tags allowed in the description: {tag_list}.', 
+            {tag_list => comma_only_list(
+              ['a', 'abbr', 'br', 'code', 'em', 'li', 'p', 'span', 'strong', 'ul'],
```
yvanzo: Please define this list along with `htmlTagName` in `root/static/scripts/common/i18n/expand2react.js`.

reosarevok: This can't really be done just yet since this is a TT file, but I added a comment here to move it to expand2react once this is converted.

PR #1209, opened 2019-09-13
MBS-10361: Move edit note help for better tabbing

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10361

Users will want to tab directly to the submit button on edit pages, which the current position of the edit note help prevents (at least in Chrome). This moves it above the note field, but makes it .small to ensure the user doesn't confuse it with an existing edit note.

### DIFF
--- a/root/edit/notes.tt
+++ b/root/edit/notes.tt
@@ -19,17 +19,16 @@
           <div class="add-edit-note edit-note"
               [% 'style="display: none"' IF hide -%]>
               <h3 class="owner">[% link_entity(c.user) %]</h3>
-              <div class="edit-note-text">
-                  <textarea class="edit-note" rows="5" placeholder="[% l('Add an edit note') %]"
-                      name="enter-vote.vote.[% index %].edit_note"></textarea>
-              </div>
-            <p>
+              <p class="edit-note-help">
                   [% l('Edit notes support {doc_formatting|a limited set of wiki formatting options}.
                         Please do always keep the {doc_coc|Code of Conduct} in mind when writing edit notes!',
                        { doc_coc => { href => doc_link('Code_of_Conduct'), target = '_blank' },
                          doc_formatting => { href => doc_link('Edit_Note'), target => '_blank' } }) %]
-            </p>
-
+              </p>
+              <div class="edit-note-text">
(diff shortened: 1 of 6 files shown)


reosarevok:
I changed it to only be cursive, but not small and grey. I feel that still differentiates it enough from the edit note text while probably making it easier to read. Opinion?
Screenshot from 2019-10-07 23-28-53
Screenshot from 2019-10-07 23-28-55

mwiencek:
The issue with editors being confused seems mostly theoretical to me - it only looks like a note at first if you glance at the screen without reading it. And it only ever appears between your own name and the edit note text area, never under another user's name (so you couldn't think that another user wrote it).

So I'd personally think it'd be fine even without the italics. That said, I don't think the italics hurt, and they do distinguish the help text from typical notes which aren't in italics somewhat.

yvanzo:
FWIW, in entity editing form, edit note field has a label and edit note help is indented.

mwiencek:
@yvanzo I'm not sure I understand what you're proposing here, can you clarify?

yvanzo:
Per yesterday’s discussion, it would be best to avoid heterogeneous styling. Maybe having a style similar to `ar-desc` class, like in pull request https://github.com/metabrainz/musicbrainz-server/pull/1217, would do?

yvanzo:
Rebased on latest master branch to get around CircleCI build errors.

reosarevok:
Why not just use the same style completely, including background? Wouldn't that be the most consistent?

yvanzo:
I did not set `background-color` probably because it made no difference in edits list. Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1209/commits/12fe00dad9bc1047d0e1746ea52be1982eade4bc.

yvanzo:
Yes because there are a few differences:
* edit note help is in a `p` tag while other forms help is in a `div`;
* edit note help has extra indent by `1em` to align with below `textarea` and to distinguish from actual edit note.

I just refactored these via the added mix-in `help-box` in commit https://github.com/metabrainz/musicbrainz-server/pull/1209/commits/b2446e93385c724ff5da27609e267875bcdd18b9 to reuse what can be.

(Note: adding `display: table` doesn’t impair relationship description display in relationship editor.)

PR #1211, opened 2019-09-16
MBS-5460/MBS-5572: Add shift-click selecting to more places

reosarevok:
https://tickets.metabrainz.org/browse/MBS-5460
https://tickets.metabrainz.org/browse/MBS-5572

The patch extends the support for shift-click (de)selecting of checkboxes to the release editor recording tab and to the artist credit list on artist/edit.

This has a bunch of eslint fixes included. I didn't fix all of the issues, just some that were annoying me, but I feel that's still better than before. If preferred, I can drop the eslint commits or split them into a separate PR.

### DIFF
--- a/root/static/scripts/release-editor/init.js
+++ b/root/static/scripts/release-editor/init.js
@@ -54,11 +54,22 @@ releaseEditor.init = function (options) {
     });
 
     /*
+     * Allow using range-select (shift-click) on the change recording artist
+     * and change recording title checkboxes in the Recordings page.
+     */
+    MB.Control.RangeSelect(
+        '#track-recording-assignation input.update-recording-title[type="checkbox"]',
+    );
+
+    MB.Control.RangeSelect(
+        '#track-recording-assignation input.update-recording-artist[type="checkbox"]',
+    );
+
+    /*
      * Allow pressing enter to advance to the next tab. The listener is added
      * to the document and not #release-editor so that other events can call
      * preventDefault if necessary.
      */
-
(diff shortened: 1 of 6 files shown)


root/static/scripts/release-editor/init.js
```diff
@@ -53,12 +53,31 @@ releaseEditor.init = function (options) {
...
+                track.titleDiffersFromRecording();
+                track.artistDiffersFromRecording();
+            });
+        });
+
+        setTimeout(() => {
+            const $checkboxes1 = $('#track-recording-assignation').find('input.update-recording-title[type="checkbox"]');
+            MB.Control.RangeSelect($checkboxes1);
+            const $checkboxes2 = $('#track-recording-assignation').find('input.update-recording-artist[type="checkbox"]');
+            MB.Control.RangeSelect($checkboxes2);
```
mwiencek: I'm actually not 100% sure this always returns different DOM nodes when it runs. Or at least I feel uneasy relying on that. If we attach duplicate event handlers on the same checkboxes, that could be bad.

I think it'd be much better to use [event delegation](https://learn.jquery.com/events/event-delegation/) for this: that is, pass the selector string (like `'#track-recording-assignation input.update-recording-title[type="checkbox"]'`)  to `MB.Control.RangeSelect`, then attach a single event handler on the body element that listens for events on that selector. The selector can also be used inside the handler to get all the relevant checkbox nodes.

With that, we wouldn't need this whole `utils.withRelease` hack, since we'd just be attaching an event handler on the body once after the page loads.

PR #1212, opened 2019-09-17
MBS-10370: Add "direction: forward" to XML WS rels

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10370

The JSON WS already outputs direction either way, but the XML one only outputs direction if it's backward. mmd specifies forward is allowed, so this should cause no issues.


reosarevok:
Conflicts fixed.

PR #1214, opened 2019-09-18
MBS-4776: Require confirmation for VA track artist

reosarevok:
https://tickets.metabrainz.org/browse/MBS-4776

We have a warning for Various Artists used as a track artist, but people keep wrongly selecting it. This adds a checkbox so that they have to explicitly confirm they want to use VA, which hopefully will make them think twice about it (we can't disallow it completely since it *is* correct in some rare cases).

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -338,6 +338,13 @@
           <strong>[% l('Warning:') | html_entity %]</strong>
           [% l('You’ve used the {valink|Various Artists} special purpose artist on some tracks below. {valink|Various Artists} should very rarely be used on tracks; please make sure the artists have been entered correctly.', { valink => va_doc_link }) %]
         </p>
+        <label>
+          <input id="confirm-va" type="checkbox" data-bind="checked: confirmedVariousArtists" />
+          [% l('I confirm this is an intentional use of Various Artists on tracks.') %]
+        </label>
+        <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: hasUnconfirmedVariousArtists">
+          [%~ l('If you’re sure the use of Various Artists is correct, confirm it above. Otherwise, please fix the data.') %]
+        </p>
       </div>
     <!-- /ko -->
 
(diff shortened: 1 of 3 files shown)


reosarevok:
I mean, the general idea is that guidelines are overriden by common sense, so if there's a case where common sense says VA is the least bad option, it should be used. Not sure how to word that too well though :)

yvanzo:
Mentioning exceptions exist with a counter-example would be appreciated.

(I don't know of any but you mentioned it exists in your PR comment.)

root/release/edit/tracklist.tt
```diff
@@ -338,6 +338,13 @@
           <strong>[% l('Warning:') | html_entity %]</strong>
           [% l('You’ve used the {valink|Various Artists} special purpose artist on some tracks below. {valink|Various Artists} should very rarely be used on tracks; please make sure the artists have been entered correctly.', { valink => va_doc_link }) %]
         </p>
+        <label>
+          <input id="format-unknown" type="checkbox" data-bind="checked: confirmedVariousArtists" />
```
mwiencek:
```suggestion
          <input type="checkbox" data-bind="checked: confirmedVariousArtists" />
```

root/release/edit/tracklist.tt
```diff
@@ -338,6 +338,13 @@
...
+          [% l('I confirm this is an intentional use of Various Artists on tracks.') %]
+        </label>
+        <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: hasUnconfirmedVariousArtists">
+          [%~ l('If you’re sure the use of Various Artists is correct, please confirm it above. Otherwise, please fix the data.') %]
```
mwiencek: Too many pleases in this sentence. :P
```suggestion
          [%~ l('If you’re sure the use of Various Artists is correct, confirm it above. Otherwise, please fix the data.') %]
```

root/static/styles/layout.less
```diff
@@ -362,9 +362,13 @@ div.warning {
     width: 60em;
     background: @warning-background;
 
-    p {
+    p, label {
         margin: 1em;
     }
+
+    label {
+        display: block
```
mwiencek:
```suggestion
        display: block;
```

PR #1217, opened 2019-09-26
MBS-10382: Explain when to use the Update checkbox

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10382

### DIFF
--- a/root/artist/ArtistMerge.js
+++ b/root/artist/ArtistMerge.js
@@ -69,6 +69,21 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
 
           <FormRowCheckbox
             field={form.field.rename}
+            help={
+              <>
+                <p>
+                  {l('You should only use the checkbox above to fix errors (e.g. typos).')}
+                </p>
+                <p>
+                  {exp.l(
+                    `If a name appears on the cover of a release, don’t check the box:
+                     the artists will still be combined if you don’t, but the 
+                     {doc_acs|artist credits} will be kept as they are now.`,
+                    {doc_acs: '/doc/Artist_Credits'},
+                  )}
+                </p>
+              </>
+            }
             label={l(`Update matching artist and relationship credits to use
                       the new artist’s name`)}
           />
--- a/root/artist/ArtistMerge.js
+++ b/root/artist/ArtistMerge.js
@@ -85,7 +85,7 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
               </>
             }
             label={l(`Update matching artist and relationship credits to use
-                      the new artist’s name`)}
+                      the target artist’s name`)}
           />
 
           <EnterEditNote field={form.field.edit_note} />
--- a/root/components/FormRowCheckbox.js
+++ b/root/components/FormRowCheckbox.js
@@ -7,21 +7,23 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-import React from 'react';
+import * as React from 'react';
 
 import FieldErrors from './FieldErrors';
 import FormRow from './FormRow';
 
 type Props = {
   +field: ReadOnlyFieldT<boolean>,
+  +help?: React.Node,
   +label: string,
   +onChange?: (event: SyntheticEvent<HTMLInputElement>) => void,
 };
 
-const FormRowCheckbox = ({field, label, onChange}: Props) => (
+const FormRowCheckbox = ({field, help, label, onChange}: Props) => (
   <FormRow hasNoLabel>
     <label className="inline">
       <input
+        aria-describedby={help ? `field-help-${field.id}` : null}
         defaultChecked={field.value}
         name={field.html_name}
         onChange={onChange}
@@ -32,6 +34,11 @@ const FormRowCheckbox = ({field, label, onChange}: Props) => (
       {label}
     </label>
     <FieldErrors field={field} />
+    {help ? (
+      <div className="form-help" id={`field-help-${field.id}`}>
+        {help}
+      </div>
+    ) : null}
   </FormRow>
 );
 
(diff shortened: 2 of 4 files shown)


root/artist/ArtistMerge.js
```diff
@@ -65,9 +65,23 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
...
             label={l(`Update matching artist and relationship credits to use
-                      the new artist’s name`)}
+                      the target artist’s name`)}
           />
 
+          <p>
+            {exp.l(
+              `By default, the merge will combine the artists but the previous
+              name(s) will still be used on the {doc_acs|artist credits}. If
+              all the name(s) being merged are incorrect (e.g. typos) and all
```
mwiencek:
```suggestion
               all the names being merged are incorrect (e.g. typos) and all
```
since names can't be singular due to "all," plus "names" being used later in the sentence

(lines needs to be indented an extra space to line up with the first, too)

root/artist/ArtistMerge.js
```diff
@@ -65,9 +65,23 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
...
+              existing credits matching those names should be updated, you can
+              click the checkbox above to do so. Keep in mind artist credits should
+              generally follow what is printed on releases. If an artist has
+              releases using several of the names being merged, do not change the
```
mwiencek:
```suggestion
               releases using any of the names being merged, do not change the
```
(Otherwise it's saying it's okay to update all of the names if only some are unused, right?)

reosarevok: I think I meant "several" as in including the one being merged *into*, but you're right this'd make it clearer.

reosarevok:
Thinking of something like this? I feel it looks a bit weird full-length, maybe the second is better.
Screenshot from 2019-10-18 00-12-58
Screenshot from 2019-10-18 00-21-15

reosarevok:
For the last commit (ReportUser):
Screenshot from 2019-10-18 15-57-33

root/user/ReportUser.js
```diff
@@ -128,20 +128,19 @@ const ReportUser = ({
...
-        </p>
-      </FormRow>
-
       <FormRowCheckbox
         field={form.field.reveal_address}
+        help={
+          <p>
+            {exp.l(
+              `If you don’t want our admins to contact you further regarding 
+              this report, you can uncheck the checkbox above.
```
mwiencek: Heh, good to see you did `s/below/above/` here already (noticed it was still "below" in the screenshot). Though the indentation is off by one:
```suggestion
               this report, you can uncheck the checkbox above.
```

root/components/FormRowCheckbox.js
```diff
@@ -32,6 +33,9 @@ const FormRowCheckbox = ({field, label, onChange}: Props) => (
       {label}
     </label>
     <FieldErrors field={field} />
+    <div className="form-help">
```
mwiencek: This should probably output `null` instead of an empty div if `help` isn't given.

Also, a nice accessibility tweak we can add: ``id={`field-help-${field.id}`}`` to the div here, and ``aria-describedby={help ? `field-help-${field.id}` : null}`` to the input above.

root/artist/ArtistMerge.js
```diff
@@ -64,8 +64,23 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
...
             field={form.field.rename}
+            help={
+              <>
+                <p>
+                  {l('You should only use the checkbox above to fix errors (e.g. typos).')}
+                </p>
+                <p>
+                  {exp.l(
+                    `If the name appears on the cover of a release, don't check the box:
+                    the artists will still be combined but the {doc_acs|artist credits}
```
mwiencek:
```suggestion
                     the artists will still be combined but the {doc_acs|artist credits}
```

root/components/FormRowCheckbox.js
reosarevok: Oooh. I like that :) We should add more of those.

yvanzo:
Oops, it is probably related to radio buttons (for target selection) instead, which has no `label` parent.

root/artist/ArtistMerge.js
```diff
@@ -69,8 +69,23 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
...
             field={form.field.rename}
+            help={
+              <>
+                <p>
+                  {l('You should only use the checkbox above to fix errors (e.g. typos).')}
+                </p>
+                <p>
+                  {exp.l(
+                    `If the name appears on the cover of a release, don't check the box:
+                     the artists will still be combined but the {doc_acs|artist credits}
```
mwiencek: Hrm, it's not totally clear if the info after the colon is what happens if you don't check the box, or if you do. As a dev it's obvious to me from the context, but I wouldn't be so sure for a regular user trying to figure this out.

```suggestion
                     the artists will still be combined if you don’t, but the {doc_acs|artist credits}
```

reosarevok: Good idea, changed.

root/artist/ArtistMerge.js
```diff
@@ -69,8 +69,23 @@ const ArtistMerge = ({$c, form, toMerge}: Props) => {
...
           <FormRowCheckbox
             field={form.field.rename}
+            help={
+              <>
+                <p>
+                  {l('You should only use the checkbox above to fix errors (e.g. typos).')}
+                </p>
+                <p>
+                  {exp.l(
+                    `If a name appears on the cover of a release, don't check the box:
```
mwiencek: one last nitpick (the other "don’t" in the string uses it so it's inconsistent now :))
```suggestion
                    `If a name appears on the cover of a release, don’t check the box:
```

PR #1219, opened 2019-09-28
MBS-5479: Add an "Edit Relationships" tab

reosarevok:
https://tickets.metabrainz.org/browse/MBS-5479

This adds an "Edit Relationships" tab to releases by the "Edit" tab, in order to make the relationship editor easier to find for everyone.

Since most users are currently used to click the link on the sidebar and it also makes sense in that context, this leaves it in there as well - we certainly can use as big a chance as possible for people
to add relationships anyway.

### DIFF
--- a/root/components/EntityTabs.js
+++ b/root/components/EntityTabs.js
@@ -105,6 +105,10 @@ function buildLinks(
     }
   }
 
+  if (entity.entityType === 'release') {
+    links.push(buildLink(l('Edit Relationships'), entity, 'edit-relationships', page));
+  }
+
   return links;
 }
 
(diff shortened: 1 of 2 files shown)


root/components/EntityTabs.js
```diff
@@ -105,6 +105,10 @@ function buildLinks(
     }
   }
 
+  if (entity.entityType === 'release') {
+    links.push(buildLink(l('Edit Relationships'), entity, 'edit-relationships', page));
```
mwiencek: In https://github.com/metabrainz/musicbrainz-server/blob/2baca997d9eb737aff377cc541d1ef0108cfe53f/root/release/edit_relationships.tt#L7 we pass `page => 'edit_relationships'` which I think needs to be changed to use a hyphen to match the third argument here, so that the tab is highlighted correctly once selected.

reosarevok: Oh yeah, thanks :) Fixed!

PR #1220, opened 2019-09-29
MBS-9920: Convert the recording fingerprints page to React

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9920

On top of https://github.com/metabrainz/musicbrainz-server/pull/1200 since it has an action column too.


root/static/scripts/common/components/FingerprintTable.js
```diff
@@ -0,0 +1,92 @@
...
+
+
+const FingerprintTable = ({recording}: {recording: RecordingT}) => {
+  const [tracks, setTracks] = useState([]);
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  // We ensure fetch only runs client-side since it's not in node
+  useEffect(() => {
+    fetch(
+      `//api.acoustid.org/v2/track/list_by_mbid?format=json&disabled=1&jsoncallback=?&mbid=${recording.gid}`,
```
mwiencek: Breaking this up with concatenation would be nice too. (For URLs with a lot of query parameters, I usually break before `?` or `&` to maintain readability.)

root/static/scripts/common/components/FingerprintTable.js
```diff
@@ -0,0 +1,92 @@
...
+                  {track.disabled ? l('Link') : l('Unlink')}
+                </a>
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    ) : isLoaded ? (
+      <p>{l('This recording does not have any associated AcoustIDs')}</p>
+    ) : null
```
mwiencek: If the component is in the process of loading things, it'd be nice to display something like `<p className="loading-message">{l('Loading...')}</p>` here. It took a while to load them while I was testing and I thought the page was broken. (It happens in production too, but seems like a pretty simple thing to add now that the page is using React.)

reosarevok: Good idea, done.

root/static/scripts/common/components/FingerprintTable.js
```diff
@@ -0,0 +1,99 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React, {useEffect, useState} from 'react';
+
+import loopParity from '../../../../utility/loopParity';
+import hydrate from '../../../../utility/hydrate';
+
+function orderTracks(tracks) {
+  tracks.sort(function (a, b) {
```
mwiencek: In my previous comment I meant just moving this inner function that accepts `(a, b)` outside, but keeping the `tracks.sort` call in the component, so that it doesn't need to create a closure at all.

root/static/scripts/common/components/FingerprintTable.js
```diff
@@ -0,0 +1,99 @@
...
+    if (!a.disabled && b.disabled) {
+      return -1;
+    }
+    if (a.id > b.id) {
+      return 1;
+    }
+    if (a.id < b.id) {
+      return -1;
+    }
+    return 0;
```
mwiencek: We can replace all of these lines with `return a.id - b.id;`.

root/static/scripts/common/components/FingerprintTable.js
```diff
@@ -0,0 +1,99 @@
...
+    fetch(
+      '//api.acoustid.org/v2/track/list_by_mbid' +
+      `?format=json&disabled=1&jsoncallback=?&mbid=${recording.gid}`,
+    ).then(
+      function (response) {
+        return response.json();
+      },
+    ).then(
+      function (data) {
+        const orderedTracks = orderTracks(data.tracks);
```
mwiencek: Something like `data.tracks.sort(compareTracks)` here is what I meant to suggest, with `compareTracks` being the function that accepts `(a, b)`. BTW, `.sort` on arrays sorts in-place, so you don't have to assign it to anything.

PR #1221, opened 2019-10-02
Bump React to 16.10.2

mwiencek:
https://reactjs.org/versions


mwiencek:
This should have no breaking changes (see the changelog and versioning policy in the link above), so should be fine to include in the next beta.

PR #1216, opened 2019-09-23
MBS-5588, MBS-5611, MBS-8016, MBS-8356, MBS-10376, MBS-10377: "Set cover art" improvements

reosarevok:
This includes fixes for:

https://tickets.metabrainz.org/browse/MBS-5588
https://tickets.metabrainz.org/browse/MBS-5611
https://tickets.metabrainz.org/browse/MBS-8016
https://tickets.metabrainz.org/browse/MBS-8356
https://tickets.metabrainz.org/browse/MBS-10376
https://tickets.metabrainz.org/browse/MBS-10377

I'm pretty sure it also fixes (thanks to the changes made, even if not intentionally):
https://tickets.metabrainz.org/browse/MBS-5612
https://tickets.metabrainz.org/browse/MBS-7793

PR #1222, opened 2019-10-03
Port the Autocomplete formatters to React

mwiencek:


mwiencek:
This can be tested in the same way as the previous one, by running `./script/compile_resources.sh tests` and navigating to `http://localhost:5000/static/scripts/tests/autocomplete2.html`.

Though it's slightly annoying to switch entity types on that page; you have to paste an MBID of a different type to do that. I could add a `<select>`, but it's extra work for a page I plan to delete soon. :)

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -0,0 +1,433 @@
...
+
+      {series.type ? (
+        showComment(bracketedText(lp_attributes(series.type.name, 'series_type')))
+      ) : null}
+    </>
+  );
+}
+
+function formatWork(work: AutocompleteWorkT) {
+  const typeName = work.typeName;
```
reosarevok: A bit confused about why this one deserves a const, but work.languages doesn't :)

.eslintrc.yaml
```diff
@@ -24,7 +24,7 @@ rules:
   no-await-in-loop: error
   no-cond-assign: warn
   no-constant-condition: [error, {checkLoops: false}]
-  no-control-regex: warn
+  no-control-regex: off
```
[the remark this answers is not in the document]
mwiencek: Moved that to a separate commit

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -0,0 +1,433 @@
...
+  return !nonLatinRegExp.test(str);
+}
+
+function showComment(children, className = 'comment') {
+  return (
+    <span className={`autocomplete-${className}`}>
+      {children}
+    </span>
+  );
+}
```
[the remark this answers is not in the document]
mwiencek: Renamed it to `showExtraInfo` (and renamed `showLineComment` to `showExtraInfoLine`).

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -0,0 +1,433 @@
...
+function formatReleaseEvent(event: ReleaseEventT) {
+  const date = formatDate(event.date);
+  const country = event.country;
+  let countryDisplay = null;
+
+  if (country) {
+    const primaryCode = primaryAreaCode(country);
+    if (primaryCode) {
+      countryDisplay = (
+        <span className={`flag flag-${primaryCode}`}>
```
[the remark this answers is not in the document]
mwiencek: Yeah, the icon CSS was just missing from the test page. Added it

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -0,0 +1,433 @@
...
+        catNos.push(releaseLabel.catalogNumber);
+      }
+    }
+
+    for (const [labelName, catNos] of catNosByLabel) {
+      catNos.sort();
+
+      const catNoDisplay = catNos.length ? (
+        catNos.length > 2 ? (
+          catNos[0] + ' ' + l('…') + ' ' + catNos[catNos.length - 1]
```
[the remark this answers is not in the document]
mwiencek: I didn't see any other uses of this kind, but added a `'{first_list_item} … {last_list_item}'` string for it anyway. Let me know if you had something else in mind.

root/static/scripts/common/components/Autocomplete2/formatters.js
mwiencek: It was to work around that Flow issue where it invalidates your refinements unless you save to a variable first, which the languages didn't need, but agreed it's nicer to have one added in any case.

root/static/scripts/common/components/Autocomplete2/formatters.js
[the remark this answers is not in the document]
mwiencek: > I was also wondering why functions rather than components, but tbh I'm still not sure when to use one over the other sometimes.

There'd be nothing technically wrong with always using components, but I sometimes just prefer to use functions in cases where they do something very simple (wrapping something in a span) and are used heavily as a utility to the point where the component syntax would become a burden. Another case I'd use a function is if it's being used as a map callback, like in `release.events.map(formatReleaseEvent)`.

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -0,0 +1,433 @@
...
+  const country = event.country;
+  let countryDisplay = null;
+
+  if (country) {
+    const primaryCode = primaryAreaCode(country);
+    if (primaryCode) {
+      countryDisplay = (
+        <span className={`flag flag-${primaryCode}`}>
+          <abbr title={country.name}>{primaryCode}</abbr>
+        </span>
```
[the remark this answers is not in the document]
mwiencek: Added the component for it, but left the markup the same for now. Looking at https://github.com/metabrainz/musicbrainz-server/pull/1216/commits/65e50c231958d29fb594597bafd84bc768f4532e, I definitely think that's just due to us not chomping whitespace somewhere (problem being it's nearly impossible to find out where). React on the other hand never outputs spaces unless you add them yourself, like `{' '}` for example.

root/static/scripts/common/components/Autocomplete2/types.js
```diff
@@ -86,14 +86,73 @@ export type Actions =
...
-  +name: string,
+type Appearances<T> = {|
+  +hits: number,
+  +results: $ReadOnlyArray<T>,
+|};
+
+export type AutocompleteAreaT = {|
+  ...AreaT,
+  +primaryAlias: string | null,
+|};
```
yvanzo: Would it be easier to have `+entity: T` like for `SearchResults<T>` instead of extending `AreaT`?

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -0,0 +1,425 @@
...
+  return (
+    <>
+      <br />
+      {showExtraInfo(children, className)}
+    </>
+  );
+}
+
+function formatGeneric(entity, extraInfo) {
+  const name = unwrapNl<string>(entity.name);
```
yvanzo: If `formatGeneric` is called from `formatItem`, then `unwrapNl<string>(entity.name)` is computed twice.

root/static/scripts/common/components/Autocomplete2/types.js
mwiencek: Well, these types just model the format outputted by `/ws/js`, which I'm not really inclined to change on the server at this point. What seems nicer to me is (eventually) just returning normal `AreaT`s etc. from the server and computing `primaryAlias` in JavaScript.

root/static/scripts/common/components/Autocomplete2/formatters.js
mwiencek: From what I can see, `formatItem` only calls `unwrapNl` inside the `if (!item.entityType) { ... }` block at the top, where it also returns, so doesn't reach `formatGeneric` below. So not sure I see that. (I guess it wouldn't matter either way, since `unwrapNl` is a no-op if the argument is already a string.)

yvanzo: Yup, sorry about for misreading (twice) the code here.

mwiencek: No need to be sorry, my code is not as clear as it could be (and a point of reviewing is figuring out what the heck it does). :)

PR #1224, opened 2019-10-05
Enable some Flow features

mwiencek:
The one massive, painful change here is switching to exact object types by default, as discussed [here](https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf), and then removing the old exact object type syntax (which we'll have to do eventually, so may as well get it over with).

The other commits allow us to use some useful stage-3 ES proposals ([optional chaining](https://github.com/tc39/proposal-optional-chaining) & [nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing)), and enable Flow's [strict mode](https://flow.org/en/docs/strict/) on some files to get that started.


root/static/scripts/common/components/TagEditor.js
```diff
@@ -623,9 +624,10 @@ function createInitialTagState(
   });
 
   // Always show upvoted user tags (affects sidebar)
-  for (const t of userTagsByName.values()) {
-    if (t.vote > 0 && !used.has(t.tag)) {
-      combined.push(t);
+  for (const tagName of Object.keys(userTagsByName)) {
+    const tag = userTagsByName[tagName];
+    if (tag.vote > 0 && !used.has(tagName)) {
+      combined.push(tag);
```
mwiencek: `userTagsByName` is a plain object rather than a `Map` now, so it doesn't have a `.values()` method. I've changed it to loop over the object keys instead.

root/account/Preferences.js
```diff
@@ -17,10 +17,10 @@ import type {PreferencesFormPropsT}
   from '../static/scripts/account/components/PreferencesForm';
 import * as manifest from '../static/manifest';
 
-type Props = {|
-  +$c: {user: EditorT} & CatalystContextT,
+type Props = {
+  +$c: {...CatalystContextT, user: EditorT},
```
mwiencek: Both `{user: EditorT} & CatalystContextT` and `{...CatalystContextT, user: EditorT}` should work (both giving exact types), but we tend to use spreads more, so I'm just standardizing on that syntax.

root/static/scripts/common/components/Autocomplete2.js
```diff
@@ -47,15 +47,13 @@ import type {
...
   indexedSearch: true,
-  inputTimeout: null,
   inputValue: '',
   isOpen: false,
   items: EMPTY_ARRAY,
   page: 1,
   pendingSearch: null,
   selectedItem: null,
   statusMessage: '',
-  xhr: null,
```
mwiencek: This (and `inputTimeout`) aren't stored in state at all, they're stored on the instance. Flow just didn't notice until exact object types were enabled.

mwiencek:
Rebased and fixed conflicts in package.json / yarn.lock.

mwiencek:
@yvanzo can we still include this in Mommy Shark? It's going to be a pain to keep concurrent React PRs in sync.

PR #1229, opened 2019-10-11
MBS-10390 and MBS-10398 Place entity edit details

anirudhjain75:


anirudhjain75:
@reosarevok I've made the changes you requested, atleast the ones I understood. I need some more hints with the remaining issues.

anirudhjain75:
Hope it makes sense with react element and text now. Currently using Object type for Content type

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,96 @@
...
+    <table className="details edit-place">
+      <tbody>
+        <tr>
+          <th>{l('Place:')}</th>
+          <td><DescriptiveLink entity={entity} /></td>
+        </tr>
+
+        {display.name ? (
+          <Diff
+            label={l('Name:')}
```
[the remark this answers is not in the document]
anirudhjain75: Done this :)

root/edit/details/AddPlace.js
```diff
@@ -0,0 +1,81 @@
...
+              <td>{display.type.name}</td>
+            </tr>
+          ) : null}
+          {display.address ? (
+            <tr>
+              <th>{l('Address:')}</th>
+              <td>{display.address}</td>
+            </tr>
+          ) : null}
+          {display.area.gid === '' ? null : (
```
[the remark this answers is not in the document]
anirudhjain75: @reosarevok I still don't understand perl well. Hope @mwiencek checks this out 😅

root/static/scripts/common/components/DescriptiveLink.js
```diff
@@ -14,19 +14,21 @@ import ArtistCreditLink from './ArtistCreditLink';
...
 
 const DescriptiveLink = ({
+  allowNew,
   content,
   entity,
   showDeletedArtists = true,
   target,
 }: DescriptiveLinkProps) => {
-  const props = {content, showDisambiguation: true, target};
+  const props = {content, showDisambiguation: true, target, allowNew};
```
reosarevok: You're not actually using allowNew anywhere, as far as I can see, so what's the point of it? :)

root/edit/details/AddPlace.js
```diff
@@ -0,0 +1,103 @@
...
+            <th>{addColonText(l('Place'))}</th>
+            <td><DescriptiveLink allowNew entity={place} /></td>
+          </tr>
+        </tbody>
+      </table>
+      <table className="details add-place">
+        <tbody>
+          <tr>
+            <th>{addColonText(l('Name'))}</th>
+            <td>{lp_attributes(display.name, 'place_type')}</td>
```
reosarevok: This is the actual name of the place, it doesn't need translating.

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,126 @@
...
+            newText={display.comment.new}
+            oldText={display.comment.old}
+            split="\s+"
+          />
+        ) : null}
+        {display.type ? (
+          <FullChangeDiff
+            label={addColonText(l('Type'))}
+            newText={display.type.new.name}
+            oldText={display.type.old.name}
```
reosarevok: I'm expecting these two should be localized? @mwiencek, please confirm.

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,126 @@
...
+        ) : null}
+        {display.address ? (
+          <Diff
+            label={addColonText(l('Address'))}
+            newText={display.address.new}
+            oldText={display.address.old}
+            split="\s+"
+          />
+        ) : null}
+        {display.area && (notDiffContent = display.area.new.gid === display.area.old.gid) ? (
```
reosarevok: I'm a bit confused about the use of assigning and passing notDiffContent here. Wouldn't be easier then to just do `display.area && !(display.area.new.gid === display.area.old.gid)`? What am I missing?

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,126 @@
...
+            oldContent={display.area.old
+                ? <DescriptiveLink entity={display.area.old} />
+                : ''}
+          />
+        ) : null}
+        {display.coordinates ? (
+          <Diff
+            label={addColonText(l('Coordinates'))}
+            newText={formatCoordinates(display.coordinates.new)}
+            oldText={formatCoordinates(display.coordinates.old)}
```
reosarevok: formatCoordinates doesn't work properly if coordinates are missing (for example, if the edit is *adding* coordinates).

Example:
Screenshot from 2019-12-06 11-53-44

root/static/scripts/edit/components/edit/Diff.js
```diff
@@ -21,7 +21,7 @@ export type DiffProps = {
 
 type Props = {
   ...DiffProps,
-  +split: string,
+  +split?: string,
```
reosarevok: Hmm. Isn't there always a split, since a default is defined below? Maybe that doesn't work like that, I'm not sure. @mwiencek?

root/edit/details/EditPlace.js
anirudhjain75: Well I've created consts oldType and newType just incase. Hope that resolved this :)

root/edit/details/EditPlace.js
anirudhjain75: I've make the assignments in the top and Im using the notDiffContent to compare if their gid are same and passing it to FullChangeDiffContent for rendering only if changes exist

root/static/scripts/edit/components/edit/Diff.js
anirudhjain75: default values will just assign default values to props. But flow checks to see if the files which is rendering the components to make sure all the props are passed. At least that's what I think it does 😓

root/static/scripts/common/components/DescriptiveLink.js
anirudhjain75: I'm using allowNew in AddPlace to ensure the place can be displayed with gid being present in the database

root/edit/details/AddPlace.js
anirudhjain75: Corrected it thanks :)

root/edit/details/EditPlace.js
anirudhjain75: Made temporary changes to fix this :)

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,131 @@
...
+            split="\s+"
+          />
+        ) : null}
+        {display.area ? (
+          <FullChangeDiffContent
+            label={addColonText(l('Area'))}
+            newContent={display.area.new
+              ? <DescriptiveLink entity={display.area.new} />
+              : ''}
+            notDiffContent={notDiffContent}
```
mwiencek: We currently don't need a prop like this to control the visibility, since `display.area` won't exist unless the area actually changed (so this would never be false).

I'm guessing you might need this sort of behavior for your in-page edit previews, though? (Assuming it passes in old/new data without checking them for equality.) Even in that case, we still don't need this prop, because we can just perform the check above where `display.area ?` is, with something like `(display.area && display.area.new?.gid === display.area.old?.gid) ?`.

root/utility/coordinates.js
```diff
@@ -8,7 +8,15 @@
  */
 
 export function formatCoordinates(coordinates: CoordinatesT) {
-  const {latitude, longitude} = coordinates;
+  // $FlowFixMe
+  if (coordinates === '') {
```
mwiencek: As I mentioned somewhere above, we need to fix `build_display_data` so that it returns `undef` instead of an empty string. Then you should remove the `$FlowFixMe` and change the function argument type to `coordinates: ?CoordinatesT` (note the added `?`).

root/utility/coordinates.js
```diff
@@ -8,7 +8,15 @@
...
+    return '';
+  }
+  let {latitude, longitude} = coordinates;
+  if (typeof latitude != 'number') {
+    latitude = Number(latitude);
+    longitude = Number(longitude);
+  }
```
mwiencek: This definitely shouldn't happen. :) So let's remove this and fix it on the Perl side rather than adding kludges. I'm guessing you need to change `TO_JSON` in lib/MusicBrainz/Server/Entity/Coordinates.pm to this, maybe

```Perl
    return {
        latitude => 0 + $self->latitude,
        longitude => 0 + $self->longitude,
    };
```

root/edit/details/EditPlace.js
mwiencek: I agree with @reosarevok here. See https://github.com/metabrainz/musicbrainz-server/pull/1229#discussion_r355174318

root/edit/details/EditPlace.js
mwiencek: By localized he meant translated, using `lp_attributes`. I'm not sure the current TT template does that, actually, but we should now in any case. Suggested that again in https://github.com/metabrainz/musicbrainz-server/pull/1229#discussion_r355173889

root/edit/details/EditPlace.js
mwiencek: This needs to be unkludged now. :) See https://github.com/metabrainz/musicbrainz-server/pull/1229#discussion_r355174613

root/static/scripts/edit/components/edit/FullChangeDiffContent.js
```diff
@@ -0,0 +1,29 @@
+/*
```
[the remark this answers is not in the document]
anirudhjain75: Ohh okay. It didn't before though. I'll change it to FullChangeDiff

root/edit/details/AddPlace.js
```diff
@@ -0,0 +1,104 @@
...
+import yesNo from '../../static/scripts/common/utility/yesNo';
+
+type AddPlaceEditT = {
+  ...EditT,
+  display_data: {
+    ...DatePeriodRoleT,
+    address: string | null,
+    area: AreaT,
+    comment: string | null,
+    coordinates: CoordinatesT,
```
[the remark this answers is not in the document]
anirudhjain75: Yea this is one of the thing I said on IRC. guess no-one read it

root/edit/details/EditPlace.js
anirudhjain75: ohh

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,131 @@
...
+  ...EditT,
+  display_data: {
+    address: CompT<string>,
+    area: CompT<AreaT>,
+    begin_date: CompT<PartialDateT>,
+    comment: CompT<string>,
+    coordinates: CompT<CoordinatesT>,
+    end_date: CompT<PartialDateT>,
+    ended: CompT<boolean>,
+    name: CompT<string>,
```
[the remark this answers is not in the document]
anirudhjain75: I'm confused with this. name as defined in Edit.pm is Optional[str]. So I don't it can ever be undefined or can it ? 🤔

root/edit/details/AddPlace.js
```diff
@@ -0,0 +1,104 @@
...
+    name: string,
+    place: PlaceT,
+    type: PlaceTypeT,
```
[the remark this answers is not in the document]
anirudhjain75: Done it :)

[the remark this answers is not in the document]
anirudhjain75: This is also taken care of :)

root/static/scripts/edit/components/edit/FullChangeDiffContent.js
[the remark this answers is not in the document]
anirudhjain75: Do have a look if I've done it right :)

root/edit/details/AddPlace.js
```diff
@@ -0,0 +1,104 @@
...
+  '../../static/scripts/common/components/DescriptiveLink';
+import isDateEmpty from '../../static/scripts/common/utility/isDateEmpty';
+import {formatCoordinates} from '../../utility/coordinates';
+import formatDate from '../../static/scripts/common/utility/formatDate';
+import yesNo from '../../static/scripts/common/utility/yesNo';
+
+type AddPlaceEditT = {
+  ...EditT,
+  display_data: {
+    ...DatePeriodRoleT,
```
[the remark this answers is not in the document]
anirudhjain75: Done

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,131 @@
...
+import FullChangeDiffContent from
+  '../../static/scripts/edit/components/edit/FullChangeDiffContent';
+import {formatCoordinates} from '../../utility/coordinates';
+import formatDate from '../../static/scripts/common/utility/formatDate';
+import yesNo from '../../static/scripts/common/utility/yesNo';
+
+type PlaceEditT = {
+  ...EditT,
+  display_data: {
+    address: CompT<string>,
```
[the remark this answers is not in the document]
anirudhjain75: Do have a look at this again please, I'm not very good at determining the diff between `?` and `| null`

root/edit/details/EditPlace.js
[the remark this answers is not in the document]
anirudhjain75: I don't think so either

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,131 @@
...
+            label={addColonText(l('Disambiguation'))}
+            newText={display.comment.new}
+            oldText={display.comment.old}
+            split="\s+"
+          />
+        ) : null}
+        {display.type ? (
+          <FullChangeDiff
+            label={addColonText(l('Type'))}
+            newText={newType.name}
```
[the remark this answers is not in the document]
anirudhjain75: set newTypeName to `display.type?.new?.name ?? ''` as the type itself may be null :)

root/edit/details/EditPlace.js
anirudhjain75: Don't know if this is still necessary now 🤔 removed the component altogether

root/utility/coordinates.js
anirudhjain75: Done it

root/utility/coordinates.js
anirudhjain75: I don't see this necessary as everything works out after editing edit.pm and create.pm of place entity

root/static/scripts/edit/components/edit/FullChangeDiff.js
```diff
@@ -8,15 +8,24 @@
...
+};
 
-const FullChangeDiff = ({label, newText, oldText}: DiffProps) => (
-  oldText === newText ? null : (
+const FullChangeDiff = ({
+  label,
+  newContent,
+  oldContent,
+}: FullChangeDiffProps) => (
+  _.isEqual(oldContent, newContent) ? null : (
```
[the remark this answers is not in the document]
anirudhjain75: Deep compare is don't because the input to FullChangeDiff can also be a text in some cases

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,134 @@
...
+  ...EditT,
+  +display_data: {
+    +address?: CompT<string>,
+    +area?: CompT<AreaT | null>,
+    +begin_date?: CompT<PartialDateT>,
+    +comment?: CompT<string>,
+    +coordinates?: CompT<CoordinatesT | null>,
+    +end_date?: CompT<PartialDateT>,
+    +ended?: CompT<boolean>,
+    +name: CompT<string>,
```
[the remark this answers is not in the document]
anirudhjain75: This is a bit confusing... but I made the changes :)

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,134 @@
...
+    +place: PlaceT,
+    +type?: CompT<PlaceTypeT> | null,
```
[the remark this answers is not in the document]
anirudhjain75: done this too

root/static/scripts/edit/components/edit/FullChangeDiff.js
```diff
@@ -8,15 +8,24 @@
  */
 
 import * as React from 'react';
+import _ from 'lodash';
 
-import {type DiffProps} from './Diff';
+type FullChangeDiffProps = {
+  +label: string,
+  +newContent: Object,
+  +oldContent: Object,
```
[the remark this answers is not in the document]
anirudhjain75: Used React$Node instead :)

root/static/scripts/edit/components/edit/FullChangeDiff.js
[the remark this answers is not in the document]
mwiencek: I'm not sure I follow. You shouldn't ever need to deeply compare objects for this - can you give a concrete example?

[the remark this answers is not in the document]
anirudhjain75: Case 1 :

> <FullChangeDiff
>           label={addColonText(l('Type'))}
>           newContent={display.type.new ? display.type.new.name : ''}
>           oldContent={display.type.old ? display.type.old.name : ''}
>         />

Case 2 :

> <FullChangeDiff
>             label={addColonText(l('Area'))}
>             newContent={area.new
>               ? <DescriptiveLink entity={area.new} />
>               : ''}
>             oldContent={area.old
>               ? <DescriptiveLink entity={area.old} />
>               : ''}
>           />

Case 1 is from EditAlias where the content type is String, whereas in Case 2 which is from EditPlace the content type is Object where gid had to be compared. I'll currently use conditions using type method and push the changes, please let me know if you have a better idea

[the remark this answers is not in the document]
mwiencek: Just display `null` if `area.new.gid === area.old.gid`. :) To clarify I wasn't sure what you meant by "If I'd have used .gid to compare then the object structure with gid would have to be enforced." We don't care about the rest of the object here. If the gids are equal, the rest of the area properties will be equal too - you can't change the properties of an area in a place edit. If there is a case where we ever need to deeply compare objects, I'd rather that be done outside the component and not by default, because it's a relatively expensive operation.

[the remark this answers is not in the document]
anirudhjain75: Yep made the changes and fixed the test

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,134 @@
...
+        ) : null}
+        {address ? (
+          <Diff
+            label={addColonText(l('Address'))}
+            newText={address.new ? address.new : ''}
+            oldText={address.old ? address.old : ''}
+            split="\s+"
+          />
+        ) : null}
+        {area && area.new?.gid === area.old?.gid ? (
```
reosarevok: Pretty sure you want a diff if they're NOT the same, here :)

anirudhjain75: taken care of it,  thanks :)

root/edit/details/EditPlace.js
```diff
@@ -0,0 +1,134 @@
...
+            oldText={formatDate(endDate.old)}
+            split="-"
+          />
+        ) : null}
+        {ended ? (
+          <FullChangeDiff
+            label={addColonText(l('Ended'))}
+            newContent={yesNo(ended.new)}
+            oldContent={yesNo(ended.old)}
+          />
```
mwiencek: There's an issue here where "ended" diffs aren't currently displayed. Looks like this is because we don't convert those to booleans in `build_display_data`, so something like `{ new: 1, old: '0' }` will cause both sides to be true. (It worked in Perl 'cause `'0'` is falsey in Perl, but not so in JS.) So you'll need something like this in `build_display_data`:

```Perl
if (exists $self->data->{new}{ended}) {
    $data->{ended} = {
        new => boolean_to_json($self->data->{new}{ended}),
        old => boolean_to_json($self->data->{old}{ended}),
    };
}
```

anirudhjain75:
@mwiencek make the changes... What I think is we need more test cases be it selenium tests or perl tests to detect these errors instantly. As I see for most of the changes I make, one of the mentors has to have look at the diff personally instead of having a test suit do it. I'd recommend increasing the amount of tests and also adding variety of possible for every entity. That should help both the contributor as well as the mentors who review the work. :)

mwiencek:
Yes, that's exactly why we discussed adding Selenium tests here before. I'm glad you agree they're needed. :)

PR #1230, opened 2019-10-11
MBS-10419: Beta: Unbreak historical attributes display in edits

mwiencek:
# Fix beta regression MBS-10419

### DIFF
--- a/root/static/scripts/edit/utility/linkPhrase.js
+++ b/root/static/scripts/edit/utility/linkPhrase.js
@@ -141,7 +141,15 @@ function _setAttributeValues<T>(
       const info = linkType.attributes[type.root_id];
       const rootName = linkedEntities.link_attribute_type[type.root_id].name;
 
-      if (info.max === 1) {
+      /*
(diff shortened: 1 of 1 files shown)


root/static/scripts/edit/utility/linkPhrase.js
```diff
@@ -141,7 +141,15 @@ function _setAttributeValues<T>(
...
-      if (info.max === 1) {
+      /*
+       * This may be a historical relationship which uses an attribute
+       * that has since been removed from the link type, but where the
+       * attribute still exists in the link_attribute_type table. In
+       * that case we assume `max` is unbounded just to be safe. (The
+       * only effect this has is passing the values to commaOnlyList
+       * for display.)
+       */
+      if (info && info.max === 1) {
```
mwiencek: Also with https://github.com/metabrainz/musicbrainz-server/pull/1224, an example of where we could do `info?.max === 1` instead without running into refinement issues. :)

PR #1231, opened 2019-10-12
MBS-10421: Update schema_fixup of ReleasePackaging

yvanzo:
# Fix critical regression MBS-10421

After having added release packaging id to MMD schema for MBS-8978, and having updated search indexer/server to follow this new schema, this patch updates MusicBrainz Server to handle the new JSON format returned by MB Solr search server to instantiate ReleasePackaging.

It keeps compatibility with former Lucene-based search server.

Since JSON WS 2 lookup/browse requests use 'packaging-id' instead of 'id' subkey to 'packaging' to preserve backward-compatibility, this patch also anticipates further fix to SOLR-121 search server issue.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -520,10 +520,29 @@ sub schema_fixup
                 name => delete $data->{status}
             )
         }
-        if ($data->{packaging}) {
-            $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
-                name => delete $data->{packaging}
-            )
+
+        my $packaging = delete $data->{packaging};
+        my $packaging_id = delete $data->{'packaging-id'};
+
+        if ($packaging) {
+            if (ref($packaging) eq 'HASH') {
+                # MB Solr search server v3.1
+                $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
+                    name => $packaging->{name},
+                    gid => $packaging->{id}
+                )
+            } elsif ($packaging_id) {
+                # MB Solr search server v3.2? (SOLR-121)
+                $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
+                    name => $packaging,
+                    gid => $packaging_id
+                )
+            } else {
+                # MB Lucene search server
+                $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
+                    name => $packaging
+                )
+            }
         }
     }
     if ($type eq 'release-group') {


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -520,10 +520,18 @@ sub schema_fixup
...
             )
         }
+
+        my $packaging = delete $data->{packaging};
+
         if ($data->{packaging}) {
-            $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
-                name => delete $data->{packaging}
-            )
+            $data->{packaging} = ($packaging->{name} && $packaging->{id})
```
mwiencek: If `$packaging` can still be a string you may want to check `ref($packaging) eq 'HASH'` before these key accesses, otherwise Perl might break with "can't use string as a hash ref."

yvanzo: Thank you, fixed in https://github.com/metabrainz/musicbrainz-server/pull/1231/commits/a42067b0fe8c616bd3164d9d04e302050b4aaf8a#diff-273b5368fc014db576aec607d9a783aaR527.

yvanzo:
Updated with @mwiencek’s suggestion to anticipate a future fix to SOLR-121 with 'packaging-id' key.

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -520,10 +520,29 @@ sub schema_fixup
                 name => delete $data->{status}
             )
         }
+
+        my $packaging = delete $data->{packaging};
+        my $packaging_id = delete $data->{'packaging-id'};
+
         if ($data->{packaging}) {
```
mwiencek:
```suggestion
        if ($packaging) {
```
since `$data->{packaging}` is deleted above

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -520,10 +520,29 @@ sub schema_fixup
...
-            $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
-                name => delete $data->{packaging}
-            )
+            if (ref($packaging) eq 'HASH') {
+                # MB Solr search server v3.1
+                $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
+                    name => $packaging->{name},
+                    gid => $packaging->{id}
+                )
+            } elsif ($data->{packaging_id}) {
```
mwiencek:
```suggestion
            } elsif ($packaging_id) {
```

lib/MusicBrainz/Server/Data/Search.pm
yvanzo: Fixed at https://github.com/metabrainz/musicbrainz-server/pull/1231/commits/39e40c1951b6b0a5c56c7c9be27258e937d95c4f#diff-273b5368fc014db576aec607d9a783aaR527

lib/MusicBrainz/Server/Data/Search.pm
yvanzo: Fixed at https://github.com/metabrainz/musicbrainz-server/pull/1231/commits/39e40c1951b6b0a5c56c7c9be27258e937d95c4f#diff-273b5368fc014db576aec607d9a783aaR534

PR #1233, opened 2019-10-14
MBS-9468: Missing check for remove rel edits in search

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9468

The query currently checks for `(data#>>'{relationship,link_type,id}')::int`.
In the removal examples I checked, the type is loaded and we need `(data#>>'{relationship,link,type,id}')::int` (comma instead of underscore) instead.
Since I'm not 100% sure whether the previous check actually matches something too, not removing it at the moment (if you're confident it should be removed, let me know).


mwiencek:
I ran `select * from edit e join edit_data ed on e.id = ed.edit where e.type in (90, 91, 92, 99) and data#>>'{relationship,link_type,id}' is not null limit 1;` on pink and it found nothing, so it seems that the old check doesn't actually match anything and can be removed.

reosarevok:
Sad yet unsurprising :p Removed it then.

PR #1234, opened 2019-10-14
MBS-10423: Handle missing ID of release packaging

yvanzo:
# Fix regression MBS-10423

Search server may sometimes return release packaging without any ID associated with.
This patch enables MusicBrainz Server to handle this corner case.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -529,7 +529,7 @@ sub schema_fixup
                 # MB Solr search server v3.1
                 $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
                     name => $packaging->{name},
-                    gid => $packaging->{id}
+                    defined $packaging->{id} ? (gid => $packaging->{id}) : ()
                 )
             } elsif ($packaging_id) {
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -527,10 +527,14 @@ sub schema_fixup
         if ($packaging) {
             if (ref($packaging) eq 'HASH') {
                 # MB Solr search server v3.1
-                $data->{packaging} = MusicBrainz::Server::Entity::ReleasePackaging->new(
+                $data->{packaging} = defined $packaging->{id}
+                ? MusicBrainz::Server::Entity::ReleasePackaging->new(
                     name => $packaging->{name},
                     gid => $packaging->{id}
```
mwiencek: This works, but there's a simpler way to make a hash entry optional in Perl without duplicating the definition:
```suggestion
                     defined $packaging->{id} ? (gid => $packaging->{id}) : (),
```

yvanzo: applied in https://github.com/metabrainz/musicbrainz-server/pull/1234/commits/c8577d080a9c4d27e19d41ab939017349ec11711

PR #1235, opened 2019-10-14
MBS-10408: Stop AC edits_pending overwriting artist one

reosarevok:
By having both  artist.edits_pending and ac.edits_pending, and processing them not very carefully later, this query was causing ac.edits_pending to be used for both the AC and the artist itself, giving the wrong result everywhere ArtistCredit::get_by_ids is used.


lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -25,7 +25,7 @@ sub get_by_ids
     my ($self, @ids) = @_;
     my $artist_columns = $self->c->model('Artist')->_columns;
     my $query = "SELECT artist, artist_credit_name.name AS ac_name, join_phrase, artist_credit, " .
-                $artist_columns . ", ac.edits_pending " .
+                $artist_columns . ", ac.edits_pending as ac_edits_pending " .
```
mwiencek: nitpick
```suggestion
                $artist_columns . ", ac.edits_pending AS ac_edits_pending " .
```

PR #1236, opened 2019-10-14
MBS-10425: Allow rel credits for instruments

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10425


reosarevok:
Whoopsio. Done.

yvanzo:
Where is that change visible?

reosarevok:
Any instrument editing page, for example. instrument/b3e097c7-c167-4436-984f-5742a4542bb2/edit should exist even in the sample DB. You can just add a rel to another instrument and see if you can set credits.

reosarevok:
Not only - I think you should also see it if you try to link an artist to an instrument, for example (with "invented" or whatnot) :) But yeah, I can see the confusion, will change wording.

yvanzo:
> if you try to link an artist to an instrument

Yes but I’m not sure this should be allowed in the first place.

reosarevok:
Well, we're going to need it for STYLE-480 :)

PR #1237, opened 2019-10-15
Eslint fixes for /account/applications

reosarevok:
Removing unused imports (and a space between imports)

### DIFF
--- a/root/account/applications/Remove.js
+++ b/root/account/applications/Remove.js
@@ -11,9 +11,11 @@ import * as React from 'react';
 
 import ConfirmLayout from '../../components/ConfirmLayout';
 
-const RemoveApplication = () => ConfirmLayout({
-  question: l('Are you sure you want to remove this application?'),
-  title: l('Remove Application'),
-});
+const RemoveApplication = () => (
+  <ConfirmLayout
+    question={l('Are you sure you want to remove this application?')}
+    title={l('Remove Application')}
+  />
+);
 
 export default RemoveApplication;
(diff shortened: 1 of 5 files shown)


root/account/applications/Remove.js
```diff
@@ -7,8 +7,6 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-import * as React from 'react';
```
mwiencek: This is unused because `ConfirmLayout` is called as a function below, like `ConfirmLayout({..})`, rather than as a React element like `<ConfirmLayout {...} />` (which would get transpiled into `React.createElement(ConfirmLayout, {...})`, so the import would be needed). Perhaps we should change it to use the React element syntax for consistency instead, but as long as we don't, well.

reosarevok: Is there an actual benefit to calling it that way? (I guess consistency is already a benefit - in which case I could just make the change)

mwiencek: I'm by no means a React expert, but I don't think it particularly matters unless we we're using performance tweaks like [React.PureComponent](https://reactjs.org/docs/react-api.html#reactpurecomponent)/`shouldComponentUpdate` that involve the component lifecycle, or features like [error boundaries](https://reactjs.org/docs/error-boundaries.html).

Using React elements changes the "component tree," which can affect [reconciliation](https://reactjs.org/docs/reconciliation.html) too, but these components in particular are for static server-rendered pages which aren't being updated.

TL;DR: It doesn't matter here, and is probably slightly faster to call it as a function even, but using a React element would be more consistent.

reosarevok: I'm changing them to be called as an element instead. Yay consistency!

yvanzo: And for future use of error boundaries!

PR #1238, opened 2019-10-15
MBS-10428: Show artist aliases for recording ACs

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10428

While at it, I also did https://tickets.metabrainz.org/browse/MBS-10429 which makes the result more consistent with the XML WS without (AFAICT) dropping any info.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
@@ -264,6 +264,7 @@ sub serialize_relationships {
          $entity->has_loaded_relationships);
 
     local $hide_tags_and_genres = 1;
+    local $hide_aliases = 1;
 
     my @relationships =
         map { serialize_entity($_, $inc, $stash) }
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
@@ -152,6 +152,18 @@ sub serialize_aliases {
 
     return unless defined $inc && $inc->aliases;
 
+    # We don't show aliases again for recording artists if they're on the release or track AC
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
```diff
@@ -152,6 +152,16 @@ sub serialize_aliases {
...
 
+    # We don't show aliases again for recording artists if they're on the release or track AC
+    if ($entity->isa('MusicBrainz::Server::Entity::Artist')) {
+        if (my $release_ac = $stash->{release_artist_credit}) {
+            if (my $track_ac = $stash->{track_artist_credit}) {
+                return if (grep { $_->artist_id == $entity->id } $release_ac->all_names);
+                return if (grep { $_->artist_id == $entity->id } $track_ac->all_names);
+            }
+        }
+    }
```
mwiencek: Can we copy the same code from https://github.com/metabrainz/musicbrainz-server/blob/6b9a6446d6ed6593ffe23e39c2df29154680771c/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm#L284-L291 here? Or factor it out into a `is_artist_used_by_track_or_release` or something?

(Not actually sure there's a case that matters for aliases where `release_artist_credit` is defined and not `track_artist_credit`. But I might be forgetting something. It'd be nice to have the code shared in any case.)

reosarevok: I specifically based this on that but changed how it works, so that it must ensure a  track_artist_credit exists. Otherwise, unless I'm getting stuff wrong, this would also apply to track artist credits and actually break stuff that is there now.

mwiencek: Got it, thanks. Can you add a small line to the comment above saying you check for both so that it doesn't apply to tracks to maintain compatibility?

PR #1240, opened 2019-10-17
MBS-10432 / MBS-10433 / MBS-10437: Block adding removal edits for areas that shouldn't be removed

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10432
https://tickets.metabrainz.org/browse/MBS-10433
https://tickets.metabrainz.org/browse/MBS-10437

I chose to add a new check for MBS-10433 because I wanted a different error text for that case.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Area.pm
+++ b/lib/MusicBrainz/Server/Data/Area.pm
@@ -24,6 +24,7 @@ use MusicBrainz::Server::Data::Utils qw(
     placeholders
     object_to_ids
 );
+use MusicBrainz::Server::Data::Utils::Cleanup qw( used_in_relationship );
 
 extends 'MusicBrainz::Server::Data::CoreEntity';
 with 'MusicBrainz::Server::Data::Role::Annotation' => { type => 'area' };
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/Area.pm
```diff
@@ -134,6 +134,15 @@ sub update
     return 1;
 }
 
+sub is_release_country_area {
```
mwiencek: I would call this in `can_delete` below too, since `can_delete` is also called by `Edit::Generic::Delete` as a final check, which might matter if someone manages to actually enter such an edit by way of another bug.

reosarevok: Probably sensible. Done!

reosarevok:
I'm happy to make that change, it just didn't seem like a bug, just a decision, unlike the others :) IMO that'd be a reasonably sensible change.

reosarevok:
Made a change to only allow removing empty areas.

reosarevok:
bitmap pointed that there's no particular reason why we can't have this inside can_delete instead, since nothing else uses is_empty anyway. So I moved the check inside that (and added a bit of info to the commit message as requested!)

PR #1241, opened 2019-10-17
MBS-10431: Remove 'yourmusic' from 7digital URLs

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10431

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -312,6 +312,17 @@ function disallow(url, id) {
  *             for an auto-selected relationship type.
  */
 const CLEANUPS = {
+  '7digital': {
+    match: [new RegExp('^(https?://)?([^/]+\\.)?(7digital\\.com|zdigital\\.com\\.au)', 'i')],
+    type: LINK_TYPES.downloadpurchase,
+    clean: function (url) {
+      // Standardise to https
+      url = url.replace(/^https?:\/\/(.*)$/, 'https://$1');
+      // Remove yourmusic + id from link for own purchases
+      url = url.replace(/^https:\/\/([^/]+\.)?(7digital\.com|zdigital\.com\.au)\/yourmusic\/(.*)\/[\d]+\/?$/, 'https://$1$2/$3');
+      return url;
+    },
+  },
   '45cat': {
     match: [new RegExp('^(https?://)?(www\\.)?45cat\\.com/', 'i')],
     type: LINK_TYPES.otherdatabases,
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -137,6 +137,12 @@ const testData = [
              input_entity_type: 'release',
     expected_relationship_type: 'downloadpurchase',
   },
+  {
+                     input_url: 'https://us.7digital.com/yourmusic/artist/falco/release/vienna-greatest-hits-311837/311837',
+             input_entity_type: 'release',
+    expected_relationship_type: 'downloadpurchase',
+            expected_clean_url: 'https://us.7digital.com/artist/falco/release/vienna-greatest-hits-311837/311837',
```
mwiencek: Hmm, just removing `/yourmusic` doesn't fix the link though. The edit in the ticket also had to remove the `/311837` at the end (otherwise it's a 404). Not sure if those are a standard part of the "yourmusic" URLs or what.

reosarevok: Oh, whoops. I'll ask.

reosarevok: Updated the regex for this.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -312,6 +312,14 @@ function disallow(url, id) {
  *             for an auto-selected relationship type.
  */
 const CLEANUPS = {
+  '7digital': {
+    match: [new RegExp('^(https?://)?([^/]+\\.)?(7digital\\.com|zdigital\\.com\\.au)', 'i')],
+    type: LINK_TYPES.downloadpurchase,
+    clean: function (url) {
+      url = url.replace(/^https?:\/\/([^/]+\.)?7digital\.com\/yourmusic\/(.*)\/[^/]+\/?$/, 'https://$17digital.com/$2');
```
yvanzo: That would be nice to clean legitimate URLs up too. An additional test would be welcome then.

yvanzo:
```suggestion
      url = url.replace(/^https?:\/\/([^/]+\.)?7digital\.com\/(?:yourmusic\/)?(.*)\/[^/]+\/?$/, 'https://$17digital.com/$2');
```

reosarevok: This would break because of only yourmusic urls having the bit at the end that needs removing, but it might make sense to just make it two parts. Can do.

reosarevok:
Changed this to have two replacements - one generic cleanup that changes to https, and then one that removes the yourmusic specific bits.

root/static/scripts/edit/URLCleanup.js
yvanzo: Thanks, that should prevent any unmindful coder to make the same mistake as I just did!

PR #1242, opened 2019-10-18
MBS-10424: Collapse release events

mwiencek:
# MBS-10424: Collapse release event list when there are too many

This truncates long release event lists in the sidebar and various release tables to 3 entries. By default, I made it show the first two and then the last, with a link to show more between them.

The "Date" and "Country" columns in the various release tables have been merged into a single "Country/Date" column for several reasons:

 1. This lets us reuse the same `ReleaseEvents` component everywhere.

 2. Managing the rowSpans for every row (depending on whether the events were expanded) was very ugly and complicated.

 3. The release event pairs and release label pairs don't really work as tabular data, since a table implies a 1:1 association between those pairs in the same row. In reality, the release events and release labels are separate lists that have no association between each other.

 4. It worked well (i.e. the data was structured/labeled clearly to me) when I tested the tables in VoiceOver.

Old TT macros and React components that are no longer used (`release_dates`, `release_countries`, etc.) have been removed.


root/cdtoc/attach_artist_releases.tt
```diff
@@ -13,8 +13,7 @@
           <thead>
             <tr>
               <th colspan="2">[% l('Release') %]</th>
-              <th>[% l('Date') %]</th>
-              <th>[% l('Country') %]</th>
+              <th>[% l('Date') _ ' / ' _ l('Country') %]</th>
```
mwiencek: I wondered if I should add a single "Date / Country" string instead of reusing the existing ones and concatenating them with "/", but the sortable headers (on release collections) kinda needed them to be separate anyway, since they're separate links. Not sure how friendly joining them with a slash is, though, i.e. if we need to localize that somehow.

mwiencek:
Showing the first two + last (sorted by date, then country) might be weird, since it's unbalanced that way (maybe showing 4 initially, the first two plus last two, would be better). Including some at the end of the list seems nice though so you can see the maximum date range without expanding the list, I guess.

root/cdtoc/attach_artist_releases.tt
reosarevok: We might want to localize the slash (possibly with a context comment of some sort) but I'd expect these to be separate, yes. I might even just keep it as two headers with a colspan 2 for the actual content cells, if that'd work visually.

reosarevok:
Two comments: 

1. There's no way to collapse the whole thing back if you show them all. Maybe that'd be nice? (but it might also not matter).
2. This changes the display on the sidebar from country first, date later to date first, country later. I'm not sure if that's better - I feel that the structure "date flag country" all in one block is a bit odd to read, and I wonder if "flag country date" wouldn't be simpler. But I guess that the order in the other, tabulated views was indeed date / country, so whichever change we make will "break" one of the two.

root/cdtoc/attach_artist_releases.tt
mwiencek: I did try keeping them as separate columns at first, but it was somewhat tricky to get the dates/countries to stay lined up with the headers in all cases. I also wasn't sure how usable it was in VoiceOver - it seemed weirder when I tried it, but I could check again.

reosarevok:
Also: it's not clear at all to me currently that this is is showing the first and last dates. A much clearer version would be, I feel, to have the two first, some collapsible "show 75 more" text in the center, then the last.

reosarevok:
And one last thing is that by changing from country abbreviations to country names, it makes the tables probably more readable, but certainly less consistent-looking (see below, where in some places the country name wraps, in some places the date, making it look kinda off).
Screenshot from 2019-10-21 19-44-06

mwiencek:
Made some changes.

1. Added a way to collapse things again.
2. Made it display abbreviated country codes in both the table and sidebar.
3. Switched the order from Date/Country to Country/Date in both places.
4. Localized the slash in Country/Date like `lp('/', 'inclusive or')`.
5. Displayed a `-` for missing dates or countries. And added a "missing data" context to that string.

reosarevok:
This looks better, although I do wonder whether it wouldn't be better to just leave the sidebar list as it was (with links to countries and whatnot) and just collapse that one too in the same way.

mwiencek:
@reosarevok I changed the release sidebar display back to how it was. Lemme know how it looks now

root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -0,0 +1,156 @@
...
+    ? events.length > COLLAPSE_THRESHOLD
+    : false;
+
+  return (
+    (events && events.length) ? (
+      <>
+        {(tooManyEvents && !expanded) ? (
+          <>
+            <ul {...containerProps}>
+              {events.slice(0, COLLAPSE_THRESHOLD - 1).map(
```
yvanzo: The `li.show-all` element has larger height than `li.release-event`. Thus, if there are `COLLAPSE_THRESHOLD + 1` events, then the release event list takes more space collapsed than before this commit. One way out would be to ensure hiding at least 2 release events as follows:

```suggestion
              {events.slice(0, COLLAPSE_THRESHOLD - 2).map(
```

root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -0,0 +1,156 @@
...
+
+import CountryAbbr from '../../../../components/CountryAbbr';
+import hydrate from '../../../../utility/hydrate';
+import {bracketedText} from '../utility/bracketed';
+import formatDate from '../utility/formatDate';
+import isDateEmpty from '../utility/isDateEmpty';
+
+import EntityLink from './EntityLink';
+
+const COLLAPSE_THRESHOLD = 3;
```
yvanzo: This follows the other comment on hiding at least two elements. If you still want to show at least the 2 first elements:
```suggestion
const COLLAPSE_THRESHOLD = 4;
```

root/components/list/ReleaseList.js
```diff
@@ -107,24 +107,21 @@ const ReleaseList = ({
...
+                {lp('/', 'inclusive or')}
+                <SortableTableHeader
+                  label={l('Date')}
+                  name="date"
+                  order={order}
+                />
+              </>
             )
-            : l('Country')}
+            : l('Country') + lp('/', 'inclusive or') + l('Date')}
```
yvanzo: I’m not sure that `inclusive or` is self-explanatory to anyone who isn’t programming.
I would suggest `table-header-column-separator` or something instead...

mwiencek: I tried to pick something more general that could be used in other places we use a slash (though the only other place I can think of is language/script pairs). `lp('/', 'and')` seems a lot better than "inclusive or" now that I think about it. Do you think that'd be useful?

root/static/scripts/common/components/ReleaseEvents.js
mwiencek: I made those changes but also reduced the padding around those links a bit (from 1em to 0.75em), since it did seem a bit much.

root/components/list/ReleaseList.js
yvanzo: I feel like it still is missing _context_ but I don’t have any better suggestion.

root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -0,0 +1,156 @@
...
+              <li className="show-all" key="show-all">
+                <a
+                  href="#"
+                  onClick={expand}
+                  title={l('Show all release events')}
+                >
+                  {texp.l('show {n} more', {
+                    n: events.length - COLLAPSE_THRESHOLD,
+                  })}
+                </a>
```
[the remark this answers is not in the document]
yvanzo: Parenthesis should be around the `a` tag, see the _Release group rating_ section for example.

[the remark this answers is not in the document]
mwiencek: I know the other place is like that, but it doesn't really make sense to me that the `(` and `)` shouldn't be clickable in this case.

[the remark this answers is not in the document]
mwiencek: Here the link is being used as a button, basically, whereas there it's just a link inside parentheses.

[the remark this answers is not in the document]
yvanzo: It’s probably because it makes a small graphical glitch with the underlining style otherwise.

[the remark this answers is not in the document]
mwiencek: I'm not sure what you mean. :) Can you post a screenshot?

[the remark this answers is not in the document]
yvanzo: If it is just about styling as a button, maybe CSS should be used instead of parenthesis?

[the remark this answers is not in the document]
yvanzo: Better now, since parenthesis are no more part of the localized string at least.

PR #1243, opened 2019-10-19
MBS-10403 / MBS-10440: Removal page improvements

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10403
https://tickets.metabrainz.org/browse/MBS-10440

This adds some info about merging / fixing to our (manual) entity removal pages.
An example of the recording removal page (everything except the last paragraph is also on all other removal pages):
Screenshot from 2019-10-19 12-46-31

It also removes series/delete (see commit message for reasoning) and standardises page titles.


reosarevok:
Also improved Merge Rather Than Delete a bit. It's still not amazing, but.

root/components/forms.tt
```diff
@@ -358,3 +358,20 @@
...
   <div id="external-links-editor-container"></div>
 [%- END -%]
+
+[%- MACRO removal_help(extra_help) BLOCK -%]
+  <div id="removal-help">
+    <p>
+      [% l('Are you sure you wish to remove {entity} from MusicBrainz?', { entity => link_entity(entity) }) %]
+    </p>
+    <p>
+      [% l('If it is a duplicate, {doc_merge|you should probably merge it instead}.', { doc_merge => doc_link('Merge_Rather_Than_Delete') }) %]
```
mwiencek:
```suggestion
      [% l('If it’s a duplicate, {doc_merge|you should probably merge it instead}.', { doc_merge => doc_link('Merge_Rather_Than_Delete') }) %]
```
sounds better with the contraction, like below

root/components/forms.tt
```diff
@@ -358,3 +358,20 @@
...
+      [% l('If it just has some small errors, it’s usually better to just fix those.') %]
```
mwiencek: Should this just be merged with the previous string? (Or if they're meant to be on separate lines, maybe in a separate `<p>` tag.)

root/recording/delete.tt
```diff
@@ -1,9 +1,9 @@
-[%- WRAPPER 'recording/layout.tt' title=l('Remove') full_width=1 -%]
+[%- WRAPPER 'recording/layout.tt' title=l('Remove Recording') full_width=1 -%]
     <h2>[% l('Remove Recording') %]</h2>
     [%- IF can_delete -%]
-        <p>
-            [%- l('Please make sure you\'re not removing a legitimate {doc_standalone|standalone recording}. If this recording is a duplicate of another, please {doc_merge|merge them instead}.', {doc_standalone => doc_link('Standalone_Recording'), doc_merge => doc_link('Merge_Rather_Than_Delete')}) -%]
-        </p>
+        [% removal_help(
+            l('Please make sure you\'re not removing a legitimate {doc_standalone|standalone recording}.', {doc_standalone => doc_link('Standalone_Recording')})
```
mwiencek:
```suggestion
            l('Please make sure you’re not removing a legitimate {doc_standalone|standalone recording}.', {doc_standalone => doc_link('Standalone_Recording')})
```
since we're changing the string anyway, why not

root/components/forms.tt
mwiencek: (Unless we're using/plan to use the strings separately. Otherwise maybe combining them will be better for context.)

reosarevok: I separated them because I thought they looked more clear like that, but you might be right :)

PR #1245, opened 2019-10-22
Get po/Makefile prereqs with git-ls-files

mwiencek:
Update: also fixes MBS-10446 and MBS-10447 (see discussion below).

I happen to have a ton of untracked files in my musicbrainz-server checkout, many of them .js files from various branches I'm working on. When I run `fab pot`, I don't want to have to worry about our gettext extraction picking up strings from these files (which does always happen), so I've modified the commands used to find source files to use `git ls-files` instead of `find`, so that only cached git files are found.

I've also modified some of the make steps to pass these very-long file lists to their respective commands using xargs.


po/Makefile
```diff
@@ -40,18 +51,18 @@ pot: mb_server.pot statistics.pot languages.pot languages_notrim.pot scripts.pot
 mb_server.pot: $(POTFILES) $(JSFILES) $(TT) $(wildcard extract_pot_templates)
 	@echo "Rebuilding the mb_server.pot file";
 	@echo "- Update templates .pot";
-	@find ../root -path ../root/statistics -prune -o -type f -name \*.tt -print | sort | ./extract_pot_templates | msguniq -s > mb_server.pot;
+	@$(GIT_LS_TT) | ./extract_pot_templates | msguniq -s > mb_server.pot;
 	@echo "- Update code .pot"
-	@xgettext --from-code utf-8 --keyword=__ --keyword=l --keyword=lp:1,2c --keyword=N_lp:1,2c --keyword=N_l --keyword=ln:1,2 --keyword=N_ln:1,2 --keyword=__x --keyword=__nx:1,2 --keyword=__n:1,2 -Lperl -o mb_server.pot --add-comments=translators -j $(POTFILES)
-	@../script/xgettext.js $(JSFILES) > javascript.pot
```
mwiencek: Note that using xargs here also fixes potential issues with spaces or parentheses in filenames, which is a minefield in GNU make.

We don't have any .js files that meet those criteria, but we actually have some .tt files with parentheses in their names (edit templates with "(historic)" appended). This is likely why the .tt command above duplicated the `find` and piped it into `extract_pot_templates` instead of passing `$(TT)` as arguments (which the `extract_pot_templates` script supports, but would have completely broken make).

yvanzo: To be really safe, you can set NUL separator using `git ls-files -z` and `xargs -0`.

yvanzo: To be really safe, you can set NUL separator using `git ls-files -z` and `xargs -0`.

mwiencek: That's true, though it'll break the make prerequisite lists which have to be delimited with spaces. (`git ls-files` without `-z` at least still quotes file names with odd characters.)

mwiencek: Though apparently git doesn't consider a space as unusual enough to quote - I'll look into this further.

mwiencek: Yeah, I was wrong that spaces were working with `xargs` alone - it only fixed things for the parentheses (which is a shell issue).

I still couldn't use `-z` for the reason I mentioned, but backslash-escaping the spaces with sed seems to work for both the prerequisites lists and within xargs. To test, I added a .js file like root/"this is (a test)".js with `l('test');` as the contents, and did `rm po/mb_server.pot && make -C po mb_server.pot` and saw that it extracted the string fine. But note that I'm on macOS, so using the BSD xargs and sed, not GNU - please test it with those if you can. :)

Of course we shouldn't be adding any spaces to source filenames to begin with, but. :)

reosarevok: I wouldn't mind changing the existing files to not have parentheses in their titles, FWIW. Even if this works with them.

mwiencek: I'd def. agree to ban spaces with some kind of git hook, and maybe parentheses too. Though we fix the parentheses issue for free here with xargs, so I'm happy enough to leave this PR as-is. I can remove the spaces escaping if we don't want it.

mwiencek: Apparently we only had one actual file with parentheses, and it wasn't even being referenced anywhere, causing MBS-10446 (kinda fun what pointless inquiries will uncover sometimes). So I pushed a commit to fix that.

reosarevok: And some of those edits were broken, causing MBS-10447 - so yet another commit for that.

mwiencek: Thank you!

PR #1246, opened 2019-10-22
MBS-10439: Convert user collections list to React

yvanzo:
# MBS-10439: Convert user collections list to React

It introduces `react-table` v7 beta dependency as a PoC for MBS-10365.

Only difference is that cells are generated with an empty `class=""` attribute instead of not having any class attribute as before.


root/user/UserCollections.js
```diff
@@ -0,0 +1,265 @@
...
+    accessor: 'collaborators',
+    id: 'collaborators',
+  };
+  const subscriptionColumn = {
+    Cell: ({cell: {value}}) => yesNo(value),
+    Header: N_l('Subscribed'),
+    accessor: 'subscribed',
+  };
+  const privacyColumn = {
+    Cell: ({row: {original}}) => collaborativePrivacy(original, $c.user),
```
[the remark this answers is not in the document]
yvanzo: I messed up the condition (fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/551fc4449c0d5b1c9500e9ae562fd01d018927f5) and probably the function name too.

root/user/UserCollections.js
```diff
@@ -0,0 +1,265 @@
...
+    accessor: 'id',
+    id: 'actions',
+  };
+  const columns = React.useMemo(
+    () => [
+      nameColumn,
+      typeColumn,
+      sizeColumn,
+      collaboratorsColumn,
+      ...viewingOwnProfile && !isCollaborative ? [subscriptionColumn] : [],
```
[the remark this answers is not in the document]
yvanzo: No particular reason, just abusive copy/paste of condition, fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/0431806e36ac93ebc9aa373c65d513e1ad442be7.

mwiencek:
> More generally: I feel this way of building tables is much less clear than just building the table in JSX like we have done so far, so can you specify what the benefits are? I understand the main thing this seems to allow for is sorting, but there's very few places where we do sorting currently and we wouldn't still need to do it server-wise for pagination.

Good questions. I hoped it would make building tables easier, but it looks more complicated. Maybe it's worth it for the features it provides. Just not sure which ones we'll use, since our pagination requires a full page reload at the moment (which we already have a component written for), and sorting tables is already trivial in vanilla React (just store the row data in state and call `.sort()` in a header click event).

> I read something about this doing pagination too, but I'm expecting clients to not be very happy if we throw 250 pages of recordings at them at once so the pagination we do should still probably be server side. Or is this something I'm misunderstanding? :)

I read through the docs and it does apparently allow for server-side pagination. We'd just need to create a ton of server endpoints that return results as JSON and figure out the right combination of options to hook it together here. It's also something our existing pagination component could be modified to support, though.

Anyway, I guess I would have liked to see this PoC done on a page that actually uses any of the `react-table` features, since this one doesn't. :) Some of it sounds like it might be useful on the statistics pages. But do we want to build all of our tables this way?

BTW, apparently v7 has a memory leak (see their issues tracker; I won't link it to avoid cross-link spam).

root/user/UserCollections.js
```diff
@@ -0,0 +1,265 @@
...
+  label: N_l('Label Collections'),
+  place: N_l('Place Collections'),
+  recording: N_l('Recording Collections'),
+  release: N_l('Release Collections'),
+  release_group: N_l('Release Group Collections'),
+  series: N_lp('Series Collections', 'plural'),
+  work: N_l('Work Collections'),
+};
+
+const CollectionsTable = (({columns, data}) => {
```
mwiencek: Just an observation: this component looks generic enough that it could be used for pretty much any table, not just collections. Maybe that's the (eventual) idea. :)

root/user/UserCollections.js
```diff
@@ -0,0 +1,265 @@
...
+  user,
+}) => {
+  const viewingOwnProfile = !!$c.user && $c.user.id === user.id;
+  const nameColumn = {
+    Cell: ({row: {original}}) => (
+      <EntityLink entity={original} />
+    ),
+    Header: N_l('Collection'),
+    accessor: 'name',
+  };
```
mwiencek: One use I can see of building tables this way is that we can move these column descriptions outside of the component, and export/import them around so we can compose them together to customize which columns are shown in different tables. (Though maybe the `hideFoo` props we have already work just as well.)

yvanzo:
Thanks for your reviews. Keep in mind this is just the conversion of user collections list to React, not the refactoring task. I picked that conversion task because:
1. It is still useful per se as it contributes to the React conversion;
2. It allows to code a table from a clean React sheet;
3. It is a proof-of-concept that it is possible to render our current `table` tags using `react-table`, which is one minimal requirement!

That is why I only mentioned empty `class` attributes as current regression.  It is not breaking anything though, and it will probably be fixed at some point since it is still in the beta stage.

> > More generally: I feel this way of building tables is much less clear than just building the table in JSX like we have done so far, so can you specify what the benefits are?
>
> Good questions. I hoped it would make building tables easier, but it looks more complicated.

It is more complicated for a single instance, that’s for sure.  The main benefit is not about sorting or such features, it is just the way to build the table, even though it then makes some features available for free. (BTW, another feature you didn’t mention: selective display of columns.)

Spoiling the refactoring task:  I started to use react-table because I was basically reinventing the wheel with the same approach: _column-definition-driven_ table component, read more in the [blog post](https://medium.com/@tannerlinsley/why-i-wrote-react-table-and-the-problems-it-has-solved-for-nozzle-others-445c4e93d4a8). For collections list, it is not very useful as there is no other collections list for now at least. It will be very useful to entity list as column instances can be defined once, either as a const or generated with a function, in utility and used for every entity table, so that coding a new table would only require to define columns as in https://github.com/metabrainz/musicbrainz-server/blob/0431806e36ac93ebc9aa373c65d513e1ad442be7/root/user/UserCollections.js#L181-L192
For example: artist’s recordings, recording search results, reported recordings, collected recordings, … all of these recording lists have almost the same columns with a few additions and exceptions.

> > I read something about this doing pagination too, but I'm expecting clients to not be very happy if we throw 250 pages of recordings at them at once so the pagination we do should still probably be server side. Or is this something I'm misunderstanding? :)
>
> I read through the docs and it does apparently allow for server-side pagination. […] It's also something our existing pagination component could be modified to support, though.

Pagination can be made either client-side or server-side. We don’t have to throw everything at once, even with client-side pagination. `react-table` has a fair support of pagination as it doesn’t support scrolling (which doesn’t seem to be something wanted anyway).

> I would have liked to see this PoC done on a page that actually uses any of the react-table features

For a PoC about `react-table` advanced features, see its fully fledged example at https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/kitchen-sink
It is nice that `react-table` comes with some features too, but its main strength is to be lightweight and headless thus versatile.  I’m confident it will keep to be like that, and that advanced features will be fixed at some point.
Again this is just a 1:1 conversion to React, but I agree sorting, filtering, editing, and even merged tables would be nice to have for this page later on.

> But do we want to build all of our tables this way?

That’s the spirit!

> BTW, apparently v7 has a memory leak

It is about features that are not used here. It is pretty concerning for further use of client-side sorting though, but again still in the beta stage.

----

Thank you for reading that far, I hope these explanations clarified the motivations of refactoring with `react-table`.  Let me know if it doesn’t make sense to you.  I will keep improving this PR and making the code a bit more modular to anticipate further refactoring of other tables.

root/user/UserCollections.js
yvanzo: Yes it is, see https://github.com/metabrainz/musicbrainz-server/pull/1246#issuecomment-545968391.

root/user/UserCollections.js
yvanzo: Exactly, answered in https://github.com/metabrainz/musicbrainz-server/pull/1246#issuecomment-545968391.

mwiencek:
Thanks for the clarification! I can see the potential now for how this might actually simplify some of our tables rather than make them more complicated, if we were able to define them in terms of (shared) column definitions. It was just hard to picture that at first from this isolated example.

I'm definitely fine with iterating on using `react-table` further, but would like to see the `class=""` and memory leak issues fixed before we use it in production at least (maybe we should open an issue for the class one, if there isn't already one, and if it isn't due to some misconfiguration on our part).

mwiencek:
In #1242 I'm combining the "Date" and "Country" columns while still allowing them to be sortable (by clicking on either word). I wonder if something like that is still possible with `react-table`.

yvanzo:
> In #1242 I'm combining the "Date" and "Country" columns while still allowing them to be sortable (by clicking on either word). I wonder if something like that is still possible with `react-table`.

It requires hidden columns to sort by date and by country, and displayed column to combine these. The header of this latter column can probably be customized to allow sorting by either date or country.

yvanzo:
Fixed bugs, got my PR for the `class=""` issue merged, the other issue has a PR under review but doesn’t affect collections list anyway.

Two more points to look at:
* flow typing: currently using `any`, should probably be replaced with `flow-typed/react-table`?
* editor data: currently server-side rendered only, should be sanitized for client-side sorting later on?

root/components/Table.js
```diff
@@ -0,0 +1,67 @@
...
+    getTableBodyProps,
+    headerGroups,
+    rows,
+    prepareRow,
+  } = useTable({
+    columns,
+    data,
+  });
+
+  /* eslint-disable react/jsx-key */
```
reosarevok: Why don't we want keys?

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+  subscriptionColumn,
+  defineActionsColumn,
+} from '../utility/tableColumns';
+
+type Props = {|
+  +$c: {user: EditorT} & CatalystContextT,
+  +collaborativeCollections: CollectionListT,
+  +ownCollections: CollectionListT,
+  +user: EditorT,
+|};
```
reosarevok: Types are exact by default now. Also IIRC the {user: EditorT} & CatalystContextT format was a problem with the latest Flow version, but @mwiencek will remember better.

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+) {
+  return !!currentUser && !!collection.editor &&
+    collection.editor.id === currentUser.id ? (
+      formatPrivacy(collection.public) + ' ' + l('(your collection)')
+    ) : (
+      formatPrivacy(collection.public)
+    );
+}
+
+const CollectionsEntitySection = ({
```
reosarevok: Why not just CollectionsTable or, if you don't like Table because it includes the heading, CollectionsList or even CollectionsSection? "Entity" just seems weird to me here.

root/components/Table.js
```diff
@@ -12,7 +12,7 @@ import {useTable} from 'react-table';
 
 import loopParity from '../utility/loopParity';
 
-const Table = (({columns, data}) => {
+const Table = (({columns, data}: any) => {
```
reosarevok: I'd really really like these not to be any, if possible.

root/user/UserCollections.js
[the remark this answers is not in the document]
reosarevok: Commit 551fc4449c0d5b1c9500e9ae562fd01d018927f5 seemingly got reverted at some point. viewingOwnProfile should also be removed (or well, at least that it's not part of the condition currently and I'd argue it doesn't make much sense for it to be).

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+    [l('Edit'), '/edit'],
+    [l('Remove'), '/delete'],
+  ]);
+  const columns = React.useMemo(
+    () => [
+      nameColumn,
+      typeColumn,
+      sizeColumn,
+      collaboratorsColumn,
+      ...viewingOwnProfile ? [subscriptionColumn] : [],
```
reosarevok: I'm actually thinking there's no real good reason not to show the subscription column for any collection I can see - I *can* subscribe to other users' public collections, so we should show the subscription status (and possibly a "subscribe"/"unsubscribe" direct link, but that'd be a big bigger a change).

root/components/Table.js
yvanzo: Because `key` is expanded from `getFooProps`.

root/user/UserCollections.js
yvanzo: This component is a _section_ (that features not only a table but also an `h3` title) and matches a specific _entity_ type (not the same as own/collaborative sections). Would `CollectionsEntityTypeSection` be more explicit?

(Note: This is just a 1:1 conversion from TT to JSX but these sections could possibly be merged in the future as tables get the client-sided ability of filtering by entity type.)

root/components/Table.js
yvanzo: See https://github.com/metabrainz/musicbrainz-server/pull/1246#issuecomment-551162606

root/user/UserCollections.js
mwiencek: `$ReadOnly<{...CatalystContextT, user: EditorT}>` (using spreads) would the correct way to extend an object type.

root/components/Table.js
```diff
@@ -0,0 +1,67 @@
...
+    data,
+  });
+
+  /* eslint-disable react/jsx-key */
+  return (
+    <table {...getTableProps({className: 'tbl'})}>
+      <thead>
+        {headerGroups.map(headerGroup => (
+          <tr {...headerGroup.getHeaderGroupProps()}>
+            {headerGroup.headers.map(column => (
```
mwiencek: Can we move some of these functions that don't depend on any variables from the closure to the module scope, here and below? That way the component can reuse the same function objects on every render.

root/components/Table.js
```diff
@@ -0,0 +1,67 @@
...
+              >
+                {column.render('Header')}
+              </th>
+            ))}
+          </tr>
+        ))}
+      </thead>
+      <tbody {...getTableBodyProps()}>
+        {rows.map(
+          (row, i) => prepareRow(row) || (
```
mwiencek: When I read this, I assumed `prepareRow` was some kind of default renderer, with `||` providing a display fallback if that produced null or undefined. But it appears that `prepareRow` has no return value and is simply mandatory to call, with `||` being used to sequence things.

The code would be much clearer if written as
```JS
(row, i) => {
  prepareRow(row);
  return (
    <tr>
      {/* ... */}
    </tr>
  );
}
```

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+  from '../static/scripts/common/utility/formatEntityTypeName';
+import {
+  defineNameColumn,
+  typeColumn,
+  subscriptionColumn,
+  defineActionsColumn,
+} from '../utility/tableColumns';
+
+type Props = {|
+  +$c: {user: EditorT} & CatalystContextT,
```
mwiencek:
```suggestion
  +$c: $ReadOnly<{...CatalystContextT, user: EditorT}>,
```

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+  +collaborativeCollections: CollectionListT,
+  +ownCollections: CollectionListT,
+  +user: EditorT,
+|};
+
+type CollectionListT = {|
+  +[string]: $ReadOnlyArray<CollectionT>,
+|};
+
+const collectionsListTitle = {
```
mwiencek: I'd keep object types plural
```suggestion
const collectionsListTitles = {
```

root/user/UserCollections.js
mwiencek: Parentheses would make the actual spread operation here much easier to parse.
```suggestion
      ...(viewingOwnProfile ? [subscriptionColumn] : []),
```

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+
+  return (
+    <UserAccountLayout
+      entity={user}
+      page="collections"
+      title={l('Collections')}
+    >
+      <h2>{l('My collections')}</h2>
+      {Object.keys(ownCollections).length > 0 ? (
+        Object.keys(ownCollections).sort().map(type => (
```
mwiencek: Can we store `Object.keys(ownCollections)` in a const instead of calling it twice?

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+          {viewingOwnProfile ? (
+            l('You have no collections.')
+          ) : (
+            exp.l('{user} has no public collections.', {user: user.name})
+          )}
+        </p>
+      )}
+      <h2>{l('Collaborative collections')}</h2>
+      {Object.keys(collaborativeCollections).length > 0 ? (
+        Object.keys(collaborativeCollections).sort().map(type => (
```
mwiencek: Same as above

root/utility/tableColumns.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export function defineActionsColumn(
+  actions: $ReadOnlyArray<[string, string]>,
+) {
+  return {
+    Cell: ({row: {original}}: CellT) => (
+      <>
+        {actions.map((actionPair, index) => (
+          <React.Fragment key={actionPair[1] + index === 0 ? '-first' : ''}>
```
mwiencek: I think the intent here is
```suggestion
          <React.Fragment key={actionPair[1] + (index === 0 ? '-first' : '')}>
```
but that's isn't what the current logic is. Why is `'-first'` needed anyway?

mwiencek:
They have some TypeScript typings we can (manually) convert pretty easily: https://github.com/tannerlinsley/react-table/blob/master/TYPESCRIPT.md

For those, apparently you're supposed to copy them into your project and remove the bits you don't use. So we can probably start by just converting the few bits we use, and maybe submit them back upstream if we end up converting them all in the future.

root/utility/tableColumns.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-b9a383135673174ada27c6a8899eba55R47

`-first` stands for _without leading separator_.

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR139

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR138

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR116-R118

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR36

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR26

root/components/Table.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-dfa9d33d3b7fc4efebab124f77237409R15-R64

root/components/Table.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-dfa9d33d3b7fc4efebab124f77237409R15-R64

lib/MusicBrainz/Server/Controller/User.pm
```diff
@@ -323,33 +322,43 @@ sub collections : Chained('load') PathPart('collections')
     $c->model('Collection')->load_entity_count(@$collections);
     $c->model('CollectionType')->load(@$collections);
 
+    my %paired_collections;
```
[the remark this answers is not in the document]
yvanzo: Yes, fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-255249d92c545f0561dab9681845c1fbR325

root/user/UserCollections.js
```diff
@@ -0,0 +1,199 @@
...
+  $c,
+  isCollaborative,
+  collections,
+  type,
+  user,
+}) => {
+  const viewingOwnProfile = !!$c.user && $c.user.id === user.id;
+  const nameColumn = defineNameColumn(N_l('Collection'));
+  const sizeColumn = {
+    Header: () => formatPluralEntityTypeName(type),
```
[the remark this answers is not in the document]
yvanzo: Fixed too, see `Header:` occurrences in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR79

root/user/UserCollections.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR25-R34

root/user/UserCollections.js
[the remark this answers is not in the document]
yvanzo: Fixed by adding `isCollaborativeSection` boolean parameter in https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/6ee5d3d424b1dda8aecefe6be1b9f7d2ac4092e1#diff-3f513fa6fa1fed40110c8c20d602c01fR25-R34

mwiencek:
I tried to add some basic Flow types here and fix a few issues I saw, but others remain:

1. `subscriptionColumn` has "subscribed" for an accessor, but there's no `subscribed` property on `CollectionT`.
2. The style of the tables is messed up for some reason - all the cells are bold and links are black.

yvanzo:
@mwiencek: Fixed (1) in commits https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/04a9f7a41e223341c370388932dad2bf1300fc27 and https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/051e6e132957dc62e22b8226ecd8d844d50c4152, and (2) in commit https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/3e75f812af700768ebd85a44d63be9f4cf00b21f.

root/user/UserCollections.js
yvanzo: Done in commit https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/0c471ca4565043b0edb75ff390e6e012e9a410d6.

mwiencek:
This looks good to me now, I think - I fixed a few small style issues with the types I added, plus dcadf4cc090b0943ad796b479ffda619c8a4391e which wasn't from this PR but bothered me. :) See the force-push.

mwiencek:
Actually, just noticed two of the `s/l/N_l/` changes I made got reverted, so I added them back.

reosarevok:
```
        #   Failed test 'other collection list has 4 cols'
        #   at t/lib/t/MusicBrainz/Server/Controller/User/Collections.pm line 62.
        #          got: '5'
        #     expected: '4'
```

Please take a look at that :)

yvanzo:
Thank you @reosarevok! Fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/1246/commits/1b83d5c968b74f550a1837ab0bb83e7dcf614715. There are so many spurious messages in https://app.circleci.com/jobs/github/metabrainz/musicbrainz-server/7284 I did not find this one.

PR #1247, opened 2019-10-23
MBS-10448: Add Dynamic Range DB to otherdbs whitelist

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10448

No favicon added because it just doesn't seem to have any.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1026,6 +1026,16 @@ const CLEANUPS = {
       return url;
     },
   },
+  'dynamicrangedb': {
+    match: [new RegExp('^(https?://)?dr\\.loudness-war\\.info', 'i')],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      return url.replace(/^https:/, 'http:');
+    },
+    validate: function (url, id) {
+      return id === LINK_TYPES.otherdatabases.release;
+    },
+  },
   'ester': {
     match: [new RegExp('^(https?://)?(www\\.)?ester\\.ee/', 'i')],
     type: LINK_TYPES.otherdatabases,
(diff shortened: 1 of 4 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1026,6 +1026,19 @@ const CLEANUPS = {
       return url;
     },
   },
+  'dynamicrangedb': {
+    match: [new RegExp('^(https?://)?dr\\.loudness-war\\.info', 'i')],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      // Seems to support http only
+      url = url.replace(/^(?:https?:\/\/)?dr\.loudness-war\.info\//, 'http://dr.loudness-war.info/');
```
mwiencek: Since we're just removing the 's' from https here, do we need the whole `dr\.loudness-war\.info\/` part of the regexp? I'd think we can assume that's already there, if it matches the `match` regexp above.

reosarevok: I guess not :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1026,6 +1026,18 @@ const CLEANUPS = {
       return url;
     },
   },
+  'dynamicrangedb': {
+    match: [new RegExp('^(https?://)?dr\\.loudness-war\\.info', 'i')],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      // Seems to support http only
+      return url.replace(/^https?:\/\//, 'http://');
```
yvanzo: Can be simplified further so as to avoid replacing matched string with the same string:
```suggestion
      return url.replace(/^https:/, 'http:');
```

PR #1244, opened 2019-10-21
JavaScript test coverage for Selenium

mwiencek:
Generates JavaScript test coverage for our Selenium tests using Istanbul. If all goes well, should be saved as a build artifact on CircleCI for viewing.

PR #1251, opened 2019-10-28
Add a script to check a specific eslint rule

mwiencek:
This can be invoked like

```sh
./script/check_eslint_rule \
  'block-scoped-var: [warn]' \
  root/
```

to check a specific rule under a specific directory

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+root/static/build/**/*.js
+root/static/lib/**/*.js
--- a/HACKING.md
+++ b/HACKING.md
@@ -205,13 +205,35 @@ the test database is clean and has t/sql/selenium.sql loaded into it.
 
 ### Code standards
 
-We use `Perl::Critic` to enforce certain code standards. The list of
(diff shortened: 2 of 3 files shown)


yvanzo:
Using NodeJS v10.10.0:

```console
$ ./node_modules/.bin/eslint root/types.js
Error: /home/yvan/musicbrainz-server/.eslintrc.yaml:
        Configuration for rule "spaced-comment" is invalid:
        Value {"null":null} should be string.
```
It might be a regression caused by merged commit https://github.com/metabrainz/musicbrainz-server/pull/1244/commits/1cc0bb28988ec0a70f9c1fcc7b02bd6f65774d10.

reosarevok:
I have a change for that configuration on the spaces PR, not sure whether that'd change anything though :)

mwiencek:
Yeah, the change in https://github.com/metabrainz/musicbrainz-server/pull/1252 works for me. I'm really confused why the rule I added broke - I tested it on several files and it was working when I added it. :\

PR #1253, opened 2019-10-30
MBS-10389: Add “IRC” link to footer

yvanzo:
# MBS-10389: Add “IRC” link to footer

### DIFF
--- a/root/layout.tt
+++ b/root/layout.tt
@@ -103,6 +103,7 @@
                 <a href="https://metabrainz.org/donate" class="internal">[% l('Donate') %]</a> |
                 <a href="//wiki.musicbrainz.org/" class="internal">[% l('Wiki') %]</a> |
                 <a href="https://community.metabrainz.org/" class="internal">[% l('Forums') %]</a> |
+                <a href="/doc/Communication/IRC" class="internal">[% l('IRC') %]</a> |
                 <a href="http://tickets.metabrainz.org/" class="internal">[% l('Bug Tracker') %]</a> |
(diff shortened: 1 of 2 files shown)


root/layout.tt
```diff
@@ -103,6 +103,7 @@
                 <a href="https://metabrainz.org/donate" class="internal">[% l('Donate') %]</a> |
                 <a href="//wiki.musicbrainz.org/" class="internal">[% l('Wiki') %]</a> |
                 <a href="https://community.metabrainz.org/" class="internal">[% l('Forums') %]</a> |
+                <a href="/doc/Communication/IRC" class="internal">[% l('IRC') %]</a> |
```
Freso: I think I’d prefer "Chat" over "IRC", FWIW.

yvanzo: There is only an “IRC” doc page for now. When there will be another doc page about another tool, e.g. Telegram , or a “Chat” doc page to carry both, then this link can be changed.

Freso: We don’t have a doc page for "Forums" either, but we still don’t call the link "community" (the domain) nor do we call it "Discourse" (the technology/platform). The "bug tracker" link is also is also not called "tickets" or "JIRA".

IRC is the technology/platform, but what we do there is chat.

Also see https://community.metabrainz.org/t/idea-rename-irc-to-chat/450955?u=freso

yvanzo: Thanks for having opened ticket MBS-10470 that can be used for follow-up.
I’m closing this conversation as this PR is already merged and forums/tickets are preferred.

PR #1255, opened 2019-11-04
Flow 0.111

mwiencek:
I asked in their Discord if the computed property stuff is a bug, but didn't get a response yet. Should be fine to merge either way.


mwiencek:
Updated this based on the recommendations from one of the Flow devs.

PR #1259, opened 2019-11-07
MBS-10465: Beta: Handle legacy time zones in formatUserDate

mwiencek:
# Fix MBS-10465 in beta

I used the alternatives suggested at https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

Additionally, we now fall back to UTC if nothing else works.

### DIFF
--- a/root/utility/formatUserDate.js
+++ b/root/utility/formatUserDate.js
@@ -61,12 +61,45 @@ const patterns = {
   }],
 };
 
+const TZ_CANCUN = 'America/Cancun';
+const TZ_CHICAGO = 'America/Chicago';
+const TZ_DENVER = 'America/Denver';
+const TZ_HONOLULU = 'Pacific/Honolulu';
+const TZ_LISBON = 'Europe/Lisbon';
+const TZ_LOSANGELES = 'America/Los_Angeles';
+const TZ_NEWYORK = 'America/New_York';
+const TZ_PARIS = 'Europe/Paris';
+const TZ_PHOENIX = 'America/Phoenix';
+const TZ_SOFIA = 'Europe/Sofia';
+
+/*
+ * These short names are selectable in /account/preferences, but
+ * unsupported by Intl.DateTimeFormat in Node. We map them to a
+ * supported "Area/Location" name that observes the referenced time
+ * zone. This should fortunately match the behavior of Perl's
+ * DateTime::TimeZone, which does not define all of these as hard-coded
+ * offsets, but takes DST into account where it makes sense to.
+ */
+const timeZoneFallbacks = {
+  CET: TZ_PARIS,
+  CST6CDT: TZ_CHICAGO,
+  EET: TZ_SOFIA,
+  EST: TZ_CANCUN,
+  EST5EDT: TZ_NEWYORK,
+  HST: TZ_HONOLULU,
+  MET: TZ_PARIS,
+  MST: TZ_PHOENIX,
+  MST7MDT: TZ_DENVER,
+  PST8PDT: TZ_LOSANGELES,
+  WET: TZ_LISBON,
+};
+
 let HAS_INTL_DATETIMEFORMAT = true;
 
 try {
   const now = new Date();
-  formatDateUsingPattern(now, '%a', 'America/Chicago', 'en-US');
-  formatDateUsingPattern(now, '%c', 'America/Chicago', 'en-US');
+  formatDateUsingPattern(now, '%a', TZ_CHICAGO, 'en-US');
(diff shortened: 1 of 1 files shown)


root/utility/formatUserDate.js
```diff
@@ -81,10 +104,22 @@ function formatDateUsingPattern(
...
-      locale,
-      Object.assign({timeZone: timezone}, options),
-    );
+    try {
+      formatter = new Intl.DateTimeFormat(
+        locale,
+        Object.assign({timeZone: timezone}, options),
+      );
+    } catch (e) {
+      if (e instanceof RangeError && /time zone/.test(e.message)) {
```
mwiencek: In Node v12.12.0 on macOS I get a slightly different error message, "RangeError: Invalid time zone specified: CET," not "RangeError: Expected Area/Location(/Location)* for time zone, got CET" as on beta (where I believe we use Node v10). So I'm just checking for "time zone" in the message here. (I do believe there are other cases where a `RangeError` might be thrown that has nothing to do with the time zone, which is why I'm checking the message at all.)

mwiencek: Also note that "CET" works fine in my browser (Firefox 70.0.1) despite not working in Node. It's implementation-dependent what time zone names are supported.

yvanzo: Node.js supports UTC, IANA time zones (Area/City), and GMT offsets (GMT±X).

root/utility/formatUserDate.js
```diff
@@ -61,12 +61,35 @@ const patterns = {
...
+  CST6CDT: TZ_CHICAGO,
+  EET: TZ_SOFIA,
+  EST: TZ_CANCUN,
+  EST5EDT: TZ_NEWYORK,
+  MET: TZ_PARIS,
+  MST: TZ_PHOENIX,
+  MST7MDT: TZ_DENVER,
+  PST8PDT: TZ_LOSANGELES,
+  WET: TZ_LISBON,
+};
```
yvanzo: That would break next daylight saving time change when user’s computers switch back to CEST or WEST, see this more complete list of [time zone abbreviations](https://www.timeanddate.com/time/zones/) (which contains homonyms).

Another option would be to use `jstimezonedetect` which doesn’t return abbreviations. (Yes I know it is yet another dependency, but it is much lighter than `moment` and it just _gives you the zone info key representing your device's time zone setting_, no less no more.)

mwiencek: I'm a bit confused about what you mean. The names in `timeZoneFallbacks` aren't being guessed or detected at all, they're what the user selected in /account/preferences. There's no option for e.g. CEST in the list; Perl's `DateTime` correctly switched to CEST where needed with CET set for the time zone.

yvanzo: Sorry, I thought that name was guessed from user’s browser:
https://github.com/metabrainz/musicbrainz-server/blob/db2845bb84eb6399395657800975ff2fbc7ea548/root/static/scripts/account/components/PreferencesForm.js#L111
but it is indeed filtered with:
https://github.com/metabrainz/musicbrainz-server/blob/db2845bb84eb6399395657800975ff2fbc7ea548/root/static/scripts/account/components/PreferencesForm.js#L115-L124
which doesn’t allow for other abbreviations.

root/utility/formatUserDate.js
```diff
@@ -61,12 +61,35 @@ const patterns = {
...
+const TZ_PARIS = 'Europe/Paris';
+const TZ_PHOENIX = 'America/Phoenix';
+const TZ_SOFIA = 'Europe/Sofia';
+
+const timeZoneFallbacks = {
+  CET: TZ_PARIS,
+  CST6CDT: TZ_CHICAGO,
+  EET: TZ_SOFIA,
+  EST: TZ_CANCUN,
+  EST5EDT: TZ_NEWYORK,
```
yvanzo:
```suggestion
  EST5EDT: TZ_NEWYORK,
  HST: TZ_HONOLULU,
```

root/utility/formatUserDate.js
```diff
@@ -61,12 +61,35 @@ const patterns = {
   }],
 };
 
+const TZ_CANCUN = 'America/Cancun';
+const TZ_CHICAGO = 'America/Chicago';
+const TZ_DENVER = 'America/Denver';
```
yvanzo:
```suggestion
const TZ_DENVER = 'America/Denver';
const TZ_HONOLULU = 'Pacific/Honolulu';
```

mwiencek:
Added HST + a comment above `timeZoneFallbacks`

PR #1260, opened 2019-11-08
MBS-9654: Show expanded ACs on recording merges 

reosarevok:
https://tickets.metabrainz.org/browse/MBS-9654

When merging recordings, it's pretty useful to be able to see whether, say, the same artist credit name hides a different artist (and the disambiguation comment for the artist).

This makes the merge page and the edit display show expanded artist credits in the artist cell to make these things more visible than just on hover.

This is on top of https://github.com/metabrainz/musicbrainz-server/pull/1162 and should probably be reviewed after it.

Examples:
_**Edit page**_
Screenshot from 2019-11-08 17-40-42

**_Merge page_**
Screenshot from 2020-01-07 19-55-32

### DIFF
--- a/root/static/scripts/common/components/ExpandedArtistCredit.js
+++ b/root/static/scripts/common/components/ExpandedArtistCredit.js
@@ -1,0 +1,60 @@
+/*
+ * @flow
+ * This file is part of MusicBrainz, the open internet music database.
+ * Copyright (C) 2019 MetaBrainz Foundation
+ * Licensed under the GPL version 2, or (at your option) any later version:
+ * http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+import commaOnlyList from '../i18n/commaOnlyList';
+
+import ArtistCreditLink from './ArtistCreditLink';
+import DescriptiveLink from './DescriptiveLink';
+
+export const ExpandedArtistCreditList = ({
+  artistCredit,
+}: {artistCredit: ArtistCreditT}) => {
+  if (!artistCredit) {
+    return null;
+  }
+
+  const names = artistCredit.names;
+  let artistList = [];
+
+  if (names.some(x => x.artist.name !== x.name || x.artist.comment)) {
+    artistList = names.map(name => {
+      if (name.artist.name === name.name) {
+        return <DescriptiveLink entity={name.artist} />;
+      }
+      return exp.l(
+        '{artist} as {name}',
+        {
+          artist: <DescriptiveLink entity={name.artist} />,
+          name: name.name,
+        },
+      );
+    });
+  }
+
+  if (artistList.length) {
+    return (
+      <span className="expanded-ac-list">{commaOnlyList(artistList)}</span>
+    );
+  }
+
+  return null;
+};
+
+const ExpandedArtistCredit = ({
+  artistCredit,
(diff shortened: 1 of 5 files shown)


root/static/scripts/common/components/ExpandedArtistCredit.js
```diff
@@ -0,0 +1,68 @@
...
+  if (!artistCredit) {
+    return null;
+  }
+
+  const names = artistCredit.names;
+  const artistList = [];
+  let showList = false;
+
+  if (reduceArtistCredit(artistCredit) !== names[0].artist.name ||
+    names.length > 1 || names[0].artist.comment) {
```
mwiencek: I think I'd just replace this condition with
```JS
if (names.some(x => x.artist.name !== x.name || x.artist.comment)) {
```
and remove the whole `showList` thing. It looks like the code is using this condition to avoid running all the stuff below if not necessary, but has some false positives making the `showList` thing is necessary. But I'd think we could just encode the complete logic here - it's probably not any slower than running `reduceArtistCredit` + dealing with any false positives.

root/static/scripts/common/components/ExpandedArtistCredit.js
```diff
@@ -0,0 +1,68 @@
...
+    names.map(name => {
```
mwiencek: You can assign the result of the map here to `artistList = ` and change `artistList.push` in the map callback to just `return`. The map will return a list of React elements then.

root/static/scripts/common/components/ExpandedArtistCredit.js
```diff
@@ -0,0 +1,68 @@
...
+            artist: <DescriptiveLink entity={name.artist} />,
+            name: name.name,
+          },
+        ));
+        showList = true;
+      }
+    });
+  }
+
+  if (showList) {
```
mwiencek:
```suggestion
  if (artistList.length) {
```

root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -70,7 +70,6 @@ const MpIcon = hydrate<MpIconProps>('span.ac-mp', (
 const ArtistCreditLink = ({
   artistCredit,
   showDeleted = true,
-  showEditsPending = true,
```
yvanzo: Isn’t that supposed to be passed to `EntityLink` on [line 88](https://github.com/metabrainz/musicbrainz-server/pull/1260/commits/653075bbf021c142942fb3480376b6a799dd2da9#diff-1e1382b906771e43beecbf0a967eb48cL88) instead?

reosarevok: It's already passed (as showEditsPending={!artistCredit.editsPending}) meaning this does nothing right now. If you mean we should not just do !artistCredit.editsPending there, then I don't know! :)

root/static/scripts/common/components/ExpandedArtistCredit.js
reosarevok: Seems sensible, changed!

root/static/scripts/common/components/ArtistCreditLink.js
yvanzo: This has been removed as part of GCI.

PR #1262, opened 2019-11-11
MBS-10471: Add option to view edits by date closed

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10471

Also adding an option for date applying.
For the closing date filter, filtering specifically for edits with a close_time set.  For the applying date filter, filtering specifically for open edits.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Query.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Query.pm
@@ -67,7 +67,7 @@ has combinator => (
 );
 
 has order => (
-    isa => enum([qw( asc desc rand )]),
+    isa => enum([qw( asc desc closed_asc closed_desc vote_closing_asc vote_closing_desc rand )]),
     is => 'ro',
     required => 1,
     default => 'desc'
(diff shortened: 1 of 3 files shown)


root/edit/search_macros.tt
```diff
@@ -19,6 +19,10 @@
     [%- sort_select_block = '<select name="order">
       <option value="desc"' _ selected_if_matches('desc', query.order, 1) _ '>' _ l('newest first') _ '</option>
       <option value="asc"'  _ selected_if_matches('asc', query.order)  _ '>' _ l('oldest first') _ '</option>
+      <option value="closedesc"' _ selected_if_matches('closeddesc', query.order) _ '>' _ l('recently closed first')     _ '</option>
+      <option value="closeasc"' _ selected_if_matches('closedasc', query.order) _ '>' _ l('earliest closed first')     _ '</option>
+      <option value="applyasc"' _ selected_if_matches('expiringasc', query.order) _ '>' _ l('closing sooner first')     _ '</option>
+      <option value="applydesc"' _ selected_if_matches('expiringdesc', query.order) _ '>' _ l('closing later first')     _ '</option>
```
mwiencek: also `s/apply/expiring/` in the value attributes here

lib/MusicBrainz/Server/EditSearch/Query.pm
```diff
@@ -67,7 +67,7 @@ has combinator => (
 );
 
 has order => (
-    isa => enum([qw( asc desc rand )]),
+    isa => enum([qw( asc desc closedasc closeddesc expiringasc expiringdesc rand )]),
```
yvanzo: 1. The term “expiring” should probably be replaced with “vote closing” since MBS-9732.
2. Aggregated values could be replaced with separate `sort_by` query parameter, or at the very least be more readable using a `-` separator.

reosarevok: Fair enough. sort_by seems annoying but - does make it much more clear :)

root/edit/search_macros.tt
```diff
@@ -19,6 +19,10 @@
     [%- sort_select_block = '<select name="order">
       <option value="desc"' _ selected_if_matches('desc', query.order, 1) _ '>' _ l('newest first') _ '</option>
       <option value="asc"'  _ selected_if_matches('asc', query.order)  _ '>' _ l('oldest first') _ '</option>
+      <option value="closed-desc"' _ selected_if_matches('closed-desc', query.order) _ '>' _ l('recently closed first')     _ '</option>
+      <option value="closed-asc"' _ selected_if_matches('closed-asc', query.order) _ '>' _ l('earliest closed first')     _ '</option>
+      <option value="voteclosing-asc"' _ selected_if_matches('voteclosing-asc', query.order) _ '>' _ l('closing sooner first')     _ '</option>
+      <option value="voteclosing-desc"' _ selected_if_matches('voteclosing-desc', query.order) _ '>' _ l('closing later first')     _ '</option>
```
yvanzo: Labels should specify it is about _vote_ closing.

reosarevok: I feel the difference for the user here is pretty much nonexistent, but it doesn't hurt, so I'll change it.

lib/MusicBrainz/Server/EditSearch/Query.pm
```diff
@@ -67,7 +67,7 @@ has combinator => (
 );
 
 has order => (
-    isa => enum([qw( asc desc rand )]),
+    isa => enum([qw( asc desc closed-asc closed-desc voteclosing-asc voteclosing-desc rand )]),
```
mwiencek: Every other field/value I can see in the edit search uses underscores instead of hyphens to separate words, so I'd prefer that for consistency. And if we're doing that for readability, we might as well separate vote closing too (`vote_closing_asc`).

PR #1268, opened 2019-11-14
Eslint fixes by rule, pt. 1

reosarevok:
I've started fixing eslint issues rule by rule (mostly the ones that are auto-fixable). This is the first bit, quite a few more to come (48 commits total).

### DIFF
--- a/root/static/scripts/jquery.flot.musicbrainz_events.js
+++ b/root/static/scripts/jquery.flot.musicbrainz_events.js
@@ -26,9 +26,8 @@
                     return !plotEvent;
             });
             return plotEvent;
-        } else {
-            return false;
-        }
+        } 
+        return false;
     }
 
     function drawCrosshairLine(plot, ctx, x, color) {
--- a/root/static/scripts/tests/release-editor/fields.js
+++ b/root/static/scripts/tests/release-editor/fields.js
@@ -55,11 +55,11 @@ fieldTest("mediums having their \"loaded\" observable set correctly", function (
 
     mediums([
         new fields.Medium({ tracks: [] }),
-        new fields.Medium({ tracks: [ {} ] }),
+        new fields.Medium({ tracks: [{}] }),
         new fields.Medium({ id: 1, tracks: [] }),
         new fields.Medium({ originalID: 1, tracks: [] }),
-        new fields.Medium({ id: 1, tracks: [ {} ] }),
-        new fields.Medium({ originalID: 1, tracks: [ {} ] })
+        new fields.Medium({ id: 1, tracks: [{}] }),
+        new fields.Medium({ originalID: 1, tracks: [{}] })
     ]);
 
     t.equal(mediums()[0].loaded(), true, "medium without id or tracks is considered loaded");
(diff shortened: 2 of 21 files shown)


root/static/scripts/jquery.flot.musicbrainz_events.js
```diff
@@ -74,4 +74,4 @@
         name: 'musicbrainzEvents',
         version: '1.0'
     });
-})(jQuery);
+}(jQuery));
```
yvanzo: Actually using default `outside` option for `wrap-life` rule, but current code follows `inside` style.

root/static/scripts/timeline.js
```diff
@@ -616,7 +616,7 @@ class TimelineLine {
             }
         }
     };
-})();
+}());
```
yvanzo: Same as above.

root/static/scripts/jquery.flot.musicbrainz_events.js
reosarevok: I checked, and there's four fixes needed for inside vs two for outside, so this is still better it seems :)

mwiencek: I've always used `}(...))`, mostly as [a habit from JSLint](https://youtu.be/taaEzHI9xyY?t=2019).

But note that for arrow functions, `((x) => { return x }(1))` is actually a syntax error; you have to write it like `((x) => { return x })(1)`. This rule doesn't apply to arrow functions, but it's interesting to note and maybe an argument for the other way if that bothers you. On the other hand, our usage of these should be extremely rare, if we ever do use them at all.

reosarevok: I dropped the wrap-iife commit to hopefully unblock this.

root/static/scripts/tests/release-editor/fields.js
```diff
@@ -55,11 +55,11 @@ fieldTest("mediums having their \"loaded\" observable set correctly", function (
...
     mediums([
         new fields.Medium({ tracks: [] }),
-        new fields.Medium({ tracks: [ {} ] }),
+        new fields.Medium({ tracks: [{}] }),
         new fields.Medium({ id: 1, tracks: [] }),
         new fields.Medium({ originalID: 1, tracks: [] }),
-        new fields.Medium({ id: 1, tracks: [ {} ] }),
-        new fields.Medium({ originalID: 1, tracks: [ {} ] })
+        new fields.Medium({ id: 1, tracks: [{}] }),
+        new fields.Medium({ originalID: 1, tracks: [{}] })
```
yvanzo: I’m not sure these changes are making the code more readable.

reosarevok: I dunno, for me they are, since they make it look the same as usual. If you want to show specifically that these are empty, we could add a comment? `[{/* empty */}]` or something? But I'm not sure there's really any doubt here.

mwiencek: There's an "objectsInArrays" exception for https://eslint.org/docs/rules/space-in-brackets but you can only enforce one style or the other, not allow both without disabling space-in-brackets entirely AFAICT.

yvanzo: Personal preference mostly, the new form seems more likely confused with `[]`.
But as told on IRC, that would be even more cumbersome to work around eslint here.

PR #1272, opened 2019-11-14
Eslint fixes by rule, pt. 5: quotes

reosarevok:
On top of https://github.com/metabrainz/musicbrainz-server/pull/1271

Ok, so this one is huge, but it really should be trivial, since the way our rule is set should mean that no string that would need escaping will be "fixed"


reosarevok:
Too many conflicts by now, and being superseded by newer PRs anyway with the new system of dual eslintrc files.

PR #1275, opened 2019-11-15
MBS-10479: Show RG types on recording page

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10479


root/recording/RecordingIndex.js
```diff
@@ -50,6 +50,7 @@ const RecordingAppearancesTable = ({
         <th>{l('Track Artist')}</th>
         <th>{l('Release Title')}</th>
         <th>{l('Release Artist')}</th>
+        <th>{l('Release Type')}</th>
```
mwiencek: Wouldn't "Release Group Type" be better?

reosarevok: It would, yes :)

PR #1277, opened 2019-11-18
MBS-10089: Limit Works tab to written/performed works

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10089

In general, an artist's work tab is meant to show the works the artist has performed or written. It doesn't make a lot of sense to show works linked with relationships such as "named after" or "commissioned" in this tab.
This limits the works listed in the tab to those with relationships in @WRITER_RELATIONSHIP_GIDS. All other works will still be shown in the artist's relationships page.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1171


mwiencek (approved):

reosarevok:
Hmm. Actually, now that we will be able to filter works where the artist appears as a writer, as a performer or as something else as part of https://github.com/metabrainz/musicbrainz-server/pull/1171, I wonder if there's a real use to removing these from the works page. I can see how having a "named after" work in there feels a bit weird but it's also not useless data as long as it's clearly indicated...

mwiencek:
It seems like a fairly minor issue overall, especially if they can be filtered in the future, so I don't have a strong opinion about removing them.

reosarevok:
I think keeping them but under its own column as we are doing in https://github.com/metabrainz/musicbrainz-server/pull/1171 makes sense, so let's drop this.

PR #1278, opened 2019-11-19
Convert some medium macros to React

mwiencek:


mwiencek:
Since these are placed under root/static/scripts, I updated pot files in each commit to avoid Jed throwing exceptions about missing keys in the tests. (Actually, I could get used to this, since it's nice to see how each commit affects the strings by itself.)

root/static/scripts/tests/relationship-editor.js
```diff
@@ -96,7 +96,10 @@ var testRelease = {
                     artistCredit: {names: []},
                 }
             ],
-            format: "Vinyl",
+            format: {
+                name: "Vinyl",
+            },
```
reosarevok: Not directly about this code, but: medium formats have MBIDs we seem to not be using at all, or am I missing something? :)

root/static/scripts/common/entity.js
```diff
@@ -412,7 +412,9 @@ import formatTrackLength from './utility/formatTrackLength';
             this.tracks = _.map(data.tracks, x => new Track(x));
 
             const positionArgs = {
-                medium_format: this.format,
+                medium_format: this.format
+                    ? lp_attributes(this.format.name, 'medium_format')
+                    : null,
```
reosarevok: Can't this use the stuff from 312a3a7b0e3290a9eea80d00ddb612f41df82798 ?

root/static/scripts/common/components/MediumDescription.js
```diff
@@ -0,0 +1,37 @@
...
+
+import * as React from 'react';
+
+import {addColonText} from '../i18n/addColon';
+import isolateText from '../utility/isolateText';
+import mediumFormatName from '../utility/mediumFormatName';
+
+type Props = {
+  +medium: MediumT,
+};
```
reosarevok: Isn't it a bit overkill not to just do `({medium}: {medium: MediumT})` ? :)

root/static/scripts/common/components/MediumLink.js
```diff
@@ -0,0 +1,30 @@
...
+import * as React from 'react';
+
+import linkedEntities from '../linkedEntities';
+
+import DescriptiveLink from './DescriptiveLink';
+import MediumDescription from './MediumDescription';
+
+type Props = {
+  +medium: MediumT,
+};
```
reosarevok: Same comment about possible prop overkill.

root/static/scripts/tests/relationship-editor.js
mwiencek: I don't think we use them in the code, yet from what I can see. So don't think you're missing anything. :)

root/static/scripts/common/entity.js
mwiencek: See 6cbcb2ee3433264f533e1ec8d4d61a873613e2c4

This changes the strings used.

reosarevok: This seems sensible and less prone to errors IMO, if @yvanzo is happy with the string change (should only change the way strings are built, but not the final result).

yvanzo: I’m happy! :grinning:

PR #1280, opened 2019-11-20
Flow any fixes

mwiencek:
This PR mainly aims to reduce the number of uses of the `any` type. (There's only one remaining after this.) Where fixing the type implicated some smelly code, I fixed the code too.


root/release/ReleaseHeader.js
```diff
@@ -41,6 +41,7 @@ const ReleaseHeader = ({release, page}: Props) => {
             <EntityLink
               content={rgLink}
               entity={release.releaseGroup}
+              nameVariation
```
reosarevok: Nitpick of course, but it seems weird to call it a name variation here - we're just overriding the name bit completely. Maybe we should call the variable "hasCustomContent" or something else? (but it might not matter at all)

root/static/scripts/edit/components/Tooltip.js
```diff
@@ -13,28 +13,38 @@
...
+    const container = this.containerRef.current;
+    const links = container?.getElementsByTagName('a');
+    if (links) {
+      for (let i = 0; i < links.length; i++) {
+        links[i].setAttribute('target', '_blank');
+      }
     }
   }
 
+  containerRef: {current: HTMLDivElement | null};
```
reosarevok: Is there a way to mark this as being a type? I'm just used to having types defined first and their usage later, so it confused me for a bit - not a big deal at all, but still.

root/release/ReleaseHeader.js
mwiencek: Hmm. `hasCustomContent` could be misleading too, since that's already specified by having a `content` prop, whereas this is more about if the content prop has text that differs from the entity name. `textContentDiffersFromEntityName` seems accurate, but might be too wordy. :)

mwiencek: Another way to look at it is that `nameVariation` just controls whether the content is wrapped in `<span className="name-variation" />` with hover text added.

root/static/scripts/edit/components/Tooltip.js
mwiencek: I think eslint required it to be down here because of the [react/sort-comp](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md) rule. Like you said, it actually makes more sense to me to have all type definitions at the start of the class, so we should probably fix the rule. Would changing the rule and moving it to the top of the class help, or did you have something else in mind?

reosarevok: I think that'd make it more clear to me, yes - @yvanzo: what do you think about this case?

yvanzo: Just do it! :athletic_shoe:

mwiencek: Fixed in the latest commit. :)

root/release/ReleaseHeader.js
reosarevok: Fiiine :)

PR #1282, opened 2019-11-21
MBS-10350: Stop highlighting Google+ links

yvanzo:
# Address MBS-10350

The social network has been shutdown in April 2019.


yvanzo:
Yes, URL cleanup is still there to allow adding old links.

PR #1283, opened 2019-11-22
MBS-7379: Add median duration to works

reosarevok:
https://tickets.metabrainz.org/browse/MBS-7379

It's pretty useful to be able to see at a glance what's the expected duration for recordings of a specific work.

This ignores recordings that are marked as partial recordings, as part of a medley, or are related to more than one work.


reosarevok:
This should probably be shown in more places eventually (inline search results maybe? possibly the works page?). But a quick check showed a lot of wrong durations, because of stuff that should be partial not being set to partial, so maybe it's better not to do that just yet to avoid people not selecting the right work because of it?

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -423,6 +424,62 @@ sub _find_recording_artists
     }
 }
 
+=method load_median_length
+
+This method will load the median length for the recordings of the work,
+ignoring any recordings marked as partial or as being a medley.
+
```
yvanzo:
```suggestion

See the only recording-work relationship type "performance" and its attributes:
https://musicbrainz.org/relationship/a3005666-a872-32c3-ad06-98af558e99b0
```

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -423,6 +424,62 @@ sub _find_recording_artists
...
+            GROUP BY lrw2.entity0
+            HAVING COUNT(*) > 1
+        )
+        AND NOT EXISTS (
+            SELECT 1
+            FROM link_attribute la
+            JOIN link_attribute_type lat ON la.attribute_type = lat.id
+            WHERE la.link = lrw.link
+            AND lat.gid IN (
+                'd2b63be6-91ec-426a-987a-30b47f8aae2d',
```
yvanzo:
```suggestion
                'd2b63be6-91ec-426a-987a-30b47f8aae2d', -- partial
```

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -423,6 +424,62 @@ sub _find_recording_artists
...
+                '37da3398-5d1b-4acb-be25-df95e33e423c'
```
yvanzo:
```suggestion
                '37da3398-5d1b-4acb-be25-df95e33e423c' -- medley
```

root/layout/components/sidebar/WorkSidebar.js
```diff
@@ -78,6 +80,15 @@ const WorkSidebar = ({$c, work}: Props) => {
               </SidebarProperty>
             ) : null}
 
+            {length ? (
+              <SidebarProperty
+                className="work-length"
+                label={addColonText(l('Median Length'))}
+              >
+                {formatTrackLength(length)}
+              </SidebarProperty>
```
yvanzo: Since there are a lot of wrong durations, I suggest to either put a disclaimer here that this is experimental, or show in work report instead of work page.

reosarevok:
Wrong durations are mostly happening in pages for main classical works in my testing (so, the main work for a sonata will be messy, but the movements will be more or less fine, and I expect non-classical to also be mostly fine). That said, a disclaimer that it is experimental seems sensible, plus maybe a report showing what, in your opinion? Works with recordings that are more than half / double the average length? (the more I think about this, the more I think we should go for average over median here).

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -425,6 +426,66 @@ sub _find_recording_artists
...
+    my ($self, $ids, $map) = @_;
+    return unless @$ids;
+
+    my $query = "
+        SELECT lrw.entity1 AS work_id, median(r.length) AS length
+        FROM l_recording_work lrw
+        JOIN recording r ON lrw.entity0 = r.id
+        WHERE lrw.entity1 IN (" . placeholders(@$ids) . ")
+        AND NOT EXISTS (
+            SELECT lrw2.entity0, COUNT(*)
```
mwiencek:
```suggestion
            SELECT 1
```

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -425,6 +426,66 @@ sub _find_recording_artists
...
+        )
+        AND NOT EXISTS (
+            SELECT 1
+            FROM link_attribute la
+            JOIN link_attribute_type lat ON la.attribute_type = lat.id
+            WHERE la.link = lrw.link
+            AND lat.gid IN (
+                'd2b63be6-91ec-426a-987a-30b47f8aae2d', -- partial
+                '37da3398-5d1b-4acb-be25-df95e33e423c' -- medley
+            )
```
mwiencek: I'd just check `la.attribute_type IN (579, 750)`, since there's not really any technical reason to prefer the gids internally here. The IDs are just as permanent, and we can save a modest couple milliseconds by not joining `link_attribute_type` on works with a huge number of recordings.

I was testing the query against https://musicbrainz.org/work/0c80db24-389e-3620-8e0b-84dc2b7c009a on pink, and  it takes somewhere between 15 and 20 ms to execute, which is very significant for one tiny number on the page, but that kind of work is also an outlier...for comparison, a work with two recordings took about 0.6 ms.

lib/MusicBrainz/Server/Entity/Work.pm
```diff
@@ -97,6 +102,7 @@ around TO_JSON => sub {
         %{ $self->$orig },
         attributes => [map { $_->TO_JSON } $self->sorted_attributes],
         languages => [map { $_->TO_JSON } $self->all_languages],
+        length  => $self->length,
```
mwiencek: Perhaps only output this if it's loaded?
```suggestion
        defined $self->length ? (length => $self->length) : (),
```

lib/MusicBrainz/Server/Data/Work.pm
reosarevok: That page takes forever to load already so I don't think this query is a huge added issue tbh... Will get rid of that join though.

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -425,6 +426,66 @@ sub _find_recording_artists
...
+            SELECT 1
+            FROM l_recording_work lrw2
+            WHERE lrw2.entity0 = lrw.entity0
+            GROUP BY lrw2.entity0
+            HAVING COUNT(*) > 1
+        )
+        AND NOT EXISTS (
+            SELECT 1
+            FROM link_attribute la
+            JOIN link_attribute_type lat ON la.attribute_type = lat.id
```
mwiencek, line 471:
```suggestion
```

reosarevok:
I added this, does it seem like enough?

Screenshot from 2020-09-02 08-28-45

I also changed it to ignore cover recordings, since those could have a completely different duration if they have been recorded in a very different style or whatever :)

lib/MusicBrainz/Server/Data/Work.pm
```diff
@@ -425,6 +426,66 @@ sub _find_recording_artists
...
+            HAVING COUNT(*) > 1
+        )
+        AND NOT EXISTS (
+            SELECT 1
+            FROM link_attribute la
+            WHERE la.link = lrw.link
+            AND la.attribute_type IN (
+                567, -- cover
+                579, -- partial
+                750 -- medley
```
yvanzo, line 475: Didn’t we have some sort of standard comment to track `id` usage instead of `gid` in code?

root/layout/components/sidebar/WorkSidebar.js
```diff
@@ -77,6 +81,26 @@ const WorkSidebar = ({$c, work}: Props): React.Element<'div'> => {
...
+                {formatTrackLength(length)}
+                {' '}
+                <InformationIcon
+                  className="align-top"
+                  title={l(
+                    `Autocalculated from all non-partial, non-medley,
+                     non-cover recordings linked only to this work.
+                     If the data seems off, it’s likely some
+                     recording relationships need to be marked as
+                     partial or medley.`,
```
yvanzo, line 98: Or it is just the work can be played a random number of times, which is quite frequent, as already mentioned in comments to the ticket.

root/layout/components/sidebar/WorkSidebar.js
```diff
@@ -77,6 +81,26 @@ const WorkSidebar = ({$c, work}: Props): React.Element<'div'> => {
               </SidebarProperty>
             ) : null}
 
+            {length == null ? null : (
+              <SidebarProperty
+                className="work-length"
+                label={addColonText(l('Median Length'))}
```
yvanzo, line 87: This should probably not be under “Work information” as this is not an information stored in the database. I would probably suggest to move it either under “Recordings” in the main panel or under a separate section in the sidebar.

“Median Length” is not about the work, it is about its recordings. If the section title refers to recordings, it would probably make more sense.

yvanzo (commented):
I’m not sure to understand why you want to exclude covers which have not been mentioned in the ticket. Maybe you are thinking about a specific musical genre? Otherwise, the label should reflect that too.

reosarevok:
Excluding covers is because usually if I'm looking for info about a work I'm curious about the original length by the original artist, IMO (except in genres where the work is meant to be played by everyone, such as classical music, but in that context the idea of cover is also not used anyway).

root/layout/components/sidebar/WorkSidebar.js
reosarevok: The recording length is under "Recording information" on the recording sidebar even though it is almost always just derived from the tracks. IMO this does belong on the sidebar for sure, but if you think we should have a separate section under "Work information", then you can propose some ideas :) I do think the information "the median duration of this work, as far as we know, is this" is about the work, though...

root/layout/components/sidebar/WorkSidebar.js
reosarevok: Fair enough. Expanded the text a bit to indicate that possibility.

lib/MusicBrainz/Server/Data/Work.pm
reosarevok: Did we? I'm honestly not sure, but happy to change if you or @mwiencek remember.

reosarevok:
Would be nice to get this merged. I still think this is 100% work information and belongs there, but open to a different location if there's a clear suggestion that sounds good :)

PR #1285, opened 2019-11-25
MBS-10489: Make EnterEditNote text less redundant

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10489

### DIFF
--- a/root/components/EnterEditNote.js
+++ b/root/components/EnterEditNote.js
@@ -29,9 +29,9 @@ const EnterEditNote = ({
           {exp.l(
             `Entering an {note|edit note} that describes where you got
              your information is highly recommended. Not only does it
-             make it clear where you got your information, but it can also
-             encourage other users to vote on your edit — thus making
-             your edit get applied faster.`,
+             make your sources clear (both now and to users who see the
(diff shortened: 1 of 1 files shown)


root/components/EnterEditNote.js
```diff
@@ -29,9 +29,10 @@ const EnterEditNote = ({
           {exp.l(
             `Entering an {note|edit note} that describes where you got
              your information is highly recommended. Not only does it
-             make it clear where you got your information, but it can also
-             encourage other users to vote on your edit — thus making
-             your edit get applied faster.`,
+             make your sources clear (both now and to users who see
+             the edit years after your edit), but it can also
```
mwiencek: this sounds a bit clearer while using less words
```suggestion
             the edit years later), but it can also
```

reosarevok: True, thanks!

mwiencek: Now it says "years after later" instead of "years later." :)

reosarevok: Sigh, I'm an idiot. *Now* it should be ok.

PR #1288, opened 2019-11-25
MBS-10422: Add ISRCs to release editor recording search

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10422

The recording search in the release editor takes ws/2 data and turns it into ws/js. This makes it so that the ws/2 formatted ISRCs also get converted to the ws/js format, so that we can actually display them in autocomplete.

### DIFF
--- a/root/static/scripts/release-editor/utils.js
+++ b/root/static/scripts/release-editor/utils.js
@@ -137,6 +137,10 @@ utils.cleanWebServiceData = function (data) {
         clean.comment = data.disambiguation;
     }
 
+    if (data.isrcs) {
+        clean.isrcs = data.isrcs.map(cleanIsrc);
+    }
+
     return clean;
 };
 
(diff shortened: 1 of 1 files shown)


root/static/scripts/release-editor/utils.js
```diff
@@ -153,6 +157,12 @@ function cleanArtistCreditName(data) {
     };
 }
 
+function cleanIsrc(data) {
+    return {
+        isrc: data.id,
```
mwiencek: This should just be
```suggestion
        isrc: data,
```
since the JSON format doesn't have any id property like the XML, it's just a list of strings

root/static/scripts/release-editor/utils.js
```diff
@@ -137,6 +137,10 @@ utils.cleanWebServiceData = function (data) {
         clean.comment = data.disambiguation;
     }
 
+    if (data["isrcs"]) {
+        clean.isrcs = _.map(data["isrcs"], cleanIsrc);
```
mwiencek: BTW, you can use `data.isrcs` instead of `data["isrcs"]` here, the brackets/quotes are only needed on sort-name/artist-credit 'cause of the hyphens.

PR #1289, opened 2019-11-27
MBS-10493: Better sizing for edit search selects

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10493

Some edit search multiple selects have *lots* of options, yet their default size is very, very small and they can't be resized.
This makes it so the largest selects (edit type, relationship type, release language and release country) start at size 15 and can be vertically resized, while two smaller selects that still were too big for the default size (edit status and vote type) get a size matching their number of entries to avoid pointless scrolling.

Screenshot from 2019-11-27 12-39-25

### DIFF
--- a/root/edit/search_macros.tt
+++ b/root/edit/search_macros.tt
@@ -56,9 +56,9 @@
              CASE 'MusicBrainz::Server::EditSearch::Predicate::Set';
                SWITCH field.field_name;
                  CASE 'type';
-                   predicate_set(field.field_name, edit_types, field);
+                   predicate_set(field.field_name, edit_types, field, 15, 1);
                  CASE 'status';
-                   predicate_set(field.field_name, status, field);
+                   predicate_set(field.field_name, status, field, 8);
                END;
              CASE 'MusicBrainz::Server::EditSearch::Predicate::ReleaseQuality';
                predicate_set(field.field_name, quality, field);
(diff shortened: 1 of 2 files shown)


reosarevok:
The reason is that all others are actually small enough to fit, so there's no reason to resize them at all AFAICT and it'd be more confusing than useful IMO.

yvanzo:
One reason to resize these all would be to make room for selecting additional conditions by vertically compacting already selected conditions. The added resize handle wouldn’t change the layout but shortening the disabled scroll bar. What would be confusing?

reosarevok:
Oh, I didn't think about making them *smaller*. Maybe! I'll try.

root/edit/search_macros.tt
```diff
@@ -181,11 +181,13 @@
...
 [% END %]
 
-[% MACRO predicate_set(field, set_contents, field_contents) WRAPPER wfield predicate="set" %]
+[% MACRO predicate_set(field, set_contents, field_contents, size) WRAPPER wfield predicate="set" %]
   [% operators([ [ '=', l('is') ]
                  [ '!=', l('is not') ] ], field_contents) %]
 
-  <select name="args" multiple="multiple">
+  <select name="args" multiple="multiple"
+          [% ' size="' _ size _ '"' IF size %]
```
mwiencek: to chomp all extraneous spaces from the HTML
```suggestion
          [%~ ' size="' _ size _ '"' IF size ~%]
```

PR #1292, opened 2019-12-02
Eliminate warnings caused by eslint rule react/jsx-handler-names

joker314:

The eslint rule **react/jsx-handler-names** requires that prop methods which handle events are prefixed so that their purpose is clear (that is, it imposes a simple naming convention on event handlers appearing in JSX). [Specifically](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md), their names must start with "handle" (or in some cases, "on").

This pull request does this in a few different ways:
  - In some cases, such as where the event handler is `console.log`, eslint is simply disabled for that line because the code is in its most readable and understandable form
  - Often, a callback is simply renamed from `x` to `handlerX`
  - In some cases, the callback is renamed to something vastly different to its original name, in an attempt to make it clearer what it does

I have verified that eslint no longer complains about this particular error message, by running
```bash
./script/check_eslint_rule 'react/jsx-handler-names: warn' root/
```
However, I have added one commit (unrelated to lint) afterwards so it may be an outdated result. I will re-run as soon as I can.

### DIFF
--- a/root/static/scripts/edit/components/ArtistCreditEditor.js
+++ b/root/static/scripts/edit/components/ArtistCreditEditor.js
@@ -81,7 +81,7 @@ class ArtistCreditEditor extends React.Component {
     this.copyArtistCredit = this.copyArtistCredit.bind(this);
     this.done = this.done.bind(this);
     this.hide = this.hide.bind(this);
-    this.onNameChange = this.onNameChange.bind(this);
+    this.handleNameChange = this.handleNameChange.bind(this);
     this.pasteArtistCredit = this.pasteArtistCredit.bind(this);
     this.removeName = this.removeName.bind(this);
     this.handleBubbleToggle = this.handleBubbleToggle.bind(this);
@@ -116,7 +116,7 @@ class ArtistCreditEditor extends React.Component {
     });
   }
 
-  onNameChange(i, update) {
+  handleNameChange(i, update) {
     this.setState(state => mutate(state, newState => {
       newState.artistCredit.names[i] =
         {...state.artistCredit.names[i], ...update};
@@ -234,7 +234,7 @@ class ArtistCreditEditor extends React.Component {
         copyArtistCredit={this.copyArtistCredit}
         done={this.done}
         hide={this.hide}
-        onNameChange={this.onNameChange}
+        onNameChange={this.handleNameChange}
         pasteArtistCredit={this.pasteArtistCredit}
         removeName={this.removeName}
         {...this.props}
--- a/root/static/scripts/edit/components/ArtistCreditEditor.js
+++ b/root/static/scripts/edit/components/ArtistCreditEditor.js
@@ -84,7 +84,7 @@ class ArtistCreditEditor extends React.Component {
     this.handleNameChange = this.handleNameChange.bind(this);
     this.pasteArtistCredit = this.pasteArtistCredit.bind(this);
     this.removeName = this.removeName.bind(this);
-    this.handleToggleBubble = this.handleToggleBubble.bind(this);
+    this.handleBubbleToggle = this.handleBubbleToggle.bind(this);
   }
 
   addName() {
@@ -138,7 +138,7 @@ class ArtistCreditEditor extends React.Component {
     this.setState({artistCredit: {names}});
   }
 
-  handleToggleBubble() {
+  handleBubbleToggle() {
     const $bubble = $('#artist-credit-bubble');
     if ($bubble.is(':visible')) {
       const inst = $bubble.data('componentInst');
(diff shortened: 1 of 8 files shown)


root/static/scripts/edit/components/ArtistCreditEditor.js
```diff
@@ -81,10 +81,10 @@ class ArtistCreditEditor extends React.Component {
     this.copyArtistCredit = this.copyArtistCredit.bind(this);
     this.done = this.done.bind(this);
     this.hide = this.hide.bind(this);
-    this.onNameChange = this.onNameChange.bind(this);
+    this.handleNameChange = this.handleNameChange.bind(this);
     this.pasteArtistCredit = this.pasteArtistCredit.bind(this);
     this.removeName = this.removeName.bind(this);
-    this.toggleBubble = this.toggleBubble.bind(this);
+    this.handleToggleBubble = this.handleToggleBubble.bind(this);
```
mwiencek: `handleBubbleToggle` for max nounage

root/static/scripts/edit/components/RemoveButton.js
```diff
@@ -13,7 +13,7 @@ class RemoveButton extends React.Component {
     return (
       <button
         className="nobutton icon remove-item"
-        onClick={this.props.callback}
+        onClick={this.props.onButtonRemoved}
```
mwiencek: This one's confusing, since no button is being removed. I'd keep it simple and name this `onClick`.

root/static/scripts/edit/externalLinks.js
```diff
@@ -278,7 +278,7 @@ export class ExternalLinksEditor
                 }
                 isOnlyLink={this.state.links.length === 1}
                 key={link.relationship}
-                removeCallback={_.bind(this.removeLink, this, index)}
+                handleRemoveCallback={_.bind(this.removeLink, this, index)}
```
mwiencek: This looks like a wrong use of `handle`, since it's a prop name and not a component method. "handleRemoveCallback" also doesn't make a lot of sense. :) `onRemove` should do fine.

joker314:
Thank you for the feedback @mwiencek ! I've addressed it (so now it should all be handle\<noun\>) but there's a failing Selenium error for a file I changed -- so I'll just quickly fix that and it'll then be ready for review.

joker314:
I'm struggling to get Selenium to run locally: here is a screenshot of what happens when I try
It prompts for the postgres password twice, and then it asks for a password so it can "Create Account"

I set the postgres password by running
```
sudo su -
psql postgres postgres
\password
```
but then I'm not sure what I should type when I'm prompted to "Create [an] account".

I also tried to test using the browser extension Selenium IDE, but the HTML tests won't run unless you're logged in to the development environment. I've managed to host a development environment, but I'm not sure how to create an account with sufficient privileges.

@mwiencek do you have any suggestions?

<hr>

Also, I just filed a commit to try to revert the change I thought was likely causing the issue. It clearly wasn't the correct one, and I don't want to try to revert all of the others unless I can do it locally, so as not to spam the CI.

If I can get Selenium to run in the browser, my hope is that I can see if there's some kind of error message caused by my renaming of some of the relevant functions.

mwiencek:
Sorry you're having trouble running the tests. :\ If you're only using PostgreSQL for development, I'd suggest removing any password from the postgres user and making sure you have "local all all trust" as the first/only line in your pg_hba.conf file. That should clear up any auth issues. (Otherwise, you have to make sure a password is set for the SYSTEM database in lib/DBDefs.pm.)

root/static/scripts/edit/components/ArtistCreditNameEditor.js
```diff
@@ -52,7 +52,7 @@ class ArtistCreditNameEditor extends React.Component {
     this.props.onChange({name: event.target.value});
   }
 
-  onJoinPhraseBlur(event) {
+  hanldeJoinPhraseBlur(event) {
```
mwiencek: This looks like it might be the source of some or all of the test failures: "hanlde"

joker314: Thank you so much for catching that!

root/static/scripts/tests/autocomplete2.js
```diff
@@ -66,7 +66,7 @@ $(function () {
             entityType={entityType}
             id="entity-test"
             key={entityType + '-autocomplete'}
-            onChange={console.log}
+            onChange={console.log} // eslint-disable-line react/jsx-handler-names
```
reosarevok: Small nitpick, but this (and the one below) are a bit over the max-len constraint of 78 characters. I'd rather put them above the line with eslint-disable-next-line instead, if you don't mind :)

joker314: Good spot! Fixed with https://github.com/metabrainz/musicbrainz-server/pull/1292/commits/40f49b0647c06281103c254cfaadaaee0e2bba44

mwiencek:
Nice, thanks again! :)

PR #1293, opened 2019-12-02
REFACTOR: Convert var or let to const

VladimirMikulic:
Changes:
- let converted to const when appropriate across multiple files

Head.js(let pager) & server.js(let timeout) files were left with let, because of code readability.

### DIFF
--- a/root/layout/components/Head.js
+++ b/root/layout/components/Head.js
@@ -33,7 +33,8 @@ function canonicalize(url) {
 }
 
 function getTitle(props) {
-  let {title, pager} = props;
+  const pager = props.pager;
+  let title = props.title;
 
   if (!props.homepage) {
     const parts = [];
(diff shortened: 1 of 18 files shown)


mwiencek:
Thanks! I think we can also fix the two left in Head.js & server.js though. :)

For Head.js, just use separate statements:
```JS
const pager = props.pager;
let title = props.title;
```

For server.js, I'd reorder the `disconnect` and `setTimeout` calls for clarity, then use `const`:

```JS
  const cleanup = Raven.wrap(function (signal) {
    const timeout = setTimeout(() => {
      for (const id in cluster.workers) {
        killWorker(cluster.workers[id]);
      }
      process.exit();
    }, DISCONNECT_TIMEOUT);

    cluster.disconnect(function () {
      clearTimeout(timeout);
      process.exit();
    });
  });
```

This is fine in any case because the timeout won't be hit until well after the `disconnect` call executes.

mwiencek:
Nice work! Thank you for your contribution. :)

PR #1294, opened 2019-12-03
Fix JS to follow the no-cond-assign eslint rule

ABCbum:

This pull request aims to make JS files in `/root` directory follows [ESLint](https://eslint.org/) rule [no-cond-assign](https://eslint.org/docs/rules/no-cond-assign). 

**Description:** 
If the assignment is intentional then wrap it with parentheses to signals readers that assigning a new value to the variable (using =) was intended, rather than it being an equality comparison (which should use == or ===) that was wrongly written.

For example: 
`while(foo = bar) // Intentional assignment` 
becomes 
`while((foo = bar)) // Wrap it with parentheses`.

### DIFF
--- a/root/static/scripts/timeline.js
+++ b/root/static/scripts/timeline.js
@@ -212,13 +212,26 @@ class TimelineViewModel {
 
         _.forEach(parts, function (part) {
             var match;
-            if (match = part.match(/^(-)?([rv])-?$/)) { // trailing - for backwards-compatibility
+
+            /** Please correct me if i'm misunderstanding this
+             * if(match = part.match(RegExp) {
+             *    // Do sth!!
+             * }
+             * 
+             * is equivalent to
+             * 
+             * if(part.match(RegExp)) {
+             *    match = part.match(RegExp);
+             *    // Do sth!!
+             * }
+             *  */ 
+            if ((match = part.match(/^(-)?([rv])-?$/))) { // trailing - for backwards-compatibility
                 var meth = match[2] === 'r' ? 'rate' : 'events';
                 self.options[meth](!(match[1] === '-'));
-            } else if (match = part.match(/^(-)?(c-.*)$/)) {
+            } else if ((match = part.match(/^(-)?(c-.*)$/))) {
                 var category = _.find(self.categories(), { hashIdentifier: match[2] });
                 if (category) { category.enabled(!(match[1] === '-')) }
-            } else if (match = part.match(/^g\/.*$/)) {
+            } else if ((match = part.match(/^g\/.*$/))) {
                 self.zoomHashPart(part);
             } else {
                 match = part.match(/^(-)?(.*)$/);
--- a/root/static/scripts/timeline.js
+++ b/root/static/scripts/timeline.js
@@ -213,18 +213,6 @@ class TimelineViewModel {
         _.forEach(parts, function (part) {
             var match;
 
-            /** Please correct me if i'm misunderstanding this
-             * if(match = part.match(RegExp) {
-             *    // Do sth!!
-             * }
-             * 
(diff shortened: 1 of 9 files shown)


root/static/scripts/timeline.js
```diff
@@ -212,13 +212,26 @@ class TimelineViewModel {
...
+             *    // Do sth!!
+             * }
+             * 
+             * is equivalent to
+             * 
+             * if(part.match(RegExp)) {
+             *    match = part.match(RegExp);
+             *    // Do sth!!
+             * }
+             *  */
```
ruinername: Yes, but it is not good practice https://stackoverflow.com/questions/2576571/assign-variable-in-if-condition-statement-good-practice-or-not

root/static/scripts/timeline.js
```diff
@@ -212,13 +212,26 @@ class TimelineViewModel {
...
+            /** Please correct me if i'm misunderstanding this
+             * if(match = part.match(RegExp) {
+             *    // Do sth!!
+             * }
+             * 
+             * is equivalent to
+             * 
+             * if(part.match(RegExp)) {
+             *    match = part.match(RegExp);
+             *    // Do sth!!
```
mwiencek: Not quite. It's equivalent to:
```JS
match = part.match(RegExp);
if (match) {
  // Do something
}
```
The difference being `part.match` is only executed once, not twice (which doesn't matter in this example, but is very important if the function has side effects).

root/static/scripts/timeline.js
```diff
@@ -212,13 +212,26 @@ class TimelineViewModel {
 
         _.forEach(parts, function (part) {
             var match;
-            if (match = part.match(/^(-)?([rv])-?$/)) { // trailing - for backwards-compatibility
+
+            /** Please correct me if i'm misunderstanding this
+             * if(match = part.match(RegExp) {
+             *    // Do sth!!
+             * }
```
mwiencek: You can remove this entire comment now, hopefully we've clarified things here. :) (Next time feel free to ask questions like this on the pull request instead of adding them as code comments.)

ABCbum: Oh thanksss i will ask through the pull request next time :D <3

root/static/scripts/timeline.js
```diff
@@ -212,13 +212,14 @@ class TimelineViewModel {
 
         _.forEach(parts, function (part) {
             var match;
-            if (match = part.match(/^(-)?([rv])-?$/)) { // trailing - for backwards-compatibility
+
```
reosarevok: Super nitpicky, but: this is adding an extra line, is it intentional?

The next line is too long (it already was, not your fault) - maybe it would make sense to move the comment to this line instead?

@mwiencek: opinion? :)

ABCbum: Hi, so i should remove the extra line (sorry i have a habit to add a line after variable declaration) and move the comment to the next line right?

ABCbum: Hi, yes the extra line was intentional since for me that makes the code just a little more readable. Should i remove it?

reosarevok: Let's see what mwiencek says :)

ABCbum: Okay :) maybe you and i have a different "code taste"

reosarevok: I mean, I don't actually dislike it or anything, just want to make sure we're being consistent.

ABCbum: I get it :D

mwiencek: I don't have a problem with the extra line. :)

mwiencek:
Excellent work, much appreciated!

PR #1295, opened 2019-12-03
Eliminate unused variables

ruinername:
Related to this ticket: https://tickets.metabrainz.org/browse/MBS-10504


reosarevok:
Hmm, just realized we're on beta 11, not 10 - so this downgrade might make some things worse. That said, they don't seem to have a beta release yet that solves this issue though. @mwiencek: could you take a look at this and give your opinion?

reosarevok:
(ignore the failing selenium test, btw - it's flaky and needs fixing)

mwiencek:
The v11 beta is or was needed to parse Flow's inexact object syntax correctly (see https://github.com/metabrainz/musicbrainz-server/pull/1079). Unfortunately this broke the no-unused-vars rule, which they've fixed in 10.0.3, but they haven't yet merged the v10 updates back into v11.

I don't think downgrading is the right solution, since that might break parsing files with `{...}` altogether. I left a polite ping on an issue in their repo asking if they can backport the changes soon, but if not, we could very easily just set up our own fork until they do.

ruinername:
This is commit with fix before 10.0.3 release. Maybe it will be useful if you will fork https://github.com/babel/babel-eslint/commit/354953da5347a005fe566cc7dd338974aab50908

ruinername:
Ohhh... My bad that I worked in master

root/tag/TagIndex.js
```diff
@@ -32,6 +32,7 @@ type Props = {
   },
 };
 
+// eslint-disable-next-line no-unused-vars
```
mwiencek: You can remove `<T>` below, then delete this comment.

root/static/scripts/tests/react-macros.js
```diff
@@ -6,6 +6,8 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+/* eslint no-unused-vars: 0 */
```
mwiencek:
```suggestion
/* eslint-disable no-unused-vars */
```

root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -70,7 +70,6 @@ const MpIcon = hydrate<MpIconProps>('span.ac-mp', (
 const ArtistCreditLink = ({
   artistCredit,
   showDeleted = true,
-  showEditsPending = true,
```
mwiencek: You should be able to remove it from `type Props` above too now.

root/vars.js
```diff
@@ -25,7 +27,6 @@ declare var N_ln: (skey: string, pkey: string) => (val: number) => string;
 declare var N_lp: (key: string, context: string) => () => string;
 
 declare var l_attributes: typeof l;
-declare var ln_attributes: typeof ln;
```
mwiencek: This line should be added back too. :)

root/types.js
```diff
@@ -15,6 +15,8 @@
  * how data is serialized for us.
  */
 
+/* eslint no-unused-vars: 0 */
```
mwiencek: Nitpick:
```suggestion
/* eslint-disable no-unused-vars */
```

PR #1296, opened 2019-12-03
REFACTOR: Enforce eslint curly brace style convention

VladimirMikulic:


package-lock.json
```diff
@@ -0,0 +1,8642 @@
+{
```
mwiencek: Whoops, you accidentally added a file here (maybe it's from using `npm` instead of `yarn`?). Can you remove this and update the PR?

root/static/scripts/timeline.js
```diff
@@ -111,11 +111,13 @@ class TimelineViewModel {
         }, 1000);
 
         self.waitToGraph = ko.computed(function () {
-            if (_.some(self.enabledCategories(), function (c) { return c.hasLoadingLines() }))
+            if (_.some(self.enabledCategories(), function (c) { return c.hasLoadingLines() })) {
                 return true;
+            }
 
-            if (self.options.events() && !self.loadedEvents())
+            if (self.options.events() && !self.loadedEvents()){
```
mwiencek: We're missing a space before the `{` here:
```suggestion
            if (self.options.events() && !self.loadedEvents()) {
```

VladimirMikulic:
> Looks almost perfect, but need to remove package-lock.json from here and add a missing space. :)

Thanks for pointing that out. My bad, I accidentally committed package-lock.json.
I should really switch to yarn :)

mwiencek:
Thank you for the fixes! :)

PR #1300, opened 2019-12-04
Fix issues relating to eslint rule no-negated-condition

joker314:
*Submitted as part of the Google Code In competition*

This pull request makes stylistic changes only. Where there is a negated condition, like the one below:
```javascript
!boolExpr ? "a" : "b"
```
it is rewritten to not be negated:
```javascript
boolExpr ? "b" : "a"
```

This is to comply with the eslint rule *no-negated-condition*, to make it easier to reason about branching.

<hr>

This pull request is not intended to change any program behaviour, so all I did was negate any conditionals.

### DIFF
--- a/root/layout/components/Head.js
+++ b/root/layout/components/Head.js
@@ -127,14 +127,14 @@ const Head = ({$c, ...props}: HeadProps) => (
 
     {manifest.js('jed-data')}
 
-    {$c.stash.current_language !== 'en' ? (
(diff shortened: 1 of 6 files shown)


reosarevok:
This looks fine, thanks! (but you might want to fix the typo on the commit name :) )

joker314:
@reosarevok good spot, thanks for changing the title! I've edited the commit message and have force-pushed to make the replacement. I didn't change any code since your initial reviews.

PR #1301, opened 2019-12-05
REFACTOR: Make line breaks inside parentheses more consistent

ABCbum:

This pull request aims to make js files in `/root` directory follows [ESLint](https://eslint.org/) rule [function-paren-newline](https://eslint.org/docs/rules/function-paren-newline) (with 'consistent' option).  

**Description** 
This rule enforces consistent line breaks inside parentheses of function parameters or arguments.

For example: 
 ```
foo(
  bar,
  baz);
```
becomes 
 ```
foo(
  bar,
  baz
);
```

### DIFF
--- a/root/static/scripts/edit/MB/CoverArt.js
+++ b/root/static/scripts/edit/MB/CoverArt.js
@@ -410,7 +410,11 @@ MB.CoverArt.FileUpload = function (file) {
                     self.updateProgress(2, 100);
 
                     var submitting = MB.CoverArt.submit_edit(
-                        self, postfields, mime_type, position);
+                        self,
+                        postfields, 
+                        mime_type, 
+                        position,
+                    );
 
                     submitting.fail(function (msg) {
                         self.status(statuses.submit_error);
@@ -593,7 +597,10 @@ MB.CoverArt.add_cover_art = function (gid) {
             if (mime_type)
             {
                 $('iframe')[0].contentWindow.upload(
-                    gid, $('#id-add-cover-art\\.id').val(), mime_type);
+                    gid, 
+                    $('#id-add-cover-art\\.id').val(),
+                    mime_type,
+                );
             }
             else
             {
--- a/root/static/scripts/edit/forms.js
+++ b/root/static/scripts/edit/forms.js
@@ -60,8 +60,12 @@ MB.forms = {
 
     linkTypeOptions: function (root, backward) {
         function getText(data) {
-            return stripAttributes(data, l_relationships(
-                backward ? data.reverse_link_phrase : data.link_phrase));
+            return stripAttributes(
+                data, 
+                l_relationships(
+                    backward ? data.reverse_link_phrase : data.link_phrase,
+                ),
+            );
         }
 
         var options = MB.forms.buildOptionsTree(root, getText, 'id');
--- a/root/static/scripts/timeline.js
+++ b/root/static/scripts/timeline.js
@@ -128,7 +128,11 @@ class TimelineViewModel {
                     var rateBounds = line.calculateRateBounds(
                         line.rateData().data,
                         line.rateData().thresholds,
-                        {min: self.zoom.xaxis.min(), max: self.zoom.xaxis.max()});
+                        {
+                          min: self.zoom.xaxis.min(),
+                          max: self.zoom.xaxis.max(),
+                        },
+                    );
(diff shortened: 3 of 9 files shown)


reosarevok:
Hi! One thing I can already see - a lot of the lines are now likely to need a [comma-dangle](https://eslint.org/docs/rules/comma-dangle) :)

mwiencek:
This looks good to me except for the missing comma-dangles as @reosarevok pointed out. :) Take a look at those when you have a moment.

ABCbum:
Hi @reosarevok  and @mwiencek!

I took a look at [comma-dangle](https://eslint.org/docs/rules/comma-dangle) and it says "require or disallow trailing commas" (a little confusing) but from my work i assume that i probably should add trailing commas for more clear diff when an item is added or removed from an object or array, right?

reosarevok:
Yes, in our case the choice is to require rather than disallow :)

ABCbum:
Hi! 

In file /edit/MB/CoverArt should i add a new line after each comma so that it is consistent with other files ?

root/static/scripts/edit/MB/CoverArt.js
```diff
@@ -410,7 +410,8 @@ MB.CoverArt.FileUpload = function (file) {
                     self.updateProgress(2, 100);
 
                     var submitting = MB.CoverArt.submit_edit(
-                        self, postfields, mime_type, position);
+                        self, postfields, mime_type, position
```
ABCbum: Should i add new line after each comma here?

ABCbum: So that it is consistent with other files changed.

mwiencek: Well, I would. :) I don't think we have a rule for it, but I find it easier to read that way.

ABCbum: okay thanks

root/static/scripts/timeline.js
```diff
@@ -128,7 +128,8 @@ class TimelineViewModel {
                     var rateBounds = line.calculateRateBounds(
                         line.rateData().data,
                         line.rateData().thresholds,
-                        {min: self.zoom.xaxis.min(), max: self.zoom.xaxis.max()});
+                        {min: self.zoom.xaxis.min(), max: self.zoom.xaxis.max()}
```
ABCbum: Here should i make the object a little more clear? 
```suggestion
                        {
                            min: self.zoom.xaxis.min(), 
                            max: self.zoom.xaxis.max()
                          }
```

mwiencek:
This looks good to me except for the last commit - let's keep the trailing commas in object literals too (when they're multiline).

root/static/scripts/edit/forms.js
```diff
@@ -60,8 +60,12 @@ MB.forms = {
 
     linkTypeOptions: function (root, backward) {
         function getText(data) {
-            return stripAttributes(data, l_relationships(
-                backward ? data.reverse_link_phrase : data.link_phrase));
+            return stripAttributes(
+                data, 
+                l_relationships(
+                  backward ? data.reverse_link_phrase : data.link_phrase
```
mwiencek: We generally indent 2 spaces but this particular file is 4, so. :) I also suggested adding in the trailing comma:
```suggestion
                    backward ? data.reverse_link_phrase : data.link_phrase,
```

ABCbum: Okay this one is my bad my setting is 2 spaces indent so i didn't notice. But when i keep the trailing commas in object literals the selenium tests do not pass and i tested with chrome dev tools that if i add trailing comma it says missing '('.

mwiencek: We use trailing commas in object literals in tons of places, so that can't be right. :) I checked https://circleci.com/gh/metabrainz/musicbrainz-server/7375 and don't see any error about a missing "(" (unless I'm missing something too). Chrome definitely supports it, too, so that must've been an unrelated syntax issue.

ABCbum: Yeah i'm so confused i know that trailing commas is supported in object literals but removing it makes the tests pass somehow i think i made a mistake somewhere else so let me doublecheck it. :DD

ABCbum:
I think it is ready now

PR #1304, opened 2019-12-06
Fix: not have empty block statements (ESLint 'no-empty' rule)

ruinername:


ruinername:
I have one question about rev-manifest. Should it be .json instead .js?

root/static/scripts/guess-case/MB/GuessCase/Handler/Track.js
```diff
@@ -91,10 +91,10 @@ MB.GuessCase.Handler.Track = function (gc) {
...
      */
     self.doWord = function () {
-        if (self.doIgnoreWords()) {
-        } else if (self.doFeaturingArtistStyle()) {
-        } else if (gc.mode.doWord()) {
-        } else {
+        if (self.doIgnoreWords());
+        else if (self.doFeaturingArtistStyle());
+        else if (gc.mode.doWord());
+        else {
```
mwiencek: Here you can combine all three conditions into the first `if`, or assign the result to a `handled` variable like above and do `if (!handled)` for the rest of the matching below.

root/static/scripts/guess-case/MB/GuessCase/Handler/Track.js
```diff
@@ -91,10 +91,7 @@ MB.GuessCase.Handler.Track = function (gc) {
      * - Handles FeaturingArtistStyle
      */
     self.doWord = function () {
-        if (self.doIgnoreWords()) {
-        } else if (self.doFeaturingArtistStyle()) {
-        } else if (gc.mode.doWord()) {
-        } else {
+        if (!self.doIgnoreWords() && !self.doFeaturingArtistStyle() && !gc.mode.doWord()) {
```
reosarevok: Nitpick: this line is too long, it should be split :)

mwiencek:
Thank you!

PR #1305, opened 2019-12-06
MBS-10365: Convert AreaList to React Table

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10365

Pretty much all @yvanzo's code, just fixed one bug I found. Putting it up since it seems good enough to merge to me as a second step on react-table work.

### DIFF
--- a/root/components/list/AreaList.js
+++ b/root/components/list/AreaList.js
@@ -9,11 +9,13 @@
 
 import React from 'react';
 
+import Table from '../Table';
 import {withCatalystContext} from '../../context';
-import loopParity from '../../utility/loopParity';
-import DescriptiveLink
-  from '../../static/scripts/common/components/DescriptiveLink';
-import SortableTableHeader from '../SortableTableHeader';
+import {
+  defineCheckboxColumn,
+  defineNameColumn,
+  defineTypeColumn,
+} from '../../utility/tableColumns';
 
 type Props = {
   +$c: CatalystContextT,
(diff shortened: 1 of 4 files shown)


root/user/UserCollections.js
```diff
@@ -88,7 +88,9 @@ const CollectionsEntityTypeSection = ({
   user,
 }) => {
   const viewingOwnProfile = !!$c.user && $c.user.id === user.id;
-  const nameColumn = defineNameColumn(l('Collection'));
+  const nameColumn =
+    defineNameColumn<CollectionT>(false, '', false, l('Collection'));
```
yvanzo: `defineNameColumn` should have optional arguments with default values to avoid such kludge I did.

root/utility/tableColumns.js
```diff
@@ -10,27 +10,82 @@
...
+  sortable: ?boolean,
   title: string,
-): ColumnOptions<CoreEntityT | CollectionT, string> {
+): ColumnOptions<T, string> {
   return {
-    Cell: ({row: {original}}) => (
-      <EntityLink entity={original} />
+    Cell: ({row: {original}}) => (descriptive
+      // $FlowFixMe
+      ? <DescriptiveLink entity={original} />
```
yvanzo: This is due to `descriptive` argument being redundant with flow type. Maybe there should be a separate function for `CollectionT`, or maybe `DescriptiveLink` can be used with collection instead of `EntityLink` below.

root/utility/tableColumns.js
```diff
@@ -10,27 +10,82 @@
...
+      ? (
+        <SortableTableHeader
+          label={l('Type')}
+          name="type"
+          order={order}
+        />
+      )
+      : l('Type')),
+    // $FlowFixMe
+    accessor: 'typeName',
```
yvanzo: Issue is `CoreEntityT` should probably be replaced with something else here as it doesn’t have a `typeName`.

root/components/list/AreaList.js
```diff
@@ -29,63 +31,24 @@ const AreaList = ({
...
+  const typeColumn = defineTypeColumn('area_type', order, sortable);
+
+  const columns = React.useMemo(
+    () => [
+      ...(checkboxColumn ? [checkboxColumn] : []),
+      nameColumn,
+      typeColumn,
+    ],
+    [],
+  );
```
mwiencek: Oh, this kinda defeats the point of `React.useMemo`. The one in root/user/UserCollections.js is wrong too, though I overlooked that.

It's supposed to be for avoiding recalculating things on every render, so to make this right, we should move all the column declarations inside the `useMemo` callback. Then, we should fix the dependencies array (the second argument to `useMemo`) to list all the variables used to create the columns... i.e. `[$c.user_exists, checkboxes, order, sortable]`.

React will then check if any of the values in the dependencies array changed to decide if it needs to recalculate the columns.

PR #1313, opened 2019-12-08
Code style: eslint consistent-return (2)

joker314:
The eslint rule [consistent-return](https://eslint.org/docs/rules/consistent-return) asks that all functions should either:
- Always return a value explicitly (using `return value;` and even `return undefined;`) or
- Never return a value explicitly (using only `return;` or simply by omission)

This means that a function which sometimes returns a value and sometimes doesn't (e.g. it depends on an `if` statement) will be in violation of this rule.

<hr>

I would say that it is possible to categorise when a rewrite in a function occurred as follows:
- *Event listener*: Boolean return value decides whether the default action should be taken by the browser. **Solution: explicitly return the 'default' Boolean)**
- *Undefined indicates that the result should be discarded*. This would often occur in a `_.map` which would be fed into a `_.compact`. **Solution: return `null` to show it's intentional**
- *Unintentional*. If code expects that it is either in state X or in state Y, then `if (X) return 1; if (B) return 2;` should be sufficient. Ideally, this would be rewritten into an `else` construct. If we are not in state A or B, we should throw an error rather than silently return `undefined`. **See review comments**

### DIFF
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -589,6 +589,8 @@ $.widget("ui.menu", $.ui.menu, {
              */
             event.stopPropagation();
             event.preventDefault();
+
+            return true;
         }
     },
 
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -554,6 +554,7 @@ $.widget("mb.entitylookup", $.ui.autocomplete, {
         if (arguments.length) {
             recentEntities[entityType] = _.take(arguments[0], MAX_RECENT_ENTITIES);
             localStorage("recentAutocompleteEntities", JSON.stringify(recentEntities));
+            return undefined;
         } else {
             return recentEntities[entityType] || [];
         }
@@ -591,6 +592,8 @@ $.widget("ui.menu", $.ui.menu, {
 
             return true;
         }
+
+        return false;
     },
 
     _create: function () {
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -592,6 +592,8 @@ $.widget("ui.menu", $.ui.menu, {
 
             return true;
         }
+
+        return undefined;
     },
 
     _create: function () {
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -1050,6 +1050,8 @@ MB.Control.EntityAutocomplete = function (options) {
                 options.entity = entity;
                 return true;
             }
+
+            return false;
         });
     }
 
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -589,8 +589,6 @@ $.widget("ui.menu", $.ui.menu, {
              */
             event.stopPropagation();
             event.preventDefault();
-
-            return true;
         }
     },
 
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -590,10 +590,10 @@ $.widget("ui.menu", $.ui.menu, {
             event.stopPropagation();
             event.preventDefault();
 
-            return true;
+            return false;
         }
 
-        return undefined;
+        return true;
     },
 
     _create: function () {
@@ -602,7 +602,7 @@ $.widget("ui.menu", $.ui.menu, {
     },
 
     select: function (event) {
-        if (!this._selectAction(event)) {
+        if (this._selectAction(event)) {
             this._super(event);
         }
         /*
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -592,8 +592,6 @@ $.widget("ui.menu", $.ui.menu, {
 
             return true;
         }
-
-        return false;
     },
 
     _create: function () {
--- a/root/static/scripts/edit/MB/Control/Bubble.js
+++ b/root/static/scripts/edit/MB/Control/Bubble.js
@@ -271,7 +271,7 @@ function bubbleControlHandler(event) {
                 bubble.hide(false);
             }
         }
-        return;
+        return true;
     }
 
     var isButton = $(control).is(":button");
--- a/root/static/scripts/edit/MB/Control/Bubble.js
+++ b/root/static/scripts/edit/MB/Control/Bubble.js
@@ -273,7 +273,7 @@ function bubbleControlHandler(event) {
                 bubble.hide(false);
             }
         }
-        return true;
+        return undefined;
     }
 
     var isButton = $(control).is(":button");
--- a/root/static/scripts/edit/confirmNavigationFallback.js
+++ b/root/static/scripts/edit/confirmNavigationFallback.js
@@ -41,11 +41,15 @@ MB.confirmNavigationFallback = function () {
                 return false;
             }
         }
+
+        return true;
     };
 
     document.onkeypress = function () {
         if (prevented) {
             return (prevented = false);
         }
+
+        return true;
     };
 };
--- a/root/static/scripts/edit/confirmNavigationFallback.js
+++ b/root/static/scripts/edit/confirmNavigationFallback.js
@@ -42,7 +42,7 @@ MB.confirmNavigationFallback = function () {
             }
         }
 
-        return true;
+        return undefined;
     };
 
     document.onkeypress = function () {
@@ -50,6 +50,6 @@ MB.confirmNavigationFallback = function () {
             return (prevented = false);
         }
 
-        return true;
+        return undefined;
     };
 };
--- a/root/static/scripts/edit/utility/guessFeat.js
+++ b/root/static/scripts/edit/utility/guessFeat.js
@@ -129,6 +129,7 @@ function bestArtistMatch(artists, name) {
             if (similarity >= MIN_NAME_SIMILARITY) {
                 return {similarity: similarity, artist: a, name: name};
             }
+            return null;
         })
         .compact()
         .sortBy('similarity')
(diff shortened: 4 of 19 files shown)


root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -554,6 +554,7 @@ $.widget("mb.entitylookup", $.ui.autocomplete, {
         if (arguments.length) {
             recentEntities[entityType] = _.take(arguments[0], MAX_RECENT_ENTITIES);
             localStorage("recentAutocompleteEntities", JSON.stringify(recentEntities));
+            return undefined;
```
joker314: This acts as either a setter or a getter depending on how many parameters it is getting. It's nicer to think of it really as two functions with the same name. This means that there was never an inconsistency, but I think explicitly returning `undefined` is the best way to show this. (Comment **A**)

root/static/scripts/edit/utility/guessFeat.js
```diff
@@ -216,6 +217,7 @@ MB.Control.initGuessFeatButton = function (formName) {
             name: function () {
                 if (arguments.length) {
                     nameInput.value = arguments[0];
+                    return undefined;
```
joker314: See comment marked **A**.

root/static/scripts/relationship-editor/common/entity.js
```diff
@@ -29,6 +29,8 @@ function getDirection(relationship, source) {
   if (source === entities[1]) {
```
joker314: Is it ever possible to be neither "forwards" nor "backwards"? If not, we should really be throwing an error rather than returning `undefined` or `null`. This is comment **B**

root/static/scripts/relationship-editor/common/fields.js
```diff
@@ -541,6 +541,8 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
             if (entity === entities[1]) {
```
joker314: Same question as **B**

joker314:
Selenium error is a result of flaky tests: the 'breaking' change was just adding a semicolon (resolving conflicts from master)

root/static/scripts/relationship-editor/common/entity.js
mwiencek: `source` should always be either entity0 or entity1, so I'd probably throw an error here at the end, yes.

root/static/scripts/relationship-editor/common/fields.js
mwiencek: Same here, I'd say it's an error if this function ever receives an entity not in the `entities` array.

root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -588,8 +589,6 @@ $.widget("ui.menu", $.ui.menu, {
              */
             event.stopPropagation();
             event.preventDefault();
-
-            return true;
```
mwiencek: There's a case below where we check `if (!this._selectAction(event)) {`, so this seems off and we should rather return false at the end.

root/static/scripts/edit/MB/Control/Bubble.js
```diff
@@ -273,7 +273,7 @@ function bubbleControlHandler(event) {
                 bubble.hide(false);
             }
         }
-        return;
+        return true;
```
mwiencek: For event handlers I think `return undefined` makes more sense, since `false` is the only value that isn't ignored. `return true` could be slightly misleading since it might imply to a reader that a `true` value is used somehow.

root/static/scripts/edit/confirmNavigationFallback.js
```diff
@@ -41,11 +41,15 @@ MB.confirmNavigationFallback = function () {
                 return false;
             }
         }
+
+        return true;
```
mwiencek:
```suggestion
        return undefined;
```

root/static/scripts/edit/confirmNavigationFallback.js
```diff
@@ -41,11 +41,15 @@ MB.confirmNavigationFallback = function () {
...
     };
 
     document.onkeypress = function () {
         if (prevented) {
             return (prevented = false);
         }
+
+        return true;
```
mwiencek:
```suggestion
        return undefined;
```

root/static/scripts/guess-case/MB/GuessCase/Handler/Artist.js
```diff
@@ -165,6 +165,8 @@ MB.GuessCase.Handler.Artist = function (gc) {
 
                 return utils.trim(_.compact(names).join(" ") + (append || ""));
             }
+
+            return null;
```
mwiencek: I believe the result of this function is string-joined by `sortCompoundName`.  It's true that `String.prototype.join` converts `null` to `''` when joining, but I think `return '';` would be a bit clearer.

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -313,7 +314,7 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
 
         keydownEvent(data, event) {
             if (event.isDefaultPrevented()) {
-                return;
+                return false;
```
mwiencek: This might be an exception to my event handler comments above, since the knockout-delegatedEvents plugin actually does use the `true` value: https://github.com/rniemeyer/knockout-delegatedEvents/blob/309b1877b3bd33bbfb20a9cf197e94e7e777a644/src/knockout-delegatedEvents.js#L158

So returning false is probably fine here, since this plugin is confusing anyway. :) (And we'll be changing all this code when it's converted to React.)

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -800,6 +805,8 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
                 return id;
             }
         }
+
+        return null;
```
mwiencek: `null` *seems* fine, but since our relationship editor code doesn't have Flow types, I'm afraid of possibly breaking conditions that use `===`. So I'd keep this as `undefined` for safety.

```suggestion
        return undefined;
```

root/static/scripts/relationship-editor/common/entity.js
```diff
@@ -29,6 +29,8 @@ function getDirection(relationship, source) {
...
     return 'backward';
   }
+
+  throw 'source should be in the entities array (which has length 2)';
```
mwiencek: shortening this to something very minimal should be fine, since it should never happen (and these strings can't be minified for production, so having long strings in our js bundles that aren't used isn't ideal)
```suggestion
  throw 'source not in the entities array';
```

root/static/scripts/relationship-editor/common/fields.js
```diff
@@ -541,6 +541,8 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
...
                 return this.entity1_credit;
             }
+
+            throw 'entity should be in the entities array (which has length 2)';
```
mwiencek:
```suggestion
            throw 'entity not in the entities array';
```

root/static/scripts/relationship-editor/release.js
```diff
@@ -68,7 +68,7 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
 
             window.addEventListener('beforeunload', function (event) {
                 if (self.redirecting) {
-                    return;
+                    return true;
```
mwiencek:
```suggestion
                    return undefined;
```

root/static/scripts/relationship-editor/release.js
```diff
@@ -77,6 +77,8 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
                     event.returnValue = l("All of your changes will be lost if you leave this page.");
                     return event.returnValue;
                 }
+
+                return true;
```
mwiencek:
```suggestion
                return undefined;
```

root/static/scripts/release-editor/trackParser.js
```diff
@@ -451,15 +451,19 @@ const trackParser = releaseEditor.trackParser = {
     },
 
     matchDataWithTrack: function (data, track) {
+        // The result of this function will be fed into _.compact so that
+        // null and undefined return values will be stripped.
```
mwiencek: We have another rule where multi-line comments should be formatted like
```JS
/*
 * The result of this function will be fed into _.compact so that
 * null and undefined return values will be stripped.
 */
```

joker314:
Hi @reosarevok ! I've just pushed some changes which address your feedback. I've now remembered why I (albeit incorrectly) removed the `return true;` in the autocorrecter, and it is in part why I would advocate for Boolean return values in event listeners despite this not being a language feature. (I don't really have strong opinions about it though).

Effectively, I think there are two "conflicting" lines:

https://github.com/metabrainz/musicbrainz-server/blob/a570d0ed9aaef0db4fe063488e5a016861f6eb10/root/static/scripts/common/MB/Control/Autocomplete.js#L598

https://github.com/metabrainz/musicbrainz-server/blob/a570d0ed9aaef0db4fe063488e5a016861f6eb10/root/static/scripts/common/MB/Control/Autocomplete.js#L602

The first one doesn't like `false` (because it's basically `preventDefault`); but the latter doesn't like `true`. This, as I understand it, means that we have to return `undefined`.

I have verified by `git grep` that `_selectAction` only appears in this file, and we've already discussed all of the instances. I would suggest inverting the `if` statement (second embedded code block above) so that it no longer has a negated predicate, and then we can `return false;` safely at the end of the function.

joker314:
The line I initially removed, the one which was `return true;`, I feel like this can all be solved if I change it to `return false;`, then invert the `if` statement.

It is intentionally trying to prevent default etc., and `_on` wants a `return false` when that happens, so this seems to make perfect sense!

mwiencek:
> Hi @reosarevok ! I've just pushed some changes which address your feedback

I'm @mwiencek but thanks. :D

> It is intentionally trying to prevent default etc., and _on wants a return false when that happens, so this seems to make perfect sense!

I'll check it again locally to be sure but this looks good to me. :)

joker314:
@mwiencek Sorry for getting your name wrong earlier!

Sounds good. Is there a way to tell if the Selenium test is a false positive? I haven't managed to get the postgresql database set up quite right yet to test locally myself.

mwiencek:
> Is there a way to tell if the Selenium test is a false positive?

Not really other than remembering which ones fail the most often (don't remember seeing this one). I ran them locally and it worked, so asked CircleCI to run them again to make sure.

PR #1314, opened 2019-12-08
fixed one-var rule

monicaq21:
The one-var rule has been fixed, so that each variable has its own separate var, let, or const declaration. All additional indent and Multiline-comment-style problems of those files are also fixed.


ruinername:
Think you should remove ESLint update from this PR and create another PR with it

monicaq21:
Thank you for the feedback. I have done so and added a commit. Do I have to create a new PR?

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
joker314: I think your IDE created this file accidentally by the way

monicaq21: Thank you, that is fixed!

root/static/scripts/edit/confirmNavigationFallback.js
```diff
@@ -9,43 +9,43 @@
...
+   *
+   * Opera < 12.10 fires both keydown and keypress events, but keypress
+   * must return false. Opera >= 12.10 doesn't fire keypress for special
+   * keys, so keydown must return false. Regular event listeners and/or
+   * preventDefault don't work for this, they must be assigned directly
+   * to document.onkeydown and document.onkeypress.
+   */
+  document.onkeydown = function (event) {
+    if (event.keyCode == 8) {
+      let node = event.srcElement || event.target; let tag = node.tagName.toLowerCase();
```
ABCbum: Hi @monicaq21 , 

Just a small detail, i think `tag`'s declaration should be in a new line just like you did in [scripts/edit/ExampleRelationships.js](https://github.com/metabrainz/musicbrainz-server/blob/efc397bd5c4119f1c36dae86ab6f6281530c863e/root/static/scripts/edit/ExampleRelationships.js#L12#L17).

I'm not a maintainer so if i'm wrong just correct me.

root/static/scripts/edit/utility/guessFeat.js
```diff
@@ -21,215 +21,215 @@ import {
...
-            } else {
-                name += remainder;
-                break;
-            }
+  return _.reduce(bracketPairs, function (accum, pair) {
+    let name = '';
+    let joinPhrase = accum.joinPhrase;
+    let credits = accum.artistCredit;
+    let remainder = accum.name;
+    let b; let m;
```
ABCbum: And this also.

root/static/scripts/guess-case/MB/GuessCase/Output.js
```diff
@@ -15,207 +15,207 @@ MB.GuessCase = MB.GuessCase ? MB.GuessCase : {};
...
+  self.capitalizeWordAtIndex = function (index, overrideCaps) {
+    overrideCaps = (overrideCaps == null ? flags.context.forceCaps : overrideCaps);
+    if ((!gc.mode.isSentenceCaps() || overrideCaps) &&
             (!self.isEmpty()) &&
             (self.getWordAtIndex(index) != null)) {
 
-            // Don't capitalize last word before puncuation/end of string in sentence mode.
-            var w = self.getWordAtIndex(index), o = w;
+      // Don't capitalize last word before puncuation/end of string in sentence mode.
+      let w = self.getWordAtIndex(index); let o = w;
```
ABCbum: And this too 
```suggestion
      let w = self.getWordAtIndex(index); 
      let o = w;
```

root/static/scripts/edit/confirmNavigationFallback.js
ABCbum: I think you can find all the similar problems by using **Ctrl + F** or **Command + F** then enter `; let` or `; const` in all the files you have changed. Of course if it is a problem.

reosarevok:
I'm closing this since it's been replaced by https://github.com/metabrainz/musicbrainz-server/pull/1316

PR #1317, opened 2019-12-09
MBS-10511: Regression: Modbot edit references are no longer linkified

mwiencek:
# Fix MBS-10511 in beta branch

This was broken in d2f9ced19a20a45e25b5a7e5beb0b7fdb4b726a4 due to all ModBot notes no longer being passed to `format_editnote`. (We don't need to do that if the note is actually localized, because it should be using our translated string syntax to link/format things in that case.)


mwiencek:
Not sure why, but the PR didn't close when I merged this (using the merge button here). So closing manually.

PR #1318, opened 2019-12-09
MBS-10497: Limit the number of cache locks

mwiencek:
This limits the number of cache locks we have open per entity type per transaction by mapping entity IDs to locks IDs with `% 50`.

This may increase contention, but cache deletions are relatively infrequent compared to reads. If desired we could add some logging in cases where it waits longer than 10ms or so for a lock.


mwiencek:
Updated with some logging

PR #1321, opened 2019-12-10
MBS-6539: Allow resolving MBIDs of unknown type

reosarevok:
https://tickets.metabrainz.org/browse/MBS-6539

This takes the quite hidden MBID "search" in the /search page and makes it possible to do /mbid/$mbid to redirect to the entity that MBID represents, whatever its type. The /search page option now just forwards to this.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/MBID.pm
+++ b/lib/MusicBrainz/Server/Controller/MBID.pm
@@ -1,0 +1,45 @@
+package MusicBrainz::Server::Controller::MBID;
+use Moose;
+
+use MusicBrainz::Server::Constants qw( entities_with );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
+use MusicBrainz::Server::Validation qw( is_guid );
+
+BEGIN { extends 'MusicBrainz::Server::Controller'; }
+
+sub base : Path('/mbid') CaptureArgs(0) { }
+
+sub show : Path('/mbid') CaptureArgs(1) {
+    my ($self, $c, $gid) = @_;
+
+    my $is_guid = is_guid($gid);
+
+    if ($is_guid) {
+        for my $model (entities_with('mbid', take => 'model')) {
+            my $entity = $c->model($model)->get_by_gid($gid) or next;
+            $c->response->redirect(
+                $c->uri_for_action(
+                    $c->controller($model)->action_for('show'),
+                    [ $gid ]));
+            $c->detach;
+        }
+    }
+
+    $c->stash(
+        component_path => 'mbid/NotFound',
+        component_props => {isGuid => boolean_to_json($is_guid), mbid => $gid},
+        current_view => 'Node',
+    );
+}
+
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2019 MetaBrainz Foundation
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
--- a/root/mbid/NotFound.js
+++ b/root/mbid/NotFound.js
@@ -1,0 +1,51 @@
+/*
+ * @flow
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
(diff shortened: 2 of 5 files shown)


reosarevok:
Note: this currently includes only a website version. Having the same for the WS would also be useful - I could make it into a different ticket, or try to add it to this PR.

lib/MusicBrainz/Server/Controller/MBID.pm
```diff
@@ -0,0 +1,47 @@
...
+use MusicBrainz::Server::Constants qw( entities_with );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
+use MusicBrainz::Server::Validation qw( is_guid );
+
+BEGIN { extends 'MusicBrainz::Server::Controller'; }
+
+sub base : Path('/mbid') CaptureArgs(0) { }
+
+sub show : Path('/mbid') CaptureArgs(1) {
+
```
mwiencek: ✂️ this newline

lib/MusicBrainz/Server/Controller/MBID.pm
```diff
@@ -0,0 +1,47 @@
...
+    my ($self, $c, $gid) = @_;
+
+    my $is_guid = is_guid($gid);
+
+    if ($is_guid) {
+      for my $model (entities_with('mbid', take => 'model')) {
```
mwiencek: This whole block is indented 6 spaces, should be 8
```suggestion
        for my $model (entities_with('mbid', take => 'model')) {
```

lib/MusicBrainz/Server/Controller/MBID.pm
```diff
@@ -0,0 +1,47 @@
...
+          $c->detach;
+      }
+    }
+
+    $c->stash(
+        component_path => 'mbid/NotFound',
+        component_props => {isGuid => boolean_to_json($is_guid), mbid => $gid},
+        current_view => 'Node',
+    );
+
```
mwiencek: ✂️

root/mbid/NotFound.js
```diff
@@ -0,0 +1,33 @@
...
+const MbidNotFound = ({isGuid, mbid}: {+isGuid: boolean, +mbid: string}) => (
+  <NotFound title={isGuid ? l('MBID Not Found') : l('Invalid MBID')}>
+    <p>
+      {isGuid ? (
+        texp.l(
+          `No MusicBrainz entities match the MBID {mbid}.
+           Either it’s incorrect, or the entity has since been deleted.`,
+          {mbid: mbid},
+        )
+      ) : (
```
yvanzo: GUID is used in MusicBrainz not only for entities : alternative release, release status, release type, media attribute type, media attribute type allowed value, and so on, that might be visible from various source (json-ld, web service, Picard-tagged files, potentially any 3rd-party using MB data) and thus legitimately looked at through this new `/mbid/` URL.

Please link to available documentation and shortly mention this other possible case.

reosarevok: What about something like: 
`No MusicBrainz entities match the MBID {mbid}. Either it’s incorrect, it was for an entity that has since been deleted, or it is for something else than an entity (for example, a relationship type).`

The only link that seems totally relevant is https://musicbrainz.org/doc/MusicBrainz_Identifier but that's pretty technical and not very good as a doc. We might want to improve those docs eventually, but is it a good idea to link to that in the meantime?

yvanzo: Yes, plus https://musicbrainz.org/doc/MusicBrainz_Entity cause entity is a developer term.

reosarevok: Ok, added a bit more info to the message as suggested.

root/mbid/NotFound.js
```diff
@@ -0,0 +1,40 @@
...
+          {
+            entity_doc: '/doc/MusicBrainz_Entity',
+            mbid: mbid,
+            mbid_doc: '/doc/MusicBrainz_Identifier',
+            rel_type_table: '/relationships',
+          },
+        )
+      ) : (
+        texp.l(
+          '{mbid} is not a valid MBID.',
```
yvanzo: * The same doc link as above would be nice to have.
* This sentence doesn’t make sense from `/mbid`, `/mbid?q=invisible_man`, `/mbid//invisible_man`, or `/mbid/reosarevok`.

reosarevok: It does make sense for `/mbid/reosarevok` (`reosarevok` is not a valid MBID!) but I added a separate option if no mbid is passed ("No MBID selected.").

PR #1250, opened 2019-10-28
Set up search for selenium/integration tests

mwiencek:
This will allow us to write tests for pages/features that make search queries.

PR #1324, opened 2019-12-11
MBS-10521: Fix format in /ws/js/medium search

mwiencek:
c4f6823e4465fb7dff9a631250d19079a4b32238 changed the release editor to no longer treat format as a string, but the `tracklist_results` subroutine in WS::js wasn't changed to match the corresponding TO_JSON changes.


mwiencek:
I can add a selenium test once we have https://github.com/metabrainz/musicbrainz-server/pull/1250, but would rather not base it on top of that giant PR.

PR #1325, opened 2019-12-11
Replace old raven SDKs

mwiencek:
According to https://github.com/getsentry/sentry-javascript these are still supported, but development of new features on them has ceased and the successors are @sentry/node and @sentry/browser.

I set up a local instance of Sentry to test this, and added some blatant errors in client/server code to make sure they were reported correctly.


root/server.js
```diff
@@ -10,9 +10,14 @@
 
 'use strict';
 
-const Raven = require('raven');
+const Sentry = require('@sentry/node');
 const DBDefs = require('./static/scripts/common/DBDefs');
-Raven.config(DBDefs.SENTRY_DSN).install();
+
+Sentry.init({
+  dsn: DBDefs.SENTRY_DSN_PUBLIC,
```
mwiencek: Note this uses `DBDefs.SENTRY_DSN_PUBLIC` instead of `DBDefs.SENTRY_DSN` now. That's intentional, as the settings page on Sentry writes:

> DSN (Deprecated)
> This DSN includes the secret which is no longer required by Sentry' newer versions of SDKs. If you are unsure which to use, follow installation instructions for your language.

I didn't change it for our Perl code because I don't think "newer versions of SDKs" applies to Sentry::Raven in cpan (we're not updating it here anyway, but I don't see any relevant changes in its git repo either).

PR #1326, opened 2019-12-12
MBS-10365: Convert entity lists to react-table (I)

reosarevok:
https://tickets.metabrainz.org/browse/MBS-10365

This is turning out a bit too big to do (and maybe review?) all in one go, so putting this one up for review and possible merge already, with a bunch of the simpler entity types.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -1018,6 +1018,10 @@ a.tagger-icon {
     width: 1px;
 }
 
+.number-column {
+    width: 1em;
+}
+
 .positive {
     background: @positive-light-bg;
 }
--- a/root/utility/tableColumns.js
+++ b/root/utility/tableColumns.js
@@ -13,6 +13,12 @@ import type {ColumnOptions} from 'react-table';
 import RatingStars from '../components/RatingStars';
 import SortableTableHeader from '../components/SortableTableHeader';
 import linkedEntities from '../static/scripts/common/linkedEntities';
+import ArtistCreditLink
+  from '../static/scripts/common/components/ArtistCreditLink';
+import ArtistRoles
+  from '../static/scripts/common/components/ArtistRoles';
+import AttributeList from '../static/scripts/common/components/AttributeList';
+import CodeLink from '../static/scripts/common/components/CodeLink';
 import DescriptiveLink
   from '../static/scripts/common/components/DescriptiveLink';
 import EntityLink from '../static/scripts/common/components/EntityLink';
@@ -145,6 +151,18 @@ export function defineEntityColumn<D>(
   };
 }
 
+export function defineSeriesNumberColumn(
+  seriesItemNumbers: {+[entityId: number]: string},
+): ColumnOptions<CoreEntityT, number> {
+  return {
+    Cell: ({cell: {value}}) => seriesItemNumbers[value],
+    Header: l('#'),
+    accessor: 'id',
+    className: 'number-column',
+    id: 'series-number',
+  };
+}
+
 export function defineTextColumn<D>(
   getText: (D) => string,
   columnName: string,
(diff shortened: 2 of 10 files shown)


root/utility/tableColumns.js
```diff
@@ -57,6 +203,41 @@ export function defineNameColumn<T: CoreEntityT | CollectionT>(
...
 
+export function defineSeriesNumberColumn(
+  seriesItemNumbers: {+[entityId: number]: string},
+): ColumnOptions<CoreEntityT, number> {
+  return {
+    Cell: ({cell: {value}}) => seriesItemNumbers[value],
+    Header: N_l('#'),
+    accessor: 'id',
+    className: 'number-column',
+    id: 'id',
```
yvanzo:
```suggestion
    id: 'series-number',
```

root/utility/tableColumns.js
```diff
@@ -91,6 +96,100 @@ export function defineTypeColumn(
...
+  getEntity: (D) => CoreEntityT,
+  columnName: string,
+  title: string,
+  order?: string = '',
+  sortable?: boolean = false,
+): ColumnOptions<D, string> {
+  return {
+    Cell: ({row: {original}}) => (getEntity(original)
+      ? <DescriptiveLink entity={getEntity(original)} />
+      : null),
```
mwiencek: I'd add the extra couple lines here and store `getEntity(original)` in a variable so we don't have to call it twice.

root/utility/tableColumns.js
```diff
@@ -143,6 +143,29 @@ export function defineEntityColumn<D>(
   };
 }
 
+export function defineTextColumn<D>(
```
mwiencek: This could probably be used for the label code too? (Maybe others)

root/utility/tableColumns.js
```diff
@@ -225,6 +243,84 @@ export const seriesOrderingTypeColumn:
...
+    Cell: ({row: {original}}) => <AttributeList entity={original} />,
+    Header: N_l('Attributes'),
+    accessor: 'attributes',
+  };
+
+export const workArtistsColumn:
+  ColumnOptions<WorkT, $ReadOnlyArray<ArtistCreditT>> = {
+    Cell: ({cell: {value}}) => (
+      <ul>
+        {value.map((artist, i) => (
```
mwiencek: I'd rename `artist` to `artistCredit` for clarity

root/static/styles/layout.less
```diff
@@ -1018,6 +1018,10 @@ a.tagger-icon {
     width: 1px;
 }
 
+.number-column {
+    width: 1em;
```
mwiencek: Seems a bit cramped since even something like "TH 5" wraps.

root/utility/tableColumns.js
```diff
@@ -143,6 +149,18 @@ export function defineEntityColumn<D>(
   };
 }
 
+export function defineSeriesNumberColumn(
+  seriesItemNumbers: {+[entityId: number]: string},
+): ColumnOptions<CoreEntityT, number> {
+  return {
+    Cell: ({cell: {value}}) => seriesItemNumbers[value],
+    Header: N_l('#'),
```
mwiencek: s/N_l/l/

root/utility/tableColumns.js
```diff
@@ -225,6 +243,84 @@ export const seriesOrderingTypeColumn:
...
+
+export const iswcsColumn:
+  ColumnOptions<{
+    +iswcs: $ReadOnlyArray<IswcT>,
+    ...,
+  }, $ReadOnlyArray<IswcT>> = {
+    Cell: ({cell: {value}}) => (
+      <ul>
+        {value.map((iswc, i) => (
+          <li key={i}>
```
mwiencek: Probably makes sense to use the iswc as the key here

mwiencek: Or well, `iswc.iswc`.

root/static/styles/layout.less
reosarevok: True, but it's what it already had. Wonder what would make more sense - do like with .actions and have it auto-wrap to as small as needed?

mwiencek: Yeah I realized after that you were just moving it here. Not sure honestly, CSS is hard. It's probably worth figuring out as a separate ticket.

root/utility/tableColumns.js
reosarevok: By making the passed getText something like `entity => formatLabelCode(entity.label_code)` ?

root/utility/tableColumns.js
reosarevok: This would still need to be inside the Cell function, right? So something like this?: 

```
    Cell: ({row: {original}}) => {
      const entity = getEntity(original);
      return (entity
        ? <DescriptiveLink entity={entity} />
        : null),
```

yvanzo: right!

root/utility/tableColumns.js
mwiencek: Looks like the current label code column handles empty values, so something like `entity => entity.label_code ? formatLabelCode(entity.label_code) : ''` seems more equivalent to what you have, but yeah.

reosarevok: Ok, done. Also changed order key code to label_code to avoid adding a new argument (and because it didn't make sense anyway)

PR #1327, opened 2019-12-13
MBS-10524: /account/change-password is broken unless logged in

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10524

The issue is ultimately that `UserAccountTabs` requires a user to function. It doesn't make sense to use user/profile/layout.tt at all without a user, but it "worked" until the React conversion because TT allows magic behavior.

This patch uses the normal layout if the user isn't logged in.

### DIFF
--- a/root/account/change_password.tt
+++ b/root/account/change_password.tt
@@ -1,7 +1,4 @@
-[% WRAPPER "user/profile/layout.tt" page="change_password" title=l("Change Password") %]
-
-    <h2>[% l("Change Password") %]</h2>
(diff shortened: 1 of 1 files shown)


mwiencek:
Oops, I didn't realize this was a duplicate of MBS-10234 (I did search first, but "public_subscriptions" failed to find anything). Actually I had no idea this was the reason some people couldn't log in (I forgot about the mandatory password reset thing) - just happened to see the error in Sentry.

PR #1330, opened 2019-12-14
Flow 0.114

mwiencek:


root/area/AreaArtists.js
```diff
@@ -18,7 +18,7 @@ import AreaLayout from './AreaLayout';
 type Props = {
   +$c: CatalystContextT,
   +area: AreaT,
-  +artists: $ReadOnlyArray<ArtistT>,
+  +artists: ?$ReadOnlyArray<ArtistT>,
```
mwiencek: The code was doing `artists && artists.length > 0` below, so either the type was wrong here or the check was unnecessary. I changed the type in this case and elsewhere, though in other cases I just removed the additional check if I was fairly sure it couldn't be null/undefined.

root/collection/CollectionIndex.js
```diff
@@ -190,7 +190,7 @@ const CollectionIndex = (props: Props) => {
         ) : null}
       </div>
       <div className="">
-        {collection.collaborators && collection.collaborators.length ? (
+        {collection.collaborators.length ? (
```
mwiencek: For example, here the check looked unnecessary, since we always output `collaborators` in `TO_JSON` (and it calls `collection.collaborators.some` above without the extra check, so it would've been already broken otherwise).

root/layout/components/sidebar/WorkSidebar.js
```diff
@@ -77,7 +77,7 @@ const WorkSidebar = ({$c, work}: Props) => {
               </SidebarProperty>
             ) : null}
 
-            {iswcs && iswcs.length ? (
+            {iswcs.length ? (
```
reosarevok: So iswcs always exists, then? :)

root/layout/components/sidebar/ArtistSidebar.js
```diff
@@ -72,9 +78,9 @@ const ArtistSidebar = ({$c, artist}: Props) => {
 
         <SidebarType entity={artist} typeType="artist_type" />
 
-        {artist.gender ? (
+        {gender ? (
           <SidebarProperty className="gender" label={l('Gender:')}>
-            {l(artist.gender.name)}
```
mwiencek: Flow doesn't like this anymore because a function is called inside an attribute (`l('Gender:')`). At run-time that function will be called before anything in the children (here `l(artist.gender.name)`), and Flow can't prove that it doesn't, say, set `artist.gender` to null before this line runs, so it invalidates the refinement. Which seems somewhat silly in this case, but it's trying to be as sound as possible.

On the other hand, if these were reversed (with `label={l(artist.gender.name)}` as the attribute and `{l('Gender:')}` as the child), Flow won't invalidate the refinement for the attribute anymore just because you call a function in the children. (It did this in 0.113 and earlier, which was a bug.) So it's better in some ways.

Declaring the variables above as `const` works around all of this, because Flow knows a const variable can't be reassigned.

root/layout/components/sidebar/WorkSidebar.js
mwiencek: Looks like it to me. I see `iswcs => [map { $_->TO_JSON } $self->all_iswcs],` is output unconditionally in `Entity::Work::TO_JSON`.

root/search/components/RecordingResults.js
```diff
@@ -125,7 +125,7 @@ export const RecordingResultsInline = ({
         <th>{l('Artist')}</th>
         <th>{l('ISRCs')}</th>
         <th>{l('Release')}</th>
-        {$c && $c.session && $c.session.tport ? <th>{l('Tagger')}</th> : null}
+        {$c?.session?.tport ? <th>{l('Tagger')}</th> : null}
```
yvanzo: I cannot understand why `$c` is optional here.
It has been made optional with commit 3e0afa96b7a1626f2fcf5c33e705c6c7a4e771ef but I don’t understand that change either.

mwiencek: @reosarevok any idea? :) you authored 3e0afa9

reosarevok: I made it optional because it was unused in, say, ArtistResultsInline. Is that not the right way to do this? Should we have different prop types for with and without context?

mwiencek: Ah okay. Perhaps, since we do that for `ResultsPropsT` / `ResultsPropsWithContextT` already. I'll leave it for a different PR though.

PR #1331, opened 2019-12-16
MBS-9919: Convert the place map tab to React

reosarevok:
This should be the last bit needed for https://tickets.metabrainz.org/browse/MBS-9919

### DIFF
--- a/root/place/PlaceMap.js
+++ b/root/place/PlaceMap.js
@@ -1,0 +1,45 @@
+/*
+ * @flow
+ * Copyright (C) 2019 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+import * as manifest from '../static/manifest';
+
+import PlaceLayout from './PlaceLayout';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


root/place/PlaceMap.js
```diff
@@ -0,0 +1,45 @@
...
+
+import PlaceLayout from './PlaceLayout';
+
+type Props = {
+  +mapDataArgs: {
+    +draggable: boolean,
+    +place: {
+      +coordinates: CoordinatesT | null,
+      +name: string,
+    },
```
reosarevok: Not too sure about this, not that it makes much of a difference what I put here in this case.

mwiencek: Which bit aren't you sure about?

reosarevok: Well, whether I've typed it right. Is \0 meant to be a boolean, or a number? (I'm assuming boolean).

mwiencek: Yeah, `\0` and `\1` get encoded as booleans. It's mentioned here https://metacpan.org/pod/JSON#other-references though using `JSON::false` and `JSON::true` is clearer.

reosarevok: Clearer as in "I might as well change to that while I'm at it"? :)

mwiencek: Heh, it was just an observation, you can change it if you feel like but otherwise don't bother. :)

PR #1338, opened 2019-12-21
Fixed files so that objects do not use non-prototype-builtins, going along with the ESLint guidelines.

ultra-programmer:
In accordance with ESLint, the files in musicbrainz-server have been edited to have no objects use non-prototype-builtins.

**Disambiguation:**
Instead of objects using the `.hasOwnProperty` method in the following way:
`object.hasOwnProperty('name')`
they would do it in the following way instead:
`Object.prototype.hasOwnProperty.call(object, 'name')`
where `object` is the object that the method is being called on.

**Command:** `./script/check_eslint_rule 'no-prototype-builtins: warn' root/`

### DIFF
--- a/root/static/scripts/common/components/Autocomplete.js
+++ b/root/static/scripts/common/components/Autocomplete.js
@@ -67,7 +67,9 @@ class Autocomplete extends React.Component {
       autocomplete.element.val(next.name);
     }
 
-    if (Object.prototype.hasOwnProperty.call(nextProps, 'isLookupPerformed')) {
+    if (Object.prototype.hasOwnProperty.call(
+      nextProps, 'isLookupPerformed'
+    )) {
       autocomplete.element.toggleClass(
         'lookup-performed',
         !!nextProps.isLookupPerformed,
--- a/root/static/scripts/common/components/Autocomplete.js
+++ b/root/static/scripts/common/components/Autocomplete.js
@@ -67,7 +67,7 @@ class Autocomplete extends React.Component {
       autocomplete.element.val(next.name);
     }
 
-    if (nextProps.hasOwnProperty('isLookupPerformed')) {
+    if (Object.prototype.hasOwnProperty.call(nextProps, 'isLookupPerformed')) {
       autocomplete.element.toggleClass(
         'lookup-performed',
         !!nextProps.isLookupPerformed,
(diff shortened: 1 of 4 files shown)


root/static/scripts/common/utility/getCookie.js
```diff
@@ -6,7 +6,8 @@ import _cookies from './_cookies';
 let defaultExport = getCookieFallback;
 
 function getCookieFallback(name, defaultValue = undefined) {
-  return _cookies.hasOwnProperty(name) ? _cookies[name] : defaultValue;
+  let cookieHasName = Object.prototype.hasOwnProperty.call(_cookies, name);
+  return cookieHasName ? _cookies[name] : defaultValue;
```
reosarevok: For this, it might be easier to just do:

```
  return Object.prototype.hasOwnProperty.call(_cookies, name)
    ? _cookies[name]
    : defaultValue;

```

root/static/scripts/common/components/Autocomplete.js
```diff
@@ -67,7 +67,9 @@ class Autocomplete extends React.Component {
       autocomplete.element.val(next.name);
     }
 
-    if (nextProps.hasOwnProperty('isLookupPerformed')) {
+    if (Object.prototype.hasOwnProperty.call(
+      nextProps, 'isLookupPerformed'
+    )) {
```
reosarevok: I would prefer this to be:

```
    if (Object.prototype.hasOwnProperty.call(
      nextProps,
      'isLookupPerformed',
    )) {
```

(same for the other one below on TagEditor.js)

root/static/scripts/common/components/TagEditor.js
```diff
@@ -290,7 +290,9 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
       };
 
       if (!this.state.positiveTagsOnly || isAlwaysVisible(t)) {
-        const isGenre = this.genreMap.hasOwnProperty(t.tag.name);
+        const isGenre = Object.prototype.hasOwnProperty.call(
+          this.genreMap, t.tag.name
+        );
```
reosarevok: This should change into two lines in the same way as the Autocomplete.js one :)

ultra-programmer: Sounds good. Thanks so much and happy holidays!

reosarevok:
Thanks, sorry for the delay with the merging! :)

PR #1342, opened 2020-01-04
MBS-10501: Display avatars for collaborators on collections

Kebabpizza:
https://tickets.metabrainz.org/browse/MBS-10501

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -422,7 +422,7 @@ sub load_for_collection {
     return unless $id; # nothing to do
 
     $self->load($collection);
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -422,12 +422,20 @@ sub load_for_collection {
     return unless $id; # nothing to do
 
     $self->load($collection);
-    my $query = "SELECT " . $self->_columns . "
+    my $query = "SELECT " . $self->_columns . ", ep.value AS prefs_value
```
mwiencek: I think I'd prefer renaming `prefs_value` to `show_gravatar` so it's clearer we're just selecting/storing that specific pref there, but just a small nitpick

yvanzo:
Thank you!

PR #1343, opened 2020-01-05
Updated the installation guide.

Zastai:
PostgreSQL changes:
- added a missing ';' to the ALTER USER
- note that for PostgreSQL 10 and later, UNENCRYPTED must be left
  off the ALTER USER
- explicitly note that the database must be dropped to restart an
  interrupted or failed run of InitDb --createdb

Translations:
- adjusted setup instructions, linking to .transifexrc docs
- noted that simply creating an account is not enough to be able
  to download the translations

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -383,16 +383,33 @@ If you intend to run a server with translations, there are a few steps to follow
 
 1.  Prerequisites
 
-    Make sure gettext is installed (you need msgmerge and msgfmt, at least),
-    and the transifex client 'tx'
-    (http://help.transifex.com/features/client/index.html):
+    Make sure gettext is installed (you need msgmerge and msgfmt, at least):
 
-        sudo apt-get install gettext transifex-client
+        sudo apt-get install gettext
 
-    You will need to make an account on transifex.com. Configure either a
-    username/password or an API key in ~/.transifexrc; see
-    (https://docs.transifex.com/client/client-configuration#-transifexrc) for
-    its contents.
+    This will enable you to compile and install the translations that are in
+    the source repository.
+    
+    If you want to get the latest translation files or partial work-in-progress
+    translations, or wish to work on translations yourself, you will need to
+    create a [Transifex](https://www.transifex.com/) account and install its
+    client software (`tx`):
+
+        sudo apt-get install transifex-client
+
+    More information (and alternative ways to install the client) can be found
+    at https://docs.transifex.com/client/introduction/.
+
+    Next, create an API token at https://www.transifex.com/user/settings/api/
+    and use it to configure your credentials in `~/.transifexrc`. Details at
+    https://docs.transifex.com/client/client-configuration#-transifexrc.
+    
+    Finally, you will need to join the
+    [MetaBrainz Foundation organization](https://www.transifex.com/musicbrainz/public/)
+    on Transifex to get access to the translations. If you wish to work on
+    translations, you will also need to join a language team at
+    (https://www.transifex.com/musicbrainz/musicbrainz/dashboard/); see also
+    https://musicbrainz.org/doc/Server_Internationalisation#Getting_started.
 
 2.  Change to the po directory
 
@@ -403,12 +420,16 @@ If you intend to run a server with translations, there are a few steps to follow
         tx pull -l {a list of languages you want to pull}
 
     This will download the .po files for your language(s) of choice to the po/
-    folder with the correct filenames.
-    If you want to get /all/ translations, you can also use `tx pull -a`.
+    folder with the correct filenames. Languages are written as an ISO language
+    code, optionally followed by an underscore and an ISO country code (e.g. `fr`
+    for French, `fr_CA` for Canadian French).
 
-    Note that you will need to join the MetaBrainz Foundation organization
-    (https://www.transifex.com/musicbrainz/public/) in order to be able to do
-    this, or you will get `Forbidden` errors from `tx pull`.
+    Or, if you want to get _all_ translations instead:
+
+        tx pull -a
+
+    If you get `Forbidden` errors from `tx pull`, you will need to make sure
+    you have joined the MusicBrainz organization and/or project (see point 1).
 
 4.  Install translations
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -251,8 +251,9 @@ Creating the database
     users (whoever your web server/crontabs run as), to one database (the
     MusicBrainz Database), as one PostgreSQL user. The PostgreSQL database name
     and user name are given in DBDefs.pm (look for the `READWRITE` key).  For
-    example, if you run your web server and crontabs as "www-user", the
-    following configuration recipe may prove useful:
+    example, if you run your web server and crontabs as "www-user", and you have
+    kept the default PostgreSQL user name ("musicbrainz"), the following
+    configuration recipe may prove useful:
 
         # in pg_hba.conf (Note: The order of lines is important!):
         local    musicbrainz_db    musicbrainz    ident    map=mb_map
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -386,8 +393,10 @@ If you intend to run a server with translations, there are a few steps to follow
 
         sudo apt-get install gettext transifex-client
 
-    Configure a username and password in ~/.transifexrc using the format listed
-    on the above page.
+    You will need to make an account on transifex.com. Configure either a
+    username/password or an API key in ~/.transifexrc; see
+    (https://docs.transifex.com/client/client-configuration#-transifexrc) for
+    its contents.
```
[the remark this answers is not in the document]
Zastai: Well, this installation guide is for developers, who might want to run with the latest translations. Not necessarily someone who wants to translate anything themselves. So some duplication is to be expected.

Plus, that getting started page does not link to the .transifexrc, nor does it link to the MetaBrainz organisation you have to join.

INSTALL.md
```diff
@@ -399,6 +408,11 @@ If you intend to run a server with translations, there are a few steps to follow
 
     This will download the .po files for your language(s) of choice to the po/
     folder with the correct filenames.
+    If you want to get /all/ translations, you can also use `tx pull -a`.
+
+    Note that you will need to join the MetaBrainz Foundation organization
+    (https://www.transifex.com/musicbrainz/public/) in order to be able to do
+    this, or you will get `Forbidden` errors from `tx pull`.
```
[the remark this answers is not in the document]
Zastai: Yeah, that would make sense.

INSTALL.md
```diff
@@ -270,10 +270,13 @@ Creating the database
     By default, the password for the user musicbrainz should be "musicbrainz",
     as stated in lib/DBDefs.pm. You can change it with `psql`:
 
-        postgres=# ALTER USER musicbrainz UNENCRYPTED PASSWORD 'musicbrainz'
+        postgres=# ALTER USER musicbrainz UNENCRYPTED PASSWORD 'musicbrainz';
+
+    For PostgreSQL 10 or later, you will need to leave off the `UNENCRYPTED`.
```
mwiencek: Huh, I can't see any reason to specify `UNENCRYPTED` in 9.5 either. According to the docs, that indicated whether the given password was to be encrypted in the system catalog (not whether the given password was already in encrypted format). I'd just remove `UNENCRYPTED` from the original line.

INSTALL.md
```diff
@@ -399,6 +408,11 @@ If you intend to run a server with translations, there are a few steps to follow
 
     This will download the .po files for your language(s) of choice to the po/
     folder with the correct filenames.
+    If you want to get /all/ translations, you can also use `tx pull -a`.
```
mwiencek: asterisks will give actual italics in markdown
```suggestion
    If you want to get *all* translations, you can also use `tx pull -a`.
```

INSTALL.md
Zastai: The 9.5 docs do not quite make it clear whether leaving it off would assume/require the value to be an md5 hash. I could switch it around and say that 9.x might require `UNENCRYPTED`.

Would it also make sense to change the main line to use `CREATE`, given this is mostly a set-up-from-scratch guide (mentioning `ALTER`In text afterwards)?

INSTALL.md
Zastai: I intended italics, but have no problem it switching to bold.

mwiencek: I know :) I meant that `*all*` (or `_all_`) will display in italics as intended.

Zastai: My bad - thought asterisks was bold. Is also the disadvantage of making the edit in vi - no markdown preview :)

INSTALL.md
mwiencek: The relevant docs for ENCRYPTED / UNENCRYPTED are actually found under [CREATE ROLE](https://www.postgresql.org/docs/9.5/sql-createrole.html) since those options are inherited from there:

> These key words control whether the password is stored encrypted in the system catalogs. (If neither is specified, the default behavior is determined by the configuration parameter password_encryption.) If the presented password string is already in MD5-encrypted format, then it is stored encrypted as-is, regardless of whether ENCRYPTED or UNENCRYPTED is specified (since the system cannot decrypt the specified encrypted password string). This allows reloading of encrypted passwords during dump/restore.

So I don't see any reason to specify `UNENCRYPTED` for 9.5.

Re: changing it to `CREATE`, this is actually kind of weird since admin/InitDb.pl creates the musicbrainz user if it doesn't exist (setting the password to the one specified in DBDefs). So maybe we should just say to make sure the desired password in specified for `READWRITE` in DBDefs before running InitDb.pl. I'm not sure we really need to document how to change the password once the user is already created.

https://github.com/metabrainz/musicbrainz-server/blob/396d0ca13a917d1b7fddd8c1a52e18bb12c15f22/admin/InitDb.pl#L200-L211

Zastai: OK, even easier, I'll drop that paragraph. I always manually created the user before running InitDB, because of where this ALTER was documented.

INSTALL.md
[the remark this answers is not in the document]
Zastai: I've now reworked the translation section.

INSTALL.md
```diff
@@ -267,13 +267,12 @@ Creating the database
 
         local   all    all    trust
 
-    By default, the password for the user musicbrainz should be "musicbrainz",
-    as stated in lib/DBDefs.pm. You can change it with `psql`:
-
-        postgres=# ALTER USER musicbrainz UNENCRYPTED PASSWORD 'musicbrainz'
+    You do not need to create this user in PostgreSQL yourself; the next step will
```
mwiencek: I'd just clarify this part a bit since "this user" doesn't really follow from the previous paragraph
```suggestion
    You do not need to create the musicbrainz user in PostgreSQL yourself; the next step will
```

INSTALL.md
```diff
@@ -380,14 +383,33 @@ If you intend to run a server with translations, there are a few steps to follow
...
+    Next, create an API token at https://www.transifex.com/user/settings/api/
+    and use it to configure your credentials in `~/.transifexrc`. Details at
+    https://docs.transifex.com/client/client-configuration#-transifexrc.
+    
+    Finally, you will need to join the
+    [MetaBrainz Foundation organization](https://www.transifex.com/musicbrainz/public/)
+    on Transifex to get access to the translations. If you wish to work on
+    translations, you will also need to join a language team at
+    (https://www.transifex.com/musicbrainz/musicbrainz/dashboard/); see also
+    https://musicbrainz.org/doc/Server_Internationalisation#Getting_started.
```
mwiencek: Replacing the URLs with textual links might read a bit nicer in rendered form, e.g. `[join a language team](https://www.transifex.com/musicbrainz/musicbrainz/dashboard/)` instead of "join at language team at [long URL]," or `see also our documentation on [getting started with i18n](https://musicbrainz.org/doc/Server_Internationalisation#Getting_started).`

INSTALL.md
Zastai: Right, will tweak that tomorrow.

INSTALL.md
Zastai: Fair point, will change those too.

INSTALL.md
Zastai: Also noticed the dump import points directly to the main ftp (which doesn't even get turned into a hyperlink), instead of the server page that also lists the mirrors; will update that too.

Zastai:
Should I squash these commits into one on my end once everything is approved, or will that be done as part of the merge?

reosarevok:
We can squash at the end, assuming they can all be one commit, so no worries.

mwiencek:
Thanks for the improvements!

PR #1349, opened 2020-01-13
Fix MBS-10366: Remove rather than renumber pregap track if DiscID is present

reosarevok:
MBS-10366

If a medium has a discID attached, the user shouldn't be able to move the pregap track to track 1 by unchecking the "has pregap" checkbox, since the number of "normal" tracks is frozen by the discID. This makes it so the pregap track is removed if there is a discID, but still renumbered otherwise, which is consistent with the way data tracks work.

I also added a toggleable help section with a second commit, that looks like this: 
Screenshot from 2020-01-18 23-19-04

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -408,14 +408,32 @@
           <p style="margin-top: 0;">
             <label>
               <input class="has-pregap" type="checkbox" data-bind="checked: hasPregap" />
-              [% l('This disc has a hidden pregap track before track 1 ({info|more info})', { info => { href => doc_link('Hidden_Track'), target => '_blank' } }) %]
+              [% l('This disc has a hidden pregap track before track 1') %]
+              (<a href="#" data-bind="click: togglePregapTrackHelp">[% l('help') %]</a>)
             </label>
+            <div class="ar-descr" data-bind="visible: showPregapTrackHelp">
+              <p>
+                [% l('Some discs contain a hidden track in the pregap section that precedes track 1. Use this to add (or remove) the special pregap track section.') %]
+              </p>
+              <p>
+                [% l('Keep in mind that unselecting this will delete the track if the medium has a disc ID! If you unselect it by mistake, please readd the pregap track before submitting.') %]
+              </p>
+            </div>
           </p>
           <p style="margin-top: 0;">
             <label>
               <input class="has-data-tracks" type="checkbox" data-bind="checked: hasDataTracks" />
-              [% l('This disc contains data tracks at the end ({info|more info})', { info => { href => doc_link('Data_Track'), target => '_blank' } }) %]
+              [% l('This disc contains data tracks at the end') %]
+              (<a href="#" data-bind="click: toggleDataTracksHelp">[% l('help') %]</a>)
             </label>
+            <div class="ar-descr" data-bind="visible: showDataTracksHelp">
+              <p>
+                [% l('Some discs contain one or more data tracks after all audio tracks. Use this to add (or remove) the special data tracks section. You should only add data tracks that contain audio or video ({info|more info})', { info => { href => doc_link('Data_Track'), target => '_blank' } }) %]
+              </p>
+              <p>
+                [% l('Keep in mind that unselecting this will delete the tracks if the medium has a disc ID! If you unselect it by mistake, please readd the data tracks before submitting.') %]
+              </p>
+            </div>
           </p>
         <!-- /ko -->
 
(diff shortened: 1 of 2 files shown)


mwiencek:
The title of the PR is misleading too ("Block adding/removing pregap if TOC present") since this doesn't appear to block adding a pregap if there's a TOC, and that's not something we want to do.

mwiencek:
> whether it has one or not is already frozen by the discID

To clarify, with the TOC we can see the length of the pregap before track 1, but not whether there's actually anything there. So having a track 0 is independent of the DiscID. We allow the user to indicate there's something there, but we can't say anything about it other than what it's maximum length would be.

reosarevok:
I'm starting to think the right answer here is just to change the way pregap works so you either delete or add it, without the renumbering, and then not block anything.

reosarevok:
Did that.

yvanzo:
Doesn’t it prevent adding a legitimate DiscID to a release after the pregap has been moved to track 1? It is unclear to me which cases are legitimate and [doc about “pregap”](https://musicbrainz.org/search?type=doc&query=pregap) is quite sparse.

reosarevok:
The current situation is that pregap tracks are always moved to track 1 when deselected. That's still fine when the release has no discid, but won't happen anymore if it does have one. Yes, having the pregap as track 1 would avoid adding the correct discid, but that's because the data would be wrong and need correcting first. It also happens now already unless I'm missing something :)

reosarevok:
One link was lost because there's nothing more of any use there. The same info is now in the help section directly. The other is still there. Will look into the comments though.

reosarevok:
>I still wonder how documentation may disappear…

The "hidden track" doc link was to:

> hidden track
> A hidden track can mean:
> 
> a track that is not listed on the release's sleeve.
> a song in the pregap of track 1 on a CD, i.e. is located before the first track (a so-called pregap track).
> one or more songs that follow another song on the same track, usually after a longer period of silence, but are not listed on the sleeve.
> See the special purpose track title guidelines and, where necessary, the guidelines for multiple titles for details about how to name these tracks.

From that I kept the only bit relevant to pregaps specifically, "Some discs contain a hidden track in the pregap section that precedes track 1", and put it on the help section, so I dropped the link :)

root/release/edit/tracklist.tt
```diff
@@ -408,14 +408,32 @@
...
-              [% l('This disc has a hidden pregap track before track 1 ({info|more info})', { info => { href => doc_link('Hidden_Track'), target => '_blank' } }) %]
+              [% l('This disc has a hidden pregap track before track 1') %]
+              (<a href="#" data-bind="click: togglePregapTrackHelp">[% l('help') %]</a>)
             </label>
+            <div class="ar-descr" data-bind="visible: showPregapTrackHelp">
+              <p>
+                [% l('Some discs contain a hidden track in the pregap section that precedes track 1. Use this to add (or remove) the special pregap track section.') %]
+              </p>
+              <p>
+                [% l('Keep in mind that unselecting this will delete the track if the medium has a discID! If you unselect it by mistake, please readd the pregap track before submitting.') %]
```
mwiencek:
```suggestion
                [% l('Keep in mind that unselecting this will delete the track if the medium has a disc ID! If you unselect it by mistake, please readd the pregap track before submitting.') %]
```

I really don't see the point of adding "if you make a mistake, undo it" to any help message. But as long as "discID" -> "disc ID" is fixed I'll approve. :)

root/release/edit/tracklist.tt
```diff
@@ -408,14 +408,32 @@
...
-              [% l('This disc contains data tracks at the end ({info|more info})', { info => { href => doc_link('Data_Track'), target => '_blank' } }) %]
+              [% l('This disc contains data tracks at the end') %]
+              (<a href="#" data-bind="click: toggleDataTracksHelp">[% l('help') %]</a>)
             </label>
+            <div class="ar-descr" data-bind="visible: showDataTracksHelp">
+              <p>
+                [% l('Some discs contain one or more data tracks after all audio tracks. Use this to add (or remove) the special data tracks section. You should only add data tracks that contain audio or video ({info|more info})', { info => { href => doc_link('Data_Track'), target => '_blank' } }) %]
+              </p>
+              <p>
+                [% l('Keep in mind that unselecting this will delete the tracks if the medium has a discID! If you unselect it by mistake, please readd the data tracks before submitting.') %]
```
mwiencek:
```suggestion
                [% l('Keep in mind that unselecting this will delete the tracks if the medium has a disc ID! If you unselect it by mistake, please readd the data tracks before submitting.') %]
```

PR #1352, opened 2020-01-14
Implement MBS-9477, MBS-10568, MBS-10569: DiscID tab improvements

reosarevok:
MBS-9477, MBS-10568, MBS-10569

See commit messages for details. Screenshots:

Screenshot from 2020-01-14 22-43-05
Screenshot from 2020-01-14 22-43-08
Screenshot from 2020-01-14 22-43-10

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -91,6 +91,9 @@ after 'load' => sub {
     my $artwork_count = $c->model('Artwork')->find_count_by_release($release->id);
     $c->stash->{release_artwork_count} = $artwork_count;
 
+    my $cdtoc_count = $c->model('MediumCDTOC')->find_count_by_release($release->id);
+    $c->stash->{release_cdtoc_count} = $cdtoc_count;
+
     # We need to load more artist credits in 'show'
     if ($c->action->name ne 'show') {
         $c->model('ArtistCredit')->load($release);
--- a/root/components/EntityTabLink.js
+++ b/root/components/EntityTabLink.js
@@ -13,13 +13,22 @@ import EntityLink from '../static/scripts/common/components/EntityLink';
 
 type Props = {
   +content: string,
+  +disabled?: boolean,
   +entity: CoreEntityT | CollectionT,
   +selected: boolean,
   +subPath: string,
 };
 
-const EntityTabLink = ({selected, ...linkProps}: Props) => (
-  <li className={selected ? 'sel' : null}>
+const EntityTabLink = ({disabled, selected, ...linkProps}: Props) => (
+  <li
+    className={
+      selected || disabled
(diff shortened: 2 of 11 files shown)


zas:
I'm not fond of `(n/a)`: while it makes sense and is fairly common, it may not translate well (=being short) in all languages.
That said, I'm not sure what would be better. Initially I was thinking about greying out the tab (disabling it) or just hiding it there are no CD medium in the release.

reosarevok:
This *also* greys it out and disables it. We could just not show anything in parens (just "Disc IDs") in that case.

zas:
> This also greys it out and disables it.

Ah ok :)

> We could just not show anything in parens (just "Disc IDs") in that case.

Perhaps the best option.

reosarevok:
Made the change to remove "(n/a)" and just leave a disabled "DiscIDs".

lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -109,6 +112,8 @@ after 'load' => sub {
         $c->model('CritiqueBrainz')->load_display_reviews($release->release_group)
             unless $returning_jsonld;
     }
+
+    $c->stash->{may_have_discids} = $release->may_have_discids;
```
mwiencek: It might be a bit cleaner to just use `release.may_have_discids` in the templates (which should work without any changes in TT, but for the JS we'd have to serialize it in `TO_JSON`). Though you may have some other reason for adding it to the stash that I've missed.

root/components/EntityTabLink.js
```diff
@@ -13,13 +13,22 @@ import EntityLink from '../static/scripts/common/components/EntityLink';
...
-const EntityTabLink = ({selected, ...linkProps}: Props) => (
-  <li className={selected ? 'sel' : null}>
+const EntityTabLink = ({disabled, selected, ...linkProps}: Props) => (
+  <li
+    className={
+      selected || disabled
+        ? (selected ? 'sel' : '') +
+          (selected && disabled ? ' ' : '') +
+          (disabled ? 'disabled' : '')
+        : null
```
mwiencek: Since a tab can't be both selected and disabled (right?), would this be equivalent?
```JS
(selected ? 'sel' : (disabled ? 'disabled' : null))
```

reosarevok: It can, actually - but just by manually going to /discids. I didn't block that because there might be a case where (due to a bug or whatnot, I think some bugs related to merges still exist currently) a release that shouldn't have discIDs ends up having some.

lib/MusicBrainz/Server/Controller/Release.pm
reosarevok: This is much better in other ways, but it breaks Flow, since in EntityTabs all we know is "CoreEntityT" (that has discids = true, but that doesn't help Flow). Suggestions?

PR #1355, opened 2020-01-15
MBS-10365: Convert entity lists to react-table (II)

reosarevok:
MBS-10365

* [ ] Artist
* [ ] Event
* [ ] Recording
* [x] Release
* [ ] Release group

### DIFF
--- a/root/components/list/ReleaseList.js
+++ b/root/components/list/ReleaseList.js
@@ -9,22 +9,24 @@
 
 import React from 'react';
 
+import Table from '../Table';
 import {withCatalystContext} from '../../context';
-import loopParity from '../../utility/loopParity';
-import InstrumentRelTypes from '../InstrumentRelTypes';
-import ReleaseCatnoList from '../ReleaseCatnoList';
-import ReleaseLabelList from '../ReleaseLabelList';
 import filterReleaseLabels
   from '../../static/scripts/common/utility/filterReleaseLabels';
 import formatBarcode from '../../static/scripts/common/utility/formatBarcode';
-import ArtistCreditLink
-  from '../../static/scripts/common/components/ArtistCreditLink';
-import EntityLink from '../../static/scripts/common/components/EntityLink';
-import ReleaseEvents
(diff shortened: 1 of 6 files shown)


root/utility/tableColumns.js
```diff
@@ -294,14 +421,20 @@ export const seriesOrderingTypeColumn:
...
-
 export const subscriptionColumn:
   ColumnOptions<{+subscribed: boolean, ...}, boolean> = {
     Cell: ({cell: {value}}) => yesNo(value),
     Header: N_l('Subscribed'),
     accessor: 'subscribed',
   };
 
+export const taggerColumn:
+  ColumnOptions<{+subscribed: boolean, ...}, boolean> = {
```
mwiencek: The `+subscribed` part of the type seems wrong.

reosarevok: Oh, yes. I changed it to ColumnOptions<RecordingT | ReleaseT, RecordingT | ReleaseT> but I'm actually not sure what the second part should even be (I don't have an accessor at all here).

yvanzo: `ColumnOptions<RecordingT | ReleaseT, void>` would probably be just fine.

root/components/list/ReleaseList.js
```diff
@@ -50,192 +52,98 @@ const ReleaseList = ({
...
+  const columns = React.useMemo(
+    () => {
+      const checkboxColumn = $c.user_exists && checkboxes
+        ? defineCheckboxColumn(checkboxes)
+        : null;
+      const seriesNumberColumn = seriesItemNumbers
+        ? defineSeriesNumberColumn(seriesItemNumbers)
+        : null;
+      const nameColumn =
+        defineNameColumn<ReleaseT>(l('Release'), order, sortable, false);
```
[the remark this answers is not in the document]
reosarevok: I added a comment for now (I'd be ok with changing this to a structured argument but then we should do it for all the functions, so maybe it's better to finish this first and then make that change in one go / one PR).

PR #1356, opened 2020-01-16
MBS-10365: Convert entity lists to react-table (III)

reosarevok:
MBS-10365
Also fixes MBS-10583

* [ ] Artist
* [x] Event
* [ ] Recording
* [x] Release
* [ ] Release group

After https://github.com/metabrainz/musicbrainz-server/pull/1355

### DIFF
--- a/root/components/list/EventList.js
+++ b/root/components/list/EventList.js
@@ -9,20 +9,23 @@
 
 import React from 'react';
 
-import commaOnlyList from '../../static/scripts/common/i18n/commaOnlyList';
+import Table from '../Table';
+import {withCatalystContext} from '../../context';
+import {commaOnlyListText}
+  from '../../static/scripts/common/i18n/commaOnlyList';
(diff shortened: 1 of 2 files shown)


root/components/list/EventList.js
```diff
@@ -52,115 +55,75 @@ const EventList = ({
...
+        : null;
+      const seriesNumberColumn = seriesItemNumbers
+        ? defineSeriesNumberColumn(seriesItemNumbers)
+        : null;
+      const nameColumn =
+        defineNameColumn<EventT>(
+          l('Event'),
+          order,
+          sortable,
+          false, // to use EntityLink without dates (separate column for that)
```
yvanzo: It seems to fix a regression, might be worth to notify users about that through a ticket.
(Event search results have been changed the same way in MBS-9765.)

reosarevok: Ok, created MBS-10583

PR #1357, opened 2020-01-17
MBS-10392 Convert Add Series edit to React

anirudhjain75:


root/edit/details/AddSeries.js
```diff
@@ -0,0 +1,71 @@
...
+    +series: SeriesT,
+    +type: SeriesTypeT,
+  },
+};
+
+type Props = {
+  edit: AddSeriesEditProps,
+};
+
+const AddSeries = ({edit}: Props) => {
```
reosarevok: Wouldn't it be easier to do 
```suggestion
const AddSeries = ({edit}: {edit: AddSeriesEditProps}) => {
```
here and avoid the second Props above?

root/edit/details/AddSeries.js
```diff
@@ -0,0 +1,71 @@
...
+          {edit.display_data.comment ? (
+            <tr>
+              <th>{addColonText(l('Disambiguation'))}</th>
+              <td>{edit.display_data.comment}</td>
+            </tr>
+          ) : null}
+          {edit.display_data.type ? (
+            <tr>
+              <th>{addColonText(l('Type'))}</th>
+              <td>{edit.display_data.type.name}</td>
```
reosarevok: We should probably translate this, even if it wasn't translated originally for some reason.

root/edit/details/AddSeries.js
```diff
@@ -0,0 +1,71 @@
...
+            </tr>
+          ) : null}
+          {edit.display_data.ordering_type ? (
+            <tr>
+              <th>{addColonText(l('Ordering Type'))}</th>
+              <td>{edit.display_data.ordering_type.name}</td>
```
reosarevok: I think this should be translatable too, but I'm not sure how exactly. @mwiencek, would this be l_attributes with ordering_type or something? (there's l_name but we shouldn't use that).

root/edit/details/AddSeries.js
```diff
@@ -0,0 +1,71 @@
...
+    <>
+      <table className="details">
+        <tbody>
+          <tr>
+            <th>{addColonText(l('Series'))}</th>
+            <td><EntityLink entity={edit.display_data.series} /></td>
+          </tr>
+        </tbody>
+      </table>
+      <table className="details add-series">
```
reosarevok: I know this is in the original code, but I can't figure out why we have these two as separate tables...

root/edit/details/AddSeries.js
reosarevok: I think you'd want:

```suggestion
              <td>{lp_attributes(edit.display_data.type.name, 'series_type')}</td>
```

root/edit/details/AddSeries.js
reosarevok: I think:

```suggestion
              <td>
                {lp_attributes(
                  edit.display_data.ordering_type.name,
                  'series_ordering_type',
                )}
              </td>
```

root/edit/details/AddSeries.js
reosarevok: On second thought, you can leave this be for now and we can fix them all in one commit later.

root/edit/details/AddSeries.js
anirudhjain75: thanks :)

root/edit/details/AddSeries.js
anirudhjain75: Okay !

root/edit/details/AddSeries.js
anirudhjain75: Thanks

mwiencek:
This would crash trying to display edits where the series was later deleted. I fixed that and some minor issues with the Flow types. See https://github.com/metabrainz/musicbrainz-server/compare/890c5a11eb1beec038004845f5b35164693d1e3e..89d79a101455933e7845df7d3cd32c5194a96f9a

PR #1358, opened 2020-01-19
MBS-10571: Fix "localized" ModBot email display

mwiencek:
ModBot's notes can be localized since d2f9ced using a special syntax. Where it was used, the notes weren't being mailed correctly, instead exposing the internal syntax to the recipient.

This passes the ModBot note text through `Entity::EditNote::localize` before sending out the email. `editor_id` and `text` are the only attributes needed to run that method.

The message is sent in English (via a new `run_without_translations` method in `Server::Translation`) because we don't know the language of the recipient. This is the behavior it's always had, of course.

### DIFF
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -18,6 +18,7 @@ use MusicBrainz::Server::Constants qw(
     :edit_status
     :email_addresses
     $CONTACT_URL
+    $EDITOR_MODBOT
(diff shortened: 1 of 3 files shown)


reosarevok:
> The message is sent in English (via a new run_without_translations method in Server::Translation) because we don't know the language of the recipient. This is the behavior it's always had, of course.

It might make sense to eventually add a preference for email languages, but for now I guess that's the best we can do.

PR #1359, opened 2020-01-19
Fix MBS-10572: Have one Jed bundle per language

mwiencek:
Prior to this patch, we had separate Jed data bundles per language per domain, and any page that used a domain other than mb_server had to specify that so it could be loaded. This was horrible, because we'd have to remember to update these domain lists on every change, and before we did, had to figure out what pages the change affected.

For example, re MBS-10572 an l_countries call was introduced into the ReleaseEvents component, which it turns out is used on a lot of pages that don't have the countries domain loaded.

This patch combines all domain-specific Jed data bundles into a single bundle per language. These new bundles are around 500KB before gzip and loaded on every page if a language is selected. Some unnecessary data will be loaded on some pages now, but on a lot of edit pages we had to load most or all of the domains anyway; this at least reduces the number of HTTP requests in those cases. In other cases, we don't have to worry about loading them now.

There is a potential for future work in reducing the size of these.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -37,6 +37,7 @@ module.exports = function (api) {
   const ignore = [
     'node_modules',
     'root/static/scripts/tests/typeInfo.js',
+    /root\/static\/build\/jed-[A-z_-]+?\.source\.js$/,
   ];
 
   return {
--- a/root/account/EditProfile.js
+++ b/root/account/EditProfile.js
@@ -30,7 +30,6 @@ const EditProfile = ({$c, ...props}: Props) => {
   return (
     <UserAccountLayout
       entity={user}
-      gettextDomains={['attributes']}
       page="edit_profile"
       title={l('Edit Profile')}
     >
--- a/root/area/create.tt
+++ b/root/area/create.tt
@@ -1,4 +1,4 @@
-[%- WRAPPER 'layout.tt' title=l('Add Area') full_width=1 gettext_domains=['attributes', 'countries', 'instruments', 'instrument_descriptions', 'languages', 'relationships'] -%]
+[%- WRAPPER 'layout.tt' title=l('Add Area') full_width=1 -%]
     <div id="content">
         <h1>[%- l('Add Area') -%]</h1>
         [%- INCLUDE "area/edit_form.tt" -%]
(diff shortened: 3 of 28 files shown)


mwiencek:
> There is a potential for future work in reducing the size of these.

For example, replacing all of the English source strings in them with integers, and mapping the gettext calls somehow that way is an idea I had (though it'd require having an actual en bundle). I'm sure that's at least 1/3rd of the total size in all cases.

These files are extremely simple to parse and execute though, so once they're cached I don't think performance is currently an issue.

mwiencek:
> Do I understand correctly that we're still keeping all the subdomains, just loading them all every time? By that time, would it make sense to stop having subdomains altogether, and also simplify stuff by getting rid of l_whatever functions?

That's correct. It still works differently on the server (in either Perl or Node), since there the domains still aren't loaded until they're used. But yes, I don't think there's any technical reason we can't have a single mb_server domain. (There are already some strings duplicated between them, which isn't ideal.)

mwiencek:
BTW, the referenced ticket MBS-10572 is triggering tens of thousands of errors on Sentry currently, so we should work to include this in today's release.

yvanzo:
> > By that time, would it make sense to stop having subdomains altogether, and also simplify stuff by getting rid of l_whatever functions?
> 
> I don't think there's any technical reason we can't have a single mb_server domain. (There are already some strings duplicated between them, which isn't ideal.)

That would certainly simplify server coding stuff, but we should also take care of potential impacts on:
* translators: Lesser duplicates is nice, but separate resources have some advantages still: these provide a bit of context (that could probably be replaced with context strings), and one can focus on a specific domain as a priority (Transifex provide separate statistics for each resource).
* related projects: MusicBrainz Picard and android app currently use database-extracted domains only.

mwiencek:
Yes, those are good points! I'd love to hear feedback from translators about it, especially where they see context being missing that domains might provide.

I do have a feeling a single `mb_server` domain would even simplify Picard's code a bit, as annoying as it'd be to have to change something that's working. They'd likely just want to filter out strings they don't need with `msggrep` first.

mwiencek:
Updated the babel config to have it ignore the jed-*.source.js files, to prevent these messages:

> [BABEL] Note: The code generator has deoptimised the styling of /musicbrainz-server/root/static/build/jed-de.source.js as it exceeds the max of 500KB.

I also changed these to use `JSON.parse` for the massive objects holding all the strings. See https://v8.dev/blog/cost-of-javascript-2019#json

PR #1361, opened 2020-01-21
Fix MBS-10576: Support errors with HTML

reosarevok:
MBS-10576

This allows passing forms errors containing HTML. We interpret the errors as HTML only when specifically requested to ensure we still escape the appropriate characters otherwise.

For now, this changes only form_row_select since that's the one that already has one error using HTML, but we might want to expand it in the future (if we add links to documentation in error texts, for example).

Even though this isn't yet needed anywhere that uses the React components, I'm adding it there already for parity - we'll need it once we convert release/merge anyway.

### DIFF
--- a/root/components/FieldErrors.js
+++ b/root/components/FieldErrors.js
@@ -9,17 +9,24 @@
 
 import React from 'react';
 
+import expand2react from '../static/scripts/common/i18n/expand2react';
 import subfieldErrors, {type FieldShape} from '../utility/subfieldErrors';
 
-const buildErrorListItem = (error, index) => (
-  <li key={index}>{error}</li>
-);
+const buildErrorListItem = (error, hasHtmlErrors, index) => {
+  if (hasHtmlErrors) {
+    return (
+      <li key={index}>{expand2react(error)}</li>
+    );
+  }
+  return <li key={index}>{error}</li>;
+};
 
 type Props = {
   +field: FieldShape,
+  +hasHtmlErrors?: boolean,
 };
 
-const FieldErrors = ({field}: Props) => {
+const FieldErrors = ({field, hasHtmlErrors}: Props) => {
   if (!field) {
     return null;
   }
@@ -27,7 +34,9 @@ const FieldErrors = ({field}: Props) => {
   if (errors.length) {
     return (
       <ul className="errors">
-        {errors.map(buildErrorListItem)}
+        {errors.map(function (error, index) {
+          return buildErrorListItem(error, hasHtmlErrors, index);
+        })}
       </ul>
     );
   }
--- a/root/components/FieldErrors.js
+++ b/root/components/FieldErrors.js
@@ -12,6 +12,7 @@ import React from 'react';
 import expand2react from '../static/scripts/common/i18n/expand2react';
 import subfieldErrors, {type FieldShape} from '../utility/subfieldErrors';
 
+// FIXME: Use expandable object instead of HTML string for safety (MBS-10632)
 const buildErrorListItem = (error, hasHtmlErrors, index) => {
   if (hasHtmlErrors) {
     return (
--- a/lib/MusicBrainz/Server/Data/Release.pm
+++ b/lib/MusicBrainz/Server/Data/Release.pm
@@ -28,6 +28,7 @@ use MusicBrainz::Server::Data::Utils qw(
 );
 use MusicBrainz::Server::Log qw( log_debug );
 use MusicBrainz::Server::Translation qw( comma_list N_l );
+use MusicBrainz::Server::Validation qw( encode_entities );
 use aliased 'MusicBrainz::Server::Entity::Artwork';
 
 extends 'MusicBrainz::Server::Data::CoreEntity';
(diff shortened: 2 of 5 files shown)


root/components/FieldErrors.js
```diff
@@ -11,23 +11,34 @@ import React from 'react';
...
 import subfieldErrors, {type FieldShape} from '../utility/subfieldErrors';
 
-const buildErrorListItem = (error, index) => (
-  <li key={index}>{error}</li>
-);
+const buildErrorListItem = (error, hasHtmlErrors, index) => {
+  if (hasHtmlErrors) {
+    return (
+      <li
+        dangerouslySetInnerHTML={{__html: error}}
```
yvanzo: It is vulnerable to XSS injection via entity’s name, e.g.
```javascript
</a><script>alert("XSS")</script><a>
```

mwiencek: I'm surprised if `Translation::expand` doesn't HTML-escape the name there. That's a pretty glaring bug which we should address.

But you should also be able to use `expand2react` in place of `dangerouslySetInnerHTML` here, and that does seem safer overall.

yvanzo: Steps to reproduce:
* In a release group, create a new release duplicating an existing release but the recordings associated with the two first tracks have to be swapped in the Recordings tab
* Rename one of these recordings `</a><script>alert("XSS")</script><a>`
* Merge releases and choose strategy “Merge mediums and recordings”

reosarevok: expand2react seems to be affected as well, unless I'm missing something.

reosarevok: This should be fixed with 363b3efa5235e2191a8e0dc6529906dff6a2ede7 on TT, not sure if that will cause issues once it moves to React though?

yvanzo: > expand2react seems to be affected as well, unless I'm missing something.

`expand2react` should only allow tags from a [short list of HTML tags](https://github.com/metabrainz/musicbrainz-server/blob/v-2020-01-20/root/static/scripts/common/i18n/expand2react.js#L49). Doesn’t it? That should prevent `script` tag at least, not sure about roguish links though.

yvanzo: > This should be fixed with [363b3ef](https://github.com/metabrainz/musicbrainz-server/commit/363b3efa5235e2191a8e0dc6529906dff6a2ede7) on TT, not sure if that will cause issues once it moves to React though?

Just tested, it works perfectly for merge error messages at least. It should not have any side-effect as `_link_recording` is only called for merge error messages, not otherwise.

But I would still recommend to remove that `dangerouslySetInnerHTML` if possible.

reosarevok: I did change that to expand2react :) (and yeah, sorry, I was being stupid when I posted about expand2react earlier - I forgot I'm not actually using React on the page with the errors and only changed the file to keep parity!

yvanzo: Alright, I only checked this conversation and the added commit, not the updated one, thanks!

yvanzo: The added commit https://github.com/metabrainz/musicbrainz-server/pull/1361/commits/363b3efa5235e2191a8e0dc6529906dff6a2ede7 is working around a bug in `Translation::expand` as @mwiencek suggested. This bug might affect pages using the TT macro `display_relationship` too.

Since TT macros will be replaced one day, it would be nice to stop using `Translation::expand` here already.

mwiencek: The current code in the PR looks reasonable to me. I do agree that the right thing for `Translation::expand` to do would be to escape the name itself, but I'm afraid we might abuse that in some places to pass HTML inside the link, or are already escaping the HTML outside of the call in some places (so we'd end up double-escaping things by changing it).

yvanzo: > The current code in the PR looks reasonable to me.

It works flawlessly but it also prepares for a future vulnerability as any added `message` in Perl code will have to be checked for correctly encoding entities. Ideally, this task should not be carried out by the data layer.

> I do agree that the right thing for `Translation::expand` to do would be to escape the name itself, but…

I thought it would be to pass fully expandable messages rather than HTML messages from Perl to React/TT and let it deal with encoding entities. Does it seem reasonable? Might be a separate PR.

mwiencek: > I thought it would be to pass fully expandable messages rather than HTML messages from Perl to React/TT and let it deal with encoding entities. Does it seem reasonable?

Well, you'd have to pass back the interpolation arguments too (in this case, recording JSON). So the error message would have to be a JSON document that you parse and pass to `l`/`ln`/`lp`.

This might work with React, but root/release/merge.tt doesn't use React for these error messages. I don't quite understand how you envision it working for TT, since we have to generate HTML at some point in that case.

root/components/FieldErrors.js
```diff
@@ -9,25 +9,34 @@
 
 import React from 'react';
 
+import expand2react from '../static/scripts/common/i18n/expand2react';
 import subfieldErrors, {type FieldShape} from '../utility/subfieldErrors';
 
-const buildErrorListItem = (error, index) => (
-  <li key={index}>{error}</li>
-);
+const buildErrorListItem = (error, hasHtmlErrors, index) => {
```
yvanzo:
```suggestion
// FIXME: Convey expandable object instead of HTML string for safety, see MBS-10632.
const buildErrorListItem = (error, hasHtmlErrors, index) => {
```

root/components/FieldErrors.js
yvanzo: Created MBS-10632 for follow-up.

PR #1362, opened 2020-01-23
MBS-10401 Convert Edit Series edit to React

anirudhjain75:


root/edit/details/EditSeries.js
```diff
@@ -0,0 +1,88 @@
...
+          <td colSpan="2">
+            <EntityLink entity={SeriesEntity} />
+          </td>
+        </tr>
+        <Diff
+          label={l('Name')}
+          newText={name.new}
+          oldText={name.old}
+          split="\s+"
+        />
```
reosarevok:
```suggestion
        {name ? (
          <Diff
            label={l('Name')}
            newText={name.new}
            oldText={name.old}
            split="\s+"
          />
        ) : null}
```

root/edit/details/EditSeries.js
```diff
@@ -0,0 +1,88 @@
...
+        <Diff
+          label={addColonText(l('Disambiguation'))}
+          newText={comment.new}
+          oldText={comment.old}
+          split="\s+"
+        />
```
reosarevok:
```suggestion
        {comment ? (
          <Diff
            label={addColonText(l('Disambiguation'))}
            newText={comment.new}
            oldText={comment.old}
            split="\s+"
          />
        ) : null}
```

reosarevok:
The rest seems fine I think once those changes are made so the edits actually load, but I'll test it once again after that :)

anirudhjain75:
> I'm a bit confused about why you didn't use the actual copy-and-paste suggestions I gave (that are consistent with your previous work in EditPlace) and chose to do this one differently. Is there an actual benefit to this other way? If so, we should change EditPlace to do the same, to keep it consistent. After a quick look, while both ways seem to work, checking before passing it to Diff (as in EditPlace and my suggestion) seems a bit more efficient to me, but if you feel I'm missing something, do let me know!
> 
> Sometimes I get the feeling you don't want to just follow suggestions as-is because you feel it's somehow "cheating", like it is too easy. If so, I totally understand that, but don't forget our goal here should still be to write good code consistently and in a way that's easy to review, even if sometimes that's not really the most challenging :)

Remade the changes, I was trying to do everything inplace. Changed it to previous approach :)

root/edit/details/EditSeries.js
```diff
@@ -0,0 +1,95 @@
...
+      <tbody>
+        <tr>
+          <th>{addColonText(l('Series'))}</th>
+          <td colSpan="2">
+            <EntityLink entity={series} />
+          </td>
+        </tr>
+        {name ? (
+          <Diff
+            label={addColonText(l('Name'))}
```
mwiencek: I fixed some minor issues with the Flow types and indentation, but other than that, the only actual display issue I found was `addColonText` missing here (also fixed).

PR #1363, opened 2020-01-24
MBS-10579: Unbreak es-es/el-gr translations

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10579

The code and associated comment here don't make sense to me. The language used in the Jed files should match what's in DBDefs, since that's what gets set in the lang cookie and is what's used to fetch the file.

I'd also be in favor of changing these to es/el (without country codes) in `MB_LANGUAGES` in production and removing the symlinks - they were to work around a bug in `Locale::Util` that's been fixed for years. But in that case we'd still remove this code as unused.

### DIFF
--- a/webpack.client.config.js
+++ b/webpack.client.config.js
@@ -62,16 +62,6 @@ function langToPosix(lang) {
   return lang.replace(/^([a-zA-Z]+)-([a-zA-Z]+)$/, function (match, l, c) {
     l = l.toLowerCase();
     c = c.toUpperCase();
-
-    /*
-     * Handle symlinks (see under po/). We want to keep the the result here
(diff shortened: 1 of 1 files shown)


mwiencek:
> What was the reason for having country-specific locales here?

They're actually es and el in Transifex. I mentioned why we have the symlinks in the commit message/PR description:

> they were to work around a bug in Locale::Util that's been fixed for years

Basically, `web_set_locale` had a bug where it wouldn't use an es_ES locale on the system if you specified just es.

yvanzo:
Thanks, I was just surprised it affected specifically these two languages and not every other (de, fr, …).

PR #1364, opened 2020-01-24
MBS-10584: Collapse work artists if too many

reosarevok:
MBS-10584

For some works (mostly classical) there can be dozens or hundreds of artists listed, making proper navigation of any work list including them pretty much impossible. This makes it so that only the top four artists are shown, with an option to see all, using the same general idea as MBS-10424 (https://github.com/metabrainz/musicbrainz-server/pull/1242).

For an example where this makes a huge difference:
Screenshot from 2020-02-10 00-17-34

### DIFF
--- a/root/static/scripts/common/components/WorkArtists.js
+++ b/root/static/scripts/common/components/WorkArtists.js
@@ -1,0 +1,101 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React, {useCallback, useState} from 'react';
+
+import hydrate from '../../../../utility/hydrate';
+import {bracketedText} from '../utility/bracketed';
+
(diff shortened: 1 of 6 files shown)


root/static/scripts/common/components/WorkArtists.js
```diff
@@ -0,0 +1,103 @@
...
+type WorkArtistsProps = {
+  +artists: ?$ReadOnlyArray<ArtistCreditT>,
+};
+
+const WorkArtists = ({artists}: WorkArtistsProps) => {
+  const [expanded, setExpanded] = useState<boolean>(false);
+
+  const expand = useCallback(event => {
+    event.preventDefault();
+    console.log('yay!');
```
mwiencek:
```suggestion
```
(though I'm equally excited to review this 😄)

root/static/scripts/common/components/WorkArtists.js
```diff
@@ -0,0 +1,103 @@
...
+    setExpanded(true);
+    console.log(expanded);
```
mwiencek:
```suggestion
```

reosarevok:
Good idea. I moved your CSS for the spacing outside of the parent container since the same formatting makes sense and the IDs are not used elsewhere.

PR #1365, opened 2020-01-26
MBS-7256: Add "Expand all mediums" to release page

reosarevok:
MBS-7256

If a release has, say, 20 mediums, expanding them all by hand is a huge pain. This makes it so that they can all be expanded and collapsed with one click.

Screenshot from 2020-02-25 09-15-27

### DIFF
--- a/root/components/MediumToolbox.js
+++ b/root/components/MediumToolbox.js
@@ -1,0 +1,40 @@
+/*
+ * @flow strict
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+const MediumToolbox = ({
+  hasMultipleMedia,
+}: {hasMultipleMedia: boolean}): React.Element<'span'> => (
+  <span id="medium-toolbox">
+    {hasMultipleMedia ? (
+      <>
+        <button
+          className="btn-link"
+          id="expand-all-mediums"
+          type="button"
+        >
+          {l('Expand all mediums')}
+        </button>
+        {' | '}
+        <button
+          className="btn-link"
+          id="collapse-all-mediums"
+          type="button"
+        >
+          {l('Collapse all mediums')}
+        </button>
+        {' | '}
+      </>
+    ) : null}
+    <button className="btn-link" id="toggle-credits" type="button" />
+  </span>
+);
+
+export default MediumToolbox;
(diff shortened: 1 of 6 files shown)


root/static/styles/layout.less
```diff
@@ -136,6 +136,11 @@ pre code {
 /* For general right aligning, and especially for barcodes */
 .align-right, .barcode-cell { text-align: right; }
 
+#expand-medium-helper {
+    color: @dark-grey;
+    opacity: 0.5;
```
mwiencek: This looks almost invisible on my screen. 😄TBH I don't see a need to change the text color or opacity.

root/components/MediumToolbox.js
```diff
@@ -0,0 +1,34 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
```
mwiencek: Should be fixed per #1418 though.

root/static/scripts/common/MB/release.js
```diff
@@ -61,7 +61,7 @@ $(function () {
       .addClass('loading-message')
       .text(l('Loading...'));
 
-    var mediumId = this.getAttribute('data-medium-id');
+    var mediumId = element.data('medium-id');
 
     $.get('/medium/' + mediumId + '/fragments')
```
mwiencek: The only concern I have here is that if you have a release with say 100 discs, there isn't any sort of rate limiting here that I can see. So we should probably update this to use root/static/scripts/common/utility/request.js too. (Sorry for the quick approval revert, the rest looks great at least. :D)

reosarevok: Ok, changed it to use request.

reosarevok:
> My main concern with the current approach is it adds two very prominent links which are only needed for releases with multiple mediums. At the very least, these should be displayed only when useful.

We don't currently display the "show rels inline / at bottom" only when there are rels, AFAICT. Why would this be different? Plus you can still collapse and expand one-medium releases, and it's more comfortable to do it from these links than from the tiny triangle by the medium, if wanted (I've done it a couple times when I only cared about rels at the bottom or whatnot).

yvanzo:
> We don't currently display the "show rels inline / at bottom" only when there are rels, AFAICT.

It’s a toggle link, so it only displays one over two link titles, but that’s a good suggestion.

reosarevok:
Made it so that the new links only show if a release has at least 2 media.

yvanzo (approved):
Thanks, that is much better.

I would prefer these links to be off when they do nothing (e.g. “expand all” should be off when all mediums are expanded already), but if coding it with a mix of TT/ko/jquery/react is too much hassle, let's postpone that for now.

PR #1366, opened 2020-01-27
Fix MBS-10586 / MBS-10587 / MBS-10588: Release Duplicates medium format fixes

reosarevok:
MBS-10586
MBS-10587
MBS-10588

### DIFF
--- a/root/static/scripts/release-editor/duplicates.js
+++ b/root/static/scripts/release-editor/duplicates.js
@@ -171,7 +171,7 @@ function formatReleaseData(release) {
 
 function combinedMediumFormatName(mediums) {
     var formats = pluck(_(mediums), "format");
-    var formatCounts = formats.countBy(_.identity);
+    var formatCounts = _.countBy(mediums, 'format');
 
     return formats
         .uniq()
--- a/root/static/scripts/release-editor/duplicates.js
+++ b/root/static/scripts/release-editor/duplicates.js
@@ -170,16 +170,16 @@ function formatReleaseData(release) {
 
 
 function combinedMediumFormatName(mediums) {
-    var formats = pluck(_(mediums), "format");
-    var formatCounts = _.countBy(mediums, 'format');
+    const getFormat = medium => medium.format || '';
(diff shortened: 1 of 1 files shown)


root/static/scripts/release-editor/duplicates.js
```diff
@@ -170,16 +170,16 @@ function formatReleaseData(release) {
 
 
 function combinedMediumFormatName(mediums) {
-    var formats = pluck(_(mediums), "format");
-    var formatCounts = formats.countBy(_.identity);
+    const getFormatOrUnknown = medium => medium.format || '(unknown)';
```
mwiencek: `'(unknown)'` is in the mb_server domain, so may not work being passed to `lp_attributes` below. Perhaps just fallback to `''` here and do something like `format ?  lp_attributes(format, 'medium_format') : l('(unknown)')` below?
```suggestion
    const getFormatOrUnknown = medium => medium.format || '';
```

reosarevok: See the commit message: 

> While this isn't currently being translated, it will eventually need to
use "medium_format" context for consistency with the rest. This is
different from the current usage for '(unknown)', but IMO preferable,
since a medium format and, say, a release status might need different
forms of the word "unknown" in different languages.

mwiencek: Sorry! I should've went through the commits first.

Gotcha, but are you suggesting we include it in the attributes domain somehow, or instead wouldn't we want to use `lp('(unknown)', 'medium_format')` at least? Since it also can't be extracted this way either, unless we pass the string directly to `lp` or `lp_attributes`.

mwiencek: Correction, only if we pass it directly to `lp` at present (since we don't extract strings from domain-specific calls).

reosarevok: Left it as l() for now but added MBS-10607 for the issue with translating all 'unknown' as the same.

PR #1367, opened 2020-01-29
Fix development scripts

yvanzo:
* Fix bug spotted by @reosarevok in `update_pot.sh`.
* Add support for `DOCKER_CMD` to allow running `sudo docker`.
* Move `Locale::PO` Perl dependency from production to development.


reosarevok:
Since the stuff this is fixing is still in beta, why not point this at beta? :)

yvanzo:
It doesn’t change anything to beta.musicbrainz.org except removing unused `Locale::PO` module. `master` can probably be updated merged into `beta` once again this evening if needed.

PR #1369, opened 2020-02-03
MBS-10591: Enable Disc IDs tab in release editor

mwiencek:
The EntityTabs component accesses `entity.may_have_discids` to toggle the Disc IDs tab, but this property requires mediums loaded to be computed on the server. Since loading mediums twice isn't a good use of resources, I've change the release editor controller to directly seed release and medium data to the page (still without tracks) rather than requesting it asynchronously.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -75,6 +75,7 @@ after 'load' => sub {
 
     # Load release group
     $c->model('ReleaseGroup')->load($release);
+    $c->model('ReleaseGroupType')->load($release->release_group);
 
     unless ($returning_jsonld) {
         $c->model('ReleaseGroup')->load_meta($release->release_group);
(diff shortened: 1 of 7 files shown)


lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -94,20 +95,14 @@ after 'load' => sub {
     my $cdtoc_count = $c->model('MediumCDTOC')->find_count_by_release($release->id);
     $c->stash->{release_cdtoc_count} = $cdtoc_count;
 
-    # We need to load more artist credits in 'show'
-    if ($c->action->name ne 'show') {
-        $c->model('ArtistCredit')->load($release);
-    }
```
mwiencek: This comment/condition was obsolete. We don't combine `$c->model('ArtistCredit')->load` calls in `show`, at least not since the change to load mediums with lots of discs asynchronously. It just had the exact same line as here, hence why I've removed it.

lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -94,20 +95,14 @@ after 'load' => sub {
...
+    $c->model('ArtistCredit')->load($release, $release->release_group);
+    $c->model('ReleasePackaging')->load($release);
+    $c->model('ReleaseStatus')->load($release);
+    $c->model('Language')->load($release);
+    $c->model('Script')->load($release);
+    $c->model('Release')->load_related_info($release);
```
mwiencek: Mediums are loaded by `load_related_info`. They were already loaded for every action except `edit` (see below).

root/static/scripts/release-editor/init.js
```diff
@@ -268,7 +268,7 @@ releaseEditor.init = function (options) {
     this.seed(options.seed);
 
     if (this.action === "edit") {
-        this.loadRelease(options.gid);
```
mwiencek: `loadRelease` is still used when `action === "add"` to switch between release duplicates.

t/lib/t/MusicBrainz/Server/Controller/Release/Aliases.pm
```diff
@@ -26,6 +26,12 @@ test all => sub {
         },
         'releaseOf' => {
             'albumReleaseType' => 'http://schema.org/AlbumRelease',
+            'byArtist' => {
+                '@id' => 'https://musicbrainz.org/artist/4b585938-f271-45e2-b19a-91c634b5e396',
+                '@type' => ['Person', 'MusicGroup'],
+                'name' => 'Kate Bush',
+            },
+            'creditedTo' => 'Kate Bush',
```
mwiencek: The release group AC is loaded unconditionally now, and happens to be outputted automatically with the release JSON-LD, which seems fine to me. But I could revert it to load the AC conditionally as before if desired.

yvanzo: Unconditionally loading AC with release works for me, it is closely tightened to release info.

mwiencek:
That release works for me. Is the medium format set on yours?

PR #1372, opened 2020-02-07
MBS-10617 : Convert Add work edit to React

anirudhjain75:


root/edit/details/AddWork.js
```diff
@@ -0,0 +1,104 @@
...
+  [key: string]: $ReadOnlyArray<WorkAttributeT>,
+  };
+
+
+const AddWork = ({edit}: {edit: AddWorkEditProps}) => {
+  const display = edit.display_data;
+  const attributes = display.attributes ?? {};
+  const type = display.type;
+  const language = display.language;
+  console.log(display.attributes);
```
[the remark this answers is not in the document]
anirudhjain75: oops forgot to remove the console log

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,104 @@
...
+        {type ? (
+          <tr>
+            <th>{addColonText(l('Type'))}</th>
+            <td>{type.name}</td>
+          </tr>
+        ) : null}
+        {language ? (
+          <tr>
+            <th>{addColonText(l('Language'))}</th>
+            <td>{language.name}</td>
```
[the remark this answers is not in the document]
anirudhjain75: Is l_language a translation of language ?

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,104 @@
...
+        ) : null}
+        {Object.keys(attributes).map((elem) => {
+          return (
+            <tr key={elem}>
+              <th>{addColonText(elem)}</th>
+              <td>
+                <ul>
+                  {attributes[elem].map((attribute) => {
+                    return (
+                      <li key={attribute.typeID}>{attribute.value}</li>
```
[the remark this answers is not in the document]
anirudhjain75: Attribute.id is undefined in many cases, https://www.irccloud.com/pastebin/DABLw0xM/ is the result of console log

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,104 @@
...
+            <td>{language.name}</td>
+          </tr>
+        ) : null}
+        {display.languages.length ? (
+          <tr>
+            <th>{addColonText(l('Lyrics Languages'))}</th>
+            <td>{commaOnlyList(display.languages)}</td>
+          </tr>
+        ) : null}
+        {Object.keys(attributes).map((elem) => {
```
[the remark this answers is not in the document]
anirudhjain75: for consistency I added a sort() to sort based on keys before being mapped upon. that solves the problem of consistencies

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+     +languages: $ReadOnlyArray<string>,
+     +name: string,
+     +type?: WorkTypeT,
+     +work: WorkT,
+   },
+ };
+
+type AttributeT = {
+  [key: string]: $ReadOnlyArray<WorkAttributeT>,
+  };
```
mwiencek: Can you inline this type into `display_data` above? It doesn't appear that we need an alias for it.

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+
+
+const AddWork = ({edit}: {edit: AddWorkEditProps}) => {
```
mwiencek:
```suggestion
const AddWork = ({edit}: {+edit: AddWorkEditProps}) => {
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
+
+import EntityLink from '../../static/scripts/common/components/EntityLink';
+import commaOnlyList from '../../static/scripts/common/i18n/commaOnlyList';
+
+
+ type AddWorkEditProps = {
```
mwiencek: To avoid confusion, I think I'd only include "props" in the name if it's for something passed directly to a react component (not just any type that appears within props).

```suggestion
 type AddWorkEdit = {
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+   ...EditT,
+   +display_data: {
+     +attributes?: AttributeT,
+     +comment?: string,
```
mwiencek: The comment is always included in display_data, but technically can be `null` unless you add `// ''` in `build_display_data`, so I'd suggest doing that.
```suggestion
     +comment: string,
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+     +iswc?: string,
```
mwiencek: Same comment as for `comment`, including adding `// ''`. See the `build_display_data` definition.
```suggestion
     +iswc: string,
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+     +language?: LanguageT,
+     +languages: $ReadOnlyArray<string>,
+     +name: string,
+     +type?: WorkTypeT,
```
mwiencek: That's always included, but can be `null` if the type is unset or can't be loaded. See `build_display_data`.
```suggestion
     +type: WorkTypeT | null,
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+
+
+ type AddWorkEditProps = {
+   ...EditT,
+   +display_data: {
+     +attributes?: AttributeT,
+     +comment?: string,
+     +iswc?: string,
+     +language?: LanguageT,
+     +languages: $ReadOnlyArray<string>,
```
mwiencek: That's included conditionally, like `language`. See `build_display_data`.
```suggestion
     +languages?: $ReadOnlyArray<string>,
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+          <tr>
+            <th>{addColonText(l('Language'))}</th>
+            <td>{language.name}</td>
+          </tr>
+        ) : null}
+        {display.languages.length ? (
+          <tr>
+            <th>{addColonText(l('Lyrics Languages'))}</th>
+            <td>
+              {commaOnlyList(display.languages.map(
```
mwiencek:
```suggestion
              {commaOnlyListText(display.languages.map(
```

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+                language => l_languages(language),
+              ))}
+            </td>
+          </tr>
+        ) : null}
+        {Object.keys(attributes).sort().map((attributeType) => {
+          return (
+            <tr key={attributeType}>
```
mwiencek: You can use the short form ` => (` above and then remove the `return` here. That removes a couple lines and a level of indentation.

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,113 @@
...
+              <th>
+                {addColonText(lp_attributes(
+                  attributeType,
+                  'work_attribute_type',
+                ))}
+              </th>
+              <td>
+                <ul>
+                  {attributes[attributeType].map((attribute) => {
+                    return (
```
mwiencek: Same here re: using an implicit return.

root/edit/details/AddWork.js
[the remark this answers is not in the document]
mwiencek: Unfortunately we don't store the id, but you can use something like `String(attribute.typeID) + '-' + String(attribute.value_id ?? attribute.value)` for the key. The `lp_attributes` change also still needs doing, so marked this as unresolved.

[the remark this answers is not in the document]
mwiencek: But, you should only use `lp_attributes` if the `type_id` is set. Otherwise it's a free-text value that should be displayed as-is.

root/edit/details/AddWork.js
anirudhjain75: Incase of this, I tried creating component with no lyrics language, the output was an empty array in the console log

root/edit/details/AddWork.js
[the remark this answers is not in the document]
anirudhjain75: Hope the changes should work fine

root/edit/details/AddWork.js
mwiencek: That's only for new edits, though. There are thousands of old edits that only have `language_id` in the data, e.g. https://musicbrainz.org/edit/17725386/data or even older ones that have neither, e.g. http://musicbrainz.org/edit/14459931/data

anirudhjain75: ok

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,118 @@
...
+import EntityLink from '../../static/scripts/common/components/EntityLink';
+import {commaOnlyListText} from
+  '../../static/scripts/common/i18n/commaOnlyList';
+
+
+ type AddWorkEdit = {
+   ...EditT,
+   +display_data: {
+     +attributes?: {
+       [key: string]: $ReadOnlyArray<WorkAttributeT>,
```
mwiencek: Let's use a more descriptive name for the key, like `attributeName`.

root/edit/details/AddWork.js
```diff
@@ -0,0 +1,118 @@
...
+          <tr>
+            <th>{addColonText(l('Lyrics Languages'))}</th>
+            <td>
+              {commaOnlyListText(display.languages ? display.languages.map(
+                language => l_languages(language),
+              ) : [])}
+            </td>
+          </tr>
+        ) : null}
+        {Object.keys(attributes).sort().map((attributeType) => (
```
mwiencek: Would also suggest using `attributeName` instead of `attributeType` here to match the type indexer name (and since we often use `fooType` when it's the actual type object, so helps disambiguate things slightly).

PR #1374, opened 2020-02-10
MBS-10625: Block even more smart URLs

yvanzo:
# Do MBS-10625: Block even more smart URLs

* `g.co`
* `pandora.app.link`
* `spoti.fi`
* `sptfy.com`


yvanzo:
Yes we should but that is not a “smart” URL and should be handled separately mainly because of other still welcome Google Play.

PR #1376, opened 2020-02-11
Fix MBS-10622, MBS-10627, MBS-10630 (regressions from MBS-7678)

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10622
https://tickets.metabrainz.org/browse/MBS-10627
https://tickets.metabrainz.org/browse/MBS-10630

### DIFF
--- a/root/components/StaticRelationshipsDisplay.js
+++ b/root/components/StaticRelationshipsDisplay.js
@@ -50,13 +50,12 @@ const StaticRelationshipsDisplay = ({
 
         phraseRows.push(
           <React.Fragment key={targetGroup.key}>
-            {groupSize > 1 &&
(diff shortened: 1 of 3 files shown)


mwiencek:
You can still check these by connecting your local MB instance to pink and loading the entities referenced in the tickets. Here's a screenshot of the changes for MBS-10622.

Screen Shot 2020-02-10 at 3 27 36 PM

mwiencek:
Opened MBS-10630. Also realized it is in fact a regression due to MBS-7678, since it doesn't happen on production. Must've been looking at the wrong commit when I did git blame.

yvanzo:
Parenthesis after ticket key suggested it was a partial fix only, thus GitHub Bot did not transition it as you would expect.

PR #1377, opened 2020-02-11
Reduce noise in website server logs

yvanzo:
Fix a bunch of warnings about uninitialized values in Perl code.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -259,7 +259,7 @@ sub subscribed : Local RequireAuth {
     my ($self, $c) = @_;
 
     my $only_open = 0;
-    if ($c->req->query_params->{open} eq '1') {
+    if (($c->req->query_params->{open} // '') eq '1') {
         $only_open = 1;
     }
 
@@ -279,7 +279,7 @@ sub subscribed_editors : Local RequireAuth {
     my ($self, $c) = @_;
 
     my $only_open = 0;
-    if ($c->req->query_params->{open} eq '1') {
+    if (($c->req->query_params->{open} // '') eq '1') {
         $only_open = 1;
     }
 
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -325,7 +325,10 @@ sub merge
         my $group_type = 2;
         my $orchestra_type = 5;
         my $choir_type = 6;
-        if ($artist_type != $group_type && $artist_type != $orchestra_type && $artist_type != $choir_type) {
+        if (
+            !defined $artist_type ||
+            ($artist_type != $group_type && $artist_type != $orchestra_type && $artist_type != $choir_type)
+        ) {
             push @$merge_columns, 'gender';
         }
         merge_table_attributes(
--- a/lib/MusicBrainz/Server/Edit/Release.pm
+++ b/lib/MusicBrainz/Server/Edit/Release.pm
@@ -15,10 +15,11 @@ sub check_event_countries {
     my $countries = {};
 
     for (@$events) {
-        if (exists $countries->{$_->{country_id} // 'undef'}) {
-            die "Duplicate release country: " . ($_->{country_id} // 'undef');
+        my $country_id = $_->{country_id} // 'undef';
+        if (exists $countries->{$country_id}) {
+            die "Duplicate release country: " . $country_id;
         }
-        $countries->{$_->{country_id} // 'undef'} = 1;
+        $countries->{$country_id} = 1;
     }
 }
 
--- a/lib/MusicBrainz/Server/Edit/Release.pm
+++ b/lib/MusicBrainz/Server/Edit/Release.pm
@@ -15,10 +15,10 @@ sub check_event_countries {
     my $countries = {};
 
     for (@$events) {
-        if (exists $countries->{$_->{country_id}}) {
+        if (exists $countries->{$_->{country_id} // 'undef'}) {
             die "Duplicate release country: " . ($_->{country_id} // 'undef');
         }
-        $countries->{$_->{country_id}} = 1;
+        $countries->{$_->{country_id} // 'undef'} = 1;
     }
 }
 
--- a/lib/MusicBrainz/Server/Edit/Release/EditReleaseLabel.pm
+++ b/lib/MusicBrainz/Server/Edit/Release/EditReleaseLabel.pm
@@ -126,9 +126,10 @@ sub build_display_data {
 
             if (exists $_->{country}) {
                 my $country = $_->{country};
+                my $country_gid_or_id = gid_or_id($country);
 
-                $event_display->{country} = $loaded->{Area}->{gid_or_id($country)} //
-                    (defined($country->{name}) && MusicBrainz::Server::Entity::Area->new($country));
+                $event_display->{country} = defined $country_gid_or_id && $loaded->{Area}->{$country_gid_or_id};
+                $event_display->{country} //= defined $country->{name} && MusicBrainz::Server::Entity::Area->new($country);
             }
 
             $event_display->{date} = MusicBrainz::Server::Entity::PartialDate->new($_->{date});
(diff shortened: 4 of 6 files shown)


lib/MusicBrainz/Server/Entity/LinkAttributeType.pm
```diff
@@ -78,7 +78,7 @@ around TO_JSON => sub {
     return {
         %{ $self->$orig },
         gid => $self->gid,
-        root_id => $self->root_id + 0,
+        root_id => defined $self->root_id ? $self->root_id + 0 : 0,
```
reosarevok: In 6b0254a65c53a3b2ca3fbfce0fdc1d92ad52edc2 you used defined () with parens. I'm ok with either, but we can try and be consistent :)

lib/MusicBrainz/Server/Controller/Edit.pm
```diff
@@ -259,7 +259,7 @@ sub subscribed : Local RequireAuth {
     my ($self, $c) = @_;
 
     my $only_open = 0;
-    if ($c->req->query_params->{open} eq '1') {
+    if (($c->req->query_params->{open} // '') eq '1') {
```
reosarevok: Why use the empty string comparison here instead of using defined like elsewhere?

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -326,6 +326,8 @@ sub collapse_whitespace {
 sub sanitize {
     my $t = shift;
 
+    return '' if not defined $t;
```
reosarevok: Maybe we can call this $t something a bit more descriptive while we're touching it? $text or whatever it wants to be?

lib/MusicBrainz/Server/Entity/MediumCDTOC.pm
```diff
@@ -34,7 +34,7 @@ sub is_perfect_match {
     my @medium_tracks = @{ $self->medium->cdtoc_tracks };
 
     return (@cdtoc_info == @medium_tracks) && all {
-      $_->[0]{length_time} == $_->[1]->length
+      ($_->[0]{length_time} // -1) == ($_->[1]->length // -1)
```
reosarevok: Should it be a perfect match if there's no length_time on both sides? Is that even a possible situation in the first place?

lib/MusicBrainz/Server/Entity/LinkAttributeType.pm
yvanzo: Parenthesis are not needed. Amended 6b0254a65c with d102a748e1 (see below).

lib/MusicBrainz/Server/Controller/Edit.pm
yvanzo: Because it is more concise, and [logicial defined-or operator](https://perldoc.perl.org/perlop.html#Logical-Defined-Or) does test if defined too. Amended other changes to use it if appropriate with d102a748e1, 6fe24002ca,  34a1b6b503, and 17fc7d607a.

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: I would certainly avoid adding more one-letter variables, but this one is already here and doesn’t seem to impair readability as `sanitize` is fairly short and obviously about text.

lib/MusicBrainz/Server/Entity/MediumCDTOC.pm
yvanzo: That was stupid to replace `undef` with `-1` here since `undef` compares to `0` already. It seems more sensible too. Fixed with 095bedfdaa.

lib/MusicBrainz/Server/Entity/Relationship.pm
```diff
@@ -279,9 +279,9 @@ around TO_JSON => sub {
         entity1_credit  => $self->entity1_credit,
         entity0_id      => $self->entity0_id,
         entity1_id      => $self->entity1_id,
-        id              => $self->id + 0,
-        linkOrder       => $self->link_order + 0,
-        linkTypeID      => $link->type_id + 0,
+        id              => ($self->id // 0) + 0,
+        linkOrder       => ($self->link_order // 0) + 0,
+        linkTypeID      => ($link->type_id // 0) + 0,
```
mwiencek: The `linkOrder` change seems fine, though it might be better for us to add `default => 0` to the Moose attribute instead (since that's also the column default at the database level).

For the other two, I'd rather not silence any warnings those are causing without identifying why the `id` and `linkTypeID` are undefined there. I'm guessing it might be from relationships that haven't been created yet or have incomplete info from edit data, but don't know for certain.

If there's a reason for them to be undefined, I do wonder if it'd be safer to output them as `undef` instead of 0 in case we have any checks that assume anything that's a number is a valid ID, e.g. https://github.com/metabrainz/musicbrainz-server/blob/87f0ab5/root/static/scripts/edit/MB/edit.js#L157

lib/MusicBrainz/Server/Entity/LinkAttributeType.pm
```diff
@@ -35,7 +35,7 @@ has 'root' => (
 sub l_name {
     my $self = shift;
     my $rootid = defined $self->root ? $self->root->id : $self->root_id;
-    if ($rootid == $INSTRUMENT_ROOT_ID) {
+    if (defined $rootid && $rootid == $INSTRUMENT_ROOT_ID) {
```
mwiencek: I'd like to understand which pages are triggering this too (I'm guessing edit pages again) - if it's not something we can fix, a short comment here describing when this happens would be nice.

lib/MusicBrainz/Server/Entity/LinkAttributeType.pm
```diff
@@ -78,7 +78,7 @@ around TO_JSON => sub {
     return {
         %{ $self->$orig },
         gid => $self->gid,
-        root_id => $self->root_id + 0,
+        root_id => ($self->root_id // 0) + 0,
```
mwiencek: Same comment as in lib/MusicBrainz/Server/Entity/LinkAttributeType.pm

lib/MusicBrainz/Server/Entity/LinkAttribute.pm
```diff
@@ -25,7 +25,7 @@ sub html {
     my $type = $self->type;
     my $value = encode_entities($type->l_name);
 
-    if ($type->root->id == $INSTRUMENT_ROOT_ID && $type->gid) {
+    if (defined $type->root->id && $type->root->id == $INSTRUMENT_ROOT_ID && $type->gid) {
```
mwiencek: This one seems odd to me too, since if we were able to load `$type->root` I'd think we should have the id...hmm.

yvanzo:
I started with fixing a few minor warnings only, making simple local changes that are unlikely to have global side-effects, but I soon got carried away fixing further warnings and the _simple_ changes soon became closer to hiding wonky stuff rather than actually fixing issues to the root...

I will split this PR into smaller chunks, easier to review, and better documented, just like #1379.

lib/MusicBrainz/Server/Entity/Relationship.pm
yvanzo: Agreed, silencing warnings here without identifying when these happen was just bad.

Note: even `link_order` might have a reason to be undefined: https://github.com/metabrainz/musicbrainz-server/blob/v-2020-02-18/root/edit/details/add_relationship.tt#L8

yvanzo:
Almost done. Most of commits have been investigated, documented, sometimes rewritten, and moved into smaller pull requests, see above mentions/references.

I’m dropping the following commits about relationships/links:

1. https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/60ccc9d9a3f27b39dce50437f00c19ea856dfbf8
2. https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/b65a98e18d0a7f02000941c43976fa4a6df01fa5
3. https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/0b796ccf6a3845dab84ff4fd139f609fbe05e826
4. https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/a996a326447f400f8217175efcc9508be431c386
5. https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/2afed99ff12d18d826513eb3577f8580d9d221a4

plus that other one about sanitize subroutine:

6. https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/95ae81f496ae1f79e4adfaaed7c84ace0f34ca4b

All of these commits wrongly silence warnings whereas these should be further investigated separately to find and solve real issues upstream, that will be for another time.

I will rebase this PR to just keep the rest, the most obvious commits I reworked too.

PR #1381, opened 2020-02-14
Fix noisy warning about autocomplete: uninitialized value in numeric gt (>)

yvanzo:
Fix a not serious but noisy warning (in website server logs) that is triggered by ws/js entity autocomplete.

Note: Since `alias_preference` is an ordered list of user preferred languages, it is expected to not necessarily intersects the set of available languages with primary alias for a result/entity.

It is picked from https://github.com/metabrainz/musicbrainz-server/pull/1377


yvanzo:
Thanks, fixed that typo.

PR #1382, opened 2020-02-14
Ignore local::lib files for docker/js tools too

yvanzo:
This patch completes pull request https://github.com/metabrainz/musicbrainz-server/pull/1319 that added `local::lib` files to the ignore list of Git only.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+.cpanm/
 .dockerignore
 .DS_Store
 blib/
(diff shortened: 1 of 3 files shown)


webpack/watchConfig.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   watch: true,
   watchOptions: {
+    ignored: /.cpanm/,
     ignored: /node_modules/,
+    ignored: /perl_modules/,
```
mwiencek: Same here, Webpack shouldn't look in these unless you tell it to, so no need to ignore

yvanzo:
(That was mostly problematic with `flow` but I tried to cover every ignore list.)
Left babel and webpack out, thanks.

PR #1384, opened 2020-02-17
MBS-10628: Make [no label] really special purpose

reosarevok:
MBS-10628

Unlike Various Artists, [no label] is currently editable, can be merged, etc. All of these are things that we should avoid, so this makes it be much more restricted, in the same way as Various Artists.

For full parity, a trigger should be edited  (schema change): MBS-10647


lib/MusicBrainz/Server/Controller/Label.pm
```diff
@@ -173,4 +176,44 @@ with 'MusicBrainz::Server::Controller::Role::Delete' => {
...
+    }
+};
+
+around _validate_merge => sub {
+    my ($orig, $self, $c, $form) = @_;
+    return unless $self->$orig($c, $form);
+    my $target = $form->field('target')->value;
+    my @all = map { $_->value } $form->field('merging')->fields;
+    if (grep { is_special_label($_) && $target != $_ } @all) {
+        $form->field('target')->add_error(l('You cannot merge a special purpose label into another label.'));
```
mwiencek: Also need `use MusicBrainz::Server::Translation qw( l );` imported for these errors to work.

PR #1387, opened 2020-02-17
MBS-7473: Allow specifying a release MBID when attaching a CD TOC

mwiencek:
If the specified MBID can't have the CD TOC attached, I figured it'd be confusing to show an empty results page, so it shows the requested release with a custom error message.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/CDTOC.pm
+++ b/lib/MusicBrainz/Server/Controller/CDTOC.pm
@@ -261,10 +261,31 @@ sub _attach_list {
             $c->detach;
         }
         elsif ($c->form_submitted_and_valid($search_release, $c->req->query_params)) {
+            my $query = $search_release->field('query')->value;
+            my ($mbid) = $query =~ m/(
+                [\da-f]{8} -
+                [\da-f]{4} -
+                [\da-f]{4} -
+                [\da-f]{4} -
+                [\da-f]{12}
+            )/ax;
             my $releases = $self->_load_paged($c, sub {
-                $c->model('Search')->search('release', $search_release->field('query')->value, shift, shift,
+                if (defined $mbid) {
+                    $c->stash->{was_mbid_search} = 1;
+                    my $release = $c->model('Release')->get_by_gid($mbid);
+                    return [] unless defined $release;
+                    return [
+                        MusicBrainz::Server::Entity::SearchResult->new(
+                            position => 1,
+                            score => 100,
+                            entity => $release,
+                        ),
+                    ];
+                }
+                $c->model('Search')->search('release', $query, shift, shift,
                                             { track_count => $cdtoc->track_count });
             });
+
             my @releases = map { $_->entity } @$releases;
             $c->model('Release')->load_related_info(@releases);
             my @mediums = map { $_->all_mediums } @releases;
--- a/root/cdtoc/attach_filter_release.tt
+++ b/root/cdtoc/attach_filter_release.tt
@@ -6,7 +6,7 @@
     [% USE r = FormRenderer(query_release) %]
     <input type="hidden" name="toc" value="[% toc %]" />
     [% WRAPPER form_row %]
-      [% r.label('query', l('Release title or MBID:')) %]
+      [% r.label('query', add_colon(l('Release title or MBID'))) %]
       [% r.text('query') %]
       [% form_submit(l('Search'), 'inline') %]
     [% END %]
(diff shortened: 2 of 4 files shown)


lib/MusicBrainz/Server/Controller/CDTOC.pm
```diff
@@ -253,10 +254,23 @@ sub attach : Local DenyWhenReadonly
             $c->detach;
         }
         elsif ($search_release->submitted_and_valid($c->req->query_params)) {
+            my $query = $search_release->field('query')->value;
             my $releases = $self->_load_paged($c, sub {
-                $c->model('Search')->search('release', $search_release->field('query')->value, shift, shift,
+                if (is_guid($query)) {
```
reosarevok: Ideally this would check if it contains a GUID like most of our other searches, so you can still paste a release link directly, but this is certainly an improvement.

lib/MusicBrainz/Server/Controller/CDTOC.pm
```diff
@@ -253,10 +254,23 @@ sub attach : Local DenyWhenReadonly
...
+                    my $release = $c->model('Release')->get_by_gid($query);
+                    return [] unless defined $release;
```
reosarevok: I would prefer returning something that specifies that it understood it's an MBID, but it didn't match anything - rather than the standard "No results found. Try refining your search query." it could be something like "We didn't find a release with that MBID" or "That MBID doesn't seem to belong to a release". Not sure what the best way of achieving that would be - I guess you could set a boolean for $was_mbid_search and pass it to the stash?

root/cdtoc/attach_list.tt
```diff
@@ -43,6 +49,13 @@
       </tr>
     </tr>
   [% END %]
+  [% UNLESS attachable_mediums.size %]
+    <tr[% ' class="even"' IF zebra % 2 == 0 %]>
+      <td class="error" colspan="7">
+        [% l('None of the mediums on this release can have the given CD TOC attached; either they have the wrong number of tracks, or the formats don’t support disc IDs.') %]
```
reosarevok: This doesn't actually check for wrong format though. If you pick an MBID for a release with the wrong format it will load it without the error, it'll just be unselectable:

Screenshot from 2020-02-22 19-32-51

lib/MusicBrainz/Server/Controller/CDTOC.pm
mwiencek: Changed it to use a regular expression to extract the MBID, which'll allow people to paste a link.

lib/MusicBrainz/Server/Controller/CDTOC.pm
mwiencek: Was hoping I could reuse the message from root/mbid/NotFound.js, but seems a bit too long and generic. Opted for "We couldn’t find a release matching that MBID." Let me know if that can be improved.

root/cdtoc/attach_list.tt
mwiencek: I changed the message to not mention formats, and also only show this when it's an MBID search now.

mwiencek:
@brainzbot retest this please

reosarevok (approved):
Seems to work fine now.

root/cdtoc/attach_filter_release.tt
```diff
@@ -6,7 +6,7 @@
     [% USE r = FormRenderer(query_release) %]
     <input type="hidden" name="toc" value="[% toc %]" />
     [% WRAPPER form_row %]
-      [% r.label('query', l('Release title:')) %]
+      [% r.label('query', l('Release title or MBID:')) %]
```
yvanzo, line 9: Since it’s a new string:

```suggestion
      [% r.label('query', add_colon(l('Release title or MBID'))) %]
```

root/cdtoc/lookup.tt
```diff
@@ -68,7 +68,7 @@
   <form action="[% c.req.uri_for_action('/cdtoc/attach') %]" method="get">
     [% USE r = FormRenderer(query_release) %]
     <input type="hidden" name="toc" value="[% toc %]" />
-    [% form_row_text(r, 'query', l('Release:')) %]
+    [% form_row_text(r, 'query', l('Release title or MBID:')) %]
```
yvanzo, line 71: Same as above:

```suggestion
    [% form_row_text(r, 'query', add_colon(l('Release title or MBID'))) %]
```

yvanzo (approved):
LGTM but new localizable message, shouldn’t require a second review though.

PR #1388, opened 2020-02-17
MBS-10594: Fail add annotation edit if entity is gone

reosarevok:
MBS-10594


reosarevok:
> It is way too complicated to test by hand since an entity removal has an irreducible voting period of a week

It doesn't in a development server :) I tested it by hand myself.

PR #1389, opened 2020-02-17
Allow number-string equality in assertEditData

mwiencek:
c7d7038e527ba6bdd0699e0ba3a1e878e997b285 broke some of our Selenium tests which expect numeric IDs to not be encoded as strings.

### DIFF
--- a/t/selenium.js
+++ b/t/selenium.js
@@ -53,12 +53,25 @@ const writeCoverage = require('../root/utility/writeCoverage');
 
 const IGNORE = Symbol();
 
-function skipIgnored(a, b) {
-  return (a === IGNORE || b === IGNORE) ? true : undefined;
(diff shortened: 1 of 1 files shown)


mwiencek:
https://github.com/metabrainz/musicbrainz-server/pull/1370 had an alternate fix since I wasn't sure why they were failing at the time. Specifically, https://github.com/metabrainz/musicbrainz-server/pull/1370/commits/64a83667a4010666286454a500ac2dc98a8d3c6f. So we can probably remove that commit from #1370 if this works fine.

mwiencek:
Okay, the "new" fix I had here didn't end up working at all, since the numbers were still being loaded as strings from `restore` even if they were saved in the edit correctly. So I'm just cherry-picking the commit from the other PR here.

PR #1337, opened 2019-12-20
MBS-10322: Drop Fabric

yvanzo:
# Summary

Resolve MBS-10322: Port fabfile

# Problem

`fabfile.py` was written for Fabric 1/Python 2 which is retiring in 2020. It was the only piece of Python in this repository, that doesn’t favor Fabric 2/Python 3 which _is not 100% feature parity_ anyway. It was missing hosts discovery for deployment.

# Solution

Rewriting tasks as independent scripts, with a few differences:
* Scripts are self-documented (with `HELP` string)
* More errors are handled and have detailed messages
* Target hosts are deduced from docker server configs

# TODO

- [x] Adding `PROD_STANDY` to DBDefs example/default
- [x] Adding `MB_POT_DB` to `extract_pot_db`
- [x] Completing `update_pot`to check these parameters

PR #1393, opened 2020-02-19
Fix blog post link in generated git tag message

yvanzo:
It fixes a regression from <https://github.com/metabrainz/musicbrainz-server/pull/1337>.

----

I also fixed blog post links that were broken for all of 2020 [tags](https://github.com/metabrainz/musicbrainz-server/tags).


yvanzo:
Good catch, this has likely changed after the blog moved from mb.o to meb.o.

PR #1394, opened 2020-02-20
Fix ubernoisy warning about artwork image suffix on homepage

yvanzo:
Suffix used to not be loaded on homepage, spamming website logs with this warning:

    Use of uninitialized value in concatenation (.) or string at lib/MusicBrainz/Server/Entity/Artwork.pm line 88

This patch mutes this warning for the homepage only by setting an arbitrary value as suffix.

It’s the 3rd commit picked from https://github.com/metabrainz/musicbrainz-server/pull/1377

----

It doesn’t solve similar warnings triggered on cover art edits’ display which are less frequent.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Release.pm
+++ b/lib/MusicBrainz/Server/Data/Release.pm
@@ -1411,6 +1411,7 @@ sub newest_releases_with_artwork {
         Artwork->new(
             id => $caa_id,
             release => $release,
+            suffix => 'spoof',
         );
     });
 }


mwiencek:
Did the logs give any indication of which pages that occurs on? 'Cause all existing cover artworks have a mime type set, and all mime types have a suffix. So it sounds like we're possibly not loading it somewhere. That could be a bug if `->suffix` is being used as if it's loaded.

yvanzo:
I just counted the number of occurrences per page during the past week. It occurs on the homepage mostly (~740K occurrences) but also on a few editing history pages. See the [exact count per page](https://gist.github.com/yvanzo/1282afb58ff02576f5b32b9ffc40bca6).

yvanzo:
> all existing cover artworks have a mime type set

Indeed, by construction `mime_type` is a [not-nullable foreign key](https://github.com/metabrainz/musicbrainz-server/blob/v-2020-02-18/admin/sql/caa/CreateTables.sql#L29).

> So it sounds like we're possibly not loading it somewhere.

Good catch, since `suffix` is stored in a separate table, it requires additional steps.

yvanzo:
Dropped the previous patch to load `suffix` for newest releases instead.

yvanzo:
(There is work-in-progress already at https://github.com/yvanzo/musicbrainz-server/commit/4857aec3fd190b08050a95ecf5250b436bd23401.)

PR #1395, opened 2020-02-20
Fix noisy warnings about event’s search: uninitialized value $text in substitution (s///)

yvanzo:
* Fix noisy warnings on `/search` and `/ws/js/event` endpoints, see commit message for details.
* Stop exporting `format_setlist` filter to templates, as events are now rendered using React.

It replaces commit https://github.com/metabrainz/musicbrainz-server/pull/1377/commits/4e5b3917c2550159898280cab6bff2a54bbe3864 submitted in #1377

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Event.pm
+++ b/lib/MusicBrainz/Server/Entity/Event.pm
@@ -16,7 +16,7 @@ with 'MusicBrainz::Server::Entity::Role::Type' => { model => 'EventType' };
 
 use MooseX::Types::Structured qw( Dict );
 use MooseX::Types::Moose qw( ArrayRef Object Str );
-use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json non_empty );
 use MusicBrainz::Server::Entity::Util::JSON qw( add_linked_entity );
 use MusicBrainz::Server::Filters qw( format_setlist );
 use MusicBrainz::Server::Types qw( Time );
@@ -109,6 +109,8 @@ around TO_JSON => sub {
     my @related_series = $self->related_series;
     add_linked_entity('series', $_->id, $_) for @related_series;
 
+    my $setlist = $self->setlist;
+
     return {
         %{ $self->$orig },
         areas => [map +{
@@ -124,7 +126,7 @@ around TO_JSON => sub {
             entity => $_->{entity},
         }, $self->all_places],
         related_series => [map { $_->id } @related_series],
-        setlist => format_setlist($self->setlist),
+        (non_empty($setlist) ? (setlist => format_setlist($setlist)) : ()),
         time => $self->formatted_time,
     };
 };
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -56,7 +56,6 @@ __PACKAGE__->config(
             'format_length' => \&MusicBrainz::Server::Filters::format_length,
             'format_wikitext' => \&MusicBrainz::Server::Filters::format_wikitext,
             'format_editnote' => \&MusicBrainz::Server::Filters::format_editnote,
-            'format_setlist' => \&MusicBrainz::Server::Filters::format_setlist,
             'locale' => \&MusicBrainz::Server::Filters::locale,
             'gravatar' => \&MusicBrainz::Server::Filters::gravatar,
             'coverart_https' => \&MusicBrainz::Server::Filters::coverart_https
--- a/root/event/EventIndex.js
+++ b/root/event/EventIndex.js
@@ -31,32 +31,36 @@ const EventIndex = ({
   event,
   numberOfRevisions,
   wikipediaExtract,
-}: Props) => (
-  <EventLayout entity={event} page="index">
-    {eligibleForCleanup ? (
-      <CleanupBanner entityType="event" />
-    ) : null}
-    <Annotation
-      annotation={event.latest_annotation}
-      collapse
-      entity={event}
-      numberOfRevisions={numberOfRevisions}
-    />
-    <WikipediaExtract
-      cachedWikipediaExtract={wikipediaExtract || null}
-      entity={event}
-    />
-    <Relationships source={event} />
-    {event.setlist ? (
-      <>
-        <h2 className="setlist">{l('Setlist')}</h2>
-        <p className="setlist">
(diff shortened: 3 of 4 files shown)


lib/MusicBrainz/Server/Entity/Event.pm
```diff
@@ -124,7 +125,7 @@ around TO_JSON => sub {
             entity => $_->{entity},
         }, $self->all_places],
         related_series => [map { $_->id } @related_series],
-        setlist => format_setlist($self->setlist),
+        setlist => non_empty($setlist) ? format_setlist($setlist) : $setlist,
```
mwiencek: Since we know it's empty in that case, `''` might be better since it could possibly be either `undef` or `''`.
```suggestion
        setlist => non_empty($setlist) ? format_setlist($setlist) : '',
```

yvanzo: That is on purpose because search doesn’t provide the same level of details as lookup.

For example:
* <https://musicbrainz.org/ws/2/event/3c28f7f8-3c69-4007-8375-913a42df7d86> provides `setlist: "..."`
* <https://musicbrainz.org/ws/2/event?query=eid:3c28f7f8-3c69-4007-8375-913a42df7d86&fmt=json> doesn’t provide any setlist field

mwiencek: Oh, for fields that aren't loaded we do generally omit the property so that it's not confused with being unset, so `defined $setlist ? (setlist => format_setlist($setlist)) : (),`. I just noticed the type for `EventT` in root/types.js should be changed to `+setlist?: string,` too.

yvanzo: I initially tried to change `root/types.js` but then remembered that it was like that to work around flow shortcomings, see MBS-10656. Is it still current?

mwiencek: I'm not actually sure what Flow shortcomings that references, but I don't think it should be an issue? You can change it and see, I guess :)

yvanzo: IIRC, the issue was about this part of `EventIndex`:

https://github.com/metabrainz/musicbrainz-server/blob/v-2020-03-02/root/event/EventIndex.js#L50-L57

Flow was not able to deduce from test at line 50 that event.setlist was actually defined at line 54.

mwiencek: Oh, right. That's something Flow does intentionally (invalidates the refinements) if you call any function before using it again. But it's silly for it to do it for `React.createElement`.

We work around that by e.g. assigning the setlist to a `const` variable above and using that in place of `event.setlist`.

mwiencek: Correction, `React.createElement` does not in fact invalidate refinements. In that case it's the `l` call, which seems less silly.

mwiencek: This is documented at https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations BTW.

yvanzo: Thanks. It is presented as a feature, but Flow invalidates such refinements because it has a limited scope while we do know `l` call obviously does not change `event.setlist`. Instead of having a way to instruct Flow that `l` is innocuous once for all, we are required to introduce dozens of code blocks around JSX and a lot more new variables just to make current Flow work, that is a _shortcoming_. Can we wait for a potential better way to resolve MBS-10656 later on?

mwiencek: Oh I agree it's a shortcoming, and would love if Flow had a way to mark functions as pure :) But I also wouldn't let that dictate the format in this case - one additional assignment isn't a big deal and we already do this in tons and tons of places.

yvanzo: > But I also wouldn't let that dictate the format in this case

It actually dictates the code format, dissolving JSX sugar in flow.
If it is about the JSON format, this is already the case, see MBS-10656.

> one additional assignment isn't a big deal and we already do this in tons and tons of places.

It’s not true, we more often don’t do it hence the linked ticket.

I just checked Flow issues, upvoted the feature request for pure functions, and noticed the tremendous amount of bugs reported about type refinements. I was still hoping for Flow improvements on this topic and then found out it is not even part of their [2020 plans](https://medium.com/flow-type/what-were-building-in-2020-bcb92f620c75) they published two days ago…

Let’s unsugar our templates rather than expect Flow to support this feature any soon.

Here is a replacement commit: https://github.com/metabrainz/musicbrainz-server/pull/1395/commits/0b46bd02feb260d1c2de68c43ea8279de7ba14a1.

Note that it removes empty setlist `""` for both ws/js, ws/2, and JSON dump.

mwiencek: > It actually dictates the code format, dissolving JSX sugar in flow.
> If it is about the JSON format, this is already the case, see MBS-10656.

Actually I still don't see how MBS-10656 is related to Flow...it's talking about /ws/2, which was designed far before we ever used Flow, and isn't modeled using Flow types even today. /ws/js is.

And I'm not too concerned about the format we use in /ws/js here TBH, I'd just like the Flow types to match reality (whether as `setlist: string | null` or `setlist?: string`) to prevent bugs where null/undefined are unexpected at runtime.

> It’s not true, we more often don’t do it hence the linked ticket.

No, I meant we do the thing where we cache property accesses to `const` variables in tons of places already, to help the Flow refinement process.

> Note that it removes empty setlist "" for both ws/js, ws/2, and JSON dump.

The /ws/js changes look good, but I'm not sure we want to make these breaking changes to /ws/2 for now...we should probably discuss if it's even worth doing before /ws/3.

yvanzo: > Actually I still don't see how MBS-10656 is related to Flow...it's talking about /ws/2

Right but I wrote it with `root/types.js` for in mind, probably because search mbjson output is returned for both React components and ws/2 search requests. Fixed the last paragraph.

> not sure we want to make these breaking changes to /ws/2 for now...

Reverted those changes in commit https://github.com/metabrainz/musicbrainz-server/pull/1395/commits/ff8886a720b90bfeea8d69390f76656f5cc7df64

> we should probably discuss if it's even worth doing before /ws/3

Fix Version for MBS-10656 is set to ws/3 already.

PR #1397, opened 2020-02-22
MBS-10618 Convert Edit work edit to React

anirudhjain75:


reosarevok:
brainzbot, add to whitelist

reosarevok:
@brainzbot, add to whitelist

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,159 @@
...
+        />
+      ) : null}
+      {iswc ? (
+        <Diff
+          label={addColonText(l('ISWC'))}
+          newText={iswc?.new ?? ''}
+          oldText={iswc?.old ?? ''}
+        />
+      ) : null}
+      {display.type ? (
```
reosarevok: You're already doing const type, no?

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,159 @@
...
+          newText={commaOnlyListText(languages.new.map(
+            language => l_languages(language),
+          ))}
+          oldText={commaOnlyListText(languages.old.map(
+            language => l_languages(language),
+          ))}
+          split=", "
+        />
+      ) : null}
+      {edit.data.new.attributes
```
reosarevok: I know it did check data in TT too, but i'm a bit confused why it would check that - we normally use only data_display. @mwiencek: do you see a clear reason we shouldn't use data_display for this?

root/edit/details/EditWork.js
anirudhjain75: Nope not yet atleast

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,159 @@
...
+            .type?.new?.name ?? '', 'work_type')}
+          oldContent={lp_attributes(display
+            .type?.old?.name ?? '', 'work_type')}
+        />
+      ) : null}
+      {language ? (
+        <FullChangeDiff
+          label={addColonText(l('Language'))}
+          newContent={language.new?.name ?? ''}
+          oldContent={language.old?.name ?? ''}
```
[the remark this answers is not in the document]
anirudhjain75: Still wondering why it wasn't done in the AddWork.js file though if language was also to be translated 🤔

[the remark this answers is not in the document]
reosarevok: Because I missed that you didn't do it (in fact, I left a comment but you seemingly didn't make the change) :) A lot of our old TT edit templates were untranslated, which is bad, but we might as well fix them while we're porting them.

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,157 @@
...
+          oldText={commaOnlyListText(languages.old.map(
+            language => l_languages(language),
+          ))}
+          split=", "
+        />
+      ) : null}
+      {edit.data.new.attributes
+        ? Object.keys(attributes).sort().map((attributeName) => (
+          <tr key={attributeName}>
+            <th>{addColonText(attributeName)}</th>
```
reosarevok: Should be translated, like in AddWork

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,157 @@
...
+            <td className="old">
+              {attributes[attributeName].old.length > 0
+                ? (
+                  <ul>
+                    {attributes[attributeName].old.map((attribute, index) => (
+                      <li key={attribute}>
+                        <DiffSide
+                          filter={DELETE}
+                          newText={attributes[attributeName].new[index] || ''}
+                          oldText={attribute}
```
reosarevok: Should probably be translated?

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,157 @@
...
+          label={addColonText(l('ISWC'))}
+          newText={iswc?.new ?? ''}
+          oldText={iswc?.old ?? ''}
+        />
+      ) : null}
+      {type ? (
+        <FullChangeDiff
+          label={addColonText(l('Work type'))}
+          newContent={lp_attributes(type?.new?.name ?? '', 'work_type')}
+          oldContent={lp_attributes(type?.old?.name ?? '', 'work_type')}
```
reosarevok: You already know type exists, so you probably don't need to do type? here, right? Same for iswc above, etc.

Also, the way you've done this here crashes the edit display (as usual, please do test the stuff locally on the site!). You need to do the same you did on https://github.com/metabrainz/musicbrainz-server/pull/1362/files#diff-fc0b1391e46fee0ae9a544b658540ee3R68

root/edit/details/EditWork.js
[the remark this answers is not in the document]
anirudhjain75: Cool. I'll get it done in sometime

root/edit/details/EditWork.js
anirudhjain75: Sure I did test all the stuff locally. I'll recheck the same. :)

anirudhjain75:
Rebased to current master

root/edit/details/EditWork.js
mwiencek: Nope, I agree we should just check `attributes`/`display.attributes` here. It should only exist if `data.new.attributes` exists already: https://github.com/metabrainz/musicbrainz-server/blob/2997b2a/lib/MusicBrainz/Server/Edit/Work/Edit.pm#L158-L159

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,186 @@
...
+type Props = {
+  ...EditT,
+  +data: {
+    +new: {
+      +attributes?: {
+        +[attributeName: string]: $ReadOnlyArray<WorkAttributeT>,
+      },
+      +languages: $ReadOnlyArray<string>,
+    },
+  },
```
mwiencek: This part can be removed pending https://github.com/metabrainz/musicbrainz-server/pull/1397/files#r382935981

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,186 @@
...
+                  <ul>
+                    {attributes[attributeName].old.map((attribute, index) => (
+                      <li key={attribute}>
+                        <DiffSide
+                          filter={DELETE}
+                          newText={
+                            lp_attributes(
+                              attributes[attributeName].new[index] || '',
+                              'work_attribute_type_allowed_value',
+                            )
```
mwiencek: The attribute names and values are already all translated in the display data (note the use of `l_value` here):
https://github.com/metabrainz/musicbrainz-server/blob/1555e503d2f10a404ec81b0314def70e89eff89e/lib/MusicBrainz/Server/Edit/Work/Edit.pm#L168-L169

So all the `lp_attributes` calls can be removed here.

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,186 @@
...
+        />
+      ) : null}
+      {edit.data.new.attributes
+        ? Object.keys(attributes).sort().map((attributeName) => (
+          <tr key={attributeName}>
+            <th>
+              {addColonText(lp_attributes(
+                attributeName,
+                'work_attribute_type',
+              ))}
```
mwiencek: Same here, this is already translated on the server:
https://github.com/metabrainz/musicbrainz-server/blob/1555e503d2f10a404ec81b0314def70e89eff89e/lib/MusicBrainz/Server/Edit/Work.pm#L23-L23

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,186 @@
...
+            </th>
+            <td className="old">
+              {attributes[attributeName].old.length > 0
```
mwiencek: Since we access `attributes[attributeName]` at least six times, it'd be nice to cache the value as `attributeDiff` or something.

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,186 @@
...
+                ? (
+                  <ul>
+                    {attributes[attributeName].old.map((attribute, index) => (
+                      <li key={attribute}>
+                        <DiffSide
+                          filter={DELETE}
+                          newText={
+                            lp_attributes(
+                              attributes[attributeName].new[index] || '',
```
mwiencek: I'd use `??` instead of `||` here and elsewhere just to be safe. I'm afraid there might be a free-text value like `0` in some cases and Perl might serialize it as a number.

```suggestion
                              attributes[attributeName].new[index] ?? '',
```

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,186 @@
...
+
+const EditWork = ({edit}: {edit: Props}) => {
+  const display = edit.display_data;
+  const comment = display.comment;
+  const iswc = display.iswc;
+  const name = display.name;
+  const type = display.type;
+  const language = display.language;
+  const languages = display.languages;
+  const attributes = display.attributes ?? {};
```
mwiencek: Checking `attributes ?` instead of `edit.data.new.attributes ?` below is probably sufficient so that it doesn't require a fallback.
```suggestion
  const attributes = display.attributes;
```

root/edit/details/EditWork.js
reosarevok: Is there a benefit of doing it like that? I'd much prefer we were consistent and either tried to always translate the stuff already on the Perl layer, or always on React...

mwiencek: Not really, just that it's already written like that so requires less changes. Another minor problem is that the translation should only be done if a `value_id` is set, like `l_value` has, but we don't have the `value_id` here to check that.

root/edit/details/EditWork.js
```diff
@@ -0,0 +1,151 @@
...
+            </td>
+            <td className="new">
+              {attributeDiff.new.length > 0 ? (
+                <ul>
+                  {attributeDiff.new.map((attribute, index) => (
+                    <li key={attribute}>
+                      <DiffSide
+                        filter={INSERT}
+                        newText={attribute}
+                        oldText={attributeDiff.old[index] ?? ''}
```
mwiencek: This was using `attributeDiff.new` instead of `attributeDiff.old`, so additions weren't being highlighted. Pushed a fix for that

PR #1398, opened 2020-02-23
MBS-10659: Always show entity credits in rel edits

reosarevok:
MBS-10659

If one or both entities on the relationship have a credit, we want to still display that credit in the edit relationship edit, even if the edit itself doesn't change the credit (because it is still relevant to know the relationship is under a specific credit). Currently, the edit will display the main artist name at all times except if the credits are changing.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/Edit.pm
@@ -91,6 +91,8 @@ has '+data' => (
         relationship_id => Int,
         type0 => Str,
         type1 => Str,
+        entity0_credit => Optional[Str],
+        entity1_credit => Optional[Str],
         link => find_type_constraint('LinkHash'),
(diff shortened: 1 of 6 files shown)


lib/MusicBrainz/Server/Edit/Relationship/Edit.pm
```diff
@@ -91,6 +91,8 @@ has '+data' => (
         relationship_id => Int,
         type0 => Str,
         type1 => Str,
+        entity0_credit => Optional[Str],
+        entity1_credit => Optional[Str],
```
yvanzo: I’m not sure about using`Optional` over `Nullable` here. In both cases, tests need to be updated.

mwiencek: I think `Optional` makes more sense since those aren't nullable in the DB. (`Nullable` is just defined as `Optional[Maybe[x]]`.)

PR #1400, opened 2020-02-27
MBS-10607: Add context to (unknown) strings

reosarevok:
MBS-10607

We have plenty of "(unknown)" strings, and the thing that is unknown can be one of many. Depending on the thing, though, the translation should vary (for example, a release format is "desconocido" in Spanish but a release itself is "desconocida" - gender variation). This adds context to the different strings to make it possible to translate them separately.

### DIFF
--- a/root/edit/details/add_isrcs.tt
+++ b/root/edit/details/add_isrcs.tt
@@ -2,7 +2,7 @@
   [% IF edit.data.client_version.defined %]
   <tr>
     <th>[% l('Client:') %]</th>
-    <td>[% (edit.data.client_version || l('(unknown)')) | html %]</td>
+    <td>[% (edit.data.client_version || lp('(unknown)', 'isrc client')) | html %]</td>
   </tr>
   [% END %]
 
--- a/root/edit/details/historic/convert_release_to_multiple_artists.tt
+++ b/root/edit/details/historic/convert_release_to_multiple_artists.tt
@@ -9,7 +9,7 @@
         [% END %]
       </ul>
       [% ELSE;
-           l('(unknown)');
+          lp('(unknown)', 'release');
          END %]
     </td>
   </tr>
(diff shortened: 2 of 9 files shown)


root/edit/details/historic/convert_release_to_multiple_artists.tt
```diff
@@ -9,7 +9,7 @@
         [% END %]
       </ul>
       [% ELSE;
-           l('(unknown)');
+          lp('(unknown)', 'release');
```
reosarevok: I'm confused about the original indentation here.

root/edit/details/add_isrcs.tt
```diff
@@ -2,7 +2,7 @@
   [% IF edit.data.client_version.defined %]
   <tr>
     <th>[% l('Client:') %]</th>
-    <td>[% (edit.data.client_version || l('(unknown)')) | html %]</td>
+    <td>[% (edit.data.client_version || lp('(unknown)', 'isrc_client')) | html %]</td>
```
mwiencek: This is probably trivial, but the underscore in the context seems weird to me. We only have underscores elsewhere where it uses the DB table name as context, but this isn't a table name or field, just a phrase. :)
```suggestion
    <td>[% (edit.data.client_version || lp('(unknown)', 'isrc client')) | html %]</td>
```

root/edit/details/historic/convert_release_to_multiple_artists.tt
mwiencek: Looks like it was indented two spaces after the start of the `ELSE` (since the file uses two spaces indentation). Though the file as a whole isn't very consistent.

root/edit/details/add_isrcs.tt
yvanzo: This applies to other context strings too: `medium_forrmat` :arrow_forward: `medium format`, and so on.

reosarevok: Will do! I honestly didn't even know we could have spaces on those :)

PR #1401, opened 2020-02-29
MBS-10666: Collapse long artist lists in work merge page

reosarevok:
MBS-10666

This reuses the React component for work artists rather than having its own implementation, allowing us to use the built-in collapsing options.

I also changed the work writers column to reuse our React component for that - the only difference it makes is showing disambiguations, which seems like unambiguously an improvement.


reosarevok:
I guess it is, yeah! :)

PR #1403, opened 2020-02-29
MBS-10633: Allow changing username capitalization

reosarevok:
MBS-10633

It should be possible to change someone's username to the same username, but with different capitalization. This changes the comparison that checks if the username has changed to allow those
capitalization--only changes, by comparing normalized usernames instead. The registration system doesn't allow two usernames to only differ in capitalization in any case, so this should be perfectly safe.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Utils.pm
+++ b/lib/MusicBrainz/Server/Form/Utils.pm
@@ -206,13 +206,14 @@ sub validate_username {
 
     my $username = $self->value;
     my $previous_username = $self->init_value;
+    my $editor_model = $self->form->ctx->model('Editor');
 
     if (defined $username) {
-        unless (defined $previous_username && $previous_username eq $username) {
+        unless (defined $previous_username && $editor_model->are_names_equivalent($previous_username, $username)) {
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Form/Utils.pm
```diff
@@ -208,7 +208,7 @@ sub validate_username {
     my $previous_username = $self->init_value;
 
     if (defined $username) {
-        unless (defined $previous_username && $previous_username eq $username) {
+        unless (defined $previous_username && uc($previous_username) eq uc($username)) {
```
mwiencek: The only possible problem I see here is that `uc($previous_username) eq uc($username)` in Perl may not be equivalent to `lower(x) = lower(y)` in Postgres. `uc` is locale-aware but our database cluster uses `--no-locale`, so they may not behave the same for Unicode characters...perhaps we should add something like `are_names_equivalent` to `Data::Editor` that just passes both names to a query like `SELECT lower(?) = lower(?)`?

reosarevok: Ok, made that change to be fully safe :)

PR #1405, opened 2020-03-05
MBS-10676: Output MD5SUMS/SHA256SUMS for JSON dumps

mwiencek:


mwiencek:
(There might be some problems with the tests though, still working those out.)

mwiencek:
Made some minor tweaks: `write_checksum_files` now exits early if no compression is set, and I removed some unnecessary `sh -c` bits from the tests, since `system` already passes the input to a shell if only one scalar is given that contains shell metacharacters.

lib/MusicBrainz/Script/MBDump.pm
```diff
@@ -204,6 +204,29 @@ sub write_file {
...
+        : $compression;
+
+    my $output_dir = shell_quote($self->output_dir);
+    for my $hash_program ('md5sum', 'sha256sum') {
+        my $hash_output_file = uc($hash_program . 's');
+        chomp (my $hash_bin = `which g$hash_program` || `which $hash_program`);
+        system "cd $output_dir && $hash_bin --binary *.tar.${tar_ext}" .
+               " | grep -v mbdump-private > $hash_output_file";
+
+        $? == 0 or die "$hash_program returned $?";
```
yvanzo: `$?` matches exit status of `grep` command only. To test exit status of both `$hash_bin` and `grep`, a simple solution is to use `pipefail` option (available in `bash`, `ksh`… but not `dash`). Check https://github.com/metabrainz/musicbrainz-server/pull/1406/commits/1eb925b44cb7aedd2cff7a40a0aee83cde8373f9

PR #1406, opened 2020-03-05
Add compression level to MBDump

yvanzo:
Allow to choose a different trade-off between computing time and dump size when server is busy.

### DIFF
--- a/lib/MusicBrainz/Script/MBDump.pm
+++ b/lib/MusicBrainz/Script/MBDump.pm
@@ -153,10 +153,9 @@ sub make_tar {
     }
 
     log("Creating $tar_file");
-    chomp (my $tar_bin = `which gtar` || `which tar`);
     system
         'bash', '-c',
-            $tar_bin .
+            'tar' .
             ' -C ' . shell_quote($self->export_dir) .
             ($compression ? " --$compression" : '') .
             ' --create' .
--- a/lib/MusicBrainz/Script/MBDump.pm
+++ b/lib/MusicBrainz/Script/MBDump.pm
@@ -145,11 +145,11 @@ sub make_tar {
     my $t0 = [gettimeofday];
     my $output_dir = $self->output_dir;
     my $compression = $self->compression;
-    my $_compression = $compression;
 
-    if ($compression eq 'xz') {
-        $compression = '';
-        $tar_file =~ s/\.xz$//;
+    my $compress_command;
+    if ($compression) {
+        $compress_command = "$compression";
+        $compress_command .= ' --threads=0' if $compression eq 'xz';
     }
 
     log("Creating $tar_file");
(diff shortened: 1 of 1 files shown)


mwiencek:
That looks useful 👍 Though I'm not sure about the refactoring commit, since `-I` appears to be something unrelated in BSD tar (and I don't see an alternative way to change the compression program with it). I can install GNU tar from Homebrew, but it'd be nice to keep compatibility if possible.

yvanzo:
The refactoring commit also makes the archive process to use lesser disk space at once, which might not be negligible to archive recordings search index that takes about 49GB uncompressed.

An alternative might be to run `$tar_bin -c … | $compression_command - > $tar_file`. It seems to be more portable. Would it work for you? Should we also revert your commit https://github.com/metabrainz/musicbrainz-server/commit/89e7740b6d08b30dfc9d81519d916801e8cb3328 for [using `gtar`](
https://github.com/metabrainz/musicbrainz-server/pull/1406/commits/ecb43f1967395339a2ba739821f02298ecda4153#diff-a580c93871cc4cf19cf5e900f2da1bbaL156)?

mwiencek:
> The refactoring commit also makes the archive process to use lesser disk space at once, which might not be negligible to archive recordings search index that takes about 49GB uncompressed.

Yeah, I didn't see what the motivation for the refactoring was at first. Good point. Can you mention that in the commit message while rebasing?

> An alternative might be to run $tar_bin -c … | $compression_command - > $tar_file. It seems to be more portable. Would it work for you? Should we also revert your commit 89e7740 for using gtar?

That'd be great. I think we can mostly revert 89e7740, but not in how it removes the hardcoded, absolute binary paths. But we shouldn't need `which` there either -- `system` should respect the `PATH` already AFAIK.

yvanzo:
Ironically, GNU tar option is actually replaced with GNU bash option…

mwiencek:
Hah. Bash is pretty much already a hard requirement given how many scripts we have with `#!/bin/bash`. And macOS has always included bash (it was the default shell for a long time), so I never had an issue with that.

yvanzo:
This commit switches from a comma-separated list to a single string with shell quotes. The next commit reuses that syntax and reuses parameters to `-I` and `--file` for piped command and output redirect, plus introduces `bash` with `pipefail` option. [Hiding whitespace changes](https://github.com/metabrainz/musicbrainz-server/pull/1406/commits/dbee57a6a7b85c126a802b6be8099e54a4ae3cc3?diff=split&w=1) makes it more visible.

mwiencek:
The overall diff LGTM, but the commits switching to `-I` make the history confused since it's immediately reverted. It can honestly make bisecting a pain, since if you're debugging an issue with the dumps and suddenly you land on the `-I` commit it might break your setup. :P

Just wondering, are you planning to use this in #1407? 'Cause the patch here doesn't expose the compression level as an option for the existing dumps, so I was wondering if you plan to add it as an option there later.

yvanzo:
OK, I reordered commits to not use `-I` option at all, and split some to improve readability.

@mwiencek: It feels weird your setup did not have `gtar` installed since https://github.com/metabrainz/musicbrainz-server/commit/89e7740b6d08b30dfc9d81519d916801e8cb3328. Maybe we should document that GNU versions of `md5sum` and `sha256sum` are required for the full export feature.

And to answer your question, my initial idea was to use MBDump for search indexes, but I gave this idea up for now since MBDump is a bit difficult to extend as-is. So I plan to release search indexes dump without it first, then to rework both JSON dump and search indexes dump to make use of this.

mwiencek:
Thanks. :) I don't like 89e7740b6d08b30dfc9d81519d916801e8cb3328 because it suggests that GNU tar is required there for some reason, while also silently falling back to BSD tar if that's not present (which would break if gtar actually was required). So I'd prefer to either require gtar and document that, or support both, and supporting both seems preferable considering it already works and doesn't require people to install an extra thing or require extra documentation.

PR #1407, opened 2020-03-05
MBS-10546: Export search indexes dump to FTP

yvanzo:
# Add MBS-10546: Dump MB Solr data along with MB DB full export

Bootstrapping a mirror with indexed search requires huge resources.
Solution here is to provide search indexes dump on FTP to start with.

Since search indexes dump must have all entities from latest mbdumps, and is allowed to have more recent entities (that will be replicated), it starts at 4:10 which is expected to be after the end of full export.

SolrCloud collections backup is pretty fast (few minutes).

However compressing 50GB takes some time, thus the dedicated service.

It adds the following five entries to DBDefs:
* SOLRCLOUD_COLLECTIONS_API that is SolrCloud Collections API endpoint.
* SOLRCLOUD_BACKUP_LOCATION that is the absolute path on SolrCloud leader filesystem to save backups to and retrieve these from.
* SOLRCLOUD_RSYNC_BANDWIDTH that is passed to `rsync --bwlimit` to prevent the search service to timeout during backup's download.
* SOLRCLOUD_SSH_CIPHER_SPEC that is passed to `ssh -c` to lower CPU load of retrieving backups. Differences can be significant according to https://blog.famzah.net/2010/06/11/openssh-ciphers-performance-benchmark/
* SEARCH_INDEXES_DUMP_COMPRESSION_LEVEL that is passed to `zstd -` to set compression level with the best trade-off between speed and size.

### DIFF
--- a/admin/BundleSearchIndexesDump
+++ b/admin/BundleSearchIndexesDump
@@ -1,0 +1,211 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+
+use Getopt::Long qw( GetOptions );
+use Pod::Usage qw( pod2usage );
+
+################################################################################
+
+=head1 NAME
+
+BundleSearchIndexesDump - Bundle search indexes dump from SolrCloud backups
+
+=head1 SYNOPSIS
+
+BundleSearchIndexesDump [options]
+
+Options:
+
+    -w, --working-dir DIRECTORY         directory that contains solr backups
+                                        to be turned into search indexes dump
+    -d, --debug                         print more progress messages
+
+    -h, --help                          show this help
+
+Environment:
+
+    BACKUP_STAMP                        timestamp used for backup directory name
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2020 MetaBrainz Foundation
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
+
+################################################################################
+
+my $working_dir;
+my $debug_flag;
+my $help_flag;
+
+GetOptions(
+    "working-dir|w=s"            => \$working_dir,
+    "debug|d"                   => \$debug_flag,
+    "help|h"                    => \$help_flag,
+);
+
+pod2usage() if $help_flag;
+pod2usage(
+    -exitval => 64, # EX_USAGE
+    -message => "$0: unrecognized arguments",
+) if @ARGV;
+
+################################################################################
+
+$SIG{'INT'} = sub { exit 3 };
+
+use File::Copy qw( copy move );
+use POSIX qw( strftime );
+use String::ShellQuote qw( shell_quote );
+
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+
+use DBDefs;
+use MusicBrainz::Script::MBDump qw( gpg_sign );
+
+################################################################################
+
+sub _debug { print "[debug] ", (sprintf shift, @_), "\n" if $debug_flag; }
+sub _info { print "[info] ", (sprintf shift, @_), "\n"; }
+sub _warn { warn "[warning] ", (sprintf shift, @_), "\n"; }
+sub _error { warn "[error] ", (sprintf shift, @_), "\n"; }
+
+my $BACKUP_STAMP = $ENV{BACKUP_STAMP};
+my $COMPRESSION_LEVEL = DBDefs->SEARCH_INDEXES_DUMP_COMPRESSION_LEVEL;
+
+my @backup_dirs;
+my @extraneous_files;
+
+opendir(my $wdh, $working_dir) or die $!;
+for my $filename (sort grep { ! /^\.\.?$/ } readdir $wdh)
+{
+    if ($filename =~ /^backup_${BACKUP_STAMP}_/)
+    {
+        push @backup_dirs, $filename;
+    }
+    else
+    {
+        push @extraneous_files, $filename;
+    }
+}
+closedir $wdh;
+
+die "$working_dir has extraneous files: @extraneous_files\n"
+    if scalar @extraneous_files;
+
+my @public_domain_collections = qw(
+    area
+    artist
+    event
+    instrument
+    label
+    place
+    recording
+    release
+    release-group
+    series
+    url
+    work
+);
+
+my @private_collections = qw(
+    editor
+);
+
+for my $backup_dir (@backup_dirs)
+{
+    my $collection = $backup_dir =~ s/^backup_\d{8}-\d{6}_//r;
+    my $dump_dir = $collection;
+    opendir(my $bdh, "$working_dir/$backup_dir") or die $!;
+    if (grep { /^$collection$/ } @private_collections)
+    {
+        system 'rm', '-fr', "$working_dir/$backup_dir";
+        next;
+    }
+    @extraneous_files = ();
+    for my $filename (sort grep { ! /^\.\.?$/ } readdir $bdh)
+    {
+        push @extraneous_files, $filename
+            if ! $filename =~ 'backup\.properties|snapshot\.shard1';
+    }
+    closedir $bdh;
+    die "$working_dir/$backup_dir has extraneous files: @extraneous_files\n"
+        if scalar @extraneous_files;
+
+    my $working_dump_dir = "$working_dir/$dump_dir";
+    move("$working_dir/$backup_dir", $working_dump_dir) or die $!;
+    move("$working_dump_dir/snapshot.shard1", "$working_dump_dir/index") or die $!;
+
+    my $license_file = (grep { /^$collection$/ } @public_domain_collections)
+        ? "$FindBin::Bin/COPYING-PublicDomain"
+        : "$FindBin::Bin/COPYING-CCShareAlike";
+    copy($license_file, "$working_dump_dir/COPYING") or die $!;
+
+    my $readme_text = <<EOF;
+The files in the directory 'index' are a snapshot of the search index
+'$collection' from musicbrainz.org, in a format suitable for
+use with an Apache Solr server.  To import them, you need a compatible
+version of the MusicBrainz simple Solr search server schema (mbsssss).
+EOF
+    _write_file($working_dump_dir, 'README', $readme_text);
+
+    my $tar_file = "$collection.tar.zst";
+    _info("Creating $tar_file");
+    system
+        'bash', '-c',
+            'set -o pipefail; ' .
+            'tar' .
+            ' -C ' . shell_quote($working_dir) .
+            ' --create' .
+            ' --group=solr:8983' .
+            ' --owner=solr:8983' .
+            ' --remove-files' .
+            ' --verbose' .
+            ' -- ' .
+            shell_quote($dump_dir) .
+            " | zstd -T0 -$COMPRESSION_LEVEL" .
+            ' > ' . shell_quote("$working_dir/$tar_file");
+    if ($? != 0)
+    {
+        _error('Tar returned ' . ($? >> 8));
+        exit 70; # EX_SOFTWARE
+    }
+
+    MusicBrainz::Script::MBDump::gpg_sign("$working_dir/$tar_file");
+}
+
+for my $hash_program ('md5sum', 'sha256sum') {
+    my $hash_output_file = uc($hash_program . 's');
+    chomp (my $hash_bin = `which g$hash_program` || `which $hash_program`);
+    my $hash_command = "cd $working_dir" .
+        " && $hash_bin --binary *.tar.zst > $hash_output_file";
+    system $hash_command;
+    if ($? != 0)
+    {
+        _error('Failed command with ' . ($? >> 8) . ": $hash_command");
+        exit 70; # EX_SOFTWARE
+    }
+
+    MusicBrainz::Script::MBDump::gpg_sign("$working_dir/$hash_output_file");
+}
+
+_info 'Successfully bundled SolrCloud backups as search indexes dump.';
+exit;
+
+################################################################################
+
+sub _write_file
+{
+    my ($parent_dir, $file, $contents) = @_;
+
+    open(my $fh, '>' . $parent_dir . "/$file") or die $!;
+    print $fh $contents or die $!;
+    close $fh or die $!;
+}
--- a/admin/RequestSolrCloudBackups
+++ b/admin/RequestSolrCloudBackups
@@ -1,0 +1,306 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+
+use Getopt::Long qw( GetOptions );
+use Pod::Usage qw( pod2usage );
+
+################################################################################
+
+=head1 NAME
+
+RequestSolrCloudBackups - Request MB SolrCloud leader to backup every collection
+
+=head1 SYNOPSIS
+
+RequestSolrCloudBackups [options]
+
+Options:
+
+    -c, --collection COLLECTION         backup given collection only
+    -d, --debug                         print more progress messages
+
+    -h, --help                          show this help
+
+Environment:
+
+    BACKUP_STAMP                        timestamp used for backup directory name
+                                        (default: date +%Y%m%d-%H%M%S GMT)
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2019 MetaBrainz Foundation
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
+
+################################################################################
+
+my $sample_collection;
+my $debug_flag;
+my $help_flag;
+
+GetOptions(
+    "collection|c=s"            => \$sample_collection,
+    "debug|d"                   => \$debug_flag,
+    "help|h"                    => \$help_flag,
+);
+
+pod2usage() if $help_flag;
+pod2usage(
+    -exitval => 64, # EX_USAGE
+    -message => "$0: unrecognized arguments",
+) if @ARGV;
+
+################################################################################
+
+$SIG{'INT'} = sub { exit 3 };
+
+use JSON::XS qw( decode_json encode_json );
+use POSIX qw( strftime );
+use Readonly;
+use URI;
+use URI::QueryParam;
+
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+
+use DBDefs;
+use MusicBrainz::Server::Context;
+
+################################################################################
+
+sub _debug { print "[debug] ", (sprintf shift, @_), "\n" if $debug_flag; }
+sub _info { print "[info] ", (sprintf shift, @_), "\n"; }
+sub _warn { warn "[warning] ", (sprintf shift, @_), "\n"; }
+sub _error { warn "[error] ", (sprintf shift, @_), "\n"; }
+
+foreach my $def (qw(
+    SOLRCLOUD_COLLECTIONS_API
+    SOLRCLOUD_BACKUP_LOCATION
+))
+{
+    if (!DBDefs->$def)
+    {
+        _error("$def is not set in DBDefs.");
+        exit 78; # EX_CONFIG
+    }
+}
+
+Readonly my $COLLECTIONS_API => DBDefs->SOLRCLOUD_COLLECTIONS_API;
+Readonly my $BACKUP_LOCATION => DBDefs->SOLRCLOUD_BACKUP_LOCATION;
+
+Readonly my $BACKUP_STAMP => $ENV{BACKUP_STAMP} // strftime "%Y%m%d-%H%M%S", gmtime;
+
+my $c = MusicBrainz::Server::Context->create_script_context;
+
+# pending   => Backup request has not been sent yet
+Readonly my $BACKUP_REQUEST_PENDING => 'pending';
+# running   => Solr Cloud is processing backup request
+Readonly my $BACKUP_REQUEST_RUNNING => 'running';
+# completed => Solr Cloud completely processed backup request
+Readonly my $BACKUP_REQUEST_COMPLETED => 'completed';
+
+Readonly my $collections => $sample_collection
+    ? [$sample_collection]
+    : [grep { $_ ne 'editor' } @{ list_solr_collections() }];
+my %backup_requests;
+prepare_backup_requests();
+_info 'Requesting SolrCloud to backup all collections...';
+do_backup_collections();
+wait_backup_complete();
+_info 'Successfully completed backup for all collections on SolrCloud.';
+exit;
+
+################################################################################
+
+sub list_solr_collections
+{
+    my $solr_response = _query_collections_api('LIST');
+    return $solr_response->{collections};
+}
+
+sub prepare_backup_requests
+{
+    my $next_request_id = 1500;
+    my $has_error;
+    foreach my $collection (sort @$collections)
+    {
+        my $request_id = $next_request_id++;
+        $backup_requests{$request_id} = {
+            name => 'backup_' . $BACKUP_STAMP . '_' . $collection,
+            collection => $collection,
+            status => $BACKUP_REQUEST_PENDING,
+        };
+        my $solr_response = _query_collections_api('REQUESTSTATUS', {
+                requestid => $request_id,
+            }
+        );
+        my $solr_request_status = $solr_response->{status}->{state};
+        if ($solr_request_status ne 'notfound')
+        {
+            _error(
+                "A request to SolrCloud with id '%d' is '%s' already: '%s'",
+                $request_id,
+                $solr_request_status,
+                encode_json($solr_response),
+            );
+            $has_error = 1;
+        }
+    }
+    exit 70 if $has_error; # EX_SOFTWARE
+}
+
+sub do_backup_collections
+{
+    foreach my $request_id (sort keys %backup_requests)
+    {
+        my $backup_name = $backup_requests{$request_id}{name};
+        my $collection = $backup_requests{$request_id}{collection};
+
+        _debug (
+            "Sending backup request '%d' of collection '%s' as '%s'.",
+            $request_id,
+            $collection,
+            $backup_name,
+        );
+        my $solr_response = _query_collections_api('BACKUP', {
+                name => $backup_name,
+                collection => $collection,
+                location => $BACKUP_LOCATION,
+                async => $request_id,
+            }
+        );
+        if (defined $solr_response->{error})
+        {
+            _error(
+                "Failed to backup request '%s' of collection '%s' as '%s': " .
+                "Error '%s'",
+                $request_id,
+                $collection,
+                $backup_name,
+                $solr_response->{error},
+            );
+            _flush_running_requests_status();
+            exit 70; # EX_SOFTWARE
+        }
+        $backup_requests{$request_id}{status} = $BACKUP_REQUEST_RUNNING;
+    }
+}
+
+sub wait_backup_complete
+{
+    while (
+        grep { $backup_requests{$_}{status} eq $BACKUP_REQUEST_RUNNING }
+        keys %backup_requests
+    )
+    {
+        my $pause = '60';
+        _debug "Sleeping $pause";
+        sleep $pause;
+        foreach my $request_id (
+            sort
+            grep { $backup_requests{$_}{status} eq $BACKUP_REQUEST_RUNNING }
+            keys %backup_requests
+        )
+        {
+            my $solr_response = _query_collections_api('REQUESTSTATUS', {
+                    requestid => $request_id,
+                }
+            );
+            _debug (
+                "Received: %s",
+                encode_json($solr_response),
+            );
+            if (exists $solr_response->{error}) {
+                _error(
+                    "Failed on backup request '%s' of collection '%s' as '%s': " .
+                    "Error '%s'",
+                    $request_id,
+                    $backup_requests{$request_id}{collection},
+                    $backup_requests{$request_id}{name},
+                    $solr_response->{error},
+                );
+                _flush_running_requests_status();
+                exit 70; # EX_SOFTWARE
+            }
+            elsif ($solr_response->{status}->{state} eq 'completed')
+            {
+                $backup_requests{$request_id}{status} = $BACKUP_REQUEST_COMPLETED;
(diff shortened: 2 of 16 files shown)


admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,307 @@
...
+
+sub _query_collections_api
+{
+    my ($action, $parameters) = @_;
+
+    my $url = $COLLECTIONS_API . '?action=' . $action;
+    foreach my $key (keys %{$parameters})
+    {
+        $url .= '&' . $key . '=' . $parameters->{$key};
+    }
```
mwiencek, line 277: `URI::QueryParam` might be nifty here, e.g. `$u->query_param(action => $action, %{$parameters});`. (You should be able to pass a `URI` object directly to `$c->lwp->get`.)

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,307 @@
...
+            $http_response->status_line,
+        );
+        exit 70; # EX_SOFTWARE
+    }
+
+    my $http_content = $http_response->decoded_content;
+    my $solr_response = decode_json($http_content);
+    if (!defined $solr_response->{responseHeader}
+        || !defined $solr_response->{responseHeader}->{status}
+        || $solr_response->{responseHeader}->{status} ne '0')
```
mwiencek, line 295: Perl will autovivify the keys as needed, so you can leave out the definedness checks for readability if desired
```suggestion
    if (($solr_response->{responseHeader}{status} // '') ne '0')
```

admin/config.default.sh
```diff
@@ -15,6 +15,13 @@ JSON_DUMP_GROUP=musicbrainz
 JSON_DUMP_DIR_MODE=755
 JSON_DUMP_FILE_MODE=644
 
+# Same, but for search indexes dumps.
+SEARCH_INDEXES_DUMP_DIR=/home/musicbrainz/search-indexes-dumps
```
mwiencek, line 19: nitpick: search-indexes-dumps sounds strange because indexes and dumps are both plural. "search-index-dumps" makes more sense to my ear (for referring to a collection of dumps). though "search indexes dump" (as used in admin/RunSearchIndexesDump for example) also makes sense for referring to the dump as a whole (or the singular dump process)

bin/rsync-solrcloud-backups
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+unset SSH_AUTH_SOCK
+
+set -u
+
+DESTINATION="$1"
+
+MB_SERVER_ROOT=$(cd "$(dirname "${BASH_RSYNC_SOLRCLOUD_BACKUPS_DIR[0]}")/../" && pwd)
```
mwiencek, line 9: I think this was intended
```suggestion
MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)
```

docker/templates/Dockerfile.search-indexes-dump.m4
```diff
@@ -0,0 +1,28 @@
...
+    docker/musicbrainz-search-indexes-dump/consul-template.service \
+    /etc/service/consul-template/run
+RUN chmod 755 /etc/service/consul-template/run
+
+COPY docker/musicbrainz-search-indexes-dump/crontab /var/spool/cron/crontabs/musicbrainz
+
+RUN chown musicbrainz:musicbrainz /var/spool/cron/crontabs/musicbrainz && \
+    chmod 600 /var/spool/cron/crontabs/musicbrainz
+
+COPY docker/scripts/start_musicbrainz_server.sh /usr/local/bin/
```
mwiencek, line 24: Does the container require plackup running? If not I'd guess this isn't needed

yvanzo: Indeed, it does not. Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/4cef7200f0d77a0df059af5e94dc752ab34afee9

bin/rsync-solrcloud-backups
yvanzo: Oops, that is most probably due to an hasty search & replace. Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/3c0ba68b0422b5b0056b90f89eed42db9cf6e8fd

admin/config.default.sh
yvanzo: To clarify (just in case):
Search index refer to a specific collection in Solr, for example: `release-group`.
Each dump saves all of the 16 search indexes.
The above directory contains 2 to 3 dumps. (The 2 latest plus the current one.)

Should I still rename it to prevent double plural? (4 occurrences in this PR, 9 in docker server configs)

admin/RequestSolrCloudBackups
yvanzo: Applied in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/b7187807e3552f60c30c97f01310c79d31a44997

admin/RequestSolrCloudBackups
yvanzo: > `URI::QueryParam` might be nifty here

Indeed, that should prevent bad surprises in the future if that code is used with parameter values that need to be escaped. Used in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/4ec2f862faf100fd12d64c962c3b705e3c3e66ef

> e.g. `$u->query_param(action => $action, %{$parameters});`

_Do NOT do this_ according to https://metacpan.org/pod/URI::QueryParam

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,307 @@
...
+}
+
+my $COLLECTIONS_API = DBDefs->SOLRCLOUD_COLLECTIONS_API;
+my $BACKUP_LOCATION = DBDefs->SOLRCLOUD_BACKUP_LOCATION;
+
+my $BACKUP_STAMP = $ENV{BACKUP_STAMP} // strftime "%Y%m%d-%H%M%S", gmtime;
+
+my $c = MusicBrainz::Server::Context->create_script_context;
+
+#my %BACKUP_REQUEST_STATUS = (
```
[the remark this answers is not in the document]
yvanzo: It was supposed to document the hard-coded values `'pending'`, `'running'`, and `'completed'`, but replacing `'pending'` with `BACKUP_REQUEST_STATUS['pending']` and so on does not seem nice either…

Would using `Readonly` instead be alright?
```perl
Readonly our $BACKUP_REQUEST_PENDING = 'pending'; # Backup request has not been sent yet
```

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,307 @@
...
+    -h, --help                          show this help
+
+Environment:
+
+    BACKUP_STAMP                        timestamp used for backup directory name
+                                        (default: date +%Y%m%d-%H%M%S GMT)
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2019 MetaBrainz Foundation
```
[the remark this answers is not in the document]
yvanzo: Started working on it in 2019…

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,210 @@
...
+}
+
+sub _gpg_sign
+{
+    my ($file_to_be_signed) = @_;
+
+    MusicBrainz::Script::MBDump::gpg_sign('dummy', $file_to_be_signed);
+}
+
+## eof BundleSearchIndexesDump
```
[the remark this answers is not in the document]
yvanzo: I just followed conventions from the 31 other existing scripts that also end with such comment. I’m not sure about the reason of this habit. This does not seem to be a documentation tag (not POD at least). Might it be a way to prevent a bug that would happen if the last line of code was not ending with a line break character or something?

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,210 @@
...
+    open(my $fh, '>' . $parent_dir . "/$file") or die $!;
+    print $fh $contents or die $!;
+    close $fh or die $!;
+}
+
+sub _gpg_sign
+{
+    my ($file_to_be_signed) = @_;
+
+    MusicBrainz::Script::MBDump::gpg_sign('dummy', $file_to_be_signed);
```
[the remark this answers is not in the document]
yvanzo: Removed `$self` in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/7e68edce3ca1b4f21b2a3a5f899f7b1af62248ee (commit to be moved ahead of this PR)

Removed `'dummy'` in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/c1931beba20170284d83847e7079ebbf6e2c56cf (commit to be squashed with main commit, like most other commits)

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,210 @@
...
+            'tar' .
+            ' -C ' . shell_quote($working_dir) .
+            ' --create' .
+            ' --group=solr:8983' .
+            ' --owner=solr:8983' .
+            ' --remove-files' .
+            ' --verbose' .
+            ' -- ' .
+            shell_quote($dump_dir) .
+            " | xz --threads=0 -$COMPRESSION_LEVEL" .
```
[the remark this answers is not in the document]
yvanzo: Not really, I choosed `xz` for convenience: because it is already installed with MusicBrainz Server (to handle `json` and `sample` dumps) and `xzcat` is already installed with Solr (Docker Official image). And also because it is very versatile: its fastest compression level is faster than `bzip2` without sacrificing as much as `gzip` and the curve up to default compression level allows to tune it significantly if needed.

If I was to pick a new algorithm up, I would probably look at Zstandard. It seems to be very efficient when it comes to compress a lot of small files such as search indexes.

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,210 @@
...
+    move("$working_dump_dir/snapshot.shard1", "$working_dump_dir/index") or die $!;
+
+    my $license_file = grep { /^$collection$/ } @public_domain_collections
+        ? "$FindBin::Bin/COPYING-PublicDomain"
+        : "$FindBin::Bin/COPYING-CCShareAlike";
+    copy($license_file, "$working_dump_dir/COPYING") or die $!;
+
+    my $readme_text = <<EOF;
+The files in this directory are a snapshot of the search index for
+'$collection' in the MusicBrainz database, in a format suitable for
```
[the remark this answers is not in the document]
yvanzo: This is supposed to be a _search index_, _MusicBrainz database_ had a general meaning:

Committed https://github.com/metabrainz/musicbrainz-server/commit/f5583dda4f2dcf36a82f6ffe3c38060410ba3539 that fixes both:
* Files actually are in directory `index`, the current one just contains metadata.
* Replaced the misleading _MusicBrainz database_ with a general _official MusicBrainz_.

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,210 @@
...
+
+for my $backup_dir (@backup_dirs)
+{
+    my $collection = $backup_dir =~ s/^backup_\d{8}-\d{6}_//r;
+    my $dump_dir = $collection;
+    opendir(my $bdh, "$working_dir/$backup_dir") or die $!;
+    for my $filename (sort grep { ! /^\.\.?$/ } readdir $bdh)
+    {
+        push @extraneous_files, $filename
+            if ! $filename =~ "backup.properties|snapshot.shard1";
```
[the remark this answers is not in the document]
yvanzo: Fixed and used single quotes in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/b5c591df79df5d2ad08b75fa1a265903792d416b

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,307 @@
...
+
+use DBDefs;
+use MusicBrainz::Server::Context;
+
+################################################################################
+
+sub _debug { print "[debug] ", (sprintf shift, @_), "\n" if $debug_flag; }
+sub _info { print "[info] ", (sprintf shift, @_), "\n"; }
+sub _warn { warn "[warning] ", (sprintf shift, @_), "\n"; }
+sub _error { warn "[error] ", (sprintf shift, @_), "\n"; }
```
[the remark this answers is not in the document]
yvanzo: See https://github.com/metabrainz/musicbrainz-server/pull/1407#discussion_r404587141 for follow-up

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,210 @@
...
+
+use DBDefs;
+use MusicBrainz::Script::MBDump;
+
+################################################################################
+
+sub _debug { print "[debug] ", (sprintf shift, @_), "\n" if $debug_flag; }
+sub _info { print "[info] ", (sprintf shift, @_), "\n"; }
+sub _warn { warn "[warning] ", (sprintf shift, @_), "\n"; }
+sub _error { warn "[error] ", (sprintf shift, @_), "\n"; }
```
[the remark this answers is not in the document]
yvanzo: Yes.

I initially defined these in Bash, then converted scripts in Perl. I noticed that most Perl scripts were using `print` instead, so I preferred to stick with proper logs. Just reported MBS-10750.

There are a pair of exceptions though, as in [RebuildCoverArt](https://github.com/metabrainz/musicbrainz-server/blob/v-2020-03-31/lib/MusicBrainz/Script/RebuildCoverArt.pm#L80).

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+sub wait_backup_complete
+{
+    while (
+        sum
+        map { $_ => 1 }
+        grep { $backup_requests{$_}{status} eq 'running' }
+        keys %backup_requests
```
mwiencek, line 199: Not sure I understand what this is doing. `sum map { $_ => 1 }` will sum all the keys and values together, but what does that mean?

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: Yeah it'd be nice to have a way to set that from the script, short of configuring `LOGGER_ARGUMENTS` in the container.

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: Cool. Comparing xz with zstd or lz4 here would be very interesting indeed.

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: Not that I'm aware of. Maybe it provides some sort of indication of where the files start/end if you cat them together. Anyway, it doesn't harm anything, just seems a bit cargo-cultish if we can't find a reason for them. :P

admin/RequestSolrCloudBackups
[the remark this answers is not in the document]
mwiencek: Ah okay, I didn't realize it was documentation (thought it might be unused code). The readonly constants sound fine, or maybe just turning the comment here into prose so it doesn't look like dead code.

admin/RequestSolrCloudBackups
mwiencek: Lol good call. That's what happens when you skim the examples and don't see the giant NOT in front of them. :D

admin/config.default.sh
mwiencek: Well, if it were up to me I'd rename it just 'cause it sounds ungrammatical, but I don't have a technical reason to force it. "Search index dumps" could mean any number of indexes per dump, so it's fine in that respect.

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
yvanzo: I did a bit of Git-archeology and found out that the very first occurrence of `# eof` has been introduce by Dave Evans with a new file that also contained a Vi modeline (https://github.com/metabrainz/musicbrainz-server/commit/ebabbd3d3fbbc6e4bf1a9667a476abef8abba43d). He added 75 occurrences (mostly with new files) until Rob started to follow (https://github.com/metabrainz/musicbrainz-server/commit/227ae8a2dd3af7ec6e62ee49e7c2184e3be26efe). Since then, it has been followed by most other contributors including you added the last occurrence (https://github.com/metabrainz/musicbrainz-server/commit/bc0a4306b9e0ade5ae2e09c51aa71c0774e0bcbe).

I thought it could related to either Vi or Vi plugin, but your hypothesis seems plausible too. These first comments have been added very early, at a time when Plackup and Catalyst were not used yet. Might it be some files were concatenated by the toolchain used in those days?

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,203 @@
...
+    move("$working_dump_dir/snapshot.shard1", "$working_dump_dir/index") or die $!;
+
+    my $license_file = grep { /^$collection$/ } @public_domain_collections
+        ? "$FindBin::Bin/COPYING-PublicDomain"
+        : "$FindBin::Bin/COPYING-CCShareAlike";
+    copy($license_file, "$working_dump_dir/COPYING") or die $!;
+
+    my $readme_text = <<EOF;
+The files in the directory 'index' is a snapshot of the search index
+'$collection' from the official MusicBrainz, in a format suitable for
```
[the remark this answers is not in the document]
yvanzo: Or just `from musicbrainz.org`?

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,203 @@
...
+    move("$working_dir/$backup_dir", $working_dump_dir) or die $!;
+    move("$working_dump_dir/snapshot.shard1", "$working_dump_dir/index") or die $!;
+
+    my $license_file = grep { /^$collection$/ } @public_domain_collections
+        ? "$FindBin::Bin/COPYING-PublicDomain"
+        : "$FindBin::Bin/COPYING-CCShareAlike";
+    copy($license_file, "$working_dump_dir/COPYING") or die $!;
+
+    my $readme_text = <<EOF;
+The files in the directory 'index' is a snapshot of the search index
```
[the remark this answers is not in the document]
yvanzo: Re-oops, reverted in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/95097a00d8f1dc3875d174f4d8914f1d56dba5e8

admin/RequestSolrCloudBackups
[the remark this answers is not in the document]
yvanzo: Used `Readonly` in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/b5c1b88f54fc81b8cca67e550c548e9576bef171 to better distinguish status in script from status in SolrCloud.

admin/RequestSolrCloudBackups
yvanzo: It just counts backup requests that are still running.

admin/config.default.sh
yvanzo: Renamed in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/e10fb6bbfdf06687cb119e9769110fa1b1ddc1c6

Please let me know if these changes are alright before I do the same in docker server configs and mb solr cloud.

admin/RequestSolrCloudBackups
mwiencek: Oh. That makes sense now, but just `while (grep { $backup_requests{$_}{status} eq 'running' } keys %backup_requests)` should suffice unless I'm missing something.

admin/config.default.sh
mwiencek: Yes, that looks alright to me 👍

yvanzo: Thanks, changed docker server configs PR, no change needed in mb solr cloud.

admin/RequestSolrCloudBackups
yvanzo: Indeed, removed count in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/728b7f7945abe38ffa1d55cd372c8aa5157c8341.

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
yvanzo: See https://github.com/metabrainz/musicbrainz-server/pull/1407#discussion_r404981901 for follow-up

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: That works for me. I think people will know it is meant in any case :)

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: Yeah, could be some sort of vi-ism, though I'm not sure I've seen this in the wild outside of MBS before either. I think we can agree it's not material to the execution of the script at least. :) I'll probably not add it to future scripts I write, personally.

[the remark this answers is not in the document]
yvanzo: @mwiencek: Removed it and reformatted comments in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/1913dad04431f7b0d2db7fc7770e92f0e536970b

admin/BundleSearchIndexesDump
```diff
@@ -189,7 +193,7 @@ for my $hash_program ('md5sum', 'sha256sum') {
 _info 'Successfully bundled SolrCloud backups as search indexes dump.';
 exit;
 
-################################################################################
+##########                          '-.((*_.-'                        ##########
```
mwiencek, line 196: Am I supposed to know what these hieroglyphs do? 😆"Make separating comments more useful" doesn't explain anything.

mwiencek: LOL! I've decoded the Easter surprise, well played. 😆

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
yvanzo: Removed for good in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/a71056236d34055f4cc9cacad7767556ebe655f9

admin/cron/daily-search-indexes-dump.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -eu
+
+MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../../" && pwd)
+cd "$MB_SERVER_ROOT"
+
+# Only do this on the nominated days (0=Sun 6=Sat)
+if date +%w | grep -q [36]
```
reosarevok, line 9: Maybe it should be 3=Wed 6=Sat instead? :)

yvanzo: The comment gives the full range of week days returned by `date +%w` assuming `C` locale.
It differs from other locales/systems that may use `(0=Mon 6=Sun)` or `(1=Sun 7=Sat)`.

reosarevok: I understood that, but why make people calculate what 3 is then? They don't need to know what Sunday is, just what the ones we use are :)

yvanzo: People need to know the full range to set another value correctly, the ones we use are not enough, for example the full range could also be `(1=Mon 7=Sun)` (as with `date +%u`).
Rephrased with both used values and smallest allowed value in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/3f8bd723029f907af61c12904488b01f2aa5ea8e.

admin/config.search-indexes-dump.sh
```diff
@@ -0,0 +1,5 @@
+RSYNC_FULLEXPORT_HOST='10.2.2.40'
+RSYNC_FULLEXPORT_DIR="$FTP_DATA_DIR/search-indexes"
```
mwiencek, line 2: Shouldn't `SEARCH_INDEX_DUMP_DIR` be set to here, not /home/musicbrainz/search-index-dumps? 'Cause right now admin/RunSearchIndexesDump dumps to `$SEARCH_INDEX_DUMP_DIR`, but then bin/rsync-fullexport-files is set to copy files from here, which would be empty unless I'm missing something.

mwiencek: Or alternatively, the other way around: set `$RSYNC_FULLEXPORT_DIR` to `$SEARCH_INDEX_DUMP_DIR` or some subdirectory thereof. I can't find where it copies anything to `$FTP_DATA_DIR` at least.

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,305 @@
...
+                    "Error '%s'",
+                    $request_id,
+                    $backup_requests{$request_id}{collection},
+                    $backup_requests{$request_id}{name},
+                    $solr_response->{error},
+                );
+                _flush_running_requests_status();
+                exit 70; # EX_SOFTWARE
+            }
+            elsif ($solr_response->{status}->{state} eq 'completed')
```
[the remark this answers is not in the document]
yvanzo: No because this occurrence is rather defined by SolrCloud Collections API.
See above https://github.com/metabrainz/musicbrainz-server/pull/1407#discussion_r405315278.

admin/RequestSolrCloudBackups
```diff
@@ -0,0 +1,305 @@
...
+            {
+                $backup_requests{$request_id}{status} = $BACKUP_REQUEST_COMPLETED;
+                _query_collections_api('DELETESTATUS', {
+                        requestid => $request_id,
+                    }
+                );
+            }
+            elsif ($solr_response->{status}->{state} ne 'running')
```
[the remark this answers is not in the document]
yvanzo: Same as above.

admin/RunSearchIndexesDump
```diff
@@ -0,0 +1,65 @@
...
+# This is to help with disk space monitoring - run "df" before and after
+echo "Disk space when RunSearchIndexesDump starts:" ; df -m
+trap 'echo "Disk space when RunSearchIndexesDump ends:" ; df -m' 0
+
+MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)
+cd "$MB_SERVER_ROOT"
+
+source admin/config.sh
+
+. ./admin/functions.sh
```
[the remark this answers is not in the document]
yvanzo: This is just the same as in all other admin scripts. I would rather change this in a later PR.

admin/RequestSolrCloudBackups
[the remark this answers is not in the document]
mwiencek: Ok, I thought you intended to use the exact same constants as there (since they are the same, and the doc comments referenced Solr Cloud)

admin/RunSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: Sure. I would've even preferred https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/3f8bd723029f907af61c12904488b01f2aa5ea8e be separate though I tested those changes and it LGTM.

admin/config.search-indexes-dump.sh
yvanzo: Right, my bad, fixed in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/b326eb240c2abff6ccbc72eb81ced42ee7c13db1.

Also fixed rsync remote host in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/bd118dd1b8f9c94254a1e65e8dc99e80dbb3f0ec as it doesn’t need to be different from the default (williams, also the target for deployment which has enought free disk space).

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
yvanzo: So I tested these 3 compression algorithms with real data (work search index) and Zstandard is by far the best compromise:

| algorithm / level | time | space savings |
| -- | -- | -- |
| lz4 / 1 | 3.65 | 89.93% |
| lz4 / 2 | 3.60 | 89.93% |
| lz4 / 3 | 56.34 | 76.54% |
| lz4 / 4 | 63.94 | 76.04% |
| lz4 / 5 | 68.01 | 75.88% |
| lz4 / 6 | 71.25 | 75.81% |
| xz / 0 | 307.24 | 62.39% |
| xz / 1 | 493.05 | 60.14% |
| xz / 2 | 795.86 | 59.06% |
| xz / 3 | 1299.95 | 58.03% |
| zstd / 1 | 11.40 | 66.59% |
| zstd / 2 | 19.64 | 64.87% |
| zstd / 3 | 31.75 | 63.90% |
| zstd / 4 | 34.64 | 63.41% |
| zstd / 5 | 69.01 | 62.43% |
| zstd / 6 | 92.93 | 61.47% |

Changed to use `zstd` in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/bfcaa355b5216f3fa5abb32e4208de4e9d4c41be (and https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/15631dbb768e9cf10d70d7b73b125a83ef07c03f)

yvanzo:
Also prevented dumping/publishing `editor` search index has it contains `bio` field we don’t export in database dump, and there is no easy way to filter it out.

admin/BundleSearchIndexesDump
```diff
@@ -0,0 +1,211 @@
...
+my @private_collections = qw(
+    editor
+);
+
+for my $backup_dir (@backup_dirs)
+{
+    my $collection = $backup_dir =~ s/^backup_\d{8}-\d{6}_//r;
+    my $dump_dir = $collection;
+    opendir(my $bdh, "$working_dir/$backup_dir") or die $!;
+    if grep { /^$collection$/ } @private_collections
```
mwiencek, line 127: Don't think you can leave off the parentheses outside of the postfix form (at least it threw a compile error when I tried this in a REPL)
```suggestion
    if (grep { /^$collection$/ } @private_collections)
```

admin/BundleSearchIndexesDump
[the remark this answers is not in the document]
mwiencek: Nice! Glad you found time to measure that, sounds like it should make the process much more efficient.

admin/BundleSearchIndexesDump
yvanzo: Re-oops, fixed in https://github.com/metabrainz/musicbrainz-server/pull/1407/commits/774b15037f940cb3c83a6e599dbca1bd5b249e79.

yvanzo:
Thanks, pushed to `test` branch for now.

yvanzo:
ftp://ftp.eu.metabrainz.org/pub/musicbrainz/data/search-indexes/ :tada:

PR #1408, opened 2020-03-06
MBS-10679 / MBS-10680: WS detail tab improvements

reosarevok:


root/entity/Details.js
```diff
@@ -21,10 +21,11 @@ type Props = {
...
 
-const XMLLink = ({
+const WSLink = ({
   entityGid,
   entityProperties,
   entityType,
+  isJson,
   isSecureConnection,
 }) => {
   const xmlInc = [];
```
mwiencek: Should probably rename this to just `inc` too :P

root/entity/Details.js
```diff
@@ -89,10 +91,23 @@ const Details = ({
         <tr>
           <th>{l('XML:')}</th>
           <td>
-            <XMLLink
+            <WSLink
               entityGid={entity.gid}
               entityProperties={entityProperties}
               entityType={entityType}
+              isJson={false}
```
yvanzo: Probably unneeded? Or if the idea is to highlight this is XML, maybe replace `isJson` with `format` and use `format={'xml'}` instead here?

reosarevok: IIRC Flow complained without it? Can check when I get home

root/entity/Details.js
```diff
@@ -16,28 +16,37 @@ import EntityLink from '../static/scripts/common/components/EntityLink';
 import chooseLayoutComponent from '../utility/chooseLayoutComponent';
 import formatUserDate from '../utility/formatUserDate';
 
-type Props = {
+type WSLinkProps = {
+  +entityGid: string,
+  +entityProperties: {[property: string]: any},
```
reosarevok: We don't have a flow type for this (ENTITIES[type]) yet, right? Should we have a more detailed one?

PR #1412, opened 2020-03-09
Upgrade leaflet

reosarevok:
On top of https://github.com/metabrainz/musicbrainz-server/pull/1410

We're running a 5 year old version of leaflet. Upgraded to the latest version, upgraded the markercluster plugin too, checked that maps for areas and places seem to work just fine still.

The yarn process did complain that _"leaflet.markercluster@1.4.1" has incorrect peer dependency "leaflet@~1.3.1"_ but it does seem to work fine with leaflet 1.6 as well - that plugin is [looking for maintainers](https://github.com/Leaflet/Leaflet.markercluster/issues/979) though and I'm not sure whether we should be looking for a replacement eventually.


yvanzo:

PR #1423, opened 2020-03-14
MBS-10391 Convert Add Release Group edit to React

anirudhjain75:


root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,85 @@
...
+            <tr>
+              <th>{l('Primary Type:')}</th>
+              <td>{type.name}</td>
+            </tr>
+          ) : null}
+
+          {display.secondary_types ? (
+            <tr>
+              <th>{l('Secondary Types:')}</th>
+              <td>{display.secondary_types}</td>
```
reosarevok: These should also be translated right now, but that's a bit trickier, since display_data only gets a string joining the names. We should probably build the string here (using lp_attributes to translate each type name), and change Edit::ReleaseGroup::Create to pass an array of types or an array of type names instead. @mwiencek: what do you think?

mwiencek: Hmm, I'd probably just change the `->name` to `->l_name` in display data for now. I get that we want a consistent place to do the translations, but we don't really have any plans to remove the Perl i18n later any time soon either, so it's just avoiding unnecessary work for now to get the core React conversion merged.

root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,85 @@
...
+
+import DescriptiveLink from
+  '../../static/scripts/common/components/DescriptiveLink';
+import ExpandedArtistCredit from
+  '../../static/scripts/common/components/ExpandedArtistCredit';
+
+type AddReleaseGroupProp = {
+  ...EditT,
+  +display_data: {
+      +artist_credit: ArtistCreditT,
```
mwiencek: This block is indented a couple spaces too many
```suggestion
    +artist_credit: ArtistCreditT,
```

root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,85 @@
...
+      +comment?: string,
+      +name: string,
+      +release_group: ReleaseGroupT,
+      +secondary_types?: string,
+      +type?: ReleaseGroupTypeT,
+  },
+};
+
+const AddReleaseGroup = ({edit}: {edit: AddReleaseGroupProp}) => {
```
mwiencek:
```suggestion
const AddReleaseGroup = ({edit}: {+edit: AddReleaseGroupProp}) => {
```

root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,85 @@
...
+import DescriptiveLink from
+  '../../static/scripts/common/components/DescriptiveLink';
+import ExpandedArtistCredit from
+  '../../static/scripts/common/components/ExpandedArtistCredit';
+
+type AddReleaseGroupProp = {
+  ...EditT,
+  +display_data: {
+      +artist_credit: ArtistCreditT,
+      +comment?: string,
```
mwiencek: `foo?: ...` means `foo` can be missing from the object, whereas this is always returned: https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm#L60

```suggestion
      +comment: string,
```

root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,85 @@
...
+      +name: string,
+      +release_group: ReleaseGroupT,
+      +secondary_types?: string,
```
mwiencek: This is always returned, too: https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm#L65
```suggestion
      +secondary_types: string,
```

root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,85 @@
...
+      +type?: ReleaseGroupTypeT,
```
mwiencek: Let's fix https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm#L61 to return `undef` instead of `''`, then tweak the type here too:
```suggestion
      +type: ReleaseGroupTypeT | null,
```

root/edit/details/AddReleaseGroup.js
```diff
@@ -0,0 +1,91 @@
...
+                {lp_attributes(type.name, 'release_group_primary_type')}
+              </td>
+            </tr>
+          ) : null}
+
+          {secondaryType ? (
+            <tr>
+              <th>{l('Secondary Types:')}</th>
+              <td>
+                {lp_attributes(secondaryType, 'release_group_secondary_type')}
```
reosarevok, line 81: As you yourself said, you didn't need to add translation stuff here :) (it's coming translated from Perl). Otherwise this looks good to me!

anirudhjain75: Ohh I forgot and added these beforehand, I'll revert this :)

anirudhjain75:
That should work up on all the comments :)

PR #1431, opened 2020-03-20
MBS-10364 (I): Convert merge pages (except release) to React

reosarevok:
This converts all entity merge pages to React, except release/merge.tt which is much more complicated and long so will take longer and would make the PR way too big. 

Also fixes MBS-10666


root/work/WorkMerge.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const WorkMerge = ({$c, form, toMerge}: Props) => (
+  <Layout fullWidth title={l('Merge works')}>
+    <div id="content">
+      <h1>{l('Merge works')}</h1>
+      <p>
+        {l(`You are about to merge the following works into a single
+            work. Please select the work which you would
+            like other works to be merged into:`)}
```
mwiencek: This changes the string FWIW. It makes it consistent with the other entities, though the old wording actually sounds better to me (this sounds like an article is missing, and the "to be" is awkward). But it's not wrong either, so it might be better to make them consistent, since both versions sound like bad tongue twisters anyway.

root/static/scripts/common/utility/sortByEntityName.js
```diff
@@ -0,0 +1,16 @@
...
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import sortBy from 'lodash/sortBy';
+
+export default function sortByEntityName(
+  entities: $ReadOnlyArray<CoreEntityT>,
+): $ReadOnlyArray<CoreEntityT> {
+  return sortBy(entities, [entity => entity.name.toLowerCase()]);
```
mwiencek: If there's only one sort function, you can pass it directly rather than as a single-item list. Although...I also wonder if we can just perform this sorting on the server instead, where it does `@entities = nsort_by { $_->id == $target ? 0 : 1 } @entities;` in `Controller::Role::Merge` to be precise, by sticking a `sort_by` in there.
```suggestion
  return sortBy(entities, entity => entity.name.toLowerCase());
```

root/event/EventMerge.js
```diff
@@ -0,0 +1,59 @@
+/*
```
mwiencek: This file shouldn't technically be added in the areas commit (it's a copy of the areas one, but then later fixed in the events commit, which was a bit confusing).

root/components/list/AreaList.js
```diff
@@ -29,25 +31,30 @@ const AreaList = ({
...
     () => {
-      const checkboxColumn = $c.user_exists && checkboxes
-        ? defineCheckboxColumn(checkboxes)
+      const checkboxColumn = $c.user_exists && (checkboxes || mergeForm)
+        ? defineCheckboxColumn(checkboxes, mergeForm)
         : null;
       const nameColumn =
         defineNameColumn<AreaT>(l('Area'), order, sortable);
       const typeColumn = defineTypeColumn('area_type', order, sortable);
+      const removeFromMergeColumn = mergeForm
```
mwiencek: Would it make sense to check if `areas.length > 2` here, so an empty column isn't shown in that case? Or check that in `defineRemoveFromMergeColumn` instead so it applies to all?

reosarevok: I thought defineRemoveFromMergeColumn did that already, but maybe I messed up. I'll check!

root/work/WorkMerge.js
reosarevok: Oh, I didn't even notice... I actually think the wording is crap too, but I didn't have any good ideas. "You're about to merge all these works into a single one. Please select the work all others should be merged into:" is probably the least bad I can think of.

root/components/list/AreaList.js
reosarevok: Yes, defineRemoveFromMergeColumn already does this.

root/static/scripts/common/utility/sortByEntityName.js
reosarevok: For now I made the suggested change on the JS.

root/components/list/AreaList.js
mwiencek: Sorry, thought I checked but must've missed that somehow!

root/work/WorkMerge.js
mwiencek: Yeah, I think your suggestion sounds 100x better than what we have, though you don't have to change anything in this PR if you don't want to. :)

reosarevok: Undid the string change here, then added a commit standardising all of these to the same more normal-sounding string.

root/static/scripts/common/utility/sortByEntityName.js
```diff
@@ -0,0 +1,16 @@
...
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import sortBy from 'lodash/sortBy';
+
+export default function sortByEntityName(
+  entities: $ReadOnlyArray<CoreEntityT>,
+): $ReadOnlyArray<CoreEntityT> {
+  return sortBy(entities, [entity => entity.name.toLowerCase(), 'id']);
```
yvanzo, line 15: This only works with ASCII characters in `en` locale.
I would expect sort to follow user selected locale instead.

root/work/WorkMerge.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const WorkMerge = ({$c, form, toMerge}: Props) => (
+  <Layout fullWidth title={l('Merge works')}>
+    <div id="content">
+      <h1>{l('Merge works')}</h1>
+      <p>
+        {l(`You are about to merge the following works into a single
+            work. Please select the work that you would like
+            the other works merged into::`)}
```
yvanzo: Typo (not in the original TT) added in commit https://github.com/metabrainz/musicbrainz-server/pull/1431/commits/4627a92a506d8aae39c5f39b3b3ecaa17b0ce705, reverted after:

```suggestion
            the other works merged into:`)}
```

root/static/scripts/common/utility/sortByEntityName.js
reosarevok: Changed it to use i18n/compare, which does sort é with e as expected :)

root/static/scripts/common/utility/sortByEntityName.js
```diff
@@ -0,0 +1,26 @@
...
+import {compare} from '../i18n';
+
+function compareEntities(a: CoreEntityT, b: CoreEntityT): number {
+  const nameComparison = compare(a.name, b.name);
+  if (nameComparison === 0) {
+    return a.id < b.id
+      ? -1
+      : (a.id > b.id ? 1 : 0);
+  }
+  return nameComparison;
```
mwiencek, line 19: If `a.id` and `b.id` are always defined, the standard way to compare numbers is just to subtract them
```suggestion
  return compare(a.name, b.name) || (a.id - b.id);
```

root/static/scripts/common/utility/sortByEntityName.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export default function sortByEntityName(
+  entities: $ReadOnlyArray<CoreEntityT>,
+): $ReadOnlyArray<CoreEntityT> {
+  return entities.slice(0).sort((a, b) => compareEntities(a, b));
```
mwiencek, line 25:
```suggestion
  return entities.slice(0).sort(compareEntities);
```

root/static/scripts/common/utility/sortByEntityName.js
mwiencek: Unrelated, but we may also want to tweak this to use the sort name for artists. In my relationship editor rewrite branch I have a file at root/utility/getSortName.js which does this: https://gist.github.com/mwiencek/e7a2946edc11de6681760197dcd0eb39

So if you'd like you can copy that file to this branch and do `compare(getSortName(a), getSortName(b))` here instead. Or we can fix that later if you want to leave it for now

reosarevok: I did that, but changed it to actually check entities.json for sort_name: true, in case we would ever add sort names to something else than artist :)

root/utility/getSortName.js
```diff
@@ -0,0 +1,22 @@
...
+import ENTITIES from '../../entities';
+
+/*
+ * Returns the sort name for entities that have them, or falls back to
+ * the name otherwise.
+ */
+
+export default function getSortName(entity: CoreEntityT) {
+  const hasSortName = ENTITIES[entity.entityType].sort_name;
+  return hasSortName && entity.sort_name
```
reosarevok, line 19: The entity.sort_name check seems to be needed to calm Flow down.

mwiencek: This might be useful too: https://flow.org/en/docs/types/comments/#toc-comment-type-include
```suggestion
  return hasSortName /*:: && entity.sort_name */
```
Flow will treat the comment as if it were code, but it won't actually be executed. Sometimes useful for nudging the type checker like here while avoiding unnecessary checks at run time.

PR #1433, opened 2020-03-25
MBS-9728: Allow new emojis in titles

mwiencek:
https://tickets.metabrainz.org/browse/MBS-9728

The `:print:` char class doesn't include code points reserved for future use, since characters that don't exist aren't printable. But clearly, new emojis are added quite often and they'll always outpace the particular version of Unicode supported by our Perl interpreter.

Instead of stripping non-`:print:` characters, I've tried to construct the desired class of characters to remove explicitly. I didn't strip `\p{Cf}` (format characters) since it does appear we allow LRM/RLM under some circumstances in `remove_direction_marks`, though we may still want to remove zero-width joiners and language tag chars.

Obviously this may allow the entry of some unassigned code points if someone manages to enter them, but I'm not sure we need to worry about preventing that at all costs.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -317,9 +317,6 @@ sub collapse_whitespace {
     # Replace all spaces with U+0020
     =~ s/\s/ /gr
 
-    # Remove non-printable characters.
-    =~ s/[^[:print:]]//gr
-
     # Compress whitespace
     =~ s/\s{2,}/ /gr
 }
@@ -328,8 +325,12 @@ sub sanitize {
     my $t = shift;
 
     $t = NFC($t);
+    # Before removing invalid characters, convert space control characters
+    # into U+0020 (or else they'll be removed).
+    $t = collapse_whitespace($t);
     $t = remove_invalid_characters($t);
     $t = remove_direction_marks($t);
+    # Collapse spaces again, since characters may have been removed.
     $t = collapse_whitespace($t);
 
     return $t;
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -328,8 +325,12 @@ sub sanitize {
     my $t = shift;
 
     $t = NFC($t);
+    # Before removing invalid characters, convert space control characters
+    # into U+0020 (or else they'll be removed).
+    $t = collapse_whitespace($t);
```
reosarevok: Are we sure this doesn't remove anything we don't want to remove? (I guess we already remove full-width spaces anyway, although see MBS-5555

Also, we have a list of chars to avoid at https://musicbrainz.org/doc/Internationalization#Unicode_issues - not sure if we should block all of them while at it, but I guess dropping Ⅻ or whatnot would be weird. Maybe we should add a "split_roman_numbers" function here.

mwiencek:
> I'm not sure whether older PSQL won't also have some issues with the emojis, but in any case this seems needed for the newest ones.

I tested the patch against a 9.5 container (based on the same image we use in production) and had no issues there.

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: It shouldn't remove anything we didn't already remove, AFAICT. (If you can think of an example where it might, I can add a test.)

Re: MBS-5555, IDEOGRAPHIC SPACE is still prohibited by the doc you linked, so we should probably make a decision on whether we want to allow that first. :) The others listed in that doc are all already stripped (or converted to U+0020 where appropriate) except for fullwidth latin and halfwidth kana/hangul, and roman numerals, but those don't make sense to remove in `sanitize` since they're valid printable characters. We'd probably want to add FormHandler validation for those instead to inform the user that they shouldn't be entered.

Freso:
Would it be worth it/make sense to go through https://github.com/minimaxir/big-list-of-naughty-strings and see if there’s anything that needs testing for (in case something `:print:` was handling is not no longer being handled)? I don’t have a good overview of how comprehensive current handling/test suite for this is. :)

mwiencek:
> Would it be worth it/make sense to go through https://github.com/minimaxir/big-list-of-naughty-strings and see if there’s anything that needs testing for (in case something :print: was handling is not no longer being handled)? I don’t have a good overview of how comprehensive current handling/test suite for this is. :)

Thanks for the link! The one thing I found there we should probably handle is U+FEFF, which was also mentioned in MBS-4606, so I added a commit for that. There are also probably things we'd want to block in "general category Cf" listed there, as I mentioned in the PR description, but after checking we don't *currently* block everything from there either, so I don't believe this PR causes a regression in that area. So I'd rather do the bare minimum to fix MBS-9728 for now, since Cf has quite a few characters and it might take a while to research which ones to strip.

PR #1434, opened 2020-03-25
MBS-10718: Always use link orders for grouping

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10718

Previously, we only avoided grouping relationships with different link orders if we were viewing it from the "orderable" side. This modifies the code to use the link orders no matter which side we're viewing them from.

However, displaying the link orders as "1. Foo", "2. Bar", etc. doesn't make any sense from the non-orderable side, because the target entities there aren't what's being ordered (the source entity is). So I've changed it to display as "Foo (order 1)", "Bar (order 2)" from that side instead. (Let me know if you have a better idea here.)

A before screenshot can be found in the linked ticket. After:
Screen Shot 2020-03-25 at 3 39 44 AM

### DIFF
--- a/root/components/StaticRelationshipsDisplay.js
+++ b/root/components/StaticRelationshipsDisplay.js
@@ -50,11 +50,18 @@ const StaticRelationshipsDisplay = ({
 
         phraseRows.push(
           <React.Fragment key={targetGroup.key}>
-            {groupSize > 1 && targetGroup.linkOrder ? (
-              exp.l('{num}. {relationship}', {
-                num: targetGroup.linkOrder,
-                relationship: relationshipLink,
-              })
+            {targetGroup.linkOrder ? (
+              targetGroup.isOrderable ? (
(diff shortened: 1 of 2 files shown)


reosarevok:
Maybe order: 5 would be clearer? Or "in position 5"?

mwiencek:
Changed it to `order: {n}` but either of those seem fine actually.

root/components/StaticRelationshipsDisplay.js
```diff
@@ -50,11 +50,18 @@ const StaticRelationshipsDisplay = ({
 
         phraseRows.push(
           <React.Fragment key={targetGroup.key}>
-            {groupSize > 1 && targetGroup.linkOrder ? (
```
mwiencek: Also, I removed the `groupSize > 1` from here so that the link order is shown even if there's only one target entity. I see no reason to hide that data from the user in case they need to fix it.

PR #1435, opened 2020-03-25
MBS-10722: Entity list regression fixes

reosarevok:
# Fix MBS-10722

### DIFF
--- a/root/components/list/EventList.js
+++ b/root/components/list/EventList.js
@@ -72,23 +72,26 @@ const EventList = ({
           false, // to use EntityLink without dates (separate column for that)
         );
       const typeColumn = defineTypeColumn('event_type', order, sortable);
-      const artistsColumn = defineArtistRolesColumn(
+      const artistsColumn = defineArtistRolesColumn<EventT>(
         entity => entity.performers,
         'performers',
         l('Artists'),
       );
-      const timeColumn = defineTextColumn(
+      const timeColumn = defineTextColumn<EventT>(
         entity => entity.time,
         'time',
         l('Time'),
       );
       const rolesOnlyColumn = artist && artistRoles
-        ? defineTextColumn(
-          entity => entity.performers.map(performer => (
-            performer.entity.id === artist.id
-              ? commaOnlyListText(localizeArtistRoles(performer.roles))
-              : null
-          )),
+        ? defineTextColumn<EventT>(
+          entity => commaOnlyListText(
+            entity.performers.reduce((result, performer) => {
+              if (performer.entity.id === artist.id) {
+                result.push(...localizeArtistRoles(performer.roles));
+              }
+              return result;
+            }, []),
+          ),
           'performers',
           l('Role'),
         )
--- a/root/components/list/ReleaseGroupList.js
+++ b/root/components/list/ReleaseGroupList.js
@@ -89,8 +89,6 @@ export const ReleaseGroupListTable = withCatalystContext(({
         entity => entity.artistCredit,
         'artist',
         l('Artist'),
-        order,
-        sortable,
       );
       const typeColumn = defineTextColumn(
         entity => entity.l_type_name,
--- a/root/components/list/ReleaseGroupList.js
+++ b/root/components/list/ReleaseGroupList.js
@@ -55,6 +55,14 @@ export const ReleaseGroupListTable = withCatalystContext(({
   showType = true,
   sortable,
 }: ReleaseGroupListTableProps) => {
+  function getFirstReleaseYear(entity: ReleaseGroupT) {
+    if (!entity.firstReleaseDate) {
+      return '—';
+    }
+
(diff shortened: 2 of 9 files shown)


reosarevok:
The last commit will eventually be reverted if we merge https://github.com/metabrainz/musicbrainz-server/pull/1437 - if we merge that soon, I guess we could always just drop the commit here.

yvanzo:
> The last commit will eventually be reverted [...] I guess we could always just drop the commit here.

It seems cleaner to fix bugs with this PR, and to rebase the other PR that implements improvements.

root/components/list/ReleaseGroupList.js
```diff
@@ -64,8 +71,8 @@ export const ReleaseGroupListTable = withCatalystContext(({
         ? defineSeriesNumberColumn(seriesItemNumbers)
         : null;
       const yearColumn = defineTextColumn(
-        entity => parseDate(entity.firstReleaseDate).year?.toString() ?? '—',
```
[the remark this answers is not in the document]
reosarevok: Actually, I was being stupid. It's firstReleaseDate which was null, not the entity. So we don't need to change that.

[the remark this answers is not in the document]
mwiencek: Oh I see. `+firstReleaseDate: string | null,` is set in root/types.js already and `parseDate` also has valid Flow types, so I believe the actual issue was that Flow can't infer `D` from the `defineTextColumn` call. But we can specify the type parameter explicitly to help it:

```JS
const yearColumn = defineTextColumn<ReleaseGroupT>(
```

root/components/list/ReleaseGroupList.js
```diff
@@ -64,8 +71,8 @@ export const ReleaseGroupListTable = withCatalystContext(({
         ? defineSeriesNumberColumn(seriesItemNumbers)
         : null;
       const yearColumn = defineTextColumn(
```
mwiencek: Re: my last comment in the resolved thread

```suggestion
      const yearColumn = defineTextColumn<ReleaseGroupT>(
```

root/components/list/ReleaseGroupList.js
```diff
@@ -82,11 +89,11 @@ export const ReleaseGroupListTable = withCatalystContext(({
         entity => entity.artistCredit,
         'artist',
         l('Artist'),
-        order,
-        sortable,
       );
-      const typeColumn = defineTypeColumn(
-        'release_group_type',
+      const typeColumn = defineTextColumn(
```
mwiencek: We can add the type param here too.

reosarevok:
Decided to add the typing to all the relevant columns (not only RG ones). Caught a couple typing issues, and flow is complaining in a very confusing manner about one more:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ root/components/list/EventList.js:87:21

Cannot call defineTextColumn with function bound to getText because array type [1] is incompatible with string [2] in
the return value.

     root/components/list/EventList.js
      84│       );
      85│       const rolesOnlyColumn = artist && artistRoles
      86│         ? defineTextColumn<EventT>(
      87│           entity => entity.performers.map(performer => (
      88│             performer.entity.id === artist.id
      89│               ? commaOnlyListText(localizeArtistRoles(performer.roles))
      90│               : ''
      91│           )),
      92│           'performers',
      93│           l('Role'),
      94│         )

     /tmp/flow/flowlib_2a0deee7/core.js
 [1] 235│     map<U>(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => U, thisArg?: any): Array<U>;

     root/utility/tableColumns.js
 [2] 425│   getText: (D) => string,


```

root/components/list/EventList.js
```diff
@@ -72,23 +72,23 @@ const EventList = ({
...
       const rolesOnlyColumn = artist && artistRoles
-        ? defineTextColumn(
+        ? defineTextColumn<EventT>(
           entity => entity.performers.map(performer => (
             performer.entity.id === artist.id
               ? commaOnlyListText(localizeArtistRoles(performer.roles))
-              : null
-          )),
+              : ''
+          )).filter(Boolean).join(', '),
```
mwiencek: Instead of using `commaOnlyListText` on each set of roles and then further joining all of those with ", ", it might make more sense if we concatenate all the roles together and use `commaOnlyListText` on the final result. You can do that pretty easily with a `reduce` call, and it also removes the need for the `filter`.
```JS
          entity => commaOnlyListText(entity.performers.reduce((result, performer) => {
            if (performer.entity.id === artist.id) {
              result.push(...localizeArtistRoles(performer.roles));
            }
          }, [])),
```

PR #1436, opened 2020-03-25
MBS-10723: Separate className for cell and header

reosarevok:
# Fix MBS-10723

On top of https://github.com/metabrainz/musicbrainz-server/pull/1435

The lenghtClass for recording merges should only apply to cells, but currently className applies to both cells and headers. This uses the existing headerProps and creates an equivalent cellProps for cell attributes.

### DIFF
--- a/root/components/Table.js
+++ b/root/components/Table.js
@@ -14,9 +14,7 @@ import loopParity from '../utility/loopParity';
 
 const renderTableHeaderCell = (column) => (
   <th
-    {...column.getHeaderProps(
-      {...column.headerProps, className: column.className},
-    )}
+    {...column.getHeaderProps(column.headerProps)}
   >
     {column.render('Header')}
   </th>
@@ -29,7 +27,7 @@ const renderTableHeaderRow = (headerGroup) => (
 );
 
 const renderTableCell = (cell) => (
-  <td {...cell.getCellProps({className: cell.column.className})}>
+  <td {...cell.getCellProps(cell.column.cellProps)}>
     {cell.render('Cell')}
   </td>
 );
--- a/root/utility/tableColumns.js
+++ b/root/utility/tableColumns.js
@@ -66,7 +66,8 @@ export function defineActionsColumn(
     ),
     Header: l('Actions'),
     accessor: 'id',
-    className: 'actions',
+    cellProps: {className: 'actions'},
+    headerProps: {className: 'actions'},
     id: 'actions',
   };
 }
@@ -99,7 +100,7 @@ export function defineArtistCreditColumn<D>(
       )
       : title),
     accessor: row => getArtistCredit(row)?.names[0].name ?? '',
-    className: 'artist',
+    headerProps: {className: 'artist'},
     id: columnName,
   };
 }
@@ -162,7 +163,7 @@ export function defineCheckboxColumn(
         />
       ),
     Header: mergeForm ? null : <input type="checkbox" />,
-    className: 'checkbox-cell',
+    headerProps: {className: 'checkbox-cell'},
     id: 'checkbox',
   };
 }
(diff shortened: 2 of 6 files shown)


root/components/Table.js
```diff
@@ -14,9 +14,7 @@ import loopParity from '../utility/loopParity';
 
 const renderTableHeaderCell = (column) => (
   <th
-    {...column.getHeaderProps(
-      {...column.headerProps, className: column.className},
-    )}
```
yvanzo: Added `cellProps` looks good, but I don’t get why `className` has been removed. It seems that no change in parameters should be required but to `lengthColumn`.

mwiencek: The `className` appears to be passed through `headerProps` in all cases now, so that a special case isn't needed AFAICT.

root/utility/tableColumns.js
```diff
@@ -533,7 +539,8 @@ export const ratingsColumn:
     Cell: ({row: {original}}) => <RatingStars entity={original} />,
     Header: N_l('Rating'),
     accessor: 'rating',
-    className: 'rating c',
+    cellProps: {className: 'rating c'},
+    headerProps: {className: 'rating c'},
```
mwiencek: Didn't we previously only add most of these classes to the headers?

root/components/Table.js
reosarevok: It just seems much more consistent to me compared with passing className in two different ways and having to check className: column.headerProps.className || column.className.

root/utility/tableColumns.js
reosarevok: Hmm. For ratings specifically, we were super inconsistent. Many didn't have classes at all, but https://github.com/metabrainz/musicbrainz-server/commit/bae4c64c859f9ab752a5e6dca0cab4627689e2c5#diff-bf2f0326c5431ebe0a1f4fc70a8635f5 for example seems to have had 'rating c' on the header and just 'c' on the cell. We could now do that again, if desired.

mwiencek: Huh, okay. I'm also pretty sure we didn't have year, barcode-cell, checkbox-cell, actions, and the rest on all the cells either (at least I don't remember seeing these classes on every table cell in the web inspector). If we can avoid adding those unless they're already used or needed, that'd be good. If they're not used by any JS/CSS then they're cluttering the HTML (though existing ones might be used by userscripts).

reosarevok: Checked all the commits converting lists to React.
Some had `.pos`: it was used for some checkbox cells but we replaced, is equivalentish to `.checkbox-cell`, and you're right was used only on headers, so deleting `cellProps` for that one.
`.artist` was used on RG lists (only, for some reason?) and also only on headers. Same change.
`.rating` is also on (RG and recording) headers only, but both cells and headers have `.c` on RG (not on recording, because we love consistency). Same for `.count` and `.year` on RG. Changed it back for cells to have just `.c`.
Works used to have `.iswc` on ISWC cells (but not headers). Readding that.
`.barcode-cell` was on cells, not headers. Deleted `headerProps` there.

root/components/Table.js
yvanzo: From the below discussion, it seems that classes should be set only to headers in most cases, thus there won’t be that much duplicates between `headerProps` and `cellProps` anyway.

root/utility/tableColumns.js
yvanzo: Good catch, changes seem reasonable but I did not review thoroughly. To ease review (and future search through code history), it would have been helpful to split this patch (which is supposed to fix heading ticket mainly) and to add a reference for each of these buggy commits you checked and fixed.

root/components/Table.js
```diff
@@ -29,7 +27,9 @@ const renderTableHeaderRow = (headerGroup) => (
 );
 
 const renderTableCell = (cell) => (
-  <td {...cell.getCellProps({className: cell.column.className})}>
+  <td
+    {...cell.getCellProps(cell.column.cellProps)}
+  >
```
mwiencek: Some line breaks were added inside the tag here, but it actually makes the original line shorter, so probably not needed

PR #1437, opened 2020-03-25
MBS-10724: Make list sorting options consistent

reosarevok:
We're super inconsistent on what columns we can sort by or not. This makes it so that if an entity can be sorted by a column, others can too.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -237,8 +237,14 @@ sub _order_by {
         "begin_date" => sub {
             return "begin_date_year, begin_date_month, begin_date_day, musicbrainz_collate(name)"
         },
+        "begin_area" => sub {
+            return "begin_area, musicbrainz_collate(name)"
+        },
         "end_date" => sub {
             return "end_date_year, end_date_month, end_date_day, musicbrainz_collate(name)"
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -228,9 +228,24 @@ sub _order_by {
         "name" => sub {
             return "musicbrainz_collate(name)"
         },
+        "area" => sub {
+            return "area, musicbrainz_collate(name)"
```
mwiencek: Shouldn't we join the area table and sort by its name instead of its ID? 'Cause I imagine the ID won't have any logical order other than putting identical areas next to each other.

reosarevok: Maybe? This is what we have done all the time, but it doesn't seem too smart. Of course the ideal sort would use containment but name might be better than id anyway

mwiencek: I guess just labels, but I see we do similar bad stuff with gender, type, etc. below. Surprised no one's complained about that. Anyway, if you're okay with keeping it how it is for now we can merge this and I'll enter a ticket for having the sort make sense.

reosarevok:
Let's do that maybe :)

PR #1438, opened 2020-03-25
MBS-10688: Show a human readable error when attaching a dupe CDTOC

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10688

If the selected medium already has the CDTOC being attached, go back to the filter/listing page and show a readable error message informing the user rather than a cryptic "Bad Request" page.
Screen Shot 2020-03-25 at 4 46 14 PM

### DIFF
--- a/root/cdtoc/attach_list.tt
+++ b/root/cdtoc/attach_list.tt
@@ -18,18 +18,25 @@
   </tr>
   [% FOR medium=release.mediums %]
     [% NEXT UNLESS
-        medium.cdtoc_track_count == cdtoc.track_count %]
+        medium.cdtoc_track_count == cdtoc.track_count;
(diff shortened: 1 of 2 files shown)


root/cdtoc/attach_list.tt
```diff
@@ -18,18 +18,25 @@
   </tr>
   [% FOR medium=release.mediums %]
     [% NEXT UNLESS
-        medium.cdtoc_track_count == cdtoc.track_count %]
+        medium.cdtoc_track_count == cdtoc.track_count;
+       this_medium_has_cdtoc = medium_has_cdtoc.defined AND (medium.id == medium_has_cdtoc) %]
```
reosarevok: Indent seems off?

mwiencek: The line I added looks okay since it's lined up with the `NEXT`. The one above should probably be indented +1 space, but not sure I care enough to fix old TT. :P

PR #1439, opened 2020-03-25
MBS-10360: Whitelist User-Agent header in CORS preflight requests

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10360


reosarevok:
Apparently you also need to change https://github.com/metabrainz/musicbrainz-server/blob/393fac6c747526bb226eb03808ff77d3b71dab6a/t/lib/t/MusicBrainz/Server/Controller/WS/2/SubmitRecording.pm#L51 but otherwise seems good :)

PR #1441, opened 2020-03-26
MBS-9169: Fix corrupted alias locales

mwiencek:
When we upgraded DateTime::Locale to version 1.00+ ages ago, one of the breaking changes was that "[t]he canonical form of the locale codes now uses dashes (-) instead of underscores (\_)." Unfortunately, this introduced an inconsistency in the way locales are stored in the database, with old ones using '\_' and new ones using '-', causing several serious issues described by Ulrich Klauer in MBS-9169.

This commit standardizes all locale codes to used underscores as before, adding a script to convert any existing ones now considered corrupt.


lib/MusicBrainz/Server/Constants.pm
```diff
@@ -955,6 +957,30 @@ Readonly our $PASSPHRASE_BCRYPT_COST => 12;
...
 Readonly our $OAUTH_WEB_APP_REDIRECT_URI_RE => qr/^https?:\/\//;
 
+=item %ALIAS_LOCALES
+
+Historically, alias locales have been stored in the database and
+returned in the web service using underscores instead of dashes, e.g.
+zh_Hant_HK instead of zh-Hant-HK. Presumably this was simply because
+that was how DateTime::Locale returned them prior to version 1.00.
+The primary reason to continue doing this is to maintain compatibility
+for data users.
```
reosarevok: Given we have had both kinds forever now, wouldn't it make sense to consider compatibility broken already and just go with the canonical form?

mwiencek: No because even though we've had the hyphened ones for a while now, the underscored ones are still vastly more prevalent (163709 with underscores vs. only 4135 with hyphens). Any downstream data users who've written code to use alias locales (whether from the DB or web service) would very likely see their code break if we changed that. Here's just one example from Picard: https://github.com/metabrainz/picard/blob/8245afbd5ba1f345e7ab61db83dc7af67e6f482c/picard/mbjson.py#L205

reosarevok: Ouch. Ok then.

lib/MusicBrainz/Server/Constants.pm
```diff
@@ -6,6 +6,7 @@ use warnings;
 use base 'Exporter';
 
 use DateTime::Duration;
+use DateTime::Locale '1.00';
```
yvanzo, line 9: Just to make sure: this means _at least version 1.00_, right?

mwiencek: That is correct, see https://perldoc.perl.org/functions/use.html

mwiencek:
> LGTM, but we should maybe test running it on a non-replicated DB first if you haven't yet.

I tested the script on my local development DB of course, though that only had the sample data dumps imported. I think the next step should be running it on the test.mb database.

mwiencek:
I ran this on floyd but replacing `COMMIT` with `ROLLBACK`, and also outputting counts of affected rows from each table before/after, plus the data from 3 sample rows per table to manually check the changes. (Complete script used is here: https://gist.github.com/mwiencek/542ff8a3ecf43916bbaf321be458c5cc.)

That produced this output: https://gist.github.com/mwiencek/0f04894949a83eccae50c8dff0c22e35 which doesn't show any issues as far as I can tell.

mwiencek:
I've rebased this onto production, since we should hotfix the code here to production & beta before running the SQL. Otherwise someone could add another hyphenated locale after the script is run. (Alternatively we could run the script twice, but it seems ideal to have everything working on all servers at once.)

PR #1442, opened 2020-03-26
MBS-7781: Add script to merge duplicate artist credits

mwiencek:
https://tickets.metabrainz.org/browse/MBS-7781

### DIFF
--- a/admin/cleanup/MergeDuplicateArtistCredits
+++ b/admin/cleanup/MergeDuplicateArtistCredits
@@ -1,0 +1,216 @@
+#!/usr/bin/env perl
+
+=head1 NAME
+
+MergeDuplicateArtistCredits - Finds and merges duplicate artist credits
+
+=head1 SYNOPSIS
+
+MergeDuplicateArtistCredits [options]
+
+Options:
+
+    -l, --limit NUM         limit the number of entities updated
+    -d, --database NAME     name of database in DBDefs.pm to connect to
+    -n, --dry-run           show any changes but don't apply them
+    -h, --help              show this help
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2020 MetaBrainz Foundation
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
+
+use strict;
+use warnings;
+
+use FindBin;
+use lib "$FindBin::Bin/../../lib";
+
+use Getopt::Long qw( GetOptions );
+use List::MoreUtils qw( natatime );
+use Pod::Usage qw( pod2usage );
+use MusicBrainz::Server::Constants qw(
+    %ENTITIES
+    entities_with
+);
+use MusicBrainz::Server::Context;
+use MusicBrainz::Server::Log qw( log_info );
+
+my $limit = 100;
+my $database = 'MAINTENANCE';
+my $dry_run;
+my $help_flag;
+
+GetOptions(
+    'limit|l=i' => \$limit,
+    'database|d=s' => \$database,
+    'dry-run|n' => \$dry_run,
+    'help|h' => \$help_flag,
+);
+
+pod2usage() if $help_flag;
+pod2usage(
+    -exitval => 64, # EX_USAGE
+    -message => "$0: unrecognized arguments",
+) if @ARGV;
+
+log_info {
+    'Performing a dry run: changes reported below will not be committed ' .
+    'to the database or cache.'
+} if $dry_run;
+
+my $c = MusicBrainz::Server::Context->create_script_context(
+    database => $database,
+);
+
+$c->sql->begin;
+
+my $rows = $c->sql->select_single_column_array(q{
+    WITH ac_names AS (
+        SELECT ac.id,
+               array_agg(
+                row(acn.position,
+                    acn.artist,
+                    acn.name,
+                    acn.join_phrase)
+                ORDER BY acn.position ASC
+               ) AS names
+          FROM artist_credit ac
+          JOIN artist_credit_name acn
+            ON acn.artist_credit = ac.id
+         WHERE ac.edits_pending = 0
+         GROUP BY ac.id
+    )
+    SELECT array_agg(id ORDER BY id ASC) AS duplicates
+      FROM ac_names
+     GROUP BY names
+    HAVING count(id) > 1
+     LIMIT ?
+}, $limit);
+
+unless (@$rows) {
+    log_info { 'No mergeable artist credits found.' };
+    exit(0);
+}
+
+my @ac_entity_models = map { $c->model($_) }
+    sort { $a cmp $b }
+    entities_with('artist_credits', take => 'model');
+
+my $delete_unused_ac_name_query =
+    'DELETE FROM artist_credit_name ' .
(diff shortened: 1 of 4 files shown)


admin/sql/updates/20200326-mbs-7781.sql
```diff
@@ -0,0 +1,84 @@
...
+               AND ac1.names = ac2.names
+        )
+        -- Remove dupes of other dupes.
+        SELECT *
+          FROM dupe_acs dac1
+         WHERE NOT EXISTS (
+           SELECT 1 FROM dupe_acs dac2
+            WHERE dac2.dupe = dac1.orig
+         )
+         ORDER BY dac1.orig ASC
```
reosarevok: So the idea here is that all dupes will be paired with every single dupe with a smaller ID, and this removes those to leave only the pairs between all dupes and the smallest ID of them all?

mwiencek: Sounds about right. But there might be a more straightforward way to do that, it was pretty late when I wrote this. :)

mwiencek: Ok yeah, a much simpler way is just to group `ac_names` by `names` and stuff all the dupe IDs into one array (with the smallest ID being the first in the array). Did that and also separated out the main query into a `duplicate_artist_credits` view, in case we want to see/inspect if duplicate artist credits exist without merging them.

mwiencek:
I did test it in my local DB, but I'll try to get a pgtap test for that added too. :)

And thinking about how we should go about running this: I should probably add a parameter to limit the number of ACs it changes in one run (to avoid stuffing all the changes into a single replication packet, which I imagine would end up being pretty large), and then add a small script to call it in hourly.sh.

mwiencek:
Added admin/cleanup/MergeDuplicateArtistCredits, which calls the SQL function and clears relevant data from the cache. Hooked that into hourly.sh with a limit of 10 ACs.

admin/sql/updates/20200326-mbs-7781.sql
```diff
@@ -0,0 +1,79 @@
...
+
+-- N.B. Affected entities and artist credits must be cleared from the cache
+-- after running this.
+
+CREATE OR REPLACE FUNCTION merge_duplicate_artist_credits(orig_ac INT)
+RETURNS void AS $$
+DECLARE
+    dupe_acs INT[];
+BEGIN
+    SELECT duplicates[2:]
```
yvanzo, line 35: Test passed on CircleCI (job [1656](https://app.circleci.com/pipelines/github/mwiencek/musicbrainz-server/465/workflows/a46b2d49-1847-41ef-90b8-10432c777c82/jobs/1656)), not sure what is causing this failure on my host:
```
psql:t/pgtap/../../admin/sql/updates/20200326-mbs-7781.sql:77: ERROR:  syntax error at or near "]"
LINE 6:     SELECT duplicates[2:]
                                ^
t/pgtap/merge_duplicate_artist_credits.sql ............ Dubious, test returned 3 (wstat 768, 0x300)
No subtests run 
```

mwiencek: Oh. I guess that's a feature in PG 9.6 (of course...). I don't see it mentioned in the 9.5 docs.

Added `array_length(duplicates, 1)` for the missing upper bound.

yvanzo:
It limits to 10 target ACs though. The very first AC is “Various Artists” and it has 10 duplicates already. The list of release-groups and the list of releases to be cleared are so large that output is truncated. Maybe there should be a way to further limit the total number of ACs to be merged?

mwiencek:
@yvanzo I've updated it so that `--limit` indicates the max number of *entities* to update. So invoking the script may not actually merge or remove any artist credits, at least not on the first invocation, unless a high limit is set; it only proceeds to do that once enough entities are updated to make an artist credit become unused.

admin/cleanup/MergeDuplicateArtistCredits
```diff
@@ -0,0 +1,210 @@
...
+        }, $orig_ac, \@dupe_acs, ($limit - $updated_entity_count));
+
+        my $entity_id_count = scalar(@$entity_ids);
+        next unless $entity_id_count > 0;
+
+        log_info {
+            "Updated $entity_id_count $entity_table " .
+            ($entity_id_count == 1 ? 'row' : 'rows') .
+            " to use artist_credit = $orig_ac."
+        };
```
yvanzo, line 144: To still append the full list of entity rows being updated, it can be logged 100 ids at a time only to keep each message below max length before message truncation:

```suggestion
        log_info {
            "Updated $entity_id_count $entity_table " .
            ($entity_id_count == 1 ? 'row' : 'rows') .
            " to use artist_credit = $orig_ac:"
        };
        my $it = natatime 100, @$entity_ids;
        while (my @next_ids = $it->()) {
            log_info {
                join(', ', @next_ids)
            };
        }
```

admin/cleanup/MergeDuplicateArtistCredits
```diff
@@ -0,0 +1,210 @@
...
+    unless ($dry_run) {
+        my $it = natatime 100, @$ids;
+        while (my @next_ids = $it->()) {
+            $model->_delete_from_cache(@next_ids);
+        }
+    }
+
+    my $id_count = scalar @$ids;
+    log_info {
+        "Cleared $id_count $table " .
```
yvanzo, line 197: Following suggestion above:
```suggestion
        "Cleared the same $id_count $table " .
```

mwiencek: Merged both suggestions, thanks!

admin/cleanup/MergeDuplicateArtistCredits
```diff
@@ -0,0 +1,216 @@
...
+                row(acn.position,
+                    acn.artist,
+                    acn.name,
+                    acn.join_phrase)
+                ORDER BY acn.position ASC
+               ) AS names
+          FROM artist_credit ac
+          JOIN artist_credit_name acn
+            ON acn.artist_credit = ac.id
+         WHERE ac.edits_pending = 0
```
mwiencek, line 86: Also, forgot to mention, but when I moved the query here I tweaked it slightly to avoid selecting artist credits with `edits_pending != 0`. I'd rather not have the script have to deal with merging the `edits_pending` columns, or breaking any open edits for that matter, and the script should pick them up again once the edits are applied.

admin/cleanup/MergeDuplicateArtistCredits
```diff
@@ -0,0 +1,216 @@
...
+
+    for my $model (@ac_entity_models) {
+        my $entity_table = $model->_main_table;
+
+        my $entity_ids = $c->sql->select_single_column_array(qq{
+            UPDATE $entity_table SET artist_credit = ?
+             WHERE id IN (SELECT id FROM $entity_table
+                           WHERE artist_credit = any(?::INT[])
+                           ORDER BY id ASC
+                           LIMIT ? FOR UPDATE SKIP LOCKED)
```
mwiencek, line 133: `FOR UPDATE` will lock the selected entity rows against concurrent updates, and `SKIP LOCKED` will skip ones we can't immediately obtain a lock for (to avoid lock contention). The `SELECT` statements in `$delete_unused_ac_name_query` will still see locked rows, so it won't try to remove the ACs if there are still locked rows left - they'll just have to be processed again in the next hour.

PR #1443, opened 2020-03-29
MBS-10730: Select DISTINCT video/standalone recordings

Kebabpizza:
Standalone and video recordings with multiple artist credits to the same artist were shown
multiple times on the artist's page

### DIFF
--- a/lib/MusicBrainz/Server/Data/Recording.pm
+++ b/lib/MusicBrainz/Server/Data/Recording.pm
@@ -301,7 +301,8 @@ sub find_standalone
 {
     my ($self, $artist_id, $limit, $offset) = @_;
     my $query ='
-        SELECT ' . $self->_columns . '
+        SELECT DISTINCT ' . $self->_columns . ',
+            musicbrainz_collate(recording.name)
           FROM ' . $self->_table . '
      LEFT JOIN track t ON t.recording = recording.id
           JOIN artist_credit_name acn
@@ -316,7 +317,8 @@ sub find_video
 {
     my ($self, $artist_id, $limit, $offset) = @_;
     my $query ='
-        SELECT ' . $self->_columns . '
+        SELECT DISTINCT ' . $self->_columns . ',
+            musicbrainz_collate(recording.name)
           FROM ' . $self->_table . '
           JOIN artist_credit_name acn
             ON acn.artist_credit = recording.artist_credit


lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -301,7 +301,8 @@ sub find_standalone
 {
     my ($self, $artist_id, $limit, $offset) = @_;
     my $query ='
-        SELECT ' . $self->_columns . '
+        SELECT DISTINCT ' . $self->_columns . ',
+            musicbrainz_collate(recording.name)
```
reosarevok: Distinct makes sense, but why add this?

yvanzo: Selecting the sort column `musicbrainz_collate(recording.name)` is not needed with PostgreSQL.

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -316,7 +317,8 @@ sub find_video
...
```
yvanzo: Same as above.

lib/MusicBrainz/Server/Data/Recording.pm
Kebabpizza: Doesn't work for me. Appears to be needed when using distinct
```
42P10 DBD::Pg::st execute failed: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 14:       ORDER BY musicbrainz_collate(recording.name) OFFSET '0...
                        ^ [for Statement "
        WITH x AS (
        SELECT DISTINCT recording.id, recording.gid, recording.name,
            recording.artist_credit AS artist_credit_id,
            recording.length, recording.comment, recording.video,
            recording.edits_pending, recording.last_updated
          FROM recording
     LEFT JOIN track t ON t.recording = recording.id
          JOIN artist_credit_name acn
            ON acn.artist_credit = recording.artist_credit
         WHERE t.id IS NULL
           AND acn.artist = ?
      ORDER BY musicbrainz_collate(recording.name) OFFSET ?)
        SELECT x.*, c.count AS total_row_count
        FROM x, (SELECT count(*) from x) c
     LIMIT ?" with ParamValues: 1='1911941', 2='0', 3='100']
```

yvanzo: Oops, it is not needed for `SELECT` but it is for `SELECT DISTINCT indeed, my bad!

yvanzo: Below comment https://github.com/metabrainz/musicbrainz-server/pull/1443#discussion_r399800203 seems to answer that too.

PR #1445, opened 2020-04-01
MBS-5380 (1st step): Contextualize “Edit”

mihkel-t:
My first steps to work on MBS-5380: Contextualize further strings

As I'm not really sure whether I did things right or not, this PR only touches the occurrences of "Edit" (which is the single worst offender currently, l10n-wise, at least for Estonian), and changes a compounded string into a regular one (which is a better practice i18n-wise).

For "Edit", I used two contexts:
1) "button/link", as these two UI elements behave similarly and should therefore also be translated similarly
2) "tab/column/page title" - or should it be just "title", perhaps?

### DIFF
--- a/root/cdstub/header.tt
+++ b/root/cdstub/header.tt
@@ -8,7 +8,7 @@
 
 [%- info_links = [
     ['index', link_cdstub(cdstub, 'show', l('Overview')) ]
-    ['edit', link_cdstub(cdstub, 'edit', lp('Edit', 'tab/column title')) ]
+    ['edit', link_cdstub(cdstub, 'edit', lp('Edit', 'tab/column/page title')) ]
 ] -%]
 
 [% INCLUDE 'components/tabs.tt' list=info_links %]
--- a/root/cdstub/header.tt
+++ b/root/cdstub/header.tt
@@ -8,7 +8,7 @@
 
 [%- info_links = [
     ['index', link_cdstub(cdstub, 'show', l('Overview')) ]
-    ['edit', link_cdstub(cdstub, 'edit', lp('Edit', 'tab title')) ]
+    ['edit', link_cdstub(cdstub, 'edit', lp('Edit', 'tab/column title')) ]
 ] -%]
 
 [% INCLUDE 'components/tabs.tt' list=info_links %]
--- a/root/cdstub/header.tt
+++ b/root/cdstub/header.tt
@@ -8,7 +8,7 @@
 
 [%- info_links = [
     ['index', link_cdstub(cdstub, 'show', l('Overview')) ]
-    ['edit', link_cdstub(cdstub, 'edit', l('Edit')) ]
+    ['edit', link_cdstub(cdstub, 'edit', lp('Edit', 'tab title')) ]
 ] -%]
 
 [% INCLUDE 'components/tabs.tt' list=info_links %]
--- a/root/release/edit/information.tt
+++ b/root/release/edit/information.tt
@@ -249,7 +249,7 @@
       </tr>
 
       <tr>
-        <td><label for="comment">[% add_colon(l('Disambiguation')) %]</label></td>
+        <td><label for="comment">[% l('Disambiguation:') %]</label></td>
         <td>
           <input id="comment" type="text" maxlength="255" data-bind="value: comment, controlsBubble: $root.commentBubble" />
         </td>
--- a/root/release/edit/recordings.tt
+++ b/root/release/edit/recordings.tt
@@ -55,7 +55,7 @@
             <!-- /ko -->
 
             <td rowspan="2" class="buttons">
-              <button class="edit-track-recording" type="button" data-bind="controlsBubble: $root.recordingBubble, css: { negative: needsRecording() }">[% lp('Edit', 'button') | html_entity %]</button>
+              <button class="edit-track-recording" type="button" data-bind="controlsBubble: $root.recordingBubble, css: { negative: needsRecording() }">[% lp('Edit', 'button/link') | html_entity %]</button>
             </td>
           </tr>
 
@@ -91,7 +91,7 @@
       <div class="edit-recording">
         <div class="buttons">
           <button type="button" data-bind="disable: loading" data-click="loadTracks">
-            [% lp('Edit', 'button') | html_entity %]
+            [% lp('Edit', 'button/link') | html_entity %]
           </button>
         </div>
 
--- a/root/release/edit/recordings.tt
+++ b/root/release/edit/recordings.tt
@@ -55,7 +55,7 @@
             <!-- /ko -->
 
             <td rowspan="2" class="buttons">
-              <button class="edit-track-recording" type="button" data-bind="controlsBubble: $root.recordingBubble, css: { negative: needsRecording() }">[% l('Edit') | html_entity %]</button>
+              <button class="edit-track-recording" type="button" data-bind="controlsBubble: $root.recordingBubble, css: { negative: needsRecording() }">[% lp('Edit', 'button') | html_entity %]</button>
             </td>
           </tr>
 
@@ -91,7 +91,7 @@
       <div class="edit-recording">
         <div class="buttons">
           <button type="button" data-bind="disable: loading" data-click="loadTracks">
-            [% l('Edit') | html_entity %]
+            [% lp('Edit', 'button') | html_entity %]
           </button>
         </div>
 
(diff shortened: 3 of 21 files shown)


mwiencek:
@brainzbot add to whitelist

root/release/edit/information.tt
```diff
@@ -249,7 +249,7 @@
       </tr>
 
       <tr>
-        <td><label for="comment">[% add_colon(l('Disambiguation')) %]</label></td>
+        <td><label for="comment">[% l('Disambiguation:') %]</label></td>
```
yvanzo, line 252: Oops, why do you think it is pointless?
The purpose of `add_colon` (name can be changed if needed) is twofold:
* to prevent duplicates between labels depending on presentation (e.g. `Name` and `Name:`)
* to uniformize localization of labels ending with a colon `Foo:`

root/release/edit/recordings.tt
```diff
@@ -91,7 +91,7 @@
       <div class="edit-recording">
         <div class="buttons">
           <button type="button" data-bind="disable: loading" data-click="loadTracks">
-            [% l('Edit') | html_entity %]
+            [% lp('Edit', 'button') | html_entity %]
```
yvanzo: Button is just the presentation, sure `button` context helps with context, but it does not help with the meaning of Edit (could still be either _an_ edit or _to_ edit). Would `action` be a better fit for example?

root/cdstub/header.tt
```diff
@@ -8,7 +8,7 @@
 
 [%- info_links = [
     ['index', link_cdstub(cdstub, 'show', l('Overview')) ]
-    ['edit', link_cdstub(cdstub, 'edit', l('Edit')) ]
+    ['edit', link_cdstub(cdstub, 'edit', lp('Edit', 'tab title')) ]
```
yvanzo: Same as above, same suggestion too.

root/release/edit/recordings.tt
mihkel-t: The question is not whether it's a verb or a noun. The point is that some languages (such as mine) would need the verb to appear differently depending on whether it's used as in a title-like position (e.g. as a tab in the UI, as a column header, as part of page title, or as menu header), or as a direct command (whether button, link or a menu command).

FWIW, I changed the context to "button/link" later on.

I feel that "action" can be interpreted a bit too ambiguously... Would "command" instead of "button/link" make more sense?

root/release/edit/information.tt
mihkel-t: I have two reasons for this, as well :)

1) Colon is not a universal symbol in this function (nor formatted universally alike), and it would be better if translators could replace it, if needed.
See these two links for examples:
https://en.wikipedia.org/wiki/Punctuation#Other_languages
https://en.wikipedia.org/wiki/Colon_(punctuation)#End_of_sentence_or_verse

2) In my language, Estonian (and surely some other languages, too), it is sometimes needed to translate a word slightly differently (e.g. using a different case ending) depending on whether it's used as a label for a text field or as title of something.

mihkel-t:
> It seems to me that both `'Edit', 'button'` and `'Edit', 'tab title'` have the same meaning and should be translated in the same way. So, I would rather use the same context string here: `action` or any better descriptive string.

Naturally they need to be translated using the same term - but in some languages (such as my native Estonian) not using the same case ending of that term.

root/release/edit/information.tt
yvanzo: >     1. Colon is not a universal symbol in this function (nor formatted universally alike), and it would be better if translators could replace it

Totally agree, and translators can fortunately replace it, see https://github.com/metabrainz/musicbrainz-server/blob/v-2020-03-31/po/mb_server.pot#L6994-L6998

>     2. In my language, Estonian (and surely some other languages, too), it is sometimes needed to translate a word slightly differently (e.g. using a different case ending) depending on whether it's used as a label for a text field or as title of something.

Can you please provide examples? I’m inclined to believe you, that is not the issue. Because we intentionally replaced colon strings with `addColon` calls until now, we need to convincingly document the reasoning and create guidelines for future coders to add properly localizable strings rather than to do the same mistake again.

reosarevok: We should probably still replace them, in any case - it's just we might need to add context in some cases.

root/release/edit/recordings.tt
yvanzo: > The question is not whether it's a verb or a noun. [...] the verb [...]

I would not assume it is always the verb as in _Edit_ tab ([example](https://musicbrainz.org/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/edit)), it could be the noun as in _Edit_ descriptive page ([example](https://musicbrainz.org/edit/65697443)).

> to appear differently depending on whether it's used as in a title-like position (e.g. as a tab in the UI, as a column header, as part of page title, or as menu header), or as a direct command (whether button, link or a menu command). [...] I feel that "action" can be interpreted a bit too ambiguously... Would "command" instead of "button/link" make more sense?

I think I got it. We just had different expectations of the context string: meaning/disambiguation vs. role/layout. It seems that we should probably have both. That would require to agree on how to formulate context strings. Since MBS-10132 was not specific enough about that, let's discuss guidelines about writing localizable strings on that ticket?

mihkel-t: > I would not assume it is always the verb as in _Edit_ tab ([example](https://musicbrainz.org/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/edit)), it could be the noun as in _Edit_ descriptive page ([example](https://musicbrainz.org/edit/65697443)).

Well, you are right of course, it's just that _that_ is a third context in addition to my proposed two :) Luckily, currently none of the nouns "edit" are confused with the verbs "edit" (or at least I haven't noticed it), and it's only the verbs that I added context to.

> I think I got it. We just had different expectations of the context string: meaning/disambiguation vs. role/layout. It seems that we should probably have both.

That would probably be a better option indeed.

> That would require to agree on how to formulate context strings. ~Since [MBS-10132](https://tickets.metabrainz.org/browse/MBS-10132) was not specific enough about that~, oops [MBS-5380](https://tickets.metabrainz.org/browse/MBS-5380), let's discuss guidelines about writing localizable strings on that ticket? Or create a new ticket to improve https://musicbrainz.org/doc/Server_Internationalisation?

Sure. Maybe it would make sense to create a new ticket, so that MBS-5380 would only have the actual requests to split certain strings?

yvanzo: Agreed, created MBS-10739.

root/release/edit/information.tt
mihkel-t: > Totally agree, and translators can fortunately replace it, see https://github.com/metabrainz/musicbrainz-server/blob/v-2020-03-31/po/mb_server.pot#L6994-L6998

Okay, that makes it somewhat better at least.

> > ```
> > 2. In my language, Estonian (and surely some other languages, too), it is sometimes needed to translate a word slightly differently (e.g. using a different case ending) depending on whether it's used as a label for a text field or as title of something.
> > ```
> 
> Can you please provide examples?

Ended: → Lõpp: ("the end" - not ideal, but had to find something that would work for both artist, label and area entities)
Ended → Pole enam aktiivselt kasutusel ("not in active use anymore", admittedly a rather loose translation, but again, that's what worked for all the various contexts of this string)

Note: → Pane tähele: (verb)
Note → Märkus (noun)

Votes for: → Poolthääli: ("[number of] favourable votes")
Votes for → Poolt ("in favour")

Votes against: → Vastuhääli:
Votes against → Vastu

These are just from a quick look, and only from server.po. Sure, there were also many otherwise identical strings where the colon really was the only difference, but I still stand by the idea that it's easier for translators if we can quickly see the _complete_ string, including any trailing colons, in order to properly localize it. Especially with short phrases (1-3 words), my experience is that it's easier to translate 3 separate versions of a string (presuming one can understand their context) than to try and find a translation that _kind of works_ for all the various contexts lumped into 1 source string :)

mihkel-t: > We should probably still replace them, in any case - it's just we might need to add context in some cases.

It's hard to know beforehand what needs context - and even what kind of context makes sense to different persons, as is apparent in the comments here.

mihkel-t:
Cancelling this PR until at least a partial agreement is reached over at MBS-10739 in Jira. (Will most likely need to change some of the added context phrases anyway.)

PR #1449, opened 2020-04-03
MBS-10742: Show right number of hidden events

reosarevok:
### Fix MBS-10742

The number being printed in "show {n} more" was events.length - COLLAPSE_THRESHOLD, but in fact the code prints two events before the collapse and one after - 3, not the 4 for COLLAPSE_THRESHOLD.

This moves the numbers to (easier to understand) constants, and then ensures we show the right number of events.

### DIFF
--- a/root/static/scripts/common/components/ReleaseEvents.js
+++ b/root/static/scripts/common/components/ReleaseEvents.js
@@ -17,7 +17,9 @@ import isDateEmpty from '../utility/isDateEmpty';
 
 import EntityLink from './EntityLink';
 
-const COLLAPSE_THRESHOLD = 4;
+const TO_SHOW_BEFORE = 2;
+const TO_SHOW_AFTER = 1;
+const TO_TRIGGER_COLLAPSE = TO_SHOW_BEFORE + TO_SHOW_AFTER + 2;
 
 const releaseEventKey = event => (
   String(event.country ? event.country.id : '') + '\0' +
@@ -103,7 +105,7 @@ const ReleaseEvents = ({
   };
 
   const tooManyEvents = events
-    ? events.length > COLLAPSE_THRESHOLD
+    ? events.length >= TO_TRIGGER_COLLAPSE
     : false;
 
   return (
@@ -112,7 +114,7 @@ const ReleaseEvents = ({
         {(tooManyEvents && !expanded) ? (
           <>
             <ul {...containerProps}>
-              {events.slice(0, COLLAPSE_THRESHOLD - 2).map(
+              {events.slice(0, TO_SHOW_BEFORE).map(
                 event => buildReleaseEventRow(event, abbreviated),
               )}
               <li className="show-all" key="show-all">
@@ -123,11 +125,13 @@ const ReleaseEvents = ({
                   title={l('Show all release events')}
                 >
                   {bracketedText(texp.l('show {n} more', {
-                    n: events.length - COLLAPSE_THRESHOLD,
+                    n: events.length - (TO_SHOW_BEFORE + TO_SHOW_AFTER),
                   }))}
                 </a>
               </li>
-              {buildReleaseEventRow(events[events.length - 1], abbreviated)}
+              {events.slice(-TO_SHOW_AFTER).map(
+                event => buildReleaseEventRow(event, abbreviated),
+              )}
             </ul>
           </>
         ) : (
--- a/root/static/scripts/common/components/WorkArtists.js
+++ b/root/static/scripts/common/components/WorkArtists.js
@@ -14,7 +14,8 @@ import {bracketedText} from '../utility/bracketed';
 
 import ArtistCreditLink from './ArtistCreditLink';
 
-const COLLAPSE_THRESHOLD = 4;
+const TO_SHOW_BEFORE = 4;
+const TO_TRIGGER_COLLAPSE = TO_SHOW_BEFORE + 2;
 
 const buildWorkArtistRow = (artistCredit: ArtistCreditT) => {
   return (
@@ -47,7 +48,7 @@ const WorkArtists = ({artists}: WorkArtistsProps) => {
   };
 
   const tooManyArtists = artists
-    ? artists.length > COLLAPSE_THRESHOLD
+    ? artists.length >= TO_TRIGGER_COLLAPSE
     : false;
 
   return (
@@ -56,7 +57,7 @@ const WorkArtists = ({artists}: WorkArtistsProps) => {
         {(tooManyArtists && !expanded) ? (
           <>
             <ul {...containerProps}>
-              {artists.slice(0, COLLAPSE_THRESHOLD).map(
+              {artists.slice(0, TO_SHOW_BEFORE).map(
                 artist => buildWorkArtistRow(artist),
               )}
               <li className="show-all" key="show-all">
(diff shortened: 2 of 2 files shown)


root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -123,7 +123,7 @@ const ReleaseEvents = ({
                   title={l('Show all release events')}
                 >
                   {bracketedText(texp.l('show {n} more', {
-                    n: events.length - COLLAPSE_THRESHOLD,
+                    n: events.length - 3,
```
mwiencek, line 126: I'd still keep the constant though, just with the offset correction added:
```suggestion
                    n: events.length - COLLAPSE_THRESHOLD + 1,
```

reosarevok: I didn't because it seemed arbitrary, but maybe? If anything, it'd feel better to have `events.length - (COLLAPSE_THRESHOLD - 2 + 1)`, since that's what we actually do. But I'll make the suggested change for now :)

reosarevok:
WorkArtists is fine, since it doesn't cut before or after. It just shows COLLAPSE_THRESHOLD artists and then the text.

root/static/scripts/common/components/ReleaseEvents.js
mwiencek: I can see what you mean. That might be more clear, particularly if we had a named constant for the `2` like `COLLAPSE_MIN` or something. I'd mainly just like that if we changed `COLLAPSE_THRESHOLD` and nothing else, the code would still work. :)

root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -18,6 +18,8 @@ import isDateEmpty from '../utility/isDateEmpty';
 import EntityLink from './EntityLink';
 
 const COLLAPSE_THRESHOLD = 4;
+const BEFORE_COLLAPSE = 2;
+const AFTER_COLLAPSE = 1;
```
yvanzo, line 22: Thanks for documenting hard-coded values but I can hardly understand the exact meaning of these constants.

I would suggest to:
1. Replace `THRESHOLD` (which is a bit ambiguous: collapse at it or after it?)
   with `TRIGGER` which is less ambiguous to me but must be set to 5 instead of 4.
2. Replace `BEFORE_COLLAPSE`
   with `EXPAND_ITEM_HEIGHT` which is the height (in number of normal items, rounded up) taken by the expand item.
3. Drop `AFTER_COLLAPSE` which might not be needed after the two above replacements.

```suggestion
const COLLAPSE_TRIGGER = 5;
const EXPAND_ITEM_HEIGHT = 2;
```

The rest of the code would need to be adjusted accordingly.

yvanzo: I think I got it: `BEFORE_COLLAPSE` and `AFTER_COLLAPSE` are for the number of items that are displayed resp. above and below the (expand) `show-all` item. Thus the above suggestions 2. and 3. make lesser sense, even though I would still rename these two added constants to `ABOVE_EXPAND_ITEM` and `BELOW_EXPAND_ITEM` resp. (_expand_ or _show all_).

I just noticed that artists are not collapsed the same way as release events: No artist is displayed below the `show-all` item, and the `show-all` item height is not taken into account. Should we uniformize to release events’ collapsed display (2+1) rather than artists’ collapsed display (4+)?

reosarevok: I would say no because the split has a clear purpose on release events ("this are the first and last releases, by date") but it wouldn't have on artists (I'm not sure even how those are sorted, just alphabetically possibly?)

mwiencek: I didn't initially intend for the number of items before & after to be configurable. `COLLAPSE_THRESHOLD` was just the max number of items before it collapsed things...so:

 * If COLLAPSE_THRESHOLD=4 it'll show 4, but if there's 5 it'll show 2 + (2 hidden) + 1
 * If COLLAPSE_THRESHOLD=5 it'll show 5, but if there's 6 it'll show 3 + (2 hidden) + 1
 * etc.

So if we decided we wanted to allow showing more initially, we could just update `COLLAPSE_THRESHOLD` and that's it. That still seems ideal to me, though a constant for the "2 hidden" might still be useful (I suggested `COLLAPSE_MIN`) above.

yvanzo: It’s more about documenting code through constants' name than allowing to configure.

* _threshold_ is ambiguous because it does not specify it actually is an upper bound (that is, nothing happens up to 4) and not a lower bound (which would be the suggested _trigger_ replacement set to 5).

* _collapse_before_ and _collapse_after_ are misleading because it is about `show-all` button to expand the list, not to collapse it (hence the suggested replacements).

reosarevok: Agree with @yvanzo here. I feel the constants make the code much more clear. I went with this in the latest version of the commit:

```
const TO_TRIGGER_COLLAPSE = 4;
const TO_SHOW_BEFORE = 2;
const TO_SHOW_AFTER = 1;
```
I think those are the clearer names for the constants here.

yvanzo: > `const TO_TRIGGER_COLLAPSE = 4;`

Should be `5` instead of `4` though, right?

root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -17,7 +17,9 @@ import isDateEmpty from '../utility/isDateEmpty';
 
 import EntityLink from './EntityLink';
 
-const COLLAPSE_THRESHOLD = 4;
+const TO_TRIGGER_COLLAPSE = 4;
```
mwiencek, line 20: But if you have `TO_SHOW_BEFORE` and `TO_SHOW_AFTER`, I'm not sure this makes sense to configure separately - I'd think you should just do `const TO_TRIGGER_COLLAPSE = TO_SHOW_BEFORE + TO_SHOW_AFTER + 1` below (or + 2 rather, since you want to hide at least two, and since that's technically the "trigger," but you'll need to change `>` to `>=` below too in that case).

root/static/scripts/common/components/ReleaseEvents.js
```diff
@@ -123,11 +125,14 @@ const ReleaseEvents = ({
...
-                    n: events.length - COLLAPSE_THRESHOLD,
+                    n: events.length - (TO_SHOW_BEFORE + TO_SHOW_AFTER),
                   }))}
                 </a>
               </li>
-              {buildReleaseEventRow(events[events.length - 1], abbreviated)}
+              {buildReleaseEventRow(
+                events[events.length - TO_SHOW_AFTER],
+                abbreviated,
+              )}
```
mwiencek, line 135: This needs to be changed to use a slice like `TO_SHOW_BEFORE` has, 'cause right now this does/means something different (an offset of a single item from the end)

reosarevok: I think it should be fine now!

root/static/scripts/common/components/WorkArtists.js
```diff
@@ -14,7 +14,8 @@ import {bracketedText} from '../utility/bracketed';
 
 import ArtistCreditLink from './ArtistCreditLink';
 
-const COLLAPSE_THRESHOLD = 4;
+const TO_SHOW_BEFORE = 4;
+const TO_TRIGGER_COLLAPSE = TO_SHOW_BEFORE + 1;
```
yvanzo, line 18: Even though it correctly reproduces existing behavior, it should be `+ 2` (as for collapsed areas) because the “Show {n} more” item takes more place in height than 1 normal item (the shortest), so it makes no sense to hide just 1 item.

```suggestion
const TO_TRIGGER_COLLAPSE = TO_SHOW_BEFORE + 2;
```

reosarevok: As you prefer :) Changed.

PR #1451, opened 2020-04-04
MBS-10740: Convert /relationship static pages to React

reosarevok:
Converts:
`/relationship/linkattributetype/in_use.tt`
`/relationship/linkattributetype/index.tt`
`/relationship/linktype/in_use.tt`
`/relationship/linktype/index.tt`
`/relationship/linktype/tree.tt`
and removes `/relationship/linkattributetype/common.tt` which was only used by `/relationship/linkattributetype/index.tt`

Also implements MBS-10741

PR #1453, opened 2020-04-05
MBS-10107: Add overview page for rel attributes

reosarevok:
### Implement MBS-10107

There's no good reason why we don't have these, and it would be good if relationship-attribute/$gid wasn't an error.

If we have them, it makes sense to go to them and not the main list after creating or adding an attribute, so I made that change too, and linked to them from the attributes list.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1451

Screenshot from 2020-04-06 01-16-04

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
+++ b/lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
@@ -26,14 +26,14 @@ sub _load_tree
 
 sub base : Chained('/') PathPart('relationship-attribute') CaptureArgs(0) { }
 
-sub index : Path('/relationship-attributes') Args(0)
+sub list : Path('/relationship-attributes') Args(0)
 {
     my ($self, $c) = @_;
 
     $self->_load_tree($c);
 
     $c->stash(
-        component_path  => 'relationship/linkattributetype/RelationshipAttributeTypesIndex',
+        component_path  => 'relationship/linkattributetype/RelationshipAttributeTypesList',
         component_props => {root => $c->stash->{root}->TO_JSON},
         current_view    => 'Node',
     );
@@ -62,7 +62,7 @@ sub create : Path('/relationship-attributes/create') Args(0) RequireAuth(relatio
             );
         });
 
-        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/index'));
+        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/list'));
         $c->detach;
     }
 }
@@ -93,7 +93,7 @@ sub edit : Chained('load') RequireAuth(relationship_editor)
             );
         });
 
-        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/index'));
+        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/list'));
         $c->detach;
     }
 }
@@ -130,7 +130,7 @@ sub delete : Chained('load') RequireAuth(relationship_editor) SecureForm
             );
         });
 
-        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/index'));
+        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/list'));
         $c->detach;
     }
 }
--- a/lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
+++ b/lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
@@ -6,7 +6,7 @@ use MusicBrainz::Server::Constants qw(
     $EDIT_RELATIONSHIP_ATTRIBUTE
     $INSTRUMENT_ROOT_ID
 );
-
+use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 use MusicBrainz::Server::Validation qw( is_guid );
 
 with 'MusicBrainz::Server::Controller::Role::Load' => {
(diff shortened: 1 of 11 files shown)


root/relationship/linkattributetype/RelationshipAttributeTypesList.js
```diff
@@ -68,6 +68,10 @@ const AttributeDetails = ({
             </a>
           </>
         )}
+        {' | '}
+        <a href={'/relationship-attribute/' + attribute.gid}>
+          {l('Documentation')}
+        </a>
```
mwiencek, line 74: I was super confused why I couldn't find this, and it's 'cause it's currently under `$c.user?.is_relationship_editor`. :) (I think it can be shown even if you're not logged in.)

lib/MusicBrainz/Server/Data/LinkType.pm
```diff
@@ -94,6 +94,18 @@ around get_by_gid => sub
...
+sub find_by_attribute
+{
+    my ($self, $attribute_id, $limit, $offset) = @_;
+    my $query = "SELECT " . $self->_columns . "
+                 FROM " . $self->_table . "
+                     JOIN link_type_attribute_type ltat ON ltat.link_type = link_type.id
+                 WHERE ltat.attribute_type = ?
+                 ORDER BY link_type.name COLLATE musicbrainz";
+
+    $self->query_to_list_limited($query, [$attribute_id], $limit, $offset);
```
mwiencek, line 106: Don't think this needs to use `query_to_list_limited`, since the number of results will always be small (and we're showing all of them in the one place that uses this).
```suggestion
    my ($self, $attribute_id) = @_;
    my $query = "SELECT " . $self->_columns . "
                 FROM " . $self->_table . "
                     JOIN link_type_attribute_type ltat ON ltat.link_type = link_type.id
                 WHERE ltat.attribute_type = ?
                 ORDER BY link_type.name COLLATE musicbrainz";

    $self->query_to_list($query, [$attribute_id]);
```

lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
```diff
@@ -26,14 +26,36 @@ sub _load_tree
...
-sub index : Path('/relationship-attributes') Args(0)
+sub show : PathPart('') Chained('load')
+{
+    my ($self, $c) = @_;
+
+    my $attribute = $c->model('LinkAttributeType')->get_tree(
+        undef, # don't filter
+        $c->stash->{link_attr_type}->{id}
+    );
+    my ($relationships, $hits) = $c->model('LinkType')->find_by_attribute($attribute->{id});
```
mwiencek, line 37: (assuming change to use `query_to_list`)
```suggestion
    my @relationships = $c->model('LinkType')->find_by_attribute($attribute->{id});
```

lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
```diff
@@ -26,14 +26,36 @@ sub _load_tree
...
+
+    my %props = (
+        attribute => $attribute,
+        relationships => $relationships,
```
mwiencek, line 41: (assuming change to use `query_to_list`)
```suggestion
        relationships => \@relationships,
```

root/relationship/linkattributetype/RelationshipAttributeTypeIndex.js
```diff
@@ -0,0 +1,197 @@
...
+          {' '}
+        </>
+      ) : null}
+      {bracketedText(attribute.child_order.toString())}
+
+      {childrenAttrs.length ? (
+        <ul>
+          {childrenAttrs
+            .slice(0)
+            .sort((a, b) => compareChildren(a, b))
```
mwiencek, line 50: Don't these already come sorted from the server? Seems like they should be if not. But anyway, you can pass the sort function directly here:
```suggestion
            .sort(compareChildren)
```

root/relationship/linkattributetype/RelationshipAttributeTypeIndex.js
```diff
@@ -0,0 +1,197 @@
...
+          ) : null}
+        </p>
+
+        {childrenAttrs.length ? (
+          <>
+            <h2>{l('Possible values')}</h2>
+            <ul>
+              {childrenAttrs
+                .slice(0)
+                .sort((a, b) => compareChildren(a, b))
```
mwiencek, line 142:
```suggestion
                .sort(compareChildren)
```

root/relationship/linkattributetype/RelationshipAttributeTypeIndex.js
```diff
@@ -0,0 +1,197 @@
...
+            </ul>
+          </>
+        ) : null}
+
+        <h2>{l('Relationship usage')}</h2>
+        {relationships.length ? (
+          <>
+            <p>
+              {l(`This relationship type is being used
+                  by the following relationships:`)}
```
mwiencek, line 158:
```suggestion
              {l(`This attribute is being used
                  by the following relationship types:`)}
```

root/relationship/linkattributetype/RelationshipAttributeTypeIndex.js
```diff
@@ -0,0 +1,197 @@
...
+                    {expand2react(l_relationships(relationship.description))}
+                  </p>
+                ) : null}
+              </React.Fragment>
+            ))}
+          </>
+        ) : (
+          <p>
+            {l(`This relationship type isn’t directly being used
+                by any relationships.`)}
```
mwiencek, line 189:
```suggestion
            {l(`This attribute isn’t directly being used
                by any relationship types.`)}
```
BTW, not sure how relevant this message is on child attribute types, since link type attributes can *only* refer to root attribute types. But it probably doesn't matter too much.

root/relationship/linkattributetype/RelationshipAttributeTypeIndex.js
reosarevok: They don't, currently, no. But did change to pass the function directly at least.

root/relationship/linkattributetype/RelationshipAttributeTypesList.js
reosarevok: That's from a different file which I was just renaming, but I can change that in a separate commit I guess :)

mwiencek (approved):

mwiencek:
Rebased this, made some TO_JSON fixes, and re-tested it

PR #1454, opened 2020-04-06
MBS-10640: Fix donation check

reosarevok:
### Fix MBS-10640

This was only saying you needed to donate if the MeB site returned a nag status of 1: "known editor, donation time has passed". It was not saying it when it returned -1 ("we don't know of this editor, has never donated before").

This change makes it so that only 0 ("no-nag period applies currently") is shown as "you don't need
to donate".

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Account.pm
+++ b/lib/MusicBrainz/Server/Controller/Account.pm
@@ -7,6 +7,7 @@ use Digest::SHA qw(sha1_base64);
 use JSON;
 use List::MoreUtils qw( uniq );
 use MusicBrainz::Server::ControllerUtils::JSON qw( serialize_pager );
+use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -689,12 +689,15 @@ sub donation : Local RequireAuth HiddenOnSlaves
...
+    # If nag is 0, don't nag - if 1 or -1 then nag
+    my $nag = $result->{nag} != 0;
+
     $c->stash(
         current_view => 'Node',
         component_path => 'account/Donation',
         component_props => {
             days => sprintf("%.0f", $result->{days}),
-            nag => $result->{nag} > 0,
+            nag => $nag,
```
mwiencek, line 700: Fix looks good, but while you're at it, can you do `boolean_to_json($nag)` here since the type in the file is `+nag: boolean,`?

PR #1455, opened 2020-04-07
MBS-10748: Convert relationship type page to React

reosarevok:
### Implements MBS-10748

On top of https://github.com/metabrainz/musicbrainz-server/pull/1451


root/static/scripts/common/components/Relationship.js
```diff
@@ -0,0 +1,178 @@
...
+    />,
+  );
+  const extraAttributes = getExtraAttributes(
+    relationship,
+    'link_phrase',
+    false, /* forGrouping */
+  );
+  return (
+    <>
+      {longPhrase}
```
yvanzo, line 143: How did you convert `verbose_phrase` and https://github.com/metabrainz/musicbrainz-server/blob/adeba5c7bf9754f40b774cf7f08b8ba62fb90ce3/root/components/common-macros.tt#L649-L651
?

reosarevok: verbose_phrase_with_placeholders is the equivalent of interpolate with long_link_phrase. The only cases where we didn't call with verbose_phrase were historical edits, which hack the link_phrase field, so I dealt with those separately in HistoricRelationshipContent.

root/static/scripts/common/components/OrderableDirection.js
```diff
@@ -0,0 +1,40 @@
...
+      directionName = l('None');
+      break;
+    case 1:
+      directionName = l('Forward');
+      break;
+    case 2:
+      directionName = l('Backward');
+      break;
+    default:
+      directionName = l('Unknown');
```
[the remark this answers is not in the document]
reosarevok: It doesn't if everything works well, but I was under the impression we were supposed to have defaults. Removed then.

root/static/scripts/common/components/Relationship.js
```diff
@@ -0,0 +1,178 @@
...
+import type {LinkPhraseProp} from '../../edit/utility/linkPhrase';
+
+import DescriptiveLink from './DescriptiveLink';
+
+type Props = {
+  +phraseField?: LinkPhraseProp,
+  +relationship: RelationshipT,
+};
+
+const RelationshipDates = ({relationship}: {relationship: RelationshipT}) => {
```
mwiencek, line 27: This looks like it re-implements root/utility/relationshipDateText.js - can we use that instead?

root/static/scripts/common/components/Relationship.js
```diff
@@ -0,0 +1,178 @@
...
+        content={relationship.entity1_credit}
+        entity={relationship.target}
+      />
+      <RelationshipDates relationship={relationship} />
+      {extraAttributes ? (
+        <>
+          {' '}
+          {bracketedText(commaOnlyListText(extraAttributes.map(
+            attribute => attribute.typeName,
+          )))}
```
mwiencek, line 102: Looks like this should use root/static/scripts/common/utility/displayLinkAttribute.js

root/static/scripts/common/components/Relationship.js
```diff
@@ -0,0 +1,178 @@
...
+    <>
+      {longPhrase}
+      {' '}
+      <RelationshipDates relationship={relationship} />
+      {extraAttributes ? (
+        <>
+          {' '}
+          {bracketedText(commaOnlyListText(extraAttributes.map(
+            attribute => attribute.typeName,
+          )))}
```
mwiencek, line 151: Also looks like this should use root/static/scripts/common/utility/displayLinkAttribute.js

root/types.js
```diff
@@ -840,7 +845,10 @@ declare type RelationshipT = {
   +id: number,
   +linkOrder: number,
   +linkTypeID: number,
+  +source: CoreEntityT,
+  +source_credit: string,
   +target: CoreEntityT,
+  +target_credit: string,
```
mwiencek, line 851: These three additions are now unused AFAICT.

root/relationship/linktype/RelationshipTypeIndex.js
```diff
@@ -0,0 +1,232 @@
...
+  from '../../static/scripts/common/utility/formatEntityTypeName';
+
+type Props = {
+  +$c: CatalystContextT,
+  +relType: LinkTypeT,
+};
+
+const RelationshipTypeIndex = ({
+  $c,
+  relType,
```
mwiencek, line 32: Up to you if you think this change makes sense, but we do tend to use `linkType` more often for naming variables that refer to link types, since that's how they're called everywhere else. Though the page/component name is  `RelationshipTypeIndex` so I can see how that follows from that, too
```suggestion
  linkType,
```

root/relationship/linktype/RelationshipTypeIndex.js
```diff
@@ -0,0 +1,232 @@
...
+  const type1 = relType.type1;
+  const formattedType0 = formatEntityTypeName(type0);
+  const formattedType1 = formatEntityTypeName(type1);
+  const possibleAttributes = [];
+  Object.keys(relType.attributes)
+    .map(attributeTypeId => (
+      possibleAttributes.push(
+        linkedEntities.link_attribute_type[parseInt(attributeTypeId, 10)],
+      )
+    ));
```
mwiencek, line 47: The callback passed to `map` will convert each value for you, so you can simplify this a bit.
```suggestion
  const possibleAttributes = Object.keys(relType.attributes).map(attributeTypeId => (
    linkedEntities.link_attribute_type[parseInt(attributeTypeId, 10)]
  ));
```

root/relationship/linktype/RelationshipTypeIndex.js
```diff
@@ -0,0 +1,232 @@
...
+  const sortedPossibleAttributes = sortBy(possibleAttributes, 'name');
```
mwiencek, line 48: Importing localizeLinkAttributeTypeName and i18n.compare might make sense here. We can also sort this in-place instead of making a copy.
```suggestion
  possibleAttributes.sort((a, b) => compare(
    localizeLinkAttributeTypeName(a),
    localizeLinkAttributeTypeName(b),
  ));
```

root/types.js
```diff
@@ -690,6 +690,11 @@ declare type LinkTypeT = {
   +cardinality1: number,
   +children?: $ReadOnlyArray<LinkTypeT>,
   +deprecated: boolean,
+  +documentation?: string,
```
mwiencek, line 693: I think this should be
```suggestion
  +documentation: string | null,
```
since `$json->{documentation} = $self->documentation;` means the property will always be present, though it may be serialized as `null` if `$self->documentation` is `undef` on the Perl side.

root/static/scripts/common/components/Relationship.js
reosarevok: I had made that change (well, displayLinkAttributes because I see no issue with it being commaList instead) in the relationship edit PR, but I guess I can move it to here already.

root/relationship/linktype/RelationshipTypeIndex.js
reosarevok: I personally prefer relType, but that's because I'd like to eventually get rid to references in the code to stuff we no longer use anywhere else. I could change to linkType for now and change everything together eventually though, if you prefer...

PR #1456, opened 2020-04-08
MBS-10719 / MBS-10751: Convert Remove PUID edit to React and fix some of them not loading

reosarevok:
### Implement MBS-10751, fix MBS-10719


mwiencek:
> I'm no longer sure what's the point of Maybe[] then, but Optional[] seems to be what we actually want here.

`Maybe` allows the property to be `undef`, but it still has to exist:

```
$ rlwrap re.pl
$ use MooseX::Types::Moose qw( Int Maybe );
$ use MooseX::Types::Structured qw( Dict );
$ my $constraint = Dict[recording_puid_id => Maybe[Int]];
MooseX::Types::Structured::Dict[recording_puid_id,Maybe[Int]]
$ $constraint->validate({recording_puid_id => 1}); # ok
$ $constraint->validate({recording_puid_id => undef}); # ok
$ $constraint->validate({}); # not ok:
Validation failed for 'Dict[recording_puid_id,Maybe[Int]]' with value "{  }, Internal Validation Error is: \n [+] Validation failed for 'Maybe[Int]' with value "NULL""
```

root/edit/details/historic/RemovePuid.js
```diff
@@ -0,0 +1,38 @@
...
+import * as React from 'react';
+
+import DescriptiveLink
+  from '../../../static/scripts/common/components/DescriptiveLink';
+
+type RemovePuidEditT = {
+  ...EditT,
+  +display_data: {
+    puid_name: string,
+    recording: RecordingT,
```
mwiencek, line 19:
```suggestion
    +puid_name: string,
    +recording: RecordingT,
```

root/edit/details/historic/RemovePuid.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+};
+
+const RemovePuid = ({edit}: {edit: RemovePuidEditT}) => (
```
mwiencek, line 23:
```suggestion
const RemovePuid = ({edit}: {+edit: RemovePuidEditT}) => (
```

root/edit/details/historic/RemovePuid.js
```diff
@@ -0,0 +1,38 @@
...
+  <>
```
mwiencek, line 24: The surrounding `<></>` can be removed since the component only returns one element

reosarevok:
Good points, thanks!

PR #1459, opened 2020-04-12
MBS-10760: Convert Remove track edit to React 

reosarevok:
### Implement MBS-10760

Changes from the TT:

* Added the edit classes to be more consistent and improve the display of the data.
* Added a mock recording to the display data when we don't have it to avoid DescriptiveLink dying.
* Call DeletedLink directly when a release is null.
* Print "Releases:" -> "No releases found." (translation exists already) rather than just an empty "Releases:" -> ""  if the releases array is empty. I'd use DeletedLink, but I'm not sure if this was also used for non-album tracks, in which case implying a release used to exist would be incorrect.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1456


mwiencek:
> Print "Releases:" -> "No releases found." (translation exists already) rather than just an empty "Releases:" -> "" if the releases array is empty. I'd use DeletedLink, but I'm not sure if this was also used for non-album tracks, in which case implying a release used to exist would be incorrect.

This doesn't seem to be what the current PR does. Was it changed?

yvanzo:
> This doesn't seem to be what the current PR does. Was it changed?

It actually displays `[removed]` which is still better than nothing.

reosarevok:
Yes, that's the outdated message, whoops. Will fix.

yvanzo:
> It looks much better with the new table than with the old bullet list.

Actually the code is the same but for some reason the bullet does not show up anymore.

reosarevok:
It's the "details" class IIRC :)

yvanzo:
Richtig! https://github.com/metabrainz/musicbrainz-server/blob/b1a4eb22816b8da00ed055c7a402991f3b2b03af/root/static/styles/layout.less#L794-L826

reosarevok:
Oh, actually nvm, I fixed the commit message, just didn't change it here. So I guess it's fine :)

PR #1460, opened 2020-04-13
MBS-10761: Disallow YouTube links at wrong levels

reosarevok:
### Implement MBS-10761

Users keep wrongly adding channel links to recordings and video links to artists. This commit disallows that, since there's no case where a video link is a channel nor a channel is a video link.

For video links we already standardise them to the "watch?v=" syntax so it should be safe to disallow all others. For channel links YouTube seems to have a ton of different options, all of them valid, so this just checks it's a YouTube video and not a watch video.

Releases are sometimes officially put up on YouTube as playlists. As such, it makes sense to allow playlist links for them.


reosarevok:
Good point, release should support playlists. Will fix.

reosarevok:
Why? Series YouTube is specifically meant to be a channel-level relationship.

yvanzo:
Maybe it should be changed to allow playlist then because there would be many real use cases for it.
For example, most radio stations have a video channel and use playlists for their radio programmes.

reosarevok:
I mean, you could also have a label that has a channel and puts stuff for each artist in its own playlist instead of having separate channels. I don't think that makes those channels, though. Maybe we should have a separate relationship for that type of thing if it's common enough...

yvanzo:
[`Series-URL  / Video channel`](https://musicbrainz.org/relationship/71774032-781b-468c-9cbf-8a9a2f8eda13) relationship type is described to link “to a channel, playlist, or user page on a video sharing site” already. I suggest to extend the description of [`Series-URL / Youtube`](https://musicbrainz.org/relationship/f23802a4-36be-3751-8e4d-93422e08b3e8) to allow for both playlist and video channel  rather than just `/user/`entry, that description is outdated anyway.

Please add a test for `series` too. Here is an example of recurring show: https://www.youtube.com/playlist?list=PL43OynbWaTMKSxLVnUF0HbHHiXEgAVm3Q

reosarevok:
Just realized we're actually already not blocking playlist for series anyway - this would only block video links. So it doesn't seem like we actually need to change anything, unless we *do* want to block playlist for everything except series.

yvanzo:
> Just realized we're actually already not blocking playlist for series anyway

Cool, please add a test to represent this use case.

PR #1463, opened 2020-04-13
MBS-10764: Convert historic Remove Label Alias edit to React 

reosarevok:
### Implement MBS-10764

The only change from the TT is to add the "remove-label-alias" class, for consistency.

Amazingly enough, it seems no info about the label itself is stored with these old edits, so all we do is literally print the alias being removed as a string.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1462

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Historic/RemoveLabelAlias.pm
+++ b/lib/MusicBrainz/Server/Edit/Historic/RemoveLabelAlias.pm
@@ -12,7 +12,7 @@ sub edit_name { N_l('Remove label alias') }
 sub edit_kind { 'remove' }
 sub historic_type { 62 }
 sub edit_type { $EDIT_HISTORIC_REMOVE_LABEL_ALIAS }
-sub edit_template { 'historic/remove_label_alias' }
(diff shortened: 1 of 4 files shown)


reosarevok:
Not separately, no. Just search for the oldest ones first. Some edit types are all grouped under one name (Edit release has a ton) so the only surefire way of finding edits is by checking the edit type and then querying the DB for edits with that type.

yvanzo:
Shouldn’t we relabel historic edits with appended “(historic)” to make those searchable?

reosarevok:
I think this is intentional - there's no reason for the users these should be split, since as far as they care, it's the same kind of edit - we just need different code for it.

yvanzo:
One reason would be for them to be aware that such edit is not abnormally incomplete, it is just historic.

reosarevok:
That's fair, and we could consider implementing a way of showing the difference on the edit display, but while still grouping them so that they don't need to search for all sorts of different "Edit release" or whatever edits :) That's outside the scope of this IMO though :)

PR #1464, opened 2020-04-13
MBS-10765: Convert historic Add, Move and Remove DiscID edits to React

reosarevok:
### Implement MBS-10765

On top of https://github.com/metabrainz/musicbrainz-server/pull/1463


reosarevok:
As in, it was being used from data instead of display_data :)

yvanzo:
Thanks but it would have been more useful in commit message. Not a big deal, just a friendly reminder. :)

PR #1469, opened 2020-04-17
MBS-10773: Convert historic Change Release Quality edit to React

reosarevok:
### Implement MBS-10773

Changes from the TT:

Use DescriptiveLink to also display the artist credit for the release if available.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1464


root/static/scripts/common/utility/formatDataQualityName.js
```diff
@@ -0,0 +1,22 @@
+/*
+ * @flow strict
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export default function formatDataQualityName(quality: QualityT) {
```
mwiencek, line 10: Hmm, we do already have a `QUALITY_NAMES` constant in root/static/scripts/common/constants.js which root/layout/components/sidebar/SidebarDataQuality.js uses by doing `const name = QUALITY_NAMES.get(quality);`. It doesn't handle the default "Unknown" case, but I guess the TT didn't either. Is there a situation where we'd show that? It'd be nice to only have one way of doing things in any case.

PR #1470, opened 2020-04-17
MBS-10359: Guess feat. adds bogus trailing join phrases

mwiencek:
I was able to reproduce this by re-importing the release in https://musicbrainz.org/edit/64567170, selecting the release artist first, then using the release-level guess-feat. button.

The problem appears to be that the artist credit name objects are getting shared internally, so modifying the join phrase on one AC name would potentially modify the AC name on an entirely unrelated track. (Note that this is a good example of why future React-based edit page rewrites should never use mutable state.)

The added test case consistently fails without the cloneDeep change.


mwiencek:
@brainzbot retest this please

PR #1473, opened 2020-04-20
Upgrade a bunch of JavaScript packages

yvanzo:
Mostly small game.


yvanzo:
@brainzbot, retest this please.

mwiencek:
I think there are some breaking changes in shelljs. The rest looks good though.

yvanzo:
@brainzbot, retest this please.

webpack.client.config.js
```diff
@@ -108,7 +108,7 @@ function createJsPo(srcPo, lang) {
    * possible terminal paths.
    */
   const scriptsDir = shellQuote.quote([dirs.SCRIPTS]);
-  const nestedDirs = shell.exec(`find ${scriptsDir} -type d`, {silent: true}).output.split('\n');
+  const nestedDirs = _(shell.exec(`find ${scriptsDir} -type d`, {silent: true}).stdout).split('\n');
```
mwiencek, line 111: I don't think we need to create another lodash chain here. Did you run into a case where that's needed?
```suggestion
  const nestedDirs = shell.exec(`find ${scriptsDir} -type d`, {silent: true}).stdout.split('\n');
```

mwiencek: Maybe when stdout is a Buffer. I'd probably use `String()` over `_()` for clarity of intent though.

yvanzo: I thought that was needed but that was just a syntax error, thanks.

PR #1475, opened 2020-04-20
MBS-10717, MBS-10778: Set SameSite=None on session cookie and add CSRF tokens to user/admin forms

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10717
https://tickets.metabrainz.org/browse/MBS-10778

~~TODO:~~

- [x] Regenerate cpanfile.snapshot
- [x] Fix Flow types
- [x] Add a test for the invalid CSRF token check


mwiencek:
@yvanzo Thanks, should be added to all of those now.

I refactored the implementation a bit so that it requires less controller changes (which were due to `$c->validate_csrf_token` previously calling `$c->detach`). Now we just need to add a `CSRFToken` attribute to the Catalyst action. Though, I did also have to refactor the `submitted_and_valid` checks we have to handle this attribute...so it probably did end up being around the same amount of changes overall, but should be a clearer implementation, I hope.

Also had to refactor some `withCatalystContext` stuff in order for `FormCsrfToken` to be able to access the sanitized context data on the client, and for the Flow types to work. :/

lib/MusicBrainz/Server/Controller/Root.pm
```diff
@@ -365,7 +365,11 @@ sub end : ActionClass('RenderView')
 {
     my ($self, $c) = @_;
 
-    return if exists $c->action->attributes->{Minimal};
+    my $attrs = $c->action->attributes;
+
+    $c->generate_csrf_token if exists $attrs->{CSRFToken};
```
mwiencek, line 370: We now generate the token in `Controller::Root::end` (which is called before the template renders on every request), but only if `CSRFToken` is set on the action. Which I think is nicer than having to manually call `generate_csrf_token` at the end of every action.

lib/MusicBrainz/Server.pm
```diff
@@ -480,6 +506,67 @@ has json_utf8 => (
...
+        unless defined $params;
+
+    return 0 unless
+        %{$params} &&
+        $form->process(params => $params) &&
+        $form->has_params;
+
+    return 0 if
+        exists $self->action->attributes->{CSRFToken} &&
+        !$self->validate_csrf_token;
```
mwiencek, line 529: `$c->validate_csrf_token` is now called as part of `$c->form_submitted_and_valid`. Hence why a lot of `$form->submitted_and_valid` checks have changed to that.

Note: The form still has to be processed first, otherwise submitted fields won't be preserved.

yvanzo: Ok, `$self->action->attributes->{CSRFToken}` is checked to validate forms that don’t need CSRF token yet since it is defined server-side. :+1:

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -134,7 +134,7 @@ sub delete_user : Path('/admin/user/delete') Args(1) RequireAuth HiddenOnSlaves
     }
 }
 
-sub edit_banner : Path('/admin/banner/edit') Args(0) RequireAuth(banner_editor) {
+sub edit_banner : Path('/admin/banner/edit') Args(0) RequireAuth(banner_editor) CSRFToken {
```
yvanzo, line 137: I was wondering why some subroutines had to separately check `validate_csrf_token` and found out that `form_posted_and_valid` just call  `form_submitted_and_valid`. That is explained in the previous commit. :+1:

mwiencek: Yeah, e.g. in `delete_user` above I manually call `validate_csrf_token`. Because apparently we only check that it's a POST request there - there's no Form to validate against.

PR #1481, opened 2020-04-28
MBS-10791: Convert historic Change RG edit to React

reosarevok:
### Implement MBS-10791

Changes from the TT:

On top of https://github.com/metabrainz/musicbrainz-server/pull/1480


reosarevok:
Not directly. For testing purposes, you could just change the edit name temporarily so that they get their own section in the edit search :)

PR #1483, opened 2020-04-28
MBS-10793: Convert historic Move Release edit to React 

reosarevok:
### Implement MBS-10793

Changes from the TT:

Actually attempt to link the artists if possible, rather than just printing their names. That might be because a fair amount of them have their IDs missing (one example had Metallica with an ID of 0) but we should still try to link them where we can.

Additionally, this actually tries to pass move_tracks to the template - it was trying to load it from display_data but AFAICT it wasn't being added to display_data at all. Not sure if it's ever not false, but just in case.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1482


reosarevok:
Oh, nice, so I guess it *did* get added sometimes, but because it wasn't passed to display_data at all, it showed nothing anyway.

PR #1485, opened 2020-04-29
Remove $c.user_exists from JS

mwiencek:
`user_exists` exists on the Perl side to check if a user is logged in without possibly fetching that user from the database. It serves no purpose in our JavaScript; if the user is logged in, `$c.user` will be defined.


mwiencek:
I just didn't like that we had two different ways of checking this. Also, checking `$c.user_exists` doesn't Flow-refine `$c.user` as being truthy for you if you need that.

PR #1474, opened 2020-04-20
MBS-10777: Convert Add/Remove Relationship edits to React

reosarevok:
### Implement MBS-10777, fix MBS-10774

On top of https://github.com/metabrainz/musicbrainz-server/pull/1455

PR #1487, opened 2020-05-01
MBS-10801, MBS-10802: Convert Reorder Relationships edit to React

reosarevok:
### Implements MBS-10801, MBS-10802

This should be the last use of the display_relationship TT macro so this removes that too.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1474


reosarevok:
Yes, this was getting an ISE when one of the entities was undefined. It should be fixed now!

PR #1488, opened 2020-05-04
MBS-6528: Modify release editor button order

navap:
# Problem
The button order in the release editor goes: Cancel, Left, Right, Finish.

On the last tab I was expecting that the Enter button would replace the Finish button, but instead it was inserted on the far left.

# Solution
I reordered the buttons so that 'enter edit' and 'finish' are in the same spot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "eslint-plugin-react": "7.20.6",
     "eslint-plugin-react-hooks": "4.1.2",
     "file-url": "2.0.2",
-    "flow-bin": "0.146.0",
+    "flow-bin": "0.147.0",
     "gettext-parser": "3.1.0",
     "http-proxy": "1.18.1",
     "jsdom": "13.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3265,10 +3265,10 @@ flatted@^2.0.0:
   resolved "https://registry.yarnpkg.com/flatted/-/flatted-2.0.2.tgz#4575b21e2bcee7434aa9be662f4b7b5f9c2b5138"
   integrity sha512-r5wGx7YeOwNWNlCA0wQ86zKyDLMQr+/RB8xy74M4hTphfmjlijTSSXGuH8rnvKZnfT9i+75zmd8jcKdMR4O6jA==
 
-flow-bin@0.146.0:
-  version "0.146.0"
-  resolved "https://registry.yarnpkg.com/flow-bin/-/flow-bin-0.146.0.tgz#fafa002663a0e13bf3c08c3972dd93d68289ccc6"
-  integrity sha512-TP8eCwltqc7fo6ad5klgsrZ2veZIK2qM1vHf1A/cnXTStPh8hLAz1cOXmlQIFZR/7fjSMI39TS3CgF6M/HUhAw==
+flow-bin@0.147.0:
+  version "0.147.0"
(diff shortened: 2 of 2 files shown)


reosarevok:
This undoes MBS-3112 (enter edit is on the left because it's the first thing that should be tabbed to from the edit note field)

navap:
I think this is related to MBS-6528.

Tab order and physical placement of the button aren't dependent on each other, if there is a tab order requirement then a `tabindex` attribute should be used to set a custom order. 

Unfortunately I don't have an actual server set up and can't test a change like that, so what do you suggest here?

reosarevok:
I've read plenty of people saying tabindex > 1 is a bad idea (https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html for one)

navap:
I'm fine with the thought that tab order should follow the physical order, in fact I agree that it's a more natural order of tabbing. So then it comes down to what _is_ the appropriate order of buttons. 

I believe my proposed order is most appropriate based on the in-page UX convention set up by the other tabs within the release editor that show the 'next' and 'finish edit' button on the right. 

I think the idea that "the 'submit' button should be the first button tabbed to" (i.e. MBS-3112) is flawed. It is certainly the first button on most (all?) other edit pages, but what is true for a "simple" edit page with just one button underneath doesn't mean the same logic applies for the release editor where there are multiple buttons required.

mwiencek:
I hate to bikeshed here, but reverting the previous change doesn't seem like a good idea to me at this point. (I can imagine many people who were annoyed by the previous change are even used to the new order by now.)

Here's an alternate proposal, if we even want to risk make further changes here and annoying people all over again...

- Remove the cancel button.
- Have the following button orders on each tab:
  - Information: Next
  - Duplicates: Next / Finish / Prev
  - Tracklist: Next / Finish / Prev / Add Medium
  - Recordings: Finish / Prev
  - Edit note: Enter / Prev

Rationale: I think the cancel button is useless and only creates unnecessary risk of losing one's work. "Next" is the most logical thing to make first to me, since it's the action someone going through each part of the RE will want to take normally. On the recordings tab, we can remove "Next" (so Finish becomes first) because it's the second-to-last tab anyway. On the edit note tab, "Enter" is now in the place where Next/Finish typically are, which I think resolves navap's main issue here.

navap:
Sorry, I oppose the idea of putting the "wanted" action first. It doesn't make any sense to have the "Previous" button come after the "Next" button.

This is not a software prompt with the default "next" option being handy and first (e.g. an OS prompt for run, save, open, etc.). This is an option in a linear navigation flow as defined by a linear workflow through the tabs, and the context provided by the current button order themselves.

Coming from a part of the world that reads Left to Right, it is extremely jarring and unexpected to see the order in this manner.

I've been away from MB for a while so my perspective is perhaps closer to that 'new user' perspective as opposed to the 'regular, power-tabber user'. I can understand if you want to prioritize the wants of the latter over the expectations of the former.

(I'm not trying to be insensitive to users who read Right to Left, but the site is very clearly designed with a Left to Right audience in mind, hence my strong opinion)

mwiencek:
If you feel strongly about it, make a thread on the forums and we can see how other people feel. :) I'm sympathetic to your perspective but am also not inclined to change the order again without feedback from more than one person...

(Though even without making any other changes, I'd still +1 removing the cancel button.)

yvanzo:
There is probably more than only one way to do it right - the one proposed by @navap doesn’t confuse me at least, but I feel much stronger about not creating the same mess that led to MBS-10426 again.

PR #1489, opened 2020-05-04
MBS-10804: Remove redundant header with editor name

navap:
# Problem
Because this header is styled the same way as the edit note headers
(the ones with an actual edit note), having this header makes it look
like there is a comment missing.

**Before:**
image

**After:**
image

# Solution
Redundant header has been removed.


reosarevok:
This seems sensible to me actually (never even stopped to think what the header was good for...) but please do add a ticket for it so we can track it and show it to users :)

PR #1493, opened 2020-05-04
PostgreSQL 12 changes

mwiencek:
I think this is mostly done, ~but I need to create some tickets and resolve any test failures~.


lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -102,7 +102,8 @@ sub _where_filter
 
     if (defined $filter) {
         if (exists $filter->{name}) {
-            push @query, "(to_tsvector('mb_simple', release.name) @@ plainto_tsquery('mb_simple', ?) OR release.name = ?)";
+            push @query,
```
reosarevok, line 105: Why the extra push here?

mwiencek: Looks like a rebase issue. Fixed, thanks!

lib/MusicBrainz/Server/Data/Alias.pm
```diff
@@ -88,8 +88,8 @@ sub find_by_entity_ids
                    end_date_year NULLS LAST,
                    end_date_month NULLS LAST,
                    end_date_day NULLS LAST,
-                   musicbrainz_collate(sort_name),
-                   musicbrainz_collate(name)";
+                   sort_name COLLATE musicbrainz,
+                   name COLLATE musicbrainz";
```
mwiencek, line 92: Unfortunately creating a function that returns `input COLLATE musicbrainz` didn't work, because that doesn't affect the expression outside the function call. So I didn't see a way to do this without breaking the code (short of defining the collation in the table definition, but changing the data type of a column isn't that easy).

admin/sql/Extensions.sql
```diff
@@ -3,7 +3,42 @@ BEGIN;
...
+-- the "superuser" is restricted in creating C language functions, e.g.
+-- Amazon RDS.
+--
+-- [1] https://stackoverflow.com/a/11007216
+
+DO $$
+BEGIN
+  CREATE OR REPLACE FUNCTION public.immutable_unaccent(regdictionary, text)
+    RETURNS text LANGUAGE c IMMUTABLE PARALLEL SAFE STRICT AS
+  '$libdir/unaccent', 'unaccent_dict';
```
yvanzo, line 44: It would probably be useful to preload `unaccent` in `docker-postgres` repo.

mwiencek:
> As long as it doesn’t sort "02" before "1" 😉

Nope, the `colNumeric=yes` options seems to be working as intended otherwise: it sorts "02" after "1", "11" after "2", etc. It seems to just be when there's zero-padding on numbers that are otherwise equal. Unless I'm missing some other ICU option...

> I would be interested in seeing these differences. 👀

Here you go: https://gist.github.com/mwiencek/bd57be4210f3d68a09a3bfa6b735c0ba

mwiencek:
And here's the relevant musicbrainz-collate source, which AFAICT only sets the `UCOL_NUMERIC_COLLATION` attribute: https://github.com/metabrainz/postgresql-musicbrainz-collate/blob/master/musicbrainz_collate.c#L95

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   musicbrainz-tests:
     docker:
-      - image: metabrainz/musicbrainz-tests:v-2020-02
+      - image: metabrainz/musicbrainz-tests:v-2020-05-18
```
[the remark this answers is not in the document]
mwiencek: Sure, I don't mind which tag is used so changed it to `v-2020-05.0-pg12`.

admin/sql/Extensions.sql
mwiencek: I'll remember to check that when I submit changes there (and in docker-postgres-cluster, too). 👍

docker/Dockerfile.tests
```diff
@@ -105,10 +105,10 @@ RUN git clone --branch $PG_AMQP_TAG --depth 1 https://github.com/omniti-labs/pg_
     make install && \
     cd /home/musicbrainz
 
-ENV SOLR_VERSION 7.7.2
+ENV SOLR_VERSION 7.7.3
```
yvanzo, line 109: Note that we still run 7.7.2 in production. There seems to be no breaking [changes](https://lucene.apache.org/solr/7_7_3/changes/Changes.html) though.

mwiencek: I assumed they follow semantic versioning...

yvanzo: 7.7.2 and 7.7.1 have upgrade notes. Since 7.x is in maintenance only, it might be they don't want to increment minor version either.

PR #1496, opened 2020-05-05
MBS-10732: Add collaborator validation to collection form

reosarevok:
### Fix MBS-10732

Up until now a user could type in an editor name in the collaborators field for a collection and try to submit without selecting a result, leading to an ISE.

The error field is special-cased to remove the margin because otherwise a margin of over 160px makes it look absurd.

Also, inputName for the autocomplete was incorrect, so this fixes that.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Collection.pm
+++ b/lib/MusicBrainz/Server/Form/Collection.pm
@@ -78,4 +78,32 @@ sub validate_type_id {
     }
 }
 
+sub validate_collaborators {
+    my $self = shift;
+
+    my @collaborators = $self->field('collaborators')->fields;
+    my $is_valid = 1;
+    for my $collaborator (@collaborators) {
+        my $id_field = $collaborator->field('id');
+        my $name_field = $collaborator->field('name');
+        if (defined $name_field->value && !(defined $id_field->value)) {
+            my $editor = $self->ctx->model('Editor')->get_by_name($name_field->value);
+            if (defined $editor) {
+                $id_field->add_error(
+                    l('To add “{editor}” as a collaborator, please select them from the dropdown.',
+                      {editor => $name_field->value})
+                );
+            } else {
+                $id_field->add_error(
+                    l('Editor “{editor}” does not exist.',
+                      {editor => $name_field->value})
+                );
+            }
+            $is_valid = 0;
+        }
+    }
+
+    return $is_valid;
+}
+
 1;
--- a/root/static/scripts/collection/components/CollectionEditForm.js
+++ b/root/static/scripts/collection/components/CollectionEditForm.js
@@ -100,7 +100,11 @@ const CollectionEditForm = ({collectionTypes, form}: Props) => {
               />
               <div className="form-row-text-list">
                 {collaborators.field.map((collaborator, index) => (
-                  <div className="text-list-row" key={collaborator.html_name}>
+                  <div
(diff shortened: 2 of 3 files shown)


lib/MusicBrainz/Server/Form/Collection.pm
```diff
@@ -78,4 +78,19 @@ sub validate_type_id {
...
 
+sub validate_collaborators {
+    my $self = shift;
+
+    my @collaborators = $self->field('collaborators')->fields;
+    for my $collaborator (@collaborators) {
+        my $id_field = $collaborator->field('id');
+        my $name_field = $collaborator->field('name');
+        if (defined $name_field->value && !(defined $id_field->value)) {
+            return $id_field->add_error(
```
yvanzo, line 89: It should probably add an error for all invalid collaborators, not just `return` the first one.

lib/MusicBrainz/Server/Form/Collection.pm
```diff
@@ -78,4 +78,19 @@ sub validate_type_id {
...
 }
 
+sub validate_collaborators {
+    my $self = shift;
+
+    my @collaborators = $self->field('collaborators')->fields;
+    for my $collaborator (@collaborators) {
+        my $id_field = $collaborator->field('id');
+        my $name_field = $collaborator->field('name');
+        if (defined $name_field->value && !(defined $id_field->value)) {
```
yvanzo, line 88: It would be nice to try to find editor’s id from the given name before rejecting it.

lib/MusicBrainz/Server/Form/Collection.pm
reosarevok: True! Made the change.

lib/MusicBrainz/Server/Form/Collection.pm
reosarevok: I don't like the idea of automatically selecting the editor, but I did change the messages to account for the editor existing and hopefully make them more useful!

lib/MusicBrainz/Server/Form/Collection.pm
```diff
@@ -78,4 +78,32 @@ sub validate_type_id {
...
 
+sub validate_collaborators {
+    my $self = shift;
+
+    my @collaborators = $self->field('collaborators')->fields;
+    my $is_valid = 1;
+    for my $collaborator (@collaborators) {
+        my $id_field = $collaborator->field('id');
+        my $name_field = $collaborator->field('name');
+        my $editor = $self->ctx->model('Editor')->get_by_name($name_field->value);
```
mwiencek, line 89: Would be better to move this line inside the block below, above `if (defined $editor) {`, since it's not used outside that block and we probably don't want to run this if `$name_field->value` is actually `undef`.

lib/MusicBrainz/Server/Form/Collection.pm
```diff
@@ -78,4 +78,32 @@ sub validate_type_id {
...
+                    l('To add “{editor}” as a collaborator, please select them from the dropdown.',
+                      {editor => $name_field->value})
+                );
+            } else {
+                $id_field->add_error(
+                    l('Editor “{editor}” does not exist.',
+                      {editor => $name_field->value})
+                );
+            }
+            my $is_valid = 0;
```
mwiencek, line 102:
```suggestion
            $is_valid = 0;
```

PR #1499, opened 2020-05-06
MBS-10811: Convert historic Edit Relationship edit to React

reosarevok:
### Implement MBS-10811

Changes from the TT:

Includes a hack to allow passing the different link phrases with the link types.
If we managed to add the attributes properly to these edits at some point we could remove the hack.

Fixes the crash mentioned by yvanzo in https://github.com/metabrainz/musicbrainz-server/pull/1474


reosarevok:
I think using interpolateText fixes that - let me know if you find it broke something else!

PR #1500, opened 2020-05-06
MBS-10393: Convert Add Standalone Recording edit to React

anirudhjain75:


root/components/ExpandedArtistCredit.js
```diff
@@ -0,0 +1,58 @@
+/*
```
mwiencek, line 1: We have an `ExpandedArtistCredit` component at root/static/scripts/common/components/ExpandedArtistCredit.js which you should be able to use. (Might've been added since you wrote those.)

root/edit/details/AddStandaloneRecording.js
```diff
@@ -0,0 +1,82 @@
...
+    +name: string,
+    +recording: RecordingT,
+    +video: boolean,
+  },
+};
+
+const AddStandaloneRecording =
+  ({edit}: {edit: AddStandaloneRecordingProps}) => {
+    const display = edit.display_data;
+    console.log(display);
```
mwiencek, line 32:
```suggestion
```

root/edit/details/AddStandaloneRecording.js
```diff
@@ -0,0 +1,82 @@
...
+    +length: number | null,
+    +name: string,
+    +recording: RecordingT,
+    +video: boolean,
+  },
+};
+
+const AddStandaloneRecording =
+  ({edit}: {edit: AddStandaloneRecordingProps}) => {
+    const display = edit.display_data;
```
mwiencek, line 31: I'd suggest something like
```suggestion
const AddStandaloneRecording = (
  {edit}: {+edit: AddStandaloneRecordingProps}
) => {
  const display = edit.display_data;
```
because that way you only have to indent 2 spaces instead of 4

root/edit/details/AddStandaloneRecording.js
```diff
@@ -0,0 +1,82 @@
+/*
+ * @flow
+ * Copyright (C) 2020 Anirudh Jain
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React from 'react';
```
mwiencek, line 10:
```suggestion
import * as React from 'react';
```

root/edit/details/AddStandaloneRecording.js
```diff
@@ -0,0 +1,82 @@
...
+        <table className="details add-standalone-recording">
+          <tbody>
+            <tr>
+              <th>{addColonText(l('Name'))}</th>
+              <td>
+                <EntityLink
+                  allowNew
+                  content={display.name}
+                  entity={display.recording}
+                  showDisambiguation
```
mwiencek, line 52: Looks like we don't show the disambiguation for the name in production, just for "Recording" above. I think that makes more sense, since the disambiguation isn't part of the name.

root/edit/details/AddStandaloneRecording.js
```diff
@@ -0,0 +1,85 @@
...
+  {edit}: {+edit: AddStandaloneRecordingProps},
+) => {
+  const display = edit.display_data;
+  return (
+    <>
+      <table className="details">
+        <tbody>
+          <tr>
+            <th>{addColonText(l('Recording'))}</th>
+            <td><EntityLink allowNew entity={display.recording} /></td>
```
reosarevok, line 40: I know this is what we use now, but maybe we should change to DescriptiveLink? That would show the artist credit here. Admittedly, the artist is shown further down, but that's the original artist credit for the edit, while this would be the current one :) @mwiencek, @yvanzo: what do you think?

root/edit/details/AddStandaloneRecording.js
```diff
@@ -0,0 +1,85 @@
...
+        <tbody>
+          <tr>
+            <th>{addColonText(l('Name'))}</th>
+            <td>
+              <EntityLink
+                allowNew
+                content={display.name}
+                entity={display.recording}
+                showDisambiguation={false}
+              />
```
reosarevok, line 54: I would change this to just print the name, same as how it's done in Add release. I see no reason to link the same recording twice. @mwiencek, @yvanzo: opinions?

anirudhjain75: It was this way according to the previous template... and yes I do thing using links twice is not of much use 🤔. It mostly depends on what experience do you want to give

root/edit/details/AddStandaloneRecording.js
yvanzo: Is there any reason to display current data along with edit data to start with? There is nothing that tells users `Recroding` is not edit data, it is quite confusing. If it is neither needed to display current edits nor needed to display past edits, I would just drop the whole `Recording` row.

root/edit/details/AddStandaloneRecording.js
yvanzo: I would prefer dropping the `Recording:` row if that is possible, see the above https://github.com/metabrainz/musicbrainz-server/pull/1500#discussion_r422478389.

root/edit/details/AddStandaloneRecording.js
anirudhjain75: @mwiencek Whats your take on this ?

reosarevok: @yvanzo: Well, it helps me figure out if it was changed again since this edit, so I personally find it useful :) But I'm not sure if that's just me or a more general situation.

mwiencek: I could see cases in which it might be useful, but I'd move it outside from under the `Changes` header (maybe above it), since it's not a change.

root/edit/details/AddStandaloneRecording.js
mwiencek: +1 to only printing the name.

root/edit/details/AddStandaloneRecording.js
mwiencek: Since that's actually outside of the edit template, maybe just leave it be for now.

reosarevok:
Changed this to add React.MixedElement as the result flow type, and pushed a commit changing the edit to print only recording name under Name. @mwiencek, @yvanzo: I think this is fine for merging, but do check :)

PR #1503, opened 2020-05-08
MBS-9340: Only allow mul and zxx as the only work language 

reosarevok:
### Implement MBS-9340

There's no case where it makes sense to have [Multiple languages] or [No lyrics] and a second, different language as a work language at the same time.

There's already a limitation on the JS layer that makes it so no more languages can be added if one of the two is selected, but that doesn't affect any languages already selected before them.

I chose not to further enforce this on the JS level because the form verification does all that is needed while making sure the selected languages do not get lost/deselected, which would seem annoying if the user selected one of these two by mistake.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Work.pm
+++ b/lib/MusicBrainz/Server/Form/Work.pm
@@ -68,6 +68,35 @@ sub is_empty_attribute {
     return !$field->field('type_id')->value && !$field->field('value')->value;
 }
 
+sub validate_languages {
+    my $self = shift;
+
+    my @languages = $self->field('languages')->fields;
+
+    my $is_valid = 1;
+
+    # If we have only one language, then it can be whatever
+    # If we have two or more, we check it's not mul (248) nor zxx (486)
+    # because combining those with anything else doesn't make sense
+    if (scalar @languages > 1) {
+        for my $language_field (@languages) {
+            if ($language_field->value == 284) {
+                $language_field->push_errors(
+                    l('You cannot select “[Multiple languages]” and specific languages at the same time.')
+                );
+                $is_valid = 0;
+            } elsif ($language_field->value == 486) {
+                $language_field->push_errors(
+                    l('You cannot select “[No lyrics]” and a lyrics language at the same time.')
+                );
+                $is_valid = 0;
+            }
+        }
+    }
+
+    return $is_valid;
+}
+
 after 'validate' => sub {
     my ($self) = @_;
 
--- a/root/static/scripts/work.js
+++ b/root/static/scripts/work.js
@@ -69,7 +69,8 @@ const store = createStore(function (state: WorkForm = form, action) {
 
     case 'EDIT_LANGUAGE':
       state = mutate<WorkForm, _>(state, newState => {
-        newState.field.languages.field[action.index].value = action.languageId;
+        newState.field.languages.field[action.index].value =
+          action.languageId;
       });
       break;
 
@@ -272,7 +273,10 @@ function renderWorkLanguages() {
       addId="add-language"
       addLabel={l('Add Language')}
       getSelectField={_.identity}
-      hideAddButton={_.intersection(form.field.languages.field.map(lang => String(lang.value)), ["486", "284"]).length > 0}
+      hideAddButton={_.intersection(
+        form.field.languages.field.map(lang => String(lang.value)),
+        ['486', '284'],
+      ).length > 0}
       label={l('Lyrics Languages')}
       onAdd={addLanguage}
       onEdit={editLanguage}


lib/MusicBrainz/Server/Form/Work.pm
```diff
@@ -68,6 +68,35 @@ sub is_empty_attribute {
...
+    my $is_valid = 1;
+
+    # If we have only one language, then it can be whatever
+    # If we have two or more, we check it's not mul (248) nor zxx (486)
+    # because combining those with anything else doesn't make sense
+    if (scalar @languages > 1) {
+        for my $language_field (@languages) {
+            if ($language_field->value == 284) {
+                $language_field->add_error(
+                    l('You can’t select “~[Multiple languages~]” and specific languages at the same time.')
```
reosarevok, line 85: Considered translating `[Multiple languages]` (and `[No lyrics]` below) separately as {multiple_languages_name} or whatnot to ensure it uses the same name as the actual language in translation, but I'm actually not sure that is safe because the translation in a sentence might not be the same as the standalone one in some languages.

yvanzo: IMHO, using `{lang}` placeholder would probably be safer because these are specific values not a free wording. Maybe @mihkel-t has a different point of view?

Otherwise, what is `~` doing here?

mihkel-t: I'd say that for translators, it's definitely better to have the full string visible, not a {lang} placeholder. Might be a good idea to have a msgctx saying that "[multiple languages]" should be translated the same way as in the language list.

reosarevok: `~` escapes the brackets - I got an error otherwise, which specifically suggested using `~`, so I did!

mihkel-t:
> There's no case where it makes sense to have [Multiple languages] ... and a second, different language as a work language at the same time.


> It seems obvious for `[No lyrics]` but `[Multiple languages]` could be used to specify there are other languages not known yet. Not sure it is frequent enough to be worth not adding annotation instead.

Eluveitie is the prime example that comes to mind: they have a number of songs partly (and some wholly) in a reconstructed version of Gaulish, which doesn't have an ISO language code, and is hence not usable on MBz (see MBS-9105). I think currently tags are used to store that information, but to me at least it would be confusing, if work language is only English and yet the work has a tag "Gaulish".

reosarevok:
I'd argue that should either be just "multiple languages" or "English + annotation". "English + Multiple" is also confusing, because there's only one language other than English :) In any case, this seems like a very obscure exception that can be explained with annotations - eventually we might just become able to add the languages without fully depending on ISO too.

lib/MusicBrainz/Server/Form/Work.pm
yvanzo: I have not been able to get that error message. You should probably find another way to get around that because this escape character is not currently used for other `[` occurrences. It would be nice to prevent introducing yet another markup syntax if possible.

lib/MusicBrainz/Server/Form/Work.pm
```diff
@@ -68,6 +68,35 @@ sub is_empty_attribute {
...
+sub validate_languages {
+    my $self = shift;
+
+    my @languages = $self->field('languages')->fields;
+
+    my $is_valid = 1;
+
+    # If we have only one language, then it can be whatever
+    # If we have two or more, we check it's not mul (248) nor zxx (486)
+    # because combining those with anything else doesn't make sense
```
yvanzo, line 80: Previous comments have shown that combining other language(s) with `mul` can be used to specify that there are more languages which:
* either have to be determined ,
* or are not supported by MusicBrainz.

It would be helpful to know how many legitimate occurrences of such combination are currently used in the database.

reosarevok: As I commented above, this seems to not be the ideal way to do this anyway, unless you specifically know there's at least two additional languages so that "multiple" is still true. But I can run a query.

mwiencek: That assumes eng + mul means "english, plus other multiple languages" rather than just "multiple languages, including english," which is how I'd interpret it, but perhaps you're right that it's not an ideal way.

yvanzo: I agree it is not ideal but it is still better than checking annotation to make sure “English” means just “English” and not “English plus possibly other languages”. Anyway it matters only if it is used.

reosarevok: We have 30 of those (well, 29 because one was a clearly wrong mess I'm getting rid of). Some seem to wrongly have [multiple] and a few seem to be intended as "something I don't know" (example annotation: "English and an unknown second language."). If the intention is to have a way to indicate "I don't know" then we should have an "[unknown]" option, but I think this is so minor that it is actually much less confusing to just use the annotation.

lib/MusicBrainz/Server/Form/Work.pm
reosarevok: Made this change. If we don't generally want to pass these through FormHandler localization, should we always use push_errors instead of add_error?

mwiencek: I'd say yes, ideally, though I'm not sure `add_error` causes any problems unless you clash with the maketext syntax by accident.

lib/MusicBrainz/Server/Form/Work.pm
```diff
@@ -68,6 +68,35 @@ sub is_empty_attribute {
...
+    if (scalar @languages > 1) {
+        for my $language_field (@languages) {
+            if ($language_field->value == 284) {
+                $language_field->push_errors(
+                    l('You can’t select “[Multiple languages]” and specific languages at the same time.')
+                );
+                $is_valid = 0;
+            } elsif ($language_field->value == 486) {
+                $language_field->push_errors(
+                    l('You can’t select “[No lyrics]” and a lyrics language at the same time.')
```
yvanzo, line 90: Nitpicking: I would probably replace “You can’t” with “You cannot” which is already used for most other strings.

PR #1504, opened 2020-05-08
MBS-10771: Block tagging for unverified users

reosarevok:
### Implement MBS-10771

People are creating email-less accounts to mass-tag their favourite boy band. While that's mostly harmless, it's still stupid, and I can't see any reason why we should allow unverified users to submit tags.

Ratings were already blocked recently by MBS-10671 / https://github.com/metabrainz/musicbrainz-server/pull/1404.

Most "changes" are just extra indentation required from splitting lines because `has_confirmed_email_address` is a pretty long name.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Tag.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Tag.pm
@@ -62,7 +62,7 @@ sub _vote_on_tags {
     my ($self, $c, $method) = @_;
 
     $c->res->headers->header('X-Robots-Tag' => 'noindex');
-    if (!$c->user_exists) {
+    if (!$c->user_exists || !$c->user->has_confirmed_email_address) {
         $c->res->status(401);
         $c->res->body('{}');
         $c->detach;
--- a/root/layout/components/sidebar/SidebarTags.js
+++ b/root/layout/components/sidebar/SidebarTags.js
@@ -54,46 +54,47 @@ const SidebarTags = ({
 }: SidebarTagsProps): React.Element<typeof React.Fragment> | null => (
   $c.action.name === 'tags' ? null : (
     <>
-      {($c.user && aggregatedTags && userTags) ? (
-        <SidebarTagEditor
-          $c={$c}
-          aggregatedTags={aggregatedTags}
-          entity={entity}
-          more={more}
-          userTags={userTags}
-        />
-      ) : (
-        <div id="sidebar-tags">
-          <h2>{l('Genres')}</h2>
-          <div className="genre-list">
+      {($c.user?.has_confirmed_email_address &&
+        aggregatedTags && userTags) ? (
+          <SidebarTagEditor
+            $c={$c}
+            aggregatedTags={aggregatedTags}
+            entity={entity}
+            more={more}
+            userTags={userTags}
+          />
+        ) : (
+          <div id="sidebar-tags">
+            <h2>{l('Genres')}</h2>
+            <div className="genre-list">
+              <p>
+                <TagList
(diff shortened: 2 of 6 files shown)


root/static/scripts/common/components/TagEditor.js
```diff
@@ -512,18 +513,20 @@ export const MainTagEditor = hydrate<TagEditorProps>(
...
+                  <p>
+                    {l(
+                      `Tags with a score of zero or below,
+                      and tags that you’ve downvoted are hidden.`,
+                    )}
+                  </p>
+                ) : (
+                  <p>
+                    {l('Tags with a score of zero or below are hidden.')}
+                    {' '}
```
yvanzo, line 527: This change seems to be completely unrelated, plus this extra space seems to be useless anyway.

t/lib/t/MusicBrainz/Server/Controller/Work/Tags.pm
```diff
@@ -40,4 +40,27 @@ test all => sub {
...
 
+test 'Cannot tag without a confirmed email address' => sub {
+    my $test = shift;
+    my $mech = $test->mech;
+    my $c = $test->c;
+
+    MusicBrainz::Server::Test->prepare_test_database($c);
+
+    $c->model('Editor')->insert({
+        name => 'iwannarate',
```
yvanzo, line 51:
```suggestion
        name => 'iwannatag',
```

t/lib/t/MusicBrainz/Server/Controller/Work/Tags.pm
```diff
@@ -40,4 +40,27 @@ test all => sub {
...
+        password => 'password'
+    });
+
+    $mech->get_ok('/login');
+    $mech->submit_form( with_fields => { username => 'iwannarate', password => 'password' } );
```
yvanzo, line 56:
```suggestion
    $mech->submit_form( with_fields => { username => 'iwannatag', password => 'password' } );
```

root/static/scripts/common/components/TagEditor.js
reosarevok: IIRC it made the line too long otherwise? We were already adding the extra space, I just moved it.

reosarevok: It does seem to be useless though, yes. Just removed it.

root/layout/components/sidebar/SidebarTags.js
```diff
@@ -55,46 +55,47 @@ const SidebarTags = ({
...
-            <p>
-              <TagList
-                entity={entity}
-                isGenreList
-                tags={aggregatedTags}
-              />
-            </p>
-          </div>
+      {($c.user && $c.user.has_confirmed_email_address &&
+        aggregatedTags && userTags) ? (
```
mwiencek, line 59: Optional chaining might be nice here. (I count 4 other places in the PR, too)
```suggestion
      {($c.user?.has_confirmed_email_address &&
        aggregatedTags && userTags) ? (
```

lib/MusicBrainz/Server/Controller/Role/Tag.pm
```diff
@@ -68,6 +68,10 @@ sub _vote_on_tags {
         $c->detach;
     }
 
+    if (!$c->user->has_confirmed_email_address) {
+        $c->detach('/error_401');
+    }
+
```
mwiencek, line 74: That returns an HTML page, right? Since this is a JSON endpoint we should probably add `|| !$c->user->has_confirmed_email_address` to the condition above instead.

PR #1505, opened 2020-05-09
MBS-10397: Convert Edit Event edit to React

anirudhjain75:


yvanzo:

lib/MusicBrainz/Server/Edit/Event/Edit.pm
```diff
@@ -110,6 +112,16 @@ sub build_display_data
         }
     }
 
+    if (exists $data->{ended}) {
+        $data->{ended}{old} = boolean_to_json($data->{ended}{old});
+        $data->{ended}{new} = boolean_to_json($data->{ended}{new});
+    }
```
reosarevok, line 118: I wouldn't bother with this, since we're not displaying it, unless we have to for some reason.

root/edit/details/EditEvent.js
```diff
@@ -0,0 +1,114 @@
...
+          newText={formatDate(endDate.new)}
+          oldText={formatDate(endDate.old)}
+          split="-"
+        />
+      ) : null}
+      {time ? (
+        <FullChangeDiff
+          label={addColonText(l('Time'))}
+          newContent={time.new}
+          oldContent={time.old}
```
reosarevok, line 99: I see in the typing that time can be null. If that's true, then we should be checking it here, no? And if it's not, we should change the typing. Unless I'm missing something.

reosarevok: I see now that the diff accepts react.node, so it *is* allowed to be null - but then it'd make sense for all others to also just pass null when it's null, no?

anirudhjain75:
@reosarevok I made the changes you suggested. Thanks :)

root/edit/details/EditEvent.js
anirudhjain75: As for this.. I've added `?? ''` to check if the value is null and pass an empty string instead. I think that should better.

PR #1506, opened 2020-05-10
MBS-10816: Convert Add Label Edit to React

anirudhjain75:


root/edit/details/AddLabel.js
```diff
@@ -0,0 +1,125 @@
...
+import formatDate from '../../static/scripts/common/utility/formatDate';
+import yesNo from '../../static/scripts/common/utility/yesNo';
+import DescriptiveLink from
+  '../../static/scripts/common/components/DescriptiveLink';
+import formatIsni from '../../utility/formatIsni';
+
+type AddLabelProps = {
+  ...EditT,
+  +display_data: {
+    ...CommentRoleT,
```
mwiencek, line 23: Technically the comment can be null here, but you can either make the suggested change or add `// ''` to `build_display_data`.
```suggestion
    +comment: string | null,
```

reosarevok:
Oh, duh, missed the sort name. Yes, we definitely should still show that when present :)

reosarevok:
Adapted this to types-first, added missing sortname field, fixed comment type issue. @yvanzo, @mwiencek: maybe we can merge this now :)

mwiencek:
This looked good except deleted labels were shown as being added. I copied the change from 20cb5c09166c8b5200ffeb580074987226205ac5 to this file, which fixed that. Also rebased on master.

PR #1508, opened 2020-05-11
MBS-10822: Change tableColumns tables to use named parameters 

reosarevok:
### Implement MBS-10822

Hopefully this makes things more readable and understandable, to begin with.

Also: a couple tiny fixes to related files.

### DIFF
--- a/root/components/list/AreaList.js
+++ b/root/components/list/AreaList.js
@@ -38,13 +38,20 @@ const AreaList = ({
   const columns = React.useMemo(
     () => {
       const checkboxColumn = $c.user && (checkboxes || mergeForm)
-        ? defineCheckboxColumn(checkboxes, mergeForm)
+        ? defineCheckboxColumn({mergeForm: mergeForm, name: checkboxes})
         : null;
-      const nameColumn =
-        defineNameColumn<AreaT>(l('Area'), order, sortable);
-      const typeColumn = defineTypeColumn('area_type', order, sortable);
+      const nameColumn = defineNameColumn<AreaT>({
+        order: order,
+        sortable: sortable,
+        title: l('Area'),
+      });
+      const typeColumn = defineTypeColumn({
+        order: order,
+        sortable: sortable,
+        typeContext: 'area_type',
+      });
       const removeFromMergeColumn = mergeForm
-        ? defineRemoveFromMergeColumn(areas)
+        ? defineRemoveFromMergeColumn({toMerge: areas})
         : null;
 
       return [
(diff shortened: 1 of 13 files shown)


root/utility/tableColumns.js
```diff
@@ -282,14 +293,15 @@ export function defineInstrumentUsageColumn(
 }
 
 export function defineNameColumn<T: CoreEntityT | CollectionT>(
-  title: string,
-  order?: string = '',
-  sortable?: boolean = false,
-  descriptive?: boolean = true,
+  props: {
+    ...OrderableProps,
+    +descriptive: boolean,
```
yvanzo, line 298:
```suggestion
    +descriptive?: boolean,
```

root/utility/tableColumns.js
```diff
@@ -282,14 +293,15 @@ export function defineInstrumentUsageColumn(
...
+    +title: string,
+  },
 ): ColumnOptions<T, string> {
   return {
     Cell: ({row: {original}}) => (
-      descriptive
+      props.descriptive
```
yvanzo, line 304:
```suggestion
    const descriptive = props.hasOwnProperty('descriptive') ? descriptive : true;
    return {
      Cell: ({row: {original}}) => (
        descriptive
```

root/components/list/AreaList.js
```diff
@@ -38,13 +38,21 @@ const AreaList = ({
...
     () => {
       const checkboxColumn = $c.user && (checkboxes || mergeForm)
-        ? defineCheckboxColumn(checkboxes, mergeForm)
+        ? defineCheckboxColumn({mergeForm: mergeForm, name: checkboxes})
         : null;
-      const nameColumn =
-        defineNameColumn<AreaT>(l('Area'), order, sortable);
-      const typeColumn = defineTypeColumn('area_type', order, sortable);
+      const nameColumn = defineNameColumn<AreaT>({
+        descriptive: true,
```
yvanzo, line 44:
```suggestion
```
and all other occurrences.

root/utility/tableColumns.js
reosarevok: Is this better though?

I feel defaulting things to true is confusing, so I thought specifying when we do want descriptive linking seemed more clear.

On the other hand, it's true that we want descriptive linking in most cases, so maybe the clearest option of them all would be change the condition from "descriptive" to "simpleLink" or something like that, and then only pass it when we specifically want to deviate from the (descriptive) norm?

yvanzo: Default value can be either `true` or `false`, in either case that will be far better than requiring to set it for the vast majority of calls. “simpleLink” is simply not as descriptive as its negation “descriptive”, but there might be a clearer wording.

reosarevok: Ok, since it seems it's just me who finds this kinda weird, then let's do it :)

yvanzo:
@brainzbot retest this please

PR #1511, opened 2020-05-12
MBS-10186: Check for missing CD TOC on SetTrackLengths

reosarevok:
### Fix MBS-10186

If a CD TOC is removed before an edit setting track times from it applies, the edit currently fails to close. This fails it with an appropriate failed dependency message.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
+++ b/lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
@@ -5,6 +5,7 @@ use namespace::autoclean;
 use MooseX::Types::Moose qw( ArrayRef Int Str );
 use MooseX::Types::Structured qw( Dict );
 use MusicBrainz::Server::Constants qw( $EDIT_SET_TRACK_LENGTHS );
+use MusicBrainz::Server::Data::Utils qw( localized_note );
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
```diff
@@ -133,6 +133,14 @@ sub accept {
         );
     }
 
+    my $cdtoc_id = $self->data->{cdtoc}{id};
+    if (!$self->c->model('CDTOC')->get_by_id($cdtoc_id)) {
+        MusicBrainz::Server::Edit::Exceptions::FailedDependency->throw(
+            'The CD TOC the track times were being set from no longer exists. '.
+            'It may have been removed since this edit was entered.'
+        );
```
mwiencek, line 141: Since this will presumably be left as a note, can we use `localized_note` here?

(I also wonder what other cases there are besides it being removed that we have to say it "may" have been removed.)

reosarevok: I'm not sure! And because I wasn't sure, I wrote "may". But it might very well be that it is the only possibility.

reosarevok:
Changed this to use localized_note, added a test.

PR #1512, opened 2020-05-12
MBS-10821: Remove orphaned recordings from collections for deletion

reosarevok:
### Fix MBS-10821

An edit medium edit is stuck because one of the recordings that would be orphaned and removed with the edit is in an editor collection.

While silently removing the recording from a collection might not seem ideal, this is already what we do with other entity removals. The alternative would be to fail edits that would cause that, but that would block removal of any recordings that should be removed but are still present on an editor's collection for whatever reason, since the collection owner would need to actively remove them from the collection to allow the medium edit to ever apply.


reosarevok:
Tried to add a test in Edit::Medium::Edit but that was a huge pain in the ass because this runs only *after the transaction ends*. So just testing it in pgtap, which should work fine anyway.

PR #1513, opened 2020-05-12
MBS-10187: Fail edits that would result in a gendered group

reosarevok:
### Fix MBS-10187

An edit that merges a group into an artist of no type set, but with a gender set will currently crash at the DB level. This adds an error and fails the edit - we could also choose to silently drop the type, but this seems better for avoiding dubious / dodgy merges.

NB: this will also cause an error in a situation where the edit would possibly have applied earlier - if both group *and* person artists are being merged into the no-type, gendered artist. That situation
seems even more dodgy though, so I think that's actually a positive change.

FWIW, the opposite edit (a gendered, no-type artist being merged into a group) results on the merge passing and the gender being silently dropped. I'm not too happy with that, though.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Artist/Merge.pm
+++ b/lib/MusicBrainz/Server/Edit/Artist/Merge.pm
@@ -3,7 +3,8 @@ use Moose;
 
 use MooseX::Types::Moose qw( ArrayRef Bool Int Str );
 use MooseX::Types::Structured qw( Dict );
-use MusicBrainz::Server::Constants qw( $EDIT_ARTIST_MERGE );
+use List::MoreUtils qw( any );
+use MusicBrainz::Server::Constants qw( $ARTIST_TYPE_GROUP $EDIT_ARTIST_MERGE );
 use MusicBrainz::Server::Data::Utils qw( boolean_to_json );
 use MusicBrainz::Server::Translation qw( N_l );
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Edit/Artist/Merge.pm
```diff
@@ -51,6 +52,24 @@ sub do_merge
...
+            'SELECT 1 FROM artist_type WHERE id = ? AND parent = ?',
+            $_->type_id, $ARTIST_TYPE_GROUP,
+        );
+    } @old_artists;
+
+    if (defined $new_artist->gender_id && !defined $new_artist->type_id && $old_groups) {
+        MusicBrainz::Server::Edit::Exceptions::GeneralError->throw(
+            'This edit would create a group artist with a gender, '.
+            'but a group cannot have a gender. '.
+            'Please correct the data as needed and re-enter the merge.'
```
yvanzo, line 69: Coding style only:

```suggestion
            'This edit would create a group artist with a gender, ' .
            'but a group cannot have a gender. ' .
            'Please correct the data as needed and re-enter the merge.'
```

mwiencek:
I still have another (older) branch for this that avoids failing the edits & leaves ModBot notes when a conflict arises. @reosarevok said he was fine with me submitting that, but sorry for the delay in doing so.

PR #1514, opened 2020-05-12
MBS-9931: Fail gracefully when trying to remove example rel

reosarevok:
### Implement MBS-9931

If someone tries to remove an example relationship, the edit gets stuck. We don't want people to actually remove these themselves, so this fails the edit and asks them to contact us.

Eventually, we might want to block entering these removals completely, but that's much harder and given the situation is pretty rare, this seems like a good first step.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
@@ -2,8 +2,10 @@ package MusicBrainz::Server::Edit::Relationship::Delete;
 use Moose;
 use Try::Tiny;
 
-use MusicBrainz::Server::Constants qw( $EDIT_RELATIONSHIP_DELETE );
+use List::MoreUtils qw( any );
+use MusicBrainz::Server::Constants qw( $CONTACT_URL $EDIT_RELATIONSHIP_DELETE );
 use MusicBrainz::Server::Data::Utils qw(
+    localized_note
     partial_date_to_hash
     type_to_model
 );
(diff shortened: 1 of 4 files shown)


reosarevok:
You should be able to accept/reject edits from a development server, including remove rel ones, no? :)

mwiencek:
With `DB_STAGING_TESTING_FEATURES` enabled you can approve any edit. You can also update the `expire_time` of the edit directly in the DB and run CheckVotes, but enabling the testing features is much easier. :)

yvanzo:
Thank you @mwiencek! It was set to 0 a few months ago. As I rarely used the “accept” button, I just forgot about it and used the auto-editor “approve” button instead, but it is not available for destructive edits. It is not the first time I mention it but always got a dismissive answer until now, anyway :champagne:

lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
```diff
@@ -248,6 +250,22 @@ sub accept {
...
+    $self->c->model('LinkType')->load_documentation($relationship->link->type);
+    my @examples = $relationship->link->type->all_examples;
+
+    if (any { $_->{relationship}->id == $relationship->id } @examples) {
+        my $error = localized_note(
+            N_l('This edit would remove a relationship that is set '.
+            'as an example of its relationship type in the documentation. '.
+            'If you still think this should be removed, please '. 
+            'contact us (see details at {contact_url}).'),
+            {contact_us => $CONTACT_URL},
```
yvanzo, line 264: The placeholder `contact_url` does not match the variable `contact_us`.
While fixing it, I would suggest to directly link from “contact us”.

```suggestion
            'If you still think this should be removed, please '.
            '{contact_url|contact us}.'),
            {contact_url => $CONTACT_URL},
```

lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
```diff
@@ -248,6 +250,22 @@ sub accept {
...
+    $self->c->model('Link')->load($relationship);
+    $self->c->model('LinkType')->load($relationship->link);
+    $self->c->model('LinkType')->load_documentation($relationship->link->type);
+    my @examples = $relationship->link->type->all_examples;
+
+    if (any { $_->{relationship}->id == $relationship->id } @examples) {
+        my $error = localized_note(
+            N_l('This edit would remove a relationship that is set '.
+            'as an example of its relationship type in the documentation. '.
+            'If you still think this should be removed, please '.
```
yvanzo, line 262: Just coding style: please vertically align strings and put a space before the concatenation operator `.`

lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
mwiencek: Agreed. However, this breaks the link because there's a bug where we run `format_editnote` on the resulting URL, so it ends up as `https://musicbrainz.org/edit/%3Ca%20href=%22https://metabrainz.org/contact%22%20rel=%22nofollow%22%3Ehttps://metabrainz.org/contact%3C/a%3E`. I pushed https://github.com/metabrainz/musicbrainz-server/pull/1514/commits/a46113cd9eb2e9ae481badd59c7bbf04723ce4a9 to resolve that.

reosarevok:
Added an actual test now too.

PR #1515, opened 2020-05-13
MBS-9516: Support custom validation errors in URLCleanup

reosarevok:
### Implement MBS-9516

This patch allows giving custom error messages for URLs that fail validation. Hopefully this will help avoid confusion when, for example, a user tries and fails to paste an Allmusic link on a release.

The patch already adds a fair amount of errors for cases that I have often seen can be confusing, but most validation failures will still return the default message. We should eventually decide just how often we want to give custom messages, but I would expect for a lot of the most obvious ones the default error message is enough.

This also changes the Wikipedia exception introduced in MBS-9515 to work as all the other newly-added error messages.

It also implements MBS-10831 and MBS-10412 for niconico improvements.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -306,10 +306,6 @@ function reencodeMediawikiLocalPart(url) {
   return url;
 }
 
-function disallow() {
-  return false;
-}
-
 function findAmazonTld(url) {
   let tld = '';
   let m;
@@ -360,14 +356,14 @@ const CLEANUPS = {
         const prefix = m[1];
         switch (id) {
           case LINK_TYPES.otherdatabases.artist:
-            return prefix === 'artist';
+            return {result: prefix === 'artist'};
           case LINK_TYPES.otherdatabases.label:
-            return prefix === 'label';
+            return {result: prefix === 'label'};
           case LINK_TYPES.otherdatabases.release_group:
-            return prefix === 'record';
+            return {result: prefix === 'record'};
         }
       }
-      return false;
+      return {result: false};
     },
   },
   '45worlds': {
@@ -382,18 +378,18 @@ const CLEANUPS = {
         const prefix = m[2];
         switch (id) {
           case LINK_TYPES.otherdatabases.artist:
-            return /^(artist|composer|conductor|orchestra|soloist)$/.test(prefix);
+            return {result: /^(artist|composer|conductor|orchestra|soloist)$/.test(prefix)};
           case LINK_TYPES.otherdatabases.event:
-            return prefix === 'listing';
+            return {result: prefix === 'listing'};
           case LINK_TYPES.otherdatabases.label:
-            return prefix === 'label';
+            return {result: prefix === 'label'};
           case LINK_TYPES.otherdatabases.place:
-            return prefix === 'venue';
+            return {result: prefix === 'venue'};
           case LINK_TYPES.otherdatabases.release_group:
-            return /^(album|cd|media|music|record)$/.test(prefix);
+            return {result: /^(album|cd|media|music|record)$/.test(prefix)};
         }
       }
-      return false;
+      return {result: false};
     },
   },
   'allmusic': {
@@ -408,18 +404,21 @@ const CLEANUPS = {
         const prefix = m[1];
         switch (id) {
           case LINK_TYPES.allmusic.artist:
-            return prefix === 'artist/mn';
+            return {result: prefix === 'artist/mn'};
           case LINK_TYPES.allmusic.recording:
-            return prefix === 'performance/mq';
+            return {result: prefix === 'performance/mq'};
           case LINK_TYPES.allmusic.release:
-            return prefix === 'album/release/mr';
+            return {result: prefix === 'album/release/mr'};
           case LINK_TYPES.allmusic.release_group:
-            return prefix === 'album/mw';
+            return {result: prefix === 'album/mw'};
           case LINK_TYPES.allmusic.work:
-            return prefix === 'composition/mc' || prefix === 'song/mt';
+            return {
+              result: prefix === 'composition/mc' ||
+                prefix === 'song/mt',
+            };
         }
       }
-      return false;
+      return {result: false};
     },
   },
   'amazon': {
@@ -455,7 +454,7 @@ const CLEANUPS = {
       return null;
     },
     validate: function (url) {
-      return /^https:\/\/www\.amazon\.(com|ca|co\.uk|fr|ae|at|de|it|sg|co\.jp|jp|cn|es|in|nl|com\.br|com\.mx|com\.au|com\.tr)\//.test(url);
+      return {result: /^https:\/\/www\.amazon\.(com|ca|co\.uk|fr|ae|at|de|it|sg|co\.jp|jp|cn|es|in|nl|com\.br|com\.mx|com\.au|com\.tr)\//.test(url)};
     },
   },
   'amazonmusic': {
@@ -490,12 +489,12 @@ const CLEANUPS = {
         const prefix = m[1];
         switch (id) {
           case LINK_TYPES.streamingpaid.artist:
-            return prefix === 'artists';
+            return {result: prefix === 'artists'};
           case LINK_TYPES.streamingpaid.release:
-            return prefix === 'albums';
+            return {result: prefix === 'albums'};
         }
       }
-      return false;
+      return {result: false};
     },
   },
   'animationsong': {
@@ -505,7 +504,10 @@ const CLEANUPS = {
       return url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?animationsong\.com\/(archives\/\d+\.html).*$/, 'http://animationsong.com/$1');
     },
     validate: function (url, id) {
-      return id === LINK_TYPES.lyrics.work && /^http:\/\/animationsong\.com\/archives\/\d+\.html$/.test(url);
+      return {
+        result: id === LINK_TYPES.lyrics.work &&
+          /^http:\/\/animationsong\.com\/archives\/\d+\.html$/.test(url),
+      };
     },
   },
   'animenewsnetwork': {
@@ -528,14 +530,14 @@ const CLEANUPS = {
         const prefix = m[1];
         switch (id) {
           case LINK_TYPES.otherdatabases.artist:
-            return prefix === 'person';
+            return {result: prefix === 'person'};
           case LINK_TYPES.otherdatabases.release:
-            return prefix === 'source';
+            return {result: prefix === 'source'};
           case LINK_TYPES.otherdatabases.recording:
-            return prefix === 'song';
+            return {result: prefix === 'song'};
         }
       }
-      return false;
+      return {result: false};
     },
   },
   'archive': {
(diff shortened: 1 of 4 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -3942,7 +3942,7 @@ _.each(testData, function (subtest, i) {
       const validationResults = _.reduce(LINK_TYPES[relationshipType],
         function (results, relUuid, entityType) {
           const rule = validationRules[relUuid];
-          const isValid = rule ? rule(cleanUrl) || false : true;
+          const isValid = rule ? rule(cleanUrl).result || false : true;
```
yvanzo, line 3945: Maybe type could be checked so as not to have to change all of the returned results without message.

reosarevok: That would introduce inconsistency though. I'd much rather everything followed a consistent style.

yvanzo: Which “inconsistency”?

reosarevok: Some rule results being an object and some being a Boolean, and us having to change them for adding messages rather than just adding error to the object. If some will be objects then I wouldn't want others not to be. I will split this in several commits later though :)

reosarevok:
Should hopefully be much easier to review now! :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1000,9 +1000,25 @@ const CLEANUPS = {
       if (m) {
         const prefix = m[1];
         if (_.includes(LINK_TYPES.videochannel, id)) {
+          if (prefix === 'video/') {
+            return {
+              error: l(
+                `Please link to a channel, not a specific video.
+                 Videos should be linked to the appropriate
+                 recordings instead.`,
```
yvanzo, line 1043: All of these error messages for video relationships should probably be defined in a set of common error messages once.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -409,6 +409,16 @@ const CLEANUPS = {
           case LINK_TYPES.allmusic.recording:
             return {result: prefix === 'performance/mq'};
           case LINK_TYPES.allmusic.release:
+            if (prefix === 'album/mw') {
+              return {
+                error: l(
+                  `Allmusic album links should be added to release groups.
+                   To find the appropriate release link for this release,
+                   check the Releases tab for the Allmusic album.`,
```
yvanzo, line 417:
```suggestion
                   please check the Releases tab for the Allmusic album.`,
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -757,7 +757,13 @@ const CLEANUPS = {
...
           switch (id) {
             case LINK_TYPES.mailorder.artist:
-              return {result: product === undefined};
+              return {
+                error: l(
+                  `Please link to the main page for the artist,
+                   not a specific product.`,
+                ),
+                result: product === undefined,
+              };
```
yvanzo, line 766: `error` should not be set when `result` is `true`.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1126,6 +1126,15 @@ const CLEANUPS = {
           case LINK_TYPES.discogs.release_group:
             return {result: prefix === 'master'};
           case LINK_TYPES.discogs.release:
+            if (prefix === 'master') {
+              return {
+                error: l(
+                  `Discogs master links group several releases,
+                   and should be added to release groups instead.`,
```
yvanzo, line 1133:
```suggestion
                   thus should be added to release groups instead.`,
```

root/static/scripts/edit/URLCleanup.js
reosarevok: I thought it wouldn't matter since we should only be loading the error when result is false :)

root/static/scripts/edit/URLCleanup.js
reosarevok: They're not all exactly the same though, since some cases are slightly different than others.

root/static/scripts/edit/URLCleanup.js
reosarevok: "thus" sounded too fancy to me, but changed to "so this should be added to the release group instead" :)

root/static/scripts/edit/URLCleanup.js
yvanzo: Most of error messages in this commit are used at least two times, if not three times. If these messages are expected to be reused for more video websites, constants seem worth to have.

root/static/scripts/edit/URLCleanup.js
reosarevok: So you'd rather have this set so that it checks whether the check is false, and then return either one way or the other? Well then. It honestly seems like a waste of lines to me, but if it makes you happy :D

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1831,8 +1852,20 @@ const CLEANUPS = {
         switch (id) {
           case LINK_TYPES.streamingfree.recording:
           case LINK_TYPES.streamingfree.release:
+            if (prefix === 'user') {
+              return {
+                error: linkToVideoMsg,
```
reosarevok, line 1933: linkToVideoMsg says "artist, label, etc.", but for some reason, niconico is restricted to `LINK_TYPES.videochannel.artist`. Do we know if this is intentional? Are labels, places, etc. not allowed to have a channel there?

yvanzo: I did not find anything about it in MBS-9944. Possibly niconico is just for artists, not sure.

By the way, you might want to check MBS-10412 again.

reosarevok: Made some changes to niconico (MBS-10831 / MBS-10412) and put them at the end of the PR, do check :)

yvanzo: Nitpicking the message of commit https://github.com/metabrainz/musicbrainz-server/pull/1515/commits/75ccb1670a00890441fd5c6dd2272826960d9d7a which seems erroneous:
* It should probably be about user links, not “channel” links.
* “videochannel entities” should probably be replaced with “videochannel rels”.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -325,6 +322,18 @@ function findAmazonTld(url) {
   }
   return tld;
 }
+
+const linkToChannelMsg = l(
```
mwiencek, line 326: This script is likely executed after the language is set, but better to not assume that's the case (or that we won't ever run this on the server, e.g. on seeded URLs once the edit forms are server-rendered by React).
```suggestion
const linkToChannelMsg = N_l(
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -325,6 +322,18 @@ function findAmazonTld(url) {
...
+  `Please link to a channel, not a specific video.
+   Videos should be linked to the appropriate
+   recordings or releases instead.`,
+);
+
+const linkToVideoMsg = l(
```
mwiencek, line 332:
```suggestion
const linkToVideoMsg = N_l(
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -983,11 +1050,20 @@ const CLEANUPS = {
       if (m) {
         const prefix = m[1];
         if (_.includes(LINK_TYPES.videochannel, id)) {
-          return prefix === undefined;
+          if (prefix === 'video/') {
+            return {
+              error: linkToChannelMsg,
```
mwiencek, line 1055:
```suggestion
              error: linkToChannelMsg(),
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1784,18 +1922,29 @@ const CLEANUPS = {
...
-          case LINK_TYPES.streamingfree.release:
-            return prefix === 'watch';
-          case LINK_TYPES.videochannel.artist:
-            return prefix === 'user';
+        const subdomain = m[1];
+        const prefix = m[2] || m[3];
+        if (_.includes(LINK_TYPES.videochannel, id)) {
+          if (prefix === 'watch') {
+            return {
+              error: linkToChannelMsg,
```
mwiencek, line 1932:
```suggestion
              error: linkToChannelMsg(),
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2348,11 +2515,20 @@ const CLEANUPS = {
       if (m) {
         const prefix = m[1];
         if (_.includes(LINK_TYPES.videochannel, id)) {
-          return prefix === undefined;
+          if (prefix === 'videos/') {
+            return {
+              error: linkToChannelMsg,
```
mwiencek, line 2520:
```suggestion
              error: linkToChannelMsg(),
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -983,11 +1050,20 @@ const CLEANUPS = {
...
+              result: false,
+            };
+          }
+          return {result: prefix === undefined};
         }
-        return prefix === 'video/';
+        return {
+          error: linkToVideoMsg,
```
mwiencek, line 1062:
```suggestion
          error: linkToVideoMsg(),
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1784,18 +1922,29 @@ const CLEANUPS = {
...
+              error: linkToChannelMsg,
+              result: false,
+            };
+          }
+          return {result: subdomain === 'ch' || prefix === 'user'};
+        }
+
+        if (subdomain === 'ch' || prefix === 'user') {
+          return {
+            error: linkToVideoMsg,
```
mwiencek, line 1941:
```suggestion
            error: linkToVideoMsg(),
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2348,11 +2515,20 @@ const CLEANUPS = {
...
+            return {
+              error: linkToChannelMsg,
+              result: false,
+            };
+          }
+          return {result: prefix === undefined};
         }
-        return prefix === 'videos/';
+        return {
+          error: linkToVideoMsg,
```
mwiencek, line 2527:
```suggestion
          error: linkToVideoMsg(),
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2603,13 +2816,28 @@ const CLEANUPS = {
...
+              `Please link to a channel, not a specific video.
+               Videos should be linked to the appropriate
+               recordings or releases instead`,
+            ),
+            result: /^https:\/\/www\.youtube\.com\/(?!watch\?v=[a-zA-Z0-9_-])/.test(url),
+          };
         case LINK_TYPES.streamingfree.recording:
-          return /^https:\/\/www\.youtube\.com\/watch\?v=[a-zA-Z0-9_-]+$/.test(url);
+          return {
+            error: linkToVideoMsg,
```
mwiencek, line 2829:
```suggestion
            error: linkToVideoMsg(),
```

PR #1516, opened 2020-05-14
Updated Facebook, Google Play, and Spotify icons.

KamranMackey:
I have updated the icons used for Spotify, Google Play, and Facebook to the latest revisions of their logos. Should look a fair bit better now. I also updated the Spotify icon to be 32px instead of 16px since it's slightly higher res and doesn't look as janky as the older icon that was previously used.

**EDIT:** It appears that there was an anomaly with the Google Play icon where it jumped to 5 KB from 1.23 KB. Not entirely sure on how to optimize this.


KamranMackey:
@yvanzo I have replaced the ICO-based PNG with a proper PNG icon.

yvanzo:
@KamranMackey, your patch has been deployed on beta servers, see “External links” in https://beta.musicbrainz.org/artist/c2f6c57f-27ba-4c0d-b2fd-3348961e79a1 for example.

PR #1517, opened 2020-05-14
MBS-10825: Fix Muziekweb cleanup, add work support

reosarevok:
### Fix MBS-10825

We've been removing the CLASSICAL/POPULAR segment from Muziekweb links, but it turns out they do have two separate data pages. Removing the segment makes them all default to POPULAR, so classical artists (and works) will not show the intended data. Additionally, artists that worked in both classical and popular contexts have different data on both URLs, so we will need to store both /CLASSICAL and /POPULAR for those anyway.

I mailed Muziekweb to make sure there's no way to get both sets of data with one link and they confirmed it's not possible.


reosarevok:
Hmm. That seems to be the same as just CLASSICAL/ without the COMPOSER/, right? Should the cleanup just drop it?

yvanzo:
> Hmm. That seems to be the same as just CLASSICAL/ without the COMPOSER/, right? Should the cleanup just drop it?

Not sure, according to Muziekweb’s answer, it is not the same presentation if the person is both a composer and a performer. We should make sure that no data is missing in such case.

reosarevok:
Ok, made that change :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1876,14 +1876,16 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?www\\.muziekweb\\.(eu|nl)/', 'i')],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:www\.)?muziekweb\.(?:eu|nl)\/(?:[a-z]{2}\/)?Link\/([A-Z]{1,3}\d+).*$/, 'https://www.muziekweb.eu/Link/$1/');
+      return url.replace(/^(?:https?:\/\/)?(?:www\.)?muziekweb\.(?:eu|nl)\/(?:[a-z]{2}\/)?Link\/([A-Z]{1,3}\d+)(?:\/)?((CLASSICAL(?:\/COMPOSER)?|POPULAR)\/)?.*$/, 'https://www.muziekweb.nl/Link/$1/$2');
```
yvanzo, line 1879: Old domain `.com` should probably be matched (above) and cleaned up (here) too.

The RegExp can be simplified to use less groups, and it seems slightly clearer about handling slashes:
```suggestion
      return url.replace(/^(?:https?:\/\/)?(?:www\.)?muziekweb\.(?:com|eu|nl)\/(?:[a-z]{2}\/)?Link\/([A-Z]{1,3}\d+(?:\/(?:CLASSICAL(?:\/COMPOSER)?|POPULAR))?).*$/, 'https://www.muziekweb.nl/Link/$1/');
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1876,14 +1876,16 @@ const CLEANUPS = {
...
       switch (id) {
         case LINK_TYPES.otherdatabases.artist:
-          return {result: /^https:\/\/www\.muziekweb\.eu\/Link\/M\d{11}\/$/.test(url)};
+          return {result: /^https:\/\/www\.muziekweb\.nl\/Link\/M\d{11}\/(CLASSICAL(?:\/COMPOSER)?|POPULAR)\/$/.test(url)};
         case LINK_TYPES.otherdatabases.release:
-          return {result: /^https:\/\/www\.muziekweb\.eu\/Link\/[A-Z]{2,3}\d{4,6}\/$/.test(url)};
+          return {result: /^https:\/\/www\.muziekweb\.nl\/Link\/[A-Z]{2,3}\d{4,6}\/$/.test(url)};
+        case LINK_TYPES.otherdatabases.work:
+          return {result: /^https:\/\/www\.muziekweb\.nl\/Link\/U\d{11}\/(CLASSICAL|POPULAR)\/$/.test(url)};
       }
```
yvanzo, line 1889: Common parts can be checked first, so as to check only the part specific to relationship types then:

```suggestion
      const subpath = /^https:\/\/www\.muziekweb\.nl\/Link\/(.*)\/$/.exec(url);
      if (subpath) {
      switch (id) {
        case LINK_TYPES.otherdatabases.artist:
          return {result: /^M\d{11}\/(CLASSICAL(?:\/COMPOSER)?|POPULAR)$/.test(subpath)};
        case LINK_TYPES.otherdatabases.release:
          return {result: /^[A-Z]{2,3}\d{4,6}$/.test(subpath)};
        case LINK_TYPES.otherdatabases.work:
          return {result: /^U\d{11}\/(CLASSICAL|POPULAR)$/.test(subpath)};
      }
      }
```

reosarevok:
Allowed label, amended the commit message.

PR #1518, opened 2020-05-15
MBS-10829: Fix SOLR syntax for recording with no length

reosarevok:
### Fix MBS-10829

This wasn't actually finding recordings without length at all, except when the track also had no length (which made the query different).

This makes it so that recordings without length can also be found. Because we're unlikely to want recordings without length and *also* with a different artist as the track, and that can cause a lot of false positives, this restricts the no-length results to ones with matching artists.

While this certainly improves the inline search, I feel that this causes more (less precise) suggested recordings to appear. If you find while testing that that's the case, please do feel free to suggest improvements to the query.

I tested with "A Soul to Bear" by Fall of Efrafa, FWIW, but ideally you'd test it with other recordings that have no times to make sure it works always :)

### DIFF
--- a/root/static/scripts/release-editor/recordingAssociation.js
+++ b/root/static/scripts/release-editor/recordingAssociation.js
@@ -85,7 +85,7 @@ function recordingQuery(track, name) {
 
   var titleAndArtists = utils.constructLuceneFieldConjunction(params);
   var justTitle = utils.constructLuceneField(params.recording, 'recording');
-  var query = '(' + titleAndArtists + ')^2 OR (' + justTitle + ')';
+  var query;
 
   var duration = parseInt(track.length(), 10);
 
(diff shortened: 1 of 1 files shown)


mwiencek:
> While this certainly improves the inline search, I feel that this causes more (less precise) suggested recordings to appear. If you find while testing that that's the case, please do feel free to suggest improvements to the query.

I noticed this too. I tried (and failed) to modify the search query to help with that, but realized the real issue is how the release editor sorts the results. Can you have a look at 9c09ba60d83525fb6fac2b8fc5873fb17a0b820a, see if it makes sense, and try it out? Some of the logic might need tweaking.

reosarevok:
Rebased, fixed indent, pushed. My own test case works fine, but it's hard to just test this further without having some ideas of stuff that didn't work well before.

yvanzo:
I pushed it a rebased version to `test` branch, currently building, will deploy this evening.

mwiencek (approved):

PR #1519, opened 2020-05-19
MBS-10834: Account flag to disable writing edit notes 

reosarevok:
### Implement MBS-10834 (and fixes MBS-10838)

The editing_disabled flag blocks a user from editing, but they can still leave edit notes on edits. This is fine in cases where the editor just makes bad edits (whether by lack of care or knowledge), but in some cases the problem is one of attitude when interacting with the community. In that situation, just blocking editing won't stop the user from leaving insulting / demeaning edit notes.

This is the goal of the new flag.

In some cases, we might want to set this *without* actually disabling editing at all, especially if a user keeps getting into a comment war and just needs to stop commenting for a while.


root/layout.tt
```diff
@@ -10,12 +10,22 @@
...
 
-        [%- IF c.user.is_editing_disabled -%]
+        [%- IF c.user.is_editing_disabled || c.user.are_notes_disabled -%]
             <div class="banner editing-disabled">
                 <p>
-                    [% l('You’re currently not allowed to edit or vote because an admin has revoked your privileges.
-                          If you haven’t already been contacted about why, please {uri|send us a message}.',
-                        { uri => { href => 'https://metabrainz.org/contact', target => '_blank' } }) %]
+                    [%- IF c.user.is_editing_disabled && c.user.are_notes_disabled -%]
+                        [% l('You’re currently not allowed to edit, vote or leave edit notes on other editors’ edits because an admin has revoked your privileges.
```
Freso, line 17: AFAICT the code currently doesn’t allow you to leave comments on your own edits either (maybe except for an initial one) (the `$editor->id == $self->editor_id ||` isn’t used for the check for the disabled flag)?

I agree with this logic, but that makes this wording seem kind of weird, as it suggests that you would still be able to leave notes on your own edits.

Maybe just
```suggestion
                        [% l('You’re currently not allowed to edit, vote or leave edit notes because an admin has revoked your privileges.
```

root/layout.tt
```diff
@@ -10,12 +10,22 @@
...
+                            If you haven’t already been contacted about why, please {uri|send us a message}.',
+                            { uri => { href => 'https://metabrainz.org/contact', target => '_blank' } }) %]
+                    [%- ELSIF c.user.is_editing_disabled -%]
+                        [% l('You’re currently not allowed to edit or vote because an admin has revoked your privileges.
+                            If you haven’t already been contacted about why, please {uri|send us a message}.',
+                            { uri => { href => 'https://metabrainz.org/contact', target => '_blank' } }) %]
+                    [%- ELSE -%]
+                        [% l('You’re currently not allowed leave edit notes on other editors’ edits because an admin has revoked your privileges.
```
Freso, line 25: Same thing.

root/layout.tt
reosarevok: Duh. I noticed this and fixed the React messages, but I guess I forgot the TT ones. Fixing!

Freso:
Oh, and you probably want to get rid of `postgresql-musicbrainz-collate` and `postgresql-musicbrainz-unaccent` too… :)

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -104,6 +104,10 @@ sub is_editing_enabled {
     (shift->privileges & $EDITING_DISABLED_FLAG) == 0;
 }
 
+sub are_notes_disabled {
```
yvanzo, line 107: This would mean that existing notes are disabled too but they are not.

```suggestion
sub is_adding_note_disabled {
```

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -104,6 +104,10 @@ sub is_editing_enabled {
...
+    (shift->privileges & $NOTES_DISABLED_FLAG) > 0;
```
yvanzo, line 108: For the same reason as above:
```suggestion
    (shift->privileges & $ADDING_NOTE_DISABLED_FLAG) > 0;
```

lib/MusicBrainz/Server/Form/User/AdjustFlags.pm
```diff
@@ -54,6 +54,10 @@ has_field 'editing_disabled' => (
     type => 'Boolean',
 );
 
+has_field 'notes_disabled' => (
```
yvanzo, line 57: Same thing:
```suggestion
has_field 'adding_note_disabled' => (
```

root/layout.tt
```diff
@@ -10,12 +10,22 @@
...
+                            { uri => { href => 'https://metabrainz.org/contact', target => '_blank' } }) %]
+                    [%- ELSIF c.user.is_editing_disabled -%]
+                        [% l('You’re currently not allowed to edit or vote because an admin has revoked your privileges.
+                            If you haven’t already been contacted about why, please {uri|send us a message}.',
+                            { uri => { href => 'https://metabrainz.org/contact', target => '_blank' } }) %]
+                    [%- ELSE -%]
+                        [% l('You’re currently not allowed to leave edit notes because an admin has revoked your privileges.
+                            If you haven’t already been contacted about why, please {uri|send us a message}.',
+                            { uri => { href => 'https://metabrainz.org/contact', target => '_blank' } }) %]
+                    [%- END -%]
```
yvanzo, line 28: Since edits should be entered with edit notes, is there a case where editor would be allowed to enter edits without being allowed to enter edit notes?

root/layout/index.js
```diff
@@ -113,14 +113,33 @@ const Layout = ({$c, ...props}: Props) => (
...
+                  {uri: {href: 'https://metabrainz.org/contact', target: '_blank'}},
+                )
+              )
+            ) : (
+              exp.l(
+                `You’re currently not allowed to leave edit notes because
+                 an admin has revoked your privileges. If you haven’t already
+                 been contacted about why, please {uri|send us a message}.`,
+                {uri: {href: 'https://metabrainz.org/contact', target: '_blank'}},
+              )
```
yvanzo, line 142: Same question as above.

root/layout.tt
reosarevok: You can actually leave notes when entering the edit, but that's the only case where they can.

yvanzo: Ok about editing, but you can still vote but not leave edit note, is it intended too?

reosarevok: I'd say yes. This should be used on its own just for one use case IMO, getting people to calm down and not get into an argument. In most cases, it will be used in combination with editing + voting blocks, it's just that there *might* be some cases where @Freso only needs to block edit notes.

lib/MusicBrainz/Server/Entity/Editor.pm
mwiencek: That sounds fine too though I'd keep `notes` plural

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -314,8 +314,11 @@ sub update_privileges {
                 + ($values->{banner_editor}    // 0) * $BANNER_EDITOR_FLAG
                 + ($values->{mbid_submitter}   // 0) * $MBID_SUBMITTER_FLAG
                 + ($values->{account_admin}    // 0) * $ACCOUNT_ADMIN_FLAG
-                + ($values->{editing_disabled} // 0) * $EDITING_DISABLED_FLAG;
+                + ($values->{editing_disabled} // 0) * $EDITING_DISABLED_FLAG
+                + ($values->{notes_disabled}   // 0) * $NOTES_DISABLED_FLAG;
 
+    use Data::Dumper;
+    print Dumper($privs);
```
mwiencek, line 321:
```suggestion
```

PR #1522, opened 2020-05-20
MBS-10839: Add merge button to recording list in artist overview

reosarevok:
### Fix MBS-10839

For some reason, we didn't have a merge option if an artist only has standalone recordings and those are shown on the artist overview.

### DIFF
--- a/root/artist/ArtistIndex.js
+++ b/root/artist/ArtistIndex.js
@@ -297,13 +297,20 @@ const ArtistIndex = ({
       ) : null}
 
       {hasRecordings ? (
-        <PaginatedResults pager={pager}>
-          <RecordingList
-            checkboxes="add-to-merge"
-            recordings={recordings}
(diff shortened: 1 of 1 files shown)


root/artist/ArtistIndex.js
```diff
@@ -297,13 +297,20 @@ const ArtistIndex = ({
...
+          <PaginatedResults pager={pager}>
+            <RecordingList
+              checkboxes="add-to-merge"
+              recordings={recordings}
+              showRatings
+            />
+          </PaginatedResults>
+          {$c.user ? (
+            <div className="row">
+              <FormSubmit label={l('Merge recordings')} />
```
yvanzo, line 310: It should probably use the same wording as in the page for artist’s recordings:

```suggestion
              <FormSubmit label={l('Add selected recordings for merging')} />
```

PR #1523, opened 2020-05-20
MBS-10842: Remove report link from deleted editors

reosarevok:
### Implement MBS-10842

There seems to be no reason to allow reporting deleted editors, since they already deleted their account. I talked to Freso and he agreed. If there's a particular problem they still want us to look at they can contact us at support.

This changes the /report page to say that - I could have blocked it if the user has been deleted, but it feels that if somebody makes the effort of trying to get to /report without the link, we might as well explain the issue.

### DIFF
--- a/root/user/ReportUser.js
+++ b/root/user/ReportUser.js
@@ -9,6 +9,7 @@
 
 import * as React from 'react';
 
+import {CONTACT_URL} from '../constants';
 import UserAccountLayout from '../components/UserAccountLayout';
 import FormCsrfToken from '../components/FormCsrfToken';
 import FormRow from '../components/FormRow';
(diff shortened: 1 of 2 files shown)


root/user/ReportUser.js
```diff
@@ -81,76 +82,92 @@ const ReportUser = ({
...
-      </FormRow>
-    </form>
-
+    {user.deleted ? (
+      <p>
+        {exp.l(
+          `This editor account has already been deleted,
+           so there's probably no need to report it.
+           If you feel there's a problem that still needs action,
+           please {link|contact us}.`,
```
mwiencek, line 91: not really important but maybe s/editor/user/ since the page is called report user
```suggestion
          `This user account has already been deleted,
           so there’s probably no need to report it.
           If you feel there’s a problem that still needs action,
           please {link|contact us}.`,
```

PR #1528, opened 2020-05-21
Prevent two statement timeouts in reports

mwiencek:
This should unbreak daily.sh for now and allow subscription emails to get out without us having to manually kill RunReports.pl.

### DIFF
--- a/lib/MusicBrainz/Server/Report/RecordingsSameNameDifferentArtistsSameName.pm
+++ b/lib/MusicBrainz/Server/Report/RecordingsSameNameDifferentArtistsSameName.pm
@@ -4,6 +4,10 @@ use Moose;
 with 'MusicBrainz::Server::Report::RecordingReport',
      'MusicBrainz::Server::Report::FilterForEditor::RecordingID';
 
+# MBS-10843: This report has been disabled since the upgrade to PG 12,
+# because its query can no longer execute in under 5 minutes in
+# production.
+
 sub query {
     "
     SELECT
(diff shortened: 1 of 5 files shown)


mwiencek:
Yes, MBS-10843 was opened to track that.

lib/MusicBrainz/Server/Report/RecordingsSameNameDifferentArtistsSameName.pm
```diff
@@ -4,6 +4,9 @@ use Moose;
 with 'MusicBrainz::Server::Report::RecordingReport',
      'MusicBrainz::Server::Report::FilterForEditor::RecordingID';
 
+# FIXME: This report has been disabled since the upgrade to PG 12, because
+# its query can no longer execute in under 5 minutes in production.
```
yvanzo, line 8:
```suggestion
# MBS-10843: This report has been disabled since the upgrade to PG 12,
# because its query can no longer execute in under 5 minutes in production.
```

root/report/ReportsIndex.js
```diff
@@ -449,12 +449,12 @@ const ReportsIndex = ({$c}: {$c: CatalystContextT}) => (
...
           </a>
         </li>
-        <li>
+        {/*<li>
           <a href="/report/RecordingsSameNameDifferentArtistsSameName">
             {l(`Recordings with the same name by different artists
                 with the same name`)}
           </a>
-        </li>
+        </li>*/}
```
yvanzo, line 457: This will likely drop all translations for this message. Could it be moved to an unused `const` instead?

mwiencek: I could just set display: none on the link too, I guess

PR #1529, opened 2020-05-22
MBS-10845: User lookup tools for account admins

mwiencek:
/admin/email-search:

Allows account admins to search users by email using a POSIX regular expression. Since I didn't bother paging the results, it's limited to returning 100 users (in case someone were to accidentally enter a pattern matching half the database) ordered by signup date.

/admin/ip-lookup/$hash:

Allows account admins to see a list of users accessing MusicBrainz from a specific IP hash (an MD5 of their IP address). This is linked from users' profile pages, which now display a list of hashed IPs "belonging" to a user to account admins. (The point of the hash isn't to be cryptographically secure, since admins can find this data in the server logs anyway; it's just to obscure the possible geographic location of the user, since it doesn't seem relevant in most cases.)

Redis keys mapping a user ID to a list of IP hashes and vice versa (an IP hash to a list of user IDs) have an expiration of 6 months, but this expiration is refreshed whenever a relevant user or IP accesses the site.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Admin.pm
+++ b/lib/MusicBrainz/Server/Controller/Admin.pm
@@ -1,5 +1,6 @@
 package MusicBrainz::Server::Controller::Admin;
 use Moose;
+use Try::Tiny;
 
 BEGIN { extends 'MusicBrainz::Server::Controller' };
 
@@ -157,6 +158,53 @@ sub edit_banner : Path('/admin/banner/edit') Args(0) RequireAuth(banner_editor) 
             component_props => {form => $form},
         );
     }
+}
+
+sub email_search : Path('/admin/email-search') Args(0) RequireAuth(account_admin) HiddenOnSlaves {
+    my ($self, $c) = @_;
+
+    my $form = $c->form(form => 'Admin::EmailSearch');
+    my @results;
+
+    if ($c->form_submitted_and_valid($form)) {
+        try {
+            @results = $c->model('Editor')->search_by_email(
+                $form->field('email')->value // '',
+            );
+        } catch {
+            my $error = $_;
+            if ("$error" =~ m/invalid regular expression/) {
+                $form->field('email')->add_error(l('Invalid regular expression.'));
+                $c->response->status(400);
+            } else {
+                die $error;
+            }
+        };
+    }
+
+    $c->stash(
+        current_view => 'Node',
+        component_path => 'admin/EmailSearch',
+        component_props => {
+            form => $form,
+            @results ? (results => \@results) : (),
+        },
+    );
+}
+
+sub ip_lookup : Path('/admin/ip-lookup') Args(1) RequireAuth(account_admin) HiddenOnSlaves {
+    my ($self, $c, $ip_hash) = @_;
+
+    my @users = $c->model('Editor')->find_by_ip($ip_hash // '');
+
+    $c->stash(
+        current_view => 'Node',
+        component_path => 'admin/IpLookup',
+        component_props => {
+            ipHash => $ip_hash,
+            users => \@users,
+        },
+    );
 }
 
 1;
--- a/root/admin/EmailSearch.js
+++ b/root/admin/EmailSearch.js
@@ -1,0 +1,70 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import FormRowText from '../components/FormRowText';
+import FormSubmit from '../components/FormSubmit';
+import {withCatalystContext} from '../context';
+import Layout from '../layout';
+
+import UserList from './components/UserList';
+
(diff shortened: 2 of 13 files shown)


t/lib/t/MusicBrainz/Server/Controller/UnconfirmedEmailAddresses.pm
```diff
@@ -84,6 +84,8 @@ test 'Paths that allow browsing without a confirmed email address' => sub {
   "Controller::Admin::delete_user",
   "Controller::Admin::edit_banner",
   "Controller::Admin::edit_user",
+  "Controller::Admin::email_search",
+  "Controller::Admin::ip_lookup",
```
reosarevok, line 88: I'm not sure why these *can* be seen with an unconfirmed email address, but I guess that's true of all admin bits anyway, so it's not inconsistent.

mwiencek: Yeah. I also thought this was silly, but it's technically true that somebody can have an admin flag set without a confirmed email address, so...

mwiencek:
> Seems fine as it is, although it might be interesting to allow admins to do an email search from a user's page, like with the IPs? Saves the effort of going check the email in edit user, then pasting it on the search page.

Added a link to "find all users of this email." Still shows (hidden) for admins otherwise until they click the link.

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -159,6 +159,43 @@ sub edit_banner : Path('/admin/banner/edit') Args(0) RequireAuth(banner_editor)
...
 
+sub email_search : Path('/admin/email-search') Args(0) RequireAuth(account_admin) {
+    my ($self, $c) = @_;
+
+    my $form = $c->form(form => 'Admin::EmailSearch');
+    my @results;
+
+    if ($c->form_submitted_and_valid($form)) {
+        @results = $c->model('Editor')->search_by_email(
+            $form->field('email')->value // '',
```
yvanzo, line 170: Even though it is accessible to admin only, it should probably be checked for potential SQL injection.

root/admin/EmailSearch.js
```diff
@@ -0,0 +1,70 @@
...
+const EmailSearch = ({
+  $c,
+  form,
+  results,
+}: Props) => (
+  <Layout fullWidth title={l('Search users by email')}>
+    <div id="content">
+      <h1>{l('Search users by email')}</h1>
+
+      <form action="/admin/email-search" method="post">
```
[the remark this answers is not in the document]
mwiencek: We can't have CSRF tokens here because `/admin/email-search` is invoked via query parameters from the user profile page. (I don't think they're useful here anyway, because this form doesn't modify the database in any way.)

lib/MusicBrainz/Server/Controller/Admin.pm
mwiencek: Hmm, can you give an example? It's passed in via a normal DBI placeholder so should be fully escaped by the driver.

mwiencek:
Added the `HiddenOnSlaves` attributes.

lib/MusicBrainz/Server/Controller/Admin.pm
yvanzo: Did not find Little Bobby Tables: https://metacpan.org/pod/DBI#Placeholders-and-Bind-Values

mwiencek:
Sounds like PG is returning `invalid regular expression` there. I pushed a small change to detect that and add a field error. But re: "No details about this error are available," I noticed this for many errors in production too. I think we have a more general issue around `finalize_error` not receiving error details somewhere.

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -159,6 +160,52 @@ sub edit_banner : Path('/admin/banner/edit') Args(0) RequireAuth(banner_editor)
...
+
+    if ($c->form_submitted_and_valid($form)) {
+        try {
+            @results = $c->model('Editor')->search_by_email(
+                $form->field('email')->value // '',
+            );
+        } catch {
+            my $error = $_;
+            if ("$error" =~ m/invalid regular expression/) {
+                $form->field('email')->add_error(l('Invalid regular expression.'));
```
mwiencek, line 177: This string already existed in root/static/scripts/release-editor/trackParser.js

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -159,6 +160,52 @@ sub edit_banner : Path('/admin/banner/edit') Args(0) RequireAuth(banner_editor)
...
+            } else {
```
yvanzo, line 178: Should probably return 400 here:
```suggestion
            if ("$error" =~ m/invalid regular expression/) {
                $form->field('email')->add_error(l('Invalid regular expression.'));
                $c->response->status(400);
            } else {
```

mwiencek:
There's a problem here in that because prod and beta have separate redis stores, the userips/ipusers keys will be separate between those domains too. (This has caused other problems, like MBS-9010.) I think we should finally merge the redis stores before deploying this. Otherwise the feature could be easily bypassed by accessing a sockpuppet only on the beta server.

PR #1530, opened 2020-05-22
MBS-10849: Check for allowNew on AddReleaseGroup

reosarevok:
### Fix MBS-10849

On the release editor preview, new RGs were being shown as removed.
WS::js::preview is already passing allowNew here for previews, we just weren't using it.

### DIFF
--- a/root/edit/details/AddReleaseGroup.js
+++ b/root/edit/details/AddReleaseGroup.js
@@ -14,7 +14,7 @@ import DescriptiveLink from
 import ExpandedArtistCredit from
   '../../static/scripts/common/components/ExpandedArtistCredit';
 
-type AddReleaseGroupProps = {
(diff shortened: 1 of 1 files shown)


root/edit/details/AddReleaseGroup.js
```diff
@@ -15,18 +15,21 @@ import ExpandedArtistCredit from
   '../../static/scripts/common/components/ExpandedArtistCredit';
 
 type AddReleaseGroupProps = {
```
mwiencek, line 17: I'd prefer renaming this to
```suggestion
type AddReleaseGroupEditT = {
```
and adding a separate type for the React props like
```JS
type Props = {
  +allowNew?: boolean,
  +edit: AddReleaseGroupEditT,
};
```
because we'll likely be using these edit type types soon

PR #1531, opened 2020-05-26
Sleep when deploying

yvanzo:
Because deploying a new beta version is more work, we need more restful sleeps! Joking aside, this is because there are only 2 servers in beta, and starting a container currently takes more than 30 seconds. This is due to `compile_resources.sh` being run by the container rather than at build time.

### DIFF
--- a/script/update_containers.sh
+++ b/script/update_containers.sh
@@ -83,6 +83,8 @@ else
   pause=30
 fi
 
+last_host=${HOSTS/* }
+
 for host in $HOSTS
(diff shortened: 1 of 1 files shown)


yvanzo:
For prod, there are 7 hosts, so having up to 3 hosts restarting their containers at the same time has little impact, not something worth slowing the whole process down by 6 minutes. In the future, deployment time should be much shorter since the plan is to not recompile all resources during container’s startup.

PR #1535, opened 2020-05-27
MBS-10863: Fix Edit barcodes edit display for no barcode -> barcode

reosarevok:
### Fix MBS-10863

While in the edit data the field that is used here is "barcode", the display data maps it to "new_barcode", so no barcode was shown when the edit just added a new barcode to releases that had none.

By the way, any reason we shouldn't just use a diff anyway and just pass an empty string if no old barcode exists?


mwiencek:
From https://github.com/metabrainz/musicbrainz-server/commit/ba11b612dbf23056113723c6d986d6db48688091 my understanding is that not all such edits store the old barcode, so displaying an empty string wouldn't be correct in those cases.

PR #1536, opened 2020-05-27
Fix new localizable messages

yvanzo:
# Problem

Many new error messages have been added with #1515 and #1529. A word is likely missing according to https://community.metabrainz.org/t/incomplete-string-is-not-translatable/478813 and I found some other phrasing issues while translating.

# Solution

# Checklist for authors

* [x] Check newly added messages for further issues
* [x] <del>Add context strings</del> (MBS-10864)

# Action

1. Get latest translations before re-uploading `server` resource to Transifex

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1453,8 +1453,8 @@ const CLEANUPS = {
         return {
           error: l(
             `Explore links are not officially connected to a location.
-             If you want to link a to a location, try and find
-             the place's Instagram profile instead, if there's one.`,
+             If you want to link a place to a location, try and find
+             the place’s Instagram profile instead, if there is one.`,
           ),
           result: false,
         };
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1066,9 +1066,12 @@ const CLEANUPS = {
           }
           return {result: prefix === undefined};
         }
+        if (prefix === 'video/') {
+          return {result: true};
+        }
         return {
           error: linkToVideoMsg(),
-          result: prefix === 'video/',
+          result: false,
         };
       }
       return {result: false};
@@ -2560,9 +2563,12 @@ const CLEANUPS = {
           }
           return {result: prefix === undefined};
         }
+        if (prefix === 'videos/') {
+          return {result: true};
+        }
         return {
           error: linkToVideoMsg(),
-          result: prefix === 'videos/',
+          result: true,
         };
       }
       return {result: false};
@@ -2853,21 +2859,30 @@ const CLEANUPS = {
         case LINK_TYPES.youtube.label:
         case LINK_TYPES.youtube.place:
         case LINK_TYPES.youtube.series:
+          if (/^https:\/\/www\.youtube\.com\/(?!watch\?v=[a-zA-Z0-9_-])/.test(url)) {
+            return {result: true};
+          }
           return {
             error: linkToChannelMsg(),
-            result: /^https:\/\/www\.youtube\.com\/(?!watch\?v=[a-zA-Z0-9_-])/.test(url),
+            result: false,
           };
         case LINK_TYPES.streamingfree.recording:
+          if (/^https:\/\/www\.youtube\.com\/watch\?v=[a-zA-Z0-9_-]+$/.test(url)) {
+            return {result: true};
+          }
           return {
             error: linkToVideoMsg(),
-            result: /^https:\/\/www\.youtube\.com\/watch\?v=[a-zA-Z0-9_-]+$/.test(url),
+            result: false,
           };
         case LINK_TYPES.streamingfree.release:
+          if (/^https:\/\/www\.youtube\.com\/(watch\?v=[a-zA-Z0-9_-]+|playlist\?list=[a-zA-Z0-9_-]+)$/.test(url)) {
+            return {result: true};
+          }
           return {
             error: l(
               'Only video and playlist links are allowed on releases.',
             ),
-            result: /^https:\/\/www\.youtube\.com\/(watch\?v=[a-zA-Z0-9_-]+|playlist\?list=[a-zA-Z0-9_-]+)$/.test(url),
+            result: false,
           };
       }
       return {result: false};
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2785,12 +2785,12 @@ const CLEANUPS = {
       if (/^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\/.*#/.test(url)) {
         return {
           error: exp.l(
-            `Links to specific sections of Wikipedia articles are not 
+            `Links to specific sections of Wikipedia articles are not
              allowed. Please remove “{fragment}” if still appropriate.
              See the {url|guidelines}.`,
             {
               fragment: (
-                <span className="url-quote" key="fragment">
+                <span className="url-quote">
                   {url.replace(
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2751,7 +2751,7 @@ const CLEANUPS = {
         return {
           error: exp.l(
             `Links to specific sections of Wikipedia articles are not 
-             allowed. Please remove “{fragment}” if still appropriate.
+             valid. Please remove “{fragment}” if still appropriate.
```
reosarevok, line 2754: I dunno. These are valid links, we just don't allow them.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2830,7 +2830,7 @@ const CLEANUPS = {
         case LINK_TYPES.streamingfree.release:
           return {
             error: l(
-              'Only video and playlist links are allowed on releases.',
+              'Only video and playlist links are valid on releases.',
```
reosarevok, line 2833: This has more of a legitimate reason to be called valid, although I still think "allowed" leaves a clearer impression that we just don't want that link.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -424,9 +424,9 @@ const CLEANUPS = {
             if (prefix === 'album/mw') {
               return {
                 error: l(
-                  `Allmusic album links should be added to release groups.
-                   To find the appropriate release link for this release,
-                   please check the Releases tab for the Allmusic album.`,
+                  `Allmusic “album” links should be added to release groups.
+                   To find the appropriate Allmusic link for this release,
+                   please check the “Releases” tab for the Allmusic “album”.`,
```
reosarevok, line 429: I was thinking whether to say "/album" links instead at first, and if you're going to quote the first one I'd probably prefer the slash to be added there. I think the second album was fine as it was, but otherwise I'd prefer `check the “Releases” tab of the Allmusic page.` or something like that. The quotation otherwise makes it sound weird to me, like "they call them albums but they're wrong".

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1137,7 +1137,7 @@ const CLEANUPS = {
             if (prefix === 'master') {
               return {
                 error: l(
-                  `Discogs master links group several releases,
+                  `Discogs “master” links group several releases,
```
reosarevok, line 1140: I'd prefer `“/master”` to make it clearer.

yvanzo: It was quoting the content of the Discogs page. To refer to an URL part, I would probably use `<code>` instead:
* either `Discogs <code>/master</code> links`
* or `Discogs links with <code>/master</code> in path`

or any better phrased variant… :)

root/static/scripts/edit/URLCleanup.js
yvanzo: Your point is valid. :wink: I felt something was amiss, but I was probably just missing the link to the guidelines at the very end of the message.

root/static/scripts/edit/URLCleanup.js
yvanzo: Just like above, my concern is probably more about the missing why.

root/static/scripts/edit/URLCleanup.js
yvanzo: * Similarly as below, I propose `<code>/album</code>` for the first one,
* For the second one, maybe we can directly use the link? 
  That would be `from {album_url}.` with `album_url: (<a href={url}>url</a>),`.

reosarevok: Maybe `from {album_url|your link}`? Rather than reprinting the URL? But I like the idea!

root/static/scripts/edit/URLCleanup.js
reosarevok: I like the first :)

root/static/scripts/edit/URLCleanup.js
reosarevok: Hmm. We have never specified the why before, but I guess this could be something like               'Only video and playlist links are valid on releases. Channel / user pages should be added to the artist, label, etc. as appropriate.' ? Of course, maybe they've pasted something that isn't a channel or user page either.

root/static/scripts/edit/URLCleanup.js
yvanzo: Fixed at https://github.com/metabrainz/musicbrainz-server/pull/1536/commits/2683b3e6da835c600ffb2dabda28a13eb0853977#diff-87814c26df342be7e4c2a85531bbd47cL1140-R1153

root/static/scripts/edit/URLCleanup.js
yvanzo: * Fixed at https://github.com/metabrainz/musicbrainz-server/pull/1536/commits/2683b3e6da835c600ffb2dabda28a13eb0853977#diff-87814c26df342be7e4c2a85531bbd47cL427-R435
* Fixed with https://github.com/metabrainz/musicbrainz-server/pull/1536/commits/b8bd4d62456eb5a2e0ba6df74641e66c5ad13c97

root/static/scripts/edit/URLCleanup.js
yvanzo: Since I don’t have a better proposition, I just dropped that change.

root/static/scripts/edit/URLCleanup.js
yvanzo: Dropped that change too.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1438,10 +1458,16 @@ const CLEANUPS = {
...
         return {
-          error: l(
-            `Explore links are not officially connected to a location.
-             If you want to link a to a location, try and find
-             the place's Instagram profile instead, if there's one.`,
+          error: exp.l(
+            `Instagram “{explore_url_pattern}” links are not officially
+             connected to a location.
+             If you want to link a place to a location, try and find
+             the place’s Instagram profile instead, if there’s one.`,
```
mwiencek, line 1465: While you're changing the string I'd spell out "is" here because the stress is on that word, not "one"
```suggestion
             the place’s Instagram profile instead, if there is one.`,
```

yvanzo:
Sure it works, after adding proper interpolation in https://github.com/metabrainz/musicbrainz-server/pull/1536/commits/2af3aa21c1fe5c12e1dc46409fa45725465847f0! Just tested it again.
Note: This error message occurs for clean album URLs only.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1137,8 +1151,14 @@ const CLEANUPS = {
             if (prefix === 'master') {
               return {
                 error: l(
-                  `Discogs master links group several releases,
+                  `Discogs “{master_url_pattern}” links group several
+                  releases,
```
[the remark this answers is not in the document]
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1536/commits/9cd63120fd9327f3f7c96d968de2219307ab21a1

root/static/scripts/edit/URLCleanup.js
yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1536/commits/9531a73781dbb17b64e7bb0f28846070fdccb0a1

yvanzo:
Since `master` must be merged into `beta` for https://github.com/metabrainz/musicbrainz-server/pull/788 and other small PRs, I rebased this PR on `master`.

PR #1538, opened 2020-05-28
MBS-10499: Alert for entity in danger of removal once add edit passes 

reosarevok:
### Implement MBS-10499

Sometimes users (especially new ones) add an entity as a votable edit (by checking the relevant checkbox) and are then not notified that the entity is in risk of removal, since their add edit means there's a pending edit keeping the entity around. Unless they have added more data, though, the entity will be removed soon after the one pending edit passes.

This checks if there's only one pending edit and that is the only thing keeping the entity from being eligible for cleanup. If that's the case, it then notifies the user with a banner.


lib/MusicBrainz/Server/Controller/Role/Cleanup.pm
```diff
@@ -2,14 +2,30 @@ package MusicBrainz::Server::Controller::Role::Cleanup;
...
+    my $model = $self->config->{model};
+    my $creation_edit = $c->model('Edit')->find_creation_edit(
+        $model_to_edit_type{$model},
+        $entity->id);
+    # If the creation edit is the only pending edit
+    # and the entity is otherwise empty,
+    # it will be removed as soon as the creation edit passes
+    my $in_cleanup_danger = $entity->edits_pending == 1 &&
+        $creation_edit && $creation_edit->is_open &&
+        $c->model($model)->is_empty($entity->id, 1);
```
reosarevok, line 24: I would like to make the `1` in `is_empty($entity->id, 1)` documented somehow, but not sure what would be the best way to do this. Just define a `my $ignore_edits = 1;` before and then pass that here?

lib/MusicBrainz/Server/Data/Event.pm
```diff
@@ -150,20 +150,23 @@ sub load_meta
...
 sub is_empty {
-    my ($self, $event_id) = @_;
-
+    my ($self, $event_id, $ignore_edits) = @_;
+
+    my $edits_pending_condition = 'AND edits_pending = 0' unless $ignore_edits;
+    my $open_edits_condition = 
+        'EXISTS (
+          SELECT TRUE
+          FROM edit_event JOIN edit ON edit_event.edit = edit.id
```
reosarevok, line 159: I'm a bit confused about why some of these `JOIN edit` at all while others don't. Is there something I'm missing or should we standardise this while I'm at it?

lib/MusicBrainz/Server/Data/Series.pm
```diff
@@ -171,7 +171,7 @@ sub update {
 }
 
 sub is_empty {
-    my ($self, $series_id) = @_;
+    my ($self, $series_id, $ignore_edits) = @_;
```
reosarevok, line 174: Since I'm passing $ignore_edits I guess I should have it here even if it's unused. I'm a bit surprised that this doesn't actually look at pending / open edits at all.

lib/MusicBrainz/Server/Controller/Role/Cleanup.pm
mwiencek: Typically you'd just pass it as a named parameter: add `%args` as the last argument to `is_empty`, pass it as `, ignore_edits => 1`, and access it like `$args{ignore_edits}` inside `is_empty`.

lib/MusicBrainz/Server/Controller/Role/Cleanup.pm
```diff
@@ -2,14 +2,30 @@ package MusicBrainz::Server::Controller::Role::Cleanup;
...
+    my $creation_edit = $c->model('Edit')->find_creation_edit(
+        $model_to_edit_type{$model},
+        $entity->id);
+    # If the creation edit is the only pending edit
+    # and the entity is otherwise empty,
+    # it will be removed as soon as the creation edit passes
+    my $in_cleanup_danger = $entity->edits_pending == 1 &&
+        $creation_edit && $creation_edit->is_open &&
+        $c->model($model)->is_empty($entity->id, ignore_edits => 1);
+    my $eligible_for_cleanup = $c->model($model)->is_empty($entity->id);
```
mwiencek, line 25: Hmm. Do we need separate banners for these two cases? It seems like we could just remove all the edit-checking code from `is_empty` and check here if `edits_pending == 0 || (edits_pending == 1 && creation_edit->is_open)`, then just show a generic message about the entity being in danger of removal within the next few days.

(But we should also change the `RemoveEmpty` script so that that's actually true; it appears to remove things that are empty after only 1 day.)

reosarevok: Yes, I think it's always been 1 day, "the next few days" was always in case it failed to run or something :smile: 

We'd still need to check if it's empty, right? It might be other edits adding more data have been made as autoedits. But that could check only for data and not edits, yes.

lib/MusicBrainz/Server/Data/Event.pm
yvanzo: This seems to be required here to test `status` below.

lib/MusicBrainz/Server/Data/Series.pm
yvanzo: This comment seems to be outdated?

lib/MusicBrainz/Server/Data/Event.pm
reosarevok: Oh. Some tables such as `edit_label` have a materialized `status` column, but some like `edit_event` do not. That explains this at least, but wtf, why?

yvanzo (approved):
Successfully tested with/without votable checkbox/related entities for a new Place.

lib/MusicBrainz/Server/Controller/Role/Cleanup.pm
mwiencek: Right, we'd still check if it's empty in any case. One banner for both cases + changing the RemoveEmpty interval to 3 days or something makes sense to me. Do you see any issue with combining them?

reosarevok: The only issue with moving the RemoveEmpty interval to 3 days is that it adds more risk of someone actually using the stuff before it gets deleted (say, if an editor empties an incorrect label in hopes of it getting removed). Not sure how common that is, though.

reosarevok: Another issue is for stuff like vandalism you'd probably want to get rid ASAP once empty, but I guess for that case we could add a manual way of doing it :)

reosarevok: So if we did this, what would remain in `is_empty`? Something like this only?

```
    my $used_in_relationship = used_in_relationship($self->c, artist => 'artist_row.id');
    return $self->sql->select_single_value(<<~"SQL", $artist_id);
        SELECT TRUE
        FROM artist artist_row
        WHERE id = ?
        AND NOT (
          EXISTS (
            SELECT TRUE FROM artist_credit_name
            WHERE artist = artist_row.id
            LIMIT 1
          ) OR
          $used_in_relationship
        )
        SQL
```

I guess the `RemoveEmpty` script does its own separate check for emptiness so that wouldn't necessarily be a problem to remove from here.

Also, I'm assuming we'd have a separate ticket for the 1 day -> 3 days RemoveEmpty change? :)

reosarevok: I tried to implement this more or less as talked, see if it looks good now :)

lib/MusicBrainz/Server/Controller/Role/Cleanup.pm
```diff
@@ -1,17 +1,35 @@
...
+    my $creation_edit = $c->model('Edit')->find_creation_edit(
+        $model_to_edit_type{$model},
+        $entity->id);
+
+    my $has_no_edits = $entity->edits_pending == 0;
+    my $has_creation_edit_only = $entity->edits_pending == 1 &&
+        $creation_edit && $creation_edit->is_open;
+    my $is_empty = $c->model($model)->is_empty($entity->id);
+    my $eligible_for_cleanup = $is_empty &&
+        ($has_no_edits || $has_creation_edit_only);
```
mwiencek, line 30: You can avoid calling `is_empty` or `find_creation_edit` entirely depending on the edits pending count. Maybe something like this?
```suggestion
    my $edits_pending = $entity->edits_pending;
    if ($edits_pending == 1) {
        my $creation_edit = $c->model('Edit')->find_creation_edit(
            $model_to_edit_type{$model},
            $entity->id,
        );
        if ($creation_edit && $creation_edit->is_open) {
            $edits_pending = 0;
        }
    }
    my $eligible_for_cleanup =
        $edits_pending == 0 &&
        $c->model($model)->is_empty($entity->id);
```

mwiencek (changes requested):
Tested this and it seems to work well. Though if we can avoid some unneeded queries in `Role::Cleanup`, let's do that.

reosarevok:

PR #1539, opened 2020-05-29
Enable types-first mode in Flow

mwiencek:
# Problem

Flow is making https://flow.org/en/docs/lang/types-first/ the default architecture and [removing the "classic" mode by January](https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb), so we have to prepare our codebase for that.

# Solution

Adds type signatures to all of our exports to make them "well-formed." Then enables `types_first` in .flowconfig.

This requires changing a lot of lines, to say the least.


root/account/EditProfile.js
```diff
@@ -22,13 +21,17 @@ type Props = {
   +$c: CatalystContextT,
 };
 
-const EditProfile = ({$c, ...props}: Props) => {
+const EditProfile = ({
+  $c,
+  ...props
+}: Props): React.Element<typeof UserAccountLayout> | null => {
```
mwiencek, line 27: `React.MixedElement` also works as a return type, as a generic way of saying it returns a React element without specifying which one. Though there is at least one case I saw where specifying it was useful.

root/components/EntityHeader.js
```diff
@@ -12,16 +12,17 @@ import * as React from 'react';
 import EntityLink from '../static/scripts/common/components/EntityLink';
 
 import EntityTabs from './EntityTabs';
+import typeof EntityTabLink from './EntityTabLink';
 import SubHeader from './SubHeader';
 
 type Props = {
-  +editTab?: React.Node,
+  +editTab?: React.Element<EntityTabLink>,
```
mwiencek, line 19: Here's an example of where it might be useful to specify the type of element we want.

root/static/scripts/common/components/AreaContainmentLink.js
```diff
@@ -14,7 +15,11 @@ import EntityLink from './EntityLink';
 
 const makeLink = (x, i) => <EntityLink entity={x} key={i} />;
 
-const AreaContainmentLink = ({area}) => (
+type Props = {
+  +area: AreaT,
+};
+
+const AreaContainmentLink = ({area}: Props) => (
```
reosarevok: Wouldn't just `{area}: {area: AreaT}` be simpler here?

mwiencek: It made the line too long once `: Expand2ReactOutput | null` was added in the last commit. I could've put the return type on a separate line, I guess, though I defaulted to making the Props type separate 'cause I think it makes adding new props a bit easer, and allows for reusing the type if we need to. But I don't mind either too much otherwise.

mwiencek:
Documented the `CatalystContext` React context in 4fc7ed2

PR #1542, opened 2020-06-01
MBS-10867: Fix "entity not in the entities array"

mwiencek:
There are easy steps to reproduce in MBS-10867. This is just a workaround until we replace this stuff with React. Also fixes MBS-10877, though that's relatively minor.


root/static/scripts/relationship-editor/common/fields.js
```diff
@@ -551,7 +551,22 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
...
-            throw 'entity not in the entities array';
+            /*
+             * We can sometimes hit this case while `this.entities` is being
+             * updated but `entity` is stale, due to the order of
+             * subscriptions being run. If this is the case, just return a
+             * function that returns an empty string. Only throw an error if
+             * we attempt to *write* a value in this situation.
+             */
+            return function () {
+                if (arguments.length > 0) {
```
mwiencek, line 562: (With knockout observables, you read values by calling the function and write values by calling the function with an argument.)

mwiencek:
BTW, this exception was added as part of 6e3446d27e9, though the `throw` itself didn't break anything; without that it just throws a different exception about `createField(...)` (the return value) not being a function.

mwiencek:
For the one linked in the ticket, it works if you change the URL in the input to use https first so that it's valid. But you can test it on any valid URL page too.

reosarevok:
This seems to work if not changing the relationship type, but still break if changing it to "download purchase page for the score of": 

```
Uncaught TypeError: Cannot read property 'target' of undefined
    at n.value (viewModel.js:142)
    at text (eval at <anonymous> (index.js:2754), <anonymous>:3:70)
    at update (index.js:3888)
    at index.js:3133
    at p (index.js:1638)
    at t.subscription.d [as callback] (index.js:1581)
    at Function.notifySubscribers (index.js:1063)
    at Function.r.valueHasMutated (index.js:1241)
    at t.r [as entities] (index.js:1226)
    at t.value (fields.js:156)
    at n.value (dialog.js:692)
    at n.value (dialog.js:245)
    at dialog.js:338
    at lodash.js:2756
    at r (helpers.js:71)
```

Which is weird to me. Probably an improvement in any case, but leaving this note in case you can reproduce and fix that one.

mwiencek:
Yeah, that looks like a different issue the ticket didn't mention, but I can reproduce it too so I'll look into it. :)

mwiencek:
c80fd36 seems to fix that one for me

PR #1543, opened 2020-06-02
MBS-10833: Avoid crashing if entity0_id is missing

reosarevok:
### Fix MBS-10833

Some old edits (at least edit 12587968) are missing this.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Historic/Relationship.pm
+++ b/lib/MusicBrainz/Server/Edit/Historic/Relationship.pm
@@ -208,6 +208,9 @@ sub _expand_relationships {
     } @$mappings;
 }
 
+# To access entities missing an ID via linkedEntities later on
+my $fake_entity_id = 1000000000;
+
 sub _display_relationships {
     my ($self, $data, $loaded) = @_;
     return [
@@ -216,7 +219,7 @@ sub _display_relationships {
             my $entity1_type = $_->{entity1_type};
             my $model0 = type_to_model( $_->{entity0_type} );
             my $model1 = type_to_model( $_->{entity1_type} );
-            my $entity0_id = $_->{entity0_id};
+            my $entity0_id = $_->{entity0_id} // $fake_entity_id++;
             my $entity1_id = $_->{entity1_id};
             my $entity0 = $loaded->{ $model0 }{ $entity0_id } ||
                 $self->c->model($model0)->_entity_class->new(
--- a/lib/MusicBrainz/Server/Entity/Relationship.pm
+++ b/lib/MusicBrainz/Server/Entity/Relationship.pm
@@ -239,8 +239,10 @@ sub _interpolate {
             return $alt1;
         }
     };
-    $phrase =~ s/{(.*?)(?::(.*?))?}/$replace_attrs->(lc $1, $2)/eg;
-    trim_in_place($phrase);
+    if (defined $phrase) {
+        $phrase =~ s/{(.*?)(?::(.*?))?}/$replace_attrs->(lc $1, $2)/eg;
+        trim_in_place($phrase);
+    }
 
     my @extra_attrs = map { @$_ } values %extra_attrs;
     return [ $phrase, comma_only_list(@extra_attrs) ];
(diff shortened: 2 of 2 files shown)


lib/MusicBrainz/Server/Edit/Historic/Relationship.pm
```diff
@@ -220,18 +220,18 @@ sub _display_relationships {
             my $entity1_id = $_->{entity1_id};
             my $entity0 = $loaded->{ $model0 }{ $entity0_id } ||
                 $self->c->model($model0)->_entity_class->new(
-                    id => $entity0_id,
+                    id => $entity0_id // 0,
```
yvanzo, line 223: Even though it fix the ISE, it still causes the warning “Use of uninitialized value `$entity0_id` in hash element”.

I would suggest to make your change once and for all when setting the `*_id` variables instead:

```
            my $entity0_type = $_->{entity0_type} // 0;
            my $entity1_type = $_->{entity1_type} // 0;
```

yvanzo:
Optionally, you might want to take a look at these other warnings which occur on the same edit page:

```log
[warning] GET http://localhost:5000/edit/12587968 caused a warning: Use of uninitialized value in addition (+) at lib/MusicBrainz/Server/Entity/Relationship.ptity/Relationship.pm line 271.
[warning] GET http://localhost:5000/edit/12587968 caused a warning: Use of uninitialized value in addition (+) at lib/MusicBrainz/Server/Entity/Relationship.pm line 271.
[warning] GET http://localhost:5000/edit/12587968 caused a warning: Use of uninitialized value $entity0_id in hash element at lib/MusicBrainz/Server/Edit/Historic/Relationship.pm line 225.
[warning] GET http://localhost:5000/edit/12587968 caused a warning: Use of uninitialized value in addition (+) at lib/MusicBrainz/Server/Entity/Relationship.pm line 271.
[warning] GET http://localhost:5000/edit/12587968 caused a warning: Use of uninitialized value in addition (+) at lib/MusicBrainz/Server/Entity/Relationship.pm line 271.
[warning] GET http://localhost:5000/edit/12587968 caused a warning: Use of uninitialized value $phrase in substitution (s///) at lib/MusicBrainz/Server/Entity/Relationship.pm line 242.
```

reosarevok:
I also fixed the other warnings, although I'm not sure if there's a better way to do so :)

lib/MusicBrainz/Server/Entity/Relationship.pm
```diff
@@ -279,9 +281,9 @@ around TO_JSON => sub {
         entity1_credit  => $self->entity1_credit,
         entity0_id      => $self->entity0_id,
         entity1_id      => $self->entity1_id,
-        id              => $self->id + 0,
-        linkOrder       => $self->link_order + 0,
-        linkTypeID      => $link->type_id + 0,
+        id              => $self->id ? $self->id + 0 : 0,
+        linkOrder       => $self->link_order ? $self->link_order + 0 : 0,
+        linkTypeID      => $link->type_id ? $link->type_id + 0 : 0,
```
mwiencek, line 286: For `id`/`linkTypeID` I think it makes more sense to output `undef` (null) if they're unset, not 0, so that we don't have a special numeric case there. For `linkOrder` outputting 0 is fine since that's the default value at the database level.

lib/MusicBrainz/Server/Edit/Historic/Relationship.pm
```diff
@@ -216,7 +219,13 @@ sub _display_relationships {
...
             my $model0 = type_to_model( $_->{entity0_type} );
             my $model1 = type_to_model( $_->{entity1_type} );
-            my $entity0_id = $_->{entity0_id};
+            my $entity0_id;
+            if (defined $_->{entity0_id}) {
+                $entity0_id = $_->{entity0_id};
+            } else {
+                $entity0_id = $fake_entity_id;
+                $fake_entity_id++;
+            }
```
mwiencek, line 228: That should be exactly equivalent to just
```suggestion
            my $entity0_id = $_->{entity0_id} // $fake_entity_id++;
```

since Perl only evaluates `$fake_entity_id++` if the left-hand side of `//` is `undef`. `$fake_entity_id++` returns the ID before the increment happens.

PR #1545, opened 2020-06-02
MBS-10856: Trim whitespace on relationship credits

reosarevok:
### Fix MBS-10856

We have a control_for_whitespace_entity0_credit constraint, so trying to add a relationship credit ending in whitespace was causing an error. This actually trims the credits before submitting, as we do with other strings.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
@@ -349,6 +349,9 @@ sub process_relationship {
     $data->{entity0} = $data->{entities}->[0];
     $data->{entity1} = $data->{entities}->[1];
 
+    trim_string($data, 'entity0_credit');
+    trim_string($data, 'entity1_credit');
+
     my $begin_date = clean_partial_date(delete $data->{begin_date});
     my $end_date = clean_partial_date(delete $data->{end_date});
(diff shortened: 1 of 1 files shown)


reosarevok:
We do the same on the entity names and whatnot, so I don't think this is a problem. If we decide we want to avoid trimming other whitespace then I expect that wouldn't be exclusive to relationship credits.

yvanzo:
See “space bugs” MBS-5555 and MBS-9576.

reosarevok:
Sure, I know of those, but having this be consistent with other bits is not a problem IMO. If we decide those are actual bugs and not wontfixes we'll want to fix the trim function, not each case separately :)

yvanzo:
IMHO those are actual bugs, so replacing one bug with two doesn’t seem right. :/

mwiencek:
They seem like STYLE issues to me, personally. We've been stripping those since 2012 AFAICT and I can see an argument about invisible characters being harder for almost everyone to edit and easier to accidentally revert. We'd also have to decide how our whitespace collapsing logic should continue to work.

I'm not against allowing them, just saying we might want to make minor improvements to the interface to make them more visible to editors first. And I don't think trimming these fields like we do almost every other (except join phrases) is incorrect just because we might change how the trimming works later. :)

PR #1546, opened 2020-06-02
MBS-10858:  Warn about feat. artists in track titles

reosarevok:
### Implement MBS-10858

This is very rarely correct (but sometimes it can be, such as in imagined title "They Read About It in an FT Article" about the good old Financial Times). As such, it makes sense to ask users if they're sure about what they're doing here.


root/static/scripts/release-editor/fields.js
```diff
@@ -294,6 +295,10 @@ class Track {
         return hasVariousArtists(this.artistCredit());
     }
 
+    hasFeatOnTitle() {
+        return this.name().match(featRegex);
```
mwiencek, line 299: `.test()` is usually preferred if we don't use the match and just want a boolean
```suggestion
        return featRegex.test(this.name());
```

yvanzo (commented):
It works in the overall but doesn’t match track titles that contain `(feat. Reo)`. It should probably use more than just `featRegex` from `guessFeat.js`.

reosarevok:
That works but expects unbracketed feat., it seems - so added a replace to ensure brackets are not there. Also changed the message a bit to explicitly mention the guess feat. button.

root/static/scripts/release-editor/fields.js
```diff
@@ -293,6 +294,10 @@ class Track {
         return hasVariousArtists(this.artistCredit());
     }
 
+    hasFeatOnTitle() {
+        return featRegex.test(this.name().replace(/[\[\]()]/g," "));
+    }
+
```
yvanzo, line 300: It should probably use `bracketPairs` which has some more bracket variants.

yvanzo (changes requested):

root/static/scripts/release-editor/fields.js
reosarevok: Isn't that the same two but with and without spaces?

yvanzo: No, these are Japanese variants.

reosarevok: Oooooh! Ok :)

reosarevok: So, bracketPairs is annoying because they're, well paired. I just copied the Japanese variants here too instead (because the pairs are only useful to check they are balanced and whatnot, which we don't care about here).

root/static/scripts/release-editor/fields.js
```diff
@@ -296,6 +297,10 @@ class Track {
         return hasVariousArtists(this.artistCredit());
     }
 
+    hasFeatOnTitle() {
+        return featRegex.test(this.name().replace(/[\[［\]］(（)）]/g,' '));
+    }
+
```
yvanzo, line 303: Did not test but that might be a way to use `bracketPairs`:

```suggestion
    const bracketRegex = new RegExp(bracketPairs.flat().join(''), 'g');

    hasFeatOnTitle() {
        return featRegex.test(bracketRegex,' '));
    }

```

root/static/scripts/release-editor/fields.js
yvanzo: So just flatten them, see below <https://github.com/metabrainz/musicbrainz-server/pull/1546#discussion_r485618374>.

root/static/scripts/release-editor/fields.js
mwiencek: They still have to be in a character class and escaped with `escapeRegExp` if you wanna do that. (Also, the brackets are meant to be removed from the name. :))

reosarevok: Do you feel it's worth the extra effort as opposed to just listing them again there?

mwiencek: Not really, but a comment saying to keep them in sync could be helpful.

reosarevok: Added comments

mwiencek (approved):

reosarevok:
My understanding is that we decided against the changes requested in the end? Does it otherwise seem ready to merge?

yvanzo:
> My understanding is that we decided against the changes requested in the end?

Do you have a link to this discussion? Or a summary? There is no such discussion here nor in comment to the ticket.

reosarevok:
Well, I was looking at the comments with @mwiencek above, where he said the changes did not seem worth the effort and keeping the two in sync with a comment was good enough :)

reosarevok:

root/static/scripts/release-editor/fields.js
yvanzo: > Do you feel it's worth the extra effort

The suggested code might need to be updated again, but I did update it immediately after the next comment, so the “extra effort” should be hopefully limited.

> Added comments

AFAICT, in the current commit https://github.com/metabrainz/musicbrainz-server/pull/1546/commits/71706724264a2ec195a6528b083e19f07f8b159b, comments have been added to the various `bracketPairs`, not to `hasFeatOnTitle`.

reosarevok: I was confused because it didn't seem to be working, but it just needed the result of `escapeRegExp` to be wrapped in `[]`. Thanks!

derat:
Awesome to see that this already exists! Does it just need a rebase (assuming that tests pass)?

mwiencek (approved):

PR #1549, opened 2020-06-04
MBS-10883: Revert Maybe->Optional for Remove PUID

reosarevok:
### Fix MBS-10883

I was overzealous with this change last time - while client_version should be Optional, recording_puid_id can be there and be undef, and it should be Maybe. I'm expecting puid_id to be the same.


reosarevok:
Yeah, we should just close this one, I guess :)

PR #1551, opened 2020-06-04
MBS-6864: Drop all PUID-related code

reosarevok:
### Implement MBS-6864

We're dropping our old PUID edits, so all this code is no longer of any use.


mwiencek:
It'd be good if WS2 requests that have `inc=puids` don't start to fail but remain a no-op as before, so people don't have to update their code all of a sudden. I haven't tested the patch to see if that's the case, but I see it being removed from tests. Ideally we leave at least one test for that.

reosarevok:
Sigh. Ok, I guess I can add a separate ticket to drop that with ws/3?

PR #1552, opened 2020-06-04
Sanitize IP hashes on staging servers

yvanzo:
# Problem
IP hashes were stored and displayed to everyone on `test.musicbrainz.org` since everyone has account admin permissions.

# Solution
Do not store IP hashes on staging servers running a sanitized database.
This is the same condition that is already used to inform that all passwords have been reset.

# Action

1. Set `DB_STAGING_SERVER_SANITIZED` to `1` for `test.musicbrainz.org`

### DIFF
--- a/lib/DBDefs.pm.sample
+++ b/lib/DBDefs.pm.sample
@@ -209,8 +209,10 @@ sub WEB_SERVER                { "www.musicbrainz.example.com" }
 # If left empty the default value will be shown.
 # sub DB_STAGING_SERVER_DESCRIPTION { '' }
 
-# Only change this if running a non-sanitized database on a dev server,
-# e.g. http://test.musicbrainz.org.
+# Only change this if running a non-sanitized database on a staging server,
+# e.g. http://beta.musicbrainz.org.
+# * It shows a banner informing that 'all passwords have been reset to "mb"'.
+# * It disables the IP lookup admin tool.
 # sub DB_STAGING_SERVER_SANITIZED { 1 }
 
 # Testing features enable "Accept edit" and "Reject edit" links on edits,
--- a/lib/MusicBrainz/Server/Controller/Root.pm
+++ b/lib/MusicBrainz/Server/Controller/Root.pm
@@ -226,14 +226,18 @@ sub begin : Private
         my $store = $c->model('MB')->context->store;
 
         if ($c->user_exists) {
-            my $ip_md5 = md5_hex($c->req->address);
-            my $user_id = $c->user->id;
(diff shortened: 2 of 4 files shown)


lib/DBDefs.pm.sample
```diff
@@ -210,7 +210,7 @@ sub WEB_SERVER                { "www.musicbrainz.example.com" }
 # sub DB_STAGING_SERVER_DESCRIPTION { '' }
 
 # Only change this if running a non-sanitized database on a dev server,
-# e.g. http://test.musicbrainz.org.
+# e.g. http://beta.musicbrainz.org.
```
mwiencek, line 213: Ah, the old message is from when test.mb used to have peoples' real passwords, a long time ago. So it makes sense to update this, but it's now a bit misleading too because beta isn't a dev server.

I'd also change "dev server" to "staging server" above, and I think it'd be helpful to add a short list of what this actually does, now that we're overloading it a bit:
 1. Makes the "all passwords have been reset to 'mb'" banner visible.
 2. Disables the IP lookup admin tool.

lib/MusicBrainz/Server/Controller/Root.pm
```diff
@@ -225,7 +225,10 @@ sub begin : Private
     try {
         my $store = $c->model('MB')->context->store;
 
-        if ($c->user_exists) {
+        if ($c->user_exists && !(
+                $c->stash->{server_details}->{staging_server} &&
+                $c->stash->{server_details}->{is_sanitized}))
+        {
```
mwiencek, line 231: This also disables the notes-received functionality, but we should keep that part enabled.

yvanzo: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/1552/commits/2d8bdb6f04cd054f3f066f3873e6527b4457c73a

lib/DBDefs.pm.sample
yvanzo: Added in https://github.com/metabrainz/musicbrainz-server/pull/1552/commits/d93249ffc34b99266e91bf2093e566eebe3409d1

yvanzo:
Just fixed checking DBDefs keys instead of not yet defined stash entries in commit <https://github.com/metabrainz/musicbrainz-server/pull/1552/commits/db5abe3ea7c808bf404aa608f4dafb0bc9f78be0>.

mwiencek:
I'd say it's preferable to do that in both places even.

yvanzo:
To be addressed in https://github.com/metabrainz/musicbrainz-server/pull/1555.

PR #1554, opened 2020-06-05
MBS-10887: fix country/date text CSS alignment

loujine:
# Problem

MBS-10887

# Solution
Fix CSS

### DIFF
--- a/root/static/styles/entity.less
+++ b/root/static/styles/entity.less
@@ -124,17 +124,11 @@ div.release-events-container {
                     box-sizing: border-box;
                     display: inline-block;
                     width: 4em;
-
-                    &.no-country {
-                        /*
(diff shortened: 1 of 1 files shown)


reosarevok:
Keep in mind this changes it everywhere, including the release sidebars (where the flag now looks out of center to me).

loujine:
I don't see it but I'm not surprised to see side-effects :)
I guess the solution would be to apply the fix only in tables. I tried also changing the 'vertical-align' property (which is what people recommend to counter the display: inline-block responsible for the mismatch in the first place) but didn't find a good solution.
I'll take another look later, but if you find a better solution in the meantime feel free :)

loujine:
New version: replacing padding-top only inside tables. This means the 3px on .no-country to align the release date column text with the release country one is not needed anymore

PR #1556, opened 2020-06-06
Mbs 8625 release data quality icons

loujine:
Draft PR for MBS-8625
I don't have a strong feeling for UI-related items usually, so I'm fine with this PR being rejected and completed by someone else.

# Problem
MBS-8625: add icons for releases data quality

# Solution
see MBS-8625: adding column "DQ" where relevant with basic unicode circles (green = high quality, orange = low quality) with full description on hover

# Action
* I will need some help to check that i18n is dealt with properly


yvanzo:
I updated PR title according to “[Contributing § Title](https://github.com/metabrainz/musicbrainz-server/blob/master/CONTRIBUTING.md#title)” so that the bot auto updates the related ticket.

root/static/styles/layout.less
```diff
@@ -604,6 +604,28 @@ span.video {
...
+    display: inline-block;
+    height: 14px;
+    vertical-align: middle;
+    width: 6px;
+}
+
+/* Don't try to add the quality icons on the release overview tab */
+.tabs .high-quality-release { display: none; }
+.tabs .low-quality-release { display: none; }
+
```
reosarevok, line 628: Is it intentional that it is shown on the release *header* though?

loujine:
Sorry I didn't see the approval before I pushed more stuff.
My todo list before I think this will be ready:
- squash commits
- go back to dots
- make sure the dot is not displayed on release page header (should be a CSS rule on h1)
- possibly look at how to display on edit pages
- possibly look at how to display on the release details sidebar

I'll try to have something ready tomorrow

yvanzo:
> My todo list before I think this will be ready

I copied it as a checklist into the initial comment, it makes easier to mark progress for you and to check the state of readiness from the list of pull requests.

loujine:
Updated the PR with the icon on the release sidebar.
I didn't see how to apply the icon in the edit window, I suggest to wait until that template is converted to React.

I think the PR is in a good state for merging

root/static/styles/layout.less
```diff
@@ -604,6 +604,34 @@ span.video {
...
+    content: '🟠';
+}
+
+/* Don't add the quality icons on the release overview tab */
+.tabs .high-data-quality { display: none; }
+.tabs .low-data-quality { display: none; }
+
+/* Don't add the quality icons on the release page title */
+h1 .high-data-quality { display: none; }
+h1 .low-data-quality { display: none; }
```
mwiencek, line 633: Would it make sense to have a `showDataQuality` prop on `EntityLink` instead so we can specify that where it should be shown, instead of having a blacklist in the CSS? I guess it depends on whether it's intended to show everywhere other than a few specific cases.

reosarevok:
While I'm bikeshedding, looking at https://test.musicbrainz.org/release-group/29a9ec3b-a9f6-45a7-89fc-ef5cdcbdac3b the circles look a bit off-center? But maybe I'm imagining it. I really like it, otherwise!

yvanzo:
@reosarevok: Circles are correctly placed on the baseline but they are slightly smaller than flags:
* Flags have the same height as uppercase letters.
* Circles have the same height as lowercase letters.

It looks fine to me. I would also be fine with discussing icon design in a further pull request/ticket.

loujine:
@reosarevok 
> the circles look a bit off-center?
Yes, they are defined as CSS property 'content' in order to use directly the unicode glyphs, so they are considered as text and aligned w.r.t. the bottom of the text. I don't think the vertical-align property can fix this.
If we want to use a background-image (as for the 'video' recordings) we need to define png or svg images instead of the glyphs themselves.. I thought we could wait a bit to see if there are other suggestions for the icon look and color before doing that.

root/static/styles/layout.less
loujine: Since the first idea of the ticket was "show as much as possible the quality to push people to improve low-quality releases", I think we want to show the icons everywhere except a few places.
I thought about having a new prop on EntityLink but didn't dare since I don't have a good grasp on the react codebase. It would probably be much faster for a core developer to do it than fix my mess :)
Oh and I copied the CSS behavior from the recording 'video' attribute, so the same discussion probably applies to it too.

PR #1559, opened 2020-06-08
 MBS-10381: Disallow using a work language multiple times

reosarevok:
### Fix  MBS-10381

While this seems to already be dropped silently on the DB level (since entering an Add work edit with the same language twice results on a work with just one language) it makes sense to not allow entering this in the first place, since it is meaningless and likely to be a user error.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1503


lib/MusicBrainz/Server/Form/Work.pm
```diff
@@ -77,20 +78,30 @@ sub validate_languages {
...
                 $is_valid = 0;
-            } elsif ($language_field->value == 486) {
+            } elsif ($language_id == 486) {
                 $language_field->add_error(
                     l('You can’t select “~[No lyrics~]” and a lyrics language at the same time.')
                 );
                 $is_valid = 0;
+            } elsif (any { $language_id == $_ } @used_languages) {
+                $language_field->add_error(
+                    l('You can’t select the same language more than once.')
```
yvanzo, line 100: Same minor comment as https://github.com/metabrainz/musicbrainz-server/pull/1503#discussion_r437701021

PR #1564, opened 2020-06-10
MBS-10895: conflicting discids report

loujine:
# Problem
MBS-10895

# Solution
New report (.pm and .js)
Add subPath to report/<ReleaseList> react class in order to have links directly to release/uuid/discids page

# Checklist for author
Ready for review.
Works on my local instance with sample data

# Action
Possibly change what the cutoff should be, I started with a 5 minutes minimal difference between discids on the same medium

### DIFF
--- a/root/report/ReleasesConflictingDiscIds.js
+++ b/root/report/ReleasesConflictingDiscIds.js
@@ -1,0 +1,54 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import Layout from '../layout';
+import formatUserDate from '../utility/formatUserDate';
+
+import ReleaseList from './components/ReleaseList';
+import FilterLink from './FilterLink';
+import type {ReportDataT, ReportReleaseT} from './types';
+
+const ReleasesConflictingDiscIds = ({
+  $c,
+  canBeFiltered,
+  filtered,
+  generated,
+  items,
+  pager,
+}: ReportDataT<ReportReleaseT>): React.Element<typeof Layout> => (
+  <Layout $c={$c} fullWidth title={l('Releases with conflicting disc IDs')}>
+    <h1>{l('Releases with conflicting disc IDs')}</h1>
+
+    <ul>
+      <li>
+        {l(`This report shows releases that have conflicting disc IDs on the
+            same medium with significant differences in duration. This usually
+            means a disc ID was applied to the wrong medium or the wrong
+            release.`)}
+      </li>
+      <li>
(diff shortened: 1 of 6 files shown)


loujine:
> Also, can't you just use leadout_offset for the total duration?
Ah, I completely missed that column. That should make the code and the explanation simpler.

> Ideally we'd compare track by track
I will think about it. I had another query in mind to detect discids shared by different releases using different recordings, and that needs to look at tracks separately too (https://github.com/loujine/musicbrainz-sql/blob/master/discid_linked_to_releases_with_different_recordings.sql#L31)

root/report/ReleasesConflictingDiscIds.js
```diff
@@ -0,0 +1,54 @@
...
+      </li>
+      <li>
+        {texp.l('Generated on {date}',
+                {date: formatUserDate($c, generated)})}
+      </li>
+
+      {canBeFiltered ? <FilterLink $c={$c} filtered={filtered} /> : null}
+    </ul>
+
+    <ReleaseList items={items} pager={pager} subPath='discids' />
```
loujine, line 49: Ideally I'd like the href to be to '/release/uuid/discids#discN' where N is read from the column 'position' of the SQL table, for the rare case where a release boxset has 300 discids.
But a link to '/release/uuid/discids' is good enough for the moment

loujine:
PR updated from @reosarevok 's comments.

I didn't update copyright notices on ReportFactory.pm and others, tell me if I should

root/report/ReleasesConflictingDiscIds.js
```diff
@@ -0,0 +1,54 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
```
yvanzo, line 8: > I didn't update copyright notices on ReportFactory.pm and others, tell me if I should

It isn’t needed except if you want to keep your rights under the copyright, see <https://github.com/metabrainz/musicbrainz-server/pull/1568#discussion_r444129691>. Please confirm you are voluntarily assigning the copyright to the MetaBrainz Foundation or make the relevant changes.

Nitpicking: The title of commit https://github.com/metabrainz/musicbrainz-server/pull/1564/commits/b49e79f29b10ec2a0c74c3f66592abc9168e1c88 is slightly misleading: we might think the linked ticket is about adding subPath property. I found similar examples in your other PRs. Please do not start commit’s message with ticket’s key if it isn’t directly related. If you want to specify this refactoring is required to implement this ticket, please mention ticket’s key later in commit’s message.

loujine: Force-pushed with fixes to the copyright notice and commit message

loujine:
Rebased on master with fixes from yvanzo's comments

I used my name on the .pm file since I wrote the SQL query, but otherwise I confirm that in commits/PR I submit to this repo, I am voluntary assigning copyright to the MetaBrainz Foundation.

PR #1566, opened 2020-06-10
Fix error message about Instagram `/explore` URLs

yvanzo:
# Problem

From <https://community.metabrainz.org/t/incomplete-string-is-not-translatable/478813/5?u=yvanzo>, Instagram `/explore` URLs are not necessarily about location/place, these can be about artist, event, label, or series instead.

# Solution

Generalize the wording as suggested from <https://chatlogs.metabrainz.org/brainzbot/metabrainz/msg/4598170/>.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1472,10 +1472,8 @@ const CLEANUPS = {
       if (/^https:\/\/www\.instagram\.com\/explore\//.test(url)) {
         return {
           error: exp.l(
-            `Instagram “{explore_url_pattern}” links are not officially
-             connected to a location.
-             If you want to link a place to a location, try and find
-             the place’s Instagram profile instead, if there is one.`,
+            `Instagram “{explore_url_pattern}” links are not allowed,
+             please link to a profile instead, if there is one.`,
```
reosarevok, line 1476: Nitpick: 
```suggestion
            `Instagram “{explore_url_pattern}” links are not allowed.
             Please link to a profile instead, if there is one.`,
```

yvanzo: Applied, thanks.

PR #1567, opened 2020-06-10
MBS-10898: Do not recompute all part numbers when swapping parts

loujine:
# Problem
MBS-10898

# Solution
Swap part numbers instead of recomputing default part numbers 1 to N.


loujine:
Needs more work, this solution will not work everywhere (at least breaks the test where two work-parts exist with linkOrder 0)

loujine:
Cancelling, the solution is not good enough

PR #1568, opened 2020-06-11
MBS-10770: add report about relationships with future dates

loujine:
# Problem
MBS-10770 new report about recordings with relationships having dates in the future

# Solution
New report (.pm + .js) to display dates / relation type / link to recording

The SQL query looks for begin dates > current year OR end date > (current year + 5y)

# Action
I tried to keep the SQL query straightforward. On the JS side I added the support of date columns on RecordingRelationshipList instead of creating a new one (even if for this report the Artist column is not needed).
Works on my local instance with the sample data so I don't expect more work to be needed (except i18n) before merging.

### DIFF
--- a/lib/MusicBrainz/Server/Report/RecordingsWithFutureDates.pm
+++ b/lib/MusicBrainz/Server/Report/RecordingsWithFutureDates.pm
@@ -1,0 +1,67 @@
+package MusicBrainz::Server::Report::RecordingsWithFutureDates;
+use Moose;
+
+with 'MusicBrainz::Server::Report::RecordingReport',
+     'MusicBrainz::Server::Report::FilterForEditor::RecordingID';
+
+sub query {
+    "
+        WITH
+        link AS (
+            SELECT
+                l.id,
+                l.begin_date_year AS begin,
+                l.end_date_year AS end,
+                lt.gid AS link_gid,
+                lt.name AS link_name,
+                lt.entity_type0,
+                lt.entity_type1
+            FROM
+                link AS l
+                JOIN link_type AS lt ON lt.id = l.link_type
+            WHERE
+                l.begin_date_year > date_part('year', now())
+                OR l.end_date_year > date_part('year', now()) + 5
+        )
+        SELECT
+            link.begin,
+            link.end,
+            link.link_gid,
+            link.link_name,
+            CASE
+                WHEN entity_type0 = 'recording' THEN l_.entity0
+                WHEN entity_type1 = 'recording' THEN l_.entity1
+                ELSE NULL
+            END AS recording_id,
+            row_number() OVER (ORDER BY link.begin, link.end)
+        FROM
+            link
+            JOIN (
+                SELECT link, entity0, entity1 FROM l_area_recording
+                UNION ALL
+                SELECT link, entity0, entity1 FROM l_artist_recording
+                UNION ALL
+                SELECT link, entity0, entity1 FROM l_label_recording
+                UNION ALL
+                SELECT link, entity0, entity1 FROM l_place_recording
+                UNION ALL
+                SELECT link, entity0, entity1 FROM l_recording_recording
+                UNION ALL
+                SELECT link, entity0, entity1 FROM l_recording_work
+            ) AS l_ ON l_.link = link.id
+    ";
+}
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2020 Jerome Roy
+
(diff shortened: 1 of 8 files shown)


root/report/RecordingsWithFutureDates.js
```diff
@@ -0,0 +1,53 @@
...
+
+    <ul>
+      <li>
+        {exp.l(
+          `This report shows recordings with relationships using dates in the
+           future. Those are probably typos (e.g. 2109 instead of 2019).`
+        )}
+      </li>
+      <li>
+        {texp.l('Total releases found: {count}',
```
reosarevok, line 39: Recordings, or probably better relationships, not releases :)

loujine:
> I wonder if we could just get both relationship entities instead
That was my first idea (cf query in the original ticket). but I got stuck on how to tell React which entity types were used.
As far as I understand, most reports only return the entities 'id' and another perl script fetches the full entity info afterwards to send it to React (so that <EntityList entity={entity}> just works).

Here I would probably have to make the SQL query much more complicated or write something like RecordingReport.pm that would read the entity type and id, but I don't speak Perl :p 

I will check if I can hide the Artist column in that report and fix the other points you mention.

lib/MusicBrainz/Server/Report/RecordingsWithFutureDates.pm
```diff
@@ -0,0 +1,77 @@
...
+                l.id,
+                l.begin_date_year AS begin,
+                l.end_date_year AS end,
+                lt.gid AS link_gid,
+                lt.name AS link_name,
+                lt.entity_type0,
+                lt.entity_type1
+            FROM
+                link AS l
+                INNER JOIN link_type AS lt ON lt.id = l.link_type
```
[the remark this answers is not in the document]
loujine: fixed

root/report/RecordingsWithFutureDates.js
loujine: fixed

loujine:
Pushed suggested fixes + hide Artist column on this report

loujine:
> the only other table I might check is l_artist_artist

From what I remember during a test on the full database a few weeks ago, there was at least a "X main conductor for orchestra Y" with dates 2020-2030, which is the kind of false positive I would expect

root/report/RecordingsWithFutureDates.js
```diff
@@ -0,0 +1,53 @@
...
+const RecordingsWithFutureDates = ({
+  $c,
+  generated,
+  items,
+  pager,
+}: ReportDataT<ReportRecordingRelationshipT>): React.Element<typeof Layout> => (
+  <Layout
+    $c={$c}
+    fullWidth
+    title={l('Recordings with dates in the future')}
```
reosarevok, line 27: The title suggests the recording itself has dates, maybe "Recordings with relationships having dates in the future" would be better? It's long, but so are other report titles.

loujine: fixed

lib/MusicBrainz/Server/Report/RecordingsWithFutureDates.pm
```diff
@@ -0,0 +1,77 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+=cut
```
yvanzo, line 77: Please use the [newest template](https://musicbrainz.org/doc/Development/Server_Coding_Style#Perl_Module_Template_and_Documentation_Guidelines) instead, see below replacement.

@loujine, I’m not sure we discussed this before: As an individual, you can assign the copyright either to yourself or to the foundation, so feel free to put your name and/or your pseudonym in place of “MetaBrainz Foundation” if you prefer.

```suggestion
=head1 COPYRIGHT AND LICENSE

Copyright (C) 2020 MetaBrainz Foundation

This file is part of MusicBrainz, the open internet music database,
and is licensed under the GPL version 2, or (at your option) any
later version: http://www.gnu.org/licenses/gpl-2.0.txt

=cut
```

loujine: I don't think we discussed it before. I updated the copyright notice to the new template and used my name

loujine:
Rebased on latest master, fixed copyright notice and squashed the fix commits

PR #1569, opened 2020-06-13
MBS-10894: Add LC code in help bubble after adding label in release editor

loujine:
# Problem
MBS-10894 Display label code in the release editor

# Solution
Add label code in the info bubble after selection of the label

# Checklist for author
Tested on local instance with sample database, looks ready to me

### DIFF
--- a/root/static/scripts/common/entity.js
+++ b/root/static/scripts/common/entity.js
@@ -11,6 +11,8 @@ import _ from 'lodash';
 import * as React from 'react';
 import * as ReactDOMServer from 'react-dom/server';
 
+import formatLabelCode from '../../../utility/formatLabelCode';
+
 import ArtistCreditLink from './components/ArtistCreditLink';
 import EditorLink from './components/EditorLink';
 import EntityLink from './components/EntityLink';
@@ -27,6 +29,7 @@ import {
 } from './immutable-entities';
 import linkedEntities from './linkedEntities';
 import MB from './MB';
+import bracketed from './utility/bracketed';
 import clean from './utility/clean';
 import formatTrackLength from './utility/formatTrackLength';
 
@@ -231,9 +234,18 @@ import formatTrackLength from './utility/formatTrackLength';
 
     class Label extends CoreEntity {
         selectionMessage() {
+            const code = this.label_code;
+            let codeSection = '';
+
+            if (code) {
+                codeSection = ' ' + bracketed(exp.l(
+                    'Label code: {code}',
+                    {code: formatLabelCode(code)},
+                ));
+            }
             return ReactDOMServer.renderToStaticMarkup(
                 exp.l('You selected {label}.', {label: this.reactElement({target: '_blank'})}),
-            );
+            ) + codeSection;
         }
     }
 


root/static/scripts/common/entity.js
```diff
@@ -231,8 +231,14 @@ import formatTrackLength from './utility/formatTrackLength';
 
     class Label extends CoreEntity {
         selectionMessage() {
+            let code = this.toJSON().label_code;
```
reosarevok, line 234: I don't think you need toJSON() here? Also, I'd rather this was a const and we had a different variable for the code. This variable can already start as the empty string, that way you don't need the `else`.

```suggestion
            const code = this.label_code;
            let codeSection = '';            
```

root/static/scripts/common/entity.js
```diff
@@ -231,8 +231,14 @@ import formatTrackLength from './utility/formatTrackLength';
...
+            if (code) {
+                code = ` (Label Code: LC-${"0".repeat(5 - String(code).length)}${code})`;
```
reosarevok, line 236: We can use the existing `formatLabelCode` for this. You'll need to import it:

`import formatLabelCode from '../../../utility/formatLabelCode';`

Also, this should be translatable. So, something like:

```suggestion
                codeSection = ' ' + bracketed(exp.l(
                    'Label Code: {code}',
                    {code: formatLabelCode(code)},
                ));
```

We use lowercase "Label code" on the label sidebar, FWIW. Not sure about that one.

root/static/scripts/common/entity.js
```diff
@@ -231,8 +231,14 @@ import formatTrackLength from './utility/formatTrackLength';
...
+            } else {
+                code = '';
+            }
```
reosarevok, line 239: The else wouldn't be needed anymore.

root/static/scripts/common/entity.js
```diff
@@ -231,8 +231,14 @@ import formatTrackLength from './utility/formatTrackLength';
...
             return ReactDOMServer.renderToStaticMarkup(
-                exp.l('You selected {label}.', {label: this.reactElement({target: '_blank'})}),
+                exp.l('You selected {label}{code}.', {label: this.reactElement({target: '_blank'}), code: code}),
```
reosarevok, line 241: This is waaay too long a line - remember our limit is 78 chars. To be fair, it was already too long before, but still, might as well improve that :)
```suggestion
                exp.l(
                    'You selected {label}{code}.',
                    {
                        label: this.reactElement({target: '_blank'}),
                        code: codeSection,
                    },
                ),
```

root/static/scripts/common/entity.js
loujine: Indeed, no need for the toJSON

root/static/scripts/common/entity.js
loujine: Added the helper function and 'bracketed' import. Fixed to lower case "Label code"

root/static/scripts/common/entity.js
loujine: Sure. I copied your fix

loujine:
Should be fixed now

root/static/scripts/common/entity.js
```diff
@@ -27,7 +27,9 @@ import {
 } from './immutable-entities';
 import linkedEntities from './linkedEntities';
 import MB from './MB';
+import bracketed from './utility/bracketed';
 import clean from './utility/clean';
+import formatLabelCode from '../../../utility/formatLabelCode';
```
reosarevok, line 32: This should be in a separate import group (does your IDE allow checking for eslint issues? this would have shown as a warning when running eslint on the file)

loujine: looks like my eslint config was broken, I see it now

root/static/scripts/common/entity.js
```diff
@@ -231,8 +234,23 @@ import formatTrackLength from './utility/formatTrackLength';
...
+            if (code) {
+                codeSection = ' ' + bracketed(exp.l(
+                    'Label code: {code}',
+                    {code: formatLabelCode(code)},
+                ));
+            }
             return ReactDOMServer.renderToStaticMarkup(
-                exp.l('You selected {label}.', {label: this.reactElement({target: '_blank'})}),
+                exp.l(
+                    'You selected {label}{code}.',
```
mwiencek, line 248: This works though I wonder it's easy enough for translators to make sense of. Maybe we should have two separate strings instead, `'You selected {label}.'` (keeping the no-label-code case the same) and `'You selected {label} (Label code: {code}).'` for the new case? @reosarevok @yvanzo

mwiencek: Or maybe just concatenating ` + codeSection` to the result of the `ReactDOMServer.renderToStaticMarkup` call is fine, since that's basically what we do with other bracketed things.

loujine: Rebased and pushed with this solution

mwiencek: My bad, I realized now we should do the same as I suggested in https://github.com/metabrainz/musicbrainz-server/pull/1570#discussion_r443694811 so React can escape any potential HTML chars there. But I pushed the fix so we can merge this now.

PR #1570, opened 2020-06-13
MBS-9502: Add search for pre-existing barcode value in release editor

loujine:
# Problem
MBS-9502
Display warning when creating a new release if the barcode is already linked to an existing release

# Solution
Add request to ws/2/release with the barcode as the same time as the UPC/EAN validation is performed
Display a sentence suggesting to check to search page results and confirm the new release is different.

# Checklist for author
Looks ready to me.
Works on my local instance with the sample database

# Action
Please confirm the call to /ws/2/ is the right way to do this.

### DIFF
--- a/root/release/edit/information.tt
+++ b/root/release/edit/information.tt
@@ -310,6 +310,7 @@
     </p>
     <!-- ko with: target() && target().barcode -->
     <p data-bind="visible: message, text: message"></p>
+    <p data-bind="visible: existing, html: existing"></p>
     <p class="field-error" data-bind="visible: error() || confirmed()">
       <label>
         <strong>
--- a/root/release/edit/information.tt
+++ b/root/release/edit/information.tt
@@ -308,6 +308,9 @@
       [% l('Please enter the barcode of the release you are entering, see <a href="{url}" target="_blank">Barcode</a> for more information.',
            { url => doc_link('Barcode') }) %]
     </p>
+    <p>
+      [% l('If you do not know whether the release has a barcode or not, just leave this blank.') %]
+    </p>
     <!-- ko with: target() && target().barcode -->
     <p data-bind="visible: message, text: message"></p>
     <p data-bind="visible: existing, html: existing"></p>
@@ -320,9 +323,6 @@
       </label>
     </p>
     <!-- /ko -->
-    <p>
-      [% l('If you do not know whether the release has a barcode or not, just leave this blank.') %]
-    </p>
   </div>
 
   <div class="bubble" data-bind="bubble: $root.annotationBubble">
--- a/root/static/scripts/release-editor/validation.js
+++ b/root/static/scripts/release-editor/validation.js
@@ -9,8 +9,11 @@
 import $ from 'jquery';
 import ko from 'knockout';
 import _ from 'lodash';
+import * as React from 'react';
+import * as ReactDOMServer from 'react-dom/server';
 
-import expand2html from '../common/i18n/expand2html';
+import MB from '../common/MB';
+import DescriptiveLink from '../common/components/DescriptiveLink'
 import expand2text from '../common/i18n/expand2text';
 import {errorsExist} from '../edit/validation';
 
@@ -128,18 +131,32 @@ $(function () {
 function searchExistingBarcode(field, barcode) {
     utils.search("release", `barcode:${barcode}`, 1).done(data => {
         if (data.releases.length) {
-            var msg = l(
-                `There already exists at least one release with that barcode.
-                 Please check on {url|this search page} whether the release you
-                 are entering is identical.`
+
+            const msg = l(
+                `The following releases with that barcode are already in the
+                 MusicBrainz database. Please make sure you are not adding an
+                 exact duplicate of any of these:`,
             );
-            var query = `barcode%3A${barcode}&type=release&method=advanced`;
-            field.existing(
-                expand2html(
-                    msg,
-                    { url: { href: `/search?query=${query}`, target: "_blank" }}
-                )
+            const releaseList = ReactDOMServer.renderToString(
+                <>
+                    {msg}
+                    <ul>
+                        {data.releases.map(release => {
+                        const cleanedRelease = new MB.entity.Release(
+                            utils.cleanWebServiceData(release),
+                        );
+                        return (
+                            <li key={release.id}>
+                                <DescriptiveLink
+                                    entity={cleanedRelease}
+                                />
+                            </li>
+                        );
+                    })}
+                    </ul>
+                </>,
             );
+            field.existing(releaseList);
         }
     });
 }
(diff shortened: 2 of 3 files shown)


root/static/scripts/release-editor/validation.js
```diff
@@ -123,13 +125,36 @@ $(function () {
 });
 
 
+// Search releases with the same barcode
+function searchExistingBarcode(field, barcode) {
+    request({
```
loujine, line 130: I preferred to split the logic in a new function since the barcode validation block is already complex

root/static/scripts/release-editor/validation.js
```diff
@@ -123,13 +125,36 @@ $(function () {
...
+        url: `/ws/2/release/?query=barcode:${barcode}&fmt=json`
```
loujine, line 131: I couldn't find a way to run this precise query with /ws/js/, so I chose /ws/2/

loujine:
cf screenshot on https://tickets.metabrainz.org/browse/MBS-9502 for how it looks

root/release/edit/information.tt
```diff
@@ -310,6 +310,7 @@
     </p>
     <!-- ko with: target() && target().barcode -->
     <p data-bind="visible: message, text: message"></p>
+    <p data-bind="visible: existing, html: existing"></p>
     <p class="field-error" data-bind="visible: error() || confirmed()">
       <label>
         <strong>
```
loujine, line 316: Not directly related to the PR, but I think it would make more sense to have the 'If you do not know whether' block a few lines lower from here to be moved line 310, before the barcode validation messages

root/static/scripts/release-editor/validation.js
loujine: Nevermind, I found the helper function for /ws/2/ requests in the release-editor

root/release/edit/information.tt
reosarevok: I'd suggest moving it in a separate commit (in the same PR) if so :)

root/static/scripts/release-editor/validation.js
```diff
@@ -123,13 +125,36 @@ $(function () {
...
+                 Please check on {url|this search page} whether the release you
+                 are entering is identical.`
+            );
+            var query = `barcode%3A${barcode}&type=release&method=advanced`;
+            field.existing(
+                expand2html(
+                    msg,
+                    { url: { href: `/search?query=${query}`, target: "_blank" }}
+                )
+            );
```
reosarevok, line 143: It would make sense to me to just list the releases here. It shouldn't make it too big, since most barcodes aren't reused in many releases, and it would save a lot of time and make it much more likely for the user to check them. I think you should be able to just do a map on the results array and use DescriptiveLink to show the releases?

root/static/scripts/release-editor/validation.js
reosarevok: @mwiencek: is there any way of doing this with ws/js? I guess we do use ws/2 for stuff in the RE already, so maybe this is expected.

root/static/scripts/release-editor/validation.js
```diff
@@ -10,7 +10,9 @@ import $ from 'jquery';
 import ko from 'knockout';
 import _ from 'lodash';
 
+import expand2html from '../common/i18n/expand2html';
 import expand2text from '../common/i18n/expand2text';
+import request from '../common/utility/request';
```
reosarevok, line 15: :scissors:

loujine: thanks

root/static/scripts/release-editor/validation.js
loujine: I will test that idea and post a screenshot

reosarevok:
```suggestion
function searchExistingBarcode(field, barcode) {
    utils.search('release', `barcode:${barcode}`, 1).done(data => {
        if (data.releases.length) {
            const msg = l(
                `The following releases with that barcode are already in the
                 MusicBrainz database. Please make sure you are not adding an
                 exact duplicate of any of these:`,
            );
            const releaseList = ReactDOMServer.renderToString(
                <ul>
                    {data.releases.map(release => {
                        const cleanedRelease = new MB.entity.Release(
                            utils.cleanWebServiceData(release),
                        );
                        return (
                            <li key={release.id}>
                                <DescriptiveLink
                                    entity={cleanedRelease}
                                />
                            </li>
                        );
                    })}
                </ul>,
            );
            field.existing(msg + releaseList);
        }
    });
}
```

loujine: Looks good. I added the relevant imports and pushed the commit. I didn't know about most of those helper functions :)

root/release/edit/information.tt
loujine: Done

root/static/scripts/release-editor/validation.js
```diff
@@ -123,13 +127,45 @@ $(function () {
...
+        if (data.releases.length) {
+
+            const msg = l(
+                `The following releases with that barcode are already in the
+                 MusicBrainz database. Please make sure you are not adding an
+                 exact duplicate of any of these:`,
+            );
+            const releaseList = ReactDOMServer.renderToString(
+                <ul>
+                    {data.releases.map(release => {
```
mwiencek, line 142: This looks great, but I'd just put the msg inside the `renderToString` call (you can do it with a React fragment, `<> </>`) so that any potential HTML in the translated string is escaped there.
```suggestion
            const releaseList = ReactDOMServer.renderToString(
                <>
                    {msg}
                    <ul>
                        {data.releases.map(release => {
```

root/static/scripts/release-editor/validation.js
```diff
@@ -123,13 +127,45 @@ $(function () {
...
+                        return (
+                            <li key={release.id}>
+                                <DescriptiveLink
+                                    entity={cleanedRelease}
+                                />
+                            </li>
+                        );
+                    })}
+                </ul>,
+            );
```
mwiencek, line 155:
```suggestion
                    </ul>
                </>,
            );
```

root/static/scripts/release-editor/validation.js
```diff
@@ -123,13 +127,45 @@ $(function () {
...
+            field.existing(msg + releaseList);
```
mwiencek, line 156:
```suggestion
            field.existing(releaseList);
```

loujine:
@mwiencek thanks for the suggestion

loujine:
@reosarevok not sure why the PR still show "Changes requested". I believe the last open comment was your question about ws/js and ws/2

reosarevok:
Because until I approve or do anything else it will show the previous status. Once I'm behind a computer again (possibly tonight) I can recheck this :)

yvanzo:
@brainzbot retest this please

PR #1571, opened 2020-06-15
MBS-10903: Update Operabase cleanup / validation

reosarevok:
### Implement MBS-10903

While old Operabase cleaned links still redirect correctly, they have changed their URL format. The new format is simpler and no longer requires the artist name.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2023,12 +2023,12 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?(www\\.)?operabase\\.com', 'i')],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:www\.)?operabase\.com\/a\/([^\/?#]+)\/([0-9]+).*$/, 'https://operabase.com/a/$1/$2');
+      return url.replace(/^(?:https?:\/\/)?(?:www\.)?operabase\.com\/(?:a\/[^\/?#]+|artists)\/(?:[^0-9]+)?([0-9]+).*$/, 'https://operabase.com/artists/$1');
     },
     validate: function (url, id) {
       return {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2023,12 +2023,15 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?(www\\.)?operabase\\.com', 'i')],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:www\.)?operabase\.com\/a\/([^\/?#]+)\/([0-9]+).*$/, 'https://operabase.com/a/$1/$2');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?operabase\.com\//, 'https://operabase.com/');
+      url = url.replace(/^https:\/\/operabase\.com\/a\/[^\/?#]+\/([0-9]+).*$/, 'https://operabase.com/artists/$1');
```
reosarevok, line 2027: Still support the older format in case users find them online, and to make it easier to update the existing ones.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2023,12 +2023,15 @@ const CLEANUPS = {
...
+      url = url.replace(/^https:\/\/operabase\.com\/artists\/[^0-9]+([0-9]+).*$/, 'https://operabase.com/artists/$1');
+      return url;
```
yvanzo, line 2029: It can still call `replace` only once as before:

```suggestion
      return url.replace(/^(?:https?:\/\/)?(?:www\.)?operabase\.com\/(?:a\/[^\/?#]+|artists)\/([0-9]+).*$/, 'https://operabase.com/artists/$1');
```

reosarevok: I guess! It feels a bit less obvious but I guess regexes aren't too obvious anyway.

PR #1572, opened 2020-06-16
MBS-7994 / MBS-10746: Add pregap icon and indicate pregap in edits

reosarevok:
### Fix MBS-7994, MBS-10746

This adds an icon for pregap tracks, which were previously only visually identified by the default track number 0. This wasn't foolproof at all, since users can override the number and make the pregap number 1 or a non-pregap number 0.

Additionally, this should fix an issue where dropping or adding a pregap track created an edit medium that displayed no changes as long as the numbers (not positions) of tracks didn't change, since positions weren't looked at at all.

The icon currently used is basically "CD with rewind arrow" since I couldn't find anything else that seemed better. Icon improvement suggestions welcome.


reosarevok:
@brainzbot, retest this

PR #1574, opened 2020-06-19
MBS-10908: Stop recording-of rel edits showing for all recordings of a work

reosarevok:
### Fix MBS-10908

While it's generally useful to show relationship changes for a work on recordings and releases connected to the work (MBS-6532), for example if writers are being added or changed, there's one very common use case (recording-work "recording of" rels) where the edits are just noise on other recording entries. The fact that other, unrelated recording is or isn't a recording of this work should never affect the other recordings of the work in any way.

This makes it so the edit is only added to the edit histories of recordings and releases connected to the work if the relationship type is *not* "recording of". The recording and release(s) that are directly connected to the edit will still be linked through the recording side of the edit data.

Also, the previous implementation of MBS-6532 had an error in which all changes made to any work also appeared on all artists who ever recorded that work. This was unintentional, and this removes it (if it gets explicitly asked for for non-"recording of" rels it can be readded in a more intentional, not side-effecty way).

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
@@ -5,9 +5,9 @@ use namespace::autoclean;
 use MusicBrainz::Server::Constants qw( %ENTITIES );
 use MusicBrainz::Server::Data::Utils qw( type_to_model );
 
-requires 'directly_related_entities';
+requires 'directly_related_entities', 'link_type';
 
-my %expand = map { $_ => 1 } qw( recording release release_group work );
+my %expand = map { $_ => 1 } qw( recording release release_group );
 
 around _build_related_entities => sub {
     my ($orig, $self) = @_;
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
```diff
@@ -25,21 +25,35 @@ around _build_related_entities => sub {
...
         }
+    }
+
+    my $link_type =
+        # Remove relationship
+        $self->data->{relationship}{link}{link_type} //
+        # Edit relationship
+        $self->data->{link}{link_type} //
+        # Add relationship
+        $self->data->{link_type};
```
mwiencek, line 36: This is probably fine since Perl should autovivify the keys if they don't exist, but it'd be slightly cleaner to add `sub link_type { shift->data->{relationship}{link}{link_type} }`, `sub link_type { shift->data->{link}{link_type} }`,  etc. to each relevant package and then do `my $link_type = $self->link_type;` here.

lib/MusicBrainz/Server/Edit/Relationship/RelatedEntities.pm
```diff
@@ -7,7 +7,7 @@ use MusicBrainz::Server::Data::Utils qw( type_to_model );
 
 requires 'directly_related_entities';
```
mwiencek, line 8: If following the above suggestion, this should cause it to error if consumers don't implement `link_type`.
```suggestion
requires 'directly_related_entities', 'link_type';
```

mwiencek:
The fixes here won't be visible on existing data, since related entities are only linked into the `edit_${entity}` tables at the time an edit is entered, so to test this you'll have to enter a "recording of" relationship between a recording and work locally, and then check the edit histories of recordings associated with that work. (It should appear in the edit history of the recording in the relationship, but not other recordings linked to the work.)

The subscription emails script gets recent entities via the `edit_${entity}` tables, so it should fix that too.

yvanzo:
Thanks, so we should probably remove illegitimate entries from `edit_${entity}` after this change has been released in production, right?

mwiencek:
We could, though we don't normally make fixes to those tables & it might be more effort than it's worth.

mwiencek:
(That seems to have revealed a bug: the remove relationship one should be `shift->data->{relationship}{link}{type}` instead of `shift->data->{relationship}{link}{link_type}`. Fixed now.)

PR #1575, opened 2020-06-19
MBS-10901: Fix ISE when historic EditRelationship data is broken

reosarevok:
### Fix MBS-10901

We seem to have some cases where historic EditRelationship edits do not have "new" for some of the "old" relationships in the old-new pairs. This makes no sense, so it's probably just data corruption caused during the NGS migration - we can't do much about it, so this just displays those old rels unchanged (diff with themselves) which is the least neurotic of the possibilities.

### DIFF
--- a/root/edit/details/historic/EditRelationship.js
+++ b/root/edit/details/historic/EditRelationship.js
@@ -13,6 +13,7 @@ import linkedEntities from '../../../static/scripts/common/linkedEntities';
 import DescriptiveLink
   from '../../../static/scripts/common/components/DescriptiveLink';
 import DiffSide from '../../../static/scripts/edit/components/edit/DiffSide';
+import Warning from '../../../static/scripts/common/components/Warning';
 import relationshipDateText from '../../../utility/relationshipDateText';
 import {INSERT, DELETE} from '../../../static/scripts/edit/utility/editDiff';
 import {interpolateText}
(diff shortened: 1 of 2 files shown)


reosarevok:
That's a very fair point. Added a big warning message :)

mwiencek:
The only thing I wish we could improve here is that it's not clear which bits of code are related to `isDataBroken`. I can see that from the diff, but not from reading the file that things like `!newSource || oldSource.id === newSource.id` and `newRel || oldRel` (which are pretty confusing without context) are for the `isDataBroken` stuff.

Maybe these changes aren't even needed though. Since the goal is just to diff the old rels with themselves, might we be able keep the code otherwise the same and just have something like this?

```JS
                let newRel = newRels[index];
                if (isDataBroken && !newRel) {
                  newRel = oldRel;
                }
```

reosarevok:
Actually @mwiencek's suggestion sounds pretty good. Made the changes :)

root/edit/details/historic/EditRelationship.js
```diff
@@ -29,126 +30,162 @@ type Props = {
...
+                /*
+                 * Some relationships seem to have broken data where newRel
+                 * might not exist for every oldRel. In those cases, we
+                 * display oldRel unchanged (diff with itself) in order
+                 * to show *something* in the least bad way possible.
+                 */
+                let newRel = newRels[index];
+                if (isDataBroken && !newRel) {
+                  newRel = oldRel;
+                }
```
kellnerd, line 69: This also happens the other way around, i.e. `oldRel` does not exist for every `newRel`: https://musicbrainz.org/edit/4573989/data
Reported as MBS-11194

PR #1576, opened 2020-06-24
Bump react-table version to 7.2.1

yvanzo:
> The low usage we have of the API is affected by neither deprecation nor breaking change.
> 
> Reference:
> * https://github.com/tannerlinsley/react-table/releases


yvanzo:

PR #1577, opened 2020-06-24
MBS-9769: add entities added stats on editor profile

loujine:
# Problem
MBS-9769
Display the number of entities created (artists, releases, etc.) on the editor main page

# Solution
See screenshot on MBS-9769: new colum on the editor main page with the count of added entities for the main entity types (artists, releases, works, events...)

# Checklist for author
Tested on local instance with sample data, looks ready to me

# Action
I'm actually not sure how what the performance will be for editors with e.g. > 1 million edits. Can we try this on a test server?

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -13,15 +13,28 @@ import UserAccountLayout from '../components/UserAccountLayout';
 import DescriptiveLink
   from '../static/scripts/common/components/DescriptiveLink';
 import {FLUENCY_NAMES} from '../static/scripts/common/constants';
+import {compare} from '../static/scripts/common/i18n';
 import commaList from '../static/scripts/common/i18n/commaList';
 import expand2react from '../static/scripts/common/i18n/expand2react';
 import bracketed, {bracketedText}
   from '../static/scripts/common/utility/bracketed';
+import * as TYPES from '../static/scripts/common/constants/editTypes';
 import escapeRegExp from '../static/scripts/common/utility/escapeRegExp';
 import commaOnlyList from '../static/scripts/common/i18n/commaOnlyList';
 import {formatCount, formatPercentage} from '../statistics/utilities';
 import formatUserDate from '../utility/formatUserDate';
 import {canNominate} from '../utility/voting';
+
+const ADDED_ENTITIES_TYPES = {
+  artist:    N_l('Artist'),
+  cover_art: N_l('Cover Art'),
+  event:     N_l('Event'),
+  label:     N_l('Label'),
+  place:     N_l('Place'),
+  release:   N_l('Release'),
+  series:    N_lp('Series', 'singular'),
+  work:      N_l('Work'),
+};
 
 function generateUserTypesList(user: EditorT) {
   const typesList = [];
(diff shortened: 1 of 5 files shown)


loujine:
At first I coded "release groups", "mediums", "standalone recordings" and "areas" also (didn't think about instruments), then thought it would be too much information for a first draft. If the concept looks good to you I can add them again.

mwiencek:
The query takes between 850ms and 1s for editor = 326637 (reosarevok) on floyd (and it took 7s on first run when caches were cold), which is quite slow, so I think we should at least cache the results in redis for 24 hours. We can also improve the query performance slightly (by about 150ms on average for editor = 326637) by removing the `WHERE type IN (...)` part of the query, since then it only needs one index scan on `edit_idx_editor_id_desc`.

loujine:
@mwiencek : thanks for the test. Do you want me to push a new commit to remove the WHERE condition?

mwiencek:
@loujine Yeah, that'd be good since it did appear to help significantly for users with lots of edits. It'll cause a bunch of edit types we don't care about to be included in the results, but perhaps you can just add `ELSE 'other'` to the `CASE` and just ignore that key.

If you're unsure how to cache the results in Redis I can help with that, but roughly something like this (untested)
```Perl
my $cache_key = "editor:$editor_id:aded_entities_counts";
my $cached_result = $self->c->cache->get($cache_key);
return $cached_result if defined $cached_result;

# fetching the results from the DB here

$self->c->cache->set($cache_key, \%result, 60 * 60 * 24);
```

loujine:
@mwiencek : looks like your redis solution directly works on my local instance. I could check that the cache was refreshed as expected

loujine:
@yvanzo : the new version shows links only if the user is logged in + the list is ordered by labels now. I think i18n should work but I haven't tested it.

root/user/UserProfile.js
```diff
@@ -13,17 +13,30 @@ import UserAccountLayout from '../components/UserAccountLayout';
...
+const ADDED_ENTITIES_TYPES = {
+  artist:    l('Artist'),
+  event:     l('Event'),
+  release:   l('Release'),
+  series:    l('Series'),
+  work:      l('Work'),
+  label:     l('Label'),
+  place:     l('Place'),
+  cover_art: l('Cover Art'),
+};
```
yvanzo, line 38: * Should be localized at render time only, so use `N_l` instead.
* Plus context string `singular` is needed for proper localization of `Series`.

```suggestion
const ADDED_ENTITIES_TYPES = {
  artist:    N_l('Artist'),
  event:     N_l('Event'),
  release:   N_l('Release'),
  series:    N_l('Series', 'singular'),
  work:      N_l('Work'),
  label:     N_l('Label'),
  place:     N_l('Place'),
  cover_art: N_l('Cover Art'),
};
```

root/user/UserProfile.js
```diff
@@ -560,12 +627,44 @@ const UserProfileStatistics = ({
...
+        <thead>
+          <tr>
+            <th colSpan="2">
+              {exp.l('Added entities')}
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          {Object.entries(ADDED_ENTITIES_TYPES)
+          .sort((a, b) => a[1] > b[1])
```
yvanzo, line 645: Following the above suggestion, `l` should be called here instead, just before `sort`.

loujine:
@yvanzo : the localization works fine now, thanks for the explanation. Still have to fix flowlib complaining about functions in an object, I'll take a look tomorrow.

root/user/UserProfile.js
```diff
@@ -560,12 +627,45 @@ const UserProfileStatistics = ({
...
+          <tr>
+            <th colSpan="2">
+              {exp.l('Added entities')}
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          {Object.keys(ADDED_ENTITIES_TYPES)
+          .map(type => [type, ADDED_ENTITIES_TYPES[type]()])
+          .sort((a, b) => a[1] > b[1] ? 1 : -1)
```
mwiencek, line 646: The `compare` function might be better here, 'cause it uses the `Intl.Collator` API for language-specific comparisons. You've already imported it but haven't used it, did it not work?
```suggestion
          .sort((a, b) => compare(a[1], b[1]))
```

loujine: It was probably lost during one rebase, I will rebase & push with it fixed

loujine:
Rebased and pushed with eslint fixes

yvanzo (approved):
Thank you @loujine!

mwiencek (approved):

jesus2099:
It's great addition to the profile, @loujine!

Shouldn't we include only effective edits, though?
[Status is Open or Applied](https://github.com/jesus2099/konami-command/commit/8e80b3cbf4a9805625e539af8a25f0736cc07eab#diff-e87827441bce14262d8816044904e49bR840), or maybe even only Applied.

loujine:
@jesus2099 can you open a MBS ticket for that? We'll probably want to add instruments/areas/etc. too, but I'd like to have some feedback on the performance impact first

jesus2099:
Hi @loujine, here I have just added this ticket now: https://tickets.metabrainz.org/browse/MBS-11123 :)

jesus2099:
I opened https://tickets.metabrainz.org/browse/MBS-11174 about this row: https://github.com/metabrainz/musicbrainz-server/pull/1577/commits/611310db0da9226034d896ad4b74e874eea0bf5a#diff-3dd03aa9dd7cd28faeef616cbfe3249ebfc4474ad177479bc032cb899985242bR14

    export const EDIT_RELEASE_CREATE: 31 = 31;

Where we should not only consider edit type `31` for Add release, but also `216` (seen as `216,31` when you use the Search edit form).

PR #1578, opened 2020-06-24
MBS-10916: showing AcousticBrainz info in the Recording sidebar

loujine:
# Problem
MBS-10916
Display AcousticBrainz info in Recording pages sidebar

# Solution
Use React hook to fetch submission count from AB API. If not zero, fetch low-level data (tonal and rhythm).
See screenshot in MBS-10916

# Checklist for author
Tested on local database with sample data, works as expected

# Action
Check that the number of requests/data transfer amount to the AcousticBrainz API is reasonable. Another solution I tested quickly would be to display the result in the figerprints tab instead: less API requests vs lower visibility of the information

### DIFF
--- a/root/static/scripts/common/components/sidebar/AcousticBrainz.js
+++ b/root/static/scripts/common/components/sidebar/AcousticBrainz.js
@@ -1,0 +1,82 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import hydrate from '../../../../../utility/hydrate';
+import {SidebarProperty, SidebarProperties} from
+  '../../../../../layout/components/sidebar/SidebarProperties';
+
+const SidebarAcousticBrainz = ({recording}: {recording: RecordingT}) => {
+  const [count, setCount] = React.useState(0);
+  const [data, setData] = React.useState(null);
+
+  function fetchAcousticBrainzCount() {
+    const countUrl = `//acousticbrainz.org/api/v1/${recording.gid}/count`;
+    fetch(countUrl).then(
+      resp => resp.json(),
+    ).then(
+      data => {
+        setCount(data.count);
+        if (data.count) {
+          fetchAcousticBrainzData();
+        }
+      },
+    );
+  }
+
+  function fetchAcousticBrainzData() {
+    const dataUrl = `//acousticbrainz.org/api/v1/${recording.gid}/low-level`;
+    fetch(dataUrl).then(
+      resp => resp.json(),
+    ).then(
+      data => {
+        setData(data);
+      },
+    );
+  }
+
+  React.useEffect(fetchAcousticBrainzCount, [recording.gid]);
+  return count ? (
+    <>
+      <h2>
+        {l('Acoustic information')}
+      </h2>
+
+      <a className="external" href={`//acousticbrainz.org/${recording.gid}`}>
+        {texp.ln(
+          '{count} submission on AcousticBrainz',
+          '{count} submissions on AcousticBrainz',
+          count,
+          {count},
+        )}
+      </a>
+
+      {data === null ? null : (
+        <SidebarProperties>
+
+          {data.tonal.key_strength > 0.5 ? (
+            <SidebarProperty className="acousticbrainz_key" label={l('Key:')}>
+              {`${data.tonal.key_key} ${data.tonal.key_scale}`}
+            </SidebarProperty>
+          ) : null}
+
+          <SidebarProperty className="acousticbrainz_bpm" label={l('BPM:')}>
+            {Math.round(data.rhythm.bpm)}
+          </SidebarProperty>
+        </SidebarProperties>
+      )}
+    </>
+  ) : null;
+};
+
+export default (hydrate<{recording: RecordingT}>(
+  'div.acousticbrainz',
+  SidebarAcousticBrainz,
+): React.AbstractComponent<{recording: RecordingT}, void>);
--- a/root/static/scripts/common/components/sidebar/AcousticBrainz.js
+++ b/root/static/scripts/common/components/sidebar/AcousticBrainz.js
@@ -64,7 +64,7 @@ const SidebarAcousticBrainz = ({recording}: {recording: RecordingT}) => {
   return count ? (
     <>
       <h2>
-        {l('Acoustic information')}
+        {l('Acoustic analysis')}
       </h2>
 
       <a className="external" href={`//acousticbrainz.org/${recording.gid}`}>
--- a/root/static/scripts/common/components/sidebar/AcousticBrainz.js
+++ b/root/static/scripts/common/components/sidebar/AcousticBrainz.js
@@ -56,6 +56,10 @@ const SidebarAcousticBrainz = ({recording}: {recording: RecordingT}) => {
     return key.includes('#') ? KEY_LABELS[key] : key;
   }
 
+  function roundedBPM(data) {
+    return Math.round(data.tonal.key_strength * 100) / 100;
+  }
+
   React.useEffect(fetchAcousticBrainzCount, [recording.gid]);
   return count ? (
     <>
@@ -64,12 +68,7 @@ const SidebarAcousticBrainz = ({recording}: {recording: RecordingT}) => {
       </h2>
 
       <a className="external" href={`//acousticbrainz.org/${recording.gid}`}>
-        {texp.ln(
-          '{count} submission on AcousticBrainz',
-          '{count} submissions on AcousticBrainz',
-          count,
-          {count},
-        )}
+        {l('AcousticBrainz entry')}
       </a>
 
       {data === null ? null : (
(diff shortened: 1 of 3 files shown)


loujine:
I didn't notice the problem, but my local instance is already very slow (several seconds to load a page), I will have to check.

loujine:
If I compare master/this branch on my laptop, a recording page loads in ~5 seconds in both cases, so I can't detect a difference.

The behavior should be the same as the FingerprintTable: I can see in firefox developer tools that the requests to AB are done after the rest (same time as the favicon) and takes <100ms.
If AB is down the fetch request fails immediately, and I don't think there's anything I want to catch() since I don't display anything if no data is found. If we want to have a "No data found in AB" displayed instead, I can prepare an error message. I don't think anything was caught for the FingerprintTable either

root/layout/components/sidebar/SidebarAcousticBrainz.js
```diff
@@ -0,0 +1,81 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import React, {useEffect, useState} from 'react';
```
reosarevok, line 10: Please use `import * as React` (I'm changing the place you copied this one from in https://github.com/metabrainz/musicbrainz-server/pull/1580/files) for this one :)

root/layout/components/sidebar/SidebarAcousticBrainz.js
```diff
@@ -0,0 +1,81 @@
...
+
+import hydrate from '../../../utility/hydrate';
+
+import {SidebarProperty, SidebarProperties} from './SidebarProperties';
+
+const SidebarAcousticBrainz = ({recording}: {recording: RecordingT}) => {
+  const [count, setCount] = React.useState(0);
+  const [isCountLoaded, setIsCountLoaded] = React.useState(false);
+  const [data, setData] = React.useState({});
+  const [isDataLoaded, setIsDataLoaded] = React.useState(false);
```
mwiencek, line 20: Might make sense to just make the default value for `data` `null` - then you can just check `data != null` to determine if it's loaded rather than relying on `isDataLoaded`.

root/layout/components/sidebar/SidebarAcousticBrainz.js
```diff
@@ -0,0 +1,81 @@
...
+
+  React.useEffect(fetchAcousticBrainzCount, [recording.gid]);
+  return (isCountLoaded && count) ? (
+    <>
+      <h2>
+        {l('Acoustic information')}
+      </h2>
+
+      <a className="external" href={`//acousticbrainz.org/${recording.gid}`}>
+        {l(`${count} submissions on AcousticBrainz`)}
```
mwiencek, line 57: To handle plural forms for numeric values we'd use `ln` instead of `l` here, and interpolate the count this way
```suggestion
        {texp.ln(
          '{count} submission on AcousticBrainz',
          '{count} submissions on AcousticBrainz',
          {count},
        )}
```

root/layout/components/sidebar/SidebarAcousticBrainz.js
```diff
@@ -0,0 +1,81 @@
+/*
```
mwiencek, line 1: One more thing is that we'll need to move this file to somewhere under root/static/scripts/common/components/, perhaps root/static/scripts/common/components/sidebar/SidebarAcousticBrainz.js. This is mentioned briefly in HACKING.md, but basically scripts that run on the client (even if they're also used on the server) generally have to go under root/static/scripts/ so that translated strings work. The reason is that when we bundle translated strings as .js for the client, we want to only include strings actually used by client scripts to minimize the bundle size, as there are a lot of strings in server-side components that we don't need in the browser. So, we only extract client strings from root/static/scripts/.

root/layout/components/sidebar/SidebarAcousticBrainz.js
loujine: Good point, fixed

root/layout/components/sidebar/SidebarAcousticBrainz.js
loujine: I moved the file to root/static/scripts/common/components/sidebar/
And I should definitely read again HACKING.md, that's on me
Thanks for the explanation

loujine:
@mwiencek : is there a way to install this branch on a test server and test the performance impact?

yvanzo:
@loujine: just by asking :genie: https://test.musicbrainz.org/recording/cdba2c78-2645-4286-a225-013f2735d415

loujine:
Rebased, fixed the tests, and the performance on test.mb.o looks good. So unless we want to change the threshold to display/hide the tonality key info, I think this is ready for production

alastair:
on my list for this week 👍

root/static/scripts/common/components/sidebar/AcousticBrainz.js
```diff
@@ -0,0 +1,82 @@
...
+        setCount(data.count);
+        if (data.count) {
+          fetchAcousticBrainzData();
+        }
+      },
+    );
+  }
+
+  function fetchAcousticBrainzData() {
+    const dataUrl = `//acousticbrainz.org/api/v1/${recording.gid}/low-level`;
```
alastair, line 35: you should use our low-data endpoint: https://acousticbrainz.readthedocs.io/api.html#get--api-v1-low-level
use `?features=tonal.key_key;tonal.key_scale`. I see that we don't support key_strength. I added it in https://github.com/metabrainz/acousticbrainz-server/pull/384

root/static/scripts/common/components/sidebar/AcousticBrainz.js
```diff
@@ -0,0 +1,82 @@
...
+          {count},
+        )}
+      </a>
+
+      {data === null ? null : (
+        <SidebarProperties>
+
+          {data.tonal.key_strength > 0.5 ? (
+            <SidebarProperty className="acousticbrainz_key" label={l('Key:')}>
+              {`${data.tonal.key_key} ${data.tonal.key_scale}`}
```
alastair, line 66: just one thing to keep in mind here - we use the sharp variations of all keys, so Bb major will be reported as A# major. This means that some people might complain that it's "wrong". We could provide mappings here, so we always report "A#/Bb major" in these cases, for example.

I'm not sure what the best strength threshold is. I was reading [the paper that describes the algorithm](https://www.researchgate.net/publication/220669391_Tonal_Description_of_Polyphonic_Audio_for_Music_Content_Processing) but it doesn't show how the accuracy increases as the strength threshold is altered. I'll discuss this with my coworkers. For now 0.5 is probably a "safe" bet.

Do we want a way for people to complain? Or at the very least a popup that says "this is automatically generated and might be wrong"?

alastair:
I collected the strengths for about 150,000 of the items in the AB database: 
```
0.0: 0 (0.0%)
0.1: 48 (0.03%)
0.2: 535 (0.35%)
0.3: 2689 (1.74%)
0.4: 9618 (6.21%)
0.5: 22158 (14.3%)
0.6: 39251 (25.33%)
0.7: 46709 (30.15%)
0.8: 31339 (20.23%)
0.9: 2596 (1.68%)
```
e.g. 0.2 means all strengths from 0.2 to 0.3. Let's keep 0.5 as the threshold for now, but I wonder if 0.6 or 0.7 may be better to be more sure. The original paper gives an accuracy of only ~66% for getting the key right. I'm not sure exactly what the accuracy of the current version of the code is.

root/static/scripts/common/components/sidebar/AcousticBrainz.js
loujine: Didn't know about that, thanks

loujine:
Regarding key strength/confidence, I also thought about comparing all submissions and make sure they all found the same key. I had one example with strength > 0.5 and different keys for some submissions, I'll try to find it again

alastair:
Yeah, the different strengths (or key values) may be from users submitting incorrect recordings for a specific MBID. until we merge the similarity work (https://github.com/metabrainz/acousticbrainz-server/pull/353) we don't really have a way of deciding which of many different submissions is the best. This is coming.

dbogdanov:
Checking the distribution of key strength values, the `>0.5` is a reasonable start to get rid of the outliers. The estimation algorithm always finds the best match, even though the music may be atonal. 

So, it is really good to put it straight and mention in the UI that this is an automatic value to avoid any misunderstandings.

root/static/scripts/common/components/sidebar/AcousticBrainz.js
```diff
@@ -32,16 +32,31 @@ const SidebarAcousticBrainz = ({recording}: {recording: RecordingT}) => {
...
       },
     );
   }
 
+  function keyLabel(key) {
+    if (!key.includes('#')) {
+      return key;
+    }
+    return {
+      'A#': 'A♯/B♭',
```
alastair, line 52: 👏

root/static/scripts/common/components/sidebar/AcousticBrainz.js
```diff
@@ -0,0 +1,97 @@
...
+      'C#': 'C♯/D♭',
+      'D#': 'D♯/E♭',
+      'F#': 'F♯/G♭',
+      'G#': 'G♯/A♭',
```
mwiencek, line 56: I'd probably move this to a `const KEY_LABELS` or something after the imports, just to avoid the extra allocation on each call.

loujine: Sure, I'll fix that

loujine:
Spent some time thinking about a better way to display the fact that the AB info is generated, and in the end I used <abbr> with the text "Automatic suggestion from entry #1/{count}". The <abbr> makes it obvious (I hope) that there's an explanation without adding too much text on the page itself. The key strength is also mentioned in the abbr title.

Also I stopped displaying the count (as "{count} submissions on AB") since I don't think the information is that relevant

root/static/scripts/common/components/sidebar/AcousticBrainz.js
```diff
@@ -0,0 +1,107 @@
...
+
+  function roundedBPM(data) {
+    return Math.round(data.tonal.key_strength * 100) / 100;
+  }
+
+  React.useEffect(fetchAcousticBrainzCount, [recording.gid]);
+  return count ? (
+    <>
+      <h2>
+        {l('Acoustic information')}
```
yvanzo, line 67: > Spent some time thinking about a better way to display the fact that the AB info is generated

@loujine, maybe just retitle the section with _analysis_:

```suggestion
        {l('Acoustic analysis')}
```

yvanzo:
Thank you @loujine for the changes, @alastair and @dbogdanov for the comments.

loujine:
Thanks everyone for the feedback!

yvanzo:
For the record, it has been amended through commit https://github.com/metabrainz/musicbrainz-server/commit/b18e3402cf2efc97ef52e5881bd6345913026516 to get string placeholders fit our localization system.

PR #1583, opened 2020-07-01
MBS-10926: add 'copy date' button on relationship editor dialog

loujine:
# Problem

MBS-10926: Add "paste date" button in relationship editor dialogs

# Solution
Minimal solution to keep the UI simple:
 * only the copy begin → end behavior
 * only on the relationship editor

The userscript still exists for more complex behavior.
I guess this part will be converted to React later, so it might be easy to apply somewhere else.

# Checklist for author
Looks ready to me

### DIFF
--- a/root/components/relationship-editor.tt
+++ b/root/components/relationship-editor.tt
@@ -125,6 +125,9 @@
       <input type="text" data-bind="value: date.year, valueUpdate: 'afterkeydown'" maxlength="4" placeholder="[% l('YYYY') %]" size="4" />-<!--
       --><input type="text" data-bind="value: date.month, valueUpdate: 'afterkeydown'" maxlength="2" placeholder="[% l('MM') %]" size="2" />-<!--
       --><input type="text" data-bind="value: date.day, valueUpdate: 'afterkeydown'" maxlength="2" placeholder="[% l('DD') %]" size="2" />
+      <!-- ko if: isBeginDate -->
+        <button class="icon copy-date" type="button" title="[% l('Copy to end date') | html %]" data-bind="click: $parent.copyDate"></button>
+      <!-- /ko -->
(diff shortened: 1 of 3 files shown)


loujine:
I will try to find a clearer icon.
I thought also about using some kind of chain/link icon (like https://daviesmediadesign.com/wp-content/uploads/2019/05/Final-Image-Resolution-Resize-Image-in-GIMP-Tutorial.jpg) but it doesn't really correspond to the button behavior

root/components/relationship-editor.tt
```diff
@@ -125,6 +125,9 @@
       <input type="text" data-bind="value: date.year, valueUpdate: 'afterkeydown'" maxlength="4" placeholder="[% l('YYYY') %]" size="4" />-<!--
       --><input type="text" data-bind="value: date.month, valueUpdate: 'afterkeydown'" maxlength="2" placeholder="[% l('MM') %]" size="2" />-<!--
       --><input type="text" data-bind="value: date.day, valueUpdate: 'afterkeydown'" maxlength="2" placeholder="[% l('DD') %]" size="2" />
+      <!-- ko if: isBeginDate -->
+        <button class="icon copy-date" type="button" title="Copy to end date" data-bind="click: $parent.copyDate"></button>
```
mwiencek, line 129: Let's localize the title here too:
```suggestion
        <button class="icon copy-date" type="button" title="[% l('Copy to end date') | html %]" data-bind="click: $parent.copyDate"></button>
```

loujine:
* added localization
* changed icon to the down arrow already in images/icons/down.png
* tried to improve the commit message :)

@yvanzo: for me having the icon on the begin date line makes more sense, since I use my Tab key to jump between begin fields then tab/enter to copy to the end dates. But I don't have a strong feeling for UX in general, so I'll change it if you think it's better

PR #1584, opened 2020-07-03
MBS-10927: Check whether entity existed before setting allowNew

reosarevok:
### Fix MBS-10927

Add Relationship edits were always assuming allowNew should be set, but in reality this should only be set where the entity didn't ever exist (usually in previews) rather than every time the entity was missing.

While in most cases I could think of either both entities should have allowNew or neither should, I made a separate allowNew for entity0 and entity1, just in case.

### DIFF
--- a/root/edit/details/AddRelationship.js
+++ b/root/edit/details/AddRelationship.js
@@ -36,7 +36,13 @@ const AddRelationship = ({edit}: Props): React.MixedElement => {
     <table className="details add-relationship">
       <tr>
         <th>{l('Relationship:')}</th>
-        <td><Relationship allowNew relationship={relationship} /></td>
+        <td>
+          <Relationship
+            allowNewEntity0={!edit.display_data.entity0?.id}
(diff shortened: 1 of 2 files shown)


root/edit/details/AddRelationship.js
```diff
@@ -36,7 +38,13 @@ const AddRelationship = ({edit}: Props): React.MixedElement => {
     <table className="details add-relationship">
       <tr>
         <th>{l('Relationship:')}</th>
-        <td><Relationship allowNew relationship={relationship} /></td>
+        <td>
+          <Relationship
+            allowNewEntity0={edit.display_data.entity0_is_new}
+            allowNewEntity1={edit.display_data.entity1_is_new}
```
mwiencek, line 44: You should be able to to check `edit.data.entity0.id` etc. directly here, I think, which would make adding `entityX_is_new` to the display data unnecessary.

reosarevok: ... That would be sensible, yes. Done!

PR #1585, opened 2020-07-03
MBS-10923: add unit test in (unsuccesful) attempt to reproduce.

ijc:
# Problem

I am seeing issues with inconsistent results listing the releases in one of my collections (described in MBS-10923). I tried to reproduce by adding a unit test but I was not successful. Posting it as a PR in case it is useful/desirable but please feel free to close if it is not.

/cc @reosarevok 

# Solution

I added some collections with a few more releases in to the corpus (the existing ones only had at most one release in them) and then added a test which repeatedly looks up a set of releases in the hopes of seeing the issue.

Sadly this didn't reproduce the problem for me.

`ẁebservice.sql` says `Generated by ../../script/webservice_test_data.pl` which exists however it references `release-group-sql-dump.pl` which doesn't exist -- therefore I just edited `webservice.sql` by hand (I'm sure that can't be right, if the PR is wanted and someone points me to the script I'm happy to rework).

# Checklist for author

* [x] Delete irrelevant parts of this template


ijc:
Obsoleted by #1591

PR #1586, opened 2020-07-05
MBS-2768: Add acoustID info in the recording merge table

loujine:
# Problem
MBS-2768

# Solution
Use React hook to fetch the acoustIDs on the fly from the merge page.

# Checklist for author
Works on my local instance (see screenshot on MBS-2768)

# Action
As for MBS-10916 the main question is to check on a test server that the performance is good

### DIFF
--- a/root/utility/tableColumns.js
+++ b/root/utility/tableColumns.js
@@ -503,6 +503,57 @@ export const isrcsColumn:
     accessor: 'isrcs',
   };
 
+const AcoustidCell = ({
+  value,
+}: {value: string}): React.Element<typeof React.Fragment> => {
+  const url = '//api.acoustid.org/v2/track/list_by_mbid' +
+    `?format=json&disabled=1&mbid=${value}`;
+  const [tracks, setTracks] = React.useState([]);
+  const [isLoaded, setIsLoaded] = React.useState(false);
+  React.useEffect(() => {
+    fetch(url).then(
+      resp => resp.json(),
+    ).then(data => {
+      setTracks(data.tracks);
+      setIsLoaded(true);
+    });
+  }, [url]);
+
+  return (
+    <React.Fragment>
+      {
+        isLoaded ? (
+          tracks.length ? (
+            <ul>
+              {tracks.map((track) => (
+                <li key={track.id}>
+                  <code>
+                    <a
+                      className={'external' +
+                        (track.disabled ? ' disabled-acoustid' : '')}
+                      href={`//acoustid.org/track/${track.id}`}
+                    >
+                      {track.id.slice(0, 6) + '…'}
+                    </a>
+                  </code>
+                </li>
+              ))}
+            </ul>
+          ) : null
+        ) : <p className="loading-message">{l('Loading...')}</p>
+      }
+    </React.Fragment>
+  );
+};
+
+export const acoustidColumn:
+  ColumnOptions<{+gid?: string, ...}, string> = {
+    Cell: ({cell: {value}}) => (<AcoustidCell value={value} />),
+    Header: N_l('AcoustIDs'),
+    accessor: 'gid',
+    id: 'acoustid',
+  };
+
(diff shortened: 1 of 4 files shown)


loujine:
@yvanzo : I'd be interested to see that PR on test.mb.o too (but no hurry)

yvanzo (approved):
Thank you @loujine! I pushed a `master`-rebased version of it to the branch `test` (with build [3765](https://app.circleci.com/pipelines/github/metabrainz/musicbrainz-server/3765/workflows/05cf3a7f-cc9f-495f-ba33-879888fcfab9) successful on CircleCI) and deployed to <https://test.musicbrainz.org/>.

loujine:
Rebased to fix conflict

root/utility/tableColumns.js
```diff
@@ -503,6 +503,49 @@ export const isrcsColumn:
...
+      const [tracks, setTracks] = React.useState([]);
+      const [isLoaded, setIsLoaded] = React.useState(false);
+      React.useEffect(() => {
+        fetch(url).then(
+          resp => resp.json(),
+        ).then(data => {
+          setTracks(data.tracks);
+          setIsLoaded(true);
+        });
+      }, [url]);
```
mwiencek, line 520: This should work now because it's run in the context of the table component and the column is always there, but I'm afraid it'll violate the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) if we make columns hideable at some point or if react-table changes how they render stuff.

We should be able to easily fix this though, just by moving all the `Cell` contents to an `AcoustidCell` component defined above the column here. Then just return `<AcoustidCell value={value} />` from `Cell` here.

loujine: Done. Please check the types I used, tests pass but I'm not sure I used the right way

mwiencek:
The `AcoustidCell` change looks good. My only concern now is that the Acoustid docs say not to make more than [3 requests per second](https://acoustid.org/webservice), but that could be hit on any merge that involves >3 recordings.

[/v2/track/list_by_mbid](https://acoustid.org/webservice#list_by_mbid) supports batching multiple MBIDs into a single request. Do you think it'd be easy to request them all in advance, in RecordingMerge.js? I guess that'd mean you'd be hydrating the `RecordingMerge` component instead, and passing an object like `{[recordingMbid]: Array<acoustid>}` to `RecordingList` that it can read the Acoustids from.

But if you don't want to spend more time on this we can merge it as-is and I can look into it later!

loujine:
> My only concern now is that the Acoustid docs say not to make more than 3 requests per second
I hadn't thought about that

> But if you don't want to spend more time on this we can merge it as-is and I can look into it later!
I would prefer to merge as-is, since I'm not sure to have much time to look at this in the near future.

I still have a branch about the same kind of feature for AcousticBrainz info (MBS-9487), so that would be a good candidate to improve this solution and move it to the RecordingMerge component and solve the API limiting rate/payload optimization at the same time

mwiencek (approved):
Sounds good, I'm happy to look into batching the calls after we merge this too

mwiencek:
Sorry for the delay. Merged this in https://github.com/metabrainz/musicbrainz-server/commit/53fc7e3c18ff7fd4231cb1461ee58f01f29c0f2e and committed https://github.com/metabrainz/musicbrainz-server/commit/c12af240f781d5d572aa0d5c07ac8d5dcf48b670 for the batching. I retested everything and it seems to work well. Thanks again! Closing this since it won't automatically.

jesus2099:
Hi @loujine,

Apparently, once AcoustIDs are loaded, React redraws, not only the AcoustIDs cells (`<td>`), but also all the other cells (Name, Artist, ISRCs, Length).
Which breaks user scripts using those cells ([collection highlighter](https://github.com/jesus2099/konami-command/issues/517), for instance).

Is there a way to limit React redraws to only AcoustIDs cells?

PR #1588, opened 2020-07-06
MBS-10931: Avoid crashing on missing ID

reosarevok:
### Fix MBS-10931

Some historic relationship edits seem to be missing the ID for some of the related entities, but we expect an ID. Rather than making the ID optional, it seems sensible to do // 0 like we already do elsewhere for similar situations.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Historic/Relationship.pm
+++ b/lib/MusicBrainz/Server/Edit/Historic/Relationship.pm
@@ -216,8 +216,8 @@ sub _display_relationships {
             my $entity1_type = $_->{entity1_type};
             my $model0 = type_to_model( $_->{entity0_type} );
             my $model1 = type_to_model( $_->{entity1_type} );
-            my $entity0_id = $_->{entity0_id};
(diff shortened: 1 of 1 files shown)


mwiencek:
This looks like a duplicate of https://github.com/metabrainz/musicbrainz-server/pull/1543 unless I'm missing something. :)

Technically the `id` is already optional (you can simply not pass it to the constructor), you just can't pass undef explicitly. Like in https://github.com/metabrainz/musicbrainz-server/pull/1543#discussion_r451021861 I don't like using 0 as a placeholder: if we actually need the id somewhere, it's very unexpected to have to do `if (id !== 0)` anywhere instead of `if (id != null)` or just `if (id)`, and two of those can also break if Perl encodes it as `"0"` in the JSON for some reason.

So I'd rather just optionally pass it to the constructor if it's defined like at https://github.com/metabrainz/musicbrainz-server/blob/4d3b777/lib/MusicBrainz/Server/Edit/Relationship/Edit.pm#L175

reosarevok:
Duh.

PR #1589, opened 2020-07-06
MBS-10930: Fix loading relationship entities for removed reltypes

reosarevok:
### Fix MBS-10930

Removed reltypes (such as "microblog") no longer have info on which entity types they expected as type0 and type1. Luckily, we have source_type and target_type on relationships which serve the same purpose and can be used as a replacement.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/Create.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/Create.pm
@@ -230,6 +230,8 @@ sub build_display_data
             grep { !exists $loaded->{LinkAttributeType}{$_->{type}{id}} }
                 @{ $self->data->{attributes} // [] }
         )),
+        source_type => $type0,
+        target_type => $type1,
         entity0 => defined $entity0->id ? undef : $entity0,
         entity1 => defined $entity1->id ? undef : $entity1,
     }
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Edit/Relationship/Create.pm
```diff
@@ -230,6 +231,8 @@ sub build_display_data
             grep { !exists $loaded->{LinkAttributeType}{$_->{type}{id}} }
                 @{ $self->data->{attributes} // [] }
         )),
+        source_type => defined $link_type->entity0_type ? undef : $type0,
+        target_type => defined $link_type->entity1_type ? undef : $type1,
```
mwiencek, line 235: I think it shouldn't hurt to always output the types here.

root/static/scripts/common/components/Relationship.js
```diff
@@ -68,15 +68,17 @@ const RelationshipContent = ({
   const linkType = linkedEntities.link_type[relationship.linkTypeID];
   let entity0 = relationship.entity0;
   let entity1 = relationship.entity1;
+  const type0 = relationship.source_type || linkType.type0;
+  const type1 = relationship.target_type || linkType.type1;
```
mwiencek, line 72: I see in `Edit::Relationship::Create` `source_type` and `target_type` are hard coded to `type0` & `type1` respectively, but in `Entity::Relationship` it depends on the direction of the relationship. So I'd think this should be something like
```suggestion
  const type0 = linkType.type0 ||
    (direction === 'backward' ? relationship.target_type : relationship.source_type); 
  const type1 = linkType.type1 ||
    (direction === 'backward' ? relationship.source_type : relationship.target_type); 
```

PR #1595, opened 2020-07-13
MBS-10939, MBS-10940: Speed up Data::Release::find_by_artist for VA, allow filtering by date/country

mwiencek:
In addition to those two tickets, I added a commit to cache the number of hits from `find_by_artist` and `find_by_label` for 30 minutes so that we don't have to run `count(*)` over the entire set of results on every page load.


mwiencek:
This was cherry-picked to prod & deployed today by @reosarevok 

> This seems to work fine. My only question would be whether we should mention on the VA releases tab something like "This is a huge list, and it is unordered for performance reasons. Feel free to use the filters above to reduce it to a more manageable size", to avoid people reporting the bug that the sort is broken.

This sounds fine by me. Or we could wait to see if anyone actually uses this page and reports it. :)

PR #1598, opened 2020-07-17
MBS-10965: Convert Add Disc ID edit to React

reosarevok:
### Implement MBS-10965

The original TT code checked `!edit.preview OR edit.display_data.medium` to decide whether to display the Medium section of the edit data. That means that it would always be displayed if there's a medium though, and if there isn't then there's nothing to display anyway, so I changed it to just a medium check.


mwiencek:
Seems to work for me in the release editor. I tested that by adding `<input name="mediums.0.toc" type="text" value="1 3 57895 150 18402 35080"/>` to root/static/tests/seed-love-bug.html. @yvanzo do you have any more info about the bug you encountered there?

The only problem is that the release hover text says "This entity has been removed, ..." when adding a release. You may just need to add support for `allowNew` as in https://github.com/metabrainz/musicbrainz-server/pull/1530.

reosarevok:
Made the change (adding it to MediumLink too, since it's the release that needs allowNew)

root/static/tests/seed-love-bug.html
```diff
@@ -57,7 +57,7 @@
         <input name="urls.0.url" type="text" value="https://www.amazon.co.jp/Love-Bug-M-Flo-Loves-Boa/dp/B0001FAD2O/ref=sr_1_1?ie=UTF8&amp;qid=1466612275&amp;sr=8-1&amp;keywords=4988064451180"/>
         <input name="urls.0.link_type" type="text" value="77"/><br />
       </fieldset>
-
+      <input name="mediums.0.toc" type="text" value="1 3 57895 150 18402 35080"/>
```
mwiencek, line 60: Good idea. :+1:

yvanzo:
@mwiencek: By clicking on “Add a new release”
from <http://localhost:5000/cdtoc/move?toc=147687&filter-release.query=Be>
the following error shows up:

> The data you’ve seeded contained the following errors:
>    Invalid mediums.0.toc: “147687”.

It isn’t specific to this pull request.

reosarevok:
That does seem like an invalid toc though, where did you get the link from?

yvanzo:
https://musicbrainz.org/cdtoc/9IJicubE2sms4mejeMo5KUlXMN8-

PR #1602, opened 2020-07-20
MBS-10972: Convert Add Instrument edit to React

reosarevok:
### Implement MBS-10972

This displays the type translated (as elsewhere), but the description (which we translate elsewhere) is left in English since the translations are not connected with the edits at all.


root/edit/details/AddInstrument.js
```diff
@@ -0,0 +1,82 @@
...
+              entity={display.instrument}
+            />
+          </td>
+        </tr>
+      </table>
+
+      <table className="details add-instrument">
+        <tr>
+          <th>{addColonText(l('Name'))}</th>
+          <td>{display.name}</td>
```
yvanzo, line 54: It would look as a localization bug in non-English UI. It would be nice to specify it is intentionally untranslated (using hard-coded tag `lang="en"`) and to allow further localizing its display (using quotes or anything appropriate for the UI locale).

```suggestion
          <th>{addColonText(l('Name'))}</th>
          <td>{untranslated(display.name)}</td>
```

with something like:

```javascript
function untranslated(variable: Expand2ReactInput): Expand2ReactOutput {
  return exp.lp(
    '{untranslated_text}',
    {untranslated_text: <span lang="en">{variable}</span>},
  );
}
```

I did not test it.

mwiencek: Hm, well we do display the localized name in the "Instrument:" link right above this, at least, and all other entities display untranslated names at present too. I'm not sure about a hard-coded `lang="en"` since not all instruments are in English or have English names. Though I'd be curious to hear from other translators if `{untranslated_text}` would be useful to them. @mihkel-t maybe? :) I would proceed with this in a separate PR if necessary.

mihkel-t: @mwiencek: I think "{untranslated_text}" on its own will only confuse translators. I'd say it needs a comment for translators, something like "Don't change the {untranslated_text} placeholder, only add e.g. quotes around it if appropriate in your language."

reosarevok: Should we leave it as is for now and add a ticket like "Mark intentionally untranslated strings"?

mihkel-t: FWIW, I'd support adding the comment to this one right away, while it's still remembered... But if there are a whole bunch of such intentionally untranslated things, then probably a good idea to have a ticket to track them.

reosarevok: IIRC we don't have a way to set comments now though?

mwiencek: Would we eventually want this for all entity names, or just entity names where we have a translation but aren't using it? (If we start displaying aliases in place of entity names, that distinction might become less useful.) Just wondering because we display raw entity names in (to make an understatement) a lot of places.

And would it cause a problem if we used `{untranslated_text}` where the entity name happens to be in the UI language, so isn't actually "untranslated?"

reosarevok: I kinda feel we should just have a note or warning that a field is intentionally left untranslated. Maybe an icon of some sort? https://github.com/metabrainz/musicbrainz-server/pull/1636 for example has untranslated text on edits that would look weird styled, and I agree that we shouldn't assume the text is always in English to begin with since things like instrument names are very international.

yvanzo: > not all instruments are in English or have English names

Correct, many are not. We don’t have any information about the language here. Let’s forget about `lang` attribute.

> Would we eventually want this for all entity names, or just entity names where we have a translation but aren't using it? 

The latter, here we do intentionally display raw name because translations were not available at edit time.

> And would it cause a problem if we used `{untranslated_text}` where the entity name happens to be in the UI language, so isn't actually "untranslated?"

Probably not. Even with quotes or whatever around it, it would still be readable to users from that language.

> I kinda feel we should just have a note or warning that a field is intentionally left untranslated. Maybe an icon of some sort? 1636 for example has untranslated text on edits that would look weird styled

But since the only use case seems to be when creating entities, maybe we can specify it in header rather than data:
```suggestion
          <th>{addColonText(l('Name (raw)'))}</th>
          <td>{display.name}</td>
```

reosarevok: I think "raw" is not too clear. What about this?

Screenshot from 2020-08-31 15-51-14

yvanzo: Seems alright or just “Field value shown untranslated as entered in this edit”

reosarevok:
@brainzbot, retest this

reosarevok:

PR #1605, opened 2020-07-20
MBS-10975: Convert Remove ISWC edit to React

reosarevok:
### Implement MBS-10975


lib/MusicBrainz/Server/Edit/Work/RemoveISWC.pm
```diff
@@ -61,7 +62,11 @@ sub build_display_data {
         ISWC->new(
             iswc => $self->data->{iswc}{iswc},
             work => $loaded->{Work}{ $self->data->{work}{id} } //
-                    Work->new( name => $self->data->{work}{name} ),
+                    Work->new(
+                        id => $self->data->{work}{id},
+                        name => $self->data->{work}{name}
+                    ),
+            work_id => $self->data->{work}{id},
```
yvanzo, line 69: Is it needed to have work id twice?

reosarevok: work_id is to load it from linkedEntities later. The work having an ID is because ID is not optional in EntityRoleT, so I always try to output it when at all possible.

PR #1611, opened 2020-07-24
MBS-10986: Convert edit error templates to React

reosarevok:
### Implement MBS-10986


root/edit/CannotApproveEdit.js
```diff
@@ -0,0 +1,56 @@
...
+  {$c, edit}: Props,
+): React.Element<typeof Layout> => {
+  const editDisplay = 'edit #' + edit.id;
+  const editLink = <EditLink content={editDisplay} edit={edit} />;
+  const editIsClosed = !edit.is_open;
+  const editIsCancelled = edit.status === EDIT_STATUS_DELETED ||
+    edit.status === EDIT_STATUS_TOBEDELETED;
+  const reason = editIsCancelled ? (
+    l('The edit has been cancelled.')
+  ) : editIsClosed ? (
```
yvanzo, line 34: The distiction between these two statuses still exist in [`root/utility/edit.js`](https://github.com/metabrainz/musicbrainz-server/blob/v-2020-08-10/root/utility/edit.js#L85-L88).

reosarevok: It does, but is that a problem? While on the edit itself showing `This edit was recently cancelled` vs `This edit was cancelled` might make more of a difference, for an error message like this one it seems irrelevant.

Are you proposing to just get rid of the distinction between these two statuses completely? I wouldn't mind that, unless `EDIT_STATUS_TOBEDELETED` is important for debug purposes (i.e. to figure out an edit is stuck).

yvanzo: I’m not proposing to get rid of the distinction, quite the opposite.

> I guess it was meant for the rare case where the edit was open but being cancelled, but the string doesn't particularly match that. I changed it to show it for actual cancelled edits too, since it can't hurt.

As you mentioned it was intended for some rare case, so maybe improving that string would be more helpful in such case?

reosarevok: I mean, it is still useful for the (I expect much more common, even if still very rare) case where an edit gets cancelled while the approving editor has it open. If you think we should have a third case for EDIT_STATUS_TOBEDELETED, I guess we can, it just seems like the difference is irrelevant to the approving editor.

yvanzo: It seems to require further investigation and to be out of the scope of this ticket, whereas keeping the distinction as-is for now would be straightforward.

reosarevok: Ok, I added a separate line for that in the unlikely chance that someone hits it. Guess it cannot hurt.

root/edit/CannotApproveEdit.js
```diff
@@ -0,0 +1,58 @@
...
+): React.Element<typeof Layout> => {
+  const editDisplay = 'edit #' + edit.id;
+  const editLink = <EditLink content={editDisplay} edit={edit} />;
+  const editIsClosed = !edit.is_open;
+  const editIsCancelled = edit.status === EDIT_STATUS_DELETED;
+  const editBeingCancelled = edit.status === EDIT_STATUS_TOBEDELETED;
+  const reason = editIsCancelled ? (
+    l('The edit has been cancelled.')
+  ) : editBeingCancelled ? (
+    l('The edit is already in the process of being cancelled.')
```
yvanzo, line 35: Thanks, maybe [the message for the same case in `root/utility/edit.js`](https://github.com/metabrainz/musicbrainz-server/blob/v-2020-08-10/root/utility/edit.js#L86) should be changed at the same time?

reosarevok: I guess we might. Made that change in a separate commit.

mwiencek: The `TOBEDELETED` status is unused. That was from before MBS-777 when edits weren't cancelled immediately. So I think these messages should be removed instead of modified.

reosarevok:
Dropped anything involving TOBEDELETED from this PR. Opened MBS-11044 for the rest of the removal.

PR #1613, opened 2020-07-25
MBS-10910: Display renamed from/into on label overview

reosarevok:
### Implement MBS-10910

This seems like a sensible way to follow the flow of labels that were renamed, and I see no negatives about showing them there.
I saw artist legal name is treated as only ever one result (which seems weird to me, but that's a separate issue). Because a label could be listed as renamed from or into multiple others, I pass these as arrays. I expect in most cases these would be data errors, but having them be visible makes them more likely to get fixed anyway.


reosarevok:
> It shows the latest previous name and the first next name only 

IMO that's correct - only those were renamed to and from this label. Any others were renamed to and from the other labels.

yvanzo:
This relationship seems to be transitive, so it seems it would be correct to show all the other names too. In either case, dates apply.

reosarevok:
Wouldn't the dates be just a duplicate of the label start and end dates?

lib/MusicBrainz/Server/Controller/Label.pm
```diff
@@ -117,11 +119,29 @@ sub show : PathPart('') Chained('load')
     $c->model('MediumFormat')->load(map { $_->all_mediums } @$releases);
     $c->model('ReleaseLabel')->load(@$releases);
  
+    my @renamed_from = map { $_->target }
+                        grep { $_->direction == $MusicBrainz::Server::Entity::Relationship::DIRECTION_BACKWARD }
+                        grep { $_->link->type->gid eq 'e6159066-6013-4d09-a2f8-bc473f21e89e' } @{ $label->relationships };
```
mwiencek, line 124: suggesting to store the first grep in a var, then use it for both `@renamed_from` and `@renamed_to`.
```suggestion
    my @rename_rels = grep { $_->link->type->gid eq 'e6159066-6013-4d09-a2f8-bc473f21e89e' } @{ $label->relationships };
    my @renamed_from = map { $_->target }
                        grep { $_->direction == $MusicBrainz::Server::Entity::Relationship::DIRECTION_BACKWARD }
                        @rename_rels;
```

mwiencek (approved):

root/label/LabelIndex.js
```diff
@@ -51,6 +59,20 @@ const LabelIndex = ({
...
     />
+    {renamedFrom.length ? (
+      <RelatedEntitiesDisplay title={l('Renamed from')}>
+        {commaOnlyList(renamedFrom.map(
+          label => <DescriptiveLink entity={label} key={label.gid} />,
+        ))}
+      </RelatedEntitiesDisplay>
+    ) : null}
+    {renamedInto.length ? (
+      <RelatedEntitiesDisplay title={l('Renamed into')}>
```
mwiencek, line 70: This doesn't seem wrong but "Renamed to" sounds a bit more common.
```suggestion
      <RelatedEntitiesDisplay title={l('Renamed to')}>
```

root/label/LabelIndex.js
```diff
@@ -51,6 +59,20 @@ const LabelIndex = ({
       entity={label}
       numberOfRevisions={numberOfRevisions}
     />
+    {renamedFrom.length ? (
+      <RelatedEntitiesDisplay title={l('Renamed from')}>
```
mwiencek, line 63: This one sounds slightly strange to me though without both parts, as in "renamed from x to y." I'd suggest "Previously named" or "Previously known as." Do you think either of those would work?

```suggestion
      <RelatedEntitiesDisplay title={l('Previously named')}>
```

mwiencek (commented):
I rebased and pushed some minor changes here to use only one `load_subset` call -- not that it's common for both directions to exist, but might as well.

But looking at it again, I'm wondering if we can improve the wording slightly.  Let me know what you think.

root/label/LabelIndex.js
reosarevok: I like "Previously known as" a bit more so using that.

mwiencek:
Thanks!  There hasn't been any follow-up about the dates in almost a year so I think it's best to merge this, as it's an improvement; it can certainly be improved further later on if people ask for dates.

PR #1615, opened 2020-07-26
MBS-10990: Limit edit displays to 50 edits per page 

reosarevok:
### Implement MBS-10990

MBS-8412 standardised all lists to 100 items, including edit lists (which used to be 25 items, except for the EditListing role, which didn't set a limit so it used the standard 100 item limit).

Some users were overwhelmed by 100 edits everywhere, so this sets all edit lists to 50 items per list, including in EditListing.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/EditListing.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/EditListing.pm
@@ -61,7 +61,7 @@ sub _list {
 
     my $type   = model_to_type( $self->{model} );
     my $entity = $c->stash->{ $self->{entity_name} };
-    my $edits  = $self->_load_paged($c, $find->($type, $entity));
+    my $edits  = $self->_load_paged($c, $find->($type, $entity), limit => 50);
 
     $c->stash(
         edits => $edits, # stash early in case an ISE occurs
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -23,6 +23,10 @@ with 'MusicBrainz::Server::Controller::Role::Load' => {
     entity_name => 'edit',
 };
 
+__PACKAGE__->config(
+    paging_limit => 50,
+);
+
 =head1 NAME
 
 MusicBrainz::Server::Controller::Moderation - handle user interaction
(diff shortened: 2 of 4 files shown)


yvanzo:
IMHO, the problem is not to satisfy some users over some others, it’s more that some lists should be shorter than others:
* Open edit lists should be rather small (25 or even 10) as it is mainly intended for voters who must take a decision for each edit.
* Other edit lists should be rather large (100 or more) as it is mainly intended to allow skimming history for old data and changes.

lib/MusicBrainz/Server/Controller/Role/EditListing.pm
```diff
@@ -4,6 +4,10 @@ use Moose::Role -traits => 'MooseX::MethodAttributes::Role::Meta::Role';
 use MusicBrainz::Server::Data::Utils qw( model_to_type load_everything_for_edits );
 use MusicBrainz::Server::Constants qw( :edit_status );
 
+MusicBrainz::Server::Controller->config(
+    paging_limit => 50,
+);
```
mwiencek, line 9: This looks like it sets it "globally" (for all controllers) rather than on the specific package implementing this role.

Can we just pass `limit => 50` to the relevant `_load_paged` calls?

mwiencek:
> IMHO, the problem is not to satisfy some users over some others, it’s more that some lists should be shorter than others:
> 
> * Open edit lists should be rather small (25 or even 10) as it is mainly intended for voters who must take a decision for each edit.
> * Other edit lists should be rather large (100 or more) as it is mainly intended to allow skimming history for old data and changes.

I don't disagree with the logic here, though I'm not sure there's an easy way to introspect whether the list contains only open edits & variably change the paging size based on that at the moment. And I think 50 is a better compromise than what we have now.

reosarevok:
> Open edit lists should be rather small (25 or even 10) as it is mainly intended for voters who must take a decision for each edit.

I do kinda disagree with that because very often our edits are better reviewed as a group. I would agree with it if we implemented edit grouping in the future though :)

lib/MusicBrainz/Server/Controller/Role/EditListing.pm
reosarevok: I guess that would be reasonable, yeah. Done.

PR #1616, opened 2020-07-27
Function to generate Flow types from edit data

mwiencek:
A SQL function, `edit_data_type_info()`, is added that generates "type info" for a single column of edit data.

A script, script/generate_edit_data_flow_type.js, is added which takes an edit type, `selects edit_data_type_info()` for every row of edit data for that type, and outputs a single merged Flow type representing the edit data overall for that type.

For example:

```sh
./script/generate_edit_data_flow_type.js --edit-type 92
```

Generates a Flow type for `EDIT_RELATIONSHIP_DELETE`.

The `PROD_STANDBY` database must be configured in DBDefs.pm to use this utility.

The underlying code used to build the Flow type may also be useful to generate Flow types for other JSON data, e.g. entities.json, so I've added script/generate_json_flow_type.js for this more general purpose too. Here's an example invoking that script:

```sh
cat entities.json | jq -c | script/generate_json_flow_type.js
```

(This expects the JSON object on a single line, so `jq -c` is used to compact the data first.)

I've committed the type generated for root/edit/details/RemoveRelationship.js as an example.

I think this will help with two things:
1. The SQL function by itself is useful to see all the different data formats an edit type has, which can make it easier to identify places we can clean up the data later.
2. The combined Flow type generated by edit_data_flow_type.js is useful for accessing the edit data safely (type-checked by Flow) in a React template. This isn't so important now, but in the future we'll likely want to move `build_display_data` etc. into JS.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -156,6 +156,33 @@ done with the following command:
 
     $ prove -l t/js.t
 
+### Flow
+
+Our JavaScript uses [Flow](https://flow.org/) for static type checking.
+If you're adding a new JS file, it should have a `@flow strict-local`
+header at the top to enable type checking.
+
+To ensure all types are correct after making a change, run Flow:
+
+    $ ./node_modules/.bin/flow
+
+Global type declarations available to all files are found in
+[root/types.js](root/types.js) and [root/vars.js](root/vars.js).
+The latter is for functions and variables that are auto-imported by
+Webpack's [ProvidePlugin](webpack/providePluginConfig.js).
+
+We have a couple of scripts you may find useful that generate Flow object
+types based on JSON data:
+
+ * `./script/generate_edit_data_flow_type.js --edit-type $EDIT_TYPE_ID`
+   will generate an object type to represent the edit data of `$EDIT_TYPE_ID`.
+   However, this requires having a `PROD_STANDBY` database configured in
+   DBDefs.pm, as it uses production data to ensure a correct type.
+
+ * `cat $JSON | ./script/generate_json_flow_type.js` will generate an object
+   type for a stream of JSON objects (which each must be on a single line).
+   Passing a single object is fine, too.
+
 ### Selenium
 
 We have a set of browser-automated UI tests running with Selenium WebDriver.
--- a/package.json
+++ b/package.json
@@ -48,6 +48,8 @@
     "less-plugin-clean-css": "1.5.1",
     "leven": "2.0.0",
     "mutate-cow": "4.0.3",
+    "pg": "8.3.3",
+    "pg-cursor": "2.3.3",
     "po2json": "0.4.1",
     "querystring": "0.2.0",
     "react": "16.13.1",
--- a/admin/sql/CreateFunctions.sql
+++ b/admin/sql/CreateFunctions.sql
@@ -1445,4 +1445,39 @@ CREATE OR REPLACE FUNCTION mb_simple_tsvector(input text) RETURNS tsvector AS $$
   SELECT to_tsvector('musicbrainz.mb_simple', musicbrainz.mb_lower(input));
 $$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE STRICT;
 
+-----------------------------------------------------------------------
+-- Edit data helpers
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION edit_data_type_info(data JSONB) RETURNS TEXT AS $$
+BEGIN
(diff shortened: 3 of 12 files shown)


mwiencek:
Note that we can use script/edit_data_flow_type.js to generate types for other JSON too, e.g. entities.json. I'd like to do that in another branch once this is merged. :)

reosarevok (approved):
LGTM but I didn't test

yvanzo (approved):
Successfully tested on production database with “Reorder relationships” edits (type ID = 99).

Would it make sense to have a wrapper script that uses `PROD_STANDBY` from `DBDefs`?

That would be nice to mention the command generating the added flow type in second commit too.

mwiencek:
> Would it make sense to have a wrapper script that uses PROD_STANDBY from DBDefs?

Added a wrapper script for that (and updated the commit message/PR description with instructions). It sets `statement_timeout = 0` so you shouldn't get timeouts anymore.

Also added `script/generate_json_flow_type.js` to generate a Flow type from other JSON.

package.json
```diff
@@ -48,6 +48,8 @@
     "less-plugin-clean-css": "1.5.1",
     "leven": "2.0.0",
     "mutate-cow": "4.0.3",
+    "pg": "8.3.3",
+    "pg-cursor": "2.3.3",
```
mwiencek, line 52: I'm also using these in my json-dumps rework.

yvanzo (changes requested):
```bash
./script/generate_edit_data_flow_type.js --edit-type 92
```
returned:

```node
/musicbrainz-server/root/utility/generateFlowType.js:37
      (isEditDataTypeInfo ?? parent.isEditDataTypeInfo);
                           ^

SyntaxError: Unexpected token ?
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/musicbrainz-server/script/generate_edit_data_flow_type.js:23:28)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
```

using Node `v10.19.0` currently installed in MusicBrainz Docker.

yvanzo (approved):
Works fine now. Code looks good too. Maybe document scripts also in `HACKING.md`?

mwiencek:
> Maybe document scripts also in HACKING.md?

I wasn't sure where to put that so I made a general Flow section. Have a quick look. :)

HACKING.md
```diff
@@ -156,6 +156,33 @@ done with the following command:
 
     $ prove -l t/js.t
 
+### Flow
+
+Our JavaScript uses [Flow](https://flow.org/) for static type checking.
+If you're adding a new JS file, it should have a `@flow` header at the top
```
yvanzo, line 162:
```suggestion
If you're adding a new JS file, it should have a `@flow strict-local` header at the top
```

HACKING.md
```diff
@@ -156,6 +156,33 @@ done with the following command:
...
+types based on JSON data:
+
+ * `./script/generate_edit_data_flow_type.js --edit-type $EDIT_TYPE_ID`
+   will generate an object type to represent the edit data of `$EDIT_TYPE_ID`.
+   However, this requires having a `PROD_STANDBY` database configured in
+   DBDefs.pm, as it uses production data to ensure a correct type.
+
+ * `cat $JSON | ./script/generate_json_flow_type.js` will generate an object
+   type for a stream of JSON objects (which each must be on a single line).
+   Passing a single object is fine, too.
```
yvanzo, line 184: Not sure how useful that will be though?

yvanzo (approved):
Each time I approve this pull request, there are new additions, so I approve it again!

HACKING.md
mwiencek: It seems more generally useful to me than the one for edit data. You commonly might want to generate a type for some request data (say, JSON from a web service endpoint like AB/CB/AcoustID). Whereas the edit data one is really only useful for core developers writing edit code.

mwiencek:
OK, I'll commit the `strict-local` addition and merge this to save you from another approval. :)

PR #1618, opened 2020-07-29
MBS-10994: Convert user/message.tt to React

reosarevok:
### Implement MBS-10994

This reuses the existing StatusPage component, which works perfectly for the use case (but should maybe be renamed by now).

email_sent.tt was effectively the same page but with the messages defined on the TT instead of the Perl side. I moved them to the Perl side so it can also use UserMessage.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User.pm
+++ b/lib/MusicBrainz/Server/Controller/User.pm
@@ -243,10 +243,13 @@ sub _check_for_confirmed_email {
 
     unless ($c->user->has_confirmed_email_address) {
         $c->stash(
-            title    => l('Send Email'),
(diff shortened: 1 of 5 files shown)


reosarevok:
It's not so much about saving code, but ensuring some consistency :) Shouldn't hurt, I hope, at least!

yvanzo:
Is “consistency” having `UserMessage` to use `StatusPage`? It looks like abusing “consistency” term.

reosarevok:
I mean, "StatusPage" has nothing to do with any status. It's used for all sort of messages, such `LostUsernameSent`, and we even have `const NotFound = StatusPage`. Maybe we should just rename it to NotificationPage or something, just not sure what'd be a good name.

PR #1619, opened 2020-07-29
MBS-10995 / MBS-10996 / MBS-11277 / MBS-11279: Convert Other Lookups pages and search index to React

reosarevok:
### Implement MBS-10995 / MBS-10996 / MBS-11277 / MBS-11279

Cherry-picked 1b21cb3f5c997fea93bec1c462e077dff329dd99 from https://github.com/metabrainz/musicbrainz-server/pull/1158/ as part of MBS-11277.


lib/MusicBrainz/Server/Form/OtherLookup.pm
```diff
@@ -28,19 +28,19 @@ has_field 'iswc'     => (
     type => '+MusicBrainz::Server::Form::Field::ISWC',
 );
 
-has_field 'artist-ipi'     => (
+has_field 'artist_ipi'     => (
```
yvanzo, line 31: That breaks all existing other lookup URLs with such parameter.

reosarevok: Fair point. Dunno how many we have but I guess I'll just do field['artist-ipi'] and cry a bit :D

reosarevok:
Expanded this to finish converting all of the otherlookups templates (a fair amount could just be removed or replaced, so there's very little display code left after all of this that is exclusive to otherlookups)

root/search/SearchIndex.js
```diff
@@ -0,0 +1,46 @@
...
+import type {TagLookupFormT} from '../taglookup/types';
+
+import SearchForm from './components/SearchForm';
+
+type Props = {
+  +$c: CatalystContextT,
+  +otherLookupForm: OtherLookupFormT,
+  +searchForm: SearchFormT,
+  +tagLookupForm: TagLookupFormT,
+
```
mwiencek, line 25:
```suggestion
```

mwiencek (approved):

root/components/list/ReleaseList.js
```diff
@@ -136,21 +159,27 @@ const ReleaseList = ({
...
         ...(instrumentUsageColumn ? [instrumentUsageColumn] : []),
+        ...(typeColumn ? [typeColumn] : []),
+        ...(statusColumn ? [statusColumn] : []),
         ...($c.session?.tport == null ? [] : [taggerColumn]),
         ...(showRatings ? [ratingsColumn] : []),
       ];
     },
     [
-      $c.session,
+      $c.session?.tport,
```
yvanzo, line 171: This change seems to be unrelated, it is not documented in commit https://github.com/metabrainz/musicbrainz-server/pull/1619/commits/1cfccb902a3b9fba2793a40e70a3fd08ff6d7a16 at least.

reosarevok: That's just part of the expected elements in that array as per eslint, I guess I can split it into its own commit if needed.

yvanzo: Why wouldn’t it depend on `$c.session?.merger`?

reosarevok: Because $c.session?.tport is the only conditional in the table return AFAICT. I don't see $c.session?.merger anywhere in the file actually? :)

yvanzo: Ok, it’s from https://github.com/metabrainz/musicbrainz-server/commit/1cfccb902a3b9fba2793a40e70a3fd08ff6d7a16#diff-3070f82643a68f5cadc699007b7fea0d20a5b989184430f6b71af9dc88f95e3cL140-R166 :)
Then why was it set to `$c.session` before?

reosarevok: No idea! :D I mean, it should work fine if it changes every time $c.session changes, but I'm not sure why that was set like that.

yvanzo: Ok, thanks :) It would be better fixed in a separate commit if possible.

yvanzo (approved):
:ship:

PR #1621, opened 2020-07-31
Install bzip2 in the test-database docker image

alastair:
This is needed for importing dumps. It was present in the postgres 9.5 base image but isn't present in the postgres 12 image.


docker/templates/Dockerfile.test-database.m4
```diff
@@ -6,7 +6,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 # install_extensions.sh removes certain build dependencies that we need, so we
 # can't install everything here.
 # Note: curl is also a dependency of carton.
-RUN apt_install(`ca-certificates curl sudo')
+RUN apt_install(`ca-certificates curl sudo bzip2')
```
yvanzo, line 9: Just keep alphabetical ordering:

```suggestion
RUN apt_install(`bzip2 ca-certificates curl sudo')
```

PR #1624, opened 2020-08-02
MBS-10999: Make adding first IPI/ISNI an auto-edit

reosarevok:
### Implement MBS-10999

We generally allow auto-edits for everyone if data is being added where it didn't earlier exist. For some reason though we didn't make that check for IPIs and ISNIs. It makes sense to make the edit go into voting for cases where an IPI or ISNI already exists, but otherwise we should act in the same way as we do for other data.

We check whether an ISNI / IPI is already in use elsewhere, and block autoediting in that case. Whether the edit was otherwise autoeditable or not, ModBot will leave a note specifying that the IPI or ISNI is already in use in x artists / labels, and linking to a search to find out which (actually building the list for display on the edit turned out to be almost impossible with localized_note, so this is the best compromise I could find)

### DIFF
--- a/lib/MusicBrainz/Server/Data/Role/IPI.pm
+++ b/lib/MusicBrainz/Server/Data/Role/IPI.pm
@@ -21,6 +21,31 @@ role
         value_type          => 'ipi',
     };
 
+    method find_reused_ipis => sub {
+        my ($self, @ipis) = @_;
+        my $query = "SELECT 'artist' AS entity_type, artist_ipi.ipi, COUNT(*) AS count
+                    FROM artist
+                        JOIN artist_ipi ON artist.id = artist_ipi.artist
+                    WHERE artist_ipi.ipi = any(?)
+                    GROUP BY artist_ipi.ipi
+                    UNION ALL
+                    SELECT 'label' AS entity_type, label_ipi.ipi, COUNT(*) AS count
+                    FROM label
+                        JOIN label_ipi ON label.id = label_ipi.label
+                    WHERE label_ipi.ipi = any(?)
+                    GROUP BY label_ipi.ipi";
+        my $results = $self->sql->select_list_of_hashes($query, \@ipis, \@ipis);
+        my %reused_ipis;
+        for my $result (@$results) {
+            my $ipi = $result->{ipi};
+            my $entity_count = $result->{count};
+            my $entity_type = $result->{entity_type};
+            $reused_ipis{$ipi}{$entity_type} = $entity_count;
+        }
+
+        return \%reused_ipis;
+    };
+
 };
 
 no Moose::Role;
--- a/lib/MusicBrainz/Server/Data/Role/ISNI.pm
+++ b/lib/MusicBrainz/Server/Data/Role/ISNI.pm
@@ -21,6 +21,31 @@ role
         value_type          => 'isni',
     };
 
+    method find_reused_isnis => sub {
+        my ($self, @isnis) = @_;
+        my $query = "SELECT 'artist' AS entity_type, artist_isni.isni, COUNT(*) AS count
+                    FROM artist
+                        JOIN artist_isni ON artist.id = artist_isni.artist
+                    WHERE artist_isni.isni = any(?)
+                    GROUP BY artist_isni.isni
+                    UNION ALL
+                    SELECT 'label' AS entity_type, label_isni.isni, COUNT(*) AS count
+                    FROM label
+                        JOIN label_isni ON label.id = label_isni.label
+                    WHERE label_isni.isni = any(?)
+                    GROUP BY label_isni.isni";
+        my $results = $self->sql->select_list_of_hashes($query, \@isnis, \@isnis);
+        my %reused_isnis;
+        for my $result (@$results) {
+            my $isni = $result->{isni};
+            my $entity_count = $result->{count};
+            my $entity_type = $result->{entity_type};
+            $reused_isnis{$isni}{$entity_type} = $entity_count;
+        }
+
+        return \%reused_isnis;
+    };
+
 };
 
 no Moose::Role;
--- a/lib/MusicBrainz/Server/Edit/Artist/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Artist/Edit.pm
@@ -193,9 +193,33 @@ around allow_auto_edit => sub {
     return 0 if exists $self->data->{old}{end_area_id} &&
         defined($self->data->{old}{end_area_id}) && $self->data->{old}{end_area_id} != 0;
 
-    return 0 if $self->data->{new}{ipi_codes};
+    if (defined $self->data->{new}{ipi_codes}) {
+        # If there's already IPIs for the artist, not an autoedit
+        if (@{ $self->data->{old}{ipi_codes} // [] }) {
+            return 0;
+        }
 
-    return 0 if $self->data->{new}{isni_codes};
+        # If there's already an entity with any of the IPIs, not an autoedit
+        my $reused_ipis = $self->reused_ipis;
+
+        if (%$reused_ipis) {
+            return 0;
+        }
+    }
+        
+    if (defined $self->data->{new}{isni_codes}) {
+        # If there's already ISNIs for the artist, not an autoedit
+        if (@{ $self->data->{old}{isni_codes} // [] }) {
+            return 0;
+        }
+
+        # If there's already an entity with any of the ISNIs, not an autoedit
+        my $reused_isnis = $self->reused_isnis;
+
+        if (%$reused_isnis) {
+            return 0;
+        }
+    }
 
     return $self->$orig(@args);
 };
--- a/lib/MusicBrainz/Server/Edit/Label/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Label/Edit.pm
@@ -171,9 +171,33 @@ around allow_auto_edit => sub {
     # Don't allow an autoedit if the area changed
     return 0 if defined $self->data->{old}{area_id};
 
-    return 0 if $self->data->{new}{ipi_codes};
+    if (defined $self->data->{new}{ipi_codes}) {
+        # If there's already IPIs for the label, not an autoedit
+        if (@{ $self->data->{old}{ipi_codes} // [] }) {
+            return 0;
+        }
 
-    return 0 if $self->data->{new}{isni_codes};
+        # If there's already an entity with any of the IPIs, not an autoedit
+        my $reused_ipis = $self->reused_ipis;
+
+        if (%$reused_ipis) {
+            return 0;
+        }
+    }
+        
+    if (defined $self->data->{new}{isni_codes}) {
+        # If there's already ISNIs for the label, not an autoedit
+        if (@{ $self->data->{old}{isni_codes} // [] }) {
+            return 0;
+        }
+
+        # If there's already an entity with any of the ISNIs, not an autoedit
+        my $reused_isnis = $self->reused_isnis;
+
+        if (%$reused_isnis) {
+            return 0;
+        }
+    }
 
     return $self->$orig(@args);
 };
--- a/lib/MusicBrainz/Server/Edit/Role/IPI.pm
+++ b/lib/MusicBrainz/Server/Edit/Role/IPI.pm
@@ -1,6 +1,17 @@
 package MusicBrainz::Server::Edit::Role::IPI;
 use 5.10.0;
 use Moose::Role;
+use utf8;
+
+use MusicBrainz::Server::Constants qw( $EDITOR_MODBOT );
+use MusicBrainz::Server::Data::Utils qw( localized_note type_to_model );
+use MusicBrainz::Server::Translation qw( N_l );
+use Set::Scalar;
+
+has 'reused_ipis' => (
+    isa => 'HashRef',
+    is  => 'rw'
+);
 
 with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
     prop_name => 'ipi_codes',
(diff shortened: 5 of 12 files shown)


reosarevok:
ISNIs are probably not going to be unique, if they do them properly (since they should identify a person rather than a persona). Of course, I know at least some cases where they do split personas, for... reasons (such as Eminem). I wouldn't make too many assumptions, in any case.

yvanzo:
> any non-unique ones should be forced as a non-autoedit, then that might be good

:+1: 

> but it would be kinda confusing too (there wouldn't be any indicator to the user about _why_ that happened, right?

Then just add an edit note by ModBot when that happens.

reosarevok:
I changed this to also force voting if the IPI or ISNI is already in use. We *could* add Data::IPI and a sub check_if_used and check both tables at once, I just felt this also was interesting in allowing us to find any other artists or labels using it and maybe printing them on the edit note. @mwiencek: how would I make ModBot add a note? I guess in allow_auto_edit I still don't know the ID for the edit, so I can't do it from here. I imagine one option would be to stash a boolean or add it to the edit data and then check that when entering the edit?

mwiencek:
> I guess in allow_auto_edit I still don't know the ID for the edit, so I can't do it from here. I imagine one option would be to stash a boolean or add it to the edit data and then check that when entering the edit?

Yes, but I don't think you need to put this in the edit data, since you only need that information when the edit is created. I'd add some Moose attributes like `reused_ipis` and `reused_isnis` to `Edit::Role::IPI` and `Edit::Role::ISNI`. In each of those roles I'd also add `after initialize => sub {}` where you set the attributes, and `after post_insert => sub {}` where you post the ModBot note.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -222,6 +222,28 @@ sub find_by_work
     $self->query_to_list_limited($query, [($work_id) x 2], $limit, $offset);
 }
 
+sub find_by_ipi
```
mwiencek, line 225: These should ideally go in `Data::Role::IPI` and `Data::Role::ISNI` respectively. And it might be better if they were plural & accepted multiple codes so that we don't have to execute SQL queries inside a loop.

reosarevok: Hmm. Would you expect it to be just "is_ipi_in_use" or whatnot and only return a boolean (per code, or even in total) then? It would seem weird to have it return artists / labels from the role IMO but maybe not?

mwiencek: Doesn't seem that weird to me. :) Why do you think so? We could do `is_ipi_in_use` etc. instead but I figured you were going to use the existing artists / labels for display. If not then that works.

reosarevok:
Well, finally made it so that this left useful notes. It was *a lot* more complicated and frustrating than I would have thought, and I ended up having to compromise on what to display, but it's certainly better than nothing IMO!

lib/MusicBrainz/Server/Edit/Label/Edit.pm
```diff
@@ -171,9 +171,40 @@ around allow_auto_edit => sub {
...
+        
+    if (defined $self->data->{new}{isni_codes}) {
+        # If there's already ISNIs for the label, not an autoedit
+        if (defined $self->data->{old}{isni_codes} &&
+            scalar @{$self->data->{old}{isni_codes}} != 0) {
+                return 0;
+        }
+
+        # If there's already an entity with any of the ISNIs, not an autoedit
+        # If there's already an entity with any of the ISNIs, not an autoedit
```
yvanzo, line 199: dupe

lib/MusicBrainz/Server/Edit/Role/IPI.pm
```diff
@@ -12,6 +23,62 @@ with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
...
+                    {
+                        artist_count => $artist_dupe_count,
+                        artist_search => "/search?query=ipi%3A$ipi&advanced=1&type=artist",
+                        ipi => $ipi,
+                        label_count => $label_dupe_count,
+                        label_search => "/search?query=ipi%3A$ipi&advanced=1&type=label",
+                    }
+                );
+            } elsif ($artist_dupe_count) {
+                $edit_note = localized_note(N_l('The IPI {ipi} is already in use on artists ({artist_count} at the time of this note). You can {artist_search|search for uses of this IPI}.'),
```
yvanzo, line 55: Using `N_ln` would allow for a more fluent phrase with “in use on another artist” and “in use on {n} other artists”.

I would suggest to make the second phrase more pressing, e.g. “Please check {artist_search|uses of this IPI}.”

lib/MusicBrainz/Server/Edit/Artist/Edit.pm
```diff
@@ -193,9 +193,39 @@ around allow_auto_edit => sub {
...
+        }
+
+        # If there's already an entity with any of the ISNIs, not an autoedit
+        my $reused_isnis = $self->c->model('Artist')->find_reused_isnis(
+            $self->data->{new}{isni_codes}
+        );
+
+        if (%$reused_isnis) {
+            return 0;
+        }
```
yvanzo, line 227: Something is wrong with this test. I tested entering an existing ISNI for another artist. Even though the ModBot added a note, the edit still passed through as an auto-edit. It seems worth adding to `t/`.

yvanzo (changes requested):

lib/MusicBrainz/Server/Edit/Role/IPI.pm
reosarevok: localized_note doesn't support N_ln, that's why this is done that way.

lib/MusicBrainz/Server/Edit/Artist/Edit.pm
reosarevok: Did you add it as an autoeditor?

yvanzo: Probably but I dropped the test DB since then. I just tested again and it isn’t applied when not having autoeditor privs.

lib/MusicBrainz/Server/Edit/Role/IPI.pm
yvanzo: Adding support seems feasible and worth it since more edit notes are being coded. Reported it under MBS-11171. That should not stop this PR though.

yvanzo (approved):

lib/MusicBrainz/Server/Data/Role/IPI.pm
```diff
@@ -21,6 +21,31 @@ role
...
     };
 
+    method find_reused_ipis => sub {
+        my ($self, @ipis) = @_;
+        my $query = "SELECT 'artist' AS entity_type, artist_ipi.ipi, COUNT(*) AS count
+                    FROM artist
+                        JOIN artist_ipi ON artist.id = artist_ipi.artist
+                    WHERE artist_ipi.ipi = any(?)
+                    GROUP BY artist_ipi.ipi
+                    UNION
```
mwiencek, line 31:
```suggestion
                    UNION ALL
```
UNION ALL should be used if uniqueness isn't an issue, 'cause UNION is slower (and there shouldn't be duplicates here).

lib/MusicBrainz/Server/Data/Role/ISNI.pm
```diff
@@ -21,6 +21,31 @@ role
...
     };
 
+    method find_reused_isnis => sub {
+        my ($self, @isnis) = @_;
+        my $query = "SELECT 'artist' AS entity_type, artist_isni.isni, COUNT(*) AS count
+                    FROM artist
+                        JOIN artist_isni ON artist.id = artist_isni.artist
+                    WHERE artist_isni.isni = any(?)
+                    GROUP BY artist_isni.isni
+                    UNION
```
mwiencek, line 31:
```suggestion
                    UNION ALL
```

lib/MusicBrainz/Server/Edit/Artist/Edit.pm
```diff
@@ -193,9 +193,39 @@ around allow_auto_edit => sub {
     return 0 if exists $self->data->{old}{end_area_id} &&
         defined($self->data->{old}{end_area_id}) && $self->data->{old}{end_area_id} != 0;
 
-    return 0 if $self->data->{new}{ipi_codes};
+    if (defined $self->data->{new}{ipi_codes}) {
+        # If there's already IPIs for the artist, not an autoedit
+        if (defined $self->data->{old}{ipi_codes} &&
+            scalar @{$self->data->{old}{ipi_codes}} != 0) {
```
mwiencek, line 199:
```suggestion
        if (@{ $self->data->{old}{ipi_codes} // [] }) {
```
is a bit more perlish

lib/MusicBrainz/Server/Edit/Artist/Edit.pm
```diff
@@ -193,9 +193,39 @@ around allow_auto_edit => sub {
...
+                return 0;
+        }
+
+        # If there's already an entity with any of the IPIs, not an autoedit
+        my $reused_ipis = $self->c->model('Artist')->find_reused_ipis(
+            $self->data->{new}{ipi_codes}
+        );
```
mwiencek, line 206: this runs after `initialize` so these should already be set, same below and in Label::Edit
```suggestion
        my $reused_ipis = $self->reused_ipis;
```

lib/MusicBrainz/Server/Edit/Role/IPI.pm
```diff
@@ -12,6 +23,62 @@ with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
     extract_value => sub { shift->ipi }
 };
 
+after initialize => sub {
+    my ($self, %opts) = @_;
+
+    my $old_ipis = $self->data->{old}{ipi_codes};
+    my $new_ipis = $self->data->{new}{ipi_codes};
```
mwiencek, line 30: Aren't these potentially `undef`? De-referencing them below will break if so.

lib/MusicBrainz/Server/Edit/Role/IPI.pm
```diff
@@ -12,6 +23,62 @@ with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
...
+    my $added_ipis_set = Set::Scalar->new(@$new_ipis) - Set::Scalar->new(@$old_ipis);
+    my @added_ipis = $added_ipis_set->members;
+    $self->reused_ipis($self->c->model($self->_edit_model)->find_reused_ipis(@added_ipis));
+};
+
+after post_insert => sub {
+    my $self = shift;
+
+    if ($self->reused_ipis) {
+        for my $ipi (keys %{ $self->reused_ipis }) {
```
mwiencek, line 40: That will always be true, but you could do
```suggestion
    my %reused_ipis = %{$self->reused_ipis};
    if (%reused_ipis) {
        for my $ipi (keys %reused_ipis) {
```
or probably just remove the `if` since the for loop will already be a no-op if `keys %reused_ipis` is empty

lib/MusicBrainz/Server/Edit/Role/IPI.pm
```diff
@@ -12,6 +23,62 @@ with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
...
+            my $artist_dupe_count = $self->reused_ipis->{$ipi}->{artist};
+            my $label_dupe_count = $self->reused_ipis->{$ipi}->{label};
+            my $edit_note;
+            if ($artist_dupe_count && $label_dupe_count) {
+                 $edit_note = localized_note(N_l('The IPI {ipi} is already in use on both artists and labels ({artist_count} and {label_count} at the time of this note respectively). You can search for uses of this IPI on {artist_search|artists} and {label_search|labels}.'),
```
mwiencek, line 45: Wouldn't it be exceptionally rare for both artists and labels to share an IPI or ISNI being added again? Not sure we need a specific message for this instead of just leaving two notes.

lib/MusicBrainz/Server/Edit/Role/IPI.pm
mwiencek: I'd probably rather just add support for `ln` here, since these strings sound kind of awkward reading them again. I can help with that.

mwiencek (changes requested):

lib/MusicBrainz/Server/Edit/Role/IPI.pm
reosarevok: Yes, it would, but I thought it would be more clear like this. I guess we could just leave two notes too.

lib/MusicBrainz/Server/Edit/Role/IPI.pm
reosarevok: Maybe? What should be done, ` // []`?

mwiencek: Yes, `// []` (otherwise the line below will throw `Can't use an undefined value as an ARRAY reference`).

lib/MusicBrainz/Server/Edit/Role/ISNI.pm
```diff
@@ -12,6 +23,62 @@ with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
     extract_value => sub { shift->isni }
 };
 
+after initialize => sub {
+    my ($self, %opts) = @_;
+
+    my $old_isnis = $self->data->{old}{isni_codes};
+    my $new_isnis = $self->data->{new}{isni_codes};
```
mwiencek, line 30: here too
```suggestion
    my $old_isnis = $self->data->{old}{isni_codes} // [];
    my $new_isnis = $self->data->{new}{isni_codes} // [];
```

lib/MusicBrainz/Server/Edit/Role/ISNI.pm
```diff
@@ -12,6 +23,62 @@ with 'MusicBrainz::Server::Edit::Role::ValueSet' => {
...
+    my $added_isnis_set = Set::Scalar->new(@$new_isnis) - Set::Scalar->new(@$old_isnis);
+    my @added_isnis = $added_isnis_set->members;
+    $self->reused_isnis($self->c->model($self->_edit_model)->find_reused_isnis(@added_isnis));
+};
+
+after post_insert => sub {
+    my $self = shift;
+
+    if ($self->reused_isnis) {
+        for my $isni (keys %{ $self->reused_isnis }) {
```
mwiencek, line 40:
```suggestion
    for my $isni (keys %{ $self->reused_isnis }) {
```

lib/MusicBrainz/Server/Edit/Role/IPI.pm
mwiencek: Dunno, it just seemed kinda weird to have a string for a situation that might happen once a year, but I could be underestimating people. :)

lib/MusicBrainz/Server/Edit/Role/IPI.pm
reosarevok: I understood the ln support was more complex than expected?

reosarevok:
Changed this to use ln now that this is supported thanks to @mwiencek's work :)

yvanzo (approved):

mwiencek (approved):

PR #1630, opened 2020-08-03
MBS-10715 / MBS-10716: Block Amazon and Apple redirect links

reosarevok:
### Implement MBS-10715 / MBS-10716 

Sadly these can't be massaged into the real link like some other ones can, so we need to just block them and ask the user to follow the redirect.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2347 for the Apple links.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -573,15 +573,18 @@ const CLEANUPS: CleanupEntries = {
     },
   },
   'amazon': {
-    match: [new RegExp(
-      '^(https?://)?' +
-      '(((?!music)[^/])+\.)?' +
-      '(amazon\\.(' + (
-        'ae|at|com\\.au|com\\.br|ca|cn|com|de|es|fr|in' +
-        '|it|jp|co\\.jp|com\\.mx|nl|se|sg|com\\.tr|co\\.uk'
-      ) + ')|amzn\\.com)',
-      'i',
-    )],
+    match: [
+      new RegExp(
+        '^(https?://)?' +
+        '(((?!music)[^/])+\.)?' +
+        '(amazon\\.(' + (
+          'ae|at|com\\.au|com\\.br|ca|cn|com|de|es|fr|in' +
+          '|it|jp|co\\.jp|com\\.mx|nl|se|sg|com\\.tr|co\\.uk'
+        ) + ')|amzn\\.com)',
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -635,6 +658,23 @@ const CLEANUPS = {
...
+                href: url,
+                rel: 'noopener noreferrer',
+                target: '_blank',
+              },
+            },
+          ),
+          result: false,
+        };
+      }
+
```
reosarevok, line 677: Frustratingly, given we don't autoselect a type but just clean up the URL at the moment, there's no way (AFAICT) to keep this validation in the current code (validation is only ran `if (cleanup && cleanup.type && cleanup.type[entityType])`. Should we add an id-agnostic validation option for cases like this? Or should we just special-case this one?

yvanzo: The real issue is MBS-9902. I would just leave it as-is in the meantime.

reosarevok: As is meaning "don't block Apple redirect links for now"? (AFAICT I can't block them normally without it going through validation)

yvanzo: `apple.co` redirect links can be blocked by creating a separate `'appleco'` rule ahead of `'applemusic'`.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -645,6 +645,27 @@ const CLEANUPS = {
       return {result: false};
     },
   },
+  'appleco': {
+    match: [new RegExp('^(https?://)?([^/]+\\.)?apple\\.co/', 'i')],
+    type: LINK_TYPES.downloadpurchase,
```
reosarevok, line 670: This works, but autoselecting to downloadpurchase, even if only to block the link, will mean that the user is likely to enter the redirect destination as downloadpurchase even if incorrect. So it might be better to wait with this.

root/static/scripts/edit/URLCleanup.js
reosarevok: I went back to use `applemusic` now since we can have validation there.

reosarevok:

mwiencek (approved):

yvanzo (approved):
:ship:

PR #1634, opened 2020-08-06
Update search plugin settings

jesus2099:
Problem
-------

Mozilla/Netscape search and OpenSearch files are outdated:

- Some icon URLs are broken
- **Their update URLs are broken**
- They are pointing to HTTP URLs (now we exclusively use HTTPS)
- They are using pre-NGS pages like search.html, textsearch.html and oldsearch.html

For the most part, it just relies on some redirects and works out well.
But I think it's cleaner to use canonical URLs.

Solution
--------

I replaced broken URLs by working URLs.
I also set canonical URLs instead of HTTP and pre-NGS URLs that were indirectly working after redirecting the user.

I am not sure that Firefox or another browser actually uses those non-standard OpenSearch `<Mox:UpdateUrl>` tags at all.
See https://developer.mozilla.org/fr/docs/Web/OpenSearch#Supporting_automatic_updates_for_OpenSearch_plugins
But having correct URLs cannot harm.

Action
------

1. I left all the detailed commits to ease the review but I'd like to eventually **squash** (?) them all in one commit.
  Apparently can be done by you who would [Squash and merge](https://github.blog/2016-04-01-squash-your-commits/) this PR, see [more details](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits).
2. It looks like my Firefox **does** use the **OpenSearch** (search_plugins/opensearch/) `.xml` files (they are provided in all MusicBrainz.org html page headers) and **not** the **Mozilla/Netscape** (search_plugins/firefox/) `.src` files.
  See https://developer.mozilla.org/fr/search?q=firefox%20search%20plugin and https://developer.mozilla.org/fr/search?q=mozilla%20netscape%20search
  Maybe we should remove the **search_plugins/firefox/** folder entirely (in another PR).

### DIFF
--- a/root/static/search_plugins/firefox/musicbrainzartist-old.src
+++ b/root/static/search_plugins/firefox/musicbrainzartist-old.src
@@ -11,8 +11,8 @@
  version="7.1"
  name="MusicBrainz: Artist (direct)"
  description="Direct search for an artist on MusicBrainz.org" 
- action="http://musicbrainz.org/search/oldsearch.html"
- searchForm="http://musicbrainz.org/search.html"
+ action="https://musicbrainz.org/search/oldsearch.html"
+ searchForm="https://musicbrainz.org/search"
  method="GET" >
 
 <input name="query" user="">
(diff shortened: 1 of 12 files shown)


jesus2099:
In fact I don't use them, I found them because I saw obsolete `handlearguments=1` in duckduckgo searches and wondered where they come from.

I don't know if the src are still used by anything because they don't seem linked by any page.
It seems we could remove them.

The xml files are referenced by all MBS pages.
I don't know if they are really that useful...

PR #1635, opened 2020-08-06
MBS-11017: Normalize IMSLP to HTTPS and add validation

reosarevok:
### Implement MBS-11017

IMSLP defaults to HTTPS, so we should clean up links to follow that.
This also adds basic validation: score links shouldn't include "Category", and artist ones should, since Category:Whatever links are their entries for a person and are never (AFAICT) used for a specific work or score.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1494,6 +1494,49 @@ const CLEANUPS = {
   'imslp': {
     match: [new RegExp('^(https?://)?(www\\.)?imslp\\.org/', 'i')],
     type: {...LINK_TYPES.score, ...LINK_TYPES.imslp},
+    clean: function (url) {
+      // Standardise to https
+      return url.replace(/^https?:\/\/(?:www\.)?(.*)$/, 'https://$1');
+    },
+    validate: function (url, id) {
+      switch (id) {
+        case LINK_TYPES.imslp.artist:
+          if (/^https:\/\/imslp\.org\/wiki\/Category:/.test(url)) {
+            return {result: true};
+          }
+          return {
+            error: exp.l(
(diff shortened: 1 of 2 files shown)


reosarevok:
I modified the test to check for cleanup. I'm not sure about how to further test for validation with our current test setup, since score already applies only to work and imslp to artist :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1494,6 +1494,43 @@ const CLEANUPS = {
...
+               for artists. Please link work pages to the specific
+               work in question.`,
+              {
+                category_url_pattern: (
+                  <span className="url-quote">{'Category:'}</span>
+                ),
+              },
+            ),
+            result: /^https:\/\/imslp\.org\/wiki\/Category:/.test(url),
+          };
```
yvanzo, line 1516: Successful validation should not set error message.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1494,6 +1494,43 @@ const CLEANUPS = {
...
+               for artists. Please link the specific work page to this
+               work instead, if available.`,
+              {
+                category_url_pattern: (
+                  <span className="url-quote">{'Category:'}</span>
+                ),
+              },
+            ),
+            result: /^https:\/\/imslp\.org\/wiki\/(?!Category:)/.test(url),
+          };
```
yvanzo, line 1530: Same as above.

yvanzo (changes requested):
LGTM but error message being returned even though validation has been successful.

Also, I updated ticket’s description, please update commit message/pull request too.

yvanzo (approved):
LGTM but commit message

mwiencek (approved):

PR #1636, opened 2020-08-06
MBS-11018: Convert Add Relationship Type edit to React

reosarevok:
### Implement MBS-11018

Data entered by the user with this edit is left intentionally untranslated (type name, description, documentation, link phrases).
Other data (entity type names, attribute names) is translated.

Changes:
* Add reverse and long link phrases.
* Fix bug that meant the entity types were not shown for link phrases.
* Remove the conditional for cardinality and orderable direction. Both were currently broken anyway and always showing, but that actually seems to me like a feature, not a bug.
* Fix bug that didn't hide the Attributes section if there are none.


lib/MusicBrainz/Server/Edit/Relationship/AddLinkType.pm
```diff
@@ -73,18 +74,33 @@ sub build_display_data {
...
+        name => $self->data->{name},
+        orderable_direction => $self->data->{orderable_direction},
+        reverse_link_phrase => $self->data->{reverse_link_phrase},
     }
 }
 
 sub _build_attributes {
     my ($self, $list, $loaded) = @_;
+    use Data::Dumper;
+    print Dumper($list);
```
mwiencek, line 94:
```suggestion
```

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,149 @@
...
+type Props = {
+  +edit: AddRelationshipAttributeEditT,
+};
+
+const AddRelationshipAttribute = ({edit}: Props): React.Element<'table'> => {
+  const display = edit.display_data;
+  const entity0Type = ENTITY_NAMES[display.entity0_type]();
+  const entity1Type = ENTITY_NAMES[display.entity1_type]();
+
+  console.log(display);
```
mwiencek, line 50:
```suggestion
```

reosarevok:
Not just you. Apparently some edits do not include the attribute ID at all (fun) so I just dropped the code that expected it to be there, since it's not needed for display.

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,147 @@
...
+      <tr>
+        <th>{l('Long link phrase:')}</th>
+        <td>
+          <em>{entity0Type}</em>
+          {' '}
+          {display.long_link_phrase}
+          {' '}
+          <em>{entity1Type}</em>
+        </td>
+      </tr>
```
yvanzo, line 94: It is barely readable, I would suggest to just keep `{entity0}` and `{entity1}` in link phrases and to add two fields: entity0 type and entity1 type. That would also help with interpreting cardinality fields correctly.

reosarevok: Hmm. I'm just restoring what was the intended display of this, what do you find barely readable about it? :) 
I expect for most people seeing such an edit, it's much more clear to see "artist recorded release" than "entity0 recorded entity1"...

reosarevok:
Rebased on 3e78055 to make this flow strict-local

root/edit/details/AddRelationshipType.js
yvanzo: It sometimes work in English but not always. Two counter-examples:

* http://localhost:5000/edit/53062894
  > Link phrase: _Label_ get the music _URL_

* http://localhost:5000/edit/52510291
  > Long link phrase: _Artist_ {entity1} was premiered by {entity0} _Work_

Plus, I suspect that inserting entity type names like that might not work as “well” in some other languages.

reosarevok: The fun thing here is that I'm not dropping entity0 and entity1 from the link phrase, it just works like that unless it is specifically changed :) But I guess I can just leave it as it was (showing only the phrase) and display the types separately, sure!

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,169 @@
...
+                {display.description}
+                {rawIconSection}
+              </>
+            ) : lp('(none)', 'description')}
+          {}
+        </td>
+      </tr>
+
+      <tr>
+        <th>{l('entity0 type:')}</th>
```
yvanzo, line 81: - `entity0` should not be translated as it is a placeholder;
- Not starting for a capital letter may mislead translator too;
- `:` can be saved here.

```suggestion
        <th>{addColonText(l('Type of {entity0}'))}</th>
```

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,169 @@
...
+        <td>{entity0Type}</td>
+      </tr>
+
+      <tr>
+        <th>{l('entity1 type:')}</th>
```
yvanzo, line 86: Same suggestion as above:

```suggestion
        <th>{addColonText(l('Type of {entity1}'))}</th>
```

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,169 @@
...
+      <tr>
+        <th>{l('Long link phrase:')}</th>
+        <td>
+          {display.long_link_phrase}
+          {rawIconSection}
+        </td>
+      </tr>
+
+      <tr>
+        <th>{l('entity0 cardinality:')}</th>
```
yvanzo, line 115: Same as above, even though that localization message preexisted:

```suggestion
        <th>{addColonText(l('Cardinality of {entity0}'))}</th>
```

If changed here, it should be changed in these other files too:
- `root/edit/details/edit_relationship_type.tt`
- `root/relationship/linktype/RelationshipTypeIndex.js`
- `root/relationship/linktype/RelationshipTypePairTree.js`
- `root/relationship/linktype/form.tt`

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,169 @@
...
+        <td>
+          <Cardinality cardinality={display.entity0_cardinality || 0} />
+        </td>
+      </tr>
+      <tr>
+        <th>{l('entity1 cardinality:')}</th>
```
yvanzo, line 121: The very same:

```suggestion
        <th>{addColonText(l('Cardinality of {entity1}'))}</th>
```

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,169 @@
...
+      <tr>
+        <th>{l('Documentation:')}</th>
+        <td>
+          {nonEmpty(display.documentation)
+            ? (
+              <>
+                {display.documentation}
+                {rawIconSection}
+              </>
+            ) : lp('(none)', 'documentation')}
```
yvanzo, line 162: :+1: to `lp`

root/edit/details/AddRelationshipType.js
```diff
@@ -0,0 +1,169 @@
...
+                  {attribute.max}
+                </li>
+              ))}
+            </ul>
+          </td>
+        </tr>
+      ) : null}
+
+      <tr>
+        <th>{l('Documentation:')}</th>
```
yvanzo, line 154: `:` can be saved here and both versions are localizable already:

```suggestion
        <th>{addColonText(l('Documentation'))}</th>
```

yvanzo (changes requested):

root/edit/details/AddRelationshipType.js
reosarevok: This just displays `Type of {entity0}` as the actual string, I assume that's not the intention? Same for the other {entity0/1} ones here. Making the other changes for now but without the {}.

yvanzo: To make the intention more obvious and prevent localization of the placeholder, we can even use:
```jsx
<th>{addColonText(l('Type of {entity_placeholder}', {
  entity_placeholder: '{entity0}',
}))}</th>
```

reosarevok: Huh. Ok, I didn't consider that. Honestly, to me it just feels like something broke :/

What about this?

```
<th>{addColonText(l('Type of {entity_placeholder}', {
  entity_placeholder: 'entity0',
}))}</th>
```

It should still make it so that it doesn't get translated but to users it won't look like the string broke. And it allows us to only translate each header once, which I do like.

yvanzo: It’s not _broken_, it is _intentionally left as it was originally entered (untranslated, unformatted)_ which is explained thanks to the :information_source: information icons you added. 

The full placeholder `{entity0}` is printed in the linking phrases as-is, so I feel like printing the full placeholder here makes the relationship more obvious. Maybe just adding the tag `code` would help to make it feel less broken?

reosarevok: I guess that's a bit better, I'm still not a fan but I can live with it.

Screenshot from 2020-10-14 21-15-51

yvanzo: Link phrases should keep `{entity0}` as [suggested above](
https://github.com/metabrainz/musicbrainz-server/pull/1636#discussion_r470461998) otherwise there is no point to have it in type label indeed.

reosarevok: As I mentioned as an answer to that, the phrases just don't include the placeholders in most cases. I could just add them by hand, but then it would break when they *do* include them...

yvanzo: Maybe test link phrases with a RegExp to avoid such breakage?

reosarevok: Ok, made that change, hopefully this seems ready now :)

yvanzo:
> Getting an ISE on edit 2591002 (possibly others). Just me?

It is still happening.

reosarevok:
> It is still happening.

Hmm, it wasn't happening in the meantime, so something must have broken it again. Will fix.

yvanzo (approved):

mwiencek (approved):

PR #1637, opened 2020-08-07
MBS-11020: Convert historic Add Release Annotation edit to React

reosarevok:
### Implement MBS-11020

Main change here is to use HistoricReleaseList, which shows [removed] rather than nothing at all if the releases are gone. Otherwise this mostly mirrors AddAnnotation.


root/edit/details/historic/AddReleaseAnnotation.js
```diff
@@ -0,0 +1,65 @@
+/*
+ * @flow
+ * Copyright (C) 2019 MetaBrainz Foundation
```
yvanzo, line 3:
```suggestion
 * Copyright (C) 2020 MetaBrainz Foundation
```

root/edit/details/historic/AddReleaseAnnotation.js
```diff
@@ -0,0 +1,65 @@
...
+        releases={display.releases}
+      />
+      <tr>
+        <th>{addColon(l('Text'))}</th>
+        <td>
+          {display.html
+            ? (
+              <span
+                dangerouslySetInnerHTML={{__html: display.html}}
+              />
```
mwiencek, line 43: Would it be possible to use `expand2react` here, or does that not support all of the HTML that's generated?

reosarevok: That's probably a good thing to check - we currently use dangerouslySetInnerHTML on every annotation-related section I think, but maybe expand2react is better. I'll try to remember to take a look tomorrow - if it works I guess we should change all of them?

mwiencek: Sure. Feel free to merge this PR as-is though!

PR #1640, opened 2020-08-07
Disable the full json dumps cache

mwiencek:
Disables reading from the json_dump.* tables (pulling cached JSON) when making full JSON dumps. This is slower, but will help mitigate MBS-10911 (so that data remains stale until, at most, a Wednesday or Sunday), which has been reported by multiple people.

### DIFF
--- a/lib/MusicBrainz/Script/JSONDump.pm
+++ b/lib/MusicBrainz/Script/JSONDump.pm
@@ -222,7 +222,10 @@ sub _fetch_entities_json {
     my $query;
     my %found;
     my @missing;
-    my ($force_update) = grep { $_ eq $entity_type } @{ $options{force_update} // [] };
+    # XXX Until MBS-10911 can be resolved, have the full dumps fetch
+    # all JSON anew to prevent it from becoming stale for too long.
+    #my ($force_update) = grep { $_ eq $entity_type } @{ $options{force_update} // [] };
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Script/JSONDump.pm
```diff
@@ -222,7 +222,10 @@ sub _fetch_entities_json {
     my $query;
     my %found;
     my @missing;
-    my ($force_update) = grep { $_ eq $entity_type } @{ $options{force_update} // [] };
+    # XXX Until MBS-10911 can be resolved, have the full dumps fetch
+    # all JSON anew to prevent it from becoming stale for too long.
+    #my ($force_update) = grep { $_ eq $entity_type } @{ $options{force_update} // [] };
+    my $force_update = $options{is_full_dump};
```
reosarevok, line 228: Would it make sense to still make it possible to force_update on request with options?

mwiencek: Good question. It probably wouldn't be that useful, since we don't use `--force-update` for incremental dumps at all and probably won't ever. (This patch already makes it a no-op for full dumps.) This is kind of an esoteric option that'll be removed entirely when the JSON dumps are reworked, and this stuff is only used in our production setup, so I'm fine with a temporary hack here.

PR #1642, opened 2020-08-10
MBS-11028: Convert historic Edit Release Language edit to React

reosarevok:
### Implement MBS-11028

Changes:

* Translate language/script names
* When new language/script is null, display '?' (like with old values) rather than nothing.
* When the release array is empty, show [removed] rather than nothing.


reosarevok:

PR #1647, opened 2020-08-12
MBS-11032: Convert Add medium edit to React 

reosarevok:
### Implement MBS-11032

Converts MediumTracklist too. The TT for that contains a lot of code for relationship display, meant for release index pages as per the commit that added it (d2bdd1a). Release index pages don't use this code at all anymore though; only medium edits and cdtoc pages do (and neither of those display relationships at all). So I dropped all that code on the React version.

I moved the macros that are only used in add_medium.tt to just be part of AddMedium.js.


root/edit/details/AddMedium.js
```diff
@@ -0,0 +1,191 @@
...
+  startNumber,
+}: CondensedTrackACsRowProps): React.Element<'tr'> => (
+  <tr className={loopParity(rowCounter)}>
+    <td className="pos t">
+      {endNumber && endNumber !== startNumber
+        ? startNumber + '-' + endNumber
+        : startNumber}
+    </td>
+    <td>
+      <ExpandedArtistCredit artistCredit={artistCredit} />
```
yvanzo, line 62: Note: Another change is that `ExpandedArtistCredit` does not insert extra `br` tags when there is nothing to expand, thus not all lines have the same height in the “Artist Credits” table.

root/edit/details/AddMedium.js
```diff
@@ -0,0 +1,191 @@
...
+  tracks,
+}: CondensedTrackACsProps):
+  Array<React.Element<typeof CondensedTrackACsRow>> => {
+  if (!tracks) {
+    return [];
+  }
+
+  let thisCredit;
+  let thisPosition = 0;
+  let rowCounter = 1;
```
yvanzo, line 77: Odd/even row classes have been shifted by one row, this fixes it:

```suggestion
  let rowCounter = 0;
```

root/edit/details/AddMedium.js
reosarevok: That's not a bad thing though, is it?

yvanzo: I have no idea whether initial height/new height are intentional or not, just noticing an undocumented change.

reosarevok: Oh, ok. Pretty sure that was a bug in the expanded_artist_credits implementation in TT :)

PR #1648, opened 2020-08-12
A few eslint config updates

mwiencek:
Just fixes a few issues I've run into while coding that always annoy me.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -330,7 +330,7 @@ rules:
   react/jsx-indent-props: [error, 2]
   react/jsx-key: warn
   react/jsx-max-props-per-line: [error, {maximum: 1, when: multiline}]
-  react/jsx-no-bind: warn
+  react/jsx-no-bind: [warn, {ignoreDOMComponents: true}]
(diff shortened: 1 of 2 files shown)


.eslintrc.yaml
```diff
@@ -330,7 +330,7 @@ rules:
   react/jsx-indent-props: [error, 2]
   react/jsx-key: warn
   react/jsx-max-props-per-line: [error, {maximum: 1, when: multiline}]
-  react/jsx-no-bind: warn
+  react/jsx-no-bind: [warn, {ignoreDOMComponents: true}]
```
yvanzo, line 334: > DOM components aren't memoized

There are 14 occurrences of `React.useMemo` in the code and there will be more as code gets moved to `react-table`. Could that be related?

mwiencek: I mean the DOM elements themselves aren't memoized by React, not that they can't be part of a larger component that's memoized (or uses `useMemo`).

But yeah, it is related in that this silences a useless warning even if you're already using `useMemo` correctly. But it's always useless on DOM elements, even outside of `useMemo`.

PR #1649, opened 2020-08-12
MBS-11034: Convert Remove medium edit to React 

reosarevok:
### Implement MBS-11034

Fixes a bug where the tracklist, which was supposed to be shown if present, was never shown because the check looked for the wrong variable.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1647


root/edit/details/RemoveMedium.js
```diff
@@ -0,0 +1,61 @@
...
+      <tr>
+        <th>{l('Medium:')}</th>
+        <td>
+          <MediumLink medium={display.medium} />
+        </td>
+      </tr>
+
+      {display.tracks ? (
+        <tr>
+          <th>{l('Tracklist:')}</th>
```
yvanzo, line 44: These two labels can be switched to `addColon`. Colonless strings are already used elsewhere as field labels.

yvanzo (approved):
LGTM but did not test. If you have time to use `addColon`, it's just bonus.

PR #1651, opened 2020-08-13
MBS-7409: Make "cannot attach discid" more obvious

reosarevok:
### Implement MBS-7409

Instead of a disabled radio icon (which e.g. recent versions of Firefox display the same as a non-disabled one!), this uses a forbidden sign. That should give a much more clear sign that you're not supposed to select those entries.

Discussion on the ticket is divided between showing these entries at all or not, and another suggestion is to move them to a separate section. But for now I feel this is the easiest way to make the difference more clear, at least.

Example: 
Screenshot from 2020-08-13 20-12-56

### DIFF
--- a/root/cdtoc/attach_list.tt
+++ b/root/cdtoc/attach_list.tt
@@ -38,7 +38,7 @@
           </label>
           <small>(<a class="toggle" style="cursor:pointer;">[% l('show tracklist') %]</a>)</small>
           [% IF this_medium_has_cdtoc %]
-            <div class="error">[% l('This CDTOC is already attached to this medium') %]</div>
+            <div class="error">[% l('This CDTOC is already attached to this medium.') %]</div>
           [% END %]
         </td>
         <td colspan="6"></td>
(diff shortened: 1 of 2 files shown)


root/cdtoc/attach_list.tt
```diff
@@ -23,11 +23,14 @@
...
-          <label
-            [%~ IF !medium.may_have_discids %] title="[% l('This medium format cannot have a disc ID attached') %]"[% END ~%]
-            [%~ IF this_medium_has_cdtoc %] title="[% l('This CDTOC is already attached to this medium') %]"[% END ~%]
-          >
-            <input type="radio" name="medium" value="[% medium.id %]"[% IF !medium.may_have_discids OR this_medium_has_cdtoc %] disabled="disabled"[% END %]/>
+          <label>
+            [% IF this_medium_has_cdtoc %]
+              <div class="cannot-attach-discid icon img" title="[% l('This CDTOC is already attached to this medium.') %]"></div>            
+            [% ELSIF !medium.may_have_discids %]
+              <div class="cannot-attach-discid icon img" title="[% l('This medium format cannot have a disc ID attached.') %]"></div>
```
yvanzo, line 30: Please add the missing full stop to the same sentence on line 41, in a separate commit.

yvanzo (approved):
Thanks for the screenshot! Just found a preexisting typo that should be fixed at the same time to avoid duplicating localization messages.

I would be happy with showing matching entries only, but this is a progress already.

PR #1652, opened 2020-08-14
Support GET requests in ssssss.psgi

mwiencek:
This was helpful in testing https://github.com/metabrainz/musicbrainz-server/pull/1216 to allow thumbnails to be displayed.

### DIFF
--- a/contrib/ssssss.psgi
+++ b/contrib/ssssss.psgi
@@ -34,6 +34,7 @@ use feature "switch";
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 use File::Copy;
+use File::Slurp qw( read_file );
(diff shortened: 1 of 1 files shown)


mwiencek:
See also https://github.com/metabrainz/CAA-indexer/pull/15

reosarevok:
The general idea seems fine, but do you have suggestions for how to test this? (such as, some get URLs :) )

mwiencek:
> The general idea seems fine, but do you have suggestions for how to test this? (such as, some get URLs :) )

Only by following HACKING-CAA and uploading some cover art that way, since this depends on thumbnail files existing in the `SSSSSS_STORAGE` dir. (I suppose you could just copy some image files there, but I don't remember what the directory structure or request format is off the top of my head.)

PR #1656, opened 2020-08-18
MBS-11039: Don't use entity id as key for seriesItemNumbers

reosarevok:
### Fix MBS-11039

An entity can appear more than once in a series with different series numbers. While it is uncommon, it is supported and has real-life use cases. With the current code, the number is overwritten and only the latest number is displayed for all rows of the series table.

This changes seriesItemNumbers to be an array that is later accessed based on the index of the react-table row (so, the entities array map).

As far as I can tell, this should work in any event where the arrays are just taken directly from Perl and sent up to React (even if we were to introduce server-side filtering or ordering later). If I'm missing a reason why this is a bad idea, do let me know! It's the most minor change I could figure out to fix this. I expect this would cause issues if we eventually decided to drop pagination etc and do filtering and ordering directly from react-table, but I trust we'd find a better alternative while researching for that change :)

### DIFF
--- a/root/utility/tableColumns.js
+++ b/root/utility/tableColumns.js
@@ -404,11 +404,11 @@ export function defineReleaseLabelsColumn(
 
 export function defineSeriesNumberColumn(
   props: {
-    +seriesItemNumbers: {+[entityId: number]: string},
+    +seriesItemNumbers: $ReadOnlyArray<string>,
   },
 ): ColumnOptions<CoreEntityT, number> {
   return {
-    Cell: ({cell: {value}}) => props.seriesItemNumbers[value],
+    Cell: ({row: {index}}) => props.seriesItemNumbers[index],
     Header: l('#'),
     accessor: 'id',
     cellProps: {className: 'number-column'},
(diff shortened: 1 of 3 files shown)


root/utility/tableColumns.js
```diff
@@ -404,11 +404,11 @@ export function defineReleaseLabelsColumn(
 
 export function defineSeriesNumberColumn(
   props: {
-    +seriesItemNumbers: {+[entityId: number]: string},
+    +seriesItemNumbers: $ReadOnlyArray<string>,
   },
 ): ColumnOptions<CoreEntityT, number> {
   return {
-    Cell: ({cell: {value}}) => props.seriesItemNumbers[value],
+    Cell: ({row: {index}}) => props.seriesItemNumbers[index + 1],
```
mwiencek, line 411: Aren't both `seriesItemNumbers` and the row index 0-based? How come the + 1 there?

reosarevok: Because I was initializing it as `my @item_numbers = [];` rather than just `my @item_numbers;`. Whoops!. Fixed.

yvanzo (approved):
Not providing a single data array is hacky and won’t work with future asynchronous calls, but if @reosarevok tested it and it solves a current bug it works for me.

PR #1658, opened 2020-08-19
MBS-11168: Convert Alias edit forms to React

anirudhjain75:
### Implement MBS-11168

Notes by @reosarevok, who wrote a fair amount of this: 
This converts both the Add/Edit alias form and the much smaller Delete alias form to React. I tried to split it as much as possible (see commit messages for details) but some commits are still pretty big.

The state usage here (with dispatch/reducer combos where some reducers just pass the action down to the component where it's most relevant to allow reuse) seems sensible to me but really needs to be checked in case we decide we want to do some things differently (the sooner we decide the less we'll need to refactor later).


reosarevok:
This has been in draft for quite a while. @anirudhjain75, want me to finish it myself? :)

anirudhjain75:
I'm not working on this anymore... Forgot to mention it. 😓. You can use templates I have on temp branch for the rest of the entities too. Thanks Reo

mwiencek:
I fixed a few bugs with the `guess-case-sortname` action: it was using the sortname as input instead of the name, was hardcoded to call `recording.guess` (when the entity type should vary, and it should call the `.sortname` method instead), and didn't pass in the artist type to handle person sort names.

Also fixed the use of `any` in `set-type` by adding some helper functions to copy values/errors between RO and RW fields, and moved repeated reducer action handling to `FormRowNameWithGuessCase` and `FormRowSortNameWithGuessCase`.

yvanzo (approved):
Works for me! Tested thoroughly from artist’s aliases tab; played with as many form fields and options as possible, including guess case, using French localization; and checked edit data.

Improvements for search hint are real nice.

Please merge commit https://github.com/metabrainz/musicbrainz-server/pull/1658/commits/0c0f91780c76c20e49d1022637eb22b46c448cb0 with into commit https://github.com/metabrainz/musicbrainz-server/pull/1658/commits/c381892d603096e41597717a629c651b0207ff42 to ease tracking converted code.

yvanzo (approved):
@mwiencek: Double-checked, guess-case button for sort name works correctly again.

mwiencek (approved):

jesus2099:
Sorry if my question is weird but, do you know why is the DOM around the edit note label and text area is rendered , then re-rendered a few milliseconds after?

```html
<div class="row" style="width: 1200px;">
	<label for="edit-note-text">Edit note:</label>
	<textarea class="edit-note" cols="80" id="edit-note-text" name="edit-alias.edit_note" rows="5" style="height: 132px; width: 98%;"></textarea>
</div>
```

I have a userscript inserting stuff after the `<label>` that is wiped out when the re-rendering occurs.
If I delay my addition, it stays. But it's empirical and I would like to know what event I could wait to add my stuff and to know if there will be no other re-rendering afterwards.

I have the exact same problem with dates fields, BTW.

reosarevok:
@jesus2099: that's probably just the hydration (we build most of the code server-side, then run `hydrate` on top for stuff that needs to run client-side). @mwiencek might know if there's a better way to know when that's done loading than just adding a delay.

jesus2099:
Oh yes, that [same `hydrate<Props>`](https://github.com/metabrainz/musicbrainz-server/pull/1658/files#diff-d2744bc6ab250ec067a4e55bacb47d047a69a525f80563a1968a7f2b9340df44R379-R382) as [on my other issue](https://github.com/metabrainz/musicbrainz-server/commit/2a3ce5f808fb907e78b3b565867a0a3b6859c974#diff-8faa5353a1c84c9ee6dcf90be9bd62c669390f97c07ad8cf8401a8cee0bc7a6fR137-R140)!

Maybe the solution is that I must add my stuff and let React know my stuff with `hydrate(mystuff)` or something like that.

I found [some doc](https://reactjs.org/docs/react-dom.html#hydrate), but well... :thinking: 

What is strange is that in my other issue, `div.recording-list` was hydrate, so I "_understand_" that my nodes are wiped out from `div.recording-list`.

But here, it's hydrate on... :open_mouth: `div.recording-name` (although I see the error in artist alias page). So it's strange.

PR #1659, opened 2020-08-20
MBS-8668: Block adding duplicates to a series

reosarevok:
### Fix MBS-8668

Elsewhere, we accept relationships as different if their link order varies. For series though, unless the number attribute is different, we probably don't want to have the same entity listed many times (and to the user is often not obvious at all that a link order even exists, especially if the series is set to automated order).

NB: this blocks adding new duplicates, but cases where the duplicates already exist will throw DuplicateViolation, which in the entity editor is hardcoded to give what is a completely unrelated error (see image below for the error, only difference is in the current version it won't show the green new rel line).

This is not only new with this code, it was already the case for any relationships that were already considered duplicate before - we're just not seeing it because the JS layer doesn't let us submit those duplicates. Not sure whether we should just purge the series rels duplicates from the DB if we merge this and otherwise ignore the issue, or throw a different exception for duplicate relationships (but using DuplicateViolation in Edit::Relationship::Edit does seem sensible in other cases). Or we could just remove the check completely from the Perl layer and just check it on JS.

Do we consider it safe to ignore link order for PART_OF_SERIES, or are there cases where this does matter?

Screenshot from 2020-08-19 15-48-18

### DIFF
--- a/lib/MusicBrainz/Server/Data/Relationship.pm
+++ b/lib/MusicBrainz/Server/Data/Relationship.pm
@@ -20,6 +20,7 @@ use MusicBrainz::Server::Data::Series;
 use MusicBrainz::Server::Data::URL;
 use MusicBrainz::Server::Data::Work;
 use MusicBrainz::Server::Data::Utils qw(
+    non_empty
     placeholders
     ref_to_type
     type_to_model
@@ -28,6 +29,7 @@ use MusicBrainz::Server::Constants qw(
     :direction
     %ENTITIES
     %ENTITIES_WITH_RELATIONSHIP_CREDITS
+    %PART_OF_SERIES
     @RELATABLE_ENTITIES
 );
 use Scalar::Util qw( weaken );
@@ -823,8 +825,18 @@ sub exists {
 
     $self->_check_types($type0, $type1);
 
-    my @props = qw(entity0 entity1 link_order link);
-    my @values = @{$values}{qw(entity0_id entity1_id link_order)};
+    my @props = qw(entity0 entity1);
+    my @values = @{$values}{qw(entity0_id entity1_id)};
+
+    # For part of series, do not count link order as different
+    # if everything else is the same in order to avoid duplicate additions
+    # (different number attributes still allowed)
+    my $ignore_link_order = grep { $_ eq $values->{link_type_gid} } values %PART_OF_SERIES;
+
+    if (non_empty($values->{link_order}) && !$ignore_link_order) {
+        push @props, 'link_order';
+        push @values, $values->{link_order};
+    }
 
     my $link = $self->c->model('Link')->find({
         link_type_id => $values->{link_type_id},
@@ -835,6 +847,7 @@ sub exists {
     });
 
     return 0 unless $link;
+    push @props, 'link';
     push @values, $link;
 
     return $self->sql->select_single_value(
--- a/lib/MusicBrainz/Server/Data/Relationship.pm
+++ b/lib/MusicBrainz/Server/Data/Relationship.pm
@@ -20,6 +20,7 @@ use MusicBrainz::Server::Data::Series;
 use MusicBrainz::Server::Data::URL;
 use MusicBrainz::Server::Data::Work;
 use MusicBrainz::Server::Data::Utils qw(
+    contains_number
     non_empty
     placeholders
     ref_to_type
@@ -29,7 +30,7 @@ use MusicBrainz::Server::Constants qw(
     :direction
     %ENTITIES
     %ENTITIES_WITH_RELATIONSHIP_CREDITS
-    %PART_OF_SERIES
+    @PART_OF_SERIES_LINK_TYPE_IDS
     @RELATABLE_ENTITIES
 );
 use Scalar::Util qw( weaken );
@@ -831,7 +832,7 @@ sub exists {
     # For part of series, do not count link order as different
     # if everything else is the same in order to avoid duplicate additions
     # (different number attributes still allowed)
-    my $ignore_link_order = grep { $_ eq $values->{link_type_gid} } values %PART_OF_SERIES;
+    my $ignore_link_order = contains_number(\@PART_OF_SERIES_LINK_TYPE_IDS, $values->{link_type_id});
 
     if (non_empty($values->{link_order}) && !$ignore_link_order) {
         push @props, 'link_order';
--- a/lib/MusicBrainz/Server/Constants.pm
+++ b/lib/MusicBrainz/Server/Constants.pm
@@ -75,7 +75,8 @@ our @EXPORT_OK = (
         $LIMIT_FOR_EDIT_LISTING
         $ARTIST_ARTIST_COLLABORATION
         $AMAZON_ASIN_LINK_TYPE_ID
-        %PART_OF_SERIES $PART_OF_AREA_LINK_TYPE $PART_OF_AREA_LINK_TYPE_ID
+        %PART_OF_SERIES @PART_OF_SERIES_LINK_TYPE_IDS
+        $PART_OF_AREA_LINK_TYPE $PART_OF_AREA_LINK_TYPE_ID
         $SERIES_ORDERING_TYPE_AUTOMATIC $SERIES_ORDERING_TYPE_MANUAL
         $ARTIST_RENAME_LINK_TYPE
         $LABEL_RENAME_LINK_TYPE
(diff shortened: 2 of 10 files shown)


yvanzo (changes requested):
I guess the DuplicateViolation was intended for newly created duplicate entity, not for duplicate relationship. It seems worth to have two distinct errors then.

Pre-existing duplicates should probably not block editing a series, just like we intentionally don’t block editing an entity if an existing URL relationship is not valid anymore.

The check should still be made on the back-end as data validation is one of its main purposes.

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -580,8 +581,18 @@ sub exists {
...
-    my @values = @{$values}{qw(entity0_id entity1_id link_order)};
+    my @props = qw(entity0 entity1);
+    my @values = @{$values}{qw(entity0_id entity1_id)};
+
+    # For part of series, do not count link order as different
+    # if everything else is the same in order to avoid duplicate additions
+    # (different number attributes still allowed)
+    my $ignore_link_order = grep { $_ eq $values->{link_type_gid} } values %PART_OF_SERIES;
+
+    if ($values->{link_order} && !$ignore_link_order) {
```
mwiencek, line 592: Should check `defined` to handle 0, right? Or `non_empty`
```suggestion
    if (defined $values->{link_order} && !$ignore_link_order) {
```

root/static/scripts/relationship-editor/common/entity.js
```diff
@@ -178,12 +178,18 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
...
-                    var obj = _.omit(rel.editData(), "id");
-
-                    obj.begin_date = mergeDates(rel.begin_date, other.begin_date);
-                    obj.end_date = mergeDates(rel.end_date, other.end_date);
+                    /*
+                     * If it is a duplicate part of series,
+                     * just ignore it completely, otherwise merge it.
+                     */
+                    if (!rel.isSeriesPart()) {
+                        var obj = _.omit(rel.editData(), "id");
```
mwiencek, line 186:
```suggestion
                        const obj = _.omit(rel.editData(), "id");
```

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -580,8 +581,18 @@ sub exists {
...
 
-    my @props = qw(entity0 entity1 link_order link);
-    my @values = @{$values}{qw(entity0_id entity1_id link_order)};
+    my @props = qw(entity0 entity1);
+    my @values = @{$values}{qw(entity0_id entity1_id)};
+
+    # For part of series, do not count link order as different
+    # if everything else is the same in order to avoid duplicate additions
+    # (different number attributes still allowed)
+    my $ignore_link_order = grep { $_ eq $values->{link_type_gid} } values %PART_OF_SERIES;
```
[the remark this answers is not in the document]
reosarevok: Those are only done with arrows though, so you can't get a duplicate order, can you?

reosarevok:
Testing this with the new relationship editor, the JS part is no longer needed because the new editor already detects these as duplicates and blocks them.

The Perl part might be worth keeping in case the JS check fails / gets bypassed.

For some reason, the exception message is not being shown anymore though - in fact `/series/476798b7-12ae-4237-9e2f-e2d4d6fab5f2` (with a duplicate release of catno 469 202-2) does not show any error message at all when trying to edit it.

mwiencek:
> Testing this with the new relationship editor, the JS part is no longer needed because the new editor already detects these as duplicates and blocks them.

I can still add duplicates without any UI error in the beta branch, but only one duplicate per entity (same as MBS-12838 describes).

1. Add a new part to any automatic series, with a number set (from the series edit page)
2. Submit the edit
3. Reload the series edit page and add the exact same relationship with the same number attributes
4. It lets you submit the edit

If I try to repeat steps 3 & 4 again, it doesn't let me (it finally detects the new relationship as a duplicate of the previous two, somehow).

So it seems like some UI changes must be needed still.

reosarevok:
Hmm, ok, I guess I was just testing adding the same rel twice, but not adding an existing rel :) I'll look into it.

lib/MusicBrainz/Server/Data/Relationship.pm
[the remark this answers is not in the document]
reosarevok: > "stuff like recording-recording compilation/DJ-mix relationships probably make sense to ignore too"

Actually if the same recording is used twice in a DJ-mix or whatnot, that would seem legit - maybe better to stick to series only when ignoring link orders.

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -778,8 +780,18 @@ sub exists {
...
```
mwiencek, line 789: I see you pass `link_type_gid` to here in two places below, but if I search for `model('Relationship')->exists`, there are 5 calls (including one in the tests). So it seems kinda unclear whether it's required. Would it make sense to just add a constant for the part-of-series row IDs instead, so that a gid isn't needed?

reosarevok: You mean just use `link_type_id` and have something like `@PART_OF_SERIES_ROW_IDS` that we look at instead? If that seems better, I can do that.

I'm confused about the "So it seems kinda unclear whether it's required" bit - if what is required? :) We need to look at this in here in some way to avoid a `MusicBrainz::Server::Edit::Exceptions::NoChanges->throw`, I understand.

reosarevok: Seems we don't use the Perl `%PART_OF_SERIES` anywhere else, not sure if it should in that case be removed or what :)

reosarevok: I pushed a commit changing to use row ids - tested quickly by disabling the JS check, seems to work fine.

mwiencek: > I'm confused about the "So it seems kinda unclear whether it's required" bit - if what is required? :) We need to look at this in here in some way to avoid a MusicBrainz::Server::Edit::Exceptions::NoChanges->throw, I understand.

I meant whether the `link_type_gid` was required, since you weren't passing it in all cases. (For example, maybe because the places it wasn't being passed never dealt with series relationships.) But I prefer we always perform this check (hence switching to the row IDs).

Now I'm a bit confused: isn't this patch intended to cause *more* `NoChanges` exceptions to be thrown? (Since `NoChanges` == this is a duplicate.)

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -823,8 +825,18 @@ sub exists {
...
 
-    my @props = qw(entity0 entity1 link_order link);
-    my @values = @{$values}{qw(entity0_id entity1_id link_order)};
+    my @props = qw(entity0 entity1);
+    my @values = @{$values}{qw(entity0_id entity1_id)};
+
+    # For part of series, do not count link order as different
+    # if everything else is the same in order to avoid duplicate additions
+    # (different number attributes still allowed)
+    my $ignore_link_order = grep { $_ eq $values->{link_type_id} } values %PART_OF_SERIES_ROW_IDS;
```
mwiencek, line 834: Should we just store them as a list like `PART_OF_SERIES_LINK_TYPE_IDS` in root/static/scripts/common/constants.js so that we don't need to convert it via `values`? (At least until we have a use for looking them up by entity type.) I guess we could also use `contains_number` from `Data::Utils`, then. :)

mwiencek (commented):
This looks good to me overall, though it's kinda subtle behavior so it seems like the kind of thing that should have at least one test on the Perl and JS side:
 * We have an existing test for "Don't consider relationships with different link orders to be the same" in t/lib/t/MusicBrainz/Server/Data/Relationship.pm which we could copy and adapt for the series case.
 * On the JS side we have a "compareRelationships: Series comparisons" test in root/static/scripts/tests/relationship-editor/utility/compareRelationships.js which we could just amend with one additional case.

lib/MusicBrainz/Server/Data/Relationship.pm
reosarevok: Oh, I didn't realize we were already doing it that way in JS. Then we might as well, sure.

reosarevok:
Added the suggested tests. Reworded the existing JS ones a bit since they were confusingly mentioning "order" when they meant "number" (so, "ordering attribute").

mwiencek:
It seems we're also lacking a JS check for the case where you edit an existing relationship to have the same number as another. The Perl still blocks the addition with a `NoChanges` exception, but there's otherwise no error message when the form is reloaded (with the dupe relationship still being visible).

Ideally the relationship dialog would block the edit to begin with. I guess the `!(isPartOfSeries && isAdd)` check should also include edits, though it will have the effect of blocking any edits to the relationship that don't also fix the dupe number, IIUC.

PR #1661, opened 2020-08-24
MBS-11048: Convert Edit Area edit to React

reosarevok:
### Implement MBS-11048

Changes:

Use the double table system for consistency and better display.


mwiencek:
> Use the double table system for consistency and better display.

Maybe I'm misremembering, but I thought we actually combined them in another PR (for a different edit type) because we couldn't find a reason they were separate. :) So since I still don't know, how do they make the display better?

reosarevok:
We thought about doing that, but didn't so far. Turns out that without that, this edit had weird alignment issues, IIRC.

yvanzo (approved):

PR #1662, opened 2020-08-24
Edit diff improvements

reosarevok:
Two small commits improving our diffs. The first is only refactoring, but makes sense to me at least. The second fixes what I think is probably a tiny bug, unless it was intended like that for some reason.


yvanzo (approved):
That was a bug indeed.

PR #1663, opened 2020-08-24
MBS-11049: Temporarily disable external links to Decoda

yvanzo:
# Problem

MBS-11049: Decoda has been reported to sometimes redirect to a scam page.

# Solution

To do no harm, this patch temporarily disables external links to Decoda.
It should be reverted whenever their situation gets resolved.

### DIFF
--- a/root/static/scripts/common/components/EntityLink.js
+++ b/root/static/scripts/common/components/EntityLink.js
@@ -17,6 +17,7 @@ import entityHref from '../utility/entityHref';
 import formatDatePeriod from '../utility/formatDatePeriod';
 import isolateText from '../utility/isolateText';
 import nonEmpty from '../utility/nonEmpty';
+import isGreyedOut from '../../../../url/utility/isGreyedOut';
 
 type DeletedLinkProps = {
   +allowNew: boolean,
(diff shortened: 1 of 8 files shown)


reosarevok (approved):
Seems good to me. Added an extra commit with an explanatory tooltip because I'm afraid otherwise it will actually encourage people to be curious about why it doesn't work in the first place and visit it! But if you don't think it's a good idea, just drop it :)

root/static/scripts/common/components/EntityLink.js
```diff
@@ -215,8 +216,17 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
...
+    ? (
+      <span
+        className="deleted"
+        title={isGreyedOut(content)
+          ? l(`This link has been temporarily disabled because
+            it has been reported as potentially harmful.`)
+          : null}
+      >
+        {isolateText(content)}
+      </span>
```
yvanzo, line 228: In this specific case, the link might not be harmful but just linked via an “ended” relationship.

yvanzo: My bad, didn’t see the `isGreyedOut` test.

yvanzo:
Fixed flow error in [build 10605](https://app.circleci.com/pipelines/github/metabrainz/musicbrainz-server/3750/workflows/b62fa62a-3b37-4918-8fce-34b06f6f8fa0/jobs/10605) with commit <https://github.com/metabrainz/musicbrainz-server/commit/ce6cdcac4901270854f538e4c2f0406a9e7d91fe>.

PR #1666, opened 2020-08-25
MBS-11052: Convert Edit Relationship Attribute edit to React 

reosarevok:
### Implement MBS-11052

Changes:

If the description has changed, but only in the underlying HTML, we display the unchanged "expanded" description without a diff and then diff the raw text.

We don't show a description line at all anymore if the description is (unchangedly) empty.

Description changes are no longer WordDiffed (since they are React elements, not strings). In most cases that might be an improvement, since any non-trivial description change is meant to be read as a whole.

On a separate commit, also use expand2react on Add Relationship Attribute (AFAICT that mirrors what it used to display before being converted).


root/edit/details/EditRelationshipAttribute.js
```diff
@@ -0,0 +1,115 @@
...
+  const newDescription = display.description.new;
+  const descriptionRawChanges = newDescription !== oldDescription;
+  const newExpandedDescription = newDescription
+    ? expand2react(newDescription)
+    : '';
+  const oldExpandedDescription = oldDescription
+    ? expand2react(oldDescription)
+    : '';
+  const descriptionChanges =
+    newExpandedDescription !== oldExpandedDescription;
```
mwiencek, line 44: If `expand2react` returns a React element (like say a fragment), then these will always be `!==`. They'd only be equal if they refer to the exact same object in memory. Or am I misunderstanding what this is doing?

In 83fd3a04a6782faed1725e7d315c1a96bfe0e0bb we started showing diffs of the raw HTML for link type edits, and I think it works well because the majority of the description edits seem to be small wording or tag adjustments and not rewrites. It'd be nice to keep some consistency between them in either case.

reosarevok: Would you also show the HTML in AddRelationshipAttribute?

I'm thinking about always displaying both raw and final, but have only raw be a real diff. That way, we'd allow people to see the actual HTML change *and* how it will look like on the site.

reosarevok:
Changed this quite a bit to leave more stuff raw (using IntentionallyUntranslatedIcon) and do some more improvements after what @mwiencek suggested. See commit messages for more info.

root/edit/details/EditRelationshipAttribute.js
```diff
@@ -0,0 +1,113 @@
...
+          extraOld={nonEmpty(oldDescription) ? rawIconSection : null}
+          label={addColonText(l('Description'))}
+          newText={newDescription}
+          oldText={oldDescription}
+        />
+      ) : nonEmpty(oldDescription) ? (
+        <tr>
+          <th>{addColonText(l('Description'))}</th>
+          <td colSpan="2">
+            {expand2react(l_relationships(oldDescription))}
```
yvanzo, line 89: Even though _raw_ doesn’t mean anything to me :wink: should this raw description be localized still?

```suggestion
            {expand2react(oldDescription)}
```

reosarevok: This one is not raw. This is the one that is shown just for extra info when it hasn't been changed, so localizing it makes sense to me.

yvanzo (approved):

mwiencek (approved):

PR #1667, opened 2020-08-26
Hydrate improvements

mwiencek:
These commits were cherry-picked from my React relationship editor branch. They're aimed at reducing the size of the JSON for `$c` and props embedded into the HTML.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -21,6 +21,7 @@ globals:
   # Fake globals; these are auto-imported upon use by Webpack's
   # ProvidePlugin, but look like real globals to eslint
   __webpack_public_path__: writable
+  GLOBAL_CATALYST_CONTEXT_NAMESPACE: readonly
   GLOBAL_DBDEFS_NAMESPACE: readonly
   addColon: readonly
   addColonText: readonly
--- a/root/entity/Tags.js
+++ b/root/entity/Tags.js
@@ -36,6 +36,7 @@ const Tags = ({
         $c={$c}
         aggregatedTags={allTags}
         entity={entity}
+        genreMap={$c.stash.genre_map}
         more={moreTags}
         userTags={userTags}
       />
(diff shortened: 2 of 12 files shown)


reosarevok:
Seems to break TagEditor, which looks for props.$c.

reosarevok (changes requested):
Seems to break TagEditor, which looks for props.$c. Only when logged in, of course :)

mwiencek:
Whoops, think I broke that while rebasing. Should be fixed now. :)

mwiencek:
BTW, this PR reduces the size of the Beatles artist page (note: the HTML only, not counting JS/CSS/images/etc.) from 1.2 MB to 315 kB. (In production the response is gzipped of course, so it's "only" 165 kB gzipped, but the uncompressed size is still 1.2 MB.)

reosarevok (approved):
Seems to work now

yvanzo (approved):
The current size of the Beatles artist page is 158.68 KB (162 490 bytes) according to my browser.

mwiencek:
> The current size of the Beatles artist page is 158.68 KB (162 490 bytes) according to my browser. Edit: uncompressed size is 776K when downloaded using wget.

The compressed response size will vary slightly between users and reloads. But 776 KB is closer to what you'd get if you're not logged in—you're likely running wget without any login cookie.

yvanzo:
> The compressed response size will vary slightly between users and reloads.

I was interested in comparing the gzipped size before/after but was not able to get something comparable to your measures.

> 776 KB is closer to what you'd get if you're not logged in—you're likely running wget without any login cookie.

Yes it a was a simple `wget` without any login cookie.

mwiencek:
I see, my 1.2 MB to 315 KB measurement was for the uncompressed difference. For the gzipped comparison, I'm not sure I can exactly emulate the gzip compression that nginx will do here, but the command below is giving 163410 on the master branch and 45878 on this one, which is ~117.5 KB smaller. (This is with pointing my local MB instance to pink, and running compile_resources.sh with `WEBPACK_MODE=production` and `NODE_ENV=production`.)
```sh
# replace $sid with actual session id
curl --cookie 'musicbrainz_server_session=$sid' 'http://localhost:5000/artist/b10bbbfc-cf9e-42e0-be17-e2c3e1d2600d' | gzip | wc
```

PR #1668, opened 2020-08-26
Autocomplete2 improvements

mwiencek:
This PR collects various commits to Autocomplete2 that I've cherry-picked from my React relationship editor and artist credit editor branches. I'd say the primary change here is converting the component into a stateless one, but there are a lot of smaller changes needed for those branches too.

You can still test the new Autocomplete2 component by running `./script/compile_resources.sh tests` (ignore any jsdom error) and navigating to `http://localhost:5000/static/scripts/tests/autocomplete2.html`.


mwiencek:
Some commits like fc0b291526271144fe6db706024bb1037299a105 are later "undone" because they aren't relevant to a stateless component. But they were part of the logical progression of the branches and removing them caused too many conflicts, so don't really see a point to get rid of them.

root/static/scripts/common/components/Autocomplete2/types.js
```diff
@@ -37,19 +37,20 @@ export type SearchAction = {
   +type: 'search-after-timeout',
 };
 
+/* eslint-disable flowtype/sort-keys */
```
reosarevok, line 46: Maybe re-enable afterwards?

reosarevok (approved):
The test seems to work at least, and the general changes seem sensible.

yvanzo (approved):
That one I tested a while ago, just forgot to approve, sorry.

PR #1669, opened 2020-08-26
Improve flow typing and add flow strict-local when possible

reosarevok:

### DIFF
--- a/root/collection/CollectionHeader.js
+++ b/root/collection/CollectionHeader.js
@@ -1,5 +1,5 @@
 /*
- * @flow
+ * @flow strict-local
  * Copyright (C) 2018 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
@@ -28,7 +28,7 @@ const CollectionHeader = ({
   user,
 }: Props): React.Element<typeof React.Fragment> => {
   const owner = collection.editor;
-  const viewingOwnCollection = user && owner && owner.id === user.id;
+  const viewingOwnCollection = Boolean(user && owner && owner.id === user.id);
   const subHeading = (
     <>
       {collection.public ? (
--- a/root/components/FormRow.js
+++ b/root/components/FormRow.js
@@ -1,5 +1,5 @@
 /*
- * @flow
+ * @flow strict
  * Copyright (C) 2017 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
@@ -16,10 +16,13 @@ type Props = {
 
 const FormRow = ({
   children,
-  hasNoLabel,
+  hasNoLabel = false,
   ...props
 }: Props): React.Element<'div'> => (
-  <div className={'row' + (hasNoLabel ? ' no-label' : '')} {...props}>
+  <div
+    className={'row' + (hasNoLabel ? ' no-label' : '')}
+    {...props}
+  >
     {children}
   </div>
 );
--- a/root/components/RelationshipsTable.js
+++ b/root/components/RelationshipsTable.js
@@ -1,5 +1,5 @@
 /*
- * @flow
+ * @flow strict-local
  * Copyright (C) 2019 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
@@ -34,12 +34,12 @@ const RelationshipsTable = ({
   entity,
   fallbackMessage,
   heading,
-  showCredits,
+  showCredits = false,
 }: Props): React.MixedElement | null => {
   const appearances = generateRelationshipAppearancesList(entity);
   const relationshipTypes = Object.keys(appearances);
   if (!appearances || relationshipTypes.length === 0) {
-    return fallbackMessage ? (
+    return nonEmpty(fallbackMessage) ? (
       <>
         <h2>{heading}</h2>
         <p>{fallbackMessage}</p>
@@ -65,7 +65,7 @@ const RelationshipsTable = ({
       if (!hasArtistColumn && relationship.target.artistCredit) {
         hasArtistColumn = 1;
       }
-      if (!hasLengthColumn && relationship.target.length) {
+      if (!hasLengthColumn && relationship.target.length != null) {
         hasLengthColumn = 1;
       }
       if (hasCreditColumn && hasAttributeColumn &&
--- a/root/components/UserAccountTabs.js
+++ b/root/components/UserAccountTabs.js
@@ -1,5 +1,5 @@
 /*
- * @flow
+ * @flow strict-local
  * Copyright (C) 2018 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
@@ -20,8 +20,8 @@ function buildTabs(
   user: EditorT,
   page: string,
 ): $ReadOnlyArray<React.Element<'li'>> {
-  const viewingOwnProfile = $c.user && $c.user.id === user.id;
-  const showAdmin = $c.user && $c.user.is_account_admin;
+  const viewingOwnProfile = Boolean($c.user && $c.user.id === user.id);
+  const showAdmin = Boolean($c.user && $c.user.is_account_admin);
   const showPrivate = showAdmin || viewingOwnProfile;
 
   const userName = encodeURIComponent(user.name);
@@ -86,7 +86,8 @@ function buildTabs(
     ));
   }
 
-  if (showAdmin || DBDefs.DB_STAGING_TESTING_FEATURES && $c.user) {
+  if (showAdmin ||
+    DBDefs.DB_STAGING_TESTING_FEATURES && $c.user) {
     tabs.push(buildTab(
       page,
       l('Edit User'),
--- a/root/components/list/ReleaseList.js
+++ b/root/components/list/ReleaseList.js
@@ -1,5 +1,5 @@
 /*
- * @flow
+ * @flow strict-local
  * Copyright (C) 2019 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
@@ -48,13 +48,13 @@ const ReleaseList = ({
   order,
   releases,
   seriesItemNumbers,
-  showInstrumentCreditsAndRelTypes,
-  showRatings,
+  showInstrumentCreditsAndRelTypes = false,
+  showRatings = false,
   sortable,
 }: Props): React.Element<typeof Table> => {
   const columns = React.useMemo(
     () => {
-      const checkboxColumn = $c.user && checkboxes
+      const checkboxColumn = $c.user && nonEmpty(checkboxes)
         ? defineCheckboxColumn({name: checkboxes})
         : null;
       const seriesNumberColumn = seriesItemNumbers
@@ -77,7 +77,9 @@ const ReleaseList = ({
       const formatColumn = defineTextColumn<ReleaseT>({
         columnName: 'format',
         getText:
-          entity => entity.combined_format_name || l('[missing media]'),
+          entity => nonEmpty(entity.combined_format_name)
+            ? entity.combined_format_name
+            : l('[missing media]'),
         order: order,
         sortable: sortable,
         title: l('Format'),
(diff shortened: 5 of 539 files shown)


root/components/FormRow.js
```diff
@@ -19,7 +19,10 @@ const FormRow = ({
   hasNoLabel,
   ...props
 }: Props): React.Element<'div'> => (
-  <div className={'row' + (hasNoLabel ? ' no-label' : '')} {...props}>
+  <div
+    className={'row' + (hasNoLabel /*:: === true */ ? ' no-label' : '')}
+    {...props}
+  >
```
mwiencek, line 25: In cases where we have an optional property like here, I think it's better for clarity to just set the default value `hasNoLabel = false,` in the destructuring above.

root/static/scripts/common/utility/bracketed.js
```diff
@@ -27,14 +27,14 @@ export default function bracketed(
   text: ?VarSubstArg,
   args?: Args,
 ): Expand2ReactOutput {
-  if (text) {
+  if (text != null && text !== '') {
```
mwiencek, line 30: Looks like the definition of `nonEmpty` to me. :)

root/utility/releaseGroupType.js
```diff
@@ -14,7 +14,7 @@ export default function releaseGroupType(
 ): string {
   const types = [];
   let id = releaseGroup.typeID;
-  if (id) {
+  if (id != null) {
```
mwiencek, line 17: Just note that this can change behavior in cases where we use an `id` of 0, which might just be some hacky places in edit classes. (Which is why I do think we think avoid that if possible.)

root/vars.js
```diff
@@ -21,7 +21,9 @@ declare var GLOBAL_DBDEFS_NAMESPACE: '__MB_DBDefs__';
 declare var addColon: (variable: Expand2ReactInput) => Expand2ReactOutput;
 declare var addColonText: (variable: StrOrNum) => string;
 declare var hyphenateTitle: (title: string, subtitle: string) => string;
-
+function nonEmpty(value: mixed): boolean %checks {
+  return value !== null && value !== undefined && value !== '';
+}
```
mwiencek, line 26: Does this work without an accompanying `declare var`? I thought it wouldn't make it global without that. I guess you can tell by if vscode shows the correct function type on hover rather than just `any`.

mwiencek:
Regarding the test failure, you have to add the path to `nonEmpty` to webpack/providePluginConfig.js.

root/utility/releaseGroupType.js
reosarevok: I specifically used this *because* I thought we might have cases where we hackishly use an ID of 0 and we'd still want to display the stuff, even if the ID is falsy.

root/components/FormRow.js
reosarevok: Didn't realize it was equivalent, but yes, that seems a lot better :) Changed.

root/utility/releaseGroupType.js
mwiencek: That can be true, though this case is a bit different since we also require the ID to be in `linkedEntities` - I don't think an ID of 0 would be in there 'cause we'd use a positive number for a fake ID.

The code here shouldn't be affected at least, I was just making a general point about how using 0 as a placeholder complicates things. :)

root/components/RelationshipsTable.js
```diff
@@ -56,7 +56,7 @@ const RelationshipsTable = ({
         ? relationship.entity1_credit
         : relationship.entity0_credit;
 
-      if (!hasCreditColumn && showCredits && sourceCredit) {
+      if (!hasCreditColumn && showCredits /*:: === true */ && sourceCredit) {
```
mwiencek, line 59: Here you can do `showCredits = false,` above

root/components/UserAccountTabs.js
```diff
@@ -22,14 +22,14 @@ function buildTabs(
 ): $ReadOnlyArray<React.Element<'li'>> {
   const viewingOwnProfile = $c.user && $c.user.id === user.id;
   const showAdmin = $c.user && $c.user.is_account_admin;
-  const showPrivate = showAdmin || viewingOwnProfile;
+  const showPrivate = showAdmin /*:: === true */ || viewingOwnProfile;
```
mwiencek, line 25: Maybe just wrap `viewingOwnProfile` and `showAdmin` in `!!( ... )` so that those are always boolean. Then you shouldn't need any of the `/*:: === true */` in this file.

root/components/list/ReleaseList.js
```diff
@@ -115,11 +119,12 @@ const ReleaseList = ({
         sortable: sortable,
         title: l('Barcode'),
       });
-      const instrumentUsageColumn = showInstrumentCreditsAndRelTypes
-        ? defineInstrumentUsageColumn({
-          instrumentCreditsAndRelTypes: instrumentCreditsAndRelTypes,
-        })
-        : null;
+      const instrumentUsageColumn =
+        showInstrumentCreditsAndRelTypes /*:: === true */
```
mwiencek, line 123: You can set this and `showRatings` (below) to false in the destructuring.

root/edit/details/AddRelationship.js
```diff
@@ -50,7 +50,7 @@ const AddRelationship = ({edit}: Props): React.MixedElement => {
           <td>{edit.display_data.relationship.linkOrder}</td>
         </tr>
       ) : null}
-      {edit.display_data.unknown_attributes ? (
+      {edit.display_data.unknown_attributes /*:: === true */ ? (
```
mwiencek, line 53: It looks like `unknown_attributes` is always outputted by `Edit::Relationship::Create::build_display_data` - maybe this should be non-optional in the type instead?

root/entity/Subscribers.js
```diff
@@ -49,7 +49,7 @@ const Subscribers = ({
         <>
           <p>
             {entityType === 'editor' ? (
-              viewingOwnProfile ? (
+              viewingOwnProfile /*:: === true */ ? (
```
mwiencek, line 52: I'd just do `!!$c.user` above so that it's a boolean.

root/layout/index.js
```diff
@@ -219,21 +219,21 @@ const Layout = ({
...
-      {!!($c.sessionid && $c.flash.message) &&
+      {!!(nonEmpty($c.sessionid) && nonEmpty($c.flash.message)) &&
         <div className="banner flash">
           <p dangerouslySetInnerHTML={{__html: $c.flash.message}} />
         </div>}
 
       <div
         className={(fullWidth ? 'fullwidth ' : '') +
-          (homepage ? 'homepage' : '')}
+          (homepage /*:: === true */ ? 'homepage' : '')}
```
mwiencek, line 229: `homepage = false,` above

root/static/scripts/common/components/ArtistListEntry.js
```diff
@@ -81,9 +81,9 @@ const ArtistListRow = ({
     <td>
       <DescriptiveLink entity={artist} />
     </td>
-    {showSortName ? <td>{artist.sort_name}</td> : null}
+    {showSortName /*:: === true */ ? <td>{artist.sort_name}</td> : null}
```
mwiencek, line 84: `showSortName = false,`

root/static/scripts/common/components/ArtistListEntry.js
```diff
@@ -95,7 +95,7 @@ const ArtistListRow = ({
     <td>
       {artist.area ? <DescriptiveLink entity={artist.area} /> : null}
     </td>
-    {showBeginEnd ? (
+    {showBeginEnd /*:: === true */ ? (
```
mwiencek, line 98: `showBeginEnd = false,`

root/static/scripts/common/components/ArtistListEntry.js
```diff
@@ -111,12 +111,12 @@ const ArtistListRow = ({
         </td>
       </>
     ) : null}
-    {showRatings ? (
+    {showRatings /*:: === true */ ? (
```
mwiencek, line 114: `showRatings = false,`

root/static/scripts/common/components/ArtistListEntry.js
```diff
@@ -111,12 +111,12 @@ const ArtistListRow = ({
...
       <td>
         <RatingStars entity={artist} />
       </td>
     ) : null}
-    {showInstrumentCreditsAndRelTypes ? (
+    {showInstrumentCreditsAndRelTypes /*:: === true */ ? (
```
mwiencek, line 119: ` = false,`

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -71,7 +71,7 @@ export const WorkListRow = ({
     <td>
       <WorkArtists artists={work.artists} />
     </td>
-    {showIswcs ? (
+    {showIswcs /*:: === true */ ? (
```
mwiencek, line 74: ` = false,`

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -97,12 +99,12 @@ export const WorkListRow = ({
         ))}
       </ul>
     </td>
-    {showAttributes ? (
+    {showAttributes /*:: === true */ ? (
```
mwiencek, line 102: ` = false,`

root/static/scripts/common/components/WorkListEntry.js
```diff
@@ -97,12 +99,12 @@ export const WorkListRow = ({
...
       <td>
         <AttributeList entity={work} />
       </td>
     ) : null}
-    {showRatings ? (
+    {showRatings /*:: === true */ ? (
```
mwiencek, line 107: ` = false,`

root/collection/CollectionHeader.js
```diff
@@ -49,7 +49,7 @@ const CollectionHeader = ({
                 '/collections'
               }
             >
-              {viewingOwnCollection ? (
+              {viewingOwnCollection /*:: === true */ ? (
```
mwiencek, line 52: `const viewingOwnCollection = !!( ... );`

root/static/scripts/common/utility/formatTrackLength.js
```diff
@@ -1,13 +1,17 @@
...
  * and is licensed under the GPL version 2, or (at your option) any
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-export default function formatTrackLength(milliseconds, placeholder='?:??') {
-    if (!milliseconds) {
+export default function formatTrackLength(
+    milliseconds?: number | null,
+    placeholder: string = '?:??',
+): string {
```
mwiencek, line 13: The `?` seems mixed up here
```suggestion
export default function formatTrackLength(
    milliseconds: number | null,
    placeholder?: string = '?:??',
): string {
```

root/static/scripts/common/utility/getBooleanCookie.js
```diff
@@ -8,7 +9,10 @@
 
 import getCookie from './getCookie';
 
-function getBooleanCookie(name, defaultValue = false) {
+function getBooleanCookie(
+  name: string,
+  defaultValue: boolean = false,
```
mwiencek, line 14:
```suggestion
  defaultValue?: boolean = false,
```

root/static/scripts/common/utility/isBlank.js
```diff
@@ -1,3 +1,12 @@
...
+/*
+ * @flow strict
+ * Copyright (C) 2015 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export default function isBlank(str: string): boolean {
```
mwiencek, line 10: I think `str` is intended to allow nullish values here, hence the `|| ''` below.
```suggestion
export default function isBlank(str: ?string): boolean {
```

yvanzo (approved):
:shipit:

PR #1670, opened 2020-08-31
Remove dependency on lodash

mwiencek:
This branch is pretty old so I finally cleaned it up for submission. It reduces the size of our minified JS by about 53.25 KB.

Many lodash methods can be replaced by native ES5+ methods (which we get for free via Babel and core-js). There are some lodash utilities that are pretty useful (`groupBy`, `uniq`, etc.), but they're also very easy to implement in modern JS with `Map`/`Set`, so I'd rather just have our own simple implementations with proper Flow types.

It's possible to reduce lodash's footprint by specifying the imports directly, as in `import groupBy from 'lodash/groupBy';`, and this is what I originally aimed to do, but I found that after I switched the obvious cases to all use native JS methods, there were only a small handful of methods left. It can also be misleading how much space the direct imports save, since those can still pull in a bunch of internal, shared collections code that we don't need.

Also, keeping lodash around for just the few useful methods it has is the slippery slope that led us to the current situation: since we knew it was there, we've used it anywhere and everywhere it's not needed, even where the native JavaScript way is more straightforward. Our JS bundles are extremely bloated, with `common-chunks` at a whopping 1.3 MB *minified*. Ideally we just have React as our primary dependency.

This PR is big, but the commits are pretty isolated. Besides running tests, to check this I left the JS console open and browsed the site, plus submitted at least one entity of every type, interacting with all of the form fields as best I could, and seeing if any errors occurred.


root/static/scripts/release-editor/edits.js
```diff
@@ -195,8 +197,10 @@ releaseEditor.edits = {
             newMediumData = _.cloneDeep(newMediumData);
 
             if (medium.id) {
-                var newNoPosition = _.omit(newMediumData, "position");
-                var oldNoPosition = _.omit(oldMediumData, "position");
+                const newNoPosition = {...newMediumData};
```
reosarevok, line 202: Just me being weird, but maybe change these to newWithoutPosition or something? I was wondering what number "No" referred to :D

root/static/scripts/tests/relationship-editor.js
```diff
@@ -158,21 +158,8 @@ function relationshipEditorTest(name, callback) {
...
-         * _.defer makes its target functions asynchronous. It is redefined
-         * here to call its target right away, so that we don't have to deal
-         * with writing async tests.
-         */
-        var _defer = _.defer;
-
-        _.defer = function (func) {
-            func.apply(null, Array.from(arguments).slice(1));
-        };
-
```
reosarevok: Does setTimeout not need any changes here to replace these?

reosarevok (commented):
Some comments (got up to e72862334e9a09f5c1776f607ab77a6c15dee1c5 / Replace _.reduce with native method so far)

root/static/scripts/tests/relationship-editor.js
mwiencek: IIRC there was only one test left that needed it, so I just made that test asynchronous (putting the `t.ok` or whatever inside a `setTimeout`).

root/static/scripts/release-editor/duplicates.js
```diff
@@ -133,35 +132,35 @@ function toggleLoadingIndicator(show) {
 }
 
 
-function pluck(chain, name) {
-    return chain.map(name).compact();
-}
```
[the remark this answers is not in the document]
mwiencek: Yeah, I'll look into adding a helper for that. But I'd probably implement it using `reduce` so that it only has to loop over the array once.

root/static/scripts/edit/components/edit/ReleaseEventsDiff.js
```diff
@@ -30,7 +30,7 @@ function areReleaseDatesEqual(a, b) {
   return formatDate(a.date) === formatDate(b.date);
 }
 
-const countryId = x => x.country ? x.country.id : null;
+const countryId = x => String(x.country?.id ?? null);
```
reosarevok, line 33: `getCountryId` since it's a function?

root/static/scripts/edit/externalLinks.js
```diff
@@ -44,6 +43,7 @@ type LinkStateT = {
 };
 
 type LinkHashT = {
+  __proto__: empty,
```
reosarevok, line 44: What's this for?

root/static/scripts/common/utility/arrays.js
```diff
@@ -111,7 +111,7 @@ export function sortByString<T>(
 export function groupBy<T>(
   array: $ReadOnlyArray<T>,
   func: (T) => string,
-): {__proto__: null, +[groupKey: string]: $ReadOnlyArray<T>} {
+): {__proto__: empty, +[groupKey: string]: $ReadOnlyArray<T>} {
```
reosarevok, line 136: Why? (sorry if these `__proto__` changes should be obvious, I've just never dealt with that at all AFAICT)

root/components/ReleaseLabelList.js
```diff
@@ -21,6 +21,8 @@ type ReleaseLabelsProps = {
   +labels?: $ReadOnlyArray<ReleaseLabelT>,
 };
 
+const getLabelGid = x => x.gid;
```
reosarevok, line 24: Why create a new const for one use?

root/static/scripts/common/utility/deepEqual.js
```diff
@@ -0,0 +1,83 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+const objectPrototype = Object.prototype;
+const hasOwnProperty = objectPrototype.hasOwnProperty;
+
+export default function deepEqual(
+  a: any,
+  b: any,
+  cmp?: (any, any) => boolean,
```
reosarevok, line 16: Any chance we can make them not `any` and this flow strict?

root/statistics/Index.js
```diff
@@ -576,22 +579,30 @@ const Index = ({
...
+                  </th>
+                  <td>{fc('medium.' + num + 'discids')}</td>
+                  <td>
+                    {fp('medium.' + num + 'discids', 'medium.has_discid')}
+                  </td>
+                </tr>,
+              );
+            }
+            return discIdRows;
+          })()}
```
reosarevok, line 605: These two could be the same function with an argument `release` or `medium`, couldn't they?

root/static/scripts/common/utility/arrays.js
```diff
@@ -122,6 +122,10 @@ export function groupBy<T>(
   }, Object.create(null));
 }
 
+export function head<T>(array: ?$ReadOnlyArray<T>): ?T {
+  return array?.length ? array[0] : undefined;
+}
+
```
reosarevok, line 150: Maybe I'm just being boring, but if the last is "last", why can't we name the first "first"? :p

root/static/scripts/work.js
```diff
@@ -268,6 +267,8 @@ function removeLanguage(i) {
 const LANG_MUL_ID = 284;
 const LANG_ZXX_ID = 486;
 
+const getSelectField = field => field;
```
reosarevok, line 270: Why not just have this directly in the attribute?

root/types.js
```diff
@@ -463,6 +463,7 @@ declare type EditT = {
     +votes: number,
   },
   +created_time: string,
+  +data: any,
```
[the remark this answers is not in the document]
mwiencek: I could make it `{[any]: any}`, actually. With https://github.com/metabrainz/musicbrainz-server/pull/1616 we'll be able to model `data` on every edit type. :)

root/static/scripts/edit/externalLinks.js
```diff
@@ -476,20 +476,22 @@ function isEmpty(link) {
...
         canRemove[index] = true;
+        canRemoveCount++;
       }
     }
   });
 
   if (emptyCount === 0) {
     return links.concat(newLinkState({relationship: _.uniqueId('new-')}));
-  } else if (emptyCount > 1 && _.size(canRemove)) {
+  } else if (emptyCount > 1 && canRemoveCount > 0) {
```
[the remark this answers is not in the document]
mwiencek: Adding a count variable didn't seem burdensome or affect readability at all, so I did that to avoid an extra allocation (since `Object.keys` has to build a new list of keys every time you call it).

root/static/scripts/common/utility/cloneDeep.js
```diff
@@ -0,0 +1,70 @@
+/*
+ * @flow
```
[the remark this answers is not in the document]
mwiencek: It'd be impossible, currently, I think. (I did make an attempt.) In fact a `clone` function is [the example they use](https://flow.org/en/docs/types/casting/#type-casting-through-any-) when demonstrating a use of `any`:
> It would be hard to create a type for this because we’re creating a new object based on another object.

root/static/scripts/common/components/TagEditor.js
```diff
@@ -249,7 +248,9 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
   flushPendingVotes(asap?: boolean) {
     const actions = {};
 
-    each(this.pendingVotes, item => {
+    ((Object.values(this.pendingVotes): any):
```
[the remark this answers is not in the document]
mwiencek: `Object.values` returns `Array<mixed>` IIRC. I could maybe replace this with a `Map` and get a better type though.

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -29,6 +28,8 @@ import {stripAttributes} from '../../edit/utility/linkPhrase';
 import isBlank from '../../common/utility/isBlank';
 import debounce from '../../common/utility/debounce';
 
+const hasOwnProperty = Object.prototype.hasOwnProperty;
```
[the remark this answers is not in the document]
mwiencek: Yeah, I was thinking of adding something like `hasOwnProp` to the ProvidePlugin config.
```JS
const objectHasOwnProperty = Object.prototype.hasOwnProperty;

function hasOwnProp(object, prop) {
  return objectHasOwnProperty.call(object, prop);
}
```

root/static/scripts/edit/externalLinks.js
mwiencek: It's possible to create objects that have a null prototype with `Object.create(null)`. That's useful if you're using them like Maps, because they won't have any conflicting properties in the prototype chain (methods like `hasOwnProperty` for example). But Flow doesn't like one object type has `__proto__: null` and another doesn't, so I'm typing it as `empty` instead to make it an error to use in any case, which Flow seems to be happy with.

mwiencek: Usually you'd use `new Map` to create a proper map, but `Object.create(null)` can be necessary if it's more convenient to use the map like a plain object. (Not sure if that's the case here, but I didn't want to refactor anything too much.)

root/static/scripts/common/utility/arrays.js
mwiencek: As explained above, it lets you use the result where `__proto__: null` would conflict with another object type that has a normal object prototype.

root/statistics/Index.js
mwiencek: Seems like it. Not sure I want to add more refactoring to this PR though since it's big enough as it is. :) Also, this comment is on an outdated diff, since I did end up adding a `mapRange` utility. So no real changes to this file at present.

root/static/scripts/work.js
```diff
@@ -266,21 +265,26 @@ function removeLanguage(i) {
...
       addLabel={l('Add Language')}
       getSelectField={identity}
-      hideAddButton={intersection(
-        form.field.languages.field.map(lang => String(lang.value)),
-        ['486', '284'],
-      ).length > 0}
+      hideAddButton={
+        selectedLanguageIds.includes(String(LANG_MUL_ID)) ||
+        selectedLanguageIds.includes(String(LANG_ZXX_ID))
+      }
```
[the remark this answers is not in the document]
mwiencek: The form data comes from Perl, so I don't expect them to be numbers in all cases. I can move the constants. :)

root/static/scripts/work.js
mwiencek: It's usually a bad idea to do that because the component could be using `React.memo`. The component would never be memoized, because you're passing a different function every time. (On HTML elements this doesn't matter.)

This function also doesn't depend on any value inside the scope of `renderWorkLanguages`, so there's no technical need to redefine it either.

root/static/scripts/common/components/TagEditor.js
```diff
@@ -38,9 +38,9 @@ const VOTE_DELAY = 1000;
 
 const getTagName = t => t.tag.name;
```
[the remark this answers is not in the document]
mwiencek: If the function is re-created often and is [always run on the same type of object](https://stackoverflow.com/a/37059056), I do like to make a constant for them, but in this case `createInitialTagState` should only be called once or a few times, so sure. :)

root/static/scripts/release-editor/recordingAssociation.js
```diff
@@ -306,7 +305,9 @@ function setSuggestedRecordings(track, recordings) {
     var lastRecording = track.recording.saved;
 
     if (!track.hasExistingRecording() && lastRecording) {
-        recordings = union([lastRecording], recordings);
+        recordings = [lastRecording].concat(
+            recordings.filter(x => x !== lastRecording),
+        );
```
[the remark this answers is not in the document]
mwiencek: Good question. It might be slightly faster but I don't think performance matters here. It does have to stay an array so maybe that was why, but `[...new Set([lastRecording, ...recordings])]` should be fine.

root/components/ReleaseLabelList.js
mwiencek: This way it can use the same (hopefully optimized, since it runs on the same type of object) function in memory every time the component is called, rather than creating a new function each time. (Probably a drop in the bucket and I'd do it the other way if it helped readability, but the named function actually seems more readable to me, too.)

root/static/scripts/common/utility/deepEqual.js
mwiencek: Not really, for the same reason implementing `clone` is hard without `any`. Flow refinement isn't good enough yet to support the kind of magic this does.

mwiencek: But luckily we shouldn't be needing to use this function much (outside of tests) in Reactified code, since we won't need to deeply-compare state objects there.

root/static/scripts/common/utility/formatDate.js
```diff
@@ -23,22 +22,22 @@ function formatDate(date: PartialDateT | null): string {
 
   if (nonEmpty(y)) {
     if (y < 0) {
-      result += '-' + padStart(String(-y), 3, '0');
+      result += '-' + String(-y).padStart(3, '0');
```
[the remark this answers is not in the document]
mwiencek: Added a `fixedWidthInteger` utility. It handles negative numbers too (here for example) by making sure the `-` stays in front of the zeroes. This actually changed the `formatDate` output slightly for negative years; it now outputs year -1 like `-0001` instead of `-001`. But I think that's fine?

reosarevok (approved):
I think this looks good as-is, but I haven't tested the site with it. We can throw it on test and ask people to do some mock editing if it seems helpful, or if the changes aren't too dangerous we can just do that in beta, given you've already tried locally.

yvanzo (approved):
Checked most of replacements and tested editing locally.

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -647,7 +647,7 @@ const RE = MB.relationshipEditor = MB.relationshipEditor || {};
                     maxLinkOrder = Math.max(maxLinkOrder, other.linkOrder.peek() || 0);
                 });
 
-                if (maxLinkOrder === 0 || !_.isFinite(maxLinkOrder)) {
+                if (maxLinkOrder === 0 || !Number.isFinite(maxLinkOrder)) {
```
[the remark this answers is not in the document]
mwiencek: Correct, that's what https://babeljs.io/docs/en/babel-preset-env does for us. It's configured in babel.config.js, reads our supported browsers from `BROWSER_TARGETS`, and injects any polyfills that are needed. Those come from core-js: the polyfill we get for `Number.isFinite` can be seen [here](https://github.com/zloirock/core-js/blob/6e3106817aa46919273b80d6f961e75a65967d22/packages/core-js/internals/number-is-finite.js).

root/static/scripts/timeline.js
```diff
@@ -231,8 +231,8 @@ class TimelineViewModel {
             let match;
 
             if ((match = part.match(/^(-)?([rv])-?$/))) { // trailing - for backwards-compatibility
-                const meth = match[2] === 'r' ? 'rate' : 'events';
-                self.options[meth](!(match[1] === '-'));
+                const method = match[2] === 'r' ? 'rate' : 'events';
+                self.options[method](!(match[1] === '-'));
```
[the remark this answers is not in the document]
mwiencek: Fair enough, I can replace both. :)

mwiencek:
@brainzbot test this please

PR #1672, opened 2020-09-02
Generalize IntentionallyRawIcon

reosarevok:
Sometimes it's not untranslated, but unformatted.

### DIFF
--- a/root/edit/components/IntentionallyRawIcon.js
+++ b/root/edit/components/IntentionallyRawIcon.js
@@ -12,13 +12,13 @@ import * as React from 'react';
 import InformationIcon
   from '../../static/scripts/edit/components/InformationIcon';
 
-const IntentionallyUntranslatedIcon =
+const IntentionallyRawIcon =
   (): React.Element<typeof InformationIcon> => (
     <InformationIcon
       className="align-top"
-      title={l(`This field is intentionally left untranslated
-                (i.e. just as entered) to facilitate edit review`)}
+      title={l(`This field is intentionally left as it was originally
+                entered (untranslated, unformatted).`)}
     />
   );
 
-export default IntentionallyUntranslatedIcon;
+export default IntentionallyRawIcon;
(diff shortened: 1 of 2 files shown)


root/edit/components/IntentionallyRawIcon.js
```diff
@@ -12,13 +12,13 @@ import * as React from 'react';
...
 
-const IntentionallyUntranslatedIcon =
+const IntentionallyRawIcon =
   (): React.Element<typeof InformationIcon> => (
     <InformationIcon
       className="align-top"
-      title={l(`This field is intentionally left untranslated
-                (i.e. just as entered) to facilitate edit review`)}
+      title={l(`This field is left as it was originally entered
+                (untranslated, unformatted) to facilitate edit review`)}
```
yvanzo, line 20: IMHO the end of the sentence is not useful (but to developers?), the sentence should be stopped still.

```suggestion
      title={l(`This field is left as it was originally entered
                (untranslated, unformatted).`)}
```

yvanzo (changes requested):

root/edit/components/IntentionallyRawIcon.js
mwiencek: Hm, I thought that was the point of the string. Since the user doesn't need a tooltip to see that it's untranslated & unformatted, just that it might not be obvious why (i.e. it's not a bug). +1 to the period!

reosarevok: Added the period. I also feel that the end of the sentence is the whole point of why we do this and seems sensible to mention :/

yvanzo: It’s obviously for edit review as it shows up for edit display only, and writing the obvious is unneeded. :wink:
Maybe just rephrase using _intentionally_ instead? e.g. “This field is intentionally shown as…).”?

reosarevok: I still think that's a worse string but I really want to merge this today, so whatever, let's do that.

yvanzo: Well, phrasing is not really my specialty either…

yvanzo (approved):
I feel familiar with this wording…

PR #1673, opened 2020-09-02
Polyfill improvements (also unbreaks IE11)

mwiencek:
While testing the lodash removal PR in IE11, I found that it was already broken in that browser even before the PR. So I started working on fixing that, since it's one of our supported browsers. Then I discovered some more disturbing stuff unrelated to IE11 (see commit "Disable core-js on plugin-transform-runtime"). That commit shaves about 502 kB off of common-chunks, or 37 kB after minification.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -30,9 +30,8 @@ module.exports = function (api) {
     '@babel/plugin-transform-react-jsx',
     '@babel/plugin-transform-react-constant-elements',
     ['@babel/plugin-transform-runtime', {
-      corejs: 3,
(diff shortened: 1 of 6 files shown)


reosarevok (approved):
Generally looks good to me, tested that it does indeed unbreak the menu in IE11 too at least.

yvanzo (approved):
LGTM but did not test with IE11.

mwiencek:
Rebased to fix conflicts. Will deploy to beta after tests pass.

PR #1674, opened 2020-09-03
MBS-11062: Link basic how-tos from the front page

reosarevok:
### Implement MBS-11062

The Quick Start section in the front page seems like a good place to direct people to the most basic editing how-tos: artists and releases. A general link to the how-to page is in the documentation menu, but having these two somewhere that might be more directly accessible to newcomers can't hurt.

While we have a lot more how tos, these two are the ones that seem the most useful for new users, and from the bottom of those two they can always find the links to the more advanced ones.

Screenshot from 2020-09-03 14-18-09

### DIFF
--- a/root/main/index.js
+++ b/root/main/index.js
@@ -178,6 +178,12 @@ const Homepage = ({
               <li>
                 <a href="/doc/Frequently_Asked_Questions">{l('FAQs')}</a>
               </li>
+              <li>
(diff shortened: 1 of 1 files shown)


mwiencek (approved):
I feel like there should be a rule where for every link you add to the homepage you have to remove two. ;)

yvanzo (approved):

PR #1675, opened 2020-09-03
MBS-11061: Disallow MusicBrainz URLs in relationships

reosarevok:
### Implement MBS-11061

I checked and there were only three more-or-less legitimate cases of this (among a larger amount of stuff clearly pasted in error such as links to the same entity the URL was added to). One of those was a link to our wiki summit page as an "official website" link for the Summit 2014, where some music got recorded (and arguably that is not an official website anyway) and the other two were artists using their own MusicBrainz user profile for a bio and then linking to their user page as their biography page (which is something I think we should disencourage anyway and feels very spammy, while being no better than just adding the same text to an annotation).

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -237,6 +237,9 @@ export class ExternalLinksEditor
               error = l('Required field.');
             } else if (!isValidURL(link.url)) {
               error = l('Enter a valid url e.g. "http://google.com/"');
+            } else if (isMusicBrainz(link.url)) {
+              error = l(`Links to MusicBrainz URLs are not allowed.
+                         Did you mean to paste something else?`);
             } else if (isShortened(link.url)) {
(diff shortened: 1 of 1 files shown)


mwiencek (approved):
The summit one doesn't sound that wrong to me, but since that's the only real use case in the whole DB, sure. :)

root/static/scripts/edit/externalLinks.js
```diff
@@ -612,6 +615,10 @@ function isShortened(url) {
   });
 }
 
+function isMusicBrainz(url) {
+  return /^https?:\/\/([^/]+\.)?musicbrainz.org/.test(url);
```
mwiencek, line 619: not that it matters unless someone tries to link to musicbrainzzorg but nitpick
```suggestion
  return /^https?:\/\/([^/]+\.)?musicbrainz\.org/.test(url);
```

yvanzo (approved):

PR #1676, opened 2020-09-03
MBS-11064: Mark ended links as such in the URL editor

reosarevok:
### Implement MBS-11064

This adds a message explaining that ended links are marked as ended, and as such not considered current, and should probably not be removed.

The color used is the information color from bootstrap, since I wasn't sure what would be a good pick.

Screenshot from 2020-09-03 16-21-36

@yvanzo mentioned as an alternative not showing the ended links at all in the editor. I personally feel that's a bit too hardcore, but it might be safer!

### DIFF
--- a/root/static/styles/colors.less
+++ b/root/static/styles/colors.less
@@ -40,6 +40,7 @@
 @light-border-background: @very-light-border;        // the background for the light-border tables
 
 @warning-background: #FFFFD0;                        // "warning" background
+@warning-text: #17A2B8;                              // "warning" text message color
 
 @positive-bg: #B1EBB0;                               // background color for positive and success states
 @positive-light-bg: #E4FBE4;                         // lighter version
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -33,6 +33,7 @@ import * as URLCleanup from './URLCleanup';
 import validation from './validation';
 
 type LinkStateT = {
+  ended: boolean,
   relationship: number | string | null,
   type: number | null,
   url: string,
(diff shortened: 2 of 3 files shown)


mwiencek:
If the idea is just to warn people not to remove these, then this seems fine (although the text color is unreadable to my eyes :P).

I wouldn't completely hide the data from people in the editor, at least not where they have to go to a different page to remove them, but perhaps have them hidden by default and add a checkbox to make the historical/obsolete links visible for editing.

root/static/styles/colors.less
```diff
@@ -40,6 +40,7 @@
 @light-border-background: @very-light-border;        // the background for the light-border tables
 
 @warning-background: #FFFFD0;                        // "warning" background
+@warning-text: #17A2B8;                              // "warning" text message color
```
mwiencek, line 43: Maybe just use black text with `@warning-background` for the box color?

zas:
Really not sure about what to do, but it seems to me letting editor removing a link marked "ended" kills the whole purpose of "ending" a link.

Hiding them by default in the edit page (with a way to show them) as suggested by yvanzo is ok for me, since they are hidden elsewhere too.

Perhaps ask on forums about that.

yvanzo (commented):
It currently is a PITA for advanced editors to end links, while it is pretty fast to mess them either by clicking the button “Remove Link” or modifying the related URL which actually replaces the current URL entity with a new one (not ended). That is why I would prefer hiding them like in other places rather than leaving a button “Nuke it” with a gentle explanation about why not to press it. Even when the URL relationship field will be improved to handle “ended” flag, it should still make harder to modify or remove ended URLs.

reosarevok:
@y-young has done a good job of showing this (and allowing editing it!) so this is no longer needed.

PR #1678, opened 2020-09-04
Fix unspecified RG type

reosarevok:
With the lodash changes for groupBy, the type here is no longer null but empty string, so currently no type string is shown at all.


mwiencek (approved):

mwiencek:
@brainzbot test this please

PR #1679, opened 2020-09-04
MBS-10835: Don't send verification mail if address in use

reosarevok:
### Implement MBS-10835

To avoid email reuse, this only sends a verification email if the address isn't being used by any other editor.

For cases where the user has a good reason to reuse the address (such as bots) they can let us know and we can manually override the whole thing by skipping verification.

I don't have email sending setup locally. If this seems generally fine, I can throw it on test and try from there, but before that let's make sure I didn't do anything too dumb.


yvanzo (approved):
LGTM

mwiencek (approved):

PR #1681, opened 2020-09-07
Eslint indent auto-fixes

reosarevok:
indent: [warn, 2, {CallExpression: {arguments: first}, SwitchCase: 1}]
https://eslint.org/docs/rules/indent

Autofixes + hand-fixing the comment blocks and nested bits the autofix ignores.


root/components/FormRowSelectList.js
```diff
@@ -28,54 +28,55 @@ type Props<S> = {
...
+    label,
+    onAdd,
+    onEdit,
+    onRemove,
+    options,
+    removeClassName,
+    removeLabel,
+    repeatable,
+  }: Props<S>): React.Element<typeof FormRow> => (
+    <FormRow>
```
reosarevok, line 46: This is the only way I found to make both indent and jsx-indent happy at the same time.

root/taglookup/Results.js
```diff
@@ -15,17 +15,18 @@ import TagLookupForm from './Form';
...
-      {props.children}
-      <TagLookupForm form={props.form} />
-    </div>
-  </Layout>
-);
+const TagLookupResults = <T>
+  (
+    props: TagLookupResultsPropsT<T>,
+  ): React.Element<typeof Layout> => (
+    <Layout $c={props.$c} fullWidth title={l('Tag Lookup Results')}>
```
reosarevok, line 22: This is the only way I found to make both indent and jsx-indent happy at the same time.

reosarevok:
Aside from the two things I commented, pretty much everything else here should be space changes only, except for a couple places where I just split things in two lines or whatever to avoid the new indent making lines too long.

mwiencek:
I pushed a commit which ignores some nodes that are either handled by the jsx-indent rule already or seem off. Been meaning to do that for a while, since those two rules conflict. That should fix the two cases you mentioned (so I reverted those changes).

mwiencek:
Also, I've bumped all the Babel/ESLint dependencies just in case they fixed any bugs that'd affect the `--fix` output. It did make one change in recordingAssociation.js that seems correct (the `var appearsOn = uniqBy(...)` call was over-indented before).

.eslintrc.yaml
```diff
@@ -174,7 +174,7 @@ rules:
   func-call-spacing: [warn, never]
   function-paren-newline: [warn, consistent]
   implicit-arrow-linebreak: [warn, beside]
-  indent: [warn, 2, {CallExpression: {arguments: first}, SwitchCase: 1}]
+  indent: [warn, 2, {CallExpression: {arguments: first}, SwitchCase: 1, ignoredNodes: ["JSXElement", "ArrowFunctionExpression"]}]
```
yvanzo, line 177: Does it just ignore these lines or does it enforce the indent style mentioned in the commit message?

yvanzo (approved):
This one was quick to review thanks to _hide whitespace changes_!

Is there any plan to make ESLint validation part of CI checks?
Would it be alright to have a script for the rules that have been auto-fixed so far?

.eslintrc.yaml
reosarevok: @mwiencek ^

reosarevok:
"Is there any plan to make ESLint validation part of CI checks?" I talked to @mwiencek once about having them auto-checked, but maybe once they're all fixed - of course, maybe it is doable (and sensible) to do it bit by bit.

yvanzo:
They might never get all fixed because we don’t auto-check that changes being made don’t break rules followed so far.

mwiencek:
I went to test this again, and to my surprise, the eslint config had errors. Not sure how it didn't show those when I tested it previously. Anyway, those should be fixed now. See the updated https://github.com/metabrainz/musicbrainz-server/pull/1681/commits/0d86d30a1b6f1fa0d2be05550e06a9f2a910cdb8

.eslintrc.yaml
mwiencek: It ignores those nodes. `JSXElement` nodes are handled separately by the react/jsx-indent rule, so that's intended. Ignoring `ArrowFunctionExpression` means it won't find any indentation errors on those, but I believe that's better than it forcing us to make unnecessary changes.

mwiencek (approved):

mwiencek:
Merging this ASAP 'cause I have a feeling it's going to cause hundreds of conflicts in my linked-entities-refactor branch and I don't want to fix those twice. :)

PR #1682, opened 2020-09-07
[WIP] MBS-10321: Convert timeline to React

mwiencek:
Replaces flot with d3. This requires handling some things previously done automatically by flot or plugins, but it's not too bad.

TODO:

- [ ] add description to event tooltips
- [ ] tooltip styling
- [ ] zooming


reosarevok:
Why does it replace flot with d3? (I mean, d3 does seem to be much more popular, but wondering what specifically didn't work)

mwiencek:
> Why does it replace flot with d3? (I mean, d3 does seem to be much more popular, but wondering what specifically didn't work)

Mostly because I'm more familiar with it (and it's more popular, so more people in general are likely familiar with it, making it hopefully easier to maintain in the long run).

Also, the flot timeline is using canvas instead of SVG, which is good for performance but makes accessibility improvements almost impossible. I know you can use flot with both canvas and SVG, but since I was changing it anyway I went with what I was familiar with.

It also doesn't depend on jQuery which is nice.

root/statistics/stats.js
```diff
@@ -1,4 +1,5 @@
 /*
+ *
```
reosarevok, line 2: Intentional?

In this file, please do check the translations for labels such as "l('Failed edits (voted down)')". They're not getting translated, and neither do they if I change that to l_statistics (which seems like the right call).

PR #1687, opened 2020-09-09
MBS-11085 / MBS-11089: Fix canHaveDiscID check / Fix missing favicons

reosarevok:
### Fix MBS-11085 / MBS-11089

MBS-11085: 
This was changed with cede01b1342c553ceb91e790f61eedcff26db5a8 - I assume under the assumption that MB.formatsWithDiscIDs is an object. It is not, though, it's just an array of IDs, so this was checking whether the position in the array matching the format ID was null. 
Changed to use includes, as it did before (just the native version).

MBS-11089:
While making the file flow strict-local I forgot to change this into a ternary, so it was being passed a boolean when className exists.

### DIFF
--- a/root/layout/components/ExternalLinks.js
+++ b/root/layout/components/ExternalLinks.js
@@ -60,7 +60,7 @@ const ExternalLink = ({
   }
 
   return (
-    <li className={nonEmpty(className) || faviconClass(url)}>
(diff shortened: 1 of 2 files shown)


yvanzo:

yvanzo (approved):
Over-tested locally. :rocket: :shipit:

PR #1688, opened 2020-09-09
MBS-11072: Improve GenerateSQLScripts.pl usability

yvanzo:
# Problem

[MBVM-60](https://tickets.metabrainz.org/browse/MBVM-60): Output of `setup-amqp-triggers` is misleading and incomplete

# Solution

MBS-11072:
- Self-document `GenerateSQLScripts` with `-h` option
- Add an option to specify a limited set of files to be parsed
- Improve output to log about successfully parsed files

# Checklist for author

* [X] Locally tested with `setup-amqp-triggers`

# Action

1. Merge the related PR to `musicbrainz-docker` when upgrading to the next release of MusicBrainz Server.

### DIFF
--- a/admin/GenerateSQLScripts.pl
+++ b/admin/GenerateSQLScripts.pl
@@ -4,6 +4,57 @@ use utf8;
 use warnings;
 use strict;
 
+use Getopt::Long qw( GetOptions );
+use Pod::Usage qw( pod2usage );
+
+################################################################################
+
+=head1 NAME
+
+GenerateSQLScripts.pl - Generate SQL DROP scripts for each script in DIRECTORY
+
+=head1 SYNOPSIS
+
+GenerateSQLScripts.pl [options] [--] [DIRECTORY]
+
+Generate a corresponding SQL DROP script for each SQL CREATE script in DIRECTORY
+(admin/sql/ by default).
+
+Options:
+
+    -h, --help                          show this help
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2009 Lukas Lalinsky
+Copyright (C) 2010 MetaBrainz Foundation
+Copyright (C) 2012 Aurélien Mino
+
+This file is part of MusicBrainz, the open internet music database,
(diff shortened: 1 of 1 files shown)


mwiencek:
I'm confused. Why does the Docker setup need to run `GenerateSQLScripts.pl` for sir when the Drop* scripts for it are [already stored in its repo](https://github.com/metabrainz/sir/tree/master/sql)? It even has its own [GenerateDropSql.pl](https://github.com/metabrainz/sir/blob/master/GenerateDropSql.pl) which shouldn't have those misleading messages. Am I missing something?

yvanzo:
Indeed, it seems that `GenerateSQLScripts.pl`’s call from SIR’s [`Makefile`](https://github.com/metabrainz/sir/blob/v1.0.2/Makefile#L15) predates the script `GenerateDropSql.pl` which is a derivative simplified version of `GenerateSQLScripts.pl`.

If I recall correctly, SIR used to call scripts directly from MBS, assuming it was installed in the same system, which is not the case with separate Docker containers.

Also I did not find any other call to `GenerateSQLScripts.pl` in our repositories even though it seems it may have been used for CAA in the past? The Git history doesn’t really explain the rationale of this script either. Is it solely for use by maintainers during schema change or something like that?

mwiencek:
> Also I did not find any other call to GenerateSQLScripts.pl in our repositories even though it seems it may have been used for CAA in the past? The Git history doesn’t really explain the rationale of this script either. Is it solely for use by maintainers during schema change or something like that?

Yeah, it's for regenerating scripts after schema changes. (The CAA schema is in this repo.) Which is mostly why it only logs when it didn't find something - that could indicate an error, whereas seeing what scripts it generated is just `git diff` / `git status`.

reosarevok:
So should we stop using this file in sir instead?

mwiencek:
We use it in musicbrainz-docker, not sir. But musicbrainz-docker should be able to use the scripts in https://github.com/metabrainz/sir/tree/master/sql AFAICT and not have to generate anything. The documentation improvements here make sense at least.

PR #1654, opened 2020-08-18
MBS-11041: Make ModBot leave a note on autoremoval edits

reosarevok:
### Implement MBS-11041

These edits often confuse users, who don't understand why their added entity has been removed without a word by some weird editor. This adds an explanatory edit note which should hopefully avoid those situations and help them understand they can readd the entity if needed.

Screenshot from 2020-08-18 15-25-18

@mwiencek: please let me know if localized_note would not work here for some reason :)

PR #1690, opened 2020-09-10
Amend d5e3f84e62 to fix localizable string

yvanzo:
# Problem

Since https://github.com/metabrainz/musicbrainz-server/pull/1654, the generated message contained line breaks and indenting spaces.

# Solution

Split into strings and concatenate with `.` operator.


reosarevok (approved):
Sorry, and thanks!

yvanzo:
Successfully tested string extraction locally, thanks for the double-check!

PR #1696, opened 2020-09-16
Enforce ESLint rules followed so far

yvanzo:
# Problem

A lot of efforts have been made to make the code follow some ESLint rules already, but nothing prevents new code to break them again. But `estlint` cannot be run with all rules currently defined in `.eslintrc` and not followed by code prior to these rules.

# Solution

Create a script that checks only the rules followed so far. This script is for continuous integration and to be run manually for convenience when creating new code. It follows PR #1681 which enforced `indent` rule but for `script/xgettext.js`.

# Action

1. Add other rules that have already been applied, if any.
2. Gradually add more rules when fixing code to follow them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,7 @@ jobs:
           sudo -E -H -u musicbrainz make -C po all_quiet deploy
           NODE_ENV=test WEBPACK_MODE=development sudo -E -H -u musicbrainz carton exec -- ./script/compile_resources.sh server web-tests
           sudo -E -H -u musicbrainz ./node_modules/.bin/flow --quiet
+          sudo -E -H -u musicbrainz ./node_modules/.bin/eslint --max-warnings 0 .
           rm /etc/service/chrome/down && sv start chrome
           sudo -E -H -u musicbrainz carton exec -- perl \
               -I lib \
--- a/.eslintrc.unfixed.yaml
+++ b/.eslintrc.unfixed.yaml
@@ -1,0 +1,105 @@
+rules:
+  no-await-in-loop: error
+  no-extra-semi: warn
+  no-unused-vars: error
+  consistent-return: error
+  eqeqeq: [warn, smart]
+  no-else-return: warn
+  no-multi-spaces: [error, {ignoreEOLComments: true}]
+  wrap-iife: warn
+  array-element-newline: [warn, consistent]
+  camelcase: [warn, {properties: never,
+                     allow: [
+                       "l_attributes",
+                       "ln_attributes",
+                       "lp_attributes",
+                       "l_countries",
+                       "ln_countries",
+                       "lp_countries",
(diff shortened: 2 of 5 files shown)


script/check_enforced_eslint_rules
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+cd "$(dirname "${BASH_SOURCE[0]}")/../" || exit 66 # EX_NOINPUT
+
+./script/check_eslint_rule \
+  'indent: [warn, 2, {CallExpression: {arguments: first}, SwitchCase: 1, ignoredNodes: ["JSXElement", "ArrowFunctionExpression"]}]' \
+  --rule 'react/jsx-indent: [error, 2]' \
+  --rule 'react/jsx-indent-props: [error, 2]' \
+  --max-warnings 0 \
+  .
```
mwiencek, line 30: Thanks! This will be very useful to have.

I wish we didn't have to duplicate all the rule configuration here. There've already been several times where I forgot to update configuration in check_eslint_rule. :(

What do you think about just commenting out all the rules in .eslintrc.yaml we haven't fixed yet? Then we can uncomment them as we verify they're fixed. The CI command becomes as simple as running `./node_modules/.bin/eslint`.

yvanzo: This new script just checks the bare minimum to pass CI tests, but I still use the full `\.eslintrc` when creating new files.

mwiencek: Right, but eventually it'll duplicate a lot of rules (otherwise we'd just fix them all at once).

What about having two configs, `.eslintrc.yaml` and `.eslintrc.unfixed.yaml`? You can specify the additional config like `-c .eslintrc.unfixed.yaml` and eslint will merge them.

yvanzo: Adopted it and added a wrapper script for checking extra rules by hand.

reosarevok: Would this two-file setup also allow us to have the Github auto-check check PRs for fixed rules?

yvanzo: Yes but CircleCI does the job, see commit https://github.com/metabrainz/musicbrainz-server/pull/1696/commits/be6616aea2ff1c11033da0712b4d68258477d9c2.

reosarevok (approved):
LGTM. @mwiencek: do you know if vscode can be told to follow two different eslint files?

mwiencek:
> LGTM. @mwiencek: do you know if vscode can be told to follow two different eslint files?

Yes. Go to the eslint extension settings -> Eslint: Options -> Edit in settings.json, and put

```
    "eslint.options": {
        "configFile": ".eslintrc.unfixed.yaml"
    }
```

Then restart vscode to take effect

mwiencek (approved):

PR #1697, opened 2020-09-16
Beta (MBS-11069): Avoid code formatting within white space sensitive area

kellnerd:
# Problem

Too many leading and trailing white spaces are rendered for the positive part of the diff for changed recordings, see the [comment on the ticket](https://tickets.metabrainz.org/browse/MBS-11069?focusedCommentId=55354&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-55354).
This is a regression of #1691 which changed the style of highlighted differences to always render white spaces inside their `<span>` elements.

# Solution

Move the code of the template that is not generating output out of the `<span>` element to avoid code formatting within the white space sensitive area.

# Action

This change needs to be verified again on a test server as I haven't installed the server on my machine.

### DIFF
--- a/root/edit/details/edit_medium.tt
+++ b/root/edit/details/edit_medium.tt
@@ -199,11 +199,11 @@
                  </td>
                  <td><span class="diff-only-a">[% IF old_track.recording; descriptive_link(old_track.recording); END %]</span></td>
                  <td>
-                   [% # If no recording_id exists, it's creating a recording
+                   [%~ # If no recording_id exists, it's creating a recording
                      SET allow_new = 1 IF !new_track.recording_id
-                   %]
+                   ~%]
                    <span class="diff-only-b">[% descriptive_link(new_track.recording) %]</span>
-                   [% SET allow_new = 0 IF !new_track.recording_id %]
+                   [%~ SET allow_new = 0 IF !new_track.recording_id ~%]
                  </td>
                </tr>
           [% END %]
--- a/root/edit/details/edit_medium.tt
+++ b/root/edit/details/edit_medium.tt
@@ -199,13 +199,11 @@
                  </td>
                  <td><span class="diff-only-a">[% IF old_track.recording; descriptive_link(old_track.recording); END %]</span></td>
                  <td>
-                   <span class="diff-only-b">
-                     [% # If no recording_id exists, it's creating a recording
-                       SET allow_new = 1 IF !new_track.recording_id
-                     %]
-                     [% descriptive_link(new_track.recording) %]
-                     [% SET allow_new = 0 IF !new_track.recording_id %]
-                   </span>
+                   [% # If no recording_id exists, it's creating a recording
+                     SET allow_new = 1 IF !new_track.recording_id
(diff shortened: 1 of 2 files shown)


reosarevok:
If this is fixing a problem with code currently in beta, can you point it against beta instead of master? :)

root/edit/details/edit_medium.tt
```diff
@@ -199,13 +199,11 @@
...
-                       SET allow_new = 1 IF !new_track.recording_id
-                     %]
-                     [% descriptive_link(new_track.recording) %]
-                     [% SET allow_new = 0 IF !new_track.recording_id %]
-                   </span>
+                   [% # If no recording_id exists, it's creating a recording
+                     SET allow_new = 1 IF !new_track.recording_id
+                   %]
+                   <span class="diff-only-b">[% descriptive_link(new_track.recording) %]</span>
+                   [% SET allow_new = 0 IF !new_track.recording_id %]
```
yvanzo, line 206: To remove whitespaces and line breaks inside the `td` tag:

```suggestion
                   [%~ # If no recording_id exists, it's creating a recording
                     SET allow_new = 1 IF !new_track.recording_id
                   ~%]
                   <span class="diff-only-b">[% descriptive_link(new_track.recording) %]</span>
                   [%~ SET allow_new = 0 IF !new_track.recording_id ~%]
```

See [chomp tag modifiers in Template Toolkit](http://template-toolkit.org/docs/manual/Config.html#section_PRE_CHOMP_POST_CHOMP).

That would work with the original code too.

yvanzo (commented):
I’m not sure why commit https://github.com/metabrainz/musicbrainz-server/pull/1697/commits/97d51cff6a5205fbd47fc929554ec6a88b33c7fb appears here whereas that change has already been merged with the previous pull request as commit f5c567fb158c87c57c0b8ae94539dde84656c99d. The second commit works for me even though more unneeded white spaces and line breaks can be removed, see suggestion.

kellnerd:
Thank you for the suggestion @yvanzo. As I have zero experience with Perl and Template Toolkit I had just used the "most obvious" way for me to fix this, but now I can see that these modifiers work similar as in other template libraries I have used in the past.

I was also surprised that this PR still shows the old commit from the previous PR. Should I have used a new branch for this one or is there anything else I should do now?

yvanzo:
> Should I have used a new branch for this one or is there anything else I should do now?

Oh, indeed, that is because you used the same branch. Your commit https://github.com/metabrainz/musicbrainz-server/pull/1697/commits/97d51cff6a5205fbd47fc929554ec6a88b33c7fb is not the same as the merged commit f5c567fb158c87c57c0b8ae94539dde84656c99d because we used the “Squash and merge” strategy (which has the benefit of saving a merge commit but is more often causing problems than solving any).

yvanzo (approved):
Thanks for fixing it anyway, I guess we can just “squash and merge” this pull request too.

mwiencek (approved):

PR #1698, opened 2020-09-17
Improve Flow error suppressions

mwiencek:
We can make error suppressions more granular by specifying which exact error code we're ignoring. (Sort of like specifying which exceptions to catch.)

https://flow.org/en/docs/errors/

In some cases I removed the suppression altogether by addressing the underlying issue, if it seemed reasonable enough to.

### DIFF
--- a/root/static/scripts/account/components/EditProfileForm.js
+++ b/root/static/scripts/account/components/EditProfileForm.js
@@ -139,8 +139,15 @@ class EditProfileForm extends React.Component<Props, State> {
     e: SyntheticEvent<HTMLSelectElement>,
     languageIndex: number,
   ) {
-    // $FlowFixMe ~ string incompatible with FluencyT's string literals
-    const selectedFluency: FluencyT = e.currentTarget.value;
+    const selectedValue = e.currentTarget.value;
+    let selectedFluency: FluencyT | null = null;
+    switch (selectedValue) {
+      case 'basic':
+      case 'intermediate':
+      case 'advanced':
+      case 'native':
+        selectedFluency = selectedValue;
+    }
     this.setState(prevState => mutate<State, _>(prevState, newState => {
       const compound = newState.form.field.languages.field[languageIndex];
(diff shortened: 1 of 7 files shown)


root/static/scripts/account/components/EditProfileForm.js
```diff
@@ -139,8 +139,15 @@ class EditProfileForm extends React.Component<Props, State> {
...
-    const selectedFluency: FluencyT = e.currentTarget.value;
+    const selectedValue = e.currentTarget.value;
+    let selectedFluency: FluencyT | null = null;
+    switch (selectedValue) {
+      case 'basic':
+      case 'intermediate':
+      case 'advanced':
+      case 'native':
+        selectedFluency = selectedValue;
+    }
```
mwiencek, line 150: This error wasn't wrong TBH, since the string we get from the HTML could be anything (for example, we could introduce a bug where the `value` attributes are incorrect). But maybe that's unlikely and the `switch` statement is overkill.

root/static/scripts/common/utility/parseIntegerOrNull.js
```diff
@@ -9,7 +9,7 @@
 
 import parseInteger from './parseInteger';
 
-export default function parseIntegerOrNull(str: string): $FlowFixMe | null {
+export default function parseIntegerOrNull(str: string): number | null {
```
mwiencek, line 12: This was just an outright bug. Concerning since this module is marked as strict.

root/static/scripts/account/components/EditProfileForm.js
mwiencek: A better solution might be to just change the type to `FieldT<string | null>,`.

reosarevok: I don't think this hurts, at least. We *want* this to be FluencyT or null, so it's probably better to make that explicit.

reosarevok (approved):

yvanzo (approved):

PR #1699, opened 2020-09-18
MBS-11112: Convert /main error pages to React

reosarevok:
### Implement MBS-11112

See each commit for details.


mwiencek:
Converting the 500 error page should probably wait for https://github.com/metabrainz/musicbrainz-server/pull/1739 since even the .tt version barely works at all at present.

reosarevok:
AFAICT this is now ready.

root/main/error/components/ErrorEnvironment.js
```diff
@@ -0,0 +1,65 @@
...
+
+import * as React from 'react';
+
+type Props = {
+  +$c: CatalystContextT,
+  +hostname?: string,
+  +useLanguages?: boolean,
+};
+
+const ErrorEnvironment = ({
```
yvanzo, line 18: Commit https://github.com/metabrainz/musicbrainz-server/pull/1699/commits/faf46ccac63eaa0d97b71c59dc05a63cdcb418e3 gives no indication of what this is being converted from.

root/main/error/components/ErrorInfo.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+import * as React from 'react';
+
+type Props = {
+  +formattedErrors?: $ReadOnlyArray<string>,
+  +message?: string,
+};
+
+const ErrorInfo = ({
```
yvanzo, line 17: Same here, I just found it comes from `/root/main/info/error.tt` but commit https://github.com/metabrainz/musicbrainz-server/pull/1699/commits/b5ef9f12c395ef4135313c62470437d5fdb48c08 gave no indication.

yvanzo (changes requested):
Please clarify the origin of converted files in commit messages, I did not review them all.

root/main/error/components/ErrorEnvironment.js
reosarevok: Fair enough, added the info to the commit message

root/main/error/500.js
```diff
@@ -0,0 +1,97 @@
...
+    <ErrorEnvironment
+      $c={$c}
+      hostname={hostname}
+      useLanguages={useLanguages}
+    />
+
+    <p>
+      {l(`We’re terribly sorry for this problem.
+          Please wait a few minutes and repeat
+          your request — the problem may go away.`)}
```
yvanzo, line 79: This will obsolete existing translations for the previous message.

root/main/error/500.js
```diff
@@ -0,0 +1,97 @@
...
+
+type Props = {
+  +$c: CatalystContextT,
+  +edits?: $ReadOnlyArray<EditT>,
+  +formattedErrors?: $ReadOnlyArray<string>,
+  +hostname?: string,
+  +useLanguages: boolean,
+};
+
+const Error500 = ({
```
yvanzo, line 28: Shouldn’t filename match component’s name like we did so far? That is `Error500.js` instead of `500.js`.

root/main/error/components/ErrorEnvironment.js
yvanzo: Thank you!

root/main/error/500.js
reosarevok: I just followed what we did for 404, but I'd prefer this too. I'll change it in an extra commit.

root/main/error/500.js
reosarevok: React would otherwise just print the &#x2014; unchanged. I guess we could avoid that by using expand in some way, but maybe it's not such a big issue?

yvanzo: That’s why, thanks. No it isn’t a big issue.

yvanzo (approved):
Thanks!

lib/MusicBrainz/Server/Controller/Root.pm
```diff
@@ -166,7 +166,10 @@ sub error_404 : Private {
 
     $c->response->status(404);
     $c->stash->{current_view} = 'Node';
-    $c->stash->{component_props}{message} = $message;
+    $c->stash->{component_props} = {
+        message => $message,
+        template => 'main/error/404',
```
mwiencek: This change seems unrelated to the commit and possibly a no-op (re the `template` addition), unless I'm missing something.

mwiencek (commented):
This looks ready for merge, just wondering about the comment on `template => 'main/error/404',` below.

lib/MusicBrainz/Server/Controller/Root.pm
reosarevok: I'm moving the file from main/404 to main/error/404, wouldn't that require specifying the template?

reosarevok: I was being an idiot, fixed that now.

mwiencek (approved):

PR #1700, opened 2020-09-19
MBS-11117: Create report for too long discIDs

loujine:
# Problem

MBS-11117: new report about discid durations above 100 minutes, that should be fixed or removed.
cf https://community.metabrainz.org/t/sacd-types-and-discid/494807/8

# Solution

cf screenshot on MBS-11117

# Action

There are a few things that need a serious review and will probably need rework:

* the report could be applied on releases but it makes more sense to have it on discids directly. Since there was no report on cdtocs yet, I had to create a lib/MusicBrainz/Server/Report/CdTocReport.pm ~~that just passes the parameters of the report SQL query. Not sure it would make more sense to pass the 'id' there, since we don't have gid on the cdtoc table.~~ (Edit: I had to use "id" in the end in order to use CDTocLink)

~~* CDTOCs are not defined with React yet. I didn't want to modify EntityLink so I just defined the href=/cdtoc/... link in root/report/components/CdTocList.js, but that should probably be changed for other conversions to React.~~

* I pass durations from SQL in seconds and convert to milliseconds in JS to use formatTrackLength. Exporting durations in millisecond in SQL would be over the 2**31 integer limit in postgres

* I don't speak Perl so I just copy-pasted lib/MusicBrainz/Server/Report/CdTocReport.pm, I'm not sure what's inside is correct

### DIFF
--- a/lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
+++ b/lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
@@ -1,0 +1,42 @@
+package MusicBrainz::Server::Report::CDTOCDubiousLength;
+use Moose;
+
+with 'MusicBrainz::Server::Report::CDTOCReport';
+
+sub table { 'cd_toc_dubious_length' }
+sub component_name { 'CDTocDubiousLength' }
+
+sub query {
+    q{
+        SELECT
+            cdtoc.id AS cdtoc_id,
+            medium_format.name AS format,
+            cdtoc.leadout_offset / 75 AS length, -- seconds
+            row_number() OVER (
+                ORDER BY medium_format.name, cdtoc.leadout_offset DESC)
+        FROM
+            cdtoc
+            JOIN medium_cdtoc ON medium_cdtoc.cdtoc = cdtoc.id
+            JOIN medium ON medium_cdtoc.medium = medium.id
+            JOIN medium_format ON medium.format = medium_format.id
+        WHERE
+            leadout_offset > 75 * 60 * 100 -- cutoff 100 minutes
+        ORDER BY
+            medium_format.name,
+            cdtoc.leadout_offset DESC
+    }
+}
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2020 Jerome Roy
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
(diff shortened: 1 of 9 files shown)


root/report/components/CdTocList.js
```diff
@@ -0,0 +1,71 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          {items.map((item, index) => {
+            return (
+              <tr className={loopParity(index)} key={item.cdtoc_id}>
+                {item.discid ? (
+                  <>
+                    <td>
+                      <a href={'/cdtoc/' + item.discid}>
```
loujine, line 46: This could be integrated into EntityLink instead

reosarevok: See CDTocLink

loujine: Uh, not sure why I didn't see it. Fixing now

loujine: BTW https://github.com/metabrainz/musicbrainz-server/blob/master/root/types.js#L753 mentions an undefined MediumCDToc that you created last month, you want me to fix it at the same time?

reosarevok: In a separate commit, but sure! (if it doesn't break anything else)

loujine: Just pushed, let's see if tests complain

yvanzo (commented):
It is currently correctly creating and filling the report table but it fails to display it on error:

```
TypeError: Cannot read property 'total_entries' of undefined
    at CDTocDubiousLength (webpack:///./root/report/CDTocDubiousLength.js?:113:18)
```

I suspect the code around `inflate_rows` did not find the cause yet.

reosarevok:
Pushed a rebased fix. Also changed the Perl files to use CDTOC as we do elsewhere :)

reosarevok (approved):

lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
```diff
@@ -0,0 +1,42 @@
+package MusicBrainz::Server::Report::CDTOCDubiousLength;
+use Moose;
+
+with 'MusicBrainz::Server::Report::CDTOCReport';
+
+sub table { 'cd_toc_dubious_length' }
```
yvanzo, line 6: I guess that was the missing bit.

yvanzo (approved):
Works flawlessly. Thanks for having fixed the letter casing too.

lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
reosarevok: I think the main issue was the link generated from the reports index was wrong, too :D But a bit of everything maybe :)

PR #1702, opened 2020-09-21
MBS-11107: Support nameVariation for other entity types

reosarevok:
### MBS-11107

We can use relationship credits with most entity types. As such, it makes sense to supply the same hover info + span class name for them as we currently do for artists and recordings.

I special-cased URLs since I expect we don't really want to have hover features for them if we print a pretty-name and not the URL.

### DIFF
--- a/root/static/scripts/common/components/EntityLink.js
+++ b/root/static/scripts/common/components/EntityLink.js
@@ -191,9 +191,8 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
     href = entity.href_url;
   }
 
-  // TODO: support name variations for all entity types?
-  if (!subPath &&
-      (entity.entityType === 'artist' || entity.entityType === 'recording')) {
+  // URLs are kind of weird and we probably don't care to set this for them
+  if (!subPath && entity.entityType !== 'url') {
     if (nameVariation === undefined && typeof content === 'string') {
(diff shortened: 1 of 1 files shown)


mwiencek (approved):

root/static/scripts/common/components/EntityLink.js
```diff
@@ -191,9 +191,8 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
     href = entity.href_url;
   }
 
-  // TODO: support name variations for all entity types?
-  if (!subPath &&
-      (entity.entityType === 'artist' || entity.entityType === 'recording')) {
+  // URLs are kind of weird and we probably don't care to set this for them
+  if (!subPath && entity.entityType !== 'url') {
```
yvanzo, line 195: The same probably goes for collection and genre. I would suggest to extend the list of entity types known to have name variation instead of setting name variation by default.

reosarevok: My thought was that anything else than url probably should show a name variation if the content is different than the name, but there might be some places we don't want this? For genre for example I'd expect to want to show name variation once we start allowing aliased genres.

yvanzo: Alright, no specific `content` is provided for genre and collection for now anyway.

yvanzo (approved):

PR #1704, opened 2020-09-21
MBS-11036: Don't show Add release button when moving CD TOC

reosarevok:
### Implement MBS-11036

This button is currently broken, but after some talk we decided that it shouldn't be fixed, but dropped. There's no way to know
that moving (as opposed to additionally adding) a CD TOC to a new release is correct, except for very specific cases.

It makes more sense to only allow moving the CD TOC to another existing release. If it has been added by mistake, it can either be removed from the wrong one (and the new release can just be added based on the CD TOC) or it can be moved after adding the release.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/CDTOC.pm
+++ b/lib/MusicBrainz/Server/Controller/CDTOC.pm
@@ -300,6 +300,7 @@ sub _attach_list {
 
             $c->stash(
                 template => 'cdtoc/attach_filter_release.tt',
+                cdtoc_action => 'add',
                 results => [sort_by { $_->entity->release_group ? $_->entity->release_group->gid : '' } @$releases]
             );
             $c->detach;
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/CDTOC.pm
```diff
@@ -300,6 +300,7 @@ sub _attach_list {
 
             $c->stash(
                 template => 'cdtoc/attach_filter_release.tt',
+                type => 'add',
```
mwiencek, line 303: This PR looks good except I usually prefer to use descriptive names for stuff in the stash, since they're basically globals and it makes it easier to grep for all the places they're referenced. Maybe something like `cdtoc_action` here?
```suggestion
                cdtoc_action => 'add',
```

yvanzo (approved):

mwiencek (approved):
Thanks!

PR #1705, opened 2020-09-21
MBS-11094: Don't block editing on pre-existing too early format

reosarevok:
### Fix MBS-11094

With other errors like this one, we allow the user to edit as long as they don't actually introduce a problem that didn't previously exist. This does the same, by not giving an error (but still displaying a warning) if the too early format was already there. It will still give an error if the user selects *another* too early format.

IT'S ALL YOUR FAULT version:
Screenshot from 2020-09-22 07-37-56

YOU'RE GOOD version:
Screenshot from 2020-09-22 07-37-42

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -309,9 +309,6 @@
           <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: needsFormat">
             [%~ l('A format is required. If you don’t know it, tick the “I don’t know” checkbox next to the format dropdown.') %]
           </p>
-          <p class="error field-error" style="padding-left: 1em" data-bind="showErrorRightAway: hasTooEarlyFormat">
-            [%~ l('This medium is set to a format that did not exist on the selected release date.') %]
-          </p>
         </td>
 
         <td class="align-right icon" >
@@ -327,6 +324,26 @@
          )
        %]
     </table>
+
+    <!-- ko if: hasTooEarlyFormat -->
+      <div class="warning">
+        [%- warning_icon %]
+        <p>
+          <strong>[% add_colon(l('Warning')) %]</strong>
+          [% l('This medium is set to a format that did not exist on the selected release date.') %]
+          <br />
+          <!-- ko if: hasUnconfirmedEarlyFormat -->
+          [% l('This should never be correct. Make sure that, say, your release set to CD isn’t meant to be a vinyl, or your vinyl isn’t actually a shellac disc.') %]
+          <!-- /ko -->
+          <!-- ko ifnot: hasUnconfirmedEarlyFormat -->
+          [% l('This problem already existed before you opened the release editor, so you don’t need to do anything about it. But if you have some time, consider trying to find out what the real format should be or if the release should be split!') %]
+          <!-- /ko -->
+        </p>
+        <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: hasUnconfirmedEarlyFormat">
+          [%~ l('This medium is set to a format that did not exist on the selected release date.') %]
+        </p>
+      </div>
+    <!-- /ko -->
 
     <!-- ko if: loading -->
       <div class="tracklist-loading">
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -320,6 +320,9 @@ class Medium {
     this.name = ko.observable(data.name);
     this.position = ko.observable(data.position || 1);
     this.formatID = ko.observable(data.format_id);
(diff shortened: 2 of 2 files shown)


root/release/edit/tracklist.tt
```diff
@@ -328,6 +325,20 @@
        %]
     </table>
 
+    <!-- ko if: hasTooEarlyFormat -->
+      <div class="warning">
+        [%- warning_icon %]
+        <p>
+          <strong>[% l('Warning:') | html_entity %]</strong>
+          [% l('This medium is set to a format that did not exist on the selected release date.') %]
```
reosarevok, line 333: Left as its own l() to reuse the error message translation.

root/release/edit/tracklist.tt
```diff
@@ -328,6 +325,20 @@
...
+          [% l('This should never be correct. Make sure that, say, your release set to CD isn’t meant to be a vinyl, or your vinyl isn’t actually a shellac disc.') %]
```
yvanzo, line 334: The release has not been entered by the current editor, so we should probably not write “your” and it should be clarified that the mistake is not the fault of the current editor either:

```suggestion
          [% l('This mistake was there before you started editing this release. Make sure that, say, the release set to CD isn’t meant to be a vinyl, or the vinyl isn’t actually a shellac disc.') %]
```

root/release/edit/tracklist.tt
```diff
@@ -328,6 +325,20 @@
        %]
     </table>
 
+    <!-- ko if: hasTooEarlyFormat -->
+      <div class="warning">
+        [%- warning_icon %]
+        <p>
+          <strong>[% l('Warning:') | html_entity %]</strong>
```
yvanzo, line 332: Optionally:

```suggestion
          <strong>[% add_colon(l('Warning')) %]</strong>
```

yvanzo (commented):
Setting the format to another wrong format causes the warning to disappear. It’s probably still better than the current situation though.

root/release/edit/tracklist.tt
reosarevok: We don't usually clarify this elsewhere (and then we would need to change the warning text depending on whether the editor added it now or not). Is it really needed? It's still true that it should never be correct. 

The 'the' is certainly better :)

reosarevok:
> Setting the format to another wrong format causes the warning to disappear. It’s probably still better than the current situation though.

That shouldn't happen, unless the other format doesn't actually have a year stored (and is as such not officially wrong)... will retest if that does happen.

root/release/edit/tracklist.tt
yvanzo: > We don't usually clarify this elsewhere (and then we would need to change the warning text depending on whether the editor added it now or not).

Changing release format is not straightforward because releases can be mixed together and then changing only one field in only one release is not solving anything, it often requires to check the release group instead. So I feel that more explanations should be given about it. Maybe it should even be discouraged to fix it if information is missing, e.g. “Please fix it if and only if you have enough information to distinguish between possible mixed information from different releases.”

> It's still true that it should never be correct.

Correct. I suggested to start with “This mistake” instead which is probably clear enough (as a mistake should never be correct)?

reosarevok: I added a separate wording for when the error predates the editing session. See the PR description for images too.

root/release/edit/tracklist.tt
```diff
@@ -328,6 +325,26 @@
...
+        [%- warning_icon %]
+        <p>
+          <strong>[% add_colon(l('Warning')) %]</strong>
+          [% l('This medium is set to a format that did not exist on the selected release date.') %]
+          <br />
+          <!-- ko if: hasUnconfirmedEarlyFormat -->
+          [% l('This should never be correct. Make sure that, say, your release set to CD isn’t meant to be a vinyl, or your vinyl isn’t actually a shellac disc.') %]
+          <!-- /ko -->
+          <!-- ko ifnot: hasUnconfirmedEarlyFormat -->
+          [% l('You didn’t cause this problem right now, so you don’t need to do anything about it. But if you have some time, consider trying to find out what the real format should be!') %]
```
yvanzo, line 339: IMHO the most common case is likely that the date for the earliest release event actually matches the original release date for that release group. In such case, the format is usually correct but the release event should be moved to a separate release. So I would rather suggest to look at possibly different releases, roughly:

```suggestion
          [% l('You didn’t cause this problem right now, so you don’t need to do anything about it. But if you have some time, consider trying to find out what the real format should be or if the release should be split!') %]
```

yvanzo (approved):
> That shouldn't happen, unless the other format doesn't actually have a year stored (and is as such not officially wrong)...

Yup, I likely used 7" Vinyl which has not date set. That is a separate issue.

mwiencek (approved):

PR #1707, opened 2020-09-23
MBS-11033: Update the Apple Music logo

reosarevok:
### Implement MBS-11033

They changed it *again*.


mwiencek (approved):

yvanzo (approved):
TBH, it is not really needed to use their latest favicon, any recognizable icon would probably be just fine, especially when a website is changing its favicon depending on seasons, events, special offers…

reosarevok:
They specifically say this (and its variations) is the only logo you can use to represent them, so...

yvanzo:
The previous icon was probably also _the only logo you can use_ at that time, and there is probably no legal obligation to update it, but I’m not lawyer.

PR #1709, opened 2020-09-23
Delete unneeded json_dump rows faster

mwiencek:
If we're running a full JSON dump, we can delete any rows older than the full dump sequence, provided there exists at least 1 row per entity newer or equal to that sequence. Previously we would be using the sequence of the /previous/ full dump, so the number of rows in these tables is double what they should be. We'll need to VACUUM FULL once the fix is in place.


reosarevok (approved):
Seems sensible to me

yvanzo (approved):
LGTM

PR #1710, opened 2020-09-24
MBS-11121: Normalize Apple Music beta links

reosarevok:
### Implement MBS-11121

They point to the same content, so just removing `beta.` from them works.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -374,6 +374,10 @@ const testData = [
             expected_clean_url: 'https://music.apple.com/us/artist/444923726',
   },
   {
+                     input_url: 'https://beta.music.apple.com/ca/artist/imposs/205021452',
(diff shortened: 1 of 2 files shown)


mwiencek (approved):

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -381,6 +381,10 @@ const testData = [
                      input_url: 'https://music.apple.com/jp/album/uchiagehanabi-single/1263790414',
             expected_clean_url: 'https://music.apple.com/jp/album/1263790414',
   },
+  {
+                     input_url: 'https://beta.music.apple.com/ca/artist/imposs/205021452',
+            expected_clean_url: 'https://music.apple.com/ca/artist/205021452',
+  },
```
yvanzo, line 387: Tests are usually grouped by entity type, this test should probably follow the first one for Apple Music.

yvanzo (approved):

PR #1711, opened 2020-09-24
MBS-8650: Lower how long an IPI must be to 5

reosarevok:
### Implement MBS-8650

5 digits was the shortest our users found in SACEM (see ticket). Given that it seems unlikely a user would randomly type 5 digits by mistake in the IPI field, going down to this seems safe to me.


mwiencek (approved):

yvanzo (approved):
Please update ticket accordingly.

PR #1712, opened 2020-09-24
MBS-11114: Allow track times to be unset

mwiencek:
0d45452121257d336c915f1b78f1114a21afe518 broke this by incorrectly removing the "useless" isNaN check, which it turns out was not "useless." The result of `unformatTrackLength` is null when the length is empty, so the bug is that we'd bail out too early with the new check and fail to set `this.length(newLength)`. The result is only NaN when the input can't be parsed. A test is now added.

### DIFF
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -159,7 +159,12 @@ class Track {
 
     var newLength = utils.unformatTrackLength(length);
 
-    if (!newLength) {
+    /*
+     * The result of `unformatTrackLength` is NaN when the length entered
+     * by the user can't be parsed. If they've *cleared* the length, it's
+     * null.
+     */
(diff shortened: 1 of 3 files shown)


root/static/scripts/release-editor/fields.js
```diff
@@ -159,7 +159,7 @@ class Track {
 
     var newLength = utils.unformatTrackLength(length);
 
-    if (!newLength) {
+    if (Number.isNaN(newLength)) {
```
mwiencek, line 162: This previously used https://lodash.com/docs/4.17.15#isNaN which is based on `Number.isNaN`, not `isNaN`:

> Note: This method is based on Number.isNaN and is not the same as global isNaN which returns true for undefined and other non-number values.

reosarevok (approved):
Whoops.

yvanzo (approved):
:shipit:

PR #1713, opened 2020-09-24
MBS-3116: Warn when merging works with different ISWCs 

reosarevok:
### Implement MBS-3116

This just mirrors the approach for recordings with different ISRCs.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Work.pm
+++ b/lib/MusicBrainz/Server/Controller/Work.pm
@@ -166,7 +166,7 @@ sub _merge_load_entities
         my $get_iswc_set = sub { Set::Scalar->new(map { $_->iswc } shift->all_iswcs) };
         my $expect = $get_iswc_set->($comparator);
         $c->stash(
-            iswcs_differ => any { $get_iswc_set->($_) != $expect } @tail
+            iswcs_differ => (any { $get_iswc_set->($_) != $expect } @tail),
         );
     }
 };
--- a/lib/MusicBrainz/Server/Controller/Work.pm
+++ b/lib/MusicBrainz/Server/Controller/Work.pm
@@ -154,9 +154,6 @@ sub _merge_load_entities
     my ($self, $c, @works) = @_;
     $c->model('Work')->load_meta(@works);
     $c->model('WorkType')->load(@works);
-    if ($c->user_exists) {
-        $c->model('Work')->rating->load_user_ratings($c->user->id, @works);
-    }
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Controller/Work.pm
```diff
@@ -153,14 +154,21 @@ sub _merge_load_entities
...
     $c->model('Language')->load_for_works(@works);
     $c->model('ISWC')->load_for_works(@works);
+
+    my @works_with_iswcs = grep { $_->all_iswcs > 0 } @works;
+    if (@works_with_iswcs > 1) {
+        my ($comparator, @tail) = @works_with_iswcs;
+        my $get_iswc_set = sub { Set::Scalar->new(map { $_->iswc } shift->all_iswcs) };
+        my $expect = $get_iswc_set->($comparator);
+        $c->stash(
+            iswcs_differ => any { $get_iswc_set->($_) != $expect } @tail
```
mwiencek, line 169: You should still add `()` around the `any` call, because if you add more keys after `iswcs_differ`, they'll be consumed by `any`:

```
$ use List::AllUtils qw( any );
$ (foo => any { $_ } (0, 0, 0), bar => 'haha');
$VAR1 = 'foo';
$VAR2 = 1;
$ (foo => (any { $_ } (0, 0, 0)), bar => 'haha');
$VAR1 = 'foo';
$VAR2 = '';
$VAR3 = 'bar';
$VAR4 = 'haha';
```

```suggestion
            iswcs_differ => (any { $get_iswc_set->($_) != $expect } @tail),
```

mwiencek (changes requested):

root/work/WorkMerge.js
```diff
@@ -35,6 +37,18 @@ const WorkMerge = ({
         {l(`You are about to merge all these works into a single one.
             Please select the work all others should be merged into:`)}
       </p>
+      {iswcsDiffer ? (
+        <div className="warning warning-iswcs-differ">
+          <p>
+            {exp.l(
+              `<strong>Warning:</strong> Some of the works you're
```
mwiencek, line 44:
```suggestion
              `<strong>Warning:</strong> Some of the works you’re
```

lib/MusicBrainz/Server/Controller/Work.pm
reosarevok: Made the change to both this and the ISRC version in a separate commit.

mwiencek (approved):

PR #1714, opened 2020-09-25
MBS-4118: Add full-icu to Node for locale strings

reosarevok:
### Fix MBS-4118

Without this, Number.toLocaleString and Date.toLocaleString always use the English locale in Node 12, at the very least.

See https://nodejs.org/docs/latest-v12.x/api/intl.html

For testing, see that (e.g.) the statistics page does have the right thousand separators (comma for English, space for Estonian, thin space for French and dots for Spanish) rather than always having comma-separators.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "detect-node": "2.0.3",
     "file-loader": "3.0.1",
     "filesize": "2.0.4",
+    "full-icu": "1.3.1",
     "generic-diff": "1.0.1",
     "he": "1.1.1",
     "imports-loader": "0.8.0",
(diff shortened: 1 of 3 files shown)


kellnerd:
I also had problems with Node.js ignoring my i18n settings and have solved them by switching from version 12 to 14 which now [includes full-icu by default](https://nodejs.org/docs/latest-v14.x/api/intl.html#intl_internationalization_support). You might want to consider taking this step too as [LTS for version 14](https://nodejs.org/en/about/releases/) starts in a few days.

yvanzo (approved):
LGTM.

_LTS for version 14  starts_ in a month, we should certainly consider it then.

mwiencek (approved):

PR #1715, opened 2020-09-25
MBS-11127: Use cardinality for work-level-rels in WS/2

reosarevok:
### Implement MBS-11127

This avoids returning all unrelated recording-work relationships for every release with work-level-rels inc, which in some cases
was over 90% of the size of the full response for no apparent benefit.

The only other relationship currently not returned because of cardinality is work-artist "artist was named after work", which also seems irrelevant to release responses anyway.

Note: we should make sure to notify users of this change when we release it (and possibly before?) in case someone specifically expected and made use of this behaviour. If needed, we could add a second, explicitly GIVE ME ALL inc that still works like before.


mwiencek (approved):

yvanzo (approved):
I clarified “_unrelated_ to what” (which explains why) in ticket’s description.

PR #1716, opened 2020-09-28
MBS-6322: Add report for releases with same barcodes in diff RGs

reosarevok:
### Implement MBS-6322

This should almost never happen, unless the label made a mistake. This skips bootleg releases, since those sometimes intentionally are all given the same barcode anyway by the bootleg label.


yvanzo (approved):
:shipit:

PR #1717, opened 2020-09-28
MBS-11129: Display mediums as Medium # (Format)

reosarevok:
### Implement MBS-11129

The current way (CD 1, DVD 2) kinda makes it sound like there should be a DVD 1.

For release tracklists, this just shows the medium position at the beginning of the header, then shows only medium format and name after the expand/collapse arrow.

Everywhere else, such as on edits, or on the relationship lists under the release, `Medium 1 (CD)`, `Medium 2 (DVD)`, etc. seems a lot less confusing and not too repetitive / silly.

### DIFF
--- a/root/cdtoc/AttachCDTocConfirmation.js
+++ b/root/cdtoc/AttachCDTocConfirmation.js
@@ -39,7 +39,8 @@ component AttachCDTocConfirmation(
       <p>
         {exp.l(
           `Are you sure that you wish to attach the disc ID
-           <code>{discid}</code> to {format} {pos} of {release} by {artist}?`,
+           <code>{discid}</code> to medium {pos} ({format})
+           of {release} by {artist}?`,
           {
             artist: <ArtistCreditLink artistCredit={release.artistCredit} />,
             discid: cdToc.discid,
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -420,7 +420,7 @@ END -%]
 [%~ MACRO format_length(n) BLOCK; n | format_length; END -%]
 
 [%~ MACRO medium_format_name(medium) BLOCK # Converted to React at root/static/scripts/common/utility/mediumFormatName.js -%]
-    [% medium.l_format_name or l("Medium") | html %]
+    [%- medium.l_format_name or l("Medium") | html -%]
 [%- END -%]
 
 [%~ MACRO release_countries_list(release_events) BLOCK;
--- a/root/static/scripts/common/components/MediumDescription.js
+++ b/root/static/scripts/common/components/MediumDescription.js
@@ -10,11 +10,32 @@
 import isolateText from '../utility/isolateText.js';
 import mediumFormatName from '../utility/mediumFormatName.js';
 
+export component MinimalMediumDescription(medium: MediumT) {
+  const mediumDescription = mediumFormatName(medium);
+  if (medium.name) {
+    return (
+      <>
+        {addColonText(mediumDescription)}
+        {' '}
+        <span className="medium-name">{isolateText(medium.name)}</span>
+      </>
+    );
+  }
+  return mediumDescription;
+}
+
 component MediumDescription(medium: MediumT) {
-  const formatAndPosition = texp.l('{medium_format} {position}', {
-    medium_format: mediumFormatName(medium),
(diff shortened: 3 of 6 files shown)


mwiencek:
This looks perfectly fine to me if there are multiple formats, though I do kinda agree with nikki's comment that "[i]t seems kinda clunky, especially when what we have works fine for the majority of releases."

By my count, only ~0.85% of releases have more than one medium format, or when excluding single-medium releases, ~11.9% of releases with multiple mediums. For the other ~88%, "CD 1 + CD 2 + ..." just makes much more sense.

What if we only downgraded to this style when there are multiple formats, or gaps in the medium sequence? (Does it have to be perfectly consistent?)

reosarevok:
That has the issue that you need to know not only the format of this medium, but the format of all others too, in order to display a medium. Seems annoying.

mwiencek:
I mean, if it's outside a release context and other mediums aren't loaded, Medium 1 (CD) seems fine too. That's mostly just the CDTOC pages and the medium search in the RE, right? It just looks clunky on the majority of release pages.

reosarevok:
And some edit pages like remove medium :) 

Honestly, I think it only looks clunky because we didn't do it like this from the beginning, and I prefer the consistency to the prettiness, but YMMV :) Let's see what @yvanzo thinks.

yvanzo (commented):
I kinda agree with previous comments about keeping the current label when all mediums of the release are in the same format, at least when it is technically feasible. Annoying code is a lesser issue than annoying page for the user.

reosarevok:
My problem is not annoying code, it's inconsistent display.

mwiencek:
I do think having one style we can apply consistently is a good idea. But the most popular suggestion in https://tickets.metabrainz.org/browse/MBS-5864 seemed to be:

 1. Showing positions relative to the medium format
 2. Hiding the relative position if there's only one of that format
 3. Showing the overall position at the start somehow

So...

1.  CD
2.  DVD 1
3.  DVD 2

There were concerns brought up about situations where you have interleaving formats or gaps in the medium sequence, but those situations seem quite rare and not worth worrying about.

Both the relative & overall position numbers do seem useful in different scenarios though:

 1. The overall position makes the total count of mediums clear and lets you identify gaps.
 2. The relative position makes the total count of each format clear and reads more naturally.

For the overall positions at the start, nikki suggested to put them "in the same column as the track number (perhaps in a lighter font)." @Aerozol might have a better idea of how to format these.

Aerozol:
I never thought about this, but it's a good point being made in the ticket.

I think this is what's been suggested, perhaps a bit simpler? It would work for me:
image

reosarevok:
That seems ok - what should be shown everywhere else we link to mediums? Something like https://musicbrainz.org/edit/123406855 which right now says "CD 1 on Release". Should it be "Medium 1 (CD) on Release"? "Medium 1 on Release"? Something else?

Aerozol:
Medium 1 (CD) seems good, contains all the information someone would need

reosarevok:
I made some basic changes here as suggested by @Aerozol, and put it on test so it can be checked a bit.

Release index display can be tested [with this release](https://test.musicbrainz.org/release/84d25ad6-8da8-496a-85d1-f6fd8a85c4d2), for example. As discussed, it's just `1	▼CD` rather than counting the different numbers. This would seem to be no worse than what we currently do (where we do not have counters per format) but less likely to lead to confusion.

Clicking "Display credits at bottom" on the same release shows the new headers. I actually like `Medium 1 (CD)` *more* as a header personally than the previous `CD 1`, not just because it avoids the previous confusion but because it feels more... headery by not being so short. But that might just be me :sweat_smile: 

Edit display can be tested [with this edit search](https://test.musicbrainz.org/search/edits?auto_edit_filter=&order=desc&negation=0&combinator=and&conditions.0.field=type&conditions.0.operator=%3D&conditions.0.args=55%2C232&conditions.0.args=52&conditions.0.args=56%2C221&conditions.0.args=54%2C220&conditions.0.args=22&field=Please+choose+a+condition).

reosarevok:
Funny thing is there *is* supposed to be a space there from the beginning, it's there in the code, but for some reason the computer says no. I replaced it with CSS now, and pushed the update to test.

PR #1718, opened 2020-09-29
MBS-7219: Hide links to empty ArtistRecording subsets

reosarevok:
### Implement MBS-7219

If there are no standalone recordings or no videos then there is no good reason to actually give the user links to such subset pages.

I hope making this check every time a recording page is loaded isn't too bad given that I'm specifically limiting it to 1 - I can't see a good way to do this otherwise.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -361,6 +361,9 @@ sub recordings : Chained('load')
     my $standalone_only;
     my $video_only;
 
+    my $has_standalone = $c->model('Recording')->has_standalone($artist->id);
+    my $has_video = $c->model('Recording')->has_video($artist->id);
+
     my %filter = %{ $self->process_filter($c, sub {
         return create_artist_recordings_form($c, $artist->id);
     }) };
@@ -368,13 +371,15 @@ sub recordings : Chained('load')
 
     if ($c->req->query_params->{standalone}) {
         $recordings = $self->_load_paged($c, sub {
-            $c->model('Recording')->find_standalone($artist->id, shift, shift);
+            return ([], 0) unless $has_standalone;
+            return $c->model('Recording')->find_standalone($artist->id, shift, shift);
         });
         $standalone_only = 1;
     }
     elsif ($c->req->query_params->{video}) {
         $recordings = $self->_load_paged($c, sub {
-            $c->model('Recording')->find_video($artist->id, shift, shift);
+            return ([], 0) unless $has_video;
+            return $c->model('Recording')->find_video($artist->id, shift, shift);
         });
         $video_only = 1;
     }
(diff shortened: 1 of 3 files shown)


root/artist/ArtistRecordings.js
```diff
@@ -89,19 +93,35 @@ const ArtistRecordings = ({
...
+          )}
+        </p>
+      ) : hasVideos ? (
+        <p>
+          {exp.l(
+            `Showing all recordings. {show_vid|Show only videos}.`,
+            {show_vid: `/artist/${artist.gid}/recordings?video=1`},
+          )}
+        </p>
+      ) : null
```
yvanzo, line 124: It makes 3 localizable quite redundant localizable messages. If a 3rd parameter gets added, it would make 7. I would suggest to split into 3 separate sentences instead.

reosarevok: Done. Since I was already making it modular, I expanded it to actually add the modular links to the standalone only and video only pages.

yvanzo (approved):

lib/MusicBrainz/Server/Controller/Artist.pm
```diff
@@ -360,6 +360,9 @@ sub recordings : Chained('load')
     my $standalone_only;
     my $video_only;
 
+    my $has_standalone = scalar $c->model('Recording')->find_standalone($artist->id, 1);
+    my $has_videos = scalar $c->model('Recording')->find_video($artist->id, 1);
```
mwiencek, line 364: We should add custom methods to the data layer for these, like `has_standalone` and `has_video`. The first query resolves to e.g.

```SQL
WITH x AS (
    SELECT DISTINCT recording.id, recording.gid, recording.name,
        recording.artist_credit AS artist_credit_id,
        recording.length, recording.comment, recording.video,
        recording.edits_pending, recording.last_updated,
        recording.name COLLATE musicbrainz
      FROM recording
    LEFT JOIN track t ON t.recording = recording.id
      JOIN artist_credit_name acn
        ON acn.artist_credit = recording.artist_credit
     WHERE t.id IS NULL
       AND acn.artist = 303
    ORDER BY recording.name COLLATE musicbrainz
)
SELECT x.*, c.count AS total_row_count
FROM x, (SELECT count(*) from x) c
LIMIT 1;
```

for artist = 303, which takes around 100 ms on pink. With a more optimized query, it only takes ~25 ms (still quite slow, but 4X faster):

```SQL
SELECT EXISTS (
    SELECT 1
      FROM recording
      JOIN artist_credit_name acn
        ON acn.artist_credit = recording.artist_credit
     WHERE acn.artist = 303
       AND NOT EXISTS (SELECT 1 FROM track WHERE track.recording = recording.id)
);
```

mwiencek (changes requested):

lib/MusicBrainz/Server/Controller/Artist.pm
reosarevok: I think it should be fine now. I thought about using these to avoid running the queries in     `if ($c->req->query_params->{standalone}) ` and whatnot, but a pager is required not to break the React display. Any suggestions for that, or should I just leave it be? :)

mwiencek: One thing you could do is return `([], 0)` from the `_load_paged` sub if those vars are set.

mwiencek (approved):

PR #1719, opened 2020-09-30
MBS-11136: Support HTML in admin/attribute descriptions

reosarevok:
### Implement MBS-11136

While the descriptions for things like EventType (the one I actually added an HTML description for) don't actually show anywhere except the admin/attribute page yet AFAICT, it still probably makes sense to show them as intended here.

Hopefully we will eventually also display them elsewhere when relevant, so that most users can see them.

### DIFF
--- a/root/admin/attributes/Attribute.js
+++ b/root/admin/attributes/Attribute.js
@@ -12,6 +12,7 @@ import * as React from 'react';
 
 import Layout from '../../layout';
 import {compare} from '../../static/scripts/common/i18n';
+import expand2react from '../../static/scripts/common/i18n/expand2react';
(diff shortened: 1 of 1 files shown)


mwiencek (approved):

yvanzo (changes requested):
Change is good but commit message, pull request, and ticket should specify “**entity** attribute”.

No given example because there currently is no HTML in entity attribute descriptions?

reosarevok:
There is one now (in event type) but when I tried to test it wasn't on pink yet. It probably is now though. You can create your own, in any case, just add an `<a>` to any description in the dev set.

yvanzo (approved):

PR #1720, opened 2020-09-30
MBS-11137: Show info on hover on frontpage's new releases

reosarevok:
### Implements MBS-11137

I suspect having the release title and artist on the <a> title was meant to show it when hovering over the cover art, but it doesn't - we need to pass it down to ArtworkImage.
I also made Artwork itself support it, since we might as well want it later and it feels more consistent.


yvanzo (approved):
Show **release** info indeed.

mwiencek (approved):

PR #1721, opened 2020-10-01
MBS-2418: Show Edit URL edits in entity edit histories

reosarevok:
### Implement MBS-2418

This should make it so that the edits show for every entity linked to a specific URL being edited.

We don't look at URL-URL relationships, since they don't exist as a concept at the moment at all.

I *think* I don't need to check direction here, but I might be wrong?

Let's make sure this won't accidentally link a ton of stuff like last time, but the general idea seems like it just makes sense IMO.


mwiencek:
> I think I don't need to check direction here, but I might be wrong?

What you have works, but it might make generating the hash ref a bit easier.

```perl
for my $rel (@relationships) {
    my $target_type = $rel->target_type;
    my $target_id = $rel->direction == 1 ? $rel->entity1_id : $rel->entity0_id;
    push @{ $related_entities->{$target_type} //= [] }, $target_id;
}
```

reosarevok:
Made that change (it does require passing the URL IDs separately), added a test.

mwiencek (approved):

reosarevok:

reosarevok:

reosarevok:

reosarevok:

PR #1722, opened 2020-10-01
Remove span.name-variation around "see all" releases link

reosarevok:
Amend d17a2fd: Drop name-variation from "see all"

Remove span with class name-variation around "see all" releases link.

We did this once before for MBS-10536, but we've since changed the behaviour while implementing MBS-11107 so that it is shown if nameVariation is undefined and the content is not the entity name. As such, explicitly state we do not want nameVariation here.

### DIFF
--- a/root/release/ReleaseHeader.js
+++ b/root/release/ReleaseHeader.js
@@ -44,6 +44,7 @@ const ReleaseHeader = ({
             <EntityLink
               content={rgLink}
               entity={release.releaseGroup}
+              nameVariation={false}
             />,
           )}
         </span>


mwiencek (approved):

yvanzo (approved):
The code works for me but PR title is bugging the Jira bot because it refers to a closed issue. Actually the commit message is misleading too, it should refer to the current issue, which is the faulty commit since there is no ticket for this regression. Instead of referring to the previous ticket in commit message title, it should refer to it in body only. I mixed it with your PR description, use it if you are happy with it:

```gfm
Amend d17a2fd: Drop name-variation from "see all"

Remove span with class name-variation around "see all" releases link.

We did this once before for MBS-10536, but we've since changed the
behaviour while implementing MBS-11107 so that it is shown if
nameVariation is undefined and the content is not the entity name.
As such, explicitly state we do not want nameVariation here.
```

PR #1724, opened 2020-10-01
MBS-11140: Build Jed language bundles in prod

mwiencek:
Language packs /probably/ need to be installed before compile_resources.sh is run in order for translations to work. Just a guess.

Broken in aed356a10c0551d26f585d03a30a95918231f9c9

### DIFF
--- a/docker/templates/Dockerfile.website.m4
+++ b/docker/templates/Dockerfile.website.m4
@@ -8,9 +8,9 @@ copy_common_mbs_files
 
 git_info
 
-install_javascript_and_templates(` --only=production')
-
 install_translations()
+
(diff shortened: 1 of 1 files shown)


mwiencek:
I built some custom Docker images based on this branch at https://ci.metabrainz.org/job/musicbrainz-docker-images/529/console

As you can see, jed-$lang bundles are now outputted by Webpack whereas they previously weren't. So this does the trick.

reosarevok (approved):

yvanzo (approved):
Thanks, would be nice to have in the upcoming release! :ship:

mwiencek:
OK, I changed the base branch to beta.

PR #1726, opened 2020-10-04
MBS-11149: Fix ironic mis(s)pelling

reosarevok:
### Fixes MBS-11149

As two commits because I'm lazy and did it through the web UI. We can squash for merging.


yvanzo (approved):
Looks good too me!

mwiencek (approved):

PR #1727, opened 2020-10-04
MBS-11139: Use HTTPS for display on Library of Congress links

reosarevok:
### Implement MBS-11139

We want to keep the URI as HTTP because that seems to be their official permalink, but they redirect to HTTPS so it makes sense for us to do the same.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/URL/LoC.pm
+++ b/lib/MusicBrainz/Server/Entity/URL/LoC.pm
@@ -5,6 +5,12 @@ use Moose;
 extends 'MusicBrainz::Server::Entity::URL';
 with 'MusicBrainz::Server::Entity::URL::Sidebar';
 
+override href_url => sub {
+    # Turn the official permalink into what LoC currently redirects to.
+    shift->url->as_string =~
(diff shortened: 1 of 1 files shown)


mwiencek (approved):

yvanzo (approved):
This was not really needed but if you do so you should update existing links to `https` so as to preserve normalization (which helps with searching for URL and preventing duplicate).

reosarevok:
I think it does make sense (we should try to always use https when possible) but agreed on the normalization - entered MBBE-16 for that. I'm going to try to set up a bot for MBBE this week, that seems like a good start issue.

yvanzo (changes requested):
I just reminded why these URLs are being enforce to `http` over `https`: it’s what is used by the Library of Congress itself.

For example, all the links from http://id.loc.gov/authorities/names/n79146554 are still using `http` either because the page is outdated but more likely because it’s considered as a permalink more than a simple URI. The same goes for VIAF URLs. I guess we should keep the normalized `http` form in the database, and link using `https` in the webpage only, just like do for VIAF and some others already.

reosarevok:
Made the change.

yvanzo (approved):

PR #1728, opened 2020-10-04
MBS-11147: Update autocomplete serialization

reosarevok:
### Fix MBS-11147

artists was changed to related_artists with commit d1d8fbaf9c97f9d8f45b4e28b70b00a12688d3da but the autocomplete_work sub was not updated for it, breaking autocomplete.


yvanzo (approved):
Make sense, did not find any other occurrence after a quick review.

mwiencek (approved):

PR #1730, opened 2020-10-05
MBS-11152: Convert entity/ratings page to React

reosarevok:
### Implement MBS-11152


root/components/RatingStars.js
```diff
@@ -23,11 +23,33 @@ const ratingURL = (entity: RatableT, rating) => (
...
+        <span
+          className="current-rating"
+          style={{width: `${rating ?? 0}%`}}
+        >
+          {starRating}
+        </span>
+      </span>
+    </span>
+  );
+};
```
yvanzo, line 50: It seems to duplicate half of `RatingStars` component, would it be possible to deduplicate this? Maybe by adding a flag parameter to `RatingStars`?

reosarevok: I personally feel it's much more understandable to have a separate component than to have even more conditionals in RatingStars. That said, one thing that could be done is to add a third component like this and reuse it as a layout for both of these, if you feel that's better (it would certainly reduce duplication): 

```
    <span className="inline-rating">
      <span className="star-rating" tabIndex="-1">
        {children}
      </span>
    </span>
```

yvanzo: It would deduplicate only a small portion of code. Couldn’t `DisplayStaticRatingProps` (not sure of the meaning of this name) have an entity parameter instead and thus be able to simply share more code?

reosarevok: Not really, because sometimes it's not getting the rating from a RatableT but from a RatingT.

mwiencek (approved):

root/components/RatingStars.js
yvanzo: `RatingT` is an addition of this PR, can it be avoided? For example, private ratings should probably not be passed to the renderer at all, the logic of counting them probably better fits the Perl back-end.

reosarevok: Passing a count of private ratings rather than the actual ratings seems fine to me if desired, but we'd still need the rating + editor combo for public ones :)

reosarevok: In general, IMO there's a reason why this was a separate macro in TT - it seems much simpler to call a static method when you want static ratings and one that allows setting ratings elsewhere :).

yvanzo: Does _static_ means _read-only_ here? If so please rename `DisplayStaticRating` to `ReadOnlyRatingStars` since it’s doing the same as `RatingStars` but read-only, rendered components are all about _display_ and `static` is mostly used for resources.

reosarevok: I guess ReadOnlyRatingStars seems like a good enough name :)

yvanzo: I failed at deduplicating code too, but there are not that many lines and all are in the same file, so not a big deal anyway.

yvanzo (approved):

root/components/RatingStars.js
mwiencek: I prefer `DisplayStaticRating` here. "Static" (vs. dynamic or editable) is an established term for UIs. "Read only" makes me think of file systems or flow types, and not a user interface. We use it for `StaticRelationshipsDisplay` already too.

https://docs.microsoft.com/en-us/windows/win32/controls/static-controls
https://developer.apple.com/documentation/swiftui/view-layout-and-presentation
https://stackoverflow.com/questions/1000878/static-ui-vs-dynamic-ui/

reosarevok: Does StaticRatingStars work as a compromise?

mwiencek: That's fine with me, I just think "Static" is less confusing that "ReadOnly"

PR #1732, opened 2020-10-06
 MBS-11146: Reviews tab for ratings + CB reviews 

reosarevok:
### Implement MBS-11146

As shown on TV, er... talked about during Summit '20.

This creates a tab for ratings, which we didn't previously have but makes sense in any case IMO. The tab is named Reviews and also displays the CB reviews.

We don't really want to display reviews without a body (rating only reviews) so I request ones with text explicitly. We do claim there's no reviews when there's just rating-only ones, but I feel that's fair when we're talking about 'writing a review'. We could add a third call for ratings but I feel this isn't really useful enough to justify three calls. Ideally eventually CB will support getting the average rating for an entity and its recent and top reviews in one call.

Flow is unhappy because it doesn't realize `entityProperties.ratings` limits entities to `RatableT` and so on, so I added two FlowIgnores for that.

Example (spam but shows the idea pretty well): 
Screenshot from 2020-10-06 12-12-30


root/static/scripts/common.js
```diff
@@ -42,6 +42,7 @@ require('./common/components/WorkArtists');
 require('./common/MB/Control/SelectAll');
 require('./common/components/TagEditor');
 require('./common/components/sidebar/AcousticBrainz');
+require('./common/components/CritiqueBrainzReview');
```
mwiencek, line 45: let's add some bundles for root/$entity/reviews.js and only load this in those files if possible -- I'm doing a bunch of work on removing common.js and edits.js so I'd prefer not to add anything new here

lib/MusicBrainz/Server/Controller/Role/Rating.pm
```diff
@@ -1,19 +1,34 @@
...
 
 sub ratings : Chained('load') PathPart('ratings')
 {
     my ($self, $c) = @_;
 
     my $entity = $c->stash->{$self->{entity_name}};
-    my @ratings = $c->model($self->{model})->rating->find_by_entity_id($entity->id);
-    $c->model('Editor')->load(@ratings);
-    $c->model('Editor')->load_preferences(map { $_->editor } @ratings);
+    my $entity_properties = $ENTITIES{ $entity->entity_type // 0 };
```
mwiencek, line 13: Do you remember why/when `$entity->entity_type` is undef there?

root/components/EntityTabs.js
```diff
@@ -124,6 +124,13 @@ function buildLinks(
     links.push(buildLink(l('Tags'), entity, 'tags', page));
   }
 
+  if (entityProperties.ratings || entityProperties.reviews) {
+    const ratingsTabTitle = entityProperties.reviews
+      ? l('Reviews')
+      : l('Ratings');
```
mwiencek, line 130: It'd be pretty cool if we displayed the number of ratings/reviews in parentheses here like we do for disc IDs.

reosarevok: I did think about it. But that means hitting CB for the amount of reviews with every page load, is this something we worry about at all?

Also, not sure how to show *both* the number of reviews and ratings if both exist. Should we even try? It might be confusing to show Reviews (0) if there's a lot of ratings.

lib/MusicBrainz/Server/Controller/Role/Rating.pm
reosarevok: It might never be, I might have just copied this as-is from elsewhere, can't remember rn. Guess we can remove it and see if it breaks.

root/static/scripts/common.js
reosarevok: So, just `manifest.js('reviews')` or whatnot?

mwiencek: Yeah. If a single reviews.js file works for all entity types even better

root/components/EntityTabs.js
mwiencek: > But that means hitting CB for the amount of reviews with every page load, is this something we worry about at all?

True, guess I was thinking they were already cached there. That might be something we could load asynchronously in the browser.

> Also, not sure how to show both the number of reviews and ratings if both exist. Should we even try? It might be confusing to show Reviews (0) if there's a lot of ratings.

This is tricky. We could revert to "Ratings" if there are ratings but no reviews (though I guess "Reviews" helps indicate that it's reviewable). Or we could change the tab name to "Ratings and Reviews" if both are supported and display a total count. Or something else.

Needs more thought, so maybe not in this PR. I still like the idea of indicating if there's content though, to save people from wasting a click.

reosarevok: > Ratings and Reviews

That seems too long, that's why I didn't do that.

> I still like the idea of indicating if there's content though, to save people from wasting a click.

I generally agree, although hey, maybe they'll want to write a review if it's empty? :p

reosarevok: I'd merge as-is for now, but if https://tickets.metabrainz.org/browse/CB-427 is implemented as mentioned then we could call that (and maybe cache it) and get all the data in one call rather than 2/3.

reosarevok: Also, all edits support reviews *and* ratings now AFAICT, since Recording is now supported in CB too. But maybe it makes sense to keep this conditional here in case we end up getting more entities that can be rated but not reviewed?

mwiencek (approved):
Rebased this on latest master branch.

This still seems to work fine and implements what we decided, so approving.  (Thought I approved this already.)

I will say that for some releases this has an opposite effect than it intends to.  In many cases where there's only one release in the RG and few relationships, on a desktop monitor you can see all CB reviews without even having to scroll at all.  They're just right in your face.

This now hides them behind a tiny tab which is easy to miss, IMO.  This is probably fine if we increase the tab sizes and add an icon to make the reviews tab stand out more (gold star?) -- perhaps in a future PR.

reosarevok:
I think part of the idea was that by always having a "Reviews" tab there, it actually serves also as a reminder that you can review whatever. I'm kinda wondering whether we should have a reviews tab for releases that points to the RG reviews tab or something too, since I think another reason why reviews are quite hidden is just that they're not on the most common entity type people look at :) But that could be a separate PR anyway.

reosarevok:
Merging for now, we can improve further as needed.

PR #1733, opened 2020-10-06
MBS-10221, MBS-11156: Track parser doesn't update track artists, unsets VA track artists

mwiencek:
### Fix MBS-10221, MBS-11156


mwiencek:
This might resolve MBS-9292 too, though I'm not certain. Some of the comments there mention using the track parser, but the original ticket summary doesn't, so I imagine multiple issues are being conflated there.

reosarevok (approved):
LGTM, didn't test locally but the tests you added pass, so.

yvanzo (approved):

PR #1737, opened 2020-10-07
Convert Selenium tests to JSON5

mwiencek:
This converts the Selenium tests from the old HTML format of the old IDE to JSON5. These are nicer to read, edit, and parse.

This is *not* the same as the new .side format that the new IDE uses (which is just JSON with a different extension). Converting these to the .side format is not possible, because all the commands are different, for one, in addition to the custom commands we've written which will _never_ exist in the IDE. But there are other reasons we wouldn't do this: (1) they can't be run in the IDE anyway because a proxy is required to set the mb-set-database header, and (2) JSON is a horrible format to write by hand, and I still write all Selenium tests that way.

However, having the tests in JSON5 makes the possibility of having a converter from .side to .json5 more feasible. That would allow people who want to create tests via the IDE to do so. I would still recommend cleaning them up by hand, because in many cases the IDE produces bad selectors that are prone to breaking and includes other extraneous commands.

The custom script I wrote to convert these can be found at https://gist.github.com/mwiencek/fecba5369a2c8e078c9562a9a778b7a4


yvanzo (approved):
LGTM but didn’t get selenium container to work with it yet.

yvanzo:
Is _converter from .side to .json5_ a future step or is it missing here?

mwiencek:
> Is converter from .side to .json5 a future step or is it missing here?

That is a future possibility, not done here.

reosarevok (approved):
Seems generally fine but I did not read the 6k lines

PR #1738, opened 2020-10-07
MBS-11158: Make relationship type rowids visible to all

reosarevok:
### Implement MBS-11158

While this *should* be an internal ID only, sadly it's needed for release editor seeding, see the docs for that at https://musicbrainz.org/doc/Development/Release_Editor_Seeding


mwiencek (approved):

yvanzo (approved):
IIUC, this is a quick fix but the proper fix would be to allow using `link_type.mbid` instead in the editor, right? It seems feasible as the MBIDs of all the URL relationship types are already included with `URLCleanup`.

Since it is entered as a temporary quick fix, please just document it as such in message.

PR #1739, opened 2020-10-08
MBS-10281, MBS-11160, MBS-11161: ISE fixes

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10281
https://tickets.metabrainz.org/browse/MBS-11160
https://tickets.metabrainz.org/browse/MBS-11161

### DIFF
--- a/admin/BuildIncrementalSitemaps.pl
+++ b/admin/BuildIncrementalSitemaps.pl
@@ -7,7 +7,7 @@ use FindBin;
 use lib "$FindBin::Bin/../lib";
 
 use MooseX::Runnable::Run;
-use MusicBrainz::Sentry qw( capture_exceptions );
+use MusicBrainz::Errors qw( capture_exceptions );
 
 capture_exceptions(sub {
     run_application 'MusicBrainz::Server::Sitemap::Incremental', @ARGV;
--- a/admin/DumpIncrementalJSON
+++ b/admin/DumpIncrementalJSON
@@ -7,7 +7,7 @@ use FindBin;
 use lib "$FindBin::Bin/../lib";
 
 use MooseX::Runnable::Run;
-use MusicBrainz::Sentry qw( capture_exceptions );
+use MusicBrainz::Errors qw( capture_exceptions );
 
 capture_exceptions(sub {
     run_application 'MusicBrainz::Script::JSONDump::Incremental', @ARGV;
--- a/admin/DumpJSON
+++ b/admin/DumpJSON
@@ -7,7 +7,7 @@ use FindBin;
 use lib "$FindBin::Bin/../lib";
 
 use MooseX::Runnable::Run;
-use MusicBrainz::Sentry qw( capture_exceptions );
+use MusicBrainz::Errors qw( capture_exceptions );
 
 capture_exceptions(sub {
     run_application 'MusicBrainz::Script::JSONDump::Full', @ARGV;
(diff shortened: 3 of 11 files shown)


reosarevok:
Something seems off here. When reproducing MBS-10277 in test.mb I get the first error below, which is about what I'd expect (except for the "Thank you, your edit (#70317788) has been automatically accepted and applied" message for an edit that seemingly does not exist - second new bug?). But in my local server with this PR on, I get the second (blank) screen (?).

Screenshot from 2020-10-13 11-16-12

Screenshot from 2020-10-13 11-16-18

mwiencek:
@reosarevok thanks, hopefully fixed the `Sentry::Raven` requirement. I'm not seeing any blank screen here, though - I was able to reproduce the same bug and get a stack trace. (You could also just insert a `die` in any controller to test.) Are there any errors logged by plackup or the template renderer? (Maybe try restarting the latter.)

mwiencek:
> Seems that the blank screen issue only happens if CATALYST_DEBUG is on.

Thanks. Yeah, the `ErrorInfo` plugin should be disabled when `CATALYST_DEBUG` is set, since that installs its own conflicting `$SIG{__DIE__}` handler IIRC and provides its own error pages. (So you'll need that disabled to test this PR in any case.) I've pushed a fix to get `CATALYST_DEBUG` working again. :)

reosarevok:
For testing MBS-10281 and MBS-11160: on a sample DB, an easy option is to try to link series http://localhost:5000/series/c82868aa-1c67-4e6a-b1f8-bf5c5337cd7e to recording http://localhost:5000/recording/52714b53-1a6f-430a-a7d2-434f0ee601d8 to trigger MBS-10277 (the search won't offer that series I think, its MBID needs to be copy-pasted). This seems to solve both for me locally.

@mwiencek: do you have an easy suggestion to trigger / test MBS-11161?

mwiencek:
> @mwiencek: do you have an easy suggestion to trigger / test MBS-11161?

I don't remember the exact issue I found that triggered this, but it's trivial to reproduce it by inserting a `die` in any Catalyst action after `current_view` is set. On the homepage for example:

```diff
diff --git a/lib/MusicBrainz/Server/Controller/Root.pm b/lib/MusicBrainz/Server/Controller/Root.pm
index 61ce5c8ce6..671798fb9c 100644
--- a/lib/MusicBrainz/Server/Controller/Root.pm
+++ b/lib/MusicBrainz/Server/Controller/Root.pm
@@ -65,6 +65,7 @@ sub index : Path Args(0)
             newestReleases => \@newest_releases,
         },
     );
+    die '';
 }

 =head2 set_language
```

If you apply this to the master branch, the homepage is rendered as if nothing happened (you wouldn't even know an error occurred). On the ise-fixes branch this correctly returns an ISE page.

reosarevok (approved):
All of these seem to work and to be decent improvements :)

yvanzo (commented):
Mojibake still occurs on database connection error, I just stopped the database to reproduce it:
```
08006 DBI connect('dbname=musicbrainz_db;host=db;port=5432','musicbrainz',...) failed: n'a pas pu traduire le nom d'hÃ´te Â« db Â» en adresse : Nom ou service inconnu
```
…
```
MusicBrainz::Server on Catalyst 5.90126
```

Edit 1: With `CATALYST_DEBUG` set to 1.
Edit 2: Setting `CATALYST_DEBUG` to 0 solved it.
Edit 3: It solved displaying the header correctly, not the error message itself which comes from Catalyst but that is probably a lesser issue that can be investigated later on.

yvanzo (approved):

mwiencek:
Hmm, it seems to be an issue with messages from DBI in particular, since I can throw other kinds of errors with unicode characters fine. I opened MBS-11207 to track that.

PR #1741, opened 2020-10-09
MBS-11163: Show type descriptions when editing more entities

reosarevok:
### Implement MBS-11163

We currently show work type descriptions when adding / editing works, but we don't do the same for other entities with types that already have descriptions, such as area, artist (ok, that didn't have them, I just added it), event, place. This extends that code to these four entities. 

Series already has the same kind of popup, implemented somewhat differently. If desired, I can try to replace the implementation there with this same one.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1740

### DIFF
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -343,6 +343,21 @@
     </div>
 [%- END -%]
 
+[%- MACRO type_bubble(form, types) BLOCK -%]
+    <div class="bubble" id="type-bubble">
+      <span id="type-bubble-default" [% IF form.value %] style="display:none" [% END %]>
+        [%- l('Select any type from the list to see its description. If the work
+               doesn’t seem to match any type, just leave this blank.') -%]
+      </span>
+      [% FOREACH id IN form.options %]
+         [%- id = id.value -%]
+         <span [% IF form.value != id %] style="display:none" [% END %] class="type-bubble-description" id="type-bubble-description-[% id %]">
+           <strong>[%- add_colon(l('Description')) -%] </strong>[%- types.$id.l_description -%]
+         </span>
+      [% END %]
+    </div>
+[%- END -%]
+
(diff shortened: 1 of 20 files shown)


root/components/forms.tt
```diff
@@ -352,7 +352,12 @@
       [% FOREACH id IN form.options %]
          [%- id = id.value -%]
          <span [% IF form.value != id %] style="display:none" [% END %] class="type-bubble-description" id="type-bubble-description-[% id %]">
-           <strong>[%- add_colon(l('Description')) -%] </strong>[%- types.$id.l_description -%]
+           <strong>[%- add_colon(l('Description')) -%] </strong>
```
yvanzo, line 355: I know this is already being displayed for work type but it seems neither needed nor helpful.

root/components/forms.tt
```diff
@@ -352,7 +352,12 @@
...
+           [% IF types.$id.l_description %]
+             [%- types.$id.l_description -%]
+           [% ELSE %]
+             [%- lp('(none)', 'description') -%]
```
yvanzo, line 359: Or “Style BDFL has been lazy”? Or just nothing?

yvanzo (approved):
IMHO it can be simplified, but it works already as-is.

root/components/forms.tt
reosarevok: Nothing would show an empty bubble. Right now, only "Other" (which arguably shouldn't exist) shows this, IIRC.

mwiencek (approved):

root/components/forms.tt
yvanzo: Reported in MBS-11177.

root/components/forms.tt
yvanzo: Reported in MBS-11177.

PR #1743, opened 2020-10-12
MBS-11167: Normalize vk.com links to HTTPS

reosarevok:
### Implement MBS-11167

AFAICT vk.com always redirects to HTTPS now.  Given there was no normalization at all before this, there's no downside I can see to standardising in the future.

MBBE-17 exists to convert preexisting VK links to HTTPS.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2615,13 +2615,6 @@ const CLEANUPS = {
       return {result: false};
     },
   },
-  'socialnetwork': {
-    match: [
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
The only downside is to convert 4K `http` for 1K `https`. By the way, there is no need to batch enter edits for this, there is a script to normalize to `https` without edit that can be run as `cron`.

mwiencek (approved):

reosarevok:
I guess if we fix them with a cron job then we have no need to do an override.

yvanzo:
Right, there are only 1K+ URLs, that should take a few hours. Will do only after this has been deployed in production though.

PR #1744, opened 2020-10-13
Convert form_row_name_with_guesscase to React

mwiencek:
Also converts the guess case options dialog with it.

I've only used the new components in the recording edit form as an initial test.

The state handling is a bit messy at the moment, since we don't have a top-level `RecordingForm` component yet (which is where the state would normally go). But eventually things should work themselves out.

~This is marked as WIP because I haven't tested it thoroughly and might want to make changes to the state handling. Also, the style of GC dialog clashes with the bubble doc style, which we may want to rectify, though the popper stuff will be used in the relationship editor too where it more closely matches the current dialog style there...so not sure.~

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 .cpanm/**/*.js
+flow-typed/npm/@sentry/*.js
 perl_modules/**/*.js
 root/static/build/**/*.js
 root/static/lib/**/*.js
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -228,7 +228,7 @@
     [% END %]
 [%- END -%]
 
-[%- MACRO form_row_name_with_guesscase(r, options) BLOCK -%]
+[%- MACRO form_row_name_with_guesscase(r, options) BLOCK # Converted to React at root/static/scripts/edit/components/FormRowNameWithGuessCase.js -%]
   [% WRAPPER form_row %]
     [% r.label('name', options.label || l('Name:')) %]
     [% r.text('name', { class => 'with-guesscase' _ (options.guessfeat ? '-guessfeat' : '') }) %]
--- a/root/static/scripts/common/components/ErrorBoundary.js
+++ b/root/static/scripts/common/components/ErrorBoundary.js
@@ -1,0 +1,45 @@
+/*
+ * @flow strict
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+import * as Sentry from '@sentry/browser';
+
+type Props = {
+  +children: React$Node,
+};
+
+type State = {
+  errorMessage: string,
+};
+
+export default class ErrorBoundary extends React.Component<Props, State> {
+  static getDerivedStateFromError(error: Error): {+errorMessage: string} {
+    return {errorMessage: error.message ?? String(error)};
+  }
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {errorMessage: ''};
+  }
+
+  componentDidCatch(error: Error, info: {componentStack: string, ...}) {
+    Sentry.withScope(function (scope) {
+      scope.setExtra('componentStack', info.componentStack);
+      Sentry.captureException(error);
+    });
+  }
+
+  render(): React.Node {
+    const errorMessage = this.state.errorMessage;
+    if (errorMessage) {
+      return <div className="error">{errorMessage}</div>;
+    }
+    return this.props.children;
+  }
+}
--- a/root/static/scripts/edit/components/GuessCaseOptions.js
+++ b/root/static/scripts/edit/components/GuessCaseOptions.js
@@ -1,0 +1,148 @@
+/*
+ * @flow
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import expand2react from '../../common/i18n/expand2react';
+import bracketed from '../../common/utility/bracketed';
+import getBooleanCookie from '../../common/utility/getBooleanCookie';
+import setCookie from '../../common/utility/setCookie';
+import * as modes from '../../guess-case/modes';
+import gc from '../../guess-case/MB/GuessCase/Main';
+
+/* eslint-disable flowtype/sort-keys */
+export type ActionT =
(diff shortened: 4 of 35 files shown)


root/static/scripts/edit/components/GuessCaseOptions.js
```diff
@@ -0,0 +1,148 @@
...
+  function handleUpperCaseRomanChanged(event) {
+    dispatch({
+      enabled: event.target.checked,
+      type: 'set-upper-case-roman',
+    });
+  }
+
+  return (
+    <div id="guesscase-options">
+      <h4>{l('Guess case options')}</h4>
```
reosarevok, line 111: Why h4 here? The existing dialog uses `.ui-dialog-title` for the title here - is there a specific decision to change the format?

root/static/scripts/edit/components/Dialog.js
```diff
@@ -0,0 +1,182 @@
...
+  );
+
+  return dialogNode ? createPortal(
+    <>
+      <div
+        id={id}
+        onKeyDown={handleInputKeyDown}
+        ref={containerRef}
+        tabIndex="-1"
+      >
```
reosarevok, line 171: FWIW, at least with the example you give here, I feel this div (so, the whole dialog content including the button) could use a bit of left padding not to be so close to the left edge of the popup :) 5 to 10 px probably?

root/static/scripts/edit/components/GuessCaseOptions.js
reosarevok: The relationship editor popups don't have any title headings, so this probably wouldn't affect those in any case - although we might want to actually add titles to at least some of them, like the batch-add ones :)

reosarevok:
Is there a reason why we shouldn't standardize to this kind of bubble for documentation as well?

root/static/styles/forms.less
```diff
@@ -751,3 +752,47 @@ div.rel-editor-dialog {
...
+
+.popper-dialog {
+    visibility: hidden;
+    padding: 0.5em;
+    position: absolute;
+    max-width: 500px;
+    background: @text-white;
+    border: 1px solid #AAA;
+    top: 0;
+    left: 0;
```
reosarevok, line 764: For the example here, at least, a bit higher than 0 would seem good, so that it doesn't cover the icon but appears to come just to the right of it.

mwiencek:
> Is there a reason why we shouldn't standardize to this kind of bubble for documentation as well?

We could do that. It feels a bit off compared to how the bubbles stay on the right half of the page without crossing the fieldset border, and these kind of just...float, but maybe that's just me being not used to it.

reosarevok:
> It feels a bit off compared to how the bubbles stay on the right half of the page without crossing the fieldset border, and these kind of just...float, but maybe that's just me being not used to it.

Isn't that just fixed by adding some left margin until it's to the right of the border?

mwiencek:
> Isn't that just fixed by adding some left margin until it's to the right of the border?

Yeah, but it'd be mildly annoying to parameterize that (it wouldn't make sense in e.g. the relationship editor where the dialogs are meant to float near the opener). Since we'd have to allow configuring the margin, maxWidth, and flippability (which is a popperjs thing where the dialog flips directions if there isn't enough space), just to make it behave like the existing bubbles which are static. And I wonder if any of that even matters. :) Having them float might be good if it lets us increase the form width a bit.

reosarevok:
Increasing the width seems good *and* having them float would probably be better than stuck on the right if we want to work on small-screen scenarios, I assume, but not sure.

root/components/forms.tt
```diff
@@ -228,7 +228,7 @@
     [% END %]
 [%- END -%]
 
-[%- MACRO form_row_name_with_guesscase(r, options) BLOCK -%]
+[%- MACRO form_row_name_with_guesscase(r, options) BLOCK # Converted to React at root/components/FormRowNameWithGuessCase.js -%]
```
reosarevok, line 231:
```suggestion
[%- MACRO form_row_name_with_guesscase(r, options) BLOCK # Converted to React at root/static/scripts/edit/components/FormRowNameWithGuessCase.js -%]
```

mwiencek:
Made a lot of changes here. Split the Dialog component into separate Modal/Popover components, improved the focus handling, and made them easier to reuse.

I'm going to mark this as non-draft and keep the state-handling as it is for now. We can make any necessary improvements as the rest of the page is converted.

reosarevok (approved):
I've been trying this while working on MBS-11168 and it seems to work fine. I did not quite read through all the additions in 080499eeef11c39aafe17d0e12981076976aff24 though.

root/static/styles/forms.less
```diff
@@ -751,3 +751,75 @@ div.rel-editor-dialog {
...
+    &.modal {
+        max-width: 500px;
+
+        @media @wide {
+            max-width: 700px;
+        }
+    }
+
+    &.popover {
+        max-width: 500px;
```
yvanzo, line 793: Maybe round corners the same as with AC bubble:
```suggestion
    &.popover {
        max-width: 500px;
        .border-radius(12px);
```

yvanzo (approved):
Works nicely here. Separate components and their properties make sense.
Did you consider _the official React wrapper_ [React Popper](https://popper.js.org/react-popper/) too?

root/static/scripts/common/components/ErrorBoundary.js
```diff
@@ -0,0 +1,45 @@
+/*
+ * @flow
```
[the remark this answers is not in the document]
mwiencek: Yeah, I will install those from flow-typed in a separate commit.

mwiencek:
> Did you consider the official React wrapper React Popper too?

That's what I tried first, but found it difficult to use and didn't need any of the features it provided. Right now our popper.js integration is pretty simple:

```JS
  React.useLayoutEffect(() => {
    const popper = createPopper(
      buttonRef.current,
      getElementFromRef(dialogRef),
      POPPER_OPTIONS,
    );
    return () => {
      popper.destroy();
    };
  }, [dialogRef, buttonRef]);
```

If there's something more advanced we need that the library can do we can switch later, but avoiding another dependency is nice too. :)

mwiencek:
Pushed a few small changes:

 * Add libdefs for `@sentry/browser` from flow-typed, so we can mark more modules as strict.
 * Copied the `NormalModuleReplacementPlugin` technique from https://github.com/metabrainz/musicbrainz-server/pull/1758 so that we can use top-level imports for popperjs on the server (they'll just be mapped to the empty no-op module).
 * Added a 12px border radius to the popover.
 * Moved root/static/scripts/recording.js to root/static/scripts/recording/form.js (I'd like the former to be for code that needs to execute on all pages using the recording layout, not for the edit form).

.eslintignore
```diff
@@ -1,4 +1,5 @@
 .cpanm/**/*.js
+flow-typed/npm/@sentry/*.js
```
mwiencek, line 2: This contained Flow syntax that eslint didn't like (mostly related to semicolons and lack of indexer key names). I've only ignored this specific libdef because other ones we've written ourselves, and I'd still like those checked.

PR #1751, opened 2020-10-16
Upgrade to Flow 0.136.0

mwiencek:
It's now an error for generics to escape outside of the scope they're defined in. Which usually just means that if a variable's type contains a generic type variable, then we can't pass that variable around without also passing the associated generic.

### DIFF
--- a/root/layout/components/Search.js
+++ b/root/layout/components/Search.js
@@ -13,7 +13,15 @@ import SearchIcon from '../../static/scripts/common/components/SearchIcon';
 import DBDefs from '../../static/scripts/common/DBDefs';
 import {compare} from '../../static/scripts/common/i18n';
 
-const TYPE_OPTION_GROUPS = [
+type SearchOptionValueT =
+  (() => string) | null;
+
+type SearchOptionGroupT = {
+  +[optionName: string]: SearchOptionValueT,
+  ...
+};
+
+const TYPE_OPTION_GROUPS: $ReadOnlyArray<SearchOptionGroupT> = [
   {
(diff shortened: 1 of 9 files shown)


root/layout/components/Search.js
```diff
@@ -46,34 +54,39 @@ const TYPE_OPTION_GROUPS = [
...
-            value={key}
-          >
-            {text}
-          </option>
-        );
-      })
-    ))}
+    {TYPE_OPTION_GROUPS.map((group) => (
+      Object.entries(group)
+        // $FlowIssue[incompatible-call]
```
reosarevok, line 75: Can't group be typed to avoid this?

reosarevok (approved):

root/layout/components/Search.js
mwiencek: It's already typed, but the issue is that `Object.entries` is hard-coded to return `Array<[string, mixed]>`. You can't use `mixed` without either refining it or casting it to `any` first, and both of those are worse solutions.

reosarevok: Well, that sucks :(

mwiencek: I suspect they'll fix the type eventually (I see they're working on improving generics checking as we speak). So the error suppression isn't so bad in the meantime: unlike `any` it doesn't prevent marking the file as strict, it only affects `incompatible-call` and not other kinds of errors, and we can run Flow with `--include-warnings` to find unneeded error suppression comments in the future.

yvanzo (approved):

PR #1752, opened 2020-10-19
MBS-11186: Add missing space between class names

reosarevok:
### Fix MBS-11186

When an editor has entered an edit and yet a vote type (approval, since others are not allowed in this case)
applies to the same editor, votename doesn't separate the applied classes, and adds the class "Approveowner" which unsurprisingly does nothing.

Also removing extra spaces that were being added to the HTML while at it.


root/components/common-macros.tt
```diff
@@ -887,11 +887,14 @@ note-[%- edit.id -%]-[%- loop.count() -%]
 [%- END -%]
 
 [% MACRO votename(note) BLOCK %]
-  [% edit = note.edit %]
+  [%- is_owner = edit.editor_id == note.editor_id -%]
+  [%- edit = note.edit -%]
```
mwiencek, line 891: Shouldn't `edit` be declared first since it's referenced in `is_owner = edit.editor_id == note.editor_id`?

reosarevok: It seemed to work fine, but I'm not quite sure how now that you mention this. Maybe the equality was only checked when is_owner was used? I'm confused. Just TT things I guess. Anyway, will make the change.

mwiencek (approved):

yvanzo (approved):

PR #1754, opened 2020-10-19
MBS-11130: Clarify it's not possible to subscribe to VA / [no label]

reosarevok:
### Implement MBS-11130

While special entities don't show the subscribers link on the sidebar, it's still possible to navigate to /subscribers manually. There, a "Subscribe" link that did nothing could be found.

This makes it so that entities that do not support subscribing just state that on that page instead.

NB: A lot of the changes here are just extra indent.

### DIFF
--- a/root/entity/Subscribers.js
+++ b/root/entity/Subscribers.js
@@ -10,11 +10,14 @@
 import * as React from 'react';
 
 import EditorLink from '../static/scripts/common/components/EditorLink';
+import isSpecialPurpose
+  from '../static/scripts/common/utility/isSpecialPurpose';
 import chooseLayoutComponent from '../utility/chooseLayoutComponent';
 
 type Props = {
   +$c: CatalystContextT,
   +entity: CoreEntityT | CollectionT | EditorT,
+  +isSpecialEntity: boolean,
   +privateEditors: number,
   +publicEditors: $ReadOnlyArray<EditorT>,
   +subscribed: boolean,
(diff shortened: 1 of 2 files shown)


mwiencek:
This seems like an odd fix if it's hidden on a page that isn't linked from anywhere. It'd make more sense to me to either mention it in the sidebar and make `/subscribers` 404, or allowing subscribing to these entities.

reosarevok:
Hmm. We don't generally block accessing /whatever pages on entities that don't use them if others of the same kind do, and I feel it might make sense to actually be able to go there and be told what happens if you really want to see it. But I'm also ok with a 404 if you prefer.

mwiencek:
I don't mind keeping the page too much, just dunno if this clarifies why the option is missing in the sidebar if you don't know to go to this page first. :) Though we also mentioned on IRC that we might want to just allow subscribing and put a disclaimer that the special-purpose labels will clog the subscription edits page?

reosarevok:
I mean, this isn't meant to clarify why it's missing from the sidebar, but to replace the current "Subscribe" link that does nothing with no explanation if you navigate to this page :) 

But we can also do that, if we prefer :) I don't think we reached a clear conclusion but that might work too.

root/entity/Subscribers.js
```diff
@@ -45,108 +48,116 @@ const Subscribers = ({
     >
       <h2>{l('Subscribers')}</h2>
 
-      {(publicEditors.length || (privateEditors > 0)) ? (
+      {isSpecialPurpose(entity) ? (
+        <p>
+          {l('This is a special entity and does not support subscriptions.')}
```
mwiencek, line 53: We use "special purpose" in most strings, so just for consistency
```suggestion
          {l('This is a special purpose entity and does not support subscriptions.')}
```

mwiencek (approved):
Let's ignore my earlier rant.  Not the worst bug in the world but hey, this seems better than the current situation at least.

reosarevok:
@mwiencek: @yvanzo actually suggested to drop this, and actually allow people to subscribe to the entities - and that if we're afraid that they'll flood the subscriptions list, we could provide anyone who subscribes to any special entities with a separate subscription page with / without them. Opinion?

mwiencek:
Yeah, I was wondering that before too (I mentioned it in an earlier comment here).  I guess I'd favor that if (1) it's ever useful for people to subscribe to those and (2) we can delete a decent amount of code for it.  But I'm a bit skeptical of (1) and not sure about (2) either since we still have things like "Deleted Artist" which we'd have to prevent, I think (unless that just 404s first?).  So I'm okay with merging the message improvement here unless you think allowing these would be helpful and make the code and UI clearer.

yvanzo:
FTR: Commented on IRC.

PR #1755, opened 2020-10-19
MBS-11192: Add voting-icon for Approve

reosarevok:
### Implement MBS-11192

It seems weird we don't actually show whether the edit note writer approved the edit, given that's even stronger than a Yes vote (which we do specifically indicate).

This adds a checkmark icon for approval, from the same icon set as the existing yes and no icons (recolored to gold by CatCat to differentiate the color, thanks!)

Screenshot from 2020-10-19 14-03-12


root/static/styles/edit.less
```diff
@@ -164,6 +164,16 @@ table.vote-tally {
         }
     }
 
+    h3.Approve {
+        a img {
+            border-color: @yes-border;
+        }
+
+        .voting-icon {
+            background-image: data-uri('../images/icons/tickextrasharp.png');
```
mwiencek, line 173: should be tick.png right?
```suggestion
            background-image: data-uri('../images/icons/tick.png');
```

reosarevok: It really should, yes. Sigh. Thanks!

mwiencek (approved):

yvanzo (approved):

PR #1757, opened 2020-10-20
MBS-11196: Whitelist saisaibatake.ame-zaiku.com for instruments

reosarevok:
### Implement MBS-11196

This is IROM's personal (but fairly complete) instrument database. Given we're planning to use their images, we might as well link back to the data as well!


lib/MusicBrainz/Server/Entity/URL/Gakki.pm
```diff
@@ -0,0 +1,23 @@
+package MusicBrainz::Server::Entity::URL::Gakki;
+
+use Moose;
+use utf8;
+
+extends 'MusicBrainz::Server::Entity::URL';
+with 'MusicBrainz::Server::Entity::URL::Sidebar';
+
+sub sidebar_name { '私家版楽器事典' }
```
yvanzo, line 9: It seems to just mean “Private musical instrument encyclopedia”. Might be worth mentioning whose it is. The whole information is given in homepage’s footer.
```suggestion
sub sidebar_name { 'IROMBOOK 私家版楽器事典' }
```

root/static/scripts/common/constants.js
```diff
@@ -149,6 +149,7 @@ export const FAVICON_CLASSES = {
   'rockipedia.no': 'rockipedia',
   'rolldabeats.com': 'rolldabeats',
   'runeberg.org': 'runeberg',
+  'saisaibatake.ame-zaiku.com': 'gakki',
```
yvanzo, line 152:
```suggestion
  'saisaibatake.ame-zaiku.com/gakki/': 'gakki',
```

reosarevok: I left it without because we don't have /wiki for generasia, but maybe we should just have that too.

lib/MusicBrainz/Server/Entity/URL/Gakki.pm
reosarevok: Apparently the actual meaning is more like "self-made" than "private", but still, that might make sense - if that's correct in Japanese. I'll ask a JP speaker.

reosarevok: Confirmed that it's grammatical in Japanese, making the change :)

mwiencek (approved):

root/static/scripts/common/constants.js
yvanzo: I suggested it because you were using it for [URL specialization](https://github.com/metabrainz/musicbrainz-server/commit/02c8301d150a37bbc9ef84c97ea8f86de386b93d#diff-c7e5aef95d55e2d6e1d96236e898f6701a422ce0ced4a2b4c701a97d3b4ff639R71) and [URL cleanup](https://github.com/metabrainz/musicbrainz-server/commit/02c8301d150a37bbc9ef84c97ea8f86de386b93d#diff-a808bbcd17cb3e956e9ac0fce554d75f6e9db02e020bc03ca543c198ac8f8e06R1356) already, and I saw it was used for [dancedb favicon class](https://github.com/metabrainz/musicbrainz-server/blob/v-2020-10-19/root/static/scripts/common/constants.js#L163) too.

> we don't have /wiki for generasia, but maybe we should just have that too.

:+1:

yvanzo (approved):

root/static/scripts/common/constants.js
reosarevok: Ok, pushed that change.

PR #1758, opened 2020-10-20
MBS-11195: Convert the artist-credit-renamer to React

mwiencek:
I tested with the extreme case of 9ddd7abc-9e1b-471d-8031-583bc6bc8be9 and tried to guarantee acceptable performance by caching the rows and keeping the checkbox state local.

Previews are now displayed at the bottom instead of inline, and use the `diffArtistCredits` utility.

The list is collapsed if there are more than 10 artist credits. (I've used the hide technique that annotations use, since actually adding/removing hundreds of items to the DOM is extremely slow.)

This also fixes MBS-6666 by hiding the fieldset if the artist name hasn't changed.

### DIFF
--- a/root/static/scripts/artist.js
+++ b/root/static/scripts/artist.js
@@ -7,6 +7,8 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import './artist/components/ArtistCreditRenamer';
+
 import typeBubble from './edit/typeBubble';
 
 const typeIdField = 'select[name=edit-artist\\.type_id]';
--- a/webpack.server.config.js
+++ b/webpack.server.config.js
@@ -40,10 +40,10 @@ module.exports = {
   externals: [
     nodeExternals({
       /*
-       * @popperjs is resolved to root/static/scripts/empty.js on the server.
-       * See NormalModuleReplacementPlugin below.
+       * jquery and @popperjs are resolved to root/static/scripts/empty.js
+       * on the server. See NormalModuleReplacementPlugin below.
        */
-      whitelist: [/@popperjs/],
+      whitelist: [/(jquery|@popperjs)/],
       modulesFromFile: true,
     }),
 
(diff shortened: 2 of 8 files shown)


webpack.server.config.js
```diff
@@ -70,8 +70,11 @@ module.exports = {
   },
 
   plugins: [
+    new webpack.NormalModuleReplacementPlugin(
+      /jquery/,
+      path.resolve(dirs.SCRIPTS, 'empty.js'),
+    ),
```
reosarevok, line 76: You're using `IgnorePlugin` for `@popperjs` in https://github.com/metabrainz/musicbrainz-server/pull/1744 - should this approach also be used there then?

root/static/scripts/artist.js
```diff
@@ -1,12 +1,14 @@
 /*
- * @flow strict-local
+ * @flow
```
reosarevok, line 2: Doesn't seem needed, unless I'm missing something? It's strict-local, not strict.

reosarevok (commented):
The way the preview displays seems strange to me - we certainly don't need to take half the available space for the "Artist Credit" label, at the very least. Ideally we'd drop the label (we know these are artist credits) and just have the two columns use the available space. If not, at least let's try to make more space available for the actual info :) 

Screenshot from 2020-10-21 12-34-05

It can be a bit confusing that the previews are shown in the order you selected them, not the order the ACs are listed on the list (so if you deselect one then select it again, it will new be at the bottom). Probably quite minor but if it's trivial to maintain a consistent sort then that would be nice.

root/static/scripts/artist.js
mwiencek: Right, I copied this from another branch where it was needed. Will revert it from here.

mwiencek:
Updated to maintain a consistent sort order, and fixed the preview display to use less space.

I found the `EditArtistCredit` component to be less than ideal in other ways -- I think the inclusion of disambiguation comments is unnecessary and makes the changes impossible to read, and the lack of diff highlighting doesn't help either. Since we have a `diffArtistCredits` utility written already, I changed it to just use that.

reosarevok (approved):
Seems to work much better now. There's a noticeable wait / freeze if I modify the name for Tchaikovsky but I guess it takes a bit to load all those names.

yvanzo (approved):

PR #1759, opened 2020-10-20
Switch to new JSX transform

mwiencek:
https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html


.eslintrc.yaml
```diff
@@ -222,7 +222,7 @@ rules:
   react/no-will-update-set-state: error
   react/prefer-es6-class: off
   react/prop-types: off
-  react/react-in-jsx-scope: error
+  react/react-in-jsx-scope: off
```
reosarevok, line 225: Maybe comment about why they're off, for future reference? We could also just remove them, not sure which one would be less confusing.

reosarevok (approved):
Seems reasonable. Navigated around a bit and it also seemed to work. Should we upgrade to React 17 next?

.eslintrc.yaml
mwiencek: A comment sounds good. I tend to prefer setting things to `off` just in case the defaults change.

mwiencek:
> Should we upgrade to React 17 next?

Yes, once this is merged I'll work on that. :) (It was released right after I opened this PR.)

mwiencek:
@brainzbot test this please

yvanzo (approved):

PR #1760, opened 2020-10-21
MBS-11194: Fix ISE when historic EditRelationship data is broken 

reosarevok:
### Fix MBS-11194

We fixed this for missing newRel with MBS-10901 (https://github.com/metabrainz/musicbrainz-server/pull/1575) but apparently it also happens the other way around in some cases.

Also reverting a couple changes from that PR that seem unneeded.


mwiencek (approved):
Let's hope there's not a case where both are null. :P

yvanzo (approved):

PR #1761, opened 2020-10-21
MBS-11197: Add validation for Mainly Norfolk links and normalize

reosarevok:
### Implements MBS-11197

The site redirects to HTTPS. I'm also normalizing to drop www. since both options are valid and the vast majority of our links don't include it (the site just accepts either without redirecting).

Works and artists are straightforward. I chose release rather than release group for 'records' because the site gives a format and catno, from what I can see.

This drops index.hml from artist pages because it's unneeded. It does not drop #sections because it seems in some cases different album pages are actually sections of a larger discography but we might still want to link to them.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1933,6 +1933,22 @@ const CLEANUPS = {
       url = url.replace(/^https:\/\/mainlynorfolk\.info\/([^/]+)(?:\/index\.html)?$/, 'https://mainlynorfolk.info/$1/');
       return url;
     },
+    validate: function (url, id) {
+      if (id === LINK_TYPES.otherdatabases.artist) {
+        return {result: /^https:\/\/mainlynorfolk\.info\/(?:[^/]+)\/$/.test(url)};
+      }
+      const m = /^https:\/\/mainlynorfolk\.info\/(?:[^/]+)\/(records|songs)\/(?:[^/]+)$/.exec(url);
+      if (m) {
+        const prefix = m[1];
+        switch (id) {
+          case LINK_TYPES.otherdatabases.release:
+            return {result: prefix === 'records'};
+          case LINK_TYPES.otherdatabases.work:
+            return {result: prefix === 'songs'};
+        }
+      }
+      return {result: false};
+    },
   },
   'maniadb': {
     match: [new RegExp('^(https?://)?(www\\.)?maniadb\\.com', 'i')],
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1929,7 +1929,26 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?(www\\.)?mainlynorfolk\\.info', 'i')],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:www\.)?mainlynorfolk\.info\//, 'https://mainlynorfolk.info/');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?mainlynorfolk\.info\//, 'https://mainlynorfolk.info/');
+      // Drop unneeded index.html from artist pages
+      url = url.replace(/^https:\/\/mainlynorfolk\.info\/([^/]+)\/index\.html$/, 'https://mainlynorfolk.info/$1/');
+      return url;
+    },
```
yvanzo: This part of the second commit would better fit the first commit which is about normalization (https & -www).
By the way, the first commit message should probably not misleadingly specify “to HTTPS” in header.
I just removed "and normalize to HTTPS" from ticket’s summary as it also drops www and index.html and as normalization is just the usual prerequisite to validation.


The comment in code here is probably not needed.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1929,7 +1929,26 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?(www\\.)?mainlynorfolk\\.info', 'i')],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:www\.)?mainlynorfolk\.info\//, 'https://mainlynorfolk.info/');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?mainlynorfolk\.info\//, 'https://mainlynorfolk.info/');
+      // Drop unneeded index.html from artist pages
+      url = url.replace(/^https:\/\/mainlynorfolk\.info\/([^/]+)\/index\.html$/, 'https://mainlynorfolk.info/$1/');
```
yvanzo, line 1934: We might want to add an ending slash `/` if it is missing:

```suggestion
      url = url.replace(/^https:\/\/mainlynorfolk\.info\/([^/]+)(?:\/index\.html)?$/, 'https://mainlynorfolk.info/$1/');
```

yvanzo (approved):

mwiencek (approved):

PR #1765, opened 2020-10-28
MBS-11204: Avoid crashing if entity1_id is missing

reosarevok:
### Fix MBS-11204

Equivalent of MBS-10833 (https://github.com/metabrainz/musicbrainz-server/pull/1543) for entity0_id.

Some old edits (at least edit 6774077) are missing this. Since this is later loaded via linkedEntities, we need to use the same "fake entity ID" method we use for entity0 to access it (and since _display_relationships could hit several cases where this is needed, the fake ID needs to be distinct every time).


mwiencek (approved):

yvanzo (approved):
LGTM

PR #1766, opened 2020-10-29
MBS-1424: Add a "First release date" field to recordings

mwiencek:
This doesn't go through `release_group_meta` (since the earliest release in an RG may not contain a particular recording in the RG), but calculates it directly from `release_country` and `release_unknown_country`.

Similar to how we added an `OAUTH2_ENABLE_PKCE` flag to enable that feature in production, I've added a new `ACTIVE_SCHEMA_SEQUENCE` flag to toggle all schema 26 related features in one go.

I included some basic display code to show it in the sidebar just as a test, but I'll leave it to @reosarevok to decide on a proper way.


admin/sql/CreateViews.sql
```diff
@@ -83,6 +83,24 @@ CREATE OR REPLACE VIEW work_series AS
...
+      JOIN release ON (release.id = medium.release)
+      LEFT JOIN (
+          SELECT release, date_year, date_month, date_day
+            FROM release_country
+           UNION
+          SELECT release, date_year, date_month, date_day
+            FROM release_unknown_country
+      ) release_dates
+        ON release_dates.release = release.id;
+
```
reosarevok, line 103: Is having a view better than just doing the query in the function as with `set_release_group_first_release_date` then?

reosarevok (commented):
Reading through the code seems sensible, at least :) I'll actually set it up later today for further checking.

admin/sql/CreateViews.sql
mwiencek: The view helps a bit because there are actually two functions added:

 * `set_recordings_first_release_dates` - this updates the release dates for specific recording IDs (when a track is inserted/updated/deleted).
 * `set_releases_recordings_first_release_dates` - this updates the release dates for all recordings for specific release IDs (when a release event is inserted/updated/deleted).

Both of these functions require mostly identical queries except for the `recording = ?` and `release = ?` parts, so I put the common part of the query into a view.

I could have also implemented it where we just pass `set_recordings_first_release_dates` a giant list of recordings in the second case, but PG can't optimize that as well and I didn't want to overload these triggers.

admin/sql/CreateIndexes.sql
```diff
@@ -172,6 +172,8 @@ CREATE INDEX event_tag_idx_tag ON event_tag (tag);
 CREATE INDEX event_tag_raw_idx_tag ON event_tag_raw (tag);
 CREATE INDEX event_tag_raw_idx_editor ON event_tag_raw (editor);
 
+CREATE INDEX first_recording_release_date_idx ON first_recording_release_date (year, month, day);
```
mwiencek, line 175: This index might be useful if we want to browse/sort recordings by year/date, but should probably be removed unless or until we have a plan for that. (That may only be useful in the context of artists, and this table doesn't have artist info.)

reosarevok: Allowing to sort by date (or filter?) doesn't seem like a terrible idea, at least.

mwiencek:
All current test failures are due to the mmd-schema from mb-solr being outdated.

reosarevok (approved):
Showing dates as year only like now seems legit. I tested this a few days ago and seemed to work fine after small fixes, so probably good to ship to beta and see if anyone screams.

reosarevok:
One thing we might want to do is to document how this is calculated in case people want to change it. Although we don't specify for RG dates, so it might not be needed and people might figure it out by comparison.

admin/sql/CreateTables.sql
```diff
@@ -847,6 +847,13 @@ CREATE TABLE event_type ( -- replicate
     gid                 uuid NOT NULL
 );
 
+CREATE TABLE first_recording_release_date (
```
yvanzo, line 850: `recording_first_release_date` would better fit the current naming scheme where _`entity_*`_ tables relate to _entity_ type; It would also better match the rest of your code in this pull request that refer to `*first_release_date*` only.

yvanzo (commented):
I released your patch to MMD Schema and made it available from MB Solr:
- https://github.com/metabrainz/mmd-schema/releases/tag/v-2020-11-12
- https://github.com/metabrainz/mb-solr/releases/tag/v3.3.1

I pushed commit https://github.com/metabrainz/musicbrainz-server/pull/1766/commits/db8eb0eed9cf85a7062ccf69f9b91c366dd8399c to pass continuous integration.

It still requires to create a new `musicbrainz-tests` image though.

yvanzo:
Note: Updating `sir` requires to create an in-house version `26.0.dev1` of `mbdata` that supports the new table. So, the next step for MBS-1424 / SEARCH-218 is to merge this pull request, so as to make final SQL code final available for `mbdata` development.

admin/sql/CreateTables.sql
mwiencek: Good point! I'll working on updating that soon.

lib/DBDefs/Default.pm
```diff
@@ -364,9 +364,15 @@ sub COVER_ART_ARCHIVE_DOWNLOAD_PREFIX { "//coverartarchive.org" };
...
+#  * PKCE for OAuth 2.0 clients.
+#    (admin/sql/updates/20200914-oauth-pkce.sql)
+#  * first_recording_release_date table.
+#    (admin/sql/updates/20201028-mbs-1424.sql)
+sub ACTIVE_SCHEMA_SEQUENCE { 25 }
+
 # Enable PKCE for OAuth 2.0 clients.
-# Currently requires a schema change: admin/sql/updates/20200914-oauth-pkce.sql
-sub OAUTH2_ENABLE_PKCE { 0 }
+sub OAUTH2_ENABLE_PKCE { shift->ACTIVE_SCHEMA_SEQUENCE == 26 }
```
yvanzo, line 375: PKCE is not going to be dropped for schema 27 but we may forget about this line before:

```suggestion
sub OAUTH2_ENABLE_PKCE { shift->ACTIVE_SCHEMA_SEQUENCE >= 26 }
```

PR #1767, opened 2020-10-29
Eslint fixes (few more rules for CI)

reosarevok:
One commit per rule, went for ones that didn't require many changes for a start so we can enforce them in CI.

### DIFF
--- a/t/selenium.js
+++ b/t/selenium.js
@@ -325,7 +325,7 @@ async function handleCommand({command, file, target, value}, t) {
   t.comment(
     command +
     ' target=' + utf8.encode(JSON.stringify(target)) +
-    ' value=' + utf8.encode(JSON.stringify(value))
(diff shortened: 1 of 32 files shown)


t/selenium.js
```diff
@@ -596,8 +596,10 @@ async function runCommands(commands, t) {
       'psql',
       [
         ...hostPort,
-        '-U', systemDb.user,
-        '-c', `
+        '-U',
+        systemDb.user,
+        '-c',
+        `
```
mwiencek, line 604: It doesn't hurt readability here so seems fine, though in general I'd ignore the rule for command line flags since it's much more logical to put the flag on the same line as the value

mwiencek (approved):

yvanzo (approved):
I checked `master` branch currently at bcb3ce62e36a9aea76a05e276cf20f12e2c7d8c4 is still passing test.

PR #1768, opened 2020-10-29
MBS-10664: Allow more tags in expand2react

reosarevok:
### Fix MBS-10664

This adds html tags legitimately generated by our MediaWiki syntax (annotations, bios, etc) to htmlTagName. While the
MediaWiki syntax parser sometimes generates invalid tags or puts them in the wrong order, this should unbreak some bios and cause no problems.

Originally part of https://github.com/metabrainz/musicbrainz-server/pull/1653


mwiencek (approved):

yvanzo (approved):

jesus2099:
https://beta.musicbrainz.org/user/jesus2099 is still broken.
There is a last problem with `<strong>` and `<em>`.

    “<strong><em>Reveal my email address</strong></em>”

reosarevok:
Yeah, that is a known bug with the Perl mediawiki syntax parser. We will need to either fix it or migrate away from it

PR #1770, opened 2020-10-29
Add flow types to utility/dates.js

reosarevok:
Added this during work on MBS-11168 but I'd rather get it checked separately and merged already, since that's going to be big anyway.

### DIFF
--- a/root/static/scripts/edit/utility/dates.js
+++ b/root/static/scripts/edit/utility/dates.js
@@ -1,4 +1,5 @@
 /*
+ * @flow strict
  * Copyright (C) 2012-2014 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/utility/dates.js
```diff
@@ -35,37 +46,40 @@ export const isDateValid = function (y, m, d) {
   }
 
   // Invalid number of days based on the year.
-  if (d < 1 || d > 31 || d > getDaysInMonth(y, m)) {
+  if (d < 1 || d > 31 || y != null && m != null && d > getDaysInMonth(y, m)) {
```
mwiencek, line 49: We have some tests for `isDateValid` in root/static/scripts/tests/utility.js -- can we add a couple that break without this change?

Some `()` are always nice to clarify the grouping even if it doesn't affect the operator precedence. (Apparently this line is exactly 79 chars, hence the added newlines.)
```suggestion
  if (
    d < 1 || d > 31 ||
    (y != null && m != null && d > getDaysInMonth(y, m))
  ) {
```

mwiencek (approved):

yvanzo (approved):
:shipit:

PR #1772, opened 2020-11-03
MBS-9388: Show relationship credits on Aliases tab

reosarevok:
### Implement MBS-9388

Two commits here.

The first we should change for sure. It changes the blurb of the artist credits section of the aliases page to make it less misleading: we also have relationship credits, which are not artist credits, so we shouldn't claim this is "all the ways the artist is credited in the database".

The second adds a list of relationship credits to the same page. We display all other names given to the entity in the DB there, so it probably makes sense to display these too. I do somewhat worry that it will make alias pages for artists and areas with tons of relationships pretty slow though.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Area.pm
+++ b/lib/MusicBrainz/Server/Controller/Area.pm
@@ -6,6 +6,7 @@ BEGIN { extends 'MusicBrainz::Server::Controller'; }
 with 'MusicBrainz::Server::Controller::Role::Load' => {
     model           => 'Area',
     relationships   => {
+        all         => ['aliases'],
         cardinal    => ['edit'],
         subset      => {
             show => [qw( area artist label place series instrument release_group url )],
--- a/lib/MusicBrainz/Server/Controller/Role/Alias.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Alias.pm
@@ -1,5 +1,9 @@
 package MusicBrainz::Server::Controller::Role::Alias;
 use Moose::Role -traits => 'MooseX::MethodAttributes::Role::Meta::Role';
+use List::AllUtils qw( uniq );
+use MusicBrainz::Server::Constants qw(
+    %ENTITIES_WITH_RELATIONSHIP_CREDITS
+);
 use MusicBrainz::Server::ControllerUtils::Delete qw( cancel_or_action );
 
 requires 'load';
@@ -47,6 +51,16 @@ sub aliases : Chained('load') PathPart('aliases')
         aliases => $aliases,
         entity => $entity,
     );
+
+    if ($ENTITIES_WITH_RELATIONSHIP_CREDITS{$self->{entity_name}}) {
+        my $relationships = $entity->{relationships};
+        my @relationship_credits = uniq sort {$a cmp $b} grep { $_ ne '' } map {
+            $_->direction == 2
+                ? $_->entity1_credit
+                : $_->entity0_credit
+        } @$relationships;
+        $props{relationshipCredits} = \@relationship_credits;
+    }
 
     $c->stash(
         # "aliases" needs to remain here for JSON-LD serialization
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -9,7 +9,7 @@ BEGIN { extends 'MusicBrainz::Server::Controller'; }
 with 'MusicBrainz::Server::Controller::Role::Load' => {
     model           => 'Artist',
     relationships   => {
-        all         => ['relationships'],
+        all         => ['aliases', 'relationships'],
         cardinal    => ['edit'],
         subset      => { split => ['artist'], show => ['artist', 'url'] },
         default     => ['url']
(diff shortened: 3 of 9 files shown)


lib/MusicBrainz/Server/Controller/Area.pm
```diff
@@ -6,6 +6,7 @@ BEGIN { extends 'MusicBrainz::Server::Controller'; }
 with 'MusicBrainz::Server::Controller::Role::Load' => {
     model           => 'Area',
     relationships   => {
+        all         => ['aliases'],
```
mwiencek, line 9: Instead of loading all relationships, we should probably write a more efficient query that just selects the distinct credit  strings from the DB.

mwiencek (commented):
Rebased this on top of latest master, tested it, and it seems to work fine (though I thing we could implement it more efficiently).  Any other reason it's a draft?

reosarevok:
It was a draft because of " I do somewhat worry that it will make alias pages for artists and areas with tons of relationships pretty slow though." - I was thinking whether it required a view or something. Maybe just a separate query is fast enough though? I can try to implement it that way.

lib/MusicBrainz/Server/Controller/Area.pm
reosarevok: Did that now.

lib/MusicBrainz/Server/Controller/Role/Alias.pm
```diff
@@ -49,6 +52,14 @@ sub aliases : Chained('load') PathPart('aliases')
         entity => $entity->TO_JSON,
     );
 
+    if ($ENTITIES_WITH_RELATIONSHIP_CREDITS{$self->{entity_name}}) {
```
mwiencek, line 55: Assuming `$self->{entity_name}` comes from `Controller::Role::Load`, I'm not sure that will always work, e.g. we use `rg` for release groups. `$entity->entity_type` should work.

lib/MusicBrainz/Server/Controller/Role/Alias.pm
```diff
@@ -49,6 +52,14 @@ sub aliases : Chained('load') PathPart('aliases')
...
+        my @relationship_credits = $c->model('Relationship')->load_credits(
+            $entity->entity_type,
+            [ $entity->id ],
```
mwiencek, line 58: This looks wrong because the function accepts `@ids`, not `$ids`. _But_, it should work accidentally (I didn't test it yet) because you're only passing one array ref, and because of the way `load_credits` is written: since you don't try to take a reference to `@ids`, only use it as a list, the single array ref in the list matches up with each `any(?)`.

Since we don't seem to have any need for `load_credits` to accept more than one ID at present, I'd suggest changing the signature to only take one `$id`, and remove the `any`s (which in some cases I've seen can measurably speed up queries, even if you're only passing one value to the `any`).

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -230,6 +230,55 @@ sub _load_related_info {
...
+            push @params, @ids;
+        } elsif ($type eq $type1) {
+            $query = $type1_query;
+            push @params, @ids;
+        }
+
+        my $results = $self->sql->select_single_column_array($query, @params);
+        push @credits, @$results;
+    }
+    my @sorted_credits = uniq sort {$a cmp $b} @credits;
```
mwiencek, line 278: `UNION` already eliminates duplicates.
```suggestion
    my @sorted_credits = sort { $a cmp $b } @credits;
```

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -230,6 +230,55 @@ sub _load_related_info {
...
+    return @sorted_credits;
```
mwiencek, line 279: I'd suggest returning an array ref, which I've read is more efficient for large lists.
```suggestion
    return \@sorted_credits;
```

lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -230,6 +230,55 @@ sub _load_related_info {
     $self->load_entities(@rels);
 }
 
+sub load_credits
```
mwiencek, line 233: `Data::` methods starting with `load_` typically load some data onto an entity object (i.e. fill some Moose attribute with data). Maybe `get_credits` is a bit clearer for indicating it's only for fetching data?

mwiencek (changes requested):

lib/MusicBrainz/Server/Data/Relationship.pm
reosarevok: But there might be many relationships with the same credit, so we still need to take care of those. I guess the ones without the UNION should do SELECT DISTINCT in that case?

reosarevok: Oh, no, actually that won't work, because the SQL queries are *per entity*. `uniq` here ensures that there are no duplicates also *between different relationship tables*. Not sure if doing DISTINCT *as well* is good tho, for `uniq` to have to run over less rows?

mwiencek: Oh, for some reason I was thinking there was only one query. Can you merge them into one? They all only select a single column of the same type, so should all `UNION` fine. (You don't need any `DISTINCT` or `uniq` in that case.)

reosarevok: Finally implemented this, less than two years later!

PR #1773, opened 2020-11-04
Align comments with new code (very minor low priority PR)

jesus2099:
# Problem

MBS-9505's 618e9109d151562e7787a1f76d87b75916d8117f changed auto edit grace period from 1h to 1 day but associated comment (just for coders) is still mentioning 1h.

# Solution

I just updated the comments

It should have no impact on MBS runtime.


yvanzo (approved):
Thank you, good catch.

PR #1774, opened 2020-11-04
MBS-11111: Set edit_pendings on recordings merged with release

reosarevok:
### Implement MBS-11111

If a release merge is going to merge recordings, we should also show those recordings as having pending edits, since otherwise it's easy to miss that they shouldn't be merged or whatnot separately.


mwiencek (approved):

yvanzo (approved):
Works nicely too!

PR #1775, opened 2020-11-04
MBS-11210: Allow more subpaths for saisaibatake.ame-zaiku.com

reosarevok:
### Implement MBS-11210

Two more subpaths (musical and musical_instrument) are used for musical instruments for some reason.

Removing the /gakki subpath rather than try and specify all four in constants.js and URL.pm since we're unlikely to want to link to other bits of the website that should be treated differently anyway.


lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -68,7 +68,7 @@ my %URL_SPECIALIZATIONS = (
     'Flattr'              => qr{^https?://(?:www\.)?flattr\.com/profile/}i,
     'FolkWiki'            => qr{^https?://(?:www\.)?folkwiki\.se/}i,
     'FonoFi'              => qr{^https?://(?:www\.)?fono\.fi/}i,
-    'Gakki'               => qr{^https?://(?:www\.)?saisaibatake\.ame-zaiku\.com/gakki/}i,
+    'Gakki'               => qr{^https?://(?:www\.)?saisaibatake\.ame-zaiku\.com/(gakki|musical|musical_instrument)/}i,
```
yvanzo, line 71: Capturing group is not needed here:

```suggestion
    'Gakki'               => qr{^https?://(?:www\.)?saisaibatake\.ame-zaiku\.com/(?:gakki|musical|musical_instrument)/}i,
```

yvanzo (approved):

mwiencek (approved):

PR #1778, opened 2020-11-06
Store name in EditLinkType attributes

mwiencek:
Companion to https://tickets.metabrainz.org/browse/MBBE-24

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/EditLinkType.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/EditLinkType.pm
@@ -34,10 +34,10 @@ sub change_fields
         description         => Nullable[Str],
         priority            => Optional[Int],
         attributes          => Optional[ArrayRef[Dict[
-            name => Optional[Str], # Used in old historic edits
+            name => Str,
             min  => Nullable[Int],
             max  => Nullable[Int],
-            type => Optional[Int], # Used in the new edits
+            type => Int,
         ]]],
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Edit/Relationship/EditLinkType.pm
```diff
@@ -34,7 +34,7 @@ sub change_fields
         description         => Nullable[Str],
         priority            => Optional[Int],
         attributes          => Optional[ArrayRef[Dict[
-            name => Optional[Str], # Used in old historic edits
+            name => Str,
             min  => Nullable[Int],
             max  => Nullable[Int],
             type => Optional[Int], # Used in the new edits
```
reosarevok, line 40: Should we also add type ID to the old edits for consistency?

reosarevok (approved):

lib/MusicBrainz/Server/Edit/Relationship/EditLinkType.pm
mwiencek: Actually, I'm not seeing a single edit without a type. Maybe they were added in at some point? I can remove the `Optional` + comment though.

reosarevok: If that's the case, then it makes sense I guess :)

yvanzo (approved):

PR #1780, opened 2020-11-08
MBS-7752: Replace referer redirects with ?returnto=

mwiencek:
The referer header is optional and may not be sent by the user's browser (particularly if they've chosen to block it). Specify a `returnto` parameter everywhere we might need one.

`redirect_back` has been improved ~to make the infinite redirect checking (currently handled by `$ignore`) more generalized, and~ to block any attempt at redirecting to an external URL.


root/user/UserProfile.js
```diff
@@ -24,6 +24,7 @@ import commaOnlyList from '../static/scripts/common/i18n/commaOnlyList';
 import {formatCount, formatPercentage} from '../statistics/utilities';
 import formatUserDate from '../utility/formatUserDate';
 import {canNominate} from '../utility/voting';
+import { returnToCurrentPage } from '../utility/returnUri';
```
reosarevok, line 27: Why the padding? :)

mwiencek: I use a vscode-js-import plugin to help import stuff and it adds those.

mwiencek:
Finally fixed the tests here, should be ready for review

reosarevok (commented):
The code looks fine - any specific tests suggested? I tried it for merges, where it seemed to work well. I guess we can't test the beta unsetting locally in any reasonable way. What other places do you feel we should try? :)

mwiencek:
I should be able to test the beta stuff on one of my development servers (I have an prod/beta setup to switch between). I'll check if MBS-6438 is resolved too.

Otherwise, login/logout redirects, plus adding stuff to collections/subscripts would be useful to check!

reosarevok (approved):
Tested those few things and they seem to work. Happy with this if you tested the beta stuff locally.

yvanzo (approved):
:shipit:

PR #1786, opened 2020-11-11
MBS-11216: Split InstrumentsWithoutAnImage report

reosarevok:
### Implement MBS-11216

Wikidata is not connected to images at all anymore, but it's still a very useful link to have. As such, it makes sense to split this report in two. Also changed the query a bit to hopefully improve it.

We could just remove InstrumentsWithoutAnImage, but hopefully we can soon use it again once we start being able to link to IROM images on StaticBrainz.

PR #1787, opened 2020-11-11
Report display refactoring + MBS-11513

reosarevok:
### Implement MBS-11217 and MBS-11513 (plus a ton of other refactoring)

Mostly refactoring to make the report pages more consistent and reduce the possibility for errors.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1786

### DIFF
--- a/root/report/CDTocDubiousLength.js
+++ b/root/report/CDTocDubiousLength.js
@@ -9,41 +9,35 @@
 
 import * as React from 'react';
 
-import Layout from '../layout';
-import formatUserDate from '../utility/formatUserDate';
-
 import CDTocList from './components/CDTocList';
+import ReportLayout from './components/ReportLayout';
 import type {ReportDataT, ReportCDTocT} from './types';
 
 const CDTocDubiousLength = ({
   $c,
+  canBeFiltered,
   generated,
+  filtered,
   items,
   pager,
-}: ReportDataT<ReportCDTocT>): React.Element<typeof Layout> => (
-  <Layout $c={$c} fullWidth title={l('Disc IDs with dubious duration')}>
-    <h1>{l('Disc IDs with dubious duration')}</h1>
-
-    <ul>
-      <li>
-        {l(`This report shows disc IDs indicating a total duration much longer
-        than what a standard CD allows (at least 88 minutes for a CD, or 30
-        minutes for a mini-CD). This usually means a disc ID was created for
-        the wrong format (SACD) or with a buggy tool.`)}
-      </li>
-      <li>
-        {texp.l('Total releases found: {count}',
-                {count: pager.total_entries})}
-      </li>
-      <li>
-        {texp.l('Generated on {date}',
-                {date: formatUserDate($c, generated)})}
-      </li>
-    </ul>
-
+}: ReportDataT<ReportCDTocT>): React.Element<typeof ReportLayout> => (
+  <ReportLayout
+    $c={$c}
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows disc IDs indicating a total duration much longer
+       than what a standard CD allows (at least 88 minutes for a CD, or 30
+       minutes for a mini-CD). This usually means a disc ID was created for
+       the wrong format (SACD) or with a buggy tool.`,
+    )}
+    entityType="discId"
+    filtered={filtered}
+    generated={generated}
+    title={l('Disc IDs with dubious duration')}
+    totalEntries={pager.total_entries}
+  >
     <CDTocList items={items} pager={pager} />
-
-  </Layout>
+  </ReportLayout>
 );
 
 export default CDTocDubiousLength;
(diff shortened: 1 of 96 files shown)


root/report/RecordingsSameNameDifferentArtistsSameName.js
```diff
@@ -28,44 +26,30 @@ const RecordingsSameNameDifferentArtistsSameName = ({
...
-    <ul>
-      <li>
-        {l(`This report shows all recordings with the same name that have
-            different artists (having different MBIDs) with the same name.`)}
-      </li>
-      <li>
-        {exp.l(`These are most likely cases where the {ac|artist credit} is
-                incorrect for at least one of the recordings.`,
-               {ac: '/doc/Artist_Credits'})}
-      </li>
```
reosarevok, line 52: I joined these two into one string here, since this was the only case where we had *three* paragraphs (and it seemed silly to support three sections just for this). I see no reason to have these two separate anyway - most reports have "this is what this report is about" in one string.

root/report/DuplicateArtists.js
```diff
@@ -27,42 +25,34 @@ const DuplicateArtists = ({
...
-                  {date: formatUserDate($c, generated)})}
-        </li>
-
-        {canBeFiltered ? <FilterLink $c={$c} filtered={filtered} /> : null}
-      </ul>
-
+    <ReportLayout
+      $c={$c}
+      canBeFiltered={canBeFiltered}
+      countText={l('Total duplicate groups: {count}')}
```
reosarevok, line 36: This is the only case where a separate countText is needed, but since I guess it does count groups and not artists, it's probably still needed to do it like this.

mwiencek (approved):
Didn't test or review every single template but I think the refactorings make sense.

root/report/CDTocNotApplied.js
```diff
@@ -23,34 +20,26 @@ const CDTocNotApplied = ({
...
+       attached, so it is highly likely that the disc ID can be applied
+       without any worries. Do make sure though that no existing durations
+       clash with the disc ID, or that any clashes are clear mistakes.`,
+    )}
+    entityType="release"
+    filtered={filtered}
+    generated={generated}
+    title={l('Disc IDs attached but not applied')}
+    totalEntries={pager.total_entries}
+  >
```
yvanzo: `'Total discIDs found: {count}'` is being replaced with `'Total releases found: {count}'`.

It is a bit unclear whether or not releases with several media can be reported several times.

yvanzo (commented):
Tested locally with sample data and proofread it. Still it seems easy to miss a mistaken copy/paste. So it might be worth a dedicated ticket to warn users about it.

I linked MBS-10843 and MBS-11217 to this PR. Theses will have to be manually transitioned when merging the PR.

MBS-11303 is not linked yet but it seems more general than what is actually addressed in commit https://github.com/metabrainz/musicbrainz-server/pull/1787/commits/ce78a9cef672917a269ef265290063b107334655.

reosarevok:
> MBS-11303 is not linked yet but it seems more general than what is actually addressed in commit ce78a9c.

Yes, there's already a second PR for that open, and a third to come eventually.

yvanzo:
> > MBS-11303 is not linked yet but it seems more general than what is actually addressed in commit [ce78a9c](https://github.com/metabrainz/musicbrainz-server/commit/ce78a9cef672917a269ef265290063b107334655).
> 
> Yes, there's already a second PR for that open, and a third to come eventually.

Since this refactoring seems to be large enough to be addressed by chunks in separate pull requests, it seems worth creating subticket for each chunk, unless you need all of them to be merged and released at the same time. Please link these commits/PRs to these tickets.

root/report/CDTocNotApplied.js
reosarevok: This is fair, and in fact the other discID report (which currently says "releases found") also should be changed to say discIDs. Fixed.

root/report/CDTocDubiousLength.js
```diff
@@ -9,41 +9,33 @@
...
+}: ReportDataT<ReportCDTocT>): React.Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows disc IDs indicating a total duration much longer
+       than what a standard CD allows (at least 88 minutes for a CD, or 30
+       minutes for a mini-CD). This usually means a disc ID was created for
+       the wrong format (SACD) or with a buggy tool.`,
+    )}
+    entityType="discId"
```
reosarevok, line 31: This actually used to show "releases found", but it doesn't show releases, it shows discIDs, so this fixes that.

yvanzo (approved):
Double-checked changes about discIDs. Thanks for the added ticket! :shipit:

reosarevok:
Changed to use $c directly again in three reports where I had added useContext on the top-level, as per @mwiencek's comment here: https://github.com/metabrainz/musicbrainz-server/pull/1850#discussion_r600602106

Will merge once tests pass.

PR #1788, opened 2020-11-12
MBS-11220: Allow external release editor seeding in user selected language

jesus2099:
🙇 Sorry for this very dirty PR where I cannot make sure that it works by muself. 🙇 

Problem from [MBS-11220](https://tickets.metabrainz.org/browse/MBS-11220)
-------

My browser setting: 

    Accept-Language: fr-FR,fr;q=0.9,en-GB;q=0.8,en;q=0.7,vi;q=0.6,ja;q=0.5

My MBS selected language:

    Cookie: {…}; lang=en

Using any external release editor seeding\*, release editor page is displayed in French instead of English.

Cause
-----

External (non-MBS) release editor seeding can no longer know
what is the user selected language because it cannot
access the lang Cookie in the request.

Solution
---

Trying to create the lang cookie as SameSite=None; Secure.
As I found that same issue was fixed for musicbrainz_server_session
by 54337c254c1cfe45d5c544be8c17be535af0b496 by @mwiencek.

Action
------

Please @mwiencek or @reosarevok (someone who knows how to run a server) could you run this version of MBS?

1. Set your preferred MBS language (language selection menu)
2. Setup your browser to prefer language to one of the available MBS languages but different from the one above
3. Try any external way of seeding\* the release editor to confirm that the release editor is using browser lang instead of MBS lang cookie
4. Run this version of MBS
5. See if the fix works

Because sorry I have really no idea if my fix would work at all.
I don't know Python, I don't know how to run my server version. 🙇 

\* External release editor seeding
---------------------------

Here is the **easiest way**:

1. Go to [MBwiki:Development/Release_Editor_Seeding](https://wiki.musicbrainz.org/Development/Release_Editor_Seeding) and save [seed-love-bug.html](https://musicbrainz.org/static/tests/seed-love-bug.html) locally
2. edit `action="/release/add"` into `action="https://musicbrainz.org/release/add"`
3. Run it locally and press the **seed** button

Or it's also the case with [Picard Add Cluster As Release plugin](https://github.com/metabrainz/picard-plugins/issues/270) or [Discogs Importer](https://github.com/murdos/musicbrainz-userscripts/issues/325), the two things that I have spotted that made me try to fix this.

### DIFF
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -261,7 +261,15 @@ sub get_collator
 
 sub set_language_cookie {
     my ($c, $lang) = @_;
-    $c->res->cookies->{lang} = { 'value' => $lang, 'path' => '/', 'expires' => time()+31536000, 'SameSite' => 'None', 'Secure' => '1' };
+    $c->res->cookies->{lang} = {
+        'value' => $lang,
+        'path' => '/',
+        'expires' => time()+31536000,
+        $c->req->secure ? (
+            'samesite' => 'None',
+            'secure' => '1',
+        ) : (),
+    };
 }
 
 sub handle_unicode_encoding_exception {
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -261,7 +261,7 @@ sub get_collator
 
 sub set_language_cookie {
     my ($c, $lang) = @_;
-    $c->res->cookies->{lang} = { 'value' => $lang, 'path' => '/', 'expires' => time()+31536000 };
+    $c->res->cookies->{lang} = { 'value' => $lang, 'path' => '/', 'expires' => time()+31536000, 'SameSite' => 'None', 'Secure' => '1' };
 }
 
 sub handle_unicode_encoding_exception {


jesus2099:
Judging from https://github.com/metabrainz/musicbrainz-server/blob/416ff1e5526e95347b2a5ae2f9118e416af9b282/lib/MusicBrainz/Server.pm#L488 maybe I should have written `'Secure' => 1` instead of `'Secure' => '1'`.

jesus2099:
I created a Jira ticket MBS-11220 as PR should be tracked from there.

lib/MusicBrainz/Server.pm
```diff
@@ -261,7 +261,7 @@ sub get_collator
 
 sub set_language_cookie {
     my ($c, $lang) = @_;
-    $c->res->cookies->{lang} = { 'value' => $lang, 'path' => '/', 'expires' => time()+31536000 };
+    $c->res->cookies->{lang} = { 'value' => $lang, 'path' => '/', 'expires' => time()+31536000, 'SameSite' => 'None', 'Secure' => '1' };
```
mwiencek, line 264: This didn't work for me, but this does:
```suggestion
    $c->res->cookies->{lang} = {
        'value' => $lang,
        'path' => '/',
        'expires' => time()+31536000,
        $c->req->secure ? (
            'samesite' => 'None',
            'secure' => '1',
        ) : (),
    };
```
 * samesite/secure have to be lowercase here (dunno why)
 * they should only be output if `$c->req->secure`, otherwise the browser will reject the cookie on dev servers where SSL is unlikely to be set up
 * added some indentation for clarity

mwiencek (changes requested):
Thanks! The solution makes sense, there is just a problem with the implementation. :)

mwiencek:
BTW, we plan on switching to SameSite=Strict for session cookies in https://github.com/metabrainz/musicbrainz-server/pull/1776. This means the user will be logged out whenever a cross-origin form submission occurs. However, there will be a confirmation page where they can continue with the seed in one click without logging in (assuming they have an active session). It's available for testing on https://test.musicbrainz.org/ if you'd like to check userscripts against it.

This doesn't affect this PR, since the lang cookie will have to remain as SameSite=None in order to keep the seed landing page in the user's language.

jesus2099 (commented):
Thanks @mwiencek, I think I included your syntax fix in my PR now.
Are you sure I shouldn't set `secure` to `1` instead of `'1'`?
Sorry, my PR is mostly time consuming for you.

mwiencek:
> Are you sure I shouldn't set secure to 1 instead of '1'?

Yes, either work fine -- Perl doesn't have explicit string and numeric types. `1` and `'1'` are interchangeable in both contexts.

> Sorry, my PR is mostly time consuming for you.

Not at all. :) I appreciate you submitting it.

mwiencek (approved):

yvanzo (approved):
Thank you @jesus2099!

jesus2099:
Wow, it works in beta! :)
I tested with `seed-love-bug.html` and Picard’s **Add Cluster As Release…** plugin.
Together with #1757 nicely done with its request parameter display and everything.

PR #1793, opened 2020-11-16
MBS-11227: Ensure /release-group/merge also works

reosarevok:
### Fix MBS-11227

The URL for RGs is usually release-group (see entities.json) but following the url in ENTITIES fails for merge because of it being defined in a generic role. This adds forwarding so that /release-group/merge works and just forwards to /release_group/merge.


mwiencek (approved):

yvanzo (approved):
LGTM but did not test.

PR #1794, opened 2020-11-16
MBS-11225: Autoselect IROMBOOK images

reosarevok:
### Implement MBS-11225

This autoselects IROMBOOK images on StaticBrainz as image links for instruments (and only for instruments).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1754,6 +1754,15 @@ const CLEANUPS = {
       return {result: false};
     },
   },
+  'irombook': {
+    match: [new RegExp('^https://staticbrainz\\.org/irombook/')],
+    type: LINK_TYPES.image,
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1754,6 +1754,15 @@ const CLEANUPS = {
       return {result: false};
     },
   },
+  'irombook': {
+    match: [new RegExp('^(https?://)?staticbrainz\\.org/irombook/', 'i')],
```
yvanzo, line 1758: If there is no need for cleanup, there is probably no need for flexible matching either:

```suggestion
    match: [new RegExp('^https://staticbrainz\\.org/irombook/')],
```

mwiencek (approved):
LGTM though the changes yvanzo suggested probably make sense

yvanzo (approved):

PR #1795, opened 2020-11-16
MBS-11221 / MBS-11234: Normalize whosampled.com links to HTTPS and add validation

reosarevok:
### Implement MBS-11221 / MBS-11234

They redirect to HTTPS (with www.) so we should do the same. MBBE-25 will convert existing links.

Also added validation for different sorts of links I could find.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3215,6 +3215,62 @@ const CLEANUPS = {
     clean: function (url) {
       return url.replace(/^(?:https?:\/\/)?(?:www\.)?whosampled\.com\/(.+)$/, 'https://www.whosampled.com/$1');
     },
+    validate: function (url, id) {
+      if (/[?#]/.test(url)) {
+        return {
+          error: l(
+            `There is an unencoded “?” or “#” character in this URL.
+             Please check whether it is useless and should be removed,
+             or whether it is an error and the URL is misencoded.`,
+          ),
+          result: false,
+        };
+      }
+      if (/^https:\/\/www\.whosampled\.com\/sample\//.test(url)) {
+        return {
+          error: l(
+            `Please do not link directly to WhoSampled sample pages.
+             Link to the appropriate artist, recording
+             or release page instead.`,
+          ),
+          result: false,
+        };
+      }
+      if (/^https:\/\/www\.whosampled\.com\/album\//.test(url)) {
+        if (id === LINK_TYPES.otherdatabases.release_group) {
+          return {result: true};
+        }
+        return {
+          error: l(
+            'Please link WhoSampled album pages to release groups.',
+          ),
+          result: false,
+        };
+      }
+      if (/^https:\/\/www\.whosampled\.com\/(?!(?:album|sample))[^/]+(?:\/)?$/.test(url)) {
+        if (id === LINK_TYPES.otherdatabases.artist) {
+          return {result: true};
+        }
+        return {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3210,6 +3209,13 @@ const CLEANUPS = {
       return url.replace(/^(?:https?:\/\/)?(?:[^/]+\.)?weibo\.com\/(u\/)?([^\/?#]+)(?:.*)$/, 'https://www.weibo.com/$1$2');
     },
   },
+  'whosampled': {
+    match: [new RegExp('^(https?://)?(www\\.)?whosampled\\.com', 'i')],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      return url.replace(/^(?:https?:\/\/)?(?:www\.)?whosampled\.com\/(.*)+$/, 'https://www.whosampled.com/$1');
```
yvanzo, line 3216: `*` is unneeded when using `+`.

```suggestion
      return url.replace(/^(?:https?:\/\/)?(?:www\.)?whosampled\.com\/(.+)$/, 'https://www.whosampled.com/$1');
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3210,6 +3209,13 @@ const CLEANUPS = {
...
+    },
+  },
```
yvanzo, line 3218: It would be nice to add validation to prevent `?` and `#` signs to be added instead of their percent encoding form.
For example, see https://musicbrainz.org/url/9b6219cf-9893-4485-a8dd-3d2f9579e970

yvanzo (changes requested):

root/static/scripts/edit/URLCleanup.js
reosarevok: How would you envision this validation working? Just reject any URLs containing those? Or clean them up to their encoded variants? Or something else?

yvanzo: _Just reject any URLs containing those_ would be nice at least. Cleaning them up can be difficult because it would require to distinguish between misencoded `?` in the slug from `?` which is used for the query part. Let’s not over-engineer it.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3215,6 +3215,62 @@ const CLEANUPS = {
     clean: function (url) {
       return url.replace(/^(?:https?:\/\/)?(?:www\.)?whosampled\.com\/(.+)$/, 'https://www.whosampled.com/$1');
     },
+    validate: function (url, id) {
+      if (/[?#]/.test(url)) {
+        return {
+          error: l(
+            `There's an unencoded “?” or “#” character in this URL.
```
yvanzo, line 3222:
```suggestion
            `There is an unencoded “?” or “#” character in this URL.
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3215,6 +3215,62 @@ const CLEANUPS = {
...
+             Please check whether it is useless and should be removed,
+             or whether it is an error and the URL is broken.`,
```
yvanzo, line 3224:
```suggestion
             or whether it is an error and the URL is misencoded.`,
```

yvanzo (approved):
Just two last small things. The rest is nice, thanks!

mwiencek (approved):

PR #1802, opened 2020-11-20
Use SameSite=Lax for login cookies (fixes MBS-11238)

mwiencek:
https://tickets.metabrainz.org/browse/MBS-11238

5ab8cfba261175f4712d89a230b2c8b0c2bc716a made it so that session/login cookies are only sent in first-party contexts from the same origin. This means the user will appear as logged out if they click on any links to MB from an external origin. In order to address this, we either have to downgrade to SameSite=Lax and ensure we don't allow any data modifications via GET requests, or add a separate non-strict cookie that points to a user session for display but doesn't authenticate as that user. I'm not sure how to do the latter with Catalyst without breaking a lot of assumptions we make throughout the codebase, so I'm "downgrading" the cookies to Lax mode and tweaking a couple cases where it might be possible to submit changes via a GET request. This is still an improvement over the pre-5ab8cfb behavior of SameSite=None and fulfills the spirit of that commit.

### DIFF
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -495,7 +495,7 @@ around make_session_cookie => sub {
     my ($orig, $self, $sid, %attrs) = @_;
 
     my $cookie = $self->$orig($sid, %attrs);
-    $cookie->{samesite} = 'Strict';
+    $cookie->{samesite} = 'Lax';
     $cookie->{secure} = 1 if $self->req->secure;
     return $cookie;
 };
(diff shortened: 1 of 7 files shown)


mwiencek:
@brainzbot retest this please

mwiencek:
(This does not address the issue of appearing logged out on the seed confirmation page, because that's only shown for POST requests.)

reosarevok (approved):
LGTM but didn't test

yvanzo (approved):

PR #1803, opened 2020-11-20
MBS-11241: Seed annotation, edit_note with textarea

mwiencek:
This should preserve newlines in the input.

The CSS changes are required in order to make the inputs and textareas have identical widths. Setting size=50 and cols=50 wasn't sufficient because the font families and font sizes were different for some reason.


reosarevok (approved):
Threw it onto test for checking, seems to work fine.

PR #1805, opened 2020-11-23
MBS-11247: Stop highlighting BBC Music links

reosarevok:
### Implement MBS-11247

The BBC artist pages were discontinued and taken down in Nov 19, 2020.
I set the relevant relationship as deprecated too.
The companion task MBBE-28 will eventually set these as ended.


reosarevok:
For a case like this one where the change removes the last reason to have an autoselect and cleanup (with the rel itself being deprecated too), should we remove those bits of code as well?

mwiencek (approved):

lib/MusicBrainz/Server/Entity/URL/BBCMusic.pm
```diff
@@ -1,26 +0,0 @@
-package MusicBrainz::Server::Entity::URL::BBCMusic;
-
-use Moose;
-
-extends 'MusicBrainz::Server::Entity::URL';
-with 'MusicBrainz::Server::Entity::URL::Sidebar';
-
-override href_url => sub {
-    shift->url->as_string =~ s{^http:}{https:}r;
-};
```
yvanzo, line 10: For the record, all BBC Music URLs in MusicBrainz Database are `https` now. This override was obsolete already.

yvanzo (approved):

lib/MusicBrainz/Server/Entity/URL/BBCMusic.pm
reosarevok: We should probably check whether an override exists (and get rid of it) when we bot- or cron- change URLs to https :)

PR #1808, opened 2020-11-24
MBS-11250 / MBS-11253: Properly support JSON user-rating / user-tags lookups

reosarevok:
### Fix MBS-11250 / MBS-11253

Asking for this as JSON didn't error, but serialized the entity instead. This changes it so that you can actually query for the rating/tag as in XML.

Note this returns the rating in a 0-100 scale, while normally user-rating is returned on the 0-5 scale. This is inconsistent with the rest of MB but consistent with the XML version of this endpoint, so probably for the best.

Added tests and all since I was feeling responsible.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/Rating.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/Rating.pm
@@ -13,6 +13,7 @@ my $ws_defs = Data::OptList::mkopt([
      rating => {
                          method   => 'GET',
                          required => [ qw(id entity) ],
+                         optional => [ qw(fmt) ],
      },
      rating => {
                          method   => 'POST',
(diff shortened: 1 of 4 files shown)


mwiencek:
> Note this returns the rating in a 0-100 scale, while normally user-rating is returned on the 0-5 scale. This is inconsistent with the rest of MB but consistent with the XML version of this endpoint, so probably for the best.

Ugh. :P I opened https://tickets.metabrainz.org/browse/MBS-11266 for that.

mwiencek (approved):

yvanzo (approved):
Is there a documentation for this?

reosarevok:
It's "documented" inasmuch as the docs say:

> We also provide an API interface for the following non-core resources:
> 
>  rating, tag, collection

But there's no info of how to do so either here or under /Examples. I'll add them to /Examples at least.

reosarevok:
@yvanzo: see https://wiki.musicbrainz.org/index.php?title=MusicBrainz_API%2FExamples&type=revision&diff=75036&oldid=74634

yvanzo:
Thanks, I added it to the transclusion table: https://musicbrainz.org/doc/MusicBrainz_API/Examples

PR #1810, opened 2020-11-28
MBS-9840: Add Overture by Doremus to the otherdbs whitelist

reosarevok:
### Implement MBS-9840

They seem to have 4 main entity types: artist (our artist), performance (the ticket suggested recording but it seems to be event), expression (our work, FRBR naming), and score (which would rather fit BB than MB).


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2508,6 +2508,28 @@ const CLEANUPS = {
     ],
     type: LINK_TYPES.otherdatabases,
   },
+  'overture': {
+    match: [new RegExp('^(https?://)?overture\\.doremus\\.org/', 'i')],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      return url.replace(/^(?:https?:\/\/)?overture\.doremus\.org\/$/, 'https://overture.doremus.org/');
```
zas, line 2515: In the `match` expression, case-insensitive match, but here and below (`regexp.exec()`) case-sensitive, imho everything should be case-insensitive (at least for domain part).

zas (changes requested):

root/static/scripts/edit/URLCleanup.js
mwiencek: We don't seem to do that elsewhere, but might be a good idea. @yvanzo ?

mwiencek (approved):

root/static/scripts/edit/URLCleanup.js
yvanzo: :+1:

yvanzo (approved):

root/static/scripts/edit/URLCleanup.js
yvanzo: Case-sensivity of the function `clean` function is not specific to the rule `doremus`, there are 138 `clean` functions. Since this is a much larger issue, it seems better to be addressed globally, the ticket MBS-11331 has been created to track it.

PR #1811, opened 2020-11-30
MBS-11262: Fix l_attributes translation calls

reosarevok:
### Fix MBS-11262

Unless I'm very wrong, l_attributes should basically never be called on its own, but as lp_attributes with the specific context. Some of these aren't even in the attributes context at all.


mwiencek (approved):

yvanzo (approved):
Good catch!

PR #1813, opened 2020-11-30
Set SameSite=None on beta cookie

mwiencek:
If someone has the beta preference on and uses an external importer, the POST request won't include the beta cookie due to it using SameSite=Lax. This switches the beta cookie to use SameSite=None so that the 307 redirect is performed before the ConfirmSeed page is shown twice (first on the production site, then again on the beta site).


mwiencek:
This isn't exactly a regression, though it resolves MBS-11092 much more nicely by performing the beta redirect on the initial POST so that the user doesn't have to confirm it twice (which is confusing and awkward).

reosarevok (approved):
LGTM but didn't test

PR #1814, opened 2020-11-30
MBS-9674: Support internationalized domains in URL forms

reosarevok:
### Fix MBS-9674

This was failing because, unsurprisingly, it couldn't find the ASCII hostname in the Unicode URL. I used punycode (which we already installed as a dependency, but I added as a requirement) to decode the punycode back into the original unicode URL for hostname comparison. 

During the last Monday meeting we also agreed to always display international URLs as Unicode (which we already did when prettifying but not in the links editor). This makes the relevant changes to externalLinks and to URL forms.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -7,6 +7,8 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import punycode from 'punycode';
+
 import $ from 'jquery';
 import ko from 'knockout';
 import * as React from 'react';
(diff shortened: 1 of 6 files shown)


mwiencek (approved):

root/static/scripts/edit/externalLinks.js
```diff
@@ -527,13 +531,27 @@ export function parseRelationships(
 const protocolRegex = /^(https?|ftp):$/;
 const hostnameRegex = /^(([A-z\d]|[A-z\d][A-z\d\-]*[A-z\d])\.)*([A-z\d]|[A-z\d][A-z\d\-]*[A-z\d])$/;
 
+export function getUnicodeUrl(url: string): string {
+  if (!isValidURL(url)) {
+    return url;
+  }
+
+  const urlObject = new URL(url);
```
yvanzo, line 539: This might not be supported using IE11 but (from [chat logs](https://chatlogs.metabrainz.org/brainzbot/metabrainz/msg/4729570/)):

> it seems to be provided by core-js
> https://github.com/zloirock/core-js#url-and-urlsearchparams

yvanzo (approved):
Works for me but did not test with IE 11.

PR #1815, opened 2020-12-02
MBS-11256: Make adding tracklist to empty medium an autoedit

reosarevok:
### Implement MBS-11256

We generally make adding something where nothing existed an autoedit. I see no reason why this shouldn't also apply to adding a tracklist if it was previously unknown.


mwiencek (approved):

yvanzo (approved):
Works fine!

PR #1817, opened 2020-12-02
MBS-11264: Guess case apostrophe fixes

reosarevok:
### Fix MBS-11264

### DIFF
--- a/root/static/scripts/guess-case/modes.js
+++ b/root/static/scripts/guess-case/modes.js
@@ -152,6 +152,8 @@ const DefaultMode = {
     );
   },
 
+  name: '',
+
   toLowerCase(str) {
     return str.toLowerCase();
   },
@@ -312,6 +314,8 @@ export const English = Object.assign({}, DefaultMode, {
     return false;
   },
 
+  name: 'English',
+
   runPostProcess(is) {
     is = DefaultMode.runPostProcess(is);
     is = this.fixEnglishKeyNames(is);
@@ -329,6 +333,8 @@ export const French = Object.assign({}, DefaultMode, {
       url: {href: 'https://musicbrainz.org/doc/Style/Language/French', target: '_blank'},
     },
   )),
+
+  name: 'French',
 
   runPostProcess(is) {
     return DefaultMode.runPostProcess(is)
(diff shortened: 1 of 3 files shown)


root/static/scripts/guess-case/modes.js
```diff
@@ -137,19 +137,24 @@ function runFixes(is, fixes) {
 const DefaultMode = {
   description: '',
 
+  fixEnglishApostrophes() {
```
mwiencek, line 140: Since this returns a boolean and doesn't perform an action, I think something like `shouldFixEnglishApostrophes` would be more clear, especially since we have other methods starting with "fix" that actually perform actions (`fixVinylSizes`, `fixEnglishKeyNames`, etc.).
```suggestion
  shouldFixEnglishApostrophes() {
```

Though this method also seems a bit weird since it's basically "is this mode English" and not really to do with apostrophes. I'd think we should just have a `gc.mode.name === 'english'` check or something where it matters.

root/static/scripts/guess-case/utils.js
```diff
@@ -206,19 +206,22 @@ export function titleString(gc, is, forceCaps) {
      * we got an 'round (Around = 'round), lowercase
      * we got a 'mon (Come on = C'mon), lowercase
      */
-  } else if (isApostrophe(gc.i.getPreviousWord()) &&
+  } else if (gc.mode.fixEnglishApostrophes() &&
+      isApostrophe(gc.i.getPreviousWord()) &&
       lc.match(/^(?:s|round|em|ve|ll|d|cha|re|til|way|all|mon)$/i)) {
```
mwiencek, line 211: This style of indentation makes it slightly harder for me to tell which lines are part of the condition and which are part of the `if` block. I like to follow our general brace style for multi-line conditions, so I'm recommending that unless you strongly prefer the current way. :)
```suggestion
  } else if (
    gc.mode.fixEnglishApostrophes() &&
    isApostrophe(gc.i.getPreviousWord()) &&
    lc.match(/^(?:s|round|em|ve|ll|d|cha|re|til|way|all|mon)$/i)
  ) {
```

mwiencek (approved):
I have some stylistic suggestions, but the PR looks good otherwise!

yvanzo (approved):

PR #1818, opened 2020-12-03
MBS-4548: Allow seeing all uses of an artist credit

reosarevok:
### Implement MBS-4548

Currently to see where an artist credit is used you need to visit three different tabs of an artist page (overview, releases and recordings) and filter each one by AC. And you still aren't able to see ACs that are used on tracks only.

This mostly apes the tags interface to show a list of artist credit uses. The main use for this is to either check it before entering an edit artist credit edit (from the aliases page) or while voting on one, so I added links from those two cases. For the edit AC edit the link is only shown while the edit is open for voting, since after that it's of no use.


root/artist_credit/ArtistCreditIndex.js
```diff
@@ -0,0 +1,136 @@
...
+    </ul>
+    <h2>{l('Uses')}</h2>
+    {/*
+      * The below use N_ln so languages with non-Germanic pluralization
+      * rules (i.e., any that make number distinctions above the
+      * threshold where we'll actually show the string) can translate
+      * properly. However, the strings are the same in English because
+      * we do not make a distinction other than for 1, which will never
+      * show in this case.
+    */}
```
mwiencek, line 116: We might want to consider moving these strings to constants.js later so they can be reused without duplicating the comment and everything. I guess it's fine for now since it's only copied to here and the tags place.

lib/MusicBrainz/Server/Data/Recording.pm
```diff
@@ -112,6 +112,18 @@ sub find_by_artist
     $self->query_to_list_limited($query, \@where_args, $limit, $offset);
 }
 
+sub find_by_artist_credit
+{
+    my ($self, $artist_credit_id, $limit, $offset) = @_;
+
+    my $query = "SELECT DISTINCT " . $self->_columns . ",
```
mwiencek, line 119: I don't think `DISTINCT` is needed? Since you don't have any joins and are selecting the primary key, each row will be unique.
```suggestion
    my $query = "SELECT " . $self->_columns . ",
```

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -245,6 +245,22 @@ sub find_by_artist
...
+
+    my $query = "SELECT DISTINCT " . $self->_columns . ",
+                   date_year, date_month, date_day, area.name AS country_name,
+                   release.name COLLATE musicbrainz AS name_collate
+                 FROM " . $self->_table . "
+                 LEFT JOIN release_event ON release_event.release = release.id
+                 LEFT JOIN area ON area.id = release_event.country
+                 WHERE artist_credit = ?
+                 ORDER BY date_year, date_month, date_day,
+                   country_name, barcode, release.name COLLATE musicbrainz";
```
mwiencek, line 260: Since we don't actually display the year, country, etc., I would probably remove all those joins and sorts (just sort by name like the rest). Then you can remove the `DISTINCT` too.

lib/MusicBrainz/Server/Data/ReleaseGroup.pm
```diff
@@ -252,6 +252,33 @@ sub find_by_track_artist
...
+                      ORDER BY name ASC
+                    ) secondary_types
+                 FROM " . $self->_table . "
+                 WHERE rg.artist_credit = ?
+                 ORDER BY
+                   rg.type, secondary_types,
+                   rgm.first_release_date_year,
+                   rgm.first_release_date_month,
+                   rgm.first_release_date_day,
+                   rg.name COLLATE musicbrainz";
```
mwiencek, line 278: Same as above...I don't think it makes sense to sort by secondary information you can't see. The queries should be faster that way too.

lib/MusicBrainz/Server/Data/Track.pm
```diff
@@ -87,6 +87,18 @@ sub load_for_mediums
     }
 }
 
+sub find_by_artist_credit
+{
+    my ($self, $artist_credit_id, $limit, $offset) = @_;
+
+    my $query = "SELECT DISTINCT " . $self->_columns . ",
```
mwiencek, line 94:
```suggestion
    my $query = "SELECT " . $self->_columns . ",
```

lib/MusicBrainz/Server/Controller/ArtistCredit.pm
```diff
@@ -0,0 +1,102 @@
...
+
+        $c->stash(
+            current_view => 'Node',
+            component_path => 'artist_credit/EntityList',
+            component_props => {
+                %{$c->stash->{component_props}},
+                entities => [map +{
+                    entity => $_,
+                    entity_id => $_->{id},
+                }, @$entities],
```
mwiencek, line 78: Why not just `entities => $entities`? The `id` is already included on the entities themselves.

mwiencek (changes requested):
The actual page works great, just some small suggestions

root/artist_credit/EntityList.js
```diff
@@ -0,0 +1,79 @@
...
+  entities,
+  entityType,
+  page,
+  pager,
+}: Props): React.Element<typeof ArtistCreditLayout> => (
+  <ArtistCreditLayout $c={$c} artistCredit={artistCredit} page={page}>
+    <h2>
+      {expand2text(
+        headingsText[entityType](pager.total_entries),
+        {num: pager.total_entries},
```
mwiencek, line 57: it'd be nice to import formatCount from root/statistics/utilities.js for these
```suggestion
        {num: formatCount($c, pager.total_entries)},
```

root/artist_credit/ArtistCreditIndex.js
```diff
@@ -0,0 +1,133 @@
...
+            ) : <EntityLink entity={entity} />}
+          </li>
+        ))}
+        {entities.count > entities.entities.length ? (
+          <li key="see-all">
+            <em>
+              <a href={url}>
+                {expand2text(
+                  seeAllMessage(entities.count),
+                  {num: entities.count},
```
mwiencek, line 67:
```suggestion
                  {num: formatCount($c, entities.count)},
```

mwiencek (approved):

root/artist_credit/EntityList.js
reosarevok: Made the change in a second commit (both to these and the tag pages these are based on).

We probably should move that utility out of statistics, but I'm not sure where. If you have any suggestions, I can move it in a third commit.

yvanzo (approved):
Thank you!

PR #1820, opened 2020-12-07
MBS-8371: Add searching in edit notes to the edit search

reosarevok:
### Implement MBS-8371

This seems kinda-sorta slow but probably not insanely so (no timeout), and given we have no other way to do anything like it, users probably will be willing to wait a bit for results.

For an example of usage I could imagine, something like http://localhost:5000/search/edits?auto_edit_filter=&order=desc&negation=0&combinator=and&conditions.0.field=edit_note_content&conditions.0.operator=includes&conditions.0.args.0=fuck+you&conditions.8.field=vote_count&conditions.8.vote=0&conditions.8.operator=%3E&conditions.8.args.0=0&conditions.8.args.1= (on the mirror DB in pink) will show 12 edits that got downvotes *and* have edit notes including "fuck you", so would be useful for the community manager to review.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/EditNoteContent.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/EditNoteContent.pm
@@ -1,0 +1,42 @@
+package MusicBrainz::Server::EditSearch::Predicate::EditNoteContent;
+use Moose;
+
+with 'MusicBrainz::Server::EditSearch::Predicate';
+
+sub operator_cardinality_map {
+  return (
+      'includes' => undef,
+  )
+}
+
+sub combine_with_query {
+    my ($self, $query) = @_;
+
+    my @patterns = map {
+      $_ =~ s/\\/\\\\/g;
+      $_ =~ s/_/\\_/g;
+      $_ =~ s/%/\\%/g;
+      '%' . $_ . '%'
+    } @{ $self->sql_arguments };
+
+    $query->add_where([
+        'EXISTS (
+          SELECT TRUE FROM edit_note
+          WHERE edit_note.text ILIKE ?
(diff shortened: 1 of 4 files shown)


jesus2099:
LOL, your example! :D

lib/MusicBrainz/Server/EditSearch/Predicate/EditNoteContent.pm
```diff
@@ -0,0 +1,35 @@
...
+sub combine_with_query {
+    my ($self, $query) = @_;
+
+    $query->add_where([
+        'EXISTS (
+          SELECT TRUE FROM edit_note
+          WHERE edit_note.text LIKE ?
+            AND edit_note.edit = edit.id
+        )',
+        [ map { '%' . $_ . '%' } @{ $self->sql_arguments } ]
```
mwiencek, line 21: You'll have to escape `%` and `_` and `\` there since those three chars have special meaning. Maybe something like
```perl
    my @patterns = map {
      $_ =~ s/\\/\\\\/g;
      $_ =~ s/_/\\_/g;
      $_ =~ s/%/\\%/g;
      '%' . $_ . '%'
    } @{ $self->sql_arguments };
```

```suggestion
        \@patterns,
```

mwiencek (changes requested):
Works fine other than the escaping issue. I contemplated adding a search index for edit_note.text, but it wouldn't work well at all for searching phrases like in your example, so an exact substring match is probably best for now if slow. We might want to go case insensitive with `ILIKE` though?

For some reason if I click on your link while logged out it uses "ID" for the edit note content field, but that might be an existing issue with the other edit note fields.

mwiencek (approved):

PR #1821, opened 2020-12-07
MBS-11267: Always show artwork info when adding/reordering

reosarevok:
### Implement MBS-11267

While generally you can see the types and comment for a specific image when hovering over the image, that's not too clear in general and it doesn't help at all when the thumbnail hasn't been generated yet (since there's no image to hover over).

This adds an information icon that displays the data on hover, and also shows if there's no thumbnail yet.

Additionally, we weren't loading types for add_cover_art.

Sample:

Screenshot from 2020-12-07 21-32-31

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -226,6 +226,7 @@ sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
     my @mime_types = map { $_->{mime_type} } @{ $c->model('CoverArt')->mime_types };
 
     my @artwork = @{ $c->model('Artwork')->find_by_release($entity) };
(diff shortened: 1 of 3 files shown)


mwiencek:
This works, but I feel like it'd be better to just display things the same way we do on the cover art page (have all the text in the box already). We could even re-use the component (with flags to hide the links/buttons). Do the boxes have to be so much smaller on the edit pages that we have to hide the text?

If we want to keep the boxes small, another option could be to use root/static/scripts/edit/components/Tooltip.js and just show an HTML tooltip immediately when you hover over the image.

mwiencek (approved):
The improvements I suggested seem quite difficult until more of these pages are converted to React, so I'm fine with merging this as-is, even if it seems a little weird. Sorry for delaying this for no reason.

PR #1822, opened 2020-12-07
MBS-11268: Show "Set track durations" on release/discids

reosarevok:
### Implement MBS-11268

We didn't use to show this, probably because IIRC  there was no indication at all of how a disc id's track times matched the medium's when applying the edit (so it made sense to force users to always go through the specific discID page). This has been changed though, so we can display the links here, the user can click them and then compare the durations on the next page. The user will still need to go through two pages to get all the info, but it makes the links easier to find (many people are unlikely to have realized you could do this by clicking the weird discid title).


mwiencek (approved):

yvanzo (approved):
Works fine :ship:

PR #1823, opened 2020-12-08
Fix broken date formatter cache set

mwiencek:
Very obviously broken and not populating the cache at all.


mwiencek:
> not populating the cache at all.

(In fact it does populate the cache with `formatter` as the key and `undefined` as the value, which just grows infinitely.)

yvanzo (approved):

reosarevok (approved):

PR #1824, opened 2020-12-09
MBS-10208: Allow merging collections

reosarevok:
### Implement MBS-10208

This allows merging collections of the same entity type, moving all entities to the destination collection. All collaborators for the collections being merged will be collaborators in the destination collection. The descriptions of all collections will be appended (as with annotations in entity merges) for the user to later edit them as desired.

To stop a user from being able to see info about other users' private collections by entering random IDs into the merge queue URL, the merge queue detaches to 403 if passed a collection with a different owner.


reosarevok:
Moving to draft since we decided to add SQL for redirects to it.

reosarevok:
Tested locally, redirects work, nothing else seems broken either.

lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -195,7 +195,7 @@ my %stats = (
         SQL => "SELECT " .
             join(' + ',
                  (map { "(SELECT COUNT(gid) FROM $_)" } entities_with('mbid', take => sub { my $type = shift; return shift->{table} // $type })),
-                 (map { "(SELECT COUNT(gid) FROM ${_}_gid_redirect)" } entities_with(['mbid', 'multiple'])))
+                 (map { "(SELECT COUNT(gid) FROM ${_}_gid_redirect)" } entities_with(['mbid', 'multiple'], take => sub { my $type = shift; return shift->{table} // $type })))
```
reosarevok, line 198: For some reason this just assumed it should use the name of the entity, not the table like the above. I assume this exception was added specifically for collection, which didn't use to have multiple.

lib/MusicBrainz/Server/Controller/Collection.pm
```diff
@@ -296,6 +297,33 @@ sub edit : Chained('own_collection') RequireAuth {
...
+    $c->model('CollectionType')->load(@collections);
+
+    for my $collection (@collections) {
+        $c->model('Editor')->load_for_collection($collection);
+    };
+
+    my @entity_types = uniq map { $_->type->item_entity_type } @collections;
+    my @privacy_settings = uniq map { $_->public } @collections;
+    use Data::Dumper;
+    print Dumper(\@privacy_settings);
```
mwiencek, line 320:
```suggestion

```

lib/MusicBrainz/Server/Controller/Collection.pm
```diff
@@ -296,6 +297,33 @@ sub edit : Chained('own_collection') RequireAuth {
...
+    $c->stash(
+        privacy_differs => @privacy_settings > 1,
```
mwiencek, line 322: "privacies_differ" would be a bit more consistent with types_differ

lib/MusicBrainz/Server/Controller/Role/Merge.pm
```diff
@@ -195,22 +206,30 @@ role {
...
+                    old_entities => [ map +{
+                        id => $entity_id{$_}->id,
+                        name => $entity_id{$_}->name,
+                        $self->_extra_entity_data($c, $form, $entity_id{$_})
+                    }, @old_ids ],
+                    (map { $_->name => $_->value } $form->edit_fields),
+                    $self->_merge_parameters($c, $form, $entities)
+                );
+            } else {
+                $c->model('Collection')->merge(
```
mwiencek, line 227: I think it's a good idea to check if the entity type is 'collection' here (even if it's the only thing for which `$params->edit_type` is unset right now).

lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -111,6 +113,77 @@ sub is_empty {
     return $non_empty ? 0 : 1;
 }
 
+sub merge {
+    my ($self, $new_id, $old_ids, $user_id, %opts) = @_;
```
mwiencek, line 117:
```suggestion
    my ($self, $new_id, $old_ids, $user_id) = @_;
```

lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -111,6 +113,77 @@ sub is_empty {
...
+    my @entity_types = uniq map { $_->type->item_entity_type } @collections;
+
+    if (@entity_types > 1) {
+        confess('Attempt to merge collections of different entity types');
+    }
+
+    my $type = $entity_types[0];
+
+    # Remove duplicate entities: for all entities that exist in multiple
+    # collections, remove them from all but one.
```
mwiencek, line 137: Although we don't really use `added`, `position`, or `comment` right now, I feel like there should be some sort of precedence when merging those...

 * Prefer the earliest value for `added`?
 * Prefer the position/comment of the target entry, or if unset then the earliest entry for which those are set (based on `added`)?

lib/MusicBrainz/Server/Entity/Collection.pm
```diff
@@ -66,6 +66,7 @@ around TO_JSON => sub {
     $json->{description_html} = format_wikitext($self->description);
     $json->{editor_is_limited} = boolean_to_json(defined $editor ? $editor->is_limited : 0);
     $json->{collaborators} = to_json_array($self->collaborators);
+    $self->type ? ($json->{item_entity_type} = $self->type->item_entity_type) : ();
```
mwiencek, line 69:
```suggestion
    $json->{item_entity_type} = $self->type->item_entity_type if $self->type;
```

t/lib/t/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
```diff
@@ -38,7 +38,7 @@ test all => sub {
     is($medium->tracks->[5]->length, 276933);
     is($medium->tracks->[6]->length, 94200);
 
-    isa_ok exception {
+    ok exception {
```
mwiencek, line 41: Not sure I follow this change.

mwiencek (changes requested):

lib/MusicBrainz/Server/Controller/Collection.pm
reosarevok: It just sounded kinda stupid :D But I guess.

lib/MusicBrainz/Server/Data/Collection.pm
reosarevok: Comments could be concatenated, like annotations are, and let the user clean up. But not sure about the others... I guess earliest added and target position makes sense. I guess we would need to take extra steps before this one to clean this up then...

t/lib/t/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
reosarevok: Not sure I follow it either :D Reverted.

lib/MusicBrainz/Server/Controller/Role/Merge.pm
reosarevok: I think the only way we have to do this is by checking `@$entities[0]->entity_type` - am I missing a better option? Made that change for now.

mwiencek: I think you can check `$c->namespace eq 'collection` too. Or what you have, but should be `$entities->[0]->entity_type` if it's an array ref.

lib/MusicBrainz/Server/Data/Collection.pm
mwiencek: Waiting is fine with me. Want to add a `# TODO` here at least?

mwiencek:
Added missing primary/foreign keys to the upgrade scripts (found via t/script/CheckSchemaMigration.sh).

mwiencek (approved):

PR #1832, opened 2020-12-17
MBS-2421: Show entity icon on relationship links

reosarevok:
### Implement MBS-2421

Putting this up since I've had it around for at least a couple months now IIRC. This works on all React-converted places I wanted it to at first (the release page rels and entities' relationship pages / overview pages). It doesn't currently work on the release page's bottom-credits section (which is still TT - I was waiting on that conversion to be finished before I put this up but I'm not sure it makes sense, since it's trivial to add to that once it's Reactified). It also doesn't currently show on the relationship editor - I'm not sure whether we'd want it there or not, open to opinions! :) 

Thanks @mwiencek for inverting the icons.


mwiencek (approved):
I've fixed the conflicts here and checked that the icons display for the bottom-credits sections on release pages (now that that's been converted to React).  Thanks!

PR #1833, opened 2020-12-28
MBS-11291: Fix wrong bootleg check in ReleasesSameBarcode

reosarevok:
### Fix MBS-11291

We were checking the same release twice, instead of checking whether both releases were not bootlegs.


mwiencek (approved):

yvanzo (approved):
LGTM but did not test.

PR #1834, opened 2020-12-28
MBS-11290: Report of works of type X part of another work of type X 

reosarevok:
### Implement MBS-11290

There's plenty of cases where new users add every movement of a concerto (or sonata, or whatever) with the same type as the parent work.  That's pretty much never correct.

This shouldn't have many (if any) false positives from popular music, since part of work relationships are very uncommon there, and even then it seems unlikely that "Song is part of Song" is correct. We could always disable this for "Song" if needed, but that would also hide correct hits where someone has, say, selected "Song" for a parent "Song-Cycle" work so I'd rather avoid it unless users complain.

On top of #1787

### DIFF
--- a/lib/MusicBrainz/Server/Report/WorkSameTypeAsParent.pm
+++ b/lib/MusicBrainz/Server/Report/WorkSameTypeAsParent.pm
@@ -1,0 +1,35 @@
+package MusicBrainz::Server::Report::WorkSameTypeAsParent;
+use Moose;
+
+with 'MusicBrainz::Server::Report::WorkReport',
+     'MusicBrainz::Server::Report::FilterForEditor::WorkID';
+
+sub query {<<~'EOSQL'}
+    SELECT DISTINCT w.id AS work_id,
+           row_number() OVER (ORDER BY w.type, w.name COLLATE musicbrainz)
+    FROM work w
+    WHERE EXISTS (
+        SELECT 1
+        FROM l_work_work lww
+        JOIN link l ON lww.link = l.id
+        JOIN link_type lt ON l.link_type = lt.id
+        JOIN work w2 ON w2.id = lww.entity0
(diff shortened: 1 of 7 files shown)


lib/MusicBrainz/Server/Report/WorkSameTypeAsParent.pm
```diff
@@ -0,0 +1,36 @@
...
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT
+This file is part of MusicBrainz, the open internet music database.
+Copyright (C) 2020 MetaBrainz Foundation
+Licensed under the GPL version 2, or (at your option) any later version:
+http://www.gnu.org/licenses/gpl-2.0.txt
+=cut
```
mwiencek, line 36:
```suggestion
=head1 COPYRIGHT AND LICENSE

Copyright (C) 2021 MetaBrainz Foundation

This file is part of MusicBrainz, the open internet music database,
and is licensed under the GPL version 2, or (at your option) any
later version: http://www.gnu.org/licenses/gpl-2.0.txt

=cut
```

lib/MusicBrainz/Server/Report/WorkSameTypeAsParent.pm
```diff
@@ -0,0 +1,36 @@
+package MusicBrainz::Server::Report::WorkSameTypeAsParent;
+use Moose;
+
+with 'MusicBrainz::Server::Report::WorkReport',
+     'MusicBrainz::Server::Report::FilterForEditor::WorkID';
+
+sub query {
+    q{
```
mwiencek, line 8: `<<~'EOSQL'` would be a bit nicer here

mwiencek (commented):
just two small stylistic comments

mwiencek (approved):

PR #1836, opened 2020-12-29
MBS-11292: Do not use UN flag for [Worldwide]

reosarevok:
### Implement MBS-11292

As suggested in the ticket, that's not quite appropriate.

This uses a globe flag - sadly, it gets so distorted at this tiny size that it is hard to understand what's going on, but that was also completely true of the previous one.


mwiencek (approved):
Either image is unintelligible to me unless my nose touches the screen

yvanzo (changes requested):
Its darkened blue color makes it less distinguishable from the European Union flag, which is even less representative. Ticket’s suggestion of either a visually similar icon (same color, different symbol) or a blue globe like the emoji would be a better fit.

reosarevok:
Updated with the flag pmepepnoute uploaded to Jira, does seem better.

yvanzo (approved):

PR #1837, opened 2020-12-29
MBS-11288: Drop decoda.com from lyrics whitelist

reosarevok:
### Implement MBS-11288

From the ticket:
"Site has been abandoned, taken over or sold off to someone else. Legal status uncertain, intentions due to reports of malicious redirects questionable at best."

I'm keeping isGreyedOut for any future need though, it seems like a useful tool to have around.


root/url/utility/isGreyedOut.js
```diff
@@ -7,10 +7,17 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-const greyRegExp = new RegExp('^https?://(?:www\.)?decoda\.com/');
+/*
+ * NB: No URLs currently need to be greyed out;
+ * this should always return false until that changes
+ */
+
+// const greyRegExp = new RegExp('');
```
mwiencek, line 15: I think I'd prefer just changing this to something like `/(?!)/`, which matches nothing, to avoid having to revert all this stuff whenever we need or don't need this.

mwiencek (approved):

root/url/utility/isGreyedOut.js
```diff
@@ -7,7 +7,12 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-const greyRegExp = new RegExp('^https?://(?:www\.)?decoda\.com/');
+/*
+ * NB: No URLs currently need to be greyed out;
+ * this should always return false until that changes
+ */
+
+const greyRegExp = new RegExp('/(?!)/');
```
mwiencek, line 15: Sorry, the `/(?!)/` was RegExp literal syntax

```suggestion
const greyRegExp = new RegExp('(?!)');
```

lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -50,8 +50,6 @@ my %URL_SPECIALIZATIONS = (
     'DAHR'                => qr{^https?://adp\.library\.ucsb\.edu/}i,
     'Dailymotion'         => qr{^https?://(?:www\.)?dailymotion\.com/}i,
     'DanceDB'             => qr{^https?://(?:www\.)?tedcrane\.com/DanceDB/}i,
-    # Temporarily grey Decoda out as it has been reported to be compromised
-    # 'Decoda'              => qr{^https?://(?:www\.)?decoda\.com/}i,
```
yvanzo, line 54: Dropping this makes Decoda URLs no longer greyed out under entity’s “Relationships” tab.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1153,12 +1153,6 @@ const testData = [
                      input_url: 'http://www.dailymotion.com/playlist/xwvuts_who-knows_top/1#video=xyztuvw',
             expected_clean_url: 'https://www.dailymotion.com/video/xyztuvw',
   },
-  // Decoda
-  {
-                     input_url: 'http://decoda.com/robi-on-ne-meurt-plus-damour-lyrics',
-             input_entity_type: 'work',
-    expected_relationship_type: 'lyrics',
-  },
```
yvanzo, line 1161: Maybe we can do more and completely blacklist `decoda.com` URLs for any type of relationship.

yvanzo (changes requested):

lib/MusicBrainz/Server/Data/URL.pm
reosarevok: I can drop all existing decoda rels before merging this (I just finished the previous MBBE bot run anyway), so if we block stuff as suggested below this should be fine.

yvanzo: In any case, there will be a period of time during which Decoda URLs can still be entered in production and seen in beta. Let’s be safe and keep these greyed out in next server release, and then drop this code as well as the file `lib/MusicBrainz/Server/Entity/URL/Decoda.pm` (missing from this PR) only after this patch has been deployed in production.

root/static/scripts/edit/externalLinks.js
```diff
@@ -240,6 +240,9 @@ export class ExternalLinksEditor
             } else if (isMusicBrainz(link.url)) {
               error = l(`Links to MusicBrainz URLs are not allowed.
                          Did you mean to paste something else?`);
+            } else if (isBlacklisted(link.url)) {
+              error = l(`Links to this website are not allowed
+                         because it is in our site-wide blacklist.`);
```
yvanzo, line 245:
```suggestion
            } else if (isBlacklisted(link.url)) {
              error = l(`Links to this website are not allowed
                         because it is known to host malware.`);
```

yvanzo (changes requested):
Works nicely: http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/relationships

Just please make the error message more explanatory.

lib/MusicBrainz/Server/Data/URL.pm
yvanzo: Thanks for having dropped `lib/MusicBrainz/Server/Entity/URL/Decoda.pm` (which was the real issue here), and still greying out  the URL in JavaScript code.

root/static/scripts/edit/externalLinks.js
reosarevok: I was thinking we'd probably want to expand this one eventually also to stuff like sites known to host illegal downloads, for example, which is why I left a wider message here.

yvanzo: Smart links are already being blocked, still we provide an explanatory message. It would be nice to do the same here rather than creating a catch-all. You might want not to refer to “blacklist” in code either.

reosarevok: Ok, if you want we can do that, and have basically an "isDecoda" section. I don't expect this malware thing is going to extend to any other sites pretty much ever though.

What's the problem with blacklist? We already use whitelist elsewhere.

yvanzo: Editors might be interested to know that a lyrics website they used to visit is actually malicious.

reosarevok: Ok, made this explicitly about malware, especially since it seems we might want to add this starzik page there too... this might be more common than I expected.

yvanzo (approved):
Thank you!

PR #1838, opened 2020-12-29
MBS-11296 / MBS-11297: More blocks for Wikipedia/Wikidata at release level

reosarevok:
### Implement MBS-11296 / MBS-11297

I looked at the most common misuses of these links on the release level and added errors rejecting them and telling the users to add them to the RG level (as we already did for the Wikipedia / Discography entry combo).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3470,7 +3470,7 @@ Object.values(LINK_TYPES).forEach(function (linkType) {
 });
 
 // avoid Wikipedia/Wikidata being added as release-level discography entry
-const originalRule = validationRules[LINK_TYPES.discographyentry.release];
+const discographyRule = validationRules[LINK_TYPES.discographyentry.release];
 validationRules[LINK_TYPES.discographyentry.release] = function (url) {
   if (/^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\//.test(url)) {
     return {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3478,7 +3481,63 @@ validationRules[LINK_TYPES.discographyentry.release] = function (url) {
...
+      ),
+      result: false,
+    };
+  }
+  return discographyRule(url);
+};
+
+// avoid Wikipedia/Wikidata being added as release-level license entry
+const licenseRule = validationRules[LINK_TYPES.license.release];
+validationRules[LINK_TYPES.license.release] = function (url) {
```
mwiencek, line 3498: Not sure I understand why we override the functions instead of adding it into the originals (or alternatively, call the added functions from the original function instead of doing it backwards).

mwiencek (approved):

yvanzo (approved):
Makes sense!

root/static/scripts/edit/URLCleanup.js
yvanzo: Maybe it wasn’t possible in the past, but since `LINK_TYPES.wikipedia` has no `release` key, it should be possible to auto-select `discographyentry` type for release using the _wikipedia_ rule and invalidate it there. Same for wikidata.

reosarevok: Want to add a ticket for the refactoring?

yvanzo: @mwiencek: A rule can possibly auto-select only one relationship type for each entity type. And rules have to be exclusive to each other. The _original function_ is created dynamically by testing all rules, so we can only modify functions like that.

@reosarevok: Actually, we cannot just _auto-select `discographyentry` type for release using the wikipedia rule and invalidate it there_ because editor would still be able to select another relationship type from the release editor.

This is more or less related to MBS-9902 which requires a deep refactoring of URL cleanup code. Possibly an idea for GSoC along with reworking URL relationship editor UI?

PR #1843, opened 2020-12-30
MBS-10830: Only show Remove link if entity can be removed

reosarevok:
### Implement MBS-10830

Passing can_delete through $c.stash seemed like the most straightforward option, but open to other better possibilities.

### DIFF
--- a/root/layout/components/sidebar/RemoveLink.js
+++ b/root/layout/components/sidebar/RemoveLink.js
@@ -9,6 +9,7 @@
 
 import * as React from 'react';
 
+import {CatalystContext} from '../../../context';
 import entityHref from '../../../static/scripts/common/utility/entityHref';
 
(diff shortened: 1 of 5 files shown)


root/layout/components/sidebar/RemoveLink.js
```diff
@@ -12,6 +12,7 @@ import * as React from 'react';
 import entityHref from '../../../static/scripts/common/utility/entityHref';
 
 type Props = {
+  +$c: CatalystContextT,
```
mwiencek, line 15: FWIW since hooks were added, it's usually easier to use https://reactjs.org/docs/hooks-reference.html#usecontext than pass `$c` everywhere.

mwiencek (approved):

yvanzo (approved):

PR #1844, opened 2021-01-04
MBS-11310: Clarify when discid would cause subsecond changes

reosarevok:
### Implement MBS-11310

This is currently confusing because it seems that there would be no changes at all if the edit was entered. In reality, it will cause sub-second changes, and we should specify that for clarity.


yvanzo (approved):

mwiencek (approved):

mwiencek:
I'd probably perform this check in the controller and pass a boolean rather than building strings & comparing them in the template, but I don't see any bugs with the approach.

PR #1847, opened 2021-01-04
MBS-11304: Add recordings’ first release date to release lookup

reosarevok:
### Implement MBS-11304

Release is needed for Picard.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Controller/WS/2/JSON/LookupRelease.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/WS/2/JSON/LookupRelease.pm
@@ -482,6 +482,7 @@ test 'release lookup with labels and recordings' => sub {
                                 length => 243000,
                                 disambiguation => "",
                                 video => JSON::false,
+                                "first-release-date" => '2004-03-17',
                             }
                         },
                         {
@@ -496,6 +497,7 @@ test 'release lookup with labels and recordings' => sub {
                                 length => 222000,
                                 disambiguation => "",
                                 video => JSON::false,
+                                "first-release-date" => '2004-03-17',
                             }
                         },
                         {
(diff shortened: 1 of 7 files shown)


yvanzo (approved):
* Checked parity with search results (which just do not return recordings but at top level)
* Checked compatibility with/without active schema 26
* Updated ticket MBS-11304 accordingly

Note: `inc=recording-rels` is not affected.

yvanzo:
@reosarevok: It requires updating https://wiki.musicbrainz.org/MusicBrainz_API though.

reosarevok:
@yvanzo: you mean the examples at https://wiki.musicbrainz.org/MusicBrainz_API/Examples ? I don't see anywhere on the basic page that this would be mentioned :)

yvanzo:
Right, please link to the corresponding wiki diff(s).

t/lib/t/MusicBrainz/Server/Controller/WS/2/JSON/LookupRelease.pm
```diff
@@ -792,6 +796,7 @@ test 'release lookup with discids' => sub {
                                 length => 213106,
                                 disambiguation => "",
                                 video => JSON::false,
+                                "first-release-date" => '2001-07-04',
```
mwiencek, line 799:
```suggestion
                                "first-release-date" => '2001-07-04',
```

mwiencek (changes requested):
Mentioned this on IRC, but I'd prefer not adding it to /ws/2/artist/?inc=recordings if Picard doesn't need it. That feature is useless (since it returns a fixed list of 25 recordings by the artist) and supplanted by proper recording browse requests. Returning more information there might signal to people in the patch notes that they should actually be using that feature.

reosarevok:
Dropped it from artists then.

reosarevok:
Ok, removed it from artist then :)

mwiencek (approved):

PR #1848, opened 2021-01-04
Upgrade Flow to 0.141.0 (fixes issues with new generics)

mwiencek:
This resolves some issues that have arisen due to the new generics implementation. See https://medium.com/flow-type/flows-improved-handling-of-generic-types-b5909cc5e3c5 for more information.

`./node_modules/.bin/flow` passes locally.

### DIFF
--- a/root/static/scripts/common/i18n/expand2react.js
+++ b/root/static/scripts/common/i18n/expand2react.js
@@ -146,6 +146,7 @@ function pushChild<T>(
   if (lastIndex >= 0 &&
       typeof match === 'string' &&
       typeof children[lastIndex] === 'string') {
+    // $FlowIssue[incompatible-type]
     children[lastIndex] += match;
   } else {
     children.push(match);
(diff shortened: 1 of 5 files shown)


root/static/scripts/common/i18n/expand2react.js
```diff
@@ -146,14 +146,17 @@ function pushChild<T>(
...
+    // $FlowIssue[incompatible-type]
     children[lastIndex] += match;
   } else {
     children.push(match);
   }
   return children;
 }
 
-function concatArrayMatch<T>(
+type MatchUpperBoundT = StrOrNum | {...};
```
reosarevok, line 157: I'm not sure I follow what this one means.

reosarevok (approved):
LGTM, assuming the bit I don't follow is right :)

root/static/scripts/common/i18n/expand2react.js
mwiencek: Added a comment here -- does that help some? `{...}` just matches any object.

Besides outputting React elements, another case where we parse a string into objects is HTML attributes:

https://github.com/metabrainz/musicbrainz-server/blob/0550d68/root/static/scripts/common/i18n/expand2react.js#L279-L282

(Then we pass that object as the second argument to `React.createElement`.)

yvanzo (approved):

PR #1849, opened 2021-01-04
MBS-11313: Fix OAuth PKCE S256 verification

mwiencek:
As the ticket describes, the code verifier is not meant to be base64url-decoded during verification; the encoding when generating the verifier is only to transform the initial random octets into something URL safe. The encoded string is in fact the code verifier and not the initial random octets.

I fixed the automated tests and ran `prove -vl t/tests.t :: --tests Controller::OAuth2` to check the change.


atj:
Thanks for fixing this so quickly!

reosarevok (approved):

yvanzo (approved):
:shipit:

PR #1850, opened 2021-01-05
MBS-11514 / MBS-11515 / MBS-11516 / MBS-11517: Use React.useContext in more places rather than passing $c around

reosarevok:
### Implements MBS-11514 / MBS-11515 / MBS-11516 / MBS-11517

Some big chunks (like Layout) left for a second PR to avoid making this huge. Reports bits done in https://github.com/metabrainz/musicbrainz-server/pull/1787 since that already refactors the whole thing.

This shaves some lines, but more importantly it makes it easier to understand and follow and avoids passing $c down several components just to reach the one N steps down the line.

I'd suggest reviewing with the whitespace ignore on, since most changes other than just the $c bit itself are to indent some components because of adding return () to them.

Tested by navigating around locally, but not extensively - plus Flow to ensure I didn't keep passing $c where I shouldn't anymore.

### DIFF
--- a/root/collection/CollectionIndex.js
+++ b/root/collection/CollectionIndex.js
@@ -53,12 +53,10 @@ type Props =
   | PropsForEntity<WorkT>;
 
 const listPicker = (
-  $c: CatalystContextT,
   props: Props,
   canRemoveFromCollection: boolean,
 ) => {
   const sharedProps = {
-    $c,
     checkboxes: canRemoveFromCollection ? 'remove' : '',
     order: props.order,
     sortable: true,
@@ -84,7 +82,6 @@ const listPicker = (
     case 'event':
       return (
         <EventList
-          $c={$c}
           events={props.entities}
           showArtists
           showLocation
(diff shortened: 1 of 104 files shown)


yvanzo (approved):
Proofreading the code is blazing fast when hiding space changes indeed. Quickly tested it too.

root/collection/CollectionIndex.js
```diff
@@ -161,8 +158,9 @@ const listPicker = (
 
 const CollectionIndex = (props: Props):
 React.Element<typeof CollectionLayout> => {
+  const $c = React.useContext(CatalystContext);
```
mwiencek, line 161: I think it's better not to do this for top-level components, since it's always passed as a prop in that case: https://github.com/metabrainz/musicbrainz-server/blob/b036579/root/server/response.js#L100

So it's actually simpler to just use the prop here. I'm not sure if this is the only example in this PR.

reosarevok: I rechecked, I think this was the only one. Amended.

mwiencek (approved):

PR #1851, opened 2021-01-06
MBS-11314: Add legends to all sort of timeline statistics

reosarevok:
### Implement MBS-11314

MBS-8169 adds a link to the timeline for almost every statistic we have, but most of them don't have a label and just give the stat name (such as "count.release.status.1"). This adds readable labels for them.


zas:
Just hit this issue, but it seems i'm late for bug report ;)
Nice to see you're working on it (btw, timeline for almost all entities is a good thing).

reosarevok:
Rebased this to be on top of https://github.com/metabrainz/musicbrainz-server/pull/2292 so I could get rid of all the descriptions.

PR #1853, opened 2021-01-08
MBS-11319: Fix adding cover art in beta

reosarevok:
This seems to fix the issue, although testing perfectly is hard locally since I don't have a CAA server set up. For testing I made it so that an edit is created even if the image fails to upload, and then made sure the edit actually got created with the change.

The second commit fixes a warning. I think it's correct (we should never need to pick a default value for a multiple, I expect, so I skipped adding an analog function to getSelectValue) but do let me know if it doesn't seem right :)

### DIFF
--- a/root/components/SelectField.js
+++ b/root/components/SelectField.js
@@ -24,29 +24,88 @@ const buildOptGroup = (optgroup, index) => (
   </optgroup>
 );
 
-type SelectElementProps = {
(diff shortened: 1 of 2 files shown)


mwiencek (approved):

yvanzo (approved):
:shipit:

PR #1855, opened 2021-01-11
MBS-11324: Trim input in the barcode otherlookups field

reosarevok:
### Implement MBS-11324

We remove all spaces when a barcode is pasted on the RE and store them without spaces. As such, we should do the same when a user pastes, say, " 9 78849 2540655" in the barcode search, so that the actual search submitted ("barcode:9788492540655") can find results.

Tested manually with the barcode above.


mwiencek (approved):

yvanzo (approved):
It looks like an issue to be also addressed in the search schema for the barcode field in release core: https://github.com/metabrainz/mbsssss/blob/v-2020-12-24/release/conf/schema.xml#L14 as `OtherLookup` is not the only way to send `barcode:…` queries to the search server. Just added SEARCH-637.

PR #1856, opened 2021-01-11
MBS-11322: Update Resident Advisor URL cleanup and add validation

reosarevok:
### Fix / Implement MBS-11322

Marking partially as bug because currently the updated URLs are not whitelisted.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2642,6 +2644,36 @@ const CLEANUPS = {
       return url.replace(/^(?:https?:\/\/)?(?:[^.]+\.)?recochoku\.jp\/(album|artist|song)\/([a-zA-Z0-9]+)(\/)?.*$/, 'http://recochoku.jp/$1/$2/');
     },
   },
+  'residentadvisor': {
+    match: [
+      new RegExp('^(https?://)?(www\\.)?ra\\.co/', 'i'),
+    ],
```
yvanzo, line 2650: It would be nice to still match old-style `residentadvisor.net` URLs and do something helpful of it: Ideally convert them to the current style `ra.co` if there is a clear mapping between old style and current style.

lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -137,7 +137,8 @@ my %URL_SPECIALIZATIONS = (
     'PsyDB'               => qr{^https?://(?:www\.)?psydb\.net/}i,
     'QuebecInfoMusique'   => qr{^https?://(?:www\.)?qim\.com/}i,
     'Rateyourmusic'       => qr{^https?://(?:www\.)?rateyourmusic\.com/}i,
-    'ResidentAdvisor'     => qr{^https?://(?:www\.)?residentadvisor\.net/}i,
+    # Remove residentadvisor.net once MBBE-31 is done
+    'ResidentAdvisor'     => qr{^https?://(?:www\.)?(ra\.co|residentadvisor\.net)/}i,
```
yvanzo, line 141: :+1:

yvanzo (changes requested):

root/static/scripts/edit/URLCleanup.js
reosarevok: I thought about it, but given it seems they all redirect, it doesn't seem worth the effort. At most, maybe we could detect *any* residentadvisor.net URL and ask the user to follow the redirect?

reosarevok: I did that in a separate commit.

yvanzo (approved):
Thank you!

mwiencek (approved):

PR #1857, opened 2021-01-11
MBS-11320 / MBS-11321: Don't add new medium if empty medium existed

reosarevok:
### Fix MBS-11320 / MBS-11321

For cases where an existing empty medium is already present, present the track parser (rather than the Add disc dialog) if the user disables the "tracklist is unknown" checkbox, and don't overwrite the medium if using Add medium.

### DIFF
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -358,6 +358,7 @@ class Medium {
     var tracks = data.tracks;
     this.tracks = ko.observableArray(utils.mapChild(this, tracks, Track));
     this.tracksUnknownToUser = ko.observable(false);
+    this.tracksWereUnknownToUser = false;
 
     var self = this;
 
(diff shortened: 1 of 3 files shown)


root/static/scripts/release-editor/fields.js
```diff
@@ -358,6 +358,7 @@ class Medium {
     var tracks = data.tracks;
     this.tracks = ko.observableArray(utils.mapChild(this, tracks, Track));
     this.tracksUnknownToUser = ko.observable(false);
+    this.tracksWereUnknownToUser = false;
```
mwiencek, line 361: How come this adds a new variable instead of just using `tracksUnknownToUser`?

reosarevok: Because we care about whether the tracks were unknown *at first*, but tracksUnknownToUser is updated when the user unchecks the checkbox.

mwiencek (approved):

yvanzo (approved):
It now has conflicts though.

PR #1858, opened 2021-01-12
MBS-10964: Prevent and handle empty artist credit name submissions in the release editor

mwiencek:
https://tickets.metabrainz.org/browse/MBS-10964

(See commits for details.)

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
@@ -42,7 +42,12 @@ use MusicBrainz::Server::Data::Utils qw(
 );
 use MusicBrainz::Server::Renderer qw( render_component );
 use MusicBrainz::Server::Translation qw( comma_list comma_only_list l );
-use MusicBrainz::Server::Validation qw( is_guid is_valid_url is_valid_partial_date );
+use MusicBrainz::Server::Validation qw(
+    is_database_row_id
+    is_guid
+    is_valid_url
+    is_valid_partial_date
+);
 use MusicBrainz::Server::View::Base;
 use Readonly;
 use Scalar::Util qw( looks_like_number );
@@ -244,7 +249,7 @@ sub process_release_events {
 sub process_artist_credits {
     my ($c, $loader, @artist_credits) = @_;
 
-    my @artist_gids;
+    my @artist_ids;
 
     for my $ac (@artist_credits) {
         my @names = @{ $ac->{names} };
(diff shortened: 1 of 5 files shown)


mwiencek:
> (Possibly) fixes MBS-10964

I should clarify that this does definitely fix one case where I can always reproduce the "empty AC name" issue -- I'm just not sure there aren't other causes of it yet, i.e. if jesus2099 is running into some other case. I've asked for browser details on the ticket.

mwiencek:
Added https://github.com/metabrainz/musicbrainz-server/pull/1858/commits/175765838004d0e23c268860f03d759025dcad1f which fixes the issue for good, even if someone manages to trigger an empty AC name by some other method in a newer browser.

yvanzo (approved):
I could not reproduce the bug with in latest Firefox but I could not find any regression either.

lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
```diff
@@ -261,26 +266,34 @@ sub process_artist_credits {
             trim_string($name, 'name');
             trim_string($artist, 'name');
 
-            if (!$artist->{id} && is_guid($artist->{gid}))  {
-                push @artist_gids, $artist->{gid};
+            if (is_database_row_id($artist->{id}))  {
+                push @artist_ids, $artist->{id};
+            } elsif (is_guid($artist->{gid})) {
+                push @artist_ids, $artist->{gid};
```
reosarevok, line 272: Why does this need to check for gid or id? The commit message doesn't talk about that and it's not obvious to me :)

reosarevok (commented):
Should the check on https://github.com/metabrainz/musicbrainz-server/blob/c27ff7398a46104189c8df428a9cbddd8bee681e/lib/MusicBrainz/Server/Edit/Utils.pm#L60-L62 be removed now or can it still be useful?

mwiencek:
> Should the check on
> 
> https://github.com/metabrainz/musicbrainz-server/blob/c27ff7398a46104189c8df428a9cbddd8bee681e/lib/MusicBrainz/Server/Edit/Utils.pm#L60-L62
> 
> be removed now or can it still be useful?

I'd keep this just in case we manage to find another way to trigger it.

lib/MusicBrainz/Server/Controller/WS/js/Edit.pm
mwiencek: Updated the commit message, hope this helps:

> You'll notice that some incidental improvements have been made to `process_artist_credits` related to id/gid loading. Previously, we only checked for and loaded gids to fill in numeric ids where they were missing. Since we now want to fill in the name if empty for *all* artists, we load all artists, regardless of the type of id, and do that. This is in addition to filling in the numeric id as before -- and the new code supports an additional case where an artist submitted with only a numeric id will have a gid added too.

PR #1859, opened 2021-01-13
MBS-10187: "Merge artists" edit is stuck

mwiencek:
Artist merges can get stuck if their merged types and genders conflict: groups can't have a gender.

We now detect this case and drop any conflicting information. This is done as opposed to failing the edit because it's less annoying to the user. ModBot will leave a note explaining what information was lost.

This also handles the case where the target doesn't have a type or gender, but the resulting merged type & gender from the source artists would result in a conflict. In that case we drop both values and leave two notes.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -3,8 +3,13 @@ use Moose;
 use namespace::autoclean;
 
 use Carp;
-use List::MoreUtils qw( uniq );
-use MusicBrainz::Server::Constants qw( $VARTIST_ID $DARTIST_ID $STATUS_OPEN );
+use List::MoreUtils qw( any uniq );
+use MusicBrainz::Server::Constants qw(
+    $VARTIST_ID
+    $DARTIST_ID
+    $STATUS_OPEN
+    $ARTIST_TYPE_GROUP
+);
 use MusicBrainz::Server::Entity::Artist;
 use MusicBrainz::Server::Entity::PartialDate;
 use MusicBrainz::Server::Data::ArtistCredit;
@@ -337,13 +342,22 @@ sub merge
     unless (is_special_artist($new_id)) {
         my $merge_columns = [ qw( area begin_area end_area type ) ];
         my $artist_type = $self->sql->select_single_value('SELECT type FROM artist WHERE id = ?', $new_id);
-        my $group_type = 2;
-        my $orchestra_type = 5;
-        my $choir_type = 6;
-        if (
-            !defined $artist_type ||
-            ($artist_type != $group_type && $artist_type != $orchestra_type && $artist_type != $choir_type)
-        ) {
+
+        my $group_types = $self->sql->select_single_column_array(q{
+            WITH RECURSIVE atp(id) AS (
+                VALUES (?::int)
+                 UNION
+                SELECT artist_type.id
+                  FROM artist_type
+                  JOIN atp ON atp.id = artist_type.parent
+            ) SELECT * FROM atp
+        }, $ARTIST_TYPE_GROUP);
+
+        my $artist_type_is_group =
+            defined $artist_type &&
+            any { $artist_type eq $_ } @$group_types;
+
+        if (!defined $artist_type || !$artist_type_is_group) {
             push @$merge_columns, 'gender';
         }
         merge_table_attributes(
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -17,6 +17,7 @@ use MusicBrainz::Server::Data::Edit;
 use MusicBrainz::Server::Data::Utils qw(
     is_special_artist
     add_partial_date_to_row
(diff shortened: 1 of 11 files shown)


lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -334,18 +340,66 @@ sub merge
...
+
+        my $group_types = $self->sql->select_single_column_array(q{
+            WITH RECURSIVE atp(id) AS (
+                VALUES (?::int)
+                 UNION
+                SELECT artist_type.id
+                  FROM artist_type
+                  JOIN atp ON atp.id = artist_type.parent
+            ) SELECT * FROM atp
+        }, $ARTIST_TYPE_GROUP);
```
mwiencek, line 378: This commit probably isn't necessary, since we hardcode the group types in `group_type_implies_null_gender` anyway. I figured if we ever have to update that, it'll be less trouble since we won't have to fix the code too. But a simple comment reminding to change both might suffice.

t/lib/t/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
```diff
@@ -104,9 +108,18 @@ test 'Fail gracefully if CD TOC has been removed' => sub {
 
     $c->sql->do("DELETE FROM cdtoc WHERE id = 1");
 
-    my $exception = exception { $edit->accept };
-    isa_ok $exception, 'MusicBrainz::Server::Edit::Exceptions::FailedDependency';
-    is $exception->message, localized_note('The CD TOC the track times were being set from has been removed since this edit was entered.');
```
mwiencek, line 109: I had to change this test because with the added version field, `localized_note` could return the message & version keys in any order and this test would randomly fail.

mwiencek:
This PR replaces https://github.com/metabrainz/musicbrainz-server/pull/1513.

mwiencek:
@brainzbot test this please

lib/MusicBrainz/Server/Edit/Artist/Merge.pm
```diff
@@ -51,11 +56,61 @@ sub do_merge
...
+    if ($dropped_columns->{type}) {
+        my $dropped_type =
+            $self->c->model('ArtistType')->get_by_id($dropped_columns->{type});
+
+        $self->c->model('EditNote')->add_note(
+            $self->id => {
+                editor_id => $EDITOR_MODBOT,
+                text => localized_note(
+                    'The “{artist_type}” type has been dropped because ' .
+                    'one of the artists has a gender.',
```
reosarevok, line 78: "The “{artist_type}” type has been dropped because one of the artists has a gender." -> The “{artist_type}” type has not been added to the destination artist because it conflicted with the gender setting on one of the artists being merged" maybe? This otherwise suggests that type and gender just can't ever coexist, to me.

lib/MusicBrainz/Server/Edit/Artist/Merge.pm
```diff
@@ -51,11 +56,61 @@ sub do_merge
...
+    if ($dropped_columns->{gender}) {
+        my $dropped_gender =
+            $self->c->model('Gender')->get_by_id($dropped_columns->{gender});
+
+        $self->c->model('EditNote')->add_note(
+            $self->id => {
+                editor_id => $EDITOR_MODBOT,
+                text => localized_note(
+                    'The “{gender}” gender has been dropped because ' .
+                    'one of the artists is a group.',
```
reosarevok, line 101: 'The “{gender}” gender has been dropped because the destination artist is a group.'? I mean, if a separate artist is a group but the destination artist is already a person, I'd expect to still assign the gender to it. I guess that wording doesn't work for the case from the test where the type is also being dropped though...

yvanzo (approved):
Work nicely with minimum annoyance in both merge directions! :twisted_rightwards_arrows:

lib/MusicBrainz/Server/Edit/Artist/Merge.pm
yvanzo: It suggests that “{artist_type}” type can’t ever coexist with a gender, at least to me.
But I agree with @reosarevok on replacing “been dropped” with “not been added to the destination artist”.

mwiencek: > This otherwise suggests that type and gender just can't ever coexist, to me.

Maybe we ought to explicitly state that "Group artists cannot have a gender" at the end. 'Cause I could also see how "... because it conflicted with the gender setting on one of the artists" might imply it works with some genders but not all.

And I wasn't sure about specifying "artists being merged" rather than just "artists" 'cause I usually only think of source artists as being merged. So what about:

> The “{artist_type}” type has not been added to the destination artist because it conflicted with the gender setting on one of the artists here. Group artists cannot have a gender.

reosarevok: Seems fine to me :)

reosarevok (approved):
LGTM but didn't test

PR #1860, opened 2021-01-13
MBS-11326: Convert edit/notes.tt to React

reosarevok:
### Implement MBS-11326

Tested by visiting edit lists, edit pages and notes-received, including notes by bot editors and notes by users who left a vote (to test the extra info display).


lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -342,6 +340,8 @@ sub TO_JSON {
         entityType => 'editor',
         gravatar => $self->gravatar,
         id => $self->id,
+        is_bot => boolean_to_json($self->is_bot),
+        is_limited => boolean_to_json($self->is_limited),
```
reosarevok, line 344: Is there any reason why these should be only in UnsanitizedEditor? It's public info on the site and it feels wrong to pass around an unsanitized editor just for them.

mwiencek: We now output `privileges`, which you can use for `is_bot`, at least. I don't see a problem with outputting `is_limited`.

lib/MusicBrainz/Server/Edit.pm
```diff
@@ -332,6 +332,7 @@ sub TO_JSON {
         display_data => $self->display_data,
         data => $self->data,
         edit_kind => $self->edit_kind,
+        edit_notes => [map { $_->TO_JSON } $self->all_edit_notes],
```
mwiencek, line 335:
```suggestion
        edit_notes => to_json_array($self->edit_notes),
```

lib/MusicBrainz/Server/Entity/EditNote.pm
```diff
@@ -79,6 +82,18 @@ sub localize {
...
 }
 
+around TO_JSON => sub {
+    my ($orig, $self) = @_;
+
+    my $json = $self->$orig;
+    $json->{editor_id} = $self->editor_id + 0;
+    $json->{editor} = $self->editor->TO_JSON;
+    $json->{post_time} = datetime_to_iso8601($self->post_time);
+    $json->{text} = $self->editor_id == 4 ? $self->localize : ($self->text | format_editnote);
```
mwiencek, line 92: `|` is TT syntax -- that does something different in Perl (bitwise OR).

We might want to name this `formatted_text` too to distinguish it from the raw text attribute
```suggestion
    $json->{formatted_text} = $self->editor_id == 4 ? $self->localize : format_editnote($self->text);
```

root/edit/components/EditNote.js
```diff
@@ -0,0 +1,85 @@
...
+  const anchor = returnNoteAnchor(edit, index);
+  const isOwner = edit.editor_id === editNote.editor_id;
+  const lastRelevantVote = edit.votes.find(vote => (
+    vote.editor_id === editNote.editor_id &&
+    !vote.superseded
+  ));
+  const showVotingIcon = lastRelevantVote && (
+    lastRelevantVote.vote_name === 'Approve' ||
+    lastRelevantVote.vote_name === 'No' ||
+    lastRelevantVote.vote_name === 'Yes'
```
mwiencek, line 62: We should probably import the `EDIT_VOTE_*` constants and compare those to `.vote`, since we might eventually stop outputting `vote_name` and calculate that in JS instead.

mwiencek (changes requested):
notes.tt can be removed now, right? also the note_anchor, votename, and edit_note_display macros

lib/MusicBrainz/Server/Entity/Editor.pm
mwiencek: Make sure to also update the "Unsanitized editor data was found on the client" check, though.

lib/MusicBrainz/Server/Edit.pm
reosarevok: Similar stuff had to happen for the React.embeds, confused me for a bit :) Hopefully it's all working now.

root/edit/components/EditNote.js
mwiencek: I added a `getVoteName` utility function.

mwiencek (approved):

PR #1863, opened 2021-01-14
MBS-11329: Return privileges as a number in internal editor JSON

reosarevok:
### Implement MBS-11329

We return all of them for unsanitized editors. For sanitized editors, we clear all the values except the ones that are public in /privileged (bot, admin and whatnot, but not untrusted, for example).

Navigated around a bit to check that the general flow seems to work fine (edits_disabled users still see the banner, the admin menu has stuff when you're an admin) but didn't test every use by hand.


mwiencek (approved):

yvanzo (approved):
:shipit:

PR #1866, opened 2021-01-15
MBS-11327: Show CAA icons on release search results

reosarevok:
### Implement MBS-11327

We show these on release pages elsewhere, it's probably sensible to show them here as well.


yvanzo (approved):
Work for me: http://localhost:5000/search?query=beau&type=release&limit=25&method=indexed

mwiencek (approved):

PR #1867, opened 2021-01-18
MBS-4782 / MBS-11333: Pass original MBID to AddCoverArt if no release

reosarevok:
### Fix MBS-4782 / MBS-11333

We store the original MBID, so we should use it. The TT used to use the original MBID straight from the data (not passing it to display data) for the filename display, but didn't use it for the actual cover art display, so that was broken. Passing the MBID to the Release->new call ensures both filename *and* image work.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Release/AddCoverArt.pm
+++ b/lib/MusicBrainz/Server/Edit/Release/AddCoverArt.pm
@@ -104,7 +104,11 @@ sub build_display_data {
     my ($self, $loaded) = @_;
 
     my $release = $loaded->{Release}{ $self->data->{entity}{id} } ||
-        Release->new( name => $self->data->{entity}{name} );
+        Release->new(
+            gid => $self->data->{entity}{mbid},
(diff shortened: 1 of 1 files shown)


yvanzo (commented):
The patch looks good to me, but I cannot understand how it resolves MBS-4782, or should this ticket be split as suggested in comments? Would MBS-11333 then be one of these parts?

reosarevok:
It resolves it in that it shows the cover art in the edit. The CAA problem seems to have been fixed with https://tickets.metabrainz.org/browse/CAA-65 AFAICT. Is there a third issue I'm missing? :)

yvanzo:
Then I don’t understand the difference between MBS-4782 and MBS-11333.

reosarevok:
MBS-11333 is about the printed filename. That only broke with the React conversion. MBS-4782 is about the image itself not loading, and has been broken forever.

yvanzo (approved):
Thanks, I updated tickets’ summaries accordingly.

mwiencek (approved):

PR #1868, opened 2021-01-18
MBS-11059: Indicate which release groups have CAA art in listings

reosarevok:
### Implement MBS-11059

Similar to https://github.com/metabrainz/musicbrainz-server/pull/1866

We don't have a great way currently to check whether a RG has cover art without loading it. If loading the cover art for all RGs in, say, a page of search results or an artist index seems like a problem we could always add a separate query that checks existence only.

### DIFF
--- a/lib/MusicBrainz/Server/Data/ReleaseGroup.pm
+++ b/lib/MusicBrainz/Server/Data/ReleaseGroup.pm
@@ -10,6 +10,7 @@ use MusicBrainz::Server::Data::Utils qw(
     hash_to_row
     load_subobjects
     merge_table_attributes
+    object_to_ids
     order_by
     placeholders
 );
@@ -155,6 +156,34 @@ sub has_materialized_artist_release_group_data {
         'SELECT 1 FROM artist_release_group LIMIT 1',
     ) ? 1 : 0;
     return $has_data;
+}
+
+sub load_has_cover_art {
+    my ($self, @release_groups) = @_;
+    my %id_to_rg = object_to_ids(@release_groups);
+    my @ids = keys %id_to_rg;
+
+    return unless @ids; # nothing to do
+
+    my $query = <<~'SQL';
+        SELECT release.release_group
+          FROM release
+          JOIN cover_art_archive.cover_art ca ON ca.release = release.id
+          JOIN cover_art_archive.cover_art_type cat ON cat.id = ca.id
+          JOIN release_meta ON release_meta.id = release.id
+         WHERE release.release_group = any(?)
+           AND cat.type_id = 1
+           AND ca.mime_type != 'application/pdf'
+           AND release_meta.cover_art_presence != 'darkened'
+        SQL
+
+    my $ids_with_art = $self->sql->select_single_column_array($query, \@ids);
+
+    for my $id (@{ $ids_with_art }) {
+        for my $rg (@{ $id_to_rg{$id} }) {
+            $rg->has_cover_art(1);
+        }
+    }
 }
 
 sub pick_status_condition
@@ -716,6 +745,31 @@ sub _hash_to_row
     });
 
     return $row;
+}
+
+=method load_ids
+Load internal IDs for release group objects that only have GIDs.
+=cut
+
+sub load_ids
+{
+    my ($self, @rgs) = @_;
+
+    my @gids = map { $_->gid } @rgs;
+    return () unless @gids;
+
+    my $query = <<~'SQL';
+        SELECT gid, id
+        FROM release_group
+        WHERE gid = any(?)
+        SQL
+
+    my %map = map { $_->[0] => $_->[1] }
+        @{ $self->sql->select_list_of_lists($query, \@gids) };
+
+    for my $rg (@rgs) {
+        $rg->id($map{$rg->gid}) if exists $map{$rg->gid};
+    }
 }
 
 sub load_meta
--- a/root/static/scripts/common/components/EntityLink.js
+++ b/root/static/scripts/common/components/EntityLink.js
@@ -292,20 +292,35 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
     );
   }
 
-  if (showCaaPresence &&
-    entity.entityType === 'release' &&
-    entity.cover_art_presence === 'present') {
-    content = (
-      <>
-        <a href={'/release/' + entity.gid + '/cover-art'}>
+  if (showCaaPresence) {
+    if (entity.entityType === 'release' &&
+        entity.cover_art_presence === 'present') {
+      content = (
+        <>
+          <a href={'/release/' + entity.gid + '/cover-art'}>
+            <span
+              className="caa-icon"
+              title={l('This release has artwork in the Cover Art Archive')}
+            />
+          </a>
+          {content}
+        </>
+      );
+    }
+
+    if (entity.entityType === 'release_group' && entity.hasCoverArt) {
+      content = (
+        <>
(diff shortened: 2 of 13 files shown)


jesus2099:
> We don't have a great way currently to check whether a RG has cover art without loading it.

I didn't understand MBS code to understand what you did but we can check the existence of release group cover art by loading them, indeed, or by using the Cover Art Archive JSON release group web service:

`//coverartarchive.org/release-group/<rg-mbid>`

- It returns a 404 error when there are no front cover arts for this release group: https://coverartarchive.org/release-group/df5fc1e9-d35a-45d0-a571-c62ceac9aa39
- It returns a JSON object where there is at least one: https://coverartarchive.org/release-group/6c9ae3dd-32ad-472c-96be-69d0a3536261
  If you want to use the image URL returned by the web service, like `thumbnails.small`, make sure to replace `http:` by `https:`

```JavaScript
{"images":[{"types":["Front"],"front":true,"back":false,"edit":24740860,"image":"http://coverartarchive.org/release/7dc5edce-ead6-41e4-9c4b-240223c9bab0/5676750691.jpg","comment":"","approved":true,"id":"5676750691","thumbnails":{"large":"http://coverartarchive.org/release/7dc5edce-ead6-41e4-9c4b-240223c9bab0/5676750691-500.jpg","small":"http://coverartarchive.org/release/7dc5edce-ead6-41e4-9c4b-240223c9bab0/5676750691-250.jpg"}}],"release":"http://musicbrainz.org/release/7dc5edce-ead6-41e4-9c4b-240223c9bab0"}
```

I use it in [my script](https://github.com/jesus2099/konami-command/blob/4668e71507e994cdc941db7179e30e8a55a78ea8/mb_FUNKEY-ILLUSTRATED-RECORDS.user.js#L62-L89) only to check CAA availability, but I now use `//coverartarchive.org/release-group/<rg-mbid>/front-250` (don't remember why) to load images and it works good too:

- 404 Error: https://coverartarchive.org/release-group/df5fc1e9-d35a-45d0-a571-c62ceac9aa39/front-250
- Cover: https://coverartarchive.org/release-group/6c9ae3dd-32ad-472c-96be-69d0a3536261/front-250

yvanzo (commented):
@reosarevok : I guess you should use `EXPLAIN ANALYZE` to start evaluating the added load, or @bitmap has another method to recommend?

lib/MusicBrainz/Server/Data/ReleaseGroup.pm
```diff
@@ -718,6 +718,29 @@ sub _hash_to_row
...
+{
+    my ($self, @rgs) = @_;
+
+    my @gids = map { $_->gid } @rgs;
+    return () unless @gids;
+
+    my $query = "
+        SELECT gid, id FROM release_group
+        WHERE gid IN (" . placeholders(@gids) . ")
+    ";
```
mwiencek, line 735: `<<~'SQL'` and maybe be rid of placeholders here?

root/static/scripts/common/components/EntityLink.js
```diff
@@ -292,20 +292,38 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
...
 
-  if (showCaaPresence &&
-    entity.entityType === 'release' &&
-    entity.cover_art_presence === 'present') {
-    content = (
-      <>
-        <a href={'/release/' + entity.gid + '/cover-art'}>
+  if (showCaaPresence) {
+    console.log(entity);
+
```
mwiencek, line 297:
```suggestion
```

root/types/releasegroup.js
```diff
@@ -20,6 +20,7 @@ declare type ReleaseGroupT = $ReadOnly<{
   ...TypeRoleT<ReleaseGroupTypeT>,
   +cover_art?: ArtworkT,
   +firstReleaseDate: string | null,
+  +hasCoverArt?: boolean,
```
mwiencek, line 23: This is always outputted, so you can remove the `?`:
```suggestion
  +hasCoverArt: boolean,
```

root/static/scripts/common/components/EntityLink.js
```diff
@@ -292,20 +292,38 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
...
+              title={l('This release has artwork in the Cover Art Archive')}
+            />
+          </a>
+          {content}
+        </>
+      );
+    }
+
+    if (entity.entityType === 'release_group' &&
+        entity.hasCoverArt /*:: === true */) {
```
mwiencek, line 314: (following the previous change)
```suggestion
    if (entity.entityType === 'release_group' && entity.hasCoverArt) {
```

mwiencek (commented):
Can you see how the `load_for_release_groups` query performs on e.g. a search results page with a lot of release groups? The query seems pretty complex, and loading all of the cover art just to output a boolean doesn't seem ideal, but it could be very fast, in which case it probably doesn't matter. I think at some point it'd even be nice to *display* the artwork on these pages, so we'd have to load them anyway in that case. But even then, I think we should make sure that the query is optimized first.

reosarevok:
A quick test seems to be pretty fast, so it doesn't seem like a huge issue, but you can test a bit and see what you think :)

yvanzo (approved):
@reosarevok: Please provide more details about how you tested it.

I would trust @mwiencek when it comes to query’s optimization.

The rest of the code looks good to me.

reosarevok:
@yvanzo: the speed comment by @mwiencek by opening "a search results page with a lot of release groups" as suggested (and generally navigating around). More generally, by navigating to artist overview pages with lots of RGs and checking, IIRC (it was a while ago which is why it didn't have testing info at the time)

mwiencek:
It seems to take an extra 33 ms to query artwork for 50 release groups IDs on pink (not counting time spent in Perl).  This runs frequently enough that I think we should have a separate query that can complete much faster.  Just looking at requests to artist index pages, there were nearly 500,000 between Nov. 6/7 (so you can do the math on how much extra time is spent serving those pages per day :)).

For example, this takes around 8 ms on pink.
```SQL
SELECT release.release_group
  FROM release
  JOIN cover_art_archive.cover_art ca ON ca.release = release.id
  JOIN cover_art_archive.cover_art_type cat ON cat.id = ca.id
  JOIN release_meta ON release_meta.id = release.id
 WHERE release.release_group = any(?)
   AND cat.type_id = 1
   AND ca.mime_type != 'application/pdf'
   AND release_meta.cover_art_presence != 'darkened';
```

lib/MusicBrainz/Server/Data/ReleaseGroup.pm
```diff
@@ -157,6 +158,32 @@ sub has_materialized_artist_release_group_data {
...
+         WHERE release.release_group = any(?)
+           AND cat.type_id = 1
+           AND ca.mime_type != 'application/pdf'
+           AND release_meta.cover_art_presence != 'darkened'
+        SQL
+
+    my $ids_with_art = $self->sql->select_single_column_array($query, \@ids);
+
+    for my $id (@{ $ids_with_art }) {
+        $id_to_rg{$id}->[0]->has_cover_art(1);
```
mwiencek, line 183: `object_to_ids` groups them in arrays in case you have multiple instances of the same entity for some reason -- looking at the current places we use `load_has_cover_art`, this might happen on series pages?
```suggestion
        for my $rg in (@{ $id_to_rg{$id} }) {
            $rg->has_cover_art(1);
        }
```

mwiencek (approved):

PR #1869, opened 2021-01-19
MBS-11337: Always link to URL entity page and link on URL reports 

reosarevok:
### Implement MBS-11337

On reports about URLs, it makes sense to always link to the URL entity (since that's where you can fix the problems). It still makes sense to have a link to the actual destination though.

At the moment UrlRelationshipList showed URLs in the same way as elsewhere on the site (destination link on click + [info] link to the entity) while the other lists being changed here only linked to the entity but did not allow following the link.

This makes both of these show both the URL link and a link to the URL entity (shown as the MBID), which allows visiting both links as needed without having to click the tiny info link that moves line by line depending on URL length.

### DIFF
--- a/root/report/components/ArtistUrlList.js
+++ b/root/report/components/ArtistUrlList.js
@@ -11,6 +11,7 @@ import * as React from 'react';
 
 import PaginatedResults from '../../components/PaginatedResults';
 import EntityLink from '../../static/scripts/common/components/EntityLink';
+import bracketed from '../../static/scripts/common/utility/bracketed';
 import type {ReportArtistUrlT} from '../types';
 
 type Props = {
@@ -44,10 +45,15 @@ const ArtistUrlList = ({
                 {lastGID === item.url.gid ? null : (
                   <tr className="even" key={item.url.gid}>
                     <td colSpan="2">
-                      <EntityLink
-                        content={item.url.href_url}
-                        entity={item.url}
-                      />
+                      <a href={item.url.name}>
+                        {item.url.name}
+                      </a>
+                      {' '}
+                      {bracketed(
+                        <a href={'/url/' + item.url.gid}>
+                          {item.url.gid}
+                        </a>,
+                      )}
                     </td>
                   </tr>
                 )}
--- a/root/report/components/LabelUrlList.js
+++ b/root/report/components/LabelUrlList.js
@@ -11,6 +11,7 @@ import * as React from 'react';
 
 import PaginatedResults from '../../components/PaginatedResults';
 import EntityLink from '../../static/scripts/common/components/EntityLink';
+import bracketed from '../../static/scripts/common/utility/bracketed';
 import type {ReportLabelUrlT} from '../types';
 
 type Props = {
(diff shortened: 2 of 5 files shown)


yvanzo (approved):
Code looks good to me but did not test, approving that. However, the issue itself doesn’t seem to be an improvement to me. Were you unhappy with consistency?

reosarevok:
Well, there's two things: one is the inconsistency (we showed the URLs in two different ways in different reports). The other is that in a report about a URL entity, the URL entity itself should be prioritized, not the external link, because that is what the report is about. The Discogs reports understood this (but were missing a way to follow the actual link, which can sometimes also be useful). The DeprecatedRels report did not, so the link to go to the actual place where you can do something about the issue was an insignificant amount of the report's visual space :) It's not like we don't already have cases where a URL links to the entity - apart from the Discogs reports, we also do it on the URL page headers (https://musicbrainz.org/url/8c23ed7f-e0dc-4e19-8e5a-9c18212a96a1)

mwiencek:
This seems weird to me since clicking [info] like everywhere else isn't difficult, but anyway: what do you think about adding a separate component for this instead of another flag to `EntityLink`? That component is already a bit messy and over-broad.

reosarevok:
> clicking [info] like everywhere else isn't difficult

Say you want to open the first 10 URL pages in a report fast. Since [info] is at the end of the URL, there's no consistent place for it, so you can't just quickly go down the lines and click, you have to stop, see where it is, go and click each time. With the change, you can click the same place (the start of the line) quickly. Another option, if you feel that's better, would be to put the URL entity link in a second column, but that still wouldn't help with the inconsistent Discogs links :)

yvanzo (changes requested):
Finally, it seems I’m not the only one to think it’s weird, maybe because it is.

> On reports about URLs, it makes sense to prioritize the URL entity (since that's where you can fix the problems). It still makes sense to have a link to the actual destination though.

It makes little sense to have a linked name after an URL that points to a different URL.

Please avoid doing that. Instead, the actual URL could be preceded with a link named after the entity’s MBID. That would probably be more intuitive.

reosarevok:
> It makes little sense to have a link named after an URL that points to a different URL. Please avoid doing that 

Should we also be changing the URL entity headers then? And if so, what to?

yvanzo:
Or possibly two columns “MusicBrainz URL” and “External link”?

reosarevok:
> Or possibly two columns “MusicBrainz URL” and “External link”?

As I answered to @mwiencek earlier: 
> Another option, if you feel that's better, would be to put the URL entity link in a second column, but that still wouldn't help with the inconsistent Discogs links :)

Since those are just one header (and only link to the URL entity right now, see https://musicbrainz.org/report/DiscogsLinksWithMultipleArtists). I always feel it's a bit silly to have a column that's literally the text "See URL entity" or something repeated on every line, but if that feels better to you I can live with that :)

yvanzo:
About naming a link with another link, `DiscogsLinksWithMultipleArtists` seems to be an exception that should probably not last. If you don’t feel comfortable with making two columns, just make one. There is no need to change the column headers either, “URL” is just fine.

mwiencek:
> Say you want to open the first 10 URL pages in a report fast. Since [info] is at the end of the URL, there's no consistent place for it, so you can't just quickly go down the lines and click, you have to stop, see where it is, go and click each time.

Good point. I just thought it was weird to use a URL as link text and not have it go to that same URL, & make it inconsistent with the [info] display everywhere else. But clearly the [info] stuff isn't great for reports.

> I always feel it's a bit silly to have a column that's literally the text "See URL entity" or something repeated on every line

We could use the literal entity URL (`https://musicbrainz.org/url/...`), or just the MBID as the link text. Not opposed to "See URL entity" either, even if it looks dumb. It's not much different from "follow link" except those don't vertically align.

yvanzo (approved):
Works for me (after deprecating BBC relationships in sample data). Thanks!

mwiencek (approved):

PR #1870, opened 2021-01-19
MBS-11336: Don't show ended rels in DeprecatedRelationshipURL

reosarevok:
### Implement MBS-11336

This report is now full of BBC Music Page relationships, which are set to ended. As such, they have already been dealt with, so there's no point showing those in the report. In general, "ended" URL relationships should be considered effectively removed for most cases (since we don't show them on the sidebar and we grey them out elsewhere).


yvanzo (approved):
Makes sense. Tested from http://localhost:5000/report/DeprecatedRelationshipURL

mwiencek (approved):

PR #1871, opened 2021-01-19
MBS-11335: Report for reused Wikidata items

reosarevok:
### Implement MBS-11335

Chose to implement one big report rather than one per entity because a common use case of this will be "same link for type: single RG and work", because of how WD works or used to work.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1787 and https://github.com/metabrainz/musicbrainz-server/pull/1869


mwiencek (commented):
https://github.com/metabrainz/musicbrainz-server/pull/1869 uses a different implementation now, so this should be rebased (after merging that).

You can also look into using `<<~EOSQL` in the report SQL if that interests you.

reosarevok:
Changed this to use `EOSQL` and a similar `map` approach as in https://github.com/metabrainz/musicbrainz-server/pull/2096, too.

mwiencek (approved):

PR #1872, opened 2021-01-21
MBS-11330: Split root/types.js into smaller files

reosarevok:
### Implement MBS-11330

This makes a lot of small files, but I'd expect it to be better to keep types together that way. That said, if we feel it's too many small files, I'd be happy to make changes to the split if a better way of putting the types together is suggested :)


yvanzo (commented):
I had no issue with using grep but I can see the benefit of having (un)sanitized variants grouped with their original types.

root/types/metabrainz.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * The definitions in this file are intended to model the output of the
+ * TO_JSON methods under lib/MusicBrainz/Server/Entity/, those are precisely
+ * how data is serialized for us.
+ */
+
+declare type BlogEntryT = {
+  +title: string,
+  +url: string,
+};
```
mwiencek, line 21: I'd just move this to root/main/index.js

root/types/metabrainz.js
```diff
@@ -0,0 +1,33 @@
...
+  +author: CritiqueBrainzUserT,
+  +body: string,
+  +created: string,
+  +id: string,
+};
+
+declare type CritiqueBrainzUserT = {
+  +id: string,
+  +name: string,
+};
```
mwiencek, line 33: I'd move these to root/static/scripts/common/components/CritiqueBrainzReview.js (exported). Only root/release_group/ReleaseGroupIndex.js needs to import CritiqueBrainzReviewT.

root/types/tablecomponents.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * The definitions in this file are intended to model the output of the
+ * TO_JSON methods under lib/MusicBrainz/Server/Entity/, those are precisely
+ * how data is serialized for us.
+ */
+
+declare type AppearancesT<T> = {
+  +hits: number,
+  +results: $ReadOnlyArray<T>,
+};
```
mwiencek, line 21: this might make more sense in entity.js since it applies to several entities and I wouldn't associate it with a table (it's used by the autocomplete)

root/types/editor.js
```diff
@@ -0,0 +1,101 @@
...
+  +gravatar: string,
+  +name: string,
+};
+
+declare type FluencyT =
+  | 'basic'
+  | 'intermediate'
+  | 'advanced'
+  | 'native'
+  ;
```
mwiencek, line 52: I think you recently fixed these so let's not re-add them :)
```suggestion
  | 'native';
```

root/types/tablecomponents.js
```diff
@@ -0,0 +1,46 @@
...
+declare type PagerT = {
+  +current_page: number,
+  +first_page: 1,
+  +last_page: number,
+  +next_page: number | null,
+  +previous_page: number | null,
+  +total_entries: number,
+};
+
+declare type SearchResultT<T> = {
```
mwiencek, line 36: should probably go in root/search/types.js and just imported in the few places it's used. though that just leaves PagerT in this file which is a bit weird

root/types/functions.js
```diff
@@ -0,0 +1,64 @@
...
+
+declare type AnchorProps = {
+  +className?: string,
+  +href: string,
+  +key?: number | string,
+  +target?: '_blank',
+  +title?: string,
+};
+
+declare type Expand2ReactInput = VarSubstArg | AnchorProps;
```
mwiencek, line 26: this file seems a bit nebulous to me (most of these aren't functions)

Maybe move KnockoutObservable/KnockoutObservable into a knockout.js and the rest into an i18n.js? StrOrNum is the only one that's really general...kind of a weird case, like PagerT.

mwiencek (changes requested):
Some of the files could probably be combined - e.g. workattributes into work, formcomponents into form, medium+track into release, vote into edit. (I guess there's a point where we split them too much and the files become annoying to switch between.)

root/types/functions.js
reosarevok: Moved PagerT and StrOrNum to a "misc" file for now, but yeah, I'm also not sure what to do.

mwiencek:
Rebased this and synced all the recent changes to root/types.js to the respective new files.

I removed the comment that was copied to the top of every file. The bit about keeping things in alphabetical order isn't so important since most files only have 1-3 types now. It's also not a hard rule, since I do think we should keep types like Foo, WritableFoo, and ReadOnlyFoo together. The other bit about the types modeling TO_JSON methods under lib/MusicBrainz/Server/Entity/ was just false in a lot of cases. I did add some inline comments saying which packages particular types came from.

mwiencek (approved):

PR #1873, opened 2021-01-21
MBS-11340: Allow more characters on Spotify user URLs

reosarevok:
### Fix MBS-11340

As per the URLs added to the tests, there's a lot more characters allowed in these URLs than just what we were limiting them to.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2866,12 +2866,12 @@ const CLEANUPS = {
           case LINK_TYPES.songkick.artist:
             return {result: prefix === 'artists'};
           case LINK_TYPES.songkick.event:
-            return  {result: prefix === 'concerts' || prefix === 'festivals'};
+            return {result: prefix === 'concerts' || prefix === 'festivals'};
           case LINK_TYPES.songkick.place:
-            return  {result: prefix === 'venues'};
+            return {result: prefix === 'venues'};
         }
       }
-      return  {result: false};
+      return {result: false};
     },
   },
   'soundcloud': {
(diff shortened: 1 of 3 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2941,11 +2941,11 @@ const CLEANUPS = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?(spotify\\.com)/user', 'i')],
     type: LINK_TYPES.socialnetwork,
     clean: function (url) {
-      url = url.replace(/^(?:https?:\/\/)?(?:play|open)\.spotify\.com\/user\/([a-zA-Z0-9_-]+)\/?(?:[?#].*)?$/, 'https://open.spotify.com/user/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:play|open)\.spotify\.com\/user\/([^\/?#]+)\/?(?:[?#].*)?$/, 'https://open.spotify.com/user/$1');
       return url;
     },
     validate: function (url) {
-      return {result: /^https:\/\/open\.spotify\.com\/user\/[a-zA-Z0-9_-]+$/.test(url)};
+      return {result: /^https:\/\/open\.spotify\.com\/user\/[^\/?#]+$/.test(url)};
```
yvanzo, line 2948: There is a slightly different pattern for user’s identifier in [`lib/MusicBrainz/Server/Entity/URL/Spotify.pm`](https://github.com/metabrainz/musicbrainz-server/blob/v-2021-01-11/lib/MusicBrainz/Server/Entity/URL/Spotify.pm#L13). It should probably be streamlined.

yvanzo (approved):
Works fine with NiN. Just found a tiny difference in user patterns.

root/static/scripts/edit/URLCleanup.js
reosarevok: Would you suggest adding & to this one or removing it from the other?

yvanzo: Just remove it please. `&` is separator between query parameters, thus it should normally be preceded with a `?` introducing the first query parameter, but some websites use `&` in place of `?`. That might explain the `&` in the pattern added in commit 4d92b8e1d7bfa30c91617f4f4b3525d31fd3f2b1. However there is no such URL for Spotify in the DB, so it is unlikely it exists in the wild.

mwiencek (approved):

PR #1874, opened 2021-01-22
MBS-8700: List genres in JSON-LD

reosarevok:
### Implement MBS-8700

This uses only the genres on top_tags (so, the ones that would be displayed on the sidebar). This sounds sensible enough to me, since if someone has added 100 genres to an artist we probably don't want to show them all.

https://schema.org/genre specifically allows only Text or URL for genre, so I went with URL to at least be able to pass an MBID for it. Ideally we would serialize a genre object with ID and name, but that doesn't seem to be supported.

I've never knowingly touched JSON-LD stuff before, so do let me know if something here seems dumb :)

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/Genre.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/Genre.pm
@@ -1,0 +1,33 @@
+package MusicBrainz::Server::WebService::Serializer::JSON::LD::Role::Genre;
+use Moose::Role;
+use MusicBrainz::Server::WebService::Serializer::JSON::LD::Utils qw( list_or_single serialize_entity );
+
+around serialize => sub {
+    my ($orig, $self, $entity, $inc, $stash, $toplevel) = @_;
+    my $ret = $self->$orig($entity, $inc, $stash, $toplevel);
+
+    my $tags = $stash->store($entity)->{top_tags};
+    my @genres = map {
+        my $genre = $_->{tag}->genre;
+        $genre ? (DBDefs->CANONICAL_SERVER . '/genre/' . $genre->gid) : ()
+    } @$tags;
+
+    if (@genres) {
+        $ret->{genre} = list_or_single(@genres);
+    }
+
+    return $ret;
+};
+
+no Moose::Role;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2020 MetaBrainz Foundation
+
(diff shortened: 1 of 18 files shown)


t/lib/t/MusicBrainz/Server/Controller/Artist/Show.pm
```diff
@@ -201,6 +201,33 @@ EOSQL
...
+    INSERT INTO artist_tag (artist, count, last_updated, tag) VALUES (1, 3, '2011-01-18 15:21:33.71184+00', 30);
+
+EOSQL
+
+    $mech->get_ok('/artist/dcb48a49-b17d-49b9-aee5-4f168d8004d9');
+    page_test_jsonld $mech => {
+        'name' => 'Group',
+        '@type' => 'MusicGroup',
+        '@id' => 'https://musicbrainz.org/artist/dcb48a49-b17d-49b9-aee5-4f168d8004d9',
+        'genre' => 'https://musicbrainz.org/genre/1b50083b-1afa-4778-82c8-548b309af783',
```
mwiencek, line 226: A separate test for multiple genres would be cool too

lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/Genre.pm
```diff
@@ -0,0 +1,30 @@
...
+around serialize => sub {
+    my ($orig, $self, $entity, $inc, $stash, $toplevel) = @_;
+    my $ret = $self->$orig($entity, $inc, $stash, $toplevel);
+
+    my $tags = $stash->store($entity)->{top_tags};
+    my @genres = map { $_->{tag}->genre } grep { $_->{tag}->genre } @{ $tags };
+
+    if (@genres) {
+      $ret->{genre} = list_or_single(map { DBDefs->CANONICAL_SERVER . '/genre/' . $_->gid } @genres);
+    }
```
mwiencek, line 14: I'd reduce this to a single map instead of two maps + a grep
```suggestion
    my @genres = map {
        my $genre = $_->{tag}->genre;
        $genre ? (DBDefs->CANONICAL_SERVER . '/genre/' . $genre->gid) : ()
    } @$tags;

    if (@genres) {
      $ret->{genre} = list_or_single(@genres);
    }
```

mwiencek: (Also, looks like the indentation inside the `if` is off, should be 4 spaces instead of two.)

mwiencek (approved):

PR #1875, opened 2021-01-23
MBS-11346: Lower duration threshold on dubious duration discID report

loujine:
Implement MBS-11346

The report (discIDs durations > 100 minutes) is now empty, Setting to 88 minutes to get more results and avoid too many false positive.

### DIFF
--- a/lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
+++ b/lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
@@ -20,8 +20,12 @@ sub query {
             JOIN medium ON medium_cdtoc.medium = medium.id
             JOIN medium_format ON medium.format = medium_format.id
         WHERE
-            leadout_offset > 75 * 60 * 88 -- cutoff 88 minutes
-            AND medium_format.name != 'CD-R'
+            -- cutoff 88 minutes
+            (leadout_offset > 75 * 60 * 88
+                AND medium_format.name != 'CD-R')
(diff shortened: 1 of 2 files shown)


yvanzo (commented):
Thanks @loujine!

90 minutes CD-R and even 100 minutes CD-R are quite common. Should the query be updated to exclude CD-R such as https://musicbrainz.org/cdtoc/ttC8YktpxuooF3AKLYsgp7Pt5A0- too? Maybe some other formats are to be excluded too?

loujine:
@yvanzo : I'll still add another commit excluding CD-R. I don't know if you looked at https://loujine.github.io/musicbrainz-sql/2021-01-13/discid_above_90_minutes.html - FYI the CD-R you mention seems to be the only one > 86 minutes

I am realizing at the same time:
- I should have a different limit for 8cm mini-CD, probably ~25 minutes (we have 29 discids above this limit in the DB)
- there are a few disc IDs on vinyl, cassette... in the database. It might deserve another report

yvanzo (approved):
Even better!

reosarevok (changes requested):
Please use format IDs, not names, for matching :) Names can change. Also, maybe leave a comment for the 8cm cutoff too?

Other than that this looks good :)

loujine:
@reosarevok : should be fixed

Thanks for the reviews!

reosarevok (approved):
Assuming those are the right IDs :)

PR #1879, opened 2021-01-26
MBS-6502: Surround catno search with quotes

reosarevok:
### Fix MBS-6502

CRD 3443 is a valid catalog number we should support, and in fact our search supports it. But by sending 'catno:CRD 3443' we actually search 3443 as a title!

Since there's one use case that quotes break (searching for a regex, where the string must start and end with slashes), we avoid quoting any query that starts and ends with slashes.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/OtherLookup.pm
+++ b/lib/MusicBrainz/Server/Controller/OtherLookup.pm
@@ -9,6 +9,29 @@ extends 'MusicBrainz::Server::Controller';
 
 use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 
+sub build_term {
+    my ($self, $value) = @_;
+
+    # We want to turn the given value into a valid Lucene term so
+    # the search server understands it properly.
+    # As such, we quote it to make it a Lucene phrase,
+    # except if it is a regular expression or the user already entered
+    # a Lucene phrase (i.e. the value is already in quotes).
+    # See https://lucene.apache.org/core/7_7_2/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Terms
+    my $is_regex = $value =~ /^\/.*\/$/;
+    my $is_quoted = $value =~ /^".*"$/;
+
+    # We escape any quotes inside the actual value so that they are
+    # still searched for when we turn it into a phrase.
+    my $escaped_value = $value =~ s/"/\\"/gr;
+
+    my $term = ($is_regex || $is_quoted)
+        ? $value
+        : "\"$escaped_value\"";
+
+    return $term;
+}
+
 sub lookup_handler {
     my ($name, $code) = @_;
 
@@ -43,7 +66,7 @@ lookup_handler 'catno' => sub {
 
     $c->response->redirect(
         $c->uri_for_action('/search/search', {
-            query => 'catno:' . $cat_no,
+            query => 'catno:' . $self->build_term($cat_no),
             type => 'release',
             advanced => '1',
         }));
--- a/lib/MusicBrainz/Server/Controller/OtherLookup.pm
+++ b/lib/MusicBrainz/Server/Controller/OtherLookup.pm
@@ -27,7 +27,7 @@ sub build_term {
 
     my $term = ($is_regex || $is_quoted)
         ? $value
-        : "\"$escaped_value\"";
+        : qq("$escaped_value");
 
     return $term;
 }


kellnerd:
The only drawback is that this would [break](https://musicbrainz.org/search?query=catno%3A%22%2FCRD34%5B0-9%5D%7B2%7D%2F%22&type=release&method=advanced) [RegEx searches](https://musicbrainz.org/search?query=catno%3A%2FCRD34%5B0-9%5D%7B2%7D%2F&type=release&method=advanced) which were submitted via one of the "other lookups" forms. We could either ignore this rare use case or check whether the input is a regular expression (wrapped in forward slashes) and omit the quotes in this case.

By the way, does the search automatically strip spaces (and other separators) from catalog numbers etc. and I am not supposed to include them in regular search expressions for these fields? Or have I misunderstood something? I am asking because I am curious how a RegEx that includes spaces can be entered into the advanced search when both spaces and enclosing quotes are breaking the search.

reosarevok:
> By the way, does the search automatically strip spaces (and other separators) from catalog numbers etc. and I am not supposed to include them in regular search expressions for these fields? Or have I misunderstood something? I am asking because I am curious how a RegEx that includes spaces can be entered into the advanced search when both spaces and enclosing quotes are breaking the search.

No, it doesn't do anything else at the moment than trimming starting and ending space and appending whatever you enter to "catno:" AFAICT.

> The only drawback is that this would break RegEx searches 

I didn't even remember this is a thing. I should use it more for my editing! That said, I expect most people who know enough to do those won't go through this method that requires navigating to a big search page in the first place so it might not matter?

jesus2099:
I think @kellnerd has a good idea:

> _check whether the input is a regular expression (**wrapped in forward slashes**) and omit the quotes in this case_

I don't know MBS language, but something like (JavaScript):

```JavaScript
query = 'catno:' + (cat_no.match(/^\s*\/.*\/\s*$/) ? cat_no.trim() : '"' + cat_no.trim() + '"');
```

reosarevok:
Made a change to not touch regexes then :) We might want to also allow using regexes for IPI and ISNI, if those are supported by the search - right now anything that is not a valid IPI or ISNI itself gets rejected.

mwiencek:
> We might want to also allow using regexes for IPI and ISNI

Sorry if I missed something, but why quote those? AFAICT we already strip spaces from the input for those.

For catalog numbers I just worry about cases where the catno has a ", like for these https://musicbrainz.org/label/d8520375-c2c9-46c8-8655-4509b314e6a5

If we're always injecting quotes, let's escape any existing quotes the user entered. (If they were already quoting things this breaks things for them either way, but I think the PR sets a better default behavior.)

jesus2099:
I think we probably remove spaces inside the catno argument.
But if we search for catno:ABC 123, 123 will not be part of the catno argument.
We must type catno:"ABC 123" then yes internally, it might search ABC123. But the quotes were required. ☺️

reosarevok:
> Sorry if I missed something, but why quote those?

For consistency, really, but I guess there's no point.

Made changes to escape quotes and to avoid re-quoting quoted stuff.

lib/MusicBrainz/Server/Controller/OtherLookup.pm
```diff
@@ -6,6 +6,26 @@ use Moose::Util qw( find_meta );
...
+
+    # We want to quote the query so Lucene understands it properly,
+    # except if it is a regular expression or the user already entered
+    # a quoted value.
+    my $is_regex = $value =~ /^\/.*\/$/;
+    my $is_quoted = $value =~ /^".*"$/;
+
+    # We escape any quotes inside the actual string so that they are
+    # still searched for when we put the whole thing in quotes.
+    my $escaped_value = $value =~ s/"/\\"/gr;
```
mwiencek, line 20: You'd only want to do this unless `$is_quoted` though, right? Otherwise you'd escape the outer quotes there.

mwiencek (approved):

lib/MusicBrainz/Server/Controller/OtherLookup.pm
```diff
@@ -6,6 +6,26 @@ use Moose::Util qw( find_meta );
 use MusicBrainz::Server::Translation qw( l );
 use MusicBrainz::Server::Constants qw( entities_with );
 
+sub quoted_query {
```
yvanzo, line 9: Please use the term _phrase_ as in Lucene documentation: https://lucene.apache.org/core/7_7_2/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Terms

yvanzo (changes requested):

yvanzo:
Naming was not the only issue, we found out it can be difficult to know how to enter a regular expression with a space in it.
It should probably be documented by adding examples next to that lookup field.

reosarevok:
@yvanzo: do you have any suggestions on how to document this further, if you still feel it needs to be done? (I feel documenting regexes might be outside the scope of this patch, but I'm happy to add it anyway if you have some good suggestions)

mwiencek:
> Naming was not the only issue, we found out it can be difficult to know how to enter a regular expression with a space in it.
It should probably be documented by adding examples next to that lookup field.

Is this an issue introduced by this patch (or exacerbated by it)?

mwiencek:
> Naming was not the only issue, we found out it can be difficult to know how to enter a regular expression with a space in it.
It should probably be documented by adding examples next to that lookup field.

I couldn't actually find any official documentation for Lucene's regular expression query syntax except for https://lucene.apache.org/core/7_1_0/core/org/apache/lucene/util/automaton/RegExp.html, which is not very user friendly.

I'm not sure what the issue with entering spaces was, but as far as I can tell, you should generally not enter them. For example, the following query returns results with spaces after "EK" without specifying any space in the query:

https://musicbrainz.org/search?query=catno%3A%2FEK%5B0-9%5D%2B%2F&type=release&limit=25&method=advanced

This also works if there are no spaces in the catno:

https://musicbrainz.org/search?query=catno%3A%2FEK%5B0-9%5D%2B%2F+AND+reid%3A7564c0c1-276d-48c9-b723-a0496377cf60&type=release&limit=25&method=advanced

(Finding a release named "Results" as the first example was entirely coincidental. 😄 It was the first release returned from `SELECT * FROM release_label WHERE catalog_number ~ '^EK[0-9]+' LIMIT 1`.)

I also found a way to explicitly specify spaces with `\s*`. (Note: `\s` and `\s+` do not work.) Confusingly, it returns more results than the first query:

https://musicbrainz.org/search?query=catno%3A%2FEK%5Cs*%5B0-9%5D%2B%2F&type=release&limit=25&method=advanced

But I didn't sift through them to identify what the differences are.

mwiencek:
> Rebased this. While maybe we want to document regex searches better in a separate PR, I see no reason why this specific PR should be blocked because of it since it seems like a straightforward improvement as-is.

I think you forgot to push the rebase, so I did that now (had to do it anyway since `yarn` was erroring otherwise), and pushed a small syntax change. Tested it again locally and it works great. Yes, let's finally merge this.

mwiencek (approved):

reosarevok:
Huh, I thought I had pushed it the second time after originally forgetting. My brain is moldy.

PR #1880, opened 2021-01-26
MBS-11351: Sort instrument reports by name before type

reosarevok:
### Implement MBS-11351

CatQuest requested this and IMO it gives a better overview of what's missing - in any case, they're the main user of this, so it makes sense to make it easier for them.


yvanzo (approved):
LGTM

mwiencek (approved):

PR #1881, opened 2021-01-26
Prevent linking release to Wikidata/Wikipedia

yvanzo:
# Problem

Only “discography entry”, “license”, and “show notes” relationship types used to be blocked at release level because these were the most common. It doesn’t prevent adding Wikidata/Wikipedia using other free relationship types such as streaming and so on.

# Solution

This patch prevents Wikidata/Wikipedia links from being added to release using any relationship type. Would it be allowed to in the future, a dedicated relationship type should be created in the meantime.

# How to test

Edit a release, try adding links such as `https://www.wikidata.org/wiki/Special:EntityPage/Q14005` and  `https://en.wikipedia.org/wiki/MusicBrainz`. Selecting any relationship type should throw an error.

Edit an entity of any other type, adding the same links should auto-select the appropriate relationship type and be accepted.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3561,76 +3561,49 @@ Object.values(LINK_TYPES).forEach(function (linkType) {
   });
 });
 
-// avoid Wikipedia/Wikidata being added as release-level discography entry
-const discographyRule = validationRules[LINK_TYPES.discographyentry.release];
-validationRules[LINK_TYPES.discographyentry.release] = function (url) {
-  if (/^(https?:\/\/)?([^.\/]+\.)?wikipedia\.org\//.test(url)) {
-    return {
-      error: l(
-        `Wikipedia is not a discography entry. Please add this Wikipedia link
-         to the release group instead.`,
-      ),
-      result: false,
-    };
-  }
-  if (/^(https?:\/\/)?([^.\/]+\.)?wikidata\.org\//.test(url)) {
-    return {
-      error: l(
-        `Wikidata is not a discography entry. Please add this Wikidata link
-         to the release group instead.`,
-      ),
-      result: false,
-    };
-  }
-  return discographyRule(url);
-};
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3561,76 +3561,47 @@ Object.values(LINK_TYPES).forEach(function (linkType) {
...
+        ),
+        result: false,
+      };
+    }
+    if (/^(https?:\/\/)?([^.\/]+\.)?wikidata\.org\//.test(url)) {
+      return {
+        error: l(
+          `Wikidata has no dedicated item per release.
+           Please add this Wikidata link to the release group instead,
+           if appropriate.`,
```
reosarevok, line 3597: I kinda feel that "Wikidata has no dedicated item per release" is a bit confusing to non-experts. What about something like "Wikidata normally has no entries that map to specific releases, so adding Wikidata links to a release is currently blocked. Please add this Wikidata link to the release group instead, if appropriate." (and same for Wikipedia).

reosarevok (approved):
Works fine, but see if you agree with my wording comment :)

root/static/scripts/edit/URLCleanup.js
yvanzo: Followed your suggestion in commit https://github.com/metabrainz/musicbrainz-server/pull/1881/commits/633930f0ea831022d5a8782ca9f841d0f26e0d25.

reosarevok: I specified "normally" because I'm sure there are a few rare exceptions, but we block them anyway because if we allowed them then 99% of the time people would add wrong stuff so it's better to miss out on them. Maybe not a big deal though.

yvanzo: Sorry, that was not intended, I just misread it, should be fixed now, see commit https://github.com/metabrainz/musicbrainz-server/commit/8fdde06e862df1aaeef7a5991515e464b9423d87.

reosarevok (approved):

mwiencek (approved):

PR #1882, opened 2021-01-26
MBS-11123: Improve the 'Added entities' table in the editor profile

loujine:
Implements MBS-11123

* use only applied edits
* add missing entity types (intruments, areas, RG...)
* add abbr tag to warn users the results are not calculated 'live' since we cache them in redis

# Action

Needs to be tested on the live database as the PR for MBS-9769 was, to make sure there are no performances issues (I imagine not because usually editors have much less of those edits as the ones already implemented)


mwiencek (approved):
I compared the new query with the old one (using reosarevok's account in the prod DB to test) and the execution time was identical or had a negligible difference. The query plans are the same except for the minor addition of a status = 2 filter on the existing `edit_idx_editor_id_desc` index scan.

root/user/UserProfile.js
```diff
@@ -661,12 +672,17 @@ const UserProfileStatistics = ({
...
+                             'need 24 hours to appear')}
+              >
+                {exp.l('Added entities')}
+              </abbr>
             </th>
           </tr>
         </thead>
         <tbody>
           {Object.keys(ADDED_ENTITIES_TYPES)
+            .filter(type => (type !== 'area' || user.is_location_editor))
```
kellnerd, line 685: The same should probably be done for instruments?

reosarevok: I wouldn't necessarily filter these out in that it's possible that a user used to be a location or relationship editor (hence added areas or instruments) and no longer is. But my only other suggestion would be to hide those if the number is 0. To be fair, we could just not show any line where the number is 0.

loujine: There's no INSTRUMENT_EDITOR_FLAG right? Anyway, hiding empty rows makes more sense, not displaying entity types that people are not allowed to create (instruments) or maybe not interested in. And this will reduce the height of this table so that it stays closer to the other tables on the page

reosarevok (changes requested):

reosarevok (approved):

root/user/UserProfile.js
```diff
@@ -682,7 +682,7 @@ const UserProfileStatistics = ({
         </thead>
         <tbody>
           {Object.keys(ADDED_ENTITIES_TYPES)
-            .filter(type => (type !== 'area' || user.is_location_editor))
+            .filter(type => (addedEntities[type] !== 0))
```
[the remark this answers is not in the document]
reosarevok: Amended this in an extra commit

mwiencek (approved):

PR #1883, opened 2021-01-29
MBS-11353: Allow empty row in multiple select

reosarevok:
### Fix MBS-11353

Without this, cover art type can only be unset in an unintuitive way involving control-clicks. We used to have this, so we should readd it.

### DIFF
--- a/root/components/SelectField.js
+++ b/root/components/SelectField.js
@@ -50,6 +50,7 @@ type SelectElementProps = {
 };
 
 type SharedFieldProps = {
+  +allowEmpty?: boolean,
   +className?: string,
   +disabled?: boolean,
   +onChange?: (event: SyntheticEvent<HTMLSelectElement>) => void,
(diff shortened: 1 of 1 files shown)


jesus2099:
Are you sure this is needed and a regression?
See forum topic from MBS-11353.

reosarevok:
Heh. Now I'm not sure. It's hard to test this because I don't have cover art set up locally, but you might be right the (horribly unintuitive) option you mention is how it used to be earlier too and this doesn't actually help :/ In which case I should drop this and make the ticket into "Add an intuitive way of blanking a multiselect"

jesus2099:
I remember some people don't like those multi-select: https://community.metabrainz.org/t/edit-coverart-types-to-have-multiple-types/437363/6?u=jesus2099
I will link this ticket as related: https://tickets.metabrainz.org/browse/MBS-10307

reosarevok:
Seems the TT form *did* have a blank line (I checked an old unrelated branch which still had that) so we should, in any case, bring this back to make it easier to deselect it.

jesus2099:
Yes, then. 
Sorry for my doubts. ☺️

mwiencek (approved):

yvanzo (approved):
This change doesn’t other multi-select fields because:

From @reosarevok:
> MultipleSelectField is only used in CoverArtFields right now

And other multi-select fields not yet converted to React (attributes in relationship editor, and secondary types in release group editor) allow for empty select already.

:shipit:

PR #1884, opened 2021-01-30
Simplify bottom links to avoid translation parameters

reosarevok:
As done to ArtistRecordings in commit 5c826bb


yvanzo (approved):
To do: hand-fix beta translations in Transifex.

mwiencek (approved):

PR #1885, opened 2021-02-01
MBS-11354: Don't show arrows unless we want ordering

reosarevok:
### Fix MBS-11354

This was changed with 72ce47118a42097872d4a15bb2407e1c9d14c9b9 in a way that caused the arrows to always show for orderable relationship types. It used to only show once there were at least two relationships and the user selected that they wanted them to be ordered (hasOrdering). The old way seems more sensible to me.


mwiencek (approved):

yvanzo (approved):
Example from sample data: http://localhost:5000/recording/70d25e69-b939-4898-a97d-780ac978afbd/edit

PR #1886, opened 2021-02-02
Link directly to the IA for CAA thumbnails

mwiencek:
A recent incident where our CAA redirect service became flooded with requests gave me reason to look into this sort of optimization. There should be no reason we can't bypass the CAA-redirect (now artwork-redirect) entirely for thumbnail `` links and reduce load on this service.

The only question in my mind would be whether CORS still works, but I see no reason it would stop working unless they're conditionally outputting the `Access-Control-Allow-Origin` header.


reosarevok (approved):
Seems to work fine. The only possible issue I can find is whether we care about having the direct IA URL in the source page (so that a scraper could skip the CAA redirect). But it's not like we are monetizing the redirect or anything, so I can't see what would be a problem with that.

lib/MusicBrainz/Server/Entity/Artwork.pm
```diff
@@ -76,6 +76,19 @@ sub _urlprefix
     return join('/', DBDefs->COVER_ART_ARCHIVE_DOWNLOAD_PREFIX, 'release', $self->release->gid, $self->id)
 }
 
+sub _iaurlprefix {
```
yvanzo, line 79: That would be more readable:
```suggestion
sub _ia_url_prefix {
```

yvanzo (approved):
Successfully tested with:
* http://localhost:5000/release-group/7c4cab8d-dead-3870-b501-93c90fd0a580
* http://localhost:5000/release/dceb6a01-3431-36af-b2e1-6462193bd67c
* http://localhost:5000/release/dceb6a01-3431-36af-b2e1-6462193bd67c/cover-art

Just one comment about identifier readability.

PR #1887, opened 2021-02-02
MBS-11357: Allow linking RateYourMusic to release groups

reosarevok:
### Fix MBS-11357

Forgot to allow for RGs in 9f56fea514cf49d2a276c5338c1961044d67d670

The difference between releases and RG links can't be determined from the URL as far as I can tell. Some include a dot (like in the test example) but some just append "_f1" or whatever to the RG URL.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -3004,10 +3004,17 @@ const testData = [
   },
   {
                      input_url: 'https://rateyourmusic.com/release/single/tori_amos/a_sorta_fairytale/',
-             input_entity_type: 'release',
+             input_entity_type: 'release_group',
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://rateyourmusic.com/release/single/tori_amos/a_sorta_fairytale/',
-        only_valid_entity_types: ['release'],
+        only_valid_entity_types: ['release', 'release_group'],
+  },
+  {
+                     input_url: 'https://rateyourmusic.com/release/single/tori_amos/a_sorta_fairytale.p/',
+             input_entity_type: 'release',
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -3004,10 +3004,17 @@ const testData = [
...
-        only_valid_entity_types: ['release'],
+        only_valid_entity_types: ['release', 'release_group'],
+  },
+  {
+                     input_url: 'https://rateyourmusic.com/release/single/tori_amos/a_sorta_fairytale.p/',
+             input_entity_type: 'release',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://rateyourmusic.com/release/single/tori_amos/a_sorta_fairytale.p/',
+        only_valid_entity_types: ['release', 'release_group'],
   },
```
yvanzo, line 3018: It would be nice if release and release group could be distinguished from RYM URLs. I guess it’s not possible?

yvanzo (changes requested):
Please add an example for `recording` as it is (legitimately) blocked as well.

All other [normal entity types](https://musicbrainz.org/doc/MusicBrainz_Entity#Normal_Entities) (but URL) are tested already.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Nope :( See PR description

yvanzo (approved):
My bad, `recording` are already tested to be rejected as they do not appear in any `only_valid_entity_types` array, and RYM doesn’t provide URL for neither tracks nor recordings.

yvanzo:
Merging for hotfix release.

PR #1888, opened 2021-02-02
MBS-11358: Allow alias sort name to be blank

reosarevok:
### Fix MBS-11358

The backend already copies the name to the sort name if left blank, so this was never intended to be marked required here. Making required an optional boolean since we will want it to remain required for artist editing.

### DIFF
--- a/root/static/scripts/alias/AliasEditForm.js
+++ b/root/static/scripts/alias/AliasEditForm.js
@@ -273,10 +273,7 @@ const AliasEditForm = ({
     dispatch({action, type: 'update-date-range'});
   }, [dispatch]);
 
-  const missingRequired = state.isTypeSearchHint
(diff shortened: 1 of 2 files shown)


reosarevok:
By the way, I noticed another change while checking for how this used to be: the "This is the primary alias for this locale" checkbox is now disabled but shown when locale is blank, while it used to be hidden. I don't think that's a change for the worse (seems about the same to me) so I didn't change that back, but I can if you want :)

yvanzo (approved):
Works for me.

Tested by adding an alias from http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/aliases

yvanzo:
@brainzbot retest this please

yvanzo:
Merging for hotfix release.

PR #1890, opened 2021-02-03
MBS-11360: Work around unconsistent name for RG type

reosarevok:
### Fix MBS-11360

For some reason, the entity_name for release group is 'rg' in Controller::ReleaseGroup, despite its ENTITIES entry being 'release_group'. This seems very weird to me but it'd probably be a mess to change it now, so I just massage it into what the JS code expects instead so that the RG alias form will work.


yvanzo (approved):
Works for me.

Tested by adding alias, then editing it from:
* http://localhost:5000/release-group/7c4cab8d-dead-3870-b501-93c90fd0a580/aliases
* http://localhost:5000/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/aliases

Would it be possible to test it through `t/lib/t/MusicBrainz/Server/Controller/ReleaseGroup/AddAlias.pm`?

yvanzo:
Merging for hotfix release. (Test can be added later on to prevent regressions in the future as this patch has been tested manually.)

PR #1892, opened 2021-02-03
MBS-11363: Fix broken returnto in merge-helper.tt

mwiencek:
## Fix MBS-11363: Regression: Bad request when cancelling a merge

returnto should be prefixed by ? not & since it's the start of the query parameters


yvanzo (approved):
Maybe target `master` branch as it has another fix that will go through `beta` and `production` as well.

reosarevok (approved):

PR #1893, opened 2021-02-04
MBS-11365: Allow Resident Advisor /podcast URLs for releases

reosarevok:
### Implement MBS-11365

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2703,10 +2703,14 @@ const CLEANUPS = {
       new RegExp('^(https?://)?(www\\.)?ra\\.co/', 'i'),
       new RegExp('^(https?://)?(www\\.)?residentadvisor\\.net/', 'i'),
     ],
-    type: {...LINK_TYPES.otherdatabases, ...LINK_TYPES.review},
+    type: {
+      ...LINK_TYPES.otherdatabases,
+      ...LINK_TYPES.review,
+      ...LINK_TYPES.discographyentry,
+    },
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(www\.)?ra\.co\//, 'https://ra.co/');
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2703,10 +2703,18 @@ const CLEANUPS = {
...
-    type: {...LINK_TYPES.otherdatabases, ...LINK_TYPES.review},
+    type: {
+      artist: LINK_TYPES.otherdatabases.artist,
+      event: LINK_TYPES.otherdatabases.event,
+      label: LINK_TYPES.otherdatabases.label,
+      place: LINK_TYPES.otherdatabases.place,
+      recording: LINK_TYPES.otherdatabases.recording,
+      release: LINK_TYPES.discographyentry.release,
+      release_group: LINK_TYPES.review.release_group,
+    },
```
yvanzo, line 2714:
```suggestion
    type: {
      ...LINK_TYPES.otherdatabases,
      ...LINK_TYPES.review,
      ...LINK_TYPES.discographyentry,
    },
```

yvanzo (changes requested):

root/static/scripts/edit/URLCleanup.js
reosarevok: "AFAICT I need to specify all the entity types for otherdatabases in order for it not to clobber discographyentry". I tried this earlier and it didn't work - or does the specific order matter here?

yvanzo: How does it happen it did not clobber `review` until now and it would clobber `discographyentry`? Yes, the specific order matter. I did not test this suggestion though.

reosarevok: Ok, then it might be because of that, and the one I want to keep needs to be last :) I guess I'll try that on Monday, thanks.

yvanzo (approved):
Tests coverage seems sufficient.

mwiencek (approved):

PR #1897, opened 2021-02-06
MBS-11174: Swap Add release edit types (216,31 instead of 31,216)

jesus2099:
# Problem

Currently, the user profile links to [`31,216`](https://musicbrainz.org/search/edits?auto_edit_filter=&conditions.0.field=editor&conditions.0.operator=%3D&conditions.0.name=jesus2099&conditions.0.args.0=285909&combinator=and&conditions.1.field=type&conditions.1.operator=%3D&conditions.1.args=31,216&conditions.2.field=status&conditions.2.operator=%3D&conditions.2.args=2&negation=0&order=desc) so the **Add release** type select option filter (value `216,31`) is not selected in the refine edit search form.
So if you click Search button, you will get an error:

# Solution

Fixes [MBS-11174](https://tickets.metabrainz.org/browse/MBS-11174).

I have just swapped the types, to match the select option value: [`216,31`](https://musicbrainz.org/search/edits?auto_edit_filter=&conditions.0.field=editor&conditions.0.operator=%3D&conditions.0.name=jesus2099&conditions.0.args.0=285909&combinator=and&conditions.1.field=type&conditions.1.operator=%3D&conditions.1.args=216,31&conditions.2.field=status&conditions.2.operator=%3D&conditions.2.args=2&negation=0&order=desc) so
the **Add release** type (value same as request: `216,31`) is now selected and you will get no error if you refine the search by clicking Search button

# Action

It is just an addition to @reosarevok's 809526d122193b5b0d8dfcf287659ec2a4383608.

I could not test on my local server because I have no stats in user profiles (except a few Add cover art, don't know why).
But theoretically it should work.

If you have a local server with statistic data, please test for me, sorry.

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -401,8 +401,8 @@ const UserEditsProperty = ({
     : entityType === 'release' ? (
       // Also list historical edits
       [
-        TYPES.EDIT_RELEASE_CREATE,
         TYPES.EDIT_HISTORIC_ADD_RELEASE,
+        TYPES.EDIT_RELEASE_CREATE,
       ].join(',')
     ) : TYPES[`EDIT_${entityType.toUpperCase()}_CREATE`];
   const searchEditsURL = (createEditTypes => (


yvanzo:
@brainzbot add to whitelist

yvanzo (commented):
I cannot reliably reproduce this issue. I don’t see any improvement with this change. Sometimes it works, sometimes it fails. It seems it doesn’t have anything to do with ordering edit types.

reosarevok:
Same here - I loaded it several times in current master without reordering and Add release seemed to be selected just fine - I could refine it further as well. I didn't even get any case where it didn't work.

yvanzo:
@reosarevok: What happens is that the “Add release” type in search conditions is not always selected.

reosarevok:
I got that - it's just it seemed to always be selected for me (Chrome FWIW). Or at least in my 10 or so tries. Not sure what could be causing this, but agreed it's probably not this order.

jesus2099:
You are right!
I loaded several times with [31/216](https://musicbrainz.org/search/edits?auto_edit_filter=&conditions.0.field=editor&conditions.0.operator=%3D&conditions.0.name=jesus2099&conditions.0.args.0=285909&combinator=and&conditions.1.field=type&conditions.1.operator=%3D&conditions.1.args=31,216&conditions.2.field=status&conditions.2.operator=%3D&conditions.2.args=2&negation=0&order=desc) and [216/31](https://musicbrainz.org/search/edits?auto_edit_filter=&conditions.0.field=editor&conditions.0.operator=%3D&conditions.0.name=jesus2099&conditions.0.args.0=285909&combinator=and&conditions.1.field=type&conditions.1.operator=%3D&conditions.1.args=216,31&conditions.2.field=status&conditions.2.operator=%3D&conditions.2.args=2&negation=0&order=desc) with inconstant results.

I did not notice that it could work with the query parameter with a different value than the select option value!

```html
<select name="conditions.1.args" multiple="multiple" size="15">
	<option value="216,31" selected="selected">Add release</option>
</select>
```

Isn't it surprising?

By the way, the type select being unselected is not a new bug for me.
It's true that it must come from something else, then.
This is the learning from these tests.

In fact, this bug of loosing parameters and the fact that it works with swapped arguments, both indicate that there is some javascript that is rebuilding the search form out of the URL query parameters.
And this javascript has issues many times to reselect everything.

I am currently searching JIRA to see if I ever filed a bug.

jesus2099:
There it is! https://tickets.metabrainz.org/browse/MBS-9573

jesus2099:
Like for Add release, Edit recording is also a **composite edit type**!

```html
<select name="conditions.1.args" multiple="multiple" size="15">
	<option value="216,31" selected="selected">Add release</option>
	<option value="245,72,204">Edit recording</option>
</select>
```

jesus2099:
I am cancelling my PR, thanks for your great review and testing!

The issue may be at this level:

https://github.com/metabrainz/musicbrainz-server/blob/16bad991cccf057a645df9b09cdc660c3733715c/root/edit/search_macros.tt#L184-L197

And is tracked by [MBS-9573](https://tickets.metabrainz.org/browse/MBS-9573).

yvanzo:
Given the difficulty to reproduce it reliably, I would bet on a caching issue.

jesus2099:
Now I work with **Add disc ID** edits as they are in the first rows.
I have changed the MBS code to display edit ID after ID name in the select box:

In fact, the Perl thing that is creating the edit_types array of objects is sorting keys randomly:

image
image

https://github.com/metabrainz/musicbrainz-server/blame/107634c5ee958dbcc1b607efd51c25e8adf17dc8/lib/MusicBrainz/Server/Controller/Edit.pm#L230-L234

I tried reverting 91e9777d1c1de64136fef7fad7211f809606e739 but the problem is not at this line.
The problem is I think on the `join map` level.
I am trying to understand Perl... :thinking:

jesus2099:
But in fact when I change this `lib/MusicBrainz/Server/Controller/Edit.pm` my changes are not taken in account.
Do I have to restart docker after changing a pm file, maybe?

yvanzo:
@jesus2099, just run `sudo docker-compose restart musicbrainz`

jesus2099:
Thanks @yvanzo, I have submitted #1900 now. :)

PR #1898, opened 2021-02-07
MBS-11369: Revert to middle vertical alignment for table.details data

david-russo:
# Problem

Detail table rows begin with colon-terminated headers which can easily wrap onto multiple lines. When this occurs, the value associated with the header (which follows in the next cell) can appear far above where the header ends when the value's cell is top-aligned, potentially making the value appear unrelated to the header and harder to parse than it needs to be.

Before

# Solution

Revert to the default theme's vertical alignment style for data cells which in most circumstances should more clearly delineate a row's values from any preceding or succeeding rows.

After

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -859,8 +859,6 @@ table.details {
     }
 
     td {
-        vertical-align: top;
-
         span.new, span.old,
         &.new, &.old {
             border: 1px solid @medium-border;


yvanzo:
@brainzbot add to whitelist

yvanzo (approved):
Thank you for fixing this. I created ticket MBS-11369 to track this change (for both beta testers and release changelog).

It reverts vertical alignment introduced with commit https://github.com/metabrainz/musicbrainz-server/commit/ec6b44197e39478ca86b497b37178b343e51d027. It isn’t clear (from commit message and ticket MBS-6276) why it has been introduced at that time.

Changes can be seen in:
* Credits at bottom of release page:
  https://musicbrainz.org/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60 
* Changes for edit:
  https://musicbrainz.org/edit/63185679

It improves readability when the header has a larger height than data (as in relationships/credits). It doesn’t impair readability otherwise.

reosarevok (approved):
Seems sensible to me, tested in the same places @yvanzo did plus the example you gave. For your specific example, *bottom* align works even better IMO, but that would require extra special-casing so that it wouldn't mess the display on edits and whatnot so this is probably good enough.

david-russo:
Thanks, guys. Yeah, I originally went with bottom alignment before I considered how widely used the details class might be, and figured middle alignment worked just as well to separate the data. Plus, less code, so that's always a win :)

kellnerd:
@yvanzo [This release event edit](https://beta.musicbrainz.org/edit/77130737) is an example where alignment to the top looks better and was probably the reason for MBS-6276.

reosarevok:
I don't mind the way that looks now either, but we could always special case that display if that's the main problem :)

kellnerd:
I would not mind for simple cases like the one above either, but for [long country lists](https://beta.musicbrainz.org/edit/74086408) it is probably better to have a separate style class for cells which contain lists.

reosarevok:
Added https://github.com/metabrainz/musicbrainz-server/pull/1905 for release event diffs. If you find any others where center is worse than top, let me know :)

PR #1899, opened 2021-02-08
MBS-11370: display track lengths of 0 ms or -1 ms as unknown 

reosarevok:
### Implement MBS-11370

Same issue as https://github.com/metabrainz/musicbrainz-server/pull/1783 but on another edit type. I'm also moving that PR's check from JS to Perl.


mwiencek (approved):

yvanzo (approved):
:shipit:

PR #1901, opened 2021-02-08
MBS-9098: Add "memberOf" relationships to JSON-LD

reosarevok:
### Implement MBS-9098

This just reuses the member code since everything is the same except for the actual property name.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
@@ -37,7 +37,18 @@ around serialize => sub {
             $ret->{groupOrigin} = serialize_entity($entity->begin_area, $inc, $stash);
         }
 
-        my @members = grep { $_->direction == 2 } @{ $entity->relationships_by_link_type_names('member of band') };
+        my $member_rels = $entity->relationships_by_link_type_names('member of band');
+        my @members;
+        my @member_ofs;
+
+        for my $member_rel (@{ $member_rels }) {
+            if ($member_rel->direction == 1) {
+                push(@member_ofs, $member_rel);
+            } elsif ($member_rel->direction == 2) {
+                push(@members, $member_rel);
+            }
+        }
+
         if (@members) {
             my %seen_members;
             for my $member (@members) {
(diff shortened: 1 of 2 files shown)


t/lib/t/MusicBrainz/Server/Controller/Artist/Show.pm
```diff
@@ -157,6 +157,27 @@ EOSQL
         ],
         '@context' => 'http://schema.org'
     };
+
+    $mech->get_ok('/artist/2a62773a-cdbf-44c6-a700-50f931504054');
```
reosarevok, line 161: This fails and I can't figure out what I'm missing. Halp?

lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
```diff
@@ -41,11 +41,20 @@ around serialize => sub {
         if (@members) {
             my %seen_members;
             for my $member (@members) {
-                $seen_members{$member->target->gid} = member_relationship($member, $inc, $stash, $seen_members{$member->target->gid});
+                $seen_members{$member->target->gid} = member_relationship($member, 'member', $inc, $stash, $seen_members{$member->target->gid});
             }
             $ret->{member} = [ map { $seen_members{$_} } sort keys %seen_members ];
         }
 
+        my @member_ofs = grep { $_->direction == 1 } @{ $entity->relationships_by_link_type_names('member of band') };
```
mwiencek, line 49: Since this runs on every artist page, I'm going to be inquisitive here. :) Can't we combine these blocks instead of running `relationships_by_link_type_names` twice and looping over everything twice? e.g. remove the `grep { $_->direction == $x }` from the block above and check the direction inside the loop?

mwiencek: If that doesn't make sense, at least save `@{ $entity->relationships_by_link_type_names('member of band') }` to a variable so both block can use it.

reosarevok: Fair enough, made the change :)

t/lib/t/MusicBrainz/Server/Controller/Artist/Show.pm
```diff
@@ -157,6 +157,27 @@ EOSQL
...
+    page_test_jsonld $mech => {
+        'name' => 'Person A',
+        '@type' => 'MusicGroup',
```
mwiencek, line 164: nitpick: since this is supposed to be a person according to the name, I assume the test data is wrong, 'cause the type isn't `['Person', 'MusicGroup']`

reosarevok: It was missing types, yes. I added them now - and changed the ID for the group from 1 to 100 because 1 is hardcoded to be VA and not support a Group type! :)

mwiencek (approved):
thanks!

PR #1902, opened 2021-02-08
MBS-11368: Set nameVariation in EditMedium if rec name != track name

reosarevok:
### Fix MBS-11368

This used to work, but broke on React conversion because content is no longer a string, but a React element. The easiest way to do this seems to be to just compare the two names when calling EntityLink.


reosarevok:

mwiencek (approved):

yvanzo (approved):

PR #1903, opened 2021-02-09
MBS-11373: Overwrite empty dates in relationship merges

reosarevok:
### Fix MBS-11373

I implemented MBS-11181 a bit too overzealously. If we add an ended relationship on top of one with completely empty dates, we should still overwrite it - otherwise it kinda makes the whole system useless.

### DIFF
--- a/root/static/scripts/common/utility/isDateEmpty.js
+++ b/root/static/scripts/common/utility/isDateEmpty.js
@@ -7,6 +7,20 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+export function isDateObservableEmpty(
+  date: {
+    day: KnockoutObservable<string | null>,
+    month: KnockoutObservable<string | null>,
+    year: KnockoutObservable<string | null>,
+  },
+): boolean {
+  return !(
+    nonEmpty(date.year()) ||
+    nonEmpty(date.month()) ||
+    nonEmpty(date.day())
+  );
+}
+
 export default function isDateEmpty(date: ?PartialDateT): boolean {
   return (date == null) ||
     (date.year == null && date.month == null && date.day == null);
--- a/root/static/scripts/relationship-editor/common/fields.js
+++ b/root/static/scripts/relationship-editor/common/fields.js
@@ -24,6 +24,7 @@ import {displayLinkAttributesText}
   from '../../common/utility/displayLinkAttribute';
 import formatDate from '../../common/utility/formatDate';
 import formatDatePeriod from '../../common/utility/formatDatePeriod';
+import {isDateObservableEmpty} from '../../common/utility/isDateEmpty';
 import request from '../../common/utility/request';
 import {fixedWidthInteger, uniqueId} from '../../common/utility/strings';
 import mbEdit from '../../edit/MB/edit';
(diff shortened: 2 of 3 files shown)


root/static/scripts/common/utility/isDateEmpty.js
```diff
@@ -7,6 +7,17 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+export function isDateObservableEmpty(
+  date: {
+    day: KnockoutObservable<number | null>,
+    month: KnockoutObservable<number | null>,
+    year: KnockoutObservable<number | null>,
+  },
```
mwiencek, line 15: Since you check `date == null` below, you might want to allow null values in the type too. (Not that it matters much since the relationship editor isn't Flow-typed.)
```suggestion
  } | null,
```

root/static/scripts/common/utility/isDateEmpty.js
```diff
@@ -7,6 +7,17 @@
...
+): boolean {
+  return (date == null) ||
+    (date.year() == null && date.month() == null && date.day() == null);
```
mwiencek, line 18: Those are generally strings, aren't they? So I'm not sure that they wouldn't be `''` if empty. I'd double check, but we perform the `tooShortYear` checks on the same observable AFAICT, which expects a string.
```suggestion
    !(nonEmpty(date.year()) || nonEmpty(date.month()) || nonEmpty(date.day());
```

root/static/scripts/common/utility/isDateEmpty.js
```diff
@@ -7,6 +7,17 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+export function isDateObservableEmpty(
+  date: {
+    day: KnockoutObservable<number | null>,
+    month: KnockoutObservable<number | null>,
+    year: KnockoutObservable<number | null>,
```
mwiencek, line 14: (These types might need changing too)
```suggestion
    year: KnockoutObservable<StrOrNum | null>,
```

root/static/scripts/common/utility/isDateEmpty.js
reosarevok: Hmm. This seems to work as is, which made me assume that this worked? But I can make the change and see.

root/static/scripts/common/utility/isDateEmpty.js
reosarevok: I was actually wondering: can this even be null? It might be that this always has a value.

root/static/scripts/common/utility/isDateEmpty.js
reosarevok: Checked, printed, etc. - it's null, not empty string.

root/static/scripts/common/utility/isDateEmpty.js
reosarevok: Seems only string actually (or null)

root/static/scripts/common/utility/isDateEmpty.js
mwiencek: Ok. Thanks for checking. :) Though I do slightly prefer `nonEmpty` in the off chance that there's some weird code path that sets it to an empty string.

mwiencek (approved):

root/static/scripts/common/utility/isDateEmpty.js
reosarevok: Ok, can make the change :)

yvanzo (approved):
:ship:

PR #1904, opened 2021-02-09
Yet more eslint rules for CI

reosarevok:
See commits for details


reosarevok:

script/.check_eslint_rule_config.yaml
```diff
@@ -0,0 +1,73 @@
...
+    onlyFilesWithFlowAnnotation: true
+  react:
+    version: detect
+    flowVersion: 0.144.0
+
+rules:
+  flowtype/define-flow-type: warn
+  flowtype/use-flow-type: warn
+  react/jsx-uses-vars: warn
+
```
mwiencek, line 73: This shouldn't cause any real problems, but just for a sake of tidyness, there's apparently some trailing whitespace but no newline at the end of the file.

mwiencek (approved):

.eslintrc.yaml
```diff
@@ -256,6 +256,11 @@ rules:
   react/require-optimization: off
   react/require-render-return: error
   react/self-closing-comp: error
+  react/sort-comp: [warn, {order: [type-annotations,
+                                   static-methods,
+                                   lifecycle,
+                                   everything-else,
+                                   render]}]
```
yvanzo, line 272: Instead of changing order for all files but `TagEditor.js`, would it be possible to change order for `TagEditor.js` only?

yvanzo (commented):
Just reviewed first commit.

.eslintrc.yaml
reosarevok: I guess it depends on whether we have a reason to want this order or not. I didn't pick the order for the rule - @mwiencek, comments? :)

mwiencek: Well, I do think the suggested order is the most natural, but I'm also okay with dropping this rule entirely: class-based components should eventually be converted to functional ones, anyway, and the order doesn't impact readability enough to enforce it IMO.

reosarevok: @yvanzo: enforce the rule for now, or drop the rule for now?

mwiencek: I dropped it.

PR #1905, opened 2021-02-10
MBS-11375: Use top-align for release event tables in Edit release

reosarevok:
### Implement MBS-11375

MBS-11369 changed the vertical align and in most places that's for the best, but release event diff tables are better with top align. Otherwise, a long list being turned into a small one (e.g. [edit #74086408](https://beta.musicbrainz.org/edit/74086408)) has the smaller value way down the page and makes understanding the change a lot harder.

### DIFF
--- a/root/static/scripts/edit/components/edit/ReleaseEventsDiff.js
+++ b/root/static/scripts/edit/components/edit/ReleaseEventsDiff.js
@@ -116,7 +116,7 @@ const ReleaseEventsDiff = ({
   }
 
   return (
-    <tr>
+    <tr className="release-events-diff">
       <th>{l('Release events:')}</th>
       <td className="old">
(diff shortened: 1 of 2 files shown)


kellnerd:
Since you explicitly asked for other tables that should be top-aligned, I had a look and found one other case where vertical centering is at least inconsistent (not necessarily worse): [Edit #77166535 - Edit medium](https://beta.musicbrainz.org/edit/77166535) aligns the cells of each track to the middle while the [relase page](https://beta.musicbrainz.org/release/6b968ac7-4393-45ab-8f33-7e7b277c6be5) aligns the tracks to the top.

yvanzo (approved):

yvanzo:
@brainzbot retest this please

PR #1906, opened 2021-02-10
MBS-9706: Normalize OCRemix URLs (including HTTPS)

reosarevok:
### Implement MBS-9706

Validation is MBS-7674 but there's (old) discussion there about whether /remix pages should be freedownload instead. Will do that later once we have more feedback.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/URL/OCReMix.pm
+++ b/lib/MusicBrainz/Server/Entity/URL/OCReMix.pm
@@ -5,6 +5,10 @@ use Moose;
 extends 'MusicBrainz::Server::Entity::URL';
 with 'MusicBrainz::Server::Entity::URL::Sidebar';
 
+override href_url => sub {
+    shift->url->as_string =~ s{^http:}{https:}r;
+};
+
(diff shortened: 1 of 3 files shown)


yvanzo (commented):
The ticket is about normalization, whereas this pull request is about `https` only.

reosarevok:
It's not though, it also has `([^\/?#]+).*$` to drop fragments and whatnot. Shouldn't that do the trick?

yvanzo (changes requested):
Why does it now contain unrelated commits for Overture? It should probably be rebased on `master`.

reosarevok:
That's unrelated and that's why it is in a new commit, but it's just a small bug I noticed when fixing this. I can give it a ticket number if you feel it's worth it.

yvanzo:
@reosarevok: Because the first commit message, I stopped reviewing after it. I did not see there were two commits or OCRemix. Maybe add part numbers `(1/2)`, `(2/2)` next to the ticket number in commit message?

mwiencek (approved):

yvanzo (approved):
Thank you @reosarevok!

PR #1908, opened 2021-02-12
MBS-11381: Add forms for admins to add and update timeline events

reosarevok:
### Implement MBS-11381

Was tired of needing to use direct DB editing for this. Not making the effort to check whether the date is a duplicate when adding or editing these, since this is an admin-only form (effectively for myself) and an ISE if I do something stupid is not a problem.

Tested by adding, editing and removing by hand locally.


mwiencek (approved):
might be useful to link in the admin menu, but lgtm otherwise

reosarevok:
Oh, fair point. Added a link.

mwiencek:
Rebased this and fixed some TO_JSON issues.

root/admin/statistics-events/StatisticsEventIndex.js
```diff
@@ -0,0 +1,70 @@
...
+                </a>
+              </td>
+            </tr>
+          )) : null}
+      </tbody>
+    </table>
+    <p>
+      <span className="buttons">
+        <a href="/admin/statistics-events/create">
+          {l('Add new event')}
```
mwiencek, line 63: Should we extract these strings into the statistics domain, or at least add some context? e.g. this could be confused with event entities

reosarevok: Oh, yes, we should add context in any case, even if it is in the statistics domain. Otherwise it will be confusing. Will do.

PR #1909, opened 2021-02-15
MBS-11386: Load area containment for indexed search

reosarevok:
### Fix MBS-11386


mwiencek (approved):

yvanzo (approved):
Tested with local search server and sample data.

PR #1911, opened 2021-02-15
MBS-11384 / MBS-11385: Ensure all country context names are translated

reosarevok:
### Fix MBS-11384 / MBS-11385

The existence of iso_3166_1 (not the area having the country type) is what matters for whether we make the area name available for translation. As such, that's also what we should be checking for.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Area.pm
+++ b/lib/MusicBrainz/Server/Entity/Area.pm
@@ -21,8 +21,9 @@ sub entity_type { 'area' }
 
 sub l_name {
     my $self = shift;
-    my $type = defined $self->type ? $self->type->id : $self->type_id;
-    if (defined $type && $type == $AREA_TYPE_COUNTRY) {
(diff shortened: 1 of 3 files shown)


reosarevok:

mwiencek (approved):

yvanzo (approved):

reosarevok:

reosarevok:
Requesting review again since the test showed that it was broken on areas coming from autocomplete (which is a mess without any flow typing).

yvanzo (commented):
What broke exactly? Would it be possible to make a test for it?

reosarevok:
> What broke exactly?

Areas shown in the "You selected X" bubble on the side of an edit form were untranslated.

> Would it be possible to make a test for it?

I found out *because* there's a test for it :D

yvanzo (approved):
Cool, please link to it next time.

mwiencek (approved):

PR #1913, opened 2021-02-16
MBS-11268 / MBS-11349 / MBS-11463: Improvements related to medium duration

reosarevok:
### Implement MBS-11268 / MBS-11463 and fix MBS-11349

This adds a view (MBS-11463) that allows us to access the durations of medium tracks without having to load all the data for every track in Moose.

Using this view, I re-implemented MBS-11268 (which seems to work fine now).

I also implemented MBS-11349 by replacing the complicated calculation of medium length based on either loaded tracks or discIDs with a calculation based on the values in this view.

A good place to test both of these seems to be http://localhost:5000/release/ab9e6f50-b248-4ed2-a591-1f175e609e44 (in the sample data).

NB: I wouldn't be surprised if there are some places where we loaded tracks or discIDs only to get medium durations, and if so those loads could now be removed. If you can think of any, do let me know! Guess there's no good way to find this programmatically.

### DIFF
--- a/admin/sql/updates/20210215-mbs-11268.sql
+++ b/admin/sql/updates/20210215-mbs-11268.sql
@@ -1,0 +1,15 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+CREATE OR REPLACE VIEW medium_track_durations AS
+    SELECT
+        medium.id AS medium,
+        array_agg(track.length ORDER BY track.position) FILTER (WHERE track.position = 0) AS pregap_length,
+        array_agg(track.length ORDER BY track.position) FILTER (WHERE track.position > 0 AND track.is_data_track = false) AS cdtoc_track_lengths,
+        array_agg(track.length ORDER BY track.position) FILTER (WHERE track.is_data_track = true) AS data_track_lengths
+    FROM medium
+    JOIN track ON track.medium = medium.id
+    GROUP BY medium.id;
+
+COMMIT;
--- a/lib/MusicBrainz/Server/Entity/Medium.pm
+++ b/lib/MusicBrainz/Server/Entity/Medium.pm
@@ -143,6 +143,21 @@ has 'cdtoc_track_count' => (
     isa => 'Int',
 );
 
+has 'cdtoc_track_lengths' => (
+    is => 'rw',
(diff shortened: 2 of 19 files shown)


lib/MusicBrainz/Server/Entity/Medium.pm
```diff
@@ -142,6 +126,21 @@ has 'cdtoc_track_count' => (
     isa => 'Int',
 );
 
+has 'cdtoc_track_lengths' => (
+    is => 'rw',
+    isa => 'ArrayRef[Maybe[Int]]'
```
reosarevok, line 131: Are these types correct? I always suck at moose typing.

mwiencek: I think they're okay; the values can be undef if the track durations aren't set, right?

mwiencek:
Looks good, I think, but the script needs adding to upgrade.json etc.

reosarevok:
Whoops. Made those changes.

admin/sql/updates/20210215-mbs-11268.sql
```diff
@@ -0,0 +1,15 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+CREATE OR REPLACE VIEW medium_track_durations AS
```
mwiencek, line 5: I removed the `SET search_path` at the beginning of this script (and pushed a commit to remove it from the other scripts). It's already injected at the beginning of the transaction.

lib/MusicBrainz/Server/Entity/Medium.pm
```diff
@@ -96,40 +97,26 @@ sub may_have_discids {
...
-        my $length = 0;
-
-        for my $trk ($self->all_tracks)
-        {
-            return undef unless defined $trk->length;
+    my @track_times = (
+        @{ $self->pregap_length // [] },
+        @{ $self->cdtoc_track_lengths // [] },
+        @{ $self->data_track_lengths // [] },
+    );
```
mwiencek, line 115: This previously used `push` to append the latter two arrays. I changed it to put them all in the declaration, which is equivalent.

PR #1915, opened 2021-02-17
MBS-10312: Remove tracking parameters from URLs

atj:
The functionality added here will not work in IE11 due to the use of the `URL()` constructor, however it should fail gracefully and just return the original URL in such cases. I feel this is a reasonable compromise, as the functionality is not crucial and the alternative would be to implement a clunky regexp based parameter parser.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -325,6 +325,47 @@ function findAmazonTld(url) {
   return tld;
 }
 
+const TRACKING_PARAMS = [
+  // https://en.wikipedia.org/wiki/Click_identifier
+  'dclid',
+  'gclid',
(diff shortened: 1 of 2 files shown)


reosarevok:
@brainzbot add to whitelist

reosarevok:
@atj: There's one error in the tests because of eslint :) See https://app.circleci.com/pipelines/github/metabrainz/musicbrainz-server/4823/workflows/ffd5d645-45d8-4dcf-b67b-ac41863710d2/jobs/11664 and do fix it, if you have the time.

atj:
@reosarevok: Thanks, should be resolved now :crossed_fingers:

mwiencek (approved):

yvanzo (approved):
Thank you @atj!

PR #1920, opened 2021-02-18
MBS-11396: Show item_entity_type on Series/CollectionType admin tables

reosarevok:
### Fix MBS-11396

Obviously the point of these was not to show "series_type" or "collection_type" for every row, but to show item_entity_type - what type of entities are allowed as part of the series or collection.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/CollectionType.pm
+++ b/lib/MusicBrainz/Server/Entity/CollectionType.pm
@@ -21,6 +21,15 @@ sub l_name {
     return lp($self->name, 'collection_type')
 }
 
+around TO_JSON => sub {
+    my ($orig, $self) = @_;
+
+    return {
+        %{ $self->$orig },
+        item_entity_type => $self->item_entity_type,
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Entity/CollectionType.pm
```diff
@@ -21,6 +21,15 @@ sub l_name {
     return lp($self->name, 'collection_type')
 }
 
+around TO_JSON => sub {
+    my ($orig, $self) = @_;
+
+    return {
+        %{ $self->$orig },
+        item_entity_type => $self->item_entity_type,
```
mwiencek, line 29: (Seems it's already in the Flow type since https://github.com/metabrainz/musicbrainz-server/commit/c53b65015d71543e44d7dbd44e1d2078208f30bd but wasn't actually added to the output until now.)

mwiencek (approved):

yvanzo (approved):

PR #1922, opened 2021-02-18
MBS-5680: Allow inc=artist-credits in combination with rels incs that have ACs

reosarevok:
### Implement MBS-5680

Currently, you can do a query including recording-rels for an artist, or a work, but the returned related recordings will not include their artist credit. You also can't just add artist-credits to the inc, because the validator rejects that.
This isn't reasonable - it's perfectly legitimate to ask for artist-credits when requesting recording, release or release group rels, since all of those have artist credits.


mwiencek (commented):
The change looks pretty simple, but since it does allow requesting new info I think some tests should be added (or add inc=artist-credits to some existing ones).

reosarevok:
Fair enough, extended one existing test at least for both JSON and XML

mwiencek (approved):

PR #1923, opened 2021-02-19
MBS-11398: Add Qobuz links to sidebar

reosarevok:
### Implement MBS-11398

We have over 15 thousand Qobuz links, so it seems sensible enough to add it to the sidebar. Not adding autoselect because it can be both download and streaming, it seems.


yvanzo (commented):
Many Qobuz URLs in the MB DB contain a locale code. Is it because content may be available from some countries only or just for UI localization? In the first case, it would make sense to show the locale code in the sidebar. In the second case, maybe these locale codes should be removed (that would be beyond the current ticket though).

reosarevok:
The former, mostly. See some recent discussion on that matter on https://musicbrainz.org/edit/77416060 :)

lib/MusicBrainz/Server/Entity/URL/Qobuz.pm
```diff
@@ -0,0 +1,20 @@
...
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+Copyright (C) 2021 MetaBrainz Foundation
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+=cut
```
mwiencek, line 20: The spacing here is inconsistent with all the ones you changed recently
```suggestion
=head1 COPYRIGHT AND LICENSE

Copyright (C) 2021 MetaBrainz Foundation

This file is part of MusicBrainz, the open internet music database,
and is licensed under the GPL version 2, or (at your option) any
later version: http://www.gnu.org/licenses/gpl-2.0.txt

=cut
```

yvanzo:
> The former, mostly. See some recent discussion on that matter on https://musicbrainz.org/edit/77416060 :)

I did not formulate it correctly: Is it they have different content depending on the locale? It does not seem so.
If this edit is relevant to the ticket, please link and summarize it in comment to this ticket.

yvanzo (approved):
Potential normalization of these URLs is a separate issue that should probably go through a STYLE decision at first. Let’s display them as-is for now.

mwiencek (approved):

PR #1924, opened 2021-02-19
MBS-10117: Fix broken maxlength password message on register page

reosarevok:
### Fix MBS-10117

Trying to pass the default message through l() breaks it, displaying "Field should not exceed [quant,_1,character]. You entered [_2]". It also can't be translated anyway, since it's not exported to .pot files. So I'm adding a custom error message that can be used for text_maxlength for both password and username, which fixes the issue and should also allow translating it. It seems unneeded to have separate error messages just changing the field name, since these get displayed just under the relevant field anyway.

### DIFF
--- a/lib/MusicBrainz/Server/Form/User/Register.pm
+++ b/lib/MusicBrainz/Server/Form/User/Register.pm
@@ -10,11 +10,27 @@ with 'MusicBrainz::Server::Form::Role::CSRFToken';
 
 has '+name' => ( default => 'register' );
 
+my $text_too_long_message = N_l('The value of this field cannot be longer than {max} characters, but you entered {actual}.');
+
+sub localize_method_with_text_maxlength {
+    my ($self, $message, @args) = @_;
+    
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Form/User/Register.pm
```diff
@@ -10,20 +10,45 @@ with 'MusicBrainz::Server::Form::Role::CSRFToken';
...
+    },
+    localize_meth => sub {
+        my ($self, $message, @args) = @_;
+        
+        if ($message eq $text_too_long_message) {
+            return l($message, { max => $args[0], actual => $args[1] });
+        }
+
+        return l($message);
+    }
```
mwiencek, line 31: I think it's worth moving this to a named subroutine so you can just do `localize_meth => \&localize_meth,` here and below.

mwiencek (approved):

yvanzo (approved):
LGTM but did not test.

PR #1927, opened 2021-02-22
MBS-10452: Implement tagger links for releases on recording pages

reosarevok:
### Implement MBS-10452

This makes the table even more crowded, but only on demand, so it shouldn't be a problem - most users won't see it, unless they have done a Picard lookup, in which case it should be of use.


mwiencek (approved):

yvanzo (approved):
Attaching a screenshot to the ticket would be welcome too.

reosarevok:
Done.

PR #1929, opened 2021-02-23
MBS-10487: Allow adding instruments and areas from inline search

reosarevok:
### Implement MBS-10487

This gives users with the right privileges the same ability to add areas and instruments from inline search as any editor has to add any other entity type (with a pop-up dialog and whatnot).


yvanzo (approved):

root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -466,9 +472,18 @@ $.widget('mb.entitylookup', $.ui.autocomplete, {
...
 
+    const isTopWindow = window === window.top;
+    const user = window.__MB__?.$c.user;
+    let userCanAdd = true;
+    if (entity === 'area') {
+      userCanAdd = isLocationEditor(user);
+    } else if (entity === 'instrument') {
+      userCanAdd = isRelationshipEditor(user);
+    }
+    const allowCreation = isTopWindow && userCanAdd;
```
mwiencek, line 485: Approving though I wouldn't both calculating `userCanAdd` if we already know `isTopWindow` is false.
```suggestion
    const user = window.__MB__?.$c.user;
    let userCanAdd = false;
    if (isTopWindow) {
      if (entity === 'area') {
        userCanAdd = isLocationEditor(user);
      } else if (entity === 'instrument') {
        userCanAdd = isRelationshipEditor(user);
      } else {
        userCanAdd = true;
      }
    }
    const allowCreation = isTopWindow && userCanAdd;
```

mwiencek (approved):

root/static/scripts/common/MB/Control/Autocomplete.js
reosarevok: If we're already having `userCanAdd` be `false` unless `isTopWindow`, should we even check `isTopWindow` again in `allowCreation`? We could just check `userCanAdd` directly and save one check, I guess?

mwiencek: Yeah, good point. You only need one variable, then, 'cause `const allowCreation = userCanAdd;` is also pointless in that case.

PR #1930, opened 2021-02-23
MBS-10478: Show useful message if too large search result page requested

reosarevok:
### Implement MBS-10478

This currently loads the page and just says "no results", which is quite misleading and also useless. We have all the info to show something useful, so we should do that and at least tell the user which page is the last they can go to.

Screenshot from 2022-07-27 21-51-20

### DIFF
--- a/root/search/components/PaginatedSearchResults.js
+++ b/root/search/components/PaginatedSearchResults.js
@@ -9,8 +9,10 @@
 
 import * as React from 'react';
 
+import {CatalystContext} from '../../context';
 import PaginatedResults from '../../components/PaginatedResults';
 import {type SearchResultT} from '../types';
+import uriWith from '../../utility/uriWith';
 
 type Props<T> = {
   +buildResult: (SearchResultT<T>, number) => React.Node,
@@ -27,6 +29,12 @@ const PaginatedSearchResults = <T>({
   query,
   results,
 }: Props<T>): React.Element<typeof PaginatedResults | 'p'> => {
+  const $c = React.useContext(CatalystContext);
+  const hasLastPage = pager.total_entries > 0;
+  const lastPageUrl = hasLastPage
+    ? uriWith($c.req.uri, {page: pager.last_page})
+    : null;
+
   return results.length ? (
     <PaginatedResults pager={pager} query={query} search>
       <table className="tbl">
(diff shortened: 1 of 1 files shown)


reosarevok:
The reported ticket is as a bug (because of saying "no results" where results exist). Would you make it into an improvement to redirect, or keep as a bug?

reosarevok:

mwiencek:
Hmm. I like the intent to do something useful here, though in the back of my mind I'm worrying about two things

 * Recursively calling `external_search`: looking at the code I myself can't see a way this would lead to infinite recursion, but if it *did* that would be pretty disastrous
 * AFAICT this basically adds a way for people to (maliciously or not) double the amount of traffic to the search server in the same number of rate-limited requests

Since I can't imagine this situation is all that common outside of clicking on an old/saved link, I wonder if it wouldn't be better to just show an error with a link to repeat the search (which will at least be rate limited) if they want to.

reosarevok:
Hmm. I did it like this because we already have pagination elsewhere that goes to the last available page if too large a number is given. If you think it's a worry for the search server, I guess we could just display "This search only has X pages, {link|go to last}?" instead or something. It's less comfortable/helpful, but it might be worth it for safety?

mwiencek:
> Hmm. I did it like this because we already have pagination elsewhere that goes to the last available page if too large a number is given. If you think it's a worry for the search server, I guess we could just display "This search only has X pages, {link|go to last}?" instead or something. It's less comfortable/helpful, but it might be worth it for safety?

I'd prefer displaying a message, personally.  I think this situation isn't that common, so it shouldn't be too much of an inconvenience.  I worry about the current patch making it more trivial for a malicious actor to overload the search server.

reosarevok:
Changed to show a message (example on PR description)

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -942,7 +948,15 @@ sub external_search
...
 
-        return { pager => $pager, offset => $offset, results => \@results, last_updated => $last_updated };
+        return {
+            pager => $pager,
+            offset => $offset,
+            results => \@results,
+            last_updated => $last_updated,
+            $show_largest_available_page ? (
+                largest_available_page => $largest_available_page
+            ) : ()
```
mwiencek, line 958: Is it possible to determine all of this from the pager given to the component already? e.g. there's `pager.last_page`, does that match `$largest_available_page`?  And does `pager.current_page` match `$page`?

reosarevok: `pager.last_page` does give the same as I needed, yes. `pager.current_page` doesn't show the passed parameter though, it is the same as `pager.last_page` if I pass any page number above that number. I could work with that just fine, the only problem is that `pager.last_page` is 1 even if there's no results at all (including in page 1) so if I ask for page 2 of a no-results search, I'd still think "the last page of results is 1".

reosarevok: That absolutely does it, so this is a looot simpler now. Thanks!

mwiencek:
@brainzbot retest this please

mwiencek (approved):

PR #1931, opened 2021-02-23
MBS-10076 / MBS-11403: Friendlier error when deleting attribute that has children 

reosarevok:
### Implement MBS-10076 / MBS-11403

I converted the in_use file to React as a more general error page that takes a message so I could reuse it for this error too.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
+++ b/lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
@@ -115,7 +115,16 @@ sub delete : Chained('attribute_base') Args(1) RequireAuth(account_admin) Secure
     $c->stash->{attribute} = $attr;
 
     if ($c->model($model)->in_use($id)) {
-        $c->stash->{template} = 'admin/attributes/in_use.tt';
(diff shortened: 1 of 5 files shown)


yvanzo (approved):

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -115,10 +117,28 @@ sub delete : Chained('attribute_base') Args(1) RequireAuth(account_admin) Secure
...
+            current_view => 'Node',
+            component_path => 'admin/attributes/CannotRemoveAttribute',
+            component_props => {message => $error_message}
+        );
+        
         $c->detach;
     }
 
+    if ($c->model($model)->has_children($id)) {
+        my $error_message = l('You cannot remove the attribute "{name}" because it is the parent of other attributes.', { name => $attr->name });
```
yvanzo, line 132: Can use typographic quotes when introducing new message:
```suggestion
        my $error_message = l('You cannot remove the attribute “{name}” because it is the parent of other attributes.', { name => $attr->name });
```

mwiencek (approved):

PR #1932, opened 2021-02-24
MBS-10312: Fix tracking parameter removal in Firefox 44

atj:
This commit fixes issues with the tracking parameter removal functionality in Firefox 44.

Unfortunately I've been unable to test the changes as I'm having issues initialising the database in my MB Docker instance.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -353,17 +353,12 @@ function stripTrackingParams(url) {
   }
 
   const params = urlObject.searchParams;
-  const paramKeys = new Set(params.keys());
-  let changed = false;
 
-  for (const param of paramKeys) {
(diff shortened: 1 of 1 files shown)


atj:
OK, this has actually highlighted some issues that weren't previously caught by the tests because the original URL was just passed through if no changes were made, whereas now we are always returning `url.href`.

I'll need to have a think about how best to handle this.

yvanzo:
@atj: Is it a patch to fix issues being introduced from <https://github.com/metabrainz/musicbrainz-server/pull/1915>? If so, please target the `beta` branch directly.

atj:
> @atj: Is it a patch to fix issues being introduced from #1915? If so, please target the `beta` branch directly.

Yes it is. Thanks, I'll close this one and open a new PR.

PR #1934, opened 2021-02-25
MBS-11407: Make Controller->error use React errors

reosarevok:
### Fix MBS-11407

This got missed when we did the conversion, and .tt error templates no longer exist.


mwiencek (approved):

yvanzo (approved):
:ship:

PR #1935, opened 2021-02-25
MBS-10012: Treat http and https version of link as same for adding

reosarevok:
### Implement MBS-10012

There's no reason I can see why we shouldn't block adding a http link if https exists already, and vice versa. I haven't seen any side effects of just making linkTypeAndUrlString normalize it.

Tested by both trying to add the two links from the ticket (http://www.example.com/path/object vs https://www.example.com/path/object) on artist creation and release creation, and by adding both to an artist (temporarily the code disallowing this) and then checking that they are detected as duplicates on loading the artist edit page once the code is changed back.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -480,7 +480,13 @@ function newLinkState(state: $Shape<LinkStateT>) {
 }
 
 function linkTypeAndUrlString(link) {
-  return (link.type || '') + '\0' + link.url;
+  /*
+   * There's no reason why we should allow adding the same relationship
+   * twice when the only difference is http vs https, so normalize this
+   * for the check.
+   */
+  const httpUrl = link.url.replace(/^https/, 'http');
+  return (link.type || '') + '\0' + httpUrl;
 }
 
 function isEmpty(link) {


yvanzo (commented):
A detrimental effect of this patch is that someone entering an `https` URL would see its addition being rejected even though the existing `http` URL might be obsolete. The error message would be no longer appropriate.
Made a comment to the ticket to see if we can use more context to better handle the most common case.

reosarevok:
FWIW, IIRC that shows "this relationship already exists" under both the new *and* the old link, so they could choose to remove (or update) the old link instead, I guess?

mwiencek:
That makes sense to me. If they know they `http` is obsolete and intend to replace it, they can just edit the existing URL.

mwiencek (approved):

yvanzo (changes requested):
I tested by adding `http://www.nin.com/` first, then `https://www.nin.com/` to a release, it has not been detected as duplicate.

yvanzo (approved):
My bad, I did not recompile scripts before testing them.

I noticed another point: If an entity already has a pair of duplicate http/https links, it is not possible to edit that entity unless removing one of the links. It is probably not an issue here as it is very simple to drop one of the two links and it is not a big deal if a newcomer drops the https link instead of the http one.

reosarevok:
Added MBS-11429 for that issue (not so relevant here, I agree, but more relevant elsewhere).

PR #1938, opened 2021-02-25
MBS-9987: JSON-LD: Use "Concept URI" for Wikidata

reosarevok:
### Implement MBS-9987

These are the IDs Wikidata itself uses for RDF / SPARQL, so it makes linked data usage a lot easier by avoiding the user
having to put the right URI together. They redirect to the standard page when loaded in the browser, so it shouldn't make things harder for data users otherwise.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/SameAs.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/SameAs.pm
@@ -87,7 +87,13 @@ sub sameas_url {
     if ($acceptable{$rel->link->type->gid // ''} ||
         $acceptable_parents{$rel->link->type->parent_id // ''} ||
         $acceptable_parents{$rel->link->type->id // ''}) {
-        return $rel->target->url->as_string;
+
+        my $url = $rel->target->url->as_string;
+
+        # Use the Wikidata RDF concept URIs rather than the default site URL (MBS-9987)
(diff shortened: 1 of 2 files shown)


yvanzo (approved):

lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/SameAs.pm
```diff
@@ -87,7 +87,13 @@ sub sameas_url {
     if ($acceptable{$rel->link->type->gid // ''} ||
         $acceptable_parents{$rel->link->type->parent_id // ''} ||
         $acceptable_parents{$rel->link->type->id // ''}) {
-        return $rel->target->url->as_string;
+
+        my $url = $rel->target->url->as_string;
+
+        # Use the Wikidata RDF concept URIs rather than the default site URL (MBS-9987)
+        $url =~ s/^https?:\/\/(?:www\.)?wikidata\.org\/wiki\//http:\/\/www\.wikidata\.org\/entity\//;
```
mwiencek, line 94: For the future, you can also use a different delimiter than `/` so that you don't have to escape it 8 times. :)

mwiencek (approved):

lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Role/SameAs.pm
reosarevok: I didn't actually know that was possible, whoops :D

PR #1939, opened 2021-02-25
MBS-9884: Use "Created" for character-type artist begin dates

reosarevok:
### Implement MBS-9884

The documentation says "Begin date represents the date (in real life) when the character concept was created." As such, "Created" is the best option for this, much better than "Begin date".

The docs also say "The End date should not be set", so no point setting a specific label for that.


mwiencek (approved):

yvanzo (approved):
:ship:

PR #1940, opened 2021-02-25
MBS-9876: Stop redirecting tag search to direct search

reosarevok:
### Implement MBS-9876

We support searching for tags with SOLR, so we should probably allow doing that. Because the response does not include tag ID, we just load the tag based on the returned name.

For what it's worth, the results are currently significantly worse than with direct search. We might want to try and improve that.
Screenshot from 2021-02-25 21-12-18


yvanzo (commented):
There seems to be no point of using indexed search here if it returns worse results.
Made a comment to the ticket and setting this pull request as draft for now.

reosarevok:
This depends on https://github.com/metabrainz/mbsssss/pull/60 being merged (and released) first.

PR #1941, opened 2021-02-25
MBS-9871: Display all useful series relationships in RelatedSeries

reosarevok:
### Implement MBS-9871

The only thing we want to avoid here is showing all the part-of-series rels on the page for every entity in a series. But for that there is no need to limit the rels shown to URLs - we have a perfectly effective `isNotSeriesPart` function we already use for series pages.

### DIFF
--- a/root/components/RelatedSeries.js
+++ b/root/components/RelatedSeries.js
@@ -13,9 +13,8 @@ import EntityLink from '../static/scripts/common/components/EntityLink';
 import linkedEntities from '../static/scripts/common/linkedEntities';
 import groupRelationships from '../utility/groupRelationships';
 
+import {isNotSeriesPart} from './Relationships';
 import StaticRelationshipsDisplay from './StaticRelationshipsDisplay';
-
-const targetEntityTypes = ['url'];
 
 type Props = {
   +seriesIds: $ReadOnlyArray<number>,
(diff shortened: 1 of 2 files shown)


mwiencek:
This might not matter right now, but wouldn't it make more sense to use the cardinality info?

root/components/RelatedSeries.js
```diff
@@ -38,7 +37,18 @@ const RelatedSeries = ({seriesIds}: Props): React.MixedElement => {
...
+            undefined,
+            (
+              relationship: RelationshipT,
+            ) => {
+              if (isNotSeriesPart(relationship)) {
+                return true;
+              }
+              return false;
+            },
+          )
```
mwiencek, line 51: Unless we foresee extending this soon, you can also just pass the function directly AFAICT.
```suggestion
            isNotSeriesPart,
```

reosarevok:
> wouldn't it make more sense to use the cardinality info?

Maybe? We don't seem to have anything right now that checks cardinality on the JS side... which is a bit weird tbh?

mwiencek (approved):

reosarevok:
Merging as it is for now because we actually have 0-0 cardinality for part of series rels anyway.

PR #1943, opened 2021-02-26
MBS-9857: Make merging HTTP->HTTPS versions of URL an auto edit

reosarevok:
### Implement MBS-9857

We generally want to make it easy to turn HTTP URLs into HTTPS ones. We already had code to allow auto-editing an URL if the change was only HTTP->HTTPS and it did not cause a merge.
But in fact, there's no reason not to allow this also for a merge, since the only reason such a change is incorrect is if the HTTPS URL is wrong, in which case the URL can easily be changed back to HTTP post-merge (there's no case I can imagine where we want to have both HTTP and HTTPS versions and keep some relationships using one version and some using the other).


lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
```diff
@@ -122,8 +122,11 @@ role {
                 $url_changes = $self->edit_relationships($c, $form, \@urls, $source);
             }
 
-            my @rels = grep { !$_->is_empty } $form->field('rel')->fields;
-            my $rel_changes = $self->edit_relationships($c, $form, \@rels, $source);
+            my $rel_changes = 0;
+            if (defined $source) {
+                my @rels = grep { !$_->is_empty } $form->field('rel')->fields;
+                $rel_changes = $self->edit_relationships($c, $form, \@rels, $source);
+            }
```
reosarevok, line 129: $source is not defined after the URL has been merged away. This change seems to work, but if there's a better option I am missing, do let me know! :)

yvanzo (approved):

lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
mwiencek: What if both the URL change + relationship changes are submitted? Are they not propagated to the merged URL?

mwiencek (changes requested):

lib/MusicBrainz/Server/Edit/URL/Edit.pm
```diff
@@ -85,7 +86,8 @@ after initialize => sub {
 override allow_auto_edit => sub {
     my $self = shift;
     return 0 unless defined $self->data->{old}{url} && defined $self->data->{new}{url};
-    return ($self->data->{old}{url} =~ s/^http:/https:/r) eq $self->data->{new}{url};
+    $self->data->{is_https_change_only} = ($self->data->{old}{url} =~ s/^http:/https:/r) eq $self->data->{new}{url} ? 1 : 0;
+    return $self->data->{is_https_change_only};
```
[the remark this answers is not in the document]
reosarevok: I had it as a function first, but then thought that it seemed better to just store it and not call the same function several times. But if that seems better :)

lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
reosarevok: Hmm, true, I tested now and that would drop that rel change. Sigh. Thanks for catching it. Seems like this only works right now because we don't auto-apply any merges, so the relationship changes are entered on time... Looking into it.

Another thing that seems weird is that these are not run in a transaction or anything - you can have the edit changing the URL pass, then an ISE happen when trying to change rels, and the URL change will still have applied.

reosarevok: I can get this to work by adding this inside `if ($edit)` above (updating the PR since it seems like an improvement):

```
                # If a URL got merged away, $source will be undefined now,
                # so we should follow the gid redirect to find the new entity
                # whose rels should change.
                if ($form_name eq "edit-url" && !$source) {
                    my $old_url_gid = $edit->{data}->{entity}->{gid};
                    $source = $c->model('URL')->get_by_gid($old_url_gid);
                }
```

But that seems to still cause issues for this case: 

URL1 `http://example.com` has "picture for: X"
URL2 `https://example.com` has "picture for: X"
URL1 is edited so that the URL is now `https://example.com` and the rel is now "homepage for: X"
The merge goes through first, but then the "picture for: X" rel that was being changed got merged with that of URL2. As such, the change to make it "homepage for: X" is silently dropped, because the rel it was meant to change is now gone.
I guess this is only a problem for URLs because no other entities can be merged *and* have their rels edited at the same time.
What should we do with that case?

mwiencek: What if we processed the rel edits on the source before doing the merge? Seems like the merge should do the correct thing in that case if they're all submitted as auto-edits at least.

reosarevok:
Converting to draft because IIRC this needs careful rechecking and possibly fixing.

PR #1944, opened 2021-02-26
MBS-10416: Remove invalid characters from edited annotations

yvanzo:
# Problem

MBS-10416: Annotation with some control characters cannot be indexed for search

# Solution

Strip such characters (and other invalid characters) when entered in annotation.

# Checklist for author

* [x] Add test

# Action

1. Clean the 216 existing annotations having such invalid characters.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -365,9 +365,10 @@ sub trim_comment {
 sub trim_multiline_text {
     my $t = shift;
 
-    # Not trimming starting spaces to avoid breaking list formatting,
-    # consider trimming again once this uses Markdown
-    $t =~ s/\s+$//;
+    # Not trimming starting spaces to avoid breaking
+    # either list formatting in Wikitext
+    # or block in Markdown.
+    $t =~ s/\s+$//gm;
 
     return $t;
 }
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -334,6 +334,7 @@ sub sanitize {
     # into U+0020 (or else they'll be removed).
     $t = collapse_whitespace($t);
     $t = remove_invalid_characters($t);
+    $t = remove_lineformatting_characters($t);
     $t = remove_direction_marks($t);
     # Collapse spaces again, since characters may have been removed.
     $t = collapse_whitespace($t);
@@ -397,17 +398,23 @@ my $noncharacter_pattern = '\x{FDD0}-\x{FDEF}\x{FFFE}\x{FFFF}';
 
 sub remove_invalid_characters {
     shift
-    # trim XML-invalid characters
+    # trim XML-invalid characters, among them:
+    # - Other, surrogate (which are UTF-16, not valid UTF-8)
     =~ s/[^\x09\x0A\x0D\x20-\x{D7FF}\x{E000}-\x{FFFD}\x{10000}-\x{10FFFF}]//gor
     # trim other undesirable characters:
-    # - zwsp
-    # - shy
     # - bom
-    # - Other, control
-    # - Other, surrogate
     # - Supplementary private use areas
     # - Noncharacters
-    =~ s/[\x{200B}\x{00AD}\x{FEFF}\p{Cc}\x{F0000}-\x{FFFFF}\x{100000}-\x{10FFFF}${noncharacter_pattern}]//gr
+    =~ s/[\x{FEFF}\x{F0000}-\x{FFFFF}\x{100000}-\x{10FFFF}${noncharacter_pattern}]//gr
+}
+
+sub remove_lineformatting_characters {
+    shift
+    # trim lasting line-formatting characters:
+    # - zwsp
+    # - shy
+    # - Other, control (including TAB \x09, LF \x0A, and CR \x0D)
+    =~ s/[\x{200B}\x{00AD}\p{Cc}]//gr
 }
 
 sub type_to_model
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -76,6 +76,7 @@ our @EXPORT_OK = qw(
     take_while
     trim
     trim_comment
+    trim_multiline_text
     type_to_model
     split_relationship_by_attributes
 );
@@ -359,6 +360,16 @@ sub trim_comment {
     $t =~ s/^\s*\(([^()]+)\)\s*$/$1/;
 
     return trim($t);
+}
+
+sub trim_multiline_text {
+    my $t = shift;
+
+    # Not trimming starting spaces to avoid breaking list formatting,
+    # consider trimming again once this uses Markdown
+    $t =~ s/\s+$//;
+
+    return $t;
 }
 
 sub remove_direction_marks {
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -365,6 +365,8 @@ sub trim_comment {
 sub trim_multiline_text {
     my $t = shift;
 
+    $t = NFC($t);
+    $t = remove_invalid_characters($t);
     # Not trimming starting spaces to avoid breaking
     # either list formatting in Wikitext
     # or block in Markdown.
--- a/t/lib/t/MusicBrainz/Server/Controller/Work/AddAnnotation.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/Work/AddAnnotation.pm
@@ -19,7 +19,7 @@ $mech->submit_form( with_fields => { username => 'new_editor', password => 'pass
 $mech->get_ok("/work/745c079d-374e-4436-9448-da92dedef3ce/edit_annotation");
 $mech->submit_form(
     with_fields => {
-        'edit-annotation.text' => "    * Test annotation for a work\r\n    * This annotation has two bullets  \t\t",
+        'edit-annotation.text' => "    * Test annotation for a work  \r\n    * This annotation has two bullets  \t\t",
         'edit-annotation.changelog' => 'Changelog here',
     }
 );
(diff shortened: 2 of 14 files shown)


lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -364,6 +364,8 @@ sub trim_comment {
 sub trim_text {
     my $t = shift;
 
+    $t = NFC($t);
+    $t = remove_invalid_characters($t);
```
reosarevok, line 368: The other place we use this has a comment:
```
    # Before removing invalid characters, convert space control characters
    # into U+0020 (or else they'll be removed).

```
Is this a worry here?

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -360,6 +361,16 @@ sub trim_comment {
     return trim($t);
 }
 
+sub trim_text {
+    my $t = shift;
+
+    # Not trimming starting spaces to avoid breaking list formatting,
+    # consider trimming again once this uses Markdown
```
reosarevok, line 370: "once this uses Markdown" is less clear in here than in Form::Annotation. Should this be called `trim_annotation` and say "once annotations use Markdown"?

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: That comment is not accurate enough, it does not give any example of such character. For example, TAB character is not removed, so I’m not sure which “space control characters” would actually benefit from that.

The other place collapses whitespaces, which we cannot do here because of possible indentation.

It seems neither needed nor appropriate to replace unwanted control characters such as bell with whitespace.

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: Annotation has a specific syntax for now but there is nothing specific to annotation’s syntax in this function.
It could (or should?) also be used for other text fields such as editor’s bio, edit notes.
Eventually, all of these fields will move to Markdown syntax.

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: It's talking about any character removed by `remove_invalid_characters` -- there's an explicit list in there (which includes TAB, LF, CR).

Were you perhaps testing this in the release editor? Because that goes through `WS::js::Edit`, not `Form::Annotation`. If you test on any `/edit_annotation` page, this patch will indeed cause it to remove TAB, LF, CR, etc.

(This also suggests we'll have to call `trim_text` where annotations are processed in `WS::js::Edit`.)

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: The only issue I have is that it's not distinguished from `trim` (which also operates on text). So I wouldn't know what the difference between `trim` and `trim_text` are from the names alone, I'd have to check usages and git history. What about something like `trim_wikitext`?

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: > there's an explicit list in there (which includes TAB, LF, CR).

This was not clear because [`remove_invalid_characters`](https://github.com/metabrainz/musicbrainz-server/blob/v-2021-03-01/lib/MusicBrainz/Server/Data/Utils.pm#L398-L411) currently starts by explicitly NOT removing TAB, LF, CR with `s/[^\x09\x0A\x0D…]//gor` but it ends with removing them with `s/[…\p{Cc}…]//g`.

This function will be split using commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/2bc4060b8f1abd524b95f2d035546273ee0ff119.

> we'll have to call `trim_text` where annotations are processed in `WS::js::Edit`.

Fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/dbea2f2446aad6aebf8e388d6b4025fe3c268247.

It’s a bit an issue that data processors in `WS::js::Edit` are not systematically synced with their `Form` counterparts. What can we do about that, commenting code with reminders?

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: Renamed as `trim_multiline_text` in commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/2b4597fa01406cc296545c4bbbcd0fd6a1c8b647 since it can be used for editor’s biography, edit note, event setlist, description… which are not necessarily wikitext.

yvanzo:
Refactored more code, fixed bugs with TAB, CR, LF and release editor, added/updated tests accordingly.
Also made removing spaces from the end of *every* line through commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/9dca2172d2bf5e1a524408a568bc827815a0de33. (That can be added to MBS-10416’s description.)

mwiencek:
> “Other, surrogate” characters were/are actually not removed.

Oops, that "Other, surrogate" comment was misleading 'cause they weren't removed below it. They are actually removed by the "trim XML-invalid characters" regular expression (which is good, because they aren't valid UTF-8).

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: Comments are probably the best we can do for now until more significant improvements can be made. If I could go back, I'd probably make /ws/js/edit submit everything through FormHandler and just serialize the field errors it returns.

mwiencek (approved):
LGTM but I'd clarify the commit message about surrogate chars.

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -400,14 +415,19 @@ sub remove_invalid_characters {
...
     # trim other undesirable characters:
-    # - zwsp
-    # - shy
     # - bom
-    # - Other, control
-    # - Other, surrogate
     # - Other, private use
     # - Noncharacters
-    =~ s/[\x{200B}\x{00AD}\x{FEFF}\p{Cc}\p{Co}${noncharacter_pattern}]//gr
+    =~ s/[\x{FEFF}\p{Co}${noncharacter_pattern}]//gr
```
mwiencek, line 421: I was going to suggest removing `\p{Co}` here to fix MBS-10976, but I submitted that separately instead (https://github.com/metabrainz/musicbrainz-server/pull/1964) so that I could additionally fix the ISE and add tests.

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: Reported as ticket MBS-11428 and added comments through commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/aee75730894e3c0bc096b5f0603b51a881afca31.

yvanzo:
Thanks, I clarified commit message and commented code about surrogate chararacters in commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/374ae51801b7dbebb34af05d1ffcd49876afb007.

yvanzo:
@brainzbot retest this please

yvanzo:
@brainzbot retest this please

yvanzo:
Rebased on `master` and resolved conflicts with merged #1964.

t/lib/t/MusicBrainz/Server/Controller/Work/AddAnnotation.pm
```diff
@@ -33,7 +33,7 @@ is_deeply(
             id => 1,
             name => 'Dancing Queen'
         },
-        text      => 'Test annotation 6. This is my annotation',
+		    text => "    * Test annotation for a work\r\n    * This annotation has two bullets",
```
reosarevok: Unintended extra spacing?

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -397,17 +398,23 @@ my $noncharacter_pattern = '\x{FDD0}-\x{FDEF}\x{FFFE}\x{FFFF}';
...
-    # - Other, surrogate
     # - Supplementary private use areas
     # - Noncharacters
-    =~ s/[\x{200B}\x{00AD}\x{FEFF}\p{Cc}\x{F0000}-\x{FFFFF}\x{100000}-\x{10FFFF}${noncharacter_pattern}]//gr
+    =~ s/[\x{FEFF}\x{F0000}-\x{FFFFF}\x{100000}-\x{10FFFF}${noncharacter_pattern}]//gr
+}
+
+sub remove_lineformatting_characters {
+    shift
+    # trim other undesirable characters:
```
reosarevok, line 427: I'd suggest making this comment more clear - I expect this is not "other undesirable characters" but "line formatting characters"? :)

lib/MusicBrainz/Server/Data/Utils.pm
reosarevok: Then I'd suggest "consider trimming again once text fields currently using wikitext start using Markdown" or whatnot maybe?

reosarevok (approved):

lib/MusicBrainz/Server/Data/Utils.pm
reosarevok: Actually I see you changed this further down the line so it's probably fine :)

yvanzo: Yes, even in Markdown we may want to preserve leading spaces.

t/lib/t/MusicBrainz/Server/Controller/Work/AddAnnotation.pm
yvanzo: Thanks, these were unintended tab chars. Removed from commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/5b36a92215f127602051b90b500827b6c269b9b7.

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: Specified “lasting line-formatting characters” in commit https://github.com/metabrainz/musicbrainz-server/pull/1944/commits/62348af4bfaa6d665ef5a76e7d945c53199ea37f because some have been removed already.

PR #1945, opened 2021-02-26
MBS-9826: Show more realistic edit closing time overviews

reosarevok:
### Implement MBS-9826

It's quite misleading to say "closes in 6 days" as soon as an edit is entered, given that it's really 6 days, 23 hours and 50-something minutes. Yes, we have a tooltip with the date, but still.

Instead of using floor, this uses round - it seems more sensible to say 7 days when it's more than 6 and a half, etc.

I start displaying hours once they hit 23 (rounded, so at 23 h and 30 minutes left), rather than still saying 1 day until there's only 12 hours left like elsewhere.

### DIFF
--- a/root/components/VotingPeriod.js
+++ b/root/components/VotingPeriod.js
@@ -28,11 +28,11 @@ const VotingPeriod = ({
 
   if (date > now) {
     const durationSeconds = Math.floor((date - now) / 1000);
-    const durationMinutes = Math.floor(durationSeconds / 60);
-    const durationHours = Math.floor(durationMinutes / 60);
-    const durationDays = Math.floor(durationHours / 24);
+    const durationMinutes = Math.round(durationSeconds / 60);
+    const durationHours = Math.round(durationMinutes / 60);
+    const durationDays = Math.round(durationHours / 24);
 
-    if (durationDays > 0) {
+    if (durationHours > 23) {
       return exp.ln(
         `Closes in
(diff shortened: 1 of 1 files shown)


yvanzo (commented):
It seems it defeats the whole point of having a short abstract duration.
Making an alternative proposition in comment to the ticket.

root/components/VotingPeriod.js
```diff
@@ -28,11 +28,11 @@ const VotingPeriod = ({
 
   if (date > now) {
     const durationSeconds = Math.floor((date - now) / 1000);
-    const durationMinutes = Math.floor(durationSeconds / 60);
-    const durationHours = Math.floor(durationMinutes / 60);
-    const durationDays = Math.floor(durationHours / 24);
+    const durationMinutes = Math.round(durationSeconds / 60);
+    const durationHours = Math.round(durationMinutes / 60);
```
mwiencek, line 32: So does 31 minutes get rounded up to 1 hour? (If there's less than an hour left, I'd always show minutes.)

reosarevok: Good point. Changed the check to minutes > 59 below.

mwiencek (approved):

mwiencek:
This matches yvanzo's suggestion in the ticket except for only showing hours when it's less than 24 (which I think I prefer) rather than 48.  It seems ready to test on beta, in any case.

PR #1947, opened 2021-03-01
compareDates / compareDatePeriods changes

mwiencek:
This is cherry-picked from my relationship editor branch. I'm not sure yet how to describe this in a ticket, if it warrants one, because there's no actual issue currently.

See the commit message for details.

### DIFF
--- a/root/static/scripts/common/utility/compareDates.js
+++ b/root/static/scripts/common/utility/compareDates.js
@@ -9,30 +9,39 @@
 
 /* eslint-disable sort-keys */
 const NULL_DATE: PartialDateT = Object.freeze({
-  year: -Infinity,
-  month: 1,
-  day: 1,
+  year: null,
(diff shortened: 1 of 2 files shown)


mwiencek:
> compareDates previously returned 0 if both dates had a null year. This is changed so that it now compares by month (or day if both months are null) in that case. This behavior is needed by the new relationship editor, where relationship comparison functions should only return 0 if both values are exactly equal.

The old behavior wasn't *wrong* because you can't say what the relative sort order of unknown dates are. But sorting them deterministically isn't wrong in that case, either. And as I mention, the new relationship editor code relies on these sorts of functions to determine if relationships have been modified in any way, are duplicates of each other, etc.

yvanzo (approved):

reosarevok (approved):

PR #1948, opened 2021-03-01
MBS-9695: Allow editing LinkAttributeType creditable/free_text via UI

reosarevok:
### Implement MBS-9695

Since adding/removing free_text would probably make a mess of existing attribute uses, we disable that if the attribute is in use.
Same for removing creditable (since we wouldn't want to accidentally break existing attribute credits) - adding creditable should always be safe unless I'm missing something.

Tested by hand by adding and removing them locally from newly created attributes, and by trying to remove them from existing ones (Task should be in use in sample data).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
+++ b/lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
@@ -7,6 +7,8 @@ use MusicBrainz::Server::Constants qw(
     $INSTRUMENT_ROOT_ID
 );
 use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
+use MusicBrainz::Server::Translation qw( l );
+
 use MusicBrainz::Server::Validation qw( is_guid );
 
 with 'MusicBrainz::Server::Controller::Role::Load' => {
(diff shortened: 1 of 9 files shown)


lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
```diff
@@ -77,24 +78,48 @@ sub edit : Chained('load') RequireAuth(relationship_editor)
...
     my $form = $c->form( form => 'Admin::LinkAttributeType', init_object => $link_attr_type );
 
+    my $attribute_in_use = $c->model('LinkAttributeType')->in_use($link_attr_type->id);
+
+    if ($attribute_in_use) {
+        $form->field('free_text')->disabled(1);
+        if ($link_attr_type->creditable) {
+            $form->field('creditable')->disabled(1);
+        }
+    }
```
reosarevok, line 88: This disables the field so that it is ignored in the backend, but the checkbox still shows undisabled in the frontend. Not sure if we have a good way of changing that with TT. Since this is for admins only, it might not matter a ton, but if you have ideas, I'd love to improve it.

mwiencek: It seems fixable by adding `disabled => $field->disabled,` to `_render_input` in FormRenderer. Want to push a commit for that?

lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
```diff
@@ -77,24 +78,48 @@ sub edit : Chained('load') RequireAuth(relationship_editor)
...
-                    child_order => $link_attr_type->child_order,
+        my %values = map { $_->name => $_->value } $form->edit_fields;
+
+        my $removes_creditable = $link_attr_type->creditable && !($values{'creditable'});
+        my $adds_free_text = $values{'free_text'} && !($link_attr_type->free_text);
+        my $removes_free_text = $link_attr_type->free_text && !($values{'free_text'});
+        if ($attribute_in_use && $removes_creditable) {
+            $form->field('creditable')->add_error(l('The attribute is in use, and this change would risk damaging existing values'));
+        } elsif ($attribute_in_use && ($adds_free_text || $removes_free_text)) {
+            $form->field('free_text')->add_error(l('The attribute is in use, and this change would risk damaging existing values'));
```
mwiencek, line 99: If you disable the checkboxes, this shouldn't be possible, so we can probably remove the error messages. But I do think we should perform these checks in `Data::LinkAttributeType::update` and `die` if they occur.

reosarevok:
Made the suggested changes, but on testing this, I found that no edit is created and no change gets made when setting "creditable" to an attribute that doesn't have it (like "step"). In fact, it seems like any change to any attribute does nothing. Clearly something got borked somewhere, but I can't find the issue - @mwiencek, maybe you can help me figure it out?

lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
```diff
@@ -99,24 +101,38 @@ sub edit : Chained('load') RequireAuth(relationship_editor)
...
+                    free_text => $link_attr_type->free_text,
                 }
             );
-        });
 
-        $c->response->redirect($c->uri_for_action('relationship/linkattributetype/show', [ $link_attr_type->gid ]));
-        $c->detach;
+            if ($edit) {
+                $c->response->redirect($c->uri_for_action('relationship/linkattributetype/show', [ $link_attr_type->gid ]));
+                $c->detach;
```
mwiencek, line 133: I see now, `detach` throws an exception (which is normal, that's how it works), which rolls back the transaction.  So we should leave the redirect code outside of `with_transaction`.  You can still declare `my $edit;` outside of `with_transaction` and do the assignment inside.

mwiencek: (The detach shouldn't be needed in either case -- though I'd still move the redirect outside.)

reosarevok: Thanks! Seems to work, would never have figured it out, sigh :)

yvanzo (approved):
LGTM but did not test

PR #1949, opened 2021-03-01
Add Flow helper to check the exhaustiveness of switch statements

mwiencek:
Calling this imaginary function from a 'flow-include' comment will cause a switch statement to error if you haven't handled all possible cases. This is especially useful for reducer functions that have many possible actions.

See https://flow.org/en/docs/tips/switch-statement-exhaustiveness/


mwiencek:
Yes, updated. At the time I wrote this patch I don't think that form was written yet. :)

yvanzo (commented):
:+1:

PR #1950, opened 2021-03-02
MBS-11419: Show YouTube Music links on the sidebar

reosarevok:
### Implement MBS-11419

Not adding autoselect because it seems to often be free streaming, but sometimes paid streaming.


mwiencek (approved):

root/static/scripts/common/constants.js
```diff
@@ -195,8 +196,8 @@ export const FAVICON_CLASSES = {
   'wikisource.org': 'wikisource',
   'worldcat.org': 'worldcat',
   'www.amazon': 'amazon',
+  'www.youtube.com': 'youtube',
   'www5.atwiki.jp/hmiku/': 'hmikuwiki',
-  'youtube.com': 'youtube',
 };
```
yvanzo, line 201: FWIW YouTube URLs are normalized with `www` subdomain since 2011 (commit f5a45e4779140b0f60e83bb0c114b35379873019) but I still searched for YouTube URLs without `www` as that change would degrade their display and found two of them, to be fixed soon:
* https://musicbrainz.org/edit/77961096
* https://musicbrainz.org/edit/77961097

yvanzo (approved):
:ship:

PR #1951, opened 2021-03-02
MBS-9836: Guess Case: stop uppercasing "the" in artist names

reosarevok:
### Implement MBS-9836

There's no documentation about why this was done this way in the first place, but most likely it was because of pre-NGS artists which would have two or more different, full artist names as part of a collaboration artist name. This is fairly uncommon with NGS, while artists which just have "the" in their names (think "Tyler, the Creator") and where "the" should be lowercase are reasonably common.

I asked the community and most feedback suggested dropping this unexplained / unexpected difference between artist guess case and guess case for other entities.

### DIFF
--- a/root/static/scripts/tests/GuessCase.js
+++ b/root/static/scripts/tests/GuessCase.js
@@ -114,8 +114,8 @@ test('Artist', function (t) {
     },
     {
       input: 'Peggy Sue And The Pirates',
-      expected: 'Peggy Sue and The Pirates',
-      bug: 'MBS-1370',
+      expected: 'Peggy Sue and the Pirates',
+      bug: 'MBS-1370 / MBS-9836',
       mode: 'Artist',
     },
   ];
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/GuessCase.js
```diff
@@ -114,8 +114,8 @@ test('Artist', function (t) {
     },
     {
       input: 'Peggy Sue And The Pirates',
-      expected: 'Peggy Sue and The Pirates',
-      bug: 'MBS-1370',
+      expected: 'Peggy Sue and the Pirates',
+      bug: 'MBS-1370 / MBS-9836',
```
reosarevok, line 118: I thought "oh, MBS-1370 will explain specifically why we uppercased The and whether it was pre-NGS", but no, that was about not uppercasing "And".

mwiencek (approved):

yvanzo (approved):
:ship:

PR #1952, opened 2021-03-02
MBS-8621: Reset openedSingleQuote for every new guess

reosarevok:
### Fix MBS-8621

There's no situation I can think of where a single-quote block would open in one call of guess() and close in another, so this resets openedSingleQuote in init().
This was causing a bug where forceCaps would be set if a track in a release contained a word ending in apostrophe (e.g. Tryin') and a previous track contained a word starting with an apostrophe (e.g. 'em).


mwiencek (approved):

mwiencek:
Want to add a test? :)

mwiencek (approved):

yvanzo (approved):
:ship:

PR #1953, opened 2021-03-03
MBS-1178: Relationships table paging

mwiencek:
https://tickets.metabrainz.org/browse/MBS-1178

Adds paging to all relationships tables, currently on the these URLs:

 * /area/\<mbid>/recordings
 * /area/\<mbid>/releases
 * /area/\<mbid>/works
 * /artist/\<mbid>/relationships
 * /label/\<mbid>/relationships
 * /place/\<mbid>/performances
 * /work/\<mbid>

Relationships are paged by link type. (Or more accurately, link type + direction -- it just so happens that no relationships between the same type need paging.) So by default, the initial table will show 25 relationships of each type. If there are more than 25 for a particular section, a link to see all of them is made available.

I tried not to touch the existing relationships implementation too much at the Data or Entity layers -- there's simply too much to break. Instead, I've forked parts of `Data::Relationship::_load` into a new method named `load_paged`. It doesn't push relationships into the standard entity `relationships` array, though. It adds them to a `paged_relationship_groups` hash on the entity, grouped by target type. This is further grouped by link type, until an `Entity::RelationshipLinkTypeGroup` instance is reached. This instance holds the relationships array for the group, plus all the necessary context needed for paging: `total_relationships`, `limit`, `offset`, etc.

`Controller::Role::Load` has been extended to support a `paged_subset` option, which works like the current `subset`, but calls the aforementioned `Data::Relationship::load_paged` instead. We also handle the additional query parameters needed for paging here, and if these are detected, we ensure that only one specific link type group is loaded.

The `RelationshipsTable` component has been largely refactored to make use of the new `paged_relationship_groups` hash and to adjust the display if a particular type is being viewed.


reosarevok:
Seems this breaks JSON-LD tests, probably because some rels aren't loaded that should be to get the correct JSON-LD data.

lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -33,6 +35,8 @@ parameter 'allow_integer_ids' => (
     default => sub { 1 },
 );
 
+Readonly our $RELATIONSHIP_PAGE_SIZE => 100;
```
reosarevok, line 38: My original suggestion was "Also, most relationship tables do not need to be paginated at all. I'd set a very high limit, 500 to 1000, if that is doable." - that said, maybe that's impractical and 100 is fine, especially now that I have a separate PR adding work filtering by name which effectively would allow me to replace my main requirement why I needed the composers' relationships tab to be easily searchable :) 

It would be nice to be able to filter rels eventually too, but it might be hard to do now :)

mwiencek: I'd definitely want you to confirm that this doesn't break any use case before we merge it (so perhaps we should get your work filtering PR in at the same time). But I'd agree that filtering is the answer rather than outputting as much data as we can, 'cause that's precisely why our pages are so slow. I get that it's better to load more at once if the user is going to view all 20 pages anyway, but filtering is still the ideal solution for that.

mwiencek:
> Seems this breaks JSON-LD tests, probably because some rels aren't loaded that should be to get the correct JSON-LD data.

Turns out I was just passing the wrong args to a function. Fixed now. :)

reosarevok (changes requested):
See failing test - all recordedAs are gone from works' JSON-LD. Tested also in a real page with http://localhost:5000/work/d94d2e8b-78d3-3d2f-a71b-b60a7dfa2660

mwiencek:
> See failing test - all recordedAs are gone from works' JSON-LD. Tested also in a real page with http://localhost:5000/work/d94d2e8b-78d3-3d2f-a71b-b60a7dfa2660

Fixed this, though it now of course only outputs the 25 recordings that were loaded. I added an additional sitemap in 22ee2d752721e06cbed5b38f88c4c346f0d4e92c to cover the rest.

mwiencek:
Update the commit message in the last commit to refer to the correct commit hash, which broke after a rebase.

mwiencek:
@brainzbot test this please

yvanzo (approved):
Manually tested with entities currently causing the most issues:
* http://localhost:5000/area/74e50e58-5deb-4b99-93a2-decbb365c07f/recordings
* http://localhost:5000/artist/38712b4c-0fd4-4c65-8c7a-45676fecc973/relationships
* http://localhost:5000/artist/38712b4c-0fd4-4c65-8c7a-45676fecc973/relationships?direction=1&link_type_id=150&page=1

Tests coverage is quite good otherwise :)

:ship:

PR #1954, opened 2021-03-03
MBS-8232: Support more Unicode quotation marks in guess case

reosarevok:
### Fix MBS-8232

This adds support for most Unicode quotation marks, to avoid the word just after the quotation mark being lowercased.

### DIFF
--- a/root/static/scripts/tests/GuessCase.js
+++ b/root/static/scripts/tests/GuessCase.js
@@ -209,7 +209,7 @@ test('Work', function (t) {
     },
     {
       input: 'acte 1, no. 7: chœur: «voyons brigadier»',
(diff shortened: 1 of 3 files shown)


root/static/scripts/tests/GuessCase.js
```diff
@@ -209,7 +209,7 @@ test('Work', function (t) {
     },
     {
       input: 'acte 1, no. 7: chœur: «voyons brigadier»',
-      expected: 'Acte 1, no. 7 : Chœur : « voyons brigadier »',
+      expected: 'Acte 1, no. 7 : Chœur : « Voyons brigadier »',
```
reosarevok, line 212: This broke with this change, but I'm fairly sure a classical title like this should be capitalized to begin with and the previous test was bad.

mwiencek (approved):

PR #1955, opened 2021-03-04
MBS-8678 / MBS-8866 / MBS-11421: Split artist improvements

reosarevok:
### Fix MBS-8678 and implement MBS-8866 / MBS-11421

See commit messages

Tested manually by visiting an artist's split page while transitioning it through various states (fully empty, only one event rel, only one recording credit, both event rels and recording credits).

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -289,6 +289,42 @@ sub update
     $self->sql->update_row('artist', $row, { id => $artist_id }) if %$row;
 }
 
+sub can_split
+{
+    my ($self, $artist_id) = @_;
+    return 0 if is_special_artist($artist_id);
+
+    # Can only split if they have no relationships at all other than collaboration
+    # These AND NOT EXISTS clauses are ordered by my estimated likelihood of a 
+    # relationship existing for a collaboration, as postgresql will not execute
+    # the later clauses if an earlier one has already excluded the lone artist row.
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -289,6 +289,42 @@ sub update
...
+{
+    my ($self, $artist_id) = @_;
+    return 0 if is_special_artist($artist_id);
+
+    # Can only split if they have no relationships at all other than collaboration
+    # These AND NOT EXISTS clauses are ordered by my estimated likelihood of a 
+    # relationship existing for a collaboration, as postgresql will not execute
+    # the later clauses if an earlier one has already excluded the lone artist row.
+    my $can_split = $self->sql->select_single_value(
+        "SELECT TRUE FROM artist WHERE id = ? " .
```
mwiencek, line 302: We can use `<<~EOSQL` to make this a bit nicer now

root/artist/split.tt
```diff
@@ -36,6 +36,15 @@
          <div id="artist-credit-editor"></div>
        </fieldset>
 
+       [% IF collaborators.size %]
+         <h3>[% add_colon(l('Collaborators on this artist')) %]</h3>
```
mwiencek, line 40: I wonder if it'd be better to show this first, before the AC editor.

mwiencek (approved):

root/artist/split.tt
reosarevok: I think so, made the change.

PR #1956, opened 2021-03-04
MBS-9581: Display relationship credits in EventLocations

reosarevok:
### Fix MBS-9581

Effectively MBS-10223 but for areas and places (EventLocations).


mwiencek (approved):

yvanzo (approved):
:shipit:

PR #1959, opened 2021-03-08
MBS-11424: Unbreak seeding multiple RG types

mwiencek:
We weren't accounting for arrays in PostParameters.js. Now we are.

Tested the change manually by seeding three types and observing that they were all selected. (If it helps, you can force the ConfirmSeed page locally by removing the `$c->is_cross_origin` check, then adding it back once you're ready to submit.)

### DIFF
--- a/root/components/PostParameters.js
+++ b/root/components/PostParameters.js
@@ -13,7 +13,7 @@ import {compareStrings} from '../static/scripts/common/utility/compare';
 import hydrate from '../utility/hydrate';
 
 export type PostParametersT = {
-  +[param: string]: string,
+  +[param: string]: string | $ReadOnlyArray<string>,
   ...
 };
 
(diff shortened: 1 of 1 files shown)


root/components/PostParameters.js
```diff
@@ -28,9 +28,12 @@ const PostParameters = ({
 }: PropsT): React.MixedElement => {
   const [expanded, setExpanded] = React.useState(false);
 
-  const sortedParams = Object.entries(params).sort(
-    (a, b) => compareStrings(a[0], b[0]),
-  );
+  const sortedParams:
+    $ReadOnlyArray<[string, string | $ReadOnlyArray<string>]> =
```
reosarevok, line 32: Remind me again, what did this type mean here? (sorry, I'm pretty sure I've asked this before)

reosarevok:
https://musicbrainz.org/doc/Development/Release_Editor_Seeding#Release_group_data seems crazy out of date, we should probably update it. Will look into it.

reosarevok:
Improved the documentation for this (https://wiki.musicbrainz.org/index.php?title=Development%2FRelease_Editor_Seeding&type=revision&diff=75236&oldid=73612) and the actual test (https://github.com/metabrainz/musicbrainz-server/pull/1961)

root/components/PostParameters.js
mwiencek: An array of tuples. `Object.entries` returns an array of `[key, value]` items, so in this case the key is `string` and the value is `string | $ReadOnlyArray<string>`.

reosarevok (approved):
LGTM. Tested via seed-love-bug (with improvements from https://github.com/metabrainz/musicbrainz-server/pull/1961).

yvanzo (approved):
:ship:

PR #1961, opened 2021-03-08
MBS-11427: Improve seed-love-bug test/showcase

reosarevok:
### Implement MBS-11427

This is not particularly good as a showcase of fields to seed. Not only is it missing several available fields (and not showing usage of most of the possible multi-option combos) but it has no useful labeling for fields and no placeholders, making it very hard to even understand what's going on (several MBID fields are listed one after another with no info on what MBIDs are being seeded!).

This makes it so that every field or related set of fields has a proper legend, and the most confusing ones also have a placeholder specifying what should (or can) be entered there.

### DIFF
--- a/root/static/tests/seed-love-bug.html
+++ b/root/static/tests/seed-love-bug.html
@@ -18,46 +18,95 @@
 
       <fieldset>
         <legend>Release information</legend>
+        <label>Release name:</label>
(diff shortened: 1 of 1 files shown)


mwiencek (approved):

yvanzo (approved):
:ship:

PR #1962, opened 2021-03-08
MBS-11426: Use medium's position, not array position, to load it

reosarevok:
### Fix MBS-11426

This was causing an issue where if a release doesn't have consecutive mediums, "disc/10" would load the 10th medium on the release, whatever its position, rather than the medium set to position 10.

Tested (with `pink` data) using the release given on the ticket. Alternatively, I guess you could add a release with 10+ mediums and ensure there's position gaps.


mwiencek (approved):

yvanzo (approved):
:ship:

jesus2099:
What is pink data? :)
It is already live, today. Thanks. :)

reosarevok:
`pink` is a read-only database server the team uses to access the live DB data for work without having the full database locally :) (it wouldn't even fit in my partition!)

PR #1963, opened 2021-03-08
Fix key warning in FooterSwitch, make small tweaks

mwiencek:
* Fixes a React key warning on the artist index page (in the `FooterSwitch` component) that's been spamming my logs, and probably the servers' logs.

 * Moves the initialization of strings in this component to where they're used. Calling into Jed isn't free, and most of these variables end up unused on every request.


reosarevok:
If this is a problem then we should probably do the same for https://github.com/reosarevok/musicbrainz-server/blob/804e34cbb5141a0200667dfecc6fb33ebb052e29/root/artist/ArtistRecordings.js#L44 ?

mwiencek:
Makes sense, changed those too

yvanzo (approved):

reosarevok (approved):
LGTM

PR #1964, opened 2021-03-08
MBS-10976: Allow some private use characters

mwiencek:
Allows the main PUA (U+E000..U+F8FF), but not any of the supplementary PUAs. This should at least cover the use case in the ticket (Klingon). If we want to be more restrictive, we could allow only (U+F8D0..U+F8FF) for Klingon, though there might be other use cases of the PUA that would be useful for MusicBrainz.

This commit also fixes an ISE that can be triggered by submitting a name consisting *only* of invalid characters. The message I added should very rarely be seen by any user; if they only enter whitespace, "Required field" is shown instead.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -405,9 +405,9 @@ sub remove_invalid_characters {
     # - bom
     # - Other, control
     # - Other, surrogate
-    # - Other, private use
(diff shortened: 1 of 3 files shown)


mwiencek:
@brainzbot retest this please

yvanzo:
@brainzbot retest this please

yvanzo (approved):

reosarevok (approved):
LGTM

PR #1965, opened 2021-03-09
MBS-10367: Allow requesting copy of sent report when reporting

reosarevok:
### Implement MBS-10367

This allows a reporter to save a copy of the reports they have sent, if so desired. That might be helpful to, for example, remember if you have already reported a user recently but the report still hasn't been acted on, and as such avoid sending another report on the same issue.

Tested only by printing the email to be sent with Data::Dumper, since I don't have email sending set up locally. All of the headers seemed correct to me though. If any of you do have this set up, can you test the actual email sending?

### DIFF
--- a/lib/MusicBrainz/Server/Email.pm
+++ b/lib/MusicBrainz/Server/Email.pm
@@ -579,19 +579,34 @@ sub send_editor_report {
 
     my $reporter = $opts{reporter};
     my $reported_user = $opts{reported_user};
-    my $subject = 'Editor ' . $reported_user->name . ' has been reported by ' . $reporter->name;
+    my $reported_user_name = $reported_user->name;
+    my $subject = 'Editor ' . $reported_user_name . ' has been reported by ' . $reporter->name;
     my $reason = $MusicBrainz::Server::Form::User::Report::REASONS{$opts{reason}};
     my $reporter_tolink = uri_escape_utf8($reporter->name);
     my $reported_user_tolink = uri_escape_utf8($reported_user->name);
-    my $body .= <<EOF;
-$subject for the following reason:
+    my $report_content = <<~"EOF";
+        $subject for the following reason:
 
-“$reason”
+        “$reason”
 
-Reporter’s account: https://musicbrainz.org/user/$reporter_tolink
-Reported user’s account: https://musicbrainz.org/user/$reported_user_tolink
+        Reporter’s account: https://musicbrainz.org/user/$reporter_tolink
+        Reported user’s account: https://musicbrainz.org/user/$reported_user_tolink
 
-EOF
+        EOF
+
+    my $message = $opts{message};
+    if ($message) {
+        $report_content .= <<~"EOF";
+            ------------------------------------------------------------------------
+            $message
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/Email.pm
```diff
@@ -624,6 +625,34 @@ EOF
...
     $self->_send_email($email);
+
+    if ($opts{send_to_self}) {
+        my $copy_subject = 'Copy of your report of ' . $reported_user_name;
+
+        my @copy_headers = (
+            'To'          => _user_address($opts{reporter}),
+            'Sender'      => $EMAIL_NOREPLY_ADDRESS,
+            'Subject'     => _encode_header($copy_subject),
+            'Message-Id'  => _message_id('editor-report-copy-%s-%s-%d', $reporter->id, $reported_user->id, time),
```
reosarevok, line 636: Wasn't 100% sure what to set as message ID - I understand it should be different from the ID for the actual report, so I added `-copy-`. I assume `time` wouldn't be 100% the same so maybe it's ok to just have this be the same? (that's what we seem to do for the copy of submitted user-to-user messages)

lib/MusicBrainz/Server/Email.pm
```diff
@@ -624,6 +625,34 @@ EOF
 
     my $email = $self->_create_email(\@headers, $body);
     $self->_send_email($email);
+
+    if ($opts{send_to_self}) {
+        my $copy_subject = 'Copy of your report of ' . $reported_user_name;
+
+        my @copy_headers = (
+            'To'          => _user_address($opts{reporter}),
```
mwiencek, line 633:
```suggestion
            'To'          => _user_address($reporter),
```

lib/MusicBrainz/Server/Email.pm
```diff
@@ -624,6 +625,34 @@ EOF
...
+            'Subject'     => _encode_header($copy_subject),
+            'Message-Id'  => _message_id('editor-report-copy-%s-%s-%d', $reporter->id, $reported_user->id, time),
+        );
+
+        push @copy_headers, 'From', _user_address($reporter, 1);
+
+        my $copy_body .= <<EOF;
+This is a copy of your report of MusicBrainz editor '$reported_user_name':
+------------------------------------------------------------------------
+$subject for the following reason:
```
mwiencek, line 644: Can we reuse `$body` here? Or save the initial `my $body` to a separate var if you don't want to include the reply info?

lib/MusicBrainz/Server/Email.pm
mwiencek: I think what you have should be fine.

lib/MusicBrainz/Server/Email.pm
reosarevok: I think so, yes. Reordered the email a bit so that all the content (inc. the message) goes first, and the you-can-reply info goes last, since that makes copying it easier too.

mwiencek (approved):

mwiencek:
@brainzbot test this please

PR #1966, opened 2021-03-09
MBS-11430: Convert format_setlist to Javascript

reosarevok:
### Implement MBS-11430

This is only used on the React layer anyway (for display), and having it be JS will later make it easier to generate a preview client-side from a not-yet-submitted setlist (MBS-8085).

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -30,7 +30,6 @@ module.exports = function (api) {
     ['@babel/plugin-transform-react-jsx', {
       runtime: 'automatic',
     }],
-    '@babel/plugin-transform-react-constant-elements',
     ['@babel/plugin-transform-runtime', {
       corejs: false,
       helpers: true,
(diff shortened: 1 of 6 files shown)


mwiencek (approved):
I was curious how much slower this is than the Perl version, since it's parsing the setlist to generate an HTML string and then parsing the HTML string to generate React elements.

Surprisingly, it was only twice as slow. I tested with the setlist from https://musicbrainz.org/event/b54fb46e-4e92-4c6c-a171-a19cc5fea8b7 and ran each format function 1000x. The Perl one took 0.15s while the JS one took 0.30s.

The output seemed to be identical other than the Perl version having an extra trailing `<br/>`.

mwiencek:
I looked at this again and wondered if we could greatly increase the performance (make it even faster than the Perl) by converting the setlist directly into React elements. Turns out it's possible. See the latest commit for details.

reosarevok:
Seems sensible, thanks!

mwiencek:
I actually broke it in cases where there are multiple links per line, or text adjacent to links. Fixed that + re-tested it against https://musicbrainz.org/event/503139c4-6b4d-46b8-a83a-109d59637c77 and added a test.

PR #1967, opened 2021-03-09
MBS-11431: /ws/js/check_duplicates is slow

mwiencek:
Adds the appropriate indexes needed by `Data::Role::Alias::search_by_names` and `Edit::Role::CheckDuplicates::is_disambiguation_needed`.

As shown in the ticket, these queries take anywhere from 1.5 to 3 seconds for artists currently. After I created the new artist and artist_alias indexes on floyd, they took less than 1 ms.

`artist_idx_lower_unaccent_name_comment` uses 70 MB on floyd; `artist_alias_idx_lower_unaccent_name` uses 10 MB. These would probably be the largest of the four entity types.

I'm also dropping `artist_idx_lower_name` and `label_idx_lower_name` here, which are completely unused according to `pg_stat_all_indexes`. That should offset a bit of the space required.


admin/sql/CreateIndexes.sql
```diff
@@ -611,9 +611,15 @@ CREATE INDEX work_tag_raw_idx_tag ON work_tag_raw (tag);
 
 CREATE UNIQUE INDEX work_type_idx_gid ON work_type (gid);
 
--- lowercase indexes for javascript autocomplete
-CREATE INDEX artist_idx_lower_name ON artist (lower(name));
-CREATE INDEX label_idx_lower_name ON label (lower(name));
```
reosarevok, line 616: What does autocomplete use, just name?

reosarevok (approved):
Seems sensible, in any case.

admin/sql/CreateIndexes.sql
mwiencek: The direct search makes use of the indexes in admin/sql/CreateSearchIndexes.sql

yvanzo (approved):
For the record, if you want to double-check or fix ticket or PR, I updated ticket’s summary (to match its Improvement status) and component (with Schema Change). Otherwise :ship:

mwiencek:
Makes sense, thanks!

PR #1968, opened 2021-03-10
MBS-11432: Allow also linking places and works to Operabase

reosarevok:
### Implement MBS-11432

I'm removing the cleanup for the old artist format because it hasn't been used for years, so it seems useless by now.

Works are just like artists, and can be reduced to works/id.
Places require the country name on the URL, so they can only be reduced to venues/country/id.


yvanzo (approved):
LGTM

mwiencek (approved):

PR #1969, opened 2021-03-10
MBS-11393: Show SetTrackLengths edit in recording edit history

reosarevok:
### Implement MBS-11393

If a medium edit changes the duration of a recording, it is shown on the recording's edit history. If a discID does (with a set track lengths edit), it does not, so it becomes hard to see where the set duration came from.

This will add the set track lenghts edit to the history of all recordings which are linked to a track that is changing duration.

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
+++ b/lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
@@ -63,6 +63,27 @@ sub foreign_keys {
     }
 }
 
+around _build_related_entities => sub {
+    my ($orig, $self) = splice(@_, 0, 2);
+    my $related = $self->$orig(@_);
+
+    my $medium = $self->c->model('Medium')->get_by_id($self->data->{medium_id});
+    $self->c->model('Track')->load_for_mediums($medium);
+
+    my $old_lengths = $self->data->{length}{old};
+    my $new_lengths = $self->data->{length}{new};
+    my $cdtoc_tracks = $medium->cdtoc_tracks;
+
+    for my $i (0 .. scalar @{$new_lengths}-1) {
+        if ($new_lengths->[$i] != $old_lengths->[$i]) {
+            my $changed_track = $cdtoc_tracks->[$i];
+            push @{ $related->{recording} }, $changed_track->{recording_id};
+        }
+    }
+
+    return $related;
+};
+
 sub build_display_data {
     my ($self, $loaded) = @_;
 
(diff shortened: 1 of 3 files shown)


mwiencek:
> We could choose to check the lengths array track by track to see if the specific track for this recording is changing, but I'm not sure it's worth it, when even from the edit display it is hard to see if a sub-second change was made for the track or not.

I'm a bit unsure about this. If we know for a fact that the edit couldn't have affected a recording, adding it to the edit history for that recording could be more confusing that not.

reosarevok:
Fair enough. I changed this to check track by track - it seems to work, but do let me know if I'm overcomplicating it.

lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
```diff
@@ -63,6 +63,26 @@ sub foreign_keys {
...
 
+around _build_related_entities => sub {
+    my ($orig, $self) = splice(@_, 0, 2);
+    my $related = $self->$orig(@_);
+
+    my $medium = $self->c->model('Medium')->get_by_id($self->data->{medium_id});
+    $self->c->model('Track')->load_for_mediums($medium);
+
+    my $old_lengths = $self->data->{length}{old};
+    my $new_lengths = $self->data->{length}{new};
```
mwiencek, line 74: `cdtoc_tracks` isn't cached, so I'd assign it to a variable first, otherwise it's recalculated for every track which itself loops over every track (n^2)

```suggestion
    my $new_lengths = $self->data->{length}{new};
    my $cdtoc_tracks = $medium->cdtoc_tracks;
```

lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
```diff
@@ -63,6 +63,26 @@ sub foreign_keys {
...
+
+    for my $i (0 .. scalar @{$new_lengths}-1) {
+        if (@$new_lengths[$i] != @$old_lengths[$i]) {
```
mwiencek, line 77:
```suggestion
        if ($new_lengths->[$i] != $old_lengths->[$i]) {
```

lib/MusicBrainz/Server/Edit/Medium/SetTrackLengths.pm
```diff
@@ -63,6 +63,26 @@ sub foreign_keys {
...
+            my $changed_track = @{$medium->cdtoc_tracks}[$i];
```
mwiencek, line 78:
```suggestion
            my $changed_track = $cdtoc_tracks->[$i];
```

mwiencek (changes requested):

reosarevok:
Made the changes, also added a test to ensure the edit appears only where it should.

mwiencek (approved):

PR #1971, opened 2021-03-10
MBS-11429: Don't error on existing but now blocked links

reosarevok:
### Implement MBS-11429

Normally it's not the editor's fault if someone before them has added a shortener link, MusicBrainz link, malware link or Google AMP link. As such, it feels unfair to force them to deal with the link before submitting their (unrelated) edits (plus the error messages all assume the link is newly added).

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -233,6 +233,7 @@ export class ExternalLinksEditor
             const oldLink = oldLinks[link.relationship];
             const isNewLink = !isPositiveInteger(link.relationship);
             const linkChanged = oldLink && link.url !== oldLink.url;
+            const isNewOrChangedLink = (isNewLink || linkChanged);
             const linkTypeChanged = oldLink && +link.type !== +oldLink.type;
             link.url = getUnicodeUrl(link.url);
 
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
LGTM but did not test. Thanks!

root/static/scripts/edit/externalLinks.js
```diff
@@ -233,25 +233,26 @@ export class ExternalLinksEditor
...
+            const isNewOrChangedLink = (isNewLink || linkChanged);
             const linkTypeChanged = oldLink && +link.type !== +oldLink.type;
             link.url = getUnicodeUrl(link.url);
 
             if (isEmpty(link)) {
               error = '';
             } else if (!link.url) {
               error = l('Required field.');
-            } else if (!isValidURL(link.url)) {
+            } else if (isNewOrChangedLink && !isValidURL(link.url)) {
```
mwiencek, line 244: The logic looks correct, so approving, though I feel like we should have a nested `if` for all these checks that depend on `isNewOrChangedLink` instead of checking it 6 times:

```suggestion
            } else if (isNewOrChangedLink) {
              if (!isValidURL(link.url)) {
              } else if (...) {
                // ...
              }
            }
```

Maybe even move all the checks the depend only on the URL to a separate function. These checks were already messy before this PR so it'd be nice to make it a bit better.

mwiencek (approved):

root/static/scripts/edit/externalLinks.js
reosarevok: Let's merge for now, but this should be improved in a later PR, maybe make it support an array of errors.

mwiencek: Added https://tickets.metabrainz.org/browse/MBS-11521

PR #1972, opened 2021-03-11
MBS-11435: Don't modify data in build_display_data for EditReleaseLabel

reosarevok:
### Fix MBS-11435

This is causing the new modifications not to get converted to JSON, causing an ISE. There's also no good reason to do this at all, since we can easily pass only the data we need via build_display_data and we even limit the amount of bandwidth needed while at it.


mwiencek (approved):

yvanzo (approved):
:ship:

PR #1973, opened 2021-03-11
MBS-11436: Don't try to call method public on unblessed JSON

reosarevok:
### Fix MBS-11436

These used to be blessed collections but are now JSON, so {public} is JSON::true/false. Dereferencing it to see if true/false.

Tested by making one of my own collections private and making sure it matched prod (so you can test with `pink` data from http://localhost:5000/user/reosarevok/subscriptions/collection).


mwiencek (approved):
If we can find a way to calculate this stuff beforehand and not via component props, that would be ideal, but this works for now.

reosarevok:
The alternative fix seems to work (tested locally) and is nicer, +1.

yvanzo (approved):
:ship:

PR #1974, opened 2021-03-11
MBS-11437: Beta: work-level rels not shown on release pages

mwiencek:
68f8ae8a49aaa22e49c4982c0ecd0563f8e8dae4 broke this for obvious reasons. Since I'm not comfortable reverting that (I did run into a case where infinite recursion was triggered -- just not sure where), I increased the max allowed depth instead.


reosarevok (approved):
Tested locally on `pink` data, works.

yvanzo (approved):
:ship:

PR #1975, opened 2021-03-12
MBS-10327 / MBS-11439 / MBS-11440: Improvements for collection subscriptions

reosarevok:
### Fix MBS-10327 / MBS-11440, implement MBS-11439

Tested manually by creating collections as one user and seeing whether I could subscribe or not, etc as another (marked as collaborator and then removed).


lib/MusicBrainz/Server/Controller/Account/Subscriptions/Collection.pm
```diff
@@ -21,7 +21,11 @@ before add => sub
         $c->detach('/error_400');
     }
 
-    $c->detach('/error_403') if (!$entity->public && $c->user->id != $entity->editor_id);
+    my $is_collection_owner = $c->user->id == $entity->editor_id;
+    my $is_collection_collaborator = $c->user_exists &&
+        $c->model('Collection')->is_collection_collaborator($c->user->id, $entity->id);
```
mwiencek, line 26: The `add` method from the role has `RequireAuth`, so I think we don't need to check `$c->user_exists` here (and we already reference `$c->user->id` above without checking it).

It would be nice to avoid running this query altogether if we already know either of `$entity->public` or `$is_collection_owner` are true. Perhaps something like:

```Perl
my $is_authorized = (
    $entity->public ||
    $c->user->id == $entity->editor_id ||
    $c->model('Collection')->is_collection_collaborator($c->user->id, $entity->id)
);
```

lib/MusicBrainz/Server/Controller/Role/Subscribe.pm
```diff
@@ -12,6 +12,15 @@ sub subscribers : Chained('load') RequireAuth {
     my $model = $self->{model};
     my $entity = $c->stash->{ $self->{entity_name} };
 
+    # Don't show private collections to people without rights to see them
+    if ($model eq 'Collection') {
+        my $is_collection_owner = $c->user->id == $entity->editor_id;
+        my $is_collection_collaborator = $c->user_exists &&
+            $c->model('Collection')->is_collection_collaborator($c->user->id, $entity->id);
+
+        $c->detach('/error_403') if (!$entity->public && !($is_collection_owner || $is_collection_collaborator));
```
mwiencek, line 21: Same comment as above

mwiencek (approved):

yvanzo (changes requested):
LGTM but did not test. Please add reproducible tests for this.

reosarevok:
Added a test for the last one (the /subscribers page). Looking into a good way to test the others - especially the collaborators one needs a fair amount of work, since we have 0 tests for collection collaborators (wonder who wrote that code and didn't add any :eyes:)

reosarevok:
Cherry-picked 12463babc22666a66029fc3affb8801946e4ec17 for release for now since it fixes a potentially troubling bug and has a test.

reosarevok:
@yvanzo: thanks (non-ironically) for forcing me to write tests, I found out this needed to include a schema change, actually! Although one we should be able to do in prod whenever. Made the change, and added a bunch of other tests for collaborator stuff.

yvanzo:
@brainzbot retest this please

lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -429,6 +429,20 @@ sub set_collaborators {
...
+        'DELETE FROM editor_subscribe_collection esc
+         USING editor_collection ec
+         WHERE esc.collection = ?
+         AND esc.collection = ec.id
+         AND ec.public IS FALSE
+         AND esc.editor != ec.editor
+         AND esc.editor NOT IN (
+             SELECT editor FROM editor_collection_collaborator
+             WHERE collection = esc.collection
+        )',
```
mwiencek, line 443: In `del_collection_sub_on_private` we do `UPDATE ... SET available = FALSE, last_seen_name = OLD.name`.  Is there a reason we delete them here instead?

reosarevok: Honestly, probably just that I didn't know that code existed back then. I should probably do the same here then, or write a SQL function for that?

mwiencek: Let's just update the query here then.  If we need it more than twice let's add a function. :)

reosarevok: Done, I think it should be working. Added a test for this use case too.

PR #1977, opened 2021-03-12
MBS-8820: Show explanation if relationships tab shows no rels

reosarevok:
### Fix MBS-8820

For an artist that only has event relationships (or, I assume, a label that only has relationships of a type that gets filtered out by groupRelationships), the relationships tab is currently just empty.
This changes it so that Relationships displays an informative message if all existing relationships have been filtered out. While at it, I also got rid of the separate "no rels" message on ArtistRelationships and LabelRelationships, and moved that to Relationships as well.

Tested locally by creating the appropriate situations (artist without any rels, artist with only event rels) and navigating around to see that the changes didn't seem to affect other relationship displays.

### DIFF
--- a/root/components/Relationships.js
+++ b/root/components/Relationships.js
@@ -10,6 +10,7 @@
 import * as React from 'react';
 
 import {PART_OF_SERIES_LINK_TYPES} from '../static/scripts/common/constants';
+import EntityLink from '../static/scripts/common/components/EntityLink';
 import linkedEntities from '../static/scripts/common/linkedEntities';
 import groupRelationships, {type RelationshipTargetTypeGroupT}
   from '../utility/groupRelationships';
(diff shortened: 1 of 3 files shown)


yvanzo (approved):
LGTM but did not test.

root/components/Relationships.js
```diff
@@ -91,18 +94,40 @@ const Relationships = ({
...
           />
         </>
+      ) : srcRels?.length ? (
+        <>
+          {heading}
+          <p>
+            {exp.l(
+              '{link} only has relationships that are displayed elsewhere.',
+              {link: <EntityLink entity={source} />},
+            )}
```
mwiencek, line 118: I do wish we could make this a bit less vague by saying where "elsewhere" is. As @yvanzo suggested in the meeting notes:

> Maybe we should just leave a note about event and work relationships to the user when the relationships tab is empty.

Having a custom string for each entity type doesn't seem ideal, but maybe a list of links to the other tabs would work? I'm not sure what's best here. Do you think this string is the best we can do until we can make more substantive UI improvements? If so I'm okay with merging it as-is.

reosarevok: I wasn't sure how to do this otherwise. I guess we could check every time whether one or more event rels exist. Work rels are actually displayed, just under the appearances table further below, same as recording rels, for example (which are also filtered out).

mwiencek: Maybe separate strings wouldn't be so bad then. Since artists and labels are the only entities with dedicated relationship tabs, the only two places you'd have to tell the user to check is (1) the events tabs for artist or (2) the index page for labels, right?

reosarevok: Hmm, actually I'm not even sure whether there's currently any rel for labels that does not appear on the relationships tab, either as in the relationship list or the table. So we probably only need this for artist, heh.

mwiencek (approved):

PR #1979, opened 2021-03-15
MBS-11444: Beta: Pass Label->new() call through to_json_object

reosarevok:
### Fix MBS-11444


reosarevok:
BTW, we do have an entity_gid stored in that edit that would allow us to link to the new merged label rather than nothing, but we seem to check entity_id only on add edits. Maybe something to reconsider at some point.

yvanzo (approved):
:ship:

PR #1981, opened 2021-03-15
MBS-9543: Avoid false negative for string '0'

reosarevok:
### Fix MBS-9543

These ifs were seeing any seeded string `'0'` as falsey and skipping seeding that field. This ensures the field is still not seeded when `_seeded_string` specifically returns `undef`, which I assume was the intended target of these checks.


mwiencek:
Should we use the `non_empty` util for this instead? That would also handle `''`.

mwiencek (approved):

yvanzo (approved):
:ship:

PR #1983, opened 2021-03-15
MBS-11442: Adding tags on add-cover-art page disables submit

mwiencek:
Surprisingly, this is due to some relationship editor code that's running. This shouldn't even be loaded on these pages, so I created a separate bundle for CoverArt.js. I also added a workaround in the relevant relationship editor code, just in case this affects other edit pages too.


reosarevok (approved):

yvanzo (approved):
Merging for upcoming beta release.

PR #1984, opened 2021-03-15
MBS-11446: Keep blank lines in annotation

yvanzo:
# Fix MBS-11446: Annotation trimming removes blank lines

Since commit 43b9dd787947471b6dbddf15baf98947fe916d2c, blank lines were removed from edited annotations because `\s` matches `\n` too.

This pull request amends that commit by not removing consecutive `\n` but if there are more than two.

Update tests accordingly.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -367,10 +367,13 @@ sub trim_multiline_text {
 
     $t = NFC($t);
     $t = remove_invalid_characters($t);
-    # Not trimming starting spaces to avoid breaking
-    # either list formatting in Wikitext
-    # or block in Markdown.
-    $t =~ s/\s+$//gm;
+
+    # Trimming each line to remove trailing spaces (or similar)
+    # - Not trimming starting spaces to avoid breaking
+    #   either list formatting in Wikitext
+    #   or block in Markdown.
+    # - Splitting on \n so that \s doesn’t match any \n
+    $t = join ("\n", map { $_ =~ s/\s+$//r } (split "\n", $t));
 
     return $t;
 }
--- a/lib/MusicBrainz/Server/Data/Utils.pm
+++ b/lib/MusicBrainz/Server/Data/Utils.pm
@@ -375,6 +375,9 @@ sub trim_multiline_text {
     # - Splitting on \n so that \s doesn’t match any \n
     $t = join ("\n", map { $_ =~ s/\s+$//r } (split "\n", $t));
 
+    # Merge consecutive blank lines together
+    $t =~ s/\n+(\n\n)/$1/g;
+
     return $t;
 }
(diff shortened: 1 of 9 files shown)


yvanzo:
Tested with local database and sample text:
```wikitext
    1. one trailing space 
    2. two trailing spaces  
    
    3. one blank line above   
    
    
    4. two blank lines above    
```
* Annotation editor: http://localhost:5000/recording/96f64611-49df-4e54-84e7-0f9a30f01766/edit_annotation
* Release editor: http://localhost:5000/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/edit

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -367,10 +367,16 @@ sub trim_multiline_text {
...
-    # either list formatting in Wikitext
-    # or block in Markdown.
-    $t =~ s/\s+$//gm;
+
+    # Trimming each line to remove trailing spaces (or similar)
+    # - Not trimming starting spaces to avoid breaking
+    #   either list formatting in Wikitext
+    #   or block in Markdown.
+    # - Splitting on \n so that \s doesn’t match any \n
+    $t = join ("\n", map { $_ =~ s/\s+$//r } (split "\n", $t));
```
mwiencek, line 376: In the initial commit you had `$t =~ s/[\t\r ]+$//gm;`, but all of the current tests pass with either piece of code. I guess that's to catch other characters matched by `\s` that aren't covered by `\t\r `?

Another option, if you prefer, is regex_sets, i.e. subtracting `\n` from `\s` like `/(?[ \s - \n ])/`. Though since it's experimental, you'll need to disable a warning: https://perldoc.perl.org/perlrecharclass#Extended-Bracketed-Character-Classes

mwiencek (approved):
This seems to work. You can disregard the suggestion below as you prefer.

reosarevok (approved):
Your choice whether you try the suggested changes or not :)

lib/MusicBrainz/Server/Data/Utils.pm
yvanzo: I tried many variations but I failed to make it work:
```perl
    $t =~ s/(?[ [\s] - [\n] ])$//gm;
```

Let’s not use experimental error-prone syntax when we have a single line code that works.

I added NBSP (U+00A0) to the tests though.

mwiencek: The working variation is `$t =~ s/(?[ \s - \n ])+$//gm;`.

> I added NBSP (U+00A0) to the tests though.

Thanks!

yvanzo: Plus I prefer to avoid using `m` for multi-line as it does not treat `\n` as I would expect.

PR #1988, opened 2021-03-16
MBS-11452 /  MBS-11462: Fixes to user tabs display

reosarevok:
### Fix MBS-11452 /  MBS-11462


mwiencek (approved):

yvanzo (approved):
LGTM but did not test. Adding tests or more details about tests would be welcome though. That can be later on.

PR #1989, opened 2021-03-17
MBS-11468: ISE on "Edit cover art" edits

mwiencek:
Passes the blessed `Release` object to `Artwork->new` as intended, rather than the unblessed JSON object.


reosarevok (approved):

yvanzo (approved):
:ship:

PR #1990, opened 2021-03-17
MBS-11470: Ensure to_json_object(type) does run in Area::Edit

reosarevok:
### Fix MBS-11470

This is checking $self->data->{new}{type} to see whether to run to_json_object on the type data, but the edit data does not have types, it has type_id from which the type is loaded for display.


yvanzo (approved):
:ship:

PR #1991, opened 2021-03-17
MBS-11467: Use sort name to order lists of artists for tags

reosarevok:
### Implement MBS-11467

It makes sense to use sort names for sorting artists also in tag lists.
This changes get_by_ids_sorted_by_name to use sort_name for artists, but this seems to be used for tags only anyway. In any case, I don't feel the name becomes incorrect, since the sort_name column is just meant as "the way to sort artists by name" in the first place.

Tested with http://localhost:5000/tag/classical/artist / http://localhost:5000/user/reosarevok/tag/classical (`pink` data).


mwiencek (approved):

yvanzo (approved):
:ship

PR #1992, opened 2021-03-17
MBS-11454: Don't consider hasTooEarlyFormat a blocking error

reosarevok:
### Fix MBS-11454

If a release already had a too early format before the editor started this release editor session, `hasTooEarlyFormat` is still set. As such, these lines were blocking editing even in that case.
The whole point of having `hasUnconfirmedEarlyFormat` as well was for that to be the one that triggers an error, while just `hasTooEarlyFormat` should not. Compare `hasUnconfirmedVariousArtists` vs `hasVariousArtists`.

### DIFF
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -1042,9 +1042,6 @@ class Release extends mbEntity.Release {
     this.hasUnconfirmedVariousArtists = errorField(
       this.mediums.any('hasUnconfirmedVariousArtists'),
     );
-    this.hasTooEarlyFormat = errorField(
-      this.mediums.any('hasTooEarlyFormat'),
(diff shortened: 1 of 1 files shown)


yvanzo (changes requested):
The code change looks good but the commit message is very unclear.

reosarevok:
What is unclear about it? :/

yvanzo:
@reosarevok: I reviewed code changes thanks to the ticket’s description. How could I suggest what is unclear to me? :D

reosarevok:
@yvanzo added some more text to the commit message, is it better now?

yvanzo (approved):
@reosarevok: Much better, thank you!

mwiencek (approved):

PR #1993, opened 2021-03-17
MBS-11455: Hide view links on stats when logged out

reosarevok:
### Implement MBS-11455

These links are already hidden for the Added entities section. Elsewhere they are shown, but they require log in on click - except for the last 24 hours link, which loads but drops the private info (user) and misleadingly shows all MB edits from the last 24 hours. I think just hiding these links unless the user is logged in is the most reasonable choice.

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -431,9 +431,7 @@ const UserEditsProperty = ({
       {$c.user ? (exp.l('{count} ({view_url|view})', {
         count: formatCount($c, addedEntities),
         view_url: searchEditsURL(createEditTypes),
-      })) : (exp.l('{count}', {
-        count: formatCount($c, addedEntities),
-      }))}
+      })) : formatCount($c, addedEntities)}
     </UserProfileProperty>
   );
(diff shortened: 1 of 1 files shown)


yvanzo (approved):

root/user/UserProfile.js
```diff
@@ -505,87 +505,101 @@ const UserProfileStatistics = ({
...
           <UserProfileProperty name={lp('Accepted', 'edit descriptor')}>
-            {exp.l(
+            {$c.user ? exp.l(
               '{count} ({view_url|view})',
               {
                 count: formatCount($c, editStats.accepted_count),
                 view_url: `/user/${encodedName}/edits/accepted`,
               },
-            )}
+            ) : (exp.l('{count}', {
```
mwiencek, line 524: Is that string useful? I figured `formatCount` we be enough to localize the number.

reosarevok: ... that's a fair point :) Will amend.

mwiencek (approved):

PR #1994, opened 2021-03-18
MBS-11473: Don't try to filter open edits for user if logged out

reosarevok:
### Fix MBS-11473

Unsurprisingly, this was giving an ISE. I think just showing all open edits in this case makes the most sense.


mwiencek (approved):

yvanzo (approved):
Good catch @reosarevok!

PR #1995, opened 2021-03-18
MBS-11476: Don't show own open edits in /edit/open

reosarevok:
### Implement MBS-11476

We already filter edits the editor has voted on, so obviously this is meant for voting. As such, it is useless to show open edits by that specific editor.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Edit.pm
+++ b/lib/MusicBrainz/Server/Data/Edit.pm
@@ -289,6 +289,7 @@ sub find_open_for_editor
         'SELECT ' . $self->_columns . '
            FROM ' . $self->_table . '
           WHERE status = ?
+            AND editor != ?
             AND NOT EXISTS (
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/Edit.pm
```diff
@@ -289,6 +289,7 @@ sub find_open_for_editor
         'SELECT ' . $self->_columns . '
            FROM ' . $self->_table . '
           WHERE status = ?
+            AND EDITOR != ?
```
mwiencek, line 292: No need to demonstrate that PG is case insensitive here. :)
```suggestion
            AND editor != ?
```

mwiencek (changes requested):

mwiencek (approved):

yvanzo (approved):
:ship:

PR #1998, opened 2021-03-19
MBS-11485: Allow irc(s):// links on expand2react

reosarevok:
### Implement MBS-11485

A user bio repeatedly fails to parse because it has some irc:// links. Unless I'm missing some reason why these are a security risk, we should allow users to link to IRC, especially given we use IRC all the time.

User in question: https://musicbrainz.org/user/xps2


yvanzo (approved):
IIRC there is a shortlist of allowed protocols to prevent exploits based on weird protocols (file, javascript, and so on). I don’t think it intentionally targeted IRC links. At least commit https://github.com/metabrainz/musicbrainz-server/commit/1d5513313d959dbc79de6ac1fcd0e414b24e9338 does not give details about it. Since it is our main channel of communication for chat, it makes sense to specifically allow it. I did not find previous implementation but I don’t recall about any shortlist so I guess it was probably just too permissive.

mwiencek (approved):

PR #1999, opened 2021-03-19
MBS-11486: Load release info in NoLanguage before JSON conversion

reosarevok:
### Fix MBS-11486


reosarevok:
This is the same as @mwiencek's code for NoScript and works, so I'm merging without review so I can put a hotfix out.

PR #2000, opened 2021-03-19
MBS-11367: Disable browser autocomplete for tags when non-empty

reosarevok:
### Fix MBS-11367

Looking at the comment here, the original intention was to activate the autocomplete only if the user hadn't typed anything yet, but we never deactivated it once the typing started.

### DIFF
--- a/root/static/scripts/common/components/TagEditor.js
+++ b/root/static/scripts/common/components/TagEditor.js
@@ -213,6 +213,8 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
 
   genreNames: $ReadOnlyArray<string>;
 
+  handleChange: () => void;
+
   handleSubmit: (SyntheticEvent<HTMLFormElement>) => void;
 
   onBeforeUnload: () => void;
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/components/TagEditor.js
```diff
@@ -365,6 +368,26 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
...
+     * history in a non-configurable way, but we want it to show here
+     * if the user hasn't typed anything yet, so flip it back on.
+     * Turn it off again if something has been typed (activating
+     * the jQuery autocomplete).
+     */
+    if (input.value === '') {
+      input.setAttribute('autocomplete', 'on');
+    } else {
+      input.setAttribute('autocomplete', 'off');
+    }
```
mwiencek, line 388: This minifies better and seems equally easy to understand:

```suggestion
    /*
     * MBS-9862: jQuery UI disables the browser's builtin autocomplete
     * history in a non-configurable way, but we want it to show here
     * if the user hasn't typed anything yet, so flip it back on.
     * Turn it off again if something has been typed (activating
     * the jQuery autocomplete).
     */
    this.tagsInput?.setAttribute('autocomplete', input.value ? 'off' : 'on');
```

mwiencek (approved):

yvanzo (approved):
:shipit:

PR #2001, opened 2021-03-19
MBS-10647: Add [no label] to b_del_label_special trigger

reosarevok:
### Implement MBS-10647

[no label] is now officially a special purpose label that should never be deleted, so this adds its rowid (3267) to b_del_label_special.


admin/sql/updates/20210319-mbs-10647.sql
```diff
@@ -0,0 +1,9 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
+DROP TRIGGER b_del_label_special ON label;
```
mwiencek, line 4: (This should exist, but best to avoid failing the upgrade if it doesn't.)
```suggestion
DROP TRIGGER IF EXISTS b_del_label_special ON label;
```

upgrade.json
```diff
@@ -82,7 +82,8 @@
     "slave": ["20200512-mbs-10821-orphaned-recording-collection.sql",
               "20200914-oauth-pkce.sql",
               "20201028-mbs-1424.sql",
-              "20210309-mbs-11431.sql"],
+              "20210309-mbs-11431.sql",
+              "20210319-mbs-10647.sql"],
```
mwiencek, line 86: This should  be moved to "standalone" since triggers aren't created on slaves

mwiencek (changes requested):

mwiencek (approved):

PR #2002, opened 2021-03-19
MBS-11464: Drop table statistics.log_statistic

reosarevok:
### Implement MBS-11464

This was never used since it got added in 2012


admin/sql/updates/20210319-mbs-11464.sql
```diff
@@ -0,0 +1,8 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
+ALTER TABLE statistics.log_statistic DROP CONSTRAINT IF EXISTS log_statistic_pkey;
```
mwiencek, line 4: We can remove this since `DROP TABLE` always drops constraints with it.

admin/sql/updates/20210319-mbs-11464.sql
```diff
@@ -0,0 +1,8 @@
...
+
+DROP TABLE statistics.log_statistic;
```
mwiencek, line 6:
```suggestion
DROP TABLE IF EXISTS statistics.log_statistic;
```

mwiencek (changes requested):

PR #2003, opened 2021-03-19
MBS-11466: Change language.frequency and script.frequency to SMALLINT

reosarevok:
### Implement MBS-11466

Values here are 0-4, no point using INTEGER.


admin/sql/updates/20210319-mbs-11466.sql
```diff
@@ -0,0 +1,14 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
+ALTER TABLE language
+    ALTER COLUMN frequency SET DATA TYPE SMALLINT,
+    ALTER COLUMN frequency SET NOT NULL,
+    ALTER COLUMN frequency SET DEFAULT 0;
```
mwiencek, line 7: We only really need the `SET DATA TYPE` bit; the others won't be affected by it.

mwiencek (approved):
Looks good but I'd remove the unnecessary (if harmless) parts.

PR #2005, opened 2021-03-19
MBS-11451: Support ratings for places

reosarevok:
### Implement MBS-11451


reosarevok:
Tested locally both on site and WS. Realized that I had forgotten the equivalent of https://github.com/metabrainz/musicbrainz-server/pull/507 so I added that (AFAICT that should include every place, so I'm doing an insert from place after creation - if that's not needed, let me know!).

admin/sql/CreateIndexes.sql
```diff
@@ -456,6 +456,9 @@ CREATE UNIQUE INDEX place_attribute_type_allowed_value_idx_gid ON place_attribut
 
 CREATE UNIQUE INDEX place_alias_type_idx_gid ON place_alias_type (gid);
 
+CREATE INDEX place_rating_raw_idx_place ON place_rating_raw (place);
```
mwiencek, line 459: This index is actually redundant with `place_rating_raw_pkey`, so it can be dropped. (PostgreSQL [automatically creates a unique index for primary keys](https://www.postgresql.org/docs/9.4/indexes-unique.html), and the primary key here already has `place` as the left-most column.)

The other `${entity}_rating_raw_idx_${entity}` indexes can be dropped too, but it's not something urgent we have to do this schema change.

admin/sql/updates/20210319-mbs-11451-standalone.sql
```diff
@@ -0,0 +1,9 @@
+\set ON_ERROR_STOP 1
+BEGIN;
+
+DROP TRIGGER IF EXISTS a_ins_place ON place;
+
+CREATE TRIGGER a_ins_place AFTER INSERT ON place
+    FOR EACH ROW EXECUTE PROCEDURE a_ins_place();
+
+COMMIT;
```
mwiencek, line 9: nitpick: missing newline at the end of the file

admin/sql/updates/20210319-mbs-11451.sql
```diff
@@ -0,0 +1,34 @@
...
+CREATE OR REPLACE FUNCTION a_ins_place() RETURNS trigger AS $$
+BEGIN
+    INSERT INTO place_meta (id) VALUES (NEW.id);
+    RETURN NULL;
+END;
+$$ LANGUAGE 'plpgsql';
+
+INSERT INTO place_meta (id)
+    (SELECT id FROM place)
+    ON CONFLICT (id) DO NOTHING;
```
mwiencek, line 29: We could leave out `ON CONFLICT (id) DO NOTHING` in this case since the `id` will always be unique (it's the primary key).

```suggestion
INSERT INTO place_meta (id)
    (SELECT id FROM place);
```

admin/sql/updates/20210319-mbs-11451.sql
```diff
@@ -0,0 +1,34 @@
...
+
+CREATE INDEX place_rating_raw_idx_place ON place_rating_raw (place);
```
mwiencek, line 31:
```suggestion
```

admin/sql/CreateIndexes.sql
reosarevok: But would it be ok to do it this schema change? Or is it something we should have announced beforehand?

reosarevok: Also: "The index covers the columns that make up the primary key or unique constraint (a multicolumn index, if appropriate), and is the mechanism that enforces the constraint." - shouldn't that mean that `place_rating_raw_idx_editor` is also covered? Or does it matter that it is the first vs second column, and why?

mwiencek: > But would it be ok to do it this schema change?

It would be okay IMO, since they're redundant.

> shouldn't that mean that place_rating_raw_idx_editor is also covered? Or does it matter that it is the first vs second column, and why?

You can use the pkey index to query for editors, but it's less efficient because it's not the leading column, so more of the table has to be scanned. For B-tree indexes (which is the default), the "why" is just the fact of that being how the tree is sorted. https://www.postgresql.org/docs/12/indexes-multicolumn.html

mwiencek: (I specified B-tree here 'cause it does depend on the index type; as that link notes, with GIN/BRIN indexes the column order doesn't affect search effectiveness.)

reosarevok: Added https://github.com/metabrainz/musicbrainz-server/pull/2091 for that

mwiencek:
Pushed a couple fixes to the upgrade scripts in https://github.com/metabrainz/musicbrainz-server/pull/2099

PR #2014, opened 2021-03-24
MBS-9310: Link to relationship type in Add Relationship Type edits

reosarevok:
### Implement MBS-9310

For old enough edits this isn't doable since we do not store entity_id nor entity_gid. Starting 2015 we have that data, so it makes sense to link to the relationship type.

Many changes are indentation - I'd suggest hiding whitespace changes.

### DIFF
--- a/root/edit/details/AddRelationshipType.js
+++ b/root/edit/details/AddRelationshipType.js
@@ -11,6 +11,7 @@ import * as React from 'react';
 
 import IntentionallyRawIcon from '../components/IntentionallyRawIcon';
 import Cardinality from '../../static/scripts/common/components/Cardinality';
+import EntityLink from '../../static/scripts/common/components/EntityLink';
 import OrderableDirection
   from '../../static/scripts/common/components/OrderableDirection';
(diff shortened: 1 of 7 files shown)


root/edit/details/AddRelationshipType.js
```diff
@@ -33,6 +34,7 @@ type AddRelationshipTypeEditT = {
     +long_link_phrase: string,
     +name: string,
     +orderable_direction?: number,
+    +relationship_type: LinkTypeT,
```
mwiencek, line 37: Is there a case where `defined($self->entity_id)` isn't true? 'Cause we don't output a `relationship_type` key in that case.
 
```suggestion
    +relationship_type?: LinkTypeT,
```

reosarevok: Oh, sorry, yes, I forgot changing this once I found out we don't have entity_id for old edits. Will amend.

mwiencek (approved):

yvanzo (approved):

PR #2015, opened 2021-03-25
MBS-5632 / MBS-11522 / MBS-11523: Convert Set Cover Art edit to React 

reosarevok:
### Fix MBS-5632 / MBS-11522, implement MBS-11523

MBS-11522 is a TO_JSON regression. Trivial to just fix, but thought I'd do the conversion anyway since it's not hard.


root/edit/details/SetCoverArt.js
```diff
@@ -0,0 +1,102 @@
...
+              <p>
+                {commaList(oldArt.types) || '-'}
+                <br />
+                {oldArt.comment}
+              </p>
+            </div>
+          ) : display.isOldArtworkAutomatic ? (
+            l(`The old image was selected automatically
+               from the earliest release in the release group.`)
+          ) : l(`We are unable to display this cover art.`)}
```
yvanzo, line 70: _We_ (developers) should avoid to use _we_  (?) as some people seem to forget this is a community driven database:

```suggestion
          ) : l(`Unable to display this cover art.`)}
```

There currently are only 3 other occurrences of _We are_ and 20 other occurrences of _We_ in the server code.

root/edit/details/SetCoverArt.js
```diff
@@ -0,0 +1,102 @@
...
+                  <ReleaseEvents events={newArt.release.events} />
+                </p>
+              ) : null}
+              <p>
+                {commaList(newArt.types) || '-'}
+                <br />
+                {newArt.comment}
+              </p>
+            </div>
+          ) : l(`We are unable to display this cover art.`)}
```
yvanzo, line 95: Same as above:

```suggestion
          ) : l(`Unable to display this cover art.`)}
```

root/edit/details/SetCoverArt.js
mwiencek: It sounds a bit too informal to me that way, because it's being kept in the active voice but missing a subject (so it's still implying "we were"/"we are").

We could write these in a passive voice, i.e. "This cover art couldn't/can't be displayed," I guess. Do you mean that "we" is too vague, or people can misinterpret it to mean the community?

yvanzo: I mean people could misinterpret “we” as the website holders and mistakenly deduce that the database is maintained by a specific team rather than an open and welcoming community. This is probably more relevant to editors than to visitors, but we can probably follow it as a general rule.

reosarevok: But it's not the community who has failed at showing the cover, it's the core team :D

yvanzo: I did not write that. The issue is just that using a non-inclusive “we” in the website may let people think they are not part of the “million monkeys with typewriters” that maintain the database or that the code is not open source, which may let them think it is relevant to ask “to speak with [our] supervisor” or to threaten to “move legally” if not “all additions [are] returned”.

lib/MusicBrainz/Server/Edit/ReleaseGroup/SetCoverArt.pm
```diff
@@ -129,6 +131,8 @@ sub build_display_data {
     $data{artwork}{old} = to_json_object($artwork_by_release_id{$old_id}) if $old_id;
     $data{artwork}{new} = to_json_object($artwork_by_release_id{$new_id}) if $new_id;
 
+    $data{isOldArtworkAutomatic} = !$old_id;
```
mwiencek, line 134:
```suggestion
    $data{is_old_artwork_automatic} = !$old_id;
```

root/edit/details/SetCoverArt.js
mwiencek: TBH I just don't see how it can be read that way. We could change it to a passive voice, but the current string sounds the most natural to me.

reosarevok: This is just following "We are unable to display history for this cover art". I'd keep this as is for now, if you really dislike our using "We" elsewhere you can always send a PR to change them all - but it seems like a bad idea to block what was (at least at the time) a regression fix because of a small wording issue.

mwiencek: This discussion seems to have ended, so I'm going to merge this as it fixes a bug.  A ticket to remove "we" from our strings might make sense, though it shouldn't block this PR.

PR #2018, opened 2021-03-26
MBS-11520: Convert Remove Relationship Type edit to React

reosarevok:
### Implement MBS-11520

This adds some more data we were storing, but not displaying (link phrases, attributes). Some of the phrases are missing in older edits, as usual, and a few times the attribute itself might have been removed.
In general, changed this to use display_data and work in a similar way as AddRelationshipType. Since these always auto-apply, there's no point linking to the type itself like in AddRelationshipType, as it will be gone.


root/edit/details/RemoveRelationshipType.js
```diff
@@ -0,0 +1,157 @@
...
+};
+
+const RemoveRelationshipType = ({edit}: Props): React.Element<'table'> => {
+  const display = edit.display_data;
+  const entity0Type = ENTITY_NAMES[display.entity0_type]();
+  const entity1Type = ENTITY_NAMES[display.entity1_type]();
+
+  // Always display entity placeholders for ease of understanding
+  let longLinkPhrase = display.long_link_phrase;
+  if (longLinkPhrase && !longLinkPhrase.match('{entity0}')) {
```
mwiencek, line 43: (match is meant to be passed a regular expression)
```suggestion
  if (longLinkPhrase && !longLinkPhrase.includes('{entity0}')) {
```

root/edit/details/RemoveRelationshipType.js
```diff
@@ -0,0 +1,157 @@
...
+    longLinkPhrase = '{entity0} ' + longLinkPhrase;
+  }
+  if (longLinkPhrase && !longLinkPhrase.match('{entity1}')) {
```
mwiencek, line 46:
```suggestion
  if (longLinkPhrase && !longLinkPhrase.includes('{entity1}')) {
```

mwiencek (changes requested):

root/edit/details/RemoveRelationshipType.js
reosarevok: Made the change, and also on `AddRelationshipType`

mwiencek (approved):

PR #2019, opened 2021-03-26
MBS-11530: Avoid common/MB/release code running on these

reosarevok:
### Fix MBS-11530

Using `.collapse-medium` here broke because `common/MB/release` actually has code that runs on clicking a `.collapse-medium`.
For now, using `collapse-re-medium` / `expand-re-medium` to ensure a quick fix, but we shouldn't be running that code inside the release editor anyway, it's only relevant to a release's display page.

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -285,7 +285,7 @@
     <table class="advanced-format">
       <tr>
         <td class="icon">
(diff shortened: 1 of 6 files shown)


yvanzo (approved):
Works for me as a quick fix.

mwiencek (approved):

yvanzo (approved):
Thank you @mwiencek!

PR #2024, opened 2021-03-31
MBS-11537: Collapse work attributes when there are too many (but for edits)

reosarevok:
### Implement MBS-11537

Since users have started adding codes for every PRO under the sun, some of these lists have gotten quite unwieldy, with 30+ entries. It makes sense to collapse them, in the same way we do for work artists and release events.

Tested with the entities mentioned on the ticket, work/be15e96f-b967-3d39-bed6-194a9e836338 / artist/4853b4f9-bcd6-4663-86bd-36d1270c6a3b/works

Not changing it for edits because there we should see all new added codes IMO.

### DIFF
--- a/root/static/scripts/common.js
+++ b/root/static/scripts/common.js
@@ -40,7 +40,6 @@ require('./common/components/FingerprintTable');
 require('./common/components/WikipediaExtract');
 require('./common/MB/Control/Autocomplete');
 require('./common/components/ReleaseEvents');
-require('./common/components/WorkArtists');
 require('./common/MB/Control/SelectAll');
 require('./common/components/TagEditor');
 require('./common/components/sidebar/AcousticBrainz');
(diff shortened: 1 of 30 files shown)


yvanzo (approved):
Code looks good to me. Tested with the same entities.

@reosarevok: I updated both PR’s title and ticket to match the current changes.

root/static/scripts/common.js
```diff
@@ -34,6 +34,7 @@ require('./common/components/CommonsImage');
 require('./common/components/FingerprintTable');
 require('./common/components/WikipediaExtract');
 require('./common/MB/Control/Autocomplete');
+require('./common/components/AttributeList');
```
mwiencek, line 37: Instead of adding this to every page, can we link it only where it's needed like was done at https://github.com/metabrainz/musicbrainz-server/pull/2309/commits/cd68deecabec27778605f7fe7aaca8758ddaa4e3?

reosarevok: Sure. Also updated the PR to do the same to the other `CollapsibleList`s

reosarevok:

reosarevok:

mwiencek (approved):

PR #2025, opened 2021-04-01
MBS-11546: Allow Resident Advisor /exchange URLs for releases 

reosarevok:
### Implement MBS-11546

RA has over 500 entries on their RA Exchange podcast series, under /exchange (we have three of them in the DB right now). These shouldn't be blocked, and should autoselect to show notes.

Since we don't support autoselecting to two different types for the same entity, this excludes them from the main RA cleanup and adds a separate cleanup entry for them, to be merged with MBS-9902.

Additionally, made some more capturing groups non-capturing (noticed the issue because it caught me when originally adapting the RA cleanup).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2717,7 +2717,7 @@ const CLEANUPS = {
   },
   'residentadvisor': {
     match: [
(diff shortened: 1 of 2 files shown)


mwiencek (approved):

yvanzo (approved):
:shipit:

PR #2029, opened 2021-04-01
MBS-11548: Show recent additions voted by auto-editors

yvanzo:
# Implement MBS-11548

Require an implicit review by auto-editors through a yes vote or a direct approval of the added cover art to display recent additions on the front page.

Implementation note: Approvals are too rare to refresh recent additions often enough. That’s why it also checks for yes votes (not necessarily the last vote though).

# Checklist for author

* [x] Tested with `pink` data

# Action

1. Inform auto-editors about this new side effect

### DIFF
--- a/lib/MusicBrainz/Server/Data/Release.pm
+++ b/lib/MusicBrainz/Server/Data/Release.pm
@@ -14,9 +14,12 @@ use List::MoreUtils qw( part );
 use List::UtilsBy qw( nsort_by partition_by );
 use MusicBrainz::Server::Constants qw(
     :quality
+    $AUTO_EDITOR_FLAG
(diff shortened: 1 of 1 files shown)


reosarevok:
I don't like this implementation because it feels that, as an autoeditor, I'm supposed to not vote for a correct edit if it contains art that I feel shouldn't be on the front page. We shouldn't discourage already very rare voting.

yvanzo:
@reosarevok: The initial proposition was to take all yes votes into account, not just auto-editors, so it does not prevent non auto-editors to vote for these edits. Let’s keep the discussion under the ticket so that the whole community can take part into it.

PR #2030, opened 2021-04-02
MBS-10534: Cleanup DNB search link

y-young:
# Problem

[MBS-10534](https://tickets.metabrainz.org/browse/MBS-10534): Automatically turn DNB search link into permalink

# Solution

Update `URLCleanup.js` to cleanup DNB search link

# Checklist for author

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1324,12 +1324,18 @@ const CLEANUPS = {
     ],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=nid%3D(\d+).*$/, 'http://d-nb.info/gnd/$1');
-      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=idn%3D(\d+).*$/, 'http://d-nb.info/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=nid%3D([\d-]+).*$/, 'http://d-nb.info/gnd/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=idn%3D([\d-]+).*$/, 'http://d-nb.info/$1');
       return url;
     },
     validate: function (url, id) {
-      return {result: id === LINK_TYPES.otherdatabases.artist};
+      switch (id) {
+        case LINK_TYPES.otherdatabases.artist:
+          return {result: /^http:\/\/d-nb\.info\/gnd\/[\d-]+$/.test(url)};
+        case LINK_TYPES.otherdatabases.release:
+          return {result: /^http:\/\/d-nb\.info\/[\d-]+$/.test(url)};
+      }
+      return {result: false};
     },
   },
   'downloadpurchase': {
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1324,10 +1324,12 @@ const CLEANUPS = {
     ],
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=nid%3D(\d+).*$/, 'http://d-nb.info/gnd/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=nid%3D(\d+).*$/, 'http://d-nb.info/gnd/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=idn%3D(\d+).*$/, 'http://d-nb.info/$1');
+      return url;
     },
-    validate: function (url) {
-      return {result: /^http:\/\/d-nb\.info\/gnd\/(\d+)$/.test(url)};
+    validate: function (url, id) {
+      return {result: id === LINK_TYPES.otherdatabases.artist};
     },
   },
   'downloadpurchase': {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1317,6 +1317,19 @@ const CLEANUPS = {
...
+      new RegExp('^(https?://)?([^/]+\\.)?d-nb\\.info', 'i'),
+      new RegExp('^(https?://)?([^/]+\\.)?dnb\\.de', 'i'),
+    ],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      return url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=nid%3D(\d+).*$/, 'http://d-nb.info/gnd/$1');
+    },
+    validate: function (url) {
+      return {result: /^http:\/\/d-nb\.info\/gnd\/(\d+)$/.test(url)};
+    },
```
reosarevok, line 1331: `/gnd` links (`nid` queries) seem to be exclusively for artists, so we should probably validate that it's only used for artists. See how we do it elsewhere on this file :) You'll also want to change the test to use `only_valid_entity_types: ['artist']`

Also, they also have `idn` queries (https://portal.dnb.de/opac.htm?method=simpleSearch&cqlMode=true&query=idn%3D1227621485), that can be cleaned up to links like http://d-nb.info/1227621485 and which work now, but this would make invalid (https://musicbrainz.org/url/a4eaabd0-55ba-4918-970f-6eeb48aa91d7 for example). Can you make this so that it cleans those up too and allows them for releases? Add a test for them too :)

reosarevok (changes requested):
Thanks for the patch! Left some notes asking for improvements :)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1334,11 +1334,6 @@ const testData = [
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'http://d-nb.info/gnd/129802433',
   },
-  {
-                     input_url: 'http://d-nb.info/gnd/129802433',
-             input_entity_type: 'artist',
-    expected_relationship_type: 'otherdatabases',
-  },
```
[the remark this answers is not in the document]
y-young: This is my initial thought, but it seems to cause test failure. Could you please tell me how should I fix it?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1317,6 +1317,19 @@ const CLEANUPS = {
       return {result: false};
     },
   },
+  'dnb': {
+    match: [
+      new RegExp('^(https?://)?([^/]+\\.)?d-nb\\.info', 'i'),
```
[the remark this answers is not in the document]
y-young: I don't understand, it seems that we need this line to retrict link type to artists. Could you explain further?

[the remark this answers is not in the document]
reosarevok: There's a separate rule in the same file, just named otherdatabases, that also matches d-nb.info. Since you're moving this to be its own rule, it shouldn't also be matched there :)

y-young:
I've modified the code according to your suggestions, but I have no idea how to fix the failed tests, please have a look :)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1327,6 +1327,35 @@ const testData = [
...
+            expected_clean_url: 'http://d-nb.info/gnd/129802433',
+       only_valid_entity_types: ['artist'],
+  },
+  {
+                     input_url: 'https://portal.dnb.de/opac.htm?method=simpleSearch&cqlMode=true&query=idn%3D1227621485',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'http://d-nb.info/1227621485',
+       only_valid_entity_types: ['artist'],
+  },
```
reosarevok, line 1344: This is not for an artist, but for a release :)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1327,6 +1327,35 @@ const testData = [
...
+            expected_clean_url: 'http://d-nb.info/gnd/129802433',
+       only_valid_entity_types: ['artist'],
+  },
+  {
+                     input_url: 'http://d-nb.info/1227621485',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'http://d-nb.info/1227621485',
+       only_valid_entity_types: ['artist'],
+  },
```
reosarevok, line 1358: Also for a release. Same issue with the repeated URL, try http://d-nb.info/1181136512 for example

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1317,6 +1317,21 @@ const CLEANUPS = {
...
+      new RegExp('^(https?://)?([^/]+\\.)?dnb\\.de', 'i'),
+    ],
+    type: LINK_TYPES.otherdatabases,
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=nid%3D(\d+).*$/, 'http://d-nb.info/gnd/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac\.htm\?.*\bquery=idn%3D(\d+).*$/, 'http://d-nb.info/$1');
+      return url;
+    },
+    validate: function (url, id) {
+      return {result: id === LINK_TYPES.otherdatabases.artist};
```
reosarevok, line 1332: You added the `idn` matching for releases, but this still only allows these links on artists :) You should have two checks:
If it matches `http://d-nb.info/gnd/(id)` then it's allowed for artist, otherwise not.
If it matches `http://d-nb.info/(id)` then it's allowed for release, otherwise not.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -1327,6 +1327,35 @@ const testData = [
...
+  {
+                     input_url: 'http://d-nb.info/gnd/129802433',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'http://d-nb.info/gnd/129802433',
+       only_valid_entity_types: ['artist'],
+  },
```
reosarevok, line 1351: The test issue here is because it expects the same clean URL. Just give it a separate one that is also valid. For example, http://d-nb.info/gnd/2026867-1 is a valid artist link :)

reosarevok (changes requested):

reosarevok (approved):
I think this looks good now. Let's see what the others think :)

yvanzo (approved):

yvanzo:
Thank you @y-young!

I squashed and merged your commits because only the first one had a relevant message, others were just making it better.

Also I wrote a more detailed commit message that describes the features:
```markdown
- Autoselect “other databases” relationship type with DNB URL for
  artist and release.

- Convert DNB Search URLs at `portal.dnb.be` to the permalink format
  `http://d-nb.info/[gnd/]<id>` used wtihin DNB if possible, that is
  when a `nid` is provided in the search `query` query parameter.

- Validate DNB URLs in permalink format with:
  - path prefix `gnd/` with MB artists
  - no path prefix with MB releases
```

PR #2031, opened 2021-04-02
Increase the default relationship page limit sizes

mwiencek:
Amends 12b43e8330b0feda64c636665735c7c3c931b7b5

Based on feedback, this increases the initial table size for each link type group from 25 to 100. Testing this against London Symphony Orchestra shows a ~1.67x slowdown from 2.193s to 3.664s.

The page size of the paged pages has been increased from 100 to 250. Testing against /artist/38712b4c-0fd4-4c65-8c7a-45676fecc973/relationships?direction=1&link_type_id=150&page=3 shows a slowdown of ~2.06x from 1.145s to 2.361s.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Load.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Load.pm
@@ -35,7 +35,7 @@ parameter 'allow_integer_ids' => (
     default => sub { 1 },
 );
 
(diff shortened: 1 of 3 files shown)


reosarevok (approved):
Seems sensible, and given it's already multiple seconds, the difference doesn't seem too significant IMO.

PR #2033, opened 2021-04-03
MBS-11542 / MBS-11552: Allow and cleanup new Classical Archives links + add validation

reosarevok:
### Implement MBS-11542, MBS-11552

This allows parsing the new version of classicalarchives.com links for Composer, Performer and Work, and maps them to the (still valid and meant to remain valid long term, according to CA themselves) old link versions we already accepted.
This implementation skips converting Album links, since those can't be automatically converted between the two versions. As such, we just allow both versions (old and new) for album/Album links for now.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1102,16 +1102,54 @@ const CLEANUPS = {
...
+       * newca entities match except Performer, where the old type
+       * is determined by the first letter of the ID: ensemble (e)
+       * or artist (p)
+       * newca Album links are blocked for now because they can't be converted
+       */
+      url = url.replace(/^https:\/\/www\.classicalarchives\.com\/newca\/#!\/Composer\/([^\/?#]+)/, 'https://www.classicalarchives.com/composer/$1.html');
+      url = url.replace(/^https:\/\/www\.classicalarchives\.com\/newca\/#!\/Work\/([^\/?#]+)/, 'https://www.classicalarchives.com/work/$1.html');
+      url = url.replace(/^https:\/\/www\.classicalarchives\.com\/newca\/#!\/Performer\/e([^\/?#]+)/, 'https://www.classicalarchives.com/ensemble/$1.html');
+      url = url.replace(/^https:\/\/www\.classicalarchives\.com\/newca\/#!\/Performer\/p([^\/?#]+)/, 'https://www.classicalarchives.com/artist/$1.html');
+      url = url.replace(/^https:\/\/www\.classicalarchives\.com\/(album|artist|composer|ensemble|work)\/([^\/?#]+)(?:.*)?$/, 'https://www.classicalarchives.com/$1/$2');
```
yvanzo, line 1133: Since you normalized URL’s scheme and authority above, and no more change is being made to them, you can just capture them instead of rewriting them, so as to make actual changes more visible:

```suggestion
      url = url.replace(/^(https:\/\/www\.classicalarchives\.com)\/newca\/#!\/Composer\/([^\/?#]+)/, '$1/composer/$2.html');
      url = url.replace(/^(https:\/\/www\.classicalarchives\.com)\/newca\/#!\/Work\/([^\/?#]+)/, '$1/work/$2.html');
      url = url.replace(/^(https:\/\/www\.classicalarchives\.com)\/newca\/#!\/Performer\/e([^\/?#]+)/, '$1/ensemble/$2.html');
      url = url.replace(/^(https:\/\/www\.classicalarchives\.com)\/newca\/#!\/Performer\/p([^\/?#]+)/, '$1/artist/$2.html');
      url = url.replace(/^(https:\/\/www\.classicalarchives\.com)\/(album|artist|composer|ensemble|work)\/([^\/?#]+)(?:.*)?$/, '$1/$2/$3');
```

yvanzo (commented):
Shouldn’t new Album URLs be allowed too?

Sure it would allow for duplicates but at least it does not get in the way of editors.

reosarevok:
I was expecting us to actually talk about this during review and figure out what to do. Should we block one of the two and ask for the other, rather than allowing both?

root/static/scripts/edit/URLCleanup.js
reosarevok: Good idea!

yvanzo:
If the only reason is to prevent duplicates, it does not seem worth blocking editors, because getting an URL from the other one requires editors to use a specific version of the CA website at the moment. These URLs could be programatically deduplicated later on. It would not be the first time we allow for some duplicates for websites that use two URL forms (YouTube channels…).

reosarevok:
Modified this to allow newca Album links as discussed.

yvanzo (approved):
Thanks!

mwiencek (approved):

PR #2036, opened 2021-04-05
MBS-11367 (II): Restore turning autocomplete on at the start

reosarevok:
### Fixes up MBS-11367

The previous patch (https://github.com/metabrainz/musicbrainz-server/pull/2000) removed this section and replaced it with an onChange, but both are needed. This ensures the autocomplete is still set to on on page load.


reosarevok:
We're doing this differently, it seems, so closing this.

PR #2040, opened 2021-04-06
MBS-11457: Drop the series ordering_attribute column

reosarevok:
### Implement MBS-11457

This was added back when we thought we would have different attributes to store item numbers for part of series relationships. We don't, so this is always one and the same, 788 (number).

This changes the views to hardcode to 788 and removes the ordering_attribute column elsewhere. It keeps the SERIES_ORDERING_ATTRIBUTE constant on JS since that seems to be used in fields and just points to the 'number' attribute, but removes it on Perl because it was only used to store the ID under ordering_attribute and never use it again, AFAICT.

# Problem

:beginner: Reference related tickets with MBS-XXX at least.

# Solution

:beginner: If you have a lot to say, be more detailed in this section.

# Checklist for author

* [ ] Delete irrelevant parts of this template
* [ ] Update the [WikiDocs pages](https://wiki.musicbrainz.org/Category:WikiDocs_Page)
  * [ ] :beginner: Page...
* [ ] :beginner: Task...

# Action

1. Make updated WikiDocs pages visible on MusicBrainz website
2. :beginner: Action...


admin/sql/updates/20210406-mbs-11457.sql
```diff
@@ -0,0 +1,81 @@
...
+    FROM l_series_work lsw
+    JOIN series s ON s.id = lsw.entity0
+    JOIN link l ON l.id = lsw.link
+    JOIN link_type lt ON (lt.id = l.link_type AND lt.gid = 'b0d44366-cdf0-3acb-bee6-0f65a77a6ef0')
+    LEFT OUTER JOIN link_attribute_text_value latv ON (latv.attribute_type = 788 AND latv.link = l.id)
+    ORDER BY series, link_order;
+
+-- II) Drop the ordering_attribute column
+
+ALTER TABLE series DROP COLUMN ordering_attribute;
```
mwiencek, line 79: Unfortunately, there are also sir functions/triggers that reference this column. I think the best course of action here would be to add `CASCADE` and provide instructions to reinstall the sir bits. (We'll need a schema change branch there in any case; tests are currently failing due to that.)

```suggestion
ALTER TABLE series DROP COLUMN ordering_attribute CASCADE;
```

mwiencek (changes requested):

admin/sql/updates/20210406-mbs-11457.sql
reosarevok: Made the change here - @yvanzo, can you deal with the sir bits?

mwiencek (approved):

mwiencek:
@brainzbot test this please

reosarevok:

reosarevok:
Rebased this on this year's branch, moved from schema 26 to 27, regenerated. Should probably give it a second look anyway, and @yvanzo, we need to look into the sir changes :)

reosarevok:
Hmm, actually, I understand the needed sir changes are already merged from last year and just need releasing?

mwiencek:
> Hmm, actually, I understand the needed sir changes are already merged from last year and just need releasing?

I think they are already released.  https://github.com/metabrainz/sir/tags says version 2.1.0 included the change to "Stop indexing ordering attribute for series."  Right now we are running v2.1.1 on floyd.

admin/sql/updates/20210406-mbs-11457.sql
yvanzo: SIR has been updated in https://github.com/metabrainz/sir/pull/117. It just requires up-to-date `mbdata` models.

yvanzo (approved):
LGTMBDNT :ship:

PR #2041, opened 2021-04-06
MBS-9376: Indicate that edit searches can be bookmarked + link docs

reosarevok:
### Implement MBS-9376

This adds some help text to the top of edit searches that directs users to a how to page for using the search (in the works) and informs them that the searches can be bookmarked to save them.

### DIFF
--- a/root/edit/search_macros.tt
+++ b/root/edit/search_macros.tt
@@ -183,7 +183,7 @@
   </span>
 
   <span>
-    (<a href="[% c.uri_for('/doc/How_to_Search_for_Edits#Date_format') %]" target="_blank">[% l('help') %]</a>)
+    (<a href="[% c.uri_for('/doc/How_to_Use_the_Edit_Search#Dates') %]" target="_blank">[% l('help') %]</a>)
   </span>
 [% END %]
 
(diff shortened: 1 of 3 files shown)


mwiencek (approved):

root/edit/search_macros.tt
```diff
@@ -4,6 +4,12 @@
 </span>
 [% END %]
 
+[% BLOCK search_heading %]
+  <h1>[% l('Search for Edits') %]</h1>
+  <p>[% l('The filters below allow you to find only edits fulfilling specific criteria. See the {doc_link|documentation} for more info about using the filters.', {doc_link => doc_link('How to Use the Edit Search')}) %]</p>
+  <p>[% l('If you want to save an edit search for future use, you can just bookmark it!') %]</p>
```
yvanzo, line 10: That would probably be more helpful to provide an actual button to bookmark it or at least to copy/share the URL.

reosarevok: Don't most browsers just have a trivial way to do that from the page itself, which would be easier than copying the URL?

yvanzo: Bookmarking seems _trivial_ to me indeed, so it did not seem worth mentioning it. Having a button to copy the URL to share with the community forum seem more useful, but that doesn’t seem _trivial_ to implement, so probably not worth it either.

yvanzo (approved):

PR #2042, opened 2021-04-06
Pass public editor privs when converting editor to JSON

reosarevok:
EditorT expects privileges to be passed, and there's no real reason why we shouldn't do this anyway.

### DIFF
--- a/lib/MusicBrainz/Server/Constants.pm
+++ b/lib/MusicBrainz/Server/Constants.pm
@@ -37,7 +37,7 @@ our %EXPORT_TAGS = (
     vote               => _get(qr/^VOTE_/),
     edit_status        => _get(qr/^STATUS_/),
     access_scope       => _get(qr/^ACCESS_SCOPE_/),
-    privileges         => _get(qr/_FLAG$/),
+    privileges         => _get(qr/_FLAGS?$/),
     language_frequency => _get(qr/^LANGUAGE_FREQUENCY/),
     script_frequency   => _get(qr/^SCRIPT_FREQUENCY/),
     election_status => [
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Constants.pm
```diff
@@ -345,6 +345,14 @@ Readonly our $LOCATION_EDITOR_FLAG          => 256;
 Readonly our $BANNER_EDITOR_FLAG            => 512;
 Readonly our $EDITING_DISABLED_FLAG         => 1024;
 Readonly our $ADDING_NOTES_DISABLED_FLAG    => 2048;
+# If you update this, also update root/utility/sanitizedEditor.js
```
mwiencek, line 348: Maybe also backreference the same in root/utility/sanitizedEditor.js

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -324,6 +328,7 @@ sub TO_JSON {
         gravatar => $self->gravatar,
         id => $self->id,
         name => $self->name,
+        privileges => $self->public_privileges,
```
mwiencek, line 331: Just in case Perl is being stupid:
```suggestion
        privileges => 0 + $self->public_privileges,
```

reosarevok: We do just `privileges => $self->privileges` above though. If we change them then I guess both should change?

yvanzo: It’s unneeded for `$self->privileges` because the class attribute `privileges` is of type `Int` (line 35). I doubt this is needed here either but did not check if `&` operands and call to subroutine `public_privileges` make any change.

mwiencek: The Moose attribute type is just for validation; it doesn't affect how JSON encoder works:

```Perl
#!/usr/bin/env perl

package Editor;
use Moose;
has privileges => ( is => 'ro', isa => 'Int' );
1;

package Main;
use JSON qw( encode_json );
my $x = Editor->new( privileges => '123' );
print encode_json({ privileges => $x->privileges });
1;
```

still outputs `{"privileges":"123"}`, for example.

The determinant factor is whether Perl views the scalar as a number, and you can force that with the `0 +` trick. (You can use a package like `Devel::Peek` to see if there's an "IV" (integer value) item associated with the scalar.)

So yes, it would make sense to add it to `privileges => $self->privileges` too.

reosarevok: Added it to both in a second commit.

mwiencek (approved):

yvanzo (approved):

PR #2043, opened 2021-04-07
MBS-11562: Actually randomize when picking rand sort for edit search

reosarevok:
### Fix MBS-11562

This was just not selecting any sort when the 'rand' order was selected.
PSQL docs used to say "If sorting is not chosen, the rows will be returned in random order" but that's misleading and has since been corrected to "the rows will be returned in an unspecified order". For at least some queries, the unspecified order is literally "ordered by ID".
This implements an actual ORDER BY RANDOM() when 'rand' is selected. This works fine for smaller queries, but takes forever and is likely to time out in bigger ones. I'd say that keeping it as is is not an option but we might also just decide to drop the random option to avoid the issue entirely. I suspect it's not particularly popular anyway.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Query.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Query.pm
@@ -166,6 +166,8 @@ sub as_string {
     } elsif ($self->order eq 'vote_closing_desc') {
         $order = 'ORDER BY edit.expire_time DESC';
         $extra_conditions = ' AND edit.close_time IS NULL';
+    } elsif ($self->order eq 'rand') {
+        $order = 'ORDER BY RANDOM()';
     }
 
     return 'SELECT edit.*, edit_data.data ' .


mwiencek:
I'm honestly not sure this is very useful. I'd be in favor of either removing the option or renaming it to any/unspecified.

yvanzo (commented):
`rand` order has been added in https://github.com/metabrainz/musicbrainz-server/commit/98fa403d2fd38b54985c46743589b8d7988cc04a.
It does not seem to match any use case: https://tickets.metabrainz.org/browse/MBS-3313?focusedCommentId=18000#comment-18000
Maybe the only use case would be to make the search a bit faster?

I would be fine with removing this option.

mwiencek:
> Maybe the only use case would be to make the search a bit faster?

Yes. If you don't care about the order then you can let PG return things in whatever order is most efficient. But I don't know how common this option is given that it's not the default and has to be explicitly selected.

reosarevok:
@yvanzo, @mwiencek: so, do we just remove the option? :)

yvanzo:
@reosarevok From the two above comments, plus chaban’s comment to the ticket, I would say yes.

mwiencek:
I'd check some of the edit searches on the voting page (in production) and see if they feel any faster with rand.  If so, we could keep it and rename it with "(possibly faster)" in the name or something.  Perhaps more people would use it then.  If it's just as slow, let's remove it, but if it helps at all then I don't see a reason to.

mwiencek:
A lot don't seem to show any difference, but some are kinda noticeable

```
All edits…related to artists in my subscriptions
difference: -32.358167886734 s
percentage difference: -163.039739635181%

All edits from beginner editors made less than 2 weeks ago
difference: -5.60369777679443 s
percentage difference: -97.0922690450369%

All "Add release/medium" edits by beginner editors made less than 2 weeks ago
difference: -11.0058159828186 s
percentage difference: -85.0589186805625%

All open relationship removals…related to artists in my subscriptions
difference: -2.17790699005127 s
percentage difference: -84.2374320080523%

All edits from beginner editors
difference: -3.72030997276306 s
percentage difference: -78.7382363078428%

All "Add release/medium" edits by beginner editors…related to artists in my subscriptions
difference: -11.3829362392426 s
percentage difference: -72.8982303967414%

All open edits
difference: -2.39630484580994 s
percentage difference: -70.724853983723%

All open unreviewed edits (0 votes)…related to artists in my subscriptions
difference: -2.2161762714386 s
percentage difference: -70.1505854166699%

All edits from beginner editors…related to artists in my subscriptions
difference: -2.46979784965515 s
percentage difference: -69.5296396827624%

Unreviewed edits (0 votes) that will close in less than a day
difference: -2.04933595657349 s
percentage difference: -61.2670938392042%
```

reosarevok:
So, rename and keep?

reosarevok:
Closing this, sending a PR just changing the name instead :)

PR #2044, opened 2021-04-07
MBS-11564: Don't block Niconi Commons links with Nicovideo cleanup

reosarevok:
### Fix MBS-11564

commons.nicovideo links are separate from the general nicovideo system and should not be autoselected as streamingfree (if they should be autoselected at all then probably as downloadfree, but that is an issue for a future PR). They also shouldn't be blocked, which is what the cleanup code was currently doing.
For now, this patch just makes the system ignore them so that they can be entered as needed.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2364,7 +2364,10 @@ const CLEANUPS = {
     type: LINK_TYPES.score,
   },
   'niconicovideo': {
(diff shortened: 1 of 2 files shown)


yvanzo (changes requested):
Please add a test with `expected_relationship_type: undefined`.

reosarevok:
Done :)

yvanzo (approved):
Thanks!

mwiencek (approved):

PR #2045, opened 2021-04-07
MBS-11566: Ensure consistent ordering of appearances sections

reosarevok:
### Fix MBS-11566

These need to be sorted by target type, otherwise the different sections jump around the page on every reload.

Tested and it seems to work - Flow hates this though, I need help fixing it. @mwiencek? :)


mwiencek (approved):
Fixed Flow

yvanzo (approved):

PR #2047, opened 2021-04-07
MBS-11563: Invalidate entity_link_type_counts on relationship inserts

mwiencek:
12b43e8330b0feda64c636665735c7c3c931b7b5 added some caching for the number of relationships for a given entity & link type, so that we don't have to recount them on every page load. While that commit made sure to invalidate the cached entries on relationship updates and deletions, it forgot to do so for insertions. This commit does that.


reosarevok (approved):
LGTM but didn't test

yvanzo (approved):

PR #2049, opened 2021-04-08
MBS-11571: Warn about e-mail as a username

reosarevok:
### Implement MBS-11571

While a user should be able to use an email address string as their username if they so desire, they should only do it if they are sure it's ok with them for it to be public.
This adds a warning for that scenario.

### DIFF
--- a/root/static/scripts/register.js
+++ b/root/static/scripts/register.js
@@ -1,0 +1,27 @@
+/*
+ * @flow
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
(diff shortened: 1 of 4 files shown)


root/static/scripts/register.js
```diff
@@ -0,0 +1,27 @@
...
+
+import $ from 'jquery';
+
+import debounce from './common/utility/debounce';
+
+$(function () {
+  function warnAboutEmailAsUsername() {
+    const username =
+    $('#id-register\\\.username').val();
+    const isPossibleEmail = /\w+@\w+\.\w+/.test(username);
```
reosarevok, line 18: This uses an intentionally wide regexp because, well, it doesn't really cause any problems if the string happens to be an invalid username - the user can just ignore the warning and move on.

mwiencek (approved):

yvanzo (approved):
LGTM but did not test.

PR #2050, opened 2021-04-09
MBS-11524: Disallow https://*.bandcamp.com/ URLs at release and recording level

y-young:
# Problem

MBS-11524: Disallow https://*.bandcamp.com/ URLs at release and recording level

# Solution

Restrict Bandcamp artist links

# Checklist for author

# Action

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3658,12 +3658,22 @@ for (const relUuid of relationshipTypesByEntityType.release) {
     }
     if (/^(https?:\/\/)?([^\/]+)\.bandcamp\.com\/?$/.test(url)) {
       return {
-        error: l(
-          `The artist page of Bandcamp normally has no entries
-           for specific releases,
-           so adding these links to a release is currently blocked.
-           Please add this Bandcamp link to the artist instead,
-           if appropriate.`,
+        error: exp.l(
+          `This is a Bandcamp profile, not a page for a specific
+           release. Even if it shows this release right now,
+           that can change when the artist releases another.
+           Please find and add the appropriate release page
+           (“{album_url_pattern}” or “{single_url_pattern}”)
+           instead, and feel free to add this profile link
+           to the appropriate artist or label.`,
+          {
+            album_url_pattern: (
+              <span className="url-quote">{'/album'}</span>
+            ),
+            single_url_pattern: (
+              <span className="url-quote">{'/track'}</span>
+            ),
+          },
         ),
         result: false,
       };
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3682,18 +3682,25 @@ for (const relUuid of relationshipTypesByEntityType.release) {
   };
 }
 
-// Disallow https://*.bandcamp.com/ URLs at release group level
-for (const relUuid of relationshipTypesByEntityType.release_group) {
+// Disallow https://*.bandcamp.com/ URLs at recording level
+for (const relUuid of relationshipTypesByEntityType.recording) {
   const relRule = validationRules[relUuid];
   validationRules[relUuid] = function (url) {
     if (/^(https?:\/\/)?([^\/]+)\.bandcamp\.com\/?$/.test(url)) {
       return {
-        error: l(
-          `The artist page of Bandcamp normally has no entries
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3653,6 +3656,38 @@ for (const relUuid of relationshipTypesByEntityType.release) {
...
     }
+    if (/^(https?:\/\/)?([^\/]+)\.bandcamp\.com\/?$/.test(url)) {
+      return {
+        error: l(
+          `The artist page of Bandcamp normally has no entries
+           for specific releases,
+           so adding these links to a release is currently blocked.
+           Please add this Bandcamp link to the artist instead,
+           if appropriate.`,
+        ),
```
reosarevok, line 3667: A page like https://gaerea.bandcamp.com/ *looks* like an album page, so we should be more explicit. What about something like:

```suggestion
        error: exp.l(
          `This is a Bandcamp profile, not a page for a specific
           release. Even if it shows this release right now,
           that can change when the artist releases another.
           Please find and add the appropriate release page
           (“{album_url_pattern}” or “{single_url_pattern}”)
           instead, and feel free to add this profile link
           to the appropriate artist or label.`,
          {
            album_url_pattern: (
              <span className="url-quote">{'/album'}</span>
            ),
            single_url_pattern: (
              <span className="url-quote">{'/track'}</span>
            ),
          },
        ),
```

@yvanzo might have more comments on this though, so let's maybe wait to see what he thinks :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3653,6 +3656,38 @@ for (const relUuid of relationshipTypesByEntityType.release) {
...
+  validationRules[relUuid] = function (url) {
+    if (/^(https?:\/\/)?([^\/]+)\.bandcamp\.com\/?$/.test(url)) {
+      return {
+        error: l(
+          `The artist page of Bandcamp normally has no entries
+           for specific release groups,
+           so adding these links to a release group is currently blocked.
+           Please add this Bandcamp link to the artist instead,
+           if appropriate.`,
+        ),
```
reosarevok, line 3687: Similar to the release comment, but mentioning release groups instead I guess?

reosarevok:
See the test - this is not being blocked for recording. It probably should be, but that one should only mention /track links :)

reosarevok (changes requested):

root/static/scripts/edit/URLCleanup.js
y-young: There're no release group pages but review pages, should we mention reviews instead?

reosarevok: Oh, actually, this is *already* disallowed for RGs, because it auto-selects to review and this is not a review. We could update the validation there to be more specific and say that profiles should be added to artists/labels and album/track pages to releases or recordings, but I'm also ok with just not touching RGs.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3653,6 +3656,55 @@ for (const relUuid of relationshipTypesByEntityType.release) {
...
+  const relRule = validationRules[relUuid];
+  validationRules[relUuid] = function (url) {
+    if (/^(https?:\/\/)?([^\/]+)\.bandcamp\.com\/?$/.test(url)) {
+      return {
+        error: exp.l(
+          `This is a Bandcamp profile, not a page for a specific
+           recording. Even if it shows a single recording right now,
+           that can change when the artist releases another.
+           Please find and add the appropriate recording page
+           “{single_url_pattern}”
```
reosarevok, line 3696: I might still keep parenthesis around this, it reads better IMO.

reosarevok (approved):
This seems basically ready to me now. Maybe make the string change I suggested? If @yvanzo agrees.

root/static/scripts/edit/URLCleanup.js
yvanzo: @y-young: yes to parenthesis, please.

yvanzo (approved):

yvanzo:
Thank you @y-young, I updated PR’s title and initial comment to match changes made to the ticket and to the code (that is replacing “release group” with “recording”). I also squashed your commits at merge time because they were just reworks of your first commit.

PR #2051, opened 2021-04-09
MBS-9251: Display ended places differently on the map

reosarevok:
### Implement MBS-9251

This shows all ended places with a grey marker in the AreaPlaces map. It also displays '(closed)' by the place names, which is sometimes the only way to tell if a place is ended at all:  if several places have the exact same coordinates (e.g. one replaced the other at the same location) they will be shown as a group marker only.


mwiencek (approved):

yvanzo (approved):
Tested with the example provided in the ticket, that looks nice!

PR #2022, opened 2021-03-27
MBS-8807 / MBS-11538: Collections data missing in some sidebars

jesus2099:
# Problem

https://tickets.metabrainz.org/browse/MBS-8807
https://tickets.metabrainz.org/browse/MBS-11538

Some tabs/sub-pages that contain sidebars and thus try to display collections have no collections data from the server, therefore display broken collection status (as if no collections, sort of).

- **Artist** missing collection data in **Subscribers** sub-page
- **Area** missing collection data in **Events** tab
- **Instrument** missing collection data in **Artists** tab
- **Label** missing collection data in **Subscribers** sub-page
- **Release** missing collection data in **Aliases** tab
- **Series** missing collection data in **Subscribers** sub-page
- **Work** missing collection data in **Aliases** tab

# Solution

I found how it works, you have to list the tabs/sub-pages that contain sidebar to send them collection data.
So I just added those keywords and also reordered **Release** and **Event** tab list (the latter was missing nothing) that was in unusual order compared to others:

    show collections details tags aliases <subscribers> <other tabs in left to right visual order>

# Misc

I have tested my code on `localhost:5000`.

I had to disable (comment out) this block, to be able to create collections (to allow `POST` forms):
https://github.com/metabrainz/musicbrainz-server/blob/37e4dfab3d9d2d42e2bf844a82e1b8fa6d76e335/lib/MusicBrainz/Server/Controller/Root.pm#L361-L376

PR #2052, opened 2021-04-09
MBS-11573: Load collections for ratings sidebars

reosarevok:
### Fix MBS-11573

This one was missed by https://github.com/metabrainz/musicbrainz-server/pull/2022/ it seems.


reosarevok:
Also made the change for places in https://github.com/metabrainz/musicbrainz-server/pull/2005 so that we don't have the same problem with those :)

mwiencek (approved):

yvanzo (approved):

PR #2054, opened 2021-04-12
MBS-9631: Add 1200px to available image sizes

reosarevok:
### Implement MBS-9631

Seems the IA has triggered a reindex to get all 1200px thumbnails generated (see CAA-88), so it seems like a good time to add this.
I expect we still want large-thumbnail, not huge-thumbnail, in root/static/scripts/common/coverart.js for the thumbnail URL, but added it everywhere else we have thumbnails - including to data as huge-thumbnail in case userscripts or whatnot want to use it.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Artwork.pm
+++ b/lib/MusicBrainz/Server/Entity/Artwork.pm
@@ -101,15 +101,17 @@ sub filename
 sub image { my $self = shift; return $self->_url_prefix . "." . $self->suffix; }
 sub small_thumbnail { my $self = shift; return $self->_url_prefix . "-250.jpg"; }
 sub large_thumbnail { my $self = shift; return $self->_url_prefix . "-500.jpg"; }
+sub huge_thumbnail { my $self = shift; return $self->_url_prefix . "-1200.jpg"; }
 
 # These accessors allow for requesting thumbnails directly from the IA,
 # bypassing our artwork redirect service. These are suitable for any <img>
 # tags in our templates, avoiding a pointless 307 redirect and preventing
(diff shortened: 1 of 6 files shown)


lib/MusicBrainz/Server/Entity/Artwork.pm
```diff
@@ -101,22 +101,26 @@ sub filename
 sub image { my $self = shift; return $self->_url_prefix . "." . $self->suffix; }
 sub small_thumbnail { my $self = shift; return $self->_url_prefix . "-250.jpg"; }
 sub large_thumbnail { my $self = shift; return $self->_url_prefix . "-500.jpg"; }
+sub huge_thumbnail { my $self = shift; return $self->_url_prefix . "-1200.jpg"; }
```
mwiencek, line 104: "huge" is probably fine as an internal-only name, though for reference we did decide against using that in the API (index.json) and just opted for 250/500/1200 as names (with "small" and "large" kept for backwards compatibility).

mwiencek (approved):

lib/MusicBrainz/Server/Entity/Artwork.pm
reosarevok: We could also rename these to thumbnail_250 etc, but it might break some scripts or something if we change it for the `data-` attributes.

yvanzo (approved):

lib/MusicBrainz/Server/Entity/Artwork.pm
mwiencek: Yeah, I'm fine with it how it is, just wanted to point it out so we don't start using it in public pages/APIs in the future.

PR #2055, opened 2021-04-12
MBS-1658 (Impl.): Add free comment field to collection entries

reosarevok:
### Implement MBS-1658

WIP, will take time, opening a draft PR so we do not forget about this.


lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -429,6 +429,17 @@ sub set_collaborators {
...
 
+sub set_comment {
+    my ($self, $collection_id, $entity_id, $entity_type, $comment) = @_;
+
+    $self->sql->do(<<~"SQL", $comment, $collection_id, $entity_id);
+        UPDATE editor_collection_$entity_type
+           SET comment = ?
+         WHERE collection = ?
+           AND $entity_type = ?
+        SQL
```
mwiencek, line 440: I understand we validate `$entity_type` in the WS endpoint by checking for its presence in `%ENTITIES`, but since it still comes from an untrusted WS query parameter I'd feel more comfortable if we quoted the identifiers here, just in case the validation fails -- perhaps after a future refactoring, or if we end up using `set_comment` in another location that doesn't validate things properly.
```suggestion
    my $dbh = $self->sql->dbh;
    my $table = $dbh->quote_identifier("editor_collection_$entity_type");
    my $entity_column = $dbh->quote_identifier("$entity_type");
    $self->sql->do(<<~"SQL", $comment, $collection_id, $entity_id);
        UPDATE $table
           SET comment = ?
         WHERE collection = ?
           AND $entity_column = ?
        SQL
```

lib/MusicBrainz/Server/Controller/WS/js/SetCollectionComment.pm
```diff
@@ -0,0 +1,71 @@
...
+    $c->forward(
+        '/ws/js/check_login',
+        l('You must be logged in to edit collection comments. {url|Log in} ' .
+         'first, and then try again.',
+         { url => { href => $c->uri_for_action('/user/login'), target => '_blank' } }
+        ),
+    );
+
+    $c->model('MB')->with_transaction(sub {
+        $c->model('Collection')->set_comment(
```
mwiencek, line 52: Unless I'm missing something, this endpoint doesn't appear to check that permission to edit the collection comment exists.  I only see it being checked by the UI.

root/components/list/AreaList.js
```diff
@@ -19,32 +22,51 @@ import {
...
 
 const AreaList = ({
   areas,
+  canEditCollectionComments,
   checkboxes,
+  collectionComments,
+  collectionId,
   mergeForm,
   order,
+  showCollectionComments = false,
```
mwiencek, line 42: Is there any case where we'd be passing `collectionComments` and want this to be false?  Seems like we can just show them if `collectionComments` is non-empty.

root/components/list/AreaList.js
```diff
@@ -19,32 +22,51 @@ import {
...
+          sortable: sortable,
+          title: l('Area'),
+        })
+      ) : (
+        defineNameColumn<AreaT>({
+          order: order,
+          sortable: sortable,
+          title: l('Area'),
+        })
+      );
```
mwiencek, line 69: This block is repeated for every entity list, so it'd be nice to make it less verbose if possible.  Can we perhaps perform the ternary check inside `defineNameAndCommentColumn` instead?

mwiencek (commented):
Not sure about submitting the comment by blurring the textarea - an actual "submit" button next to the field might be more obvious/usable?

root/components/list/AreaList.js
reosarevok: IIRC my worry was that this should still work if no entities in the collection have a comment yet.

root/components/list/AreaList.js
reosarevok: You mean making `defineNameAndCommentColumn` call `defineNameColumn` if there's no comments and whatnot?

lib/MusicBrainz/Server/Controller/WS/js/SetCollectionComment.pm
reosarevok: Whoops. Good catch.

reosarevok:
> Not sure about submitting the comment by blurring the textarea - an actual "submit" button next to the field might be more obvious/usable?

IIRC we talked about this back in the day and you liked this option :D But I'm not against a submit button either. What would happen on blur in that case? Nothing? (if so then we'd need a cancel button too)

Aerozol:
Hmm, it’s not making me scream, but I don’t know about adding a ‘Add Comment’ line to every item. Seems very busy if you’re not interested in adding comments.

Hate to get all column heavy, but seems like another column might be better… with the header ‘Comments’ and then ‘_Edit_’ in each row (no pencil icon please)

reosarevok:
IIRC we did it this way because most collection types already have ridiculous amounts of columns. Ideally I guess we'd have a way to filter those so that you don't need to show all the columns when you don't need them, but that's a lot of work 😅

Aerozol:
Definitely a lot of columns… but if anything I would take others out to make room for the comment column, or group some others together. No perfect solution, but some of those other columns are surely not used as often.

My mockups for the redesign have custom columns but I also though “wow glad I don’t have to code this” ;)

Maxr1998:
The extra line below the entity name doesn't seem too bad considering that (for example) label disambiguations are so long that each line takes up a lot of space already anyway. Of course, hiding or grouping columns might be another solution.

Either way, would love to have this added eventually, was just about to request this on the issue tracker earlier.

elliotchance:
I can't wait for this feature! It will be an absolute game changer for keeping notes against stuff in MusicBrainz.

elliotchance:
@reosarevok what's holding this up?

reosarevok:
Mostly the same thing that's holding most PRs back - there's just too many and the team cannot find time to go through the backlog :/

reosarevok:
Rebased the code and fixed conflicts at least so it's actionable again.

PR #2060, opened 2021-04-13
Fix Catalyst warning: ensure parent format id is not undef

reosarevok:
Some top level formats (like SACD) have no parent format, yet don't appear on the medium format map since they aren't supported by the schema.org enum (https://schema.org/MusicReleaseFormatType). As such, we were trying to access the map with an undef parent format id.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Release.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Release.pm
@@ -100,9 +100,10 @@ sub medium_format {
     # case of it not being one of these few formats. I'm not sure of the
     # best mitigation for either problem.
     my $name;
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Release.pm
```diff
@@ -100,9 +100,10 @@ sub medium_format {
     # case of it not being one of these few formats. I'm not sure of the
     # best mitigation for either problem.
     my $name;
+    my $parent_format_id = $format->parent ? $format->parent->id : $format->parent_id;
```
mwiencek, line 103: I also don't understand why `$format->parent_id` wouldn't always be identical to `$format->parent->id` here, but that's unrelated.

mwiencek (approved):

yvanzo (approved):

PR #2061, opened 2021-04-13
Fix Catalyst warning: don't pass undef review text to markdown

reosarevok:
CB supports rating-only reviews (without a review text). Those have an undef body, but we're still passing that through `markdown()`, causing warnings.

Warnings:
```
[warning] GET http://localhost:5000/release-group/cc053745-c447-3566-8f27-bed5438c9133 caused a warning: Use of uninitialized value $self in string ne at /home/reosarevok/perl5/lib/perl5/Text/Markdown.pm line 190.
[warning] GET http://localhost:5000/release-group/cc053745-c447-3566-8f27-bed5438c9133 caused a warning: Use of uninitialized value $text in substitution (s///) at /home/reosarevok/perl5/lib/perl5/Text/Markdown.pm line 270.
[warning] GET http://localhost:5000/release-group/cc053745-c447-3566-8f27-bed5438c9133 caused a warning: Use of uninitialized value $text in substitution (s///) at /home/reosarevok/perl5/lib/perl5/Text/Markdown.pm line 271.
```

### DIFF
--- a/lib/MusicBrainz/Server/Data/CritiqueBrainz.pm
+++ b/lib/MusicBrainz/Server/Data/CritiqueBrainz.pm
@@ -7,6 +7,7 @@ use JSON;
 use Text::Markdown qw( markdown );
 use Text::Trim qw( trim );
 use URI;
+use MusicBrainz::Server::Data::Utils qw( non_empty );
 use MusicBrainz::Server::Translation qw( l );
 use MusicBrainz::Server::Validation qw( encode_entities );
 use aliased 'MusicBrainz::Server::Entity::CritiqueBrainz::Review';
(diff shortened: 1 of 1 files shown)


yvanzo (approved):

lib/MusicBrainz/Server/Data/CritiqueBrainz.pm
```diff
@@ -63,7 +63,7 @@ sub _parse_review {
     return Review->new(
         id => $data->{id},
         created => DateTime->from_epoch(epoch => str2time($data->{created})),
-        body => markdown($data->{text}),
+        body => $data->{text} ? markdown($data->{text}) : '',
```
mwiencek, line 66: I think it'd be good practice to use `non_empty` here to avoid problems with `'0'` (but I'm guessing there's a character minimum on reviews anyway).

mwiencek (approved):

PR #2062, opened 2021-04-14
Fix Catalyst warning: don't try undef linktype id to access $loaded

reosarevok:
Some link types in edits only have entity0_type and entity1_type. In those cases, we shouldn't be using the nonexistent id field to access the loaded hash. It will never work + causes a warning.

Example edit: http://localhost:5000/edit/19292557

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
@@ -133,13 +133,14 @@ sub build_display_data
 
     # If no link type exists, we use the fake ID and ensure the next one doesn't clash
     my $link_type_id = $link_type->{id} // $link_type_fake_id++;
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Edit/Relationship/Delete.pm
```diff
@@ -133,13 +133,14 @@ sub build_display_data
 
     # If no link type exists, we use the fake ID and ensure the next one doesn't clash
     my $link_type_id = $link_type->{id} // $link_type_fake_id++;
+    my $loaded_link_type = $loaded->{LinkType}{$link_type->{id}} if $link_type->{id};
```
mwiencek, line 136: I had a recent PR adding spaces where the hash key isn't a simple variable or string, so we should probably continue with that.
```suggestion
    my $loaded_link_type = $loaded->{LinkType}{ $link_type->{id} } if $link_type->{id};
```

mwiencek (approved):

yvanzo (approved):

PR #2064, opened 2021-04-14
Fix catalyst warning: don't try to trim/sanitize undef "strings"

reosarevok:
This causes warnings once it gets passed to NFC. If $t is empty, I expect there's no need to do any of the sanitizing / trimming anyway, so we might as well just return it unchanged.

This is triggered at least by _clean_link_attribute with no credit or text value, but looking at logs, I think it's triggered elsewhere too.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm
+++ b/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm
@@ -393,7 +393,7 @@ sub _seeded_label
         $result->{label} = { name => trim($name) };
     }
 
-    $result->{catalogNumber} = trim($params->{catalog_number} // '');
+    $result->{catalogNumber} = trim($params->{catalog_number});
     return $result;
 }
 
@@ -542,7 +542,7 @@ sub _seeded_artist_credit_name
     my $result = {};
 
     my $name = _seeded_string($params->{name}, "$field_name.name", $errors);
-    $result->{name} = trim($name // '');
+    $result->{name} = trim($name);
 
     if (my $gid = $params->{mbid}) {
         my $entity = $c->model('Artist')->get_by_gid($gid);
@@ -556,7 +556,7 @@ sub _seeded_artist_credit_name
     }
 
     my $join = _seeded_string($params->{join_phrase}, "$field_name.join_phrase", $errors);
-    $result->{joinPhrase} = sanitize($join // '');
+    $result->{joinPhrase} = sanitize($join);
 
     $result->{artist} //= _seeded_hash($c, \&_seeded_artist, $params->{artist},
         "$field_name.artist", $errors);
--- a/lib/MusicBrainz/Server/Data/ArtistCredit.pm
+++ b/lib/MusicBrainz/Server/Data/ArtistCredit.pm
@@ -162,7 +162,7 @@ sub find_or_insert
     my ($self, $artist_credit) = @_;
 
     for my $name (@{ $artist_credit->{names} }) {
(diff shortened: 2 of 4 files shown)


lib/MusicBrainz/Server/ControllerUtils/Relationship.pm
```diff
@@ -12,8 +12,8 @@ our @EXPORT_OK = qw(
 sub _clean_link_attribute {
     my ($data) = @_;
 
-    my $credited_as = trim($data->{credited_as});
-    my $text_value = trim($data->{text_value});
+    my $credited_as = trim($data->{credited_as}) if non_empty($data->{credited_as});
+    my $text_value = trim($data->{text_value}) if non_empty($data->{text_value});
```
mwiencek, line 16: No need for these changes if you're modifying `trim`, I guess?

mwiencek:
There's a test failure in `t::MusicBrainz::Server::Controller::URL::Edit`, can you check if it's valid?

reosarevok:
It's valid, fixed now. Turns out the sanitizing was turning undef into '', so changed it to do the same while skipping the sanitizing process.

yvanzo (changes requested):
May the same error happen at some place where there is no test for it?

Some calls to `trim` do similar checks already, either before passing the argument, or after, or expect either `''` or `undef`. See for example:
* https://github.com/metabrainz/musicbrainz-server/blob/v-2021-04-05/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm#L147-L149
* https://github.com/metabrainz/musicbrainz-server/blob/v-2021-04-05/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm#L155-L157
* https://github.com/metabrainz/musicbrainz-server/blob/v-2021-04-05/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm#L396

It seems unnecessary to keep the `// ''`when passing an argument to `trim`, that is easy to remove. But checking if returned values still work seems to require more investigation.

reosarevok:
Given I changed it to always return '' which seems to be what it was doing before, I doubt anything should break? But we could drop the // '' I guess, sure.

mwiencek:
The returned values were already `''` given `undef`:

```sh
$ rlwrap re.pl
push @INC, 'lib';
use MusicBrainz::Server::Data::Utils qw( sanitize trim trim_multiline_text );
defined trim(undef) && trim(undef) eq '';
Use of uninitialized value in subroutine entry at lib/MusicBrainz/Server/Data/Utils.pm line 333, <FIN> line 3.
Use of uninitialized value in subroutine entry at lib/MusicBrainz/Server/Data/Utils.pm line 333, <FIN> line 3.
1
defined trim_multiline_text(undef) && trim_multiline_text(undef) eq '';
Use of uninitialized value in subroutine entry at lib/MusicBrainz/Server/Data/Utils.pm line 368, <FIN> line 4.
Use of uninitialized value in subroutine entry at lib/MusicBrainz/Server/Data/Utils.pm line 368, <FIN> line 4.
1
defined sanitize(undef) && sanitize(undef) eq '';
Use of uninitialized value in subroutine entry at lib/MusicBrainz/Server/Data/Utils.pm line 333, <FIN> line 5.
Use of uninitialized value in subroutine entry at lib/MusicBrainz/Server/Data/Utils.pm line 333, <FIN> line 5.
1$
```

yvanzo:
Thanks for having checked. These `// ''` should still be removed though.

yvanzo (approved):
Thanks!

mwiencek (approved):

PR #2065, opened 2021-04-14
MBS-11582: Add a sortname method for instrument guess case

reosarevok:
### Fix MBS-11582

This is needed for alias sortname guessing, which just breaks completely otherwise. Also replacing the comment with something a bit more useful.


mwiencek (approved):

yvanzo (approved):
:rofl:

PR #2067, opened 2021-04-15
MBS-10899: Report for releases with catnos that look like ISRCs

reosarevok:
### Implement MBS-10899

### DIFF
--- a/root/report/CatNoLooksLikeIsrc.js
+++ b/root/report/CatNoLooksLikeIsrc.js
@@ -1,0 +1,83 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import PaginatedResults from '../components/PaginatedResults';
+import loopParity from '../utility/loopParity';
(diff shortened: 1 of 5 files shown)


mwiencek (approved):

root/report/CatNoLooksLikeIsrc.js
```diff
@@ -0,0 +1,82 @@
...
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows releases which have catalog numbers that look
+       like ISRCs. Assigning ISRCs to releases is almost always wrong,
+       but still happens sometimes, especially for releases added to
+       MusicBrainz by an artist/label. But ISRCs are codes assigned
+       to recordings, and should be linked to the appropriate recording
+       instead. That said, do make sure this is not a legitimate catalog
+       number that just happens to look like an ISRC!`,
+    )}
```
yvanzo, line 38: I suggest to link to the documentation page to help with learning what an ISRC looks like. Untested:

```suggestion
    description={exp.l(
      `This report shows releases which have catalog numbers that look
       like ISRCs. Assigning ISRCs to releases is almost always wrong,
       but still happens sometimes, especially for releases added to
       MusicBrainz by an artist/label. But ISRCs are codes assigned
       to recordings, and should be linked to the appropriate recording
       instead. That said, do make sure this is not a legitimate catalog
       number that just happens to look like an {doc_link|ISRC}!`,
      {doc_link: '/doc/ISRC'},
    )}
```

yvanzo (approved):
Successfully tested with sample data and works for me but for one thing: There is no indication of what an ISRC looks like; Linking to the documentation can fix that.

PR #2068, opened 2021-04-15
MBS-10711: Convert most report lists to react-table

reosarevok:
### Implement MBS-10711

This converts most report lists (the ones that do not insert extra rows based on the content of the items, such as the Discogs reports) to use react-table. This already significantly limits duplication (allowing using the default lists + extra columns parameters for several reports that implemented a fully separate table before, and getting rid of separate tables for annotations and relationships for each entity).

The still unconverted reports should be possible to refactor to use react-table too (I found some info about inserting extra rows, e.g. https://pastebin.com/eeCi1DYZ), but this is huge enough as it is, so those can be done in a separate PR later on.

Thanks to @mwiencek for helping with Flow issues!


root/report/components/ArtistAnnotationList.js
```diff
@@ -22,43 +26,53 @@ type Props = {
...
-    </table>
-  </PaginatedResults>
-);
+}: Props): React.Element<typeof PaginatedResults> => {
+  const existingArtistItems = items.filter(item => item.artist != null);
+  const columns = React.useMemo(
+    () => {
+      const nameColumn = defineEntityColumn<ReportArtistAnnotationT>({
+        columnName: 'artist',
+        getEntity: result => result.artist ?? null,
```
reosarevok, line 35: Flow doesn't know I'm filtering out results with null artist, so it needs this.

root/report/components/ArtistList.js
```diff
@@ -10,51 +10,47 @@
...
+    .filter(item => item.artist != null)
+    .map(item => item.artist);
+  const columns = React.useMemo(
+    () => {
+      const nameColumn = defineNameColumn<ArtistT>({
+        title: l('Artist'),
+      });
+      const typeColumn = defineTypeColumn({
+        typeContext: 'artist_type',
+      });
```
reosarevok, line 39: Not sure whether it would be better to use the simplified approach I took here (just map the artists first) or just use the same column definitions as in the ArtistAnnotationList.

root/report/components/ArtistAnnotationList.js
mwiencek: You should be able to use `.reduce` instead of filter to work around that, though it's slightly more verbose.

root/report/components/ArtistList.js
mwiencek: If it can save lines of code and make maintenance easier, sharing the column definitions seems like a good idea. (But that may not be true until more are converted -- just looking at these two I'd say either approach is fine.)

mwiencek:
> I thought I'd just filter out any rows missing its entity, since that should mean they are effectively useless and it seems easier to do with react-table than making a colSpan 2 column saying the entity is gone.

I'm fine with filtering them out.

root/report/components/ArtistAnnotationList.js
reosarevok: I tried that but it still gives the same Flow issues.

mwiencek (approved):

PR #2069, opened 2021-04-20
MBS-11598: Do not generate image link if we don't have a suffix

reosarevok:
### Fix MBS-11598

AFAICT this happens if the file has been removed from CAA. The thumbnails still work, at the moment, but I think that's just because of a bug where deleted images won't get purged from the IA.
This also removes a concatenation-with-undef Catalyst warning.


yvanzo:
@brainzbot retest this please

mwiencek (approved):

yvanzo (approved):

PR #2072, opened 2021-04-22
MBS-11611: Do not keep/add trailing slash in OC ReMix URLs

reosarevok:
### Implement MBS-11611

They don't have trailing slashes in cases where they don't append names to the URLs (e.g. https://ocremix.org/remix/OCR00002) so it might be sensible to do the same.

This would probably cause the risk of duplicate links though, so we might want to do a MBBE cleanup if we merge this? :)

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2465,7 +2465,7 @@ const CLEANUPS = {
     type: LINK_TYPES.otherdatabases,
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?ocremix\.org\//, 'https://ocremix.org/');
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
That works for me.

I checked the format of the current OCRemix URLs in the MusicBrainz database, see the ticket.

I did not see it when reviewing #1906 but normalizing these URLs should imply adding a `validate` function too.

reosarevok:
@yvanzo: is it ok to merge this and add a second ticket for validation?

yvanzo:
Sure.

mwiencek (approved):

reosarevok:
@yvanzo: I remembered now why I didn't add validation last time - see MBS-7674. I wasn't sure what to validate some of them to. But we could at the very least add validation to some cases.

yvanzo:
@reosarevok: Oh right. Don’t bother with validating half of OCRemix URLs, that can be done after MBS-9902.

PR #2073, opened 2021-04-22
MBS-11612: Restore link to genre/delete to genre sidebar

reosarevok:
### Fix MBS-11612

MBS-10830 stopped showing removal links on the sidebar unless the entity can be deleted, but genres didn't have a can_delete method. I could just hardcode genres to always pass can_delete true or skip the can_delete check for genres, but in case we decide to add blocks to deletion later, maybe it's better to just do this like for other entities.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Genre.pm
+++ b/lib/MusicBrainz/Server/Data/Genre.pm
@@ -47,7 +47,7 @@ sub update {
     $self->sql->update_row('genre', $row, { id => $genre_id });
 }
 
-sub can_delete {1}
+sub can_delete { 1 }
 
 sub delete {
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Data/Genre.pm
```diff
@@ -47,6 +47,8 @@ sub update {
     $self->sql->update_row('genre', $row, { id => $genre_id });
 }
 
+sub can_delete {1}
```
mwiencek, line 50:
```suggestion
sub can_delete { 1 }
```

mwiencek (approved):

yvanzo (approved):

PR #2074, opened 2021-04-26
MBS-11619: Ignore periods and +tags in e-mail admin searches

reosarevok:
### Implement MBS-11619 / MBS-11620

These are often used by spammers to get "new" emails for more accounts, since many providers (such as Gmail) allow to add periods and/or +tags to the address to create aliases for the same account. We don't want to fully block this (many people use email+musicbrainz@provider to receive their MB emails, for example) but it should be easy to find people using it for nefarious purposes.

Also made the results show up to the *newest* 100 editors that match, rather than the oldest.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -195,7 +195,7 @@ sub search_by_email {
     my $query = 'SELECT ' . $self->_columns .
         ' FROM ' . $self->_table .
         q" WHERE regexp_replace(regexp_replace(email, '(\+.*@)', '@'), '\.(.*@)', '\1', 'g') ~ ?" .
-        ' ORDER BY member_since LIMIT 100';
+        ' ORDER BY member_since DESC LIMIT 100';
 
     $self->query_to_list($query, [$email]);
 }
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -186,8 +186,15 @@ sub find_by_ip {
 sub search_by_email {
     my ($self, $email) = @_;
 
+    # Remove periods (\.), and anything preceded by a + sign on the user side
+    # for email matching, since both of these are often used as "free aliases"
+    # by email providers
+    $email =~ s/\\\+.*@/@/;
+    $email =~ s/\\\.(.*@)/$1/g;
+
     my $query = 'SELECT ' . $self->_columns .
-        ' FROM ' . $self->_table . ' WHERE email ~ ?' .
+        ' FROM ' . $self->_table .
+        q" WHERE regexp_replace(regexp_replace(email, '(\+.*@)', '@'), '\.(.*@)', '\1', 'g') ~ ?" .
         ' ORDER BY member_since LIMIT 100';
 
     $self->query_to_list($query, [$email]);
--- a/root/admin/EmailSearch.js
+++ b/root/admin/EmailSearch.js
@@ -12,6 +12,7 @@ import * as React from 'react';
 import FormRowText from '../components/FormRowText';
 import FormSubmit from '../components/FormSubmit';
 import Layout from '../layout';
+import expand2react from '../static/scripts/common/i18n/expand2react';
 
 import UserList from './components/UserList';
 
@@ -40,6 +41,23 @@ const EmailSearch = ({
               link: 'https://www.postgresql.org/docs/12/' +
                 'functions-matching.html#FUNCTIONS-POSIX-REGEXP',
             },
+          )}
+        </p>
+        <p>
+          {expand2react(
+            `Since periods (<code>.</code>) and tags preceded with a
+             <code>+</code> sign on the user side of the address (that is,
+             before the <code>@</code> sign) are often used as “free aliases”
+             by email providers, both of these are ignored by this search
+             to help you find address aliases. This will only get triggered
+             if your query includes <code>@</code> (so, both
+             <code>example@example\\.com</code> and <code>example\\+mb@</code>
+             will match the address <code>exam.ple+mb@example.com</code>,
+             but simply <code>example\\+mb</code> will not).
+             <br/>
+             Don’t forget you still need to escape these symbols
+             (<code>\\.</code> and <code>\\+</code> respectively), otherwise
+             they’ll be understood as special regular expression characters!`,
           )}
         </p>
 


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -186,8 +186,13 @@ sub find_by_ip {
 sub search_by_email {
     my ($self, $email) = @_;
 
+    # Remove periods (\.), and anything preceded by a + sign for email matching,
+    # since both of these are often used as "free aliases" by email providers
+    $email =~ s/(\\.|\+[\d]+)//g;
```
yvanzo, line 191: * It should probably affect the user info only.
* `+` sign should be removed too if followed by nothing or by non-digits.

```suggestion
    $email =~ s/\+.*@/@/;
    $email =~ s/\\.(.*@)/$1/g;
```

The above comment should updated accordingly.

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -186,8 +186,13 @@ sub find_by_ip {
...
+
     my $query = 'SELECT ' . $self->_columns .
-        ' FROM ' . $self->_table . ' WHERE email ~ ?' .
+        ' FROM ' . $self->_table .
+        ' WHERE regexp_replace(email, \'(\.|\+[\d]+)\', \'\', \'g\') ~ ?' .
```
yvanzo, line 195: Same as above but in PostgreSQL:

```suggestion
        ' WHERE regexp_replace(regexp_replace(email, \'\(+.*@)\', \'@\'), \'\.(.*@)\', \'\1\', \'g\') ~ ?' .
```

yvanzo (changes requested):

lib/MusicBrainz/Server/Data/Editor.pm
mwiencek: Also, let's use a different quoting operator like `q""` or `q{}` so we don't have to escape `'` 10 times. :)

mwiencek: (Alternatively, you can use `?` and pass the regexp string as a parameter; not sure which would be clearer.)

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -186,9 +186,16 @@ sub find_by_ip {
 sub search_by_email {
     my ($self, $email) = @_;
 
+    # Remove periods (\.), and anything preceded by a + sign on the user side
+    # for email matching, since both of these are often used as "free aliases"
+    # by email providers
+    $email =~ s/\\\+.*@/@/;
+    $email =~ s/\\\.(.*@)/$1/g;
```
yvanzo, line 193: Since `$email` can be any POSIX regexp, let’s not assume it necessarily contains a `@` symbol:

```suggestion
    $email =~ s/\\\+[^@]*//;
    $email =~ s/\\\.([^@]*)/$1/g;
```

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -186,9 +186,16 @@ sub find_by_ip {
...
+
     my $query = 'SELECT ' . $self->_columns .
-        ' FROM ' . $self->_table . ' WHERE email ~ ?' .
-        ' ORDER BY member_since LIMIT 100';
+        ' FROM ' . $self->_table .
+        q" WHERE regexp_replace(regexp_replace(email, '(\+.*@)', '@'), '\.(.*@)', '\1', 'g') ~ ?" .
```
yvanzo, line 197: This is fine to assume a `@` symbol here as `email` column contains actual email addresses.

root/admin/EmailSearch.js
```diff
@@ -42,6 +42,15 @@ const EmailSearch = ({
             },
           )}
         </p>
+        <p>
+          {exp.l(
+            `Don’t forget to escape <code>\\.</code> and <code>\\+</code>!
+             And keep in mind the search ignores periods (<code>.</code>)
+             and any sections after <code>+</code> in the user side of the
+             address (before the <code>@</code>)
+             to help you find address aliases.`,
```
yvanzo, line 51: * Only one backslash (`\`) is needed to escape special characters in search input field.
* “Keep in mind” probably refers to the comment in the Perl code, or to the ticket. Maybe better explains it once instead.
* It “ignores periods (`.`)” from the `email` column, but it ignores escaped periods (`\.`) from the input.
* Same for tags following an (escaped) plus symbol.

I tried to rephrase it as follows to fix the above issues. I did neither test nor lint it.

```suggestion
            `Since periods (<code>.<code>) and tags preceded with a <code>+</code> sign
             on the user side of the address (that is before the <code>@</code> sign)
             are often used as “free aliases” by email providers, both of these are ignored
             by this search to help you find address aliases.<br/>
             Don’t forget to escape <code>\.</code> and <code>\+</code>
             still as they will be considered as special characters otherwise!`,
```

yvanzo (changes requested):
Made comments to fix main issues.

Other less urgent changes that can be made later on:
* Refactor: Rename `$email` as `$email_regexp` for clarity in `search_by_email` code.
* Improvement: Allow admin to copy/paste a raw email address without requiring them to escape special characters. It can be implemented through an additional checkbox or option.

lib/MusicBrainz/Server/Data/Editor.pm
reosarevok: I'm not sure we necessarily want to remove everything after a `+` if there's no `@`, I'd leave that as-is. Removing periods everywhere is probably kinda-sorta fine, but I'm also not sure - what happens if we search for a specific provider (that is known for spam) for example? It wouldn't match since we don't remove the periods from the provider on the PSQL, right?

root/admin/EmailSearch.js
reosarevok: > Only one backslash (\) is needed to escape special characters in search input field.

Yeah, but you need two here so that one is printed, otherwise it's interpreted as an escape character in the first place.


> “Keep in mind” probably refers to the comment in the Perl code, or to the ticket. Maybe better explains it once instead.

The idea was that we expect any admin would be aware of this in theory, but fine, I guess it doesn't hurt to change the way it's explained :)

reosarevok: Updated the text, see how it looks to you :)

lib/MusicBrainz/Server/Data/Editor.pm
yvanzo: Ok, but then we should explain that searching for an email by user info should be followed with an `@`. For example: `user.info+tag@example.org`can be found with either `user\.info\+tag@` or `userinfo@` but not `user\.info\+tag`.

root/admin/EmailSearch.js
yvanzo: Good catch for the double backslash, I thought that `\.`would be enough because that is not a valid escape here.

mwiencek (approved):

lib/MusicBrainz/Server/Data/Editor.pm
yvanzo: Explanation added through https://github.com/metabrainz/musicbrainz-server/pull/2074/commits/f6228ab9ebddeaf5ab703b81c6580b714bec447a

root/admin/EmailSearch.js
```diff
@@ -42,6 +42,18 @@ const EmailSearch = ({
             },
           )}
         </p>
+        <p>
+          {exp.l(
```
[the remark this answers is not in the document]
yvanzo: Changed in https://github.com/metabrainz/musicbrainz-server/pull/2074/commits/f6228ab9ebddeaf5ab703b81c6580b714bec447a

yvanzo (approved):
LGTM but did not retest.

PR #2075, opened 2021-04-26
CAA fixes

mwiencek:
Addresses several CAA issues reported by Ruben Opdebeeck (@ROpdebee) via email. All credit goes to him for researching these and proposing solutions.

I've tested the changes locally with the help of some additions to contrib/ssssss.psgi.

### DIFF
--- a/contrib/ssssss.psgi
+++ b/contrib/ssssss.psgi
@@ -41,6 +41,9 @@ use Plack::Request;
 use Plack::Response;
 use Log::Dispatch;
 
+use lib "$FindBin::Bin/../lib";
+use DBDefs;
+
 my $log = Log::Dispatch->new(outputs => [[ 'Screen', min_level => 'info' ]] );
 my $imgext = 'gif|jpg|pdf|png';
 my $ssssss_storage = $ENV{SSSSSS_STORAGE}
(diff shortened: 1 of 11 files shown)


yvanzo:
@mwiencek: Should tests be updated?

mwiencek:
@yvanzo Yes, I've pushed some fixes, let's see if they pass now.

reosarevok (approved):
LGTM but didn't test. First two fixes are trivial in any case. Third is a lot more complex and I'm not fully aware of how every step of the way works for CAA uploads, but seems sane and if
@ROpdebee has checked it too I'm ok with it :)

ROpdebee:
I haven't tested these changes locally either, and I don't have a lot of experience with MBS, but this looks good to me.

contrib/ssssss.psgi
```diff
@@ -80,17 +93,53 @@ sub handle_options
...
+        $response->content_type('text/xml');
+        $response->body(q{<?xml version='1.0' encoding='UTF-8'?>
+            <Error><Code>BucketAlreadyExists</Code></Error>});
+        return $response;
+    } else {
+        create_bucket($request->path_info);
+    }
 
     return $request->new_response(204);
 }
```
yvanzo, line 145: Can we log these errors using Sentry? Or do we have access to them elsewhere already?

yvanzo (approved):
Reviewed closely but could not test; Hopefully there are mechanized tests.

Thanks again to @ROpdebee for deeply investigating this issue!

contrib/ssssss.psgi
mwiencek: We'd need to add some methods to MusicBrainz::Errors to allow logging arbitrary messages to Sentry. I don't have the means to test that right now, but I could look into it after the schema change.

(But we wouldn't want to log all `BucketAlreadyExists` errors; only cases where the ownership is wrong. And that should rarely ever happen, hopefully, other than the one case we know about.)

ROpdebee:
MBS-11624

PR #2081, opened 2021-05-01
MBS-10211: Formatting Tweet links for recordings

y-young:
# Problem

MBS-10211: Formatting Tweet links for recordings

# Solution

Format tweet link to a general form

e.g.:

https://twitter.com/mountain_goats/status/1062342708470132738 -> https://twitter.com/i/web/status/1062342708470132738

# Checklist for author

* [x] Delete irrelevant parts of this template

# Action

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3244,7 +3244,7 @@ const CLEANUPS = {
       return url;
     },
     validate: function (url, id) {
-      const m = /^https:\/\/twitter\.com\/(?:[^\/?#]+|(i\/web\/status\/\d+)?)$$/.exec(url);
+      const m = /^https:\/\/twitter\.com\/(?:[^\/?#]+|(i\/web\/status\/\d+))$/.exec(url);
       if (m) {
         const isATweet = !!m[1];
         if (Object.values(LINK_TYPES.streamingfree).includes(id)) {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3230,13 +3230,21 @@ const CLEANUPS = {
...
+            return website + username;
+          }
+          return website;
+        },
       );
       return url;
     },
     validate: function (url, id) {
-      const m = /^https:\/\/twitter\.com\/[^\/?#]+(\/status\/\d+)?$/.exec(url);
+      const m = /^https:\/\/twitter\.com\/(?:[^\/?#]+|(i\/web\/status\/\d+)?)$$/.exec(url);
```
reosarevok, line 3247: Is the double `$$` intentional?
Also, if I'm reading this correctly, then the previous version rejected https://twitter.com/ (the main Twitter page) while this accepts it. Is the `?` after `(i\/web\/status\/\d+)` needed? IMO it makes sense to block linking to just Twitter without a profile or tweet.

y-young: Sorry, the double `$` was a mistake.
And yes you're right, but I found that though `https://twitter.com/` was blocked, `https://twitter.com` wasn't because it is not matched, should we do something about it?

reosarevok: Not sure what @yvanzo thinks but I see no reason not to also block `https://twitter.com` by making the end slash optional on the original matching regex or removing it entirely.

yvanzo:
Thanks @y-young for looking into this ticket and implementing it, the code changes look good overall.
However, we decided last month with @reosarevok to not follow this suggestion, see ticket’s comments for details.
Sorry that your good work doesn’t get merged, I hope that it allowed you to get more familiar with the code at least.

PR #2085, opened 2021-05-04
MBS-11630: Do not break on empty edits array

reosarevok:
### Fix MBS-11630

There is at least one case where the edits.releaseGroup array is empty: when the release group has been renamed elsewhere after the editor opened this release editor session to the same name the editor is trying to apply here. As such, we try to access edit_type of an undefined edit. This just uses optional chaining to avoid that attempted call entirely.

### DIFF
--- a/root/static/scripts/release-editor/edits.js
+++ b/root/static/scripts/release-editor/edits.js
@@ -608,7 +608,11 @@ releaseEditor.orderedEditSubmissions = [
     callback: function (release, edits) {
       var edit = edits[0];
 
-      if (edit.edit_type == MB.edit.TYPES.EDIT_RELEASEGROUP_CREATE) {
+      /*
+       * edit can be undef if the only change is RG rename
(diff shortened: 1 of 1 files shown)


root/static/scripts/release-editor/edits.js
```diff
@@ -608,7 +608,7 @@ releaseEditor.orderedEditSubmissions = [
     callback: function (release, edits) {
       var edit = edits[0];
 
-      if (edit.edit_type == MB.edit.TYPES.EDIT_RELEASEGROUP_CREATE) {
+      if (edit?.edit_type == MB.edit.TYPES.EDIT_RELEASEGROUP_CREATE) {
```
mwiencek, line 611: This looks fine, but can we add a short comment specifying when `edit` is undefined so we don't forget?

mwiencek (approved):

yvanzo (approved):

PR #2087, opened 2021-05-04
MBS-11614: Allow merging releases when a medium is empty

reosarevok:
### Implement MBS-11614

If a release has two mediums, each with 1 track, and another has two mediums, one empty and one with 1 track, we should allow merging the empty one into the non-empty one. After all, that data doesn't mean the medium actually has 0 tracks, just that we don't have enough info about the tracks on that medium.
Merging the other way around (into the release with the 0-track medium) should still fail, but with a more explanatory error message.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Release.pm
+++ b/lib/MusicBrainz/Server/Data/Release.pm
@@ -55,6 +55,7 @@ Readonly::Hash our %RELEASE_MERGE_ERRORS => (
     medium_missing              => N_l('Some mediums being merged don’t have an equivalent on the target release: either the target release has less mediums, or the positions don’t match.'),
     medium_positions            => N_l('The medium positions conflict.'),
     medium_track_counts         => N_l('The track counts on at least one set of corresponding mediums do not match.'),
+    merging_into_empty          => N_l('Merging a medium with tracks into one without them is not currently supported. You can always merge in the other direction!'),
     pregaps                     => N_l('Mediums with a pregap track can only be merged with other mediums with a pregap track.'),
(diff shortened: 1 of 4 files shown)


mwiencek:
> Merging the other way around (into the release with the 0-track medium) should still fail, but with a more explanatory error message.

The error doesn't really explain it to me. :) If the releases are the same, why shouldn't either direction be valid and have the same outcome for the empty medium?

mwiencek:
I was thinking of it more like filling in a null value with a non-null one, rather than overwriting. (Like I wouldn't expect it to overwrite the medium format or name of the empty one.)

The two tracklisty -> one empty-tracklist merge example is a good point, since the final tracklist to use is ambiguous in that case. Agreed that an error makes sense there.

Though the case where it's just two releases being merged seems to have clear enough intent and is unambiguous to me. Maybe it's still inconsistent somehow? The error message just isn't very satisfying. :)

reosarevok:
I guess the main benefit of this way is that it just works as is, while the alternative would require a lot of extra changes ;p

reosarevok:
Changed the message a bit, added a test.

mwiencek (approved):

lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -898,9 +899,23 @@ sub can_merge {
             });
         }
 
+        my $merging_into_empty_medium = $self->sql->select_single_value(
+            "$mediums_query
+             WHERE s.track_count > 0
+             AND new_medium.track_count = 0
+             LIMIT 1",
```
mwiencek, line 906: Might be a good opportunity for `<<~EOSQL` :)

PR #2088, opened 2021-05-04
MBS-11609: Show genre disambiguation on autocomplete dropdown

reosarevok:
### Implement MBS-11609

Rather than just showing the value as the label, this defines specific labels for the genres - just the name in most cases, but when there's a disambiguation comment we show it too. It won't be actually added to the tag list since it's not part of the value.

jQuery Autocomplete doesn't really support HTML labels by default, took the code for that from https://stackoverflow.com/questions/3488016/using-html-in-jquery-ui-autocomplete which seems to just take it from an extension last updated in 2013. Given that fact, we might as well just skip the extension.

Screenshot from 2021-05-04 17-21-35

### DIFF
--- a/root/static/scripts/common/components/TagEditor.js
+++ b/root/static/scripts/common/components/TagEditor.js
@@ -7,13 +7,14 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import he from 'he';
 import * as React from 'react';
 
 import hydrate, {minimalEntity} from '../../../../utility/hydrate';
 import loopParity from '../../../../utility/loopParity';
 import {unwrapNl} from '../i18n';
 import {keyBy, sortByNumber} from '../utility/arrays';
-import bracketed from '../utility/bracketed';
+import bracketed, {bracketedText} from '../utility/bracketed';
 import {compareStrings} from '../utility/compare';
 import debounce from '../utility/debounce';
 import isBlank from '../utility/isBlank';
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/components/TagEditor.js
```diff
@@ -451,15 +467,20 @@ class TagEditor extends React.Component<TagEditorProps, TagEditorState> {
...
-            x => x.startsWith(last) ? 0 : 1,
+            x => x.value.startsWith(last) ? 0 : 1,
           );
 
         response(filteredTerms);
       },
-    });
+    }).data('ui-autocomplete')._renderItem = function (ul, item) {
+      return $('<li></li>')
+        .data('item.autocomplete', item)
```
mwiencek, line 480: It appears to work without this line, and I checked the jQuery UI source and couldn't find any use of it. Odd.
```suggestion
```

mwiencek (approved):

root/static/scripts/common/components/TagEditor.js
reosarevok: Huh. I guess it might be some sort of line that has gone through stackoverflow comments for years without anyone checking whether it was needed. It does work without, made the change.

root/static/scripts/common/components/TagEditor.js
```diff
@@ -37,6 +37,15 @@ const cmpTags = (a, b) => (
   (b.count - a.count) || compareStrings(a.tag.name, b.tag.name)
 );
 
+function formatGenreLabel(genre) {
+  if (!genre.comment) {
+    return genre.name;
+  }
+
+  return genre.name + '<span class="small"> ' +
+         bracketedText(genre.comment) + '</span>';
```
mwiencek, line 46: (you'll have to import the he module here)
```suggestion
  return he.encode(genre.name) +
    '<span class="small"> ' +
    he.encode(bracketedText(genre.comment)) +
    '</span>';
```

mwiencek (changes requested):
sorry one thing we should fix

root/static/scripts/common/components/TagEditor.js
mwiencek: Never mind, made the change!

mwiencek (approved):

PR #2090, opened 2021-05-05
MBS-11638 / MBS-11639 / MBS-11640: Fixes for the work attribute form

reosarevok:
### Fix MBS-11638 / MBS-11639 / MBS-11640

Tested manually by creating the relevant situations on local data.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Work.pm
+++ b/lib/MusicBrainz/Server/Form/Work.pm
@@ -126,6 +126,7 @@ after 'validate' => sub {
         values %$attribute_types
     );
 
+    my %used_attributes;
     for my $attribute_field ($attributes->fields) {
         next if is_empty_attribute($attribute_field);
         next if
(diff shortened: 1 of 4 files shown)


root/static/scripts/work.js
```diff
@@ -121,36 +121,46 @@ class WorkAttribute {
...
-      if (!typeID || this.allowsFreeText()) {
+      if (this.allowsFreeText(typeID)) {
         return [];
       }
       return this.parent.allowedValuesByTypeID[typeID] ?? [];
     });
 
+    this.typeID.subscribe((previousTypeID => {
+      this.previousTypeID = previousTypeID;
+    }), this, 'beforeChange');
```
reosarevok, line 140: This causes an issue with our flow typing for subscribe - the typing needs to be updated to support these extra arguments, AFAICT. Not quite sure how to do so.

mwiencek (changes requested):
Works well. I was thinking it'd make sense to prevent submission and show an error with JS, but that's kind of annoying with Knockout and this probably isn't a super common error, so it could wait.

lib/MusicBrainz/Server/Form/Work.pm
```diff
@@ -145,11 +146,23 @@ after 'validate' => sub {
             next;
         }
 
+        my @used_attribute_values = @{ $used_attributes{$attribute_type} // [] };
```
mwiencek, line 149: Try `//= []`, then you can remove the line `$used_attributes{$attribute_type} = \@used_attribute_values;` below.

mwiencek (approved):

yvanzo (commented):
Commit about MBS-11638 works for me. I did not find how to reproduce MBS-11639. There is no mention of MBS-11640 in commits. (Reran CI test to make it pass.)

reosarevok:
@yvanzo: for MBS-11639 you can just pick "Key", select whatever value, then change the selection to one of the free text attributes (any of the identifiers). You should see the value ID for the previously selected key as the free text value (at least on Chrome). I seemingly squashed MBS-11640 into the MBS-11639 commit by mistake, will split.

PR #2094, opened 2021-05-07
MBS-960: Allow mass-deleting and mass-changing user tags

reosarevok:
### Implement MBS-960

Tested manually with several different combinations of situations (changing tags into some that do not exist and some that do, deleting upvotes, deleting downvotes).


lib/MusicBrainz/Server/Data/Tag.pm
```diff
@@ -96,6 +96,55 @@ sub get_cloud
     return $data;
 }
 
+sub rename_for_user {
```
mwiencek, line 99: The problem with this is that it doesn't update the aggregate tag tables, only the raw ones. So the new tags will have broken counts and not display in the editing UI at all in some cases.

Ideally we'd handle this via triggers, but that's not the current situation unfortunately. This should really only be using `Data::EntityTag` methods to perform updates (which is accessible via `->tags` on each entity model), but those only operate on one entity and one tag at present. They'd have to be modified to operate on a list of entities.

root/tag/DeleteTag.js
```diff
@@ -0,0 +1,80 @@
...
+        {' '}
+        {bracketed(
+          <a
+            href={'/user/' + encodeURIComponent(user.name) +
+                  '/tag/' + encodeURIComponent(tag.name) +
+                  '?show_downvoted=' + (deleteDownvoted ? '1' : '0')}
+          >
+            {l('see list')}
+          </a>,
+        )}
```
mwiencek, line 62: maybe make "all your votes against the tag foo" the link text pointing to the list? probably no need to link the tag itself there since it's already in the heading

root/tag/MoveTag.js
```diff
@@ -0,0 +1,81 @@
...
+          </a>,
+        )}
+      </p>
+      <p>
+        {l(`If you want to replace the tag with several others,
+            enter them separated by commas.`)}
+      </p>
+      <form action={$c.req.uri} method="post">
+        <div className="half-width">
+          <FormRowTextLong
```
mwiencek, line 64: This displays weirdly. :) The input is really long and on a separate line from the label, which is indented almost to the middle of the input.

root/tag/DeleteTag.js
```diff
@@ -0,0 +1,80 @@
...
+    >
+      <h2>{title}</h2>
+      <p>
+        {exp.l(
+          deleteDownvoted ? (
+            `Are you sure you want to remove
+             all your votes against the tag “{tag}”?`
+          ) : (
+            `Are you sure you want to remove
+             all your uses of the tag “{tag}”?`
```
mwiencek, line 49: I think this message confused me, because I was looking for a way to delete only upvotes, and thought this would delete all votes.  But I don't know if replacing "uses" with "upvotes" would be more confusing or less.

mwiencek (changes requested):

root/tag/DeleteTag.js
reosarevok: "all your votes for the tag" has the same issue, since it's not clear it's for vs against. "all your votes in favor of the tag"? Too weird?

root/tag/MoveTag.js
reosarevok: It displayed on one line to me, but it might be better now?

lib/MusicBrainz/Server/Data/Tag.pm
reosarevok: So basically use upvote/downvote/withdraw?

mwiencek: Yeah, except for the last bit I mentioned there ("those only operate on one entity and one tag at present. They'd have to be modified to operate on a list of entities").

root/tag/DeleteTag.js
mwiencek: We do use "upvotes" in the vote filters on user tag pages now, so maybe that's not so bad.  "uses (upvotes)" might work too.

reosarevok:
Converting to draft for now: the best way to do this is probably to first make sure we have triggers to keep raw and grouped tag tables in sync, rather than doing it all via code. See also stuff like MBS-5359 about why this is a mess.

reosarevok:
The triggers are now present since the 2022 Q2 schema change, so this should work fine now.

reosarevok:
I wrote some tests for this too, but do check in case you feel it needs more test cases :)

root/user/UserTagList.js
```diff
@@ -61,84 +97,113 @@ const UserTagList = ({
...
+        sortBy={sortBy}
+      />
+
+      <div id="all-tags">
+        {(genres.length > 0 || tags.length > 0) ? (
+          <>
+            <h3>{l('Genres')}</h3>
+
+            <div id="genres">
+              {genres.length > 0 ? (
```
mwiencek, line 124: I split the tag "pop rap" into "pop" and "rap" and for some reason pop is now displayed twice under genres.  Is that a new issue?

root/tag/DeleteTag.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const DeleteTag = ({
+  $c,
+  deleteDownvoted = false,
+  tag,
+}: Props): React.Element<typeof TagLayout> | null => {
+  const title = deleteDownvoted
+    ? l('Delete tag downvotes')
+    : l('Delete tag uses');
```
mwiencek, line 29: elsewhere we use "delete my upvotes," or "uses (upvotes)".  we should probably use upvotes consistently to avoid confusion - what do you think?

root/tag/MoveTag.js
```diff
@@ -0,0 +1,71 @@
...
+  tag,
+}: Props): React.Element<typeof TagLayout> | null => {
+  const user = $c.user;
+
+  return user ? (
+    <TagLayout page="delete" tag={tag} title={l('Change tag uses')}>
+      <h2>{l('Change tag uses')}</h2>
+      <p>
+        {exp.l(
+          `This will change {tag_list|all your votes for tag “{tag}”}
```
mwiencek, line 36: here we use "votes for," neither "uses" nor "upvotes" - seems confusing

root/user/UserTagList.js
reosarevok: I cannot reproduce that myself, weird. Can you?

root/tag/DeleteTag.js
reosarevok: True. I went with uses (upvotes) everywhere for now.

lib/MusicBrainz/Server/Data/Tag.pm
```diff
@@ -120,6 +120,55 @@ sub get_cloud
...
+    }
+
+    for my $entity_type (entities_with('tags')) {
+        my $entity_ids = $self->sql->select_single_column_array(<<~"SQL", $editor_id, $tag_id);
+            DELETE FROM ${entity_type}_tag_raw
+                  WHERE editor = ? AND tag = ? AND is_upvote = TRUE
+              RETURNING $entity_type
+            SQL
+
+        for my $new_tag_id (@new_tag_ids) {
```
mwiencek, line 148: Paranoid question: are we worried that allowing to split 1 tag into N tags all linked to the same entities makes for a very efficient way to amplify data and/or spam the database? Is the example given ("australian hip-hop" -> "australian", "hip hop") more than just a very niche use case? Perhaps we should restrict the splitting to max N = 2, but even that seems iffy to me.

reosarevok: There's a fair amount of tags around that could be split into genre tags (or even better, automatically "split" into "australian hip hop, hip hop" :) ). There's some risk of spam, but people can already submit tags using the API in large amounts, and we can already ban them if they seem to be abusing the system, so I'm not too worried? Any bad actors would still have to have added at least one tag to all the entities they want to spam anyway, and by that time, why wouldn't they just add all the tags there?

lib/MusicBrainz/Server/Data/Tag.pm
```diff
@@ -120,6 +120,55 @@ sub get_cloud
...
+    for my $new_tag (@$new_tags) {
+        my $tag_entity = $self->get_by_name($new_tag);
+        if ($tag_entity) {
+            push @new_tag_ids, $tag_entity->id;
+        } else {
+            my $new_tag_id = $self->sql->select_single_value(<<~'SQL', $new_tag);
+                INSERT INTO tag (name)
+                     VALUES (?)
+                  RETURNING id
+                SQL
```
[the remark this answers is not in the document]
reosarevok: Can the whole risk be avoided by running both inside one `with_transaction` block?

root/user/UserTagList.js
mwiencek: Can't reproduce.

lib/MusicBrainz/Server/Data/Tag.pm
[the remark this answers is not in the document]
mwiencek: It's already run from inside a transaction (in `Controller::tag::move`), but it doesn't avoid the risk. We'd have to use heavier locks.

mwiencek (approved):
I'm not sure if the UI or strings are ideal, but I've reached my limit on what I can suggest there. :) Otherwise this seems to work fine. (I skimmed the code again, too, but didn't do a thorough re-review.)

lib/MusicBrainz/Server/Data/Tag.pm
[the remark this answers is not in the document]
reosarevok: Is it worth it looking into those heavier locks?

root/tag/DeleteTag.js
yvanzo: Would it be alright to just use _tag upvotes_ and avoid introducing a new unneeded verbiage with _tag uses_?

_upvote_ is used everywhere already, [documentation](https://musicbrainz.org/search?type=doc&query=upvote#gsc.tab=0&gsc.q=upvote&gsc.page=1) included. I didn’t find any occurrence of _use_ for tags so far.

lib/MusicBrainz/Server/Data/Tag.pm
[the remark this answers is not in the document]
yvanzo: @mwiencek: Do you refer to the `get_by_name` call 4 lines above? What would be the scenario?

lib/MusicBrainz/Server/Data/Tag.pm
yvanzo: IIUC this API endpoint doesn’t have any limit on the number of entities and tags that can be submitted at once. Or is there a size limit on POST requests received by MB Server?

Does it pose any question about service availability beyond the potential spam issue?

This patch would reproduce the current webservice issues into the website, making it even wider.

root/tag/DeleteTag.js
```diff
@@ -0,0 +1,71 @@
...
+          {
+            tag: tag.name,
+            tag_list: '/user/' + encodeURIComponent(user.name) +
+                      '/tag/' + encodeURIComponent(tag.name) +
+                      '?show_downvoted=' + (deleteDownvoted ? '1' : '0'),
+          },
+        )}
+      </p>
+      <p>
+        {l('This can take a long time if you delete a large amount of tags.')}
```
yvanzo, line 53: Did you test the time it is taking or is it a guess?
Is there any way to know the amount of tags (actually tag votes) to be deleted?

reosarevok: IIRC we did check the time and it's fairly slow (in the same way the site sometimes times out when deleting an account with many thousands of tags). But it was ages ago and I don't have the exact numbers anywhere.

reosarevok: We have a way to know the amount of tags a person has (we show them on their profile page with other stats) so there's no way in theory why we couldn't do the same with a filter for the specific tag :)

root/tag/DeleteTag.js
reosarevok: Ok, I'll make the change. It feels a bit confusing in that as a user I'd expect "upvote" to be when I click to agree with an existing tag, but not when I enter a tag by writing it, but it's entirely true that's a much more general issue than here so we should be consistent for now.

reosarevok: I changed these. I also added `folksonomy` contexts where relevant and changed some "votes against" to "downvotes of" for parity.

PR #2095, opened 2021-05-07
MBS-11644: Allow specifying a release MBID when moving a CD TOC

reosarevok:
### Implement MBS-11644

MBS-7473 / https://github.com/metabrainz/musicbrainz-server/pull/1387, but for moving rather than attaching a new CD TOC.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/CDTOC.pm
+++ b/lib/MusicBrainz/Server/Controller/CDTOC.pm
@@ -413,10 +413,32 @@ sub move : Local Edit
         $c->stash( template => 'cdtoc/move_search.tt' );
 
         if ($c->form_submitted_and_valid($search_release, $c->req->query_params)) {
+            my $query = $search_release->field('query')->value;
+            my ($mbid) = $query =~ m/(
+                [\da-f]{8} -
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/CDTOC.pm
```diff
@@ -413,10 +413,32 @@ sub move : Local Edit
...
+                [\da-f]{4} -
+                [\da-f]{4} -
+                [\da-f]{4} -
+                [\da-f]{12}
+            )/ax;
+
             my $releases = $self->_load_paged($c, sub {
-                $c->model('Search')->search('release', $search_release->field('query')->value, shift, shift,
+                if (defined $mbid) {
+                    $c->stash->{was_mbid_search} = 1;
```
mwiencek, line 427: We should handle `was_mbid_search` the same way we for the attach pages, I guess?

reosarevok: This is sent to an attach page, so I think that it requires no further changes. Correct me if wrong :)

mwiencek (approved):

PR #2096, opened 2021-05-10
MBS-11650: Add tag and rating statistics to profile page

reosarevok:
### Implement MBS-11650

It seems interesting to see a sum of tags (up- and downvoted) and ratings for each user, if they are public (or the user's own). Ideally this could also show a count of CB reviews (MBS-11653).

Screenshot from 2021-05-10 14-44-21

Tested with `pink` data, logged in vs out, modifying preferences to make sure that works as it should. I can't guarantee this doesn't cause a slowdown though.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -714,6 +714,69 @@ sub added_entities_counts {
     return \%result;
 }
 
+sub secondary_counts {
+    my ($self, $editor_id, $viewing_own_profile) = @_;
+
+    my $editor = $self->get_by_id($editor_id);
+    $self->load_preferences($editor);
+
+    my %result;
+
+    if ($viewing_own_profile || $editor->preferences->public_tags) {
+        $result{upvoted_tag_count} = 0;
+        $result{downvoted_tag_count} = 0;
+
+        my @tag_tables = entities_with(
+            'tags',
+            take => sub { shift . '_tag_raw' },
+        );
+        my $tag_inner_query = join(
+            ' UNION ALL ',
+            map { "SELECT is_upvote FROM $_ WHERE editor = ?" } @tag_tables
+        );
+
+        my $query = <<~SQL;
+            SELECT x.is_upvote, count(*)
+            FROM ($tag_inner_query) x
+            GROUP BY x.is_upvote
+            SQL
+
+        my $rows = $self->sql->select_list_of_lists(
+            $query,
+            ($editor_id) x scalar @tag_tables,
+        );
+
+        for my $row (@$rows) {
+            my ($is_upvote, $count) = @$row;
+            if ($is_upvote) {
+                $result{upvoted_tag_count} = $count + 0;
+            } else {
+                $result{downvoted_tag_count} = $count + 0;
+            }
+        }
+  }
+
+    if ($viewing_own_profile || $editor->preferences->public_ratings) {
+        my @rating_tables = entities_with(
+            'ratings',
+            take => sub { shift . '_rating_raw' },
+        );
+        my $rating_inner_query = join(
+            ' UNION ALL ',
+            map { "SELECT 1 FROM $_ WHERE editor = ?" } @rating_tables
+        );
+
+        my $query = "SELECT count(*) FROM ($rating_inner_query) x";
+
+        $result{rating_count} = $self->sql->select_single_value(
+            $query,
+            ($editor_id) x scalar @rating_tables,
+        );
+    }
+
+    return \%result;
+}
+
 sub last_24h_edit_count
 {
     my ($self, $editor_id) = @_;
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -701,6 +701,47 @@ sub added_entities_counts {
...
+
+            for my $row (@$rows) {
+                my ($is_upvote, $count) = @$row;
+                if ($is_upvote) {
+                    $upvote_count += $count;
+                } else {
+                    $downvote_count += $count;
+                }
+            }
+        }
```
mwiencek, line 727: I think we should be able to reduce the number of round-trips to the server by doing this in one query. Maybe something like:

```SQL
SELECT x.is_upvote, count(*) FROM (
  SELECT is_upvote FROM work_tag_raw WHERE editor = ?
  UNION ALL
  SELECT is_upvote FROM recording_tag_raw where editor = ?
  UNION ALL
  ...
) x
GROUP BY x.is_upvote;
```

etc. though you'd have to build the inner part of the query of course.

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -701,6 +701,47 @@ sub added_entities_counts {
...
+        $result{'upvoted_tag_count'} = $upvote_count + 0;
+        $result{'downvoted_tag_count'} = $downvote_count + 0;
+    }
+
+    if ($viewing_own_profile || $editor->preferences->public_ratings) {
+        my $count = 0;
+        for my $type (entities_with('ratings')) {
+            my $query = "SELECT COUNT(*) FROM ${type}_rating_raw WHERE editor = ?";
+            $count += $self->c->sql->select_single_value($query, $editor_id);
+        }
```
mwiencek, line 738: (Similar comment as above.)

mwiencek (commented):
Do you think we need the "Tags" and "Ratings" tabs on the profile anymore since these basically duplicate those? (We sure have a lot of tabs there.)

Unrelatedly: If we add yet another stats box for the thanks/kudos then we may need to start organizing these vertically rather than horizontally. :)

reosarevok:
Not sure about the tabs - it might be a bit weird to have these pages linked only from the stats (and not linked at all if empty) but on the other hand it would save some space I guess.

We don't have that many tabs when viewing someone else's page:
Screenshot from 2021-05-20 11-32-22
Although from those I could see us merging subscriptions and subscribers, since they are a close enough concept.

Viewing your own profile, there's a lot more of them:
Screenshot from 2021-05-20 11-33-29
There "Change Password" and "Delete Account" could both a part of "Edit Profile" (renamed to "Edit Account" maybe?) IMO.

lib/MusicBrainz/Server/Data/Editor.pm
reosarevok: My only worry about this was to hardcode the tables, but maybe that's not a problem.

reosarevok: Oh, actually, I guess I can do something similar to https://github.com/metabrainz/musicbrainz-server/blob/49f0a33667130378141f081ad4511284aa95358c/lib/MusicBrainz/Server/Data/Editor.pm#L477

reosarevok: That seemed to work :)

root/user/UserProfile.js
```diff
@@ -709,6 +722,58 @@ const UserProfileStatistics = ({
...
+              <th colSpan="2">
+                {l('Tags and ratings')}
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            {hasPublicTags ? (
+              <>
+                <UserProfileProperty name={l('Tags upvoted')}>
+                  {$c.user && upvotedTagCount > 0 ? exp.l(
```
mwiencek, line 743: If I have no upvoted tags and no ratings, it shows a "Ratings: 0" row but nothing for tags (which I guess only seems odd since the heading is "Tags and ratings.") Maybe we should always show the "Tags upvoted" row even if it's 0?

mwiencek: Actually, this is probably because the materialized columns in my sample data say I have ratings. I see it's not supposed to show anything at all!

reosarevok: Are your tags public? My idea was that it would show 0 if they are public but you have none, and not show at all if they are set to private. But maybe my check doesn't work as well as I thought (I only tested it with ratings). Will look.

reosarevok: Yeah, on review, we do need to set a default of 0 for the tags if tags are public (or you're viewing your profile) in Perl, they otherwise remain undef if there's 0.

mwiencek: Did you push the change? I don't see anything in the PR timeline.

reosarevok: I did now, sorry :)

mwiencek:
@brainzbot test this please.

PR #2102, opened 2021-05-14
MBS-11660: Fix ignoring several periods in admin search by email

yvanzo:
# Fix MBS-11660: Periods are not always ignored in admin e-mail searches

Editor search by email (for admin) was failing at removing more than one period `.` from email search and addresses despite the `g` flag because it relied on capture group `()` that would be processed only once. That is why only one period `.` was removed from user info part.

This patch fixes it by splitting email over `@` symbol, for processing user info (removing `+suffix` and periods `.` used for aliases) and host name (not removing anything as these are not for aliases) separately, then contatenating both parts again using `||`.

It also requires admin to not include `+suffix` and periods `.` in user info parts of its regular expression search. This is because it is not possible to programmatically distinguish user info parts in complex regular expression that may include several user info parts such as `^(huey@meb\.o|duey@.+\.meb\.o|.+\.louie\+green\.cap@meb\.o)$` (or it would require a regular expression parser).
    
Add continuous integration tests for the `search_by_email` subroutine.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -186,15 +186,9 @@ sub find_by_ip {
 sub search_by_email {
     my ($self, $email) = @_;
 
-    # Remove periods (\.), and anything preceded by a + sign on the user side
-    # for email matching, since both of these are often used as "free aliases"
-    # by email providers
-    $email =~ s/\\\+.*@/@/;
-    $email =~ s/\\\.(.*@)/$1/g;
-
     my $query = 'SELECT ' . $self->_columns .
         ' FROM ' . $self->_table .
-        q" WHERE regexp_replace(regexp_replace(email, '(\+.*@)', '@'), '\.(.*@)', '\1', 'g') ~ ?" .
+        q" WHERE (regexp_replace(regexp_replace(email, '[@+].*', ''), '\.', '', 'g') || regexp_replace(email, '.*@', '@')) ~ ?" .
         ' ORDER BY member_since DESC LIMIT 100';
 
     $self->query_to_list($query, [$email]);
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -184,14 +184,14 @@ sub find_by_ip {
 }
 
 sub search_by_email {
-    my ($self, $email) = @_;
+    my ($self, $email_regexp) = @_;
 
     my $query = 'SELECT ' . $self->_columns .
         ' FROM ' . $self->_table .
         q" WHERE (regexp_replace(regexp_replace(email, '[@+].*', ''), '\.', '', 'g') || regexp_replace(email, '.*@', '@')) ~ ?" .
         ' ORDER BY member_since DESC LIMIT 100';
 
-    $self->query_to_list($query, [$email]);
+    $self->query_to_list($query, [$email_regexp]);
 }
 
 sub find_by_area {
--- a/root/admin/EmailSearch.js
+++ b/root/admin/EmailSearch.js
@@ -48,16 +48,19 @@ const EmailSearch = ({
             `Since periods (<code>.</code>) and tags preceded with a
              <code>+</code> sign on the user side of the address (that is,
              before the <code>@</code> sign) are often used as “free aliases”
-             by email providers, both of these are ignored by this search
-             to help you find address aliases. This will only get triggered
-             if your query includes <code>@</code> (so, both
-             <code>example@example\\.com</code> and <code>example\\+mb@</code>
-             will match the address <code>exam.ple+mb@example.com</code>,
-             but simply <code>example\\+mb</code> will not).
-             <br/>
-             Don’t forget you still need to escape these symbols
-             (<code>\\.</code> and <code>\\+</code> respectively), otherwise
-             they’ll be understood as special regular expression characters!`,
+             by email providers, both of these are ignored from the registered
+             email addresses by this search to help you find address aliases.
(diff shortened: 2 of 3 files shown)


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -194,7 +194,7 @@ sub search_by_email {
 
     my $query = 'SELECT ' . $self->_columns .
         ' FROM ' . $self->_table .
-        q" WHERE regexp_replace(regexp_replace(email, '(\+.*@)', '@'), '\.(.*@)', '\1', 'g') ~ ?" .
+        q"WHERE (regexp_replace(regexp_replace(email, '[@+].*', ''), '\.', '', 'g') || regexp_replace(email, '.*@', '@')) ~ ?" .
```
reosarevok, line 197: I'm not sure what each bit here is doing (not sure if it is complex or I am slow). Mind explaining step by step? :)

yvanzo: This is explained in commit message (with a few typos):
It splits `email` value on `@` character to process separately:
* user info (to remove `+suffix` and periods `.`),
* domain name (no change).

Is it more clear like that?

reosarevok: OH. Duh. I was thinking of || as OR, not concatenation. Sorry.

yvanzo: Step-by-step example with `email = 'a.b.c+d@f.g'`:
1. `regexp_replace('a.b.c+d@f.g', '[@+].*', '')` ⇒ `'a.b.c'`
2. `regexp_replace('a.b.c', '\.', '', 'g')` ⇒ `'a'`
3. `regexp_replace('a.b.c+d@f.g', '.*@', '@')` ⇒ `'@f.g'`
4. `'a' || '@f.g'` ⇒ `'a@f.g'`

reosarevok (approved):

lib/MusicBrainz/Server/Data/Editor.pm
reosarevok: Thanks! Seems sensible.

yvanzo: Updated commit message to fix typo (s/prefix/suffix/) and mention contatenation operator `||`: https://github.com/metabrainz/musicbrainz-server/pull/2102/commits/06d4763535f34db8287f1a82dd7aff9fb03caf11.

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -194,7 +194,7 @@ sub search_by_email {
 
     my $query = 'SELECT ' . $self->_columns .
         ' FROM ' . $self->_table .
-        q" WHERE regexp_replace(regexp_replace(email, '(\+.*@)', '@'), '\.(.*@)', '\1', 'g') ~ ?" .
```
mwiencek, line 197: Just to make sure, does it run? I see you removed the leading space, but `$self->_table` is just `'editor'`, so I'd think that'd result in broken syntax (`editorWHERE`).

reosarevok (changes requested):
Yeah, this is actually broken, please add the space back :) 

Edit: @yvanzo, mind adding a test for this use case (with several periods and a + section) to t/lib/t/MusicBrainz/Server/Data/Editor.pm? :) I should have added it myself when I first added this, sorry - but by now it's probably easier if you do it?

lib/MusicBrainz/Server/Data/Editor.pm
yvanzo: Good catch. Fixed it and added tests in <https://github.com/metabrainz/musicbrainz-server/pull/2102/commits/b19fc87e0475d2af29e79e0ebfffaf58ef6eb135>.

I got distracted by the uneven indentation of SQL keywords. Maybe we could use [indented here-documents](https://perldoc.perl.org/perlop#Indented-Here-docs) `<<~EOSQL` to embed SQL code? That would just require a couple of extra variables to replace `->` attribute/function calls.

yvanzo:
Fixed and found that the issue in PostgreSQL is the same in Perl: Capture group is processed only once despite `g` flag.

So I removed the code processing the email regular expression (since we don't want to build a regular expression parser) and put the admin in charge of removing `+suffix` and periods `.` from the regular expression input. Only actual email addresses are processed. Also added tests for the `search_by_email` subroutine, and renamed its parameter to `email_regexp`.

reosarevok (approved):
Seems sensible, even if it adds a little bit more work for the admin :)

yvanzo:
> Seems sensible, even if it adds a little bit more work for the admin :)

Actually, removing `+prefix` and periods `.`is not more work than removing them. An additional feature that would save time for admins (already mentioned in the previous PR) would be an alternate search field for raw email (rather than regular expression).

reosarevok:
True, but I'm fine with merging this for now, if @mwiencek is ok with it, and maybe adding a ticket for that extra option if you can be bothered :)

lib/MusicBrainz/Server/Data/Editor.pm
mwiencek: I've used regular here-docs a lot in the past (grep `EOSQL`) and agree stripped ones would be preferable. :) The only reason I hadn't used those is that [they were added in Perl 5.26](https://www.effectiveperlprogramming.com/2016/12/strip-leading-spaces-from-here-docs-with-v5-26/) AFAICT.

mwiencek (approved):
This seems to provide more useful functionality than currently exists, and the tests are appreciated!

PR #2105, opened 2021-05-17
MBS-11664 / MBS-11666: Enable discID tab for releases that wrongly have discids + add report

reosarevok:
### Implement MBS-11664 / MBS-11666

Tested locally with sample data by changing hybrid SACD to not allow discids and then checking release/ca9de00b-3934-431b-809c-e07347f99801 and release/e7502670-ca88-4ec3-aa1b-eb5cba687d54

### DIFF
--- a/root/report/ShouldNotHaveDiscIds.js
+++ b/root/report/ShouldNotHaveDiscIds.js
@@ -1,0 +1,41 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import ReleaseList from './components/ReleaseList';
+import ReportLayout from './components/ReportLayout';
+import type {ReportDataT, ReportReleaseT} from './types';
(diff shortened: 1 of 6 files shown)


root/report/ShouldNotHaveDiscIds.js
```diff
@@ -0,0 +1,41 @@
...
+  items,
+  pager,
+}: ReportDataT<ReportReleaseT>): React.Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows releases that have at least one medium with a
+       format that does not support disc IDs, yet have disc IDs attached.
+       Usually this means the disc IDs ended up here because of a bug
+       and should be moved or removed.`,
```
mwiencek, line 29: Couldn't it also mean that the format is wrong?

mwiencek: I guess that's less likely given the current ways the bug can happen, but. Hopefully we can fix these + add a database constraint and remove this report one day.

mwiencek (approved):

root/report/ShouldNotHaveDiscIds.js
reosarevok: It could mean that the format is wrong, yes - something was always meant to be a CD, but got merged into a vinyl by mistake? But I think "usually" is still fair :)

PR #2107, opened 2021-05-18
Fix double JSON dump gpg-signing

mwiencek:
The production JSON dumps deployment currently hangs indefinitely with the following type of message:

> File '/tmp/mbscript-xxx/json-dump-xxx/MD5SUMS.asc' exists.
> Overwrite? (y/N)

This may be due to a gpg upgrade in the recent switch to Ubuntu 20.04.

To resolve this, for one, I'm adding `--batch --yes` to the gpg command to always overwrite existing .asc files.

Secondly, I'm fixing the code to not run `write_checksum_files` multiple times per dump.


reosarevok (approved):
I haven't used this before, but the changes seem sensible at least.

yvanzo (approved):

PR #2108, opened 2021-05-18
MBS-1801 (I): SQL additions to support in-site social aspects

reosarevok:
This adds five tables to their own messaging namespace.

`message` is for user-to-user messages - as such, it has a sender and a receiver, plus a title and content.

`hidden_message` allows for either the sender or receiver (or both) to "remove" a message from their messaging list. This is less complex than removing the message outright (on demand from which of the users?) and makes sure that the content is available to an admin to verify abuse without the user needing to keep the abusive message in their inbox.

`notification` is for system notifications. These could be implemented as messages from ModBot, but having them in a separate table seems simpler since we are likely to want to display them separately. It also makes it easier to deal with them separately if we decide to, for example, make notifications be removed some time after being read.
In preparation for such a possibility, `read` is a timestamp for `notification`, but just a boolean for `message` where we expect to only remove (hide) messages on request by the sender or receiver.

`edit_thanks` and `edit_note_thanks` allow editors to thank others for their edits and edit notes, respectively. The thanking itself
is stored in these tables, and I expect it should be possible for a user to check all their sent or received thanks later on.
The user being thanked should also be notified, so on a new thanking a notification should be inserted in the notification table. Whether this will happen via a trigger or through code can be decided later.


admin/sql/messaging/CreateTables.sql
```diff
@@ -0,0 +1,42 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+SET search_path = 'messaging';
+
+CREATE TABLE edit_note_thanks (
+  thanker             INTEGER NOT NULL, -- PK, references editor.id
+  edit_note		        INTEGER NOT NULL, -- PK, references edit_note.id
```
mwiencek, line 9: Might want to check your editor settings: several lines here have tab characters mixed in with spaces, leading to the weird indentation. :)

admin/sql/messaging/CreateTriggers.sql
```diff
@@ -0,0 +1,11 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+CREATE TRIGGER b_ins_hidden_message BEFORE INSERT ON hidden_message
+    FOR EACH ROW EXECUTE PROCEDURE ensure_editor_is_connected_to_message();
```
mwiencek, line 6: One trigger should suffice (using `INSERT OR UPDATE`). While I don't see any triggers named e.g. `b_ins_or_upd_*`, the ones that do apply to multiple operations seem to all use the function name, so I guess we can do that.
```suggestion
CREATE TRIGGER ensure_editor_is_connected_to_message BEFORE INSERT OR UPDATE ON hidden_message
    FOR EACH ROW EXECUTE PROCEDURE ensure_editor_is_connected_to_message();
```

admin/sql/messaging/CreateTriggers.sql
```diff
@@ -0,0 +1,11 @@
...
+
+CREATE TRIGGER b_upd_hidden_message BEFORE UPDATE ON hidden_message
+    FOR EACH ROW EXECUTE PROCEDURE ensure_editor_is_connected_to_message();
+
+COMMIT;
```
mwiencek, line 11: (missing newline)

admin/sql/messaging/CreateTables.sql
mwiencek, line 9: This is mostly a stylistic suggestion, but usually when we have a compound table like `foo_bar` then `foo` is the first column and `bar` is the second. Same for other tables below.
```suggestion
CREATE TABLE edit_note_thanks (
  edit_note		        INTEGER NOT NULL, -- PK, references edit_note.id
  thanker             INTEGER NOT NULL, -- PK, references editor.id
```

mwiencek (commented):
This looks really good overall. I think the indexes are fine, but we can always tweak them later if not.

mwiencek (approved):

PR #2110, opened 2021-05-19
MBS-11671: force hourCycle: 'h23' for %H in formatUserDate

mwiencek:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle

This seems to have broken with the upgrade to Node v16.

(I've honestly never seen 1-24 before so don't know why that's a default for en.)

### DIFF
--- a/root/static/scripts/tests/utility.js
+++ b/root/static/scripts/tests/utility.js
@@ -10,6 +10,7 @@ import test from 'tape';
 
 import formatDate from '../common/utility/formatDate';
 import * as age from '../../../utility/age';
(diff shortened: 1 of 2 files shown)


reosarevok:
Wanna add some tests? ;)

mwiencek:
> Wanna add some tests? ;)

I didn't add tests, but I did add a test. :)

reosarevok (approved):
Looks sensible, but do we really have zero formatUserDate tests? That's... not a lot. Well, a lot more now, I guess.

yvanzo (approved):

PR #2112, opened 2021-05-20
Standardize (uppercase) caps on SQL strings

reosarevok:
This fixes as many lowercase 'from', 'select', 'on', 'as', etc. as I could find. I'm sure I missed some. I also intentionally skipped issues in old SQL upgrade packages, since those probably shouldn't be changed.

### DIFF
--- a/po/extract_pot_db
+++ b/po/extract_pot_db
@@ -44,7 +44,7 @@ my $domain = $domain[0];
 # and ctx, which specifies a column to use for a msgctxt entry (unused at present).
 my @DBDEFS = (
               {'domain' => 'statistics', 'table' => 'statistics.statistic_event', 'columns' => ['title', 'description'], 'id' => 'date'},
-              {'domain' => 'countries', 'table' => 'area JOIN iso_3166_1 iso on iso.area = area.id', 'columns' => ['area.name'], 'comment' => ['iso.code']},
+              {'domain' => 'countries', 'table' => 'area JOIN iso_3166_1 iso ON iso.area = area.id', 'columns' => ['area.name'], 'comment' => ['iso.code']},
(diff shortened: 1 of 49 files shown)


po/extract_pot_db
```diff
@@ -44,7 +44,7 @@ my $domain = $domain[0];
 # and ctx, which specifies a column to use for a msgctxt entry (unused at present).
 my @DBDEFS = (
               {'domain' => 'statistics', 'table' => 'statistics.statistic_event', 'columns' => ['title', 'description'], 'id' => 'date'},
-              {'domain' => 'countries', 'table' => 'area JOIN iso_3166_1 iso on iso.area = area.id', 'columns' => ['area.name'], 'comment' => ['iso.code']},
+              {'domain' => 'countries', 'table' => 'area JOIN iso_3166_1 iso ON iso.area = area.id', 'columns' => ['area.name'], 'comment' => ['iso.code']},
```
reosarevok, line 47: I'm afraid this will cause issues with the .po(t) files, since they seem to include the db query? Hopefully it wouldn't cause new translations to be required, if so I'll drop this line.

mwiencek (approved):

po/extract_pot_db
yvanzo: It just changes an optional comment that helps translators with locating the source string in the database. It won’t unmap translations.

yvanzo (approved):
Found a few more occurrences:

* `from` at https://github.com/metabrainz/musicbrainz-server/blob/v-2021-05-19-hotfixes/po/extract_pot_db#L128
* `from` at https://github.com/metabrainz/musicbrainz-server/blob/v-2021-05-19-hotfixes/lib/MusicBrainz/Server/Sitemap/Overall.pm#L269

PR #2114, opened 2021-05-24
MBS-11680: Group editing URL relationships by external link

y-young:
# Problem

The current external links editor has the following drawbacks:

* It does not make obvious a URL can be pasted a second time to add another relationship type
* It is not easy to find if a URL has been pasted already (except when using the same relationship type)
* It requires double editing to change a URL having two relationships
* Error’s position is the same for both relationship type and URL, it isn’t always obvious which field must be edited to fix the error.

# Solution

MBS-11680: Group editing URL relationships by external link

Implement a new UI for external links editor, where links with the same URL are grouped together, relationship types are shown under the URL as a list, and URLs are numbered for reference in error messages.

## Details

* Additional relationship can be added in two different ways:
  * Either by pasting the same URL a second time (as before);
  * Or by clicking the 'Add another relationship' button.

* Additional input elements are shown only when needed:
  * 'Add another relationship' button only if permitted, (auto-selected relationship types currently don’t allow for a second relationship);
  * 'Remove relationship' button is displayed only if there are several relationships as at least one relationship is required per URL.

* Linked URL is highlighted when focused to ease keyboard navigation.

* Reference the position of the original URL through a notice when the pasted URL is a duplicate.  Users can then press enter or select a different type to confirm merging with the existing one.

* Improve layout to prevent relationship type description `?` button to hide other controls and to be unintentionally triggered.

* Improve workflow with links that have error by instantly appending new input boxes and not freezing them.

* Enable validating URLs in URL editing popover and display URL-related errors only (without re-rendering the whole list).

* Fix opening links in a new tab (to prevent editing data loss)

* Fix error target of restricted link types (to `RELATIONSHIP`)

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -604,21 +604,11 @@ const ExternalLinkRelationship =
   (props: ExternalLinkRelationshipProps): React.Element<'tr'> => {
     const {link, hasUrlError} = props;
     const linkType = link.type ? linkedEntities.link_type[link.type] : null;
-    let faviconClass: string | void;
     const backward = linkType && linkType.type1 > 'url';
 
     const showTypeSelection = (link.error || hasUrlError)
       ? true
       : !(link.urlMatchesType || isEmpty(link));
-
-    if (!showTypeSelection && link.urlMatchesType) {
-      for (const key of Object.keys(FAVICON_CLASSES)) {
-        if (link.url.indexOf(key) > 0) {
-          faviconClass = FAVICON_CLASSES[key];
-          break;
-        }
-      }
-    }
 
     return (
       <tr className="relationship-item" key={link.relationship}>
@@ -641,10 +631,6 @@ const ExternalLinkRelationship =
                   </LinkTypeSelect>
                 ) : (
                   <label className="relationship-name">
-                    {faviconClass &&
-                    <span
-                      className={'favicon ' + faviconClass + '-favicon'}
-                    />}
                     {linkType ? (
                       backward
                         ? l_relationships(linkType.reverse_link_phrase)
@@ -728,10 +714,22 @@ export class ExternalLink extends React.Component<LinkProps> {
     });
     const firstLink = props.relationships[0];
 
+    let faviconClass: string | void;
+    for (const key of Object.keys(FAVICON_CLASSES)) {
+      if (props.url.indexOf(key) > 0) {
+        faviconClass = FAVICON_CLASSES[key];
+        break;
+      }
+    }
+
     return (
       <React.Fragment>
         <tr className="external-link-item">
           <td>
+            {faviconClass &&
+            <span
+              className={'favicon ' + faviconClass + '-favicon'}
+            />}
             <label>
               {props.isOnlyLink
                 ? addColonText(l('Add link'))
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -119,12 +119,22 @@ export class ExternalLinksEditor
     });
   }
 
+  cleanupUrl(url: string): string {
+    if (url.match(/^\w+\./)) {
+      url = 'http://' + url;
+    }
+    return URLCleanup.cleanURL(url) || url;
+  }
+
   handleUrlChange(
     linkIndexes: Array<number>,
-    event: SyntheticEvent<HTMLInputElement>,
+    rawUrl: string,
   ) {
-    const rawUrl = event.currentTarget.value;
     let url = rawUrl;
+    if (url === '') {
+      this.removeLinks(linkIndexes);
+      return;
+    }
 
     this.setState(prevState => {
       let newLinks = [...prevState.links];
@@ -133,10 +143,7 @@ export class ExternalLinksEditor
 
         // Allow adding spaces while typing, they'll be trimmed on blur
         if (url.trim() !== link.url.trim()) {
-          if (url.match(/^\w+\./)) {
-            url = 'http://' + url;
-          }
-          url = URLCleanup.cleanURL(url) || url;
+          url = this.cleanupUrl(url);
         }
 
         let newLink = Object.assign({}, newLinks[index], {url, rawUrl});
@@ -308,19 +315,117 @@ export class ExternalLinksEditor
     }
   }
 
-  render(): React.Element<'table'> {
-    this.props.errorObservable(false);
-
+  validateLink(link: LinkStateT): ErrorT {
     const oldLinks = this.getOldLinksHash();
-    const linksArray = this.state.links;
-    const linksByUrl = Array.from(groupLinksByUrl(linksArray));
     const linksByTypeAndUrl = groupBy(
       uniqBy(
-        linksArray.concat(this.props.initialLinks),
+        this.state.links.concat(this.props.initialLinks),
         link => link.relationship,
       ),
       linkTypeAndUrlString,
     );
+    let error = null;
+
+    const linkType = link.type
+      ? linkedEntities.link_type[link.type] : {};
+    const checker = URLCleanup.validationRules[linkType.gid];
+    const oldLink = oldLinks[link.relationship];
+    const isNewLink = !isPositiveInteger(link.relationship);
+    const linkChanged = oldLink && link.url !== oldLink.url;
+    const isNewOrChangedLink = (isNewLink || linkChanged);
+    const linkTypeChanged = oldLink && +link.type !== +oldLink.type;
+    link.url = getUnicodeUrl(link.url);
+
+    if (isEmpty(link)) {
+      error = null;
+    } else if (!link.url) {
+      error = {
+        message: l('Required field.'),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (isNewOrChangedLink && !isValidURL(link.url)) {
+      error = {
+        message: l('Enter a valid url e.g. "http://google.com/"'),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (isNewOrChangedLink && isMusicBrainz(link.url)) {
+      error = {
+        message: l(`Links to MusicBrainz URLs are not allowed.
+                Did you mean to paste something else?`),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (isNewOrChangedLink && isMalware(link.url)) {
+      error = {
+        message: l(`Links to this website are not allowed
+                because it is known to host malware.`),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (isNewOrChangedLink && isShortened(link.url)) {
+      error = {
+        message: l(`Please don’t enter bundled/shortened URLs,
+                enter the destination URL(s) instead.`),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (isNewOrChangedLink && isGoogleAmp(link.url)) {
+      error = {
+        message: l(`Please don’t enter Google AMP links,
+                since they are effectively an extra redirect.
+                Enter the destination URL instead.`),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (!link.type) {
+      error = {
+        message: l(`Please select a link type for the URL
+                you’ve entered.`),
+        target: URLCleanup.ERROR_TARGETS.RELATIONSHIP,
+      };
+    } else if (
+      linkType.deprecated && (isNewLink || linkTypeChanged)
+    ) {
+      error = {
+        message: l(`This relationship type is deprecated 
+                and should not be used.`),
+        target: URLCleanup.ERROR_TARGETS.RELATIONSHIP,
+      };
+    } else if (
+      (linksByTypeAndUrl[linkTypeAndUrlString(link)] ||
+        []).length > 1
+    ) {
+      error = {
+        message: l('This relationship already exists.'),
+        target: URLCleanup.ERROR_TARGETS.RELATIONSHIP,
+      };
+    } else if (isNewOrChangedLink && checker) {
+      const check = checker(link.url);
+      if (!check.result) {
+        error = {
+          message: '',
+          target: URLCleanup.ERROR_TARGETS.NONE,
+        };
+        error.target = check.target ||
+          URLCleanup.ERROR_TARGETS.NONE;
+        if (error.target === URLCleanup.ERROR_TARGETS.URL) {
+          error.message = l(
+            `This URL is not allowed for the selected link type,
+            or is incorrectly formatted.`,
+          );
+        }
+        if (error.target ===
+          URLCleanup.ERROR_TARGETS.RELATIONSHIP) {
+          error.message = l(`This URL is not allowed 
+                    for the selected link type.`);
+        }
+        error.message = check.error || error.message;
+      }
+    }
+    return error;
+  }
+
+  render(): React.Element<'table'> {
+    this.props.errorObservable(false);
+
+    const linksArray = this.state.links;
+    const linksByUrl = Array.from(groupLinksByUrl(linksArray));
 
     return (
       <table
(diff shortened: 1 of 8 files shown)


reosarevok:
Re: "Merge branch 'metabrainz:master' into MBS-11680" - generally, you'll want to just rebase instead (git rebase master).

root/static/scripts/edit/externalLinks.js
```diff
@@ -439,6 +439,7 @@ export class ExternalLink extends React.Component<LinkProps> {
               />}
           </td>
         </tr>
+        {isEmpty(props) ||
```
reosarevok, line 438: The indenting needs fixing here :) Also, we generally do `isEmpty(props) ? null : component` - both ways should work, but it seems more consistent to stick with that.

root/static/scripts/edit/externalLinks.js
```diff
@@ -401,68 +405,86 @@ export class ExternalLink extends React.Component<LinkProps> {
...
+      <React.Fragment>
+        <tr>
+          <td>
+            <label>
+              {props.isOnlyLink
+                ? l('Add link:')
+                : (
+                  props.isLastLink
+                    ? l('Add another link:')
+                    : props.index + 1
```
reosarevok, line 417: I'm not sure this is useful - it doesn't really say anything about the link, and people can see it's the first in the list or whatever :) If we have nothing good to show, do we need to show anything at all?

root/static/scripts/edit/externalLinks.js
```diff
@@ -480,6 +477,10 @@ export class ExternalLink extends React.Component<LinkProps> {
                   {l('video')}
                 </label>
               </div>}
+            {props.errorMessage &&
+              <div className="error field-error" data-visible="1">
+                {props.errorMessage}
+              </div>}
```
reosarevok, line 483: This is generally good, but: 
Screenshot from 2021-05-25 18-48-26
This use case shows a situation where it feels off - it seems like the required field would be the (already filled in) type, but what it means is the actual URL. Can we add the error to the appropriate input (type if it fails a validation, text if it's an error unrelated to the type) instead, and have both options possible? :)

reosarevok (changes requested):
The idea seems nice and works pretty well, good job! Left a few comments that either need fixes (linting) or could IMO be improved.

root/static/scripts/edit/externalLinks.js
y-young: In fact I was referring to the original code on line 431 for consistency, it is not my usual way, too. If we agree that this could be better we could change both lines. :)

reosarevok: Oh, I see. Unless @yvanzo or @mwiencek see a reason to have it like this, I'd do the usual. But wait until they confirm :)

root/static/scripts/edit/externalLinks.js
y-young: It is intended to show the index of a link in a list, since originally there're indentation in each table row, so to align with the fieldset above I kept this indentation, we can still discuss and optimize it.

root/static/scripts/edit/externalLinks.js
y-young: Makes sense, I'll dig into the code later to see if we can realize it.

root/static/scripts/edit/externalLinks.js
reosarevok: Yeah, I understand the reasoning, it just seems kinda... not useful to have "1", "2" etc :) But I also understand it might feel weird to just have nothing at all there as a label. Since the move is intended to eventually allow selecting more than one type here, showing the favicon or something to that effect is probably not ideal, so I'm not sure what would be a good solution that still keeps a label. @mwiencek, @yvanzo: any ideas?

mwiencek:
I'm not sure I understand the motivation behind the changes here.  You can already paste the same link multiple times (and select a different type for each), so we already support what the ticket suggests. The current UI doesn't suggest to me that you can't do that.

The proposed UI seems more limited, since if you want to change the URL for a specific type you have to remove + re-add it, whereas currently you just edit a single entry.  Am I missing an important advantage of grouping them like this?

root/static/scripts/edit/externalLinks.js
yvanzo: I find it useful to see the total number of links in the list. Numbers also help with not getting lost in long lists of external links. And we don’t have anything better to show at the moment.

root/static/scripts/edit/externalLinks.js
yvanzo: It used to be the usual. @mwiencek can better explain why it changed. That is related to Flow IIRC.

root/static/scripts/edit/externalLinks.js
yvanzo: This would be nice to have but can probably be added later on.
It requires a lot of rework to make errors be more specific about which input(s) is (are) in cause.
This is probably worth a separate ticket.

yvanzo:
> I'm not sure I understand the motivation behind the changes here. You can already paste the same link multiple times (and select a different type for each), so we already support what the ticket suggests. The current UI doesn't suggest to me that you can't do that.

I never noticed that I can do that, so I guess it doesn’t suggest to me that I can do that either. But the ticket is linked to MBS-9902 (Support auto-select/cleanup/validation of more than one relationship type for external links), so it is more than supporting hand-picked relationship types for unconstrained URLs.

> The proposed UI seems more limited, since if you want to change the URL for a specific type you have to remove + re-add it, whereas currently you just edit a single entry.

I see this as an advantage, because the current UI does not make obvious that editing an URL field actually creates a new URL and drops/adds relationships rather than modify the current URL. Besides, do you often need to change an URL for a specific type?

But it is more limited because it doesn’t support multiple relationship types where the current UI does (which I didn’t know). That should be fixed.

> Am I missing an important advantage of grouping them like this?

There is no immediate advantage, this first step mainly rearranges the layout for later improvements such as:
* URL cleanup should be the same for all occurrences of an URL, so it makes sense to group them by URL so you can handle cleanup options only once.
* It is easier to check if an URL has all appropriate relationships to an entity if they are grouped at the same place.
* It is easier to modify (e.g. mark as ended) all relationships to an URL if they are grouped at the same place.
* Auto-selected relationship types and validation errors will be more easily visible too.

More generally, grouping by relationship type seems to make more sense when editing data than when displaying data. This is more specifically the case for URL relationships because their types are learned from the webpage itself, rarely from a full-text description, unlike artist credits from booklet.

mwiencek:
> I never noticed that I can do that, so I guess it doesn’t suggest to me that I can do that either.

I guess because you didn't try it. :smile:  172,938 releases have at least one URL pointing to multiple types, which is 8.2% of releases with any URLs at all.  So it's not something nobody figured out.

But to put that into perspective, 171,969 (99.4%) of those are "get the music" type relationships.  So this is optimizing the UI for a single relationship type when the current UI is much simpler for all other types.

We could just as easily have a custom UI for "get the music." For example, we can allow only selecting the parent type and add checkboxes underneath the URL to select download, streaming, etc.

> I see this as an advantage, because the current UI does not make obvious that editing an URL field actually creates a new URL and drops/adds relationships rather than modify the current URL.

That's kinda intended though.  Most casual users aren't aware that URLs are actual entities you can modify independent of their relationships, and probably shouldn't be -- it's confusing, and they might mess something up if they try to edit the URL itself.  I even consider it a weird implementation detail sometimes.

y-young:
> if you want to change the URL for a specific type you have to remove + re-add it

I'm afraid I didn't understand your question, you can still change the URL directly in the input box as usual.

> But it is more limited because it doesn’t support multiple relationship types where the current UI does (which I didn’t know). That should be fixed.

Also, I think you can still copy-paste the same URL and choose different relationship types, I don't see where's the problem.

> But to put that into perspective, 171,969 (99.4%) of those are "get the music" type relationships. So this is optimizing the UI for a single relationship type when the current UI is much simpler for all other types.
>
> We could just as easily have a custom UI for "get the music." For example, we can allow only selecting the parent type and add checkboxes underneath the URL to select download, streaming, etc.

It needs further investigation to determine the possible combinations of relationships, now I'm only searching for the most flexible approach.

This ticket mainly aims to optimize the layouts instead of changing the logics. Besides, it aims to support MBS-3774 by sparing extra space to display other attributes like begin/end dates.

mwiencek:
> I'm afraid I didn't understand your question, you can still change the URL directly in the input box as usual.

Sorry that wasn't clear. My understanding is that you (eventually) intend to group relationships under a single URL item, where multiple types can be listed/selected. If that were the case, editing a URL would logically apply the change to all types. But say you only wanted to edit the URL for one of the types, because e.g. the streaming page was moved to a separate URL, or something like that.

> Also, I think you can still copy-paste the same URL and choose different relationship types, I don't see where's the problem.

Correct, I can still do this on your branch. That's why I was confused by MBS-11680, which says:

> Before we support multiple relationships for a single external link, the UI must be optimized to display relationships as a list.

(but we already support multiple relationships to a single external link, as you now note. :smile:)

> This ticket mainly aims to optimize the layouts instead of changing the logics. Besides, it aims to support MBS-3774 by sparing extra space to display other attributes like begin/end dates.

I fully support these tickets, just questioning the right way to optimize the UI for them. :)  For example, with regards to MBS-3774, I didn't expect that you intended to display these attributes inline, but maybe have a small icon to open a relationship-editing dialog. Before GSoC started, @yvanzo had asked me about using the new React-based relationship editor dialog code here, which is why I thought that.

y-young:
> If that were the case, editing a URL would logically apply the change to all types. But say you only wanted to edit the URL for one of the types, because e.g. the streaming page was moved to a separate URL, or something like that.

I doubt that the frequency of these situations are relatively low, and display clarity might weigh over that.

> Correct, I can still do this on your branch. That's why I was confused by [MBS-11680](https://tickets.metabrainz.org/browse/MBS-11680), which says:
> 
> > Before we support multiple relationships for a single external link, the UI must be optimized to display relationships as a list.
> 
> (but we already support multiple relationships to a single external link, as you now note. 😄)

Well, maybe we have some misunderstandings here :D But I agree with @yvanzo that it is more resonable to group relationships by links.

> For example, with regards to [MBS-3774](https://tickets.metabrainz.org/browse/MBS-3774), I didn't expect that you intended to display these attributes inline, but maybe have a small icon to open a relationship-editing dialog. Before GSoC started, @yvanzo had asked me about using the new React-based relationship editor dialog code here, which is why I thought that.

Actually I intended this implementation from the beginning, in the proposal I put a UI prototype displaying those attributes inline, with begin/end dates behind the relationship name, and grayed out text if relationship is ended. As for the dialog, IMO that was merely for editing purposes. :)

mwiencek:
Right, I did see that you displayed the dates inline in your proposal and have a pencil icon for editing. That's closer to what I expected. :)  (For a second I thought the *editing* was going to happen inline.)

I'm happy with any UI changes that the community is.  It does seem a bit like we're over-optimizing for use cases that don't apply to most URLs -- the UI here takes up a significant amount of more space which the majority of URLs don't need, and that's my main concern.

So I just suggest that you get some feedback from the editing community too, because my feedback alone is useless. Perhaps link to your proposal in the MusicBrainz section of the forums where more people will see it.

yvanzo:
The relationship type display and the new button “Add another relationship” look fine.

List items should probably labeled with “URL:” instead of order, just like list items of release events.

A couple of blocking issues I noticed after a quick test:
1. There should not be any removal button for relationship type when there is only one type, as it is mandatory to have at least one relationship type per URL.
2. When adding a second relationship type and removing the first one, a new blank URL is added to the list.

There probably are more issues, I did not test any further as I guess further issues might have the same cause.

yvanzo (changes requested):
It should not be possible to add another relationship type when it is auto-selected;

Just did a quick test, did not review the code yet.

y-young:
> It should not be possible to add another relationship type when it is auto-selected. It can currently be done in two different ways:
> 
> * either click the button “Add another relationship”,
> * or add the same URL again and select another relationship type.
> 
> Just did a quick test, did not review the code yet.

We can hide "Add another relationship" button, but I think it would either be impossible or too complex to block the second way. IIUC on musicbrainz.org it's not blocked, you'll be able to select another type but there's just no suitable type.

yvanzo:
Right, the second was an already existing issue and it does not allow to bypass validation anyway; Let's address it later on.

yvanzo (changes requested):
Two display issues:

1. When adding `https://www.deezer.com/album/356913` to <http://localhost:5000/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/edit> the relationship type is not correctly interpolated: It should be `stream for free` instead of `stream {video} for free`.

2. For the same editing <http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/edit> using the French localization, there is a glitch: When relationship type is too long, it is put on a separate line from “Type:”.
Adding a link

y-young:

y-young:

y-young:
> Two display issues:
> 
> 1. When adding `https://www.deezer.com/album/356913` to http://localhost:5000/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/edit the relationship type is not correctly interpolated: It should be `stream for free` instead of `stream {video} for free`.
> 2. For the same editing http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/edit using the French localization, there is a glitch: When relationship type is too long, it is put on a separate line from “Type:”.
>    Adding a link

The second issue is fixed. I checked the first issue on musicbrainz.org, and it turned out to be `stream {video} for free`, is this an issue?

yvanzo:
> > Two display issues:
> > 
> > 1. When adding `https://www.deezer.com/album/356913` to http://localhost:5000/release/0ce274e3-3b89-4d15-94ac-4f92e497ea60/edit the relationship type is not correctly interpolated: It should be `stream for free` instead of `stream {video} for free`.
> 
> I checked the first issue on musicbrainz.org, and it turned out to be `stream {video} for free`, is this an issue?

Right, that is yet another existing issue that can be addressed later on.

> > 2. For the same editing http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/edit using the French localization, there is a glitch: When relationship type is too long, it is put on a separate line from “Type:”.
> >    Adding a link
> 
> The second issue is fixed. 

Yes but the fix created another glitch: The type selector is too thick now as the error message is no longer on a separate line. This happens when adding an URL without type auto-select such as `https://www.alexisffrenchmusic.com/`.

reosarevok:
From quick testing: agreed with @yvanzo's comment about errors being on the same line not being a good idea: 
Screenshot from 2021-08-04 13-12-10

Also confusing: 
Screenshot from 2021-08-04 13-12-36
This error actually applies *only* to the setlist.fm relationship, but there's no way to tell from the current display. If we display URLs grouped like this, then the error should probably be moved to be a relationship error, not a URL error.

reosarevok:
Another thing that, while understandable, is super annoying: if I don't realize that I'm entering a link that is already there, it will at first show as a new link, but as soon as I'm trying to select a relationship type, it will detect (I assume on link field blur?) that it's the same link, and group it. Which is great in theory, but means I'm clicking at air :D See:

Peek 2021-08-04 13-22

Would it be possible to make it so that it doesn't move *yet* if the user only leaves the URL field to select a relationship type for the link?

reosarevok:
Other than those things, I do like how this works. I'm still not a fan of the 1/2/3 labels, I think they don't give any benefit over having nothing at all. I notice that the favicon is now displayed by the relationship type, but that's actually not that connected to the relationship, it's connected to the URL itself. What about showing the favicon instead of the 1/2/3 label, and just showing the type name under Type? It would look more like on the sidebar, and we can use the generic "no favicon" link for URLs that do not match any of our favicons :) Something like:

Screenshot from 2021-08-04 13-36-19

Notice that also would show the favicon for Apple Music, which we have but were not showing at all right now :)

root/static/scripts/edit/externalLinks.js
reosarevok: I'm still not a fan of the 1/2/3 labels, I still feel they don't give any benefit over having nothing at all (I don't generally have a reason to care about the number of links). I notice that the favicon is now displayed by the relationship type, but that's actually not that connected to the relationship, it's connected to the URL itself. What about showing the favicon instead of the 1/2/3 label, and just showing the type name under Type? It would look more like on the sidebar, and we can use the generic "no favicon" link for URLs that do not match any of our favicons :) Something like:

Screenshot from 2021-08-04 13-36-19

Notice that also would show the favicon for Apple Music, which we have but were not showing at all right now :)

yvanzo:
See also the continued thread about icons vs numbers: https://github.com/metabrainz/musicbrainz-server/pull/2114/files/4f8d19d24c5405f630e87dd3adcd16a8f6240fcb#diff-20a3affccc96c36d0a67c0ca98ebc7010590a1adc067eeea57147b0302bf1add

root/static/scripts/edit/externalLinks.js
yvanzo: 1. About external favicons:
   1. Icons match either the external website (e.g. Spotify) or the relationship type (e.g. official homepage) - sometimes it is the same (e.g. Wikidata),
   2. most of icons match external websites,
   3. there is currently no icon for many websites (there is a generic icon though) and most of relationship types.

   Thus icon should be displayed at the right place indeed which is most of the time ahead of the link rather than the relationship type.

2. About icons vs numbers:
   1. It can be used in error message as a reference to point to when grouping items together (e.g. to address the below issue <https://github.com/metabrainz/musicbrainz-server/pull/2114#issuecomment-892546511>),
   2. it can be useful to navigate through very long lists of links (e.g. the Beatles have 56 links, test example taken from the forum),
   3. it can help with distinguishing between similar links (e.g. MB artist matching 3 Discogs entries) if provided in addition to icon.

   So numbers can have some benefit and can also be used along with icons which look much nicer indeed. But my main concern about numbers is: It might let editors think numbers are actual data (likewise track numbers) which is not the case. It would probably be less confusing to find another way to refer to other list items from error messages.

P.S. The 1. is a just list a statements, while the 2. is more thinking out loud.

y-young: screenshot
Here's the new layout, put favicon before index numbers for better alignment (when there's no favicon).

y-young:
Here's a brief description on the requirements and behaviors of the new external links editor:

## Requirements
- URLs in the list are displayed as clickable links
- Links with the same URL are grouped together
- Users can add new links continuously by clicking newly appearing input boxes
- Users can add a new relationship to an exisiting link by pasting the URL again
- Tab sequence should make sense
- Focus should not be lost

## To add link to the list(submit the link)
- Press Enter / Tab (With type unselected or auto-selected)
    - Pressing Tab won't work if it's a duplicate
- Fill the URL and then select the type
- Click anywhere outside the input box
    - Won't work if it's a duplicate

## To add new relationship to an exsisting link
- Click 'Add another relationship' button (focus should be directed to the type select)
- Paste the same URL in the input box again
    - A message should be shown to notify the user of the existing URL
    - The two URLs are merged after selecting type or pressing Enter, and focus is directed to the empty input box

## To delete an exising link
- Click 'x' icon beside the relationship type to delete single link
- Click 'x' icon beside the URL to delete alls link with this URL
- Clear the input box to delete all links with this URL
- For a link that is not submitted yet, the delete icon is still shown and working

yvanzo (changes requested):
It looks much better, and merging is much smoother too, thanks!

Beyond the issue about label’s display discussed with @mwiencek (label is not correctly displayed for me either), here is a pair of small regressions that should be fixed:

1. Opening external links in a separate tab would prevent loss of editing data. (Previously, external links from the bubble were opened in a separate tab too.)
2. Adding a focus indicator on external links would help with key navigation. (Previously, all focusable elements of this field set had native focus indicators.)

yvanzo (approved):
Thanks, I cannot reproduce the tabbing issue anymore, did stress it a bit but could not find other issue either.

PR #2118, opened 2021-05-25
Indent heredocs (<<EOSQL) properly

reosarevok:
From v5.26 onwards, Perl no longer requires heredocs (which we mostly use for EOSQL entries) having no indentation at all for the end line. Instead, all spaces up to the start of the end line text are removed from all lines in the heredoc. This allows us to do proper spacing.

I'm also standardizing some other things on the blocks I touch, such as VALUES and indentation tabs.

Split this into three commits for three slightly different situations, see commits / commit messages.

Note: some editors, like VS Code, don't really understand this properly (yet). A fix was merged a while ago on what seems to be the upstream from VS Code Perl support (https://github.com/textmate/perl.tmbundle/pull/42) so hopefully that will get into VS Code Perl soonish.


t/lib/t/Edit.pm
```diff
@@ -6,12 +6,11 @@ around run_test => sub {
...
 
-    $self->c->sql->do(<<'EOSQL');
-INSERT INTO editor (id, name, password, ha1, email, email_confirm_date)
-VALUES
-  (1, 'editor', '{CLEARTEXT}pass', '3f3edade87115ce351d63f42d92a1834', 'foo@example.com', now()),
-  (200, 'editor200', '{CLEARTEXT}pass', '3f3edade87115ce351d63f42d92a1834', 'foo2@example.com', now());
-EOSQL
+    $self->c->sql->do(<<~'EOSQL');
+        INSERT INTO editor (id, name, password, ha1, email, email_confirm_date)
+            VALUES (1, 'editor', '{CLEARTEXT}pass', '3f3edade87115ce351d63f42d92a1834', 'foo@example.com', now()),
```
mwiencek, line 11: Looking at this again, since we align SELECT, FROM, WHERE, AND, etc. all at the same level, wouldn't it make sense for INSERT and VALUES to be at the same level too?

reosarevok: I dunno, as you prefer :) I guess I see FROM/WHERE as equally important than SELECT but VALUES as less important than INSERT, but I'm not too bothered by either way.

mwiencek: Not sure about the relative importance, but it is true that VALUES is a standalone command (like how you can do INSERT INTO ... SELECT) whereas FROM/WHERE are tied to the SELECT.

mwiencek (approved):

yvanzo (approved):
Thanks

PR #2120, opened 2021-05-26
Upgrade Babel dependencies

mwiencek:
Via `yarn upgrade --scope @babel --latest --exact`.

The plugin-transform-react-constant-elements update should address a serious bug I ran into a while ago:

https://github.com/mwiencek/babel-issue-11686


mwiencek:
(This shouldn't require any review, so merging to master.)

PR #2121, opened 2021-05-26
Replace direction prop with backward

mwiencek:
We often need to know whether a relationship is backward (meaning: its target is entity0), and have a lot of `direction === "backward"` checks throughout the code. It would be much simpler to replace these with a boolean, as we don't support more than two directions anyway.


reosarevok (approved):
Seems sensible and all tests pass, so :)

PR #2122, opened 2021-05-26
Small Flow type changes from the relationship editor branch

mwiencek:
Collects some small commits from my relationship editor branch which tweak our Flow types.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm
+++ b/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm
@@ -588,6 +588,7 @@ sub _seeded_url
     _report_unknown_fields($field_name, $params, $errors, @known_fields);
 
     my $result = {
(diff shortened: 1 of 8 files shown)


reosarevok (approved):
> We only made it optional to deal with seeded URLs in the release editor, but we can just have a separate type for those.

Can't we just add an empty attributes array to those when parsing the seeded data instead? I understand our plan is anyway to allow better URL editing in the RE, so we should also allow seeding dates then, eventually :)

mwiencek:
> Can't we just add an empty attributes array to those when parsing the seeded data instead? I understand our plan is anyway to allow better URL editing in the RE, so we should also allow seeding dates then, eventually :)

We could def. do that in the future if we need it, though it'd probably still be a separate type, since we only send `id`/`target`/`linkTypeID` and not a full `RelationshipT`.

PR #2123, opened 2021-05-27
a_upd_release_event: check for changes before doing work

mwiencek:
Avoid re-calculating `*first_release_date` data if the release event didn't actually change.

This can happen if the trigger fired on the `release_country` table and only the country changed. (Rather than checking for that, it's still correct to test the data we're concerned about: `UPDATE` triggers fire even when no data changed.)

Additionally, make sure we're only calling `set_release_first_release_date` once if the release column is unchanged.  (This would occur on merges, and probably no other time.)


mwiencek:
I noticed this while adding CAA triggers on the `release_first_release_date` table (which I don't want to fire unnecessarily).

reosarevok (approved):
Has this shown to be pretty slow or? Anyway, seems sensible enough, AFAICT.

mwiencek:
> Has this shown to be pretty slow or? Anyway, seems sensible enough, AFAICT.

No, but it might trigger unnecessary updates on the `release_first_release_date` table like I mentioned above (which could push unnecessary updates to the artwork indexer).

> Do we have pgtap tests for any of these?

Yes: t/pgtap/first_release_dates.sql

yvanzo (approved):
(need rebasing)

PR #2124, opened 2021-05-27
Even more Autocomplete2 improvements

mwiencek:
More commits cherry-picked from my relationship editor branch, but these all have to do with improving/fixing bugs in Autocomplete2. Most of these are simply minor issues that I ran into while testing, but other commits involve a lot of changes in order to support the following features:

 * Paginating static item lists (important for link attribute types / instruments).
 * Showing recently-used entities (including a fix for MBS-10498, which is achieved by re-fetching entities via a new endpoint, `/ws/js/entities`).
 * Implementing the "add new entity" option and converting the dialog for that to React.

To test, run `./script/compile_resources.sh tests` (ignoring any errors that show up) and navigate to http://localhost:5000/static/scripts/tests/autocomplete2.html


mwiencek:
It's also be helpful to test whether the add-new-entity functionality still works in the old autocompletes.

root/static/scripts/common/components/Autocomplete2.js
```diff
@@ -434,18 +437,18 @@ export default function Autocomplete2<+T: EntityItem>(
...
       items.length > 0 &&
-      cleanInputValue === clean(props.inputValue)
+      cleanInputValue === clean(inputValue)
     ) {
       dispatch(SHOW_MENU);
     } else if (nonEmpty(cleanInputValue)) {
       if (staticItems) {
         dispatch({
-          searchTerm: inputValue,
+          searchTerm: newInputValue,
```
reosarevok: Is this meant to be new rather than clean?

root/static/scripts/common/components/Autocomplete2/reducer.js
```diff
@@ -201,7 +201,7 @@ export function generateStatusMessage<+T: EntityItem>(
   return '';
 }
 
-function filterStaticItems<+T: EntityItem>(
+export function filterStaticItems<+T: EntityItem>(
```
reosarevok, line 239: You export this, but you don't seem to be using it elsewhere?

root/static/scripts/common/MB/Control/Autocomplete.js
```diff
@@ -487,19 +488,32 @@ $.widget('mb.entitylookup', $.ui.autocomplete, {
...
-            },
-          });
+          const containerNode = self._getAddEntityContainer()[0];
+
+          const closeAndReturnFocus = () => {
+            ReactDOM.unmountComponentAtNode(containerNode);
+            self.element.focus();
+          };
+
+          /* eslint-disable react/jsx-no-bind */
```
reosarevok, line 503: You have mentioned wanting to disable this rule globally before, FWIW.

reosarevok (commented):
From 0545e890cffff02cdc3e4761b720d1e87ab0546a

> This commit allows re-initiating the search in this scenario (by hitting key down, or pressing the icon)

Wouldn't it make sense to stop the search for the old type and start a search for the new type if the user changes the type? (without it needing to be hand-requested)

reosarevok:
From the test page, I'm getting:
```
arrays.js:155 Uncaught TypeError: Cannot read property 'reduce' of undefined
    at keyBy (arrays.js:155)
    at Object../root/static/scripts/tests/autocomplete2.js (autocomplete2.js:21)
    at __webpack_require__ (bootstrap:19)
    at bootstrap:83
    at bootstrap:83
```

mwiencek:
You might have to remove root/static/scripts/tests/typeInfo.js and re-run `compile_resources.sh tests`, since normally we skip re-generating that file if it already exists.

mwiencek:
> Wouldn't it make sense to stop the search for the old type and start a search for the new type if the user changes the type? (without it needing to be hand-requested)

If a search is in progress, I think -- it might already do this, but it's kinda hard to test. :)  But not sure it should initiate a new search on its own if you only change the type while text is entered (but there's no active search).

root/utility/invariant.js
```diff
@@ -0,0 +1,18 @@
+/*
+ * @flow strict
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export default function invariant(
```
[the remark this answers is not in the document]
mwiencek: I'm adding a comment, but don't see any dedicated Flow doc.  The only place I found that they mention it is in https://flow.org/en/docs/linting/rule-reference/#toc-unnecessary-invariant (which is a lint related to it).

root/static/scripts/common/components/Autocomplete2.js
```diff
@@ -95,11 +96,18 @@ function doSearch<T: EntityItem>(
...
 
     const actions = [];
-    let newItems = JSON.parse(searchXhr.responseText);
-    const pager = newItems.pop();
+    const entities = JSON.parse(searchXhr.responseText);
+    const pager = entities.pop();
     const newPage = parseInt(pager.current, 10);
     const totalPages = parseInt(pager.pages, 10);
 
+    let newItems: Array<Item<T>> = entities.map((entity: T) => ({
```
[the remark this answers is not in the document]
mwiencek: On this line it is, but it causes Flow issues if I change that because `state.results` is `Item<T>`.

root/static/scripts/common/components/Autocomplete2/constants.js
```diff
@@ -40,57 +40,68 @@ export const DISPLAY_NONE_STYLE = {display: 'none'};
 
 export const EMPTY_ARRAY: $ReadOnlyArray<empty> = Object.freeze([]);
 
+/* eslint-disable sort-keys */
```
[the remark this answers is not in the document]
mwiencek: Just more readable to me that way. If I'm looking at an item definition the first thing I want to know is its type.

root/static/scripts/common/components/Autocomplete2/reducer.js
```diff
@@ -95,6 +95,32 @@ function showError<+T: EntityItem>(
...
+  let index = startingIndex;
+  let count = items.length;
+
+  while (true) {
+    if (index >= items.length) {
+      index = 0;
+    }
+    const item = items[index];
+    // $FlowIgnore[sketchy-null-bool]
+    if (!item.disabled) {
```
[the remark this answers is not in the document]
mwiencek: Yea, but I prefer disabling the lint 'cause it's clearer what the intent is.  With `/*:: === true */` I feel like we'd need another comment saying `// silence sketchy-null-bool` to explain what that's even doing, which defeats the purpose.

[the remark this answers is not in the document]
reosarevok: Hmm, ok. I've done that elsewhere, without any extra commments, but :)

root/static/scripts/tests/autocomplete2.js
```diff
@@ -8,37 +10,47 @@ const {
...
+  let level = 0;
+  let parentId = type.parent_id;
+  let parentType =
+    parentId == null ? null : attributeTypesById[String(parentId)];
+  while (parentType) {
+    level++;
+    parentId = parentType.parent_id;
+    parentType =
+      parentId == null ? null : attributeTypesById[String(parentId)];
+  }
```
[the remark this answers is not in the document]
mwiencek: Sure, but we're calculating it from what we're given in `./typeInfo`, which other tests use too. So I didn't want to change a bunch of stuff just to make one test look nice.

root/static/scripts/common/components/Autocomplete2.js
mwiencek: I don't remember but it looks like this code got removed or rewritten in a later commit.

root/static/scripts/common/components/Autocomplete2/reducer.js
mwiencek: Oh, true, though I do import it somewhere in my relationship editor branch, so probably fine to leave it.

root/static/scripts/common/MB/Control/Autocomplete.js
mwiencek: Well, it's not that bad since we set `ignoreDOMComponents: true` on it. (In fact this is the only place I can see that we disable it anymore.)

root/static/scripts/common/components/Autocomplete2.js
mwiencek: Since it's filtering, it probably did intend to use `newInputValue` though since spaces affect that.

root/static/scripts/common/components/Autocomplete2/reducer.js
```diff
@@ -116,25 +142,40 @@ export function runReducer<+T: EntityItem>(
...
+        if (!item.disabled) {
+          state.highlightedItem = item;
+          break;
+        }
+        count--;
+        if (count <= 0) {
+          break;
+        }
+        index--;
       }
```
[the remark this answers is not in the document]
mwiencek: Pushed a commit for that

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -238,6 +241,23 @@ function formatLinkAttributeType(type: LinkAttrTypeT) {
...
 }
 
+function formatLinkType(linkType: LinkTypeT) {
+  let description = linkType.l_description;
+  if (description != null) {
+    // We want to strip html from the non-clickable description
+    const div = document.createElement('div');
+    div.innerHTML = description;
+    description = div.textContent;
+  }
```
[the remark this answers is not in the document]
mwiencek: Pushed a commit for that

reosarevok (approved):
Testing with the test page didn't show any obvious problems. I looked through the code and it seemed fine (except for a few things I commented on earlier and were mostly changed), but there's a ton of commits so I might have missed some stuff :)

root/static/scripts/common/components/Autocomplete2/formatters.js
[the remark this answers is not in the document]
yvanzo: https://github.com/metabrainz/musicbrainz-server/pull/2124/commits/b95b6dfaffdb86454e6813e17f24a517addf5b09

lib/MusicBrainz/Server/Controller/WS/js.pm
```diff
@@ -441,6 +446,94 @@ sub entity : Chained('root') PathPart('entity') Args(1)
...
 
+Readonly our $MAX_FETCHED_ENTITIES => 25;
+
+Readonly our $MAX_FETCHED_ENTITIES_URL_LENGTH => (
+    # 36 => Length of an MBID.
+    # 24 => Number of separators.
+    ($MAX_FETCHED_ENTITIES * 36) + ($MAX_FETCHED_ENTITIES - 1)
+);
+
+sub entities : Chained('root') PathPart('entities') Args(2)
```
yvanzo, line 457: This endpoint will be publicly accessible, even though it is neither promoted nor supported as public API. Is there anything to prevent potential abuse?

root/static/scripts/common/components/Autocomplete2/recentItems.js
```diff
@@ -0,0 +1,272 @@
...
+        return null;
+      }
+      return resp.json();
+    }).then((data: WsJsEntitiesDataT | null) => {
+      if (!data) {
+        return cachedList;
+      }
+
+      const results = data.results;
+      invariant(cachedList);
```
yvanzo, line 189: It’s the first usage ever of this function you just introduced. Please explain (in commit message?) why is that needed here.

yvanzo (commented):
I didn’t see any localization happening on the Autocomplete2 test page, otherwise didn’t see any bug either. Tested `/ws/js/entities` on test.mb.o which should probably be protected.

lib/MusicBrainz/Server/Controller/WS/js.pm
reosarevok: Isn't that already the case with all of ws/js?

root/static/scripts/common/components/Autocomplete2/recentItems.js
reosarevok: That's explained on https://github.com/metabrainz/musicbrainz-server/pull/2124/commits/84363005bf7013adef02f257538365a28f6fe6f9 (not saying it's not confusing, though...)

lib/MusicBrainz/Server/Controller/WS/js.pm
mwiencek: Well, it's limited to 25 entities.  You can request far more information from /ws/2.  And yeah, all endpoints under /ws/js are public -- some are probably already used by userscripts.

root/static/scripts/common/components/Autocomplete2/recentItems.js
mwiencek: It's not actually needed here, so I removed it.  (Must've been left from a previous version of the code.)  I added a small example to the commit message where I introduce the function.

yvanzo: Thanks!

lib/MusicBrainz/Server/Controller/WS/js.pm
yvanzo: Yes it provides the same info as `/ws/js/entity` already does, but at a higher rate.
Rate limiting or statistics are not handled from this repository anyway, just asking.

yvanzo (approved):

PR #2125, opened 2021-06-01
MBS-11688: The external link section will disappear once any link starts with //

y-young:
# Problem

MBS-11688: The external link section will disappear once any link starts with //

# Solution

Fix `isValidURL` to correctly validate links starting with "//"


reosarevok (approved):
Seems to work fine.

mwiencek (approved):

PR #2127, opened 2021-06-03
MBS-10844: Report for same URL linked to many entities 

reosarevok:
### Implement MBS-10844

While these can sometimes be correct, a run on sample data shows a lot more to fix than to keep, so I feel it's still worthwile.
This excludes different links that already get their own reports (while we could subsume them all into this, I think it's useful to have those separate since they are a lot more focused), and the license relationship, where we would expect a lot of URL reuse.
After a first test I realized it also needs to ignore cases where the same URL is linked multiple times from the same entity (such as a bandcamp link being both "stream" and "download").

On top of https://github.com/metabrainz/musicbrainz-server/pull/1871 from which it reuses a fair amount of the query.

### DIFF
--- a/root/report/LinksWithMultipleEntities.js
+++ b/root/report/LinksWithMultipleEntities.js
@@ -1,0 +1,98 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import PaginatedResults from '../components/PaginatedResults';
+import {CatalystContext} from '../context';
+import {formatCount} from '../statistics/utilities';
+import loopParity from '../utility/loopParity';
+
+import ReportLayout from './components/ReportLayout';
+import type {ReportDataT} from './types';
+
+type ReportEntryT = {
+  +count: number,
+  +row_number: number,
+  +url: ?UrlT,
+  +url_id: number,
+};
+
+const LinksWithMultipleEntities = ({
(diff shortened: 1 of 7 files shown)


root/report/LinksWithMultipleEntities.js
```diff
@@ -0,0 +1,98 @@
...
+import type {ReportDataT} from './types';
+
+type ReportEntryT = {
+  +count: number,
+  +row_number: number,
+  +url: ?UrlT,
+  +url_id: number,
+};
+
+const WikidataLinksWithMultipleEntities = ({
```
mwiencek, line 27:
```suggestion
const LinksWithMultipleEntities = ({
```

root/report/LinksWithMultipleEntities.js
```diff
@@ -0,0 +1,98 @@
...
+         whether they are really a good match: for example, a generic
+         link to a store page could be replaced with specific
+         per-release store links, and a link attached to all movements
+         of a classical work might be a better fit for the main
+         parent work only.`,
+      )}
+      entityType="url"
+      filtered={filtered}
+      generated={generated}
+      title={l('Wikidata URLs linked to multiple entities')}
```
mwiencek, line 53:
```suggestion
      title={l('URLs linked to multiple entities')}
```

root/report/LinksWithMultipleEntities.js
```diff
@@ -0,0 +1,98 @@
...
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      </PaginatedResults>
+    </ReportLayout>
+  );
+};
+
+export default WikidataLinksWithMultipleEntities;
```
mwiencek, line 98:
```suggestion
export default LinksWithMultipleEntities;
```

mwiencek (changes requested):

mwiencek (approved):

PR #2128, opened 2021-06-05
MBS-11698: More specific target of external link error messages

y-young:
# Problem

MBS-11698: More specific target of external link error messages

# Solution

Add an extra property `target` to the validation result, possible values are none, url, relationship and entity. The default error message "This URL is not allowed for the selected link type, or is incorrectly formatted." is splitted according to error target.

I noticed that we're only performing a rough check on the validity of URL, and since "This URL is incorrectly formatted." states a quite serious issue and would confuse users, we should fallback to "This URL is not allowed for the selected link type." if the URL passed the check or didn't go through any check. But meanwhile, as said by CatQuest in IRC, this message may be annoying sometimes because there's likely no valid relationship type for this combination of URL and entity, so it's a waste of time trying other options.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -297,6 +297,13 @@ const RESTRICTED_LINK_TYPES = [
   return result.concat(Object.values(linkType));
 }, []);
 
+export const ERROR_TARGETS = {
+  NONE: 'none',
+  URL: 'url',
+  RELATIONSHIP: 'relationship',
+  ENTITY: 'entity',
+};
+
 function reencodeMediawikiLocalPart(url) {
   const m = url.match(/^(https?:\/\/[^\/]+\/wiki\/)([^?#]+)(.*)$/);
   if (m) {
@@ -376,14 +383,24 @@ const CLEANUPS = {
         const prefix = m[1];
         switch (id) {
           case LINK_TYPES.otherdatabases.artist:
-            return {result: prefix === 'artist'};
+            return {
+              result: prefix === 'artist',
+              target: ERROR_TARGETS.RELATIONSHIP,
+            };
           case LINK_TYPES.otherdatabases.label:
-            return {result: prefix === 'label'};
+            return {
+              result: prefix === 'label',
+              target: ERROR_TARGETS.RELATIONSHIP,
+            };
           case LINK_TYPES.otherdatabases.release_group:
-            return {result: prefix === 'record'};
+            return {
+              result: prefix === 'record',
+              target: ERROR_TARGETS.RELATIONSHIP,
+            };
         }
+        return {result: false, target: ERROR_TARGETS.RELATIONSHIP};
       }
-      return {result: false};
+      return {result: false, target: ERROR_TARGETS.URL};
     },
   },
   '45worlds': {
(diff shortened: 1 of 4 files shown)


root/static/scripts/edit/externalLinks.js
```diff
@@ -437,9 +458,14 @@ export class ExternalLink extends React.Component<LinkProps> {
...
-            <div className="error field-error" data-visible="1">
-              {props.errorMessage}
-            </div>}
+            <>
+              <div className="error field-error" data-visible="1">
+                {props.errorMessage}
+              </div>
+              <div className="error-target">
+                {`(${props.errorTarget})`}
+              </div>
```
reosarevok, line 467: I'm a bit confused about why we would print this :) Is it just temporary until this is put on top of code that actually has separate error fields for each? :)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1041,12 +1184,19 @@ const CLEANUPS = {
...
-            return {result: prefix === 'people'};
+            return {
+              result: prefix === 'people',
+              target: ERROR_TARGETS.RELATIONSHIP,
+            };
           case LINK_TYPES.otherdatabases.recording:
-            return {result: prefix === 'files'};
+            return {
+              result: prefix === 'files',
+              target: ERROR_TARGETS.RELATIONSHIP,
```
reosarevok, line 1194: There's a ton of `target: ERROR_TARGETS.RELATIONSHIP` additions - would it maybe make sense to have this one as default, and only add a target if it should be a different one?

root/static/scripts/edit/externalLinks.js
```diff
@@ -241,39 +242,55 @@ export class ExternalLinksEditor
...
             } else if (isNewOrChangedLink && checker) {
               const check = checker(link.url);
               if (!check.result) {
-                error = check.error ||
-                  l(`This URL is not allowed for the selected link type, 
-                     or is incorrectly formatted.`);
+                errorTarget = check.target ||
+                  URLCleanup.ERROR_TARGETS.NONE;
+                if (errorTarget === URLCleanup.ERROR_TARGETS.URL) {
+                  error = l(`This URL is incorrectly formatted.`);
```
reosarevok, line 287: As you mentioned in your PR, "incorrectly formatted" can be confusing since it kinda sounds like the user can fix it by editing the formatting, which often won't be the case (otherwise, we would do it ourselves with a cleanup). Not that I'm sure what would be better, though. @mwiencek?

root/static/scripts/edit/externalLinks.js
y-young: Yes, it is just a temporary solution.

root/static/scripts/edit/URLCleanup.js
y-young: Indeed. But @yvanzo said:
> The default should probably be no value though as it might affect either URL or relationship or both.

root/static/scripts/edit/externalLinks.js
yvanzo: It would better fit as a label by displaying it in front of the error message with a colon `:` as separator.

root/static/scripts/edit/externalLinks.js
yvanzo: This is where the default case I mentioned should be used for ambiguous situations.

root/static/scripts/edit/URLCleanup.js
yvanzo: Having a catch-all for default value is unfortunately more verbose but safer than having the most common value.

root/static/scripts/edit/externalLinks.js
mwiencek: Maybe something saying that the domain is valid, but the page it points to isn't?  What other situations do we show this in?

mwiencek:
Reading the ticket, this'll allow us to display errors under either the URL input or the type selector, which makes sense.

But `ERROR_TARGETS.NONE`, `ERROR_TARGETS.ENTITY` -- where will these be displayed?  If there are only two places to display the error, I'm not sure how these help.  Or why, say, an ENTITY error isn't just a URL or RELATIONSHIP error, since the user can't change the entity and it's still a URL issue the context of the page.

root/static/scripts/edit/externalLinks.js
mwiencek: I assume we'll want to merge this so you can base other PRs (adding the separate error fields) on top of it, since basing PRs on top of other PRs is not a good development or code review experience. :)  If so, you'll have to remove this temporary display so that the UI doesn't appear broken or incoherent after we merge it.

y-young:
> Reading the ticket, this'll allow us to display errors under either the URL input or the type selector, which makes sense.
> 
> But `ERROR_TARGETS.NONE`, `ERROR_TARGETS.ENTITY` -- where will these be displayed? If there are only two places to display the error, I'm not sure how these help. Or why, say, an ENTITY error isn't just a URL or RELATIONSHIP error, since the user can't change the entity and it's still a URL issue the context of the page.

You're right. These are just special cases that should be dealt later.

mwiencek:
> These are just special cases that should be dealt later.

Does that mean you plan to remove these, or use them for display somehow?  I'm curious how they will be handled.

y-young:
> > These are just special cases that should be dealt later.
> 
> Does that mean you plan to remove these, or use them for display somehow? I'm curious how they will be handled.

In my opinion I'll keep them and display somewhere, but I need words from @yvanzo .

root/static/scripts/edit/externalLinks.js
yvanzo: This temporary display has been replaced with a div class in commit https://github.com/metabrainz/musicbrainz-server/pull/2128/commits/ad4fcbeb73f1e8bc9be965f222beed21b85761da.

root/static/scripts/edit/externalLinks.js
y-young: This message has been changed back to the old one as for now.

yvanzo (approved):
Thank you. All mentioned issues have been addressed so far. Previously existing `URL` error message should be improved later on.

@mwiencek: For example `ENTITY` can be used to link to the editing form of the most appropriate linked entity with seeded URL.

PR #2129, opened 2021-06-05
MBS-11371: Group members in JSON-LD by date, do not squash

reosarevok:
### Fix MBS-11371

The code used to squash all "member of" rels between a group and a person into one relationship for JSON-LD, whether that made sense or not (for example, it'd happily squash two relationships into "start date 2020, end date 2010").
We talked to Google people and they said they thought having one role entry per date set was better, so this code does that. We still group several relationships from the same date (to display "guitar, piano" together if the same member played both with the same dates).

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
@@ -41,7 +41,18 @@ around serialize => sub {
         if (@members) {
             my %seen_members;
             for my $member (@members) {
-                $seen_members{$member->target->gid} = member_relationship($member, $inc, $stash, $seen_members{$member->target->gid});
+                # We separate different dates, but keep together
+                # different roles with the same dates
+                my $key = $member->target->gid;
+                my $begin_date = $member->link->begin_date;
+                my $end_date = $member->link->end_date;
+                if ($begin_date && $begin_date->defined_run) {
+                    $key .= '-begin-' . join('-', $begin_date->defined_run);
+                }
+                if ($end_date && $end_date->defined_run) {
+                    $key .= '-end-' . join('-', $end_date->defined_run);
+                }
+                $seen_members{$key} = member_relationship($member, $inc, $stash, $seen_members{$key});
             }
             $ret->{member} = [ map { $seen_members{$_} } sort keys %seen_members ];
         }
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
```diff
@@ -41,7 +41,16 @@ around serialize => sub {
...
-                $seen_members{$member->target->gid} = member_relationship($member, $inc, $stash, $seen_members{$member->target->gid});
+                # We separate different dates, but keep together
+                # different roles with the same dates
+                my $key = $member->target->gid;
+                if ($member->link->begin_date && $member->link->begin_date->defined_run) {
+                    $key .= '-' . join('-', $member->link->begin_date->defined_run);
+                }
+                if ($member->link->end_date && $member->link->end_date->defined_run) {
+                    $key .= '-' . join('-', $member->link->end_date->defined_run);
+                }
```
mwiencek, line 52: I'd suggest storing `$begin_date`/`$end_date` to variables to reduce clutter.
```suggestion
                my $begin_date = $member->link->begin_date;
                my $end_date = $member->link->end_date;
                if ($begin_date && $begin_date->defined_run) {
                    $key .= '-' . join('-', $begin_date->defined_run);
                }
                if ($end_date && $end_date->defined_run) {
                    $key .= '-' . join('-', $end_date->defined_run);
                }
```

mwiencek (approved):

lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Artist.pm
```diff
@@ -41,7 +41,18 @@ around serialize => sub {
...
+                # different roles with the same dates
+                my $key = $member->target->gid;
+                my $begin_date = $member->link->begin_date;
+                my $end_date = $member->link->end_date;
+                if ($begin_date && $begin_date->defined_run) {
+                    $key .= '-' . join('-', $begin_date->defined_run);
+                }
+                if ($end_date && $end_date->defined_run) {
+                    $key .= '-' . join('-', $end_date->defined_run);
+                }
```
mwiencek, line 54: if one rel has begin_date: 1999 and another has end_date: 1999 wouldn't they end up with the same key here? maybe we should prefix the key parts with begin/end.

reosarevok: Oh, damn, yes, good point. Will do :)

mwiencek (approved):

PR #2130, opened 2021-06-06
MBS-11682: Fix deadlocks in apply_artist_*_pending_updates

mwiencek:
`TRUNCATE` requires an `ACCESS EXCLUSIVE` lock on the target table. This means it will block on any concurrent operation on the table (which, for the `artist_*_pending_update` tables, means `INSERT` and other `TRUNCATE` statements).

A deadlock may occur as follows:

  1. Txn 1 does `INSERT INTO artist_release_group_pending_update` (e.g. in the `a_del_release` trigger).
  2. Txn 2 does `INSERT INTO artist_release_group_pending_update`.
  3. Txn 2 does `TRUNCATE artist_release_group_pending_update` in the deferred "apply" trigger. This blocks on the `INSERT` from txn 1.
  4. Txn 1 does `TRUNCATE artist_release_group_pending_update` in the deferred "apply" trigger. This blocks on txn 2.
  5. Deadlock.

I'm resolving this by using `DELETE` instead, which I believe works because in this case it only blocks concurrent deletions on the same rows and not concurrent insertions.

### DIFF
--- a/admin/sql/CreateFunctions.sql
+++ b/admin/sql/CreateFunctions.sql
@@ -1822,9 +1822,13 @@ DECLARE
 BEGIN
     -- DO NOT modify any replicated tables in this function; it's used
     -- by a trigger on slaves.
+    WITH pending AS (
+        DELETE FROM artist_release_pending_update
+        RETURNING release
+    )
     SELECT array_agg(DISTINCT release)
     INTO release_ids
-    FROM artist_release_pending_update;
+    FROM pending;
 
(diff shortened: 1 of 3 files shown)


reosarevok (approved):
Sounds sensible.

mwiencek:
This requires very precise timing which I'm not sure how to reproduce in a test. Though I did verify the lock behavior manually with two psql shells.

yvanzo (approved):

mwiencek:
Deployed to production.

PR #2131, opened 2021-06-07
Convert the Release index page to React + add track paging

mwiencek:
Fixes MBS-9921
Fixes MBS-8844
Fixes MBS-11703

In addition to the 10-medium limit we've already had, I've added a 100 track limit, which covers the case of absurdly large mediums. When a medium like that is expanded, the track requests are paged (via a new endpoint, /ws/js/tracks) and can be fetched 100 at a time. I provided a no-script fallback as before.


reosarevok:
One thing I just considered - do we know whether there are unloaded tracks still from the release index? Assuming we do, which I hope so, we should give some warning that there might be more unloaded relationships when viewing rels in "view at bottom".

reosarevok:
This might be expected for now, but just in case:

```
Warning: Text content did not match. Server: "Pealkiri" Client: "Title"
    at th
    at tr
    at tbody
    at table
    at http://localhost:5000/static/build/release/index.js:1416:26
    at http://localhost:5000/static/build/release/index.js:2118:24
```

reosarevok:
Also: on http://localhost:5000/release/b5d024d2-ca9a-46a7-887c-301bce20294f I get "This medium has too many tracks to load at once; currently showing 100 out of 512 total. Näita rohkem..." but clicking the show more / näita rohkem brings a "Laadimine..." / Loading notice up, but then nothing happens.

mwiencek:
> One thing I just considered - do we know whether there are unloaded tracks still from the release index? Assuming we do, which I hope so, we should give some warning that there might be more unloaded relationships when viewing rels in "view at bottom".

Yeah, I added a message for that.  Let me know if it looks okay.

> This might be expected for now, but just in case:

This should be fixed once we regenerate POT files.

> Also: on http://localhost:5000/release/b5d024d2-ca9a-46a7-887c-301bce20294f I get "This medium has too many tracks to load at once; currently showing 100 out of 512 total. Näita rohkem..." but clicking the show more / näita rohkem brings a "Laadimine..." / Loading notice up, but then nothing happens.

Fixed. I broke that somehow in a recent push.

reosarevok:
Fairly minor, but: 

Screenshot from 2021-06-15 14-43-14

This seems to be a lot taller than it needs to be (thanks for adding it, in any case!).

The extra loading seems to not be loading work artists - the first one does though. See: 

Screenshot from 2021-06-15 14-44-48

Also, for: 

Screenshot from 2021-06-15 14-45-07

Would it make sense to have two options, load 100 more / load until all are loaded? With the caveat that the latter takes longer. I can't imagine clicking load 10 times for a 1000 track release, but I'd happily click load once, then go navigate elsewhere while it loads.

reosarevok (changes requested):
Once the work rels bit is working, this seems ready for beta at least. Other comments are fairly minor possible improvements.

reosarevok (approved):
Looked through + tested, seems ok now. Since it's a big change, let's put it in beta as soon as possible so it is there as long as possible :)

jesus2099:
Very nice track paging system!

Hem, it may sound selfish userscript requirement but:

- The _Expand all mediums_ button is like `button#expand-all-mediums`
- Each _Expand medium_ arrow link is like `a.expand-medium`
- But this new _Load all tracks_ link has no obvious CSS selector

I still can use:

```css
table.medium > tbody > tr:not(.subh):not(.even):not(.odd) > td[colspan] > a[href='#']
```

But it feels less safe on the long term.

PR #2132, opened 2021-06-08
MBS-11696: Disallow Twitter documentation links

reosarevok:
### Implement MBS-11696

These three at least have the same format as standard Twitter usernames, but redirect to documentation pages. There might be more, these are the three I found right now.


mwiencek (approved):

yvanzo (approved):
_These three_ are `/privacy`, `/rules` and `/tos` I guess.

Would be nice to mention them in commit message and update ticket accordingly.

PR #2133, opened 2021-06-08
MBS-11293: Add streaming rels to the downloads in non-digital releases report

reosarevok:
### Implement MBS-11293

These are also meant to not exist in physical releases, since a physical release cannot be streamed. They should only exist on digital versions of the releases.

Also fixes a small oops from when I was fixing lowercase SQL.


mwiencek (approved):

mwiencek:
(I'd suggest pushing the s/PN/ON/ thing straight to master.)

yvanzo (approved):
:shipit:

PR #2137, opened 2021-06-08
Revert "When doing a direct database search, also do an exact full text search too"

mwiencek:
This reverts commit 6a49115d488c6de7c11f454981c5b8a101e0b913.

Rationale: Direct search queries frequently exceed `log_min_duration_statement` and spam our PG logs. I was trying to find out why, and discovered that the `OR name = ?` bits of the query cause a much slower plan to be used. Testing against an artist search, on the order of 2200 ms down to sub-300 ms with the condition removed.

Additionally, a condition like:

```
  mb_simple_tsvector(name) @@
    plainto_tsquery('mb_simple', mb_lower('姫咲ゆずる'))
```

still returns exacts matches for '姫咲ゆずる' or any other name I've tried.

The original commit message didn't provide any reasoning for this change, so since it has a very negative impact on performance, I'm reverting it.


yvanzo (approved):
LGTM

PR #2138, opened 2021-06-09
MBS-11697: Show historical primary types on old RG edits

reosarevok:
### Implement MBS-11697

Old release group (and even older release) edits that involved one of the primary types that became secondary types with the NGS change were not showing the types, or even the change at all in some cases.
Luckily, we had a perfectly good map of type ids to names in the seemingly unused Server/Release.pm file, so I could use that to restore the information enough to actually display something useful.
I added the list to Server/Constants.pm since we might want to remove Server/Release.pm anyway, unless I'm missing something.
Just the types not having a gid seemed like a perfectly valid way of distinguishing these "mocked up" old types, so I haven't bothered trying to figure out any other option.

Testing edits, with pink data: edit/11088448 - edit/16294643/ - edit/8137486

### DIFF
--- a/lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm
+++ b/lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm
@@ -5,6 +5,7 @@ use MooseX::Types::Moose qw( ArrayRef Int Str );
 use MooseX::Types::Structured qw( Dict Optional );
 use MusicBrainz::Server::Constants qw( $EDIT_RELEASEGROUP_CREATE );
 use MusicBrainz::Server::Edit::Types qw( Nullable ArtistCreditDefinition );
+use MusicBrainz::Server::Edit::Historic::Utils qw( get_historic_type );
 use MusicBrainz::Server::Edit::Utils qw(
     load_artist_credit_definitions
     artist_credit_preview
@@ -59,7 +60,8 @@ sub build_display_data
         artist_credit => to_json_object(artist_credit_preview($loaded, $self->data->{artist_credit})),
         name          => $self->data->{name} || '',
         comment       => $self->data->{comment} || '',
-        type          => $type ? to_json_object($loaded->{ReleaseGroupType}{$type}) : undef,
+        # Older edits store historic (pre-split) "secondary" types
+        type          => get_historic_type($type, $loaded),
         release_group => to_json_object((defined($self->entity_id) &&
                               $loaded->{ReleaseGroup}{ $self->entity_id }) ||
                                   ReleaseGroup->new( name => $self->data->{name} )),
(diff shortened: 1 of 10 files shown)


mwiencek:
> Just the types not having a gid seemed like a perfectly valid way of distinguishing these "mocked up" old types, so I haven't bothered trying to figure out any other option.

Checks like these make me uneasy, because it means we can never really change or fix the data in the future without breaking hacks like this. Would it be feasible to use the edit ID or open time instead?

mwiencek:
Well, using the date has the same problem or worse, so disregard that.  In my head I was thinking more of a version field like we have for other edit types. Maybe it wouldn't be difficult to be proactive and add something like that into these edits now.

lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm
```diff
@@ -59,7 +60,8 @@ sub build_display_data
         artist_credit => to_json_object(artist_credit_preview($loaded, $self->data->{artist_credit})),
         name          => $self->data->{name} || '',
         comment       => $self->data->{comment} || '',
-        type          => $type ? to_json_object($loaded->{ReleaseGroupType}{$type}) : undef,
+        # Older edits store historic (pre-split) "secondary" types
+        type          => get_historic_type($type, $loaded),
```
mwiencek, line 64: What if we had `get_historic_type` return whether it's a historic type, and add something like `has_historic_type` to the display data?

root/edit/details/historic/EditReleaseAttributes.js
```diff
@@ -18,17 +18,29 @@ type EditReleaseAttributesEditT = {
...
   },
 };
 
 type Props = {
   +edit: EditReleaseAttributesEditT,
 };
 
+function getTypeName(type) {
+  return type
+    // If there's no type.gid then it's a historic "secondary" type
```
mwiencek, line 34: Since this is itself a historic edit type, does this ever happen here?

reosarevok: Yes, for stuff that remained a primary type (Album, Single, EP, Other) this uses $loaded, but for the rest (Compilation, Live, etc) it uses the type mocks.

lib/MusicBrainz/Server/Edit/ReleaseGroup/Create.pm
reosarevok: So, have get_historic_type return ($type, $has_historic_type), and then use that? That could work. What'd be the best way of doing that assignment inside the map at https://github.com/metabrainz/musicbrainz-server/pull/2138/files#diff-50d2fc9c2cb554797610757d6cff4d398976e3022b4aca65265ff2ed6cfe4284R59-R70 without just calling it twice?

mwiencek: Maybe adding a `historic` attribute to `Entity::ReleaseGroupType` would be simpler.

reosarevok: Made that change, it does seem simple.

PR #2139, opened 2021-06-09
MBS-11709: Beta: Edit notes with { show raw HTML

mwiencek:
`expand2react` is intended to expand our i18n syntax, so if it encounters a `{`, it expects it to be the start of a variable substitution. Edit notes are not translated strings; the HTML we're given from the server should be displayed as-is, so we should be using `dangerouslySetInnerHTML` here.

### DIFF
--- a/root/edit/components/EditNote.js
+++ b/root/edit/components/EditNote.js
@@ -16,7 +16,6 @@ import {
 } from '../../constants';
 import {CatalystContext} from '../../context';
 import EditorLink from '../../static/scripts/common/components/EditorLink';
-import expand2react from '../../static/scripts/common/i18n/expand2react';
 import getVoteName from '../../static/scripts/edit/utility/getVoteName';
 import formatUserDate from '../../utility/formatUserDate';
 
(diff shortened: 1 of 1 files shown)


reosarevok (approved):

yvanzo (commented):
Is `dangerouslySetInnerHTML` safe to use with edit note entered by users?
Is using `dangerouslySetInnerHTML` as tolerant to invalid HTML as `edit/notes.tt` was?

Or would it be preferable to just revert pull request #1860 and seek for another solution like having an edit note parser in JS?

mwiencek:
> Is dangerouslySetInnerHTML safe to use with edit note entered by users? Is using dangerouslySetInnerHTML as tolerant to invalid HTML as edit/notes.tt was?

The HTML we're outputting comes from `format_editnote` [1], which encodes any `<` and `>` characters that the user enters. [2]  This would be the exact same HTML that was outputted by the old TT template [3], so there's no actual increased danger.

[1] https://github.com/metabrainz/musicbrainz-server/blob/4028205/lib/MusicBrainz/Server/Entity/EditNote.pm#L130
[2] https://github.com/metabrainz/musicbrainz-server/blob/32de157/lib/MusicBrainz/Server/Filters.pm#L34-L35
[3] https://github.com/reosarevok/musicbrainz-server/blob/d9b2014/root/components/common-macros.tt#L911

yvanzo (commented):
If the note is by ModBot, is it still localized as it used to be?

See https://github.com/reosarevok/musicbrainz-server/blob/d9b2014/root/components/common-macros.tt#L909

yvanzo (approved):
Thanks!

PR #2141, opened 2021-06-09
Wrap all localized_note strings in N_l or N_ln

mwiencek:
We were missing the appropriate no-op functions on some of these, so they were never extracted into the POT files.

The IPI/ISNI ones have to be done outside of the localized_note call because the way the plural string is passed doesn't match the function signature of N_ln.

I've regenerated the POT files so that you can see the changes.


reosarevok (approved):

yvanzo (approved):
Good catch!

PR #2143, opened 2021-06-11
MBS-11711: Show informative error if no relationships in group

reosarevok:
### Fix MBS-11711

Sometimes a link to a specific relationship type group page might become outdated, if all relationships have been moved away from it. Right now, we just detach 400, which seems unfair - the request is not bad, there's just nothing to show.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Load.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Load.pm
@@ -6,6 +6,7 @@ use MusicBrainz::Server::Data::Utils 'model_to_type';
 use MusicBrainz::Server::Validation qw( is_guid is_positive_integer );
 use MusicBrainz::Server::Constants qw( :direction %ENTITIES );
 use Readonly;
+use aliased 'MusicBrainz::Server::Entity::RelationshipLinkTypeGroup';
 
 no if $] >= 5.018, warnings => "experimental::smartmatch";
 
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/Role/Load.pm
```diff
@@ -117,15 +117,15 @@ role
 
                 if (defined $pager) {
                     if (!scalar @$lt_groups) {
-                        $c->detach('/error_400');
-                    }
-                    if (scalar @$lt_groups > 1) {
+                        $c->stash->{paged_link_type_group_is_empty} = 1;
```
mwiencek, line 120: Would it work to just create an empty RelationshipLinkTypeGroup here to stash in paged_link_type_group, then check total_relationships == 0?  That would avoid having to pass a new prop everywhere, at least.

reosarevok: Probably. That was my original plan more or less but I wasn't sure whether it would work.

Ideally we should also have a different error (possibly 400 with a message) if the type id selected doesn't exist.

reosarevok: Updated this a bit, does it seem better? It does certainly seem simpler, anyway.

mwiencek: Thanks!  Looks good now, I think.

PR #2145, opened 2021-06-16
MBS-11722: Don't preselect basic as language proficiency

reosarevok:
### Implement MBS-11722

Users are apparently often selecting "basic" when it seems incorrect, probably because it is selected by default. "Arabic" is also selected by default, and we seem to have 900+ users with "Arabic (basic)". This might be true, but I find it unlikely.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Role/UserProfile.pm
+++ b/lib/MusicBrainz/Server/Form/Role/UserProfile.pm
@@ -3,7 +3,7 @@ package MusicBrainz::Server::Form::Role::UserProfile;
 use HTML::FormHandler::Moose::Role;
 use List::MoreUtils qw( any all );
 use MusicBrainz::Server::Form::Utils qw( language_options select_options_tree validate_username );
-use MusicBrainz::Server::Translation qw( l ln );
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Form/Role/UserProfile.pm
```diff
@@ -53,11 +53,13 @@ has_field 'languages' => (
 
 has_field 'languages.language_id' => (
     type => 'Select',
+    messages => { required => N_l('A language is required') },
     required => 1
 );
 
 has_field 'languages.fluency' => (
     type => 'Select',
+    messages => { required => N_l('A fluency level is required') },
```
mwiencek, line 62: Nitpick, but I'd prefer all error messages ended with periods:
```suggestion
    messages => { required => N_l('A language is required.') },
    required => 1
);

has_field 'languages.fluency' => (
    type => 'Select',
    messages => { required => N_l('A fluency level is required.') },
```

reosarevok: I was copying User::Login, but I actually prefer this too, so changed it for these.

PR #2146, opened 2021-06-16
MBS-11717: Also show edits pending in ArtistCreditUsageLink

reosarevok:
### Fix MBS-11717

This only shows it when requested, since it doesn't make sense everywhere `ArtistCreditUsageLink` is used (such as on every tab of `ArtistCreditLayout`.

@mwiencek: there's an issue here where `Tooltip` is inside the `h1` for the `ArtistCreditLayout` header, meaning the text in the tooltip is also at h1 size and doesn't really work well. This seems like a more widespread issue with `Tooltip`, in that it probably doesn't make sense that it takes on the style of its container. What do you suggest for this? A `tooltip` class that overrides it?

Specifically added to `artist-credit/$id` (on the header) and `report/ArtistCreditsWithDubiousTrailingPhrases` (and any other AC reports)

### DIFF
--- a/root/artist_credit/ArtistCreditLayout.js
+++ b/root/artist_credit/ArtistCreditLayout.js
@@ -59,6 +59,7 @@ const ArtistCreditLayout = ({
             'Artist credit “{artist_credit}”',
             {artist_credit: reduceArtistCredit(artistCredit)},
           )}
+          showEditsPending
         />
       </h1>
       <Tabs>
--- a/root/report/components/ArtistCreditList.js
+++ b/root/report/components/ArtistCreditList.js
@@ -37,7 +37,10 @@ const ArtistCreditList = ({
             {item.artist_credit ? (
               <>
                 <td>
-                  <ArtistCreditUsageLink artistCredit={item.artist_credit} />
+                  <ArtistCreditUsageLink
+                    artistCredit={item.artist_credit}
+                    showEditsPending
+                  />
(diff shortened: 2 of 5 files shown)


yvanzo (commented):
Code changes to `ArtistCreditUsageLink.js`look good to me.
Is the tooltip commit related to these changes?

Please be more specific about pages affected (or not) by this bugfix, in both commit message (for the future you) and in tickets (for editors); That will also help with reviewing it fully.

reosarevok:
> Is the tooltip commit related to these changes?

It's related in that otherwise a tooltip by a header, like the one below, would have the same font size and style as the header:
credit

(I fixed the typo in the translation already, btw)

yvanzo:
> It's related in that otherwise a tooltip by a header, like the one below, would have the same font size and style as the header:

Ok then please mention that example (tooltip for pending edits) in commit message.

> The PR text says that
>
> Specifically added to artist-credit/$id (on the header) and report/ArtistCreditsWithDubiousTrailingPhrases (and any other AC reports)

Thank you but I did read that. Please clarify the commit message.

reosarevok:
Oh, ok, I got the impression you didn't know at all what to test, but you just wanted it to be more documented. Sure, no problem!

yvanzo (approved):
Yes, thank you.

mwiencek (approved):

PR #2149, opened 2021-06-21
Upgrade Flow to 0.155.0

mwiencek:
See associated commits for details

### DIFF
--- a/root/static/scripts/release-editor/edits.js
+++ b/root/static/scripts/release-editor/edits.js
@@ -114,10 +114,10 @@ releaseEditor.edits = {
     var edits = [];
 
     for (let newLabel of newLabels) {
-      const id = newLabel.release_label;
+      const id = getReleaseLabel(newLabel);
 
       if (id) {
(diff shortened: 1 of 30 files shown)


root/static/scripts/release-editor/edits.js
```diff
@@ -114,10 +114,10 @@ releaseEditor.edits = {
     var edits = [];
 
     for (let newLabel of newLabels) {
-      const id = newLabel.release_label;
+      const id = getReleaseLabel(newLabel);
```
mwiencek, line 117: One thing to note with `Map` is that `1` (the number) and `"1"` (the string) are two different keys.  With regular objects these would be the same key.  So I'm using the same function that was passed to `keyBy` here to get the (string) id for lookup.

reosarevok:

reosarevok (commented):
This seems generally sensible, but selenium keeps failing, please check it :) Also check if 0.154 is easy enough to update to, maybe? :)

PR #2150, opened 2021-06-21
MBS-11737: Return different error if IA metadata is empty

mwiencek:
Recently the "we don't own the associated item" message has been triggering for users when it's not true. This error can erroneously occur if the IA metadata API returns an empty response, which just means we need to wait for it to populate.  We can return a different message in this case asking the user to retry later, rather than the current, incorrect message.

While I was trying to debug this, I added some logging to Sentry which I've kept in case it's useful in the future.

### DIFF
--- a/lib/MusicBrainz/Errors.pm
+++ b/lib/MusicBrainz/Errors.pm
@@ -16,6 +16,7 @@ our @EXPORT_OK = qw(
     capture_exceptions
     get_error_message
     send_error_to_sentry
+    send_message_to_sentry
     sentry_enabled
     sig_die_handler
 );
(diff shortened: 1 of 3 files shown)


mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

yvanzo:
@brainzbot retest this please

yvanzo:
@brainzbot retest this please

yvanzo (approved):
Could not test against the CAA, but the rest looks good to me.

Selenium failures are unrelated: timeout on either log out or guess case.

mwiencek:
@brainzbot test this please

PR #2151, opened 2021-06-25
MBS-11391: Show changes made to external link when editing URL relationship

y-young:
# Problem

MBS-11391: Show changes made to external link when editing URL relationship

# Solution

Implemented a popover for URL editing, which contains two fields - raw URL and clean URL.

Links aren't added to list automatically, users will have to press enter or let the input box blurred.

### DIFF
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -8,11 +8,12 @@
  */
 
 import * as React from 'react';
+import type {LinkStateT} from '../externalLinks';
 import ButtonPopover from '../../common/components/ButtonPopover';
 
 type PropsT = {
-  errorMessage: React.Node,
-  onCancel: (number, LinkStateT) => void,
+  errorMessage: string,
+  onCancel: ($Shape<LinkStateT>) => void,
   onChange: (number, SyntheticEvent<HTMLInputElement>) => void,
   rawUrl: string,
   url: string,
@@ -60,7 +61,7 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
             </td>
             <td>
               <input
-                className="value"
+                className="value raw-url"
                 onChange={props.onChange}
                 style={{width: '336px'}}
                 type="url"
@@ -81,7 +82,7 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
               {l('Cleaned up to:')}
             </td>
             <td>
-              <a href={props.url}>{props.url}</a>
+              <a className="clean-url" href={props.url}>{props.url}</a>
             </td>
           </tr>
         </tbody>
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -60,7 +60,6 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
                 className="value raw-url"
                 onChange={props.onChange}
                 style={{width: '336px'}}
-                type="url"
                 value={props.rawUrl}
               />
               {props.errorMessage &&
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -56,7 +56,7 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
         <tbody>
           <tr>
             <td className="section">
-              {l('Raw URL:')}
+              {l('URL:')}
             </td>
             <td>
               <input
@@ -77,8 +77,8 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
             </td>
           </tr>
           <tr>
-            <td className="section">
-              {l('Clean URL:')}
+            <td className="section" style={{whiteSpace: 'nowrap'}}>
+              {l('Cleaned up to:')}
             </td>
             <td>
               <a href={props.url}>{props.url}</a>
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -1,5 +1,5 @@
 /*
- * @flow strict-local
+ * @flow
  * Copyright (C) 2020 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
@@ -9,9 +9,12 @@
 
 import * as React from 'react';
 import ButtonPopover from '../../common/components/ButtonPopover';
+import type {ErrorTarget} from '../externalLinks';
+import {ERROR_TARGETS} from '../URLCleanup';
 
 type PropsT = {
   errorMessage: string,
+  errorTarget: ErrorTarget,
   onCancel: () => void,
   onChange: (SyntheticEvent<HTMLInputElement>) => void,
   onToggle: (boolean) => void,
@@ -66,6 +69,7 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
                 value={props.rawUrl}
               />
               {props.errorMessage &&
+                props.errorTarget === ERROR_TARGETS.URL &&
                 <div
                   className="error field-error target-url"
                   data-visible="1"
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -13,7 +13,7 @@ import ButtonPopover from '../../common/components/ButtonPopover';
 type PropsT = {
   errorMessage: string,
   onCancel: () => void,
-  onChange: (number, SyntheticEvent<HTMLInputElement>) => void,
+  onChange: (SyntheticEvent<HTMLInputElement>) => void,
   onToggle: (boolean) => void,
   rawUrl: string,
   url: string,
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -57,7 +57,7 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
         <tbody>
           <tr>
             <td className="section">
-              {l('URL:')}
+              {addColonText(l('URL'))}
             </td>
             <td>
               <input
@@ -79,7 +79,7 @@ const URLInputPopover = (props: PropsT): React.MixedElement => {
           </tr>
           <tr>
             <td className="section" style={{whiteSpace: 'nowrap'}}>
-              {l('Cleaned up to:')}
+              {addColonText(l('Cleaned up to'))}
             </td>
             <td>
               <a className="clean-url" href={props.url}>{props.url}</a>
--- a/root/static/scripts/edit/components/URLInputPopover.js
+++ b/root/static/scripts/edit/components/URLInputPopover.js
@@ -1,0 +1,126 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2020 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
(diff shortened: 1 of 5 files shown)


reosarevok:
On a quick test: one thing the previous system allowed, but this one does not, is still entering more links even if one of them has an error. You still need to eventually deal with the erroring link before submitting, of course. Is there a reason this is no longer allowed? 

Also: if I blank the URL from one of the existing relationships, and then I close the popover, I get the following: 

Screenshot from 2021-06-30 18-11-15

Maybe it should bring the text field back in that case, as it does for the new link at the bottom?

y-young:
Part of the reason is I'm only showing input box on the last link, another reason might be a unified appearance, to display all links  in list as texts (IIRC this wasn't stated explicitly in my discussion with yvanzo though, I need to check the backlogs). I have also noticed the blank URL issue, but if we bring back the input box, there would be 2 ways to edit a link, so I just keep it as it is at least for now. 

> one thing the previous system allowed, but this one does not, is still entering more links even if one of them has an error.

Concerning the unified appearance mentioned above, if this is allowed then it should be displayed as text, but meanwhile editing a link in the list requires more effort than before (you'll have to open the popover first, edit and save), so we prevent submitting if there's an error. In fact I've never thought of a situation like that, it should probably be better handled.

reosarevok:
My idea was to show the text field again only if the URL was blanked (since then there's nothing to edit in the popover anyway), not in every case :)

root/static/scripts/edit/components/URLInputPopover.js
```diff
@@ -0,0 +1,127 @@
...
+      onSubmit={(event) => {
+        event.preventDefault();
+        setIsOpen(false);
+      }}
+    >
+      <table>
+        <tbody>
+          <tr>
+            <td className="section">
+              {l('URL:')}
```
yvanzo, line 60:
```suggestion
              {addColonText(l('URL'))}
```

root/static/scripts/edit/components/URLInputPopover.js
```diff
@@ -0,0 +1,127 @@
...
+                  data-visible="1"
+                >
+                  {props.errorMessage}
+                </div>
+              }
+            </td>
+          </tr>
+          <tr>
+            <td className="section" style={{whiteSpace: 'nowrap'}}>
+              {l('Cleaned up to:')}
```
yvanzo, line 82:
```suggestion
              {addColonText(l('Cleaned up to'))}
```

yvanzo (commented):
`addColonText` can be used for every label that ends with a colon.

yvanzo (approved):
Thank you.

(PR should be squashed on merge)

yvanzo (changes requested):
Sorry I did not get the steps to reproduce the regression @reosarevok mentioned above correctly. This has to be fixed of course. URLs with errors should use the same components (link with pencil) as URLs with no error.

t/selenium/External_Links_Editor.json5
```diff
@@ -414,14 +414,9 @@
...
-      target: "xpath=(//table[@id='external-links-editor']//input[@type='url'])[1]",
-      value: '${KEY_HOME}${KEY_SHIFT}${KEY_END}${KEY_SHIFT}${KEY_BKSP}',
-    },
-    {
-      command: 'select',
-      target: "xpath=//table[@id='external-links-editor']//tr[2]//select[contains(@class, 'link-type')]",
-      value: 'label=regexp:\\B',
+      command: 'click',
+      target: "xpath=//table[@id='external-links-editor']//tr[2]//button[contains(@class, 'remove-item')]",
+      value: '',
```
yvanzo: Does it still “duplicate the first URL”?

y-young: I think so, please see the latest total diff. This comment is for the next action to edit an exising link to duplicate the deleted first URL.

yvanzo: Oh ok, you restored it, thanks.

yvanzo (approved):

root/static/scripts/edit/components/URLInputPopover.js
```diff
@@ -0,0 +1,127 @@
...
+            </td>
+          </tr>
+          <tr>
+            <td className="section" style={{whiteSpace: 'nowrap'}}>
+              {addColonText(l('Cleaned up to'))}
+            </td>
+            <td>
+              <a className="clean-url" href={props.url}>{props.url}</a>
+            </td>
+          </tr>
```
mwiencek, line 87: Maybe we should hide this row if there's no URL entered or no clean URL?  It looks a bit strange as "Cleaned up to: \<nothing\>".

mwiencek (commented):
Something small I noticed: if I edit a URL and click cancel, it reverts back to the old URL but keeps whatever new type it selected.  Not sure if that's intended.  It might be less visually distracting to avoid updating the list at all until the user hits "Done."

I'm also wondering about the `[?]` icon:

Screenshot_20210708_132012

It looks a bit cramped and out of places when misaligned.  Also, one of the purposes of it besides showing the link type documentation was to present a clickable URL.  If we're now always showing a clickable URL in the list, we can probably remove it from the tooltip.  (IIRC this should fix a bug where long URLs clip outside the tooltip, unless we fixed that already.)  And maybe then we should move the icon to be next to the type selector since it's only for type documentation.

y-young:
@mwiencek Nice point!
Link type should be reset when cancelled now. The list is keep unchanged while editing. The icons are aligned to the right and the URL inside type description popover is removed.

mwiencek (commented):
Thanks!  Thinking about my previous review again: if you enter a URL for a different type (i.e. that changes the type), we don't display the new type anywhere in the popover, so it might make sense to do that too.

root/static/scripts/edit/components/URLInputPopover.js
```diff
@@ -0,0 +1,127 @@
...
+      <table>
+        <tbody>
+          <tr>
+            <td className="section">
+              {addColonText(l('URL'))}
+            </td>
+            <td>
+              <input
+                className="value raw-url"
+                onChange={props.onChange}
```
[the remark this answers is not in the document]
y-young: You're right, but the point here is to get instant error message while editing. The error is determined in ExternalLinkEditor and passed into ExternalLink, if we want to get error message directly in the popover, it would require a large refactoring.

y-young:
> Thanks! Thinking about my previous review again: if you enter a URL for a different type (i.e. that changes the type), we don't display the new type anywhere in the popover, so it might make sense to do that too.

So do you mean we display a label in the popover when the type is changed, or

> Thanks! Thinking about my previous review again: if you enter a URL for a different type (i.e. that changes the type), we don't display the new type anywhere in the popover, so it might make sense to do that too.

So do you mean displaying a label of link type in the popover, or put a select box?

yvanzo:
> > Thanks! Thinking about my previous review again: if you enter a URL for a different type (i.e. that changes the type), we don't display the new type anywhere in the popover, so it might make sense to do that too.
> 
> So do you mean displaying a label of link type in the popover, or put a select box?

Displaying link type would not be satisfactory because it would possibly display erroneous link type with no way to fix it.
I don’t think that @mwiencek suggested to add a select box either, that would not fit the ongoing project either.

If I understood correctly, the issue here is that relationship type gets changed automatically without the editor noticing that change. However the popover doesn’t hide the type selector, so it seems it would not be worse than the current UI at least. Showing changes to relationship types is certainly a must-have though. @mwiencek: Would it be alright to put one step after the other and look into that additional improvement later on?

mwiencek:
> If I understood correctly, the issue here is that relationship type gets changed automatically without the editor noticing that change. However the popover doesn’t hide the type selector, so it seems it would not be worse than the current UI at least.

Well, the difference is that the type selector currently updates as soon as you paste the URL, whereas here it doesn't update until you hit Done.  If there's an error due to the link type (perhaps it's not valid for the current entity type), you can't see what type is causing the error.

> Displaying link type would not be satisfactory because it would possibly display erroneous link type with no way to fix it.

I guess you'd fix it by changing the URL or hitting cancel.  But it'd be hard to know what the issue is without seeing what type it wants to use, right?  (You can still ignore the error, hit "Done," and see/change the type then, but that seems like an awkward flow.)

> @mwiencek: Would it be alright to put one step after the other and look into that additional improvement later on?

I think @reosarevok should recheck this too but I'm okay with incremental improvements.  Perhaps we can create a feature branch that @yyoung can merge into?  There's some good ideas in https://community.metabrainz.org/t/a-discussion-on-a-new-ui-of-external-links-editor/540324/16 which I think should be investigated.

yvanzo (changes requested):
> I guess you'd fix it by changing the URL or hitting cancel. But it'd be hard to know what the issue is without seeing what type it wants to use, right? (You can still ignore the error, hit "Done," and see/change the type then, but that seems like an awkward flow.)

I just realize now that errors related to relationship type or entity type would be displayed in the popover too. These are the most frequent errors and cannot be fixed by editing the URL, it requires to select a (different) relationship type or to link to another entity.

@y-young: The popover should probably make use of `ERROR_TARGETS` to show `URL` errors only.

yvanzo (approved):
Thank you.

reosarevok:
One thing I noticed: it seems like closing the dialog does cancel (does not apply the changes), but most popups and whatnot in MB apply changes on closing (see for example the "edit artist credit" popup, where if you change something and then click on something else, it will apply).

Another issue: with the current implementation, if I take a Wikidata link, it autoselects Wikidata and doesn't let me change it. If I change it to something like "https://www.wikidata.org/wtfiki/Q552377", it lets me select any type (and checks it for errors accordingly), but if I change it back to https://www.wikidata.org/wiki/Q552377 then it autoselects Wikidata again and locks it again. Your code does the first and second just right, but once the URL is corrected again, the previously selected type stays selected and it does *not* automatically lock the relationship to Wikidata again :)

reosarevok (approved):
... I must have hallucinated the autoselection thing, because on re-test it doesn't seem to happen. So, this seems good now :)

mwiencek:
A couple other small things...

* If I enter a URL like "www.45worlds.com/classical/soloist/yehudi-menuhin" in the popup and hit enter, Firefox doesn't accept the input, saying "Please enter a URL."  I guess it doesn't validate as a URL due to the lack of scheme.  So I have to manually click "Done," which is annoying if you edit with the keyboard.

  Since the raw URL input allows "unclean" entries which may not validate by the browser as a URL, I guess we shouldn't be using `type="url"` for it.

* If a paste a URL in the "Add another link" input, the edit/remove icons don't appear until I blur the field.  I find this odd, but the bigger issue is that it breaks the tab key for keyboard users (try pasting something and pressing tab; the current focus is lost, whereas it should go to the next button).

y-young:
@brainzbot retest this please

y-young:
@brainzbot retest this please

root/static/scripts/edit/components/URLInputPopover.js
```diff
@@ -0,0 +1,125 @@
...
+              }
+            </td>
+          </tr>
+          {props.url &&
+            <tr>
+              <td className="section" style={{whiteSpace: 'nowrap'}}>
+                {addColonText(l('Cleaned up to'))}
+              </td>
+              <td>
+                <a className="clean-url" href={props.url}>{props.url}</a>
```
mwiencek, line 82: (This should have target="_blank" so people don't lose their editing progress by clicking on it.)
```suggestion
                <a className="clean-url" href={props.url} target="_blank">{props.url}</a>
```

mwiencek (approved):
Thanks, this looks good now except for the comment below.  I'm approving but please do fix that before merging. :)

yvanzo:
> IIRC this should fix a bug where long URLs clip outside the tooltip, unless we fixed that already.

Now tracked as MBS-11795.

PR #2155, opened 2021-06-29
MBS-11753: Fix selectors for RangeSelect in ArtistEdit

reosarevok:
### Fix MBS-11753

It seems this broke during the conversion to React (MBS-11195), since the selectors changed. Tested, and this change seems to work both for collapsible and non-collapsible editor sections.


reosarevok:

PR #2156, opened 2021-07-01
MBS-11756: Collapse artist roles when there are too many

reosarevok:
### Implement MBS-11756

This applies to artists for events and to writers for works.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2024

For a works example, see artist/164f0d73-1234-4e2c-8743-d77bf2191051/works and for an events example, area/525d4e18-3d00-31b9-a58b-a146a916de8f/events

### DIFF
--- a/root/search/components/EventResults.js
+++ b/root/search/components/EventResults.js
@@ -9,6 +9,7 @@
 
 import * as React from 'react';
 
+import * as manifest from '../../static/manifest';
 import ArtistRoles from '../../static/scripts/common/components/ArtistRoles';
 import EntityLink from '../../static/scripts/common/components/EntityLink';
 import EventLocations
@@ -41,6 +42,10 @@ function buildResult(result, index) {
       </td>
       <td>
         <ArtistRoles relations={event.performers} />
+        {manifest.js(
+          'common/components/ArtistRoles',
+          {async: 'async'},
+        )}
       </td>
       <td>
         <EventLocations event={event} />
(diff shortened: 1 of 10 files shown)


root/static/scripts/common/components/ArtistRoles.js
```diff
@@ -9,30 +9,55 @@
...
-      </li>
-    ))}
-  </ul>
+type ArtistRolesProps = {
+  +relations: $ReadOnlyArray<RelationT>,
+};
+
+const buildArtistRoleRow = (relation: RelationT) => {
+  return (
+    <li key={relation.entity.id + '-' + relation.credit}>
```
reosarevok, line 31: Credit needed to be added to the key to avoid clashes when the same artist appeared under two different credits.

yvanzo (approved):
Same as for the pull request it depends on.

@reosarevok: I noticed that event’s artists are not sorted (neither by role nor by name). See the example given in the ticket: `collection/0fca972f-9211-4f20-970f-c74bbfe716ed` which currently contains `event/8074ce30-d107-402d-a8d2-a6305aff3ab1`. Is that a reported issue?

reosarevok:
There's MBS-8063 (which assumes they're currently sorted by name, which they might be on event pages?). It would be good to have a consistent sort I guess.

root/search/components/EventResults.js
```diff
@@ -41,6 +42,10 @@ function buildResult(result, index) {
       </td>
       <td>
         <ArtistRoles relations={event.performers} />
+        {manifest.js(
+          'common/components/ArtistRoles',
+          {async: 'async'},
+        )}
```
mwiencek, line 48: These should only get included once per page (so probably output it with the table instead of every cell).

mwiencek (approved):

PR #2159, opened 2021-07-02
MBS-11760: Add missing delete_unused_tag triggers

reosarevok:
### Fix MBS-11760

For some reason, we did not have delete_unused_tag triggers for every taggable entity type. This was causing tags deleted from the missing entity types to remain in the DB with 0 uses.

Tested that after adding the trigger for event, removing a tag from an event does remove it if that's the last use.


t/pgtap/unused-tags/unused-area-tags-deleted.sql
```diff
@@ -0,0 +1,49 @@
...
+SELECT set_eq(
+  'SELECT id FROM tag', '{4, 5}'::INT[],
+  'Tag deleted after commit by AFTER INSERT ON tag'
+);
+
+-- Delete tag after commit to test AFTER DELETE ON area_tag trigger
+DELETE FROM area_tag WHERE tag = 5;
+
+-- Simulate the commit
+SET CONSTRAINTS ALL IMMEDIATE;
```
mwiencek, line 41: I'm not sure this is needed, since the previous instance of this command already set the behavior for the rest of the transaction (that constraints will trigger after every statement).

reosarevok: Oh, duh. I should probably have googled what that exactly does, huh. Thanks.

mwiencek:
Thanks!  Do you want to write a small script that cleans up existing ones, or should we do that in a later PR?

reosarevok:
On it.

reosarevok:
I looked into the script, and it turns out that we have other issues somewhere too: some tags are on entity_tag_raw but not on entity_tag. I think these triggers should still be fine to add though?

mwiencek:
Created the triggers on prod.

PR #2160, opened 2021-07-05
MBS-11764: Make user tags sortable by count

reosarevok:
### Implement MBS-11764

I don't see a benefit of reverse alphabetical sorting, so this only adds sorting on counts (both directions). The button is kept inline for cases where we only need one of the two (currently, that's always the votes selector, but this should also work if we want only the other one).

We currently do the sort in Perl - I considered moving all sorting to the DB level with ORDER BY but I wasn't sure if that would be an improvement, so I added these to Perl too.

Tested at user/reosarevok/tags - it only applies to user/username/tags since that's (AFAICT) the only page where we list counts like this.


root/user/components/UserTagFilters.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const InlineSubmitButton = (
+  <>
+    {' '}
+    <button type="submit">
+      {l('Submit')}
+    </button>
+  </>
+);
```
mwiencek, line 27: `l('Submit')` won't work here, so better to either declare this inside `UserTagFilters` or make it an actual component.

(Not sure if we have an eslint rule for this, but if we keep it declared as an inline element, I think `inline` should start with a lowercase letter since it's neither a class or component.)

reosarevok: I changed it to a component and checked it does work in French.

mwiencek (approved):

PR #2165, opened 2021-07-06
MBS-11775: Do not break if relatedArtists is undefined

reosarevok:
### Fix MBS-11775

I could add relatedArtists to ReleaseGroup in entity.js, but none of the artist relationship types seem relevant at all to a featured artist. Given we already have problems with this being overzealous elsewhere (MBS-10477), it seems sensible to be conservative here.


mwiencek:
Would it be difficult to add a small test for this to root/static/scripts/tests/guessFeat.js? :)

mwiencek:
@brainzbot test this please

PR #2167, opened 2021-07-07
MBS-11778: Convert edit header to React

reosarevok:
### Implement MBS-11778

Tested with all sorts of edit( list)s but do test more - remember to also check /user/votes on top of edit pages, edit searches, entity edit lists.


root/edit/index.tt
```diff
@@ -1,8 +1,11 @@
 [%- WRAPPER 'layout.tt' title=l('Edit #{id}', { id => edit.id }) -%]
     <div id="content">
 
-        [% INCLUDE 'edit/edit_header.tt' %]
         [%- edit_json_obj = React.to_json_object(edit) -%]
+        [%- edit_json_obj_no_data = edit_json_obj -%]
```
mwiencek, line 5: Not sure this works (i.e. makes a copy) actually -- I'm getting `TypeError: Cannot read property 'relationship' of undefined` on /edit/$id pages.

Unless I'm missing some other obvious way, we could probably use http://template-toolkit.org/docs/manual/VMethods.html#section_import by creating an empty object and "importing" the other object into it.

root/utility/edit.js
```diff
@@ -45,6 +45,13 @@ const STATUS_NAMES = {
   [EDIT_STATUS_OPEN]:           N_l('Open'),
 };
 
+export function getEditCssClassName(edit: EditT): string {
```
mwiencek, line 48: I *think* `kebabCase` from root/static/scripts/common/utility/strings.js does the same thing.

mwiencek (approved):

PR #2168, opened 2021-07-08
MBS-11767: Show track-level artists on loaded mediums

mwiencek:
It's intended that we don't store or modify the medium itself in state, as this would lead to a lot of unnecessary updates due to how `React.memo` works.  We try to keep object references the same at all costs and only store the bits of information that actually change in state.

We weren't following this correctly with the `mediumHasMultipleArtists` call.  It was memoized based on the medium, but that reference doesn't change.  The tracks do, so modify the function to accept tracks and memoize based on that reference instead.


reosarevok (approved):
Looks sensible, seems to work.

PR #2170, opened 2021-07-15
MBS-11786: Add a text view of statistics events 

reosarevok:
### Implement MBS-11786

These are the closest we have to a written timeline of MusicBrainz history, so it makes sense to display them in a way that allows users to follow the evolution of MusicBrainz without having to hover all around the timeline graph.

Sample of the current display, for which improvement suggestions are encouraged: 
Screenshot from 2021-07-15 20-12-22


yvanzo (commented):
Would it be preferable to show this list of events under the graph at the page `/statistics/timeline`(`/main`) as originally suggested in the ticket instead of a new page?

reosarevok:
Dunno. I really like the idea of having an actual history page for these rather than hiding them under the timeline where people are less likely to find them :) But I'm open to changing it if needed.

yvanzo:
This is what the ticket suggested at least. It could also match the selected period of time in the graph.

root/statistics/MusicBrainzHistory.js
```diff
@@ -0,0 +1,51 @@
...
+                {event.link ? (
+                  <a href={event.link}>{title}</a>
+                ) : title}
+              </h2>
+              <p>{expand2react(event.description)}</p>
+              <hr />
+            </div>
+          );
+        }) : (
+        l('It seems we have no history to show at all!')
```
mwiencek, line 45: Would it make sense to leave this untranslatable, since it'll probably only get shown on dev servers?
```suggestion
        'It seems we have no history to show at all!'
```

root/statistics/MusicBrainzHistory.js
```diff
@@ -0,0 +1,51 @@
...
+import expand2react from '../static/scripts/common/i18n/expand2react';
+
+type PropsT = {
+  +events: Array<StatisticsEventT>,
+};
+
+const MusicBrainzHistory = ({
+  events,
+}: PropsT): React.Element<typeof Layout> => {
+  const sortedEvents = events.reverse();
```
mwiencek, line 22: Can the Perl send them in the correct order?

mwiencek (commented):
This looks good to me if CatQuest is okay with the implementation (since they added the ticket).

mwiencek:
> This looks good to me if CatQuest is okay with the implementation (since they added the ticket).

Oh, I guess it was originally requested in a comment by @jesus2099.

jesus2099:
It seems great! :)

yvanzo:
I always found the MusicBrainz events difficult to point at and to read from the timeline, so this “text view” addresses it.

However, “timeline” and “history” are quasi-synonyms, and they would actually display the same events, just in a different way. Does that look like a rationale to have both presentations on the same page? Should that be done in a later PR?

jesus2099:
Yes would be good to have both.

reosarevok:
I personally find having a dedicated page for this a lot better than having them in the (already crowded) timeline page, FWIW :)

root/statistics/MusicBrainzHistory.js
reosarevok: Sure.

root/statistics/MusicBrainzHistory.js
reosarevok: Maybe? I mean, I don't know if we want our files to be usable by forks or whatnot, but if not this might be fine. Of course, I guess this would also be shown if we have an error loading the content, but in that case an English error might not be weird.

reosarevok:

root/statistics/MusicBrainzHistory.js
```diff
@@ -0,0 +1,47 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import Layout from '../layout/index.js';
+import expand2react from '../static/scripts/common/i18n/expand2react.js';
+
+type PropsT = {
+  +events: Array<StatisticsEventT>,
```
mwiencek, line 16:
```suggestion
  +events: $ReadOnlyArray<StatisticsEventT>,
```

root/statistics/MusicBrainzHistory.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const MusicBrainzHistory = ({
+  events,
+}: PropsT): React.Element<typeof Layout> => (
+  <Layout fullWidth title={l('History')}>
+    <h1>{l('Our Glorious History')}</h1>
+    {events ? events
```
mwiencek, line 24: It doesn't look like `events` can be falsey here, judging by the Perl (and existing Flow type). So I think this can lose the ternary.

mwiencek (approved):
> I always found the MusicBrainz events difficult to point at and to read from the timeline, so this “text view” addresses it.
>
> However, “timeline” and “history” are quasi-synonyms, and they would actually display the same events, just in a different way. Does that look like a rationale to have both presentations on the same page? Should that be done in a later PR?

I'm not opposed to showing a text view there too (but perhaps truncated to the N most recent ones, with a link to this full page to view older ones?). But I think this PR is also nice enough to be merged for now.

root/statistics/MusicBrainzHistory.js
reosarevok: Rather than losing it, it should probably be checking events.length. Will amend.

PR #2173, opened 2021-07-16
MBS-6140: Allow (recording|work)-level-rels for release browse

reosarevok:
### Implement MBS-6140

I can see no reason why we would prefer users to make one extra call per release rather than requesting this via browse when they need it. Recording already supports work-level-rels in browse, too, so this adds consistency.

Added tests too.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Controller/WS/2/BrowseRelease.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/WS/2/BrowseRelease.pm
@@ -382,6 +382,102 @@ ws_test 'browse releases via recording',
     </release-list>
 </metadata>';
 
+ws_test 'browse releases via recording, with recording and work rels',
+    '/release?inc=recordings+artist-rels+work-rels+recording-level-rels+work-level-rels&status=official&recording=c43ee188-0049-4eec-ba2e-0385c5edd2db' =>
+    '<?xml version="1.0" ?>
+<metadata xmlns="http://musicbrainz.org/ns/mmd-2.0#">
+    <release-list count="1">
+        <release id="ec0d0122-b559-4aa1-a017-7068814aae57">
+            <title>Soup</title>
+            <status id="4e304316-386d-3409-af2e-78857eec5cfe">Official</status>
+            <quality>normal</quality>
+            <text-representation>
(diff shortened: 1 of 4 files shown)


reosarevok:

mwiencek:
This should be fine since it still hits the `$max_recording_relationships` limit.  Can you add a test for `work-level-rels` too though?

t/lib/t/MusicBrainz/Server/Controller/WS/2/BrowseRelease.pm
```diff
@@ -382,6 +382,79 @@ ws_test 'browse releases via recording',
...
+            </text-representation>
+            <barcode>0208311348266</barcode>
+            <cover-art-archive>
+            <artwork>false</artwork>
+            <count>0</count>
+            <front>false</front>
+            <back>false</back>
+            </cover-art-archive>
+            <medium-list count="1">
+            <medium>
```
mwiencek, line 406: nitpick, but some of the indentation here is off (text-representation, cover-art-archive, medium-list, pregap, etc.)

reosarevok:
Added the test

yvanzo (changes requested):
The code looks good, but the commit message doesn’t:

> I can see no reason why we would prefer users to make one extra call per release rather than requesting this via browse when they need it. Recording already supports work-level-rels

The reasons seem to be (have been) that some releases have so many tracks that it could cause load issues, and that extra calls are rate-limited. This can be overcome since the above-mentioned `$max_recording_relationships` is used. On the other hand, it will return 500 recordings at most.

> Recording already supports work-level-rels in browse, too, so this adds consistency.

Recording is the exception so far, this does not add consistency at all.

reosarevok:
> Recording is the exception so far, this does not add consistency at all.

Recording is the only other place where we have the -level-rels implementation, and it supports it on both lookup and browse. Release only supports it on lookup. As such, yes, I'd argue it adds consistency. But I can remove that from the commit message if it's a problem somehow.

reosarevok:
@yvanzo: Updated the commit message, do let me know if that looks better.

yvanzo (approved):
Yes, thank you!

PR #2176, opened 2021-07-20
MBS-11796: Add Internet Archive logo for sidebar

reosarevok:
### Implement MBS-11796

For some reason, we were still using "no-favicon" for the IA. It seems sensible to use this for archive.org links. It might sometimes also show for Wayback Machine links in addition to direct links (given that AFAICT this just picks the last match it finds in the URL, I guess it will do that for any Wayback Machine links that do not already match another favicon). I don't see that as a negative at all though.


abridgedpause:
This wasn't suggested with regards to Wayback Machine links on MusicBrainz but rather for Releases or Recordings that have been added to the Internet Archive for preservation. These are allowed on MusicBrainz as they count as another place where the item can be streamed. For example, this release, which is available for streaming on Internet Archive, shows up in the right-hand sidebar https://musicbrainz.org/release/a3d8a0ab-537d-425e-94ee-59248c1e4e66

PR #2183, opened 2021-07-23
MBS-11804: Allow filtering edit search by time of edit note

reosarevok:
### Implement MBS-11804

I was hoping to reuse `transform_user_input` and `valid` from `MusicBrainz::Server::EditSearch::Predicate::Date`, but it didn't work, so for now I reused the code. Would be thankful for a hint of what I've missed, because I assume it must be possible.

This does not filter by *last* note, just by any note. I expect filtering by last edit note would be even slower since I'd need to make even more checks to see if it's the last - given we want this already for both this and the sort added in MBS-11443, wonder if we should have a materialized view that stores (edit, last_edit_note_date) and gets updated with triggers on note entry?


mwiencek:
> I was hoping to reuse transform_user_input and valid from MusicBrainz::Server::EditSearch::Predicate::Date, but it didn't work, so for now I reused the code. Would be thankful for a hint of what I've missed, because I assume it must be possible.

I guess putting it in a role would be easiest.  But if this is the only thing that's duplicated between them, I probably wouldn't bother since the function is nearly a one-liner. :)

reosarevok:
Rebased this and tested a bit more with a ssh tunnel, it seems quite a bit faster now than I remembered it. Maybe the newer PSQL versions have really helped. Anyway, I think it's probably fine to merge for now (if it looks good) and it can always be improved with a materialized view if we want later on.

PR #2184, opened 2021-07-24
MBS-9902: Support auto-select & validation of multiple relationships for external links

y-young:
# Problem
MBS-9902: Support auto-select/cleanup/validation of more than one relationship type for external links

# Solution
Changed `cleanup.type` to `cleanup.types: []`, which may contain items like `{release: '9896ecd0-6d29-482d-a21e-bd5d1b5e3425'}` and `{release: ['9896ecd0-6d29-482d-a21e-bd5d1b5e3425', '98e08c20-8402-4163-8970-53504bb6a1e4']}`, modified `validationRules` and `guessType` to support validation of multiple possible types / type combinations, added new logic to append new links in `handleUrlChange` to implement auto-selection.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1,13 +1,26 @@
...
  *
  * This file is part of MusicBrainz, the open internet music database,
  * and is licensed under the GPL version 2, or (at your option) any
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+type EntityType =
+ | 'area' | 'artist' | 'event' | 'instrument' | 'label' | 'place'
+ | 'recording' | 'release' | 'release_group' | 'series' | 'work';
+
```
reosarevok, line 13: I'd suggest to just use `CoreEntityTypeT`. We don't currently have url-url relationships, but we could choose to have them, and we are hoping to add genre-url relationships eventually, so it makes sense to not disallow those either IMO, and it simplifies this.

reosarevok:
Not checking all the code through yet, but on quick testing, one comment: Adding a link like https://www.mainlynorfolk.info/frankie.armstrong/ autoselects both lyrics and other databases, whether you add it to an artist (where only the other databases link is allowed anyway) or to a work (where both relationships make sense, but this link is invalid for both of them). Maybe it should only add types that are actually valid? My original idea was that this wouldn't actively select *all* the possible links, but it would give the user a dropdown/select limited to only the possible valid values (and if they wanted to select several they would just add the URL several times - but that bit I'm sure we can improve on).

y-young:
> Not checking all the code through yet, but on quick testing, one comment: Adding a link like https://www.mainlynorfolk.info/frankie.armstrong/ autoselects both lyrics and other databases, whether you add it to an artist (where only the other databases link is allowed anyway) or to a work (where both relationships make sense, but this link is invalid for both of them). Maybe it should only add types that are actually valid? My original idea was that this wouldn't actively select _all_ the possible links, but it would give the user a dropdown/select limited to only the possible valid values (and if they wanted to select several they would just add the URL several times - but that bit I'm sure we can improve on).

If you're talking about auto-selection logics that are related to URL, we can add another function for this, which accepts an URL as input and output a relationship type or types combination. So the entire logic might be:
1. Check if there is an auto-selection function, if so, use it to guess the type
2. Check if types are specified
    1. If only it only contains one type, use it blindly (like the current one)
    2. Otherwise limit possible options to the listed types/combinations
3. Otherwise check the relationship and URL against restricted types

reosarevok:
That sounds sensible to me, in principle, but check it with @yvanzo :)

yvanzo:
Just a reminder of the situation: Since 2016, the `type` field is used for both type auto-selection and relationship validation. It seemed to be be smart at first because it prevented duplicating a fair amount of code, but it has shown to have shortcomings by not allowing to auto-select multiple types and not allowing to validate more than one relationship for the same URL.

Yes the proposed changes sound sensible to me too, especially:
* Replacing `type` field with `types` to allow editors to select relationship types among a subset of valid ones (e.g. for Bandcamp pages not necessarily offering both download and streaming);
* Adding another field (or function) to actively auto-select more than one relationship type when appropriate (e.g. for Jamendo always offering both download and streaming);
* The entire logic you just described.

I didn’t go further into reviewing the code as it is going to change a lot again.

Only one question to @y-young: Does it have to be a function? Or would another field just work?

y-young:
> Only one question to @y-young: Does it have to be a function? Or would another field just work?

Adding a function provides the most flexibility, other options like an array of regex expression and type pairs also works, but it requires iterating the array for a match, hence the order and precision of regexp matter a lot.
```javascript
[{ regexp, types }, { regexp, types }, ...]
```
Besides, it seems that most of current `validate` functions do the same thing: extra information from URL using regexp, and use it with entity type, so maybe an regexp and a mapping to types also work:
```javascript
{
  regexp: /^https:\/\/music\.amazon\.com\/(albums|artists)/,
  match: {
    album: ...,
    artists: ...
  }
}
```

yvanzo:
I thought that the added field could be simpler and have the same structure as `types`.

> it seems that most of current `validate` functions do the same thing: extra information from URL using regexp, and use it with entity type, so maybe an regexp and a mapping to types also work

They also check that the URL follows the normalized format that is supposed to be enforced using `clean` function.

y-young:
Added feature to limit possible type options through `types` property and `RESTRICTED_LINK_TYPES`.
Implemented MBS-8932 as an example.

y-young:
Add support of custom auto-select function, enhanced auto-selection of mainlynorfolk links as an example, implemented auto-selection of Jamendo links as an example of non-custom logic.

mwiencek:
From what I understand, this will auto-select (i.e. add multiple relationships) for every valid relationship type for the URL.  Is that desired in most cases?  At first I expected it'd just keep a single entry with the type unselected, asking the user to select a type, with only the valid types available in the dropdown -- and if they want to add it to both types, they can paste it again (which, if there's only one valid type left that can be auto-selected, would then be selected).  Or is this feature only meant to be used where we know every type should be added?

y-young:
> From what I understand, this will auto-select (i.e. add multiple relationships) for every valid relationship type for the URL. Is that desired in most cases? At first I expected it'd just keep a single entry with the type unselected, asking the user to select a type, with only the valid types available in the dropdown -- and if they want to add it to both types, they can paste it again (which, if there's only one valid type left that can be auto-selected, would then be selected). Or is this feature only meant to be used where we know every type should be added?

I think it'll only be used when we're sure that both types should be selected. We can also limit possible options and leave the choice to user -- as you said, but I'm afraid we won't support auto-selecting the only option left, since it requires mutual information between different fields, but the user can see there're only 2 options available and they can select another themselves.

yvanzo (commented):
Great progress! I mostly checked code and tested a little bit (with Jamendo and Mainly Norfolk).

Additionally to threaded comments, your commits should be reorganized and documented with proper message for this PR. There are so many changes of different kind (renaming, refactoring, flow typing, adding field, ...), squashing it would make them difficult to review as a whole, and probably even more difficult to understand them from git log in the future.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1,13 +1,29 @@
...
+ * @flow
  * Copyright (C) 2010 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
  * and is licensed under the GPL version 2, or (at your option) any
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import $ from 'jquery';
+import * as React from 'react';
```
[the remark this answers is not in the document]
y-young: It's for the type annotation on https://github.com/metabrainz/musicbrainz-server/pull/2184/files#diff-a808bbcd17cb3e956e9ac0fce554d75f6e9db02e020bc03ca543c198ac8f8e06R4752, is there a better approach?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -733,6 +779,12 @@ const CLEANUPS = {
   },
   'archive': {
     match: [new RegExp('^(https?://)?([^/]+\\.)?archive\\.org/', 'i')],
+    types: [
+      LINK_TYPES.score,
+      LINK_TYPES.lyrics,
+      LINK_TYPES.streamingfree,
+      LINK_TYPES.downloadfree,
+    ],
```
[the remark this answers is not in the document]
y-young: Do you mean limiting possible type options?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2238,7 +2290,7 @@ const CLEANUPS = {
   },
   'jamendo': {
     match: [new RegExp('^(https?://)?([^/]+\\.)?jamendo\\.com', 'i')],
-    type: LINK_TYPES.downloadfree,
+    types: [multiple(LINK_TYPES.downloadfree, LINK_TYPES.streamingfree)],
```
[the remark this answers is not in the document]
y-young: Do you mean that `types` should only be used to limit type options, and `select` function takes full charge of auto-selection?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2554,16 +2606,43 @@ const CLEANUPS = {
...
-    type: LINK_TYPES.lyrics,
+    types: [LINK_TYPES.lyrics],
   },
   'mainlynorfolk': {
     match: [new RegExp('^(https?://)?(www\\.)?mainlynorfolk\\.info', 'i')],
-    type: LINK_TYPES.otherdatabases,
+    types: [
+      LINK_TYPES.otherdatabases,
+      {work: [LINK_TYPES.otherdatabases.work, LINK_TYPES.lyrics.work]},
+    ],
```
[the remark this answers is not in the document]
y-young: We need it so that these 2 options are shown in the type selector.

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: Yes

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: Is a type selector needed here? I understood that all Mainly Norfolk song pages have lyrics. Am I missing something?

Or do you mean `types` values will always be redundant with `select`results?

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: I mean that restricting options up to two possible relationship types is not the same as auto-selecting two relationship types.

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
y-young: It's related to the UI, since `types` control what options are shown in the select dropdown, otherwise the type select will show nothing after auto-selection. And yes, `types` values will always be redundant with `select` results?

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: I didn’t notice this piece of UI code at the bottom of this file implementing mostly logic otherwise. It should probably be split later on. Definitely not an issue of this PR.

@mwiencek: Would there be a way not to import these just for Flow definitions?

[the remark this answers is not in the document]
mwiencek: Flow provides global references to the React types if you replace `.` with `$`, so `React$Node` is usable without importing anything.

y-young:

y-young:

reosarevok:
Testing this a bit, it doesn't seem too bad, but we should not automatically select lyrics for mainlynorfolk (see the example given on the STYLE-1466 ticket, https://www.mainlynorfolk.info/john.kirkpatrick/songs/johnball.html, that specifies that one has no lyrics because of copyright issues). My preference with autoselect would be to never autoselect all types, just let the user choose but limit the choices to the possible "autoselect" options.

y-young:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2636,30 +2636,6 @@ const CLEANUPS: CleanupEntries = {
...
-            break;
-          default: // artist
-            if (sourceType === 'artist') {
-              return LINK_TYPES.otherdatabases.artist;
-            }
-            break;
-        }
-      }
-      return false;
-    },
```
yvanzo: Your commit https://github.com/metabrainz/musicbrainz-server/pull/2184/commits/4fed82e7fe908cf87902361c1ed05543004f4d6f removes auto-selecting “other databases” too. From comments https://github.com/metabrainz/musicbrainz-server/pull/2184#issuecomment-901833627 and https://github.com/metabrainz/musicbrainz-server/pull/2184#issuecomment-901888214 it should just not auto-select “lyrics” instead.

When adding `https://mainlynorfolk.info/folk/songs/acrossthewesternocean.html` to http://localhost:5000/work/aaaab3b7-b743-4bf4-978c-bb3c02cf9913/edit no type is currently selected, while it seems “other databases” should be auto-selected and “lyrics” should be selectable.

yvanzo (changes requested):

root/static/scripts/edit/URLCleanup.js
reosarevok: https://github.com/metabrainz/musicbrainz-server/pull/2184#issuecomment-891475945 suggests neither of the two to be autoselected if there's more than one - if that's the easiest implementation and it would need heavy refactoring to have "always" vs "optional", then I'd rather keep it as all optional. After all, most cases (bandcamp for example) can be one of the possible options, several at the same time, or rarely none of them even.

yvanzo: Auto-selecting only one relationship type is implemented already, “select” is for “always” and “types” for “allowed” if I recall correctly.

y-young: The current logic is whether we use auto-selection and block user from adding other relationships, or we don't auto-select at all, as discussed before.

y-young:

y-young:

y-young:

y-young:

y-young:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4497,154 +4591,317 @@ function testAll(tests, text) {
...
-          },
-        ),
         result: false,
-        target: ERROR_TARGETS.URL,
+        error: texp.l(
+          'This type must be used with {types}.',
+          {types: this.buildSentence(mustBeCombinedWith)},
+        ),
+        target: ERROR_TARGETS.RELATIONSHIP,
       };
```
yvanzo, line 4840: For example, when selecting the optional “lyrics” relationship type for a Mainly Norfolk URL, it returns “This type must be used with ‘other databases’.” just under the URL.

1. The actual cause of the error is not the selected relationship, it is the missing relationship. So the error target should not be `TARGET_ERRORS.RELATIONSHIP` and the error message should probably be something like “The relationship type ‘other databases’ is missing for this URL.”.
2. Maybe the error message could be displayed under the set of relationships, and even under the button “Add a new relationship” which can be used by the user to address this error. It seems it could be worth creating a new error target: `TARGET_ERRORS.RELATIONSHIP_SET`.

The 1. is a bug to be fixed, but the 2. is just a suggested improvement that can wait for further refactoring as [discussed on IRC](https://chatlogs.metabrainz.org/libera/metabrainz/msg/4850037/).

yvanzo (changes requested):

root/static/scripts/edit/URLCleanup.js
yvanzo: Implemented in commit <https://github.com/metabrainz/musicbrainz-server/pull/2184/commits/448ea12d62103fb5fb808d600f33143b6e7f0650>, thanks.

Note: The above linked discussion mentioned both approaches, and auto-selecting multiple relationship types was also part of the final roadmap. Not having selected any type in this case would have require editors to manually select the mandatory relationship type each time. Now editors just have to select the optional relationship type if needed.

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: After having discussed this further on IRC (a few days ago?), no change is required here. When `types` contains only one possible combination of types, this combination is auto-selected. It requires less code (no duplication from `types` to `select`) even though it makes the code slightly more complicated to follow.

y-young:

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: Actually not all Mainly Norfolk song pages have lyrics, so “lyrics” relationship is allowed as an additional relationship to “other databases”. This has been fixed with commits https://github.com/metabrainz/musicbrainz-server/pull/2184/commits/895dd1183d1cb35dd75a52f392f6251509cd8233 and <https://github.com/metabrainz/musicbrainz-server/pull/2184/commits/b436019f728aee0a0e1f52f3b971c7c420809194>.

root/static/scripts/edit/URLCleanup.js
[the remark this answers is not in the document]
yvanzo: This has been fixed in commit <https://github.com/metabrainz/musicbrainz-server/pull/2184/commits/e85dfecb0e2bc39c67c4762accb4b6bb4b9359d3>.

root/static/scripts/edit/URLCleanup.js
yvanzo: Bug 1. is now fixed in commit <https://github.com/metabrainz/musicbrainz-server/pull/2184/commits/231b6c481e72e0e328901cc6183ec6698c6fb1fd>.

yvanzo: Suggestion 2. was twofold: updating the error message will be part of MBS-11899, renaming `types` (and/or other properties of `CLEANUPS`?) will be decided at the same time as the larger renaming of `URLCleanup` and `ERROR_TARGETS` that was planned anyway.

yvanzo (commented):
Thanks, all errors coming from this pull request have been resolved. It mostly requires a better organization of git commits to have a more useful git log (as requested above).

root/static/scripts/edit/URLCleanup.js
```diff
@@ -355,13 +373,36 @@ const linkToVideoMsg = N_l(
...
+
+type CleanupEntry = {
+  clean?: (url: string) => string,
+  match: Array<RegExp>,
+  select?:
+    (url: string, sourceType: CoreEntityTypeT) =>
+    | string // Single type
+    | Array<string> // A type combination
+    | false, // No match
+  types?: Array<EntityTypesMap>,
```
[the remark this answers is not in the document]
y-young: Now renamed to 'restrict'.

t/selenium/release-editor/Seeding.json5
```diff
@@ -289,17 +289,17 @@
     {
       command: 'assertEval',
       target: "Array.from(document.querySelectorAll('#external-links-editor tr[class=\"external-link-item\"]')).slice(0,3).map(x => x.querySelector('a.url') ? x.querySelector('a.url').href : x.querySelector('input').value).join('\\n')",
-      value: 'http://foo.bar.baz/\n\nhttp://foo.bar.baz/foo/',
+      value: 'http://foo.bar.baz/\nhttp://www.amazon.co.jp/\n',
```
yvanzo, line 292: Did you replace this example to trigger a different error message? Just asking because I’m not sure.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -358,14 +366,24 @@ const linkToVideoMsg = N_l(
...
+ *          If there's only one type or type combination,
+ *          it will be auto-selected.
+ * - select: Custom function for auto-selection of relationship type(s).
+ *          To select a single type, return a string,
+ *          e.g: LINK_TYPES.otherdatabases.release.
+ *          To select multiple types, return an array of strings,
+ *          e.g: [
+ *            LINK_TYPES.otherdatabases.work,
+ *            LINK_TYPES.lyrics.work,
+ *          ].
```
yvanzo: Please fix indent for this commit too as suggested in earlier comment to the first commit.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4625,6 +4676,27 @@ entitySpecificRules.recording = function (url) {
...
 };
 
+/**
+ * Merge multiple link types into one object.
+ *
+ * e.g: multiple(LINK_TYPES.otherdatabases, LINK_TYPES.lyrics)
+ * returns: {
+ *  release: [LINK_TYPES.otherdatabases.release, LINK_TYPES.lyrics.release],
+ *  ...
+ * }
```
yvanzo, line 4687: There is no `LINK_TYPES.lyrics.release`. Please provide a full real example.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4646,12 +4718,22 @@ export class Checker {
...
+    const sourceType = this.entityType;
     const types = this.filterApplicableTypes();
     // If not applicable to current entity
     if (types.length === 0) {
       return false;
     }
+    // If there is a `select` function, use its return value directly
+    if (cleanup && cleanup.select) {
+      const result = cleanup.select(this.url, sourceType);
+      // `select` function may return false if it can't determine
```
yvanzo, line 4732: That should be documented in `CLEANUPS` comment instead.

root/static/scripts/edit/externalLinks.js
```diff
@@ -1032,13 +1109,30 @@ export class ExternalLink extends React.Component<LinkProps> {
...
+            <td />
+            <td>
+              {texp.l(
+                `The following type(s) will be automatically selected
+                 when submitting: {types}`,
+                {
+                  types: firstLink.pendingTypes
+                    .map(type => linkedEntities.link_type[type].link_phrase)
+                    .join(', '),
+                },
```
yvanzo, line 1126: Please use `commaOnlyList` instead, as it supports localization.

yvanzo (commented):
Thank you, this new commits partition makes a lot of sense. Added comments a few trivial issues about coding style, documentation, and localization. Additionally:
1. The first commit seems to gather a lot of different changes. If these can be easily pushed as separate commits, please do so. If that is too intricate, let it as-is.
2. Please drop `archive.org` as discussed on IRC and replace it with a simple example.
3. Replace reference to STYLE-1466 with MBS-11912 as @reosarevok just added it.
4. When commit is partial implementation (of MBS-9902), please add a part identifier in parenthesis (I) after it.
5. Your commit messages rightfully explain _what_ changed. Please expand them to explain _why_ this has been changed, see <https://chris.beams.io/posts/git-commit/>.

yvanzo:
Just added MBS-11913 to track improvements of Jamedo’s auto-select. Can be referenced in commit message too.

t/selenium/release-editor/Seeding.json5
y-young: No, it's just because there's no ASIN option after restricting type options, it'll only appear when an Amazon link is entered.

root/static/scripts/edit/externalLinks.js
```diff
@@ -966,13 +1110,33 @@ export class ExternalLink extends React.Component<LinkProps> {
...
             />
         ))}
+        {firstLink.pendingTypes &&
+          <tr>
+            <td />
+            <td>
+              {exp.l(
+                `The following type(s) will be automatically selected
+                 when submitting: {types}`,
+                {
```
mwiencek, line 1123: I don't quite understand this message.  It doesn't seem I can override this behavior at all, so I'm not sure why it's telling me this instead of just selecting every type.  The "when submitting" part is also confusing, since "submit" to me is when I submit the form, but these are just selected when I move focus.

PR #2189, opened 2021-07-30
Autocomplete2 editor fixes

reosarevok:
Fixes some issues I found while I was trying to figure out if MBS-11588 was an issue in Autocomplete2. Which I'm still not sure about, but that's a different question (@mwiencek, do you know the answer?).

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -78,8 +78,12 @@ function doSearch<T: EntityItemT>(
     });
   });
 
+  const entityWebServicePath = state.entityType === 'editor'
+    ? 'editor'
+    : ENTITIES[state.entityType].url;
+
   const url = (
-    '/ws/js/' + ENTITIES[state.entityType].url +
+    '/ws/js/' + entityWebServicePath +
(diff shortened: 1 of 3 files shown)


mwiencek:
> Fixes some issues I found while I was trying to figure out if MBS-11588 was an issue in Autocomplete2. Which I'm still not sure about, but that's a different question (@mwiencek, do you know the answer?).

We don't use Autocomplete2 anywhere yet, so it'd be as issue in the old one still.

root/static/scripts/common/components/Autocomplete2.js
```diff
@@ -78,8 +78,12 @@ function doSearch<T: EntityItemT>(
     });
   });
 
+  const entityWebServicePath = state.entityType === 'editor'
+    ? 'editor'
+    : ENTITIES[state.entityType].url;
```
mwiencek, line 83: Heh, that's annoying.  Maybe we should alias it to ws/js/user at some point.

reosarevok:
> We don't use Autocomplete2 anywhere yet, so it'd be an issue in the old one still.

I know that, but since it's relatively minor, it might not be worth fixing if it no longer happens with Autocomplete2.

PR #2193, opened 2021-08-04
MBS-11843: Add report for "Events with annotations"

reosarevok:
### Implement MBS-11843

AnnotationReports already supported this, even, but for some reason we didn't have it yet. Not sure how useful these annotation reports actually are, but we should make them consistent at least.

### DIFF
--- a/root/report/AnnotationsEvents.js
+++ b/root/report/AnnotationsEvents.js
@@ -1,0 +1,48 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2018 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import {ANNOTATION_REPORT_TEXT} from './constants';
+import EventList from './components/EventList';
+import ReportLayout from './components/ReportLayout';
+import useAnnotationColumns from './hooks/useAnnotationColumns';
+import type {ReportDataT, ReportEventAnnotationT} from './types';
+
+const AnnotationsEvents = ({
+  canBeFiltered,
+  filtered,
+  generated,
+  items,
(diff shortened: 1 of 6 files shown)


reosarevok:

reosarevok:

reosarevok:

root/report/AnnotationsEvents.js
```diff
@@ -0,0 +1,48 @@
...
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import {ANNOTATION_REPORT_TEXT} from './constants';
+import EventList from './components/EventList';
+import ReportLayout from './components/ReportLayout';
+import useAnnotationColumns from './hooks/useAnnotationColumns';
+import type {ReportEventAnnotationT, ReportDataT} from './types';
```
yvanzo, line 16: Nit-picking: It should be alphabetically ordered (just like it already is in other `root/report/Annotations*.js`):
```suggestion
import type {ReportDataT, ReportEventAnnotationT} from './types';
```

yvanzo (approved):
Tested on sample data, reporting 67 annotations.
Production `event_annotation` has 2758 rows, even though not all of them are still current annotations, that report will probably report a fair number of events.

mwiencek:
> Not sure how useful these annotation reports actually are, but we should make them consistent at least.

Yeah, I'm okay with merging this to make things consistent, though I'd prefer removing them if they're not useful.

You can already list annotations for a particular entity type via search: https://musicbrainz.org/search?query=type%3Aevent&type=annotation&limit=25&method=advanced and this seems infinitely more useful because you can actually search the text for different kinds of credits. (The "Artist annotations" report is e.g. 167 pages and not searchable at all, so I'm not sure why you'd ever use that over a search.)

mwiencek:
chaban pointed out in the ticket that the search is often unreliable (old annotations being listed) or inconsistent (order changing between pages or on reload), which seems fair enough to me

mwiencek (approved):

PR #2196, opened 2021-08-05
MBS-11850: Make footer links more visible 

reosarevok:
### Implement MBS-11850

By following the same general style as in the header menu, adding color, making them bold and adding a bit more space between them, these links will hopefully stand out a bit more.
This also moves all the explanatory text for slaves/staging servers to the right paragraph to let the links stand alone on the left, although that shouldn't be a problem in production anyway.

screencapture-localhost-5000-2021-08-05-14_17_09


reosarevok:

PR #2198, opened 2021-08-09
MBS-11830: Add more collection statistics

reosarevok:
### Implement MBS-11830

We're collecting (hah!) basically no collection stats at all. This adds statistics for number of collections, public vs private, with collaborators, and of each type.


yvanzo (commented):
LGTM but did not test. Has it been somewhat tested?

reosarevok:
Well, I checked the numbers in the statistics page, and they didn't seem very odd, but no further testing tbh.

root/statistics/stats.js
```diff
@@ -164,6 +164,132 @@ const stats = {
...
+    category: 'collection',
+    color: '#ff0000',
+    description: l('Collections of artists'),
+    label: l('Artist collections'),
+  },
+  'count.collection.type.attending': {
+    category: 'collection',
+    color: '#ff0000',
+    description: l('Collections with type set to Attending'),
+    label: l('Collections of type Attending'),
```
mwiencek, line 207: These are saying the exact same thing with slightly different wording, and I'm not sure why one wording is considered a description and another is a label. :D  Maybe we should leave description blank/null if there's nothing meaningful to add beyond the label?  Even "Artist collections" vs. "Collections of artists," and probably some others feel the same way, i.e. they're minimal re-wordings that don't feel useful.  Though I'm sure we already have existing items that do this.

lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -1941,7 +1941,72 @@ my %stats = (
...
+
+    "count.collection.type.release" => {
+        CALC => sub {
+            my ($self, $sql) = @_;
+
+            my $data = $sql->select_list_of_lists(
+                "SELECT type, COUNT(*) AS count
+                FROM editor_collection
+                GROUP BY type
+                ",
```
mwiencek, line 1959: The hanging `",` looked a little funny so I thought I'd suggest using `<<~'SQL'` here if you feel like it.

lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -1941,7 +1941,72 @@ my %stats = (
...
+
+            +{
+                "count.collection.public" => $dist{1} || 0,
+                "count.collection.private"  => $dist{0} || 0,
+            };
+        },
+    },
+    
+    "count.collection.has_collaborators" => {
+        DESC => "Count of collections with at least one collaborator",
```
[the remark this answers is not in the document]
reosarevok: Doesn't look like it - I think it's just a way of explaining what it is meant to be for devs using this file? Because it's otherwise unused AFAICT.

root/statistics/stats.js
reosarevok: Yes, a ton of these seem to be "label is just shorter to fit on the sidebar better", but with stuff like this where it's hard to make a proper short label, it starts to lose the point a bit...

lib/MusicBrainz/Server/Data/Statistics.pm
reosarevok: Made that change - a lot of these could use SQL blocks tbh, but dunno if you want me doing it now? I could add it as a second commit.

root/statistics/stats.js
yvanzo: Why the label isn’t just “of type Attending”? Like for “Artists” “of type Person” in [Overview](https://musicbrainz.org/statistics).

reosarevok: The label is (IIRC) for the timeline sidebar, so it needs to be complete.

lib/MusicBrainz/Server/Data/Statistics.pm
mwiencek: Maybe not all in this PR, but there are some more being added here that could be changed.

root/statistics/stats.js
yvanzo: Most of the current descriptions are useless (e.g. just adding “Count of” ahead of the label text), but some labels are not clear enough. They should probably be replaced with the corresponding descriptions if having long labels doesn’t cause display issues.
For example:
* “Artists not in ACs” isn’t telling as much as “Artists with no appearances in artist credits” (and even then “artist credits” should probably link to the documentation),
* “Unknown-gender Artists” is clarified with the description “Artists with gender unset (non-group artists)”,
* “Failed edits (prerequisite)” isn’t as clear as “All edits that have failed prerequisite checks” either.

I didn’t even check all the 130+ descriptions. It seems that “description” was expected to be self-explanatory while “label” depends on the grouped display. Some git archeology might help with reminding its expected purpose.

Would you be alright with moving this change into a new PR so the current one can be merged sooner?

mwiencek (approved):

yvanzo (approved):

PR #2199, opened 2021-08-09
MBS-11854: Recognize unicode hyphen in guess case

reosarevok:
### Fix MBS-11854

Words with a unicode hyphen were not being recognized as split words: the hyphen was just being considered one more character. This changes that, and then ensures that guess case recognizes the hyphen as such and does the same as with a hyphen-minus.

Sadly, part of my fix for MBS-10156 (for re‐mode with unicode hyphen) will no longer work now since it is no longer considered
one single word, but I think this makes a lot more sense for now. The current code already didn't do anything with re-mode with a hyphen-minus, so this actually adds consistency.


root/static/scripts/guess-case/MB/GuessCase/Handler/Base.js
```diff
@@ -332,7 +332,7 @@ MB.GuessCase.Handler.Base = function (gc) {
    */
   self.doHyphen = function () {
     if (!gc.regexes.HYPHEN) {
-      gc.regexes.HYPHEN = '-';
+      gc.regexes.HYPHEN = /-|‐/;
```
mwiencek, line 335: If it's a string, matchWordAtIndex appears to perform an exact match, so to be equivalent you'd need
```suggestion
      gc.regexes.HYPHEN = /^[\-‐]$/;
```
but I'm not sure if this would matter in practice

reosarevok: Might not, but might as well change it.

PR #2200, opened 2021-08-09
Test loading DBDefs before starting server

yvanzo:
# Problem

Occasionally a server instance would continuously error, we assume it likely did not start with a valid `lib/DBDefs.pm`. In production, this file is generated from `docker/templatesDBDefs.pm.ctmpl` and a JSON config file using git2consul/consul-agent/consul-template.

It is not clear which of these components is the cause of this problem. The content of the `lib/DBDefs.pm` is not known either, but it isn’t missing or empty. (Different error messages would be thrown then.)

# Solution

This patch just tries loading DBDefs module before running the server. If the module cannot be loaded, consul-template will try starting a new instance automatically, hopefully using a correct configuration provided by consul-agent in the meantime. It assumes to fix a potential bug with consul-template not restarting the server when erroring continuously.

Disclaimer: This does NOT test anything beyond loading DBDefs module such as: validity of configuration values, connection to the database or any other service.

Impl. note: Server won’t reload `lib/DBDefs.pm` if it changed, so there is no need to check DBDefs all the time while running. Consul is in charge of starting/stopping/restarting the server.

This is probably a naive approach but that might save waiting for someone to restart containers when consul-agent misbehaves.

# Checklist for author

* [x] Test script works in an existing container
* [x] Build a test image
* [x] Deploy to test instance
* [x] Check it continues to run under normal conditions
* [x] Test deploying invalid configuration through docker server configs

No idea how to reproduce consul-agent not sending proper values in time though.

Note:  Tests helped with identifying that `lib/DBDefs.pm` was neither missing nor empty at least.

### DIFF
--- a/docker/scripts/start_musicbrainz_server.sh
+++ b/docker/scripts/start_musicbrainz_server.sh
@@ -21,6 +21,10 @@ if [ -s "$PID_FILE" ]; then
     fi
 fi
 
+chpst -u musicbrainz:musicbrainz \
+    carton exec -- ./script/dbdefs_exists \
+    || exit $?
+
 exec chpst -u musicbrainz:musicbrainz \
     carton exec -- \
         start_server --port 5000 --pid-file "$PID_FILE" -- \
--- a/docker/templates/macros.m4
+++ b/docker/templates/macros.m4
@@ -144,6 +144,7 @@ copy_mb(``admin/ admin/'')
 copy_mb(``app.psgi entities.json ./'')
 copy_mb(``bin/ bin/'')
 copy_mb(``lib/ lib/'')
+copy_mb(``script/functions.sh script/dbdefs_exists script/'')
 copy_mb(``script/functions.sh script/git_info script/'')')
 
 m4_define(
--- a/script/dbdefs_exists
+++ b/script/dbdefs_exists
@@ -1,0 +1,10 @@
+#!/usr/bin/env perl
+
+use utf8;
+use strict;
+use warnings;
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+use DBDefs;
+
+print "DBDefs exists\n"


docker/scripts/start_musicbrainz_server.sh
```diff
@@ -21,6 +21,10 @@ if [ -s "$PID_FILE" ]; then
     fi
 fi
 
+chpst -u musicbrainz:musicbrainz \
+    carton exec -- ./script/dbdefs_exists \
+    || exit $?
+
```
mwiencek, line 27: I was looking at some of the server logs and reading through consul-template docs to understand this a bit more: https://github.com/hashicorp/consul-template/blob/master/docs/modes.md#exec-mode

In exec mode, consul-template doesn't supervise the `start_server` process -- if it dies then consul-template dies with it.  In that case, runit should restart consul-template, which would be good.  But that doesn't seem to be what's happening; `start_server` is staying alive while plackup is failing:

```
[error] Caught exception in engine "There is no configuration in DBDefs for database READWRITE but one is required at lib/MusicBrainz/Server/DatabaseConnectionFactory.pm line 47.
MusicBrainz::Server::DatabaseConnectionFactory::get_connection("MusicBrainz::Server::DatabaseConnectionFactory", "READWRITE", "fresh", 0) called at lib/MusicBrainz/Server/Context.pm line 46
...
```

So I'm not sure this bit of code being added will actually run when we want, because `start_musicbrainz_server.sh` isn't being invoked multiple times.

Another issue, I think, is that DBDefs isn't actually non-existent or corrupt here -- as mentioned on IRC, just the `register_databases` call is receiving empty input.  The actual exception occurs in DatabaseConnectionFactory when it can't find READWRITE.  So script/dbdefs_exists won't actually indicate that anything is wrong, because the DBDefs module itself will load fine.

yvanzo: Where does the above error message comes from? It was not in yesterday’s log of MB WS container on `trille`.

mwiencek: > So we should rather have start_server to call database_exists before running plackup?

Well, start_server isn't our code, so I don't think this is possible.  But start_server isn't being invoked again -- the service itself never dies or restarts.

> Where does the above error message come from? It was not in yesterday’s log of MB WS container on trille.

It's from all website containers (from 8/6 on trille), so not the same day or the exact same issue, but same behavior.  The service didn't restart when it encountered "Can't use an undefined value as a subroutine reference" either, otherwise you'd see it from the logs.

yvanzo: Sorry, I did not intend to change `start_server` itself but rather having a wrapper or something to make `start_server` die when “`register_databases` call is receiving empty input”.

yvanzo:
Definitely too naive to be useful. Let’s try to replace `consul-template` instead as @mwiencek suggested on IRC.

PR #2201, opened 2021-08-09
MBS-11846: Display release artist on release group view

reosarevok:
### Implement MBS-11846

Right now, there's no way to see if a release inside a release group has a different artist credit than the release group itself.
This adds a release artist column to that view. It also adds an "artist-credit-variation" class for cases where the release and release group artist credits differ, to help userscript users highlight that (matching what we do for track artists on recording pages).


mwiencek:
Would it make sense to only show these when there are multiple or are different from the RG, somewhat like we do for tracklists?

reosarevok:
We could do that, but I'm not sure it makes sense - we show release artist unconditionally on all other release lists, including lists for an artist, even if all the releases there are by that artist alone.

PR #2203, opened 2021-08-09
MBS-11825: Use same order for art types when editing vs adding

reosarevok:
### Implement MBS-11825

The Add Cover Art page currently uses a knockout form that is sorted by ID (notice CoverArtFields is display: none). The Edit Cover Art page uses the CoverArtFields display that is sorted by l_name. This is actually one of the rare cases where an ID sort seems preferable, since that ensures Front, Back and Booklet (the most relevant and common types) are sorted first whatever the language. As such, this forces sort by ID for the CoverArtType OptionsTree.


reosarevok:

reosarevok:

mwiencek:
We have a `child_order` column which we could use...

But at least this makes them consistent for now.

PR #2204, opened 2021-08-09
MBS-2221: Add date docs to relationship editor

reosarevok:
### Implement MBS-2221

This adds a bit of documentation about how to indicate a relationship happened in one day (or month/year) only and about the ended checkbox.

Only adding it to the relationship editor and not DateRangeFieldset since it seems mostly relationship specific (does not apply much to aliases, for example).

Screenshot from 2021-08-09 21-28-21

### DIFF
--- a/root/components/relationship-editor.tt
+++ b/root/components/relationship-editor.tt
@@ -128,6 +128,9 @@
       <!-- ko if: isBeginDate -->
         <button class="icon copy-date" type="button" title="[% l('Copy to end date') | html %]" data-bind="click: $parent.copyDate"></button>
       <!-- /ko -->
+      <!-- ko if: !isBeginDate -->
+        (<a href="#" data-bind="click: $parent.toggleDatesHelp.bind($parent)">[% l('help') | html %]</a>)
+      <!-- /ko -->
       <div class="error" data-bind="text: $parent.dateError(date)"></div>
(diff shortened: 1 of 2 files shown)


root/components/relationship-editor.tt
```diff
@@ -128,6 +128,9 @@
       <!-- ko if: isBeginDate -->
         <button class="icon copy-date" type="button" title="[% l('Copy to end date') | html %]" data-bind="click: $parent.copyDate"></button>
       <!-- /ko -->
+      <!-- ko if: !isBeginDate -->
+        (<a href="#" data-bind="click: $parent.toggleDatesHelp.bind($parent)">[% l('help') %]</a>)
```
mwiencek, line 132:
```suggestion
        (<a href="#" data-bind="click: $parent.toggleDatesHelp.bind($parent)">[% l('help') | html %]</a>)
```

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -375,6 +375,10 @@ class Dialog {
     relationship.entities(relationship.entities().slice(0).reverse());
   }
 
+  toggleDatesHelp() {
+    this.showDatesHelp(!this.showDatesHelp.peek());
```
mwiencek, line 379: I'm assuming you don't need to call preventDefault here, since `toggleAttributesHelp` doesn't appear to (since with `href="#"` the scroll position would normally jump to the top of the page).

PR #2205, opened 2021-08-10
MBS-11861: improve loopParity classes for tablesorter

reosarevok:
### Fix MBS-11861

This was only dealing with "even", but that meant that some rows ended up with both "even" and "odd" classes, while some that should be "odd" ended with nothing. This makes it all work more like the actual loopParity function, where you either have "odd" or "even".


mwiencek (approved):
thanks!

reosarevok:

PR #2206, opened 2021-08-10
MBS-11862: Do not show deprecated relationship types with 0 uses

reosarevok:
### Implement MBS-11862

We've historically just removed link types once we no longer wanted them used and they had 0 uses, but that is a bad idea. It means the historical edits automatically show less info, since there's no type to load anymore, and it's just an unwanted loss of historical data.
As such, this allows us to keep them in the database, but not show them except when there's something to gain from doing so: for type editing / documenting, and for the edit search.

### DIFF
--- a/lib/MusicBrainz/Server/Data/LinkType.pm
+++ b/lib/MusicBrainz/Server/Data/LinkType.pm
@@ -114,14 +114,28 @@ sub load
 
 sub get_tree
 {
-    my ($self, $type0, $type1) = @_;
+    my ($self, $type0, $type1, %opts) = @_;
 
     my %id_to_obj;
     my @objs;
+    my $extra_condition = '';
+
+    unless ($opts{get_deprecated_and_empty}) {
+        $extra_condition = <<~'EOSQL';
+            AND (
+                is_deprecated = FALSE
+                OR
+                EXISTS (
+                    SELECT 1 FROM link WHERE link.link_type = id
+                )
+            )
+            EOSQL
+    }
+
     for my $row (@{
         $self->sql->select_list_of_hashes(
-            'SELECT '  .$self->_columns . ' FROM ' . $self->_table . '
-             WHERE entity_type0=? AND entity_type1=?
+            'SELECT ' . $self->_columns . ' FROM ' . $self->_table . '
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/LinkType.pm
```diff
@@ -114,14 +114,27 @@ sub load
...
 sub get_tree
 {
-    my ($self, $type0, $type1) = @_;
+    my ($self, $type0, $type1, $get_deprecated_and_empty) = @_;
 
     my %id_to_obj;
     my @objs;
+    my $extra_condition = '';
+
+    unless ($get_deprecated_and_empty) {
```
mwiencek, line 123: This might make things a little clearer:

```suggestion
    my ($self, $type0, $type1, %opts) = @_;

    my %id_to_obj;
    my @objs;
    my $extra_condition = '';

    unless ($opts{get_deprecated_and_empty}) {
```
then instead of doing `get_tree($type0, $type1, 1)` you would do `get_tree($type0, $type1, get_deprecated_and_empty => 1)`, which is more self-documenting.

lib/MusicBrainz/Server/Data/LinkType.pm
```diff
@@ -114,14 +114,27 @@ sub load
...
+        $extra_condition = <<~'EOSQL';
+            AND NOT (
+                is_deprecated = TRUE
+                AND NOT EXISTS (
+                    SELECT 1 FROM link WHERE link.link_type = id
+                )
+            )
```
mwiencek, line 130: The double negative is kinda hard to parse. How about:
```suggestion
            AND is_deprecated = FALSE
            AND EXISTS (SELECT 1 FROM link WHERE link.link_type = id)
```

reosarevok: That's not what we want, though, it should show deprecated things as long as 1 exists in link, but I guess it could be simpler with those two conditions with an OR. Will do.

lib/MusicBrainz/Server/Data/LinkType.pm
reosarevok: Thanks, I wanted that but always forget how it goes :D

PR #2209, opened 2021-08-11
MBS-11836: Convert Edit Barcodes edit to React

reosarevok:
### Implement MBS-11836

We can just pass the barcode as a string | null, since we were only making them into a new Barcode to use Barcode::format. This just uses formatBarcode instead on the JS side.

Tested by looking around an edit search for this relationship type, both from the newest and oldest edits.


reosarevok:

mwiencek (approved):

PR #2210, opened 2021-08-11
Update HACKING to create the SELENIUM DB directly

reosarevok:
If I've understood this right, then we do not need to copy the DBs like this, but we can just run the create_test_db script with SELENIUM to create the appropriate DB.


reosarevok:

PR #2214, opened 2021-08-12
MBS-11823: Don't break tags and genres mid-word

reosarevok:
### Implement MBS-11823

This was set to word-break: break-all to avoid long one-word tags such as weird URLs. That actually means all tags with normal words break weirdly. @mwiencek proposed overflow-wrap: anywhere to cater to both of these issues, and that seems to work - the only worry is that this is not supported in older browsers.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -481,7 +481,7 @@ div.warning img.warning {
         display: flex;
         a {
             display: inline-block;
-            word-break: break-all;
+            overflow-wrap: anywhere;
             padding: 4px;
             flex-grow: 1;
             -webkit-flex-grow: 1;


reosarevok:
I'm not sure how :/ Do you have any ideas? I think it's better to cater to the most common issue with words, personally, but maybe you can think of a way to split it into lines if there's one and only one word?

mwiencek:
Does either of `overflow-wrap: anywhere` or `overflow-wrap: break-word`  work?  Seems like `word-break: break-word` does too, but MDN says that's deprecated.

reosarevok:
`overflow-wrap: anywhere` does work. It's not supported in older browser versions, but I guess the main problem with those will be that the Facebook tag and whatnot won't wrap properly, which doesn't seem like a huge issue?

mwiencek:
Agreed, it's a pretty rare occurrence and doesn't make the page completely unusable in older browsers at least.

PR #2215, opened 2021-08-13
Beta: Allow submitting form from empty URL input

y-young:
# Problem
After MBS-11680, Enter key is used for submitting links in URL input box, hence users can no longer submit the form from external links editor when they only want to add a single link.

# Solution
Redirect focus to the next item in external links list (either link or input), and prevent default onKeyDown handler only when the field is not empty, therefore users can follow the following steps to submit the form:
1. Paste a URL
2. Press Enter, this submits the link and jumps to the next empty input box
3. Press Enter again, the form is submitted

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -202,15 +202,22 @@ export class ExternalLinksEditor
     if (url !== '' && !error) {
       link.submitted = true;
       this.setLinkState(index, link, () => {
-        /*
-         * Redirect focus to the next item (either input or link)
-         * instead of staying on the current link.
-         */
-        $(this.tableRef.current)
-          .find(`tr.external-link-item:eq(${urlIndex + 1})`)
-          .find('a,input')
-          .eq(0)
-          .focus();
+        // Redirect focus instead of staying on the current link
+        if (link.type) {
+          // If type is selected, jump to the next item(either input or link)
+          $(this.tableRef.current)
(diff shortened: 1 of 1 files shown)


reosarevok (approved):
On quick test, this seems like a clear improvement to me. Please change the base branch to beta though :)

yvanzo (commented):
When the link is valid but the type isn’t auto-selected, the focus should likely move to the type selector instead. Would it be possible to check for type selector before jumping to the next link input?

root/static/scripts/edit/externalLinks.js
```diff
@@ -202,15 +202,22 @@ export class ExternalLinksEditor
...
+          // If type is selected, jump to the next item(either input or link)
+          $(this.tableRef.current)
+            .find(`tr.external-link-item:eq(${urlIndex + 1})`)
+            .find('a,input')
+            .eq(0)
+            .focus();
+        } else {
+          // If type is not selected, jump to type selector
+          $(this.tableRef.current)
+            .find(`tr.external-link-item:eq(${urlIndex})
```
yvanzo, line 216: There is an unneeded trailing white space on this line. Please check your changes using `./node_modules/.bin/eslint` and update your commits accordingly (rather than pushing a commit that specifically fixes that).

yvanzo (commented):
Tested, that works for me, thank you.

yvanzo (approved):

PR #2216, opened 2021-08-15
Beta: Fix text overflow caused by long URL

y-young:
# Problem
Long URLs in ExternalLinks and URLInputPopover cause text overflow and width change.

# Solution
Update `overflow-wrap` to `anywhere` instead of `break-word`.


y-young:

yvanzo (approved):
Good catch, tested with
`http://www.blanditiismagnamdoloresveniamdictaarchitectoeiusassumendainventore.loremipsum`
Thank you

PR #2218, opened 2021-08-16
More Perl::Critic fixes and code standardization

reosarevok:


reosarevok:

mwiencek:
This seems fine, I think, though some conflicts have been generated since you opened it (sorry).

reosarevok:
Fixed those. IIRC you mentioned at some point that the test did not seem to be actually giving warnings and that's why we disabled it - do you know if that is the case?

lib/MusicBrainz/Server/Report/AnnotationReports.pm
```diff
@@ -7,26 +7,26 @@ for my $type (grep { $_ ne 'instrument' && $_ ne 'area' } entities_with('annotat
...
+    eval <<~"EOF";
+        package MusicBrainz::Server::Report::Annotations$plural;
+        use Moose;
 
-with 'MusicBrainz::Server::Report::${model}Report',
-$subs_section
-     'MusicBrainz::Server::Report::AnnotationReport';
+        with 'MusicBrainz::Server::Report::${model}Report',
+            $subs_section
+            'MusicBrainz::Server::Report::AnnotationReport';
```
yvanzo, line 21: There is extraneous indent for `$subs_section`maybe?

lib/MusicBrainz/Server/Report/CDTOCDubiousLength.pm
```diff
@@ -6,33 +6,28 @@ with 'MusicBrainz::Server::Report::CDTOCReport';
...
-            -- custom cutoff 30 minutes for 8cm CDs
-            (medium_format.id = 34
-                AND leadout_offset > 75 * 60 * 30)
-        ORDER BY
-            medium_format.name,
-            cdtoc.leadout_offset DESC
-    }
-}
+sub query {<<~'EOSQL'}
+    SELECT cdtoc.id AS cdtoc_id, medium_format.name AS format,
```
yvanzo, line 10: Having selected columns on separate lines was probably more readable here.

lib/MusicBrainz/Server/Report/CDTOCNotApplied.pm
```diff
@@ -7,45 +7,27 @@ with 'MusicBrainz::Server::Report::CDTOCReport',
...
-            disc
-            JOIN medium AS m ON m.id = disc.medium
-            JOIN release AS r ON r.id = m.release
-        ORDER BY
-            r.name
-    }
-}
+sub query {<<~'EOSQL'}
+    WITH mc AS (
+        SELECT UNNEST(ARRAY_AGG(cdtoc)) AS cdtoc, medium
```
yvanzo, line 12: Same as above

yvanzo (commented):
* I proofread changes to Perl quotes and that part looks good to me, thanks a lot!

* The change from `EOSQL` to `SQL` doesn’t seem appropriate: Even though any string can be used to end a here-document, the form `EO…` is the most common notation, even in MeB repositories. It seems TextMate should be configured to match `EOSQL` if possible instead.

* There are many formatting changes to the SQL code, not mentioned in commit messages but one. After some comments, I stopped reviewing these changes cause I would not have time to review declared changes to the Perl code otherwise. These changes are sometimes improving the readability, sometimes not, sometimes decreasing it. It might be worth a separate topic.

mwiencek:
> There are many formatting changes to the SQL code, not mentioned in commit messages but one. After some comments, I stopped reviewing these changes cause I would not have time to review declared changes to the Perl code otherwise. These changes are sometimes improving the readability, sometimes not, sometimes decreasing it. It might be worth a separate topic.

We don't have a style guide for formatting our SQL, and probably need one, so I sympathize with this. Some of the SQL changes here can seem arbitrary and slightly less readable (if there was more indentation before, e.g. alignment of keywords, or more whitespace before). Having a style guide written somewhere would help a lot here. :)

> The change from EOSQL to SQL doesn’t seem appropriate: Even though any string can be used to end a here-document, the form EO… is the most common notation, even in MeB repositories. It seems TextMate should be configured to match EOSQL if possible instead.

If it's possible to configure easily I'd agree, but it seemed like something that is [hardcoded in some syntax file](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide).  (If not I'll be pleasantly surprised.)  So I'd prefer `SQL` if it's not easily configured, because it significantly improves readability in VSCode which I think is a bit more important than maintaining the `EO...` notation.

reosarevok:
I did check for a way to configure it first, and yeah, all I found was it seems to be hardcoded. We could send a PR for it, I guess, not sure whether they accept those happily or not :)

yvanzo:
A global change for a limited tool seems weird. It will be more readable to VSCode users only, `EO…` are recognized as termination strings. It won’t be less readable to the three of us at least: I can configure vim with a custom `~/.vim/after/syntax/perl.vim` file.

reosarevok:
I mean, I use VSCode for almost everything except small changes, so clearly it's a benefit to me :) AFAICT it's the most commonly used IDE. But I lived with EOSQL just fine before I noticed this highlighting option, so I can live with it in the future too.

mwiencek:
> A global change for a limited tool seems weird. It will be more readable to VSCode users only, EO… are recognized as termination strings.

That might just be because you're used to it; `<<~'SQL'` is just as easily recognized as a termination string to me.  Is there a good reason besides tradition, when it would be quite helpful to us if changed?

> It won’t be less readable to the three of us at least: I can configure vim with a custom ~/.vim/after/syntax/perl.vim file.

2/3 team members use VSCode, though. :) And I don't think `<<~'SQL'` is less readable or recognizable in vim (which I also switch between), especially if you can configure syntax highlighting for it just the same.

yvanzo:
> That might just be because you're used to it

Not just me, there is no heredocumentsfight.com to make statistics, but `EOF` and `END` are probably the most common termination strings in the wild, and `SQL` doesn’t indicate a termination. Anyway using here-documents is already an uncommon thing, so using a lesser common termination string is probably not the most difficult part to dive into the code.

>  [...] 2/3 team members [...]

I was thinking about new contributors mostly. I can even deal with `WAHWAH` as termination string if needed ;)

> > It won’t be less readable [...]
>
> [...] And I don't think `<<~'SQL'` is less readable [...]

Probably a miscommunication here as we keep repeating the same thing :)

mwiencek:
Yes, I think I misunderstood what you were saying there; it read as "EOSQL won't be less readable to all of us: I can configure syntax highlighting for it in my editor." :laughing:

yvanzo (changes requested):
Please revert unrelated changes to the SQL code.

yvanzo:
> vim (which I also switch between)

@mwiencek: Here is the config snippet to put in `~/.vim/after/syntax/perl.vim`:

```vim
" Here-documents with SQL end-word are assumed to embed SQL code
unlet b:current_syntax
syntax include @Sql syntax/sql.vim
let b:current_syntax = "perl"
if exists("perl_fold")
  syntax region perlSQL start=+<<\~\? *\"SQL\".*$+ end=+^ *SQL+ contains=perlVarBlock,@perlInterpDQ,@Sql keepend fold extend
  syntax region perlSQL start=+<<\~\? *\'SQL\'.*$+ end=+^ *SQL+ contains=perlVarBlock2,@perlInterpSQ,@Sql keepend fold extend
  syntax region perlSQL start=+<<\~\? *\\SQL.*$+ end=+^ *SQL+ contains=perlVarBlock2,@perlInterpSQ,@Sql keepend fold extend
  syntax region perlSQL start=+<<\~\? *[^\"\'\\]SQL.*$+ end=+^ *SQL+ contains=perlVarBlock2,@perlInterpSQ,@Sql keepend fold extend
else
  syntax region perlSQL start=+<<\~\? *\"SQL\".*$+ end=+^ *SQL+ contains=perlVarBlock,@perlInterpDQ,@Sql keepend
  syntax region perlSQL start=+<<\~\? *\'SQL\'.*$+ end=+^ *SQL+ contains=perlVarBlock2,@perlInterpSQ,@Sql keepend
  syntax region perlSQL start=+<<\~\? *\\SQL.*$+ end=+^ *SQL+ contains=perlVarBlock2,@perlInterpSQ,@Sql keepend
  syntax region perlSQL start=+<<\~\? *[^\"\'\\]SQL.*$+ end=+^ *SQL+ contains=perlVarBlock2,@perlInterpSQ,@Sql keepend
endif
```

reosarevok:
> Please revert unrelated changes to the SQL code.

That would take a long time of looking through all the SQL bits, so it would seem like a much better use of that dev time if we agreed on an SQL style all together, merged this (since it's going to get changed to match the new style anyway), and then I used the time to start applying whatever SQL style we agreed on :) (otherwise I'd spend time undoing this, *then* time adapting it to the new style). Would that seem like an acceptable option?

yvanzo:
We didn’t agree on an SQL style so far. @mwiencek suggested <https://github.com/meadmaker/sql-style-guide>. I would suggest having “select targets on separate lines” and “leading commas” on top of it; See <https://www.reddit.com/r/SQL/comments/6o6dcq/style_of_sql_indentation_and_alignment/dkf6whw/> for example.

Wouldn’t it require to review these SQL changes anyway or are these whitespace changes only? How can you be sure it doesn’t introduce bug otherwise? _It would be a much better use of dev time_ not having unrelated changes in these commits for both reviewers and the future you, in case you need to revert any of these commits later on for example.

reosarevok:
I'm ok with one line per select target but I really, really dislike those leading commas (I have never seen it before, but it just feels super confusing to me and if the main benefit is when commenting out the last column, which we barely ever do, then I don't think it's needed). Anyway, yes, I was hoping to talk about it today :)

Most SQL changes are spacing / indenting / new lines only, yes. There are a few cases where I did move some section to a variable and then used the variable, especially for a couple extra conditions that were added as a JOIN, but nothing that should change the SQL itself IIRC.

Also, these changes are only made to SQL strings that needed to be changed anyway for the right quotes (and mostly just involve moving them to [EO]SQL), except for https://github.com/metabrainz/musicbrainz-server/pull/2218/commits/6aed3bd1c3bb3e7dd7efb8d1fd7435082f1c67b6 where I did try to just get all the files standardized because we had like 4 or 5 different SQL styles in reports.

yvanzo:
> I'm ok with one line per select target but I really, really dislike those leading commas (I have never seen it before, but it just feels super confusing to me and if the main benefit is when commenting out the last column, which we barely ever do, then I don't think it's needed). Anyway, yes, I was hoping to talk about it today :)

It also keeps the river flowing on both sides, making it easier to spot a missing comma, especially with only one select target per line. It is very common too; See https://hoffa.medium.com/winning-arguments-with-data-leading-with-commas-in-sql-672b3b81eac9

> Most SQL changes are spacing / indenting / new lines only, yes. There are a few cases where I did move some section to a variable and then used the variable, especially for a couple extra conditions that were added as a JOIN, but nothing that should change the SQL itself IIRC.
> 
> Also, these changes are only made to SQL strings that needed to be changed anyway for the right quotes (and mostly just involve moving them to [EO]SQL), except for [6aed3bd](https://github.com/metabrainz/musicbrainz-server/commit/6aed3bd1c3bb3e7dd7efb8d1fd7435082f1c67b6) where I did try to just get all the files standardized because we had like 4 or 5 different SQL styles in reports. Actually, for that commit, I could just change it to standardize to whatever we decide on if we can pick a style today :)

Looking at the first commit <https://github.com/metabrainz/musicbrainz-server/pull/2218/commits/b70339d8aa0e9b27316d1f29378fc53058010beb> only, the four first SQL queries were following the coding style that has been suggested. The only expected change for these lines was to remove the opening double quote and the closing double quote. Merging that would not only mean this commit would be more difficult to revert, it would also require additional work to revert only the indent part. It seems it would be less dev time overall to remove these unrelated changes first.

yvanzo:
@brainzbot retest this please

yvanzo:
@brainzbot retest this please

reosarevok:

yvanzo:

yvanzo (approved):
Thanks!

mwiencek (approved):

PR #2219, opened 2021-08-17
MBS-3774: Add URL with begin and end dates

y-young:
### Implement MBS-3774

Display date period next to relationship name, add a popover for setting begin and end dates of URL, enabling date period editing for entity edit pages other than URL.

Screenshot

### DIFF
--- a/root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
+++ b/root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
@@ -1,0 +1,222 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+import mutate from 'mutate-cow';
+import ButtonPopover from '../../common/components/ButtonPopover';
+import type {LinkRelationshipT} from '../externalLinks';
+import DateRangeFieldset, {
+  partialDateFromField,
+  runReducer as runDateRangeFieldsetReducer,
+  type ActionT as DateRangeFieldsetActionT,
+} from './DateRangeFieldset';
+import {
+  createCompoundField,
+  createField,
+} from '../../edit/utility/createField';
+import {l} from '../../common/i18n';
+import {
+  applyAllPendingErrors,
+  hasSubfieldErrors,
+} from '../../../../utility/subfieldErrors';
+import {copyDatePeriodField} from '../utility/copyFieldData';
+
+type PropsT = {
+  onConfirm: (DatePeriodRoleT) => void,
+  relationship: LinkRelationshipT,
+};
+
+type StateT = {
+  +datePeriodField: DatePeriodFieldT,
+};
+
+type WritableStateT = {
+  ...StateT,
+  datePeriodField: WritableDatePeriodFieldT,
+};
+
+type ActionT =
+| {
+  +action: DateRangeFieldsetActionT,
+  +type: 'update-date-period',
+  }
+| {+type: 'reset'}
+| {+type: 'show-all-pending-errors'};
+
+const createInitialState = (props: PropsT): StateT => {
+  const relationship = props.relationship;
+  const beginDate = relationship.begin_date;
+  const endDate = relationship.end_date;
+
+  return {
+    datePeriodField: {
+      errors: [],
+      has_errors: false,
+      field: {
+        begin_date: createCompoundField(
+          'period.begin_date',
+          {
+            day: beginDate?.day ?? null,
+            month: beginDate?.month ?? null,
+            year: beginDate?.year ?? null,
+          },
+        ),
+        end_date: createCompoundField(
+          'period.end_date',
+          {
+            day: endDate?.day ?? null,
+            month: endDate?.month ?? null,
+            year: endDate?.year ?? null,
+          },
+        ),
+        ended: createField('period.ended', relationship.ended),
+      },
+      html_name: '',
+      id: 0,
+      type: 'compound_field',
+    },
+  };
+};
+
+const ExternalLinkAttributeDialog = (props: PropsT): React.MixedElement => {
+  const buttonRef = React.useRef<HTMLButtonElement | null>(null);
+  const [open, setOpen] = React.useState(false);
+  const initialState = createInitialState(props);
+
+  const resetReducer = (state: WritableStateT): void => {
+    copyDatePeriodField(initialState.datePeriodField, state.datePeriodField);
+  };
+
+  const reducer = (state: StateT, action: ActionT): StateT => {
+    return mutate<WritableStateT, StateT>(
+      state, (newState) => {
+        switch (action.type) {
+          case 'update-date-period':
+            runDateRangeFieldsetReducer(
+              newState.datePeriodField,
+              action.action,
+            );
+            break;
+          case 'reset':
+            resetReducer(newState);
+            break;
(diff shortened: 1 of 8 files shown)


reosarevok:
I'm playing with this a bit, and it seems to work well! :) Two comments, mostly visual:

1) When editing other relationships, the "ended" checkbox gets selected (and locked) when entering an end date. This makes sense, since you cannot have a relationship that has an end date yet isn't ended (even if you don't mark it yourself, it will be marked as ended on database insert of the end date):
Peek 2021-08-19 14-01

We should try to do the same here.

2) The dates are shown grayed out when ended is marked, and black when not:
Peek 2021-08-19 14-03

Showing ended ones in grey doesn't seem desirable to me, because it makes them harder to notice / read (we already have an end date or ???? to mark the endedness), but if we keep the difference, we should at least ensure that it applies to everything with an end date, since it's by definition ended (see point 1). I'd just keep it all black, personally.

root/static/scripts/edit/utility/copyFieldData.js
```diff
@@ -32,3 +32,34 @@ export default function copyFieldData<T>(
   targetField.value = sourceField.value;
   copyFieldErrors(sourceField, targetField);
 }
+
+
+export function copyPartialDateField(
+  sourceField: PartialDateFieldT,
+  targetField: WritablePartialDateFieldT,
+) {
```
mwiencek, line 40: Usually Flow requires an annotated return type for exported functions (not sure why it doesn't complain here -- maybe it can infer void ones).  But better to be explicit anyway.
```suggestion
export function copyPartialDateField(
  sourceField: PartialDateFieldT,
  targetField: WritablePartialDateFieldT,
): void {
```

root/static/scripts/edit/utility/copyFieldData.js
```diff
@@ -32,3 +32,34 @@ export default function copyFieldData<T>(
...
+  copyFieldData(sourceSubfields.year, targetSubfields.year);
+  copyFieldData(sourceSubfields.month, targetSubfields.month);
+  copyFieldData(sourceSubfields.day, targetSubfields.day);
+  copyFieldErrors(sourceField, targetField);
+}
+
+export function copyDatePeriodField(
+  sourceField: DatePeriodFieldT,
+  targetField: WritableDatePeriodFieldT,
+) {
```
mwiencek, line 52:
```suggestion
): void {
```

root/static/scripts/edit/externalLinks.js
```diff
@@ -35,6 +35,10 @@ import URLInputPopover from './components/URLInputPopover';
 import {linkTypeOptions} from './forms';
 import * as URLCleanup from './URLCleanup';
 import validation from './validation';
+import isDateEmpty from '../common/utility/isDateEmpty';
+import formatDatePeriod from '../common/utility/formatDatePeriod';
+import
+ExternalLinkAttributeDialog from './components/ExternalLinkAttributeDialog';
```
mwiencek, line 41:
```suggestion
import ExternalLinkAttributeDialog
  from './components/ExternalLinkAttributeDialog';
```

root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
```diff
@@ -0,0 +1,222 @@
...
+            break;
+          case 'show-all-pending-errors':
+            applyAllPendingErrors(newState.datePeriodField);
+            break;
+          default:
+            throw new Error('Unknown action: ' + action.type);
+        }
+      },
+    );
+  };
```
mwiencek, line 117: I would store props.relationship in state and move these functions outside the component so each render can reuse them.

root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
```diff
@@ -0,0 +1,222 @@
...
+      id: 0,
+      type: 'compound_field',
+    },
+  };
+};
+
+const ExternalLinkAttributeDialog = (props: PropsT): React.MixedElement => {
+  const buttonRef = React.useRef<HTMLButtonElement | null>(null);
+  const [open, setOpen] = React.useState(false);
+  const initialState = createInitialState(props);
```
mwiencek, line 90: can remove this once the relationship is stored in state, since you can access it from 'reset' that way
```suggestion
```

root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
```diff
@@ -0,0 +1,222 @@
...
+  datePeriodField: WritableDatePeriodFieldT,
+};
+
+type ActionT =
+| {
+  +action: DateRangeFieldsetActionT,
+  +type: 'update-date-period',
+  }
+| {+type: 'reset'}
+| {+type: 'show-all-pending-errors'};
```
mwiencek, line 50:
```suggestion
  | {
      +action: DateRangeFieldsetActionT,
      +type: 'update-date-period',
    }
  | {+type: 'reset'}
  | {+type: 'show-all-pending-errors'};
```

root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
```diff
@@ -0,0 +1,222 @@
...
+import DateRangeFieldset, {
+  partialDateFromField,
+  runReducer as runDateRangeFieldsetReducer,
+  type ActionT as DateRangeFieldsetActionT,
+} from './DateRangeFieldset';
+import {
+  createCompoundField,
+  createField,
+} from '../../edit/utility/createField';
+import {l} from '../../common/i18n';
```
mwiencek, line 23: (this is auto-imported by our webpack ProvidePlugin config)
```suggestion
```

root/static/scripts/edit/externalLinks.js
```diff
@@ -727,14 +771,28 @@ const ExternalLinkRelationship =
                 </div>}
               {link.url && !link.error && !hasUrlError &&
                 <TypeDescription type={link.type} url={link.url} />}
+              {(
+                !isDateEmpty(link.begin_date) ||
+                !isDateEmpty(link.end_date) ||
+                link.ended
+              ) &&
+                <span className="date-period">
+                  {` (${formatDatePeriod(link)})`}
```
mwiencek, line 780: you'll have to import bracketedText, but
```suggestion
                  {' '}
                  {bracketedText(formatDatePeriod(link))}
```

mwiencek (changes requested):

root/static/scripts/edit/externalLinks.js
```diff
@@ -727,14 +771,29 @@ const ExternalLinkRelationship =
...
               {link.url && !link.error && !hasUrlError &&
                 <TypeDescription type={link.type} url={link.url} />}
+              {(
+                !isDateEmpty(link.begin_date) ||
+                !isDateEmpty(link.end_date) ||
+                link.ended
+              ) &&
+                <span className="date-period">
+                  {' '}
+                  {bracketedText(formatDatePeriod(link))}
```
yvanzo, line 781: Break http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/edit with
```
Uncaught ReferenceError: bracketedText is not defined
ExternalLinkRelationship externalLinks.js:781
[...]
```

yvanzo (changes requested):
The “ended” checkbox should be automatically updated when modifying end date.

reosarevok:
@yvanzo: I have https://github.com/metabrainz/musicbrainz-server/pull/2227 for the ended checkbox.

y-young:

mwiencek (approved):

yvanzo (approved):
Works great for me.

It would be nice to have an additional checkbox in relationship dialog to set the dates for all relationships linked to the same URL at the same time. That is probably worth a separate ticket/pull request. Edit: Just agreed on IRC, then added MBS-11894.

PR #2221, opened 2021-08-17
MBS-11839: Convert Remove Relationship Attribute edit to React

reosarevok:
### Implement MBS-11839

Mostly kept the same, but added the IntentionallyRawIcon and actually used display data.

Tested by checking the edit search - we only have 32 of these and they're trivial, so.


root/edit/details/RemoveRelationshipAttribute.js
```diff
@@ -0,0 +1,65 @@
...
+        <th>{addColonText(l('Description'))}</th>
+        <td>
+          {nonEmpty(display.description)
+            ? (
+              <>
+                {display.description}
+                {rawIconSection}
+              </>
+            ) : lp('(none)', 'description')}
+          {}
```
mwiencek, line 58: Is something suppose to be here? :)
```suggestion
```

lib/MusicBrainz/Server/Edit/Relationship/RemoveLinkAttribute.pm
```diff
@@ -24,6 +25,15 @@ has '+data' => (
     ]
 );
 
+sub build_display_data {
+    my ($self, $loaded) = @_;
+
+    return {
+        description => $self->data->{description},
+        name => $self->data->{name},
+    }
```
mwiencek, line 34:
```suggestion
    };
```

PR #2223, opened 2021-08-17
MBS-11928: Do not render DBDefs.pm with consul-template

mwiencek:
# MBS-11928

# Problem

consul-template frequently writes invalid DBDefs.pm files in production (possibly due to services flapping in consul) and sometimes doesn't recover without manual intervention.  Since we're not sure why it does this, and because DBDefs.pm rarely actually changes, removing it altogether would greatly lessen the sysadmin burden of constantly dealing with 50x errors due to mysterious consul behavior.

# Solution

The production flow following this commit will be to store a DBDefs.pm for each type of container in docker-server-configs, copy them into the containers once started, run dbdefs_to_js.pl, and start either the website or webservice runit service.

Note: this is currently tested and running on test.musicbrainz.org, ~though JS/CSS is broken because I didn't build the images via Jenkins.  Working on that.~


script/update_dbdefs.sh
```diff
@@ -0,0 +1,75 @@
...
+  fi
+  HOSTS="$*"
+fi
+
+last_host=${HOSTS/* }
+
+for host in $HOSTS; do
+  echo "$host: Updating containers..."
+  ssh "$host" \
+    sudo -H -S -- /root/docker-server-configs/scripts/update_dbdefs.sh $CONTAINER_SUFFIX
```
yvanzo, line 69: From `shellcheck script/update_dbdefs.sh` (<https://www.shellcheck.net/wiki/SC2086>):

```suggestion
    sudo -H -S -- /root/docker-server-configs/scripts/update_dbdefs.sh "$CONTAINER_SUFFIX"
```

script/update_dbdefs.sh
```diff
@@ -0,0 +1,75 @@
...
+  if [[ $host != "$last_host" ]]; then
+    sleep 30
```
yvanzo, line 71: Pause for updating production containers is 30 seconds too, but we raised pause for updating beta containers to 90 seconds because there are lesser instances to serve beta. It would probably require having container suffixes as optional arguments to this script.

script/update_dbdefs.sh
```diff
@@ -0,0 +1,75 @@
...
+Environment:
+
+  LIST_METABRAINZ_HOSTS
+    Path to the script that lists MetaBrainz hosts by service.
+    Default: $LIST_METABRAINZ_HOSTS_DEFAULT
+
+  CONTAINER_SUFFIX
+    A "musicbrainz-" container suffix to limit changes to, like "json-dump"
+    or "webservice-test".
+    Default: (none)
```
yvanzo, line 38: Why is it passed through environment rather than arguments?

yvanzo (commented):
It looks fairly good, but I did not test it either.

I’m not sure why the script to update dbdefs is in this repository and not in docker server configs directly, since it is mostly useful after updating that other repository.

The removal of consul template is probably worth a ticket. It is not breaking mirrors using MusicBrainz Docker, but it may break custom deployments using Consul Template.

mwiencek:
> I’m not sure why the script to update dbdefs is in this repository and not in docker server configs directly, since it is mostly useful after updating that other repository.

Either works, I think, but I'd expect it to be in the same place as script/update_containers.sh.

script/update_dbdefs.sh
mwiencek: That's intentional because if `$CONTAINER_SUFFIX` is empty, `"$CONTAINER_SUFFIX"` will pass an argument.  But open to suggestions there.

script/update_dbdefs.sh
mwiencek: I'd prefer a single pause to keep the script simple, but I picked 30 because a `SIGHUP` on the services doesn't take as much time to recover as restarting the containers.

script/update_dbdefs.sh
mwiencek: It's optional, and the script already takes in a variable-length hosts list, so I didn't want to have to handle argument parsing.  In a situation like flipping `DB_READ_ONLY`, we'll be updating every container -- I think it's less common to have to update a single type of container, but it does happen sometimes.

yvanzo: The main point was to allow setting multiple container suffixes at once. But then you made it work by replacing `CONTAINER_SUFFIX` with `CONTAINER_SUFFIXES`. 
To improve help message and error checking, I checked them with the list of valid container suffixes, did the same for hosts. Since these are not overlapping, there is no need for separation, so both are just arguments now.

yvanzo (approved):
Thanks, it looks fine.

I added a ticket from/to the PR’s description, and pushed commits to that other PR.

I guess it should be merged at next production release time?

mwiencek:
> I guess it should be merged at next production release time?

I merged it & put in on beta for now just to give a bit more testing. Before next prod release we should remove the `^(beta|test)$` checks from the `add_dbdefs` calls (or uncomment them)  in docker-server-configs.

PR #2224, opened 2021-08-18
MBS-11875: Don't consider Braille an unlikely script

reosarevok:
### Implement MBS-11875

Braille is specifically meant to be used as a script for many languages and as such it shouldn't be considered as unlikely in combination with them.


CyberSkull:
Just in case it isn’t covered in the patch, Braille and mathematical notation is also a legitimate combination.

reosarevok:
The report only cares about some large languages (you can see the list in the code) so it was already not looking at anything like that, don't worry :)

CyberSkull:
👍🏻

PR #2225, opened 2021-08-18
MBS-11834: Convert Add Release edit to React

reosarevok:
### Implement MBS-11834

Tested with edit search, new and old (for old, you should check edits after 14459462 - earlier ones use the historic type)


root/edit/details/AddRelease.js
```diff
@@ -0,0 +1,151 @@
...
+            <th>{addColonText(l('Artist'))}</th>
+            <td>
+              <ExpandedArtistCredit
+                artistCredit={display.artist_credit}
+              />
+            </td>
+          </tr>
+
+          <tr>
+            <th>{addColonText(l('Release Group'))}</th>
```
mwiencek, line 82: The old template has group lowercased (Release group), which is consistent with "Release events"

reosarevok: I copied this from AddReleaseGroup, but most edits seem to use "Release group" so I changed the three that did not in a second commit.

mwiencek (approved):
Thanks!

PR #2227, opened 2021-08-19
MBS-11888: Automatically select / disable ended in DateRangeFieldset

reosarevok:
### Implement MBS-11888

We do this on relationship editor date fields, but not here. It seems entirely sensible and shows the editor that ended is automatically derived from having an end date anyway.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2219 to make sure it works there too.

Tested on an Edit alias page and when editing URL dates in the new URL editor.


y-young:

root/static/scripts/common/utility/isDateEmpty.js
```diff
@@ -7,12 +7,20 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+declare type PartialDateObservableT = {
+  +day: KnockoutObservable<string | null>,
+  +month: KnockoutObservable<string | null>,
+  +year: KnockoutObservable<string | null>,
+};
+
+declare type PartialDateStringT = {
```
mwiencek, line 16: I feel these type names are slightly more clear plural since the types themselves aren't strings/observables, the contained values are.
```suggestion
declare type PartialDateObservablesT = {
  +day: KnockoutObservable<string | null>,
  +month: KnockoutObservable<string | null>,
  +year: KnockoutObservable<string | null>,
};

declare type PartialDateStringsT = {
```

mwiencek (approved):

PR #2228, opened 2021-08-19
MBS-11890: Beta: Fix alias "guess sort name" for persons

mwiencek:
Reverts the parts of 5f1b33c that removed the typeID, which is needed for person sort names to be guessed correctly.

### DIFF
--- a/root/static/scripts/guess-case/MB/GuessCase/Main.js
+++ b/root/static/scripts/guess-case/MB/GuessCase/Main.js
@@ -28,7 +28,7 @@ class GuessCase {
   entities: {
     [entityType: string]: {
       guess: (string) => string,
-      sortname: (string, typeId?: ?number) => string,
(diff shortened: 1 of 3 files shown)


mwiencek:
The `Controller::Account` test has been flaky lately.  Not sure why.  That failure is unrelated to the PR, in any case.

root/static/scripts/guess-case/MB/GuessCase/Main.js
```diff
@@ -28,7 +28,7 @@ class GuessCase {
   entities: {
     [entityType: string]: {
       guess: (string) => string,
-      sortname: (string) => string,
+      sortname: (string, typeId?: ?number) => string,
```
reosarevok, line 31: I'm confused about this - seems only one guessSortName (artist) uses an extra parameter, and it's a boolean. That's why I removed this, because it seemed to be unused. What am I missing?

PR #2234, opened 2021-08-23
MBS-11841: Convert Set Track Lengths edit to React

reosarevok:
### Implement MBS-11841

There are three different kinds of edits here: one with only one medium, one with several releases based on the now deprecated concepted of the shared tracklist, and one with several releases based on one pre-NGS release. As such, the template needs to somehow support them all.

The track_duration_changes macro is a mess, this tries to simplify it. It will need to be passed pre-calculated arrays of lengths rather than a prop to access the length once we use it elsewhere too.

As usual, tested with oldest and newest edits, but also with oldest edits after ID 14460214 (to find the intermediate case based on shared tracklists).


reosarevok:

mwiencek (approved):

PR #2235, opened 2021-08-24
MBS-9454: Autoremove unused urls with daily.sh

reosarevok:
### Fix MBS-9454

While most of these get removed with triggers, in some cases (e.g. having open edits) unused URLs can remain behind as orphans. This ensures the daily cron will remove them like other unused entities. Since we don't usually leave edits when URLs are autoremoved by triggers, we also don't create an edit here.

Tested by running it locally and seeing it did remove one URL (in test data but it might have been one that I accidentally orphaned), at least. It did not remove the URLs I just inserted to test, which makes sense because it does have a "last_updated over a day ago" check...

### DIFF
--- a/admin/cron/daily.sh
+++ b/admin/cron/daily.sh
@@ -50,6 +50,9 @@ echo `date`" : Removing unused release groups"
 echo `date`" : Removing unused series"
 ./admin/cleanup/RemoveEmpty series
 
+echo `date`" : Removing unused urls"
+./admin/cleanup/RemoveEmpty url
+
 echo `date`" : Removing unused works"
 ./admin/cleanup/RemoveEmpty work
 
--- a/entities.json
+++ b/entities.json
@@ -719,6 +719,9 @@
         "model": "URL",
         "plural": "urls",
         "plural_url": "urls",
+        "removal": {
+            "automatic": {}
+        },
         "url": "url"
     },
     "work": {
--- a/lib/MusicBrainz/Script/RemoveEmpty.pm
+++ b/lib/MusicBrainz/Script/RemoveEmpty.pm
@@ -116,33 +116,40 @@ sub run {
 
         }
         else {
-            Sql::run_in_transaction(sub {
-                my $edit = $self->c->model('Edit')->create(
-                    edit_type => $edit_class{$entity},
-                    to_delete => $e,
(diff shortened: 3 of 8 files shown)


yvanzo (commented):
> Since we don't usually leave edits when URLs are autoremoved by triggers, we also don't create an edit here.

Is there a specific reason for not entering an edit to autoremove URLs in the first place?

> Tested by running it locally and seeing it did remove one URL (in test data but it might have been one that I accidentally orphaned), at least. It did not remove the URLs I just inserted to test, which makes sense because it does have a "last_updated over a day ago" check...

Does it skip removing URLs with pending edits too?

Having this script to perform destructive actions without even an edit seems worth having some test for it at `t/script/RemoveEmpty.t`.

reosarevok:
> Is there a specific reason for not entering an edit to autoremove URLs in the first place?

I guess mostly because we don't usually see URLs as similarly "real" - and AFAICT, since they're only removed when empty, those edits wouldn't be found from anywhere (other than, I guess, a specific edit search). IIUC there has never been a "Remove URL" edit at all.

> Does it skip removing URLs with pending edits too?

It uses the same query as other entities, so yes (https://github.com/metabrainz/musicbrainz-server/blob/78be844e46d921a370add2348e3f46023825d5c3/lib/MusicBrainz/Script/RemoveEmpty.pm#L83-L98)

> seems worth having some test for it at t/script/RemoveEmpty.t.

Sounds sensible to me, will look into that.

reosarevok:
Added a test with c90fb4f12fe964e92d04c8ea156667407ceb3412 - I put it at `t/lib/t/MusicBrainz/Script/RemoveEmptyURLs.pm` because that folder is where we were already testing `RebuildCoverArt` which is on the same folder as RemoveEmpty, hopefully that makes sense?

yvanzo (approved):
Yes, thank you.

mwiencek:
There's a test failure in `t::MusicBrainz::Server::Controller::Artist::EditExternalLinks`, can you check?

mwiencek:
@brainzbot test this please

reosarevok:
Yeah, sigh, an ID changed because of my addition to `url.sql` and broke the test. Working now.

mwiencek (approved):

PR #2236, opened 2021-08-24
MBS-10639: Convert release merge edit to React

reosarevok:
### Implement MBS-10639

Everything uses display_data now. Also, this simplifies emptyReleases, since there's no point having it as an array of dicts that only contain one key.

Tested on newest and oldest (after edit 14459468 - older are historical). Since I didn't easily find a merge edit with an empty medium nor a multi-release merge with more than one recording being merged for the same track I just generated those locally, seemed to work fine. I haven't explicitly tried to find a "version 2" edit to test though.


lib/MusicBrainz/Server/Edit/Generic/Merge.pm
```diff
@@ -65,16 +65,36 @@ sub build_display_data
...
     my $model = $self->_merge_model;
 
-    my $new_entity = (
+    # For release merges, we need to load additional data
+    # before turning the releases to JSON
+    if ($model eq 'Release') {
+        $self->c->model('Label')->load(
+            grep { $_->label_id && !defined($_->label) }
+            map { $_->all_labels }
+            values %{ $loaded->{Release} }
```
mwiencek, line 74: Can you store this in a `my @releases = values %{ $loaded->{Release} };` first? (Since it's used four times.)

root/edit/details/MergeReleases.js
```diff
@@ -0,0 +1,363 @@
...
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+
```
mwiencek, line 12:
```suggestion
```

mwiencek (approved):

mwiencek:
Pre-approved but left two nitpicky comments.

mwiencek:
The selenium failure might be legitimate, since it times out at `assertTextMatches target="css=p.merge-error" value="A merge cycle exists"`.  Is it not displaying `cannot_merge_recordings_reason` correctly?

reosarevok:
Good catch - it was in fact not displaying that. While checking that, I also found another, pre-existing bug where the errors weren't displayed correctly in the merge queue page either. Will submit the fix for that on a separate commit.

PR #2237, opened 2021-08-25
Updated favicons

akshaaatt:
The favicons currently used don't showcase the actual logo. This feels better for usage and is more accurate.


akshaaatt:
Discarding this PR based on the point that we have come to a mutual understanding of how we are gonna manage the icons as is which perfectly makes sense :)

PR #2241, opened 2021-08-30
Improvements to Edit Release Label edits

reosarevok:
This hides "Date:" if there's no date at all to show (which was meant to happen already, but we were passing an empty string as the date in that situation), fixes a regression where we wouldn't show the label diff if the label was being removed (because it was detecting a new label being missing as the label just not changing), and fixes hiding the label and catno sections when they are empty, rather than just unchanged.


mwiencek (approved):

mwiencek:
@brainzbot test this please

PR #2245, opened 2021-09-01
MBS-11932: Also exclude work licenses from LinksWithMultipleEntities

reosarevok:
### Fix MBS-11932

I forgot these existed when I wrote the report, whoops.


yvanzo (approved):
Added comments to the SQL query to identify all relationship UUIDs.

PR #2246, opened 2021-09-01
MBS-11852: Only call TO_JSON on editor if it exists

reosarevok:
### Fix MBS-11852

When trying to approve an edit with a No vote + note, we load edit notes for the edit (to see if one of the editor_ids matches the approver's) but we do not load the actual editors themselves.
On doing TO_JSON for edit when passing it to NoteIsRequired, we were (correctly) running to_json_array on the edit notes, and this was running TO_JSON directly on the note's editor, which, of course, had not been loaded. This changes it to run to_json_object, which is smarter and actually ensures TO_JSON can be called before calling it.

To test, create an open edit with one editor, vote No and leave a note with a second editor, and then try and approve the edit with a third editor. Of course, editor 2 will need to be allowed to vote/leave notes (so, not a beginner) which can be annoying to trigger with sample data, but certainly doable (worst case, just do what I did and enter 10 open edits, then accept them).

### DIFF
--- a/lib/MusicBrainz/Server/Entity/EditNote.pm
+++ b/lib/MusicBrainz/Server/Entity/EditNote.pm
@@ -9,6 +9,7 @@ use namespace::autoclean;
 use MusicBrainz::Server::Constants qw( $EDITOR_MODBOT );
 use MusicBrainz::Server::Data::Utils qw( datetime_to_iso8601 );
 use MusicBrainz::Server::Entity::Types;
+use MusicBrainz::Server::Entity::Util::JSON qw( to_json_object );
 use MusicBrainz::Server::Filters qw( format_editnote );
 use MusicBrainz::Server::Types qw( DateTime );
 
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Entity/EditNote.pm
```diff
@@ -125,7 +126,7 @@ around TO_JSON => sub {
 
     my $json = $self->$orig;
     $json->{editor_id} = $self->editor_id + 0;
-    $json->{editor} = $self->editor->TO_JSON;
+    $json->{editor} = to_json_object($self->editor);
```
mwiencek, line 129: I guess we should change `EditNoteT` to use `+editor: EditorT | null,` then (it'll be loaded in most cases, but).

reosarevok: Alternatively, we could load the editor here, but I guess it's not particularly useful just for consistency and having a check if needed on the React side is better.

mwiencek: If the editor doesn't need to be loaded there I guess I'd prefer changing the type and seeing how much Flow likes that -- but if it causes a million errors we could always-load if it's more convenient.

reosarevok: Didn't seem too problematic, made that change :)

mwiencek (approved):
thanks!

yvanzo (approved):
LGTM but did not test.

PR #2247, opened 2021-09-01
MBS-11931, MBS-11929, MBS-11927: Beta fixes for URL editing

y-young:
## Fix:
- MBS-11931: Beta: Link type option "Discogs" appears under the "get music" heading
- MBS-11929: Beta: Untouched URL missing a relationship type may block editing
- MBS-11927: Beta: URL relationship editor doesn't display invalid relationship type

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -883,12 +883,13 @@ const ExternalLinkRelationship =
                       {props.typeOptions.map((option, index) => {
                         const nextOption = props.typeOptions[index + 1];
                         if (!option.disabled ||
-                          // Ignore empty groups
+                          /*
+                           * Ignore empty groups by checking
+                           * if the next option is an item in current group,
+                           * if not, then it's an empty group.
+                           */
                           (nextOption &&
-                            // Is not a group delimiter
-                            !nextOption.disabled &&
-                            // Is an item in group
-                            nextOption.text !== nextOption.text.trim())) {
+                            nextOption.data.parent_id === option.value)) {
                           return (
                             <option
                               disabled={option.disabled}
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -95,10 +95,13 @@ export class ExternalLinksEditor
 
   generalLinkTypes: $ReadOnlyArray<LinkTypeOptionT>;
 
+  oldLinks: LinkMapT;
+
   constructor(props: LinksEditorProps) {
     super(props);
     this.state = {links: withOneEmptyLink(props.initialLinks)};
     this.tableRef = React.createRef();
+    this.oldLinks = this.getOldLinksHash();
     this.generalLinkTypes = props.typeOptions.filter(
       // Keep disabled options for grouping
       (option) => option.disabled ||
@@ -512,7 +515,6 @@ export class ExternalLinksEditor
     link: LinkStateT,
     checker?: URLCleanup.Checker,
   ): ErrorT | null {
-    const oldLinks = this.getOldLinksHash();
     const linksByTypeAndUrl = groupBy(
       uniqBy(
         this.state.links.concat(this.props.initialLinks),
@@ -527,7 +529,7 @@ export class ExternalLinksEditor
     // Use existing checker if possible, otherwise create a new one
     checker = checker ||
       new URLCleanup.Checker(link.url, this.props.sourceType);
-    const oldLink = oldLinks.get(String(link.relationship));
+    const oldLink = this.oldLinks.get(String(link.relationship));
     const isNewLink = !isPositiveInteger(link.relationship);
     const linkChanged = oldLink && link.url !== oldLink.url;
     const isNewOrChangedLink = (isNewLink || linkChanged);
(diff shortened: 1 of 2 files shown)


y-young:

y-young:

y-young:

yvanzo:
I changed the title so that our bot forwards status changes to the respective tickets; See [CONTRIBUTING](https://github.com/metabrainz/musicbrainz-server/blob/v-2021-07-19/CONTRIBUTING.md#title).

root/static/scripts/edit/externalLinks.js
```diff
@@ -508,11 +508,15 @@ export class ExternalLinksEditor
     }
   }
 
+  getOldLink(relationship: StrOrNum | null): ?LinkStateT {
+    const oldLinks = this.getOldLinksHash();
+    return oldLinks.get(String(relationship));
+  }
```
mwiencek, line 514: I'd prefer not to call `getOldLinksHash` in here if possible, since it has to recompute it each time this is called, which appears to be inside a loop (`links.some`).

root/static/scripts/edit/externalLinks.js
```diff
@@ -1261,7 +1272,7 @@ export function parseRelationships(
         rawUrl: target.name,
         relationship: data.id,
         submitted: true,
-        type: data.linkTypeID ?? null,
+        type: parseInt(data.linkTypeID, 10) ?? null,
```
mwiencek, line 1275: So the server is sending these as strings?  Seems like something we should fix on the server.

reosarevok (approved):
Tested on pink data, seems to work fine.

mwiencek (commented):
For MBS-11927, if I deselect Allmusic then it disappears or turns into a blank option, which seems kind of odd (I'd expect to be able to undo the change so I could submit the form).  Maybe it's fine for now though.

root/static/scripts/edit/externalLinks.js
y-young: I found this when I was testing seedings, they're not parsed to int after submitting.

reosarevok:
That's a fair point about not having a way to "untouch" the invalid link if you touch it. If it's easy to change, I'd suggest to do so, but if not, this is already an improvement so we can start with it.

root/static/scripts/edit/externalLinks.js
y-young: This is how it was written before. Do you mean we should simply `find` the old link by `relationship`?

reosarevok: If we can load oldLinks once, then just use the same oldLinks constant repeatedly, that would seem ideal. I assume oldLinks do not change anyway?

mwiencek: It was bad before too, and now it's done in two places. :)

Using `find` could work, or possibly caching `oldLinks` in the constructor.

root/static/scripts/edit/externalLinks.js
mwiencek: I'd guess we should change https://github.com/metabrainz/musicbrainz-server/blob/0ec0ddc/lib/MusicBrainz/Server/Controller/ReleaseEditor.pm#L606 to `$id + 0` so Perl treats it as a number.  If you can revert this & test that I'd prefer making the change on the server, because `linkTypeID` should never be a string according to its Flow type -- that's a server bug.

y-young:
> For [MBS-11927](https://tickets.metabrainz.org/browse/MBS-11927), if I deselect Allmusic then it disappears or turns into a blank option, which seems kind of odd (I'd expect to be able to undo the change so I could submit the form). Maybe it's fine for now though.

I'm afraid it's not easy to implement, after changing link type the possible options change, and the old type is lost.

root/static/scripts/edit/externalLinks.js
y-young: I'll go with caching, it would be useful for implementing highlights in the future.

root/static/scripts/edit/externalLinks.js
y-young: It is in the release editor.

root/static/scripts/edit/externalLinks.js
mwiencek: Yeah that should be fine if `this.props.initialLinks` is constant.  Just not if the component renders again with different a different `initialLinks` prop, since the constructor is only run when the component mounts.

root/static/scripts/edit/externalLinks.js
mwiencek: Looks like in that case it's being parsed from the URL query string, so it's expected to be a string.

y-young:

mwiencek (approved):
thanks! looks better now

yvanzo (commented):
* Successfully tested the fix to MBS-11931 by pasting an existing Discogs link again and looking at the selector options.
* Successfully tested the fix to MBS-11929 by editing a release already linked to a Jamendo URL through “download for free" relationship type only.
* I tested the fix to MBS-11927 by replacing a Discogs album URL linked to a MusicBrainz release through a “Discogs” relationship with other URLs (still linked to the same release). I noticed that if the URL has no validation rule, the bug is resolved. But when the URL has another validation rule (for example a Deezer artist URL), then it is possible to select any relationship type without an error being thrown. This is not neither neither a blocking bug nor a destructive bug, it just allows to set an invalid relationship type instead of another already invalid relationship type. Seems to be worth fixing though.

yvanzo (approved):
Sorry, it is an existing bug, not a regression, I just tested with `v-2021-07-19` to make sure. Will open a ticket. Thank you!

PR #2248, opened 2021-09-03
dbmirror2 support

mwiencek:
Schema change ticket: https://tickets.metabrainz.org/browse/MBS-12250

dbmirror2 is a replacement for dbmirror written in PL/pgSQL: https://github.com/metabrainz/dbmirror2

I started writing it in 2017 when [replication suddenly broke](https://blog.metabrainz.org/2017/06/06/broken-replication-packet-fix-104949/).  Although we managed to fix the broken packet and worked around what triggered it in musicbrainz-server, the underlying cause still exists in dbmirror.  (It's a difficult issue to fix, especially in C; you can find more info about it [here](https://github.com/metabrainz/dbmirror2/blob/master/test/cascading_triggers/BUG.md).) Since it was no longer an urgent concern, I postponed work on it.  Later, when I started working on improvements to JSON dumps, I realized it would be help a lot if replication packets stored more data, including the entire OLD row for updates so it could be determined which columns changed in an update. 

More recently, I was thinking about how we could take advantage of this extra data in sir. Right now, sir operates by generating a ton of [functions](https://github.com/metabrainz/sir/blob/master/sql/CreateFunctions.sql) and triggers that push changes to rabbitmq. (This setup causes problems in certain environments: there is an [old PR](https://github.com/metabrainz/sir/pull/108) to work around the pg_amqp requirement.) These triggers essentially overlap with the recordchange triggers we already have for dbmirror, but as mentioned, we aren't always logging the OLD data required by sir. (Also note that [even the existing triggers have issues](https://tickets.metabrainz.org/browse/SEARCH-609) with how they filter events.) If we did, as dbmirror2 does, we can simplify sir's architecture. For example, we can remove all of sir's existing triggers, and have only one trigger on `dbmirror2.pending_data` that copies the changes to another table which would act as our queue.

The plan for migrating to this would probably look something like this:
 * Install the dbmirror2 schema in production
 * Generate '-v2' packets alongside the current ones for a while
 * Test the new packets on our own slave servers
 * Create the dbmirror2 schema on all slaves in a future schema change release
 * Prepare mbslave to work with them
 * Have people in the community test the '-v2' packets on their slave instances
 * Make them the default once we're confident enough in them


reosarevok:
The general idea seems good to me, but we really need to test this properly - can we put it on test and try to generate some updates first? Basically, I'm happy with putting it in prod and seeing what happens as soon as we know it's not going to interfere with the dbmirror(1) versions - we can always improve the results as we find issues since we're not going to start using it just yet.

mwiencek:
Definitely, my next step is to set this up on test.mb (alongside dbmirror(1)) and have another slave server replicating from that.

mwiencek:
This is running on test.mb (so please don't update that container for now :)) alongside the old dbmirror. The slave database is musicbrainz_schema_change_20210513_slave (also on aretha) and the logs are inside the musicbrainz-website-test container at /home/musicbrainz/log (cron is running inside this container.)

reosarevok:
I understand this has been tested for quite a while in test and it seems to work fine? Do we need anything else before merging?

mwiencek:
> I understand this has been tested for quite a while in test and it seems to work fine? Do we need anything else before merging?

I added a commit ("Refresh dbmirror2.column_info in ugprade.sh") but it should otherwise be ready.

reosarevok:
I'm a bit confused - we have a CreateReplicationTriggers2 file, but we are not generating a DropReplicationTriggers2 file to go with it. Is there a reason for it or did we just forget?

PR #2249, opened 2021-09-03
MBS-11907: Block vyd.co link aggregator

reosarevok:
### Implement MBS-11907


yvanzo (approved):
Thanks! Small enough to not require a second review for merge.

PR #2250, opened 2021-09-03
MBS-11819: Show to admins whether account has oAuth uses

reosarevok:
### Implement MBS-11819

As an admin, it's useful to know how an account has been used before deleting it. If no oAuth authorizations nor applications exist, then there's no worries that deleting it might cause issues elsewhere nor that there's spam somewhere else we have not yet seen.

Ideally we would also link to the user's BB, LB and CB profiles, at least, but right now that's not trivial at all since for example BB does not currently allow finding users by MB username nor MB user id.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User.pm
+++ b/lib/MusicBrainz/Server/Controller/User.pm
@@ -441,6 +441,10 @@ sub profile : Chained('load') PathPart('') HiddenOnSlaves
     $c->stash->{subscriber_count} = $subscr_model->get_subscribed_editor_count($user->id);
     $c->stash->{votes}            = $c->model('Vote')->editor_statistics($user);
 
+    my ($tokens, $token_count) = $c->model('EditorOAuthToken')->find_granted_by_editor($user->id);
+
+    my ($applications, $application_count) = $c->model('Application')->find_by_owner($user->id);
+
     $c->model('Gender')->load($user);
     $c->model('EditorLanguage')->load_for_editor($user);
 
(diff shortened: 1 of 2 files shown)


alastair:
Why do you think that it's not trivial to link to user accounts on other sites? LB user accounts use the musicbrainz login name in the URL. We're planning on adding this URL structure for CB users too. You could special-case the prod application key for these apps for this case, without making an external webservice call.

reosarevok:
@alastair: currently LB does not change the name if the MB username gets changed, so even that can get out of sync. That said, it's mostly safe I guess - if we can assume that a LB account exists if there's a specific oAuth authorisation, at least.

BB does not allow anything like this at all at the moment, and they don't even enforce the same usernames as in MB. I guess neither does CB yet (how does CB deal with user renames?)

yvanzo (commented):
I’m not sure how to test this using sample data, and I could not test this using the production read-only database since I’m not an admin. There is no automated test under `t/` either. Has it been tested and how?

reosarevok:
I checked my profile and a couple others users' and saw there were numbers that seemed sensible there. You should be able to at least test it on your own profile with prod data :)

mwiencek (approved):

alastair:
@reosarevok right, I guess you're correct here - in the future it will be possible to link, but right at the moment there are limitations in both LB and CB that mean that it's not possible.
When we add this functionality we'll let you know and we can add the links in here too.

yvanzo:
@reosarevok, this error show up when visiting http://localhost:5000/user/yvanzo (logged in or not) (rebased on master or not):
```
42501 DBD::Pg::st execute failed: ERROR:  permission denied for table place_rating_raw [for Statement "SELECT count(*) FROM (SELECT 1 FROM place_rating_raw WHERE editor = ? UNION ALL SELECT 1 FROM label_rating_raw WHERE editor = ? UNION ALL SELECT 1 FROM artist_rating_raw WHERE editor = ? UNION ALL SELECT 1 FROM work_rating_raw WHERE editor = ? UNION ALL SELECT 1 FROM recording_rating_raw WHERE editor = ? UNION ALL SELECT 1 FROM event_rating_raw WHERE editor = ? UNION ALL SELECT 1 FROM release_group_rating_raw WHERE editor = ?) x"
```

reosarevok:
Hmm. That seems weird in that I don't think anything here should affect `place_rating_raw` (I'm not getting this locally either). That's quite a new table, can it be that something weird happened with your local setup? (unless you're using a tunnel to pink, in which case I'm even more confused)

yvanzo:
I’m accessing the database `musicbrainz_db` on the server `pink` using the username `musicbrainz_ro`.
The issue might be the table `place_rating_raw` doesn’t have proper permissions set for that user.

mwiencek:
> The issue might be the table place_rating_raw doesn’t have proper permissions set for that user.

That sounds correct to me.  We're not very good about keeping the permissions for that in-sync during schema changes (mainly because it's a production-only user, so a bit disconnected from the MBS upgrade scripts, which are meant to apply to all mirrors).  Something we should improve on, probably in a schema-change checklist somewhere...

yvanzo (approved):
Have been able to test as a unprivileged editor, and as a not logged-in user. Did not test as admin.

Added MBS-11962 for the issue about updating `musicbrainz_ro` privileges.

PR #2251, opened 2021-09-03
Allow disabling the template renderer

mwiencek:
If `RENDERER_SOCKET` is empty, MBS will now avoid communicating with the template renderer at all. This will prevent socket timeouts ("Couldn't connect to the renderer") from occurring if certain non-WS requests leak into our WS containers in production (as these containers don't have any renderer process).

A blank response is returned if this happens, because it's not an error that's even intended to be /allowed/ to be triggered normally (and I don't suspect anyone will use this flag outside of us), but I improved the logging to include the URL that triggered it. (You can already find which URLs are causing this in Sentry.)

### DIFF
--- a/lib/DBDefs/Default.pm
+++ b/lib/DBDefs/Default.pm
@@ -407,6 +407,11 @@ sub HTML_VALIDATOR { 'http://validator.w3.org/nu/?out=json' }
 # We use a small Node.js server (root/server.js) to render React.js
 # templates. RENDERER_SOCKET configures the local (UNIX) socket path it
 # listens on.
+#
+# If this is set to '' or undef, MBS will not try to communicate with the
+# template-renderer. This can be useful if you only make web service requests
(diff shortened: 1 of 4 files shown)


reosarevok:
How does this even happen???

mwiencek:
Basically it's through the subdomain/vhost we have set in the openresty config.  It points directly to the WS containers and let's you make any sort of request to them.  (I already mentioned it to Zas, so not sure if there's an easy way to avoid it.)

BTW, we could also just make it so that setting `RENDERER_SOCKET` to an empty string or undef disables it, instead of adding a new option.  That might be preferable.

reosarevok (approved):
The general idea seems fine, anyway - if you prefer to just check whether a RENDERER_SOCKET is not undef, that's also fine I guess.

reosarevok:

PR #2252, opened 2021-09-06
MBS-11866: Interface for mass deletion of accounts [WIP]

reosarevok:
### Implement MBS-11866

Account admins could sometimes use a way to remove a lot of accounts doing the same thing (for example, spammily rating or tagging an entity). This allows them to do that, by giving a bunch of editor names. This gets turned into a list of editors with more data, in order to check that the editors should in fact get deleted, after which the account admin can proceed to the deletion.

This does not allow account name reuse because from talking to Freso, it's almost never desired for a case like this, and if it ever is then the usernames can be unlocked elsewhere in the admin tools.

Draft for now since we should ideally display more data - the ticket asks for stuff like number of edits, tags, ratings. We have a way to load this for one editor for UserProfile, but should we do the same here and load all stats for all editors, then take the wanted bits? Or should we write a different method to only get the specific stuff we want?


reosarevok:
This needs a lot of work and tests to be ready, but I'm not sure it makes sense anymore with accounts moving to MeB in the medium term. Also, the main use we had for it (deleting spammers) is no longer relevant since we mark them as such instead.

It is possible that parts of this can be repurposed as a "mass-mark as spam" tool, which certainly could be of use sometimes, although most of the time it might be enough to have a one-click "mark as spam" button for users, and in any case right now we still have to do other things (like deleting their forum profiles) separately by hand.

PR #2253, opened 2021-09-06
MBS-11942: Deleted users can still use pending verification links

reosarevok:
### Fix MBS-11942

If a user gets deleted after they get send a verification link, but before they click on it, the email is added to the now-deleted user. This is certainly not intended, so this checks the user is not deleted before actually updating anything.

This isn't tested because I can't actually send verification emails from my local system, but I don't see why it wouldn't work. Scream if I did a dumb.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Account.pm
+++ b/lib/MusicBrainz/Server/Controller/Account.pm
@@ -126,6 +126,10 @@ sub verify_email : Path('/verify-email') ForbiddenOnSlaves DenyWhenReadonly
         $c->detach;
     }
 
+    if ($editor->deleted) {
+        $c->detach('/user/not_found');
+    }
+
     $c->model('Editor')->update_email($editor, $email);
 
     if ($c->user_exists) {


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -286,21 +286,25 @@ sub update_email
...
-            $self->sql->do('UPDATE editor SET email=NULL, email_confirm_date=NULL
-                      WHERE id=?', $editor->id);
-            delete $editor->{email};
-            delete $editor->{email_confirmation_date};
-        }
-    }, $self->sql);
+    if ($editor->deleted) {
+        $c->detach('/user/not_found');
+    } else {
+        Sql::run_in_transaction(sub {
```
mwiencek, line 292: No need to put this in an `else` actually since `$c->detach` always dies.  Would make the diff more readable. :)

reosarevok:
This was originally added to Data::Email, but that was dumb - one, there's no $c->detach there, and two, the problem is only for verification, so we should just detach in the verification code, before using `update_email` at all.

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -303,7 +303,7 @@ sub update_email
         }
         else {
             $self->sql->do('UPDATE editor SET email=NULL, email_confirm_date=NULL
-                      WHERE id=?', $editor->id);
+                    WHERE id=?', $editor->id);
```
mwiencek, line 306: intentional?  (doesn't appear to align things any better, not that it was good before.)

mwiencek (approved):

lib/MusicBrainz/Server/Data/Editor.pm
reosarevok: Nope, reverting.

PR #2254, opened 2021-09-07
MBS-11944: Fix label on button for place merges

reosarevok:
### Fix MBS-11944

... good job, me from the past :)


mwiencek (approved):

yvanzo (approved):
... good review too, me from the past :)

PR #2255, opened 2021-09-07
MBS-10902: Stop changing zxx language name at the Perl level

reosarevok:
### Fix MBS-10902

This was causing issues when both work and release edits were being loaded, in that releases would appear with the language
[No lyrics]. Now that all relevant edits are React anyway, there's no need to massage the work on the Perl level at all. Instead, this just passes the languages (rather than their names) to JS and lets localizeLanguageName change the displayed name only.

Tested this by visiting add work / edit work edits, work searches, artists' work pages.

I didn't change the calls for l_languages for release/stats stuff right now, although I guess we could if that seems more consistent / better :)

### DIFF
--- a/root/static/scripts/common/i18n/localizeLanguageName.js
+++ b/root/static/scripts/common/i18n/localizeLanguageName.js
@@ -1,0 +1,24 @@
+/*
+ * @flow strict
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+function localizeLanguageName(
+  language: LanguageT | null,
+  isWork?: boolean = false,
+): string {
+  if (!language) {
+    return l('[removed]');
+  }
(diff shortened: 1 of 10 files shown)


root/static/scripts/common/i18n/localizeLanguageName.js
```diff
@@ -0,0 +1,24 @@
...
+
+function localizeLanguageName(
+  language: LanguageT | null,
+  isWork?: boolean,
+): string {
+  if (!language) {
+    return l('[removed]');
+  }
+  // For works, "No linguistic content" is meant as "No lyrics"
+  if (isWork /*:: === true */ && language.iso_code_3 === 'zxx') {
```
mwiencek, line 18: You can default `isWork` to false so that it's not undefined if not specified:

```suggestion
  isWork?: boolean = false,
): string {
  if (!language) {
    return l('[removed]');
  }
  // For works, "No linguistic content" is meant as "No lyrics"
  if (isWork && language.iso_code_3 === 'zxx') {
```

reosarevok: True, true, I keep forgetting about this about half the time, sigh! Done :)

yvanzo (approved):

mwiencek (approved):

PR #2256, opened 2021-09-08
MBS-11948: More strict favicon class assignment

reosarevok:
### Fix MBS-11948

We have had several issues in the past of favicons being assigned incorrectly. The one triggering this issue was kenjimiura.com
being matched with the residentadvisor "ra.co" string.

There's no reason why we should stick to this naive approach to matching when we have regular expressions and URL objects, so this makes use of both to actually validate the full hostnames of URLs and, when needed, the start of their paths.

The regexes are mostly based on the existing matchup/cleanup we do. The few we don't validate/clean up at all I based on what the structure seems to be via a quick database check, although admittedly I might have missed some edge cases.

As an additional bonus, I added testing for the favicon classes, to make sure that at least all the URLs we were already testing
do match the expected favicon class (that should also serve as a reminder to change the regexes there if we change how a site is cleaned up).


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -3392,34 +3784,39 @@ const testData = [
...
   {
                      input_url: 'https://ra.co/tracks/544258',
              input_entity_type: 'recording',
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://ra.co/tracks/544258',
        only_valid_entity_types: ['recording'],
+        expected_favicon_class: 'residentadvisor-favicon',
   },
   {
                      input_url: 'http://www.ra.co/reviews/7636',
```
yvanzo, line 3822: Maybe test RA reviews’ favicon too?

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -3435,56 +3832,65 @@ const testData = [
...
+        expected_favicon_class: 'residentadvisor-favicon',
   },
   {
                      input_url: 'http://www.ra.co/exchange/552',
              input_entity_type: 'release',
     expected_relationship_type: 'shownotes',
             expected_clean_url: 'https://ra.co/exchange/552',
        only_valid_entity_types: ['release'],
+        expected_favicon_class: 'residentadvisor-favicon',
   },
```
yvanzo, line 3844: I suggest testing the reported URL’s favicon for that bug:

```suggestion
  {
                     input_url: 'http://www.ra.co/exchange/552',
             input_entity_type: 'release',
    expected_relationship_type: 'shownotes',
            expected_clean_url: 'https://ra.co/exchange/552',
       only_valid_entity_types: ['release'],
        expected_favicon_class: 'residentadvisor-favicon',
  },
  {
                     input_url: 'https://www.kenjimiura.com/',
             input_entity_type: 'artist',
    expected_relationship_type: undefined,
            expected_clean_url: 'https://www.kenjimiura.com/',
        expected_favicon_class: 'no-favicon',
  },
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4848,6 +5405,11 @@ testData.forEach(function (subtest, i) {
       }
       tested = true;
     }
+    if (subtest.expected_favicon_class) {
+      const actualFaviconClass = getFaviconClass(actualCleanUrl);
+      st.equal(actualFaviconClass, subtest.expected_favicon_class, 'Correct favicon');
```
yvanzo, line 5410: Should it test the input URL too? For existing unclean URLs.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4848,6 +5405,11 @@ testData.forEach(function (subtest, i) {
...
+      tested = true;
```
yvanzo, line 5411: Does that imply that a test can just consist of `input_url` and `expected_favicon_class` ?
If it is tested against the clean URL, `expected_clean_url` should be mandatory.
Maybe the documentation about test properties in dcabded48be065374c55bfb789ad588daa363729 commit’s message should be added to this file?

yvanzo (commented):
It looks very nice overall.

* If @mwiencek confirms `URL` usage is appropriate here, it is probably not useful to add unit tests for the `isValidURL` function as its new implementation seems to be browser-dependent.
* MBS-11948’s implementation looks nice, and I did check some favicon classes but not all of them since you added unit tests for these and any regression here would not be blocking editors. My only comments are about tests’ implementation.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Yes, IMO that should be ok, shouldn't it? It's tested against the return of cleanURL, which is the original if there's no cleanup function, so effectively, it is tested against whatever will actually be added to the DB.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: I was expecting we wouldn't necessarily care if the favicon is not shown for existing URLs that need fixing anyway :) (that said, wonder if we could / should have some sort of either report or automatic script for fixing those).

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: I don't think they have a different favicon for reviews? Am I missing something? :)

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Sounds sensible, will add

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Oh, or is your idea (seeing the miura.com example above) that there should be an `expected_clean_url` even if the expectation is there's no cleanup and you get the same that came in?

reosarevok:
A note: I did not add tests for the URLs not currently being tested at all in the URLCleanup tests (a bunch of them we don't clean up at all, we just autoselect, and I think there might be one or two stores we don't even autoselect to begin with). We could do that, but then we'd need to decide for sure how to do the testing for non-cleaned URLs :)

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: Right. I was initially looking for a favicon test of `kenjimiura.com` when I noticed you added a favicon test to all RA tests but that one; I just wondered why. Now looking at the code again ([commit `86ef099` - `root/static/scripts/common/constants.js` - line 156](https://github.com/metabrainz/musicbrainz-server/pull/2256/commits/86ef0995c19be175295f8b4af667c43ecf087d1f#diff-a0ecd03ab7967706f21d4f2a99034bd55a8daef0272fbace5c6b63d02c1ba01cR156)), setting this favicon doesn’t depend on the path indeed. So maybe it’s not useful to have 7 favicon tests for RA URLs, having 1 might be enough?

reosarevok: Oh, now I see :) I added it to all of them but the one with no valid entity types (that's what I generally did elsewhere too) just in case we changed some specific cleanup and it no longer matched the favicon, but admittedly it might be overdoing it :) I can remove it from the rest if desired.

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: > Yes, IMO that should be ok, shouldn't it?

Not sure what “that” and “it” refer to.

> It's tested against the return of `cleanURL(), which is the original if there's no cleanup function`

It is hiding the URL being actually tested for favicon class, from both the test file and the test output.

The clean URL may significantly differ from the input URL, that’s why I suggested to make `expected_clean_url` mandatory for use with `expected_favicon_class`.

> so effectively, it is tested against whatever will actually be added to the DB.

So effectively, it doesn’t test against actual unclean URLs already in the DB.
Let’s use the above thread https://github.com/metabrainz/musicbrainz-server/pull/2256#discussion_r709123871 for follow-up on this point.

yvanzo: > Oh, or is your idea (seeing the miura.com example above) that there should be an `expected_clean_url` even if the expectation is there's no cleanup and you get the same that came in?

Yes. (Did post the previous comment before seeing that one.)

reosarevok: Ok, we can always do that and always have `expected_clean_url` - I just thought that was a bit misleading because it suggested a cleanup was being run, but maybe not :)

yvanzo:
After [further discussion on IRC](https://chatlogs.metabrainz.org/libera/metabrainz/msg/4860476/), it seems that existing `match` properties in `URLCleanup`could be used instead of new regular expressions in `FAVICON_CLASSES`, so deduplicating code and displaying favicon for non-clean URLs too (as it is for now). Favicon tests would then have to run against both `input_url` and clean URL (`expected_clean_url` if available, otherwise `cleanURL()`).

I did investigate for potential edge cases: Found `unwelcomeimages` in `CLEANUPS`; So it seems `faviconify` should be optional.

It would also require to import `URLCleanup.js` with non-edit pages. @mwiencek: Does that seem reasonable?

root/url/utility/isUrlValid.js
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (C) 2021 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export default function isUrlValid(url: string): boolean {
+  try {
+    new URL(url);
```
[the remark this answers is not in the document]
mwiencek: Well, when I wrote that in 2014 the `URL` API was brand new (if supported at all outside of Chrome) and we didn't have core-js polyfills or anything back then.

One difference that probably exists, though, is that something like `new URL('/haha')` is valid, whereas the current function requires a fully-specified URL.  This seems like something that should be looked into, if it matters.

mwiencek:
> It would also require to import URLCleanup.js with non-edit pages. @mwiencek: Does that seem reasonable?

It's a very large file, so if we can't avoid it. :\  Can we put the match regular expressions into a separate file and import them into `URLCleanup.js` instead?

yvanzo:
> > It would also require to import URLCleanup.js with non-edit pages. @mwiencek: Does that seem reasonable?
> 
> It's a very large file, so if we can't avoid it. :\ Can we put the match regular expressions into a separate file and import them into `URLCleanup.js` instead?

That would make `URLCleanup.js` more difficult to maintain because `clean` and `validate` are often closely related to the `match` regular expressions.

* If `URLCleanup.js` builds `const FAVICON_CLASSES` (from `CLEANUPS`), exports that separately, and `root/static/scripts/common/constants.js` imports that only from `URLCleanup.js`, would it avoid loading the whole `URLCleanup.js`?
* Or is there a way to automatically generate a `faviconClasses.js` file from `URLCleanup.js` at build time?

mwiencek:
> If URLCleanup.js builds const FAVICON_CLASSES (from CLEANUPS), exports that separately, and root/static/scripts/common/constants.js imports that only from URLCleanup.js, would it avoid loading the whole URLCleanup.js?

It'd still load the whole file in that case.

> Or is there a way to automatically generate a faviconClasses.js file from URLCleanup.js at build time?

This wouldn't be too difficult, actually, since [you can access .source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/source) to rebuild the expressions.

root/url/utility/isUrlValid.js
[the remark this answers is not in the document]
reosarevok: That's not quite true - https://developer.mozilla.org/en-US/docs/Web/API/URL/URL "If url is a relative URL, base is required, and will be used as the base URL". Otherwise, it errors anyway.

PR #2257, opened 2021-09-08
MBS-11941: Normalize Worldcat identities URLs

reosarevok:
### Implement MBS-11941

This ensures the permalinks to Worldcat identities pages include the ending slash, since that's their default.
Additionally, wcidentities pages get converted into identities pages (as far as I can tell they are both the same content with a slightly different design).


yvanzo (approved):
Last commit looks good to me but needs to be rebased on `master` after merging the pull request it depends on.

mwiencek (approved):

PR #2258, opened 2021-09-08
MBS-11940: Properly error for invalid XML sent to /ws/2/tag

mwiencek:
https://tickets.metabrainz.org/browse/MBS-11940

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/Tag.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/Tag.pm
@@ -66,9 +66,16 @@ sub tag_submit : Private
     $self->_validate_post($c);
 
     my $xp = MusicBrainz::Server::WebService::XML::XPath->new( xml => $c->request->body );
+    my @nodelist;
+    eval {
+        @nodelist = $xp->find('/mb:metadata/*/*')->get_nodelist;
+    };
+    if ($@) {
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/WS/2/Tag.pm
```diff
@@ -66,9 +66,16 @@ sub tag_submit : Private
     $self->_validate_post($c);
 
     my $xp = MusicBrainz::Server::WebService::XML::XPath->new( xml => $c->request->body );
+    my @nodelist;
+    eval {
+        @nodelist = $xp->find('/mb:metadata/*/*')->get_nodelist;
+    };
+    if ($@) {
```
reosarevok, line 73: I'm a bit confused, what does this mean here? :)

mwiencek: Just one of those special perl vars you have to memorize.  If there's an error inside the `eval`, Perl sets `$@` to the error: https://perldoc.perl.org/perlvar#%24%40

reosarevok: So it's effectively a try-catch, ok :)

reosarevok (approved):

yvanzo (approved):
Thanks

PR #2260, opened 2021-09-13
MBS-11957: Clean up twitch.com to twitch.tv

reosarevok:
### Implement MBS-11957

Twitch allows both but automatically redirects. We had 5 of these in the DB right now, which I fixed by hand (including the test example).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -4036,10 +4036,10 @@ const CLEANUPS: CleanupEntries = {
     },
   },
   'twitch': {
-    match: [new RegExp('^(https?://)?([^/]+\\.)?(twitch\\.tv/)', 'i')],
+    match: [new RegExp('^(https?://)?([^/]+\\.)?twitch\\.(?:com|tv)/', 'i')],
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
Last commit looks good to me but needs to be rebased on `master` after merging the pull request it depends on.

yvanzo:
@reosarevok: Pushed a small fix for a sneaky bug in Twitch link’s validation; See <https://github.com/metabrainz/musicbrainz-server/pull/2260/commits/4892294a5c87104cb95d53215dc9f0d995e3e613>. It doesn’t seem worth a ticket as no “streaming free” relationship has been used to link to Twitch so far; See <https://musicbrainz.org/ws/2/url?query=url:*twitch*&fmt=json>.

reosarevok:
Good catch, thanks!

mwiencek (approved):

mwiencek:
Does this conflict with / need to be based on https://github.com/metabrainz/musicbrainz-server/pull/2256?

reosarevok:
Well, right now it just uses the favicon tests. But if that PR doesn't get ready this week, I can modify the other ones not to depend on it and merge them, that shouldn't be a huge issue :) Same for anything else now on top of that.

PR #2261, opened 2021-09-13
MBS-11959: Allow RYM links for music video recordings

reosarevok:
### Implement MBS-11959

Since RYM doesn't seem to otherwise have recording level data, this specifically limits recording links to music videos (which have recently been added).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3429,9 +3429,10 @@ const CLEANUPS: CleanupEntries = {
       return url.replace(/^(?:https?:\/\/)?(?:www\.)?rateyourmusic\.com\//, 'https://rateyourmusic.com/');
     },
     validate: function (url, id) {
-      const m = /^https:\/\/rateyourmusic\.com\/(\w+)\//.exec(url);
+      const m = /^https:\/\/rateyourmusic\.com\/(\w+)\/(?:(\w+)\/)?/.exec(url);
       if (m) {
         const prefix = m[1];
+        const subPath = m[2];
         switch (id) {
           case LINK_TYPES.otherdatabases.artist:
             return {
@@ -3451,6 +3452,12 @@ const CLEANUPS: CleanupEntries = {
           case LINK_TYPES.otherdatabases.place:
             return {
               result: prefix === 'venue',
+              target: ERROR_TARGETS.RELATIONSHIP,
+            };
+          case LINK_TYPES.otherdatabases.recording:
+            return {
+              error: l('Only RYM music videos can be linked to recordings.'),
+              result: prefix === 'release' && subPath === 'musicvideo',
               target: ERROR_TARGETS.RELATIONSHIP,
             };
           case LINK_TYPES.otherdatabases.release:
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -3311,6 +3311,13 @@ const testData = [
         only_valid_entity_types: ['place'],
   },
   {
+                     input_url: 'https://rateyourmusic.com/release/musicvideo/metallica/one/',
+             input_entity_type: 'recording',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://rateyourmusic.com/release/musicvideo/metallica/one/',
+        only_valid_entity_types: ['recording', 'release', 'release_group'],
+  },
+  {
                      input_url: 'https://rateyourmusic.com/release/single/tori_amos/a_sorta_fairytale/',
              input_entity_type: 'release_group',
     expected_relationship_type: 'otherdatabases',


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3453,6 +3454,12 @@ const CLEANUPS: CleanupEntries = {
               result: prefix === 'venue',
               target: ERROR_TARGETS.RELATIONSHIP,
             };
+          case LINK_TYPES.otherdatabases.recording:
+            return {
+              error: l('Only RYM music videos can be linked to recordings.'),
```
yvanzo, line 3459: Is it really necessary to have a special message here? Other RYM links were already denied for linking to recordings.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -3695,6 +3695,14 @@ const testData = [
         only_valid_entity_types: ['place'],
         expected_favicon_class: 'rateyourmusic-favicon',
   },
+  {
+                     input_url: 'https://rateyourmusic.com/release/musicvideo/metallica/one/',
+             input_entity_type: 'recording',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://rateyourmusic.com/release/musicvideo/metallica/one/',
+        only_valid_entity_types: ['recording', 'release', 'release_group'],
```
yvanzo, line 3703: Allowing these three MB entity types to link with RYM music video seems legitimate indeed because music video may be considered either as a standalone recording or as a single, and both MB release and MB release group overlap RYM album.
It might be worth mentioning in the commit message that RYM music videos were already (and remain) allowed as external link to release and release-group.

reosarevok: Actually, right now the official line is "a music video is a recording, not a release", so it might make sense to block them from release and release-group instead...

yvanzo: I assumed that was intentional because this “official line” has exceptions. IIRC, the main reason some videos are entered as releases is to attach cover art. Is there an alternative to this use case? Otherwise it seems fine to allow either type for now.

reosarevok: Well, in theory that's not supposed to happen, but I also don't know in which cases we have cover art for videos :) In any case, it seems like it'd be still fine to have those have the relationship on the recording, since that's still a video recording?

yvanzo: Videos may have (large and good quality) thumbnails that can be used as cover art. Even streaming pages may have images associated with, https://soundcloud.com/hunnytheband/xbox-luvr for example. In this situation, it makes sense to link the MB release to the page the music video is available from (e.g. Tidal). But yes, other types of link such as other databases (e.g. RYM) should probably be linked to the recording instead. In this case, Recording-URL should probably be made visible from the release page too?

reosarevok: Hmm. I'd argue a track like that one you added could be added as a single, but that a video from YouTube or whatever should not be added as a release (thumbnails for videos are as far as I've seen just stills from the videos). That said, there is a confusing level where, say, a full concert published as a DVD would get a release, but a full concert uploaded to YouTube wouldn't.

There's been talk in the forums about what to do with videos though, for example https://community.metabrainz.org/t/adding-covers-uploaded-to-youtube-as-a-standalone-recordings-or-singles/544236/1 - the general idea seems to be "we would like them to be better represented while not (generally) lumped into the general discography".

reosarevok: > In this case, Recording-URL should probably be made visible from the release page too?

I probably wouldn't want to do that for all, but I could see having a small set of icons for linked streaming sites? And maybe lyrics sites for the works? It's complicated because that view is already full of stuff, but :)

yvanzo: > thumbnails for videos are as far as I've seen just stills from the videos

* https://www.youtube.com/watch?v=Sp3dFF-Bts0
* SoundCloud doesn’t host videos and still supports thumbnails

reosarevok: Ok, I guess you can also have a separate cover-ish thumbnail then :) I stand corrected! (the soundcloud thing I already said seems fine to use as a cover for a release). We really need a better way to do videos than "it's a release", though, but I guess we could for now allow video links for releases and then consolidate whatever happens later.

yvanzo: For “get the music” relationship types only (streaming such as Tidal, Vimeo…)? Or for all relationship types (other databases such as RYM too)?

reosarevok: What would be your preference? Because the whole thing is a bit of a hack, I'm not sure what's the best option. Maybe all would be less confusing?

yvanzo: OK then, in doubt I prefer `URLCleanup` not to get in the way of editors too.

reosarevok: So should I change anything here? The commit message?

yvanzo: Yes, only the commit message.

reosarevok: Done!

root/static/scripts/edit/URLCleanup.js
reosarevok: I think it might be useful in case people try to link a single to a recording or something? :)

yvanzo (approved):
Thanks! (Tested it again)

PR #2262, opened 2021-09-13
MBS-11952: Autoselect Vimeo On Demand URLs correctly

reosarevok:
### Fix MBS-11952

Vimeo has a Vimeo On Demand service which allows video makers to rent out and sell their videos. These give a smaller rental price for 24 hours, allowing you to stream from the website, and a larger price to purchase outright, allowing streaming and downloads.

It seems in a few cases only one or the other are available, so rather than autoselecting, this just restricts the options to one (or both) of those.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -4291,7 +4291,10 @@ const CLEANUPS: CleanupEntries = {
     },
   },
   'vimeo': {
-    match: [new RegExp('^(https?://)?([^/]+\\.)?(vimeo\\.com/)', 'i')],
+    match: [new RegExp(
+      '^(https?://)?([^/]+\\.)?vimeo\\.com/(?!ondemand)',
+      'i',
+    )],
```
yvanzo, line 4297: It should not match `/store/ondemand` path either, for example:
* `https://vimeo.com/store/ondemand/buy/91410`
* `https://vimeo.com/store/ondemand/rent/91410`

yvanzo (changes requested):
It seems that `/store/ondemand` URLs need special handling too.

root/static/scripts/edit/URLCleanup.js
reosarevok: Yeah. But I guess there's no way to turn these back into the original non-store link, so we should reject these under the ondemand regexp and ask people to add the main page instead.

yvanzo: > I guess there's no way to turn these back into the original non-store link, so we should reject these under the ondemand regexp

Yes.

> ask people to add the main page instead.

Just mentioning the expected URL format should be fine. It can be a generic error message for other websites too.

lib/MusicBrainz/Server/Entity/URL/VimeoOnDemand.pm
```diff
@@ -0,0 +1,26 @@
...
+
+use Moose;
+
+extends 'MusicBrainz::Server::Entity::URL';
+with 'MusicBrainz::Server::Entity::URL::Sidebar';
+
+override href_url => sub {
+    shift->url->as_string =~ s{^http:}{https:}r;
+};
+
```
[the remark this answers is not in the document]
reosarevok: Thanks! Removed this from here, and sent https://github.com/metabrainz/musicbrainz-server/pull/2274 to remove a bunch more.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4300,6 +4303,36 @@ const CLEANUPS: CleanupEntries = {
...
+      '^(https?://)?([^/]+\\.)?vimeo\\.com/(?:ondemand|store/ondemand)',
+      'i',
+    )],
+    restrict: [LINK_TYPES.downloadpurchase, LINK_TYPES.streamingpaid],
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?vimeo\.com\/ondemand\/([^\/?#]+)(?:.*)$/, 'https://vimeo.com/ondemand/$1');
+      return url;
+    },
+    validate: function (url) {
+      const isStoreLink = /^https:\/\/vimeo\.com\/store\/ondemand/.test(url);
```
yvanzo, line 4317: Because `/store/ondemand` links are not modified by the function `clean`, the function `validate` should catch unclean URLs too:
```suggestion
      const isStoreLink = /vimeo\.com\/store\/ondemand/.test(url);
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4300,6 +4303,36 @@ const CLEANUPS: CleanupEntries = {
...
+              video_url_pattern: (
+                <span className="url-quote">{'/ondemand'}</span>
+              ),
+            },
+          ),
+          result: false,
+          target: ERROR_TARGETS.URL,
+        };
+      }
+      return {result: true};
```
yvanzo, line 4333: Further validation should happen here to prevent linking to either `artist` or `label`.

root/static/scripts/common/constants.js
```diff
@@ -44,164 +44,165 @@ export const DARTIST_ID = 2;
...
+  twitch: {host: /^www\.twitch\.tv$/},
+  twitter: {host: /^twitter\.com$/},
+  utaitedb: {host: /^utaitedb\.net$/},
+  utamap: {host: /^(?:[^/]+\.)?utamap\.com$/},
+  utanet: {host: /^www\.uta-net\.com$/},
+  utaten: {host: /^utaten\.com$/},
+  vgmdb: {host: /^vgmdb\.net$/},
+  viaf: {host: /^viaf\.org$/},
+  videogamin: {host: /^(?:www\.)?videogam\.in$/},
+  vimeo: {host: /^vimeo\.com$/, path: /^\/(?!(?:ondemand|store\/ondemand))/},
```
yvanzo, line 193: It is probably fine not to match `store/ondemand` here, or it should be matched in `lib/MusicBrainz/Server/Data/URL.pm` too.

```suggestion
  vimeo: {host: /^vimeo\.com$/, path: /^\/(?!(?:ondemand))/},
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4261,95 +4763,130 @@ const testData = [
...
              input_entity_type: 'recording',
     expected_relationship_type: 'streamingfree',
             expected_clean_url: 'https://vimeo.com/1109226',
+        expected_favicon_class: 'vimeo-favicon',
+  },
+  {
+                     input_url: 'https://vimeo.com/ondemand/inconcert/193518106?autoplay=1',
+            expected_clean_url: 'https://vimeo.com/ondemand/inconcert',
+        expected_favicon_class: 'vimeoondemand-favicon',
+  },
```
yvanzo, line 4878: Just a note for later on: This would really need additional tests when there is type restriction but no auto-select. For follow-up: https://tickets.metabrainz.org/browse/MBS-11980

yvanzo (changes requested):
Almost fixed. Found some lesser issues.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Agreed, thanks for adding the ticket.

root/static/scripts/common/constants.js
reosarevok: This is not matching it, this is avoiding matching it so that it doesn't show the favicon while it gets rejected.

root/static/scripts/edit/URLCleanup.js
reosarevok: I guess that might make sense, yes :) Not sure whether we should just restrict it to recording only?

yvanzo: Or still allow release? Discussion is still going on at https://github.com/metabrainz/musicbrainz-server/pull/2261#discussion_r711110828

yvanzo: The outcome of https://github.com/metabrainz/musicbrainz-server/pull/2261#discussion_r713729787 is “for now allow video links for releases and then consolidate whatever happens later”.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4300,6 +4303,48 @@ const CLEANUPS: CleanupEntries = {
...
+        };
+      }
+      switch (id) {
+        case LINK_TYPES.downloadpurchase.recording:
+        case LINK_TYPES.downloadpurchase.release:
+        case LINK_TYPES.streamingpaid.recording:
+        case LINK_TYPES.streamingpaid.release:
+          return {
+            result: /^https:\/\/vimeo\.com\/ondemand\/([^\/?#]+)*$/.test(url),
+          };
```
yvanzo, line 4343: It seems to be missing `target`.

mwiencek (approved):

yvanzo (approved):

PR #2263, opened 2021-09-13
MBS-10621: Handle Tidal URLs

yvanzo:
# MBS-10621

* Display Tidal URLs in the sidebar
* Auto-select (purchase for) “streaming” relationship type for Tidal URLs
* Normalize Tidal URLs
* Validate relationship types w.r.t. Tidal URLs

# Checklist for author

* [x] Get confirmation for https://tickets.metabrainz.org/browse/MBS-10621?focusedCommentId=52589#comment-52589
* [x] Complete tests accordingly, update code accordingly if needed
* [x] Rebase on `master` branch
* [ ] Rebase on `master` if #2280 is merged first

# Action

1. Check for existing URLs that require normalization, and create an MBBE ticket if appropriate.


Xythium:
does this do anything about tidal.com vs listen.tidal.com? im not entirely sure it makes sense to have listen.tidal.com links since that just seems to be a host for the desktop application from what ive seen?

yvanzo:
> does this do anything about tidal.com vs listen.tidal.com?

It does remove `desktop.`, `listen.`, `stage.`, `store.` and `www.`subdomains of `tidal.org`; See the specific [line of code](https://github.com/yvanzo/musicbrainz-server/commit/f18e2de151896880a782a52959ae8bfdc737dcbd#diff-a808bbcd17cb3e956e9ac0fce554d75f6e9db02e020bc03ca543c198ac8f8e06R2153).

> that just seems to be a host for the desktop application from what ive seen

Since you had a look at Tidal, can you please (dis)confirm my guesses from the related [ticket’s comment](https://tickets.metabrainz.org/browse/MBS-10621?focusedCommentId=52589#comment-52589)?

Xythium:
> @Xythium Since you had a look at Tidal, can you please (dis)confirm my guesses from the related [ticket’s comment](https://tickets.metabrainz.org/browse/MBS-10621?focusedCommentId=52589#comment-52589)?

I've never used Tidal as a platform but I managed to reverse engineer their internal API to make tools to get metadata from Tidal (exclusive) releases such as barcodes and ISRCs, so that's all I really know. I don't see any wrong assumptions but I'm not an expert.

I have seen, that on the listen.tidal.com API, they link to albums with the incorrect URL so I don't really know whats up there. For example this says http://www.tidal.com/album/191185943 but will redirect to https://tidal.com/browse/album/191185943
image
Maybe it is supposed to be http://listen.tidal.com/album/191185943 but they link to the wrong subdomain?

Also, this may not be relevant to MusicBrainz but listen.tidal.com has Credits pages that I think the other subdomains don't have? https://listen.tidal.com/album/191185943/credits

yvanzo:
Thanks @Xythium: I did complete the patch as follows:

> Add support for editing Tidal URLs:
> 
> * Auto-select “paid streaming” relationship type
> * Normalize URLs (see test examples) to the following canonical form `https://tidal.com/<type>/<id>` even though some are redirects.
> * Validate canonical URLs with the corresponding entity type only:
>   * `album` with MB Release
>   * `artist` with MB Artist
>   * `track` and `video` with MB Recording
> 
> Specifically ignore (that is neither auto-select nor block) URLs having either `playlist` path or `beyonce` subdomain.

I didn’t retain the casual 2-letters code because it seems to match user interface language rather than any country store.

reosarevok:
> Specifically ignore (that is neither auto-select nor block) URLs having either `playlist` path or `beyonce` subdomain.

Maybe add a test for those cases checking they do not get autoselected? (if we can do that at all, that is, I forget right now if we can test for "not blocked, just not selected")

reosarevok (approved):
Seems good to merge to me in general, but left comments.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3976,6 +3976,45 @@ const CLEANUPS: CleanupEntries = {
...
+    validate: function (url, id) {
+      const m = /^https:\/\/tidal\.com\/([a-z]+)\/\d+$/.exec(url);
+      if (m) {
+        const prefix = m[1];
+        switch (id) {
+          case LINK_TYPES.streamingpaid.artist:
+            if (prefix === 'artist') {
+              return {result: true};
+            }
+            return {result: false, target: ERROR_TARGETS.ENTITY};
```
reosarevok, line 4001: Is there no use on having an error message you think?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3976,6 +3976,45 @@ const CLEANUPS: CleanupEntries = {
...
+          case LINK_TYPES.streamingpaid.release:
+            if (prefix === 'album') {
+              return {result: true};
+            }
+            return {result: false, target: ERROR_TARGETS.ENTITY};
+          case LINK_TYPES.streamingpaid.recording:
+            if (prefix === 'track' || prefix === 'video') {
```
reosarevok, line 4008: You said in https://github.com/metabrainz/musicbrainz-server/pull/2261#discussion_r710119259 that videos should be supported for releases, but here they are not. We should be consistent maybe? But more discussion there.

Xythium:
There are actually some more of the Beyonce pages for other artists (not on specific subdomains per se), such as https://tidal.com/deadmau5 who is also one of the "owners"/founding members of Tidal

root/static/scripts/edit/URLCleanup.js
yvanzo: Actually, I assumed the code was intentional and the commit message incomplete; See comment.
But yes, whatever is decided, it should be the same for videos here.

root/static/scripts/edit/URLCleanup.js
yvanzo: I thought there would be a generic one instead, but it's not the case, will have to open a pull request for that too.

yvanzo:
> There are actually some more of the Beyonce pages for other artists (not on specific subdomains per se), such as https://tidal.com/deadmau5 who is also one of the "owners"/founding members of Tidal
> 
> https://jayz.tidal.com/ https://tidal.com/jayz
> […]

Thanks. All of these URLs (any subdomain but `desktop`, `listen`, `stage`, `store`, `www`, or any path not containing `album/`, `artist/`, `track/`, `video/`) will be ignored by the current patch: that is they are not blocked and relationship type has to be manually selected by the editor. They seem to not be frequent enough to be worth automation. Suggestions are welcome though.

root/static/scripts/edit/URLCleanup.js
yvanzo: Created https://tickets.metabrainz.org/browse/MBS-11977 and started working on it.

yvanzo: Cherry-picked https://github.com/metabrainz/musicbrainz-server/pull/2263/commits/e4da3d6caa84d9307a1123a8b02b8f49523cde41 from pull request #2280.

root/static/scripts/edit/URLCleanup.js
reosarevok: I guess given that discussion videos should be unblocked for releases too?

reosarevok (approved):
LGTM but didn't test. Do consider the video question though :)

reosarevok:

root/static/scripts/edit/URLCleanup.js
yvanzo: My bad, I forgot about it. Updated with commit <https://github.com/metabrainz/musicbrainz-server/pull/2263/commits/004f5b86f68bdb4c928b6b2dc4c673b9647fb6b5>.

yvanzo:
Thanks!

mwiencek (approved):

PR #2264, opened 2021-09-13
MBS-11961: Rename "random" in edit search since it's not random

reosarevok:
### Implement MBS-11961

This just does not select any sort when the 'rand' order is selected.

PSQL docs used to say "If sorting is not chosen, the rows will be returned in random order" but that's misleading and has since been corrected to "the rows will be returned in an unspecified order". For at least some queries, the unspecified order is literally "ordered by ID", which means people will be confused about the clear non-randomness.

Actually making this random seems to have little use, and makes things very slow (see https://github.com/metabrainz/musicbrainz-server/pull/2043). Since the most useful thing about the current option is that skipping sorting sometimes makes things quite a bit faster, this just renames the option to be less misleading.


yvanzo (approved):
LGTMBDNT™
:ship:

mwiencek (approved):

PR #2265, opened 2021-09-15
Allow limiting compression threads from config

yvanzo:
# Problem

To compress JSON dumps (using `xz`) and search index dumps (using `zstd`), the number of threads is currently not limited so dumps can be made available faster. However, it now has detrimental impacts to the server’s availability when dumps are built.

# Solution

This patch adds the key `DUMP_COMPRESSION_THREADS` to the configuration so it can be changed at deployment time. It doesn’t change the default though.

# Action

1. Set the value for `DUMP_COMPRESSION_THREADS` key to `4` in docker server configs for both `musicbrainz-json-dump` and `musicbrainz-search-indexes-dump`.

### DIFF
--- a/lib/DBDefs/Default.pm
+++ b/lib/DBDefs/Default.pm
@@ -472,6 +472,10 @@ sub SOLRCLOUD_RSYNC_BANDWIDTH { undef }
 sub SOLRCLOUD_SSH_CIPHER_SPEC { undef }
 sub SEARCH_INDEXES_DUMP_COMPRESSION_LEVEL { undef }
 
+# Controls the number of threads to use when compressing JSON dumps or search
+# index dumps. Passed as `-T` or `--threads` to xz and zstd.
+sub DUMP_COMPRESSION_THREADS { 0 }
+
 sub WIKIMEDIA_COMMONS_IMAGES_ENABLED { 1 }
 
 # On release browse endpoints in the webservice, we limit the number of
--- a/lib/MusicBrainz/Script/MBDump.pm
+++ b/lib/MusicBrainz/Script/MBDump.pm
@@ -59,6 +59,12 @@ has compression_level => (
     isa => 'Maybe[Str]',
 );
 
+has compression_threads => (
+    is => 'rw',
+    isa => 'Maybe[Int]',
+    default => DBDefs->DUMP_COMPRESSION_THREADS,
+);
+
 has replication_sequence => (
     is => 'ro',
     isa => 'Maybe[Int]',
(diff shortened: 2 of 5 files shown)


lib/MusicBrainz/Script/MBDump.pm
```diff
@@ -157,7 +163,7 @@ sub make_tar {
     my $compress_command;
     if ($compression) {
         $compress_command = "$compression";
-        $compress_command .= ' --threads=0' if $compression eq 'xz';
+        $compress_command .= ' --threads=$compression_threads' if $compression eq 'xz';
```
mwiencek, line 166: A declaration for `my $compression_threads` seems to be missing here.

lib/DBDefs/Default.pm
```diff
@@ -472,6 +472,9 @@ sub SOLRCLOUD_RSYNC_BANDWIDTH { undef }
 sub SOLRCLOUD_SSH_CIPHER_SPEC { undef }
 sub SEARCH_INDEXES_DUMP_COMPRESSION_LEVEL { undef }
 
+# Used to compress dump using xz or zstd.
```
mwiencek, line 475: Maybe something slightly more specific like this:
```suggestion
# Controls the number of threads to use when compressing JSON dumps or search
# index dumps. Passed as `-T` or `--threads` to xz and zstd.
```

mwiencek (changes requested):

lib/MusicBrainz/Script/MBDump.pm
yvanzo: Thanks, fixed that and changed to double quotes too.

lib/DBDefs/Default.pm
yvanzo: Thanks, applied it.

lib/MusicBrainz/Script/MBDump.pm
```diff
@@ -59,6 +59,12 @@ has compression_level => (
     isa => 'Maybe[Str]',
 );
 
+has compression_threads => (
+    is => 'rw',
+    isa => 'Maybe[Str]',
```
mwiencek, line 64:
```suggestion
    isa => 'Maybe[Int]',
```

mwiencek (approved):
One more tiny suggestion that isn't very impactful, but LGTM otherwise

yvanzo:
Additionally added timestamped logs for copying dump files around, since there currently are alerts to be resolved.

PR #2267, opened 2021-09-17
MBS-11919: Remove numeric labels from URLs on new editor

y-young:
# Problem
- MBS-11919: Remove numeric labels from URLs on new editor

# Solution
- Remove link index numbers for simplicity, add a link in duplicate notice to jump to the referenced link. (MBS-11919)

image

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -668,14 +668,7 @@ export class ExternalLinksEditor
 
             // Check duplicates and show notice
             const duplicate = links[0].submitted
-              ? false : linksGroupMap.get(url);
-            const duplicateNotice = duplicate
-              ? texp.l(
-                `Note: This link already exists 
-                 at position #{position}. 
-                 To merge, press enter or select a type.`,
-                {position: duplicate[0].urlIndex + 1},
-              ) : '';
+              ? null : linksGroupMap.get(url);
 
             let urlError: ErrorT | null = null;
             let hasError = false;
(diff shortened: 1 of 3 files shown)


y-young:

y-young:

reosarevok:
One thing to keep in mind: if the duplicate link added is a valid autoselectable link (so, there's no need to select a type) then when trying to click the link it will automatically merge it anyway since the field lost focus. That seems a bit confusing, we should probably either show a slightly different error there, or make it so that it's not saved/merged on clicking the link.

y-young:
> One thing to keep in mind: if the duplicate link added is a valid autoselectable link (so, there's no need to select a type) then when trying to click the link it will automatically merge it anyway since the field lost focus.

I guess it'll trigger a "This relationship already exists" error? In that case we should probably block the merge.

reosarevok:
Yeah, it will. It might make sense to block any merge that will trigger that, if that's easy enough to do :)

y-young:
That's an existing issue and it seems not easy to fix, so maybe later in another PR.

yvanzo (changes requested):
This change is twofold, it is worth two separate pull requests (or at the very least two commits).
1. Moving edit icons works nicely; It could be merged faster if in a separate pull request. It is worth its own ticket too.
2. The link pointing to the position does not seem to work as clicking on it doesn’t scroll to the target item. Maybe adding an highlighting border around the item at the target position would be help with finding it.

y-young:
OK, so renaming source branch automatically closes the PR. :|

PR #2269, opened 2021-09-17
MBS-11960: Also assign seeded URL to rawURL

reosarevok:
### Fix MBS-11960

Without this, the seeded URL is stored and can be submitted, but it does not get displayed to the user, which seems less than ideal.

On a second commit, I make it so seeded links always come after existing ones.

Tested with the seeding options suggested on the ticket, plus with a couple extra links (just the URL, just the type) to make sure multi-seeding works.


y-young:
LGTM.

mwiencek (approved):

PR #2272, opened 2021-09-18
Timestamp all logs produced by cron scripts

yvanzo:
# Problem

Scripts ran from crontab do timestamp some messages but not all of them, and - more problematic - they do not timestamp error messages.

# Solution

Cron command was already redirecting to the log file. It now timestamps every line of log using the command `ts` from the package `moreutils`.

Since cron commands timestamp the output of scripts, the timestamps added by scripts run using cron are redundant and thus removed here. Scripts for slave servers are not modified though, because their cron command is set by their maintainer, and they might be started manually.

Also, replace `print(f)` in scripts with `Log` subroutines.

# Checklist for author

* [x] Test modified scripts

# Action

1. Update containers dedicated to cron jobs, sitemaps, JSON dump, and search indexes dump.

### DIFF
--- a/admin/RunReports.pl
+++ b/admin/RunReports.pl
@@ -7,6 +7,7 @@ use lib "$FindBin::Bin/../lib";
 
 use List::AllUtils qw ( any );
 use MusicBrainz::Server::Context;
+use MusicBrainz::Server::Log qw( log_info log_debug );
 use MusicBrainz::Server::ReportFactory;
 use POSIX qw( SIGALRM );
 $| = 1;
@@ -17,13 +18,13 @@ my $c = MusicBrainz::Server::Context->create_script_context();
 my $errors = 0;
 for my $name (MusicBrainz::Server::ReportFactory->all_report_names) {
     unless (any { $name =~ /$_/i } @ARGV) {
-        print localtime() . " : Not running $name\n";
+        log_debug { "Not running $name" };
         next;
     }
 
     my $report = MusicBrainz::Server::ReportFactory->create_report($name, $c);
 
-    print localtime() . " : Running $name\n";
+    log_info { "Running $name" };
     my $t0 = time;
     my $ONE_HOUR = 1 * 60 * 60;
     my $exit_code = eval {
@@ -59,11 +60,11 @@ for my $name (MusicBrainz::Server::ReportFactory->all_report_names) {
     }
     my $t = time() - $t0;
 
-    print localtime() . " : $name finished; time=$t\n";
+    log_info { "$name finished; time=$t" };
 }
 
-print localtime() . " : Completed with 1 error\n" if $errors == 1;
-print localtime() . " : Completed with $errors errors\n" if $errors != 1;
+log_info { 'Completed with 1 error' } if $errors == 1;
+log_info { "Completed with $errors errors" } if $errors != 1;
 
 exit($errors ? 1 : 0);
 
--- a/admin/cleanup/FixTrackLength.pl
+++ b/admin/cleanup/FixTrackLength.pl
@@ -19,6 +19,7 @@ use MusicBrainz::Server::Constants qw(
     $EDIT_MEDIUM_EDIT
 );
 use MusicBrainz::Server::Constants qw( $AUTO_EDITOR_FLAG );
+use MusicBrainz::Server::Log qw( log_debug log_info log_warning );
 use MusicBrainz::Server::Track qw( format_track_length );
 
 use Getopt::Long;
@@ -49,7 +50,7 @@ EOF
 my $c = MusicBrainz::Server::Context->create_script_context;
 
 # Find mediums with at least one track to fix
-print localtime() . " : Finding candidate mediums\n" if $verbose;
+log_info { 'Finding candidate mediums' } if $verbose;
 my @medium_ids = @{ $c->sql->select_single_column_array(
     'SELECT DISTINCT m.id
        FROM medium m
@@ -59,9 +60,9 @@ my @medium_ids = @{ $c->sql->select_single_column_array(
       WHERE t.length IS NULL OR t.length = 0 AND m.track_count > 0
         AND (mf.has_discids = TRUE OR mf.has_discids IS NULL)'
 ) };
-printf localtime() . " : Found %d medium%s\n",
+log_info { sprintf 'Found %d medium%s',
     scalar(@medium_ids), (@medium_ids == 1 ? '' : 's')
-    if $verbose;
+} if $verbose;
 
 my $tracks_fixed = 0;
 my $tracks_set = 0;
@@ -76,7 +77,7 @@ my $modbot = $c->model('Editor')->get_by_id($EDITOR_MODBOT);
 
 for my $medium (@mediums)
 {
-    printf "%s : Fixing medium #%d\n", scalar(localtime), $medium->id
+    log_info { sprintf 'Fixing medium #%d', $medium->id }
         if $verbose;
 
     my @cdtocs = grep { $_->edits_pending == 0 } $c->model('MediumCDTOC')->find_by_medium($medium->id);
@@ -88,19 +89,19 @@ for my $medium (@mediums)
     my @tracks = $medium->all_tracks;
 
     if ($debug) {
-        print "TOCs:\n";
+        log_debug { 'TOCs:' };
         for my $cdtoc (@cdtocs) {
-            print '  ' . $cdtoc->toc . "\n";
-            printf "    (%s)\n", format_track_length($_->{length_time})
+            log_debug { '  ' . $cdtoc->toc };
+            log_debug { sprintf '    (%s)', format_track_length($_->{length_time}) }
                 for @{ $cdtoc->track_details };
         }
 
-        print "Tracks:\n";
-        printf "  #%02d : %10d %-8s  %12d\n",
+        log_debug { 'Tracks:' };
+        log_debug { sprintf '  #%02d : %10d %-8s  %12d',
             $_->position, $_->length || 0,
             $_->length ? format_track_length($_->length) : '',
             $_->id
-                for @tracks;
+        } for @tracks;
     }
 
     # Easy case: there is one disc ID, we have exactly the correct set of
@@ -135,9 +136,9 @@ for my $medium (@mediums)
             if ($bad == 0) {
                 # All track lengths are wrong, so we change them with a
                 # SetTrackLengths edit
-                printf "Set track durations from CDTOC #%d for medium #%d\n",
+                log_info { sprintf 'Set track durations from CDTOC #%d for medium #%d',
                     $cdtoc->id, $medium->id
-                        if $verbose;
+                } if $verbose;
 
                 unless ($dry_run) {
                     Sql::run_in_transaction(sub {
@@ -198,7 +199,7 @@ for my $medium (@mediums)
             $sqdiff /= $num_tracks;
             $sqdiff = sqrt($sqdiff) / 1000;
 
-            print "Skew for @$p = $sqdiff\n" if $debug;
+            log_debug { "Skew for @$p = $sqdiff" } if $debug;
             push @skew, $sqdiff;
         }
 
@@ -217,7 +218,7 @@ for my $medium (@mediums)
             $sqdiff /= $num_tracks;
             $sqdiff = sqrt($sqdiff) / 1000;
 
-            print "Skew for existing tracks = $sqdiff\n" if $debug;
+            log_debug { "Skew for existing tracks = $sqdiff" } if $debug;
 
             if ($sqdiff < 5) {
                 unless (@tracks) {
@@ -270,25 +271,25 @@ for my $medium (@mediums)
         }
     }
 
-    printf "Don't know what to do about medium #%d\n", $medium->id;
-    print " - multiple TOCs\n" if @cdtocs > 1 and keys(%c) == 1;
-    print " - multiple conflicting TOCs\n" if @cdtocs > 1 and keys(%c)>1;
-    print " - no TOCs with correct track count\n" if @cdtocs == 0;
+    log_warning { sprintf 'Don\'t know what to do about medium #%d', $medium->id };
+    log_warning { ' - multiple TOCs' } if @cdtocs > 1 and keys(%c) == 1;
+    log_warning { ' - multiple conflicting TOCs' } if @cdtocs > 1 and keys(%c)>1;
+    log_warning { ' - no TOCs with correct track count' } if @cdtocs == 0;
 
     if (keys(%c) == 1) {
         my $ideal_track_count = $cdtocs[0]->track_count;
         my $want_tracks = join q(,), 1 .. $ideal_track_count;
         my $have_tracks = join q(,), sort { $a<=>$b } map { $_->position }
             @tracks;
-        print " - got tracks $have_tracks\n" if $want_tracks ne $have_tracks;
+        log_warning { " - got tracks $have_tracks" } if $want_tracks ne $have_tracks;
     }
 
     my $withlength = grep { $_->length && $_->length > 0 } @tracks;
-    print " - $withlength tracks have length\n" if $withlength;
+    log_warning { " - $withlength tracks have length" } if $withlength;
 }
 
-print localtime() . " : Fixed $tracks_fixed tracks on $mediums_fixed mediums\n";
-print localtime() . " : ($tracks_set had no previous length)\n";
+log_info { "Fixed $tracks_fixed tracks on $mediums_fixed mediums" };
+log_info { "($tracks_set had no previous length)" };
 
 =head1 COPYRIGHT AND LICENSE
 
--- a/admin/run
+++ b/admin/run
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+set -e -u
+
+SCRIPT_NAME=$(basename "$0")
+
+if ((BASH_VERSINFO[0] < 4))
+then
+  echo >&2 "$SCRIPT_NAME: at least version 4 of bash is required"
+  echo >&2 "Make sure this version of bash comes first in \$PATH"
+  exit 69 # EX_UNAVAILABLE
+fi
+
+MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)
+cd "$MB_SERVER_ROOT"
+
+HELP=$(cat <<EOH
+Usage: $SCRIPT_NAME <command>
+
+This script is meant to be used in production only, for both cron
+tasks and occasional run of script in the appropriate container.
+
+Run the given command with 'carton', after:
+
+1. Setting the environment variables PATH AND PERL_CARTON_PATH;
+2. Setting the current working directory to musicbrainz-server;
+3. Checking that the corresponding log subdirectory exists;
+4. Sudoing to the user 'musicbrainz' if it exists and if not current;
+5. Timestamping and redirecting the output to a file in the
+   corresponding log subdirectory. For any command other than
+   'admin/cron/*.sh', the log subdirectory will be 'occasionally'.
+EOH
+)
+
+if [[ $# -ne 0 && $1 =~ ^-*h(elp)?$ ]]
+then
+  echo "$HELP"
+  exit 0 # EX_OK
+elif [[ $# -eq 0 ]]
+then
+  echo >&2 "$SCRIPT_NAME: missing argument"
+  echo >&2 "Try '$SCRIPT_NAME help' for usage."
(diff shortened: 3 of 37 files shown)


mwiencek:
> Scripts ran from crontab do timestamp some messages but not all of them, and - more problematic - they do not timestamp error messages.

Consistent timestamps on stderr would be nice.

> the timestamps added by scripts run using cron are redundant and thus removed here.

The concern I have with this is that the timestamps from the application are accurate, whereas the timestamps from ts are just from whenever Perl flushes stdout, which is normally buffered (unless you're outputting to a terminal, in which case it's line-buffered, or unless you enable autoflush, which we do in some scripts but not all). I've found the timestamps useful when comparing last_updated columns in the DB for json dumps or sitemaps, so if they're skewed it can make matching those up more iffy.

Here's an example to demonstrate my point

```Perl
#!/usr/bin/env perl

# test.pl

use Time::HiRes qw( sleep );

for (my $i = 0; $i < 10; $i++) {
        print localtime . " - $i\n";
        sleep 0.5;
}
```

```sh
$ ./test.pl 2>&1 | ts \%FT\%TZ > test.txt
$ cat test.txt
2021-09-20T15:59:48Z Mon Sep 20 15:59:43 2021 - 0
2021-09-20T15:59:48Z Mon Sep 20 15:59:43 2021 - 1
2021-09-20T15:59:48Z Mon Sep 20 15:59:44 2021 - 2
2021-09-20T15:59:48Z Mon Sep 20 15:59:44 2021 - 3
2021-09-20T15:59:48Z Mon Sep 20 15:59:45 2021 - 4
2021-09-20T15:59:48Z Mon Sep 20 15:59:45 2021 - 5
2021-09-20T15:59:48Z Mon Sep 20 15:59:46 2021 - 6
2021-09-20T15:59:48Z Mon Sep 20 15:59:46 2021 - 7
2021-09-20T15:59:48Z Mon Sep 20 15:59:47 2021 - 8
2021-09-20T15:59:48Z Mon Sep 20 15:59:47 2021 - 9
```

Now, we could enable autoflush on stdout and stderr in every Perl script, which helps. In some scripts we already do this for stdout (via `$| = 1`), though most scripts don't.  I'm not sure I like this solution much though: buffering is done for a reason (avoiding unnecessary IO) and having the application generate the timestamp seems ideal to me in either case.  Would having `MusicBrainz::Server::Log` add the timestamps (so we don't have to repeat `localtime` in the Perl at least) and only using `ts` for stderr make sense?

yvanzo:
> Now, we could enable autoflush on stdout and stderr in every Perl script, which helps. In some scripts we already do this for stdout (via `$| = 1`), though most scripts don't.

It is currently used in the following scripts only:
```shell
admin/MBImport.pl
admin/RunReports.pl
admin/replication/ImportReplicationChanges
admin/replication/ProcessReplicationChanges
admin/sql/updates/20130313-bcrypt-passwords.pl
admin/sql/updates/20130906-materialise-recording-lengths.pl
lib/MusicBrainz/Script/DatabaseDump.pm
lib/MusicBrainz/Server/Sitemap/Builder.pm
```

> I'm not sure I like this solution much though: buffering is done for a reason (avoiding unnecessary IO)

It seems the main IO issue currently is dump builders writing large files and moving them around. Would these extra IO really make a difference in the case of scripts with limited logs being run occasionally?

Otherwise, maybe we could buffer the output of `ts` before writing to file, e.g. using [`sponge`](https://manpages.ubuntu.com/manpages/xenial/en/man1/sponge.1.html) from the same package `moreutils`?

> and having the application generate the timestamp seems ideal to me in either case.

It is more obvious to me for client applications like Picard that do not run in a controlled environment. But it seems that timestamps can be delegated otherwise as all logging systems handle timestamps (docker, systemd, syslogd, dmesg).

>  Would having `MusicBrainz::Server::Log` add the timestamps (so we don't have to repeat `localtime` in the Perl at least) and only using `ts` for stderr make sense?

Using `ts`for `stderr` at least makes sense since it is auto-flushed by default in Perl.

But adding timestamps via `MusicBrainz::Server::Log` would add timestamps in server logs too, which would be redundant with existing timestamps from docker container logs.

mwiencek:
> It seems the main IO issue currently is dump builders writing large files and moving them around. Would these extra IO really make a difference in the case of scripts with limited logs being run occasionally?

It probably wouldn't make much difference.

> It is more obvious to me for client applications like Picard that do not run in a controlled environment. But it seems that timestamps can be delegated otherwise as all logging systems handle timestamps (docker, systemd, syslogd, dmesg).

Well, I'm thinking about the situations outside of these logging systems...the two I have in mind are:

 * Piping cron scripts' output to files.  Sometimes I have to run these scripts (ModBot, json dumps, and others) by hand, and it's nice to have the timestamps always outputted in that case.  After the changes here, when I have to run these manually I'll have to remember to include `| ts`, and probably have to look up the crontab first to get the format right.  Maybe the answer here is to have a wrapper script that does these things...though ensuring every script properly unbuffers stdout does seem kinda annoying still.
 * Running plackup locally in development.  The logged requests already included timestamps like `127.0.0.1 - - [21/Sep/2021:16:39:03 -0500]` and it'd be nice if `[warning]` and `[error]` lines did too.  (They don't currently, but it'd be nice if they did!)  Is the answer here another wrapper script that does `| ts`?

> But adding timestamps via MusicBrainz::Server::Log would add timestamps in server logs too, which would be redundant with existing timestamps from docker container logs.

If you use the `-t` switch, I guess. :)  But it doesn't seem uncommon for applications to output their own timestamps: just looking at a few containers on aretha: consulagent, cron, keydb, postgres, and redis all output their own timestamps.

I'd love to have something easy and consistent, and perhaps what you have here is the correct direction -- I'm just a bit uncertain how I feel about having to rely on an external tool in every case, and making sure stdout is always unbuffered.

If @alastair has a moment to chime in, I'd be really interested to hear if the python projects handle this in any particular way!

yvanzo:
> If you use the `-t` switch, I guess. :) But it doesn't seem uncommon for applications to output their own timestamps: just looking at a few containers on aretha: consulagent, cron, keydb, postgres, and redis all output their own timestamps.

By the way, I wrote a script to collect timestamped logs from containers of interest given a time interval, I should probably clean it and add it to our docker server configs.

alastair:
We have logs made by our startup (runit) scripts, which log with a date and time: https://github.com/metabrainz/listenbrainz-server/blob/d24f1a372a53befd4dfaee3c65b0933ed11fe74e/docker/lb-startup-common.sh#L3

In all of our python scripts, we try and use a standardised python logger instantiated in the webserver, which also includes dates: https://github.com/metabrainz/listenbrainz-server/blob/dfb0fcff6bef52195b3c5fd9a2bea0804420a0b4/listenbrainz/__init__.py#L11

If cron runs a python script then it'll have the python logging. I don't think we have any specific date logging in bash scripts that are called by python ([for example](https://github.com/metabrainz/listenbrainz-server/blob/563bdd9c803d2ae763f9e1ba18e87bcd14ea8ef2/admin/create-dumps.sh))

yvanzo:
> * Piping cron scripts' output to files.  Sometimes I have to run these scripts (ModBot, json dumps, and others) by hand, and it's nice to have the timestamps always outputted in that case.  After the changes here, when I have to run these manually I'll have to remember to include `| ts`, and probably have to look up the crontab first to get the format right.  Maybe the answer here is to have a wrapper script that does these things...

When I run one of these scripts by hand in the container `musicbrainz-production-cron`, I have a cheat-sheet to run commands:
```shell
export PERL_CARTON_PATH=/home/musicbrainz/carton-local
cd $HOME/musicbrainz-server
carton exec -- admin/RunReports.pl BadAmazonURLs > $HOME/log/occasionally/$(date --utc +\%FT\%TZ).log 2>&1
```

>   though ensuring every script properly unbuffers stdout does seem kinda annoying still.

It seems less annoying than ensuring every script properly logs, given scripts are heterogeneous (Bash & Perl, plus called commands: rsync, tar, xz…).

> * Running plackup locally in development.  The logged requests already included timestamps like `127.0.0.1 - - [21/Sep/2021:16:39:03 -0500]` and it'd be nice if `[warning]` and `[error]` lines did too.  (They don't currently, but it'd be nice if they did!)  Is the answer here another wrapper script that does `| ts`?

Running `plackup` locally in `musicbrainz-docker` development setup provides timestamps for both standard output and standard error.

Here is a wrapper script that could be used for running manually either scripts in production or `plackup` in development; See <https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/ed5d3f326fc4464f228ba43fbebe97c96f3b5fd8#diff-9f58197316593d0c2f58373f4ca464a24d61f8c193de27c7f386e9e6bf545dcd>.

admin/RunReports.pl
```diff
@@ -17,13 +18,13 @@
...
     unless (any { $name =~ /$_/i } @ARGV) {
-        print localtime() . " : Not running $name\n";
+        log_debug { "Not running $name" };
         next;
     }
 
     my $report = MusicBrainz::Server::ReportFactory->create_report($name, $c);
 
-    print localtime() . " : Running $name\n";
+    log_info { "Running $name" };
```
mwiencek, line 27: Prefixing all messages with `[info]` doesn't look nice or seem very useful.  Debug, warning, and error are the exceptions.  Should we start omitting the log level for informational messages?

mwiencek (commented):
What I'm ambivalent about:

 * Adding all timestamps with an external tool makes me feel slightly uneasy due to possible input/output buffering.  (The application/script is the most trustworthy source for the timestamps.)
 * Having to wrap all script calls in `script/timestamp` is not a nice administrative experience IMO -- carton exec is bad enough.  It makes sense for cron, but I'd like if manual script runs didn't require a cheat sheet.

What I like here:

 * Consistently using `MusicBrainz::Server::Log`.
 * Standardizing the way we output timestamps everywhere.
 * Outputting timestamps for things like external commands and unhandled exceptions.

What if we had `MusicBrainz::Server::Log` prefix all messages with timestamps, and had `script/timestamp` only prepend timestamps to lines that didn't already have them?  This way you'd still get timestamps by default (minus external commands and unhandled exceptions) and we wouldn't have to muck with or worry about autoflush.

lib/MusicBrainz/Script/DatabaseDump.pm
```diff
@@ -89,7 +89,7 @@ around begin_dump => sub {
     $self->$orig;
 
     $| = 1;
```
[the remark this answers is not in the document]
yvanzo: Added commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/a9f0ec3a99d0b8d1edf056c97fdf9261550330f0

script/timestamp
```diff
@@ -0,0 +1,35 @@
...
+  echo "$HELP"
+  exit 0 # EX_OK
+elif [[ $# -eq 0 ]]
+then
+  echo >&2 "$SCRIPT_NAME: missing command"
+  echo >&2 "Try '$SCRIPT_NAME help' for usage."
+  exit 64 # EX_USAGE
+fi
+
+exec "$@" 2> >(TZ=Z ts %FT%.TZ) 1> >(TZ=Z ts %FT%.TZ)
```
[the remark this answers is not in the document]
yvanzo: Good question. Yes, it is guaranteed that neither `bash` redirects nor `ts` have buffering.

admin/RunReports.pl
yvanzo: It doesn’t seem very useful indeed but not that bad either, it can even be useful to grep against: `grep -vF '[info]'`. There is no option in [`Log::Dispatch`](https://metacpan.org/pod/Log::Dispatch) to omit the log level for informational messages, even through callbacks. It would require to extend this class instead of just instantiating it. It looks like an increased complexity for a small eye-candy improvement to me.

yvanzo:
> * Adding all timestamps with an external tool makes me feel slightly uneasy due to possible input/output buffering.  (The application/script is the most trustworthy source for the timestamps.)

I understand this concern. Unbuffering log stdout with commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/a9f0ec3a99d0b8d1edf056c97fdf9261550330f0 should address it. Also note that `ts` itself is written in Perl too, so it is a limited overhead.

> * Having to wrap all script calls in `script/timestamp` is not a nice administrative experience IMO -- carton exec is bad enough.  It makes sense for cron, but I'd like if manual script runs didn't require a cheat sheet.

As stated above, it already requires a cheat sheet not only to use `carton` but also to write logs in `occasionally`. A more general wrapper can be added to do it all if that helps with improving the administrative experience.

> What if we had `MusicBrainz::Server::Log` prefix all messages with timestamps, and had `script/timestamp` only prepend timestamps to lines that didn't already have them?

If timestamps were formatted the same then it would be possible, I’m not sure this is the case.

> This way you'd still get timestamps by default (minus external commands and unhandled exceptions) and we wouldn't have to muck with or worry about autoflush.

It would be a big _minus_. Why would you want not to have timestamps for external commands and unhandled exceptions? This is often the most informative logs: When something unexpected happens during a long process.

admin/RunReports.pl
mwiencek: I thought this is what `_prefix_message` in `MusicBrainz::Server::Log` does -- maybe I'm misreading that code.

I'm sure I could live with it, but the status quo is that we omit `[info]` in most cases (and will continue to omit it for bash scripts).  So if there's an easy way to omit it `_prefix_message` and keep the status quo, that would be nice!

mwiencek:
> As stated above, it already requires a cheat sheet not only to use carton but also to write logs in occasionally. A more general wrapper can be added to do it all if that helps with improving the administrative experience.

This sounds like a good idea.

> If timestamps were formatted the same then it would be possible, I’m not sure this is the case.

If my note about `_prefix_message` above is correct, my thinking would be that we'd just format the timestamps identically ourselves (in `_prefix_message`).

> It would be a big minus. Why would you want not to have timestamps for external commands and unhandled exceptions? This is often the most informative logs: When something unexpected happens during a long process.

I'd just like to continue to have the timestamps we have now when I'm running scripts on development/mirror servers, without having to wrap every script I call.  (I don't use carton locally either.)

I do want timestamps for external commands and unhandled exceptions in production, and we'd have them there: cron runs would use `script/timestamp`, and if we add a helper script to combine all the production-related carton/logging incantations, then manual runs in prod should have them too.

yvanzo:
I replaced `timestamp` with a more general `run` wrapper to run scripts in production, that can be used even occasionally; See https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/bbf57a087ad4c6f043ccf659c584101048665e18#diff-7e8713025422748b67407b3804bfadae6f53053be08e39e15e99759453f74379.

For development purpose, I’ve set Perl logger and bash scripts to add local relative timestamps when outputting directly on the terminal; See https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/dd50ef5408ae991c5060a439050900b6ed3f45a6#diff-2b6a5688b53933fdef13587713e4d660e99b1dfa62a1a7d81c4e7882200b1cf3 and https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/9486787404f837c252679aa4fba00049c5bcbf50#diff-851c75069f48060ed9f568df409d60c28ece935d08266e44cdc1df91c0abbd9b.

mwiencek:
This looks good though I didn't really get a chance to test it, will do so this week!

admin/cleanup/FixTrackLength.pl
```diff
@@ -58,10 +59,10 @@
        JOIN track t ON t.medium = m.id
       WHERE t.length IS NULL OR t.length = 0 AND m.track_count > 0
         AND (mf.has_discids = TRUE OR mf.has_discids IS NULL)'
-) };
-printf localtime() . " : Found %d medium%s\n",
```
mwiencek, line 62: There seem to be some syntax errors introduced in this script, but I didn't full investigate: try running `./admin/cleanup/FixTrackLength.pl --dry-run`.

lib/MusicBrainz/Script/RemoveEmpty.pm
```diff
@@ -153,12 +154,10 @@ sub run {
...
-        printf "%s : Found %d unused $entity%s.\n",
-            scalar localtime,
-            $count, ($count==1 ? '' : 's');
-        printf "%s : Successfully removed %d $entity%s\n",
-            scalar localtime,
-            $removed, ($removed==1 ? '' : 's')
+        log_info { sprintf 'Found %d unused $entity%s.',
+            $count, ($count==1 ? '' : 's') };
+        log_info { sprintf 'Successfully removed %d $entity%s',
+            $removed, ($removed==1 ? '' : 's') }
```
mwiencek, line 160: Since switching doubles quotes to single quotes, `$entity` is no longer interpolated.

docker/templates/macros.m4
```diff
@@ -71,6 +71,7 @@ libicu66 m4_dnl
 libpq5 m4_dnl
 libssl1.1 m4_dnl
 libxml2 m4_dnl
+moreutils m4_dnl
```
mwiencek, line 74: Thanks -- I didn't have this installed on my local system, so it helped me figure out what package I needed. :)

I saw you mention you didn't modify any scripts for mirror servers, so it sounds like this shouldn't create any additional dependency for people outside of production.

mwiencek (changes requested):
Sorry for the delay!  I had to keep coming back to this over multiple days because a lot of these scripts don't have tests and I wanted to check them carefully.  But overall this seems great.

admin/replication/BundleReplicationPackets
```diff
@@ -11,6 +11,7 @@ use Getopt::Long;
 use File::Temp qw( tempdir );
 use File::Path qw( rmtree );
 use File::Copy qw( copy );
+use MusicBrainz::Server::Log qw( log_info );
```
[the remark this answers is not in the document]
yvanzo: Restored in commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/51d7fd0c4545bfe5b2282b836b523c1c08cd1d9d.

admin/RunReports.pl
```diff
@@ -17,13 +18,13 @@
 my $errors = 0;
 for my $name (MusicBrainz::Server::ReportFactory->all_report_names) {
     unless (any { $name =~ /$_/i } @ARGV) {
-        print localtime() . " : Not running $name\n";
+        log_debug { "Not running $name" };
```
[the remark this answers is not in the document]
yvanzo: Imported in commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/93e0057975d10cb95cae1ab71972431f360fad7e.

admin/run
```diff
@@ -0,0 +1,74 @@
...
+  echo >&2 "Make sure to run in the appropriate container."
+  exit 73 # EX_CANTCREAT
+fi
+log_timestamp=$(date --utc +%FT%TZ)
+log_file_path="$log_dir_path/$log_timestamp.log"
+
+echo "Running '$*' and logging at '$log_file_path'..."
+if [[ -t 1 ]]
+then
+  exec carton exec -- "$command_full_path" "$@" 2>&1 | TZ=Z ts %FT%.TZ | tee -a "$log_file_path"
```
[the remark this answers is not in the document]
yvanzo: Sudo to the user `musicbrainz` if it exists in commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/ff21883200593e8daada378a47087737c90657b8, run with the current user otherwise.
This is because the `musicbrainz` user may not exist in development setup; It doesn’t exist in `musicbrainz-docker` for now.

admin/cleanup/FixTrackLength.pl
yvanzo: Good catch, fixed all of those in commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/cf59e5b53fd15776698710ebe9902ac348e5ba2b.

lib/MusicBrainz/Script/RemoveEmpty.pm
yvanzo: Restored double quotes in commit https://github.com/metabrainz/musicbrainz-server/pull/2272/commits/a0893dff205833db03c013d73c7dff9754f5efb6.

docker/templates/macros.m4
yvanzo: Right, it is not required for mirrors, but it is required for development if one wants to test running scripts using `ts`, so I opened https://github.com/metabrainz/musicbrainz-docker/pull/221.

yvanzo:
@mwiencek: Thanks for reviewing, I addressed every reported issue, please close the conversations you opened if you are satisfied with the answers.

mwiencek (approved):

yvanzo:
Rebased to apply amending commits.

PR #2273, opened 2021-09-20
MBS-11978: Move edit icons of external links editor to the left side

y-young:
### Implement MBS-11978
Move edit icons of external links editor to the left side

PR #2275, opened 2021-09-20
MBS-11919: Remove numberings from ext links editor

y-young:
### Implement MBS-11919
Remove link index numbers for simplicity, add a link in duplicate notice to jump to the referenced link and display highlight.

> This PR depends on #2273.


yvanzo (commented):
There seems to be a display issue: URL doesn’t vertically align with relationship type and is almost displayed vertically instead (1 or 2 characters per line). Some `colspan` adjustments might be needed here.

y-young:
> There seems to be a display issue: URL doesn’t vertically align with relationship type and is almost displayed vertically instead (1 or 2 characters per line). Some `colspan` adjustments might be needed here.

Could you provide a screenshot please?

yvanzo:
From [Nine Inch Nails’ artist editing page](http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/edit):
screenshot-localhost_5000-2021 09 22-16_14_40

y-young:
Sorry, it went wrong during rebasing, now it should be all right.

reosarevok (approved):
This seems fine to me now, other than my comments on https://github.com/metabrainz/musicbrainz-server/pull/2267#issuecomment-921671920 but if that's hard to fix, let's get this in for now :)

yvanzo:
I might be doing something wrong but I still have the same issue as for <https://github.com/metabrainz/musicbrainz-server/pull/2267>:
> The link pointing to the position does not seem to work as clicking on it doesn’t scroll to the target item. Maybe adding an highlighting border around the item at the target position would be help with finding it.

reosarevok:
The highlighting and whatnot worked for me *except* if the link is an autoselected duplicate, in which case it just merges them instead.

yvanzo:
Indeed I tested duplicating `https://nineinchnails.tumblr.com/` with auto-selected relationship type “blogs” for Nine Inch Nails.

reosarevok:
Yeah, so that's apparently because it's hard to detect that situation and make it not automerge - ideally IMO we'd catch that case and either just reject it, or at least not merge it at all until the already-exists error is fixed (by editing the URL or something) if we want to avoid rejecting it outright in case the user wants to change the URL for some reason.

y-young:
> The highlighting and whatnot worked for me _except_ if the link is an autoselected duplicate, in which case it just merges them instead.

This is fixed in #2288 .

PR #2276, opened 2021-09-21
MBS-11971: Always load URL rels when loading paged subsets

reosarevok:
### Fix MBS-11971

This is needed in order for them to still show on the sidebar for pages like /artist/mbid/relationships?link_type_id=168
that would otherwise only load that specific link type. The issue was introduced by https://github.com/metabrainz/musicbrainz-server/pull/1953/commits/12b43e8330b0feda64c636665735c7c3c931b7b5

Tested manually on the page listed on the ticket.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Load.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Load.pm
@@ -138,7 +138,10 @@ role
                 $loaded = 1;
             }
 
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
Tested with NiN artist page using sample data: http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/relationships?direction=1&link_type_id=168&page=1

mwiencek (approved):

PR #2277, opened 2021-09-21
MBS-11981: Don't break Ko-fi shop URLs get in URL cleanup

reosarevok:
### Fix MBS-11981

It's not fully clear to me whether these can only be purchase for mail order, can also be a download, or maybe even more options, so for now this just avoids cleaning them up / selecting any specific type for them

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2421,7 +2421,7 @@ const CLEANUPS: CleanupEntries = {
     },
   },
   'kofi': {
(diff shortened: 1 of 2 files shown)


reosarevok:
Sigh, my bad. Fixed. I didn't see any reason why `[^/?#]` was needed for the matching so I just got rid.

yvanzo (approved):
From the original commit 97db62b8fe924d1503acbedbc5f9567d41e040e6, it seems it was just mimicking older code. The reason was to require at least one extra character in the path part. It makes sense for websites that can be seen as their own label for example. It doesn’t seem useful in the case of Ko-Fi.

mwiencek (approved):

PR #2279, opened 2021-09-22
MBS-11979: Improve example URL

yvanzo:
# Problem

MBS-11979: The previous example `google.com` was a valid URL but search pages are not actually wanted to link to any MB entity.
    

# Solution

This patch replaces it with `example.com` to get rid of this ambiguity in both editing UI and tests. It also tries to improve the phrasing and the formatting.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -569,6 +569,15 @@ export class ExternalLinksEditor
                        {example_url: <span className="url-quote">{'http://example.com/'}</span>}),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
+    } else if (isNewOrChangedLink && isExample(link.url)) {
+      error = {
+        message: exp.l(
+          `“{example_url}” is just an example.
+          Please enter the actual link you want to add.`,
+          {example_url: <span className="url-quote">{link.url}</span>},
+        ),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
     } else if (isNewOrChangedLink && isMusicBrainz(link.url)) {
       error = {
         message: l(`Links to MusicBrainz URLs are not allowed.
@@ -1592,6 +1601,10 @@ function isShortened(url) {
 
 function isGoogleAmp(url) {
   return /^https?:\/\/([^/]+\.)?google\.[^/]+\/amp/.test(url);
+}
+
+function isExample(url) {
+  return /^https?:\/\/(?:[^/]+\.)?example\.(?:com|org|net)(?:\/.*)?$/.test(url);
 }
 
 function isMusicBrainz(url) {
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -565,7 +565,8 @@ export class ExternalLinksEditor
       };
     } else if (isNewOrChangedLink && !isValidURL(link.url)) {
       error = {
-        message: l('Enter a valid url e.g. "http://google.com/"'),
+        message: exp.l('Please enter a valid URL, such as “{example_url}”.',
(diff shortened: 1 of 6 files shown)


reosarevok (approved):
The changes seem sensible to me. We might want to actually block example.com in validation (like we do for MusicBrainz links) and have a message reminding users that it's an example only :D

root/static/scripts/edit/externalLinks.js
```diff
@@ -549,6 +549,12 @@ export class ExternalLinksEditor
                        {example_url: <span className="url-quote">{'http://example.com/'}</span>}),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
+    } else if (isNewOrChangedLink && isExample(link.url)) {
+      error = {
+        message: l(`Example links are not allowed.
+                Did you mean to paste something else?`),
```
reosarevok, line 555: Indent under Example I guess?

You could also say "{example_url} is just an example. Did you mean to paste something else?" or even "{example_url} is just an example. Please enter the actual link you want to add." But I'm also ok with this, especially since you're blocking other example domains :)

root/static/scripts/edit/externalLinks.js
```diff
@@ -1447,6 +1453,10 @@ function isGoogleAmp(url) {
   return /^https?:\/\/([^/]+\.)?google\.[^/]+\/amp/.test(url);
 }
 
+function isExample(url) {
+  return /^https?:\/\/(?:[^/]+\.)?example\.\w+(?:\/.*)$/.test(url);
```
reosarevok, line 1457: Keep in mind that *technically* only .com, .org, .edu and .net are actually official example URLs - example.es or example.fr could be legit addresses.

yvanzo: Right but they seem to be currently parked by ads, no chance it could be legit any time soon. ;)

root/static/scripts/edit/externalLinks.js
yvanzo: Used your last suggestion for commit <https://github.com/metabrainz/musicbrainz-server/pull/2279/commits/49bcd9ad7aa1cdabbd372e9912a4417340c30b53>. Thanks!

root/static/scripts/edit/externalLinks.js
```diff
@@ -545,7 +545,17 @@ export class ExternalLinksEditor
...
+                       {example_url: <span className="url-quote">{'http://example.com/'}</span>}),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
+    } else if (isNewOrChangedLink && isExample(link.url)) {
+      error = {
+        message: exp.l(
+          `“{example_url}” is just an example.
+          Please enter the actual link you want to add.`,
+          {example_url: <span className="url-quote">{link.url}</span>},
+        ),
```
mwiencek, line 558: Do we think people are going to try to enter `http://example.com/` just because the previous message references it?

I think they'd have to be incredibly confused about what they're doing to attempt that -- so much so that I dunno if having a dedicated message for this will suddenly enlighten them. :D

reosarevok: Well, given a bunch of people seemed to enter the Google link from the previous example, it can't hurt to play it safe...

mwiencek (approved):
I'm not sure about the utility of the message but the code looks fine.

root/static/scripts/edit/externalLinks.js
yvanzo: I agree it might be seen as insulting the intelligence of editors, but mistakes happen and it seems worth explaining why the URL is being rejected.

mwiencek:
Looks like the selenium test failure is legitimate, since it's using "http://example.com/" in the deprecated-URL test. We can probably modify that to test both cases (that it gets the new example error, and then change the MusicMoz one to an actual MusicMoz URL).

PR #2280, opened 2021-09-22
MBS-11977: Fix targeting ENTITY for URL relationship error

yvanzo:
# Problem

MBS-11977: Mismatched entity type in external links editor doesn't trigger the appropriate error message

(see the ticket for details)

# Solution

Fix targeting `ENTITY` for URL relationship error.

It also required adding a generic error message for `ENTITY` error target (which was associated to custom error message only so far).

# Notes

Further improvement would be to specify which entity type should be linked to the entered URL, and which URL pattern would be suitable for the current entity, as proposed in “Implementation notes” of the ticket. Actually, it is probably worth a separate ticket.

# Checklist for author

* [ ] Rebase on `master` if #2263 is merged first

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -551,32 +551,32 @@ export class ExternalLinksEditor
     } else if (isNewOrChangedLink && isMusicBrainz(link.url)) {
       error = {
         message: l(`Links to MusicBrainz URLs are not allowed.
-                Did you mean to paste something else?`),
+                    Did you mean to paste something else?`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
     } else if (isNewOrChangedLink && isMalware(link.url)) {
       error = {
         message: l(`Links to this website are not allowed
-                because it is known to host malware.`),
+                    because it is known to host malware.`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
     } else if (isNewOrChangedLink && isShortened(link.url)) {
       error = {
         message: l(`Please don’t enter bundled/shortened URLs,
-                enter the destination URL(s) instead.`),
+                    enter the destination URL(s) instead.`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
     } else if (isNewOrChangedLink && isGoogleAmp(link.url)) {
       error = {
         message: l(`Please don’t enter Google AMP links,
-                since they are effectively an extra redirect.
-                Enter the destination URL instead.`),
+                    since they are effectively an extra redirect.
+                    Enter the destination URL instead.`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
     } else if (!link.type) {
       error = {
         message: l(`Please select a link type for the URL
-                you’ve entered.`),
+                    you’ve entered.`),
         target: URLCleanup.ERROR_TARGETS.RELATIONSHIP,
       };
     } else if (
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/externalLinks.js
```diff
@@ -615,6 +616,14 @@ export class ExternalLinksEditor
           error.message = l(`This URL is not allowed 
                     for the selected link type.`);
         }
+        if (error.target ===
+          URLCleanup.ERROR_TARGETS.ENTITY) {
+          error.message = exp.l(
+            `This URL is not allowed for this type of entity:
+             {entity_type_name}.`,
```
reosarevok, line 623: Maybe "This URL is not allowed for “{entity_type_name}” entities" is more straightforward? (while still being translatable).

reosarevok:

reosarevok (approved):
Seems sensible, in general. I do wonder whether this error should only be shown if the link is valid for *other* entity type, but I expect in most cases this is going to be fine.

root/static/scripts/edit/externalLinks.js
yvanzo: Applied with https://github.com/metabrainz/musicbrainz-server/pull/2280/commits/277ad5922874f54f4f9077b13d8ac195aaec2764 commit.

yvanzo:
Yes it should “be fine in most cases” and `URL` should probably be the target instead “if the link is _not_ valid for any other entity type”.

root/static/scripts/edit/externalLinks.js
```diff
@@ -615,6 +616,13 @@ export class ExternalLinksEditor
           error.message = l(`This URL is not allowed 
                     for the selected link type.`);
         }
+        if (error.target ===
+          URLCleanup.ERROR_TARGETS.ENTITY) {
```
mwiencek, line 620:
```suggestion
        if (error.target === URLCleanup.ERROR_TARGETS.ENTITY) {
```

root/static/scripts/edit/externalLinks.js
```diff
@@ -615,6 +616,13 @@ export class ExternalLinksEditor
...
+          error.message = exp.l(
+            `This URL is not allowed for “{entity_type_name}” entities.`,
+            {entity_type_name: formatEntityTypeName(checker.entityType)},
```
mwiencek, line 623: I think a separate string for each entity type (like we do for SEARCH_PLACEHOLDERS, "Add a new x," "Add another x," etc. would be nice here; saying `“artist” entities` instead of `artists` just looks and sounds awkward. Like `Yesterday I watched a movie of the genre “horror”` or something. :laughing:

reosarevok: That is fair, it shouldn't be a crazy amount of strings and they should be easy to translate anyway :)

yvanzo: Thanks, I was just too lazy here. Split in the commit <https://github.com/metabrainz/musicbrainz-server/pull/2280/commits/cb27c59e6d85b62f1ef77aa4129f637c6f9fd1e6>.

mwiencek (approved):
Skimming the changes, they look reasonable, though I can't test all the changes here.  I guess the automated tests don't check the expected error target yet?

yvanzo:
> Skimming the changes, they look reasonable, though I can't test all the changes here.

I didn’t test all of them either. I fixed the most common anti-patterns only. Some of the functions `validate` will require larger refactoring, but that will be more closely double-checked when improving error messages.

> I guess the automated tests don't check the expected error target yet?

Right, most of the new summer features are roughly tested using Selenium only. The file `root/static/scripts/tests/Control/URLCleanup.js` should be updated with further tests. Just added  MBS-11995. (See also MBS-11980 for rel types restriction.)

PR #2281, opened 2021-09-22
MBS-11974: Fix track relationship grouping for attributes/dates

mwiencek:
This is a follow-up to MBS-11806.

Relationships with different attributes or dates should not be grouped into a single track range.

For example, if we have the relationships:

```
phonographic copyright by: Aniplex Inc. (in 2007) (track 1)
phonographic copyright by: Aniplex Inc. (in 2008) (track 2)
```

we are currently grouping these as:

```
phonographic copyright by: Aniplex Inc. (in 2007, in 2008) (tracks 1–2)
```

which is not ideal, because it's ambiguous which tracks' copyrights are in 2007 and which are in 2008. (But the bug may apply to any type of relationship.)

Instead of unconditionally appending tracks to every similar target group, we first ensure that target groups are not merged for different tracks. Next, once we've run `mergeDatedExtraAttributes` on each target group, we merge ones with equal date/attribute lists, combining their track sets. This is done in `mergeTargetGroupsByTracks`.


reosarevok:
Seems to work, although on the example release (/release/a1b57866-9366-49f8-97e2-e551ea21f500):

`Warning: Encountered two children with the same key, `92196Aniplex Inc.00`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

reosarevok:

mwiencek:
Fixed the key warning, which was due to the generated `key` on `RelationshipTargetGroupT` not accounting for the list of tracks. Instead of adding the tracks there, I just generate a unique key for each group. And instead of using the key for comparisons, we just run the comparator function again. The key is now only used for the React key prop. This required some minor refactoring.

reosarevok (approved):

PR #2283, opened 2021-09-23
Enable more Perl::Critic conditions

reosarevok:


mwiencek (approved):

mwiencek:
There's a difference in the implementation of `nsort_by` in UtilsBy vs SomeUtils causing the test failure. UtilsBy does `return @_[ ... ]` while SomeUtils does `return map { ...} @list`. The former is returning an array slice while the latter is returning an array, and there's a difference between these when evaluated in scalar context: "[s]lices in scalar context return the last item of the slice,"  while arrays in scalar context return the length of the array. https://perldoc.perl.org/perldata

We could easily fix the assignment in `_serialize_release_group` to work with the new behavior, but it might be better to make sure we continue to use the same underlying modules where possible for now to avoid further subtle bugs.

reosarevok:
I checked and `sort_by` and `nsort_by` are the only ones in both `List::SomeUtils` and `List::UtilsBy`. `List::AllUtils` says "going forward, we will always prefer backwards compatibility", which suggests that even if `List::SomeUtils` added more `_by` functions, they would still keep the ones in `List::UtilsBy`. As such, we could still choose to use `List::AllUtils` for all, but we'd need to make sure no bugs are caused by the different `sort_by` and `nsort_by` implementation. Alternatively, we could keep importing `sort_by` and `nsort_by` from `List::UtilsBy` and use `List::AllUtils` for everything else. @mwiencek, your preference?

mwiencek:
I think in the long run, always importing from AllUtils will make things easier for us. So I'd just fix the few cases (or only case?) where we're assigning the result to a scalar and use AllUtils.

Here are some stats I gathered earlier when I was looking at how often each utility is imported from each module in our code:

```
List::Util:
      1 any
      3 max
      3 min
      5 sum
     10 first

List::UtilsBy:
      1 rev_sort_by
      3 rev_nsort_by
      9 uniq_by
     14 nsort_by
     14 partition_by
     43 sort_by

List::MoreUtils:
      1 mesh (zip)  (In Util 1.56+ as mesh)
      1 minmax
      3 all         (In Util)
      3 pairwise
      3 zip         (In Util 1.56+ as mesh)
      5 part
      6 any         (In Util)
      8 natatime
     29 uniq        (In Util)

List::AllUtils:
      1 max         (In Util)
      1 pairs       (In Util)
      1 part        (In MoreUtils)
      2 sum         (In Util)
      2 zip         (In Util 1.56+ as mesh)
      3 all         (In Util)
     24 uniq        (In Util)
     32 any         (In Util)
```

reosarevok:
mesh/zip is in SomeUtils *and* Util but AllUtils specialcases that to use SomeUtils to maintain compatibility since they used to have it there only. The SomeUtils version is the same as the MoreUtils one :) (so, it's still fine, but I thought I'd specify)

mwiencek:
I saw, I just wanted to see how many utilities we actually use that aren't in `List::Util` (since that's a standard Perl module).

admin/replication/ConvertReplicationPacketToXML
```diff
@@ -57,43 +57,43 @@ while (my $p_line = shift_p())
 
 
     my $p_row = decode_row($p_line, qw( SeqId TableName Op XID ));
-    my $op = $p_row->{"Op"} or die "No Op in $p_line";
+    my $op = $p_row->{'Op'} or die "No Op in $p_line";
```
mwiencek, line 60: There's no need to use quotes here either way FWIW.  Same for the other cases below.

PR #2285, opened 2021-09-24
MBS-11975: Clean up CDJapan detailview URLs

reosarevok:
### Implement MBS-11975

CDJapan has detailview.html URLs that seem to just redirect to the product ones, but have been added a few times and are not currently detected at all by the code. This changes the CDJapan autoselect to detect them and clean them to their final product page form.

Since this takes care of the only kind of http URLs we still were getting for CDJapan, it should be safe to remove the http override too (existing ones are being fixed by hand).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1315,7 +1315,7 @@ const CLEANUPS: CleanupEntries = {
     ],
     restrict: [LINK_TYPES.patronage],
     clean: function (url) {
-      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?changetip\.com\/tipme\/([^\/?#]+)(?:.*)?$/, 'https://www.changetip.com/tipme/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?changetip\.com\/tipme\/([^\/?#]+).*$/, 'https://www.changetip.com/tipme/$1');
       url = url.replace(/^(?:https?:\/\/)?([^\/?#]+)\.tip\.me(?:[\/?#].*)?$/, 'https://www.changetip.com/tipme/$1');
       return url;
     },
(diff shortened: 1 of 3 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1296,12 +1296,14 @@ const CLEANUPS: CleanupEntries = {
...
-      '^(https?://)?(www\\.)?cdjapan\\.co\\.jp/(product|person)/',
+      '^(https?://)?(www\\.)?cdjapan\\.co\\.jp/(detailview|product|person)',
       'i',
     )],
     restrict: [LINK_TYPES.mailorder],
     clean: function (url) {
-      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?cdjapan\.co\.jp\/(person|product)\/([^\/?#]+)(?:.*)?$/, 'https://www.cdjapan.co.jp/$1/$2');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?cdjapan\.co\.jp\//, 'https://www.cdjapan.co.jp/');
+      url = url.replace(/^https:\/\/www\.cdjapan\.co\.jp\/detailview\.html\?KEY=([^\/?#]+)(?:.*)?$/, 'https://www.cdjapan.co.jp/product/$1');
+      url = url.replace(/^https:\/\/www\.cdjapan\.co\.jp\/(person|product)\/([^\/?#]+)(?:.*)?$/, 'https://www.cdjapan.co.jp/$1/$2');
```
yvanzo, line 1306: * Since the scheme and the domain name are unchanged, they can be captured to shorten the replacement.
* That was in the old code already but the last non-capturing group is unneeded: `(?:.*)?` is the same as `.*`.
```suggestion
      url = url.replace(/^(https:\/\/www\.cdjapan\.co\.jp)\/detailview\.html\?KEY=([^\/?#]+).*$/, '$1/product/$2');
      url = url.replace(/^(https:\/\/www\.cdjapan\.co\.jp)\/(person|product)\/([^\/?#]+).*$/, '$1/$2/$3');
```

yvanzo (commented):
It works but it seems the code can be simplified.

root/static/scripts/edit/URLCleanup.js
reosarevok: Sounds good. I'll remove the other (?:.*)? regexes on a separate commit.

yvanzo (approved):

PR #2286, opened 2021-09-25
MBS-11989: Also correct rg to release_group for delete_alias

reosarevok:
### Fix MBS-11989

We already do this for add_alias and edit_alias, but I guess we forgot to do it for delete_alias, and that is causing an error when loading the layout on React.
Why we use "rg" I don't know, but in any case it seems sensible to have all of these work until we maybe decide to change it.


yvanzo (commented):
I get that it is supposed to fix a bug, but I don’t get why it should be corrected here in the execution flow and not elsewhere. Maybe that could be explained in the commit message?

reosarevok:
Well, there's no bug as such (other than this line being missing), it's just that we use "rg" for the entity_type on alias edits historically for reasons I don't really know, but other parts of the code (including the layout picker) expect release_group. Which is the reason we do this during the alias page building.

yvanzo (approved):
Sorry, not sure why I didn’t see your commit message but the headline.

PR #2287, opened 2021-09-26
MBS-11901: Add highlights to the links in external links editor

y-young:
### Implement MBS-11901
MBS-11901: Add highlights to the links in external links editor


mwiencek:
This seemed to work, but the `link.deleted = !link.deleted;` stuff shouldn't be allowed by Flow. I tried to fix the read-onlyness of the state in https://github.com/metabrainz/musicbrainz-server/pull/2287/commits/139ba283ee49223222907d5ae6a5db10447081d4. Can you re-test the highlighting on your end?

(Unrelated to this PR, but we'll need to perform a lot more refactoring in this file; editing is still extremely slow, as in there's a huge lag after each keystroke if the entity has a lot of links. We're not using React correctly here, unfortunately.)

y-young:
> I tried to fix the read-onlyness of the state in [139ba28](https://github.com/metabrainz/musicbrainz-server/commit/139ba283ee49223222907d5ae6a5db10447081d4). Can you re-test the highlighting on your end?

Yes, it still works for me.

PR #2288, opened 2021-09-27
MBS-11990: External links editor automatically merges auto-selected links even if it's a duplicate

y-young:
### Fix MBS-11990
MBS-11990: External links editor automatically merges auto-selected links even if it's a duplicate


y-young:

y-young:

reosarevok (changes requested):
This still says "Note: This link already exists at position #35. To merge, press enter or select a type." - but it won't merge even if you press enter, since it's a duplicate, and you can't select any other type :)

y-young:
> This still says "Note: This link already exists at position #35. To merge, press enter or select a type." - but it won't merge even if you press enter, since it's a duplicate, and you can't select any other type :)

This is currently handled as a special case, so I'm afraid the error message can't react accordingly.

reosarevok:
How come not? `duplicateNotice` is declared on that same component, so there shouldn't be a problem overriding it or even declaring it later and having a check for canMerge, no? :)

y-young:
> How come not? `duplicateNotice` is declared on that same component, so there shouldn't be a problem overriding it or even declaring it later and having a check for canMerge, no? :)

You're right, I didn't give it a thorough thought. :)

y-young:

mwiencek:
I'm not sure why we need new props/state canMerge and blockMerge. Can't we just always block merging if there's an error? Even if some of the errors are impossible to trigger, it seems confusing to merge or allow merging things if there's an error in any case (since the feedback gets moved out of view of the user). And the extra variables introduce more state/complexity we have to think about. Am I missing something?

mwiencek:
@brainzbot test this please

PR #2289, opened 2021-09-27
Try to fix flaky Selenium tests (fixes MBS-11983)

mwiencek:
Hopefully resolves MBS-11983 and flakiness related to the search server. See associated commits for details.


reosarevok:
Should we really only retry those and not all/most of them?

mwiencek:
> Should we really only retry those and not all/most of them?

Made this change, since I imagine it can only help.

I also pushed another commit that waits for sir queues to empty before proceeding to the next command. We may have to revert this though if it slows down the tests too much.

t/selenium.js
```diff
@@ -794,6 +810,18 @@ async function runCommands(commands, t) {
...
+              const finishTupStats = await getSeleniumDbTupStats();
+              didDatabaseChange = (
+                finishTupStats.tup_inserted > startTupStats.tup_inserted ||
+                finishTupStats.tup_updated > startTupStats.tup_updated ||
+                finishTupStats.tup_deleted > startTupStats.tup_deleted
+              );
+            }
+
+            // eslint-disable-next-line require-atomic-updates
+            shouldCleanSeleniumDb = didDatabaseChange;
```
reosarevok, line 824: https://eslint.org/docs/rules/require-atomic-updates says this can cause race conditions, is this one safe then?

reosarevok (approved):
The general ideas here seem good, and at the very least they'll help figure out what's messed up better, if they don't fix it outright.

t/selenium.js
mwiencek: It looks like a false positive to me, since the tests never run concurrently with each other. Added a comment.

yvanzo (approved):
Thanks a lot
!
The checks against message queues seem reasonable given SIR can be pretty fast after that.

Successfully tested setting `DISABLE_LAST_LOGIN_UPDATE` with `musicbrainz-docker` in development setup and logging in with production database on `pink`.
Will include that in next release too.

PR #2290, opened 2021-09-28
MBS-11995: Test error messages in external links editor

reosarevok:
### Implement MBS-11995

We should ensure we are returning the expected errors for tests that we expect to fail, including both specific message (when given) and target.
This just uses a substring, rather than the whole error message, because it seems like overkill otherwise and it's easier for oneswe expand with arguments.

Adds a new optional test property `expected_error`. This requires one of `input_relationship_type` or `expected_relationship_type` (since the error might vary depending on the relationship selected) and `input_entity_type` (since the same URL might be allowed for one entity type but not another).

This adds two examples:

One for our block of Wikipedia pages as discography entries for releases, testing a specific error message and `input_relationship_type`.

One for blocking non-specific 45cat pages, testing a generic error message (undefined error) and `expected_relationship_type`.

Further tests to cover other existing blocks and errors will be added in subsequent commits.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -44,6 +44,10 @@ const testData = [
              input_entity_type: 'artist',
     expected_relationship_type: 'otherdatabases',
        only_valid_entity_types: [],
+                expected_error: {
+                                  error: undefined,
+                                  target: 'url',
+                                },
   },
   // 45worlds
   {
@@ -4856,6 +4860,10 @@ limited_link_type_combinations: [
     expected_relationship_type: undefined,
        input_relationship_type: 'discographyentry',
        only_valid_entity_types: [],
+                expected_error: {
+                                  error: 'no entries for specific releases',
+                                  target: 'entity',
+                                },
   },
   // Wikisource
   {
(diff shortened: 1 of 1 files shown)


reosarevok:
@yvanzo: can you check whether this seems sensible? If so, I can add further errors to tests.

yvanzo (commented):
It looks good so far :+1: Testing with only one input relationship type seems to be sufficient as the error message for selecting an invalid combination of two or more relationship types is generic.

reosarevok:
@yvanzo: found some issues while trying to expand this (it did not support testing for errors where a custom message isn't given). Added a second example case for those, and allowed using `expected_relationship_type` for cases where we do autoselect but then reject. I'd probably merge this as-is (if it seems good) and then expand its usage further, if that seems sensible?

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -5279,6 +5287,46 @@ testData.forEach(function (subtest, i) {
...
+          validationResult.error.includes(subtest.expected_error.error),
+          'Error message contains expected string',
+        );
+      }
+      st.equal(
+        validationResult.target,
+        subtest.expected_error.target,
+        'Error target matches expected target',
+      );
+    }
```
yvanzo, line 5329: It seems that the flag `tested` should be raised after error has been tested.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -5279,6 +5287,46 @@ testData.forEach(function (subtest, i) {
...
+        st.ok(
+          validationResult.error.includes(subtest.expected_error.error),
+          'Error message contains expected string',
+        );
+      }
+      st.equal(
+        validationResult.target,
+        subtest.expected_error.target,
+        'Error target matches expected target',
+      );
```
yvanzo, line 5328: I suggest to move this part of the code to a separate function to improve readability.
(The same should have been done with validation tests really.)
Can be done after merging this pull request if in a hurry.

reosarevok: I made the change in a second commit, but do let me know if that looks like what you expected.

yvanzo (approved):

PR #2292, opened 2021-09-28
Drop statistic descriptions and improve labels

reosarevok:
We don't actually use the descriptions at all, as far as we have found.
Some labels were a lot worse than the description, and for those cases I switched them.
Some labels were bad but didn't have a better description either, and I reworded them.


yvanzo:
If I understood correctly, this pull request has been opened separately because it was expected to contain just removals and thus to be fast-tracked for merge. It actually changes existing messages as well, requiring usual review. Is it still useful to have it separate from #1851 then? Either this pull request should be closed (as this commit seems to be commented in both pull requests), or the other pull request’s main comment should be updated. Thanks to clarify :sunglasses:

reosarevok:
This was separate because this changes existing ones, and the other adds new ones (and the other one had been open forever on its own, I just rebased it on this). But if we don't care about the distinction enough, we can just use one of the two.

PR #2293, opened 2021-09-29
MBS-11986 / MBS-12128: New account flag: Spammer

reosarevok:
### Implement MBS-11986 / MBS-12128

For now this blocks the user from editing and leaving notes (a combo of editing disabled and notes disabled) and from logging in once logged out.

Future improvements:
* [ ] Automatically log out the editor when marked as spam, so that the login block has a stronger effect. MBS-12129
* [ ] Send user deletion signal to discourse when marked as spam, if desired (but if so, we should figure out a way to restore this if the flag turns out to be added in error). MBS-12130

It was suggested to block profile pages as a 403 but that kinda suggests that there's an issue with the user viewing it - maybe instead detach to a "this user was blocked as spam" template if not account admin? Something like "This user was blocked and their profile is hidden because they were deemed to be spamming. If you see spam in MusicBrainz, please do let us know by reporting the spammer from their user page.".

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Editor.pm
+++ b/lib/MusicBrainz/Server/Entity/Editor.pm
@@ -109,6 +109,10 @@ sub is_adding_notes_disabled {
     (shift->privileges & $ADDING_NOTES_DISABLED_FLAG) > 0;
 }
 
+sub is_spammer {
+    (shift->privileges & $SPAMMER_FLAG) > 0;
+}
+
 sub public_privileges {
     shift->privileges & $PUBLIC_PRIVILEGE_FLAGS;
 }
--- a/root/admin/edit_user.tt
+++ b/root/admin/edit_user.tt
@@ -11,6 +11,7 @@
         [% form_row_checkbox(r, 'banner_editor', l('Banner message editor')) %]
         [% form_row_checkbox(r, 'editing_disabled', l('Editing/voting disabled')) %]
         [% form_row_checkbox(r, 'adding_notes_disabled', l('Edit notes disabled')) %]
+        [% form_row_checkbox(r, 'spammer', l('Spammer')) %]
 
         <h2>[%- l('Technical flags') -%]</h2>
         [% form_row_checkbox(r, 'bot', l('Bot')) %]
--- a/root/admin/edit_user.tt
+++ b/root/admin/edit_user.tt
@@ -9,13 +9,15 @@
         [% form_row_checkbox(r, 'link_editor', l('Relationship editor')) %]
         [% form_row_checkbox(r, 'location_editor', l('Location editor')) %]
         [% form_row_checkbox(r, 'banner_editor', l('Banner message editor')) %]
+
+        <h2>[%- l('User sanctions') -%]</h2>
+        [% form_row_checkbox(r, 'spammer', l('Spammer')) %]
         [% form_row_checkbox(r, 'editing_disabled', l('Editing/voting disabled')) %]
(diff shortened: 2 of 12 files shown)


Freso:
> Send user deletion signal to discourse when marked as spam, if desired (but if so, we should figure out a way to restore this if the flag turns out to be added in error).

As I mentioned during the meeting, this is currently fairly manageable to do manually. I would probably make a new ticket for this instead of holding this one back on that account (once MBS-111987 gets implemented, this would get a bit more unwieldy :)). Also note that it might be possible to suspend the user instead of deleting them. That would still keep them in the forum db, but would prevent them from doing anything meaningful and would be very easy to restore in case of a mistaken flag.

root/admin/edit_user.tt
```diff
@@ -9,13 +9,15 @@
         [% form_row_checkbox(r, 'link_editor', l('Relationship editor')) %]
         [% form_row_checkbox(r, 'location_editor', l('Location editor')) %]
         [% form_row_checkbox(r, 'banner_editor', l('Banner message editor')) %]
+
+        <h2>[%- l('User sanctions') -%]</h2>
         [% form_row_checkbox(r, 'editing_disabled', l('Editing/voting disabled')) %]
         [% form_row_checkbox(r, 'adding_notes_disabled', l('Edit notes disabled')) %]
         [% form_row_checkbox(r, 'spammer', l('Spammer')) %]
```
Freso, line 16: Maybe put Spammer either at top or bottom? (Also, I had been meaning to make a ticket specifically for this, I just hadn’t gotten around to it yet. Ah well. :joy:)

reosarevok: You can still make a ticket and I'll link it from the commit message ;)

Freso:
FWIW, it would be nice to be able to mark accounts as Spam even if it doesn’t do much else than disabling editing and notes. The sooner the flag gets in, the less accounts I’ll have to go back and reflag when it does. :)

root/admin/edit_user.tt
Freso: https://tickets.metabrainz.org/browse/MBS-12128

lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -102,13 +102,18 @@ sub is_editing_disabled {
 }
 
 sub is_editing_enabled {
-    (shift->privileges & $EDITING_DISABLED_FLAG) == 0;
+    my $mask = $EDITING_DISABLED_FLAG ^ $SPAMMER_FLAG;
```
mwiencek, line 105: It's kinda odd how `is_editing_disabled` is now not always equal to the negation of `is_editing_enabled`.  I'd expect that to be true.

Is there any downside to applying the same logic to `is_editing_disabled` too?  Or even `is_adding_notes_disabled` for that matter?

Alternatively, would it make sense to just set `$EDITING_DISABLED_FLAG` and `$ADDING_NOTES_DISABLED_FLAG` automatically whenever the spammer flag is added?

reosarevok: I implemented the latter suggestion as per our IRC chat.

mwiencek (approved):
Meant to approve this after my last review.

yvanzo (approved):
Tested by pretending to address spam on my own local server, not difficult to guess who the spammer was. Nothing critical, better merge it now. Please deploy and quickly try it on beta first.

PR #2295, opened 2021-10-04
Beta: MBS-12000: Fix error msg for incompatible entity

y-young:
# Problem
MBS-12000: "This URL is not allowed for works." when trying to add a Napster (track) URL to a release (single)

Switch case fall-throughs introduced in MBS-11977 (#2280) cause all error message to be 'This URL is not allowed for works.'

# Solution
Add `break`s to all cases.


yvanzo (approved):
Oops, thanks for the fix!

PR #2296, opened 2021-10-04
MBS-9752: Add autoselect + sidebar for Tower Records URLs

y-young:
### Implement MBS-9752
MBS-9752: Add autoselect + sidebar for Tower Records URLs


reosarevok (approved):
They also have https://tower.jp/artist/discography/280635 links that we could clean up to https://tower.jp/artist/280635 as well, I guess. Maybe do that and add a test? This looks good, otherwise :)

PR #2298, opened 2021-10-05
MBS-12007: Allow and autoselect bandcamp crowdfunding

reosarevok:
### Fix MBS-12007

This stops the Bandcamp autoselect and cleanup from breaking their /campaign crowdfunding URLs.
Since these are used to press vinyl, they only make sense for releases. To differentiate them from the usual Bandcamp pages for releases, the sidebar will display "Bandcamp Campaign".


root/static/scripts/edit/URLCleanup.js
```diff
@@ -833,7 +833,11 @@ const CLEANUPS: CleanupEntries = {
     },
   },
   'bandcamp': {
-    match: [new RegExp('^(https?://)?([^/]+)\\.bandcamp\\.com', 'i')],
+    match: [new RegExp(
+      '^(https?://)?([^/]+)\\.bandcamp\\.com' +
+      '(?:[?#]+.*|/(?!campaign)([^?#]*)?.*)?$',
```
mwiencek, line 838: Probably unintended, but this also blocks `https://foo.bandcamp.com/campaignfoo`.

Not sure I understand why the query/hash strings need to be matched there -- can't we just check that `/campaign/` doesn't follow the TLD?

```suggestion
      '^(https?://)?([^/]+)\\.bandcamp\\.com' +
      '(?!/campaign/)',
```

reosarevok: Oh... I guess we could just do that, lol. dunno why I never thought of that and was obsessed with making it work against ?!campaign

PR #2301, opened 2021-10-06
[WIP] MBS-12012: Allow "dupe" URL rels with diff dates

reosarevok:
### Fix MBS-12012 / Implement MBS-12013

2026: Updating this for the new external link code.

# Problem
We have no way to indicate that a link stopped working, then came back. See MBS-13605 for a real life example.

# Solution
This changes link validation to only consider links duplicates if they have the same date period.

It also allows adding (and removing) additional relationships even for autoselected links, as long as the first one has dates or ended set (MBS-12013).

# AI usage
None

# Testing
Manually:
* Add two "official homepage" links, one with dates and one without, and see they can be submitted and edited.
* Check that the "Add another relationship" button appears for autoselected links but only after adding dates to them (MBS-12013)
* Check that validation accepts multiple relationships of the same type for a link, even for autoselected ones, as long as they are not duplicates.
* Check that complete duplicates (same dates) are still detected as such.
* Check that different relationships with different dates but the same type are no longer dropped when merging links.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -8138,11 +8138,23 @@ export class Checker {
       }
       return {result: true};
     }
-    // Multiple types are selected
+    /*
+     * Multiple types are selected
+     * We make sure that the types match one of the sets
+     * of allowed types, allowing for multiple uses of a type
+     * (to support dated relationships)
+     */
+    const selectedTypesSet = new Set(selectedTypes);
     const result = allowedTypes.some(
-      (allowedType) => Array.isArray(allowedType) &&
-        selectedTypes.length === allowedType.length &&
-        (new Set(selectedTypes)).isSubsetOf(new Set(allowedType)),
+      (allowedType) => {
+        const allowedTypeSet = new Set(Array.isArray(allowedType)
+          ? allowedType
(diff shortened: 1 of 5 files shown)


reosarevok:
Cherry-picked 068b6c7a821ecf7432eeb7db546cac69b9dcf771 into its own PR since it's blocking users from editing so we should fix that soonish: https://github.com/metabrainz/musicbrainz-server/pull/2322

yvanzo:
@reosarevok: Can you please rebase it on current `master` to see if CI tests pass again?

reosarevok:
I can, but I doubt they will, as I said above I think that's because of it hitting MBS-12032. You've played with this code more, do you have any suggestions on how to fix that bug?

yvanzo:
I probably have to review #2322 in the first place?

reosarevok:
Well, that's been merged but you can take a look :) The bug existed before that though.

atj:
Please ping me when this ends up on beta ([related irc chat](https://chatlogs.metabrainz.org/libera/metabrainz/2023-05-05/?msg=5169490&page=1)). Thanks!

reosarevok:
Changing the implementation of MBS-12012 so it applies to the new external links code (the other ticket here is now fixed by said code).

PR #2302, opened 2021-10-08
MBS-12017 / MBS-12018: Correctly capitalize YouTube on guess case and set /g

reosarevok:
### Implement MBS-12017 and fix MBS-12018

We do this for stuff like iPad, so we might as well do it for YouTube.

When adding a test I used "youtube" twice, and was confused when it didn't pass; then I noticed that all of these words should be marked `/g` or they will only be corrected the first time they appear on a title, so I amended the rest of the ones that seemed relevant with the second commit here.


yvanzo (approved):
Test’s coverage looks great!

:ship:

PR #2303, opened 2021-10-08
MBS-11903: Restore subscriptions when collection is made public

reosarevok:
### Implement MBS-11903

We have been marking collection subscriptions as not available, rather than removing them, whenever a collection is made private. This takes what seems like the obvious next step and restores them when the collection is made public again.

Schema change but non-replicated. Tested with the test I added, but did not run the schema change otherwise.

On top of https://github.com/metabrainz/musicbrainz-server/pull/1975 to use (extend) the tests already added there.


yvanzo (approved):
* Code looks good to me
* Schema change files look good too
* Added tests cover everything I would test otherwise

@reosarevok : Please update the linked ticket’s summary, type, and description to follow this pull request. (It’s definitely not a bug, maybe an improvement?)

Still requesting another review because it deals with schema change files.

admin/sql/CreateTriggers.sql
```diff
@@ -52,6 +52,9 @@ CREATE TRIGGER del_collection_sub_on_delete BEFORE DELETE ON editor_collection
 CREATE TRIGGER del_collection_sub_on_private BEFORE UPDATE ON editor_collection
     FOR EACH ROW EXECUTE PROCEDURE del_collection_sub_on_private();
 
+CREATE TRIGGER restore_collection_sub_on_public BEFORE UPDATE ON editor_collection
+    FOR EACH ROW EXECUTE PROCEDURE restore_collection_sub_on_public();
```
mwiencek, line 56: I think `AFTER` makes more sense here, since we're propagating the changes to another table and want to be sure we have the final version of the row.
```suggestion
CREATE TRIGGER restore_collection_sub_on_public AFTER UPDATE ON editor_collection
    FOR EACH ROW EXECUTE PROCEDURE restore_collection_sub_on_public();
```

admin/sql/CreateFunctions.sql
```diff
@@ -955,6 +955,21 @@ RETURNS trigger AS $$
...
+  BEGIN
+    IF NEW.public = TRUE AND OLD.public = FALSE THEN
+      UPDATE editor_subscribe_collection sub
+         SET available = TRUE,
+             last_seen_name = NEW.name
+       WHERE sub.collection = OLD.id
+         AND sub.available = FALSE;
+    END IF;
+
+    RETURN NEW;
```
mwiencek, line 969: (If changing to `AFTER`, the return value is ignored, but better to return `NULL` by convention I think.)
```suggestion
    RETURN NULL;
```

mwiencek (approved):
Tested the trigger locally.

reosarevok:

PR #2305, opened 2021-10-12
MBS-12019: Do not list downvoted tags on sidebar when logged out

reosarevok:
### Fix MBS-12019

We already hid these when logged in, but we were not filtering them out at all when the user is logged out and we simply use `TagList`.

Tested by hand with Ed Sheeran, which has some tags with 0 votes such as 'ed sheeran does not make sillyname tracks ಠ_ಠ'' which were still showing on the sidebar.


yvanzo (approved):
:ship:

PR #2306, opened 2021-10-12
MBS-12021: Properly override primary aliases in code

reosarevok:
### Fix MBS-12021

We used to do this with triggers, but that was problematic so we got rid of them. But we forgot to actually change the insert/update code to adapt to that, so they kept hitting ISEs because the old `primary_for_locale` flag was not being unset before they tried to set it for the alias in question.

Tested locally, plus test added.


yvanzo (approved):
:ship:

PR #2307, opened 2021-10-14
Run MergeDuplicateArtistCredits daily at 03:00 UTC

mwiencek:
This currently runs hourly, which is more than needed -- looking at the logs, I can't find any recent cases where we've found multiple duplicates in a single day.

### DIFF
--- a/admin/cron/hourly.sh
+++ b/admin/cron/hourly.sh
@@ -30,6 +30,11 @@ fi
 
 ./admin/CheckElectionVotes.pl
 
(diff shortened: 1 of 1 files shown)


mwiencek:
Yeah, 'cause otherwise I'd have to add a custom entry for it to the crontab and give it its own log file, which is odd because we don't currently have log files for individual scripts.

We do something similar when deciding whether to run a full dump in daily.sh (which happens twice per week).

PR #2309, opened 2021-10-18
Improvements to AcoustIdCell and TaggerIcon

mwiencek:
A while ago I was looking into removing random stuff from common.js which shouldn't be there (in this case, tagger.js). While doing that, I noticed we have to embed an unfortunate amount of JSON for hydrating RecordingList, 99% of which isn't needed, just for the AcoustIdCell. So I improved that situation too.


yvanzo (approved):
Did not test but code and commit messages look good to me. If not adding tests (that were missing already), please consider opening a ticket to recall adding them.

reosarevok:
Added a ticket at MBS-12066 - will merge this and put on beta.

PR #2311, opened 2021-10-21
MBS-12001: Use allowNew for recording links when adding a new track

reosarevok:
### Fix MBS-12001

For some reason, we did this check for track changes (so, when a new recording is assigned to an existing track) but not for new track adds that create a new recording.

### DIFF
--- a/root/edit/details/EditMedium.js
+++ b/root/edit/details/EditMedium.js
@@ -156,7 +156,9 @@ const TracklistChangesAdd = ({
             {' '}
           </>
         ) : null}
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
LGTMBDNT.

The bug report just mentions what the editor observed.
The commit rightfully mentions what is being changed to follow previous changes.
But there is no explanation of what is actually going on with this bug. Is it just a display issue?

reosarevok:
Yeah, this only changes whether the recording is displayed as no longer existing or as being added until the edit applies. The edit itself does the same thing either way.

PR #2312, opened 2021-10-21
MBS-12034: Require confirmation to enter a "Disc n" medium title

reosarevok:
### Fix MBS-12034

Users keep entering CD X or Disc X as medium titles, despite that being generally useless. This tries to avoid that.

Tested by adding a release with the appropriate medium name and seeing that it does get pre-approved when editing, errors when changed, and is still approved when changed back to the original.

I would appreciate two reviews here, just in case :)

### DIFF
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -338,6 +338,23 @@
        %]
     </table>
 
+    <!-- ko if: hasUselessMediumTitle -->
+      <div class="medium-title warning">
+        [%- warning_icon %]
+        <p>
+          <strong>[% l('Warning:') | html_entity %]</strong>
+          <span id="useless-medium-title-warning" data-bind='html: uselessMediumTitleWarning()' />
+        </p>
+        <label>
+          <input id="confirm-medium-title" type="checkbox" data-bind="checked: confirmedMediumTitle" />
+          <span id="confirm-medium-title-message" data-bind="text: confirmMediumTitleMessage()" />
+        </label>
+        <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: hasUnconfirmedUselessMediumTitle">
+          [%~ l('If you’re sure the entered medium title is correct, confirm it above. Otherwise, please fix the data.') %]
+        </p>
+      </div>
+    <!-- /ko -->
+
     <!-- ko if: hasTooEarlyFormat -->
       <div class="warning">
         [%- warning_icon %]
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -7,6 +7,7 @@
  */
 
 import ko from 'knockout';
+import * as ReactDOMServer from 'react-dom/server';
 
 import mbEntity from '../common/entity';
 import {cloneObjectDeep} from '../common/utility/cloneDeep';
@@ -342,6 +343,7 @@ class Medium {
   constructor(data, release) {
     this.release = release;
     this.name = ko.observable(data.name);
+    this.originalName = data.name;
     this.position = ko.observable(data.position || 1);
     this.formatID = ko.observable(data.format_id);
     this.originalFormatID = data.format_id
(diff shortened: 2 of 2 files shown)


root/release/edit/tracklist.tt
```diff
@@ -338,6 +338,23 @@
        %]
     </table>
 
+    <!-- ko if: hasUselessMediumTitle -->
```
yvanzo, line 341: IIRC @mwiencek is currently working on removing knockout dependency, please make sure there is no conflict or do arrange to prioritize one change over the other.

root/release/edit/tracklist.tt
```diff
@@ -338,6 +338,23 @@
...
+        <p>
+          <strong>[% l('Warning:') | html_entity %]</strong>
+          [% l('“CD X” / “Disc X” should only be used as a medium title if it’s the actual title for the medium (not just to indicate this is the disc number X in a set).') %]
+        </p>
+        <label>
+          <input id="confirm-medium-title" type="checkbox" data-bind="checked: confirmedMediumTitle" />
+          <span id="confirm-medium-title-message" data-bind="text: confirmMediumTitleMessage()" />
+        </label>
+        <p class="error field-error" style="padding-left: 1em" data-bind="showErrorWhenTabIsSwitched: hasUnconfirmedUselessMediumTitle">
+          [%~ l('If you’re sure the entered medium title is correct, confirm it above. Otherwise, please fix the data.') %]
```
yvanzo, line 353: As an editor, from this warning I would not understand when to do it or not do it. A more detailed documentation with examples would be helpful. I asked for real examples in comment to the ticket.

root/static/scripts/release-editor/fields.js
```diff
@@ -445,6 +446,10 @@ class Medium {
                 self.release.earliestYear() < mediumFormatDate);
     });
     this.confirmedEarlyFormat = ko.observable(this.hasTooEarlyFormat());
+    this.hasUselessMediumTitle = ko.computed(function () {
+      return /^(CD|Disc)\s*\d+/i.test(self.name());
```
yvanzo, line 450: Maybe match more common medium formats if that is likely to happen too?

```suggestion
      return /^(Cassette|CD|Disc|DVD|Vinyl)\s*\d+/i.test(self.name());
```

yvanzo (commented):
Examples needed.

root/static/scripts/release-editor/fields.js
reosarevok: I've seen it much more often with CD, but I guess it can't hurt to extend it a bit, DVD especially of this sounds likely. :)

root/release/edit/tracklist.tt
reosarevok: I expect it's easy to change this one when changing all the others, but I'd also be happy to change this if that code got merged earlier. Whatever I can do to get rid of knockout :D

root/release/edit/tracklist.tt
yvanzo: The place for the detailed documentation is [https://musicbrainz.org/doc/Style/Release#Medium_title]. Maybe link to it?

Also, if not entering “CD X”, editor may think that something is amiss. Maybe explain that numbering should just be set by moving mediums using up/down arrows?

reosarevok: So, something like this instead of the first section?

> “CD X” / “Disc X” should only be used as a medium title if it’s the actual title for the medium (see {release_style|the guidelines}). To indicate ordering, simply reorder the medium using the up/down arrows.

yvanzo: Trying something:
* Exposing the issue first, then possible courses of action;
* Displaying matched part of the title can be displayed instead of “CD X / Disc X”.

> “{matched_text}” seems to indicate medium ordering rather than medium title. If that is the case, please use up/down arrows on the right side to set medium position instead of title (see {release_style|the guidelines}). Otherwise please confirm that is not the case using the checkbox below.

mwiencek: that sounds good to me but I'd tweak the grammar/wording slightly
> “{matched_text}” seems to indicate medium ordering rather than a medium title. If this is the case, please use the up/down arrows on the right side to set the medium position instead of adding a title (see {release_style|the guidelines}). Otherwise, please confirm that this is the actual title using the checkbox below.

reosarevok: IIRC the reason I didn't do `{matched_text}` originally was to avoid having to create this in another page as with `confirmMediumTitleMessage()` (@mwiencek suggested it wasn't needed in both places). But I think with this wording it does make sense to have it, so I'll look into it.

reosarevok: I think it works fine, so updating with it:

Screenshot from 2021-11-17 20-51-05

yvanzo (approved):

mwiencek (approved):

PR #2314, opened 2021-10-25
MBS-9356: Add script to remove completely empty editors

reosarevok:
### Implement MBS-9356

This allows removing editors who haven't logged in for over a year, have no verified email, and have never used
pretty much anything in the DB. There's around 700k of those, so a very significant chunk of our users, and many might be dormant spammers.

### DIFF
--- a/admin/RemoveEmptyAccounts.pl
+++ b/admin/RemoveEmptyAccounts.pl
@@ -1,0 +1,199 @@
+#!/usr/bin/env perl
+
+use warnings;
+use strict;
+
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+use open ':std', ':encoding(UTF-8)';
+
+use Getopt::Long;
+use MusicBrainz::Server::Context;
+use MusicBrainz::Server::Constants qw( entities_with );
+
+my $dry_run = 0;
+my $limit = 500;
+
+GetOptions(
+    'dry-run|d'    => \$dry_run,
+) or usage();
+
+sub usage {
+    warn <<EOF;
+Usage: $0 [options]
+
+OPTIONS
+    -d, --dry-run       Perform a trial run without removing any account
+
+EOF
+    exit(2);
+};
+
+my $c = MusicBrainz::Server::Context->create_script_context(
+    database => 'MAINTENANCE',
+);
+my $c_cursor = MusicBrainz::Server::Context->create_script_context(
+    database => 'MAINTENANCE',
+    fresh_connector => 1,
+);
+
+my $editor_collection_subquery = join(
+    ' AND ',
+    map {<<~"SQL"} entities_with('collections'),
+        NOT EXISTS (
+            SELECT TRUE
+              FROM editor_collection ec
+              JOIN editor_collection_$_ ece ON ece.collection = ec.id
+             WHERE ec.editor = e.id
+             LIMIT 1
+        )
+        SQL
+);
+
+my $editor_rating_subquery = join(
+    ' AND ',
+    map {<<~"SQL"} entities_with('ratings'),
+        NOT EXISTS (
+            SELECT TRUE
+              FROM ${_}_rating_raw err
+             WHERE err.editor = e.id
+             LIMIT 1
+        )
+        SQL
+);
+
+my $editor_subscription_subquery = join(
+    ' AND ',
+    map {<<~"SQL"} entities_with('subscriptions'),
+        NOT EXISTS (
+            SELECT TRUE
+              FROM editor_subscribe_$_ ese
+             WHERE ese.editor = e.id
+             LIMIT 1
+        )
+        SQL
+);
+
+my $editor_subscription_deleted_subquery = join(
+    ' AND ',
+    map {<<~"SQL"} entities_with(['subscriptions', 'deleted']),
+        NOT EXISTS (
+            SELECT TRUE
+              FROM editor_subscribe_${_}_deleted esed
+             WHERE esed.editor = e.id
+             LIMIT 1
+        )
+        SQL
+);
+
+my $editor_tag_subquery = join(
+    ' AND ',
+    map {<<~"SQL"} entities_with('tags'),
+        NOT EXISTS (
+            SELECT TRUE
+              FROM ${_}_tag_raw etr
+             WHERE etr.editor = e.id
+             LIMIT 1
+        )
+        SQL
+);
+
+my $query = <<~"SQL";
+          SELECT e.id,
+                 e.name
+            FROM editor e
+           WHERE e.email IS NULL
+             AND deleted IS false
+             AND privs = 8192
+             AND (   member_since < NOW() - INTERVAL '12 months'
+                  OR member_since IS NULL)
+             AND (   last_login_date IS NULL
+                  OR last_login_date < NOW() - INTERVAL '12 months')
+  AND NOT EXISTS (SELECT 1
+                    FROM application
+                   WHERE application.owner = e.id)
+  AND NOT EXISTS (SELECT 1
+                    FROM editor_oauth_token
+                   WHERE editor_oauth_token.editor = e.id)
+  AND NOT EXISTS (SELECT 1
+                    FROM vote
+                   WHERE vote.editor = e.id)
+  AND NOT EXISTS (SELECT 1
+                    FROM edit
+                   WHERE edit.editor = e.id)
+  AND NOT EXISTS (SELECT 1
+                    FROM edit_note
+                   WHERE edit_note.editor = e.id)
+  AND NOT EXISTS (SELECT 1
+                    FROM annotation
+                   WHERE annotation.editor = e.id)
+  AND $editor_subscription_subquery
+  AND $editor_subscription_deleted_subquery
+  AND $editor_tag_subquery
+  AND $editor_rating_subquery
+  AND $editor_collection_subquery
+  AND NOT EXISTS (   SELECT 1
+                       FROM editor_collection ec
+                       JOIN editor_collection_collaborator ecc
+                         ON ecc.collection = ec.id
+                      WHERE ec.editor = e.id)
+  AND NOT EXISTS (   SELECT 1
+                       FROM editor_collection_collaborator ecc
+                      WHERE ecc.editor = e.id)
+  SQL
+
+$c_cursor->sql->begin;
+$c_cursor->sql->do("DECLARE cursor NO SCROLL CURSOR FOR $query");
+
+my $pull = sub {
+    $c_cursor->sql->select("FETCH FORWARD $limit FROM cursor");
+};
+while ($pull->()) {
+    $c->sql->begin;
+    my @row;
+    while (@row = $c_cursor->sql->next_row) {
+        my ($editor_id, $editor_name) = @row;
+
+        if ($dry_run) {
+            print 'Removing account ' . $editor_name . " (dry run)\n";
+        } else {
+            print 'Removing account ' . $editor_name . "\n";
+            # Remove preferences
+            $c->sql->do(
+                'DELETE FROM editor_preference WHERE editor = ?',
+                $editor_id
+            );
+            # Remove languages
+            $c->sql->do(
+                'DELETE FROM editor_language WHERE editor = ?',
+                $editor_id
+            );
+            # Remove any subscriptions *to* this editor
+            $c->sql->do(<<~'SQL', $editor_id);
+                DELETE FROM editor_subscribe_editor
+                      WHERE subscribed_editor = ?
+                SQL
+            # Remove any collections by the editor (they should be all empty)
+            $c->sql->do(
+                'DELETE FROM editor_collection WHERE editor = ?',
+                $editor_id
+            );
+            # Actually delete the editor, which should not trigger any FKs now
+            $c->sql->do('DELETE FROM editor WHERE id = ?', $editor_id);
+        }
+    }
+    $c->sql->commit;
+}
+
+$c_cursor->sql->do('CLOSE cursor');
+$c_cursor->sql->rollback;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2021 MetaBrainz Foundation
+
+This file is part of MusicBrainz, the open internet music database,
+and is licensed under the GPL version 2, or (at your option) any
+later version: http://www.gnu.org/licenses/gpl-2.0.txt
+
+=cut
--- a/admin/cron/monthly.sh
+++ b/admin/cron/monthly.sh
@@ -11,5 +11,6 @@ MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../../" && pwd)
 cd "$MB_SERVER_ROOT"
 
 ./admin/RunSampleDataDump
+./admin/RemoveEmptyAccounts.pl
 
 echo Monthly jobs complete!


mayhem (approved):
Sorry for the delay, but this query looks good to me. I wonder if it would make sense to do a dry run and post a sample of the accounts that would get deleted and then let people have a sanity check?

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,189 @@
...
+    fresh_connector => 1,
+);
+
+my $query = <<~'SQL';
+          SELECT e.id,
+                 e.name
+            FROM editor e
+           WHERE e.email IS NULL
+             AND deleted IS false
+             AND (   member_since < NOW() - INTERVAL '2 weeks'
```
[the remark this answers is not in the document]
reosarevok: I changed both to 12 months, which it seems like a sensible amount.

reosarevok:
@mayhem: A small sample: https://gist.github.com/reosarevok/7aa23ce7a16ea74f2b1b06e4939200b3 - the whole dry run returns 785181 accounts.

admin/RemoveEmptyAccounts.pl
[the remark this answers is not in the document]
reosarevok: Now I even remembered to push the change :sweat_smile:

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,189 @@
...
+};
+while ($pull->()) {
+    $c->sql->begin;
+    my @row;
+    while (@row = $c_cursor->sql->next_row) {
+      if ($dry_run) {
+          print 'Removing account ' . $row[1] . " (dry run)\n";
+      } else {
+          print 'Removing account ' . $row[1] . "\n";
+          $c->model('Editor')->delete($row[0]);
```
mwiencek, line 172: Given that these accounts were never actually used, should we change this to allow name reuse by default?

reosarevok: That actually seems sensible to me. @Freso, what do you think?

Freso: Yeah, I’m good with that. 👍

Freso:
Once we have the Spammer account flag, would it make sense to exclude accounts with that from this script too? The reason for the flag (over outright deleting them to begin with) is to be able to (eventually…) feed them to SpamBrainz for training, so would we want to keep them around until then?

reosarevok:
In that case, would it make sense to just keep anyone who doesn't have privs = 0? If any flag has been set, that probably means the user has been interacted with in a way that might mean we want to keep it around for manual check anyway.

reosarevok:
Made the change. It doesn't make a significant difference in the number of accounts removed (would be 785838 now - so the number has grown with the passed time, even discounting the ones with privs set).

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,191 @@
...
+while ($pull->()) {
+    $c->sql->begin;
+    my @row;
+    while (@row = $c_cursor->sql->next_row) {
+      if ($dry_run) {
+          print 'Removing account ' . $row[1] . " (dry run)\n";
+      } else {
+          print 'Removing account ' . $row[1] . "\n";
+          # We allow reuse, since these accounts are all unused
+          $c->model('Editor')->delete($row[0], 1);
```
mwiencek, line 174: Ignoring the actual editor row, is there anything else that'd need deleting besides the defaults collections?  Maybe editor preferences/languages too?  (I'd guess it's not a lot in any case.)

After looking at this again, I think I'd prefer avoiding `$c->model('Editor')->delete` for the prod run and being explicit about which rows we're deleting.  Although the query above looks good, the `delete` method is still dangerous in that it'll delete the editor even if it's not empty.  If there's something we missed, or if the query malfunctions somehow, I'd rather have it fail the deletion (due to existing FKs) than purge everything without question.

reosarevok: So, just do all in SQL? I guess that could work, but we should make it so that it does not crash the script if the delete fails. Is there a good way to tell it "on failure, store the username to a file and continue"?

mwiencek: IIUC the script shouldn't fail unless there's a bug in the query, or an FK we forgot to delete (but in either case we should fix the script).  The other case would be a transient error (like a timeout), but in that case we can just re-run it, right?

reosarevok: Oh, ok, so we'd just let it fail, then fix the script. I was thinking let it delete all it can with no FK clashes, then let's fix, but I guess that'd work too.

reosarevok: I think I updated all that was needed, but do make sure I didn't do something stupid, please. I also added a check for no collaborators on collections to the search.

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,233 @@
...
+            print 'Removing account ' . $editor_name . " (dry run)\n";
+        } else {
+            print 'Removing account ' . $editor_name . "\n";
+            # Remove preferences
+            $self->sql->do(
+                'DELETE FROM editor_preference WHERE editor = ?',
+                $editor_id
+            );
+            # Remove languages
+            $self->sql->do(
```
mwiencek, line 184: All occurrences of `$self->sql` should be `$c->sql`

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,233 @@
...
+                                  (type != 5 OR name != 'Attending')
+                                  AND
+                                  (type != 6 OR name != 'Maybe attending')
+                            )
+                            OR EXISTS (SELECT 1
+                                         FROM editor_collection_event ece
+                                        WHERE ece.collection = ec.id)))
+                  OR EXISTS (SELECT 1
+                               FROM editor_collection_collaborator ecc
+                              WHERE ecc.collection = ec.id)))
```
mwiencek, line 157: There seems to be an extra `)` (or two?) causing a syntax error here.

mwiencek (changes requested):

mwiencek (approved):

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,238 @@
...
+                       FROM editor_collection ec
+                      WHERE ec.editor = e.id
+                        AND ((
+                                  (type != 5 OR name != 'Attending')
+                                  AND
+                                  (type != 6 OR name != 'Maybe attending')
+                            )
+                            OR EXISTS (SELECT 1
+                                         FROM editor_collection_event ece
+                                        WHERE ece.collection = ec.id)))
```
mwiencek, line 154: I'm finding this hard to understand, but also wondering: was there a reason we don't just check if all existing collections are empty? Looks like it's trying to detect the presence of default collections, but I don't remember why that was necessary.

admin/RemoveEmptyAccounts.pl
```diff
@@ -0,0 +1,238 @@
...
+            $c->sql->do(
+                'DELETE FROM editor_language WHERE editor = ?',
+                $editor_id
+            );
+            # Remove any subscriptions *to* this editor
+            $c->sql->do(<<~'SQL', $editor_id);
+                DELETE FROM editor_subscribe_editor
+                      WHERE subscribed_editor = ?
+                SQL
+            # Remove any collections by the editor (they should be all empty)
```
mwiencek, line 198: I can't actually find where we ensure that all the collections are empty, other than for `editor_collection_event`.

admin/RemoveEmptyAccounts.pl
reosarevok: IIRC the idea was that we wanted to avoid autoremoving editors who had created a collection themselves, even if empty. Although I'd agree it's fine to remove them if they added an empty collection over a year ago tbh.

reosarevok:
Made some updates here, including constructing the query using `entities_with` where possible to make it more future-proof.

This query currently finds 834933 editors.

The full constructed query with the current `%ENTITIES` data:

```
SELECT e.id,
               e.name
          FROM editor e
         WHERE e.email IS NULL
           AND deleted IS false
           AND privs = 0
           AND (   member_since < NOW() - INTERVAL '12 months'
                OR member_since IS NULL)
           AND (   last_login_date IS NULL
                OR last_login_date < NOW() - INTERVAL '12 months')
AND NOT EXISTS (SELECT 1
                  FROM application
                 WHERE application.owner = e.id)
AND NOT EXISTS (SELECT 1
                  FROM editor_oauth_token
                 WHERE editor_oauth_token.editor = e.id)
AND NOT EXISTS (SELECT 1
                  FROM vote
                 WHERE vote.editor = e.id)
AND NOT EXISTS (SELECT 1
                  FROM edit
                 WHERE edit.editor = e.id)
AND NOT EXISTS (SELECT 1
                  FROM edit_note
                 WHERE edit_note.editor = e.id)
AND NOT EXISTS (SELECT 1
                  FROM annotation
                 WHERE annotation.editor = e.id)
AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_series ese
     WHERE ese.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_collection ese
     WHERE ese.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_label ese
     WHERE ese.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_editor ese
     WHERE ese.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_artist ese
     WHERE ese.editor = e.id
     LIMIT 1
)

AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_series_deleted esed
     WHERE esed.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_label_deleted esed
     WHERE esed.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_subscribe_artist_deleted esed
     WHERE esed.editor = e.id
     LIMIT 1
)

AND NOT EXISTS (
    SELECT TRUE
      FROM series_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM place_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM label_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM release_group_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM instrument_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM recording_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM artist_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM area_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM release_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM event_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM work_tag_raw etr
     WHERE etr.editor = e.id
     LIMIT 1
)

AND NOT EXISTS (
    SELECT TRUE
      FROM place_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM label_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM release_group_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM recording_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM artist_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM event_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM work_rating_raw err
     WHERE err.editor = e.id
     LIMIT 1
)

AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_genre ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_series ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_place ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_label ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_release_group ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_instrument ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_recording ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_artist ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_area ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_release ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_event ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)
 AND NOT EXISTS (
    SELECT TRUE
      FROM editor_collection ec
      JOIN editor_collection_work ece ON ece.collection = ec.id
     WHERE ec.editor = e.id
     LIMIT 1
)

AND NOT EXISTS (   SELECT 1
                     FROM editor_collection ec
                     JOIN editor_collection_collaborator ecc
                       ON ecc.collection = ec.id
                    WHERE ec.editor = e.id)
AND NOT EXISTS (   SELECT 1
                     FROM editor_collection_collaborator ecc
                    WHERE ecc.editor = e.id);
```

PR #2316, opened 2021-10-27
MBS-12056: Convert the edit data page to React

reosarevok:
### Implement MBS-12056

This uses `formatPluralEntityTypeName` as-is rather than adding a separate function to get lowercase names or
running a lowercasing function on the result of the format function because I can't see any good reason why the names need to be lowercase here.

Also, I'm not actively sorting the entity IDs on `relatedEntities` (TT used `nsort`). We could do a number sort here,
but they seem to come in the right order already and additionally I'm not sure it's even particularly important/useful.

Tested by visiting edit/83937903/data (small change) and edit/81690949/data (tons of linked entities).

Not sure if there's a much simpler option than my `to_json_hash` function that I'm missing, if so do let me know.


yvanzo (approved):
Successfully tested with some edits and looks good overall; one more TT file out .:)

Using `formatPluralEntityTypeName` seems to be perfectly fine for the of related entities at the bottom of the page, no need to lowercase here indeed.

As for `nsort`, it seems easier to find an ID in a sorted list. It’s been initially implemented in commit <ea8a413f12e6c77ba103ec8f26d4cd3f8fc1fa28> but the message doesn’t explain in which cases `nsort` needs to be used. Would there be that IDs are always ordered already?

lib/MusicBrainz/Server/Entity/Util/JSON.pm
```diff
@@ -23,6 +24,17 @@ sub to_json_array {
     return undef;
 }
 
+sub to_json_hash {
```
mwiencek, line 27: I was worried the name might be a bit too similar to `to_json_object` to distinguish them, but I can't think of anything better.  (If we made it more specific, I'd also want to change `to_json_array` for symmetry.)

root/edit/EditData.js
```diff
@@ -0,0 +1,105 @@
...
+
+const EditIndex = ({
+  edit,
+  relatedEntities,
+}: Props): React.Element<typeof Layout> => {
+  const title = texp.l('Edit data for edit #{id}', {id: edit.id});
+  const jsonData = JSON.parse(edit.raw_data);
+  const displayData = jsonData
+    ? JSON.stringify(jsonData, null, 2)
+    : edit.raw_data;
```
mwiencek, line 36: Not sure I understand this part.  How come we need to send `raw_data` to do `JSON.parse(edit.raw_data)` instead of just using `edit.data`?

In the next line, how come a ternary is needed to check `jsonData`?

root/edit/EditData.js
```diff
@@ -0,0 +1,105 @@
...
+      </p>
+
+      <p><strong>{l('Data:')}</strong></p>
+      <pre id="edit-raw-data">{displayData}</pre>
+
+      {Object.keys(relatedEntities).length ? (
+        <>
+          <p><strong>{l('Related entities:')}</strong></p>
+          <ul>
+            {Object.keys(relatedEntities).sort().map((type, index) => {
```
mwiencek, line 65: Since we have to use the keys twice and sort them, it seems worth putting them in a variable first:
```js
const relatedEntityTypes = Object.keys(relatedEntities).sort();
```

```suggestion
      {relatedEntityTypes.length ? (
        <>
          <p><strong>{l('Related entities:')}</strong></p>
          <ul>
            {relatedEntityTypes.map((type, index) => {
```

reosarevok:
> Would there be that IDs are always ordered already?

It kinda felt that way to me, but I didn't look into it too much, just saw the few examples I tried did come ordered.

root/edit/EditData.js
```diff
@@ -0,0 +1,105 @@
...
+import {bracketedText} from '../static/scripts/common/utility/bracketed';
+import {formatPluralEntityTypeName}
+  from '../static/scripts/common/utility/formatEntityTypeName';
+import {getEditStatusName} from '../utility/edit';
+
+type Props = {
+  +edit: {...EditT, +id: number},
+  +relatedEntities: {[type: CoreEntityTypeT]: {
+    [entityId: string]: CoreEntityT,
+  }},
```
mwiencek, line 25:
```suggestion
  +edit: $ReadOnly<{...EditT, +id: number}>,
  +relatedEntities: {
    +[type: CoreEntityTypeT]: {
      +[entityId: string]: CoreEntityT,
    },
  },
```

mwiencek, line 25: Didn't know eslint accepted stuff like

```js
{foo: {
  1}}
```
but it looks kinda messy to me. I added some newlines, `$ReadOnly`, and some `+` sigils.

```suggestion
  +edit: $ReadOnly<{...EditT, +id: number}>,
  +relatedEntities: {
    +[type: CoreEntityTypeT]: {
      +[entityId: string]: CoreEntityT,
    },
  },
```

mwiencek (changes requested):

root/edit/EditData.js
mwiencek: Or actually, isn't that `GenericEditWithIdT` now?

root/edit/EditData.js
reosarevok: I was just copying what `root/edit/data.tt` used, since I assumed it was useful, but I guess if edit.data is already JSON, that might be easier :)

root/edit/EditData.js
reosarevok: Yes, it is, I was making the change already but thanks for the reminder in case :)

mwiencek (approved):

PR #2318, opened 2021-11-02
MBS-12050: Add "edit note does not include" to edit search

reosarevok:
### Implement MBS-12050

Tested by filtering the edits by a specific user with "includes" / "does not include": "imported".


yvanzo (approved):
LGTMBDNT.

:ship:

PR #2319, opened 2021-11-02
MBS-12057: Show entity type on sidebar "No collections!" message

reosarevok:
### Implement MBS-12057

This has been confusing to users who do not realize they're on an RG page instead of a release at first
or whatnot. It doesn't seem to have any downside to specify, so we might as well.


yvanzo (approved):
LGTMBDNT

reosarevok:

PR #2323, opened 2021-11-03
MBS-12060: Block google.com/search URLs on URL relationships

reosarevok:
### Implement MBS-12060

We currently have 32 of these for google.com, but I'm pretty sure that's because a lot of them get removed soon after they're added - I've certainly seen this happen.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -603,6 +603,13 @@ export class ExternalLinksEditor
                     Enter the destination URL instead.`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
+    } else if (isNewOrChangedLink && isGoogleSearch(link.url)) {
+      error = {
+        message: l(`Please don’t enter links to search results.
+                    If you’ve found any links through your search
+                    that seem useful, do enter those instead.`),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
     } else if (!link.type) {
       error = {
         message: l(`Please select a link type for the URL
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/externalLinks.js
```diff
@@ -1603,6 +1610,10 @@ function isGoogleAmp(url) {
   return /^https?:\/\/([^/]+\.)?google\.[^/]+\/amp/.test(url);
 }
 
+function isGoogleSearch(url) {
+  return /^https?:\/\/([^/]+\.)?google\.[^/]+\/search/.test(url);
```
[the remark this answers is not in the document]
reosarevok: This wouldn't match `https://link.example?ref=http://google.com/search?q=foo` as it is (notice it expects a dot at the end of whatever is before Google) but it would match something like `https://link.example?ref=www.google.com/search?q=foo`. While that doesn't seem too likely to happen, we might as well use your version to be safe :)

root/static/scripts/edit/externalLinks.js
```diff
@@ -603,6 +603,13 @@ export class ExternalLinksEditor
                     Enter the destination URL instead.`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
+    } else if (isNewOrChangedLink && isGoogleSearch(link.url)) {
+      error = {
+        message: l(`Please don’t enter links to Google searches.
+                    If you’ve found any links through your search
+                    that seem useful, do enter those instead.`),
```
yvanzo, line 610: Found at least one Bing search using http://musicbrainz.org/ws/2/url?query=url:*bing*search*&fmt=json

A more generic message would allow blocking pages from other popular search engines too:

```suggestion
        message: l(`Please don’t enter links to search results.
                    If you’ve found any links through your search
                    that seem useful, do enter those instead.`),
```

reosarevok: Good point! I won't implement Bing blocking yet, but I will make the suggested change so the string doesn't have to change later :)

reosarevok:

PR #2324, opened 2021-11-03
MBS-12064: Add jazzmusicarchives.com to the database whitelist

reosarevok:
### Implement MBS-12064


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2337,6 +2337,34 @@ const CLEANUPS: CleanupEntries = {
...
+              result: type === 'artist',
+              target: ERROR_TARGETS.ENTITY,
+            };
+          case LINK_TYPES.otherdatabases.release_group:
+            return {
+              result: type === 'album',
+              target: ERROR_TARGETS.ENTITY,
+            };
+        }
+        return {result: false, target: ERROR_TARGETS.ENTITY};
```
yvanzo, line 2363: The Metal Archives are allowed to link to labels and releases. Should Jazz Music Archives be the same?

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2240,6 +2240,21 @@ const testData = [
     expected_relationship_type: ['downloadfree', 'streamingfree'],
             expected_clean_url: 'http://www.jamendo.com/album/56372',
   },
+  // Prog Archives
```
yvanzo, line 2243: Typo from copy/paste:
```suggestion
  // Jazz Music Archives
```

yvanzo (commented):
Details about which entity types may be linked (or not linked if something has to be specifically avoided) and example links would be welcome in the ticket’s description, for editors not to dig the code. Not sure about labels and releases. The rest looks good!

root/static/scripts/edit/URLCleanup.js
reosarevok: Jazz Music Archives does not seem to have label pages, so those are out. 

http://www.jazzmusicarchives.com/album/john-coltrane/my-favorite-things (for example) does say "Atlantic 1361(US)" which makes it seem like these map to releases, but http://www.jazzmusicarchives.com/album/miles-davis/kind-of-blue for example says "Sony CD Reissue in 1997 includes 1 bonus track" and nothing else - they don't seem to have an entry for said reissues. So I think RG is probably the most reasonable choice to map to.

yvanzo: Ok thanks, just asking because it is the same network of websites.

reosarevok: Actually, I think the original user was a bit confused - the site in the same network is http://www.metalmusicarchives.com/ as they linked, but we do not have that - we have https://www.metal-archives.com/ which is a completely different one :) People are just not very original.

yvanzo (approved):
:ship: 

It’s part of the STYLE workflow in Jira, but it’s not mentioned in the pull request, so just a friendly reminder: The documentation page about other databases has to be updated and will have to be transcluded at release time.

reosarevok:
Updated the page, now I need to remember to transclude.

PR #2326, opened 2021-11-04
MBS-12047: Support browsing artist RGs in /ws/2 as in "official" website index 

reosarevok:
### Implement MBS-12047

This is useful to be able to easily get through the API the same list of release groups we show by default on the MusicBrainz site itself.

I chose to use a new 'official-only' parameter rather than the existing 'status' parameter already in use for releases because 'status' can be used to filter the releases via inc=, and "official" for RGs is a wider concept than for releases (including also for example RGs with only no-status-set releases).

Split tests into a separate commit for clarity - I thought the easiest way to change this would be to move an existing release to bootleg, which also makes it so we have bootlegs in more places on tests (we had no Bootleg in the WS tests at all). We can later squash them if we want.


reosarevok:

phw:
What's the status here? This would be really useful, I just had wished for that once again. And AFAIK could also help for CritiqueBrainz (which currently shows e.g. bootlegs in the album list)

reosarevok:
Waiting for reviews, hopefully after the schema change then :)

nsgeek:
Thanks for implementing this, @reosarevok! I've also been waiting eagerly for this to be approved as well (I'm the user discocat on MB Jira mentioned in [MBS-12047](https://tickets.metabrainz.org/browse/MBS-12047)). It looks pretty good to me but I really hope someone with more codebase knowledge than me (a low bar) steps up and reviews it soon. :)

reosarevok:
Hmm. I could see that, but then we should still have something like `website-default` as an option because a fair amount of the people asking seem to want to have the same as MB specifically.

mwiencek:
> Hmm. I could see that, but then we should still have something like website-default as an option because a fair amount of the people asking seem to want to have the same as MB specifically.

Sounds fine to me if people want that. We could also support an option like 'unset' or 'null' so you could do official+unset, but I guess people specifically want whatever's the default on the website, which could change?

mwiencek:
That's true, though we already perform other types of joins in `_where_filter`, so I think it should be fine to join with release to get the status.  But, I didn't want to delay this any further either, so what about just adding support for website-default for now (giving us room to expand it in the future)?

mwiencek (approved):
Tested locally, seems to work well.

PR #2327, opened 2021-11-04
MBS-12055: Allow admins to see private editor-editor subscriptions

reosarevok:
### Implement MBS-12055

@Freso asked for an easier way to see if a sockpuppet editor subscribes to other editors (in order to find more sockpuppets).

They specifically asked for a way to know the tab contains private data and should not be open when streaming so I added a class (to be able to hide it) and a warning / highlight.

Screenshot from 2021-11-04 14-07-07

They also specified they did not want to be able to see any other subscriptions, so this hides even the summary
from the user subscription page in this case.

Also, since the /artists page is private, we need to specifically make the tab lead to /editors in this case.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User/Subscriptions.pm
+++ b/lib/MusicBrainz/Server/Controller/User/Subscriptions.pm
@@ -27,7 +27,22 @@ with 'MusicBrainz::Server::Controller::User::SubscriptionsRole' => {
 sub subscriptions : Chained('/user/load') {
     my ($self, $c) = @_;
     my $user = $c->stash->{user};
-    $c->response->redirect($c->uri_for_action('/user/subscriptions/artist', [ $user->name ]));
+    $c->model('Editor')->load_preferences($user);
+
+    my $is_admin_viewing_private = defined $c->user &&
+                                   $c->user->is_account_admin &&
+                                   $c->user->id != $user->id &&
+                                   !$user->preferences->public_subscriptions;
+
+    if ($is_admin_viewing_private) {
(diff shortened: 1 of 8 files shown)


yvanzo (commented):
LGTMBDNT. Has it been tested? CI tests would be really appreciated since it deals with private data.

reosarevok:
I tested it manually, but adding a test does seem smart. Will look into it.

reosarevok:
Added a test now (and found a bug where you'd get an ISE when logged out, now fixed).

lib/MusicBrainz/Server/Controller/User/Subscriptions.pm
```diff
@@ -27,7 +27,21 @@ with 'MusicBrainz::Server::Controller::User::SubscriptionsRole' => {
 sub subscriptions : Chained('/user/load') {
     my ($self, $c) = @_;
     my $user = $c->stash->{user};
-    $c->response->redirect($c->uri_for_action('/user/subscriptions/artist', [ $user->name ]));
+    $c->model('Editor')->load_preferences($user);
+
+    my $is_admin_viewing_private = $c->user && $c->user->is_account_admin &&
```
yvanzo, line 32: Not sure if `defined` is not missing here?

```suggestion
    my $is_admin_viewing_private = defined $c->user && $c->user->is_account_admin &&
```

yvanzo (approved):
That’s nice when writing tests immediately pay off, even though it’s mostly for the long term.

mwiencek (approved):

PR #2331, opened 2021-11-12
MBS-12079: JSON-LD for releases with a lot of [silence] is too big

mwiencek:
The most-referenced [silence] recording has over 2,000 gid redirects.  These are all outputted in the `sameAs` property in its JSON-LD, including on release pages that reference it.  If the release contains [silence] multiple times, as https://musicbrainz.org/release/86427535-9abb-4681-8c88-5c651e55d62a does for example, then this is multiplied.  (The linked release references [silence] more than 50 times, and contains a 7.75 MB chunk of JSON-LD.)

It also so happens that Google reports "Structured data with syntax errors" on this page, with "Incorrect value type" as the reason (but the part of the source it highlights as having an error makes no sense).  So we're wondering if this is just due to the size of the page hitting some internal limits that Google has.

This patch limits the number of gid redirects outputted for recordings' `sameAs` properties on release pages to 100 each. It also skips outputting them entirely if the recording appears multiple times (only outputting them for the first occurrence).

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Release.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD/Release.pm
@@ -52,12 +52,30 @@ around serialize => sub {
 
         if ($entity->all_mediums) {
             my @tracks;
+            my %seen_recordings;
             for my $medium ($entity->all_mediums) {
                 if ($medium->all_tracks) {
                     for my $track ($medium->all_tracks) {
-                        if ($track->recording) {
-                            $stash->store($track->recording)->{trackNumber} = join(q(.), $medium->position, $track->position);
-                            push(@tracks, serialize_entity($track->recording, $inc, $stash));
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
Seems sensible. I guess a test for this would look kinda absurdly big, so I'm happy just testing it with this release locally, where it looked fine.

yvanzo (approved):
Perl code looks good. Generated JSON-LD code looks good too. Cannot check if that will actually resolve the Google report though.

Would there be a way (a boolean flag or something) to indicate that the list of redirects has been truncated? (Can be a further improvement.)

mwiencek:
> Would there be a way (a boolean flag or something) to indicate that the list of redirects has been truncated?

I'm not sure we can add such a flag in the JSON-LD itself since it's validated against a schema, but perhaps it can be indicated somewhere else in the markup?  I'd probably wait and see if someone has a use case for it though.

PR #2007, opened 2021-03-22
MBS-11508: Move Qobuz favicon to external-favicons

reosarevok:
### Fix MBS-11508

This ended up in the wrong folder after a rebase, I think.

PR #2332, opened 2021-11-14
MBS-12082: Move Tidal favicon to external-favicons

kellnerd:
### Fix MBS-12082

Apparently Qobuz was not the only external icon which has not been moved (see MBS-11508), Tidal (https://staticbrainz.org/images/external-favicons/tidal-32.png) is also missing, so this is the same problem and fix as in #2007. I also checked the remaining files in `favicons/` and it should have been the last one in there.


reosarevok:
@brainzbot, add to whitelist

reosarevok (approved):

PR #2333, opened 2021-11-15
 MBS-12084: Use URL to build the details API links

reosarevok:
### Fix MBS-12084

The details link was broken for JSON for (at least) genres, because they have no default includes but we were using `&` for `fmt`.
This just builds an `URL` object, which seems like the safest option.

### DIFF
--- a/root/entity/Details.js
+++ b/root/entity/Details.js
@@ -47,12 +47,20 @@ const WSLink = ({
   (entityType === 'recording' || entityType === 'release_group') &&
     inc.push('releases');
   entityType === 'release' && inc.push('labels', 'discids', 'recordings');
+  const searchParams = new URLSearchParams();
+  if (inc.length) {
+    searchParams.set('inc', inc.join('+'));
+  }
+  if (isJson) {
+    searchParams.set('fmt', 'json');
+  }
   const protocol = isSecureConnection ? 'https://' : 'http://';
-  const link = '/ws/2/' + entityTypeForUrl + '/' + entityGid +
-               (inc.length ? '?inc=' : '') + inc.join('+') +
-               (isJson ? '&fmt=json' : '');
+  const urlObject = new URL(protocol + DBDefs.WEB_SERVER +
+                            '/ws/2/' + entityTypeForUrl + '/' + entityGid);
+  urlObject.search = searchParams.toString();
+
   return (
-    <a href={link}>{protocol + DBDefs.WEB_SERVER + link}</a>
+    <a href={urlObject.href}>{urlObject.href}</a>
   );
 };
 


root/entity/Details.js
```diff
@@ -49,8 +49,12 @@ const WSLink = ({
   entityType === 'release' && inc.push('labels', 'discids', 'recordings');
```
alastair, line 49: do you have a style guide that recommends this style for conditionals instead of an if statement? just curious - to me this is harder to read than
```js
if (entityType === 'release') {inc.push('labels', 'discids', 'recordings');}
```

root/entity/Details.js
```diff
@@ -49,8 +49,12 @@ const WSLink = ({
...
   const protocol = isSecureConnection ? 'https://' : 'http://';
   const link = '/ws/2/' + entityTypeForUrl + '/' + entityGid +
-               (inc.length ? '?inc=' : '') + inc.join('+') +
-               (isJson ? '&fmt=json' : '');
+               (inc.length ? '?inc=' + inc.join('+') : '') +
+               (isJson && inc.length
+                 ? '&fmt=json'
+                 : isJson
+                   ? '?fmt=json'
+                   : '');
```
alastair, line 57: to remove scary double-ternary operator, what about making an array of query params and joining it?
```js
const params = []
if (inc.length) {
  params.push('inc=' + inc.join('+'))
}
if (isJson) {
  params.push('fmt=json')
}
if (params.length) {
  // or do a ternary when making const.link as you currently do?
  link = link + '?' + params.join('&')
}
```

a bit longer, but more explicit as to what's happening. I'd have suggested [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) but it's still unsupported in a few older browsers

mwiencek: This component only runs on the server, so we could use `URLSearchParams` as it's supported by Node.

reosarevok: Fine, converted this to use URL in general.

root/entity/Details.js
mwiencek: I don't think we have an eslint rule for that yet, but an `if` would look more clear to me too.

mwiencek (approved):

root/entity/Details.js
reosarevok: I'd also prefer it. A rule doesn't seem to exist? But we could still try to find these somehow and fix them in one go at least, since we're unlikely to reintroduce them if nobody likes them :)

yvanzo (approved):
:ship:

root/entity/Details.js
reosarevok: Opened https://github.com/metabrainz/musicbrainz-server/pull/2341 for the cases I could find. There might be more (I did not check every `&&` by hand, just the ones followed by `<`)

PR #2335, opened 2021-11-16
MBS-12085: Always show a favicon in the external links editor

reosarevok:
### Implement MBS-12085

Since the same URL can have several relationships, and the favicon is at the URL level, the URL favicon takes priority. If the URL doesn't have an assigned favicon, then we check whether any of the assigned relationships is an official site one (with the same check as in `layout/components/ExternalLinks`) and show the homepage favicon if so. If not, we check if it is a blog and show the blog favicon. And if it's neither, we show the no favicon icon, as we would on the sidebar.


reosarevok:

mwiencek (approved):

PR #2336, opened 2021-11-16
MBS-11910: Allow editing rel credits in the external links editor

reosarevok:
### Implement MBS-11910

# Problem
URL relationships can have credits for the non-URL entity, if it is of a creditable type, but there is no way currently to change the credit from the entity page (only from the URL `/edit` page). This is confusing, and we really should allow editing this data from the page URL rels are most commonly edited, same as we now do for relationship dates.

# Solution
This makes it possible to change the credited entity in the same popup where we currently change the dates.
Note that this doesn't implement any changes in the release editor since release-url rels are not currently creditable - if that changes then [we should] make relevant changes to `externalLinkRelationship` in `root/static/scripts/edit/MB/edit.js`

# Testing
Tested manually by adding, editing and removing a credit for the artist for a URL relationship in an `/artist/edit` page (you can do this with any Artist-URL relationship from the artist edit page), including that the row for the relationship is highlighted when the credit changes.

Tested that work edit pages do not crash and that the dialog still allows editing dates for their URLs (works are not creditable so the credit bit is missing there). Same for release pages. Also added one selenium test for display of the credited info (but that doesn't test it being submitted).

# Known issues
Submitting the popup leaves a console warning: "Form submission canceled because the form is not connected". This already happens in production.

### DIFF
--- a/root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
+++ b/root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
@@ -193,10 +193,16 @@ const ExternalLinkAttributeDialog = (props: PropsT): React$MixedElement => {
     }
   };
 
-  const handleSubmit = (event: SyntheticEvent<HTMLFormElement>) => {
+  const handleSubmit = (
+    event: SyntheticEvent<HTMLFormElement>,
+    closeAndReturnFocus: () => void,
+  ) => {
+    event.stopPropagation();
+    event.preventDefault();
     if (hasErrors) {
       dispatch({type: 'show-all-pending-errors'});
-      event.preventDefault();
+    } else {
+      handleConfirm(closeAndReturnFocus);
     }
   };
 
@@ -206,7 +212,7 @@ const ExternalLinkAttributeDialog = (props: PropsT): React$MixedElement => {
         <form
           className="external-link-attribute-dialog"
           onKeyDown={handleKeyDown}
-          onSubmit={handleSubmit}
+          onSubmit={(event) => handleSubmit(event, closeAndReturnFocus)}
         >
           <DateRangeFieldset
             dispatch={dateDispatch}
@@ -234,7 +240,6 @@ const ExternalLinkAttributeDialog = (props: PropsT): React$MixedElement => {
               <button
                 className="positive"
                 disabled={hasErrors}
-                onClick={() => handleConfirm(closeAndReturnFocus)}
                 type="submit"
               >
                 {l('Done')}
--- a/root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
+++ b/root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
@@ -30,13 +30,17 @@ import DateRangeFieldset, {
   partialDateFromField,
   runReducer as runDateRangeFieldsetReducer,
 } from './DateRangeFieldset.js';
+import UrlRelationshipCreditFieldset
+  from './UrlRelationshipCreditFieldset.js';
 
 type PropsT = {
+  creditableEntityProp: 'entity0_credit' | 'entity1_credit' | null,
   onConfirm: ($ReadOnly<Partial<LinkStateT>>) => void,
   relationship: LinkRelationshipT,
 };
 
 type StateT = {
+  +credit: FieldT<string | null>,
   +datePeriodField: DatePeriodFieldT,
   +initialDatePeriodField: DatePeriodFieldT,
 };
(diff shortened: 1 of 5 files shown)


reosarevok:

root/static/scripts/edit/components/UrlRelationshipCreditFieldset.js
```diff
@@ -0,0 +1,53 @@
...
+      type: 'update-relationship-credit',
+    });
+  }
+
+  return (
+    <fieldset>
+      <legend>{l('Credit')}</legend>
+      <FormRowText
+        field={field}
+        label={addColonText(l('Credited as'))}
```
mwiencek, line 42: This seems a little confusing because it's inside a dialog for the URL relationship type but there's no context of which entity is being credited.

 * Can we set a placeholder attribute on the input with the source entity name, like the relationship editor dialogs do?
 * What do you think about renaming the field to "Credited to:" instead?  That seems clearer to me because the dialog is for the URL, so it's who the URL is credited to.

root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
```diff
@@ -215,7 +242,7 @@ const ExternalLinkAttributeDialog = (props: PropsT): React.MixedElement => {
                 className="positive"
                 disabled={hasErrors}
                 onClick={() => handleConfirm(closeAndReturnFocus)}
-                type="submit"
+                type="button"
```
mwiencek, line 245: Is this change needed to resolve something?  We have an `onSubmit` handler and I don't believe that it would be triggered with this anymore (so pressing enter to submit the dialog is likely broken by this).

reosarevok: IIRC this was meant to avoid the "Form submission canceled because the form is not connected" console error.

reosarevok: Yeah, I tested this, and you're right it breaks pressing enter. I changed it back, and it works, but prints "Form submission canceled because the form is not connected" to the console every time, it seems. Still better than breaking the dialog submission with Enter, but ideally we can figure a way to avoid that.

root/static/scripts/edit/externalLinks.js
```diff
@@ -1227,6 +1260,15 @@ const ExternalLinkRelationship =
                   {bracketedText(formatDatePeriod(link))}
                 </span>
               ) : null}
+              {nonEmpty(creditedName) ? (
+                <span className="entity-credit">
+                  {' '}
+                  {bracketedText(texp.l(
+                    'credited as “{credited_name}”',
```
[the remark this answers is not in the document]
reosarevok: Possibly. I'm not 100% sure that artist credit credits and relationship credits should be translated in the same way, but I guess we could have the same string and optionally separate context comments?

root/static/scripts/edit/components/UrlRelationshipCreditFieldset.js
reosarevok: When editing URL relationships, in the same popup as dates + ended.

reosarevok: I changed that item to "Credited to". Did not add a placeholder for now since we no longer do that for credits in the new relationship editor.

reosarevok:
I retested this, found and fixed a thing I had missed earlier (highlights for edited credits), and added some more info about the testing (and the reasonings).

root/static/scripts/edit/components/UrlRelationshipCreditFieldset.js
```diff
@@ -0,0 +1,51 @@
...
+    dispatch({
+      credit: event.currentTarget.value,
+      type: 'update-relationship-credit',
+    });
+  }
+
+  return (
+    <fieldset>
+      <legend>{l('Credit')}</legend>
+      <FormRowText
```
mwiencek, line 38: Should we add some short help text here to indicate when to use this field? It looks a bit sparse with just "Credit." I guess we could just link to https://wiki.musicbrainz.org/Relationships

root/static/scripts/edit/components/ExternalLinkAttributeDialog.js
mwiencek: I pushed a commit which seems to fix that (I guess it was the `onClick` handler).

root/static/scripts/edit/components/UrlRelationshipCreditFieldset.js
reosarevok: It says "Credited to:" just by the field, but not sure if that's better. To be honest, I'm not sure the one line in the doc page about credits is a lot better either...

mwiencek (approved):
I'm not sure about the strings still, but my comments don't seem enough to block the PR, so I'm approving.

PR #2337, opened 2021-11-17
MBS-12087: Fix TT TaggerIcon call

reosarevok:
### Fix MBS-12087

`TaggerIcon` was updated to take entity type and MBID instead of an entity, but the TT `tagger_icon` macro wasn't updated for it, so it was linking to `?id=undefined`

Tested on the URL mentioned on the ticket.

### DIFF
--- a/root/components/common-macros.tt
+++ b/root/components/common-macros.tt
@@ -522,7 +522,7 @@ END -%]
 END -%]
 
 [%~ MACRO tagger_icon(entity) BLOCK -%]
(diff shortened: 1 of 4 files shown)


reosarevok:

yvanzo:
(I skip reviewing this one as it is about recent changes I’m not aware about.)

reosarevok:
Merging this since @mwiencek checked the fix and the test passes.

PR #2338, opened 2021-11-17
MBS-11917: New report: Releases with "Disc n" medium titles

reosarevok:
### Implement MBS-11917

This checks for any releases that have mediums where the name is "CD n", "Disc n" or similar, with n being
the medium position. These should only very, very rarely be correct, and only if they can be proven to really be intended as a disc *title*.


yvanzo (approved):

reosarevok:
@mwiencek: would like your review just in case you can see some better options for the query :)

mwiencek (approved):
the query looks good I think, it only takes about 2 seconds on pink which isn't horrible for a report

PR #2321, opened 2021-11-02
MBS-7727: Restrict Soundcloud to only "get the music" rels

reosarevok:
### Implement MBS-7727

Soundcloud shouldn't be used for license nor crowdfunding relationships (nor, probably, any other that gets added
further down the line). This restricts it to only allow "get the music" relationships (all of them, because apparently some stuff on SoundCloud requires a SoundCloud Go+ subscription, whatever that involves  - not available in my country).

On top of https://github.com/metabrainz/musicbrainz-server/pull/2339 (which used to be part of this PR)

PR #2339, opened 2021-11-18
MBS-11980: Test relationship types restriction in external links editor

reosarevok:
### Implement MBS-11980

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -2559,6 +2559,7 @@ const testData = [
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://mainlynorfolk.info/watersons/',
        only_valid_entity_types: ['artist'],
+limited_link_type_combinations: ['otherdatabases'],
   },
   {
                      input_url: 'http://www.mainlynorfolk.info/martin.carthy/records/themoraloftheelephant.html',
@@ -2566,6 +2567,7 @@ const testData = [
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://mainlynorfolk.info/martin.carthy/records/themoraloftheelephant.html',
        only_valid_entity_types: ['release'],
+limited_link_type_combinations: ['otherdatabases'],
   },
   {
                      input_url: 'https://www.mainlynorfolk.info/watersons/songs/countrylife.html',
@@ -2573,6 +2575,10 @@ const testData = [
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://mainlynorfolk.info/watersons/songs/countrylife.html',
        only_valid_entity_types: ['work'],
+limited_link_type_combinations: [
+                                  'otherdatabases',
+                                  ['lyrics', 'otherdatabases'],
+                                ],
   },
   // maniadb
   {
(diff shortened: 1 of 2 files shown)


reosarevok:
This *should* now be correct - the new array features aren't really used by the test here, though, just by the one in https://github.com/metabrainz/musicbrainz-server/pull/2321

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4984,6 +4989,55 @@ function doMatchSubtest(
...
+    }
+    return optionNames;
+  });
+
+  if (relationshipOptions.length > 0) {
+    actualRelationshipOptions = relationshipOptions;
+  }
+
+  const msg = 'Match ' + label + ' URL relationship options for ' +
+  entityType + ' entities';
```
mwiencek, line 5022:
```suggestion
  const msg = 'Match ' + label + ' URL relationship options for ' +
    entityType + ' entities';
```

mwiencek (approved):

reosarevok:
@yvanzo: given you've worked with this more, I'd still like another look from you before I merge :)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4529,14 +4529,18 @@ const testData = [
...
+ expected_relationship_options: ['downloadpurchase', 'streamingpaid'],
   },
   {
                      input_url: 'https://vimeo.com/ondemand/inconcert#comments',
             expected_clean_url: 'https://vimeo.com/ondemand/inconcert',
+             input_entity_type: 'recording',
        input_relationship_type: 'streamingpaid',
        only_valid_entity_types: ['recording', 'release'],
+ expected_relationship_options: ['downloadpurchase', 'streamingpaid'],
   },
```
yvanzo, line 4544: These examples are wrong. The given URL `https://vimeo.com/ondemand/inconcert` should be linked using both `downloadpurchase`and `streamingpaid` relationship types. I don’t know if that applies to all Vimeo pages though. Sure things are:

* We need to know if all of Vimeo On Demand pages provide always both options or just any combination of these.
* Relationship types auto-selection for Vimeo On Demand URLs must be fixed to allow both relationship types at the same type (or to enforce selecting both if that matches Vimeo On Demand’s offer).
* Options example must be updated accordingly.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4984,6 +4989,55 @@ function doMatchSubtest(
   previousMatchTests.push(entityType + '+' + url);
 }
 
+function doOptionsSubtest(
```
yvanzo, line 4992: I suggest naming tests of `CLEANUPS.restrict` with a more closely related name:

```suggestion
function doRestrictSubtest(
```

(“Options” may also refer to relationship attributes.)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4984,6 +4989,55 @@ function doMatchSubtest(
...
+) {
+  const checker = new Checker(cleanURL(url), entityType);
+  const possibleOptions = checker.getPossibleTypes();
+  let actualRelationshipOptions = possibleOptions || undefined;
+  if (expectedRelationshipOptions === undefined) {
+    actualRelationshipOptions = undefined;
+  }
+
+  // The options can be just one relationship type, or arrays of 2+ types
+  const relationshipOptions = possibleOptions.map(optionSet => {
```
yvanzo, line 5007: I would suggest to consistenly use the term “combination” that is already used for this matter: `relationshipTypeCombinations`.

yvanzo (changes requested):
The given example are not actually correct. It would be nice to have a better coverage of this new subtest with more diverse example allowing at least two different combinations and at least one combination of at least two relationship types.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: It has to support all three options, as per the old discussion on https://tickets.metabrainz.org/browse/MBS-11952 - will fix accordingly.

yvanzo (approved):
Tested both tickets, that works nicely :) Made minor comments only. Also a more detailed commit message about the new `limited_link_type_combinations` test property would help me with writing a full documentation of URLCleanup tests from commit <https://github.com/metabrainz/musicbrainz-server/commit/dcabded48be065374c55bfb789ad588daa363729>’s message.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4984,6 +5003,55 @@ function doMatchSubtest(
...
+      expectedTypeCombinations.sort(),
+      (a, b) => {
+        if (Array.isArray(a) && Array.isArray(b)) {
+          return arraysEqual(a.sort(), b.sort());
+        } else if (typeof a === typeof b) {
+          return a === b;
+        }
+        return false;
+      },
+    ),
```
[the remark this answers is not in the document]
reosarevok: Good point, I think the sorts in the inner `arraysEqual` are no longer needed. Removing.

reosarevok:
Updated the commit message with a lot more info, which is hopefully useful enough? :)

PR #2341, opened 2021-11-19
Replace && with if/ternary

reosarevok:
@alastair [asked](https://github.com/metabrainz/musicbrainz-server/pull/2333#discussion_r749438055) why we use 'conditions && action' instead of an if statement or ternary in some files, and both @mwiencek and I didn't have any good reasoning and prefer the alternative option. So this changes the few cases I could find to use if statements or ternaries. I feel this makes things a lot more clear.

I also moved some conditions to a constant in index.js for readability.

### DIFF
--- a/root/layout/components/MergeHelper.js
+++ b/root/layout/components/MergeHelper.js
@@ -61,7 +61,7 @@ const MergeHelper = ({merger}: Props): React.Element<'div'> => {
           }
         </p>
         <div className="buttons" style={{display: 'table-cell'}}>
-          {merger.ready_to_merge &&
+          {merger.ready_to_merge ? (
             <button
               className="positive"
               name="submit"
@@ -69,7 +69,8 @@ const MergeHelper = ({merger}: Props): React.Element<'div'> => {
               value="merge"
             >
               {l('Merge')}
-            </button>}
+            </button>
+          ) : null}
           <button name="submit" type="submit" value="remove">
             {l('Remove selected entities')}
           </button>
--- a/root/entity/Details.js
+++ b/root/entity/Details.js
@@ -42,11 +42,18 @@ const WSLink = ({
   const inc = [];
   const entityTypeForUrl = entityProperties.url
     ? entityProperties.url : entityType;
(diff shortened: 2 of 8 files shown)


root/layout/components/MergeHelper.js
```diff
@@ -61,15 +61,16 @@ const MergeHelper = ({merger}: Props): React.Element<'div'> => {
           }
         </p>
         <div className="buttons" style={{display: 'table-cell'}}>
-          {merger.ready_to_merge &&
+          {merger.ready_to_merge ? (
```
alastair, line 64: Maybe @MonkeyDo could chime in on this too - 
My specific question was about doing `condition && statement` when the statement had a side effect, such as in the case of `entityProperties.aliases && inc.push('aliases');`

In jsx, I've always used `condition && <something>` when deciding if something should be output conditionally. So I'm not sure if it makes sense to use a ternary operator here. I guess the && syntax is pretty well accepted as a way of rendering react.

That is,
* in code use an if statement if the body performs an action.
* use a ternary if you want to choose between 2 values
* use && in jsx when choosing to conditionally display something

MonkeyDo: I agree, in JSX the format `{ myCondition && <render something>}` is preferable.

Falsy values will be ignored by React and nothing will be rendered.

reosarevok: Eww :) Well, we don't use that format almost anywhere, so I'd much prefer to be consistent. I also find it confusing and ugly, but that's a secondary issue.

alastair: If you use the ternary operator everywhere in JSX then I guess that's another thing, and being consistent would be better. so feel free to proceed!

mwiencek: IIRC we were only bit by the `&&` a few times when `myCondition` was the number `0` (since React would render that).  We do use the ternaries pretty consistently at this point.

mwiencek (approved):

PR #2342, opened 2021-11-22
MBS-12097: Show total edit count on user pages

reosarevok:
### Implement MBS-12097

We already have the number anyway (it's just a sum of the different categories) so I see no reason not to show this and save people the effort of summing it themselves. Moved the (view) link from the header to this row, since it makes no sense to duplicate it IMO.


reosarevok:

mwiencek (approved):

PR #2345, opened 2021-11-23
MBS-7912: Add edit search option "edit [not] in subscriptions"

reosarevok:
### Implement MBS-7912

This is one of the last (the last?) options that has a dedicated page, but cannot be used in edit searches, meaning we can't actually refine the edit lists on the /subscribed page using edit search.

This search checks (AFAICT) every single condition that /subscribed does: artist, label and series subscriptions, and all the possible collection subscriptions. It does *not* check editor subscriptions, which have a separate page (/subscribed_editors), and can already be searched within the edit search separately.

Tested by creating a bunch of collections locally, and by checking my current subscriptions in pink to see the results seemingly match /subscribed (keep in mind that page is ordered with oldest edits from the last two weeks first, newest last).

The query doesn't seem slower than /subscribed itself, but I didn't test quantitatively. If you feel it's needed I can look into that.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/EditSubscription.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/EditSubscription.pm
@@ -1,0 +1,203 @@
+package MusicBrainz::Server::EditSearch::Predicate::EditSubscription;
+use Moose;
+
+with 'MusicBrainz::Server::EditSearch::Predicate';
+
+has user => (
+    is => 'ro',
+    isa => 'MusicBrainz::Server::Authentication::User',
+    required => 1
+);
+
+sub operator_cardinality_map {
+  return (
+    'subscribed' => undef,
+    'not_subscribed' => undef,
+  )
+}
+
+sub combine_with_query {
+    my ($self, $query) = @_;
+
+    $query->add_where([ <<~'SQL', [ ($self->user->id) x 14 ] ]);
+      EXISTS (
+        SELECT TRUE
+          FROM edit_area
+         WHERE edit_area.area IN (
+               SELECT area
+                 FROM editor_collection_area
+                WHERE EXISTS (
+                      SELECT TRUE 
+                        FROM editor_subscribe_collection
+                       WHERE collection = editor_collection_area.collection
+                         AND editor = ?
+                      )
+               )
+           AND edit_area.edit = edit.id
+      ) OR EXISTS (
+        SELECT TRUE
+          FROM edit_artist
+         WHERE edit_artist.artist IN (
+               SELECT artist
+                 FROM editor_subscribe_artist
+                WHERE editor = ?
+                UNION
+               SELECT artist
+                 FROM editor_collection_artist
+                WHERE EXISTS (
+                      SELECT TRUE 
+                        FROM editor_subscribe_collection
+                       WHERE collection = editor_collection_artist.collection
+                         AND editor = ?
+                      )
+               )
+           AND edit_artist.edit = edit.id
+      ) OR EXISTS (
+        SELECT TRUE
+          FROM edit_event
+         WHERE edit_event.event IN (
+               SELECT event
+                 FROM editor_collection_event
+                WHERE EXISTS (
+                      SELECT TRUE 
+                        FROM editor_subscribe_collection
+                       WHERE collection = editor_collection_event.collection
+                         AND editor = ?
+                      )
+               )
+           AND edit_event.edit = edit.id
+      ) OR EXISTS (
+        SELECT TRUE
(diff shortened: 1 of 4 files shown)


jesus2099:
Hi @reosarevok.
Wow, this is amazing all the super nice features you add!

> This is one of the last (the last?) options that has a dedicated page, but cannot be used in edit searches, meaning we can't actually refine the edit lists on the /subscribed page using edit search.

There is also https://musicbrainz.org/collection/f0574792-6af6-481e-8b2f-d5ea9fc59c4d/edits that cannot be refined.
But this is just to answer your _**the last?**_ wondering because, for me, this MBS-7912 is already super incredible! it's just up to me to **subscribe only to what I want to review** (which is **already the case!**).

reosarevok:
Good point, and that's also something I want to add eventually. We'll get there :)

root/edit/search_macros.tt
```diff
@@ -424,6 +432,7 @@
                    [ 'vote_count', l('Vote tally') ],
                    [ 'edit_note_author', l('Edit Note Authors'), {requires_login => 1} ],
                    [ 'edit_note_content', l('Edit Note Content'), {requires_login => 1} ],
+                   [ 'edit_subscription', l('Edit'), {requires_login => 1} ],
```
yvanzo, line 435: Just “Edit” in an edit search form doesn’t give any clue what this condition is about.

```suggestion
                   [ 'edit_subscription', l('Subscribed entity'), {requires_login => 1} ],
```

root/edit/search_macros.tt
```diff
@@ -384,6 +386,12 @@
   </span>
 [% END %]
 
+[% MACRO predicate_edit_subscription(field, field_contents) WRAPPER wfield predicate="edit_subscription" %]
+  [% operators([ [ 'subscribed', l('is in my subscriptions') ]
+                 [ 'not_subscribed', l('is not in my subscriptions') ]
```
yvanzo, line 391: To follow and complete the previous comment:
```suggestion
  [% operators([ [ 'subscribed', l('is directly affected') ]
                 [ 'not_subscribed', l('is not directly affected') ]
```

Also suggested to add “directly” in case the edit search doesn’t match the same edits as the subscriptions email.

lib/MusicBrainz/Server/EditSearch/Predicate/EditSubscription.pm
```diff
@@ -0,0 +1,203 @@
...
+               SELECT artist
+                 FROM editor_collection_artist
+                WHERE EXISTS (
+                      SELECT TRUE 
+                        FROM editor_subscribe_collection
+                       WHERE collection = editor_collection_artist.collection
+                         AND editor = ?
+                      )
+               )
+           AND edit_artist.edit = edit.id
```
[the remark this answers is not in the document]
reosarevok: Well, the email query seems to exclude anything that wasn't applied nor is open, at the very least, so it's not quite the same.

root/edit/search_macros.tt
reosarevok: I'm not sure about the wording but you're right about the 'Edit' name being confusing.

yvanzo: It would be even clearer indeed.

root/edit/search_macros.tt
yvanzo: > I don't understand directly affected

Some edits are related to entities but not directly affecting these. For example, editing the work composed by an artist is not directly affecting the artist, but it still is related and appears in artist’s editing history.

> nor the relation with the subscription email.

I wondered if the edits matched for building the subscriptions email are the same as the edits matched by this search.

Anyway, my suggestion is no longer relevant since _Edit_ is about to be replaced with _Edited entity_ rather than _Subscribed entity_.

yvanzo (approved):
:ship:

root/edit/search_macros.tt
jesus2099: Ok now I understand, thanks!

kellnerd:
I've just tested the new filter on beta and unfortunately it always returns the same results for me (subscribed entities as far as I can see), regardless of whether I select *is in my subscriptions* or *is not in my subscriptions* ... At least it works for the common case 😉

jesus2099:
Yes indeed, the not in my subscription criteria displays the are in my subscription, indeed.

reosarevok:
Seems I managed to forget to implement that option. Yay me! Thanks for the report, I'm working on it.

reosarevok:
This should be fixed in beta, can you test again? :)

kellnerd:
Thanks, it's working as expected now.

PR #2346, opened 2021-11-23
MBS-12083: Add caching to get_types_for_timeline

reosarevok:
### Fix MBS-12083

This takes forever unless cached, so we should cache it and avoid everyone suffering that pain every time.

@mwiencek: I haven't used cache much before, so I tried to copy the way `get_all` responses get cached. Please do let me know if I (probably) did a dumb.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Relationship.pm
+++ b/lib/MusicBrainz/Server/Data/Relationship.pm
@@ -525,21 +525,11 @@ sub get_types_for_timeline
 
     for my $t ($self->all_pairs) {
         my $table = join('_', 'l', @$t);
-        my $data = $self->sql->select_list_of_hashes(<<~"SQL", @$t);
-                SELECT link_type_filtered.id,
-                       link_type_filtered.name,
-                       link_type_filtered.parent,
-                       count(l_table.id)
-                  FROM $table l_table
-            RIGHT JOIN link ON l_table.link = link.id
-            RIGHT JOIN (SELECT *
-                          FROM link_type
-                         WHERE entity_type0 = ?
-                           AND entity_type1 = ?) link_type_filtered
-                    ON link.link_type = link_type_filtered.id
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Data/Relationship.pm
```diff
@@ -551,6 +556,7 @@ sub get_types_for_timeline
         }
     }
 
+    $cache->set('types_for_timeline', \%rel_types);
```
mwiencek, line 559: I guess we should set a timeout here in case it's not evicted before 24 hours or so.  Does that seem like a reasonable amount of time?
```suggestion
    $cache->set('types_for_timeline', \%rel_types, 60 * 60 * 24);
```

reosarevok: Either that, or it should just never expire and then be expired by hand (well, by Perl) when editing or adding a relationship type, since that often happens rarely (there might be entire months where nothing changes)

mwiencek: I was going to say that expiring by hand might be annoying since the `get_types_for_timeline` query also touches the `link` table to get a count of rels.  But looking at the query again I'm not sure why it needs a count, or why it joins the link table at all if it only uses `$_->{name}` below.  Can you explain how that works?

reosarevok: ... I probably took that from the `count.ar.links.table.type_name` query I started from, and never took it out, which does seem dumb, now that you point it out. Wonder how much faster the whole thing will be if I drop that.

reosarevok: A lot. The query needed to be like 1/5th of what it was right now. Updated the PR with that in another commit.

mwiencek (approved):

PR #2347, opened 2021-11-24
MBS-9375: Restrict Apple Music to paid streaming and/or purchase 

reosarevok:
### Implement MBS-9375

This has been wanted forever, but now the restrict tool finally allows us to do it. For now, this just allows paid streaming, paid download or a combination of both for all entities (artist, label, recording, release). If we eventually decide to allow for Apple Music relationships for artists and/or labels, this should be modified accordingly.

I also readded the validation that got dropped with https://github.com/metabrainz/musicbrainz-server/pull/1643

On top of https://github.com/metabrainz/musicbrainz-server/pull/2339 (which allows testing it).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -799,6 +799,45 @@ const CLEANUPS: CleanupEntries = {
       url = url.replace(/^(https:\/\/music\.apple\.com)\/([a-z-]{3,})\//, '$1/us/$2/');
       return url;
     },
+    validate: function (url, id) {
+      const m = /^https:\/\/music\.apple\.com\/[a-z]{2}\/([a-z-]{3,})\/[0-9]+$/.exec(url);
+      if (m) {
+        const prefix = m[1];
+        switch (id) {
+          case LINK_TYPES.downloadpurchase.artist:
+          case LINK_TYPES.streamingpaid.artist:
+            if (prefix === 'artist') {
+              return {result: true};
+            }
+            return {
+              result: false,
+              target: ERROR_TARGETS.ENTITY,
+            };
+          case LINK_TYPES.downloadpurchase.label:
+          case LINK_TYPES.streamingpaid.label:
+            if (prefix === 'label') {
+              return {result: true};
+            }
+            return {
+              result: false,
+              target: ERROR_TARGETS.ENTITY,
+            };
+          case LINK_TYPES.downloadpurchase.recording:
+          case LINK_TYPES.streamingpaid.recording:
+            return {
+              result: prefix === 'music-video',
+              target: ERROR_TARGETS.ENTITY,
+            };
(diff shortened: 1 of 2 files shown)


reosarevok:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -787,12 +788,76 @@ const CLEANUPS: CleanupEntries = {
...
+                 to the appropriate release or recording instead.`,
+                {
+                  artist_url_pattern: (
+                    <span className="url-quote">{'/artist'}</span>
+                  ),
+                },
+              ),
+              result: false,
+              target: ERROR_TARGETS.ENTITY,
+            };
```
yvanzo, line 825: Now that there are [URL error messages for mismatched entity types ](https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-29/root/static/scripts/edit/externalLinks.js#L653-L691), this specific message seems unneeded:

```suggestion
            return {
              result: false,
              target: ERROR_TARGETS.ENTITY,
            };
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -787,12 +788,76 @@ const CLEANUPS: CleanupEntries = {
...
+                 to the appropriate release or recording instead.`,
+                {
+                  label_url_pattern: (
+                    <span className="url-quote">{'/label'}</span>
+                  ),
+                },
+              ),
+              result: false,
+              target: ERROR_TARGETS.ENTITY,
+            };
```
yvanzo, line 844: Same as above:

```suggestion
            return {
              result: false,
              target: ERROR_TARGETS.ENTITY,
            };
```

yvanzo (approved):
LGTMBDNT. Just suggested to remove error messages specific to this external website that were implemented before generic error messages were available for mismatched entity types.

reosarevok:
Hmm. We have those for Apple Books and for iTunes, so it would be weird not to have them for Apple Music - we could remove them from all, if you don't feel they give more useful information than the generic option?

yvanzo:
Yes the message about Apple Books `/author` URLs doesn’t seem to be very useful.
But the other message about `/book` provides more useful info by pointing to BookBrainz.

reosarevok:
Ok, updated this with an extra commit removing those, except for `/book`

mwiencek (approved):
I'm not sure I understand why the error messages being removed are less useful than something like https://github.com/metabrainz/musicbrainz-server/pull/2279#discussion_r722257302, since the generic message "This URL is not allowed for labels." doesn't provide any useful info compared to the old ones.  If they're too much work to maintain I'm okay with removing them, but perhaps it's possible to improve the generic ones to say which entity types the entered URL applies to in the future?

reosarevok:
Merging this for now, although I also kinda feel the error messages are better than the generics? But we can always undo the removal of those if we change our mind before release.

PR #2348, opened 2021-11-25
MBS-12102: Fix cover art check on RG TO_JSON

reosarevok:
### Fix MBS-12102

This broke in 7d28f5c7efb224b670ee3853c1336ad3e1c53abc (https://github.com/metabrainz/musicbrainz-server/pull/1868) since I changed the meaning of has_cover_art but did not change this check to still use the original variable.

Tested manually.


yvanzo (approved):
:ship:

PR #2349, opened 2021-11-26
MBS-12105: Block editors from sending messages if they are blocked from adding notes

reosarevok:
### Implement MBS-12105

The whole point of blocking notes is to block abusive communication, which obviously also includes possible abusive emails.


yvanzo (approved):
:ship:

PR #2350, opened 2021-11-29
Lower I/O load caused by dump scripts

yvanzo:
# Problem

The server running these scripts is almost unavailable for anything else when these scripts are running because they require too much I/O operations at the same time.

# Solution

* Use `ionice -c3` to decrease the priority of these I/O operations;
* Create temporary files in the same volume as their final destination, so that moving files (with `mv`) doesn’t require to actually copy their content.

# Notes

It seems safe to use `ionice`, but the decrease of performance caused by not using `overlay2` driver for temporary files may possibly overcome the benefits of not copying files being moved.

# Action

1. Update the specialized containers running these scripts.


admin/RunExport
```diff
@@ -50,6 +49,16 @@ chown "$FTP_USER:$FTP_GROUP" \
     "$FTP_DATA_DIR"/replication \
     "$FTP_DATA_DIR"/replication/{daily,weekly}
 
+# Create temporary directory under the same volume as the final
+# destination so that moving files does not require copying those.
+# If we're not doing a full export, just use default tmp instead.
+if [ "$WITH_FULL_EXPORT" ]
+then
+    make_temp_dir "$FTP_DATA_DIR"/fullexport/
```
zas, line 57: If I understand correctly this is a volume, then the data is copied somehow to final destination, right?
Can you explain the whole process?

https://github.com/metabrainz/musicbrainz-server/blob/771f3cc3362b42fdd1e364cf0a6ee136f35a7de9/admin/config.default.sh#L5
https://github.com/metabrainz/docker-server-configs/blob/bc51522124eb0f7dc6dd83ba676a996b93374230/scripts/services.sh#L1197
https://github.com/metabrainz/docker-server-configs/blob/bc51522124eb0f7dc6dd83ba676a996b93374230/scripts/services.sh#L315

lib/MusicBrainz/Script/MBDump.pm
```diff
@@ -172,13 +172,13 @@ sub make_tar {
     system
         'bash', '-c',
             'set -o pipefail; ' .
-            'tar' .
+            'ionice -c3 tar' .
```
zas, line 175: I think it would be safer to use `--ignore` `ionice` option to let the script run even if it fails to set the requested priority.
This capability may be disabled (user, docker, kernel, or whatever).
Also a check for `ionice` command existence is likely safer too.

Basically we should make `ionice` optional, if it is there and is working, we use it, else just run stuff without it.

Also I would recommend full option names for readability.

Something like:

```bash
ioniceit() { if type ionice >/dev/null 2>&1; then ionice --class 3 --ignore $@ ; else $@; fi }

ioniceit command_to_ionice
```

zas (changes requested):

admin/RunExport
yvanzo: Right. This patches makes `make_temp_dir` to create a temporary directory in this volume (instead of in /tmp).

1. It dumps the database to files (under the temporary directory): https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-15/admin/RunExport#L53-L61
2. It copies these files to the backup destination (in another volume): https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-15/admin/RunExport#L65-L70
3. After removing files for private tables, it moves the rest to the final destination (in the same volume): https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-15/admin/RunExport#L77-L80
4. After organizing files and deleting old exports, it pushes changes to the `fullexport` directory to the FTP server: https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-15/admin/RunExport#L95

Note: “Full export” is the main database dump, but the code has been generalized to work with other kind of dumps (sample data, json dumps, incremental json dumps, search index dumps).

lib/MusicBrainz/Script/MBDump.pm
yvanzo: Added (with quoted "$@") in commit <https://github.com/metabrainz/musicbrainz-server/pull/2350/commits/96f56fbfac57290527537d71eaaa02a8a3cbf655>.

bin/rsync-fullexport-files
```diff
@@ -15,6 +15,7 @@ if [ -n "$RSYNC_FULLEXPORT_HOST" ]; then
         --archive \
         --delete \
         --exclude='latest-is*' \
+        --exclude='mbscript-*' \
```
zas, line 18: Is this related to this patch?

admin/BundleSearchIndexesDump
```diff
@@ -187,7 +193,7 @@ for my $hash_program ('md5sum', 'sha256sum') {
     my $hash_output_file = uc($hash_program . 's');
     chomp (my $hash_bin = `which g$hash_program` || `which $hash_program`);
     my $hash_command = "cd $working_dir" .
-        " && $hash_bin --binary *.tar.zst > $hash_output_file";
+        " && ionice -c3 $hash_bin --binary *.tar.zst > $hash_output_file";
```
zas, line 196: `ioniceit` is missing here.
It would be more convenient to have it as a small executable script instead of a function, to avoid duplication.

zas (changes requested):

bin/rsync-fullexport-files
yvanzo: Yes; See the message of commit <https://github.com/metabrainz/musicbrainz-server/pull/2350/commits/5d3a2af1e87d089aaaf94019d0b118625fdaa194>:

> To prevent publishing temporary files, those are excluded from 'rsync'.

Indeed the name of the created temporary directories start iwth `mbscript-`; See https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-29/admin/functions.sh#L9

admin/BundleSearchIndexesDump
yvanzo: Added missing call and moved definition to a file in commit https://github.com/metabrainz/musicbrainz-server/pull/2350/commits/7da6795190217c968ae83855bbe6bf08068a8681.

yvanzo:
It seems that JSON dumps are creating their own temporary directory (based on default `tmpdir()`):
https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-29/lib/MusicBrainz/Script/JSONDump.pm#L74-L78

Moreover, it is outputting to another temporary directory (which is moved to the final destination’s volume in this patch): https://github.com/metabrainz/musicbrainz-server/blob/v-2021-11-29/admin/RunJSONDump#L34-L35

`TMP_EXPORT_DIR` cannot be defined using a `make_temp_dir` call since it can be called only once.

@mwiencek: Does it seem worth simplifying/unifying this cascade of temporary directories? Do you have any suggestion?

zas (approved):
LGTM

mwiencek:
> @mwiencek: Does it seem worth simplifying/unifying this cascade of temporary directories? Do you have any suggestion?

I don't remember this code perfectly, but I think the idea was that since the big entity files (which contain one entity per line) are built in batches, they won't be copied to the main temp dir (the output dir) until they're completed.  So we won't end up with partial/incomplete files in the output dir.

Are these temp directories in separate volumes?  If so I guess we should at least change the `DIR => tmpdir(),` bit to go under the same volume.

lib/MusicBrainz/Script/JSONDump.pm
```diff
@@ -71,10 +71,19 @@ has output_dir => (
...
-our $TMP_EXPORT_DIR = tempdir(
-    'json-dump-XXXXXX',
-    DIR => tmpdir(),
-    CLEANUP => 0,
+has tmp_export_dir => (
+    is => 'ro',
+    isa => 'Str',
+    lazy => 1,
+    default => sub {
+      my @output_path = splitdir($self->output_dir);
```
mwiencek, line 79: You'll need `my $self = shift;` above here to use `$self`.

lib/MusicBrainz/Script/JSONDump.pm
```diff
@@ -71,10 +71,19 @@ has output_dir => (
...
+      pop @output_path;
```
mwiencek, line 80: Maybe we can stick with `tmpdir()` by default but allow overriding it via an argument to the script?  If I use `~/musicbrainz-server` as the output dir while testing, I'm not sure I want it using my home dir as a temp dir and accidentally deleting stuff there.  In other cases there might not be permission to access `../`.

yvanzo: It seems less clunky. Added option in commit <https://github.com/metabrainz/musicbrainz-server/pull/2350/commits/9ba591e2802857fb0b1ffc542b7e78473b043d0d>.

yvanzo:
@mwiencek: When you have time, please review this pull request. (It seems that a review request cannot be refreshed.) It has been manually deployed to production five months ago by patching the containers directly. It was not enough to solve the alerts but it did not cause any harm either. Since the containers have been recreated, it is not currently in use. I just resolved the conflicts that occurred with the timestamp refactoring that has been merged recently.

yvanzo:
Resolve new conflicts again, this time caused by 0d3866005f7abdf4f1bc59d31ae5a6c4cba4a162 on `MBDump`.

mwiencek:
Sorry for the lack of response previously, I'll review it this week.

yvanzo:
Just rebased on the current `master` to resolve conflicts without any other change.

lib/MusicBrainz/Script/MBDump.pm
```diff
@@ -172,13 +173,13 @@ sub make_tar {
     system
         'bash', '-c',
             'set -o pipefail; ' .
-            'tar' .
+            "$FindBin::Bin/ioniceit tar" .
```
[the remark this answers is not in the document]
yvanzo: I made the suggested change in commit <https://github.com/metabrainz/musicbrainz-server/pull/2350/commits/4bfd590017e729edfc3ceff618f311aa87a4baaa>, in `lib/MusicBrainz/Script/MBDump.pm` only (which is under `lib/`) not in `admin/BundleSearchIndexesDump` (which is under the same `admin/` as `ioniceit`).

admin/RunExport
```diff
@@ -55,6 +54,16 @@ chown "$FTP_USER:$FTP_GROUP" \
     "$FTP_DATA_DIR"/replication \
     "$FTP_DATA_DIR"/replication/{daily,weekly}
 
+# Create temporary directory under the same volume as the final
+# destination so that moving files does not require copying those.
+# If we're not doing a full export, just use default tmp instead.
```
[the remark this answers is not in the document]
yvanzo: 1. The default temporary directory usually has better performances.
2. It would make assumptions from `admin/RunExport` on `admin/ExportAllTables`.
    When `WITH_FULL_EXPORT` is set to `1`, their final destination under `$FTP_DATA_DIR` is known in `admin/RunExport`.
    In the opposite case (when `WITH_FULL_EXPORT` is set to `0`), it isn’t clear what happens. I guess that it is used to make replication packets only, and these seem to be written under this temporary directory too. But I didn’t find what happens next.

yvanzo:
> Did you test both changes in isolation and see what effects each has? These changes make sense in theory, but it's hard to know what to think about both of them without any numbers. I only saw "did not cause any harm." :)

No, I deployed both at the same time in production in Nov 2021.

> > It was not enough to solve the alerts but it did not cause any harm either.
> 
> Does this mean it only a small effect, or no effect at all? (If small, do we know which change had the effect?)

IIRC, there were two main observations:
- Did not have to wait to log in through SSH. But I did it manually and that might just be luck.
- Less alerts have been triggered at that time. But load depends on many factors.

> > It seems safe to use ionice, but the decrease of performance caused by not using overlay2 driver for temporary files may possibly overcome the benefits of not copying files being moved.
> 
> This suggests the change to avoid copying files may be a wash, unless I misunderstood. So I'm not sure whether it's good or bad (without numbers).
> I wonder the same about `ionice`, especially after reading [this Stack Overflow post](https://stackoverflow.com/questions/48367857/ionice-idle-not-having-the-expected-effects). It sounds promising but if it only did no harm, that suggests it's not doing what we expect.

The only reason why `aretha` is currently not throwing `HostUnusualDiskWriteLatency` alerts two days per week is because we silenced these since April. However we did not silence the still current load alerts.

- It is hard to test these changes outside of production servers as it would require to have a similar server setup which none of us have.
- In production, there are many other factors that may vary over time.

So I don’t think that any “numbers” can be collected to give any certitude about the real impact of these changes. We can only make indirect observations such as alerts being thrown, the data used for these alerts, maybe automate measuring the login time at regular intervals.

We can also split this pull request and deploy the two changes one after the other if you prefer, to observe their presumed impact after a month.

admin/RunExport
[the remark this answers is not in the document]
mwiencek: OK, I guess it doesn't matter much since the packets aren't large to begin with.

admin/RunIncrementalJSONDump
```diff
@@ -43,6 +47,7 @@ echo Disabling all database triggers
 echo Making incremental JSON dumps
 ./admin/DumpIncrementalJSON \
     --output-dir "$TEMP_DIR" \
+    --tmp-dir "$(make_temp_dir --extra "$DUMP_DIR")" \
```
mwiencek, line 50: Would this directory be cleaned up/removed?  I don't think I see where it would be.  Just want to make sure we're not accidentally syncing temporary files to FTP.

mwiencek: With `--extra` it doesn't set up the rmdir trap AFAICT, though I'm not sure that'd work anyway if there are still empty subdirectories.

mwiencek: OK, I forgot about the rsync `--exclude`, so syncing them shouldn't be an issue, but maybe they aren't deleted still.

mwiencek:
>  We can also split this pull request and deploy the two changes one after the other if you prefer, to observe their presumed impact after a month.

This sounds reasonable to me.  If you'd like to keep a single PR you can just cherry-pick the commits you want to production & rebase this.

lib/MusicBrainz/Script/JSONDump.pm
```diff
@@ -423,7 +433,7 @@ sub run {
     my $exit_code = $self->run_impl($c);
 
     $c->connector->disconnect;
-    rmtree($TMP_EXPORT_DIR);
+    rmtree($self->tmp_export_dir);
```
mwiencek, line 436: I think this probably answers my previous question -- I was confusing `--tmp-dir` with `tmp_dir` from lib/MusicBrainz/Script/MBDump.pm

mwiencek: Running `rmtree` on what is now a configurable directory (and now under the final volume) made me a bit uneasy, so I pushed a1cc4e8

yvanzo:
OK, I opened MBS-12538 and added a schedule for deployment. I will just cherry-pick the changes and apply these directly to the container.

yvanzo:
Just rebased on current `master` branch’s `HEAD` commit 154240528e8183b6741320ae63c204d91902ff9a.

yvanzo:
Tests shown that these changes had no effect on disk latency so far.
However, it might due be an infrastructure setup issue: [MBH-581](https://tickets.metabrainz.org/browse/MBH-581).

yvanzo:
I scrapped most of the commits as agreed on with @bitmap. See notes in PR’s description and in comment to the ticket MBS-12538.

mwiencek (approved):
LGTM

PR #2352, opened 2021-11-30
Fix updating settng maintenance mode (again)

yvanzo:
# Problem

Deployment failure occurred on `yehudi` (also preventing deployment on next node: `zappa`) because the website container is currently disabled on that node.

# Solution

In the (current) event where a service is disabled on a specified host, this patch prevents deployment failure due to trying to set maintenance mode for a non-existing service/container.

### DIFF
--- a/script/update_containers.sh
+++ b/script/update_containers.sh
@@ -48,47 +48,63 @@ shift
 
 SERVICES="musicbrainz-webservice musicbrainz-website"
 
-if [ $# -eq 0 ]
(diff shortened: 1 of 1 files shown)


mwiencek (approved):
I haven't tested this but the changes seem sensible.

reosarevok (approved):
I understand the last fixes to the last release were done with this already, so if it works and @mwiencek is happy with it, I say we merge.

yvanzo:
Yes, used to fix deployment as follows:

```bash
./script/update_containers.sh prod yehudi zappa
```

PR #2354, opened 2021-12-01
MBS-12114: Account for the "Disk" alternative spelling to "Disc"

reosarevok:
### Implement MBS-12114

This is a reasonably common use, even if it's not as common as "Disc", so it makes sense to check for it too in the release editor and for the report.

Tested by hand (using it and seeing it was detected in the RE, then detected by the report).


mwiencek (approved):

reosarevok:

PR #2360, opened 2021-12-10
MBS-12121: Actually show parent changes in EditRelationshipType

reosarevok:
### Fix MBS-12121

This was actually passing `parent`, not `parentId`, so the previous setup didn't work at all. `parent` seems better anyway (if we can show a link why not show a link), so this just actually uses it.

Tested by just checking the edit search for Edit relationship type changes.


yvanzo (approved):
Tested with `pink`’s database too.

:ship:

PR #2361, opened 2021-12-10
MBS-12124: Add report for license URLs

reosarevok:
### Implement MBS-12124

Most of these are going to be wrong, because people keep misusing the relationship (even the sample data already contains a dozen incorrect usages!). This is sorted with least uses first, rather than last like other URL reports, since the most used ones are also the most likely to be correct.

screencapture-reolappy-5000-report-LicenseLinks-2021-12-10-11_05_35


yvanzo (commented):
It seems that this report isn’t specific enough to be possibly ever empty. Would it be reasonable to filter out the most used correct license URLs with a pattern to avoid having a too many false positive?

reosarevok:
It's not designed to ever be empty, yeah. I guess we could filter off Creative Commons links, but there would still be other valid license links left. This is meant to be a "start from the beginning where rels have only one use and are probably mostly but not all bad, stop whenever you feel it all seems legit" kind of report, but I wouldn't dare to block any with more than X uses because someone might spam the same link in many places without it being a license :/

reosarevok:
Actually, this report would also be useful specifically to find valid license URLs and see all the music under each license; I just used it for that purpose for a request in #musicbrainz.

PR #2362, opened 2021-12-10
MBS-12125: Allow admins to see all users with specific privileges

reosarevok:
### Implement MBS-12125

This gives admins a good way to, for example, find all users who have been blocked from editing. Since I saw no reason to limit the amount of searchable flags, this includes all of them.


mwiencek (changes requested):
If you hit search without selecting anything, it returns every editor in the DB (on a single page) -- should probably fix that. :)

reosarevok:

PR #2363, opened 2021-12-11
Use a valid SPDX License expression

akshaaatt:
More information is available at https://spdx.org/licenses/


yvanzo (commented):
* Is SPDX identifier expected by some tool you are using? It might be worth mentioning which error does it fix for which tool.
* The letter D of SPDX is missing in your commit’s message (and thus in your pull request’s title too).

mwiencek (approved):
> Is SPDX identifier expected by some tool you are using? It might be worth mentioning which error does it fix for which tool.

I'm curious about that too, but it seems sensible to follow the standard in any case.

(At the same time, this isn't a real package being published; `"private": true` is set and we're only using it to track dependencies. So it's unclear what files the license field here is even referring to.)

yvanzo:
> I'm curious about that too, but it seems sensible to follow the standard in any case.

Yes

akshaaatt:
Corrections made!

package.json
```diff
@@ -93,5 +93,5 @@
     "utf8": "2.1.2",
     "xgettext-js": "3.0.0"
   },
-  "private": true
+  "private": false
```
mwiencek, line 96: Sorry for the misunderstanding, but I wasn't requesting for this to be changed.  Please revert this, as this isn't meant to be published.

akshaaatt: Fixed in https://github.com/metabrainz/musicbrainz-server/pull/2377

PR #2365, opened 2021-12-13
MBS-12126: Add URL search to the site

reosarevok:
### Implement MBS-12126

This is already available via the WS so I see no reason not to allow users to use it from the site. The results
are kinda mediocre with normal indexed but still a lot better than not having the option at all, and return only the correct URL with advanced and url:'urlstring', so it seems like a no-brainer to me.

Note: the `UrlResults` component is almost the same as `reports/components/UrlList` - but the two results are built differently (with the url under `entity` for search vs `url` for reports) and it seemed a bit crazy to refactor either completely right now. Eventually we might want to use either `entity` or `$entity_type` in both places.


root/search/components/UrlResults.js
```diff
@@ -0,0 +1,67 @@
...
+  form,
+  lastUpdated,
+  pager,
+  query,
+  results,
+}: ResultsPropsT<UrlT>):
+React.Element<typeof ResultsLayout> => {
+  const columns = React.useMemo(
+    () => {
+      const nameColumn = defineLinkColumn<SearchResultT<UrlT>>({
```
mwiencek, line 29: Should we be adding `target="_blank" rel="noopener noreferrer"` to these?

reosarevok: We don't currently do it on the report list at root/report/components/UrlList.js but that doesn't mean we shouldn't. I added the change (to both files) in a second commit. That said, https://chromestatus.com/feature/6140064063029248 suggests in most browsers `noopener` is not required if you already have `_blank`.

PR #2368, opened 2021-12-14
MBS-9385: Warn when using non-"None" packaging for digital media

reosarevok:
### Implement MBS-9385

This warns at the medium level mostly because that's what comes last in the workflow (so it might avoid going back to the first tab for cases where the error is not the packaging). I'm not warning on unset packaging because I guess that's not wrong per se and I'm not sure we want to pretty much enforce packaging for digital.


root/static/scripts/release-editor/fields.js
```diff
@@ -447,6 +447,16 @@ class Medium {
                 self.release.earliestYear() < mediumFormatDate);
     });
     this.confirmedEarlyFormat = ko.observable(this.hasTooEarlyFormat());
+    this.hasStrangeDigitalPackaging = ko.computed(function () {
+      const isFormatDigital = self.formatID() &&
+                              self.formatID().toString() === '12';
+      return !!(isFormatDigital &&
+                self.release.packagingID() != null &&
+                self.release.packagingID().toString() !== '7'); // "None"
```
yvanzo, line 455: Commenting all hard-coded values even though their meaning is obvious when writing and reviewing it at first:

```suggestion
      const isFormatDigital = self.formatID() &&
                              self.formatID().toString() === '12'; // "Digital Media"
      return !!(isFormatDigital &&
                self.release.packagingID() != null &&
                self.release.packagingID().toString() !== '7'); // "None"
```

It will help with finding them when they must be changed.

yvanzo (commented):
LGTMBDNT. Made some minor comments still.

reosarevok:

yvanzo (approved):
:ship:

reosarevok:

reosarevok:

reosarevok:

reosarevok:
The selenium failure was legitimate, this was actually also accidentally blocking Digital Media with no packaging when the value of the field was `''` and not `undefined`. Now fixed.

reosarevok:

PR #2369, opened 2021-12-14
Respond with Bad Request for "invalid session ID"

mwiencek:
If someone submits a ton of requests with invalid `musicbrainz_server_session` cookies, each request will trigger an internal server error (due to an exception being thrown by Catalyst).  We'd rather respond to these with Bad Request, while still logging them to Sentry, so they at least don't trigger any 5xx alerts.


reosarevok (approved):
LGTMBDNT

mwiencek:
> Should we have a test for this?

Yep, I've added one

yvanzo (approved):
heheh

PR #2370, opened 2021-12-15
MBS-12133: Allow % in jazzmusicarchives URLs

reosarevok:
### Fix MBS-12133

I didn't find any URL like this when originally writing the validation, but they do use % for escaping and those URLs are currently deemed invalid.


yvanzo:
@BrainzBot retest this please

yvanzo (approved):
:ship:

PR #2330, opened 2021-11-10
MBS-12069: Convert EditList (and most usages) to React

reosarevok:
### Implement MBS-12069

On top of https://github.com/metabrainz/musicbrainz-server/pull/2315

PR #2371, opened 2021-12-15
MBS-12134: Convert the cancel edit page to React 

reosarevok:
### Implement MBS-12134

No significant changes. This reuses `EnterEditNote` which means an "Edit Note" heading is added to that section - that doesn't seem particularly problematic so I didn't add a toggle for that. The TT allowed passing `{extra}` under the cancel button but I didn't find any use or reason for it, so this does not allow children.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2330


mwiencek (approved):
worked for me locally

PR #2372, opened 2021-12-16
MBS-12135: Convert the notes received page to React

reosarevok:
### Implement MBS-12135

On top of https://github.com/metabrainz/musicbrainz-server/pull/2371


lib/MusicBrainz/Server/Entity/EditNote.pm
```diff
@@ -125,6 +125,7 @@ around TO_JSON => sub {
     my ($orig, $self) = @_;
 
     my $json = $self->$orig;
+    $json->{edit} = to_json_object($self->edit);
```
reosarevok, line 128: We could choose to make edits part of linkedEntities if we worry that this is too much.

mwiencek (approved):
page works locally with/without new notes, checkbox & cookie setting also work

PR #2373, opened 2021-12-16
MBS-11655 / MBS-12136: Separate tag cloud into genre and non-genre tags, and add a text list

reosarevok:
### Implement MBS-11655 / MBS-12136

See commit messages for details. Tested manually with `pink` data.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Tag.pm
+++ b/lib/MusicBrainz/Server/Data/Tag.pm
@@ -56,20 +56,27 @@ sub load
     load_subobjects($self, 'tag', @objs);
 }
 
-sub get_cloud
+sub _get_cloud_data
 {
-    my ($self, $limit) = @_;
+    my ($self, $get_genres, $limit) = @_;
 
-    my $cache = $self->c->cache('tag');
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/Data/Tag.pm
```diff
@@ -91,6 +98,23 @@ sub get_cloud
...
+    my ($self, $genre_limit, $tag_limit) = @_;
+
+    my $cache = $self->c->cache('tag');
+    my $data = $cache->get('tag_cloud');
+    return $data if defined $data;
+
+    $genre_limit ||= 200;
+    $tag_limit ||= 50;
+
+    $data = {
```
mwiencek, line 114: This'll cause an ISE when it's released on beta/production, since it uses the same cache key but changes the format of the data being stored there.  An easy way to avoid that would be to just change the cache key (genre_and_other_tag_clouds?).

reosarevok: Just `'tag_cloud_data'` seems fine yet different, think I'll go with that

PR #2374, opened 2021-12-17
MBS-12140: Reject empty tag names from /ws/2/tag

mwiencek:
### MBS-12140: Reject empty tag names from /ws/2/tag
Empty or blank strings could be submitted as tag names via the web service and would be accepted by the system.  We now reject these.  A script has been added to delete any existing ones from the database.


reosarevok:
In general looks good, although I'm confused why the script is under /updates if it is not a schema change script. Which reminds me, do we want a DB constraint for this?

mwiencek:
> I'm confused why the script is under /updates if it is not a schema change script.

The ones that are for schema changes get compiled to admin/sql/updates/schema-change/.  But there are a ton of one-off scripts that aren't schema changes under updates/.

> Which reminds me, do we want a DB constraint for this?

Yeah, probably.  With proper validation I think it's low priority, since tag names are unique (so any mess from creating empty ones is limited).  But it does make sense to have some `control_for_whitespace` and `only_non_empty` constraints.

reosarevok:
If so you could already add a change here for the next schema change, adding the constraint (assuming it's easy)

reosarevok (approved):
Seems sensible to me.

yvanzo:
@BrainzBot retest this please

t/lib/t/MusicBrainz/Server/Controller/WS/2/Authenticated.pm
```diff
@@ -269,4 +270,49 @@ EOXML
...
+    $mech->default_header('Accept' => 'application/xml');
+    $mech->credentials('localhost:80', 'musicbrainz.org', 'new_editor', 'password');
+
+    $mech->request(xml_post('/ws/2/tag?client=post.t-0.0.2', $content =~ s/TAG_NAME//r));
+    is($mech->status, 400);
+    is_xml_same($mech->content, $error_message);
+
+    $mech->request(xml_post('/ws/2/tag?client=post.t-0.0.2', $content =~ s/TAG_NAME/ /r));
+    is($mech->status, 400);
+    is_xml_same($mech->content, $error_message);
```
yvanzo, line 315: You might want to add descriptive arguments to the `is` and `is_xml` calls; See @reosarevok 's recent pull request about it.

yvanzo (approved):
LGTMBDNT. The added tests seem to be appropriate but could have description strings.

t/lib/t/MusicBrainz/Server/Controller/WS/2/Authenticated.pm
reosarevok: I will eventually improve these when I get to that folder, if not before :)

PR #2375, opened 2021-12-17
MBS-12132: Allow Apple Music music-video links for releases

reosarevok:
### Implement MBS-12132

These have their own barcodes, so users feel they should be added as releases.
For the future, it would be good if `only_valid_entity_types` could also be validated against all options in `limited_link_type_combinations`.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -821,7 +821,7 @@ const CLEANUPS: CleanupEntries = {
           case LINK_TYPES.downloadpurchase.release:
           case LINK_TYPES.streamingpaid.release:
             return {
-              result: prefix === 'album',
+              result: prefix === 'album' || prefix === 'music-video',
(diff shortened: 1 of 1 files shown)


mwiencek:
> For the future, it would be good if only_valid_entity_types could also be validated against all options in limited_link_type_combinations.

Does that mean we can't test this properly right now, or?

reosarevok:
Nope, it requires us picking a specific reltype to test against.

yvanzo (approved):
Which is possible here: selecting either `downloadpurchase` or `streamingpaid` through `input_relationship_type` allows to test validation, only multiple selections cannot be tested against validation.

reosarevok:
Sure, but we can't test it works for all types in one go, which is probably what we should do :)

mwiencek (approved):

reosarevok:
Added MBS-12177 for that possible improvement of test options.

PR #2376, opened 2021-12-23
MBS-12022: Update URL cleanup of Lantis.jp

y-young:
### Implement MBS-12022

MBS-12022: Update URL cleanup of Lantis.jp


reosarevok (changes requested):
Supporting the new version seems good, but given we have two different accepted regexes for lantis now, I'd split it into its own section, if you can be bothered.

Also, if something like the `.html` in `https://www.lantis.jp/release-item/LACA-15193.html` is unneeded and can be removed with a cleanup, we should probably do so to avoid duplicates being added :)

y-young:
It seems the intended format is the one with `.html` (although the other one also works), so I'll just follow the official format.

y-young:

reosarevok (approved):
Looks good now. You could also add the favicon and whatnot if you want :)

reosarevok:
This won't show up on the sidebar unless it gets an Entity::URL page too - not sure if we want it to appear there? I don't think it'd hurt, at least :)

y-young:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2582,6 +2577,25 @@ const CLEANUPS: CleanupEntries = {
...
+      ),
+      new RegExp(
+        '^(https?://)?(www\\.)?lantis\\.jp/release-item/' +
+        '[A-Z]+-\\d+(\\.html)?$',
+        'i',
+      ),
+    ],
+    restrict: [LINK_TYPES.discographyentry],
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?lantis\.jp\/release-item\/([A-Z]+-\d+)(?:\.html)?$/, 'https://www.lantis.jp/release-item/$1.html');
```
yvanzo, line 2595: * Any particular reason for normalizing the scheme to `https`? And only the `/release-item/` URLs? There are currently 600+ `http`Lantis URLs for 200+ `https` Lantis URLs. The preexisting test has no `http`.
* Same questions for the subdomain name `www`. There are currently less than 100 Lantis URLs without  `www`. The preexisting test has no `www`.

Not sure these changes are intentional as they are mentioned neither in the ticket nor in the commit messages.

The other change which auto-add the ending `.html` looks good to me. There is a way to do only this change:

```suggestion
      url = url.replace(/^((?:https?:\/\/)?(?:www\.)?lantis\.jp)\/release-item\/([A-Z]+-\d+)(?:\.html)?$/, '$1/release-item/$2.html');
```

yvanzo (commented):
@y-young: Sorry it did wait for so long. Your patch addresses the issue described in the ticket but it makes other changes too.

Given that these other changes are minor compared to the benefit of allowing editors to enter new URLs, if we don’t get your input today, we may decide to release it as-is. Then it would have to be fixed in a corrective patch.

root/static/scripts/edit/URLCleanup.js
y-young: The reason for "https" and "www" is that on the website of its parent company - [Bandai Namco Arts](https://www.bandainamcoarts.co.jp/#sites), they link Lantis to https://www.lantis.jp/top.html (Although lantis.jp works with both http/https and with or w/o www).

yvanzo (approved):
Thank you!

I completed the tests by pushing commit https://github.com/metabrainz/musicbrainz-server/pull/2376/commits/738cf28db2394188bea3c774f966600bfab40990.

(Note: The failure of Selenium tests is incidental and unrelated to these changes.)

PR #2378, opened 2022-01-11
Fix various warnings found in the musicbrainz-website-prod logs on hip

mwiencek:
In another PR I'll resolve the "Failed to lookup cover art" warnings by removing support for Amazon cover art and other providers.

PR #2379, opened 2022-01-11
MBS-12157, MBS-12200: Remove support for cover art providers

mwiencek:
Resolves MBS-12157, MBS-12200.

I checked the `release_coverart` table in production, and there's only a single archive.org URL.  There are no ozon.ru URLs.  At some point we should probably drop this table, but perhaps the existing cached data would also be of use to someone in the future, and we should back it up at the same time.


lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -1450,19 +1450,6 @@ sub load_meta
         $obj->amazon_store($row->{amazon_store}) if defined $row->{amazon_store};
```
reosarevok, line 1450: We also don't seem to be using this for anything, what am I missing?

reosarevok (approved):
This seems all good, I'll test it a bit manually later to make sure I see no breakages too. We should probably give at least two weeks' warning before merging this, possibly even longer, and clearly point users to the cover art report (especially since we are removing it with this patch), so that they can actually have some time to add the images to the CAA if they want to. AFAICT we never had a clear blog or forum post saying "we're dropping Amazon support" so far, just some discussion on IRC plus a ticket.

I understand removing this fixes some warnings too - is it possible to fix the warnings earlier while still giving some time for users to add the cover art?

lib/MusicBrainz/Server/Data/Release.pm
mwiencek: Nothing that I can see.  Removed it.

mwiencek:
> I understand removing this fixes some warnings too - is it possible to fix the warnings earlier while still giving some time for users to add the cover art?

I pushed a commit for that to https://github.com/metabrainz/musicbrainz-server/pull/2378

yvanzo:
What about announcing it as part of the next schema change? So we can also drop this table at the same time.

reosarevok:
If that seems fine to @mwiencek and @mayhem then +1 from me.

mayhem:
This is a passage of time I really approve of.

reosarevok:
Tiny conflict to fix here, seems ready otherwise for me, but requesting formal @yvanzo review too :)

reosarevok:

yvanzo:
@reosarevok:

* There are still [141323 releases that have Amazon cover art but no Cover Art Archive front cover](https://musicbrainz.org/report/ReleasesWithAmazonCoverArt). Is that expected?

* The schema change blog post was mostly intended for mirror maintainers and developers rather than editors. Has the removal of support for Amazon cover art been echoed to editors?

I wonder whether most editors don’t know about it or they just don’t mind. The former seems to be the most likely to me.

reosarevok:
We've mentioned on the forums a few times as well - honestly, there's only so much we can do to get the word there, and it's quite annoying to add the images en masse so I expect that's part of the reason we haven't gotten a lot of people working very hard on it. I think it's still fine to drop this.

yvanzo:
> We've mentioned on the forums a few times as well

Found the only editors’ discussion about it on the forums: https://community.metabrainz.org/t/385703/40?u=yvanzo

> I think it's still fine to drop this. As an example, I still have around 2000 releases by labels I follow in that report, I might add a couple dozen more between here and the dropping time, but I certainly cannot be bothered to do more. It's just too time consuming, even with scripts.

For a smoother transition, we can give editors more time by keeping this report up until this number decreases significantly. Since editors can filter releases of this report by their subscriptions, it allows them to add the cover art to the releases they worry about.

yvanzo:
@reosarevok:
* Rebased on the latest `schema-change-2022-q2`.
* Reverted the code for updating and displaying the report in commit https://github.com/metabrainz/musicbrainz-server/pull/2379/commits/c98ebe5138ebb89f54a1c4e3411565678d6747a2.
* Updated the report’s description in commit https://github.com/metabrainz/musicbrainz-server/pull/2379/commits/f42e8acf010eabc1b936a39c6bfa9d5abaef8641.

root/report/ReleasesWithAmazonCoverArt.js
```diff
@@ -24,9 +24,9 @@ const ReleasesWithAmazonCoverArt = ({
     canBeFiltered={canBeFiltered}
     description={l(
       `This report shows releases which have cover art from Amazon, but have
-       no front cover in the Cover Art Archive. The use of Amazon art
-       is going to be discontinued eventually, so these releases will lose
-       their front cover unless one is added to the Cover Art Archive.`,
+       no front cover in the Cover Art Archive. The use of Amazon art has been
+       discontinued since the 16th of May 2022, so these releases have no
+       front cover unless one is added to the Cover Art Archive.`,
```
reosarevok, line 29: What about something like:

```
      `This report shows releases which have cover art on Amazon, but have
       no front cover in the Cover Art Archive. The use of Amazon art has been
       discontinued since the 16th of May 2022, so these releases have no
       front cover art anymore until one is added to the Cover Art Archive.`,
```

reosarevok: ("from Amazon" suggests they still have it, and I feel "until" is more clear than "unless" about what we expect to happen :) )

yvanzo: Included your changes in commit https://github.com/metabrainz/musicbrainz-server/pull/2379/commits/1600f4938060f7caf7bd8c19ae0b82fad7c94d78

yvanzo (approved):
Tested but the migration (which will be tested with the `schema-change-2022-q2` branch).

PR #2380, opened 2022-01-14
MBS-12159: Add support for Amazon.pl

reosarevok:
### Implement MBS-12159

This allows also music.amazon.pl links, but those do not seem to exist yet. It seems easier to just accept them and assume users won't add them until they are available, though.
Many allowed domains already don't seem to actually work with music.amazon, such as cn, se, com.tr.


yvanzo (approved):
LGTMBDNT. It is sensible to process `pl` links the same as `cn` and all.

PR #2381, opened 2022-01-14
MBS-12147: State that an email search had no results

reosarevok:
### Implement MBS-12147

Currently there's no clear way to differentiate an empty search from a not-yet-submitted one - both just show the email address in the form. This makes it so that after a search that returned no results a message stating that fact is shown.


reosarevok:

mwiencek (approved):

PR #2383, opened 2022-01-15
MBS-12167: Use returnToCurrentPage in UserSubscriptions action

reosarevok:
### Implement MBS-12167

Other subscribe / unsubscribe links go back to the same page, but this was going to the front page. Returning here seems a lot more user-friendly.


yvanzo (approved):
LGTMBDNT. Compared with other usages of `returnToCurrentPage`. :ship:

PR #2384, opened 2022-01-15
MBS-12161: Expand, not replace, FormHandler Text's validate

reosarevok:
### Fix MBS-12161

`MusicBrainz::Server::Form::Field::Text` was completely replacing `sub validate` from the `HTML::FormHandler::Field::Text` package it extends. This means the checks for both `maxlength` and `minlength` were not running, making `maxlength` in `Form::Annotation`'s changelog completely useless.


yvanzo (approved):
LGTMBDNT. Thanks for the test.
Checked [`HTML::FormHandler::Field::Text`](https://metacpan.org/dist/HTML-FormHandler/source/lib/HTML/FormHandler/Field/Text.pm) for reference.

mwiencek (approved):

PR #2385, opened 2022-01-15
MBS-12155: Filter out downvoted tags from JSON WS user-tags/genres

reosarevok:
### Fix MBS-12155

These are already filtered out in `_serialize_user_tag` and `_serialize_user_genre` in `XMLSerializer`, but they were not being looked into in the JSON serializer. This just adds an extra `grep` for `is_upvote` to `serialize_tags` to effectively do the same there.

For some reason, we had no tests at all for JSON user-tags, too. Added a JSON version of some of the ones we have for XML, and modified the test data so that it will fail if this bug is broken again.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
@@ -314,7 +314,9 @@ sub serialize_tags {
     if ($inc->user_tags) {
         $into->{'user-tags'} = [
             sort { $a->{name} cmp $b->{name} }
-            map +{ name => $_->tag->name }, @{ $opts->{user_tags} }
+            map +{ name => $_->tag->name },
+                grep { $_->is_upvote }
+                @{ $opts->{user_tags} }
         ];
     }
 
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
```diff
@@ -314,14 +314,17 @@ sub serialize_tags {
...
+                @{ $opts->{user_tags} }
         ];
     }
 
     if ($inc->user_genres) {
         $into->{'user-genres'} = [
             sort { $a->{name} cmp $b->{name} }
             map +{ disambiguation => $_->tag->genre->comment, id => $_->tag->genre->gid, name => $_->tag->name },
+            grep { $_->is_upvote }
                 @{ $opts->{user_genres} }
```
yvanzo, line 328: Coding style looks weird. Personally I would always indent. Not sure about line returns and `,` separator.
@mwiencek: any suggestion?

yvanzo (approved):
LGTMBDNT. Thanks for the tests. Would like to have a second review on this.

lib/MusicBrainz/Server/WebService/Serializer/JSON/2/Utils.pm
mwiencek: I guess I would indent the `grep` since `@{ $opts->{user_genres} }` is indented, and it's a continuation of the `map`.  The comma is needed because there are two forms of `map`: https://perldoc.perl.org/functions/map and here it's using `map EXPR,LIST`.  Since it's returning a hash ref, it avoids double brackets in the other form like `map { { id => ... } } @list`.

mwiencek (approved):

PR #2386, opened 2022-01-17
MBS-12160: Use 'canonical' to display sorted edit data JSON

reosarevok:
### Implement MBS-12160

'`canonical`' ensures the JSON is sorted by key. This makes things relatively slower, but it makes it easier to compare different edits' data and to figure out if something is missing or weird. Since the main target of the edit data page are humans, it makes sense IMO to make it easier for them to use by adding this consistency, even if speed goes down a bit.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Edit.pm
+++ b/lib/MusicBrainz/Server/Controller/Edit.pm
@@ -7,6 +7,7 @@ BEGIN { extends 'MusicBrainz::Server::Controller' }
 
 use Data::Page;
 use DBDefs;
+use JSON;
 use MusicBrainz::Server::EditRegistry;
 use MusicBrainz::Server::Edit::Utils qw( status_names );
 use MusicBrainz::Server::Constants qw( :quality );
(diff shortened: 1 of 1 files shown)


lib/MusicBrainz/Server/Controller/Edit.pm
```diff
@@ -102,7 +102,7 @@ sub data : Chained('load') RequireAuth
         component_path => 'edit/EditData',
         component_props => {
             edit => $edit->TO_JSON,
-            rawData => $c->json->pretty->encode($decoded_raw_data),
+            rawData => $c->json->canonical->pretty->encode($decoded_raw_data),
```
mwiencek, line 105: Oops, I'm pretty sure doing `->canonical->pretty->` permanently sets that on `$c->json` for the rest of the request.  Which might be fine if this is the only use during the request, but it's not meant to have settings changed on it. (Note: suggestion needs `use JSON;`.)

```suggestion
            rawData => JSON->new->canonical->pretty->encode($decoded_raw_data),
```

mwiencek: `$c->json` also has `allow_blessed` and `convert_blessed` enabled, but I don't think those are needed here.

reosarevok: Oh! Ok, updated :)

yvanzo:

mwiencek (approved):

PR #2387, opened 2022-01-17
MBS-12162: Do not require auth for non-private tag WS queries

reosarevok:
### Fix MBS-12162

We were requiring auth for anything related to the tag resource, but as per `WS::2::Tag` only `tag_lookup` and `tag_submit`
should require auth (since they actually interact with the user's own tags). Tag search, which just searches for tag names matching a string, should not require any sort of authentication, since it's equivalent to other public searches.

Tested by actually going to both `/ws/2/tag/?query=shoegaze` and `/ws/2/tag?id=ed35bc92-2b5a-4ddf-96d2-51af9ab239e7&entity=artist` and making sure the first no longer requires logging in, while the second still does.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Validator.pm
+++ b/lib/MusicBrainz/Server/WebService/Validator.pm
@@ -319,7 +319,8 @@ role {
 
             # Check if authorization is required.
             $c->stash->{authorization_required} = $inc->{user_tags} || $inc->{user_genres} || $inc->{user_ratings} ||
-                $resource eq 'tag' || $resource eq 'rating' ||
+                $resource eq 'rating' ||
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
LGTMBDNT but we are missing tests for this `validate` method. @reosarevok: Not sure how to add it to the MBS Test Review doc?

reosarevok:
I'll try and add a test for this - but feel free to add a new section at the end maybe for stuff we *know* is missing tests?

reosarevok:
I started https://github.com/metabrainz/musicbrainz-server/pull/2411 for adding `WebService::Validator` tests.

PR #2388, opened 2022-01-19
MBS-12168: Avoid breaking list formatting on user bios and collections

reosarevok:
### Fix MBS-12168

This is the same as what we already do for annotations, which had the same problem (https://github.com/metabrainz/musicbrainz-server/pull/1160/). Tested by hand.


reosarevok:

mwiencek (approved):

PR #2389, opened 2022-01-20
Update HACKING.md with recent changes

reosarevok:
This updates HACKING.md to refer to the new types folder, the now split sets of ESlint rules, `React.useContext` and the `to_json` functions, which were all added in the last year or so.

I'm probably missing more things that should be added or changed, and I'd like to clarify the part about where components go since they don't all go under `/common`, but it gets the ball rolling.


HACKING.md
```diff
@@ -156,6 +156,10 @@ done with the following command:
 
     $ prove -l t/js.t
 
+We also have a set of tests of client-side functions that can be run (after
+compiling) by opening http://localhost:5000/static/scripts/tests/web.html
+on your local development server.
+
```
yvanzo, line 162: FWIW, this is the way I’m running URL Cleanup tests since these have been ported to ES6 and thus can no longer be run directly in Node (without Chrome). But it can also be run from the command-line by relying on a headless instance of Chrome in the background, which is how it runs in CircleCI:

```suggestion
We also have a set of tests of client-side functions that can be run (after
compiling) by opening http://localhost:5000/static/scripts/tests/web.html
on your local development server.

Provided your system is running a (headless) chrome instance, these client-side
tests can also be run from the command-line, using:

    $ node t/web.js
```

HACKING.md
```diff
@@ -289,11 +304,13 @@ specify its name in an argument:
 
 ### Adding a new report
 
- 1. Create new module in */lib/MusicBrainz/Server/Report/*.
- 2. Add created module into [ReportFactory.pm](https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/ReportFactory.pm).
+ 1. Create new module in `/lib/MusicBrainz/Server/Report/`.
```
yvanzo, line 307: Additionally, the leading `/` can be removed since all paths are relative to the `musicbrainz-server` directory. It was confusing with the `/lib` directory of the FHS.

```suggestion
 1. Create new module in `lib/MusicBrainz/Server/Report/`.
```

yvanzo (approved):
LGTM. Made complementary optional suggestions only. Feel free to merge with or without these.

HACKING.md
```diff
@@ -367,9 +387,15 @@ Common instructions for porting:
     };
     ```
     Where `convert_to_json` is a function that converts `$self->prop_name` to
-    its appropriate JSON representation. If `prop_name` is an entity that has
-    a `TO_JSON` method defined, it will be called automatically and you can
-    simply do `prop_name => $self->prop_name`.
+    its appropriate JSON representation.
+    
+    If `prop_name` is an entity that has a `TO_JSON` method defined, it will be
+    called automatically and you can simply do `prop_name => $self->prop_name`.
```
mwiencek, line 393: This is no longer true since https://github.com/metabrainz/musicbrainz-server/pull/1958 -- you have to call `prop_name => $self->prop_name->TO_JSON` (or `prop_name => to_json_object($self->prop_name)` if you want to handle undef values).

reosarevok:
Made the fixes. Any other suggestions on stuff that should be updated? We can always merge now and do more edits later, as well :)

yvanzo:
Is `http://localhost:5000/static/scripts/tests/all.html` broken just for me?

reosarevok:
No, that's also broken for me. @mwiencek said it should work, maybe it's a recent issue?

HACKING.md
```diff
@@ -156,6 +156,15 @@ done with the following command:
 
     $ prove -l t/js.t
 
+We also have a set of tests of client-side functions that can be run (after
+compiling) by opening http://localhost:5000/static/scripts/tests/web.html
```
mwiencek, line 160: Yeah, re: the previous comments about all.html, it hasn't worked in a browser in years from what I can tell.  But, it's useless anyway because web.html replaced it.  (all.html is broken because it currently tries to run the Node tests; there is nothing to fix because the web tests are already run by web.html.)

So the additions here aren't wrong, but kinda confusing when read after the previous sentence about all.html, which is broken and has an unclear purpose.  I've proposed merging the Node and browser tests in https://github.com/metabrainz/musicbrainz-server/pull/2403 and made some small changes to HACKING.md there; if that looks okay we can revert the changes here so that they don't conflict.

mwiencek: That's been merged, so you can rebase this PR & remove the changes here.

reosarevok: Done, thanks!

mwiencek (approved):

PR #2391, opened 2022-01-21
Improve URLCleanup test docs

reosarevok:
This adds a large, detailed comment on top of the URLCleanup tests file that specifies what each property means and how they can be combined. Happy to improve it further with any suggestions :)

While writing this I noticed a small bug with `expected_error` and that the `limited_link_type_combinations` properties were not ordered  as expected, fixed in a separate commit for each.

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -15,6 +15,106 @@ import {
   Checker,
 } from '../../edit/URLCleanup';
 
+/*
+ * This file tests the cleanups and autoselect / restrictions defined in
+ * root/static/scripts/tests/Control/URLCleanup.js
+ *
+ * The main part of the file is the testData object, which contains
+ * the expected results for all the tests. The following properties
+ * are supported in testData:
+ *
+ * input_url:
+ *      The raw URL "entered by the user" (a string).
+ *
+ *      This is always mandatory.
+ *
+ * input_entity_type:
+ *      The entity type selected by the user (a string, such as 'release').
+ *      This is important because the same URL can be assigned to different
+ *      relationship types (or blocked altogether) depending on the selected
+ *      entity type.
+ *
+ *      Required by 'expected_relationship_type', 'expected_error'
+ *      and 'limited_link_type_combinations'. Forbidden if none are present.
+ *
+ * expected_relationship_type:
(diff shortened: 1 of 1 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -15,6 +15,106 @@ import {
...
+ *      might be autoselected, while others are just available for selection
+ *      to the user. To test for autoselection, also use
+ *      'expected_relationship_type' alongside this.
+ *      Each of the allowed combinations can be just a string
+ *      or an array of strings. For example, a possible value for this
+ *      property would be the following:
+ *      [
+ *        'downloadpurchase',
+ *        'streamingpaid',
+ *        ['downloadpurchase', 'streamingpaid'],
```
mwiencek, line 62: IIRC the nested array means "both can be selected at the same time" but it'd be nice to mention that explicitly -- I seem to always forget.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -15,6 +15,106 @@ import {
...
+ *      relationship types (or blocked altogether) depending on the selected
+ *      entity type.
+ *
+ *      Required by 'expected_relationship_type', 'expected_error'
+ *      and 'limited_link_type_combinations'. Forbidden if none are present.
+ *
+ * expected_relationship_type:
+ *      The relationship type (or types) we expect the URL
+ *      to get autoselected to. Either a string, such as 'downloadfree',
+ *      or an array of strings, such as ['downloadfree', 'streamingfree'].
```
mwiencek, line 43: I'd perhaps mention what these strings are or where they come from: they're keys from the `LINK_TYPES`/`CLEANUPS` objects in the code.  Otherwise it'd be reasonable to assume these are link_type names.

mwiencek (approved):
looks great, I just suggested two possible improvements

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Done in 13d9e10

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Done in 13d9e10

yvanzo (approved):
Crystal clear!

PR #2393, opened 2022-01-22
Don't follow type tables for incremental sitemaps

mwiencek:
Incremental sitemaps ingest replication packets and follow the tables therein to find changes to pages in the sitemaps.  If a type table is updated, like release_group_primary_type for example, we'd end up trying to fetch every page for every release group using a modified type.  (It doesn't matter which columns were changed.)  I can't think of a situation where changing a type would significantly affect the content of so many pages that we'd want search engines to re-index them all, so I'm having the code skip these type tables entirely.


reosarevok (approved):
The closest I can think of is if we change a type name, but even that might not merit a full reindex of it all anyway.

PR #2394, opened 2022-01-24
Enable Plack::Middleware::AccessLog in production

mwiencek:
As noted in https://github.com/metabrainz/syswiki/pull/45/files#r781430759, it's often helpful to see the last request preceding an error or warning in production.  Or the last request before a plackup process became stuck.

The AccessLog middleware is normally only enabled in development; if you use plackup in development, you should be used to seeing requests logged.  This enables the same logging in production.

This will increase the amount of logs generated by the website and web service containers by quite a lot.  However, assuming we aggressively fix spammy warnings and set the appropriate log-opts on the containers, this patch would help make debugging alerts much easier.

### DIFF
--- a/app.psgi
+++ b/app.psgi
@@ -65,6 +65,18 @@ builder {
     }
 
     enable 'Static', path => qr{^/(static/|browserconfig\.xml|favicon\.ico$)}, root => 'root';
+
+    # AccessLog is a default middleware when PLACK_ENV=development
+    # (the default) [1], but we also want it enabled in production to
+    # aid debugging.  The conditional ensures it's not enabled twice in
+    # development -- only in production when PLACK_ENV=deployment.
+    #
+    # [1] https://metacpan.org/pod/plackup#-E,-env,-the-PLACK_ENV-environment-variable.
+    enable_if {
+        my $plack_env = $ENV{PLACK_ENV};
+        defined $plack_env && $plack_env eq 'deployment'
+    } 'Plack::Middleware::AccessLog', format => 'combined';
+
     MusicBrainz::Server->psgi_app;
 };
 


mwiencek:
> Seems sensible. Did you see if this brings up a lot more cruft to fix?

Well, it won't cause any additional warnings/errors to be logged, just what requests are being made to the container alongside the IP, user agent, etc.  It will make existing warnings/errors easier to track down, but I haven't deployed this in production yet to check.

reosarevok:
Oh, ok, now I realized what we're talking about exactly :) Seems sensible indeed.

mwiencek:
> However, assuming we [...] set the appropriate log-opts on the containers

Apparently there are default settings for these that I wasn't aware of: https://docs.docker.com/config/containers/logging/local/

@yvanzo do you think we need to tweak these opts at all on the containers before merging this?

mwiencek:
@brainzbot test this please

yvanzo (commented):
> do you think we need to tweak these opts at all on the containers before merging this?

Not immediately. These options are set at daemon level already. Even in the event this change would spam the logs, it would not cause any issue with disk usage. Given the current max size, it is unlikely to be reached within two weeks (usual lifespan for a container) or even within months. We can tweak these options or revert this patch if it happens to be really too verbose.

I deployed this patch in beta containers already.

However before merging this, is it going to affect mirrors too? Maybe consider having a specific flag (either in DBDefs or in environment) to turn it on/off?

mwiencek:
> However before merging this, is it going to affect mirrors too? Maybe consider having a specific flag (either in DBDefs or in environment) to turn it on/off?

If `--env deployment` is passed to plackup, then yes.  If they're not doing that, then requests were already being logged.  We don't document to do that anywhere, but [musicbrainz-docker does it](https://github.com/metabrainz/musicbrainz-docker/blob/16565197fe2b4f798b6c3b45d3eda44ffc0c44b6/build/musicbrainz-dev/scripts/start.sh#L12) at least -- I guess copied from musicbrainz-server (though it probably shouldn't be set for development anyway).  I'm okay with adding a DBDefs flag if someone requests one, but it seems a bit premature to add one right now?  Otherwise we'll end up with flags nobody uses.

yvanzo:
Yes it should probably not be used in development setup, but it is also [used in normal setup](https://github.com/metabrainz/musicbrainz-docker/blob/v-2022-01-31/build/musicbrainz/scripts/start.sh#L20). It comes [from the past](https://github.com/metabrainz/musicbrainz-docker/commit/5f6b9e46fd0bc6b2dd2751880b91fcabb26cc5dd) indeed but it seems to be appropriate.

yvanzo (approved):
@brainzbot retest this please

yvanzo:
@brainzbot retest this please

PR #2396, opened 2022-01-25
Fix regression in medium track count comparison

reosarevok:
The order in this row was switched for (I assume) readability (track_count first like elsewhere), but for some reason < turned into >=. We do not want to show there are unloaded tracks if the track count is exactly the same as the amount of loaded tracks though!


yvanzo (approved):
Also tested locally with `pink` database and various [releases with more than 100 tracks](https://beta.musicbrainz.org/search?query=tracks%3A%5B100+TO+*%5D&type=release&limit=25&method=advanced). :ship:

PR #2397, opened 2022-01-25
MBS-2313: Avoid pointless empty annotation edits

reosarevok:
### Fix MBS-2313

This was only checking if latest_annotation existed before happily using 'text', but 'text' is actually nullable. It seems like emptying an annotation through the release editor will not null it, but emptying it through the usual edit annotation page will. As such, the release editor will replace the null annotation with an empty string one, in what is a completely pointless exercise.

It might be preferable to just make annotation NOT NULL and always ensure it's an empty string when blanked.

### DIFF
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -977,7 +977,7 @@ class Release extends mbEntity.Release {
     this.packagingID = ko.observable(data.packagingID);
     this.barcode = new Barcode(data.barcode);
     this.comment = ko.observable(data.comment);
-    const annotationText = data.latest_annotation
+    const annotationText = data.latest_annotation?.text
       ? data.latest_annotation.text
       : '';
     this.annotation = ko.observable(annotationText);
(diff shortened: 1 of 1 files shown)


root/static/scripts/release-editor/fields.js
```diff
@@ -977,7 +977,7 @@ class Release extends mbEntity.Release {
     this.packagingID = ko.observable(data.packagingID);
     this.barcode = new Barcode(data.barcode);
     this.comment = ko.observable(data.comment);
-    const annotationText = data.latest_annotation
+    const annotationText = data.latest_annotation?.text
       ? data.latest_annotation.text
       : '';
```
kellnerd, line 982:
```suggestion
    const annotationText = data.latest_annotation?.text ?? '';
```
Could be shortened a bit since this will always be a string, null or undefined. Thanks for finding and fixing this, by the way. (I should really try to set up my own local development server soon, just need to find a partition with enough free space and a bit of time...)

mwiencek (approved):
(I applied kellner's suggestion, should be squashed when merging.)

reosarevok:

yvanzo:

PR #2398, opened 2022-01-25
Add a default editor avatar for display

reosarevok:
Removing Gravatar was great, but the voting icons we use on edit notes expect an avatar to exist, even if just a placeholder - without it, the display is a broken mess (see below). We could change the concept of how we show those, but we do want to use (discourse) avatars in the future, so it's probably best to not change it too much.

Screenshot from 2022-01-26 00-39-17

aerozol came up with a placeholder icon we can use; for now it'll show for all editors, but we could still use it for editors without an actual avatar once we have the discourse option.

I haven't made the same changes to the `link_editor` TT macro because that's not breaking display and, honestly, I didn't feel like figuring out how to import the image on soon to be removed code.

The icon is displayed as 12x12 on edit notes which is tiny, we might want to slightly change that, or we might just leave it be for now. It's much larger on editor pages.


mwiencek:
This looks pretty good to me -- it might be hard to make out what it is at 12px (I only have no trouble because I already know what it is).

What do you think about increasing it to 16px (this also requires removing the `max-height: 1em` rule) and removing the white background/border?

Before:
Screenshot_20220128_104213
After:
Screenshot_20220128_104925

reosarevok:
I don't mind, as long as it doesn't cause problems elsewhere for display :) Feel free to push the change, if you have it already?

mwiencek:
15px looked a little better so I used that.  Here's a screenshot (first with the default avatar, another where it's defined).
Screenshot_20220128_114540

mwiencek (approved):

PR #2400, opened 2022-01-28
Remove parallel processing on prove

mwiencek:
We've been getting spurious failures on CircleCI which I have a feeling are due to running certain tests in parallel; usually it's one of the dump scripts under t/script/ that fail, but I'm not sure which tests are conflicting with them.  Let's try disabling this and seeing if it significantly affects the run-time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,6 @@ jobs:
               --source pgTAP \
               --source Perl \
               -I lib \
-              -j 9 \
               t/critic.t \
               t/pgtap/* \
               t/pgtap/unused-tags/* \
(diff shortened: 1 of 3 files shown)


mwiencek:
@brainzbot test this please

mwiencek:
No failures yet.  The tests run about 3 minutes slower.  That's kinda significant, but probably saves time overall if we're re-running the tests less often.  I think it's worth merging this, since we can always revert the first commit if the failures reappear.

reosarevok (approved):
IMO, let's.

mwiencek:
@brainzbot test this please

PR #2402, opened 2022-01-31
MBS-12183: Display previous value of annotation in AddAnnotation edits

reosarevok:
### Implement MBS-12183

We call everything "AddAnnotation", since they add a new revision of the annotation, but there's quite a big difference between cases where an annotation replaces another, deletes it, or adds one where nothing existed.
This makes it so that a diff is shown when modifying an existing annotation, similarly to the annotations-differences page.

Screenshot from 2022-01-31 14-04-28

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Annotation/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit/Annotation/Edit.pm
@@ -3,7 +3,7 @@ use strict;
 use Carp;
 use MooseX::Role::Parameterized;
 use MooseX::Types::Moose qw( Int Str );
-use MooseX::Types::Structured qw( Dict );
+use MooseX::Types::Structured qw( Dict Optional );
 use MusicBrainz::Server::Constants qw( %ENTITIES );
 use MusicBrainz::Server::Data::Utils qw( model_to_type );
 use MusicBrainz::Server::Edit::Types qw( Nullable NullableOnPreview );
@@ -38,6 +38,7 @@ role {
                 id   => Int,
                 name => Str
             ]],
+            old_annotation_id => Optional[Nullable[Int]],
         ],
     );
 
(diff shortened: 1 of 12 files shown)


lib/MusicBrainz/Server/Edit/Annotation/Edit.pm
```diff
@@ -38,6 +38,7 @@ role {
                 id   => Int,
                 name => Str
             ]],
+            old_annotation_id => Optional[Nullable[Int]],
```
reosarevok, line 41: Optional because old edits don't have it, nullable because many entities do not yet have an annotation, and the first addition will be undef.

yvanzo:
@BrainzBot retest this please

reosarevok:
The test failure is legit, because selenium checks edit data for some reason! :smile: Will fix.

yvanzo (commented):
What changes are expected for the following situations?
- Edit adding an initial annotation: https://musicbrainz.org/edit/87442617
- Edit modifying an existing annotation: https://musicbrainz.org/edit/87233025
- Edit erasing an annotation: https://musicbrainz.org/edit/87442629

I see no change here.

yvanzo:
Thanks, that explains why the only examples would be screenshots and no examples from the current database. :)

What do you think about displaying the content of “Annotation comparison” instead of the content of “Text” for all new edits?

It seems that displaying both fields when available is redundant and useless. It would make the display more compact and more similar to other edits (when editing the fields of an entity for example).

reosarevok:
They're not quite the same, because the diff is the raw text, but the text section displays it as the site would (so, with markup converted to links and whatnot). IIRC I asked on IRC back when I wrote this and someone (atj?) suggested that it seemed fine to have the diff look at the raw text (sometimes you wouldn't see the difference at all otherwise).

yvanzo (approved):
Ok, thanks for the details. It’s not obvious when looking at simple examples with no markup indeed.

I don’t think we need both but that might just be me. Your current patch is indubitably an improvement anyway. Let’s merge it as-is and amend it in beta if people think it’s too redundant.

PR #2403, opened 2022-02-01
Merge Node and browser JS tests

mwiencek:
Prior to 814c5d4e7b7b28c938490018cdcb6909037ae8dd, we used to run JS tests from the command line using PhantomJS.  This was considered clunky, so we switched to running them under Node.js instead (using jsdom to emulate the DOM).

A lot of the reasoning in that commit became obsolete.  For one, since moving to Webpack and taking advantage of its plugins (ProvidePlugin and DefinePlugin mainly), we ended up having to bundle our tests into giant files anyway.  It was nice for the short while we could simply run them with babel-node, but that is no longer the case.

We've also replaced jsdom in the meantime; headless Chrome became usable some time in 2017 and replaced it in
67fc46ab356b58f9e1fedb6b3ac000951d5ef136.  This is when web.html appeared.  Nevertheless, we continued to run some JS tests that didn't require a browser in Node.  At some point all.html stopped working in the browser and became useless, because anything that needed the DOM was now in web.html.

Since it's quite easy in 2022 to run JS tests from the command line in a headless browser, this commit removes support for running them in Node. Any tests that were *only* run from Node (tests/index.js) have been moved into tests/index-web.js.  all.html, which I believe hasn't worked in years, has been removed in favor of web.html.  t/js.t disappears too as it only served to execute Node.


reosarevok (approved):
Seems sensible.

yvanzo (approved):
IIRC, another reason for running tests with Node was that SSR is not done in the browser, and some functions can behave differently. But I don’t see any SSR-specific test either. Maybe something to add to our test coverage doc?

mwiencek:
> IIRC, another reason for running tests with Node was that SSR is not done in the browser, and some functions can behave differently. But I don’t see any SSR-specific test either. Maybe something to add to our test coverage doc?

Well, SSR is tested somewhat generally through the Perl Test::WWW::Mechanize tests and Selenium tests (as long as there is a test requesting a particular page), but we don't have any SSR-specific unit tests.  I guess there are not many utility functions that run only on the server, but it's not something I'd rule out.

PR #2404, opened 2022-02-01
MBS-12188: Allow SNAC links for places

reosarevok:
### Implement MBS-12188

These exist, as per the added example, so we should support them.


yvanzo (approved):
Makes sense. LGTMBDNT. Tests seem to cover it. :ship:

PR #2406, opened 2022-02-02
MBS-12189: Support "names" prefix for DAHR artists

reosarevok:
### Implement MBS-12189

DAHR now has a permalink with the `/names` prefix for artists. It also has a new `/mastertalent` prefix for artists, which they
redirect to, but they do not call a permalink. As such, this accepts both, but cleans `/mastertalent` to `/names`.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1484,16 +1484,28 @@ const CLEANUPS: CleanupEntries = {
     },
   },
   'dahr': {
-    match: [new RegExp(
-      '^(https?://)?adp\\.library\\.ucsb\\.edu/index\\.php/' +
-      '(matrix|objects|talent)',
-      'i',
-    )],
+    match: [
+      new RegExp(
+        '^(https?://)?adp\\.library\\.ucsb\\.edu/index\\.php/' +
+        '(mastertalent|matrix|objects|talent)',
+        'i',
+      ),
+      new RegExp('^(https?://)?adp\\.library\\.ucsb\\.edu/names/', 'i'),
+    ],
     restrict: [LINK_TYPES.otherdatabases],
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?adp\.library\.ucsb\.edu\/index\.php\/([a-z]+)\/[a-z]+\/([\d]+).*$/, 'https://adp.library.ucsb.edu/index.php/$1/detail/$2');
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1484,16 +1484,27 @@ const CLEANUPS: CleanupEntries = {
...
+      new RegExp('^(https?://)?adp\\.library\\.ucsb\\.edu/names/', 'i'),
+    ],
     restrict: [LINK_TYPES.otherdatabases],
     clean: function (url) {
-      return url.replace(/^(?:https?:\/\/)?adp\.library\.ucsb\.edu\/index\.php\/([a-z]+)\/[a-z]+\/([\d]+).*$/, 'https://adp.library.ucsb.edu/index.php/$1/detail/$2');
+      url = url.replace(/^(?:https?:\/\/)?adp\.library\.ucsb\.edu\/index\.php\/([a-z]+)\/[a-z]+\/([\d]+).*$/, 'https://adp.library.ucsb.edu/index.php/$1/detail/$2');
+      url = url.replace(/^(?:https?:\/\/)?adp\.library\.ucsb\.edu\/names\/([\d]+).*$/, 'https://adp.library.ucsb.edu/names/$1');
+      // mastertalent URLs match 1:1 to a names permalink so we use that
+      url = url.replace(/^(?:https?:\/\/)?adp\.library\.ucsb\.edu\/index\.php\/mastertalent\/detail\/([\d]+).*$/, 'https://adp.library.ucsb.edu/names/$1');
+      return url;
```
yvanzo, line 1501: To simplify the code a bit, the scheme can be replaced once and for all, then the scheme and the domain name can be captured rather than repeatedly rewritten in the code.

```suggestion
      url = url.replace(/^(?:https?:\/\/)?adp\.library\.ucsb\.edu\//, 'https://adp.library.ucsb.edu/');
      url = url.replace(/^(https:\/\/adp\.library\.ucsb\.edu)\/index\.php\/([a-z]+)\/[a-z]+\/([\d]+).*$/, '$1/index.php/$2/detail/$3');
      url = url.replace(/^(https:\/\/adp\.library\.ucsb\.edu)\/names\/([\d]+).*$/, '$1/names/$2');
      // mastertalent URLs match 1:1 to a names permalink so we use that
      url = url.replace(/^(https:\/\/adp\.library\.ucsb\.edu)\/index\.php\/mastertalent\/detail\/([\d]+).*$/, '$1/names/$2');
      return url;
```

yvanzo (approved):
LGTMBDNT. Test seems to be appropriate. Suggested a small optional code simplification.

PR #2408, opened 2022-02-04
MBS-12191: Allow admins to see spammer profiles

reosarevok:
### Implement MBS-12191

I blocked these a bit too well, to the point where Freso himself cannot see them either anymore.
This allows admins to see the page, but adds a visible warning that they are already banned so that they don't have to go check the flags to make sure.

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -14,6 +14,7 @@ import UserAccountLayout, {
 } from '../components/UserAccountLayout';
 import DescriptiveLink
   from '../static/scripts/common/components/DescriptiveLink';
+import Warning from '../static/scripts/common/components/Warning';
 import {FLUENCY_NAMES} from '../static/scripts/common/constants';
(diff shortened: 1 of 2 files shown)


root/user/UserProfile.js
```diff
@@ -860,14 +861,15 @@ const UserProfile = ({
   addedEntities,
 }: UserProfileProps): React.Element<typeof UserAccountLayout> => {
   const viewingOwnProfile = !!$c.user && $c.user.id === user.id;
+  const adminViewing = Boolean($c.user) && isAccountAdmin($c.user);
```
mwiencek, line 864: Only commenting since I noticed the previous lines uses `!!` already. :)  `!= null` might be clearer than either since that's really what we're guarding against?
```suggestion
  const viewingOwnProfile = $c.user != null && $c.user.id === user.id;
  const adminViewing = $c.user != null && isAccountAdmin($c.user);
```

mwiencek (approved):

PR #2409, opened 2022-02-04
Upgrade to Flow 0.171.0

mwiencek:
In some class-based components, we use a technique where we bind certain methods to `this` in the constructor so that we can pass them as event handlers directly.

Flow doesn't like how we're doing this anymore.  For one, we previously had to declare the method signature again to be able to assign it in the constructor; Flow now sees this as a duplicate member declaration and errors.  But even once that's fixed, it also doesn't like `this.foo.bind(this)` anymore where `this.foo` is a method, because it views it as unbinding it from the class context (which isn't really true).

We could use `$FlowIssue` comments for these, but it's easier to just change them to arrow functions, and give a different name to the bound functions (I appended `Bound` to all of the names).

Some of these bound functions aren't or weren't even needed; I removed a few, but others could just use arrow functions in `render`.  There are still cases where we'd want to keep the same function reference though: when passing to `removeEventListener` and when passing to a `React.memo` component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "eslint-plugin-react": "7.20.6",
     "eslint-plugin-react-hooks": "4.1.2",
     "file-url": "2.0.2",
-    "flow-bin": "0.169.0",
+    "flow-bin": "0.171.0",
     "gettext-parser": "3.1.0",
     "http-proxy": "1.18.1",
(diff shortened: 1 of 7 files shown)


reosarevok (approved):
LGTMBDNT. How's the test coverage for the files changed here? :)

PR #2410, opened 2022-02-06
Remove MooseX::ABC

mwiencek:
According to the documentation at https://metacpan.org/pod/MooseX::ABC,

> This module is almost certainly a bad idea. You really want to just be using a role instead!

In most cases we can't easily convert ABC packages into roles, because they extend `MusicBrainz::Server:Edit`, which would also have to become a role.  But then we run into issues where required methods are provided by adjacent roles (triggering compile-time errors), and where attributes can no longer be extended with `+` (common for `data`), which is not possible in a role.

So instead, in most cases I've replaced `required` method lists with stubs that die.  However this means that it's no longer checked that these methods are implemented by the non-abstract classes at compile-time, only at run-time.

This patch incidentally fixes the following warning on plackup start:

> Argument "Moose::Meta::Method=HASH(0x556ed49cc448)" isn't numeric in numeric eq (==) at /home/michael/perl5/lib/perl5/MooseX/ABC/Trait/Class.pm line 61.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -63,7 +63,6 @@ requires 'Method::Signatures::Simple'                 => '1.07';
 requires 'MIME::Base64'                               => '3.13';
 requires 'Module::Pluggable'                          => '5.1';
 requires 'Moose'                                      => '2.1600';
-requires 'MooseX::ABC'                                => '0.06';
 requires 'MooseX::Clone'                              => '0.05';
 requires 'MooseX::Getopt'                             => '0.59';
 requires 'MooseX::MethodAttributes'                   => '0.29';
--- a/lib/MusicBrainz/Server/CoverArt.pm
+++ b/lib/MusicBrainz/Server/CoverArt.pm
@@ -13,7 +13,7 @@ has 'image_uri' => (
 );
 
 has 'provider' => (
-    isa => 'MusicBrainz::Server::CoverArt::Provider',
+    does => 'MusicBrainz::Server::CoverArt::Provider',
     is  => 'ro',
 );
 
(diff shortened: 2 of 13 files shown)


reosarevok:
The cover art changes seem to be failing circleci :)

yvanzo (changes requested):
It seems sane to drop a dependency which has such note and is unmaintained for 10 years!

> So instead, in most cases I've replaced `required` method lists with stubs that die.  However this means that it's no longer checked that these methods are implemented by the non-abstract classes at compile-time, only at run-time.

Having compile-time checks was the main reason for using ABC I guess.

After fixing Selenium tests, this lack of checks has to be documented.

> This patch incidentally fixes the following warning on plackup start: Argument ...

I’m so used to wait for this warning! :wink: Will have to look for `starting new work` now.

mwiencek:
Do you mean document it in the code with comments, or somewhere else?  Any suggestions on the best way/place to do that?

yvanzo:
Good question since it is about code that is not yet written! Even though that cannot replace compile-time checks, there might be two places for documenting this requirement: Comments in classes containing deadly stubs; `HACKING.md`.

reosarevok:
Is the lack of compile-time checks an issue if we ensure we have proper tests for all of these files?

mwiencek:
> Is the lack of compile-time checks an issue if we ensure we have proper tests for all of these files?

No, it's not an issue if we have proper test coverage (because the unimplemented methods would die once called).

But I updated the PR with some code comments.

mwiencek:
@brainzbot test this please

yvanzo (approved):
@brainzbot retest this please (even though the timeout error seems to be unrelated)

yvanzo:
@brainzbot retest this please

reosarevok:
@brainzbot retest this please

PR #2411, opened 2022-02-07
Test WS validation

reosarevok:
We were not testing this at all, AFAICT - if we test some bits somewhere it's certainly not in a consistent way.
This adds a set of tests for most of the validation checks made by the WS validator.

I could not figure out a way to test `validate_required` (thought `/ws/2/tag?id=1946a82a-f927-40c2-8235-38d64f50d043` without `entity` would do, and it might actually hit this, but because there's no actual error messaged linked, it looks the same as if I passed a random `a=b` parameter instead of `id`).

I'm also not sure how `validate_linked` works - I understand it's used for collections, is it what checks that you're not requesting areas for a release collection? If so, we already test that on `Controller::WS::2::Collection`.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Validator.pm
+++ b/lib/MusicBrainz/Server/WebService/Validator.pm
@@ -190,7 +190,7 @@ sub validate_inc
     my ($c, $version, $resource, $inc, $def) = @_;
 
     if (ref($inc)) {
-        $c->stash->{error} = 'Inc arguments must be combined with a space, but you provided multiple parameters';
+        $c->stash->{error} = 'Inc arguments must be combined with a space (+ or %20), but you provided multiple parameters';
         return;
     }
 
--- a/t/lib/t/MusicBrainz/Server/Controller/WS/2/Collection.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/WS/2/Collection.pm
@@ -323,9 +323,6 @@ test 'collection lookup errors' => sub {
     ws2_test_xml_invalid_mbid 'lookup 400s if mbid is invalid',
         '/collection/29611d8b-b3ad-4ffb-acb5-xxxxxxxxxxxx';
 
-    ws2_test_xml_invalid_mbid 'lookup 400s if mbid is invalid',
-        '/collection/29611d8b-b3ad-4ffb-acb5-xxxxxxxxxxxx';
-
     my $bad_entity_response = sub {
         my ($entity_type) = @_;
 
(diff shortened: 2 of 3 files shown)


lib/MusicBrainz/Server/WebService/Validator.pm
```diff
@@ -190,7 +190,7 @@ sub validate_inc
     my ($c, $version, $resource, $inc, $def) = @_;
 
     if (ref($inc)) {
-        $c->stash->{error} = 'Inc arguments must be combined with a space, but you provided multiple parameters';
+        $c->stash->{error} = 'Inc arguments must be combined with a space or plus sign, but you provided multiple parameters';
```
mwiencek, line 193: Nitpicking here:

Technically the `+` is a space: in the form-urlencoded media type, spaces are encoded as `+`.  You can also use `%20`, but both represent a space.

So this now reads like "must be combined with a space or encoded space" which is kinda confusing since it always has to be encoded -- as either `+` or `%20`.

We could instead say something like "must be combined with a space (i.e. + or %20)," giving both encodings as examples, if that also helps?

mwiencek:
> I'm also not sure how validate_linked works - I understand it's used for collections, is it what checks that you're not requesting areas for a release collection? If so, we already test that on Controller::WS::2::Collection.

It's used generally by all browse requests.  e.g. in `WS::2::Artist` we have
```
linked   => [ qw(area recording release release-group work collection) ],
```
which lists the query parameters you can browse by.  One of these is required for the request to be considered a browse request, so `validate_linked` is just checking that one of these is present in the query string (and returning the key/values).

lib/MusicBrainz/Server/WebService/Validator.pm
reosarevok: I guess I was thinking that the browsers automatically encode the spaces to `%20` before submission so you can just send an actual space, but I guess that's not true of all ways of hitting the WS maybe :) Added the specific wording instead.

reosarevok:
So I'm guessing the only way to test `validate_linked` in itself is to test the function directly rather than having a standard WS call? It seems it just returns a generic "The given parameters do not match any available query type" error otherwise, which doesn't seem useful for testing. Maybe I'm missing something though.

mwiencek:
Yeah, it's kind of tested indirectly by testing the WS generally; if browse requests work then `validate_linked` works.  We could test a badly-constructed browse query like `?artists=` (plural) and see if it returns "[t]he given parameters do not match," but it's going to be hard to test it directly.

reosarevok:
I guess the same is true of `validate_required` since that also just does `next`. I guess I'll add a test for each.

mwiencek (approved):

PR #2414, opened 2022-02-08
MBS-12195: Drop AcousticBrainz-related code

reosarevok:
### Implement MBS-12195

We display some AcousticBrainz data on recording sidebars and also a link on recording detail pages.
The project is being discontinued, although it is still going to run until 2023. That said, given the reason to discontinue it was specifically that the data isn't of the expected quality, there seems to be no reason to keep showing it for a year until it goes away.


yvanzo (approved):
LGTM :ship:

While double-checking the repo for `AcousticBrainz`/`AB` I noticed some occurrences the timeline events. It seems worth adding an event for the end of the project. :coffin: :hammer:

reosarevok:
Good idea, I'll add one (luckily I can add it from the site itself now)

PR #2416, opened 2022-02-11
cdtocs type fixes

mwiencek:
* cdtocs is outputted as an array of strings
 * cdtocs is not a count

It doesn't seem like either of these lines caused an issue in practice.


reosarevok (approved):
I'm not sure why we only pass the TOC string, but this seems correct as of now anyway.

PR #2419, opened 2022-02-16
MBS-11742: Also clean non-/user YouTube channels

reosarevok:
### Implement MBS-11742

YouTube has channels using /user, /channel, /c and no prefix. No idea *why*, but in any case while you can remove /c
you can't remove /channel, for example. So to be safe I just deal with them separately here. While for /channel, like for /user, it's probably fine to just remove anything at the end, for the ones with the username only it's harder to be confident on a wide regex, so just removing the main tabs from the YouTube channel page.


reosarevok:

PR #2420, opened 2022-02-16
MBS-10668: Add autoselect + sidebar for TikTok URLs

reosarevok:
### Implement MBS-10668

Limiting to artist, label, place and series for now as per discussion on STYLE-1993.
For allowed username characters see https://support.tiktok.com/en/getting-started/setting-up-your-profile/changing-your-username
There are pages for the recording (excerpts) that are used on videos, but not allowing linking to those now, mostly because I can't see a good relationship that would fit.


yvanzo (approved):
The commit message is explanatory and the code looks good too!

Why would the streaming relationship type to a recording not qualify? A reply to the [example](https://tickets.metabrainz.org/browse/STYLE-1993?focusedCommentId=61086#comment-61086) provided by the ticket’s reporter sammyrayy would be welcome. Even though Tiktok is supposed to be a favorite of our age group according to <https://influencermarketinghub.com/tiktok-video-size/>. I have no strong take about whether entering Tiktok videos as recordings should be forbidden or not. Checking the database, I found 1000+ Tiktok user pages for 10 videos only.

PR #2421, opened 2022-02-16
MBS-12138: Drop parens from Amazon Music pretty name

reosarevok:
### Implement MBS-12138

We don't use these in any of the other pretty names that include countries (Apple and Napster just do "Site CountryCode", and ASIN does CountryCode: ASIN). As such, it seems more consistent to drop them from here too.


reosarevok:

reosarevok:

yvanzo (approved):
:ship:

PR #2422, opened 2022-02-16
MBS-12201: Show any non-part-of rels on series page

reosarevok:
### Fix MBS-12201

We already filter out part-of rels, so there's no good reason not to show any entities here which have non part-of rels to series. Right now that seems to be RG and Work. Recording only has a relationship hidden by cardinality from the series page, but also preemptively allowing it.


yvanzo (approved):
Agreed. Thanks for the detailed rationale! :ship:

PR #2426, opened 2022-02-17
Improve test documentation (V): Controller::Instrument (adds multiple tests)

reosarevok:
This also adds a lot of tests because we had pretty much *nothing* of use here. I also expanded some analogous area tests before copying them for instruments.


yvanzo (approved):
:ship:

PR #2429, opened 2022-02-22
MBS-12213: Allow VGMDb artist links on places

reosarevok:
### Implement MBS-12213

VGMDb uses artist pages for studios, so we need to support them on place entities. I added a new relationship type for it,
and now unblocking it in the code too. There's a suggestion that org pages might also be used for studios but I didn't find any, so not unblocking those just yet.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -4573,19 +4574,24 @@ const CLEANUPS: CleanupEntries = {
               result: prefix === 'artist' || prefix === 'org',
               target: ERROR_TARGETS.ENTITY,
             };
-          case LINK_TYPES.vgmdb.release:
+          case LINK_TYPES.vgmdb.event:
             return {
-              result: prefix === 'album',
+              result: prefix === 'event',
```
reosarevok, line 4579: Fixed the order to be alphabetical per entity type, maybe I should have made this a second commit, but hopefully it's not too annoying :)

yvanzo: Hopefully it's not too annoying to split this commit either.

reosarevok: Fiiiiiine, I'll reorder separately :)

yvanzo:
Thank you! :ship:

PR #2430, opened 2022-02-22
MBS-12214: Allow BBC /events links as otherDBs for events

reosarevok:
### Implement MBS-12214

I'm also adding the favicon back that we removed when we dropped BBC Music links, but I'm not displaying these on the sidebar of the event for now since I'm not sure it's needed. The favicon will still be displayed when adding / editing the URL.


yvanzo (approved):
It could be split into two commits as well.

reosarevok:
What would be the second one? (we normally have only one page for cleanup and favicons)

yvanzo:
> we normally have only one page for cleanup and favicons

Not really, it even matches two components in tickets: URL cleanup and Data display.
It's often mixed when adding new entries, not when modifying existing ones.

reosarevok:
Well, this does not really modify existing ones, since it's a new link that we never supported before, but it's just a BBC link so we just bring back the old favicon :) But I don't mind splitting it into two, so sure, will do.

reosarevok:

PR #2431, opened 2022-02-22
MBS-12215 / MBS-12216: Load annotation entity when possible on search results

reosarevok:
### Fix MBS-12215 / MBS-12216

This was just creating a generic entity based on indexed data, but that means we lack a lot of useful info, such as sort names for artists. As such, this now loads the entity if possible and only shows the mock version if we cannot find more.

This also effectively resolves MBS-12216, an old search ticket that called for the entity disambiguation being displayed.

Not sure if results for deleted entities should actually be returned (that's a separate open search ticket) but I guess that even then, there's likely to be a few cases where the index hasn't caught up with an entity removal yet.


yvanzo (approved):
LGTMBDNT :ship:

PR #2432, opened 2022-02-24
Implement "not in my subscriptions" predicate

reosarevok:
Somehow I managed to miss that I hadn't actually implemented the negative predicate here.
An entity is not in subscriptions only if every single one of these predicates is false.


yvanzo (approved):
Tested too and works but requires additional search conditions not to timeout. Maybe we should not wait for timeout and warn when selecting this condition or even require at least two conditions. Just a note for later on.

:ship:

reosarevok:
We have a significant amount of predicates that time out on their own :/ I'm afraid edit search needs either a lot more indexes or a rethink, but this seems better than nothing for now.

PR #2433, opened 2022-02-25
Fixes/improvements to the MBS-12141 upgrade script

mwiencek:
* We need to explicitly delete empty_tag_ids from the tag table in case
   they were already unreferenced before the script ran (in which case
   delete_unused_tag won't trigger).

 * We now handle cases where a cleaned-up tag name already exists.  This
   code path likely won't be hit in the production upgrade, because only
   the two blank/empty tags that would be deleted via above would
   otherwise hit it.  But the schema upgrade should generally work no
   matter what the database contents are; this includes standalone
   databases which may have unclean data.  And there is a small chance
   that a bug in musicbrainz-server will allow someone to tag " rock" or
   "rock " in the meantime.

 * Existing control_for_whitespace and only_non_empty constraints are
   now dropped first, since we'll be running this script in production
   before schema change day and don't want the second run to fail.

### DIFF
--- a/admin/sql/updates/20211216-mbs-12140-12141.sql
+++ b/admin/sql/updates/20211216-mbs-12140-12141.sql
@@ -2,14 +2,25 @@
 
 BEGIN;
 
+-- NOTE: Make sure this script runs *before* any that recalculates
+-- count/ref_count for the schema change.
+
 DO $$
 DECLARE
   empty_tag_ids INTEGER[];
+  -- An "uncontrolled for whitespace" tag.
+  ufw_tag RECORD;
+  -- An existing "controlled for whitespace" tag ID that would conflict with
+  -- ufw_tag if it were cleaned.
+  existing_cfw_tag_id INTEGER;
+  tag_cursor REFCURSOR;
 BEGIN
   SELECT array_agg(id)
     FROM tag
    WHERE name ~ E'^\\s*$'
     INTO empty_tag_ids;
+
+  RAISE NOTICE 'Deleting empty tag IDs: %', empty_tag_ids;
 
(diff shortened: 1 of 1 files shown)


admin/sql/updates/20211216-mbs-12140-12141.sql
```diff
@@ -30,16 +41,381 @@ BEGIN
...
+
+    IF existing_cfw_tag_id IS NULL THEN
+      UPDATE tag
+         SET name = regexp_replace(btrim(name), E'\\s{2,}', ' ', 'g')
+       WHERE id = ufw_tag.id;
+    ELSE
+      RAISE NOTICE 'Conflicting tag with controlled whitespace found: id=%',
+        existing_cfw_tag_id;
+
+      OPEN tag_cursor NO SCROLL FOR SELECT * FROM area_tag_raw WHERE tag = ufw_tag.id FOR UPDATE;
```
mwiencek, line 81: I tried to find a way to templatize this block and simply loop over the entity type names, but in the end it was easier to just copy and paste it for each type.

 * You can't use `WHERE CURRENT OF` inside `EXECUTE`.
 * The `tag_cursor.ctid` doesn't match the underlying row's `ctid`, so can't use that either.
 * We can finally use the primary keys to identify the current row.  You can't cast a record type to a table type directly, so something like `($1).%I` where `$1` is the current cursor record and `%I` is e.g. "release_group" doesn't work.  There is a workaround where you can cast to `::TEXT` first.  But the primary keys of the *_tag and *_tag_raw tables still differ, so ultimately the templatized code would be hard to read.

reosarevok (approved):
LGTMBDNT

mwiencek:
This has been run in production (since it's preventing people from importing database dumps).

PR #2435, opened 2022-03-01
MBS-12226: Add allowed prefix of niconico video ID

rinsuki:
# Problem

See https://tickets.metabrainz.org/browse/MBS-12226

# Solution

Add allowed prefix of niconico video ID to regular expression.

currently, video that have a these prefix can be watch.

- sm
- so
- nm
- ax
- ca
- cw
- cw
- nl
- /z[a-d]/

other prefix also exist, but they are not available to watch currently.
(mostly due to channel that owns video are closed)

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -3221,11 +3221,11 @@ const CLEANUPS: CleanupEntries = {
     restrict: [{...LINK_TYPES.streamingfree, ...LINK_TYPES.videochannel}],
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?ch\.nicovideo\.jp\/([^\/]+).*$/, 'https://ch.nicovideo.jp/$1');
-      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?nicovideo\.jp\/(user\/[0-9]+|watch\/sm[0-9]+).*$/, 'https://www.nicovideo.jp/$1');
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
This seems sensible. I kinda wonder if it wouldn't even be easier to allow any `[a-z]{2}`, but it's an improvement anyway, AFAICT.

reosarevok:
@brainzbot, add to whitelist

reosarevok:

yvanzo (approved):
Thanks @rinsuki for your patch!

Don’t mind excluding closed video channels, URL relationships can be set as _ended_ if needed. If there is no other reason to use a limited list of prefixes, please consider using `[a-z]{2}` instead.

We will be happy to merge this PR regardless of your choice. Looking forward for your answer.

rinsuki:
Thanks for review!

there is a another reasons to use limited lists:

- niconico using same format /[a-z]{2}[0-9]+/ on many other services
  - like live broadcast https://live.nicovideo.jp/watch/lv335776540
  - news https://news.nicovideo.jp/watch/nw10603210
  - quiz https://q.nicovideo.jp/watch/nq28662
  - I think it would be good to allow only expected things.
- it seems most dead prefix aren't containing contents that suitable for links with MusicBrainz
  - (estimated by channel purpose)
- If someone want to add excluded prefix videos, we can add after hear their request.
  - (but I think it is really rare case, because they are only used in 2007~2009? videos and already deleted)

reosarevok:
Ok! Merging as it is for now, we can always make changes later if we see a need :)

PR #2437, opened 2022-03-01
MBS-12225: Change terminology from "slave" to "mirror"

reosarevok:
### Implement MBS-12225


admin/cron/mirror.sh
```diff
@@ -3,16 +3,16 @@
 MB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../../" && pwd)
```
mwiencek, line 3: This file will be referenced from people's crontabs, so we should probably include a symbolic link from slave.sh to mirror.sh in the repository.

admin/sql/updates/20220302-mbs-12225-mirror_only.sql
```diff
@@ -0,0 +1,118 @@
...
+ALTER TRIGGER apply_artist_release_group_pending_updates_slave ON release_group_meta RENAME TO apply_artist_release_group_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_group_pending_updates_slave ON release_group_secondary_type_join RENAME TO apply_artist_release_group_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_group_pending_updates_slave ON track RENAME TO apply_artist_release_group_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_pending_updates_slave ON release RENAME TO apply_artist_release_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_pending_updates_slave ON release_country RENAME TO apply_artist_release_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_pending_updates_slave ON release_first_release_date RENAME TO apply_artist_release_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_pending_updates_slave ON release_label RENAME TO apply_artist_release_pending_updates_mirror;
+ALTER TRIGGER apply_artist_release_pending_updates_slave ON track RENAME TO apply_artist_release_pending_updates_mirror;
+
+DROP TRIGGER IF EXISTS a_ins_release_slave ON release;
```
mwiencek, line 16: You can just rename these instead of dropping and re-creating, because with triggers PG references the functions they call by oid, not name.  Drop & re-create is only really needed for functions that call renamed functions, since the definitions are stored as text.

admin/cron/mirror.sh
reosarevok: I *think* I did that right?

admin/sql/updates/20220302-mbs-12225-mirror_only.sql
reosarevok: Oh, TIL. Amended.

admin/sql/updates/schema-change/20.mirror.sql
```diff
@@ -10,8 +10,10 @@
 -- 20140318-series.sql
 -- 20140418-series-instrument-functions.sql
 -- 20140429-area-view.sql
-\set ON_ERROR_STOP 1
+\\set ON_ERROR_STOP 1
```
mwiencek, line 13: Huh, how come extra slashes are being added here all of a sudden?  That's not valid syntax.

admin/sql/updates/schema-change/27.mirror_only.sql
```diff
@@ -0,0 +1,124 @@
...
+ALTER FUNCTION a_del_release_group_secondary_type_join_slave() RENAME TO a_del_release_group_secondary_type_join_mirror;
+ALTER FUNCTION a_ins_release_label_slave() RENAME TO a_ins_release_label_mirror;
+ALTER FUNCTION a_upd_release_label_slave() RENAME TO a_upd_release_label_mirror;
+ALTER FUNCTION a_del_release_label_slave() RENAME TO a_del_release_label_mirror;
+ALTER FUNCTION a_ins_track_slave() RENAME TO a_ins_track_mirror;
+ALTER FUNCTION a_upd_track_slave() RENAME TO a_upd_track_mirror;
+ALTER FUNCTION a_del_track_slave() RENAME TO a_del_track_mirror;
+
+
+CREATE TRIGGER a_ins_release_mirror AFTER INSERT ON release
```
mwiencek, line 61: This needs to be regenerated for the updates you made (but also regardless given the invalid `\\set`).

admin/sql/updates/schema-change/20.mirror.sql
mwiencek: looks like 12104a19eccae34b66910b36ecb8c723be92252e changed `<<EOSQL` to `<<'EOSQL'` which broke the escape sequence.  should be an easy fix.

mwiencek (approved):

mwiencek:
Some of the changes under admin/sql/updates/schema-change/ are pointless (regenerating old, unused scripts to set `search_path` and `statement_timeout`, which is unrelated to the slave -> mirror changes) but are fairly minimal.

I'm going to merge this now to avoid some nasty conflicts in current/future PRs.

mwiencek:
I made a few additional changes related to this -- please review when you get a chance :) https://github.com/metabrainz/musicbrainz-server/compare/ff203d9a20e1a02df1200174770d7edd1396a101...321eef80d417b0a10f220c4f4e86504e0d697e1c

reosarevok:
Looks good to me, thanks!!!

I thought I only did regenerate historical upgrades in case they were named slave but maybe not, sorry if I missed some :) The point was just to ensure someone updating ancient servers would still be able to do it.

PR #2438, opened 2022-03-02
Explain how to debug circleci tests with SSH

reosarevok:
This is not secret, but it only applies to our prod setup, so adding it to HACKING-PROD.

Feel free to suggest any improvements if my explanations are mediocre, they're mostly based on an old convo with @mwiencek  which didn't go too much in depth :)


mwiencek (approved):
LGTM.  Thanks!

PR #2441, opened 2022-03-04
Improve test documentation (VII): Shared tests (Aliases)

reosarevok:
I'm changing a lot of tests that we already had for several entities, because it seems a lot easier to see all the updates to them in one commit each rather than slowly doing them per entity.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Controller/Label/AddAlias.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/Label/AddAlias.pm
@@ -5,69 +5,122 @@ use MusicBrainz::Server::Test qw( capture_edits html_ok );
 
 with 't::Mechanize', 't::Context';
 
-test all => sub {
+=head2 Test description
 
-my $test = shift;
-my $mech = $test->mech;
-my $c    = $test->c;
+This test checks whether alias adding for labels works, including whether
+the sort name defaults to the name when not explicitly entered.
 
-MusicBrainz::Server::Test->prepare_test_database($c);
+=cut
 
-$mech->get_ok('/login');
-$mech->submit_form( with_fields => { username => 'new_editor', password => 'password' } );
+test 'Adding alias with sort name' => sub {
+    my $test = shift;
+    my $mech = $test->mech;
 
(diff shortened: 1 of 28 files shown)


t/lib/t/MusicBrainz/Server/Controller/Label/AddAlias.pm
```diff
@@ -5,69 +5,122 @@ use MusicBrainz::Server::Test qw( capture_edits html_ok );
...
-isa_ok($edit, 'MusicBrainz::Server::Edit::Label::AddAlias');
-is($edit->data->{sort_name}, 'Another alias', 'sort_name defaults to name');
+=head2 Test description
 
+This test checks whether alias adding for labels works, including whether
+the sort name defaults to the name when not explicitly entered.
+
+=cut
+
+test 'Test adding alias with sort name' => sub {
```
mwiencek, line 15: I would omit 'Test' from the test descriptions, since it's redundant.
```suggestion
test 'Adding alias with sort name' => sub {
```

t/lib/t/MusicBrainz/Server/Controller/Label/AddAlias.pm
```diff
@@ -5,69 +5,122 @@ use MusicBrainz::Server::Test qw( capture_edits html_ok );
...
+        'Edit page contains artist link',
+    );
+    $mech->content_contains('An alias', 'Edit page contains alias name');
+    $mech->content_contains(
+        'An alias sort name',
+        'Edit page contains the selected alias sort name',
+    );
 };
 
+test 'MBS-6896: Test adding alias without sort name' => sub {
```
mwiencek, line 83: For test descriptions with ticket numbers, I think they should describe what the issue or desired behavior is.  For example, let's say we wanted the page to throw an error if you don't give a sort name.  This description is a bit too vague because, since it doesn't say what should happen, it could apply to either situation.

```suggestion
test 'MBS-6896: Adding alias without sort name defaults it to the name' => sub {
```

mwiencek (approved):
Overall this looks like a massive improvement, though I would write some of the descriptions slightly differently

PR #2442, opened 2022-03-04
MBS-12231: Don't try to load last annotation if there's no entity

reosarevok:
### Fix MBS-12231

Previews are allowed to have no entity, but we were unconditionally trying to call for its id, which unsurprisingly fails.


yvanzo (approved):
LGTMBDNT :ship:

PR #2443, opened 2022-03-08
Add DataStore::RedisMulti

mwiencek:
If the `DataStore::RedisMulti` module is active, then `DATASTORE_REDIS_ARGS` may return an array ref of connection details. (How do you know if it's active? grep for `DataStore::RedisMulti->new`. We may revert back to plain-old `DataStore::Redis` if multiple instances aren't needed.)

This module is useful when Redis service needs to be migrated to a new server. We'll attempt to read from each connection in order (returning the first non-empty result), and also distribute writes to all connections. This allows time to copy any keys that don't exist on the new instance from the old instance.

### DIFF
--- a/t/lib/t/MusicBrainz/DataStore/RedisMulti.pm
+++ b/t/lib/t/MusicBrainz/DataStore/RedisMulti.pm
@@ -1,0 +1,127 @@
+package t::MusicBrainz::DataStore::RedisMulti;
+
+use utf8;
+
+use Test::Routine;
+use Test::Moose;
+use Test::More;
+use Test::Deep qw( cmp_bag );
+use MusicBrainz::Server::Test;
+use MusicBrainz::DataStore::Redis;
+use MusicBrainz::DataStore::RedisMulti;
+use DBDefs;
+
+=head2 Test description
+
+This test checks basic tasks for the RedisMulti store (including adding,
+deleting and expiring keys), which distributes queries to multiple underlying
+stores. Consistency among all the stores is checked after each operation.
+
+=cut
+
+# Initialize tests
+my $args1 = DBDefs->DATASTORE_REDIS_ARGS;
+$args1->{database} = $args1->{test_database};
+
+my $args2 = DBDefs->DATASTORE_REDIS_ARGS;
+$args2->{database} = $args2->{test_database} + 1;
+
+my $redis1 = MusicBrainz::DataStore::Redis->new($args1);
+my $redis2 = MusicBrainz::DataStore::Redis->new($args2);
+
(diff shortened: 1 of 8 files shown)


mwiencek:
Outside of our current type of ordeal (migrating one standalone Redis instance to another server), this likely won't be as useful if we have a multi-master Redis/KeyDB cluster in the future.  But, it could be used for migrating to such a setup more easily.  Or perhaps we may need to migrate from one cluster to another for some reason?

yvanzo (approved):
Indeed, having a more adaptive server for deployment is useful to help with any future maintenance operations (switching to HAProxy/KeyDB, redeploying in emergency...).

LGTMBDNT. The added tests have a good coverage.

t/lib/t/MusicBrainz/DataStore/RedisMulti.pm
```diff
@@ -0,0 +1,127 @@
...
+    is($redis2->exists('does-not-exist'), 0, 'exists returns 0 for non-existent key in database 2');
+
+    is($redis_multi->exists('string'), 1, 'exists returns 1 for existing key');
+    is($redis1->exists('string'), 1, 'exists returns 1 for existing key in database 1');
+    is($redis2->exists('string'), 1, 'exists returns 1 for existing key in database 2');
+
+    $redis_multi->delete('string');
+    is($redis_multi->exists('string'), 0, 'exists returns 0 for deleted key');
+    is($redis1->exists('string'), 0, 'exists returns 0 for deleted key in database 1');
+    is($redis2->exists('string'), 0, 'exists returns 0 for deleted key in database 2');
```
reosarevok, line 76: What should happen if you delete from `$redis1` but not `$redis2`? Should we have a test for that case?

mwiencek: It'd only be deleted from `$redis1`,  but any servers with `RedisMulti` enabled would still see the key if it exists in `$redis2`.

This shouldn't really happen in practice, because we'd be deploying the `RedisMulti` configuration on all servers at once.  But there is a minor race condition where a key is set on a server with a `RedisMulti` configuration, and then deleted on a server that doesn't have the new configuration yet (because we're in the middle of deploying it).  The only situation I can see this happening is if a user logs in and then logs out; this would have to be perfectly timed and forwarded to the right servers.  In that case there may be a stale session key on the new Redis instance, but it would expire within some hours anyway.

Test-wise we do have the "string-1-only"/"string-2-only" ones here, which check that keys set in only one of the servers will be retrieved through `$redis_multi` (doesn't matter if they're deleted first).

PR #2444, opened 2022-03-09
MBS-12241: Drop the whitespace_collapsed database constraint

reosarevok:
### Implement MBS-12241

While we don't usually want titles and names to have blocks of more than one space, that can be useful in cases where the multiple spaces are in fact artist intent. Additionally, this constraint is currently broken by several entries in the DB.
For now, this is still enforced by Perl. Further changes will be needed there if we want to allow it in some cases (e.g. track titles), and we probably want to keep blocking it in others (e.g. tag names).


mwiencek (approved):
:shipit:

PR #2447, opened 2022-03-10
MBS-12243: Block smart links: found.ee

reosarevok:
### Implement MBS-12243

These can be shortlinks or geolinks, but in either case we should not link to them: https://blog.found.ee/en/articles/4202319-make-a-shortlink-or-a-geolink


yvanzo (approved):
LGTMBDNT :ship:

PR #2448, opened 2022-03-10
MBS-12244: Block Amazon customer video URLs

reosarevok:
### Implement MBS-12244

These were happily getting matched as ASIN, but that does not seem correct at all. AFAICT, these shouldn't be added to MB at all anyway, so this blocks them and gives the user a hopefully useful message.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -614,7 +614,7 @@ const CLEANUPS: CleanupEntries = {
         return 'https://www.amazon.' + tld + '/gp/product/' + asin;
       }
 
-      return '';
+      return url;
     },
     validate: function (url) {
       if (/amzn\.to\//i.test(url)) {
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -635,6 +635,17 @@ const CLEANUPS: CleanupEntries = {
         };
       }
 
+      if (/^https:\/\/www\.amazon\.[^\/]+\/vdp\//i.test(url)) {
+        return {
+          error: l(
+            `This is a link to a user video and should not be added. Please
+             add the product link instead, if relevant.`,
+          ),
+          result: false,
+          target: ERROR_TARGETS.URL,
+        };
+      }
+
       // If you change this, please update the BadAmazonURLs report.
       return {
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -635,6 +635,17 @@ const CLEANUPS: CleanupEntries = {
         };
       }
 
+      if (/amazon\.com\/vdp\//i.test(url)) {
```
yvanzo, line 638: To match video samples that exist in other Amazon stores too:

```suggestion
      if (/^https:\/\/www\.amazon\.[^\/]+\/vdp\//i.test(url)) {
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -431,6 +431,17 @@ const testData = [
...
+                     input_url: 'https://www.amazon.com/vdp/08c6c18fc7bb4822a166db4834e123f1?ref=dp_vse_rvc_0',
+             input_entity_type: 'release',
+       input_relationship_type: 'amazon',
+    expected_relationship_type: undefined,
+       only_valid_entity_types: [],
+                expected_error: {
+                                  error: 'link to a user video',
+                                  target: 'url',
+                                },
+  },
```
yvanzo, line 444: I guess Amazon tests were missing `expected_clean_url` because it was returning an empty string in some case. It can now be tested that unwanted URLs are untouched:

```suggestion
  {
                     input_url: 'https://www.amazon.com/vdp/08c6c18fc7bb4822a166db4834e123f1?ref=dp_vse_rvc_0',
             input_entity_type: 'release',
       input_relationship_type: 'amazon',
    expected_relationship_type: undefined,
            expected_clean_url: 'https://www.amazon.com/vdp/08c6c18fc7bb4822a166db4834e123f1?ref=dp_vse_rvc_0',
       only_valid_entity_types: [],
                expected_error: {
                                  error: 'link to a user video',
                                  target: 'url',
                                },
  },
```

root/static/scripts/edit/URLCleanup.js
reosarevok: Oh, duh, fair point :)

yvanzo (approved):
:shipit:

PR #2449, opened 2022-03-11
Improve test documentation (VIII): Shared tests (Tags/Ratings)

reosarevok:
I'm changing a lot of tests that we already had for several entities, because it seems a lot easier to see all the updates to them in one commit each rather than slowly doing them per entity.


mwiencek (commented):
My comments from https://github.com/metabrainz/musicbrainz-server/pull/2441 would apply here too (but I'll stop repeating it).

yvanzo (approved):
LGTM :arrow_upper_right:

mwiencek (approved):

PR #2454, opened 2022-03-14
MBS-12252 / MBS-12253 / MBS-12254 / MBS-12255: Genre-related schema additions for consistency

reosarevok:
### Implements MBS-12252 / MBS-12253 / MBS-12254 / MBS-12255

See commit messages for details. This makes genre more consistent with other entities - most of these changes have been requested anyway by the community already, as well.

I am not yet adding genre_attribute_type since those tables are unused and we have no current use case for genre either, but I can add it if @yvanzo feels it would make his work easier in the future :)


admin/sql/CreateTables.sql
```diff
@@ -953,15 +953,56 @@ CREATE TABLE genre ( -- replicate (verbose)
...
+    gid                 uuid NOT NULL
+);
+
 CREATE TABLE genre_alias ( -- replicate (verbose)
-    id                  SERIAL,
+    id                  SERIAL, --PK
     genre               INTEGER NOT NULL, -- references genre.id
     name                VARCHAR NOT NULL,
     locale              TEXT,
-    edits_pending       INTEGER NOT NULL DEFAULT 0 CHECK (edits_pending >= 0),
```
mwiencek, line 955: micro-nit, but we can probably keep the space after `=` if that's the only change on this line (I realize this was probably copied from another entity)

admin/sql/updates/20220314-mbs-12255.sql
```diff
@@ -0,0 +1,85 @@
...
+);
+
+ALTER TABLE genre_alias ADD CONSTRAINT genre_alias_pkey PRIMARY KEY (id);
+
+CREATE INDEX genre_alias_idx_genre ON genre_alias (genre);
+CREATE UNIQUE INDEX genre_alias_idx_primary ON genre_alias (genre, locale) WHERE primary_for_locale = TRUE AND locale IS NOT NULL;
+
+INSERT INTO genre_alias (id, genre, name, locale, edits_pending, type, sort_name)
+SELECT id, genre, name, locale, edits_pending, 1, name -- sortname = name, type = genre name
+FROM tmp_genre_alias;
```
mwiencek, line 83: This can also copy `last_updated` from the old table, just to be complete.

mwiencek (commented):
LGTM overall on first review.

yvanzo (commented):
No need to add a table if it is unused, which is why there is no `url_attribute_type` table either.

admin/sql/updates/20220314-mbs-12255.sql
reosarevok: I was thinking that since a type is added by default then the time of last update would be NOW() but on second thought maybe your way is better, will do.

PR #2455, opened 2022-03-15
Use release_coverart_backup

mwiencek:
Columns in the release_coverart table have been getting cleared since eb9637f469e18a236dd35f37651e22b48423731a was deployed.  I'm not sure exactly how this is happening, but likely through the RebuildCoverArt script, which I thought was disabled.

Unfortunately, this table is only in the private dump (where none is available prior to 1/31), but I managed to find an older database dump from May 2021 with a snapshot of the table.  For now I've imported that into release_coverart_backup so that we can get the reports working again.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Release.pm
+++ b/lib/MusicBrainz/Server/Data/Release.pm
@@ -1436,6 +1436,7 @@ sub load_ids
     }
 }
 
+my $has_release_coverart_backup = undef;
 sub load_meta
 {
     my $self = shift;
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Data/Release.pm
```diff
@@ -1451,16 +1452,24 @@ sub load_meta
...
+            q(SELECT 1 FROM pg_tables WHERE schemaname = 'musicbrainz' AND tablename = 'release_coverart_backup'),
+        ) ? 1 : 0;
+    }
+
+    if ($has_release_coverart_backup) {
+        my @ids = keys %id_to_obj;
+        if (@ids) {
+            for my $row (@{
+                $self->sql->select_list_of_hashes(
+                    'SELECT * FROM release_coverart_backup WHERE id IN ('.placeholders(@ids).')',
```
mwiencek, line 1466: The change in this file is to allow the saved Amazon artwork to display on release pages.  I check if the release_coverart_backup table exists first, since it only exists in production.  There's no point falling back to release_coverart because it's a private table.

lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -467,12 +467,12 @@ my %stats = (
             my ($self, $sql) = @_;
 
             my $data = $sql->select_list_of_lists(<<~'SQL');
-                SELECT (cover_art_url ~ '^https?://.*.images-amazon.com')::int AS is_amazon, COUNT(*) FROM release_coverart
+                SELECT (cover_art_url ~ '^https?://.*.images-amazon.com')::int AS is_amazon, COUNT(*) FROM release_coverart_backup
```
mwiencek, line 470: I don't check if the table exists here or in the report below, since these don't really need to run outside of production.

reosarevok (approved):
LGTMBDNT

yvanzo (approved):
Not better than @reosarevok.

For the record, @mwiencek tested it with and without the release coverart backup.

:ship:

PR #2456, opened 2022-03-16
Improve test documentation (XI): Shared tests (Edit)

reosarevok:
I'm changing a lot of tests that we already had for several entities, because it seems a lot easier to see all the updates to them in one go each rather than slowly doing them per entity.


yvanzo (approved):
:ship:

PR #2457, opened 2022-03-16
MBS-12251: Sort primary alias before others for same locale

reosarevok:
### Implement MBS-12251

The primary for locale alias should by definition be the main alias for a specific locale. As such, it makes sense that once grouped by locale, the next alias sorting will be so that the primary alias sorts first in that group.


yvanzo (approved):
Tested with `artist/58a51d12-1e22-48f3-a75c-005791fb9b04/aliases`.

PR #2458, opened 2022-03-16
MBS-12258: Skip updating up-to-date containers

yvanzo:
# MBS-12258

Skip updating containers which are running the latest built image for this service already.  This change only helps with resuming an update, by fastening further runs of the script after it has been interrupted for some reason.

### DIFF
--- a/script/update_containers.sh
+++ b/script/update_containers.sh
@@ -44,6 +44,7 @@ then
 fi
 
 DEPLOY_ENV=$1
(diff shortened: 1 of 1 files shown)


reosarevok (approved):
Tested this locally and it worked fine (for some reason any sshing into zappa took forever for me, but that seems to be unrelated to this script anyway).

mwiencek (approved):

PR #2460, opened 2022-03-17
MBS-12263: Only display "VIAF: " on VIAF pretty links when on sidebar

reosarevok:
### Implement MBS-12263

This was always showing as `VIAF: $id`, which makes sense on the sidebar but looks a bit ridiculous elsewhere when the full display is `VIAF ID: VIAF: $id`. Changing it to only add `VIAF: ` for the sidebar, as `URL::Wikidata` already does.


yvanzo (approved):
:ship:

PR #2463, opened 2022-03-21
MBS-12272: Split unicode text properly in DiffSide

mwiencek:
Splitting a string by `''` to get an array of characters doesn't work here, because it splits the string by UTF-16 code units, not characters. Adding the `u` flag to our split `RegExp` tells it to handle 4-byte characters properly.


MonkeyDo:
I learned something useful today!

reosarevok (approved):
Works nicely. I guess we don't currently have any tests for diff creation?

PR #2466, opened 2022-03-23
MBS-11879: Fail better when trying to add deleted entity to collection

reosarevok:
### Fix MBS-11879

I don't think there's a point creating a specific `NotFound` template for this; it's a fairly rare situation, and I think giving the user an understandable error message is good enough here.

Screenshot from 2022-03-23 12-11-36

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Collection.pm
+++ b/lib/MusicBrainz/Server/Controller/Collection.pm
@@ -1,7 +1,9 @@
 package MusicBrainz::Server::Controller::Collection;
 use Moose;
+use utf8;
 use Scalar::Util qw( looks_like_number );
 use List::AllUtils qw( first uniq );
+use MusicBrainz::Server::Translation qw( l );
 
 BEGIN { extends 'MusicBrainz::Server::Controller' };
 
@@ -69,6 +71,11 @@ sub _do_add_or_remove {
 
     if ($entity_id) {
         my $entity = $c->model(type_to_model($entity_type))->get_by_id($entity_id);
+
+        unless ($entity) {
+            $c->stash->{message} = l('“{id}” is not a valid row ID; the entity might have been merged or deleted.', { id => $entity_id });
+            $c->detach('/error_400');
+        }
 
         $c->model('Collection')->$func_name($entity_type, $collection->id, $entity_id);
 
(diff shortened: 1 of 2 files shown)


reosarevok:
I want to add a test for this, but the main place where we are currently testing stuff like this is `Data::Collection` and this is done at the controller level. Should we add a new `Controller::Collection::AddAndRemove` test and also move a bunch of the stuff that is now on `Data::Collection` to actually use add/remove links and make sure it works properly all the way rather than just on a direct `add_entities_to_collection` call?

mwiencek:
Yeah, adding `Controller::Collection::AddAndRemove` makes sense to me.  Not sure about moving/removing the existing `Data::Collection` tests though, since those `add_entities_to_collection` are prerequisites for the following tests for `merge_entities`, `contains_entity`, etc.

lib/MusicBrainz/Server/Controller/Collection.pm
```diff
@@ -70,6 +70,11 @@ sub _do_add_or_remove {
     if ($entity_id) {
         my $entity = $c->model(type_to_model($entity_type))->get_by_id($entity_id);
 
+        unless ($entity) {
+            $c->stash->{message} = 'This ID does not match any entity; the entity might have been merged or deleted';
```
yvanzo, line 74: It should use the same message (either the current message or the new one) for [`invalid_mbid`](https://github.com/metabrainz/musicbrainz-server/blob/v-2022-03-28/lib/MusicBrainz/Server/Controller.pm#L33) or be more specific; for example: _No entity has been added to the collection, because this ID does not match any entity._ (Removing an already deleted entity from a collection can probably be silent.)

yvanzo (approved):
LGTMBDNT. Suggested to unify with `invalid_mbid` but this is an improvement already.

lib/MusicBrainz/Server/Controller/Collection.pm
reosarevok: I don't think a delete should be silent, because the user might have just incorrectly entered the ID and want to be notified (for example if they're doing this through a bot or something).

This is a row id, not an MBID, which is why I didn't reuse the invalid MBID message, but we could be more precise with the message :)

mwiencek: I wouldn't mind if it was phrased similarly to `invalid_mbid`, like `'{id}' is not a valid row ID; the entity might have been merged or deleted`.  Though for `invalid_mbid`, invalid there means "not a UUID," whereas here the message could be triggered by either a "not an ID" situation (like `0` or `abc123`), or a 404-type situation (merged or deleted), from what I can tell, so the explanation at the end makes sense to me.

mwiencek (approved):
Tests LGTM.  Approved this with a minor punctuation suggestion.

reosarevok:

PR #2467, opened 2022-03-24
MBS-12279 / MBS-12283 / MBS-12284: Drop more manual deletion options and fix related issues

reosarevok:
### Fix MBS-12283 and implement MBS-12279 / MBS-12284

See commit messages, but this tries to make the way we use (or do not use) /delete pages more sensible.

Documented removal at https://wiki.musicbrainz.org/How_to_Remove_Entities

One thing I did notice while looking into this is that the conditions for `can_delete` in entities' `Data` pages often differ from the conditions for them to be removed when empty. This probably makes sense for entities that actually do support manual removal, but it feels strange for those which only support autoremoval. Series for example allows deletion unconditionally, but then it is actively expected to fail if the deletion would involve removing relationships. Others allow deletion unconditionally (such as work) but the only supported deletion path is `RemoveEmpty` which won't remove them if they have relationships anyway. We should probably improve this, open to suggestions and happy to add further commits to this or a separate PR :)


mwiencek (approved):

mwiencek:
I didn't investigate the `can_delete` stuff too closely, but I do think it'd make sense if they accurately calculated whether the entity could really be deleted (based on what it tries to delete).

reosarevok:
Added MBS-12306 for that.

PR #2469, opened 2022-03-29
MBS-12289: Avoid crashing on untyped series form

reosarevok:
### Fix MBS-12289

This was crashing because we now do not select a type by default (even though a type is eventually still required).
I amended the check to disallow *all* part of series reltypes if there's no series type selected, and I'm also adding an explanatory message saying the user can't add parts until they select a type.
Also made changes to avoid a similar crash on `updateAllowedTypes`.

### DIFF
--- a/root/components/relationship-editor.tt
+++ b/root/components/relationship-editor.tt
@@ -116,6 +116,13 @@
       <!-- /ko -->
     </td>
   </tr>
+  <!-- ko if: isUntypedSeries() -->
+    <tr>
+      <td colspan="2">
+        [% l('To add parts to this series, please select a series type first.') %]
+      </td>
+    </tr>
+  <!-- /ko -->
 </script>
 
 <script type="text/html" id="template.dialog-date">
--- a/root/static/scripts/relationship-editor/common/dialog.js
+++ b/root/static/scripts/relationship-editor/common/dialog.js
@@ -20,6 +20,7 @@ import linkedEntities from '../../common/linkedEntities';
 import MB from '../../common/MB';
 import * as URLCleanup from '../../edit/URLCleanup';
 import * as dates from '../../edit/utility/dates';
+import isPositiveInteger from '../../edit/utility/isPositiveInteger';
 import {stripAttributes} from '../../edit/utility/linkPhrase';
 import {groupBy} from '../../common/utility/arrays';
 import isBlank from '../../common/utility/isBlank';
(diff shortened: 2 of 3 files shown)


root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -431,21 +431,31 @@ class Dialog {
     });
   }
 
+  isUntypedSeries() {
+    return this.source.entityType === 'series' &&
+           this.source.typeID() === '\xA0';
```
yvanzo, line 436: Where does this value come from?

yvanzo (commented):
Has it been tested? Would it be worth adding an automated test?

root/static/scripts/relationship-editor/common/dialog.js
reosarevok: From https://github.com/metabrainz/musicbrainz-server/blob/89d33473dadc93cb165ba8bece1c08117ccb9c30/lib/MusicBrainz/Server/Plugin/FormRenderer.pm#L203

reosarevok:
I did test adding/editing/removing rels by hand, yes. We could add a test, I guess that'd need to be selenium?

root/static/scripts/relationship-editor/common/dialog.js
mwiencek: A more robust check might be something like `&& !isPositiveInteger(this.source.typeID())` in case the empty placeholder changes.

root/static/scripts/relationship-editor/common/dialog.js
```diff
@@ -431,21 +431,31 @@ class Dialog {
...
     if (this.source.entityType === 'series') {
-      var itemType =
-        MB.seriesTypesByID[this.source.typeID()].item_entity_type;
+      const itemType = this.isUntypedSeries()
+        ? undefined
+        : MB.seriesTypesByID[this.source.typeID()].item_entity_type;
+
+      const allowedPartType = itemType
+        ? PART_OF_SERIES_LINK_TYPES[itemType]
+        : undefined;
```
mwiencek, line 452: semantic nitpick, but I think null makes more sense than undefined to explicitly indicate "no value" (since we're defining it) unless you need undefined.  also uses less characters :)

basically I'd expect undefined only where we access a property that may not exist, or where we want to pass an "empty" value to a function that has default arguments (since undefined will make it use the default, but null won't)

```suggestion
      const itemType = this.isUntypedSeries()
        ? null
        : MB.seriesTypesByID[this.source.typeID()].item_entity_type;

      const allowedPartType = itemType
        ? PART_OF_SERIES_LINK_TYPES[itemType]
        : null;
```

root/components/relationship-editor.tt
```diff
@@ -116,6 +116,13 @@
       <!-- /ko -->
     </td>
   </tr>
+  <!-- ko if: isUntypedSeries() -->
+    <tr>
+      <td colspan="2">
+        [% l('To add parts to this series, please select a type first') %]
```
mwiencek, line 122: Maybe specify series type here in case it's confused with relationship type, since that's in the currently open dialog?
```suggestion
        [% l('To add parts to this series, please select a series type first.') %]
```

mwiencek (approved):

PR #2470, opened 2022-03-30
Log bucket creation failures to Sentry

mwiencek:
We have at least one editor consistently getting "InternalError" / "Failed to get necessary short term bucket lock", and I'd like to collect more data on this.


mwiencek:
I tested this locally & on test.mb and it seems to work fine.  I'll deploy it on beta as another test, and then to production, so I can gather more logs about this error while I'm in communication with Zas & the IA.

PR #2472, opened 2022-04-02
Header component updated

akshaaatt:
The Header/Navbar has been fixed and updated finally to a single row functionality!

Screenshot from 2022-04-03 01-11-16

Screenshot from 2022-04-03 01-20-34


reosarevok:
Just so I understand this, what is the benefit in your opinion of single row? I've never found the double row problematic :)

akshaaatt:
> Just so I understand this, what is the benefit in your opinion of single row? I've never found the double row problematic :)

The single row is better to manage for mobile screens and is pretty much the standard used everywhere plus in my opinion, it looks clean! :)

reosarevok:
I have a few issues so far with the header with a quick test on test.mb: 


It doesn't wrap anywhere as soon as it should, making part of it invisible in medium-small screens:
Screenshot from 2022-04-26 11-36-56

The search bar gets smaller the smaller you make the window, making it useless fairly quick even when still visible:
Screenshot from 2022-04-26 11-38-13

Even at full size for me, the search field fits a lot less text than the previous one, compare:
Screenshot from 2022-04-26 11-36-02
Screenshot from 2022-04-26 11-39-08

There's also clearly too many menus to fit in one line: My Data and About Us going into two lines looks a bit silly.
If you don't want to have two rows (which I still think is the most elegant solution and I would not change, for what it's worth) then you should figure how to combine menus so we have less of them.

The order of menu items also seems odd - I understand it's just "top row followed by bottom row", but "About Us" in the center for example is quite strange. If you want one row, you should think of a way in which the menu order makes more sense IMO.

PR #2473, opened 2022-04-02
Remove jquery styles

akshaaatt:
The latest versions of Bootstrap don't use JQuery and hence we can look to remove these custom styles.


mwiencek (commented):
The code/files you're removing are for jquery-ui, not jquery.  jquery-ui is still used in a lot of places on MusicBrainz, so we can't do this yet.

mwiencek:
Here's the places I can remember off the top of my head that use it, for context --
Entity autocomplete - https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/common/MB/Control/Autocomplete.js
Tag/genre editor - https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/common/components/TagEditor.js
Relationship editor dialog - https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/relationship-editor/common/dialog.js
Artwork viewer - https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/common/artworkViewer.js
(might've missed one or two)

reosarevok:
To which degree can jqueryui be replaced with bootstrap? Ideally we'd probably get rid of it, eventually, but :)

PR #2476, opened 2022-04-04
MBS-12256: Keep meta rating columns up-to-date with triggers

mwiencek:
MBS-12256

This mostly follows a similar implementation to https://github.com/metabrainz/musicbrainz-server/pull/2434.

### DIFF
--- a/admin/sql/CreateFunctions.sql
+++ b/admin/sql/CreateFunctions.sql
@@ -1255,6 +1255,86 @@ BEGIN
 END;
 $$ LANGUAGE 'plpgsql';
 
+-------------------------------------------------------------------
+-- Ratings
+-------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION update_aggregate_rating(entity_type ratable_entity_type, entity_id INTEGER)
+RETURNS VOID AS $$
+BEGIN
+  -- update the aggregate rating for the given entity_id.
+  EXECUTE format(
+    $SQL$
+      UPDATE %2$I
+         SET rating = agg.rating,
+             rating_count = nullif(agg.rating_count, 0)
+        FROM (
+          SELECT count(rating)::INTEGER AS rating_count,
+                 -- trunc(x + 0.5) is used because round() on REAL values
+                 -- rounds to the nearest even number.
+                 trunc((sum(rating)::REAL /
+                        count(rating)::REAL) +
+                       0.5::REAL)::SMALLINT AS rating
+            FROM %3$I
+           WHERE %1$I = $1
+        ) agg
+       WHERE id = $1
+    $SQL$,
+    entity_type::TEXT,
+    entity_type::TEXT || '_meta',
+    entity_type::TEXT || '_rating_raw'
+  ) USING entity_id;
+END;
+$$ LANGUAGE 'plpgsql';
+
+CREATE OR REPLACE FUNCTION update_aggregate_rating_for_raw_insert()
+RETURNS trigger AS $$
+DECLARE
+  entity_type ratable_entity_type;
+  new_entity_id INTEGER;
+BEGIN
+  entity_type := TG_ARGV[0]::ratable_entity_type;
+  EXECUTE format('SELECT ($1).%s', entity_type::TEXT) INTO new_entity_id USING NEW;
+  PERFORM update_aggregate_rating(entity_type, new_entity_id);
+  RETURN NULL;
+END;
+$$ LANGUAGE 'plpgsql';
+
+CREATE OR REPLACE FUNCTION update_aggregate_rating_for_raw_update()
+RETURNS trigger AS $$
+DECLARE
+  entity_type ratable_entity_type;
+  new_entity_id INTEGER;
+  old_entity_id INTEGER;
+BEGIN
+  entity_type := TG_ARGV[0]::ratable_entity_type;
+  EXECUTE format('SELECT ($1).%s', entity_type) INTO new_entity_id USING NEW;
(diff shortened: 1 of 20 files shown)


yvanzo:
Tests need to be updated.

admin/sql/CreateFunctions.sql
```diff
@@ -1255,6 +1255,86 @@ BEGIN
...
+  -- update the aggregate rating for the given entity_id.
+  EXECUTE format(
+    $SQL$
+      UPDATE %2$I
+         SET rating = agg.rating,
+             rating_count = nullif(agg.rating_count, 0)
+        FROM (
+          SELECT count(rating)::INTEGER AS rating_count,
+                 -- trunc(x + 0.5) is used because round() on REAL values
+                 -- rounds to the nearest even number.
```
reosarevok, line 1274: So the idea is that it always truncates down, so by adding 0.5 it will round down to the *next* int if it was .5 or higher? I guess it makes sense.

admin/sql/CreateFunctions.sql
```diff
@@ -1255,6 +1255,86 @@ BEGIN
...
+  entity_type := TG_ARGV[0]::ratable_entity_type;
+  EXECUTE format('SELECT ($1).%s', entity_type) INTO new_entity_id USING NEW;
+  EXECUTE format('SELECT ($1).%s', entity_type) INTO old_entity_id USING OLD;
+  IF (old_entity_id = new_entity_id AND OLD.rating != NEW.rating) THEN
+    -- Case 1: only the rating changed.
+    PERFORM update_aggregate_rating(entity_type, old_entity_id);
+  ELSIF (old_entity_id != new_entity_id OR OLD.rating != NEW.rating) THEN
+    -- Case 2: the entity or rating changed.
+    PERFORM update_aggregate_rating(entity_type, old_entity_id);
+    PERFORM update_aggregate_rating(entity_type, new_entity_id);
```
reosarevok, line 1319: Do we ever change the entity or is this just in case? Not that I think it's bad to have, just trying to make sure.

admin/sql/updates/20220322-mbs-12256-mirror.sql
```diff
@@ -0,0 +1,168 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+UPDATE artist_meta
+   SET rating = agg.rating
+  FROM (
+      SELECT artist,
+             round(sum(rating)::REAL /
+                   count(rating)::REAL)::SMALLINT AS rating
```
reosarevok, line 10: Shouldn't this also use `trunc()` then?

admin/sql/CreateFunctions.sql
mwiencek: Yeah.  This is basically copied from what the Perl was previously doing (`int($rating_sum / $rating_count + 0.5)`).  (Though in that case I think we could've just used `POSIX::round`).

admin/sql/CreateFunctions.sql
mwiencek: I believe only for merges.

admin/sql/updates/20220322-mbs-12256-mirror.sql
mwiencek: Looks like I changed the implementation at some point and didn't recompile the scripts.  Thanks, will fix.

t/lib/t/MusicBrainz/Server/Controller/Artist/Ratings.pm
```diff
@@ -56,7 +56,7 @@ test 'Test artists rating page' => sub {
     );
     $tx->is(
         '(//span[@class="current-rating"])[2]',
-        3.5,
+        3,
```
mwiencek, line 59: This looks like it was just wrong, but maybe someone can confirm my math here.

The SQL at the top adds two raw ratings, 20 and 100.  (20 + 100) / 2 / 20 = 3.  So I'm not even sure where it was getting 3.5 from.

admin/sql/CreateFunctions.sql
```diff
@@ -1255,6 +1255,86 @@ BEGIN
...
+
+CREATE OR REPLACE FUNCTION update_aggregate_rating(entity_type ratable_entity_type, entity_id INTEGER)
+RETURNS VOID AS $$
+BEGIN
+  -- update the aggregate rating for the given entity_id.
+  EXECUTE format(
+    $SQL$
+      UPDATE %2$I
+         SET rating = agg.rating,
+             rating_count = nullif(agg.rating_count, 0)
```
mwiencek, line 1270: The `nullif` preserves the behavior of the Perl, BTW, where no ratings = both the rating and rating_count columns are NULL.

t/lib/t/MusicBrainz/Server/Controller/Artist/Ratings.pm
reosarevok: The whole issue here seems to be that the controller_artist sql file does:

`UPDATE artist_meta SET rating=70, rating_count=4 WHERE id = 3;`

Since we were not earlier updating the meta tables with triggers, so that was just set separately. You can probably remove that as well, if you haven't (and check for any other tests that update the meta tables).

yvanzo (approved):
Rebased on the latest `schema-change-2022-q2`, dropping 3 merge/picked commits in the process. Manually resolved conflicts in `upgrade.json` and reran `admin/CompileSchemaScripts.pl 27`.

LGTMBDNT. Selenium failure seems to be unrelated.

PR #2477, opened 2022-04-05
Upgrade to Flow 0.175.0

mwiencek:
We're currently on 0.173.0.  These two releases introduced some new errors, mostly related to sketchy null checks IIRC.

Changelogs:

https://github.com/facebook/flow/releases/tag/v0.174.0
https://github.com/facebook/flow/releases/tag/v0.175.0


root/components/PartialDateInput.js
```diff
@@ -114,15 +111,15 @@ const PartialDateInput = ({
       event: SyntheticEvent<HTMLInputElement>,
       fieldName: 'year' | 'month' | 'day',
     ) => {
-      dispatch({
+      props.dispatch({
```
mwiencek, line 114: Needs to be accessed via `props.` alongside `props.uncontrolled` so that Flow refines `props` correctly, which is a union type.

root/components/PartialDateInput.js
```diff
@@ -159,7 +156,6 @@ const PartialDateInput = ({
         size={4}
         type="text"
         {...yearProps}
-        {...inputProps}
```
mwiencek, line 162: I haven't found anywhere that we use this.  But if we do, I think it's better we define the specific props we can pass through.

reosarevok (approved):

root/utility/formatUserDate.js
```diff
@@ -163,11 +163,11 @@ export function formatUserDateObject(
 
   const preferences = $c && $c.user ? $c.user.preferences : null;
   const timezone =
-    (options && options.timezone) ||
-    (preferences && preferences.timezone) ||
+    (options ? options.timezone : null) ||
```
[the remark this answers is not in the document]
mwiencek: Yeah, I guess except returning null instead of undefined, but Flow doesn't like it:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ root/utility/formatUserDate.js:166:5

Sketchy null check on string [1] which is potentially an empty string. Perhaps you meant to check for null or
undefined [1]? [sketchy-null-string]

 [1] 152│   timezone?: string,
        :
     163│
     164│   const preferences = $c && $c.user ? $c.user.preferences : null;
     165│   const timezone =
     166│     options?.timezone ||
     167│     preferences?.timezone ||
     168│     'UTC';
     169│   let format =

```

Not sure why it accepts it like I have it, maybe because it's more explicit?

root/components/PartialDateInput.js
mwiencek: I tried, but Flow also doesn't like this. :(  I'll keep it as-is for now...

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ root/components/FormRowPartialDate.js:63:6

Cannot create PartialDateInput element because: [incompatible-type]
 • Either undefined [1] is incompatible with function type [2] in property dispatch.
 • Or function type [1] is incompatible with undefined [3] in property dispatch.

     root/components/FormRowPartialDate.js
 [1] 51│   dispatch,
       :
     60│       label={label}
     61│       required={required}
     62│     />
     63│     <PartialDateInput
     64│       disabled={disabled}
     65│       dispatch={dispatch}
     66│       field={field}

     root/components/PartialDateInput.js
 [2] 33│       +dispatch: (ActionT) => void,
       :
 [3] 38│       +dispatch?: void,



Found 1 error
```

PR #2482, opened 2022-04-11
MBS-12303: Display non-diff annotation data over all columns 

reosarevok:
### Fix MBS-12303

There's no reason to stick all the data into the first column when having a Diff element adds a second one. The easiest solution seems to be to just add a colspan of 2; even when the Diff is not present, all shows as normal since everything then has the same colspan.


mwiencek (approved):
(Diff was much easier to parse with "Hide whitespace" enabled.)

PR #2484, opened 2022-04-11
MBS-12302: Speed up large annotation diffing

mwiencek:
# MBS-12302

Large annotation diffs like for edit #80101929 push the limits of generic-diff, which isn't really designed to diff arrays of thousands of characters.  It's so slow at doing this that edits like the one referenced will simply error due to the template renderer socket timing out and returning incomplete JSON.

The generic-diff README recommends fast-diff for efficient diffing of text.  This is very fast as the name suggests, but doesn't allow for custom "split" props, e.g. "\s+" for word diffing.  So we fall back to fast-diff if the text to be diffed is greater than some arbitrary "large string" length.  I picked 256 characters.

It also makes sense to use fast-diff if the "split" prop is an empty string, since we only perform character diffing in that case.

### DIFF
--- a/lib/MusicBrainz/Server/Sitemap/Overall.pm
+++ b/lib/MusicBrainz/Server/Sitemap/Overall.pm
@@ -323,7 +323,8 @@ sub construct_url_lists($$$@) {
 
         if ($suffix_info{paginated}) {
             # 100 items per page, and the first page is covered by the base.
-            my $paginated_count = ceil($id_info->{$suffix_info{paginated}} / 100) - 1;
+            my $item_count = $id_info->{ $suffix_info{paginated} } // 0;
+            my $paginated_count = ceil($item_count / 100) - 1;
 
             # Since we exclude page 1 above, this is for anything above 0.
             if ($paginated_count > 0) {
--- a/root/static/scripts/edit/components/edit/DiffSide.js
+++ b/root/static/scripts/edit/components/edit/DiffSide.js
@@ -15,8 +15,15 @@ import editDiff, {
   EQUAL,
   CHANGE,
   CLASS_MAP,
+  stringEditDiff,
   type EditType,
 } from '../../utility/editDiff';
+
+/*
+ * The max string length before we fall back from generic-diff
+ * to fast-diff (which doesn't support the custom 'split' prop!).
+ */
+const FAST_DIFF_FALLBACK_LENGTH = 1024;
 
 function splitText(text, split = '') {
   if (split !== '') {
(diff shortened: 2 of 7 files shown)


root/static/scripts/edit/components/edit/DiffSide.js
```diff
@@ -41,10 +48,23 @@ const DiffSide = ({
...
   const stack = [];
   const splitMatch = new RegExp('^(?:' + split + ')$');
-  const diffs = editDiff(
-    splitText(oldText, split),
-    splitText(newText, split),
-  );
+
+  let diffs;
+  if (
+    split === '' ||
```
mwiencek, line 54: This line should actually be sufficient to fix the ticket, since AddAnnotation doesn't seem to use word diffing.  But perhaps we might want to change it to use word diffing, in which case the length check would be helpful.  (Note that splitting the text by \s+ instead of '' would also mean there are less items to compare, so it would be faster.)

reosarevok:
256 chars is shorter than some album titles (https://en.wikipedia.org/wiki/The_Boy_Bands_Have_Won) but I'm not sure if that's a problem or not :) That one is 865 chars - would diffing 1k be unreasonably slow?

yvanzo (approved):
LGTMBDNT. However it doesn’t seem to be worth a short-cut to production. Beta would help with adjusting the value of `FAST_DIFF_FALLBACK_LENGTH` and spotting potential regressions, especially as there is little to zero test coverage.

reosarevok (approved):

mwiencek:
> LGTMBDNT. However it doesn’t seem to be worth a short-cut to production. Beta would help with adjusting the value of FAST_DIFF_FALLBACK_LENGTH and spotting potential regressions, especially as there is little to zero test coverage.

Updated this with a test after you approved BTW (in case you want to look).

lib/MusicBrainz/Server/Sitemap/Overall.pm
```diff
@@ -323,7 +323,8 @@ sub construct_url_lists($$$@) {
 
         if ($suffix_info{paginated}) {
             # 100 items per page, and the first page is covered by the base.
-            my $paginated_count = ceil($id_info->{$suffix_info{paginated}} / 100) - 1;
+            my $item_count = $id_info->{ $suffix_info{paginated} } // 0;
+            my $paginated_count = ceil($item_count / 100) - 1;
```
mwiencek, line 327: Oops, this commit was unrelated to the PR, but something I pushed to production earlier (and it stuck here because I rebased this PR on master, and it wasn't in master yet).  Doesn't matter though.

PR #2485, opened 2022-04-11
MBS-12305: Allow editing/translating instrument (root) rel attribute

reosarevok:
### Fix MBS-12305

See commit messages


mwiencek (approved):

mwiencek:
Maybe try updating POT files and see that it's extracted correctly?

reosarevok:
I did test that earlier already, it worked :)

PR #2486, opened 2022-04-12
MBS-12190: Add Mood support (SQL)

reosarevok:
### Implement MBS-12190

Pretty much 100% a copy of genre

On top of https://github.com/metabrainz/musicbrainz-server/pull/2454


yvanzo (approved):
It still doesn’t make sense to me as moods are still not defined, just like genres, but it is technically harmless to the rest of the database. Added a reminder comment to the ticket.

PR #2487, opened 2022-04-13
MBS-10165: Allow using the edit system for genres

reosarevok:
### Implement MBS-10165

On top of https://github.com/metabrainz/musicbrainz-server/pull/2454


lib/MusicBrainz/Server/Edit/Genre/Create.pm
```diff
@@ -0,0 +1,77 @@
...
+        ),
+    };
+}
+
+sub _insert_hash
+{
+    my ($self, $data) = @_;
+    return $data;
+};
+
```
mwiencek, line 61: Doesn't seem needed since it's identical to the method from Edit::Generic::Create.

root/edit/details/EditGenre.js
```diff
@@ -0,0 +1,57 @@
...
+    <>
+      <table className="details">
+        <tbody>
+          <tr>
+            <th>{addColonText(l('Genre'))}</th>
+            <td><DescriptiveLink entity={display.genre} /></td>
+          </tr>
+        </tbody>
+      </table>
+      <table className="details edit-genre">
```
mwiencek, line 35: I forget, was there a reason we use two separate table.details instead of combining them?

root/genre/DeleteGenre.js
```diff
@@ -9,33 +9,39 @@
...
   +$c: CatalystContextT,
-  +genre: GenreT,
+  +entity: GenreT,
+  +form: GenreDeleteFormT,
 };
 
 const DeleteGenre = ({
   $c,
-  genre,
+  entity,
```
mwiencek, line 27: You can supply an alias when destructuring to keep the previous variable name, too.

```suggestion
  entity: genre,
```

mwiencek (approved):
Tested all three edit types locally, seems to work fine!

root/edit/details/EditGenre.js
reosarevok: I don't remember, we might want to eventually look through these and consistently change them one way or the other.

lib/MusicBrainz/Server/Edit/Genre/Create.pm
reosarevok: Fair enough. I had copied it from Edit::Area::Create which also doesn't seem to need it, so I removed it from there in a separate commit.

MD5SUMS
```diff
@@ -0,0 +1,10 @@
+6e70876751dd04accd96a95e08f3d0fe *mbdump-cdstubs.tar.bz2
```
mwiencek, line 1: Some extra files got accidentally committed here (SHA256SUMS too).

reosarevok: Whoops. That'd be from my testing the issues with MBS-12400. Fixed.

yvanzo (approved):
Added tests look good and are passing! :ship:

PR #2488, opened 2022-04-13
MBS-12308: Clarify the missing artist errors

reosarevok:
### Implement MBS-12308

# Problem
Beginners keep getting confused by "artist is missing" errors in the release editor, because they have typed an artist, so clearly it is not missing! They don't really understand the artist needs to be selected from the DB.

# Solution
This adds errors that specify what exactly means to select an artist.

I couldn't see any reason why we combine missing title and missing artist, which are very different issues with very different solutions. As such, I split the error into two to give a properly useful error message for each.

# Testing
Tested manually in the release editor by leaving titles empty and artists unselected and making sure I got the correct errors.

### DIFF
--- a/root/release/edit/information.tt
+++ b/root/release/edit/information.tt
@@ -27,7 +27,11 @@
         <td class="release-artist" data-bind="artistCreditEditor: $data"></td>
       </tr>
 
-      [% table_row_error(2, 'showErrorWhenTabIsSwitched: needsArtistCredit', l('You haven’t entered a complete artist credit.')) %]
+      [% table_row_error(
+        2,
+        'showErrorWhenTabIsSwitched: needsArtistCredit',
+        l('You must select an artist for the release (the background color of the artist field should be green). If you’ve entered an artist but the background remains uncolored, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it by selecting “{add_a_new_artist}” from the bottom of the search results drop-down.', { add_a_new_artist => l('Add a new artist') }),
+      ) %]
 
       <tr>
         <td><label for="release-group">[% l('Release Group:') %]</label></td>
--- a/root/release/edit/tracklist.tt
+++ b/root/release/edit/tracklist.tt
@@ -432,8 +432,12 @@
         [%~ l('Tracks are required. If you don’t know them, but do know some info (such as the format or number of media), tick the “I don’t know the tracklist for this medium” checkbox above and enter the available info. If you have no information about this release’s media, remove all media.') ~%]
       </p>
 
-      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackInfo">
-        [%~ l('You must enter a title and select an artist for every track.') ~%]
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackTitles">
+        [%~ l('You must enter a title for every track (if the track is untitled, enter “{untitled_track}”).', { untitled_track => '<code>[untitled]</code>' }) ~%]
+      </p>
+
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackArtists">
+        [%~ l('You must select an artist for every track (the background color of every artist field should be green). If you’ve entered an artist but the background remains uncolored, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it by selecting “{add_a_new_artist}” from the bottom of the search results drop-down.', { add_a_new_artist => l('Add a new artist') }) ~%]
       </p>
 
       <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: hasInvalidPregapLength">
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -313,9 +313,12 @@ class Track {
     this.recordingValue(value);
   }
 
-  hasNameAndArtist() {
-    return !isBlank(this.name()) &&
-            isCompleteArtistCredit(this.artistCredit());
+  hasArtist() {
+    return isCompleteArtistCredit(this.artistCredit());
+  }
+
+  hasTitle() {
+    return !isBlank(this.name());
   }
 
   hasVariousArtists() {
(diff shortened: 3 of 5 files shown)


Aerozol:
What a great ticket!! I'm wondering if this could be expanded to something like:

Every track must have a title entered.
Every track must have a MusicBrainz artist selected (green background) - if it is filled in but the background is white, click the magnifying glass to match a MusicBrainz artist.

Maybe even something on the end about entering a new artist? I'm not sure brevity is a requirement here.

More generally speaking I think we can also help solve this by highlighting required fields better, it's on my list somewhere...

reosarevok:
I guess there's no reason to be brief, so sure.

root/release/edit/information.tt
```diff
@@ -27,7 +27,11 @@
         <td class="release-artist" data-bind="artistCreditEditor: $data"></td>
       </tr>
 
-      [% table_row_error(2, 'showErrorWhenTabIsSwitched: needsArtistCredit', l('You haven’t entered a complete artist credit.')) %]
+      [% table_row_error(
+        2,
+        'showErrorWhenTabIsSwitched: needsArtistCredit',
+        l('You must select an artist for the release (the artist background should be green). If you have entered an artist but the background is white, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from the search results.'),
```
mwiencek, line 33: I think that "the artist background should be green" could be slightly more specific ("background" on its own feels like design or CSS terminology to me). What do you think?

```suggestion
        l('You must select an artist for the release (the background color of the artist input field should be green). If you’ve entered an artist but the background is white, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from the search results.'),
```

reosarevok: Sure, made the change.

root/release/edit/tracklist.tt
```diff
@@ -432,8 +432,12 @@
...
       </p>
 
-      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackInfo">
-        [%~ l('You must enter a title and select an artist for every track.') ~%]
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackTitles">
+        [%~ l('You must enter a title for every track (if the track is untitled, enter “[untitled]”).') ~%]
+      </p>
+
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackArtists">
+        [%~ l('You must select an artist for every track (the artist background should be green). If you have entered an artist but the background is white, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from the search results.') ~%]
```
mwiencek, line 440: Thanks, making a similar suggestion for the tracklist one below
```suggestion
        [%~ l('You must select an artist for every track (the background color of every artist field should be green). If you’ve entered an artist but the background is white, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from the search results.') ~%]
```

root/release/edit/tracklist.tt
```diff
@@ -432,8 +432,12 @@
...
       </p>
 
-      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackInfo">
-        [%~ l('You must enter a title and select an artist for every track.') ~%]
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackTitles">
+        [%~ l('You must enter a title for every track (if the track is untitled, enter “[untitled]”).') ~%]
+      </p>
+
+      <p class="error field-error" style="padding: 1em" data-bind="showErrorWhenTabIsSwitched: needsTrackArtists">
+        [%~ l('You must select an artist for every track (the background color of every artist field should be green). If you’ve entered an artist but the background is white, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from the search results.') ~%]
```
yvanzo, line 440: 1. It is “white” only using the current style, but it can be styled using external tools already, and after MBS-10444 in the future. “uncolored” seems to be a more long-standing drop-in replacement.
2. It can be more specific about the location to “Add a new artist”.

```suggestion
        [%~ l('You must select an artist for every track (the background color of every artist field should be green). If you’ve entered an artist but the background remains uncolored, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from below the search results in the drop-down.') ~%]
```

root/release/edit/information.tt
```diff
@@ -27,7 +27,11 @@
         <td class="release-artist" data-bind="artistCreditEditor: $data"></td>
       </tr>
 
-      [% table_row_error(2, 'showErrorWhenTabIsSwitched: needsArtistCredit', l('You haven’t entered a complete artist credit.')) %]
+      [% table_row_error(
+        2,
+        'showErrorWhenTabIsSwitched: needsArtistCredit',
+        l('You must select an artist for the release (the background color of the artist field should be green). If you’ve entered an artist but the background is white, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from the search results.'),
```
yvanzo, line 33: Same as comment to `tracklist.tt`:
```suggestion
        l('You must select an artist for the release (the background color of the artist field should be green). If you’ve entered an artist but the background remains uncolored, click the magnifying glass to match the name to a MusicBrainz artist. If the artist doesn’t exist yet, you can create it from below the search results in the drop-down.'),
```

root/release/edit/tracklist.tt
mwiencek: The uncolored suggestions sounds good to me.

For the second suggestion, maybe we might as well specify the name of the menu item? We could use a parameter to ensure the string is identical: `you can create it by selecting “{add_a_new_artist}” from the bottom of the search results drop-down.', {add_a_new_artist => l('Add a new artist')}`.

reosarevok: Will that work? I am having issues with a similar idea in https://github.com/metabrainz/musicbrainz-server/pull/2778

reosarevok: Can I just say "is not green"? "uncolored" seems like a more complicated word than we need (you can kinda guess what it means as a person with bad English, sure, but you might never have heard it before).

reosarevok: Hmm. After sleeping on this for a couple days, uncolored is probably fine anyway, so I'll use that :)

yvanzo (approved):
LGTMBDNT

mwiencek (approved):

PR #2489, opened 2022-04-13
MBS-12309: Clarify /edit/subscribed also includes collection edits

reosarevok:
### Implement MBS-12309

This seems to confuse people, who wonder how edits for entities they are not subscribed to (directly) end up in this list.

### DIFF
--- a/root/edit/SubscribedEdits.js
+++ b/root/edit/SubscribedEdits.js
@@ -27,6 +27,13 @@ const SubscribedEdits = ({
   <Layout fullWidth title={l('Edits for Your Subscribed Entities')}>
     <div id="content">
       <h1>{l('Edits for Your Subscribed Entities')}</h1>
+
+      <p>
+        {l(`This page lists edits linked to entities you are directly
+            subscribed to, as well as edits linked to entities which are
+            part of a collection you are subscribed to.`)}
+      </p>
+
       <EditList
         editCountLimit={editCountLimit}
         edits={edits}


root/edit/SubscribedEdits.js
```diff
@@ -27,6 +27,13 @@ const SubscribedEdits = ({
   <Layout fullWidth title={l('Edits for Your Subscribed Entities')}>
     <div id="content">
       <h1>{l('Edits for Your Subscribed Entities')}</h1>
+
+      <p>
+        {l(`This page lists edits linked to entities you are directly
+            subscribed to, as well as edits linked to all entities which are
+            part of a collection you are subscribed to.`)}
```
mwiencek, line 34: Would it be helpful to link the "collection you are subscribed to" text to `/user/foo/subscriptions/collection`?

root/edit/SubscribedEdits.js
```diff
@@ -27,6 +27,13 @@ const SubscribedEdits = ({
   <Layout fullWidth title={l('Edits for Your Subscribed Entities')}>
     <div id="content">
       <h1>{l('Edits for Your Subscribed Entities')}</h1>
+
+      <p>
+        {l(`This page lists edits linked to entities you are directly
+            subscribed to, as well as edits linked to all entities which are
```
mwiencek, line 33: I think "all" is implied here. It sounds better to me without it because the first part of the sentence says "linked to entities" but this one says "linked to all entities" which makes it sound like there's a difference?

```suggestion
            subscribed to, as well as edits linked to entities which are
```

mwiencek (approved):

root/edit/SubscribedEdits.js
reosarevok: Hmm. Possibly? We don't pass user info here, but I guess we could, if you think it's worth it.

mwiencek: Right, since we don't have user info here at the moment and I'm not sure how helpful it will be, I think we can pass on doing that for now.

mwiencek (approved):

PR #2491, opened 2022-04-14
Mbs 12312

reosarevok:
### Implement MBS-12312

See commit messages. Worked on this because it's needed for genre annotations (well, either that or writing a layout.tt for genre, but...)


root/static/scripts/annotation/AnnotationHistoryTable.js
```diff
@@ -0,0 +1,156 @@
...
+              </>
+            ) : null}
+            <td>
+              <EditorLink editor={annotation.editor} />
+            </td>
+            <td>
+              {formatUserDate($c, annotation.creation_date)}
+            </td>
+            <td>
+              <a href={`${baseUrl}/annotations/${annotation.id}`}>
```
mwiencek, line 137:
```suggestion
              <a href={`${baseUrl}/annotation/${annotation.id}`}>
```

root/static/scripts/annotation/AnnotationHistoryTable.js
```diff
@@ -0,0 +1,156 @@
...
+    selectedNew: annotations[0]?.id ?? 0,
+    selectedOld: annotations[1]?.id ?? 0,
+  };
+}
+
+function reducer(state: StateT, action: ActionT): StateT {
+  return mutate<WritableStateT, StateT>(state, newState => {
+    switch (action.type) {
+      case 'update-new': {
+        console.log(action);
```
mwiencek, line 51:
```suggestion
```

root/static/scripts/annotation/AnnotationHistoryTable.js
```diff
@@ -0,0 +1,156 @@
...
+
+type StateT = {
+  +selectedNew: number,
+  +selectedOld: number,
+};
+
+type WritableStateT = {
+  selectedNew: number,
+  selectedOld: number,
+};
```
mwiencek, line 38: One trick (which is more helpful for types with a lot of properties) is that you can just spread the read-only type to remove all the variance/read-onlyness.
```suggestion
type WritableStateT = {...StateT}; // this has writable properties
```

root/static/scripts/annotation/AnnotationHistoryTable.js
```diff
@@ -0,0 +1,156 @@
...
+
+function createInitialState(annotations) {
+  return {
+    selectedNew: annotations[0]?.id ?? 0,
+    selectedOld: annotations[1]?.id ?? 0,
+  };
+}
+
+function reducer(state: StateT, action: ActionT): StateT {
+  return mutate<WritableStateT, StateT>(state, newState => {
```
mwiencek, line 48: The `mutate` utility is useful when there's a lot of deeply-nested state, though for simple objects it's easier to just make a copy.  I even avoided this utility in the relationship editor, since I think it made things more confusing than it helped. :(

```suggestion
  const newState: WritableStateT = {...state};
  // ...
  return newState;
```

mwiencek (changes requested):
This works great!  But I noticed that the "View this version" links are broken.

mwiencek (approved):

PR #2494, opened 2022-04-19
MBS-11556: Standardise / clean up YouTube Music URLs

reosarevok:
### Implement MBS-11556

As far as I can tell, YouTube Music uses mostly five kinds of URLs:
1) watch (YouTube videos) for tracks
2) playlist for releases (and also other playlists)
3) channel for artists
4) search for search
5) browse for things like discography lists and, apparently, users' own videos and private stuff.

This supports watch for recording, playlist for release, channel for artist, and does not support browse or search links at all.
I didn't find label, series, etc channels, so I didn't touch those.
I added a new YouTube Music relationship for artists to avoid users having to check for each artist whether there's some paid content or all is free (which might anyway change later).
Label reltypes are not affected at all rn, meaning that streamingfree.label is technically still allowed. We could choose to use the restrict block here to "support" label and then reject the links, but I'm not sure that's useful, so I didn't now.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -292,6 +292,9 @@ export const LINK_TYPES: LinkTypeMap = {
     place: '22ec436d-bb65-4c83-a268-0fdb0dbd8834',
     series: 'f23802a4-36be-3751-8e4d-93422e08b3e8',
   },
+  youtubemusic: {
+    artist: '631712a0-7525-42ba-b7a3-605aa7a238c4',
+  },
 };
 
 // See https://musicbrainz.org/doc/Style/Relationships/URLs#Restricted_relationships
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -5087,6 +5090,88 @@ const CLEANUPS: CleanupEntries = {
...
+          }
+          return {
+            error: l(
+              `Please link to a specific release (playlist) link. Add track
+               (video) links to the relevant recordings and channel links
+               to the relevant artists instead.`,
+            ),
+            result: false,
+            target: ERROR_TARGETS.ENTITY,
+          };
```
yvanzo, line 5170: Someone pasting a `browse` URL such as `https://music.youtube.com/browse/MPREb_UaQXOH57k99` would be prompted this message about track video and recordings which doesn’t seem appropriate.

yvanzo (commented):
The mapping looks code, so does the code. I tested editing entities with some `music.youtube.com` URLs. Only the last error message doesn’t seem to match.

root/static/scripts/edit/URLCleanup.js
reosarevok: Hmm. Ok, that's interesting, I didn't realize `/browse` could be a redirect to `/playlist`. Wonder if it should be explicitly blocked as a redirect? Is it always a redirect? This thing is confusing.

yvanzo: I don’t know if there is a general rule, I just found it while searching for potential label pages.

reosarevok: Split this error into three, with the `/browse` block being an `ERROR_TARGETS.URL` error instead - does that seem better?

I did notice that all the album links in the https://music.youtube.com frontpage are `/browse` links redirecting to `/playlist` ones so this is probably very common. It doesn't seem like the `/browse` link ID and the `/playlist` link ID are related in any way, so we can't do a smarter thing.

yvanzo: Not sure what have been the code changes in the meantime, but at least an appropriate message is now returned about the link being redirect, which seems to be the smartest thing to do indeed. :clap:

tigerman325:
I'm not sure what this is going to do. But I want to add that YouTube Music is not free whereas regular YouTube is with commercials.

yvanzo:
> But I want to add that YouTube Music is not free whereas regular YouTube is with commercials.

Apparently some can be freely streamed but maybe not all. The PR allows for either or both. For examples, see the [tests](https://github.com/metabrainz/musicbrainz-server/pull/2494/commits/210167e7f3ac82c93c6efb145f82e4b2dc9d864e#diff-90a9599fca0aea31f86817a5990642af4a57a9d351f0b12d2231b7e870a0e2e7).

PR #2495, opened 2022-04-19
MBS-12318 / MBS-12323 / MBS-12324: Improvements to ArtistsThatMayBe reports

reosarevok:
### Implement MBS-12318 / MBS-12323 / MBS-12324

# Problem
The ArtistsThatMayBe(Persons|Groups) reports are kind of a mess. They have a fair amount of false positives and do not cover a lot of the relationships and artist types added in the last few years.

# Solution
This updates the reports to check for (or ignore) additional relationship types, plus it makes it so that the Persons report does not try to turn all characters into people. See each commit message for additional details.

# Testing
Tested with sample data to check the new results of the reports seem to make sense.

### DIFF
--- a/lib/MusicBrainz/Server/Report/ArtistsThatMayBeGroups.pm
+++ b/lib/MusicBrainz/Server/Report/ArtistsThatMayBeGroups.pm
@@ -4,16 +4,30 @@ use Moose;
 with 'MusicBrainz::Server::Report::ArtistReport',
      'MusicBrainz::Server::Report::FilterForEditor::ArtistID';
 
-sub query {
-    q{SELECT DISTINCT ON (artist.id) artist.id AS artist_id,
-       row_number() OVER (ORDER BY artist.name COLLATE musicbrainz, artist.id)
-     FROM artist
-     JOIN l_artist_artist ON l_artist_artist.entity1=artist.id
-     JOIN link ON link.id=l_artist_artist.link
-     JOIN link_type ON link_type.id=link.link_type
-     WHERE (artist.type NOT IN (2, 5, 6) OR artist.type IS NULL)
-       AND link_type.name IN ('collaboration', 'member of band', 'conductor position')};
-}
+sub query {<<~'SQL'}
+  WITH possible_groups_entity0 AS (
+         SELECT artist.id, artist.name
+         FROM artist
+         JOIN l_artist_artist laa ON laa.entity0 = artist.id
+         JOIN link ON link.id = laa.link
+         JOIN link_type ON link_type.id = link.link_type
+         WHERE (artist.type NOT IN (2, 5, 6) OR artist.type IS NULL)
+         AND link_type.name IN ('subgroup')),
+       possible_groups_entity1 AS (
+         SELECT artist.id, artist.name
+         FROM artist
+         JOIN l_artist_artist laa ON laa.entity1 = artist.id
+         JOIN link ON link.id = laa.link
+         JOIN link_type ON link_type.id = link.link_type
+         WHERE (artist.type NOT IN (2, 5, 6) OR artist.type IS NULL)
+         AND link_type.name IN ('member of band', 'collaboration', 'conductor position', 'subgroup'))
+  SELECT possible_groups.id AS artist_id,
+         row_number() OVER (ORDER BY possible_groups.name COLLATE musicbrainz, possible_groups.id)
+  FROM
+    (SELECT * FROM possible_groups_entity0
+        UNION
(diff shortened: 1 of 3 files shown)


reosarevok:

lib/MusicBrainz/Server/Report/ArtistsThatMayBePersons.pm
```diff
@@ -12,31 +12,31 @@ WITH groups_entity0 AS (
          JOIN l_artist_artist laa ON laa.entity0 = artist.id
          JOIN link ON link.id = laa.link
          JOIN link_type ON link_type.id = link.link_type
-         WHERE artist.type IS DISTINCT FROM 1
+         WHERE artist.type NOT IN (1, 4)
```
mwiencek, line 15: `null is distinct from 1` will return true, but `null not in (1, 4)` will return null, so this would appear to now exclude all artists with null types?

reosarevok: Hmm, I did not know that, will add OR IS NULL :)

lib/MusicBrainz/Server/Report/ArtistsThatMayBeGroups.pm
```diff
@@ -4,16 +4,30 @@ use Moose;
...
+         AND link_type.name IN ('subgroup')),
+       possible_groups_entity1 AS (
+         SELECT DISTINCT ON (artist.id) artist.id, artist.name
+         FROM artist
+         JOIN l_artist_artist laa ON laa.entity1 = artist.id
+         JOIN link ON link.id = laa.link
+         JOIN link_type ON link_type.id = link.link_type
+         WHERE (artist.type NOT IN (2, 5, 6) OR artist.type IS NULL)
+         AND link_type.name IN ('member of band', 'collaboration', 'conductor position', 'subgroup'))
+  SELECT DISTINCT ON (possible_groups.id) possible_groups.id AS artist_id,
```
mwiencek, line 24: Just a small thing, but the `UNION` below already gets rid of duplicates, so `DISTINCT ON (possible_groups.id)` isn't needed here.

lib/MusicBrainz/Server/Report/ArtistsThatMayBePersons.pm
```diff
@@ -21,23 +21,23 @@ WITH groups_entity0 AS (
          JOIN link ON link.id = laa.link
          JOIN link_type ON link_type.id = link.link_type
          WHERE artist.type IS DISTINCT FROM 1
-         AND link_type.name IN ('member of band', 'collaboration', 'conductor position', 'subgroup')),
+         AND link_type.name IN ('member of band', 'collaboration', 'conductor position', 'artistic director', 'composer-in-residence', 'subgroup')),
```
mwiencek: We should probably use link type IDs here rather than names. :) (Could be a future improvement though.)

mwiencek (approved):

lib/MusicBrainz/Server/Report/ArtistsThatMayBeGroups.pm
```diff
@@ -4,16 +4,45 @@ use Moose;
...
+            artist.type IS NULL
+         )
+         AND link.link_type IN (
+            103, -- member of band
+            102, -- collaboration
+            305, -- conductor position
+            965, -- artistic director
+            855, -- composer-in-residence
+            722  -- subgroup
+         )
```
mwiencek, line 37: Prior to the latest force-push, you just had `AND link_type.name IN ('member of band', 'collaboration', 'conductor position', 'subgroup'))` here. Was adding "artistic director" and "composer-in-residence" intentional?

lib/MusicBrainz/Server/Report/ArtistsThatMayBePersons.pm
```diff
@@ -6,38 +6,95 @@ with 'MusicBrainz::Server::Report::ArtistReport',
...
+            305, -- conductor position
+            965, -- artistic director
+            855, -- composer-in-residence
+            847, -- teacher
+            108, -- is person
+            111, -- married
+            110, -- sibling
+            109, -- parent
+            112  -- involved with
+         )
```
mwiencek, line 65: Looks like several more types were added here, too, since the last force-push -- I assume also intentional?

reosarevok: Yes, this added the subsets of supporting musician - I *think* we probably thought those were covered by the previous thing since the partially matched and that's why they weren't selected separately (although I don't think that was ever true).

PR #2496, opened 2022-04-20
MBS-12249: Add a materialized area_containment table kept up-to-date with triggers

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12249


lib/MusicBrainz/Server/Data/Place.pm
```diff
@@ -37,7 +37,7 @@ sub _type { 'place' }
 
 sub _columns
 {
-    return 'place.id, place.gid, place.name, place.type, place.address, place.area, place.coordinates[0] as coordinates_x, ' .
+    return 'place.id, place.gid, place.name COLLATE musicbrainz, place.type, place.address, place.area, place.coordinates[0] as coordinates_x, ' .
```
reosarevok, line 39: Maybe too nitpicky, but this for example is like a very long line (146 chars) which probably would benefit from a split. Might be true of all lines here though...

lib/MusicBrainz/Server/Data/Event.pm
```diff
@@ -46,6 +46,8 @@ sub _columns
            'event.comment, event.last_updated';
 }
 
+sub _area_cols { [qw( event_area.area )] }
```
reosarevok, line 49: We have `_id_column` under this, so maybe `_area_columns` would be better? I don't see a huge use of shortening here. I understand we call it cols right now in the role, but maybe it's better to be clear than concise :)

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -48,7 +48,7 @@ sub _type { 'artist' }
 
 sub _columns
 {
-    return 'artist.id, artist.gid, artist.name, artist.sort_name, ' .
+    return 'artist.id, artist.gid, artist.name COLLATE musicbrainz, artist.sort_name, ' .
```
reosarevok, line 51: If we're already doing this, maybe we should also `COLLATE` `sort_name` here and have a proper look at how we use `COLLATE` elsewhere, but I guess that might be a lot of extra work to do it properly? If you think it's of use but not right now, I can make a ticket I guess. In any case, we have plenty of cases where we order by collated sort name - wondering if we shouldn't be doing that for all artist queries where we use name now even?

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -254,33 +254,22 @@ sub is_valid_token {
...
+        $has_materialized_data = $sql->select_single_value(
+            'SELECT 1 FROM area_containment LIMIT 1',
+        ) ? 1 : 0;
+    }
+
+    if ($has_materialized_data) {
+        return 'area_containment';
+    }
+
+    return 'get_area_parent_hierarchy_rows(NULL)';
```
reosarevok, line 272: You've suggested we should be using DISTINCT with `get_area_parent_hierarchy_rows`, but AFAICT you don't when using this as a fallback?

lib/MusicBrainz/Server/Data/Event.pm
reosarevok: I see we already use `_area_cols` for some entities so this has to be named like this, but I'd still suggest changing it everywhere in a followup commit. AFAICT it's the only place we use `_cols` instead of `_columns` outside of a small use in `extract_pot_db`.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -98,25 +98,6 @@ sub find_by_subscribed_editor
...
-        $containment_query,
-        @containment_query_args,
-    ) = get_area_containment_query('$2', 'any(array[area, begin_area, end_area])', check_all_levels => 1);
-    my $query = 'SELECT ' . $self->_columns . '
-                 FROM ' . $self->_table . "
-                 WHERE \$1 IN (area, begin_area, end_area) OR EXISTS (
-                    SELECT 1 FROM ($containment_query) ac
-                     WHERE ac.descendant IN (area, begin_area, end_area) AND ac.parent = \$1
-                 )
-                 ORDER BY artist.name COLLATE musicbrainz, artist.id";
```
reosarevok, line 113: I'd consider using:

```
sub _find_by_area_order {
    return 'sort_name, id';
}
```
here while you're at it, but not a big deal either way I guess :)

reosarevok (changes requested):
Couple broken test issues need changing, most other things are comments rather than any serious change requests.

admin/sql/updates/20220408-mbs-12249-standalone.sql
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE OR REPLACE TRIGGER a_upd_l_area_area AFTER UPDATE ON l_area_area
+    FOR EACH ROW EXECUTE PROCEDURE a_upd_l_area_area_mirror();
+
+CREATE OR REPLACE TRIGGER a_del_l_area_area AFTER DELETE ON l_area_area
+    FOR EACH ROW EXECUTE PROCEDURE a_del_l_area_area_mirror();
+
+CREATE OR REPLACE TRIGGER deny_deprecated BEFORE INSERT ON link
+    FOR EACH ROW EXECUTE PROCEDURE deny_deprecated_links();
+
```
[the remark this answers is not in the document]
mwiencek: Yes, specifying multiple firing conditions doesn't create separate triggers.  It's still one trigger, just a different value in `pg_trigger.tgtype`.

admin/sql/CreateFunctions.sql
```diff
@@ -1529,10 +1529,10 @@ $$ LANGUAGE 'plpgsql';
 CREATE OR REPLACE FUNCTION deny_deprecated_links()
 RETURNS trigger AS $$
 BEGIN
-  IF (TG_OP = 'INSERT' OR (TG_OP = 'UPDATE' AND OLD.link_type <> NEW.link_type))
+  IF TG_OP = 'INSERT'
```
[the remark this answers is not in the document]
mwiencek: Updating this to remove the `TG_OP` check.

admin/BuildMaterializedTables
```diff
@@ -155,12 +156,49 @@ if ($build_all || $options{artist_release_group}) {
...
+    log_info { 'Building area_containment ...' };
+    $c->sql->begin;
+    if (
+        !$force &&
+        $c->sql->select_single_value('SELECT 1 FROM area_containment LIMIT 1')
+    ) {
+        log_info { 'Table not empty, skipping.' };
+        $c->sql->rollback;
+    } else {
+        $c->sql->do('TRUNCATE area_containment');
```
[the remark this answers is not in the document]
mwiencek: You can also specify `--force` to rebuild even if it's not empty.

lib/MusicBrainz/Server/Data/Place.pm
mwiencek: Agreed, but it was already well over the length limit, I just made it slightly worse. :)

I think we should change all of the `_columns` implementations to do something like `join q(, ), (...)` so we can put each item on its own line.  But then they should also probably be changed to lazy Moose attributes rather than subroutines, so that the joined value is cached.  And I thought of doing all this, but didn't really want to complicate this PR any further.

lib/MusicBrainz/Server/Data/Artist.pm
mwiencek: I think it makes sense to include here, but would prefer looking at it later for now.  But the nice thing is that we can add `COLLATE` here without breaking any `COLLATE` expressions elsewhere in the queries.  At worst we just have some useless expressions lying around.

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: Good point!

lib/MusicBrainz/Server/Data/Artist.pm
mwiencek: (To be clear, I changed it to `artist.sort_name COLLATE musicbrainz` here, just haven't properly looked into our use of `COLLATE` elsewhere.)

lib/MusicBrainz/Server/Data/Place.pm
reosarevok: When around, maybe you can make a ticket for your idea and we can implement it later on? :) For now we can just merge as-is

lib/MusicBrainz/Server/Data/Artist.pm
reosarevok: We can look into this further later.

reosarevok (approved):
I tested this locally a bit and the area containment seemed to work just fine (events for the child area appeared on the parent area after creating the link).

yvanzo (approved):
Rebased, re-rebased, tested, fixed rebasing, retested.

PR #2492, opened 2022-04-14
MBS-12311: Allow adding annotations to genres

reosarevok:
### Implement MBS-12311

On top of https://github.com/metabrainz/musicbrainz-server/pull/2487 and https://github.com/metabrainz/musicbrainz-server/pull/2491

PR #2497, opened 2022-04-21
MBS-11599: Allow adding URL relationships to genre pages

reosarevok:
### Implement MBS-11599

Not implementing a full relationship editor yet since that has no easy React implementation and mwiencek is working on that for all entities, by which point it should be trivial to also extend it to genres. In the meantime, being able to link genres to Wikidata, Rate Your Music and whatnot seems like a great first step.

We do need a "mock" relationship editor section for hidden inputs to be attached to, though, so one is added.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2492


root/static/scripts/genre/components/GenreEditForm.js
```diff
@@ -0,0 +1,96 @@
...
+  $c,
+  attrInfo,
+  form,
+  sourceEntity,
+  typeInfo,
+}: Props): React.Element<'form'> => {
+  const externalLinksEditorContainerRef = React.useRef(null);
+
+  const handleSubmit = () => {
+    console.log('preparing submission');
```
mwiencek, line 39: :scissors:

root/static/scripts/genre/components/GenreEditForm.js
```diff
@@ -0,0 +1,96 @@
...
+    prepareSubmission('edit-genre');
+  };
+
+  React.useEffect(() => {
+    exportTypeInfo(typeInfo, attrInfo);
+    invariant(externalLinksEditorContainerRef.current != null);
+    // $FlowIgnore
+    MB.sourceExternalLinksEditor = createExternalLinksEditor({
```
mwiencek, line 47: This effect should also have a cleanup handler where we unmount the external links editor root it creates. 
 (`createExternalLinksEditor` will have to return the root too.)

You can already see that an error is logged due to this effect running twice in development mode.

To register a cleanup function, just return a function from the effect. I guess something like this is the idea:

```JS
const {root, externalLinksEditorRef} = createExternalLinksEditor(...);

MB.sourceExternalLinksEditor = externalLinksEditorRef;

return () => {
  root.unmount();
};

Comment (reosarevok) [reply root/static/scripts/genre/components/GenreEditForm.js:47]: I tried that, but it only replaced the warning with a different one:

`Warning: Attempted to synchronously unmount a root while React was already rendering. React cannot finish unmounting the root until the current render has completed, which may lead to a race condition.`

File: root/types/relationship.js [hunk 3]
```diff
@@ -34,6 +36,10 @@ declare type LinkTypeAttrTypeT = {
   +min: number | null,
 };
 
+declare type LinkTypeOptionsT = {
+  [entityCombination: string]: $ReadOnlyArray<LinkTypeT>,
```
mwiencek, line 40:
```suggestion
  +[entityCombination: string]: $ReadOnlyArray<LinkTypeT>,
```

mwiencek (approved):
The new URL editing features worked fine for me when I had tested this locally while adding the fixup commits (which should be squashed before merging).

reosarevok:

PR #2498, opened 2022-04-21
MBS-12479: Add aliases for genres

reosarevok:
### Implement MBS-12479

This just implements the basic ability to add genre aliases. It does not yet do anything of more use with them than just storing them (that is, it won't associate them with tags yet) but it seems sensible to already allow storing the aliases so that we have them already whenever we want to use them further.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2497


yvanzo (approved):
Successfully tested adding/editing/removing “classical” aliases with type and date. :100:

yvanzo:
The linked ticket is a top-level ticket for the whole genre alias feature, but there are two sides of this feature that would be expected but are missing:
* Supporting genre aliases in API, that is `inc=aliases` for `genre` endpoint.
* Supporting genre aliases in search, that is in search results, as a search field, and in default request parameters.

I would instead link to a “sub-new feature” more specifically about editing and displaying genre aliases through the website.

reosarevok:
Fair enough, added MBS-12478 and MBS-12479 (will mark this as MBS-12479). SEARCH-681 added for the search bit, which AFAICT cannot be a subtask because of it being a different project.

PR #2499, opened 2022-04-25
MBS-12335: Link to attribute docs from reltype docs

reosarevok:
### Implement MBS-12335

Since we now have pages for the attributes, it seems useful to link to them. At the very least, this provides an easy way to get the attribute MBID if someone wants to use it.

I also changed the formatting a bit because having `<h3>` inside a `<p>` seemed odd (is that even valid?). It should make more sense now.


reosarevok:

yvanzo (approved):
Tested locally and it looks good to me, thanks! :ship: 

Well done with `p` tag.

PR #2500, opened 2022-04-25
MBS-12334: Collapse ISRCs and ISWCs when there are too many

reosarevok:
### Implement MBS-12334

### DIFF
--- a/root/components/list/RecordingList.js
+++ b/root/components/list/RecordingList.js
@@ -169,6 +169,7 @@ const RecordingList = ({
         data={recordings}
       />
       {manifest.js('common/components/AcoustIdCell', {async: 'async'})}
+      {manifest.js('common/components/IsrcList', {async: 'async'})}
     </>
   );
 };
--- a/root/components/list/WorkList.js
+++ b/root/components/list/WorkList.js
@@ -105,6 +105,7 @@ const WorkList = ({
       <Table columns={columns} data={works} />
       {manifest.js('common/components/ArtistRoles', {async: 'async'})}
       {manifest.js('common/components/AttributeList', {async: 'async'})}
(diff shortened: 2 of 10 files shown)


yvanzo (approved):
Tested with the example given in the linked ticket: https://musicbrainz.org/recording/21bc4ba0-3b43-43b8-9702-f5f066c894e8

I noticed that the inserted “show … more” (ISRCs) paragraph has more vertical spacing than the “show … more” (Release events) list item has in release sidebar (for example see <https://musicbrainz.org/release/1173188d-3fa7-416b-8c5e-8c5e75872748>). It should probably be a `<div>` tag rather than a `<p>` tag. Not sure how to do that though.

Otherwise the code looks good to me and the overall is already an improvement over the current display.

reosarevok:
Changing it to <div> is trivial but it didn't seem to help much, so will leave as-is for now.

yvanzo:
@reosarevok: I used the inspector tool to change `p` to `div` and it did solve the issue locally. Can you please provide the trivial change you tried?

reosarevok:
That'd be just changing `p` to `div` on https://github.com/metabrainz/musicbrainz-server/blob/83fd1488c90e3d7c9d363f97cfccc0aa6d614368/root/static/scripts/common/components/IsrcList.js#L45 (and I guess `IswcList` too) but I didn't really notice any significant difference. Maybe I'm blind though, if you find it differs just push the change straight to master if you want :)

yvanzo:
This is not changing the `p` around “(show … more)”.

yvanzo:
Was probably a cache issue, probably not the flow type issue. Anyway, it does the intended change, see the pull request #2612.

PR #2501, opened 2022-04-25
MBS-12333: Block smart links: share.amuse.io

reosarevok:
### Implement MBS-12333

These are the same people we were already blocking as amu.se. They claim to also run a label, so not blocking stuff outside the share. subdomain in case they actually eventually have legit pages.


yvanzo (approved):
The partial block seems to be legitimate. Selenium tests failure can be ignored for this change. :ship:

PR #2502, opened 2022-04-25
MBS-12328: Add #content to entity/Edits for artwork display

reosarevok:
### Fix MBS-12328

This got fixed with MBS-12294 (https://github.com/metabrainz/musicbrainz-server/pull/2474) by also checking for artwork inside #edits but it turns out a bunch of pages have both #content and #edits, leading to images being loaded into boxes twice. On a quick check, it seems entity/Edits is the only edit list missing #content anyway, for no clear reason, so it seems a lot easier to just add a #content div to it and revert the previous patch.

### DIFF
--- a/root/entity/Edits.js
+++ b/root/entity/Edits.js
@@ -82,20 +82,22 @@ const Edits = ({
 
   return (
     <Layout fullWidth title={pageTitle}>
-      <div className={className}>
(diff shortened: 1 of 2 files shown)


reosarevok:

mwiencek (approved):

PR #2503, opened 2022-04-26
MBS-11920: Report for releases with low data quality

reosarevok:
### Implement MBS-11920

# Problem
There's no easy way for a user to find all low quality releases for entities in their subscriptions in order to help improve them.

# Solution
This adds a report listing all low quality releases, which as usual can be filtered to show only releases for entities in the user's subscriptions.

This existing should also make it less stressful to just mark something as low quality if an editor has no time or skill to fix it, since now they will know other users (or even themselves when they have the time) will be able to find it and hopefully work on it.

# Testing
By hand, running the report and checking that the releases in it (for the sample data) are all low quality.

### DIFF
--- a/root/report/LowQualityReleases.js
+++ b/root/report/LowQualityReleases.js
@@ -1,0 +1,43 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2022 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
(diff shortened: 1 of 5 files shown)


root/report/LowQualityReleases.js
```diff
@@ -0,0 +1,43 @@
...
+  generated,
+  items,
+  pager,
+}: ReportDataT<ReportReleaseT>): React.Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows releases that have been marked as low quality.
+       If you have some time, you can review them and try to improve the data
+       as much as possible before changing its quality back to Normal
```
mwiencek, line 28: you can review them [...] before changing their

```suggestion
       as much as possible before changing their quality back to Normal
```

mwiencek (approved):

PR #2504, opened 2022-04-27
MBS-11916: Report for recordings marked both karaoke and instrumental

reosarevok:
### Implement MBS-11916

# Problem
A recording should never be marked both karaoke and instrumental: lyrics are relevant for karaoke recordings, and proper instrumental recordings should not be linked as karaoke. This is a relatively common mistake though, and we currently have no good way to find it when it happens.

# Solution
This adds a report for users to be able to clean up these (it usually requires figuring out which of the options is correct in a case by case basis so a report is a good fit).

# Testing
By hand, running the report and checking that the recordings in it (for the sample data) seem to fit the requirements indeed.

### DIFF
--- a/root/report/KaraokePlusInstrumentalRecordings.js
+++ b/root/report/KaraokePlusInstrumentalRecordings.js
@@ -1,0 +1,63 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2022 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import * as React from 'react';
+
+import RecordingList from './components/RecordingList.js';
+import ReportLayout from './components/ReportLayout.js';
+import type {ReportDataT, ReportRecordingT} from './types.js';
+
+component KaraokePlusInstrumentalRecordings(...{
+  canBeFiltered,
+  filtered,
+  generated,
+  items,
+  pager,
+}: ReportDataT<ReportRecordingT>) {
+  return (
+    <ReportLayout
+      canBeFiltered={canBeFiltered}
+      description={exp.l(
+        `This report shows recordings which have a
+         {doc_karaoke_rel|“karaoke” relationship} to another recording,
+         but are linked to a work with the
+         {doc_instrumental_attr|“instrumental” attribute}, or the other
+         way around, with an
+         {doc_instrumental_rel|“instrumental” relationship} and the
+         {doc_karaoke_attr|“karaoke” attribute}.`,
+        {
+          doc_instrumental_attr: '/relationship-attributes#instrumental',
+          doc_instrumental_rel:
+            '/relationship/9fc01a58-7801-4bd2-b07d-61cc7ffacf90',
+          doc_karaoke_attr: '/relationship-attributes#karaoke',
+          doc_karaoke_rel:
+            '/relationship/39a08d0e-26e4-44fb-ae19-906f5fe9435d',
+        },
+      )}
+      entityType="relationship"
+      extraInfo={l(
+        `Keep in mind that “instrumental” in MusicBrainz implies the lyrics
+         are not relevant to the recording. Since lyrics are by definition
+         relevant to karaoke recordings, “instrumental” should not be used
+         on them (use “karaoke” instead). Alternatively, if this is not
+         a karaoke recording, but just a standard instrumental recording,
+         it shouldn’t be linked to another recording with a “karaoke”
+         relationship, but with an “instrumental” one.`,
+      )}
+      filtered={filtered}
+      generated={generated}
+      title={l('Recordings marked as both karaoke and instrumental')}
+      totalEntries={pager.total_entries}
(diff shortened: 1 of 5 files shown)


root/report/KaraokeRecordingsMarkedInstrumental.js
```diff
@@ -0,0 +1,44 @@
...
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows recordings which have a “karaoke” relationship to
+       another recording, but are also linked to a work with the
+       “instrumental” attribute. Since lyrics are relevant to karaoke
+       recordings, “instrumental” should not be used on them. Alternatively,
+       if this is not a karaoke recording, but just an instrumental, it
+       shouldn’t be linked to another recording with a karaoke relationship.`,
+    )}
```
yvanzo, line 33: It isn’t clear how an instrumental with lyrics should be handled. I started this discussion in [comment to the ticket](https://tickets.metabrainz.org/browse/MBS-11916?focusedCommentId=63224&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-63224) so as to include issue’s reporter too.

reosarevok: That's a huge edge case, it's not clear how it should be handled because there's been no discussion before (I've never seen the concept), but it doesn't seem problematic for all the clear-cut cases :)

yvanzo: It helped to narrow the issue: The terms “instrumental” and “karaoke” are widely used outside of MusicBrainz in different ways. It seems it would be worth reminding in the report’s description that these terms have a specific meaning in the context of MusicBrainz.

reosarevok: Added `extraInfo`, do check if that seems better :)

yvanzo: It is slightly better, but still confusing, especially the undefined “standard instrumental”. I failed to improve the wording though, likely because the definition of “instrumental” in MusicBrainz is needlessly too far from the common definition. Both attributes “karaoke” and “instrumental” could be allowed at the same time if the definition of “instrumental” was broadened to match linking recordings without vocals to works with lyrics. It would more be a STYLE issue though.

mwiencek: Can this link to the guidelines where karaoke and instrumental are defined?

reosarevok: I did that now.

root/report/KaraokePlusInstrumentalRecordings.js
```diff
@@ -0,0 +1,63 @@
...
+}: ReportDataT<ReportRecordingT>) {
+  return (
+    <ReportLayout
+      canBeFiltered={canBeFiltered}
+      description={exp.l(
+        `This report shows recordings which have a
+         {doc_karaoke_rel|“karaoke” relationship} to another recording,
+         but are linked to a work with the
+         {doc_instrumental_attr|“instrumental” attribute}, or the other
+         way around, with a
```
mwiencek, line 31:
```suggestion
         way around, with an
```

root/report/KaraokePlusInstrumentalRecordings.js
```diff
@@ -0,0 +1,63 @@
...
+      )}
+      entityType="relationship"
+      extraInfo={l(
+        `Keep in mind that “instrumental” in MusicBrainz implies the lyrics
+         are not relevant to this recording. Since lyrics are by definition
+         relevant to karaoke recordings, “instrumental” should not be used
+         on them (use “karaoke” instead). Alternatively, if this is not
+         a karaoke recording, but just a standard instrumental recording,
+         it shouldn’t be linked to another recording with a karaoke
+         relationship, but with an “instrumental” one.`,
```
mwiencek, line 51: it looks like you're mainly using quotes around karaoke/instrumental when it refers to the specific relationship or attribute, except in this one case
```suggestion
         it shouldn’t be linked to another recording with a “karaoke”
         relationship, but with an “instrumental” one.`,
```

mwiencek (approved):
LGTM except for a few possible grammatical tweaks.

root/report/KaraokePlusInstrumentalRecordings.js
```diff
@@ -0,0 +1,63 @@
...
+            '/relationship/9fc01a58-7801-4bd2-b07d-61cc7ffacf90',
+          doc_karaoke_attr: '/relationship-attributes#karaoke',
+          doc_karaoke_rel:
+            '/relationship/39a08d0e-26e4-44fb-ae19-906f5fe9435d',
+        },
+      )}
+      entityType="relationship"
+      extraInfo={l(
+        `Keep in mind that “instrumental” in MusicBrainz implies the lyrics
+         are not relevant to this recording. Since lyrics are by definition
```
mwiencek, line 46: (github ate this comment when I submitted the previous review somehow?)

"the" sounds a bit better to me here since it's talking about any instrumental recording, not a specific one
```suggestion
         are not relevant to the recording. Since lyrics are by definition
```

PR #2505, opened 2022-04-27
MBS-12230 / MBS-12342 / MBS-12229 / MBS-12341: Allow pasting / parsing dates on date fields

reosarevok:
### Implement MBS-12230 / MBS-12342 / MBS-12229 / MBS-12341

A mostly straightforward basic implementation of the paste-a-date concept. This first PR only supports basic straightforward cases where there is no possible clash (YYYY-MM-DD, both Western and East Asian). We can add support for other date strings that do require user choices (such as DD/MM/YYYY vs MM/DD/YYYY, where the user needs to pick what they meant for a date such as "01/02/2020") in a second PR once we know the basic concept works fine.

# Testing
This adds a fair amount of tests for the parsing functions.

### DIFF
--- a/root/static/scripts/common/utility/parseNaturalDate.js
+++ b/root/static/scripts/common/utility/parseNaturalDate.js
@@ -8,8 +8,8 @@
  */
 
 
-const ymdRegex = /^\W*([0-9]{4})(?:\W+(0?[1-9]|1[0-2])(?:\W+(0?[1-9]|[12][0-9]|3[01]))?)?\W*$/;
-const cjkRegex = /^\W*([0-9]{2}|[0-9]{4})(?:(?:\u5E74|\uB144\W?)(0?[1-9]|1[0-2])(?:(?:\u6708|\uC6D4\W?)(0?[1-9]|[12][0-9]|3[01])(?:\u65E5|\uC77C)?)?)?\W*$/;
+const ymdRegex = /^[^\w?]*([0-9]{4}|\?{4})(?:[^\w?]+(0?[1-9]|1[0-2]|\?{2})(?:[^\w?]+(0?[1-9]|[12][0-9]|3[01]|\?{2}))?)?[^\w?]*$/;
+const cjkRegex = /^[^\w?]*([0-9]{2}|[0-9]{4})(?:(?:\u5E74|\uB144[^\w?]?)(0?[1-9]|1[0-2])(?:(?:\u6708|\uC6D4[^\w?]*?)(0?[1-9]|[12][0-9]|3[01])(?:\u65E5|\uC77C)?)?)?[^\w?]*$/;
 
 function cleanDateString(
   str: string,
@@ -69,11 +69,14 @@ export default function parseNaturalDate(
                 cleanedString.match(ymdRegex) ||
                 [];
 
+  const year = match[1] === '????' ? '' : match[1];
+  const month = match[2] === '??' ? '' : match[2];
+  const day = match[3] === '??' ? '' : match[3];
   return {
     /* eslint-disable sort-keys */
-    year: match[1] || '',
-    month: match[2] || '',
-    day: match[3] || '',
+    year: year || '',
+    month: month || '',
+    day: day || '',
     /* eslint-enable sort-keys */
   };
 }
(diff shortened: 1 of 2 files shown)


mwiencek (commented):
This is very useful and seems to work well.

When the fields are initially loaded, or when you modify any of the year, month, or day fields, I was wondering if we should

 * update the paste-a-date field with the new formatted date (in ISO format), to provide an easy way to copy existing dates, and
 * enable the [HTML autocomplete attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) on the paste-a-date field so that it shows a dropdown of previous dates you've entered, for quick access.

reosarevok:
Hmm. The first one seems like a no-brainer, and I wish I would have thought of it! The second might be annoying at times, but also probably useful at times... maybe worth a shot if it's not complicated (which I guess it should not be).

root/static/scripts/edit/components/PartialDateInput.js
```diff
@@ -188,6 +209,21 @@ const PartialDateInput = (props: Props): React$Element<'span'> => {
...
       />
+      {props.uncontrolled ? null : (
+        <>
+          {' '}
+          <input
+            className="partial-date-parser"
+            disabled={disabled}
+            id={'id-' + field.html_name + '.partial-date-parser'}
+            name={field.html_name + '.partial-date-parser'}
+            placeholder={l('Enter date string to parse')}
```
mwiencek, line 220: I'm not sure if "string to parse" is the best phrasing to use for the average user, but don't really have any better suggestions. :)

Maybe just something like "Paste a full date here"?

reosarevok: "or paste a full date here"?

mwiencek: That sounds good to me.

mwiencek:
> The second might be annoying at times, but also probably useful at times... maybe worth a shot if it's not complicated (which I guess it should not be).

True...at least in Firefox you can delete individual autofill entries with shift+delete (on macOS).Doesn't seem to work in Chrome anymore, though...

reosarevok:
To update the paste-a-date field with the new formatted date, I guess we would need the paste-a-date field to be part of the state, which it doesn't seem to be right now, right? Possibly a stupid question but, would that trigger a new set-date because the paste-a-date field changed? If so I guess we'd need a way to stop that to avoid silly lioops.

julian45 (dismissed):
The link in this commit message is dead; after some browsing in the Internet Archive, I found that they deprecated this site as of 5/31/22. I was able to find some copies of the page earlier than that date, though. Could we maybe use an IA link instead? Such as http://web.archive.org/web/20220330023649/https://reference.discogs.com/wiki/japanese-release-dates

julian45:
I was able to find at least one release with proof of the special Japanese date formatting; see the back cover art [here](https://musicbrainz.org/release/30d2a53a-765a-45d6-8b39-8d4aca4c82b6/cover-art). so I guess it's good to know that this _was_, in fact, used on real releases. I'd still like to find better standards documentation at some point, but that's definitely not blocking for this PR.

root/static/scripts/common/utility/parseNaturalDate.js
```diff
@@ -0,0 +1,81 @@
...
+    /[０-９－]/g,
+    function (fullwidthDigit) {
+      return String.fromCharCode(
+        fullwidthDigit.charCodeAt(0) -
+        ('０'.charCodeAt(0) - '0'.charCodeAt(0)),
+      );
+    },
+  );
+
+  // See https://reference.discogs.com/wiki/japanese-release-dates#date-format
```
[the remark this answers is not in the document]
reosarevok: I entirely missed that I had a comment too, thanks!

mwiencek (commented):
A couple issues I've noted after quickly testing:

 * The date parser fields are missing from the artist form.
 * Entering YYYY=1999, MM=(no value), DD=09 in the date fields correctly renders `1999-??-09` in the parser field, but pasting `1999-??-09` in the parser parses it as YYYY=1999, MM=09 instead of YYYY=1999, DD=09.

PR #2506, opened 2022-04-27
MBS-12131: PostgreSQL 14 compatibility

mwiencek:
This patch contains the schema changes required for PostgreSQL 14 to
work.  Note that these changes are also compatible with PostgreSQL 12.

For context, refer to the first incompatibility listed under
https://www.postgresql.org/docs/release/14.0/.

Because anycompatiblearray doesn't exist in v12, the ideal solution (and
the one used here) is to use more specific types.

 * The _median function is only used by the median aggregate, which is
   only used to calculate median recording lengths -- an integer value.

 * The array_accum aggregate was added in 33bd14e for tracklist_index, a
   table which was removed years ago.  The aggregate is unused and can
   be removed too.

 * array_cat_agg is only used internally by dbmirror.  (See
   https://github.com/metabrainz/dbmirror/blob/ca536c7/pending.c#L405.)
   Since pg_constraint.conkey is of type int2[], we can use that.

### DIFF
--- a/admin/sql/ReplicationSetup.sql
+++ b/admin/sql/ReplicationSetup.sql
@@ -8,9 +8,9 @@ BEGIN;
 --CREATE FUNCTION "recordchange" () RETURNS trigger AS
 --'$libdir/pending', 'recordchange' LANGUAGE C;
 
-CREATE AGGREGATE array_cat_agg(anyarray)(
+CREATE AGGREGATE array_cat_agg(int2[]) (
       sfunc       = array_cat,
-      stype       = anyarray,
+      stype       = int2[],
       initcond    = '{}'
 );
 
--- a/upgrade.json
+++ b/upgrade.json
@@ -123,6 +123,7 @@
                    "20220314-mbs-12253-standalone.sql",
                    "20220314-mbs-12254-standalone.sql",
                    "20220314-mbs-12255-standalone.sql",
-                   "20220322-mbs-12256-standalone.sql"]
+                   "20220322-mbs-12256-standalone.sql",
+                   "20220426-mbs-12131.sql"]
   }
 }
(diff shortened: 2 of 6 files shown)


mwiencek:
@reosarevok @yvanzo should we include this in the schema change release and update the blog post? I don't see any issue with it since these are all internal-only functions/aggregates. It'd be nice to be able to support PostgreSQL 14 for development, which is now the current version in Arch's repo.

admin/sql/ReplicationSetup.sql
```diff
@@ -8,9 +8,9 @@ BEGIN;
 --CREATE FUNCTION "recordchange" () RETURNS trigger AS
 --'$libdir/pending', 'recordchange' LANGUAGE C;
 
-CREATE AGGREGATE array_cat_agg(anyarray)(
+CREATE AGGREGATE array_cat_agg(int2[]) (
       sfunc       = array_cat,
-      stype       = anyarray,
+      stype       = int2[],
```
reosarevok, line 13: I'm a bit confused about why we use smallint elsewhere and int2 here, but this file does use int4 so :man_shrugging:

reosarevok (approved):

reosarevok:
I think it makes sense to release this during this schema change, yes (as long as we've tested it properly and it breaks nothing)

upgrade.json
```diff
@@ -111,6 +111,7 @@
                    "20210924-mbs-10327.sql",
                    "20211008-mbs-11903.sql",
                    "20211216-mbs-12140-12141.sql",
-                   "20220309-mbs-12241.sql"]
+                   "20220309-mbs-12241.sql",
+                   "20220426-mbs-112131.sql"]
```
kellnerd, line 115:
```suggestion
                   "20220426-mbs-12131.sql"]
```
I guess you wanted to name the file after the ticket number, right?

mwiencek: Whoops, thanks.  Fixed.

mwiencek:
@brainzbot retest this please

yvanzo (approved):
Tested locally and works for me. :ship:

PR #2509, opened 2022-05-04
MBS-12336: Validate that Predicate::Set is being passed integers

reosarevok:
### Fix MBS-12336

We always want an integer here (they're mostly rowIDs, except for privileges and data quality which are also ints). Default data quality is -1, so we can't restrict to *positive* integers.

Notice from the test that for some reason we semi-supported passing the arguments in a comma-only list ('4,5' rather than two arguments 4 and 5). That *did* get the intended results returned from the DB, but it breaks display (no actual values are shown as selected in the search form) and it seems like a huge hack, so I'm disallowing it with this rather than special-casing it.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/Set.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/Set.pm
@@ -1,6 +1,8 @@
 package MusicBrainz::Server::EditSearch::Predicate::Set;
 use Moose;
 use namespace::autoclean;
+use List::AllUtils qw( any );
+use MusicBrainz::Server::Validation qw( is_integer );
 
 with 'MusicBrainz::Server::EditSearch::Predicate';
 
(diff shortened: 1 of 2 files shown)


reosarevok:

lib/MusicBrainz/Server/EditSearch/Predicate/Set.pm
```diff
@@ -13,6 +14,12 @@ sub operator_cardinality_map {
 
 sub valid {
     my ($self) = @_;
+
+    # If you want to allow non-integer sets, please create ::IntegerSet, etc
+    foreach my $argument ($self->arguments) {
+        return 0 unless is_integer($argument);
+    }
```
mwiencek, line 21: If you want to save one line of code :)
```suggestion
    use List::AllUtils qw( any ); # above
    return 0 if any { !is_integer($_) } $self->arguments;
```

mwiencek (approved):
Didn't test this, but the code/tests look fine and I can't think see any potential problems since all current set predicates are integers.

yvanzo (approved):
LGTMBDNT, suggested changes look good too, :ship: otherwise.

PR #2510, opened 2022-05-04
MBS-12351: Also trim space only disambiguations

reosarevok:
### Fix MBS-12351

The transform was not being applied on spaces-only disambiguation fields, because FormHandler considers it empty. As such, I added validate_when_empty to enforce always running the transform on the comment field.


t/lib/t/MusicBrainz/Server/Controller/Artist/EditRelationships.pm
```diff
@@ -35,6 +35,7 @@ test 'Test adding a relationship' => sub {
         $mech->post('/artist/745c079d-374e-4436-9448-da92dedef3ce/edit', {
             'edit-artist.name' => 'Test Artist',
             'edit-artist.sort_name' => 'Artist, Test',
+            'edit-artist.comment' => 'Yet Another Test Artist',
```
mwiencek, line 38: Hmm, this doesn't seem ideal.  I think it makes sense to ignore the field if it's omitted rather than assume it's being set to an empty string.  (I guess this probably won't cause issues on the website, but might for scripts that are submitting things manually.)

https://metacpan.org/dist/HTML-FormHandler/view/lib/HTML/FormHandler/Manual/Validation.pod says that the form's (and I assume the field's) `validate` method will always be called.  Instead of setting `validate_when_empty`, which seems to force the value to `''`, maybe we can implement a `validate` method instead (which you can set the value from, so we could check if it's set before running `trim_comment` on it).

reosarevok:

mwiencek (approved):
(fixup commit should be squashed when merging)

yvanzo (approved):
LGTMBDNT. Squash :raised_hand_with_fingers_splayed: and merge it!

PR #2511, opened 2022-05-04
MBS-12353: Actually check proposed ratings are allowed

reosarevok:
### Fix MBS-12353

This just checked whether the rating is a number, but that doesn't help avoid ISEs when someone tries to rate, say, 700 instead of just asdfg.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Rating.pm
+++ b/lib/MusicBrainz/Server/Controller/Rating.pm
@@ -3,7 +3,8 @@ use Moose;
 BEGIN { extends 'MusicBrainz::Server::Controller' }
 
 use MusicBrainz::Server::Data::Utils qw( type_to_model );
-use Scalar::Util qw( looks_like_number );
+use MusicBrainz::Server::Translation qw( l );
+use MusicBrainz::Server::Validation qw( is_integer );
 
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Controller/Rating.pm
```diff
@@ -33,8 +34,11 @@ sub rate : Local RequireAuth DenyWhenReadonly
     my $entity_id = $c->request->params->{entity_id};
     my $rating = $c->request->params->{rating};
 
-    unless (looks_like_number($rating)) {
-        $self->error( $c, message => 'rating must be a number', status => 400 );
+    unless (is_integer($rating) && $rating >= 0 && $rating <= 100) {
+        $c->stash(
+            message  => l('Rating should be an integer between 0 and 100'),
```
mwiencek, line 39: sorry if pedantic, but if translating it I think we might as well use a full sentence.  (I see this message is just copied from rating_submit, but it's not translated there so doesn't really matter)
```suggestion
            message  => l('The rating should be an integer between 0 and 100.'),
```

mwiencek (approved):

reosarevok:

PR #2512, opened 2022-05-04
MBS-12354: Check if AC IDs are valid before passing to the DB

reosarevok:
### Fix MBS-12354

This was ISEing when sent whatever other than an integer (someone tried to pass it "undefined"). Copied what we do in Data::Entity.


mwiencek (approved):
Doesn't look like we have any tests for the artist-credit endpoint.  This seems look a good opportunity to add one. :)

reosarevok:
@yvanzo: AFAICT this brought up some sort of bug in AC gid_redirect creation, but I'm not quite sure what it is. When running the tests without any recordings / release groups, the result of 

```
    print Dumper($c->sql->select_list_of_lists(
        'SELECT * FROM artist_credit_gid_redirect'
    ));

```
after the merge is

```
$VAR1 = [
          [
            'c44109ce-57d7-3691-84c8-37926e3d41d2',
            4,
            '2022-06-02 12:39:58.364162+00'
          ],
          [
            // This GID is generated during the merge so it changes each time
            'a475df81-1197-4084-bc09-b4f81d3e91a7', 
            5,
            '2022-06-02 12:39:58.364162+00'
          ],
          [
            '949a7fd5-fe73-3e8f-922e-01ff4ca958f7',
            5,
            '2022-06-02 12:39:58.364162+00'
          ]
        ];
```

But if a recording or RG credited to the "Queen & David Bowie" credit exists, then the result is just:

```
$VAR1 = [
          [
            'c44109ce-57d7-3691-84c8-37926e3d41d2',
            4,
            '2022-06-02 12:42:18.752126+00'
          ]
        ];
```

mwiencek:
The issue is that `_swap_artist_credits` queries the `artist_credit` table (to update `artist_credit_gid_redirect`) after the artist credits are swapped.  If they were the last credits remaining, the related `artist_credit` rows will have been deleted because their ref_counts would be 0.  So this method should probably insert into/update `artist_credit_gid_redirect` before doing anything else.

reosarevok:
I pushed the change. What do you mean with a dedicated test for this, a separate sql file? I don't expect we'd remove these from here, since we now use them for two different tests. IMO leaving a comment to not delete them in the sql file would be a good enough compromise - I've done that now.

yvanzo (approved):
LGTMBDNT, thanks for fixing that bug.

PR #2513, opened 2022-05-10
MBS-4960 / MBS-12361: Allow navigation from donation check page and improve error handling

reosarevok:
### Implement MBS-4960 / MBS-12361

The donation check page can be reached as a tab from your own user page, but didn't itself contain the user header and tab navigation, so the user couldn't continue navigating elsewhere from it as expected. Having the header doesn't add much complexity, so we might as well.

We were just detaching to 500 with no info if hitting the MeB donations endpoint failed. It seems a lot more sensible to give the user a proper warning instead.

### DIFF
--- a/root/account/Donation.js
+++ b/root/account/Donation.js
@@ -9,16 +9,28 @@
 
 import * as React from 'react';
 
+import UserAccountLayout, {
+  sanitizedAccountLayoutUser,
+} from '../components/UserAccountLayout';
 import {CONTACT_URL, DONATE_URL} from '../constants';
-import StatusPage from '../components/StatusPage';
 
 type Props = {
   +days: number,
   +nag: boolean,
+  +user: UnsanitizedEditorT,
 };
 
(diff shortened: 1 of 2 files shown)


root/account/Donation.js
```diff
@@ -9,55 +9,76 @@
...
+  user,
+}: Props): React.Element<typeof UserAccountLayout> => (
+  <UserAccountLayout
+    entity={sanitizedAccountLayoutUser(user)}
+    page="donation"
+    title={l('Donation Check')}
+  >
+    <h2>{l('Donation Check')}</h2>
+    {checkFailed ? (
+      <Warning
```
mwiencek, line 38: This is more like an error than a warning, but not sure we have anything better at the moment.

mwiencek (approved):
Tested by modifying `donation_check` locally.

root/account/Donation.js
reosarevok: It just felt better than a 500 with a message :)

reosarevok:

PR #2514, opened 2022-05-10
MBS-12350 / MBS-12352 / MBS-12383 / MBS-12396 / MBS-12401: Block more "smart" links

reosarevok:
### Implement MBS-12350 / MBS-12352 / MBS-12383 / MBS-12396 / MBS-12401


root/static/scripts/edit/externalLinks.js
```diff
@@ -1568,86 +1568,94 @@ function isValidURL(url: string) {
...
-].map(host => new RegExp('^https?://([^/]+\\.)?' + host + '/.+', 'i'));
+  /* eslint-disable sort-keys */
+  {host: 'adf.ly', blockHost: false},
+  {host: 'album.link', blockHost: false},
+  {host: 'ampl.ink', blockHost: false},
+  {host: 'amu.se', blockHost: false},
+  {host: 'artist.link', blockHost: false},
+  {host: 'band.link', blockHost: false},
+  {host: 'biglink.to', blockHost: false},
+  {host: 'bio.link', blockHost: true},
```
mwiencek, line 1579: It might be better to just add the hosts that are blocked to a Set, and check membership of the set, since the added property doesn't vary much.  (I guess most should probably be true, and the set can be inverted at that point.)

mwiencek (approved):

root/static/scripts/edit/externalLinks.js
reosarevok: Is that significantly faster? (this seems clearer to me, but if it's slower then we could do it the other way).

mwiencek: It's not faster, but the code can minify better.  How many of these do you estimate should actually be `blockHost: false`?  I would guess only a few...

reosarevok: Probably only a few, yes, it's just the annoyance of looking through them all. But maybe I should just do that, and then it's done for the future...

reosarevok: Looked through them in b126ea4ee4ca8cff092b65f48446fd92f516c43b

PR #2515, opened 2022-05-10
MBS-12356: Correctly select + clean up Tidal store pages

reosarevok:
### Fix MBS-12356

The Tidal store is not a streaming page, but a download one. We were automatically dropping the store bit and turning it into a streaming link (not sure if the two always exist, but I wouldn't be too confident, so we might have broken some links). Since it's a legitimate download store, we should deal with it as such.
Like iTunes and others, Tidal seems to have separate stores per country, and removing the country code from a link turns it into a redirect to the user's local store (which might or might not have the content). As such, we should keep the country and indicate it on the sidebar.


mwiencek (approved):

reosarevok:

yvanzo (commented):
Fixing the auto-selected relationship type looks good, but the mixed changes with stores should probably be discussed separately as it isn’t clear it is either needed or an improvement.

Even though Tidal is operating in 61 countries ([source: Tidal](https://support.tidal.com/hc/en-us/articles/202992312-About-TIDAL)), the website remains available under a `.com` domain only and it seems that the two letter-code has no influence on content, it can even be replaced with random letters such as `xz`, to follow the example given in test: <https://store.tidal.com/xz/artist/160>. I initially thought it was related to the website UI language (see [comment to MBS-10621](https://tickets.metabrainz.org/browse/MBS-10621?focusedCommentId=60392&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-60392)) but it doesn’t seem to be the case anymore/either.

reosarevok:
Different stores charge in different currencies, at the very least (https://store.tidal.com/us/album/108715503 is in USD but https://store.tidal.com/ee/album/108715503 is in EUR). Also, this isn't changing any *current* links, since store ones weren't usable anyway, only the new ones :)

yvanzo:
> Different stores charge in different currencies, at the very least (https://store.tidal.com/us/album/108715503 is in USD but https://store.tidal.com/ee/album/108715503 is in EUR).

* The currency is not displayed on these pages for me, maybe because I don’t have an account on that website.
* When I click on “Download” from any of these pages, it redirects me to the same checkout page on the local Tidal store of my current country, in EUR.

reosarevok:
Huh, wtf. Now it's showing me my local store, but when I left that comment it did show it differently. Now that's confusing.

reosarevok:
Well then, rewrote this to remove the country code altogether (which seems to also work and just uses the user's own store) :)

yvanzo (approved):
LGTMBDNT :ship:

I added a comment to the ticket about the extra cleanup for the two-letters code.

PR #2517, opened 2022-05-10
MBS-12347: Also show RG artist in autocomplete if type is null

reosarevok:
### Fix MBS-12347

The artist credit is useful in the inline search results whether or not the release group has a type, so something should still be displayed when there's no `typeName`. I went with just "Release group by {artist}" since that seems the simplest while still not being misleading.


mwiencek (approved):
That string is already translated and already used by root/static/scripts/common/components/Autocomplete2/formatters.js (I guess I fixed it there on accident already), so LGTM.

PR #2518, opened 2022-05-11
MBS-12368: Add annotations, rels and aliases to genre WS

reosarevok:
### Implement MBS-12368

This also allows includes in the non-text genre list, since we had specifically talked about that as the reason we wanted to have limit / offset there while having a names-only text list.

mmd-schema changes (merged): https://github.com/metabrainz/mmd-schema/pull/27

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/2/Genre.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/2/Genre.pm
@@ -12,18 +12,19 @@ use MusicBrainz::Server::Validation qw( is_guid );
 my $ws_defs = Data::OptList::mkopt([
     genre => {
         method   => 'GET',
-        # TODO: Add include when implementing MBS-10062
-        # inc      => [ qw(aliases) ],
+        inc      => [ qw( aliases annotation _relations ) ],
         optional => [ qw(fmt limit offset) ],
         linked   => [ qw( collection ) ],
     },
     genre => {
         method   => 'GET',
+        inc      => [ qw( aliases annotation _relations ) ],
         optional => [ qw(fmt limit offset) ],
     },
     genre => {
         action   => '/ws/2/genre/lookup',
         method   => 'GET',
+        inc      => [ qw( aliases annotation _relations ) ],
         optional => [ qw(fmt) ],
     },
 ]);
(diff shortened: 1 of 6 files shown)


reosarevok:
By now the only perl test failure seems to be the expected validation one because of MMD changes.

yvanzo:
For testing `mmd-schema` is retrieved through `mb-solr` to make sure it returns similar results. So it requires updating `mb-solr` first. It seems worth doing the related changes to `sir` and `mbsssss` at the same time.

reosarevok:
mbsssss: https://github.com/metabrainz/mbsssss/pull/61
sir: https://github.com/metabrainz/sir/pull/136 / https://github.com/metabrainz/sir/pull/137 (needs running scripts)

First try and I might have missed some stuff though :)

petitminion:
hello @reosarevok :) 

I don't know if you got our query in there : https://tickets.metabrainz.org/browse/MBS-12368. So I tried it in here because this seems to have been left in a old milestone and maybe forgotten. 

Do you plan to implement this ? We would love to use it in funkwhale. If we can do anything to help let me know. 

Thanks ! :)

reosarevok:
It's not forgotten, but sadly it cannot be merged until we can update the search server to work alongside it, and that is quite a process since it's still in Python 2 and needs to be migrated... Sorry about the wait!

petitminion:
Okey thanks a lot for answering :) And good luck with the migration of the search server o/

arli2001:
Hi, any news on this pull request? Any help needed?
Thanks a lot :)

mwiencek:
Tests are already running mb-solr v4.1.0 which should have the necessary mmd-schema changes. This doesn't depend on any search changes otherwise.

reosarevok:
Rebased this on our latest code - @mwiencek, IIRC the point was we wanted to wait and do the search bits linked from https://github.com/metabrainz/musicbrainz-server/pull/2518#issuecomment-1180734068 at the same time although to be honest it might make sense to release this if it works and just update search to follow it later, if it's hard to do now.

phw:
@reosarevok I'm all in favor of at least releasing this part. Being able to get genre details in a direct lookup would already be great.

PR #2519, opened 2022-05-12
MBS-11694: Cleanup /intent/user Twitter URLs

reosarevok:
### Implement MBS-11694

These seem to just give the visitor a popup to follow the profile, seem otherwise trivial to clean up to the profile itself.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -4465,6 +4465,10 @@ const CLEANUPS: CleanupEntries = {
         'https://twitter.com/',
       );
       url = url.replace(
+        /^(https:\/\/twitter\.com)\/intent\/user\/?\?screen_name=([^\/?#]+)/,
+        '$1/$2',
+      );
+      url = url.replace(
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -4464,6 +4464,10 @@ const CLEANUPS: CleanupEntries = {
         /^(?:https?:\/\/)?(?:(?:www|mobile)\.)?twitter\.com(?:\/#!)?\//,
         'https://twitter.com/',
       );
+      url = url.replace(
+        /^(https:\/\/twitter\.com)\/intent\/user\?screen_name=([^\/?#]+)/,
```
mwiencek, line 4468: Is it worth handling an optional slash after /user?  Not sure if these are only linked from Twitter or other people can make them too.  (Interestingly, I don't get a popup with the slash added.)
```suggestion
        /^(https:\/\/twitter\.com)\/intent\/user\/?\?screen_name=([^\/?#]+)/,
```

mwiencek (approved):

root/static/scripts/edit/URLCleanup.js
reosarevok: Since both work, we might as well.

PR #2525, opened 2022-05-13
MBS-12042: Allow providing an edit note when editing annotation

reosarevok:
### Implement MBS-12042

Sometimes you want to provide a source for some data you add to the annotation, or a detailed reason why you're changing something; the changelog (which is supposed to be a short thing anyway) does not seem too appropriate for this.
More generally, I do wonder whether we should drop the changelog completely; almost nobody uses it anyway AFAICT as anything else than a place to put sources, which can now be put on the edit note.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2491

### DIFF
--- a/root/annotation/EditAnnotation.js
+++ b/root/annotation/EditAnnotation.js
@@ -9,6 +9,7 @@
 
 import * as React from 'react';
 
+import EnterEditNote from '../components/EnterEditNote';
 import EnterEdit from '../components/EnterEdit';
 import FormRowText from '../components/FormRowText';
 import FormRowTextArea from '../components/FormRowTextArea';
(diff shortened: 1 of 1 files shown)


mwiencek:
> More generally, I do wonder whether we should drop the changelog completely; almost nobody uses it anyway AFAICT as anything else than a place to put sources, which can now be put on the edit note.

TBH I think the changelog should be mandatory, but explained better on the edit page - it should be like a commit title.  Just a few words on what was changed so that we have something useful to display in the revisions table.  (Truncating an edit note, which might just be a list of links, wouldn't be very informative there.)

mwiencek (approved):
This needs to be rebased but it looks very straightforward. :ship:

PR #2526, opened 2022-05-13
Improvements to t/script/CheckSchemaMigration.sh and some documentation

mwiencek:


HACKING.md
```diff
@@ -232,6 +232,121 @@ back the actions in the browser window you have open, so doesn't require any
...
+that first.
+
+To do this, first create an empty "musicbrainz_prod_schema" database with the
+proper roles and privileges. Configure this as `PROD_SCHEMA_CHANGE_TEST`, or
+some other name of your choice, in DBDefs.pm.
+
+```sh
+# You may require custom host/port connection parameters to your local DB.
+psql -U postgres -d template1 -c 'CREATE DATABASE musicbrainz_prod_schema;'
+# It's okay if these already exists.
```
reosarevok, line 256:
```suggestion
# It's okay if these already exist.
```

reosarevok (approved):
This seems very useful! Thanks

HACKING.md
```diff
@@ -232,6 +232,121 @@ back the actions in the browser window you have open, so doesn't require any
...
+```
+
+Next, you can dump the schema of the production standby, import it locally to
+`musicbrainz_prod_schema`, and run upgrade.sh against
+`PROD_SCHEMA_CHANGE_TEST`:
+
+```sh
+ssh pink docker exec postgres-pink sudo -E -H -u postgres \
+  pg_dump --schema-only musicbrainz_db \
+  > musicbrainz_prod_schema.dump
```
yvanzo, line 271: I would suggest to:

- refer to the type of Postgres instance instead of the current host which is changing over time,
- compress the dump for transmission over SSH (`gzip` is the default with `ssh -C`),

as follows:

```suggestion
prod_standby_pg_host=
prod_standby_pg_container=
ssh -C $prod_standby_pg_host docker exec $prod_standby_pg_container sudo -E -H -u postgres \
  pg_dump --schema-only musicbrainz_db \
  | gunzip \
  > musicbrainz_prod_schema.dump
```

yvanzo (approved):
LGTMBDNT. To be tested some day not in a schema change rush.

HACKING.md
mwiencek: I tested the new command locally and couldn't get the compression to work; gunzip would give `gzip: stdin: not in gzip format`, and running just the ssh command (including `-C`) did not return compressed output for me.

I did copy the other improvements you made, though.

yvanzo: Oops, the `gunzip` command is unneeded as ssh is uncompressing it already.

mwiencek: Oh, okay.  I'll just add the `-C` in and push that to master then. :)

yvanzo: For the record, the commit is https://github.com/metabrainz/musicbrainz-server/commit/2cd0f55bcec4ebd3839b488fb379b1be3d52b69e.

PR #2527, opened 2022-05-14
MBS-12372: Implement mood support

reosarevok:
### Implement MBS-12372

This copies genre support, and generally adds mood to any places where we would already have genre. The autoselect list when entering tags for an entity now combines all available genres and moods.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2518


reosarevok:
Turning this to a draft since it requires: 
* @MonkeyDo coming up with a list of moods to implement
* The search server being updated so that the WS bit of this can be put out
* Probably a ton of rebasing and updated to mirror our current genre code rather than the bits we had a year and a half ago

PR #2528, opened 2022-05-15
Add editor_collection_deleted_entity to @PRIVATE_TABLE_LIST

mwiencek:
Although this table is empty and seems to be currently unused (see MBS-8287 for context), it should at least be included in the private dump so that if we start using it a backup exists.


yvanzo (approved):
:ship:

reosarevok (approved):

PR #2530, opened 2022-05-15
Update translations before schema change again

reosarevok:
We added one new line, so we might as well.


yvanzo (approved):
Proof-read the changes to source strings only (`mb-server.pot`).

PR #2531, opened 2022-05-15
Update dbmirror2 MasterSetup.sql with warning fix

mwiencek:
The README should make it self-explanatory that I'm just copying MasterSetup.sql from the dbmirror2 repository.

The only change from there is https://github.com/metabrainz/dbmirror2/commit/7ba642243b34ac22e02fd2a0a15eaf1543386a35.


yvanzo (approved):
:ship:

PR #2532, opened 2022-05-16
Fix echoing the command to be run

yvanzo:
# Problem

`admin/run` prints command arguments only.

# Solution

Save the command itself in a variable before shifting arguments, then use it when printing the command to be run.

# Checklist for author

* [x] Try `admin/run ls /` in a test container
* [x] Try `admin/run tac` in a test container


mwiencek (approved):
Thanks!

One other unrelated issue I noticed: if you run admin/run as root, it will attempt to run the script as musicbrainz for you as expected.  But it will still try to save the log files to /root/log/ (and fail, because the directory doesn't exist).  The log file location should probably be made consistent with the user that the command is invoked under; i.e. I'd expect it to log to /home/musicbrainz/log even if run as root.

PR #2535, opened 2022-05-17
MBS-12385: Add report for digital releases with mail order rels

reosarevok:
### Implement MBS-12385

Similar to non-digital releases with download rels, this should never be correct AFAICT.


lib/MusicBrainz/Server/Report/ReleasesWithMailOrderRelationships.pm
```diff
@@ -0,0 +1,48 @@
...
+    ';
+}
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2013 MetaBrainz Foundation
```
mwiencek, line 42:
```suggestion
Copyright (C) 2022 MetaBrainz Foundation
```

mwiencek (approved):

PR #2536, opened 2022-05-17
JSON dumps: ignore more useless primary/foreign keys

mwiencek:
These cause the incremental JSON dumps process to hang if they're not skipped.

When upgrading the musicbrainz_json_dump DB to schema 27 yesterday, I discovered that it didn't have any foreign keys (or at least all the tables I checked were missing them), which might be why this didn't cause an issue before. It must have had them at some point, and I don't know when or how they were dropped. musicbrainz_json_dump is a replicated mirror, but requires foreign keys in order for the incremental JSON dumps script to find links between tables.


reosarevok (approved):
Looks legit

mwiencek:
@brainzbot test this please

PR #2538, opened 2022-05-18
ExportAllTables.t: remove --nodbmirror2 flags

mwiencek:
Although the tests in ExportAllTables.t don't test dbmirror2 packets (t/script/dbmirror2.t exists for that), it's useful to have them enabled in order to check that having them enabled doesn't interfere with the production of old packets.

Note that removing the flag from ExportAllTables doesn't actually produce v2 packets in this test, because the v1 packets are created by manually inserting rows into the dbmirror_pending* tables, not via replication triggers (so there is no data to be dumped from the dbmirror2 schema).  This is why I've kept the --nodbmirror2 flag on the LoadReplicationChanges invocation: there is no v2 packet to load here.

### DIFF
--- a/script/create_test_db.sh
+++ b/script/create_test_db.sh
@@ -26,6 +26,7 @@ echo "
   DROP SCHEMA IF EXISTS wikidocs CASCADE;
   DROP SCHEMA IF EXISTS sitemaps CASCADE;
   DROP SCHEMA IF EXISTS json_dump CASCADE;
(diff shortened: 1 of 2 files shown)


reosarevok:
That did not seem to work, check tests :)

yvanzo (approved):
LGTM but did not test as exporting with `dbmirror2` has been successfully tested by @mwiencek. :ship:

PR #2539, opened 2022-05-18
Run all sql before enabling replication triggers

mwiencek:
I see no reason why replication triggers should be re-enabled *before* the .standalone.sql file is run.  If the standalone file happens to modify data in replicated tables, then the current order causes those changes to be logged in the dbmirror pending tables, which is not ideal. I've fixed the order so that replication is re-enabled only after all migration SQL is run.

### DIFF
--- a/upgrade.sh
+++ b/upgrade.sh
@@ -97,6 +97,18 @@ then
 fi
 
 ################################################################################
+# Add constraints that apply only to master/standalone (FKS)
+
+if [ "$REPLICATION_TYPE" != "$RT_MIRROR" ]
+then
+    echo `date` : 'Running upgrade scripts for master/standalone nodes'
+    ./admin/psql "$DATABASE" < $SQL_DIR/${NEW_SCHEMA_SEQUENCE}.standalone.sql || exit 1
+
+    echo `date` : Enabling last_updated triggers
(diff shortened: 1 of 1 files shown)


reosarevok (approved):
Seems sensible

yvanzo (commented):
LGTM but is the test `./t/script/CheckSchemaMigration.sh` still passing with this change?

mwiencek:
Yes, I can confirm that CheckSchemaMigration.sh still passes for all three replication types.  I had to modify it since the production branch now has schema 27 merged into it:

```diff
diff --git a/t/script/CheckSchemaMigration.sh b/t/script/CheckSchemaMigration.sh
index e0f05f2e7c..f9d6dbad58 100755
--- a/t/script/CheckSchemaMigration.sh
+++ b/t/script/CheckSchemaMigration.sh
@@ -59,8 +59,8 @@ fi
 
 # DB2 is the schema we want to upgrade from. Here we source it
 # from the production branch.
-git restore --source=production -- admin/sql
-git restore --source=production -- admin/InitDb.pl
+git restore --source=v-2022-04-18 -- admin/sql
+git restore --source=v-2022-04-18 -- admin/InitDb.pl
 ./admin/InitDb.pl --database $DB2 --createdb --clean --reptype $REPLICATION_TYPE "${WITH_PENDING_ARGS[@]}"
 ./admin/psql $DB2 < t/sql/initial.sql
 ./admin/psql $DB2 < admin/sql/SetSequences.sql
```

PR #2540, opened 2022-05-20
Document running Selenium tests that use search

mwiencek:
We should probably also document a way to run these by building a musicbrainz-tests image, since that does all of this for you, in addition to setting up CAA processes for the cover art tests. But this is still useful info, since developers that run selenium tests often will certainly want to run them manually instead of rebuilding a Docker image after every change.


mwiencek:
Guess it can't hurt to merge this, since it's only HACKING documentation and the steps work for me. If someone wants to try them later and report any issues, feel free...

PR #2541, opened 2022-05-25
MBS-12400: fix non-musicbrainz-schema dumps

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12400

Due to the `$table_noschema` addition in https://github.com/metabrainz/musicbrainz-server/commit/0d3866005f7abdf4f1bc59d31ae5a6c4cba4a162, we were failing to include the schema prefix in any dumped tables' file names.  Because MBImport.pl expects the file names to match the table names as specified in `@FULL_TABLE_LIST` (which includes schemas for non-musicbrainz-schema tables), this causes the import to fail at finding data files for those schemas.

The faulty change referenced above also has the side effect of causing tables with the same names in different schemas to clobber each other in the dumps.  For example, the cover-art-archive dump's art_type file would contain rows from the event_art_archive.art_type table instead!

The only purpose of the `$table_noschema` change was to ensure that the file names inside dbmirror v2 packets didn't include the "dbmirror2." prefix.  It wouldn't cause any issue if they did, but it was preferred  to match how dbmirror v1 packets are dumped (whose tables are in the musicbrainz schema, so are not schema-qualified).  I moved the re-mapping from dump_table to ExportAllTables where it's needed.

### DIFF
--- a/admin/ExportAllTables
+++ b/admin/ExportAllTables
@@ -228,7 +228,15 @@ if ($fDoReplication)
         if ($dump_dbmirror2) {
             # NULL columns that are useless on slaves.
             $sql->do('UPDATE dbmirror2.pending_data SET oldctid = NULL, trgdepth = NULL');
-            $mbdump->dump_table($_) for @dbmirror2_tables;
+
+            for my $table (@dbmirror2_tables) {
+                # Map the schema-qualified dbmirror2 table names to file
+                # names without the schema inside the packet.
+                my $table_noschema = $table =~ s/^dbmirror2\.//r;
+                $mbdump->table_file_mapping->{$table} = $table_noschema;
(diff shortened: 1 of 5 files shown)


reosarevok (commented):
The changes seem sensible, but do check the failing test, it seems related.

mwiencek:
I should note that this also fixes not being able to import the caa/eaa tables -- the files for these are expected to be schema-qualified in the dumps, and this fixes that.

mwiencek:
I updated the commit title and description to reflect that this is a general issue with dumping non-musicbrainz-schema tables.  The art_type clobbering is really an obscure side-effect of that (and only discoverable if you try to import these files by hand).

PR #2542, opened 2022-05-25
MBS-12404: Replace Text::Unaccent with Text::Unaccent::PurePerl

mwiencek:
### Fix [MBS-12404](https://tickets.metabrainz.org/browse/MBS-12404)

As the ticket notes, Text::Unaccent is unmaintained and doesn't compile against recent version of Perl [1].  I'm not sure since what version, but the linked bug report is from 2018, and I can attest that I've been having to patch Text::Unaccent manually on my development machines for years.

I'm replacing it with Text::Unaccent::PurePerl, which as its name suggests is written in Perl.  It passes all of our existing `normalise_strings` tests, which contained the only reference to Text::Unaccent.  We use `normalise_strings` mainly for two things:

 * Determining whether a title change is an auto-edit, to allow "simple" changes to go through automatically.
 * Determining whether a disambiguation comment is required when adding a new entity.

[1] https://rt.cpan.org/Public/Bug/Display.html?id=124815


reosarevok (approved):
https://metacpan.org/pod/Text::Unaccent::PurePerl#Comparison-to-Text::Unaccent specifies this follows Text::Unaccent very closely, including:

```
One might expect the following conversions

… → ...  U+2026 HORIZONTAL ELLIPSIS
Œ → OE   U+0152 LATIN CAPITAL LIGATURE OE
œ → oe   U+0153 LATIN SMALL LIGATURE OE
′ → '    U+2032 PRIME
″ → "    U+2033 DOUBLE PRIME
and more, but these aren't implemented in Text::Unaccent,
so they aren't implemented in Text::Unaccent::PurePerl either.
This might change in the future.
```

As such, I think we can reasonably assume that nothing should be different right now. We might want to just use PurePerl exclusively already - do we have a test for our use of it? If we do, and PurePerl works fine, I'd just make the switch directly.

mwiencek:
We have some tests for `normalise_strings` at least, which seem to all pass with the PurePerl module.  I do think it probably makes sense to switch to it fully.  There is always the risk of bugs or small differences in behavior, and I'm not sure the new module is well-maintained either (its last release was in 2014), but at least it's unlikely to break being Perl-only.  The only other options are writing our own or doing all normalization calls through PG (via the `musicbrainz_unaccent` and `regexp_replace` functions).  In the meantime it seems reasonable to merge this to unblock development.

reosarevok:
If you decide to just merge as-is, please add a ticket to revisit further? : )

mwiencek:
> If you decide to just merge as-is, please add a ticket to revisit further? : )

I updated the PR to just replace the module, but also added a ticket.

reosarevok (approved):
All tests pass, the changes seem sensible, happy with a merge if you want.

PR #2543, opened 2022-05-25
MBS-12398: Update release_meta.amazon_asin

mwiencek:
### Fix MBS-12398

This was previously updated by our cover art provider code (despite ASINs being unrelated to cover art, other than Amazon being a provider of both), which was deleted in https://github.com/metabrainz/musicbrainz-server/pull/2379.

I've added the essential code back, refactoring it to disassociate it from the previous cover art implementation.  I've also improved the test coverage.

Note: We'll have to run a script to fix existing columns.  This one is probably still usable: admin/sql/updates/20130724-refresh-release_meta-asin.sql


reosarevok (approved):
Seems good - only thing, if you can be bothered, could you document the tests a bit more? :) For example:

```
is($release1->amazon_asin, 'B00005CDNG');
```
could become
```
is(
    $release1->amazon_asin,
    'B00005CDNG',
    'ASIN was restored after changing the link type back',
);
```

to make it easier to figure out what the test is doing when using `-lv`

mwiencek:
Good point, should be fixed now.

yvanzo (approved):
LGTMBDNT :ship: Tests coverage and description looks particularly good.

PR #2544, opened 2022-05-26
Update react, react-dom to v18

mwiencek:
Please see https://reactjs.org/blog/2022/03/29/react-v18.html and https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html.

 * Support for IE11 has been dropped from React. As it's EOL, I've dropped it from our Babel config too.

 * flow-typed definitions for react-dom have been updated.

 * References to `ReactDOM.render` and `ReactDOM.unmountComponentAtNode` have been updated to use the new `ReactDOMClient.createRoot` API.

 * As there should only be one React root created per DOM element, any code that renders a React root more than once now saves the root onto the DOM element via jQuery's data API.

 * Callbacks that were previously passed to `ReactDOM.render` (to run after rendering was completed) are now passed as refs to the component root. See "What about the render callback?" on https://github.com/reactwg/react-18/discussions/5.

 * Code that used the return value of `ReactDOM.render` to get a reference to the *component class instance* now obtain the instance also by using refs.

 * `ReactDOM.hydrate` has been replaced with `ReactDOMClient.hydrateRoot`. StrictMode has been enabled on all hydrated elements.

 * I did not change any use of `ReactDOMServer.renderToString` in the template renderer or elsewhere. The new `renderToPipeableStream` API is *not* currently usable by us, because with our current architecture it doesn't make sense to stream HTML back to Perl.


reosarevok:

root/static/scripts/edit/utility/guessFeat.js
```diff
@@ -241,7 +241,7 @@ export default function guessFeat(entity) {
     names: artistCredit.concat(match.artistCredit),
   });
 
-  entity.artistCreditEditorInst && entity.artistCreditEditorInst.setState({
+  entity.artistCreditEditorInst?.current?.setState({
```
reosarevok, line 244: I didn't find docs for the "current" thing, what's the change? Or did you just decide to do this to simplify stuff?

reosarevok (commented):
Some quick testing around didn't find anything broken. Not sure why Perl tests never ran here?

I see `renderToString` is _discouraged_, which thankfully is not the same as deprecated. But do you feel there's a worry that'll change eventually? Is there a way to support the new APIs while still having Perl in the backend?

mwiencek:
> Some quick testing around didn't find anything broken. Not sure why Perl tests never ran here?

There was a Flow error that prevented them from running (should be fixed now).

> I see renderToString is discouraged, which thankfully is not the same as deprecated. But do you feel there's a worry that'll change eventually? Is there a way to support the new APIs while still having Perl in the backend?

I'm not sure if they plan to deprecate it, but I'm not too worried about it.  Right now it's "discouraged" because it "has limited Suspense support and does not support streaming."  We use neither of those features.

It should be possible to use streaming in some cases in the future.  Currently the Perl expects a JSON response from the template renderer, so it needs the full response to be able to parse it.  (The JSON includes HTML and a response code.)  For TT templates that use `React.embed`, streaming doesn't make sense because you can't stream into an executing TT template -- it requires a full string to embed synchronously.  Streaming could make sense where we're rendering a full page, but we'll have to change how data is sent back.

root/static/scripts/edit/utility/guessFeat.js
mwiencek: Those are refs: https://reactjs.org/docs/refs-and-the-dom.html (they're not DOM specific though, you can store any value in them).

A ref is just an object with a single property, `current`, that holds some value.  The idea is that the actual ref doesn't change (`entity.artistCreditEditorInst` stays constant), but the value stored in `current` can change.

mwiencek: We used to get the component instance like `entity.artistCreditEditorInst = ReactDOM.render(...)`, which is no longer supported in the new API.  Now we basically do this:

```js
entity.artistCreditEditorInst = React.createRef();

// ...

<ArtistCreditEditor ref={entity.artistCreditEditorInst} {...props} />
```

reosarevok: Oh, great, I didn't follow that the artistCreditEditorInst thing was being created differently on a second file :)

mwiencek:
The Selenium failure was apparently legitimate, but should be fixed now.

mwiencek:
I'll merge this then, and if something breaks I promise to add a test for it. :angel:

PR #2546, opened 2022-05-30
MBS-12280 (II): Add target _blank to more edit form links

reosarevok:
### Fix MBS-12280 (even more)

Somehow I missed quite a few last time!


yvanzo (approved):
LGTMBDNT :ship:

Just an extra note that could be added on merge/squash: `noopener`and `noreferrer` wouldn’t make sense here as these links are internal to the MB website.

reosarevok:
Added your comment on squash

PR #2547, opened 2022-05-30
MBS-12412: Fail gracefully if requesting an invalid ID in edit-cover-art 

reosarevok:
### Fix MBS-12412

This would ISE if the user requested an ID that could not be found by the `first{}` call, since we later tried to access `$artwork->types`.

This fails more gracefully now, like we already did when the release had no artwork at all; I also added specific messages for the 404 page for both cases.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -599,10 +599,14 @@ sub edit_cover_art : Chained('load') PathPart('edit-cover-art') Args(1) Edit {
 
     my @artwork = @{
         $c->model('Artwork')->find_by_release($entity)
-    } or $c->detach('/error_404');
+    } or $c->detach('/error_404', [ l('This release has no artwork.') ]);
+
     $c->model('CoverArtType')->load_for(@artwork);
 
-    my $artwork = first { $_->id == $id } @artwork;
+    my $artwork = first { $_->id == $id } @artwork or $c->detach(
+        '/error_404',
+        [ l('Found no artwork with ID “{id}”.', { id => $id }) ],
+    );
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
Is it not possible to pass message through subroutine argument? Or is there another reason for using the stash instead?

LGTMBDNT anyway :ship:

reosarevok:
Honestly, I'm not sure, maybe you can pass the message as a second argument to detach(). It seemed easier to just make it work like the error 400 that already supported passing a message like this, that's why I changed it.

mwiencek:
> Is it not possible to pass message through subroutine argument?

It's possible: you pass them as an array ref in the second argument (same as `forward`):

```suggestion
        $c->detach('/error_404', [ l('This release has no artwork.') ]);
```

yvanzo:
It raises a broader issue of lower importance which doesn’t even matter for this PR as `error_400` proved that stashed messages are working too: Which approach to use? The most efficient one, if any? Or are there other considerations?

mwiencek:
Well, `stash` is basically global variables.  Depending on how you're using it, you risk overwriting stash values with the same key, and it can make the code harder to follow because it's not always clear where the stash value is coming from.  So passing the arguments directly will always be clearer in that regard.  It's not much of an issue here, since we're detaching to a 404 page (so don't care about any existing values).  But the existing code already supported passing a message via arguments, so I think it'd make sense to just use that.

reosarevok:
Ok, reverted that bit and passed the messages directly then.

yvanzo (approved):
Still LGTMBDNT :shipit:

mwiencek (approved):
lgtmbdnt

PR #2552, opened 2022-05-31
MBS-12417: Update Soundcloud cleanup to remove ? parameters

reosarevok:
### Implement MBS-12417

This makes an exception for search/tags pages since we want to keep the ? there to drop them more easily. Otherwise, these seem to be all tracking fluff.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -4083,7 +4083,9 @@ const CLEANUPS: CleanupEntries = {
       },
     ],
     clean: function (url) {
-      return url.replace(/^(https?:\/\/)?((www|m)\.)?soundcloud\.com(\/#!)?/, 'https://soundcloud.com');
+      url = url.replace(/^(https?:\/\/)?((www|m)\.)?soundcloud\.com(\/#!)?/, 'https://soundcloud.com');
+      url = url.replace(/^(https:\/\/soundcloud\.com\/)(?!(?:search|tags)[\/?#])([^?]+).*$/, '$1$2');
+      return url;
     },
     validate: function (url) {
       return {
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -4099,11 +4099,6 @@ limited_link_type_combinations: [
   },
   // SoundCloud
   {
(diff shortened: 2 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -4083,7 +4083,9 @@ const CLEANUPS: CleanupEntries = {
       },
     ],
     clean: function (url) {
-      return url.replace(/^(https?:\/\/)?((www|m)\.)?soundcloud\.com(\/#!)?/, 'https://soundcloud.com');
+      url = url.replace(/^(https?:\/\/)?((www|m)\.)?soundcloud\.com(\/#!)?/, 'https://soundcloud.com');
+      url = url.replace(/^(https:\/\/soundcloud\.com\/)(?!(?:search|tags)[\/?#])([^?]+)(?:\?.*)?$/, '$1$2');
```
yvanzo, line 4087: Correct but can be simplified as follows since the [quantifier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers) `+` is greedy (matching the maximum number of times) by default:

```suggestion
      url = url.replace(/^(https:\/\/soundcloud\.com\/)(?!(?:search|tags)[\/?#])([^?]+).*$/, '$1$2');
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4121,6 +4121,12 @@ limited_link_type_combinations: [
     expected_relationship_type: 'soundcloud',
             expected_clean_url: 'https://soundcloud.com/alisonwonderland',
   },
+  {
+                     input_url: 'https://soundcloud.com/erin-thomson-776648435?utm_source=clipboard&utm_medium=text&utm_campaign=social_sharing',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'soundcloud',
+            expected_clean_url: 'https://soundcloud.com/erin-thomson-776648435',
+  },
```
yvanzo, line 4129: The earlier test with `metro-luminal` might be no longer needed as it’s quite redundant with this one.

reosarevok:
Changes made

yvanzo (approved):

yvanzo:
:ship:

PR #2553, opened 2022-05-31
MBS-9188: Improve LinkedIn URL cleanup

reosarevok:
### Implement MBS-9188

The subdomain is apparently just for display language and whatnot but does not change the content nor block it, so we should
standardize to the www. default.
It seems that no ? parameters are useful (seem to be for tracking only) so we should drop them too.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2731,7 +2731,7 @@ const CLEANUPS: CleanupEntries = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?linkedin\\.com/', 'i')],
     restrict: [LINK_TYPES.socialnetwork],
     clean: function (url) {
-      return url.replace(/^https?:\/\/([^/]+\.)?linkedin\.com/, 'https://$1linkedin.com');
+      return url.replace(/^https?:\/\/(?:[^/]+\.)?linkedin\.com\/([^?#]+).*$/, 'https://www.linkedin.com/$1');
     },
   },
   'livefans': {
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -2727,12 +2727,13 @@ limited_link_type_combinations: [
                      input_url: 'http://www.linkedin.com/pub/trevor-muzzy/5/282/538',
              input_entity_type: 'artist',
     expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.linkedin.com/pub/trevor-muzzy/5/282/538',
   },
   {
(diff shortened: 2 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2731,7 +2731,8 @@ const CLEANUPS: CleanupEntries = {
     match: [new RegExp('^(https?://)?([^/]+\\.)?linkedin\\.com/', 'i')],
     restrict: [LINK_TYPES.socialnetwork],
     clean: function (url) {
-      return url.replace(/^https?:\/\/([^/]+\.)?linkedin\.com/, 'https://$1linkedin.com');
+      url = url.replace(/^https?:\/\/(?:[^/]+\.)?linkedin\.com\/([^?#]+)(?:[?#].*)?$/, 'https://www.linkedin.com/$1');
```
yvanzo, line 2734: Correct but can be simplified as follows since the [quantifier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers) `+` is greedy (matching the maximum number of times) by default:

```suggestion
      url = url.replace(/^https?:\/\/(?:[^/]+\.)?linkedin\.com\/([^?#]+).*$/, 'https://www.linkedin.com/$1');
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2734,6 +2734,12 @@ limited_link_type_combinations: [
     expected_relationship_type: 'socialnetwork',
             expected_clean_url: 'https://www.linkedin.com/in/legselectric',
   },
+  {
+                     input_url: 'https://ca.linkedin.com/in/didier-charette-0630b1b6?original_referer=https%3A%2F%2Fduckduckgo.com%2F',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'socialnetwork',
+            expected_clean_url: 'https://www.linkedin.com/in/didier-charette-0630b1b6',
+  },
```
yvanzo, line 2742: * The previous test with `legselectric` might be no longer needed as it’s quite redundant with this one.
* The earlier test with `trevor-muzzy` would be worth an `expected_relationship_type`.

reosarevok: It already has `expected_relationship_type` so I'm going to assume you meant `expected_clean_url`?

reosarevok:
Updated

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: Yes, thank you!

yvanzo (approved):
LGTMBDNT 🚢

PR #2554, opened 2022-05-31
MBS-12376: Don't show spammers on area pages

reosarevok:
### Implement MBS-12376

When searching for users in a specific area, it's probably best not to show "Online Medical Marijuana Card in DC" or whatnot.


yvanzo (approved):
LGTMBDNT. The tests coverage seems to be good, even with contained area.

It will hide spammers from the users tab of areas, even to admins. Please update the ticket accordingly if that is intentional.

reosarevok:
Left a comment specifying I'm also blocking them for admins.

PR #2555, opened 2022-05-31
MBS-12418: Also format artists in setlists if there's no MBID link

reosarevok:
### Implement MBS-12418:

I think having no "Artist:" nor `<strong>` for artist lines unless there is an MBID link was actually not really intended, and in any case it seems silly - it means that unless a link is added, there's literally no point in adding `@ ` vs `* ` in the setlist.
This makes it so that artist lines are always displayed starting with "Artist: " and in bold.

### DIFF
--- a/root/static/scripts/common/utility/formatSetlist.js
+++ b/root/static/scripts/common/utility/formatSetlist.js
@@ -25,18 +25,22 @@ function setlistLink(entityType, entityGid, content) {
 const linkRegExp =
   /^\[([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})(?:\|([^\]]+))?\]/i;
 
-function formatSetlistArtist(entityGid, content) {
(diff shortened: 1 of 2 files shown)


root/static/scripts/common/utility/formatSetlist.js
```diff
@@ -25,18 +25,22 @@ function setlistLink(entityType, entityGid, content) {
 const linkRegExp =
   /^\[([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})(?:\|([^\]]+))?\]/i;
 
-function formatSetlistArtist(entityGid, content) {
+function formatSetlistArtist(content, entityGid = undefined) {
```
mwiencek, line 28: Parameters always default to undefined if you don't pass them, so `entityGid = undefined` doesn't actually do anything, but if you want to indicate that it's optional you can make that explicit with Flow:
```suggestion
function formatSetlistArtist(content: string, entityGid?: string) {
```

mwiencek (approved):

PR #2556, opened 2022-06-01
MBS-12420: Support comma-separated IDs for edit search types again

reosarevok:
### Fix MBS-12420

This was removed as a hack, but it was actually intended; it was only intended for edit type searches though, where the same edit type can have one current and one or more historical IDs and we want to search them as a group.
As such, rather than re-allow it for every set, I made a separate predicate for edit types. This still makes sure that either integers or comma-separated integers are passed.


mwiencek (approved):
Works for me locally.  Added tests look good.

PR #2557, opened 2022-06-01
MBS-12415 / MBS-12416: Limit overenthusiastic overflow-wrap

reosarevok:
### Fix MBS-12415 / MBS-12416

The main intention of the `overflow-wrap` changes was to wrap absurdly long entity names, and very long words in edit notes. By setting it everywhere, it was also affecting things like the tag button on the sidebar, or barcodes in entity lists, which does not seem useful. As such, this sets it only on `<a>` and on edit notes.

It also explicitly unsets it from `<a>` in the header. Wrapping here would make sense if it avoided scrolling, but it does not - it just makes the headers harder to read without adding anything remotely responsive. As such, let's stop the wrapping for now and recheck this once we have implemented the responsive header that is in the works.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -158,6 +158,11 @@ pre code {
     margin-right: 16px;
     clear: both;
 
+    a {
+        // So menu links don't wrap (TODO: review with responsive header)
+        overflow-wrap: normal;
+    }
+
     img.logo {
         width: @logo-width;
         height: @logo-height;


mwiencek (approved):
This appears to fix both of the linked issues for me locally (though I could only reproduce them in Firefox -- guess that's why I didn't notice them before).  Hopefully it doesn't turn out we need more exceptions for `a`.

PR #2558, opened 2022-06-01
MBS-12393: Change TOWER RECORDS to all-caps as per store Japanese usage

reosarevok:
### Implement MBS-12393

That's what they call themselves, so it makes sense to use it.


yvanzo (approved):
⛵

PR #2559, opened 2022-06-02
Use flushSync to avoid render lag (fixes MBS-12424)

mwiencek:
React v18 uses a new method of rendering that can be interrupted and doesn't block the page from performing other work. This can make updates appear laggy where we're trying to render many successive roots -- there is a noticeable delay between rendering each root where React yields its time. (In the case of MBS-12424, each "root" would be a track artist credit container.) This isn't really a problem with React, but a side effect of how we're gluing it into non-React code. Normally we wouldn't have so many roots.

In order to make our haphazard way of rendering things into non-React pages appear more snappy, I'm wrapping all calls to `render` and `hydrateRoot` with `flushSync`, which causes all pending updates to the DOM to occur immediately. This keeps that behavior in line with how React v17 worked.

Wrapping `hydrateRoot` in particular may affect the timing of when user scripts are allowed to interact with the page and avoid hydration errors.

### DIFF
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -9,6 +9,7 @@
 import he from 'he';
 import $ from 'jquery';
 import ko from 'knockout';
+import {flushSync} from 'react-dom';
 import * as ReactDOMClient from 'react-dom/client';
 
 import AddEntityDialog, {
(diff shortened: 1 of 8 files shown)


reosarevok:
Is this intended to be permanent, or to be removed once we don't use React embeds in TT files anymore?

https://reactjs.org/docs/react-dom.html#flushsync says "flushSync can significantly hurt performance. Use sparingly." so it makes it sound less than ideal.

mwiencek:
> Is this intended to be permanent, or to be removed once we don't use React embeds in TT files anymore?

I'd suggest permanent for the `hydrateRoot` call, to keep the timing of hydration consistent for user scripts.  The ones around the `render` calls wouldn't be permanent, because eventually those would go away and just become part of a normal React component tree (with only one root).

> https://reactjs.org/docs/react-dom.html#flushsync says "flushSync can significantly hurt performance. Use sparingly." so it makes it sound less than ideal.

It defeats the purpose of the concurrency features they developed for React 18.  It'd be an issue if within a normal component we were wrapping a bunch of `setState` or `dispatch` calls in `flushSync`, since that's not letting React batch updates together or interrupt rendering if needed.  For `render` calls, this just reverts back to the React v17 behavior, and it's not an area we can use concurrency at all (since we're mixing it with knockout).

mwiencek:
Oops, that was my intent.  Fixed the PR to target beta instead.

mwiencek:
@brainzbot test this please

PR #2561, opened 2022-06-07
MBS-12437 / MBS-12438: Limit overflow-wrap even more

reosarevok:
### Fix MBS-12437 / MBS-12438

Still a bunch of things are wrapping that probably should not, so let's do this the other way around and start from the minimum
amount of wrapped things. This wraps only entity headers (where anything too long will break the interface for no good use), and URL entity links, which can get really long, both on EntityLink and in the Edit URL edit display.
Edit note texts and tags on tag lists already wrap in what seems to be a useful way without complaints, so leaving those in too.

Examples that *should* still wrap: 
* https://musicbrainz.org/edit/78435051 (edit URL display with a very long URL)
* https://musicbrainz.org/edit/87598821 (remove URL rel display, with long URL that has been removed)
* https://musicbrainz.org/release/77b12c1f-9fa7-4985-b291-9b717b7454b8 (release page with an absurdly long one-word title)
* https://musicbrainz.org/edit/31405687 (edit with a crazy long one-word edit note)

Examples that no longer wrap, but do in beta:
* https://musicbrainz.org/artist/9f16acb6-58de-4690-82dd-14771cfc3d08/recordings
* https://musicbrainz.org/edit/90192268

### DIFF
--- a/root/components/EntityHeader.js
+++ b/root/components/EntityHeader.js
@@ -35,7 +35,7 @@ const EntityHeader = ({
   subHeading,
 }: Props): React.Element<typeof React.Fragment> => (
   <>
-    <div className={headerClass}>
+    <div className={'entityheader ' + headerClass}>
       {preHeader || null}
       <h1>
         {heading || <EntityLink entity={entity} />}
--- a/root/edit/details/EditUrl.js
+++ b/root/edit/details/EditUrl.js
@@ -38,7 +38,7 @@ const EditUrl = ({edit}: Props): React.Element<typeof React.Fragment> => {
           <tr>
             <th>{addColonText(l('URL'))}</th>
             <td className="old">
-              <a href={uri.old}>
+              <a className="url-entity-link" href={uri.old}>
                 <DiffSide
                   filter={DELETE}
                   newText={uri.new}
@@ -47,7 +47,7 @@ const EditUrl = ({edit}: Props): React.Element<typeof React.Fragment> => {
               </a>
             </td>
             <td className="new">
-              <a href={uri.new}>
+              <a className="url-entity-link" href={uri.new}>
                 <DiffSide
                   filter={INSERT}
                   newText={uri.new}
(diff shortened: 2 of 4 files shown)


yvanzo (commented):
I could test the added examples locally through adaptive view.

The two examples that no longer wrap are not regressions compared to production at least.

It isn’t clear what is reverted and what is new. Splitting into two separate commits can help with.
It isn’t obvious either which previous commit(s) or pull request(s) are referred to.

LGTM otherwise.

reosarevok:
Originally, https://github.com/metabrainz/musicbrainz-server/pull/2516 made it so the whole site is allowed to wrap, but that was too messy. Then (the version currently in beta, https://github.com/metabrainz/musicbrainz-server/pull/2557), every `<a>` was allowed to wrap. Now, only the specific cases I wrote down on the commit message are allowed to wrap, with everything else behaving like it used to. It's not ideal, since it will still cause issues with super long words in places like tables, but it's the best I could come up with without actively checking for long words on render and adding a class to cause wrapping only if they are present (which seems very inefficient).

yvanzo:
Thanks for the references and details but it is not for me, please add these to the commit message and :ship: it!

root/components/EntityHeader.js
```diff
@@ -35,7 +35,7 @@ const EntityHeader = ({
   subHeading,
 }: Props): React.Element<typeof React.Fragment> => (
   <>
-    <div className={headerClass}>
+    <div className={'entityheader ' + headerClass}>
```
mwiencek, line 38: any reason url-entity-link is hyphenated but entityheader isn't?

mwiencek (approved):
Fixes the issues in the tickets for me and I like the new approach.

root/components/EntityHeader.js
mwiencek: OK, just wondering. :)  No need to fix them here.

PR #2562, opened 2022-06-08
MBS-12395: Report for videos in mediums that shouldn't support video

reosarevok:
### Implement MBS-12395

This will help find both videos that have been wrongly reused in audio-only releases, and stuff like video CDs marked as just "CD". I hand-picked a list of formats that do not support video rather than a list of those that can because I'd rather have missing entries if we add a new format that doesn't support video than false positives if we add a new format that does.

### DIFF
--- a/lib/MusicBrainz/Server/Report/VideosInNonVideoMediums.pm
+++ b/lib/MusicBrainz/Server/Report/VideosInNonVideoMediums.pm
@@ -1,0 +1,91 @@
+package MusicBrainz::Server::Report::VideosInNonVideoMediums;
+use Moose;
+
+with 'MusicBrainz::Server::Report::RecordingReport',
+     'MusicBrainz::Server::Report::FilterForEditor::RecordingID';
+
+my $NON_VIDEO_FORMATS = join(',', (
+    1, # CD
+    3, # SACD
+    6, # MiniDisc
+    7, # Vinyl
+    8, # Cassette
+    10, # Reel-to-reel
+    14, # Wax Cylinder
+    15, # Piano Roll
+    16, # DCC
+    25, # HDCD
+    29, # 7" Vinyl
+    30, # 10" Vinyl
+    31, # 12" Vinyl
+    34, # 8cm CD
+    36, # SHM-CD
+    37, # HQCD
+    38, # Hybrid SACD
+    42, # Enhanced CD
+    44, # DTS CD
+    50, # Edison Diamond Disc
+    51, # Flexi-disc
+    52, # 7" Flexi-disc
+    53, # Shellac
+    54, # 10" Shellac
+    55, # 12" Shellac
+    56, # 7" Shellac
+    57, # SHM-SACD
+    58, # Pathé disc
+    61, # Copy Control CD
+    63, # Hybrid SACD (CD layer)
+    64, # Hybrid SACD (SACD layer)
+    67, # DualDisc (CD side)
+    70, # DVDplus (CD side)
+    73, # Phonograph record
+    74, # PlayTape
+    75, # HiPac
+    81, # VinylDisc (Vinyl side)
+    82, # VinylDisc (CD side)
+    83, # Microcassette
+    84, # SACD (2 channels)
+    85, # SACD (multichannel)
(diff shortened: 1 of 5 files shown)


lib/MusicBrainz/Server/Report/VideosInNonVideoMediums.pm
```diff
@@ -0,0 +1,91 @@
+package MusicBrainz::Server::Report::VideosInNonVideoMediums;
+use Moose;
+
+with 'MusicBrainz::Server::Report::RecordingReport',
+     'MusicBrainz::Server::Report::FilterForEditor::RecordingID';
+
+my @NON_VIDEO_FORMATS = join(',', (
```
yvanzo, line 7: `join` is returning a string, so it doesn’t need to be assigned to an array.
```suggestion
my $NON_VIDEO_FORMATS = join(',', (
```

lib/MusicBrainz/Server/Report/VideosInNonVideoMediums.pm
```diff
@@ -0,0 +1,91 @@
...
+            r.name AS rname
+        FROM
+            recording r
+            JOIN artist_credit ac ON r.artist_credit = ac.id
+            JOIN track t ON t.recording = r.id
+            JOIN medium m ON t.medium = m.id
+        WHERE
+            r.video IS TRUE
+            AND t.is_data_track IS FALSE
+            AND m.format IN (@NON_VIDEO_FORMATS)
```
yvanzo, line 75: To follow the above comment:

```suggestion
            AND m.format IN ($NON_VIDEO_FORMATS)
```

yvanzo (commented):
The code looks good to me, except a minor comment about.

:+1: for using “a list of formats that do not support video” (rather than a list of those that can) which I just skimmed. These formats should probably be mentioned in the ticket for advanced editors to double-check.

It is not obvious what to do about these recordings. Splitting these and linking via the Recording-Recoding [music video](https://musicbrainz.org/relationship/ce3de655-7451-44d1-9224-87eb948c205d) relationship? Maybe `extraInfo` would be useful here.

Did not test this report otherwise. Has it been tested?

reosarevok:
Fair enough about the lack of guidance - I added some, hopefully it's clear :) 

Yes, I did test it with sample data and got 43 hits that seemed legit.

root/report/VideosInNonVideoMediums.js
```diff
@@ -0,0 +1,46 @@
...
+       at least one medium that does not support videos.`,
+    )}
+    entityType="recording"
+    extraInfo={exp.l(
+      `There are two main possibilities here: either the recording being
+       marked as a video is correct, but the format is not (a CD should be a
+       VCD, for example), or the recording is being used for both a video and 
+       and audio-only recording, in which case the two should be split since
+       video recordings should always be separate.`,
+    )}
```
yvanzo, line 36: Extra guidance if the two recordings should always be linked through a music video relationship:

```suggestion
    extraInfo={exp.l(
      `There are two main possibilities here: either the recording being
       marked as a video is correct, but the format is not (a CD should be a
       VCD, for example), or the recording is being used for both a video and 
       and audio-only recording, in which case the two should be split since
       video recordings should always be separate and linked through a
       {doc_link|music video relationship}.`,
      {doc_link: '/relationship/ce3de655-7451-44d1-9224-87eb948c205d'},
    )}
```

yvanzo (approved):
Thanks for all!

A couple more open questions about editing:
Should the two recordings be linked through a music video relationship?
Or does it depend on the editing history and the possibly attached AcoustIDs?

root/report/VideosInNonVideoMediums.js
reosarevok: Only if it's a video for the same audio, I guess :) It's probably often the case, but not always - I will mention the relationship, anyway, that's a good point.

yvanzo: Your [changes](https://github.com/metabrainz/musicbrainz-server/compare/d2e84139ed368bb0759488d35f23edbd163f5ecc..bf30f276b118774508b73179bbf4544a9c6c28a9) look good, thanks!

PR #2564, opened 2022-06-14
MBS-12453 / MBS-12455 / MBS-12456: Annotation react fixes

reosarevok:
This fixes two issues (including MBS-12453) with the React implementation (which weren't problems with TT because TT just didn't care), and two issues (MBS-12455 and MBS-12456) that already existed before the React move. See commit comments for more details.

Example of empty annotation revision page:
Screenshot from 2022-06-14 13-46-26

### DIFF
--- a/root/annotation/AnnotationComparison.js
+++ b/root/annotation/AnnotationComparison.js
@@ -50,8 +50,8 @@ const AnnotationComparison = ({
           <p>
             <DiffSide
               filter={DELETE}
-              newText={newAnnotation.text}
-              oldText={oldAnnotation.text}
+              newText={newAnnotation.text ?? ''}
+              oldText={oldAnnotation.text ?? ''}
               split="\s+"
             />
           </p>
@@ -59,8 +59,8 @@ const AnnotationComparison = ({
           <p>
             <DiffSide
               filter={INSERT}
-              newText={newAnnotation.text}
-              oldText={oldAnnotation.text}
+              newText={newAnnotation.text ?? ''}
+              oldText={oldAnnotation.text ?? ''}
               split="\s+"
             />
           </p>
--- a/root/static/scripts/annotation/AnnotationHistoryTable.js
+++ b/root/static/scripts/annotation/AnnotationHistoryTable.js
@@ -21,8 +21,8 @@ type Props = {
 
 /* eslint-disable flowtype/sort-keys */
 type ActionT =
-  | {+type: 'update-new', +annotationId: number}
-  | {+type: 'update-old', +annotationId: number}
+  | {+type: 'update-new', +index: number}
+  | {+type: 'update-old', +index: number}
 ;
 /* eslint-enable flowtype/sort-keys */
 
@@ -33,10 +33,11 @@ type StateT = {
 
 type WritableStateT = {...StateT}; // this has writable properties
 
-function createInitialState(annotations) {
+function createInitialState() {
   return {
-    selectedNew: annotations[0]?.id ?? 0,
-    selectedOld: annotations[1]?.id ?? 0,
+    // These default indices are only used if canCompare is true below
+    selectedNew: 0,
+    selectedOld: 1,
   };
 }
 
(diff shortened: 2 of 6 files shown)


root/static/scripts/annotation/AnnotationHistoryTable.js
```diff
@@ -33,22 +33,22 @@ type StateT = {
...
 
-function createInitialState(annotations) {
+function createInitialState() {
   return {
-    selectedNew: annotations[0]?.id ?? 0,
-    selectedOld: annotations[1]?.id ?? 0,
+    selectedNew: 0,
+    selectedOld: 1,
   };
 }
```
yvanzo, line 41: The bug and the example mentioned in commit https://github.com/metabrainz/musicbrainz-server/pull/2564/commits/1032bd8a7a182edbf7920585b694cdd7a8042c2a make perfect sense, but I don’t understand this particular change.

root/static/scripts/annotation/AnnotationHistoryTable.js
```diff
@@ -101,8 +101,9 @@ const AnnotationHistoryTable = ({
                 <td>
                   <input
                     className="old"
+                    data-index={index}
                     defaultChecked={index === 1}
-                    disabled={annotation.id >= state.selectedNew}
+                    disabled={index <= state.selectedNew}
```
yvanzo, line 106: Same here and below.

root/annotation/AnnotationComparison.js
```diff
@@ -50,17 +50,17 @@ const AnnotationComparison = ({
...
             />
           </p>
           <h3>{l('New annotation')}</h3>
           <p>
             <DiffSide
               filter={INSERT}
-              newText={newAnnotation.text}
-              oldText={oldAnnotation.text}
+              newText={newAnnotation.text ?? ''}
+              oldText={oldAnnotation.text ?? ''}
```
yvanzo, line 63: The displayed diff is not the same as it currently is in production. Why? Is it intended? See https://musicbrainz.org/release/225f760b-f8d8-47a3-b74c-a3ca1a0c0e96/annotation/539363 for example.

yvanzo (commented):
Tested every commit by following the examples provided either in commit message or in ticket’s description (added examples when missing). It works fine locally. Just have a few open questions.

root/static/scripts/annotation/AnnotationHistoryTable.js
reosarevok: The whole idea is that rather than the annotation id (which doesn't guarantee sorting by time, probably because of some pre-NGS issues), we just use the index of the annotation in the list, since that's sorted by date anyway. Seemed easier than actually doing the date comparisons.

root/annotation/AnnotationComparison.js
reosarevok: It is, it's the same as it would be currently in beta :) See https://beta.musicbrainz.org/release/225f760b-f8d8-47a3-b74c-a3ca1a0c0e96/annotations-differences?new=539363&old=276825

The reasoning is in https://github.com/metabrainz/musicbrainz-server/pull/2491/commits/1002579308b4a1699331f3956492833e26f59fff

root/static/scripts/annotation/AnnotationHistoryTable.js
yvanzo: I had it, but it assumes that there are at least two annotations, whereas the previous code had fallback values in this case. See https://musicbrainz.org/release/1a80e304-aebb-4e2c-9165-ae30eec21c6d/annotations for example.

reosarevok: It does, but it seemed irrelevant because there's no selection mechanism unless there's `canCompare` :) But I could always pass `canCompare` and default both to 0 if that's `false` if you feel that's better? We could not even create the state at all in that case, I guess, assuming flow won't mind...

root/static/scripts/annotation/AnnotationHistoryTable.js
yvanzo: The unclear bit here was the change of direction in the comparison. IIUC, annotation IDs were mostly sorted by ascending date, while indexes are strictly sorted by descending date. If correct, the commit message should probably mention it.

reosarevok: Yeah, that's the point, since they're sorted from new to old, the sort here changes :) I can add it to the message.

yvanzo (approved):
It seems my questions were mainly relevant to the React conversion which I didn’t review and is not referred to from commits fixing beta regression. The lack of distinction with commits fixing production bug added to the confusion.

:ship:

PR #2566, opened 2022-06-14
Add custom hydration event for userscripts

mwiencek:
Currently it's difficult for userscripts to determine when hydration is completed so that it's safe to modify the page.  This adds a custom mb-hydration event that scripts can listen for.

Based on a suggestion by @ROpdebee in https://github.com/jesus2099/konami-command/issues/596#issuecomment-1154386261

I tested this by creating a new Greasemonkey script with the following code and observing that several events were logged.

```JS
document.addEventListener('mb-hydration', (event) => {
  console.log('Got mb-hydration event:', event);
});
```

### DIFF
--- a/root/utility/hydrate.js
+++ b/root/utility/hydrate.js
@@ -139,6 +139,8 @@ export default function hydrate<
               </React.StrictMode>,
             );
           });
+          // Custom event that userscripts can listen for.
+          root.dispatchEvent(new Event('mb-hydration', {bubbles: true}));
         }
       }
     });


mwiencek:
@ROpdebee I can't request your review, but let me know if this looks okay.  I know you also suggested additional hooks, so feel free to open tickets for other specific cases where an event is needed.  (The tickets don't have to be super detailed, I'm just not sure what else is needed at the moment.)

yvanzo (approved):
LGTMBDNT. As long as @ROpdebee is happy with it.

mwiencek:
This is now running on https://test.musicbrainz.org/

ROpdebee (approved):
The new events do the trick, thanks! I tested both @jesus2099's INLINE STUFF and my affected userscript on test.MB.

>  I know you also suggested additional hooks, so feel free to open tickets for other specific cases where an event is needed.

I can't think of any at the moment, and I don't think I need any for my own scripts, but I'll keep an eye out for any places where they might be of added benefit over timeouts/intervals/mutation observers.

mwiencek:
> I can't think of any at the moment, and I don't think I need any for my own scripts, but I'll keep an eye out for any places where they might be of added benefit over timeouts/intervals/mutation observers.

OK, sounds good.  Thanks for testing!

jesus2099:
It looks interesting!
@mwiencek, does it bubble up to the `<body>`?
And also do you know if we have pages where hydration?/hydratation? may happen not only once at the beginning but several times, for reasons?

jesus2099:
Thanks very much for everything!

jesus2099:
Ah, @mwiencek another question.
On recording merge page, for instance, where there is hydration to make the deferred display of AcoustID, for each recording row.
Will there be only a final `mb-hydration` event (on the table or tbody?) or one event for each recording row or for each AcoustID cell?

mwiencek:
> Ah, @mwiencek another question.
> On recording merge page, for instance, where there is hydration to make the deferred display of AcoustID, for each recording row.
> Will there be only a final `mb-hydration` event (on the table or tbody?) or one event for each recording row or for each AcoustID cell?

Right now there is a different event for each AcoustIdCell.  But this isn't a rule...it's technically possible to hydrate the whole table at once.  It just depends on how we code it.  Right now you can assume they're separate, but we can't promise it will never change. :)

jesus2099:
Thank you very much!
I will check if I can avoid [`setInterval` _daemons_](https://github.com/jesus2099/konami-command/issues/596#issuecomment-1154447347) with this new event.

PR #2567, opened 2022-06-14
MBS-12459: Mimic native tooltip position offset

jesus2099:
# Problem

Timeline custom tooltips are very nice but they are always partially covered by my mouse cursor.
I have noticed that the timeline custom tooltips were closer to the mouse than were the native tooltips (`<*** title="native tooltip">`).

See [MBS-12459](https://tickets.metabrainz.org/browse/MBS-12459) for before/after photos on Windows 10.

# Solution

I changed the timeline.js custom tooltip position offset to match the position, relative to mouse cursor, of the native tooltip.

# Checklist for review

- [x] Test on Windows 10
- [x] Test on Linux
- [ ] Test on Mac (if anyone can do this)
- [x] Test on mobile (maybe custom tooltip will look too far away from the triggering line) (no native tooltip on mobile)

### DIFF
--- a/root/static/scripts/timeline.js
+++ b/root/static/scripts/timeline.js
@@ -495,8 +495,8 @@ class TimelineLine {
       .css({
         'position': 'absolute',
         'display': 'none',
-        'top': y + 5,
-        'left': x + 5,
+        'top': y + 16,
+        'left': x + 12,
         'border': '1px solid #fdd',
         'padding': '2px',
         'background-color': '#fee',


jesus2099:
Wow, @yvanzo! You have tested the change on Linux, already? :)
Was it bad before? and better after?

yvanzo (approved):
Yes, it is slightly better than before as the cursor was sometimes (not always, it depends on when the event was triggered) hiding the beginning of the bubble.

Tested locally on Linux but I don’t think there will be any issue with desktop platforms anyway. It is more about dealing with smaller resolutions: tablets and mobiles.

I pushed this commit to the `test` branch and deployed it in the hope that everyone could test it with their own device from <https://test.musicbrainz.org/statistics/timeline/main>. Unfortunately, I couldn’t load the statistics data in the test database for now, so you will have to wait a bit.

jesus2099:
On mobile, it was looking better before my change:

Arrow
----------

IMG_20220616_161905
Before

IMG_20220616_161825
After

Hand
---------

Here is a little bit special because you already cannot really use this feature conveniently.
As soon as you click a milestone (red vertical line), a new tab opens to show you the related detailed webpage.
It's when you close this info tab, that you can see the tooltip.

IMG_20220616_161922
Before

IMG_20220616_161945
After

yvanzo:
> On mobile, or more precisely, when there is no mouse cursor, it was clearly looking better before my change

This page is clearly not supporting mobile at the moment; See the overlapping date labels on the horizontal axis for example.

Being able to read the first letters of the tooltips on desktop seems to overcome the slightly increased distance on mobile.

yvanzo:
Tests really are about desktop/mobile thus skipped testing on mac as it has already been tested on two desktop platforms.

PR #2568, opened 2022-06-15
MBS-12458: Error if trying to load annotation for wrong entity

reosarevok:
### Fix MBS-12458

The code was more than happy to load whichever annotation the passed ID was for, and then present it as if it was an annotation for the entity in question. This actually checks that the requested annotation is in the annotation history for that entity, and returns a bad request error otherwise.
I also added an actual message to the not found error returned if the ID does not exist, since the default "message: $id" was useless.


yvanzo (approved):
LGTMBDNT :ship:

PR #2571, opened 2022-06-23
Remove seemingly useless test sql files

reosarevok:
These used to be useful 10 years ago but all the needed data is now in the initial database set, so they do effectively nothing. `create_tracklist` isn't even called anywhere.


yvanzo (approved):
:ship:

PR #2572, opened 2022-06-23
MBS-12470: Fix broken unknown language statistics

reosarevok:
### Fix MBS-12470

See commit messages. No getting back the 10 years of lost stats, but...

The number collected in the meantime is not 0 because I guess there's some languages `WHERE l.iso_code_2t IS NOT NULL OR l.frequency > 0` which do not have an `iso_code_3` but still have releases or works using them? Didn't look too deep into it, but maybe we should.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Statistics.pm
+++ b/lib/MusicBrainz/Server/Data/Statistics.pm
@@ -999,7 +999,7 @@ my %stats = (
                 q{SELECT COALESCE(l.iso_code_3::text, 'null'), COUNT(r.gid) AS count
                 FROM release r FULL OUTER JOIN language l
                     ON r.language=l.id
-                WHERE l.iso_code_2t IS NOT NULL OR l.frequency > 0
+                WHERE l.iso_code_2t IS NOT NULL OR l.frequency > 0 OR l.id IS NULL
                 GROUP BY l.iso_code_3},
             );
 
--- a/lib/MusicBrainz/Server/Data/Statistics.pm
+++ b/lib/MusicBrainz/Server/Data/Statistics.pm
@@ -776,10 +776,11 @@ my %stats = (
             my ($self, $sql) = @_;
 
             my $data = $sql->select_list_of_lists(
-                q{SELECT COALESCE(l.iso_code_3::text, 'null'), COUNT(wl.work) AS count
-                FROM work_language wl FULL OUTER JOIN language l
-                    ON wl.language=l.id
-                WHERE l.iso_code_2t IS NOT NULL OR l.frequency > 0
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -995,13 +995,16 @@ my %stats = (
...
-                GROUP BY l.iso_code_3},
-            );
+            my $data = $sql->select_list_of_lists(<<~'SQL');
+                         SELECT COALESCE(l.iso_code_3::text, 'null'),
+                                COUNT(r.gid) AS count
+                           FROM release r
+                FULL OUTER JOIN language l ON r.language=l.id
+                          WHERE l.iso_code_2t IS NOT NULL
+                             OR l.frequency > 0
+                             OR l.id IS NULL
```
yvanzo, line 1009: Among the ten lines of changed code in the first commit <https://github.com/metabrainz/musicbrainz-server/pull/2572/commits/0db6f1543444c6d9deabc008db887218db6c442a>, it seems to be the only one related to the commit message. It seems worth splitting coding style changes off from this commit.

lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -775,13 +775,17 @@ my %stats = (
...
-            );
+            my $data = $sql->select_list_of_lists(<<~'SQL');
+                         SELECT COALESCE(l.iso_code_3::text, 'null'),
+                                COUNT(w.gid) AS count
+                           FROM work w
+                      LEFT JOIN work_language wl ON wl.work=w.id
+                FULL OUTER JOIN language l ON wl.language=l.id
+                          WHERE l.iso_code_2t IS NOT NULL
+                             OR l.frequency > 0
+                             OR l.id IS NULL
```
yvanzo, line 786: Same comment as for the the first commit: It seems worth splitting the large coding style changes off from this third commit <https://github.com/metabrainz/musicbrainz-server/pull/2572/commits/ee31d890210621a95a89021e265986839f018cee> to make more visible this only change actually related to the commit message.

yvanzo (commented):
The code looks good to me but the coding style changes are mostly hiding the actual changes.

lib/MusicBrainz/Server/Data/Statistics.pm
reosarevok: This is not the only change here, but that makes it obvious that mixing the two is confusing :) Splitting.

reosarevok:
Updated this and split style changes to their own commit :)

yvanzo (approved):
I did miss querying the `work` table indeed. Thanks! :ship:

PR #2573, opened 2022-06-24
MBS-12464: Also display event-work rels in Related Works

reosarevok:
### Fix MBS-12464

Not sure what was the reason to restrict it to these `targetEntityTypes` but it probably makes sense to list event here, if only for premiered at


yvanzo (approved):
LGTMBDNT :boat: :rocket:

PR #2577, opened 2022-06-28
MBS-12475: Add autoselect for Wikidata and RYM genre URLs

reosarevok:
### Implement MBS-12475


reosarevok:
Fair enough, I guess we could have one in case it is useful :)

yvanzo (approved):
LGTMBDNT :ship:

PR #2578, opened 2022-06-29
Fix warning when displaying blank annotation text

yvanzo:
# Problem

Annotation text is stored as `Maybe[Str]` in the model `Entity/Annotation.pm` thus returning as undefined if blank. This triggered the following warning when displaying annotation edits:

    Use of uninitialized value $old_annotation in concatenation (.) or
    string at lib/MusicBrainz/Server/Edit/Annotation/Edit.pm line 81.

# Solution

This patch appropriately defines `$old_annotation` so that it can be directly used as a string in the rest of the subroutine.

# Checklist for author

* [x] Tested with a local database by getting edits history after adding an annotation, reverting it and re-adding it.


reosarevok (approved):
This seems sensible. I do wonder whether annotation should always be NOT NULL if it exists, but that'd be a schema change I guess :)

PR #2582, opened 2022-07-04
[WIP] Convert the relationship editors to React

mwiencek:
This is currently running on https://test.musicbrainz.org/.

Sorry for the large diff.  (I guess in this case large is an understatement.)  We'll have to discuss how to review this.  I'll definitely need to take the time to explain parts of the code and add comments, and change things based on feedback.  It's currently WIP because I haven't added, updated, or fixed any tests -- I'm hoping to add some Selenium ones soon.

This resolves a bunch of tickets, but I haven't found or gathered them all yet.

The most user visible change in the release relationship editor (RRE) is, at least, MBS-7981.  You can also collapse mediums, and hopefully now open the RRE for releases with tons of mediums, as it now shares the lazy medium loading code from the release index page.

Another change in the RRE is that batch-created works aren't created immediately after submitting the dialog.  Edits to create them are entered in the final submission.  Eventually we can allow changing the name/type etc. of individual works while they're still pending.

Edits in the RRE are now sent in chunks (IIRC by recording/work) to /ws/js/edit, whereas they were previously sent all in one request.  The mini-relationship-editors still submit everything all at once (as HTML form data).

The mini-relationship-editors are of course rewritten now too and share much of the same code.  I recall there being weird bugs in the old code related to reordering work or series parts, so those should be resolved now.  I'm definitely forgetting other stuff.


root/static/scripts/common/utility/isDateEmpty.js
```diff
@@ -31,7 +31,7 @@ export function isDateObservableEmpty(
 
 export default function isDateEmpty(
   date: ?PartialDateT | ?PartialDateStringsT,
-): boolean {
+): boolean %checks {
```
reosarevok, line 18: Remind me, what did this do?

root/types/relationship.js
```diff
@@ -100,8 +100,8 @@ declare type RelationshipT = $ReadOnly<{
   +linkOrder: number,
   +linkTypeID: number,
   +source_id: number | null,
-  +source_type: string,
+  +source_type: CoreEntityTypeT,
   +target: CoreEntityT,
-  +target_type: string,
+  +target_type: CoreEntityTypeT,
```
reosarevok, line 106: This seems sensible, but please amend the commit title "Change source/type_type on RelationshipT to CoreEntityTypeT" :)

root/static/scripts/common/utility/isDatabaseRowId.js
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (C) 2022 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+const MAX_POSTGRES_INT = 2_147_483_647;
+
+export default function isDatabaseRowId(input: mixed): boolean %checks {
```
reosarevok, line 12: Is this going to be removed from Perl eventually? If not, maybe link the two with a comment?

root/static/scripts/common/utility/strings.js
```diff
@@ -46,8 +48,6 @@ export function unaccent(str: string): string {
     .replace(combiningDiacriticalMarks, '');
 }
 
-let NEXT_UNIQUE_ID = 1;
-
 export function uniqueId(prefix?: string = ''): string {
-  return prefix + String(NEXT_UNIQUE_ID++);
```
reosarevok, line 52: Is the `String()` call useless here?

root/static/scripts/common/entity2.js
```diff
@@ -0,0 +1,364 @@
...
+ */
+
+// This file can be renamed to entity.js once that file is removed.
+
+type CommonPropsT = {
+  +id?: number,
+  +name?: string,
+};
+
+export function createNonUrlCoreEntityObject(
```
reosarevok, line 17: Can this be auto-generated somehow? If not (and maybe even if it can) should we add a comment somewhere to indicate to update this when updating the types?

root/static/scripts/common/utility/createFastObjectCloneFunction.js
```diff
@@ -0,0 +1,33 @@
...
+
+/*
+ * See ./createFastObjectCloneFunction.benchmark.js
+ *
+ * node v16.6.0:
+ * spread x 30,049,382 ops/sec ±0.60% (92 runs sampled)
+ * fastClone x 182,189,219 ops/sec ±0.04% (100 runs sampled)
+ */
+
+export default function createFastObjectCloneFunction<T: {...}>(
```
reosarevok, line 18: This looks like something that could be its own PR where you also update any existing places that would benefit from it to use it :) Same is true of a bunch of other commits here, really.

lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
```diff
@@ -449,9 +452,11 @@ role {
         while (my ($key, $relationship_order) = each %reordered_relationships) {
             my ($link_type_id) = split /-/, $key;
 
+            my $link_type = $link_types_by_id->{$link_type_id};
+
             push @edits, $self->reorder_relationships(
                 $c, $form,
-                link_type_id => $link_type_id,
+                link_type => $link_type,
```
reosarevok, line 443: Why is this better? I don't have a problem with the change, but maybe explain the reasoning in the commit message?

.flowconfig
```diff
@@ -15,6 +15,7 @@
 ./root/vars.js
 
 [lints]
+unnecessary-invariant=error
```
reosarevok, line 18: unnecessary-optional-chain seems good too add to while at it, IMO?

root/static/scripts/common/utility/isDateEmpty.js
mwiencek: The documentation is hard to find, but - https://flow.org/en/docs/types/functions/#toc-predicate-functions

Basically it preserves any refinements that are made inside the function to apply at the call site (but they have to be very simple checks for this to work).

root/static/scripts/common/utility/isDatabaseRowId.js
mwiencek: I don't think it will be removed from Perl, but I added some comments.

lib/MusicBrainz/Server/Controller/Role/EditRelationships.pm
mwiencek: Added a commit message here.  (P.S. You left this comment twice so I deleted one.  :))

mwiencek:
BTW, in case you're wondering why there is so much more code than before...

 * React is a *lot* more verbose than knockout.
 * React reducers require a lot of boilerplate code.
 * Flow types take up a lot of space, too.  (And not just the several files that consist entirely of type definitions.)  There is a cost to having everything type checked.
 * Things that knockout handled for us, like propagating state changes, now have to be done manually.
 * The code is split into a lot more files.  License headers and "import" statements take up a great deal of space.
 * In some cases, the code just does more than before.
 * It's also possible that there's unused code from the old relationship editor I haven't yet removed.

reosarevok:
Testing it on test:
* It seems the reltype search also looks at the description, is that intentional? It seems annoying in most cases. Same for instrument descriptions - because of that, a search for "piano" gives mbira as the first result!
* When selecting "instrument" an empty instrument field is currently shown in prod, but not on your version. I think it's kind of confusing not to have it and might lead to more people not realizing they can specify.
* If you enter an invalid date period, the error only appears on blur, but the relationship preview already changes immediately to "Please fill out all required fields" - that's confusing, since the field is filled, it's just wrong.
* The "This relationship has ended." checkbox should probably be locked if an end date is present. It's not the case presently either IIRC, but it should be trivial with React anyway, so maybe we can do it? :) 
* [No lyrics] is almost at the top of the selections for lyrics language in prod, but not at all in test. Same issue with not having any field to start typing in the first place discouraging people from even entering a language, btw :) 
* It's a good idea to not add a work until you submit the editor, I think, but then it'd make sense to be able to see and edit the work data before you submit - right now I have no way to see whether I submitted the wrong data when batch-adding works. I couldn't do that earlier either, of course, but that seems to be the main benefit to adding pending works here - right now I can't fix mistakes other than by removing and re-adding the "work".
* When there's only one type available, like for recording-work rels, it could be preselected rather than forcing me to pick the one option from the list.
* For some reason, when trying to link an existing work to a recording by hand, I'm getting "Uncaught Error: The selected link type is invalid for these entity types" - sometimes, at least - trying again later worked somehow?
* The very useful "Add related work" link that opens a work relationship popup and preloads the search field with the work name is missing. That makes linking works a lot slower (even if it worked, see above :D)
* While somewhat minor and probably complicated, it would be great to be able to connect one pending work to another (if they are, say, arrangements or whatever) before submission.
* Moving from having the recording-work rel in one place on the works column to having it on both columns but not so distinctively does not seem like an improvement. At the moment, the recording rel does not get confused with other rels for the work, but now it might.
* Relatedly to the above, it's strange that the newly-added recording-work rel is the only one that gets shown for the work in question, even if it has a ton of recordings. I do generally agree with not showing them, but it's another reason why I feel the current way works best for that.
* Relatedly to *the above*, when linking to existing works, it doesn't seem to load existing rels for them at all. As such, you don't even see (nor can edit) their composer info and whatnot. That's really useful (sometimes for example the release you're editing will provide a not yet present composition date), and in any case not seeing that makes it harder to understand whether you chose the right work to begin with.
* Batch-adding rels shows them for all recordings/works, but it only actually submits it for the first of them.

mwiencek:
I think I've addressed all your comments (and updated test.mb again).  Let me know if I missed something.

> While somewhat minor and probably complicated, it would be great to be able to connect one pending work to another (if they are, say, arrangements or whatever) before submission.

You can do this now by copying the work link and pasting it into an autocomplete.  (The created work headings are now links that point to the same headings.)  It's not the most obvious thing, but seems to work.

kellnerd:
First of all, I really like the new relationship editor, especially the new autocomplete with support for relationship types (and inline descriptions of them!) and the delayed batch-creation of works (I hate it when they have no initial edit notes using the production server).

I've been testing the new relationship editor a bit now and I've noticed a few quirks:

- Dragging one of the page's scrollbars immediately closes a popup dialog (probably because it triggers a mousedown event, using the scrollwheel keeps the dialog open).
  This is especially annoying with the "Batch-create new works" dialog that appears on the right of the button, which can be partially outside of the viewport for a narrow browser window (two windows side by side).
- Adding a related work, selecting its checkbox and immediately deleting the (pending) new relationship leaves the status "Related Works (1 work selected)" unaffected. Selecting other works also counts the invisible "ghost work".
- The preview of a "Batch-add a relationship to works" dialog always uses the placeholder "[removed] (This entity will be created by this edit)", even if the selected works already exist. There should be a better placeholder, maybe something like "[selected]".
- Adding an "arranger" relationship to a recording and changing it to "instrument arranger" (without typing into the autocomplete input!) automatically selects all attribute checkboxes although they were not selected before. This happens regardless whether the dialog has been confirmed before choosing "instrument arranger" or if the type has been changed immediately, but it does not happen anymore as soon as you type something into the input (hence you can only click the search icon or hit the down arrow key to select a type which also matches the current search term).
- For [this huge release](https://test.musicbrainz.org/release/6c3a4d7e-1b2b-4dae-ab07-a9935c659ed2/edit-relationships) only the first of 24 mediums (1509 tracks in total) is loading (and there is no trace of a "load more" button or something similar).

Actually I wanted to test with a few different multiple media releases and releases with lots of relationships, but currently I'm getting 504 errors on test every time. Will continue testing later.

Additional remark: On the one hand I like the "Add another ..." shortcut buttons for each already related entity type, on the other hand this wastes a lot of screen space and sometimes you even have to scroll just to see the next(!) recording. Generally the larger font sizes and margins are reducing the number of tracks which are visible at the same time almost by half of the amount on the production servers.

P.S. Of course the new relationship editor breaks almost all userscripts which are running on the release relationship editor page, as the global `MB` object no longer exposes the relevant objects on which they rely. I hope it is possible to expose replacements for the most important of these methods once the new relationship editor is ready.

mwiencek:
@kellnerd Thanks for the feedback, it's really helpful!

I think those issues should be fixed, and test.mb is updated again, but the server is pretty slow right now so may not be usable until later today (sorry).

To help with the wasted screen space issue you brought up, I reduced the padding/margin around the buttons, which seems to help some.  If you have other ideas which can still keep the UI consistent with the mini relationship editor, let me know.

I also added the missing medium-toolbox with links to expand/collapse all mediums.

> P.S. Of course the new relationship editor breaks almost all userscripts which are running on the release relationship editor page, as the global MB object no longer exposes the relevant objects on which they rely. I hope it is possible to expose replacements for the most important of these methods once the new relationship editor is ready.

This should be possible -- if you know of some scripts that do this, can you link me to them?  I'm not sure which state/functions they rely on that need to be exposed.

reosarevok:
The scripts I use on the release relationship editor all the time:
https://github.com/loujine/musicbrainz-scripts/blob/master/mb-reledit-copy_dates.user.js
https://github.com/loujine/musicbrainz-scripts/blob/master/mb-reledit-guess_works.user.js
https://github.com/loujine/musicbrainz-scripts/blob/master/mb-reledit-set_instruments.user.js
https://github.com/loujine/musicbrainz-scripts/blob/master/mb-reledit-set_relation_attrs.user.js

Also paste-a-date but I'm already porting that to the server.

kellnerd:
In addition to the scripts @reosarevok already mentioned, I know of and am using...
https://github.com/loujine/musicbrainz-scripts/blob/master/mb-reledit-clone_relations.user.js
... plus two of my own scripts:
https://github.com/kellnerd/musicbrainz-scripts/blob/main/dist/parseCopyrightNotice.user.js
https://github.com/kellnerd/musicbrainz-scripts/blob/main/dist/voiceActorCredits.user.js
Since these are quite lengthy you can also have a look at the relevant source modules instead:
https://github.com/kellnerd/musicbrainz-scripts/blob/main/src/relationshipEditor.js
https://github.com/kellnerd/musicbrainz-scripts/blob/main/src/copyrightRelationships.js

Mostly it's about the ability of creating relationship dialogs programmatically, filling/opening/confirming them and obtaining the recordings/works which are currently selected.
Oh, and `MB.getSourceEntityInstance` and the new internal API from 29627fd4006adfd3160f27ed6365cc1caee60647 will definitely be really useful to have for some of these userscripts, too 😏

reosarevok:
Testing the artist mini-editor while adding a new artist:
* I can now select the credit for the artist I'm adding in the relationship (sweet, that fixes MBS-11180 so you can add it to your list of fixes if you didn't have it already). But once I add the rel, I have no way of telling it has a credit, it just shows as usual. It should probably show "(as whatever)" by it. To be fair, it seems that's already an issue with the current relationship editors though - seems like a trivial fix so maybe you can fix it? :) I couldn't find a ticket so I added it: MBS-12489
* MBS-10919 is still unchanged - is that intentional or should we be showing it?
* "Credited as" feels more visible than before for some reason. Maybe it's just me? But it might be good to have an info tooltip or something that says you don't need to fill this if it's credited with the same name as the artist name, unless we *do* want it to get filled in any case.
* "Change credits for other ArtistX relationships on the page." now only appears if I just changed a credit. Not a big deal, but sometimes I've realized that I should change other credits after I first apply the rel with the credit - right now I can just reopen and click the checkbox, but with this system I'd have to make an "edit" for it to show. Maybe show it always if a credit is entered, but not if it's not? Unless we want to allow also just blanking the credits like that, of course. That said, if you think it's less confusing to only show it if I just made a change, that's fine, I can always press space and backspace :D
* A similar issue as what kellnerd reported above for works: "Franz Liszt is/was a member of [removed] (principal and woodwind)" ([removed] being my new artist)
* Blanking relationship credits from an existing relationship shows them as gone, but no edit is submitted

kellnerd:
While I was playing around and manipulating the layout with the inspector, I found the following error: Clicking the edit icon of the "recording of" rel of track A1 on [this release](https://test.musicbrainz.org/release/b84ee12a-09ef-421b-82de-0441a926375b/edit-relationships) reliably triggers an "invalid relationship source" error (does also happen for some but not all of the other tracks' works) which is also reproduceable after reloading the page.
<details>
<summary>Full error message</summary>
```
Error: invalid relationship source
    r invariant.js:22
    i isRelationshipBackward.js:18
    it RelationshipDialogContent.js:121
    React 2
    st RelationshipDialogContent.js:444
    React 15
    onClick ButtonPopover.js:97
    React 11
    r helpers.js:71
    _wrapEventTarget trycatch.js:74
    ce instrument.js:328
    React 5
    i hydrate.js:133
    React 2
    i hydrate.js:132
    y hydrate.js:111
    jQuery 5
    y hydrate.js:109
    EVVwKNU ReleaseRelationshipEditor.js:1514
    Webpack 6
instrument.js:102:45
```
</details>

And I also noticed that only recording and work relationships are shown for this release, but not the existing release and release group relationships.

mwiencek:
@reosarevok 

> I can now select the credit for the artist I'm adding in the relationship (sweet, that fixes [MBS-11180](https://tickets.metabrainz.org/browse/MBS-11180) so you can add it to your list of fixes if you didn't have it already). But once I add the rel, I have no way of telling it has a credit, it just shows as usual. It should probably show "(as whatever)" by it. To be fair, it seems that's already an issue with the current relationship editors though - seems like a trivial fix so maybe you can fix it? :) I couldn't find a ticket so I added it: [MBS-12489](https://tickets.metabrainz.org/browse/MBS-12489)

This should be fixed now.

> [MBS-10919](https://tickets.metabrainz.org/browse/MBS-10919) is still unchanged - is that intentional or should we be showing it?

Artist credits are displayed now.  I kept it so that it hides disambiguation comments though, since it seems kind of obnoxious when the same entity with a long disambiguation has hundreds of credits.  What do you think?

> "Credited as" feels more visible than before for some reason. Maybe it's just me? But it might be good to have an info tooltip or something that says you don't need to fill this if it's credited with the same name as the artist name, unless we do want it to get filled in any case.

I made the field take up a bit less space and added a help icon next to it.  Let me know if that seems better.

> "Change credits for other ArtistX relationships on the page." now only appears if I just changed a credit. Not a big deal, but sometimes I've realized that I should change other credits after I first apply the rel with the credit - right now I can just reopen and click the checkbox, but with this system I'd have to make an "edit" for it to show. Maybe show it always if a credit is entered, but not if it's not? Unless we want to allow also just blanking the credits like that, of course. That said, if you think it's less confusing to only show it if I just made a change, that's fine, I can always press space and backspace :D

Now it should show those controls in two situations: the credit differs from the entity name, or the credit differs from the initial credit.

> A similar issue as what kellnerd reported above for works: "Franz Liszt is/was a member of [removed] (principal and woodwind)" ([removed] being my new artist)

Now it uses whatever name you've typed into the creation form.

> Blanking relationship credits from an existing relationship shows them as gone, but no edit is submitted

Fixed.

@kellnerd 

> While I was playing around and manipulating the layout with the inspector, I found the following error: Clicking the edit icon of the "recording of" rel of track A1 on [this release](https://test.musicbrainz.org/release/b84ee12a-09ef-421b-82de-0441a926375b/edit-relationships) reliably triggers an "invalid relationship source" error (does also happen for some but not all of the other tracks' works) which is also reproduceable after reloading the page.
> [...]
> And I also noticed that only recording and work relationships are shown for this release, but not the existing release and release group relationships.

These two issues should be fixed now.

Please report more issues as you find them!

mwiencek:
todo list:

 * fix current test failures
 * add a selenium test for a non-release page
 * convert a few more of the old unit tests from root/static/scripts/tests/relationship-editor.js over to the new code (though some aren't really relevant anymore)
 * add some hooks to give userscripts access to most of the internals

hoping to get these items done by the end of the week.

reosarevok:
Just noticed that the current relationship editor links to the entity in "Change credits for other A relationships on the page" but the version on test.mb does not. Is the change intentional? (I wanted to suggest to make it always open on a new tab because it causes external iframe uses in add entity popups, but if we've decided to remove the link that might also be fine).

mwiencek:
> Just noticed that the current relationship editor links to the entity in "Change credits for other A relationships on the page" but the version on test.mb does not. Is the change intentional? (I wanted to suggest to make it always open on a new tab because it causes external iframe uses in add entity popups, but if we've decided to remove the link that might also be fine).

I might've did that intentionally because I think clicking an input's label shouldn't perform any other action than selecting the the input no matter where you click.  And you can already access the target entity link from the preview below if needed.

mwiencek:
@brainzbot test this please

mwiencek:
Since a lot of relationship editor strings moved around between files, and since a few tests depend on the translated versions of those strings working, I had to update some .po files (after updating the .pot files) to associate strings with their new source paths using the following command:

```sh
find po -name 'mb_server.*.po' -type f -exec msgmerge --no-wrap --no-fuzzy-matching --update '{}' po/mb_server.pot \;
```

I'm not sure by what exact logic Transifex uses to wrap strings in the .po files we download from there, but at a glance it looks like they wrap msgid but not msgstr.  I didn't see an option in `msgmerge` to have different wrapping settings for those, so `--no-wrap` is specified for all.  This explains why so many lines are removed in the diff.

Once I push new .pot files to Transifex I'll probably merge this to beta by hand and force-update the last commit with the .po files as downloaded from Transifex to avoid diff churn.

reosarevok (approved):
I did look through the code, and test this on test finding some more issues (now fixed). Given the size of this though, I say we should merge it and use a prolonged beta period (as talked with the team) to find more hidden issues that our ad-hoc testing in test.mb might not have found. That does require us reacting fast to hotfix or revert in case of data-affecting bugs, though :)

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

PR #2586, opened 2022-07-13
MBS-12499: Error gracefully on non-implemented TXT API call

reosarevok:
### Fix MBS-12499

This was ISEing when people requested text/plain for non-genre-list calls.


mwiencek (approved):

yvanzo (commented):
I found the following calls in the Sentry issue linked from the ticket, and tested these locally (with sample data) as these seem to be related to this patch:

These are erroneous queries for a genre list:

- http://localhost:5000/ws/2/genre
- http://localhost:5000/ws/2/genre/
- http://localhost:5000/ws/2/genre?fmt=txt

The patch successfully makes the server to return an error message instead of a blank page as before. Nothing appears in server logs either.

However the commit message is misleading:

- These are not all “TXT API calls”, it just happens that `/ws/2/genre` is the only endpoint to use `TXTSerializer`. More specifically, the “non-implemented” part was just about formatting the error message in plain text.
- These are not really “non-genre-list” calls either.

----

The Sentry issue is also about other calls which were (and still are) not triggering any ISE but are unrelated to this pull request, mostly search queries: http://musicbrainz.org/ws/2/genre?query=rock

In this case, the returned 501 status code and the error message are appropriate since genre search will be implemented some day and already in progress if I’m not wrong.

----

Not knowing which calls to look for, I tried the following TXT API calls at first:

- Genre list: http://localhost:5000/ws/2/genre/all?fmt=txt (still 200 as expected)
- Non-genre list: http://localhost:5000/ws/2/recording?release=0ce274e3-3b89-4d15-94ac-4f92e497ea60&fmt=txt (still 406 as before, which is not really appropriate as it is not related to the `Accept` headers)
- Non-genre item: http://localhost:5000/ws/2/release=0ce274e3-3b89-4d15-94ac-4f92e497ea60&fmt=txt (still 404 as before, but probably not appropriate either)
- Genre item: http://localhost:5000/ws/2/genre/bf9955d6-c3eb-466d-95ca-30f3848fc9ea?fmt=txt (still 200 - with a text list of a single item containing Perl hash - as before, the status code might be appropriate if we wanted to support this in plain text too, otherwise it is inappropriate too)

Most often (but maybe for the latest) no implementation in the future should be expected (contrarily to what the “yet” suggests in error message which isn’t new to this pull request).

Nothing either new to this pull request or specific to TXT API calls, but I noticed these issues while reviewing this pull request. Not sure what would be the appropriate status code though, 422?

reosarevok:
A fair comment is that maybe it should specifically say "this is not supported" rather than "this has not been implemented yet".

It's possible that we should allow `fmt=txt` for all calls (as in, not reject it with the 406 as we do elsewhere, just reject it with a "not supported" message).

http://localhost:5000/ws/2/genre/bf9955d6-c3eb-466d-95ca-30f3848fc9ea?fmt=txt seems like a bug and we should also return not supported there.

yvanzo:
> Re: other sentry things, that sentry issue was someone specifically calling with a text/plain accept header, and this PR's goal is to avoid that from causing an ISE. A non-text query is non-erroring, so I assume that's not saved in sentry?

Genre non-text queries are erroring, as shown above with the first examples. The “other sentry things” are events for the same Sentry issue you linked to the ticket. If you go to the Sentry issue, click on the button “Older” on the same line as “Event” to browse through events, you will find the calls I mentioned above as the GET request is provided for each event.

> Anyway, happy to modify the commit message here, but we should still decide whether to change the message itself as well at the very least :)

These are separate issues which can be resolved separately:
- Your patch is providing the same error message as with XML or JSON, which is far better than the current ISE. The only issue is that both the ticket and the pull request/commit message are erroneous about the issue being resolved.
- Changing error messages (and status codes) would have require a lot more work as it has to be done for all serializers and possibly require to split or reorganize code depending on the situation. I was just sharing my findings here.

reosarevok:
> The Sentry issue is also about other calls which were (and still are) not triggering any ISE but are unrelated to this pull request, mostly search queries: http://musicbrainz.org/ws/2/genre?query=rock

Those were also triggering the ISE when sent with an Accept: text/plain header :) That's why they ended up in Sentry AFAICT.

yvanzo:
> Those were also triggering the ISE when sent with an Accept: text/plain header :) That's why they ended up in Sentry AFAICT.

Not necessarily. For example in the event [`a511dcd`](https://sentry.metabrainz.org/organizations/metabrainz/issues/64430/events/a511dcdfe9ed4d53bcfb85b15ab42b34/?project=12&referrer=jira_integration), `application/json` comes before `text/plain` in the `Accept` headers.

In the case where `text/plain` would come first (but I did not find any), the status code 406 would be appropriate, not in the case of `fmt=txt` is used. That’s an example where “split” might be necessary for whenever we want to fix the returned status codes.

PR #2587, opened 2022-07-13
Fix various hydration errors due to broken i18n

mwiencek:
Client scripts that use i18n functions must go under
root/static/scripts/.  Hydration errors can occur if the user has a
non-English translation enabled: the server will render the hydratable
component in the correct language, but once the client script executes,
it'll revert to English due to missing strings.

Note: This patch won't have any effect until a new mb_server.pot file is
pushed *and* new .po files with updated paths are downloaded.


reosarevok:
I'm a bit confused. `RatingStars` for example doesn't seem to be using `hydrate` - how come it's hydratable?

reosarevok (approved):
Seems good. Is there any way we can add tests for this so it does not happen again?

mwiencek:
@reosarevok I added a test for this if you want to check it, at t/hydration_i18n.t

PR #2591, opened 2022-07-21
Upgrade Flow to 0.183.0

mwiencek:
https://github.com/facebook/flow/releases/tag/v0.183.0

The first changelog entry is a bit vague but we did encounter some new
errors which I've fixed here; mostly by adding additional annotations to
help Flow and fixing some incorrect type parameters.


reosarevok (approved):
LGTM

PR #2593, opened 2022-07-22
MBS-12497: Drop AC redirects when last use is removed

reosarevok:
### Fix MBS-12497

We automatically remove empty ACs after the last usage is removed, using the `dec_ref_count` function. The `artist_credit_gid_redirect` table has FKs to `artist_credit` though, and it does not affect this process in any way. That means that when an AC with only 1 usage has a redirect, any edit that would remove it is blocked because the redirect is not being removed and the associated FK still exists.

I talked with @mwiencek and we decided to implement an intermediate table where rows can be entered when the `ref_count` gets to 0, rather than immediately deleting them. We then wait 7 days to see if the artist credit got reused (`ref_count` went above 0 again) in the meantime. This avoids losing redirects from an AC that would get used again soon after, for example because it is being removed from one recording and then added to another.

After 7 days the artist credit gets removed, including any redirects. If it did get reused in the meantime, we just drop it from the temporary table, obviously without removing the AC.

The temporary table setup should also work with the alternative tracklist code in the future, which is why the table is generic and stores a table name rather than being specifically AC-related.


reosarevok:
This only contains a test with what I expect should happen for now, but I'm not really sure how to implement this.

The whole deletion process works automatically with triggers. Should the `dec_ref_count` function also remove any redirects when needed? Should the `artist_credit_gid_redirect` table be updated to `CASCADE`? Should we try to move the redirect to whatever the new AC is if the AC was changed? (this still leaves us the same problem when the last recording with the AC is deleted or whatnot).

mwiencek:
I think you're right that any good solution is going to be a schema change.

I was also thinking we should maybe move the redirect, but then I thought of this scenario: say you want to swap the artist credits on two entities, and each entity is the last reference to their respective artist credit.  The first artist credit that's de-referenced will be deleted and have its redirect moved to the other AC, even though the intent is to add it back in the next edit or statement.  It's possible they may not even be related.

So I don't see any point in immediately deleting anything.  We could have a cleanup script that does it instead, but perhaps we shouldn't be deleting redirected ACs at all.  If we kept them in the database, that would allow us to see or even display to users *what* artist credits are redirecting.  The only issue with this is what happens if someone tries to use an AC that's been redirected, since it would still resolve to something in the database.  We could trivially modify `Data::ArtistCredit::_find` to follow any redirect, but I wonder if it wouldn't be confusing that a different artist credit gets used than the one they entered.  It's not clear to me if the intent of a redirect is to always be unconditional, or if sometimes it's only meaningful to redirect it in the context of the release or other entity where it was originally used.

reosarevok:
I mean, I think it potentially then makes sense to redirect ACs when artists are merged or whatnot, but keep them with a ref_count of 0 when they are just to be deleted without a redirect? But I'm not sure tbh.

That said, this is currently getting a fair amount of edits stuck and releases that ISE when one tries to change them. So we should probably look for a short term fix at least that allows us to solve that issue.

reosarevok:
Had a long chat with @mwiencek about this, and implemented it with an intermediate `unreferenced_row_log` table. Test seems to work fine, but happy to amend as needed.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,112 @@
...
+               AND id = any(?)
+            SQL
+
+        my $unused_ids = $self->c->sql->select_single_column_array(
+            $query,
+            $row_ids,
+        );
+
+        for my $id (@$unused_ids) {
+            ++$count;
```
mwiencek, line 68: AFAICT the count is incremented unconditionally here, so it's equivalent to `my $count = scalar @$unused_ids;`.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,112 @@
...
+                Sql::run_in_transaction(sub {
+                    if ($table_name eq 'artist_credit') {
+                        $self->sql->do(<<~'SQL', $id);
+                            DELETE FROM artist_credit_gid_redirect
+                                  WHERE new_id = ?
+                            SQL
+                    }
+                    $self->c->sql->do(<<~"SQL", $id);
+                        DELETE FROM $table_name
+                              WHERE id = ?
```
mwiencek, line 86: Expanding on my previous comment, one way to handle the race condition would be to do `WHERE id = ? AND ref_count = 0` here, then check `$self->c->sql->row_count` to see if anything was deleted.  But we'll have to `die` manually if nothing was deleted so that the `artist_credit_gid_redirect` deletion gets rolled back.  It might make sense to perform the `unreferenced_row_log` deletions individually in this transaction too.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,112 @@
...
+            $self->c->sql->do(<<~'SQL', $table_name, $row_ids);
+                DELETE FROM unreferenced_row_log
+                      WHERE table_name = ?
+                        AND row_id = any(?)
+                SQL
+        }, $self->c->sql);
+    }
+
+    if ($self->summary) {
+        log_info { sprintf 'Found %d unused unreferenced row%s.',
```
mwiencek, line 104:
```suggestion
        log_info { sprintf 'Found %d row%s unreferenced for 7 or more days.',
```

t/lib/t/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -7,6 +7,7 @@ use Test::Moose;
 use Test::Fatal;
 use Test::More;
 
+use MusicBrainz::Script::RemoveUnreferencedRows;
```
mwiencek, line 10: This works for now, but it probably makes sense to move the test under t/script/, since it's really testing the script now and not anything in the data layer.

admin/sql/CreateTables.sql
```diff
@@ -3890,6 +3890,12 @@ CREATE TABLE medium_index ( -- replicate
     toc                 CUBE
 );
 
+CREATE TABLE unreferenced_row_log (
+    table_name      VARCHAR NOT NULL, -- PK
+    row_id          INTEGER NOT NULL, -- PK
+    inserted        TIMESTAMP WITH TIME ZONE DEFAULT NOW()
```
mwiencek, line 3896: nitpick: can we align the column types here with the tabstop most of the other definitions are using

mwiencek (changes requested):

admin/cleanup/RemoveUnreferencedRows
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+use FindBin;
+use lib "$FindBin::Bin/../../lib";
+
+use DBDefs;
+
```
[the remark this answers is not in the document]
reosarevok: Heh. I took this from RemoveEmpty where I guess it's probably not needed either :)

t/lib/t/MusicBrainz/Server/Data/ArtistCredit.pm
reosarevok: Well, it's also testing that it got added to the table to begin with, but I can move it to script in any case.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
reosarevok: I understand this approach and the catch_exceptions one are not needed in combination? Or do you think they both make sense?

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,112 @@
...
+                        $self->sql->do(<<~'SQL', $id);
+                            DELETE FROM artist_credit_gid_redirect
+                                  WHERE new_id = ?
+                            SQL
+                    }
+                    $self->c->sql->do(<<~"SQL", $id);
+                        DELETE FROM $table_name
+                              WHERE id = ?
+                        SQL
+                    ++$removed
```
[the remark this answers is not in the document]
reosarevok: I tried that, but I'm not sure if this will keep processing or still just die. Can you take a look? :)

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,136 @@
...
+                            }
+                            $self->c->sql->do(<<~"SQL", $id);
+                                DELETE FROM $quoted_table_name
+                                      WHERE id = ?
+                                SQL
+                            $self->c->sql->do(<<~'SQL', $table_name, $id);
+                                DELETE FROM unreferenced_row_log
+                                      WHERE table_name = ?
+                                        AND row_id = ?
+                                SQL
```
mwiencek, line 98: This now won't run if the deletion above fails due to existing FKs (ref_count != 0), but it should.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,136 @@
...
+                        sprintf "id=%s from $table_name is no longer unreferenced",
+                        $id,
+                    } if $self->verbose;
+                } else {
+                    # If no longer unreferenced, drop from the unreferenced table
+                    Sql::run_in_transaction(sub {
+                        $self->c->sql->do(<<~'SQL', $table_name, $id);
+                            DELETE FROM unreferenced_row_log
+                                  WHERE table_name = ?
+                                    AND row_id = ?
```
mwiencek, line 119: This has a different race condition in that the row can become de-referenced again in between when the `$unused_ids` query runs and now.  (Then it will never be cleaned up.)

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
mwiencek: Now that I looked at the latest code I think this is needed.  But it'd actually simplify the code (you wouldn't need the `$unused_ids` query at all) and be free of race conditions.  Maybe just keep a separate list of IDs that should be deleted from the log table at the end.

admin/cleanup/RemoveUnreferencedRows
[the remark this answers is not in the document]
mwiencek: I marked this as unresolved since it's still there. :)

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
reosarevok: I tried to use `$self->c->sql->row_count` but it's not working, getting this: `Error while processing id 1 from artist_credit: Cannot delegate row_count to rows because the value of sth is not defined at inline delegation in Sql for sth->rows`. Pushed it so you can maybe take a look.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,130 @@
...
+                        unless ($self->dry_run) {
+                            if ($table_name eq 'artist_credit') {
+                                $self->sql->do(<<~'SQL', $id);
+                                    DELETE FROM artist_credit_gid_redirect
+                                          WHERE new_id = ?
+                                    SQL
+                            }
+
+                            $self->c->sql->do(<<~"SQL", $id);
+                                DELETE FROM $quoted_table_name
```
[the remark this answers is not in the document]
reosarevok: Isn't that set to `CASCADE` anyway?

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,130 @@
...
+            });
+        }
+    }
+
+    if ($self->summary) {
+        log_info { sprintf 'Found %d row%s in the unreferenced table.',
+            $count, ($count==1 ? '' : 's') };
+        log_info { sprintf 'Successfully removed %d unreferenced row%s.',
+            $removed, ($removed==1 ? '' : 's') }
+                unless $self->dry_run;
```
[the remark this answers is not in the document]
reosarevok: But isn't it useful to know if it found 1+, but removed 0?

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,130 @@
...
+
+has dry_run => (
+    isa => 'Bool',
+    is => 'ro',
+    default => 0,
+    traits => [ 'Getopt' ],
+    cmd_flag => 'dry-run'
+);
+
+has summary => (
```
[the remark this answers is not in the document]
reosarevok: Having it be per-table sounds sensible, actually, will do that.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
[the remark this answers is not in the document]
reosarevok: I changed it to log "No rows were removed." instead in that case.

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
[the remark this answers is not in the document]
mwiencek: Yeah, the DB I have is missing foreign keys for some reason

mwiencek (approved):
LGTM.  Thanks!

lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
```diff
@@ -0,0 +1,123 @@
+package MusicBrainz::Script::RemoveUnreferencedRows;
+use Moose;
+
```
yvanzo, line 3: I suggest adding a description to the script, and the usual copyright and license section:

```suggestion

=head1 DESCRIPTION

This script process the rows of the table 'unreferenced_row_log' that have been
inserted more than 7 days ago.

This table weakly references rows (of other tables) that have been temporarily
unreferenced but might be referenced again by an edit, hence the 7 days delay.

When processing an 'unreferenced_row_log' row, the script removes the weakly
referenced row if it is still unreferenced by checking 'ref_count', and removes
the 'unreferenced_row_log' row in any case.

=cut

=head1 COPYRIGHT AND LICENSE

Copyright (C) 2022 MetaBrainz Foundation

This file is part of MusicBrainz, the open internet music database,
and is licensed under the GPL version 2, or (at your option) any
later version: http://www.gnu.org/licenses/gpl-2.0.txt

=cut

```

yvanzo (approved):
* Please document the script through a comment at least (made an optional suggestion).
* Please add a plan of action for deployment to either the pull request’s description (see [template](https://github.com/metabrainz/musicbrainz-server/blob/master/.github/PULL_REQUEST_TEMPLATE.md)) or to the ticket itself.

Look excellent otherwise! Thoroughly read script and tests but did not run.
Already approved as it should be straightforward and not involve any code.

reosarevok:
@mwiencek: how would you go around releasing this? I expect just running the specific SQL upgrade file in prod, or is there a better option I'm missing? I understand we can't really have a beta period for this in any sensible way.

mwiencek:
> @mwiencek: how would you go around releasing this? I expect just running the specific SQL upgrade file in prod, or is there a better option I'm missing? I understand we can't really have a beta period for this in any sensible way.

We'll have to just run the upgrade script in production, since the bug is only fixed by replacing the `dec_ref_count` function, which can't be beta-tested.  However, the rest of the code changes (including the added script) can wait to be deployed until the next release.

PR #2597, opened 2022-08-09
MBS-12525: Add fallback sucrase loader for Node.js < 16.12.0

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12525

The bin/sucrase-node script added in 6aefc88 doesn't work in versions of Node.js prior to 16.12.0.  (16.13.0 was the first LTS release, and from the changelog appears to be identical to 16.12.0.)  This is because the ESM loader hooks API changed.

Since our INSTALL guide doesn't specify that you need an LTS release of v16, we should support older versions too.  A fallback loader using the pre-v16.12 API has been added for this cause.

I tested this inside a musicbrainz-docker dev container which still has Node v16.1.0 installed by simply running `./script/compile_resources.sh`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -72,7 +72,18 @@ Prerequisites
 
     Node.js is required to build (and optionally minify) our JavaScript and CSS.
     If you plan on accessing musicbrainz-server inside a web browser, you should
-    install Node and the package manager Yarn. Do this by running:
+    install Node and the package manager Yarn.
+
+    We currently run Node.js v16.16.0 in production.  While we try to support
+    all 16.x versions of Node, it's recommended to install one greater than or
+    equal to v16.13.0, as this is when the LTS line started and better matches
+    what we use and know works.  If your release of Ubuntu doesn't have such a
+    version of Node.js in its repositories, we can recommended the NodeSource
+    binary distributions, which we also use in production:
+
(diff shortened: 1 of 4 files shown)


reosarevok (approved):
If you've tested this and it works, then we should merge. @yvanzo: we probably need an updated docker release with this or with a higher Node version? :)

yvanzo (approved):
Successfully tested loading the front page with Node.js v16.0.0.

More generally, it’s unfortunate that the version required in installation prerequisites and the version used in production/continuous integration tests do not match. As we probably don’t want to revert to Node v16.0.0 in production, maybe we could at least add to the installation prerequisites a mention of the version being used in production/tested with as a “recommended version”?

yvanzo:
@reosarevok:
- dependency updates made to this repository must be propagated to `musicbrainz-docker` (now in release [`v-2022-08-09`](https://github.com/metabrainz/musicbrainz-docker/releases/tag/v-2022-08-09));
- upgrading MusicBrainz Server in `musicbrainz-docker` must be tested before committing, like any commit/merge.

mwiencek:
@yvanzo Agreed, I pushed a change to INSTALL.md in the latest commit.  Would you like to check that it sounds okay?

yvanzo (approved):
Very good, thank you!

PR #2600, opened 2022-08-11
Add entities.mjs & generate entities.json from it

mwiencek:
Importing entities.json into a JS file using Flow produces an any-typed
value.  Defining a Flow type for this JSON and keeping it up-to-date is
much harder than just defining the ENTITIES structure in JS and letting
Flow create a literal object type for it.  We can then generate
entities.json (which is still used by the Perl) from this file using an
added script.

### DIFF
--- a/entities.json
+++ b/entities.json
@@ -1,4 +1,5 @@
 {
+    "": "Automatically generated, do not edit. Refer to entities.mjs for instructions.",
     "annotation": {
         "model": "Annotation"
     },
--- a/entities.json
+++ b/entities.json
@@ -16,7 +16,16 @@
             "id": 1
         },
         "collections": true,
-        "custom_tabs": ["artists", "events", "labels", "releases", "recordings", "places", "users", "works"],
+        "custom_tabs": [
+            "artists",
+            "events",
+            "labels",
+            "releases",
+            "recordings",
+            "places",
+            "users",
+            "works"
+        ],
         "date_period": true,
         "disambiguation": true,
         "edit_table": true,
@@ -65,7 +74,12 @@
             "id": 3
         },
         "collections": true,
-        "custom_tabs": ["releases", "recordings", "works", "events"],
+        "custom_tabs": [
+            "releases",
+            "recordings",
+            "works",
+            "events"
+        ],
         "date_period": true,
         "disambiguation": true,
         "edit_table": true,
@@ -86,7 +100,10 @@
         "ratings": true,
         "removal": {
             "automatic": {
-                "exempt": [1, 2],
+                "exempt": [
+                    1,
+                    2
+                ],
                 "extra_fks": {
                     "artist_credit_name": "artist"
                 }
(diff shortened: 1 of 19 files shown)


root/components/EntityTabs.js
```diff
@@ -133,7 +133,10 @@ function buildLinks(
     links.push(buildLink(l('Tags'), entity, 'tags', page));
   }
 
-  if (entityProperties.ratings || entityProperties.reviews) {
+  if (
+    (/*:: entityProperties.ratings != null && */ entityProperties.ratings) ||
+    entityProperties.reviews
+  ) {
```
mwiencek, line 139: Flow didn't like `reviews` being accessed here without checking that `ratings` isn't null first (not exactly sure why).

entities.json
```diff
@@ -1,4 +1,5 @@
 {
+    "": "Automatically generated, do not edit. Refer to entities.mjs for instructions.",
```
mwiencek, line 2: I think this file is only accessed by Perl now, so if we wanted to, we could just generate a Perl module instead (so it doesn't have to read/parse JSON). But I kept it simple for now...

mwiencek:
@brainzbot retest this please

entities.mjs
```diff
@@ -0,0 +1,882 @@
...
+  },
+  work_type: {
+    cache: {
+      id: 38,
+    },
+    model: 'WorkType',
+  },
+};
+
+deepFreeze(ENTITIES);
```
reosarevok, line 880: I understand that this effectively makes ENTITIES immutable to the point where flow does not require a declared type because it's just literally what is there? If so, this might allow us to `flow strict` or `flow strict-local` some of the files which were accessing it, FWIW.

reosarevok (approved):
The idea seems good and tests pass. This will probably get a few conflicts once I merge https://github.com/metabrainz/musicbrainz-server/pull/2598 though.

root/components/EntityTabs.js
reosarevok: Did it also complain if you assigned `entityProperties.ratings || entityProperties.reviews` to a constant?

entities.mjs
mwiencek: Not quite.  This just prevents `ENTITIES` from being modified at runtime (an error will be thrown).  Flow doesn't know what this means, though, so the actual type will remain writable.  This module is marked as `@flow strict` either way, so it doesn't matter if the type is writable or read-only.

root/components/EntityTabs.js
mwiencek: That doesn't help because the error is on the `entityProperties.reviews` access (so it's just moving the location of the error).  I decided to mark it with `$FlowIssue` instead though, since that should at least tell us if it's fixed.

entities.mjs
mwiencek: Made it so the exported type is now read-only too.

PR #2601, opened 2022-08-11
MBS-11303: Access $c via React context APIs

mwiencek:
https://tickets.metabrainz.org/browse/MBS-11303

When I updated root/utility/hydrate.js in
9ef8b79b8e5e4f3c41fb8087b744f8ed90f3ff2d, I forgot to also update its
global type signature in root/vars.js.  When I got around to doing that,
it caused a ton of errors.

Trying to express a proper type that accounts for the component
receiving an unsanitized `$c` on the server and a sanitized one on the
client is pretty hard.  I'd rather components only access `$c` via the
React context APIs, which allows us to keep the `hydrate` type simple.
Server components always being passed `$c` through props wasn't very
obvious or documented, anyway.

### DIFF
--- a/root/label/LabelIndex.js
+++ b/root/label/LabelIndex.js
@@ -13,6 +13,7 @@ import CleanupBanner from '../components/CleanupBanner.js';
 import FormRow from '../components/FormRow.js';
 import FormSubmit from '../components/FormSubmit.js';
 import PaginatedResults from '../components/PaginatedResults.js';
+import {SanitizedCatalystContext} from '../context.mjs';
 import DescriptiveLink
   from '../static/scripts/common/components/DescriptiveLink.js';
 import WikipediaExtract
@@ -27,7 +28,6 @@ import {returnToCurrentPage} from '../utility/returnUri.js';
 import LabelLayout from './LabelLayout.js';
 
 type Props = {
-  +$c: CatalystContextT,
   +eligibleForCleanup: boolean,
   +label: LabelT,
   +numberOfRevisions: number,
@@ -39,7 +39,6 @@ type Props = {
 };
 
 const LabelIndex = ({
-  $c,
   eligibleForCleanup,
   label,
   numberOfRevisions,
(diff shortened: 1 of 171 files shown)


reosarevok:
I expect this completes MBS-11303 then?

root/static/scripts/common/components/TagEditor.js
```diff
@@ -558,18 +553,22 @@ export const MainTagEditor = (hydrate<TagEditorProps>(
...
-                    `Tags with a score of zero or below,
-                     and tags that you’ve downvoted are hidden.`,
-                  )}
-                </p>
-              ) : (
-                <p>
-                  {l('Tags with a score of zero or below are hidden.')}
-                </p>
-              )}
+              <SanitizedCatalystContext.Consumer>
```
reosarevok, line 556: Why add several `Consumer` here rather than assigning `$c` once?

root/static/scripts/common/components/TagEditor.js
```diff
@@ -20,6 +20,7 @@ import debounce from '../utility/debounce';
 import isBlank from '../utility/isBlank';
 
 import TagLink from './TagLink';
+import {SanitizedCatalystContext} from '../../../../context.mjs';
```
reosarevok, line 23: Wrong place - it's failing eslint :)

reosarevok (changes requested):

root/static/scripts/common/components/TagEditor.js
mwiencek: These create components that subscribe to context changes.  So the difference would be the entire component re-rendering when the context changes vs. just the sections that use it.  We don't currently change the `$c` context at all, but maybe we will in the future, and I think it's bad form to code the app in such a way that context changes would degrade performance.

(If you mean why not use `useContext`, you can't use hooks inside class-based components.)

mwiencek: Since all the Consumer components are adjacent I merged those together at least -- the diff is nicer if you enable "Hide whitespace"

mwiencek:
> I expect this completes [MBS-11303](https://tickets.metabrainz.org/browse/MBS-11303) then?

Yes, I believe so (I didn't know there was a ticket, thanks).

mwiencek:
Uh, actually, I vastly underestimated what was needed for that, but it should be completed now.  You'll definitely want to view the diff with whitespace changes hidden (which collapses the diff down to  +686 −610).

root/label/LabelIndex.js
```diff
@@ -26,8 +27,8 @@ import {returnToCurrentPage} from '../utility/returnUri.js';
 
 import LabelLayout from './LabelLayout.js';
 
+
```
reosarevok, line 30: Intentional?

root/layout/components/BottomMenu.js
```diff
@@ -9,7 +9,7 @@
 
 import * as React from 'react';
 
-import {CatalystContext} from '../../context.mjs';
+import {CatalystContext, SanitizedCatalystContext} from '../../context.mjs';
```
reosarevok, line 12: Can we allow `stash.server_languages` in `SanitizedCatalystContext` and always use that? It doesn't seem like something that wouldn't belong and using sanitized always when possible seems like a good plan anyway.

root/search/components/InstrumentResults.js
```diff
@@ -9,17 +9,18 @@
...
   from '../../static/scripts/common/components/InstrumentListEntry.js';
 import {isRelationshipEditor}
   from '../../static/scripts/common/utility/privileges.js';
-import type {ResultsPropsWithContextT, SearchResultT} from '../types.js';
+import type {ResultsPropsT, SearchResultT} from '../types.js';
 
 import PaginatedSearchResults from './PaginatedSearchResults.js';
 import ResultsLayout from './ResultsLayout.js';
 
+
```
reosarevok, line 22: Intentional?

root/search/components/LabelResults.js
```diff
@@ -49,39 +50,41 @@ function buildResult(result: SearchResultT<LabelT>, index: number) {
...
-      <p>
-        {exp.l('Alternatively, you may {uri|add a new label}.', {
-          uri: '/label/create?edit-label.name=' + encodeURIComponent(query),
-        })}
-      </p>
-    ) : null}
-  </ResultsLayout>
-);
+}: ResultsPropsT<LabelT>):
+React.Element<typeof ResultsLayout> => {
```
reosarevok, line 59: These should fit in one line now, which would look a lot better :)

root/search/components/PlaceResults.js
```diff
@@ -46,39 +47,41 @@ function buildResult(result: SearchResultT<PlaceT>, index: number) {
...
-      <p>
-        {exp.l('Alternatively, you may {uri|add a new place}.', {
-          uri: '/place/create?edit-place.name=' + encodeURIComponent(query),
-        })}
-      </p>
-    ) : null}
-  </ResultsLayout>
-);
+}: ResultsPropsT<PlaceT>):
+React.Element<typeof ResultsLayout> => {
```
reosarevok, line 56: Same: one line? :)

root/search/components/ReleaseGroupResults.js
```diff
@@ -46,37 +47,39 @@ function buildResult(result: SearchResultT<ReleaseGroupT>, index: number) {
...
-        {exp.l('Alternatively, you may {uri|add a new release group}.', {
-          uri: '/release-group/create?edit-release-group.name=' +
-            encodeURIComponent(query),
-        })}
-      </p>
-    ) : null}
-  </ResultsLayout>
-);
+}: ResultsPropsT<ReleaseGroupT>):
+React.Element<typeof ResultsLayout> => {
```
reosarevok, line 56: One line? :)

root/search/components/ReleaseResults.js
```diff
@@ -133,29 +133,31 @@ React.Element<typeof PaginatedSearchResults> => {
...
-        {exp.l('Alternatively, you may {uri|add a new release}.', {
-          uri: '/release/add',
-        })}
-      </p>
-    ) : null}
-    {manifest.js('common/components/TaggerIcon', {async: 'async'})}
-  </ResultsLayout>
-);
+}: ResultsPropsT<ReleaseT>):
+React.Element<typeof ResultsLayout> => {
```
reosarevok, line 142: One line? :)

root/search/components/SeriesResults.js
```diff
@@ -40,35 +41,38 @@ function buildResult(
...
-      <p>
-        {exp.l('Alternatively, you may {uri|add a new series}.', {
-          uri: '/series/create?edit-series.name=' + encodeURIComponent(query),
-        })}
-      </p>
-    ) : null}
-  </ResultsLayout>
-);
+}: ResultsPropsT<SeriesT>):
+React.Element<typeof ResultsLayout> => {
```
reosarevok, line 50: One line? :)

root/search/components/WorkResults.js
```diff
@@ -37,39 +38,41 @@ function buildResult(
...
-      <p>
-        {exp.l('Alternatively, you may {uri|add a new work}.', {
-          uri: '/work/create?edit-work.name=' + encodeURIComponent(query),
-        })}
-      </p>
-    ) : null}
-  </ResultsLayout>
-);
+}: ResultsPropsT<WorkT>):
+React.Element<typeof ResultsLayout> => {
```
reosarevok, line 47: One line? :)

reosarevok (approved):
Passes and seems like a nice change, left a few suggestions and whatnot, but feel free to merge as-is if you don't agree with the suggestions :) 

I also like that this seems to get some more files to indent at 2 rather than 4 spaces, which seems like less of a waste of space.

PR #2602, opened 2022-08-12
MBS-11758: Fix more hydration errors due to missing strings

mwiencek:
https://tickets.metabrainz.org/browse/MBS-11758

Follow-up to https://github.com/metabrainz/musicbrainz-server/pull/2587

I fixed a small issue in t/hydration_i18n.t, which revealed even more scripts that need to be moved.


reosarevok (approved):
Tests pass, the concept seems sane, so go for it I guess? :) (I did quickly look through the changes and saw nothing weird either, at least)

PR #2603, opened 2022-08-13
Eslint: autosort imports by path

reosarevok:
I was talking to @mwiencek yesterday about wanting to sort imports by path (most plugins sort by import name) and I eventually found https://github.com/lydell/eslint-plugin-simple-import-sort/ which seems to do the trick fairly well.

I added a custom sorting config so that we still sort same-folder imports last. It's based on one of their default examples, removing things that we don't seem to need.

`import/order` is superseded by this (and could conflict depending on the configuration chosen) so I'm removing it.


.eslintrc.yaml
```diff
@@ -357,3 +358,27 @@ rules:
...
+  ####################################
+  # eslint-plugin-simple-import-sort #
+  ####################################
+
+  simple-import-sort/imports: [warn, {groups: [
+      # Node.js builtins. You could also generate this regex if you use a `.js` config.
+      # For example: `^(${require("module").builtinModules.join("|")})(/|$)`
+      ["^(assert|buffer|child_process|cluster|console|constants|crypto|dgram|dns|domain|events|fs|http|https|module|net|os|path|punycode|querystring|readline|repl|stream|string_decoder|sys|timers|tls|tty|url|util|vm|zlib|freelist|v8|process|async_hooks|http2|perf_hooks)(/.*|$)"],
+      # Packages. `react` related packages come first.
+      ["^react", "^@?\\w"],
```
mwiencek, line 371: Not sure I see any reason to sort react stuff first.  I've always sorted it alphabetically with all the other package imports.

.eslintrc.yaml
```diff
@@ -357,3 +358,27 @@ rules:
...
+      # Internal packages.
+      ["^(@|@company|@ui|components|utils|config|vendored-lib)(/.*|$)"],
```
mwiencek, line 373: We don't use this, so can remove it

.eslintrc.yaml
```diff
@@ -357,3 +358,27 @@ rules:
...
+      # Side effect imports.
+      ["^\\u0000"],
+      # Parent imports. Put `..` last.
+      ["^\\.\\.(?!/?$)", "^\\.\\./?$"],
+      # Other relative imports. Put same-folder imports and `.` last.
+      ["^\\./(?=.*/)(?!/?$)", "^\\.(?!/?$)", "^\\./?$"],
+      # Style imports.
+      ["^.+\\.s?css$"],
```
mwiencek, line 381: We don't use this either (& if we did they'd be .less files)

mwiencek (commented):
`import/order` should be turned off since it conflicts with this plugin.  Also, there's a comment above `sort-imports: off` which references us using `import/order`; should be changed to mention this plugin instead

this looks nice otherwise!

reosarevok:
AFAICT 'import/order' doesn't actually conflict in the current usage we have, but I guess it doesn't do anything this already doesn't so I might as well remove it.

PR #2604, opened 2022-08-13
MBS-11758: Use current_language instead of lang cookie in Node.js

mwiencek:
https://tickets.metabrainz.org/browse/MBS-11758

The Node template renderer currently sets its locale from the lang
cookie.  This may conflict with the locale set by Perl, which would not
be based on the lang cookie if it's not set, but rather the
Accept-Language header.  We should respect the locale determined and set
by Perl, which it stores in current_language in the stash.

This fixes an issue where certain strings (anything rendered by Node)
would be displayed in English even if the browser locale was not
English.

### DIFF
--- a/root/server/response.mjs
+++ b/root/server/response.mjs
@@ -14,7 +14,6 @@ import {
   CatalystContext,
   SanitizedCatalystContext,
 } from '../context.mjs';
-import getRequestCookie from '../utility/getRequestCookie.mjs';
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
Tested locally by switching language and resetting it, on various pages. :ship:

reosarevok (approved):
LGTMBDNT

PR #2605, opened 2022-08-19
MBS-8876: Fix the banner message on UI language switch

yvanzo:
# Problem

MBS-8876: The translation hint in the banner message was not shown in the newly selected UI language.

# Solution

Split the banner message into two parts:
- The language change notice, still shown in the previously used language
  * Additionally add a direct link for the user to revert the change in case of mistake
- The translation hint, in the newly selected UI language
  * Additionally hide this message when the selected UI language is English (since it is also the source language for translations).

Note: Changes are easier to read with hidden whitespace.

# Checklist for author

* [x] Tested by
  * updating source messages in `po/mb_server.pot`
  * translating messages in `po/mb_server.*.po`
  * installing translations with `cd po && make install`
  * Switching UI language from the front page and from the editor profile

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Root.pm
+++ b/lib/MusicBrainz/Server/Controller/Root.pm
@@ -1,11 +1,13 @@
 package MusicBrainz::Server::Controller::Root;
 use Carp qw( croak );
+use DateTime::Locale;
 use Digest::MD5 qw( md5_hex );
 use Moose;
 use Try::Tiny;
 use List::AllUtils qw( max );
 use Readonly;
 use Scalar::Util qw( blessed );
+use URI::Escape qw( uri_escape_utf8 );
 use URI::QueryParam;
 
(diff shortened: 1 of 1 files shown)


yvanzo:

yvanzo:

reosarevok (approved):
Tested, seems to work fine. Given we don't have the line translated and you say you already tested translating it, I didn't check that it translates as expected - I trust your test.

Would it make sense to split the two sections with a horizontal line or something so that the different language mix is not confusing? If you don't feel it's useful (you've seen the translated version and know better how it feels), feel free to merge as-is :)

yvanzo:
It is split by a line break already so as to support both LTR and RTL locales. Don’t feel that an extra separator would be needed.

reosarevok:
In that case, feel free to merge :) If anyone finds it unclear they can always report it later.

PR #2608, opened 2022-08-25
MBS-12544: Support Egypt and Saudi Arabia ASINs

reosarevok:
### Implement MBS-12544

These do not seem to have working Amazon Music pages yet, so not adding it to those.


yvanzo (approved):
LGTMBDNT. Just grepped the code for potential other occurrences to be updated, only found code for Amazon Music left unchanged as expected. :ship:

PR #2609, opened 2022-08-25
MBS-12543: Add a TikTok exception for guess case

reosarevok:
### Implement MBS-12543

It seems sensible to treat this the same way we already do YouTube.


yvanzo (approved):
LGTMBDNT :ship:

PR #2611, opened 2022-08-25
MBS-12435: Block smart links: linkfly.to

reosarevok:
### Implement MBS-12435

https://linkfly.to/ - doesn't seem needed as a host either.


yvanzo (approved):
LGTMBDNT :ship:

PR #2612, opened 2022-08-26
Fix the spacing below ISRCs/ISWCs (show ... more)

yvanzo:
# Problem

In previous implementation of MBS-12334 by pull request #2500, ISRCs/ISWCs “(show ... more)” - which is displayed to shorten a long list of ISRCs/ISWCs - was rendered as a `p` element in the sidebar causing unintentional extra space under it.

# Solution

This patch uses `div` instead which better fits in a definition list.

# Checklist for author

* [x] Tested https://musicbrainz.org/recording/21bc4ba0-3b43-43b8-9702-f5f066c894e8


reosarevok (approved):
Does seem slightly better spaced I guess, and makes nothing worse, so go for it.

PR #2613, opened 2022-08-29
MBS-12556: Don't enter annotation edits unless text changes

reosarevok:
### Fix MBS-12556

Right now you can go to /edit_annotation and just re-enter the form as-is and it will create an edit. That doesn't match what we do elsewhere and it's also not useful in any way (even adding a changelog means nothing if nothing was changed).

As such, this blocks entering the edit if the text is the same as the text from the last annotation (regardless of whether a changelog or edit note is added).

Tested locally, plus added a test.


yvanzo (approved):
This single commit should have been split into two: one to refactor the test (with proper name and `create_edit` subroutine), a second one for the rest.

Otherwise LGTMBDNT but you did! :ship:

PR #2614, opened 2022-08-30
MBS-12558: Block linking to Wikipedia User: pages

reosarevok:
### Implement MBS-12558

User: pages are not really articles. Since the namespace is language-dependent, this will only block enwiki user pages, but that's probably most of them anyway.


yvanzo (approved):
LGTMBDNT :ship:

PR #2615, opened 2022-08-30
MBS-12528 / MBS-12530: More user-friendly reviews UX

reosarevok:
### Implement MBS-12528 / MBS-12530

This means every page load will currently hit CB, which might not be ideal. My understanding is that these API queries are actually cached in CB and as such it shouldn't be a huge issue, but we could add caching on our side too - the only problem is that it might be infuriating to be told "no releases, go add one?" and then do that, come back and still be told "no releases".

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -118,7 +118,7 @@ after 'load' => sub {
 
     if ($c->action->name ne 'edit') {
         # Only needed by pages showing the sidebar
-        $c->model('CritiqueBrainz')->load_display_reviews($release->release_group)
+        $c->model('CritiqueBrainz')->load_review_count($release->release_group)
             unless $returning_jsonld;
     }
 };
--- a/lib/MusicBrainz/Server/Controller/Role/Rating.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Rating.pm
@@ -6,6 +6,20 @@ use MusicBrainz::Server::Constants qw( %ENTITIES );
 
 requires 'load';
 
+after 'load' => sub {
+    my ($self, $c) = @_;
+
+    my $entity = $c->stash->{$self->{entity_name}};
+    my $entity_properties = $ENTITIES{ $entity->entity_type };
+    my $returning_jsonld = $self->should_return_jsonld($c);
+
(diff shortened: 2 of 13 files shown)


yvanzo (commented):
The code looks do to what the commit messages say. But systematically hitting CB for each reviewable MB entity page might have too many drawbacks:
1. An unprecedented increase of traffic for CB API, despite the caching.
2. A increased loading time for the whole page to load peripheral data only.

Since you worked on the Reviews tab. How was `review_count` update for release group before the implementation of these tabs?

In the case it would not affect the scalability of the CB API (1), I would still recommend to address (2) by fetching CB reviews (count) asynchronously from the client, like we already do with WP articles.

Aerozol:
Cross-seeding LB + CB + MB data more is a big general goal imo. An efficient way to make it work opens a lot of doors, thanks for your work on this team! Fingers crossed.

Aerozol:
I know this is probably the equivalent of me saying ‘let’s paint this server red to make it go faster’…
But can we ‘just’ update a table item somewhere for all or most of these kind of entities? +1 when one is added, -1 when removed, and then call that?
Would save so much pointless clicking on so many pages if we could do it widely.
If that’s already what we’re doing sorry and thank you ;D

reosarevok:
The problem is that the review data is stored in CritiqueBrainz, so to get it to MusicBrainz we need to ask CB about it. I'll look today into making it call after page load rather than wait until CB answers, at least, as @yvanzo suggested.

reosarevok:
Oh, I missed your question about how it worked earlier, @yvanzo: AFAICT it just hit CB each time a release group overview page was requested. And as you can see from this PR, it also was hitting CB on every release load already, unless I'm missing something :) This does however add artist, recording etc loads too, not sure what's the proportion of artist page hits vs releases for example in MB.

jesus2099:
> @Aerozol
> _But can we ‘just’ update a table item somewhere for all or most of these kind of entities? +1 when one is added, -1 when removed, and then call that?_

> @reosarevok
> _The problem is that the review data is stored in CritiqueBrainz, so to get it to MusicBrainz we need to ask CB about it._

Or we could have an MB notification web service, that CB would called only each time a review is added or removed.

mwiencek:
@jesus2099 has a good point.  But another idea would be to just have MB keep another database connection to the critiquebrainz_db open and select the review count directly.

PR #2617, opened 2022-08-31
MBS-12562: Redirect row id url to gid url for works too

jesus2099:
# Problem

[MBS-12562](https://tickets.metabrainz.org/browse/MBS-12562): Work row ID URL is not redirected to GID (MBID) URL

All entities are using `LoadWithRowID` except works.

https://github.com/metabrainz/musicbrainz-server/search?q=LoadWithRowID

[lib/MusicBrainz/Server/Controller/Role/LoadWithRowID.pm](https://github.com/metabrainz/musicbrainz-server/blob/master/lib/MusicBrainz/Server/Controller/Role/LoadWithRowID.pm) allows redirection from row ID entity URL to GID (MBID) entity URL.

So, before this commit, https://musicbrainz.org/work/12431696 did display Ricochet work but it did not change URL to https://musicbrainz.org/work/3ba92b70-ef66-48b5-9c14-18ed6aa8afcc as it does for all other entity types.

# Solution

I just link `LoadWithRowID` in `lib/MusicBrainz/Server/Controller/Work.pm` as is already done with other entity types:

```perl
with 'MusicBrainz::Server::Controller::Role::LoadWithRowID';
```

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Work.pm
+++ b/lib/MusicBrainz/Server/Controller/Work.pm
@@ -35,6 +35,7 @@ with 'MusicBrainz::Server::Controller::Role::Load' => {
         },
     },
 };
+with 'MusicBrainz::Server::Controller::Role::LoadWithRowID';
 with 'MusicBrainz::Server::Controller::Role::Annotation';
 with 'MusicBrainz::Server::Controller::Role::Alias';
 with 'MusicBrainz::Server::Controller::Role::Details';


jesus2099:
Tested on localhost, now it not only displays the work page, but it also updates the URL to MBID URL instead of row ID.

reosarevok (approved):
I guess this makes sense for consistency at least.

PR #2618, opened 2022-09-05
MBS-12567: Fail gracefully when loading unsupported class

reosarevok:
### Fix MBS-12567

Calls to `ws/js/entity/$track_mbid` were failing with a can't locate `WS::js::Track` error. Tracks are apparently unsupported
and that's fine, but we should error gracefully. This uses the same `critical_error` method we already use elsewhere in `ws/js`.


reosarevok:

mwiencek (approved):

PR #2619, opened 2022-09-07
MBS-12566: Fix incorrect URL shortener detection

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12566

As the ticket explains, our RegExps should escape their dots (t.co
should not match taco), and also match the hostname correctly (t.co
should not match t.co.example).

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -33,6 +33,7 @@ import {compareDatePeriods} from '../common/utility/compareDates.js';
 import {isMalware} from '../url/utility/isGreyedOut.js';
 
 import isPositiveInteger from './utility/isPositiveInteger.js';
+import isShortenedUrl from './utility/isShortenedUrl.js';
 import HelpIcon from './components/HelpIcon.js';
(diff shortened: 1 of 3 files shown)


root/static/scripts/tests/index-web.js
```diff
@@ -11,7 +11,7 @@ require('./GuessCase');
 require('./guessFeat');
 require('./i18n');
 require('./i18n/expand2');
-require('./relationship-editor');
+//require('./relationship-editor');
```
reosarevok, line 14: This seems unrelated from the other changes? (it also causes an eslint warning)

reosarevok (changes requested):
One issue, but other than that this seems like a big improvement, thanks! Every time I added a new shortener I thought "we should add a test for this" but got lazy, that's what I get! (surprising this bug wasn't reported a lot sooner).

root/static/scripts/tests/index-web.js
mwiencek: Oops, I had commented out that test locally because it was failing for some reason, then accidentally committed it.

PR #2621, opened 2022-09-08
MBS-12572: Convert cdtoc/info to React

reosarevok:
### Implement MBS-12572

`first_track` and `last_track` seemed completely useless, so I'm just removing them and using a `map` on `track_details`. Since they didn't seem to be used elsewhere either, this also removes the methods.


lib/MusicBrainz/Server/Entity/CDTOC.pm
```diff
@@ -143,10 +131,18 @@ with 'MusicBrainz::Server::Entity::Role::TOC';
...
-    return $json;
+    my $data = {
+        %{ $self->$orig },
+        discid          => $self->discid,
+        freedb_id       => $self->freedb_id,
+        leadout_offset  => $self->leadout_offset,
+        length          => $self->length,
+        track_count     => $self->track_count,
+        track_details   => $self->track_details,
+        track_offset    => $self->track_offset,
```
mwiencek, line 142: Should probably add `0 + ...` to all the numeric ones to ensure Perl doesn't encode them as strings.  (Not sure if any can be `undef` though.)

reosarevok: In theory they can't, by the definition (if some of our data is bad, that I can't say)

mwiencek (approved):
tested locally, seems to work fine

mwiencek (approved):

PR #2622, opened 2022-09-08
MBS-12574: Convert cdtoc/set_durations to React

reosarevok:
### Implement MBS-12574

On top of https://github.com/metabrainz/musicbrainz-server/pull/2621


mwiencek (approved):
tested entering a set track durations edit locally, seems to work fine

PR #2623, opened 2022-09-08
Fix building CSS in production Docker images

mwiencek:
* CSS is generated via webpack/server.config.mjs.
 * 8ef906fb40b4d2b8edc20e026e99229a574fc5c6 changed our image build steps so that server resources are no longer built.  The consequence is that CSS is no longer synced to the gateways.
 * 8ef906fb40b4d2b8edc20e026e99229a574fc5c6 is reverted and a script is added to run the server build with lib/DBDefs.pm.sample instead.

### DIFF
--- a/docker/scripts/compile_resources_for_image.sh
+++ b/docker/scripts/compile_resources_for_image.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# Explanatory production notes:
+#
+# Building webpack/server.config.mjs is necessary in order to generate
+# *.css files in the Docker image, which are extracted from the image
+# in a Jenkins job and synced to a static resources volume on our gateway
+# servers.
+#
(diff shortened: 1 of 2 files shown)


mwiencek:
I tested this PR on the `test` branch while deploying relationship-editor-react to test.mb.

reosarevok (commented):
Cannot we just build CSS only instead of creating and then removing useless JS?

mwiencek:
Well, we use https://webpack.js.org/guides/asset-modules/ to handle outputting CSS.  Webpack converts .less imports to a URL and outputs CSS (piped from less-loader) to the build directory.  So it can't build CSS without at least parsing the JS for imports.  It's likely possible to have it suppress bundling the JS, but would require some Webpack configuration changes I'm not sure about.

reosarevok (approved):
If it's tested to work and the alternative is a lot more complex, go for it.

PR #2624, opened 2022-09-09
Cleaning up after edit details React conversion

reosarevok:
This removes/renames stuff now to make more sense now that all edit details files are React.


reosarevok:

mwiencek (approved):
Makes sense.  Thanks for cleaning that up!

PR #2628, opened 2022-09-09
MBS-12580 / MBS-12581 / MBS-12582: Convert cdstub/index to React

reosarevok:
### Implement MBS-12580 / MBS-12581 / MBS-12582

On top of my CD TOC work from https://github.com/metabrainz/musicbrainz-server/pull/2627 since it uses a shared function.

Tested manually with /cdstub/BjDm_3hGVComa8hVLX_zXF6fWb0- (single artist) and /cdstub/90F2R5UAcfzzrD7u4TuEdvQU2SA- (VA)

PR #2631, opened 2022-09-12
MBS-12587: Convert cdstub/error to React 

reosarevok:
### Implement MBS-12587

Split this into two components using `StatusPage`, rather than overloading one component with entirely different strings like we did with TT.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2628 (but could be merged separately if we prefer by cherry-picking on master).


mwiencek (approved):
I tested triggering some of these errors locally but not all of them.  The code looked fine on review.

PR #2633, opened 2022-09-12
MBS-12589: Return if direct search query times out

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12589

Return early if an existing Internal Server Error occurs within the direct/external search methods, instead of attempting to render the results page with an undef pager, triggering another error.

This also ensures a "Request Timed Out" page is shown to the user instead of "Internal Server Error" in cases where a statement timeout occurs, since it doesn't get confused by multiple errors in `$c->error`.


reosarevok (approved):

reosarevok:
Tested locally (with `alter database musicbrainz_db set statement_timeout = 2;`), worked great.

PR #2634, opened 2022-09-13
MBS-12590 / MBS-12591: Convert cdstub/logged_in to React and fix related bugs

reosarevok:
### Implement MBS-12590, fix MBS-12591

This also makes it so that you can reuse the already provided TOC while adding the release, which seems like an obvious thing to support. Sadly, since that's a `POST`, I don't think it's possible to do so without breaking the translation.
I'm just using `StatusPage` since this has the same generic "this is an error" concept as other CD stub errors. Also, "Add CD Stub" doesn't make sense as a title for a page which tells you you cannot do that, so I improved that.

See also previous two commits for bundled improvements / bug fixes (this page wasn't being shown *at all* right now AFAICT).

On top of https://github.com/metabrainz/musicbrainz-server/pull/2631


mwiencek (approved):
I didn't test submitting an actual cdstub, but I tested browsing the relevant pages as logged out/logged in and seeding the release editor from one.  The proposed changes seem sane to me.

PR #2635, opened 2022-09-13
MBS-12592: Convert cdstub/import to React 

reosarevok:
### Implement MBS-12592

This adds an actual button to submit the artist search form (it could only be submitted by pressing enter on the field before).

On top of https://github.com/metabrainz/musicbrainz-server/pull/2634


mwiencek (approved):
Tested the seeding locally while connected to pink, seemed to work fine.

root/cdstub/ImportCDStub.js
```diff
@@ -0,0 +1,105 @@
...
+              value={track.artist}
+            />
+          ) : null}
+        </>
+      ))}
+
+      <PaginatedResults pager={pager}>
+        <ul>
+          {artists.map(artist => (
+            <li key={artist.entity.gid}>
```
mwiencek, line 84: I saw a warning logged about items needing unique key props here.  I guess using the loop index provided by `map` would make more sense, since it's a static list anyway.

root/cdstub/ImportCDStub.js
```diff
@@ -0,0 +1,105 @@
...
+    </form>
+
+    <form action="/release/add" method="post">
+      <input name="name" type="hidden" value={cdstub.title} />
+      <input name="barcode" type="hidden" value={cdstub.barcode} />
+      <input name="mediums.0.toc" type="hidden" value={cdstub.toc} />
+      <input name="mediums.0.format" type="hidden" value="CD" />
+
+      {cdstub.tracks.map((track, index) => (
+        <>
```
mwiencek, line 60: Sorry, I think this is where the warning occurred.  Though the changing the artist one probably makes sense too in case an artist is duplicated.
```suggestion
        <React.Fragment key={index}>
```

PR #2637, opened 2022-09-15
Fix relationship editor bugs causing crashes or preventing URLs from being edited

mwiencek:
Fixes:
https://tickets.metabrainz.org/browse/MBS-12596
https://tickets.metabrainz.org/browse/MBS-12606
https://tickets.metabrainz.org/browse/MBS-12607


mwiencek:
@brainzbot test this please

reosarevok:
I see you implemented MBS-12608 (thanks!). Marked it at such. Are the two relationship editors always expected to exist in combination? (I expect there's no real cases where we'll want one but not the other, but just in case).

PR #2638, opened 2022-09-19
MBS-10861: Change time before auto-deletion to 48 hours

reosarevok:
### Implement MBS-10861

# Problem
Most entities get autoremoved in 24 hours, but merging them keeps them around for a minimum of 48 hours. As such, if you want to get rid of something as an autoeditor, it's very tempting to empty it and let it be autoremoved, while a merge would be preferable.

Additionally, it's not uncommon for people to add artists for their release or whatnot, get sidetracked, and find out they've been autoremoved by the time they try to finish the whole thing the next day. Or a relationship gets removed, and suddenly the entity is empty, and gets autoremoved before the editor can react (even if the editor who removed the relationship is nice and leaves a note!).

Relatedly, but on the other direction, we have had reports that incorrect artists are buggy because they do not get autoremoved. The reason is that the artist credits they were used in take 7 days to get removed, a lot longer than the 24 hours the artist itself will need once the AC is gone.

# Solution
This sets the `RemoveEmpty` time to 48 hours / 2 days, up from 1 day, giving users a bit more time to react to edits that would lead to autoremoval.
It also sets the `RemoveUnreferencedRows` time to 48 hours / 2 days, down from 7 days, which makes it more consistent and hopefully less confusing to users.

48 hours was chosen as this is the minimum time an open merge/remove edit needs to apply, so manual and automatic removals will be consistent.

# Testing
No specific testing done but the changes are trivial.

### DIFF
--- a/lib/MusicBrainz/Script/RemoveEmpty.pm
+++ b/lib/MusicBrainz/Script/RemoveEmpty.pm
@@ -84,7 +84,7 @@ sub run {
         "SELECT id
          FROM $entity T
          WHERE edits_pending = 0
-           AND (last_updated < now() - '1 day'::interval OR last_updated IS NULL)
+           AND (last_updated < now() - '2 day'::interval OR last_updated IS NULL)
            AND NOT (
                EXISTS (
                    SELECT 1
--- a/lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
+++ b/lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
@@ -4,10 +4,10 @@ use Moose;
 =head1 DESCRIPTION
 
 This script process the rows of the table 'unreferenced_row_log' that have been
-inserted more than 7 days ago.
+inserted more than 2 days ago.
 
 This table weakly references rows (of other tables) that have been temporarily
-unreferenced but might be referenced again by an edit, hence the 7 days delay.
+unreferenced but might be referenced again by an edit, hence the 2 days delay.
 
 When processing an 'unreferenced_row_log' row, the script removes the weakly
 referenced row if it is still unreferenced by checking 'ref_count', and removes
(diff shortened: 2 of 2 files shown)


mwiencek (approved):
Makes sense to me.

yvanzo (commented):
Changes to `RemoveUnreferencedRows` are completely unrelated to the ticket and should probably tracked with a separate ticket.

Just one question about this change: If an edit is made to reference a row again but that edit stays open to votes for a week, what happens? Might it be related to the reason for initially setting the waiting period to 7 days in this script?

reosarevok:
Re: your second question: that's a very good one! You're entirely right that the script doesn't take that into consideration, but it should. I don't think this is connected to this change, because the same issue can happen now (someone can enter an open edit changing a release to the unreferenced AC 6 days after it gets unreferenced, and the script will happily remove it the next day). But it should be fixed, I added MBS-12636 and will look into it now.

reosarevok:

yvanzo:

yvanzo:
I have probably been confused by the ticket which had an outdated title and a vague description.

Now that you replaced “Extend” with “Standardize” in the ticket’s title, it correctly includes the above-mentioned change to `RemoveUnreferencedRows`.

The ticket’s description, that can be useful to editors too, should probably be improved with actual changes, probably inspired by your very detailed section “Solution” above in this pull request’s description.

More generally, to ease following changes made to tickets, or to fix these if incomplete, a comment can be made about changes and the motivation for these.

yvanzo:

yvanzo (approved):
The commit message should be clarified too:
* It doesn’t explain why the same value is used for both. (Consistency is not a goal per se, it has a purpose, specific here.)
* It could also reuse some terms such as “Standardize” from better documented parts (pull request’s description/ticket’s title).

Otherwise the code looks good to me. :ship:

reosarevok:
@yvanzo: Made changes to ticket description and commit message, please confirm they look better now :)

yvanzo:
They do, thanks!

PR #2639, opened 2022-09-20
MBS-12623: Editor icon is too large

Aerozol:
Making the editor icon match the size of the other entity icons, at least until we get avatar support.

### DIFF
--- a/root/components/UserAccountLayout.js
+++ b/root/components/UserAccountLayout.js
@@ -66,7 +66,7 @@ const UserAccountLayout = ({
     {...layoutProps}
   >
     <h1>
-      <EditorLink avatarSize={31} editor={user} />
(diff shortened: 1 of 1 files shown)


root/components/UserAccountLayout.js
```diff
@@ -66,7 +66,7 @@ const UserAccountLayout = ({
     {...layoutProps}
   >
     <h1>
-      <EditorLink avatarSize={54} editor={user} />
+      <EditorLink avatarSize={31} editor={user} />
```
reosarevok, line 69: FWIW our CSS for other headers says `background-size: 32px`

Aerozol: Thanks, changed!

reosarevok (approved):
Seems sensible to me for now.

PR #2640, opened 2022-09-21
MBS-12550: Don't enter useless annotation edit in merge

reosarevok:
### Fix MBS-12550

The entity merge system was trying to merge the annotations in entities, but if all entities had the same annotation already, it still tried to enter an edit, triggering the `NoChanges` exception added by MBS-12556.

There seems to be no reason why we would even try to enter an edit in this case, since the annotation won't change. I'd also argue it's misleading to claim it's the "Result of `$type` merge" since it's just the same as before. This just checks the current annotation in the target entity, and skips entering an edit if the annotation resulting from the merge process would have the same text.

Also added a test for the case.

This needs to mostly be put on the cron container for now, so pointing it to beta, we can merge there and patch cron, as suggested by @yvanzo.

### DIFF
--- a/lib/MusicBrainz/Server/Data/EntityAnnotation.pm
+++ b/lib/MusicBrainz/Server/Data/EntityAnnotation.pm
@@ -102,6 +102,7 @@ sub merge
     my ($self, $new_id, @old_ids) = @_;
     my $table = $self->table;
     my $type = $self->type;
+    my $model = type_to_model($type);
 
     my @ids = ($new_id, @old_ids);
     my %entity_to_annotation = map { @$_ } @{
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/Data/EntityAnnotation.pm
```diff
@@ -118,18 +119,24 @@ sub merge
         )
     };
 
+    my $current_target_annotation =
+        $self->c->model($model)->annotation->get_latest($new_id);
+    my $current_target_annotation_text = $current_target_annotation
+        ? $current_target_annotation->text
+        : '';
```
mwiencek, line 126: Is there a reason this wouldn't also work?
```suggestion
    my $current_target_annotation_text = $entity_to_annotation{$new_id} // '';
```

reosarevok: ... I totally didn't see that option, but that seems perfect and so obvious in hindsight :eyes: Thanks.

mwiencek (approved):

PR #2641, opened 2022-09-21
MBS-12614: Better placement for Add collaborator button

reosarevok:
### Implement MBS-12614

The autocomplete input is very short, so it doesn't make sense for `form-row-add` to use a full standard `form-input-width` here. Since we already have `form-input-width-short` which fits better, I'm just adding a `form-row-add.short` combo to match.

image

### DIFF
--- a/root/static/styles/forms.less
+++ b/root/static/styles/forms.less
@@ -95,6 +95,8 @@ div.form-row-add {
     margin: (@form-margin / 2) 0;
 }
 
+div.form-row-add.short { width: @form-input-width-short; }
+
 /* these rows are missing a <label> on the left and use a margin instead. */
 div.form-row-add.nolabel { margin-left: @form-label-width + @form-margin; }
 
--- a/root/static/styles/forms.less
+++ b/root/static/styles/forms.less
@@ -2,7 +2,7 @@
 @form-margin: 12px;
 @form-label-width: 150px;
 @form-input-width: @form-half-width - @form-label-width - (@form-margin * 3);
-@form-input-width-short: 220px;
+@form-input-width-short: 260px;
 @form-icon-size: 16px;
 @form-bubble-label-width: @form-label-width;
 @form-bubble-input-width: 176px;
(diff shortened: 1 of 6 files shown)


reosarevok:

root/static/styles/forms.less
```diff
@@ -93,6 +93,8 @@ div.form-row-add {
     margin: (@form-margin / 2) 0;
 }
 
+div.form-row-add.short { width: @form-input-width-short + 2; }
```
mwiencek, line 96: What does the `+ 2` do? (It actually looks better if I remove that, since the buttons align perfectly.)

mwiencek (commented):
I was going to ask why they aren't aligned in the screenshot you gave, but I tested it locally and the icons are aligned for me (other than the 2 pixel offset), so dunno if that was a before screenshot or what :)

root/static/styles/forms.less
reosarevok: I just followed the +2 from form-row-add, but I'm fine with removing that if you feel it looks better :)

mwiencek: It looks better in FF at least.  In Chrome I get the same as your screenshot, but in FF if I remove the `+ 2` I see this, which looks perfect:
image
Dunno why the text input is 202px in FF and 158px in Chrome.

reosarevok: Me neither tbh. Wonder if we could port this to Autocomplete2 and get it to be consistent that way? :)

mwiencek: Pushed a commit for that.  If you'd prefer we can merge this first and I can move it into a separate PR.

reosarevok:
For some reason this seems to be breaking removing removed collaborators from private collection subscriptions? I can't easily figure out why tbh.

reosarevok:
Ok, the issue was with the test itself - what used to work to submit the form earlier no longer did now. Amended the test as needed, seems fine now.

mwiencek:
Fixed the `useLayoutEffect` warning too.

mwiencek (approved):

PR #2642, opened 2022-09-21
MBS-12593: Convert cdtoc/attach_filter_release to React 

reosarevok:
### Implement MBS-12593

This converts `attach_filter_release` to React, and since `move_search` was effectively meant to be the same, this just gets rid of that and combines the two pages.

Also actually linking to the discID rather than just printing it when moving because why wouldn't we when we have the data.


reosarevok:

root/static/scripts/common/components/CDTocReleaseListRow.js
```diff
@@ -0,0 +1,181 @@
...
+                  {' '}
+                  {mediumFormatName(medium)}
+                  {' '}
+                  {medium.position}
+                  {nonEmpty(medium.name) ? (
+                    <>
+                      {': '}
+                      {medium.name}
+                    </>
+                  ) : null}
```
mwiencek, line 125: Should `addColonText` be used here somehow?

root/medium/MediumTracklist.js
```diff
@@ -62,6 +62,7 @@ const MediumTracklist = ({
       <tr
         className={loopParity(index) + (track.editsPending ? ' mp' : '')}
         id={track.gid}
+        key={index}
```
mwiencek, line 65: I think you'll need to add a `key` to the data track "tr" above, too, otherwise you'll get warnings in that case (I saw some in the console and assume that is the reason).

root/static/scripts/common/components/CDTocReleaseListRow.js
```diff
@@ -0,0 +1,181 @@
...
+  associatedMedium,
+  cdTocTrackCount,
+  countInReleaseGroup,
+  release,
+  showArtists = false,
+  wasMbidSearch = false,
+}: Props): React.Element<typeof React.Fragment> => {
+  const $c = React.useContext(SanitizedCatalystContext);
+
+  const [hidden, setHidden] = React.useState<boolean>(true);
```
mwiencek, line 47: On the TT page it looks like you were able to expand/collapse each medium individually, whereas this controls all of them? (Even though there are still separate links for each medium, they all expand all mediums, which seems unintended.)

mwiencek (changes requested):

root/static/scripts/common/components/CDTocReleaseListRow.js
reosarevok: I guess we could construct a `const mediumHeading = mediumFormatName(medium) + ' ' + medium.position` string first and then do:
```
                  {nonEmpty(medium.name) ? (
                    <>
                      {addColonText(mediumHeading)}
                      {' '}
                      {medium.name}
                    </>
                  ) : mediumHeading}

```

Does that seem preferable?

root/static/scripts/common/components/CDTocReleaseListRow.js
reosarevok: Hmm. I cannot set the state for each medium inside `attachableMediums.map`, right? - would you suggest having some sort of object with id and a boolean for hidden, or having a different component for the medium display that gets called inside `attachableMediums`, or something else entirely?

root/static/scripts/common/components/CDTocReleaseListRow.js
yvanzo: If it is about i18n support:
- `root/static/scripts/common/components/MediumDescription.js` already has a message `{medium_format} {position}` that can likely be reused for `mediumHeading`.
- A similar message can probably be added for the quoted code: `{medium_heading}: {name}`.

root/static/scripts/common/components/CDTocReleaseListRow.js
reosarevok: Moved the code to display each medium to a new component with its own state which means these actually work separately from each other now.

root/static/scripts/common/components/CDTocReleaseListRow.js
reosarevok: Followed @yvanzo's suggestion here :)

root/static/scripts/common/components/CDTocReleaseListRow.js
```diff
@@ -0,0 +1,191 @@
...
+  });
+
+  const formatPositionAndName = nonEmpty(medium.name)
+    ? texp.l('{medium_format_and_position}: {medium_name}', {
+      medium_format_and_position: formatAndPosition,
+      medium_name: medium.name,
+    }) : formatAndPosition;
+
+  return (
+    <React.Fragment key={index}>
```
mwiencek, line 58: This key doesn't do anything because there's no list here. Keys are only needed on the `CDTocReleaseListRowMediums` elements themselves where they're used in a list. It appears this is the only use of the `index` prop, so it can be removed if so

mwiencek (commented):
Thanks, this seems to works nicely overall. I did notice a few small differences with the old page which may or may not be intentional:

 1. On the new /cdtoc/move search page you can't expand tracklists for media that can't have disc IDs. It doesn't show the expand link at all, which seems better -- on the old page it shows the link, but it expands to an empty tracklist, which was bugged. However, on the new /cdtoc/attach search page you *can* expand tracklists for such media.
 2. Previously the loop parity would alternate even between each subheading, but now it only alternates within each subheading.
 3. On the new page some of the table cells are not top-aligned, while on the old one all of them are.
 4. On the old /cdtoc/move search page, the header says "Attach CD TOC" even though I'm on /move. This one definitely seems like a bug that you fixed, I'm just confused how to even get the old "move" template to show up.

2 & 3 might only make sense if you load the old & new pages side by side. Honestly 2 isn't a big deal but 3 does look a little worse and is probably something simple.

mwiencek (approved):
We can ignore the table style issue, since I'm not sure the previous alignment was intentional, and it doesn't look that bad. Looks good to me otherwise, thanks!

reosarevok:
1 comes from an inconsistency where we didn't load tracks for `/move` but we did for `/attach` when the medium cannot have disc ids. There seems to be no good reason for that, so I added the same `grep` for `/attach`
2 is hopefully not a big deal, as you said (it could probably be fixed by passing indices around but not sure if worth it?
3 was actually unintentional in the old code - we found out that's actually coming through from `medium/tracklist.tt`. That said, I had (also unintentionally) failed to top-align medium tracklists when converting that to React - an extra commit here fixes that.
4 I didn't even notice :eyes:  but I guess it's fixed!

mwiencek (approved):

PR #2644, opened 2022-09-22
MBS-12594: Add shift-click selection in relationship editor

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12594


reosarevok (approved):
Tested it and it seems to work fine :)

PR #2645, opened 2022-09-22
MBS-12595: Set deletedCaption for batch relationship preview

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12595

Instead of showing "This entity will be created by this edit," which isn't true, show the "{n} recordings/works selected" message we already use elsewhere.


reosarevok (approved):
Seems to work nicely. Not sure if it being bracketed in the hover tooltip is ideal, but I don't really care too much either.

PR #2646, opened 2022-09-22
MBS-12597: Set min size for relationship editor works column

mwiencek:
I couldn't find a way to allow the recording side to expand more if needed while defaulting to 50/50.  Setting them to 60/40 just made the recordings always use 60%, even if there were more work relationships. Making them always equal at least makes the positions of the UI elements more consistent.


reosarevok (approved):
Seems sensible enough to me tbh.

PR #2647, opened 2022-09-22
MBS-12598: In relationship editor, open track ACs in new tab

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12598

I added 'target' to the list of shared props in DescriptiveLink, so that setting target="_blank" also applies it to any artist credit links. Then I changed the relationship editor's TrackLink component to make use of DescriptiveLink instead of re-implementing the artist credit display on top of EntityLink.

### DIFF
--- a/root/static/scripts/common/components/DescriptiveLink.js
+++ b/root/static/scripts/common/components/DescriptiveLink.js
@@ -47,6 +47,7 @@ const DescriptiveLink = ({
     showDisambiguation,
     showEditsPending,
     showIcon,
+    target,
   };
 
   const props = {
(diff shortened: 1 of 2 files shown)


mwiencek:
I also cherry-picked 7263c69ded3a690f095d2768d10f7f33d38401c1 here to avoid conflicts. (Not sure if this was reported, but I think it's correct?  There is a related issue, [MBS-12612](https://tickets.metabrainz.org/browse/MBS-12612), which I have a separate PR for.)

reosarevok (approved):
Seems to work fine. AFAICT your second commit actually fixes MBS-12612 anyway, but it might make sense to change back the default as per your other PR too.

PR #2648, opened 2022-09-22
MBS-12612: Default showDisambiguation to false in ArtistCreditLink

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12612

As far as I can tell, this was the behavior prior to 536a5e89c47f494a48e33eb0ecfede5797240300


mwiencek:
I don't think the second commit there fixes this, since it only applies to the release relationship editor (recording comments and track AC comments, specifically), whereas this applies to every page showing artist credits.

reosarevok:
Oh, my bad, I didn't notice that was not just about the relationship editor :) Anyway, fine to merge IMO!

PR #2649, opened 2022-09-22
MBS-12599: Relationship editor no longer warns of losing changes

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12599

Factor out the code that builds the edits for submission into `getAllRelationshipEdits`, so we can check if any changes actually exist without submitting them.  The new function is a generator, so we can bail out after one edit is seen and not have to build the whole lot.

### DIFF
--- a/root/static/scripts/release/components/ReleaseRelationshipEditor.js
+++ b/root/static/scripts/release/components/ReleaseRelationshipEditor.js
@@ -354,12 +354,21 @@ async function submitWorkEdits(
   }
 }
 
-async function submitRelationshipEdits(
-  dispatch: (ReleaseRelationshipEditorActionT) => void,
+function* getAllRelationshipEdits(
   state: ReleaseRelationshipEditorStateT,
-): Promise<void> {
+): Generator<
+  {
+    +edits: Generator<
+      [Array<RelationshipStateT>, WsJsEditRelationshipT],
+      void,
+      void,
+    >,
+    +entity: CoreEntityT,
+  },
+  void,
+  void,
+> {
   const seenRelationships = new Map();
-  let editCount = 0;
 
   function linkAttributeEditData(
     attr: LinkAttrT,
@@ -398,11 +407,6 @@ async function submitRelationshipEdits(
         editData.gid = entity.gid;
       }
       return editData;
-    } else if (entity.entityType === 'work') {
-      invariant(
-        entity._fromBatchCreateWorksDialog !== true &&
-        isDatabaseRowId(entity.id),
-      );
     }
     return {
       entityType: (entity.entityType: NonUrlCoreEntityTypeT),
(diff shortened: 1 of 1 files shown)


root/static/scripts/release/components/ReleaseRelationshipEditor.js
```diff
@@ -398,11 +407,6 @@ async function submitRelationshipEdits(
         editData.gid = entity.gid;
       }
       return editData;
-    } else if (entity.entityType === 'work') {
-      invariant(
-        entity._fromBatchCreateWorksDialog !== true &&
-        isDatabaseRowId(entity.id),
-      );
```
mwiencek, line 405: This invariant checked that batch-created works were submitted (i.e. created) before any relationships for them were, but it has to be removed because that can't be true before any edits are submitted.

root/static/scripts/release/components/ReleaseRelationshipEditor.js
```diff
@@ -1552,6 +1563,29 @@ let ReleaseRelationshipEditor: React.AbstractComponent<{}, void> = (
...
+
+  React.useEffect(() => {
+    const beforeUnload = function (event: BeforeUnloadEvent) {
+      if (state.submissionInProgress) {
+        return undefined;
+      }
+      if (hasPendingEdits) {
+        // Modern browsers don't actually display this string for security.
+        event.returnValue =
+          'All of your changes will be lost if you leave this page.';
```
mwiencek, line 1578: This is the string that the old code used, but I removed the `l()` call from it, because as noted by the comment, browsers don't actually display it anymore.

root/static/scripts/release/components/ReleaseRelationshipEditor.js
reosarevok: Do we still check that in any way?

root/static/scripts/release/components/ReleaseRelationshipEditor.js
mwiencek: You have to return a string for it to work, so you could just return `''`, but I kept the old one since it's self-documenting (and maybe some non-mainstream browsers still show it -- it wasn't that long ago that the mainstream ones did).

root/static/scripts/release/components/ReleaseRelationshipEditor.js
mwiencek: Well, it's checked by /ws/js/edit, so the request would fail.  But that's probably better since it would show an error message instead of crashing the whole page.

reosarevok: Can the user do anything if that happens? I assume the whole thing is sorted to happen in the right order elsewhere anyway and this just made sure bugs were detected? In which way detecting them via ws/js/edit seems fine too, at least if the error message returned is easy to report.

reosarevok (approved):
Seems to work nicely :)

root/static/scripts/release/components/ReleaseRelationshipEditor.js
mwiencek: Yeah, the submission is done in the right order, so I was mostly using it for debugging.

It should stop submitting edits as soon as a /ws/js error occurs, then show whatever error message it got at the bottom of the page near the edit note, and allow you to re-submit from where it left off (in case removing the offending edits helps).

PR #2650, opened 2022-09-22
MBS-12603: Allow removing work relationships from the works column

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12603

Adds a red "X" icon next to work names in the right column.  Clicking
that will remove all relevant recording-work links.  (Pending links will
be removed outright, and existing links will be marked for removal.)

### DIFF
--- a/root/static/scripts/relationship-editor/components/RelationshipEditor.js
+++ b/root/static/scripts/relationship-editor/components/RelationshipEditor.js
@@ -279,6 +279,9 @@ export const reducer: ((
   const writableState: {...RelationshipEditorStateT} =
     cloneRelationshipEditorState(state);
   runReducer(writableState, action);
+  if (__DEV__) {
+    deepFreeze(writableState);
+  }
   return writableState;
 });
 
@@ -476,10 +479,6 @@ export function runReducer(
     default: {
       /*:: exhaustive(action); */
     }
-  }
-
-  if (__DEV__) {
-    deepFreeze(writableState);
   }
 }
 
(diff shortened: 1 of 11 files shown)


root/static/scripts/relationship-editor/components/RelationshipEditor.js
```diff
@@ -477,10 +480,6 @@ export function runReducer(
       /*:: exhaustive(action); */
     }
   }
-
-  if (__DEV__) {
-    deepFreeze(writableState);
-  }
```
mwiencek, line 483: Since `runRelationshipEditorReducer` is now called in a loop, I had to move this to `reducer`.  This just makes sure we're not performing direct writes to any state (and is only enabled in development).

reosarevok:
I would suggest also showing something on red on the right side (probably the title line of the work?). Right now, nothing on the right side lets us know the work is the one being removed :)

mwiencek:
@brainzbot test this please

reosarevok:

reosarevok (approved):
The change itself looks good. One selenium test is failing, let's see if it passes on a re-test, but if not please take a look at it :)

reosarevok:

reosarevok:
Hmm. Tests break in different places different times, making me wonder if it's just selenium messing up again, but do check when you can :)

mwiencek:
@brainzbot test this please

mwiencek:
I fixed a couple tests locally but I think aretha is just overloaded right now.

PR #2651, opened 2022-09-22
MBS-12601: Limit batch work creation to where works aren't linked

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12601

### DIFF
--- a/root/utility/invariant.js
+++ b/root/utility/invariant.js
@@ -22,3 +22,8 @@ export default function invariant(
     throw new Error(msg);
   }
 }
+
+export function expect<T>(value: ?T): T {
+  invariant(value != null, 'Expected a non-null value');
+  return value;
+}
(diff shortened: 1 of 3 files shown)


root/utility/invariant.js
```diff
@@ -22,3 +22,8 @@ export default function invariant(
     throw new Error(msg);
   }
 }
+
+export function expect<T>(value: ?T): T {
+  invariant(value != null, 'Expected a non-null value');
+  return value;
+}
```
mwiencek, line 29: This small little utility allowed me to avoid a bunch of ugly `?.`, and nest some calls (avoiding temporary assignments/if statements) for `tree.find` above, which normally can return `null` if an item isn't found in the list, but here we know they exist -- just have to convince Flow.

mwiencek: Maybe `nonnull` is a better name (but I'd like to avoid something overly long).

reosarevok (approved):
Works as it should now :)

PR #2654, opened 2022-09-22
MBS-12610: Discard selected link type name when changing entity type

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12610

When changing the target entity type in the relationship editor dialog, discard the link type name if it belonged to a selection.  If there wasn't a selection, assume it may have been user-entered and preserve it.


reosarevok (approved):
Tested, works fine

PR #2656, opened 2022-09-22
MBS-12615: Fix compactEntityJson for other-document data

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12615

`compactEntityJson` is used to minify the relationship editor state before putting it into sessionStorage during submission.

If you create a new entity from an autocomplete search, the returned entity object's constructor will be from a different document (the iframe), so we can't compare it by reference to our `Object`.  To determine if something is a plain object, we need another way.  The simplest is to check if the constructor function converted to a string matches our `Object` converted to a string.


reosarevok (approved):
If I understood the bug properly, then it works now

PR #2657, opened 2022-09-22
MBS-12616: Fix crash in relationship dialog searching for release

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12616

The issue here was it trying to execute `ADD_NEW_ENTITY_TITLES.release()`, which doesn't exist.  I've made it check that that's a function first, and additionally changed `determineIfUserCanAddEntities` to always return `false` for releases.


reosarevok (approved):
Works great now

PR #2658, opened 2022-09-22
MBS-12619: <Enter> on a button shouldn't accept the relationship dialog

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12619


reosarevok (approved):
Works great.

PR #2659, opened 2022-09-22
MBS-12620: Autocomplete2: "Show more" missing

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12620

Previously we weren't using the returned number of pages in the results. I also fixed a minor issue where the highlighted index isn't preserved after clicking "Show more."


reosarevok (approved):
Tested it, works fine

PR #2660, opened 2022-09-22
MBS-12624: Autocomplete2: Improve static search results

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12624

This implements basic n-gram search for static-item lists in Autocomplete2.  Results are sorted by how many grams they match (proportional to the matching gram sizes) and further weighted by the size and position of the match (so prefix matches, or matches closer in length to the actual result, will be preferred).

The exact implemention was a bit of trial-and-error to see what worked well, so I'm sure we'll need to make some tweaks along the way.


reosarevok (approved):
A quick test shows this being an improvement and the matching comments seem sensible. Let's put it on beta and see if people find weird shit?

PR #2661, opened 2022-09-23
MBS-12636: Do not remove unreferenced ACs with pending edits

reosarevok:
### Fix MBS-12636

# Problem
We do not want an unreferenced artist credit to be removed if there's any pending edits editing them - that will fail the edit and go against the intent of letting people reuse credits.

# Solution
For `artist_credit` table ids, we also check `edits_pending` and skip the id, without removing it from the table, if it has open edits. Whenever it no longer has open edits, it will either be deleted as empty, or dropped from the table if it now has a `ref_count`.

# Notes
There's a second issue which probably cannot be solved in a trivial way, of artist credits not being edited themselves but being used in open edits. Is there even a way to check this without looking through all the `edit_data` for any edits which might use ACs? We could of course use `edits_pending` here too if we made these edits affect the AC's `edits_pending` in the first place, but that'd mean almost all popular ACs would be highlighted all the time and it sounds less than ideal. We could add a *second* temporary table for ACs being used in edits? Seems like a hack, but... I'm out of ideas otherwise. I guess we could also just let those be removed and not worry about it that much.

# Testing
Test added for this case.

### DIFF
--- a/lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
+++ b/lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
@@ -75,23 +75,28 @@ sub run {
         my $quoted_table_name =
             $self->c->sql->dbh->quote_identifier($table_name);
 
+        my $columns = 'ref_count';
+
+        if ($table_name eq 'artist_credit') {
+            $columns .= ', edits_pending';
+        }
+
         for my $id (@$row_ids) {
             capture_exceptions(sub {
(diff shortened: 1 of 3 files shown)


mwiencek:
For either case I think we should skip the AC if it has pending edits, not removing it from `unreferenced_row_log`, leaving it for processing again later.  But I don't think \#2 is going to be feasible and hopefully would be quite rare anyway.

mwiencek (approved):

PR #2663, opened 2022-09-24
MBS-12625: Autocomplete2: Show recent instruments

mwiencek:
This was already implemented, but there was an issue with 3ccc1d2656aec0c01beec78286f6750bf37945e1 where the filtering compared item IDs rather than entity IDs.  For the recent entities list, the item IDs will be strings of the form `entity.id + '-recent'` so that they don't conflict with the same entity in the list below.

The `recentItems` Flow type had to be changed to use `OptionItemT`, because `ItemT` doesn't have an `entity` property.


reosarevok (approved):
Seems to work fine, thanks!

reosarevok:

PR #2664, opened 2022-09-25
MBS-12629: Adding new instrument doesn't clear type-info cache

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12629

I found some related issues with `Data::LinkType` and `Data::Instrument`, so fixed those too.


mwiencek:
I added a test for MBS-12629 that checks if the cache is cleared after inserting the instrument, at least.

mwiencek:
@brainzbot test this please

reosarevok (approved):
LGTMBDNT

PR #2665, opened 2022-09-26
Relationship editor dialog layout improvements

mwiencek:
This pull request tries to improve elements of the relationship editor dialog to take up less space.  Some of the feedback came from MBS-11847 and MBS-12635, though I'm not sure if it resolves the latter issue.

### DIFF
--- a/root/static/scripts/relationship-editor/components/DialogEntityCredit.js
+++ b/root/static/scripts/relationship-editor/components/DialogEntityCredit.js
@@ -74,6 +74,8 @@ const DialogEntityCredit = (React.memo<PropsT, void>(({
 }: PropsT): React.MixedElement => {
   const creditedAsOrName = state.creditedAs || entityName;
   const origCredit = React.useRef(creditedAsOrName);
+  const inputRef = React.useRef(null);
+  const inputId = React.useId();
 
   function handleCreditedAsChange(event: SyntheticEvent<HTMLInputElement>) {
     dispatch({
@@ -100,9 +102,6 @@ const DialogEntityCredit = (React.memo<PropsT, void>(({
       value: (event.currentTarget.value: any),
     });
   }
-
-  const inputRef = React.useRef(null);
-  const inputId = React.useId();
 
   let changeCreditsSection;
   if (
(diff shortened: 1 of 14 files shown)


root/static/scripts/relationship-editor/components/DialogEntityCredit.js
```diff
@@ -190,35 +204,35 @@ const DialogEntityCredit = (React.memo<PropsT, void>(({
...
+            id={inputId}
+            onChange={handleCreditedAsChange}
+            placeholder={entityName}
+            ref={inputRef}
+            type="text"
+            value={state.creditedAs}
+          />
+          <HelpIcon content={helpText} />
+          {changeCreditsSection}
+        </>
```
reosarevok, line 229: Hmm. If we want to hide "Credited as" behind a click, I'd say it should be called "Add credit" or something, not "Credited as", which is not an action.

That said, since the box is already more compact now once you click, wouldn't it be easier to have them always on, but taking less space as they do now? I don't feel the button is saving any space now and it asks for an extra click. Something like this:
Screenshot from 2022-09-27 12-31-54

```suggestion
        <>
          <label
            className="credit-field"
            htmlFor={inputId}
          >
            {addColonText('Credited as')}
          </label>
          {' '}
          <input
            aria-description={helpText}
            aria-label={l('Credited as')}
            className="entity-credit"
            id={inputId}
            onChange={handleCreditedAsChange}
            placeholder={entityName}
            ref={inputRef}
            type="text"
            value={state.creditedAs}
          />
          <HelpIcon content={helpText} />
          {changeCreditsSection}
        </>
```

See also there's another bug here, added MBS-12640 for it :)

reosarevok (changes requested):
This all seems a lot better, thanks! I have one issue with the credited as, see comment - it should have one of two changes made :)

root/static/scripts/relationship-editor/components/DialogEntityCredit.js
mwiencek: Thanks for trying it!

I was mainly trying to reduce visual clutter with the button.  I'm fine with keeping the input visible, but then not sure how MBS-12635 is resolved.  It's true that I did remove some padding/border-spacing, but that doesn't seem to be what the ticket is asking.  (More like remove the input entirely, so I tried to find a middle ground.)

Clicking the button does auto-focus the input at least, but if you use the keyboard it does require an extra keystroke.

The only thing I don't like about the old (production) layout is that the size and position of the credited-as input depends on the length of the label text ("Credited as:"), which also varies per language.  So it looks a bit chaotic to me, since it's not aligned or sized consistently.  That's why I moved the "Credited as:" label to go in the left column before.

What if we also put the relationship credit inputs to the right of the target autocomplete/source entity link, just like the instruments have it?

reosarevok: That might work. I have no idea why that dialog is so small anyway - it's already too big to fit on a phone comfortably anyway, so why not make it take the space it needs :)

reosarevok: I'm also fine with not fully solving MBS-12635 anyway - I think the main issue was just how prominent these were, since the credit fields were already shown before, right?

mwiencek:
@brainzbot test this please

root/static/scripts/relationship-editor/components/DialogEntityCredit.js
mwiencek: Moved the credit inputs back to under the fields, though I dropped the label and set the placeholder to "Credited as" instead, like the instrument credits have it.

reosarevok (approved):
Looks better. Let's put this (and any other small fixes you have ready) on beta and write a forum post, then have it up for one more week or two before releasing it?

PR #2666, opened 2022-09-27
MBS-12639: Also update countries.pot from the DB 

reosarevok:
### Fix MBS-12639

This was removed from here with 1feff5c since it was looking at the wrong table.
ef1e431 made it look in the right place but I guess Ian forgot to actually add countries.pot back. Whoops.


phw (approved):

mwiencek:
@brainzbot test this please

mwiencek (approved):

PR #2667, opened 2022-09-27
MBS-12573: Drop unused degraded column from cdtoc

reosarevok:
### Implement MBS-12573

This was apparently used ages ago to store "a previous algorithm (less accurate) or things imported from freedb that didn't have the accuracy of our ids", said Rob, but "we may have dropped all those back in the day and never nuked the column".
It contains always `FALSE`, at the moment, and we seem to use it exactly nowhere.


yvanzo (changes requested):
1. The files under `docs/database_schema_diagrams/graphs/` and `docs/database_schema_diagrams/images/` should not be included as this change is breaking `cdtoc` layout and these files have to be regenerated for schema 28 anyway.
2. This is a breaking change for `mbdata` which is used by `sir`. It has to be tracked to ensure updating SIR at the same time as the database schema change.

reosarevok:
For 1) I dropped the change. For 2) I added SEARCH-706 and started https://github.com/reosarevok/mbdata/tree/schema-change-2023-q2

yvanzo (approved):
LGTMBDNT

mwiencek (approved):

PR #2668, opened 2022-09-30
MBS-12645: Add guess case to genre edit form

reosarevok:
### Implement MBS-12645

Genre names should always be lowercased, so it'd be nice to have a guess case button to click rather than fiddling with the name by hand.

### DIFF
--- a/root/static/scripts/genre/components/GenreEditForm.js
+++ b/root/static/scripts/genre/components/GenreEditForm.js
@@ -1,5 +1,5 @@
 /*
- * @flow
+ * @flow strict-local
  * Copyright (C) 2019 MetaBrainz Foundation
  *
  * This file is part of MusicBrainz, the open internet music database,
--- a/root/static/scripts/genre/components/GenreEditForm.js
+++ b/root/static/scripts/genre/components/GenreEditForm.js
@@ -7,13 +7,27 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import mutate from 'mutate-cow';
 import * as React from 'react';
 
 import {SanitizedCatalystContext} from '../../../../context.mjs';
-import type {GenreFormT} from '../../../../genre/types.js';
+import type {
+  GenreFormT,
+  WritableGenreFormT,
+} from '../../../../genre/types.js';
+import isBlank from '../../common/utility/isBlank.js';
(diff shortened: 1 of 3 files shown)


root/static/scripts/genre/components/GenreEditForm.js
```diff
@@ -26,17 +40,88 @@ type Props = {
...
+/* eslint-disable flowtype/sort-keys */
+type ActionT =
+  | {+type: 'update-name', +action: NameActionT};
+/* eslint-enable flowtype/sort-keys */
+
+type StateT = {
+  +form: GenreFormT,
+  +guessCaseOptions: GuessCaseOptionsStateT,
+  +isGuessCaseOptionsOpen: boolean,
+  +previousForm?: GenreFormT | null,
```
mwiencek, line 52: Is `previousName` used at all for genres?  I can only find references in root/static/scripts/alias/AliasEditForm.js.

root/static/scripts/genre/components/GenreEditForm.js
```diff
@@ -26,17 +40,88 @@ type Props = {
...
+function createInitialState(form: GenreFormT) {
+  return {
+    form,
+    guessCaseOptions: createGuessCaseOptionsState(),
+    isGuessCaseOptionsOpen: false,
+  };
+}
+
+function reducer(state: StateT, action: ActionT): StateT {
+  return mutate<WritableStateT, StateT>(state, newState => {
```
mwiencek, line 71: For simple reducers like this you can forgo using `mutate` if you want, since it only really helps for deeply-nested assignments, like say `newState.form.field.name = 'foo';`.  (IIRC, I actually didn't use it at all in the new relationship editor code.)
```suggestion
  const newState = {...state};
  // ...
  return newState;
```

root/static/scripts/genre/components/GenreEditForm.js
reosarevok: Nope. Whoops.

root/static/scripts/genre/components/GenreEditForm.js
reosarevok: This actually complains in `runNameReducer` so since it works I'd leave it as-is (it also seems more consistent for other editors which might need `mutate` more)

reosarevok:

PR #2670, opened 2022-10-03
MBS-12646: Can't change time on artist-event relationship

mwiencek:
The first issue here was that in `pushRelationshipHiddenInputs`, we weren't outputting attribute data for existing attributes that had changed text values.  Now we output all new attributes unconditionally. (Attributes are only removed by specifying a 'removed' flag.)

The second issue was that in `splitRelationshipByAttributes`, we were resetting the ID on existing relationships, which actually caused it to submit a new relationship instead of editing the existing one.

Test cases have been added for both issues.


reosarevok (approved):
Quickly tested and it worked, a quick check of the code seemed sensible enough, so if tests pass, feel free to merge.

PR #2672, opened 2022-10-04
MBS-11063: Convert event edit form to React

reosarevok:
### Implement MBS-11063

Since event forms do not display the "Ended" section of the `DateRangeFieldset` (we do store ended, but it's meaningless and we don't use it) this adds a way to hide it on request.

We also need to pass time as a child to `DateRangeFieldset`, since we want it to be part of the same fieldset as it is now.

This adds a basic React `Bubble` component. For now, it keeps using JQuery for positioning, for simplicity - eventually, we will probably want to drop that too.


mwiencek:
Tested this and fixed a few issues I found:
* Fixed an issue preventing seeded relationships from working
* Fixed an issue preventing the Enter key from working in the setlist textarea
* Added a client-side "required field" error message for the name field
* Converted the new Bubble component to use floating-ui instead of jQuery UI
* Added back event/layout.tt since it's needed by the new EAA forms
* Rebased on latest master

All the fields I tested seem to be working. For some reason, the browser isn't displaying the native "required field" popup if I attempt to submit the form without a name, so I added an error message for this in the form state.

mwiencek (approved):

yvanzo (changes requested):
Quickly tested and noticed that the localization is partly broken.

It might be indicative of some deeper issue about components’ hydration, possibly related to the missing name field error that @mwiencek worked around.

In details:

* The form title “Add event” is localized :white_check_mark: 
* The below “For more information, check the documentation.” isn’t localized :x:
* In the first fieldset:
  * The title “Event details” isn’t localized either :x:
  * The title of each field is localized :white_check_mark:
  * In the field Type:
    * When blanked, the content of the bubble “Select any type from the list to see its description. If no type seem to fit, just leave this blank.” isn’t localized :x:
    * Otherwise, the type names and descriptions are localized :white_check_mark:
    * The option “This event was cancelled.” isn’t localized :x:
  * The help paragraphs below the field Setlist aren’t localized :x:
* The second fieldset “Date period” is localized :white_check_mark:
* The third fieldset “Relationships” is localized :white_check_mark:
* The fourth fieldset “External Links” isn’t localized :x:
* The last fieldset “Edit note” is localized :white_check_mark:

To take two examples that work correctly: the component `DateRangeFieldset` uses a hook-based approach while the component `RelationshipEditorWrapper` is _non-hydrated_.

@mwiencek: Are these two approaches relevant case by case? Or is that a legacy due to different implementations made with different versions of React?

Some errors in the browser’s JavaScript console might be related too.

mwiencek:
> Quickly tested and noticed that the localization is partly broken.

I think it's just because the locations of some strings changed, but the .po files still contain the old locations.

Testing with the "es" translation, it works if I follow these steps first:

```sh
./po/update_pot.sh
msgmerge -U po/mb_server.es.po po/mb_server.pot
./script/compile_resources.sh
```

After following this, there are still some untranslated strings (e.g., "Select any type from the list ..."), but that seems to be because po2json doesn't include fuzzy matches by default.

These issues are a common source of pain when testing new PRs in development. I wonder if we should update the webpack config to perform the above `msgmerge` to temporary files, and also change `po2json` to include fuzzy strings?

yvanzo:
> I think it's just because the locations of some strings changed, but the .po files still contain the old locations.

I thought that the locations were just indicative for translators, but our webpack config does actually perform a [`msggrep` by message location](https://github.com/metabrainz/musicbrainz-server/blob/v-2024-07-11.0/webpack/client.config.mjs#L170).

> it works if I follow these steps first

Thank you, it works for me too.

> After following this, there are still some untranslated strings (e.g., "Select any type from the list ...")

This string has been modified without notice in this pull request, that’s the main issue and also a common issue.

> I wonder if we should update the webpack config to perform the above `msgmerge` to temporary files,

That is probably needed in development environment only though, as it should be the only place where PO files are not synchronized with POT files.

> and also change `po2json` to include fuzzy strings?

Fuzzy string should not be included, because these can come from Weblate where nearest source string’s translation can be automatically suggested and target strings can be manually marked as “Needs editing”; in both cases the fuzzy translation can have a distant if not unrelated meaning.

----

It would be helpful to have an automated way to follow string changes for any commit or pull request.

It could be done through a pre-commit hook, a GitHub webhook, or a GitHub action.

Alternatively or complementarily, we could have the development setup to update Gettext resources on any change to the source code. So far the automation we have is:
* For Perl only, [`plackup --reload` option](https://metacpan.org/dist/Plack/view/script/plackup#-r,-reload) at https://github.com/metabrainz/musicbrainz-docker/blob/v-2024-07-11.0/build/musicbrainz-dev/scripts/start.sh#L12
* For JS only, [webpack configuration watch option](https://webpack.js.org/configuration/watch/) via [`run.mjs`](https://github.com/metabrainz/musicbrainz-server/blob/v-2024-07-11.0/webpack/run.mjs#L79-L82) via [`compile_resources.sh --watch` option](https://github.com/metabrainz/musicbrainz-server/blob/v-2024-07-11.0/script/compile_resources.sh#L65) at https://github.com/metabrainz/musicbrainz-docker/blob/v-2024-07-11.0/build/musicbrainz-dev/scripts/update-javascript.sh#L11

Would it be reasonable to implement a third watcher using [`inotifywait`](https://manpages.debian.org/bookworm/inotify-tools/inotifywait.1.en.html) to trigger updating POT files (from the source code only), updating PO files and triggering the two other watchers to update Perl MO files and JS Jed files (and recompile)?

yvanzo:
Now I’m noticing that the relationship types’ name and description in the relationship editor are still not localized, even though the translations didn’t change for a long time and are still shown in the event edit form currently in production.

Might it be related to the different approach for the component `RelationshipEditorWrapper` mentioned in my initial comment?

mwiencek:
> Fuzzy string should not be included, because these can come from Weblate where nearest source string’s translation can be automatically suggested and target strings can be manually marked as “Needs editing”; in both cases the fuzzy translation can have a distant if not unrelated meaning.

Yeah, I was suggesting it for development only, in case a string is tweaked slightly and "fuzzy" is added by `msgmerge`. Though I guess we wouldn't be able to distinguish between cases where the strings actually should not be reused...

> It would be helpful to have an automated way to follow string changes for any commit or pull request.
>
> It could be done through a pre-commit hook, a GitHub webhook, or a GitHub action.

It would be very cool if after a PR was opened, or whenever a commit was pushed, a GitHub comment could be posted on the PR with a summary of the string changes.

> Would it be reasonable to implement a third watcher using inotifywait to trigger updating POT files (from the source code only), updating PO files and triggering the two other watchers to update Perl MO files and JS Jed files (and recompile)?

I think it would be useful.

 1. IIRC, Webpack currently does not watch .po files for changes, only JS/CSS (this could probably be fixed).

 2. There are already so many processes to run for a full MBS development setup (Webpack with `--watch` mode, plackup, the template renderer, ssssss.psgi, artwork-redirect, artwork-indexer, mb-solr, ...). I guess we can ignore the external ones for now, but we should probably have some kind of superdaemon that can manage all these watchers for us. 😅

yvanzo (changes requested):
When you rebased this pull request, you missed some changes that have been made in the meantime to the files being converted here.

That partly explains why some strings have changed and are not translated anymore.

Most of these changes can be found through those commits:
```sh
git log v-2022-09-06..master -- root/components/forms.tt root/event/create.tt root/event/edit_form.tt
```

root/static/scripts/edit/components/DateRangeFieldset.js
```diff
@@ -205,15 +207,18 @@ component _DateRangeFieldset(
...
-        onChange={hooks.handleEndedChange}
-      />
+      {hideEnded ? null : (
+        <FormRowCheckbox
+          disabled={
+            disabled ||
+            !isDateEmpty(partialDateFromField(subfields.end_date))
+          }
+          field={subfields.ended}
+          label={endedLabel ?? l('Ended')}
```
yvanzo, line 217: This would be a new string. Context is just lacking to match the existing string:
```suggestion
          label={endedLabel ?? lp('Ended', 'label ended boolean')}
```

root/event/EditEvent.js
```diff
@@ -0,0 +1,41 @@
...
+  eventDescriptions: {+[id: string]: string},
+  eventTypes: SelectOptionsT,
+  form: EventFormT,
+) {
+  return (
+    <EventLayout
+      entity={entity}
+      fullWidth
+      page="edit"
+      title={l('Edit event')}
```
yvanzo, line 28: This would be a new string. But the previous string “Edit” was lacking context for localization.

yvanzo (commented):
Spotted a couple more of changed strings

root/static/scripts/edit/components/DateRangeFieldset.js
reosarevok: No, this is not a MB label. This is anything-with-dates ended (label here is the *HTML* label). That said, that's not great for a context... we *could* actually have a label picker that looks at entity type and picks the right context?

root/event/EditEvent.js
reosarevok: Yeah I think this is a worthy change.

yvanzo: I would suggest to keep the old string but to add the context `verb, header`.

reosarevok: Why keep it though? I think it's really weird that we have "`Add event`" for creation but just "`Edit`" for editing. The only reason I can see is that we were reusing one file in `entity/edit.tt` so we didn't want to make it complicated, but if we're splitting them, `Edit event` seems a lot better to me. That said, it should still get context.

mwiencek: I wouldn't mind keeping "Edit event" since the current page title doesn't mention "event" anywhere.

yvanzo: I actually replied before @reosarevok’s reply was visible to me. I don’t mind the addition either, but it is still worth a context for localization.

root/static/scripts/edit/components/DateRangeFieldset.js
yvanzo: Or just keep the component’s property `endedLabel` mandatory instead of making it optional?

root/static/scripts/event/components/EventEditForm.js
```diff
@@ -0,0 +1,345 @@
...
+               respectively.`,
+            )}
+          </p>
+        </fieldset>
+
+        <DateRangeFieldset
+          dispatch={dispatchDateRange}
+          field={state.form.field.period}
+          hideEnded
+        >
```
yvanzo, line 270: To follow the suggestion to keep `endedLabel` mandatory at https://github.com/metabrainz/musicbrainz-server/pull/2672#discussion_r1688691359

```suggestion
        <DateRangeFieldset
          dispatch={dispatchDateRange}
          endedLabel={lp('Ended', 'event ended boolean')}
          field={state.form.field.period}
          hideEnded
        >
```

reosarevok: I just rechecked this and remembered that the whole point was to have a default to avoid passing a label here that will never be used, since we do `hideEnded`. I guess I could just pass the string 'Ended', untranslated, for a fallback if we hit a bug with `hideEnded`, and still keep it mandatory?

root/static/scripts/edit/components/DateRangeFieldset.js
```diff
@@ -162,10 +162,12 @@ export function reducer(
 }
 
 component _DateRangeFieldset(
+  children?: React.Node,
   disabled: boolean = false,
   dispatch: (ActionT) => void,
-  endedLabel: string,
+  endedLabel?: string,
   field: DatePeriodFieldT,
+  hideEnded?: boolean = false,
```
yvanzo, line 170: What is the reason for having two properties here?
Can’t `endedLabel` be optional and `hideEnded` not added?

root/static/scripts/event/components/EventEditForm.js
yvanzo: It seems to be needlessly complicated and dodgy. I replied through another conversation: https://github.com/metabrainz/musicbrainz-server/pull/2672#pullrequestreview-2196297666

root/static/scripts/edit/components/DateRangeFieldset.js
reosarevok: Just hide the whole ended section unless you pass a label? I guess that could work.

yvanzo (approved):
Changes look good. Tested it again. Thanks! :ship:

PR #2673, opened 2022-10-04
MBS-12652: Convert entity collection tab to React

reosarevok:
### Implement MBS-12652

Tested manually with the different possible combinations.


lib/MusicBrainz/Server/Controller/Role/Collection.pm
```diff
@@ -109,11 +109,16 @@ View a list of collections that this work has been added to.
 
         $c->model('Editor')->load(@public_collections);
 
-        $c->stash
-          (entity_type => $entity_type,
-           public_collections => \@public_collections,
-           private_collections => $private_collections,
-           template => 'entity/collections.tt',
+        my %props = (
+            entity => $c->stash->{$entity_name // $entity_type}->TO_JSON,
```
mwiencek, line 113: `$entity_name` is already defined as `my $entity_name = $params->entity_name // $entity_type;` so it doesn't seem like the fallback is needed here?
```suggestion
            entity => $c->stash->{$entity_name}->TO_JSON,
```

root/entity/Collections.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const Details = ({
+  entity,
+  publicCollections,
+  privateCollectionCount,
+}: DetailsProps): React$MixedElement => {
+  const entityType = entity.entityType;
+  const LayoutComponent = chooseLayoutComponent(entityType);
+  const publicCollectionCount = publicCollections.length ?? 0;
```
mwiencek, line 27: Assuming this is always an array, which the Flow type suggests, the length should always be a number.
```suggestion
  const publicCollectionCount = publicCollections.length;
```

mwiencek (approved):
Tested locally and seems to work exactly the same as before. Only difference I can see is that the editor links all show the pencil "avatar" icons now, but that seems fine.

reosarevok:
The selenium error doesn't seem to be in a page connected with changes here, but we might still need to take a look anyway in general :) Maybe it is connected and I'm missing how.

reosarevok:
> the editor links all show the pencil "avatar" icons now

That's expected from all React conversions, yes :)

yvanzo:
> > the editor links all show the pencil "avatar" icons now
> 
> That's expected from all React conversions, yes :)

I tested it locally and I don’t see where. Can you please provide screenshots (in ticket maybe)?

reosarevok:
In a page like `/artist/ae0b2424-d4c5-4c54-82ac-fe3be5453270/collections` "Cleaned up composers by reosarevok" will now how the editor entity icon before "reosarevok", like we do elsewhere when linking the editor entity in React pages. This is the standard - we just never implemented the icons on TT because it was a lot of extra work and we are getting rid of the TT anyway.

image_2023-04-18_154038487

(adding it here because it doesn't seem relevant enough to add to the ticket to me)

yvanzo:
Thanks @reosarevok , I was not even looking at the right place.

yvanzo (approved):
LGTMBDNT :ship:

reosarevok:

PR #2677, opened 2022-10-09
Handle versions of Node.js > 16 in bin/sucrase-node

mwiencek:
Currently I cannot compile static resources with Node.js v18+ installed.  This changes how we handle such version numbers in bin/sucrase-node to be more tolerant.


mwiencek:
@brainzbot retest this please

reosarevok (approved):
LGTMBDNT

PR #2678, opened 2022-10-11
Make Autocomplete2 flow strict

reosarevok:
This also sets to strict files that depended only on this change.

I expect it's safe to ignore the 'any' issues since it is done elsewhere in the relationship editor code.


mwiencek (approved):
Thanks!

PR #2680, opened 2022-10-12
Fix warning when starting the server

yvanzo:
Amend commit dbfd10b from pull request #2669.


reosarevok:
Just in case, did you test it actually works? IIRC I tried that at first and it failed, but I might have typoed something :)

reosarevok (approved):
Ok, tested, does work, so I must have done a dumb last time :) Thanks!

yvanzo:
I didn’t test it but this notation is widely used in MBS already.

PR #2681, opened 2022-10-12
MBS-12552 (I),MBS-12661: Refactor Moose CoreEntity and fix URL last updated display

yvanzo:
# Problem

As part of MBS-12552, it has been decided to uniformize “core” (entity type) to “relatable” = “linkable”. The current Moose class `CoreEntity` doesn’t follow this meaning thus needing some refactoring.

Note: There still are other occurrences of “core”, only the ones in Moose classes are examined in this part I.

# Solution

* Move `name` Moose attribute from `CoreEntity` class to `Name` role;
* Also fix the name of the Moose role for pending edits by renaming it from `Editable` to `PendingEdits`;
* Remove collection and track from core entities in Moose as these are not relatable;
* Add the role `Linkable` to `CoreEntity` class as core entities are relatable;
* Add the role `LastUpdated` to `CoreEntity` as core entities are editable;
  Incidentally fix the bug MBS-12661 in URL’s sidebar.

# Checklist for author

* [x] Check the impact of reordering roles into the JSON output.
      (Possibly update tests or revert changes.)

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Collection.pm
+++ b/lib/MusicBrainz/Server/Entity/Collection.pm
@@ -6,7 +6,11 @@ use MusicBrainz::Server::Entity::Types;
 use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 use MusicBrainz::Server::Filters qw( format_wikitext );
 
-extends 'MusicBrainz::Server::Entity::CoreEntity';
+extends 'MusicBrainz::Server::Entity';
+with 'MusicBrainz::Server::Entity::Role::Editable';
+with 'MusicBrainz::Server::Entity::Role::GID';
+with 'MusicBrainz::Server::Entity::Role::Name';
+
 
 with 'MusicBrainz::Server::Entity::Role::Type' => { model => 'CollectionType' };
 
--- a/lib/MusicBrainz/Server/Entity/Collection.pm
+++ b/lib/MusicBrainz/Server/Entity/Collection.pm
@@ -7,7 +7,6 @@ use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 use MusicBrainz::Server::Filters qw( format_wikitext );
 
 extends 'MusicBrainz::Server::Entity';
-with 'MusicBrainz::Server::Entity::Role::Editable';
 with 'MusicBrainz::Server::Entity::Role::GID';
 with 'MusicBrainz::Server::Entity::Role::Name';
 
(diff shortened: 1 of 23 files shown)


reosarevok:
There's a `Data::Role::Editable` which is also about pending edits and it might make sense to rename it with `Entity::Role::Editable` as well :)

reosarevok:
There's plenty other entities that have `name` defined separately but could use the role you create in the first commit, btw - would it make sense to also add it to them? :)

yvanzo:
@reosarevok: I followed your second suggestion on updating `name` attribute in commits https://github.com/metabrainz/musicbrainz-server/pull/2681/commits/886a44cf51e0946d63d0ab2deaf794e5528de169 and https://github.com/metabrainz/musicbrainz-server/pull/2681/commits/b45fa76ecba3383f0eb1f3ed74d012c6c1102ab5.

Your first suggestion is sensible too but is beyond the scope of this pull request which is limited to Moose stuff, that is under `lib/MusicBrainz/Server/Entity`. It is never ending otherwise. Will look into it when addressing `Data` stuff.

reosarevok (approved):
What we have here seems sensible to me, unless I'm missing something. Given tests pass too, I'm going to expect I'm not missing something :)

reosarevok:
I'd really prefer not to have one Editable and one PendingEdits dealing with the same thing though :) Can you at least add a ticket so we don't forget we should deal with the clash? I can try and make the changes myself next week or whatnot.

lib/MusicBrainz/Server/Entity/Collection.pm
```diff
@@ -6,7 +6,11 @@ use MusicBrainz::Server::Entity::Types;
 use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 use MusicBrainz::Server::Filters qw( format_wikitext );
 
-extends 'MusicBrainz::Server::Entity::CoreEntity';
+extends 'MusicBrainz::Server::Entity';
+with 'MusicBrainz::Server::Entity::Role::GID';
+with 'MusicBrainz::Server::Entity::Role::Name';
+with 'MusicBrainz::Server::Entity::Role::PendingEdits';
```
mwiencek, line 12: `editor_collection` doesn't have an `edits_pending` column, so this can be omitted unless I'm missing something

mwiencek (approved):
> Check the impact of reordering roles into the JSON output.

This reminds me, but we aren't actually consuming roles correctly in MBS.  We should only be using a single `with` statement: https://metacpan.org/dist/Moose/view/lib/Moose/Manual/Roles.pod#CONSUMING-ROLES

Just something to note for future refactoring.

reosarevok:
I would like to think the JSON output changing in any significant way would have been caught by a test.

lib/MusicBrainz/Server/Entity/Collection.pm
yvanzo: Good catch, removed in commit https://github.com/metabrainz/musicbrainz-server/pull/2681/commits/3eea2724ecefc706a24a6f3b9663f2c2dd4fb2c3.

yvanzo:

PR #2683, opened 2022-10-12
MBS-12660: Convert user contact form to React

reosarevok:
### Implement MBS-12660


root/user/ContactUser.js
```diff
@@ -0,0 +1,86 @@
...
+}: Props): React.Element<typeof UserAccountLayout> => (
+  <UserAccountLayout
+    entity={user}
+    page="report"
+    title={l('Send Email')}
+  >
+    <h2>{l('Send Email')}</h2>
+
+    <form className="contact-form" method="post">
+      <FormCsrfToken form={form} />
```
mwiencek, line 50: Apparently `Form::User::Contact` is missing the `Form::Role::CSRFToken` role, so this doesn't actually do anything. But that was a problem with the TT page too.

mwiencek (approved):
Tested this with a local SMTP server and it seems to work fine.

root/user/ContactUser.js
yvanzo: @mwiencek: Shouldn’t this role be added to this class then? As a separate ticket/pull request?

reosarevok: I added it in a second commit. @mwiencek: Could you test it in your local setup again to make sure it works as expected? :)

mwiencek: It reloads the form now if the token is invalid, but doesn't display any error message. Probably because you are calling `$form->TO_JSON` before the form is validated.

reosarevok: I understand this just means we need to move the Node stash data setting below the `if ($c->form_posted_and_valid($form))` block, so I did that now.

reosarevok:

PR #2685, opened 2022-10-12
MBS-12663: Convert Reset password page to React

reosarevok:
### Implement MBS-12663

# Description
Just converts the Reset password page to React, adding `autoComplete` as per Chrome's automatic suggestion (seems like it would not hurt) but leaving it otherwise the same.

# Testing
Not tested since I don't have the actual setup I'd need to send emails and test the valid reset password process. @mwiencek, maybe you could test it with your setup? :)


yvanzo (approved):
I couldn’t test email sending either, but it seems safe to merge and then test it in beta by ourself.

Same comment about `autoComplete` as in pull request #2684. :ship:

PR #2686, opened 2022-10-13
MBS-12665: Widen DNB regex

reosarevok:
### Fix MBS-12665

# Problem
The DNB regex has been problematic for a while. It used to be too wide, and we tried to standardize it following the regex Wikidata uses, but it turns out that is instead too strict and is blocking some valid links, such as https://d-nb.info/97248485X

# Solution
outsidecontext got in touch with DNB and asked for better clarity on what the IDs can look like. The answer wasn't super clear, but the best we got from it is that it's probably safe to assume 9 or 10 character long IDs containing digits, hyphen or X, so that's what I'm implementing here. It could be possible to be more specific, maybe, but we've already been bitten by that, and honestly I think this is probably good enough for our needs.

# Testing
All the current tests pass, plus added a new one for the previously blocked URL.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1963,8 +1963,8 @@ const CLEANUPS: CleanupEntries = {
     restrict: [LINK_TYPES.otherdatabases],
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?d-nb\.info\//, 'http://d-nb.info/');
-      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac(?:\.htm\?)?.*\bquery=nid%3D(1[012]?\d{7}[0-9X]|[47]\d{6}-\d|[1-9]\d{0,7}-[0-9X]|3\d{7}[0-9X]).*$/, 'http://d-nb.info/gnd/$1');
-      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac(?:\.htm\?)?.*\bquery=idn%3D(1[012]?\d{7}[0-9X]|[47]\d{6}-\d|[1-9]\d{0,7}-[0-9X]|3\d{7}[0-9X]).*$/, 'http://d-nb.info/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac(?:\.htm\?)?.*\bquery=nid%3D([0-9X-]{9,10}).*$/, 'http://d-nb.info/gnd/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:[^\/]+\.)?dnb\.de\/opac(?:\.htm\?)?.*\bquery=idn%3D([0-9X-]{9,10}).*$/, 'http://d-nb.info/$1');
       return url;
     },
     validate: function (url, id) {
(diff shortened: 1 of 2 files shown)


reosarevok:
We got the following answer from DNB:

> Jeder bibliografische Metadatensatz und auch Normdatensatz (GND) ist mittels einem Primärschlüssel/Identifier eindeutig zu identifizieren.
> 
> Für bibliografische Metadaten wird der Identifier auch mit den Kürzeln IDN o. PPN benannt, für z. B. CD https://d-nb.info/357016572, lautet der Identifier/PPN/IDN „357016572“. Dieser ist immer enthalten im „Link zu diesem Datensatz“ (Permalink) in der Vollanzeige in unserem Katalog.
> 
> Die Struktur ist wie folgt:
> 
> Protokoll/Schema: https
> 
> Host: [d-nb.info](http://d-nb.info/)
> 
> Identifier: Nummer Identifier des Titeldatensatzes
> 
>  
> Bei Normdaten wird der Primärschlüssel auch GND-Nummer oder GND-ID genannt (z. B. Körperschaft/Musikgruppe https://d-nb.info/gnd/10298370-7 oder Werke https://d-nb.info/gnd/1135594082). Die Struktur enthält daher den Präfix „gnd“:
> 
> Protokoll/Schema: https
> 
> Host: [d-nb.info](http://d-nb.info/)
> 
> Präfix/Pfad: gnd
> 
> Identifier: GND-Nummer des Normdatensatzes

The suggestion from outsidecontext based on this was:

> They basically distinguish between IDs for bibliographic data and "Normdaten" (authority data) from the GND.
> 
> For bibliographic data the link structure is "[https://d-nb.info/{ID}](https://d-nb.info/%7BID%7D)", e.g. this is a CD: https://d-nb.info/357016572 .
> 
> The GND is the "Integrated Authority File" [1], basically a combined catalogue for persons, organizations, works and a few other entities. The GND URLs are always "[https://d-nb.info/gnd/{ID}](https://d-nb.info/gnd/%7BID%7D)". The German Wikipedia has a few more words about the format of the IDs for the GND [2]. Translated by Google Translate:
> 
> Each entity can be clearly identified by at least one identifier (GND-ID). Each identifier consists of the characters 0 to 9, X and a hyphen. An identifier for an entity of entity type p consists of 9 or 10 characters, it always starts with a 1 followed by 8 or 9 digits and ends with a check character 0-9 or X.
> 
> So the characters are limited to [0-9X-]. But above that it looks to me we should not make too many assumptions. It gives the example for entities of type p (persons), so this could be validated more strictly. But besides persons there are 6 other entity types. If we would need to restrict this further the valid formats for each entity type would need to be researched, I'm not sure this is worth it. Maybe it is safe to assume the last character is always [0-9X], the X in general is only allowed at the end and a hyphen can not appear at the start or end. Maybe something like [0-9][0-9-]*[0-9X] is a safe assumption to make.
> 
> The initial example you gave me was http://d-nb.info/97248485X , and with the above example it is clear that this is not a GND ID, so [3] does not even apply here. It belongs to the first category of bibliographic data (which would be mostly for "releases", containing things like CDs, audio books and books). They did not specify a format. But given the example it might be always numerical with X allowed, but that's just a guess. Maybe again we are on the safe side using the simplified regex I gave above (which would also include the hyphen).
> 
> [1] https://en.wikipedia.org/wiki/Integrated_Authority_File
> [2] https://de.wikipedia.org/wiki/Gemeinsame_Normdatei#Entit%C3%A4tsidentifikator
> [3] https://www.wikidata.org/wiki/Property:P227#P1793

PotcFdk:
> But given the example it might be always numerical with X allowed, but that's just a guess.

Not *always*: https://d-nb.info/998642312
I have not yet seen any bibliographic data id that did not follow the same format as the one specified for GND-IDs, but this is an anecdotal observation.

yvanzo (changes requested):
Can you explain in commit message why some entity types are still disallowed to link with DNB URLs having either `/gnd` or `/dnbn` as path prefix?

reosarevok:
Tried to clarify that a bit.

yvanzo (approved):
Thank you for the summary, it is much more clear now! :boat:

PR #2687, opened 2022-10-13
MBS-12666: Better edit display for stream for free external links

reosarevok:
### Implement MBS-12666

See commit messages.

Talked with @Aerozol, he thought these seem like reasonable improvements.

Proposed change:
Screenshot from 2022-10-13 08-59-24

Current:
Screenshot from 2022-10-13 09-02-56

I decided against changing the label of the checkbox to "has video" or "contains video" because we just show the attribute name elsewhere AFAICT, but I'm happy to change that if we decide it's worth it.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -1216,26 +1216,6 @@ const ExternalLinkRelationship =
                     ) : null
                   )
               }
-              {linkType &&
-                hasOwnProp(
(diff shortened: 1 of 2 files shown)


Aerozol:
Great work! I think we can workshop this (and a bunch more relationship UI stuff) more in future but this is a definite improvement no matter what imo

reosarevok:

mwiencek (approved):

PR #2688, opened 2022-10-13
MBS-12664: Convert the Register page to React

reosarevok:
### Implement MBS-12664

I added a `hidden` class rather than setting style because I was seeing issues with hydrating the styles, but actually it does seem a lot cleaner to do this anyway. We might want to use that elsewhere where we pass a style explicitly now.

I'm not `debounce`ing the email username checker because I don't think that's particularly useful and just complicates the code IMO.

Added `autoComplete` to the form because Chrome console suggested it and I can't see how it hurts in any case.


yvanzo (changes requested):
<del>The warning about username when it looks like an email is not shown anymore. Tested by creating an account with email address in username field.</del>

Edit: Actually it’s not showing up for me in `musicbrainz-docker` environment but it’s not new to this pull request. Still wondering why. Possibly need to tune CSP?

yvanzo (approved):
Tested successfully (by setting `Content-Security-Policy` unconditionally to `''` in `MusicBrainz::Server`), even tested (unintentionally) submitting the form after it expired.

The `debounce` feature is neither useful nor harmful for this warning, so either way is fine.

The `hidden` approach seems to be appropriate at first glance.

:ship:

PR #2689, opened 2022-10-13
MBS-12640: Preview always on for instrument relationships

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12640

The `LinkAttrT` type has optional `credited_as` and `text_value` fields.  In some cases, empty values will be undefined, and in other cases they'll be an empty string.  In the first case, fall back to an empty string before performing a comparison when calculating changes for previews.


reosarevok (approved):
Works nicely.

PR #2692, opened 2022-10-13
MBS-12643: Hide non-editable link orders in the relationship editor

mwiencek:
https://tickets.metabrainz.org/browse/MBS-12643


reosarevok:
This seems to work in the series page, but it still shows the number in the works pane of the release relationship editor :)

reosarevok (approved):
Nevermind, I retried and it indeed worked (a few different times too), maybe I didn't refresh the page properly earlier or something.

PR #2693, opened 2022-10-13
MBS-12667: Add Bugs! and Melon links to sidebar

Maxr1998:
Implementation for MBS-12667.

Used the Qobuz PR at #1923 for orientation.


yvanzo:
@brainzbot, add to whitelist

lib/MusicBrainz/Server/Entity/URL/Bugs.pm
```diff
@@ -0,0 +1,23 @@
...
+
+sub sidebar_name { 'Bugs!' }
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2022 MetaBrainz Foundation
```
yvanzo, line 17: It is fine but since it is your first contribution I have to ask you for confirmation: Do you want to assign copyright to the MetaBrainz Foundation or retain your own copyright instead? Both options are perfectly acceptable.

lib/MusicBrainz/Server/Data/URL.pm
```diff
@@ -106,6 +107,7 @@ my %URL_SPECIALIZATIONS = (
     'LyricEvesta'         => qr{^https?://lyric\.evesta\.jp/}i,
     'MainlyNorfolk'       => qr{^https?://(?:www\.)?mainlynorfolk\.info/}i,
     'Maniadb'             => qr{^https?://(?:www\.)?maniadb\.com/}i,
+    'Melon'               => qr{^https?://www\.melon\.com/}i,
```
yvanzo, line 110: In case someone is unintentionally copying the part `melon.com/…` only, we still want it to match as a Melon URL. So I suggest following the general scheme here:

```suggestion
    'Melon'               => qr{^https?://(?:www\.)?melon\.com/}i,
```

yvanzo (commented):
Hi @Maxr1998 and thank you for this contribution, it is very welcomed!

I made two comments, mostly a formality.

If you want to go further and also add support for automated selection of relationship type, it can be done in a separate pull request. The related files are:
```
./root/static/scripts/edit/URLCleanup.js
./root/static/scripts/tests/Control/URLCleanup.js
```

lib/MusicBrainz/Server/Entity/URL/Bugs.pm
Maxr1998: I'm happy with assigning copyright to the MetaBrainz Foundation, so we can keep it as is.

yvanzo (approved):
CircleCI is not running but I tested it locally by adding two URLs to an artist. Thank you!

PR #2695, opened 2022-10-14
MBS-12670: Convert the Delete user page to React

reosarevok:
### Implement MBS-12670

# Description
Mostly a straightforward React conversion (main visual difference is that since this is React now, the user links have entity icons).

We don't want form rows to be in the middle of the page, but it would be good to reuse `Form` components here, so I made it so `FormRow` can take a `hasNoMargin` parameter to add the existing no-margin class, and so `FormSubmit` can pass a class to its input.

# Testing
Manually, by navigating to the page as an admin and actually deleting a user, navigating to the page as the logged in user and making sure the content is the expected text, navigating to a different user page as a non-admin and ensuring we get the right forbidden message, and making sure navigating to a non-existent user name doesn't crash.

# Comments
Note that the page title and in-page h2 title are different. I'm unsure if this is intended, since we don't usually do this, and I wouldn't mind making both the same if you feel that'd be better :)


yvanzo (approved):
Roughly tested by locally deleting my own test account. :ship:

<del>Your manual test of trying to delete another user account as non-admin would be a good addition to the current CI tests `DeleteEditor`.</del>

yvanzo:
> Note that the page title and in-page h2 title are different. I'm unsure if this is intended, since we don't usually do this, and I wouldn't mind making both the same if you feel that'd be better :)

If we have to keep only one, I would prefer “Delete Account” which is more accurate.

PR #2696, opened 2022-10-17
MBS-12674: Don't assume offset data exists

reosarevok:
### Fix MBS-12674

Search results do not contain offset data, which was causing a warning for the `0 + leadout_offset` sum, and an ISE when trying to use a null `track_offset` as an array.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/CDStub.pm
+++ b/lib/MusicBrainz/Server/Entity/CDStub.pm
@@ -125,13 +125,17 @@ around TO_JSON => sub {
     $json->{date_added} = datetime_to_iso8601($self->date_added);
     $json->{discid} = $self->discid;
     $json->{last_modified} = datetime_to_iso8601($self->last_modified);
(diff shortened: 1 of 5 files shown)


mwiencek (approved):
A Selenium test would be nice, since we have the capability to test search pages there and apparently lack coverage.  Maybe open a ticket for it if you'd like to merge this first.

reosarevok:
How does the search testing work? It doesn't seem to be documented in HACKING.md - do we create our own search results or does it test based on the actual search server? (if the latter, then I guess any test would break if someone imported whatever CD Stub we're testing with and made it disappear from the list?)

mwiencek:
It uses a separate local search server.  I had a PR to document the setup a bit more at https://github.com/metabrainz/musicbrainz-server/pull/2540.

If you add the cdstub in t/sql/selenium.sql (which is data shared by all Selenium tests) or in a test-specific SQL file, it should be available in the index before the test runs.  (This is all handled by reset_selenium_env.sh.)

reosarevok:
Added a test now :)

PR #2697, opened 2022-10-17
MBS-12672: Add "Genre" to edit search like other entities

reosarevok:
### Implement MBS-12672


mwiencek (approved):
Tested and seems to work fine.

PR #2698, opened 2022-10-17
Upgrade Flow to 0.190.0

mwiencek:
Only minor changes necessary: https://github.com/facebook/flow/releases

### DIFF
--- a/root/static/scripts/common/i18n/expand2.js
+++ b/root/static/scripts/common/i18n/expand2.js
@@ -12,15 +12,9 @@ import * as React from 'react';
 
 /*
  * Flow doesn't have very good support for Symbols, so we use a unique
- * singleton class to fill its role.
+ * class instance to fill its role.
  */
-export class NO_MATCH {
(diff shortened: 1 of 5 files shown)


root/static/scripts/common/i18n/expand2.js
```diff
@@ -12,15 +12,9 @@ import * as React from 'react';
...
 /*
  * Flow doesn't have very good support for Symbols, so we use a unique
- * singleton class to fill its role.
+ * class instance to fill its role.
  */
-export class NO_MATCH {
-  static instance: ?NO_MATCH;
-
-  constructor(): NO_MATCH {
-    return NO_MATCH.instance || (NO_MATCH.instance = this);
```
mwiencek, line 21: Flow doesn't like this anymore and now requires all class constructors to return `void` since 0.188.0, to avoid confusing behavior.  The singleton stuff didn't really matter here anyway, since we never actually create more than one instance of this class.

PR #2701, opened 2022-10-18
Use Perl::Critic::Moose::ProhibitNewMethod

reosarevok:
We don't have anything hitting this now, but it seems like a good thing to keep in mind and a good introduction for Critic::Moose.

Eventually we'll probably want to look into https://metacpan.org/pod/Perl::Critic::Policy::Moose::ProhibitMultipleWiths and possibly some other rules in https://metacpan.org/pod/Perl::Critic::Moose

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   musicbrainz-tests:
     docker:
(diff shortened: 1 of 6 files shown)


reosarevok:

mwiencek (approved):

PR #2702, opened 2022-10-18
Use Perl::Critic::Moose::RequireCleanNamespace 

reosarevok:
See https://metacpan.org/pod/Perl::Critic::Policy::Moose::RequireCleanNamespace

I used `namespace::autoclean` rather than `no Moose` wherever possible, as per https://metacpan.org/pod/Moose::Manual::BestPractices


mwiencek (approved):
merge if tests pass

PR #2704, opened 2022-10-19
Avoid using multiple with in Entity:: files

reosarevok:
https://metacpan.org/pod/Perl::Critic::Policy::Moose::ProhibitMultipleWiths

Since this involves a lot of files, doing it in parts to ensure it's easier to find what broke if something does - this just for Entity::


mwiencek (approved):
it would be nice to have the style changes (removing newlines) in a separate commit

PR #2705, opened 2022-10-19
MBS-12673: Make it clearer that genres are added as tags

reosarevok:
### Implement MBS-12673

Some people were confused because Genres and Other tags were both top level headings on the sidebar, but only Other tags had a field to enter tags. This makes it so that the sidebar tag component has a Tags header, under which smaller Genres and Other tags subheaders appear.

When logged out:
Screenshot from 2022-10-19 10-15-08

When logged in:
Screenshot from 2022-10-19 10-14-57

I used `h3` as it's the obvious next step from `h2` - ~~that's currently orange, while the mockup was black. We can change the color of these if that seems helpful~~ and made it black so that it's more clearly different from the top headers.

### DIFF
--- a/root/layout/components/sidebar/SidebarTags.js
+++ b/root/layout/components/sidebar/SidebarTags.js
@@ -67,7 +67,9 @@ const SidebarTags = ({
             />
           ) : (
             <div id="sidebar-tags">
-              <h2>{l('Genres')}</h2>
+              <h2>{l('Tags')}</h2>
+
+              <h3>{l('Genres')}</h3>
               <div className="genre-list">
                 <p>
                   <TagList
@@ -78,7 +80,7 @@ const SidebarTags = ({
                 </p>
               </div>
 
-              <h2>{l('Other tags')}</h2>
+              <h3>{l('Other tags')}</h3>
               <div id="sidebar-tag-list">
                 <p>
                   <TagList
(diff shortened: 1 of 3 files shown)


Aerozol (approved):
Great!
re. making H3 black (still bold etc), looking at what it would change in the sidebar, I think that might be a good idea. Then I realised H3 is black on all the other pages… maybe it was changed here for a reason? But imo a good idea, the orange conflicts a bit.

reosarevok:
They're all orange on the sidebar - see the collections section, where "My collections" / "Other collections" are also orange h3s :)

Aerozol:
Yeah, but other pages, like: https://musicbrainz.org/user/aerozol/collections

I stopped being lazy and had a look for the special sidebar rule: #sidebar h3, .sidebar h3.
When changed:

image

Nicer imo.
I feel like making it all orange may have been by accident or something? I can’t think of a reason otherwise...

reosarevok:
Ok, I added a second commit to no longer make sidebar `h3` elements orange then :)

Aerozol:
Looks good to me!

reosarevok:
I know this is heavy bikeshedding by now, but do you feel "Genres:" or just "Genres" would be best?

atj:
> I know this is heavy bikeshedding by now, but do you feel "Genres:" or just "Genres" would be best?

FWIW the colon looks wrong to me - they're clearly sub-headers of "Tags" so you don't need them.

Aerozol:
Nothing wrong with building an insanely good shed :D 

Good points though, we don’t use them anywhere else so lets drop them.

PR #2707, opened 2022-10-19
MBS-12671: Convert Edit user page to React

reosarevok:
### Implement MBS-12671

# Description
Generally standard conversion. We needed a checkbox to take an `abbr` label so I made the form elements accept a `React$Node` label rather than just a `string`.

# Testing
Manually, by editing a user profile, including that editing stuff actually works and that field errors are shown if they exist (by updating the JSON form after validation) - for that, enter an invalid email, for example, or a - for us - invalid URL such as a localhost one.

# Action
* [x] If all previous React PRs are merged before this, add a second commit to delete unused `user/profile/layout.tt`


reosarevok:

yvanzo (approved):
We are really lacking CI tests for this sensitive form.

The conversion looks good otherwise. Roughly tested manually too.

:ship:

reosarevok:
Indeed, `t/lib/t/MusicBrainz/Server/Controller/User/Edit.pm` exists but does not test the admin-side form, just the user-side one. Added MBS-13184 to track this - not sure if the test file should be the same? Also, what is the right component for test-only tickets? :)

yvanzo:
It’s about account admin form, so I’ve just set the Account component.
The test is mainly about the Perl controller (preventing mess to happen) than just the React rendering.

PR #2708, opened 2022-10-19
Require use strict / use warnings with Perl::Critic

reosarevok:
See:

https://metacpan.org/pod/Perl::Critic::Policy::TestingAndDebugging::RequireUseStrict
https://metacpan.org/pod/Perl::Critic::Policy::TestingAndDebugging::RequireUseWarnings

These are always recommended to use, since they help write less crappy code. Adding this found a bunch of errors (fixed with commits before making everything strict) and warnings (fixed with commits after adding use warnings).

Some warnings I saw on test results that I haven't fixed yet, with some possible ways to deal with them, are at https://gist.github.com/reosarevok/cf05e577e2298dfe0c84c41cadba1c6c


mwiencek (approved):
Did not test, but reviewed each commit separately and it looks correct.

PR #2709, opened 2022-10-21
MBS-12681: Support YouTube Handle

yvanzo:
# Problem

## MBS-12681
New unique identifiers called handles will be available at YouTube.
These are almost the same as usernames and channel names but preceded with an at symbol (`@`).

# Solution

Add cleanup for YouTube handle.
No change is needed for sidebar display.
See the specs at: https://support.google.com/youtube/answer/11585688#tns


reosarevok (approved):
The theory at least seems correct. I assume that handles don't actually need to match channel names, so we can't cleanup one to the other, right?

yvanzo:
Right.

PR #2710, opened 2022-10-27
Fix Perl type constraint issues in Selenium tests

reosarevok:
I think the lack of a `use MusicBrainz::Server::Entity::Types` call on `Genre::DeleteAlias` is what was causing an error when running Selenium tests: `The type constraint 'Genre' has already been created`.

Moose seems to generate a constraint based on `isa => 'Genre'` calls, which then clashes with the one in `Types`, or at least that is my understanding from https://stackoverflow.com/questions/41550691/the-type-constraint-xyz-has-already-been-created


mwiencek (approved):
Thanks!

PR #2712, opened 2022-10-28
MBS-12684: Block trying to add a collaborator twice

reosarevok:
### Fix MBS-12684

We allowed submitting a collection form with the same collaborator listed twice, which then ISEd when we tried to insert it twice.

This makes sure each collaborator is only on the list the one time, and shows an error otherwise.


lib/MusicBrainz/Server/Form/Collection.pm
```diff
@@ -102,18 +102,31 @@ sub validate_collaborators {
...
-            my $editor = $self->ctx->model('Editor')->get_by_name($name_field->value);
+        my $name = $name_field->value;
+
+        if (defined $id) {
+            my $count =
+                grep { $_->field('id')->value eq $id } @collaborators;
+            if ($count > 1) {
+                $id_field->add_error(
+                    l('Please add each editor only once.')
+                );
```
mwiencek, line 115: Do you think it's necessary to show an error in this case?  We could also just uniquify the IDs without bothering the user.

reosarevok: I think it's useful because it feels unlikely the user did this willingly - so maybe they just clicked the wrong editor, and then if we accept it without erroring they don't realize something went wrong. But maybe I'm overthinking it.

mwiencek: Then maybe it should specify what the issue might be, like "“{collaborator}” was added more than once. Did you intend to select someone else?" I would certainly be confused why the server would bother to tell me this if I did it by accident.

reosarevok: That seems like a good text, used it.

mwiencek (approved):
Thanks! The only other improvement I was going to suggest is to only show the error for the first or last duplicate field (it looks a bit silly repeated multiple times). But you can also ignore that suggestion since I doubt this will even be triggered that often.

reosarevok:
I agree in theory but it seems more effort than it's worth for how common this should be, so let's improve it further if someone complains :)

PR #2714, opened 2022-11-02
Follow more Perl::Critic::Policy::Variables rules

reosarevok:

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -34,5 +34,6 @@ allow=refs
 
 [Variables::ProhibitUnusedVarsStricter]
 allow_unused_subroutine_arguments = 1
+[Variables::ProtectPrivateVars]
 [Variables::RequireLexicalLoopIterators]
 [Variables::RequireNegativeIndices]
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -32,6 +32,7 @@ allow=refs
 [ValuesAndExpressions::ProhibitMismatchedOperators]
 [ValuesAndExpressions::ProhibitMixedBooleanOperators]
 
+[Variables::ProhibitAugmentedAssignmentInDeclaration]
 [Variables::ProhibitUnusedVarsStricter]
 allow_unused_subroutine_arguments = 1
 [Variables::ProtectPrivateVars]
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -34,6 +34,7 @@ allow=refs
 
 [Variables::ProhibitAugmentedAssignmentInDeclaration]
 [Variables::ProhibitConditionalDeclarations]
+[Variables::ProhibitPunctuationVars]
 [Variables::ProhibitUnusedVarsStricter]
 allow_unused_subroutine_arguments = 1
 [Variables::ProtectPrivateVars]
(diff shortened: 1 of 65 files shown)


lib/MusicBrainz/Server/Data/ArtistCredit.pm
```diff
@@ -112,8 +112,8 @@ sub _find
     my @names = @{ $artist_credit->{names} };
 
     # remove unused trailing artistcredit slots.
-    while (!defined $names[$#names]->{artist}->{id} &&
-           (!defined $names[$#names]->{name} || $names[$#names]->{name} eq ''))
+    while (!defined $names[-1]->{artist}->{id} &&
+           (!defined $names[-1]->{name} || $names[-1]->{name} eq ''))
```
mwiencek, line 116: If you search for `- 1\]` in Perl files there are a couple more of the form `$foo->[scalar(@$foo) - 1]`.

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -85,13 +85,14 @@ sub search
...
+    my $extra_columns = '';
 
     if ($type eq 'artist') {
-
         my $where_deleted = 'WHERE entity.id != ?';
         $deleted_entity = $DARTIST_ID;
 
-        my $extra_columns = 'entity.gender, entity.area, entity.begin_area, entity.end_area,' if $type eq 'artist';
+        $extra_columns = 'entity.gender, entity.area,
+                          entity.begin_area, entity.end_area,';
```
mwiencek, line 95: My only problem with this is that it adds random indentation and whitespace to the string which can show up in query logs and make things harder to read.  Maybe concatenate the string across two lines instead?

lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -139,7 +139,12 @@ sub _load_mediums_limited {
     my ($self, $c, $mediums, $paged_medium, $medium_page_number) = @_;
 
     my @mediums = @{$mediums};
-    my $user_id = $c->user->id if $c->user_exists;
+
```
mwiencek, line 142: The added newline between variable declarations doesn't seem necessary.

.perlcriticrc
```diff
@@ -34,6 +34,7 @@ allow=refs
 
 [Variables::ProhibitAugmentedAssignmentInDeclaration]
 [Variables::ProhibitConditionalDeclarations]
+[Variables::ProhibitPunctuationVars]
```
mwiencek, line 37: This one will take me the longest to get used to since I know what all the symbols mean and don't know any of the English names without looking them up, but I can't argue that it's not more readable in the long run.

mwiencek (approved):

lib/MusicBrainz/Server/Data/Search.pm
reosarevok: Does that still happen with a heredoc? https://metacpan.org/pod/Perl::Critic::Policy::ValuesAndExpressions::ProhibitImplicitNewlines recommends concatenation over split lines and heredocs over concatenation, but if the heredoc is problematic on query logs then we can just concatenate.

reosarevok: I changed this to use heredocs since AFAICT that should be an improvement over the multiline stuff that the file was already using in most places.

.perlcriticrc
reosarevok: I would hope the English names at least seem connected with the things the symbols do :)

PR #2715, opened 2022-11-03
Follow more Perl::Critic::Policy::BuiltinFunctions rules

reosarevok:

On top of https://github.com/metabrainz/musicbrainz-server/pull/2714

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -7,6 +7,7 @@ severity=1
 [BuiltinFunctions::ProhibitShiftRef]
 [BuiltinFunctions::ProhibitStringyEval]
 allow_includes=1
+[BuiltinFunctions::ProhibitStringySplit]
 
 [CodeLayout::ProhibitHardTabs]
 [CodeLayout::ProhibitTrailingWhitespace]
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -3,6 +3,7 @@ severity=1
 
 [BuiltinFunctions::ProhibitBooleanGrep]
 [BuiltinFunctions::ProhibitLvalueSubstr]
+[BuiltinFunctions::ProhibitReverseSortBlock]
 [BuiltinFunctions::ProhibitShiftRef]
 [BuiltinFunctions::ProhibitStringyEval]
 allow_includes=1
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -9,6 +9,7 @@ severity=1
 allow_includes=1
 [BuiltinFunctions::ProhibitStringySplit]
 [BuiltinFunctions::ProhibitUniversalIsa]
+[BuiltinFunctions::ProhibitVoidGrep]
 
 [CodeLayout::ProhibitHardTabs]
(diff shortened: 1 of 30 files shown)


lib/MusicBrainz/Server/dbmirror.pm
```diff
@@ -3,6 +3,8 @@
 use warnings;
 use strict;
 
+use English;
```
mwiencek, line 6: This commit looks like a mistake, since `use English;` is already declared below (or it should be merged into the previous commit if it should be moved).

.perlcriticrc
```diff
@@ -5,6 +5,7 @@ severity=1
 [BuiltinFunctions::ProhibitLvalueSubstr]
 [BuiltinFunctions::ProhibitReverseSortBlock]
 [BuiltinFunctions::ProhibitShiftRef]
+[BuiltinFunctions::ProhibitSleepViaSelect]
```
mwiencek, line 8: I'm not sure we should add every random rule that we don't need TBH, since every extra rule it has to check surely slows down perlcritic incrementally?

lib/MusicBrainz/Server/Controller/ArtistCredit.pm
```diff
@@ -79,9 +83,9 @@ map {
         );
     };
 
-    find_meta(__PACKAGE__)->add_method($_ => $method);
+    find_meta(__PACKAGE__)->add_method($entity_type => $method);
     find_meta(__PACKAGE__)->register_method_attributes($method, [q{Chained('load')}, "PathPart('$url')"]);
-} entities_with('artist_credits');
+};
```
mwiencek, line 88: The semi-colon here is not needed

lib/MusicBrainz/Server/WebService/Validator.pm
```diff
@@ -204,7 +204,11 @@ sub validate_inc
     my %extra;
     for my $i (@inc)
     {
-        map { $extra{$_} = 1 } @{ $extra_inc{$i} } if (defined $extra_inc{$i});
+        if (my $extra_incs = $extra_inc{$i}) {
+            for my $extra_inc (@$extra_incs) {
+                $extra{$extra_inc} = 1
```
mwiencek, line 209: Missing semi-colon here.

.perlcriticrc
reosarevok: I can drop this one if you're not worried about us ever using it :)

lib/MusicBrainz/Server/dbmirror.pm
reosarevok: Yeah, that's a mistake on rebasing, thanks for catching it!

lib/MusicBrainz/Server/Controller/ArtistCredit.pm
reosarevok: I think later critic rules will catch any other that we missed here, but fixed this :)

mwiencek (approved):

reosarevok:

PR #2716, opened 2022-11-07
SEC-900: Upgrade nodemon to use fixed minimatch

yvanzo:
# Problem: SEC-900

For website rendering nodemon@2.0.19 was used, which depends on vulnerable minimatch@3.0.4 (CVE-2022-3517).

# Solution

Even though it seems unlikely that it might affect the MusicBrainz website, upgrading to nodemon@2.0.20 makes it use the fixed minimatch@3.1.2 instead.

# Notes

Other dependencies to minimatch@3.0.4 are used in a development context only (even glob@7.0.0 via @babel/cli@7.14.8 and shelljs@0.8.5).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "less-plugin-clean-css": "1.5.1",
     "leven": "2.0.0",
     "mutate-cow": "4.1.1",
-    "nodemon": "2.0.19",
+    "nodemon": "2.0.20",
     "pg": "8.3.3",
     "pg-cursor": "2.3.3",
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
Changing this seems sensible, all tests pass, if you tested this locally at all then IMO this is fine to merge.

yvanzo:
> if you tested this locally at all

Thanks for the reminder. I tested it locally just now and didn’t notice anything weird while browsing the website, even in server logs.

PR #2717, opened 2022-11-07
MBS-12552 (III): Move Entity::Role::Linkable to Relatable

yvanzo:
# Problem: Part of MBS-12552

Linkable is used in entityHref.js only and in a way broader meaning, whereas Relatable is used in a lot of files (including the very central Constants.pm and entities.mjs) and in the same meaning as in this role: “that can have relationships”.

# Solution

Move Entity role Linkable to Relatable

# Checklist for author

* [x] Check case-insensitive occurrences of linkable and relatable in the code base.
* [x] Rebase on master after #2706 has been merged.


reosarevok (approved):
Seems good, and especially given we use linkable elsewhere with a significantly different meaning, as you say :)

yvanzo:
Thanks!

PR #2718, opened 2022-11-07
MBS-9417 / MBS-12680: Fix some multiselect attribute bugs in the relationship editor

mwiencek:
* MBS-9417: It's now allowed to add the same instrument or vocal multiple times with different credits.  (These can't exist on the same relationship per our database schema, but they will be split into separate relationships by `splitRelationshipByAttributes`).

 * MBS-12680: A duplicate relationship is no longer added when changing an instrument or vocal attribute on an existing relationship.


reosarevok (approved):
Tested it locally, seems to work.

PR #2719, opened 2022-11-08
MBS-12694: Fix error submitting a recording artist relationship

mwiencek:
The issue stems from the recording AC still using the old Autocomplete component, which pushes a class instance (Artist, from root/static/scripts/common/entity.js) onto the recent items cache (which is shared with Autocomplete2).

Selecting this recent item causes the class instance to wind up in the relationship editor state, which is expected to only contain plain objects.  Later, when the state is serialized to sessionStorage before submission, `compactEntityJson` throws an error upon seeing the class instance, which it doesn't know how to convert to JSON.

### DIFF
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -339,7 +339,7 @@ $.widget('mb.entitylookup', $.ui.autocomplete, {
     if (hasID) {
       // Add/move to the top of the recent entities menu.
       pushRecentItem({
-        entity: data,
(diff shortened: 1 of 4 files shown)


reosarevok (approved):
This seems fine for now, given tests pass and we don't want to drop people's edits silently.

How big of a change would be to make the AC editor use Autocomplete2? If it's not trivial, can you make a ticket at least? :)

mwiencek:
Sure, I opened https://tickets.metabrainz.org/browse/MBS-12715 since it's not quite trivial. :)

PR #2720, opened 2022-11-08
MBS-12695: Fix encoding issue in "Add a new entity" dialog

mwiencek:
This was a regression introduced by 1786909dd2aa2a082a63b5dd6704eb21e8c71d69.  Fix it by decoding the path's query_form before forwarding it to the dialog action.


reosarevok (approved):
LGTMBDNT

PR #2721, opened 2022-11-08
MBS-12703: Add genre l_ tables to @CORE_TABLE_LIST

reosarevok:
### Fix MBS-12703

These were missing, so they were not being exported in data dumps.


yvanzo:

yvanzo (approved):
Good catch! Double-checked the list but did not test. :shipit:

yvanzo:
Hi @elliotchance, since you seem to need it we will make it appear starting from the next Wednesday dumps.

PR #2724, opened 2022-11-09
MBS-12679: Refresh dialog attributes when entity type changes

mwiencek:
This commit ensures that attributes in the relationship editor dialog are properly updated when:

 * The dialog opens and a link type is pre-selected (for cases where it's the only link type available).

 * The target entity type is changed.


reosarevok (approved):
Tested this and it seemingly worked fine - do mind the two unused imports triggering eslint though :)

reosarevok:
Removed the imports on a second commit, feel free to squash on merge if you want. Also rebased on the non-borked-for-selenium master

PR #2725, opened 2022-11-11
Follow Perl::Critic::Policy::CodeLayout rules

reosarevok:
See commit messages for details.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2723

PR #2726, opened 2022-11-11
Follow Perl::Critic::Policy::ValuesAndExpressions rules 

reosarevok:

On top of https://github.com/metabrainz/musicbrainz-server/pull/2725


reosarevok:

lib/MusicBrainz/Server/dbmirror.pm
```diff
@@ -3,6 +3,8 @@
 use warnings;
 use strict;
 
+use English;
```
mwiencek, line 6: Was this commit merged already or something? Not sure why `use English` in a single file is the only change here.

lib/MusicBrainz/Script/Role/IncrementalDump.pm
```diff
@@ -433,7 +433,7 @@ sub get_current_replication_sequence {
     my $replication_info_uri = $self->replication_access_uri . '/replication-info';
     my $response = $c->lwp->get("$replication_info_uri?token=" . DBDefs->REPLICATION_ACCESS_TOKEN);
 
-    unless ($response->code == 200) {
+    unless ($response->code == HTTP_OK) {
```
mwiencek, line 436: I wrote an awk script to check the diff for this commit (so, if a removed line contained "200", then the next added line should contain "HTTP_OK', etc. for each code/name pair defined in https://metacpan.org/pod/HTTP::Status#CONSTANTS). It didn't work in cases where the line was broken up into multiple, but it turned out that those were the only ones I had to check (other than a couple false positives). Everything else was apparently fine.

t/lib/t/MusicBrainz/Server/Controller/Area/Aliases.pm
```diff
@@ -33,7 +33,7 @@ test 'Area alias appears on alias page content and on JSON-LD' => sub {
 
     page_test_jsonld $mech => {
         '@id' => 'http://musicbrainz.org/area/106e0bec-b638-3b37-b731-f53d507dc00e',
-        'alternateName' => ["\x{30aa}\x{30fc}\x{30b9}\x{30c8}\x{30e9}\x{30ea}\x{30a2}"],
+        'alternateName' => ["\N{KATAKANA LETTER O}\N{KATAKANA-HIRAGANA PROLONGED SOUND MARK}\N{KATAKANA LETTER SU}\N{KATAKANA LETTER TO}\N{KATAKANA LETTER RA}\N{KATAKANA LETTER RI}\N{KATAKANA LETTER A}"],
```
mwiencek, line 36: In this commit, I'm not reviewing any of the changes under t/, because I assume the tests would fail if the wrong characters were used.

.perlcriticrc
```diff
@@ -44,6 +44,7 @@ allow=refs
 [TooMuchCode::ProhibitUnusedImport]
 [TooMuchCode::ProhibitUnusedInclude]
 
+[ValuesAndExpressions::ProhibitConstantPragma]
```
mwiencek, line 47: Re: the commit message, I don't see any uses of `REPLICATION_ACCESS_URI` that would benefit from compiler optimization.

lib/MusicBrainz/Server/dbmirror.pm
reosarevok: Yeah, two older commits (in unfixed versions) had slipped in. Dropped.

.perlcriticrc
mwiencek: While changing this one might be fine, changing the other `RT_*` constants in Replication.pm would break existing DBDefs files that reference them, so I think it's best to leave the file alone for now.

t/lib/t/MusicBrainz/Server/Controller/WS/js/Edit.pm
```diff
@@ -1432,7 +1432,7 @@ test 'MBS-13309: Cross-origin requests are limited to bot accounts' => sub {
     my ($c, $mech) = ($test->c, $test->mech);
 
     MusicBrainz::Server::Test->prepare_test_database($c);
-    MusicBrainz::Server::Test->prepare_test_database($c, <<~SQL);
+    MusicBrainz::Server::Test->prepare_test_database($c, <<~'SQL');
         INSERT INTO editor (id, name, password, privs, email, email_confirm_date, member_since, ha1)
         VALUES (2, 'bot_editor', '{CLEARTEXT}password', 2, 'bot\@editor.org', '2023-11-04', '2023-11-04', '');
```
mwiencek, line 1438: Don't need to escape `@` anymore.

mwiencek (approved):

PR #2727, opened 2022-11-14
MBS-12704: Remove historical "Watch artist" code

reosarevok:
### Implements MBS-12704

Two commits, one removes the code, another the database tables. I assume since these are not replicated, it's fine to drop all the tables and whatnot whenever we want.


admin/sql/updates/schema-change/28.all.sql
```diff
@@ -31,4 +32,17 @@ $$ LANGUAGE 'plpgsql';
...
+
+DROP TRIGGER IF EXISTS a_ins_editor ON editor;
+
+DROP FUNCTION a_ins_editor();
+
+DROP TABLE editor_watch_artist;
+DROP TABLE editor_watch_preferences;
+DROP TABLE editor_watch_release_group_type;
+DROP TABLE editor_watch_release_status;
+
```
yvanzo, line 47: It appropriately targets all instances and avoids failing if the trigger doesn't exist (as in mirrors).

yvanzo (approved):
Found and removed a couple more files in an extra commit.
LGTMBDNT

mwiencek (approved):

PR #2733, opened 2022-11-14
Add an easy way to set dialog attributes from userscripts

mwiencek:
This currently requires a lot of nested actions and is much more difficult than in the old relationship editor. (Based on feedback from @kellnerd.)

Example:

```JS
MB.relationshipEditor.relationshipDialogDispatch({
  type: 'set-attributes',
  attributes: [
    {
      type: {gid: 'd3a36e62-a7c4-4eb9-839f-adfebe87ac12'},
      credited_as: 'Narrator',
    },
  ],
});
```

### DIFF
--- a/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
+++ b/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
@@ -45,6 +45,7 @@ import {
   RelationshipSourceGroupsContext,
 } from '../constants.js';
 import type {
+  ExternalLinkAttrT,
   LinkAttributesByRootIdT,
   RelationshipDialogStateT,
   RelationshipStateT,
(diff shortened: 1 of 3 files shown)


reosarevok:
@kellnerd: should I throw this branch on test.mb or can you test it locally? :)

kellnerd:
Having it on test.mb would be helpful, I still haven't set up a local development server 😅

reosarevok:
This should be up on https://test.musicbrainz.org now

kellnerd (approved):
Thanks a lot! I have now adapted my userscript and my bookmarklet which require this feature and have tested them on test.mb. The relevant part (i.e. filling a relationship dialog) is working as expected for both attributes with `credited_as` or `text_value` properties.

Now I only have to figure out why the equivalent process of automatically adding relationships (without opening a dialog) fails as soon as I try to set attributes 😖 Any help with [debugging my sample code](https://gist.github.com/kellnerd/dc39236e58d5fb39b41a53a883eacb09) would be appreciated!

reosarevok (approved):
If @kellnerd is happy then we should release this, unless the second part of the review brings up something @mwiencek thinks needs to be fixed as part of this PR that is.

kellnerd:
I've seen that you've been discussing the next steps on IRC to finally get the React relationship editor into production and would love to get this PR released with it.

@mwiencek have you already looked into programatically creating new relationships with attributes using the `'update-relationship-state'` action? I'm not sure if this requires additional changes which should be part of this PR, but that's the last piece of my userscripts which is not working so far, although my sample code (see link above) looks plausible to me. If you could help me to solve this issue, I'm ready to finally publish my adapted userscripts and would be able to help other userscript authors to adapt theirs (most of them also need relationship attributes unfortunately).

mwiencek:
@kellnerd very sorry for not getting back to you here earlier

I believe the problem is that attributes are stored in the state as a tree rather than an array.  If you wrap the attributes array in `MB.tree.fromDistinctAscArray` it should work (note that this function expects the array to be sorted by link attribute type ID, but there's only one item here):

```
createRelationship({
	target: targetRG,
	linkTypeID: 742, // RG "part of" RG series
	attributes: MB.tree.fromDistinctAscArray([{
		type: { // full copy, I also tried to specify only the gid
			"free_text": true,
			"name": "number",
			"id": 788,
			"creditable": false,
			"entityType": "link_attribute_type",
			"parent_id": null,
			"child_order": 0,
			"root_id": 788,
			"description": "This attribute indicates the number of a work in a series.",
			"root_gid": "a59c5830-5ec7-38fe-9a21-c7ea54f6650a",
			"gid": "a59c5830-5ec7-38fe-9a21-c7ea54f6650a",
		},
		text_value: '18',
		typeName: 'number', // not sure if this attribute is really mandatory
		typeID: 788, // not sure if this attribute is really mandatory
	}]),
});
```

kellnerd:
Thanks for the hint, I'll have a look whether I'm able to write a lookup function which gives me the full attribute object for a given gid. I want to avoid hardcoding the required attribute objects into any usercript if possible :)

mwiencek:
We store all of the link attribute types in the linkedEntities module, but this isn't currently exported to the window.  I'll update the PR so that you can just do:
```js
MB.linkedEntities.link_attribute_type['a59c5830-5ec7-38fe-9a21-c7ea54f6650a']
```

PR #2734, opened 2022-11-15
MBS-12708: Report for lonely pseudo-releases

reosarevok:
### Implement MBS-12708

A pseudo-release should never be in a release group by itself or only with other pseudos.

Made this a release report, rather than RG, to allow it to be filtered by subscriptions.


mwiencek (approved):
Tested locally, works for me.

PR #2735, opened 2022-11-15
Move extends out of BEGIN blocks

reosarevok:
AFAICT this doesn't provide any particular benefit. The main reason we were probably doing this is because method attributes such as `Private`, `Path`, etc made Moose crash, but [`use MooseX::MethodAttributes;`](https://metacpan.org/pod/MooseX::MethodAttributes) seems to be all that is needed to avoid that.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Account/Subscriptions/Label.pm
+++ b/lib/MusicBrainz/Server/Controller/Account/Subscriptions/Label.pm
@@ -2,7 +2,7 @@ package MusicBrainz::Server::Controller::Account::Subscriptions::Label;
 use Moose;
 use namespace::autoclean;
 
-BEGIN { extends 'MusicBrainz::Server::Controller' }
+extends 'MusicBrainz::Server::Controller';
 
 with 'MusicBrainz::Server::Controller::Account::SubscriptionsRole';
 
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -2,7 +2,8 @@ package MusicBrainz::Server;
 
 use Moose;
 use namespace::autoclean;
-BEGIN { extends 'Catalyst' }
+
+extends 'Catalyst';
 
 use Class::Load qw( load_class );
 use DBDefs;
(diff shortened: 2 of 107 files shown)


mwiencek:
> AFAICT this is worse when writing code (I'd like Moose to tell me if something is borked ASAP, not just at runtime) and doesn't provide any particular benefit.

Not sure I follow this.  `BEGIN` runs during "compile time" (before any other code is executed), so moving the `extends` outside of `BEGIN` would be making it run later during runtime than it was before, unless I misunderstood.

> The main reason we were probably doing this is because method attributes such as Private, Path, etc made Moose crash, but [use MooseX::MethodAttributes;](https://metacpan.org/pod/MooseX::MethodAttributes) seems to be all that is needed to avoid that.

This is probably correct, since `Catalyst::Controller` itself imports `MooseX::MethodAttributes` - https://fastapi.metacpan.org/source/JJNAPIORK/Catalyst-Runtime-5.90130/lib/Catalyst/Controller.pm

Also, there is an example in the `Catalyst::Controller` documentation doing this, so I guess it's fine - https://metacpan.org/pod/Catalyst::Controller#CaptureArgs

It just feels a bit weird and undocumented. :)

lib/MusicBrainz/Server/Controller/Account/Subscriptions/Label.pm
```diff
@@ -1,7 +1,8 @@
 package MusicBrainz::Server::Controller::Account::Subscriptions::Label;
 use Moose;
+use MooseX::MethodAttributes;
```
mwiencek, line 3: Do we actually need this in files that don't have any appearances of method attributes, like here?

reosarevok: I guess not, I just missed that some of them were using BEGIN without needing it.

reosarevok:
Sigh, so I was confused again about how BEGIN works :D Well, anyway, it seems like having it all be checked together is the easiest and probably the most likely to find actual issues if they exist.

mwiencek (approved):

PR #2736, opened 2022-11-15
SEC-904: Upgrade babel-loader to drop loader-utils

yvanzo:
# Problem: SEC-904

For website rendering babel-loader@8.2.2 was used, which depends on vulnerable loader-utils@1.4.0 ([CVE-2022-37601](https://nvd.nist.gov/vuln/detail/CVE-2022-37601)).

I did not investigate how it might affect the MusicBrainz website exactly but it is used in production.

# Solution

Upgrading to babel-loader@9.1.0 makes it not to use loader-utils anymore.

From changes at https://github.com/babel/babel-loader/releases the new requirements by the v9 are already met by MBS. (In details: babel version, node version, and webpack version. There is no breaking change otherwise.)

# Checklist for author

* [x] Tested with a local instance

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "@popperjs/core": "2.5.3",
     "@sentry/browser": "5.10.2",
     "@sentry/node": "5.10.2",
-    "babel-loader": "8.2.2",
+    "babel-loader": "9.1.0",
     "balanced-match": "0.2.0",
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
Thanks!

yvanzo:

PR #2737, opened 2022-11-16
SEC-123: Upgrade @babel packages to drop kind-of

yvanzo:
# Problem: SEC-123

For website rendering the `@babel` packages that were used indirectly depended on vulnerable `kind-of@6.0.2 `(CVE-2022-37601).

# Solution

Even though it is unknown how it might affect the MusicBrainz website, upgrading to the latest `@babel` packages makes these not to use kind-of anymore.

Note that `@babel/register@7.18.9` still depends on `clone-deep@4.0.1` which in turn depends on `kind-of@6.0.2` or above so `kind-of@6.0.3` is actually used.

Note finally that the vulnerable `kind-of@3.2.2` still in-use is required by cliui only which isn’t used for website rendering in production.

# Checklist for author

* [ ] Tested with a local instance
  * [ ] Test editing a release


yvanzo:
Rebased on master for Selenium tests, but CI tests are known to fail anyway.

reosarevok:
You should be able to sidestep the bug this is hitting by changing the following section on init.js

```
  // Keep in sync with is_valid_edit_note in Server::Validation
  this.rootField.invalidEditNote = function () {
    return self.action === 'add' && (
      empty(self.rootField.editNote()) ||
      /^[\p{White_Space}\p{Punctuation}]+$/u.test(self.rootField.editNote()) ||
      /^\p{ASCII}$/u.test(self.rootField.editNote())
    );
  };
```

to this:

```
  /*
   * We precompile this regex to avoid a bug in regexpu-core transpiling,
   * since it only transpiles regular expression literals.
   */
  const onlyWhitespaceOrPunctuationSet =
    String.raw`[\p{White_Space}\p{Punctuation}]+`;
  const onlyWhitespaceOrPunctuationRegex = new RegExp(
    `^${onlyWhitespaceOrPunctuationSet}$`,
    'u',
  );

  // Keep in sync with is_valid_edit_note in Server::Validation
  this.rootField.invalidEditNote = function () {
    return self.action === 'add' && (
      empty(self.rootField.editNote()) ||
      onlyWhitespaceOrPunctuationRegex.test(self.rootField.editNote()) ||
      /^\p{ASCII}$/u.test(self.rootField.editNote())
    );
  };
```

yvanzo:
Thanks @reosarevok, can you please push your tested changes as a commit? I will cleanup my commit message after the PR has successfully passed tests.

mwiencek:
To be clear, pre-compiling or transpiling the regular expression is pretty important, isn't it?  From https://caniuse.com/?search=Unicode%20property%20escape we support browsers that don't have Unicode property escapes, so the suggested fix would cause the page to crash in these browsers AFAICT.

reosarevok:
Hmm, ok, I did not consider that it might crash not-so-old browsers. Do you have any good ideas? Some PRs suggest this is fixed in newer babel (someone said starting on 15.something), but we're seeing it on the latest version so that does not seem correct :/

mwiencek:
I noticed there were some conflicting babel packages (two different versions, even though they could be combined) in yarn.lock. The commit "Regenerate yarn.lock" resolves that and seems to fix the RegExp error too.

yvanzo:
I just dropped the commits that had been reverted. Thank you both for looking into it!

(I cannot self-approve my own pull request but I approve your respective commits which in turn approve mine.)

PR #2738, opened 2022-11-16
Fix warnings in DAOLogger

reosarevok:
We had a bunch of uninitialized values here, and they were triggering warnings I did not see earlier because I don't usually run `CATALYST_DEBUG`.


yvanzo (approved):
Tested by browsing a few pages, including the release editor. Not a single warning anymore. :ship:

PR #2739, opened 2022-11-17
Add another missing Types module import for Genre

mwiencek:
Follow-up to a case that was missed in a7d71920e6723be81b04abfb3551f026a4120078.

The PruneCache error is triggered randomly, but it's possible to reproduce by running the script in a loop.  I confirmed that this commit fixes the error, because the script will loop indefinitely, whereas normally it will fail in under a minute.


reosarevok (approved):
Thanks for catching this! Will merge already and rebase other PRs onto this, since it's obviously correct.

PR #2742, opened 2022-11-18
Validate invalid session ID cookies earlier

mwiencek:
If a request contains an invalid (as in malformed) session ID cookie paired with a `beta=on` cookie, the request will die before the beta redirect response is returned and result in a closed connection on the gateway.

In daf0647c3fdeebd1cc857c14678a619ebd04a3cf we started catching "invalid session ID" errors in `Controller::Root::begin` and logging them to Sentry.  The `beta=on` issue arises because a redirect is issued immediately instead of dispatching the request through `Controller::Root`.

The potential for an invalid ID to be returned is now handled at the source by wrapping `get_session_id`.  Since this is called very early in the dispatch process, we can't render an HTML bad-request page because there's no view available, so a text/plain error is returned instead. This is fine because 100% of requests with malformed session IDs are spam.

I no longer think we should log these errors to Sentry for the same reason, and because having them in the Docker logs is enough for forensics.

### DIFF
--- a/docker/musicbrainz-tests/DBDefs.pm
+++ b/docker/musicbrainz-tests/DBDefs.pm
@@ -148,6 +148,7 @@ sub DISABLE_LAST_LOGIN_UPDATE { 1 }
 # a different hostname alias.
 sub WEB_SERVER { 'mbtest:5000' }
 sub STATIC_RESOURCES_LOCATION { '//mbtest:5000/static/build' }
+sub BETA_REDIRECT_HOSTNAME { 'mbtest-beta:5000' }
 
 sub REPLICATION_USE_DBMIRROR2 { 1 }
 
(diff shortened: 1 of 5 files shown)


reosarevok:
Updated your commit to drop the unused imports that were triggering Critic.

yvanzo (approved):
Thank you! Tested locally by fiddling with `musicbrainz_server_session` and resetting the session store. :shipit: 


I confirm it isn’t an issue to return a plain text error since [`validate_session_id`](https://metacpan.org/pod/Catalyst::Plugin::Session#validate_session_id-SID) only checks the format of the session ID. Even if we lose or reset the session store, this plain text error won’t ever be returned to legitimate users (having outdated but still valid session IDs).

PR #2744, opened 2022-11-18
MBS-12713: Add KBR to the other databases whitelist

reosarevok:
### Implement MBS-12713

Tested locally + tests added


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2701,6 +2701,26 @@ limited_link_type_combinations: [
...
+             input_entity_type: 'label',
+    expected_relationship_type: 'otherdatabases',
+       only_valid_entity_types: ['artist', 'label'],
+  },
+  {
+                     input_url: 'https://opac.kbr.be/LIBRARY/doc/SYRACUSE/17060572',
+             input_entity_type: 'release',
+    expected_relationship_type: 'otherdatabases',
+       only_valid_entity_types: ['release'],
+  },
```
yvanzo, line 2723: There is a `clean` function so it seems worth testing it didn’t make any unwanted change.

yvanzo (approved):
The mapping looks good, especially since these are permalinks supported as such by the library. The code looks good too but did not test. For completeness expected clean URL should be tested even if unchanged (unless when there is no `clean` function).

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Hmm, I thought you had told me in the past that we shouldn't always use `expected_clean_url` because it's redundant, but maybe I misremembered :) Ok, added!

yvanzo: Right, it is redundant if it is the same input URL. It isn’t if it is a different URL pattern.

Maybe it could be added to the current documentation for this test field: https://github.com/metabrainz/musicbrainz-server/blob/v-2022-09-06/root/static/scripts/tests/Control/URLCleanup.js#L85-L90

yvanzo (approved):
Thank you for having added these optional but recommended test fields. :shipit:

PR #2745, opened 2022-11-19
Fix empty response when hitting a temporary IA delay

mwiencek:
`_detach_with_temporary_delay` incorrectly invokes `detach_with_error` by not wrapping the error key/value pairs in a hash ref.  In Perl, `=>` is the same as a comma except it auto-quotes bare words on the LHS.  So the result was that `'message'` was being passed as `$error`, and the `l(...)` string as `$status`.  This malformed `$status` value triggers a warning:

```
Argument "We\x{2019}ve hit a temporary del..." isn't numeric in numeric lt (<)
  at /home/musicbrainz/carton-local/lib/perl5/Catalyst/Plugin/Cache/HTTP.pm line 228.
Argument "We\x{2019}ve hit a temporary del..." isn't numeric in numeric ge (>=)
  at /home/musicbrainz/carton-local/lib/perl5/Plack/Middleware/FixMissingBodyInRedirect.pm line 18.
```

...and causes MBS to return an empty response and a 502 on the gateway server.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js.pm
@@ -292,14 +292,12 @@ sub _detach_with_ia_server_error {
 sub _detach_with_temporary_delay : Private {
     my ($self, $c) = @_;
 
-    $self->detach_with_error(
-        $c,
+    $self->detach_with_error($c, {
         message => l(
             'We’ve hit a temporary delay while trying to fetch metadata ' .
             'from the Internet Archive. Please wait a minute and try again.',
         ),
-        500,
-    );
+    }, 500);
 }
 
 sub cover_art_upload : Chained('root') PathPart('cover-art-upload') Args(1)


reosarevok:

reosarevok (approved):
The correction seems fine. We have plenty of `detach_with_error` calls in the form `detach_with_error($c, 'string', XXX)` too, maybe we should consolidate them to avoid issues like this one coming up by mistake again?

mwiencek:
> The correction seems fine. We have plenty of detach_with_error calls in the form detach_with_error($c, 'string', XXX) too, maybe we should consolidate them to avoid issues like this one coming up by mistake again?

Hmm, well not all of the calls can pass only a string, since at least once appears to include an `error_details` prop alongside the `message`.

reosarevok:
Yes, I meant the other way around, consistently using a props object rather than just a string :)

mwiencek:
I can change it for consistency, though not sure how much it'll help to prevent the same mistake.  It'll just cause an exception (trying to de-reference a string) instead of a warning, which I guess could make the error more obvious. :)

reosarevok:
Crashing sounds a lot more clear than warning to me :D

PR #2746, opened 2022-11-21
MBS-12716: Output the right JSON-LD @context

reosarevok:
### Fix MBS-12716

In order to make `@context` properly machine-readable, we should be pointing at the actual JSON-LD context file, as per https://schema.org/docs/developers.html

I'm keeping the link as HTTP because that's what we do all over our JSON-LD and schema.org say it's fine.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD.pm
+++ b/lib/MusicBrainz/Server/WebService/Serializer/JSON/LD.pm
@@ -3,7 +3,9 @@ use Moose;
 
 sub serialize {
     my ($self, $entity, $inc, $stash, $toplevel) = @_;
-    return $toplevel ? {'@context' => 'http://schema.org'} : {};
+    return $toplevel
+        ? {'@context' => 'https://schema.org/docs/jsonldcontext.json'}
+        : {};
 }
 
 __PACKAGE__->meta->make_immutable;
--- a/t/lib/t/MusicBrainz/Server/Controller/Area/Aliases.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/Area/Aliases.pm
@@ -34,7 +34,7 @@ test 'Area alias appears on alias page content and on JSON-LD' => sub {
     page_test_jsonld $mech => {
         '@id' => 'http://musicbrainz.org/area/106e0bec-b638-3b37-b731-f53d507dc00e',
         'alternateName' => ["\x{30aa}\x{30fc}\x{30b9}\x{30c8}\x{30e9}\x{30ea}\x{30a2}"],
-        '@context' => 'http://schema.org',
+        '@context' => 'https://schema.org/docs/jsonldcontext.json',
         'name' => 'Australia',
         '@type' => 'Country'
     };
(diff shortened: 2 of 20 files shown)


ahankinson:
I was the original reporter of the bug to @alastair . I would suggest using `https` directly, since the `http` version of the `@context` document returns a 301 to `https` anyway, and some clients may not be configured to follow redirects.

```
❯ curl -vvv "http://schema.org/docs/jsonldcontext.json"
*   Trying 142.250.81.238:80...
* Connected to schema.org (142.250.81.238) port 80 (#0)
> GET /docs/jsonldcontext.json HTTP/1.1
> Host: schema.org
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Location: https://schema.org/docs/jsonldcontext.json
< X-Cloud-Trace-Context: 9381d9eb98a14000e7c0d053b20b319f
< Date: Wed, 23 Nov 2022 18:59:04 GMT
< Content-Type: text/html
< Server: Google Frontend
< Content-Length: 0
<
* Connection #0 to host schema.org left intact
```

reosarevok:
Hmm. That makes sense, although such clients will be stumped with all the other links too, since all the MusicBrainz links are listed as `http` but should be redirects as well - and probably also many (but not all) external links. @mwiencek: opinions?

ahankinson:
The other URIs in the response *should* be just namespaced identifiers, so are not expected to be resolvable, whereas the document listed in the `@context` needs to be dereferenced in order to actually work in JSON-LD clients; it's what makes the "LD" part of JSON-LD happen. 

To expand: if you take a MBID like: https://musicbrainz.org/release/1834eae1-741b-3c03-9ca5-0df3decb43ea and retrieve the JSON-LD, you get a field like this:

`"@type": "MusicRelease",`

The Schema.org context, in a JSON-LD client like the [JSON-LD Playground](https://json-ld.org/playground/) is retrieved and, when parsed, contains two parts; one maps the value `MusicRelease` to `schema:MusicRelease`: 

```
"MusicRelease": {
    "@id": "schema:MusicRelease"
},
```

and the second expands the `schema:` [CURIE](https://en.wikipedia.org/wiki/CURIE) to a full namespace:

```
"schema": "http://schema.org/"
```

So the resulting value, when expanded, for the value in the MB `@type` field, is `http://schema.org/MusicRelease`. This is simply a namespaced URI that ensures this identifier is distinct from any other vocabularies' "MusicRelease" type. But it can't get expanded if it can't retrieve the context document, so until then it's just the literal text value "MusicRelease". 

Whether the namespace is http or https doesn't really matter, since it's only used as the identifier bit, not the actual locator. (URI not URL). 

To see this work, copy and paste the JSON-LD response into the JSON-LD Playground. By default it won't work until you change the value of `@context` to `https://schema.org/docs/jsonldcontext.json` in the JSON-LD. Then you will see the JSON-LD response get expanded in the tabs below; The easiest way I've found to visualize this is to flip to the "Table" tab where the expanded form produces triple statements like:

image

reosarevok:
Ok then :) Seems like a sensible reasoning, so updated that link to HTTPS.

mwiencek (approved):

PR #2747, opened 2022-11-22
MBS-12720: Set the remember_login cookie to HttpOnly

mwiencek:
This should not be accessible to JavaScript and is prone to XSS attacks otherwise.

Note: Once this is deployed, existing `remember_login` cookies will be overridden once the current session expires (which I believe only lasts 3 hours), so no other action needs to be taken on those.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/User.pm
+++ b/lib/MusicBrainz/Server/Controller/User.pm
@@ -269,6 +269,7 @@ sub _renew_login_cookie
             : '',
         samesite => 'Lax',
         $c->req->secure ? (secure => 1) : (),
+        httponly => 1,
     };
 }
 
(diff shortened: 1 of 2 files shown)


reosarevok:
This seems sensible.

* Do we have any other cookies that would benefit from the same thing?
* Should we have a test that checks this is not undone?

derat:
I saw this while looking through some of my old tickets. Is there any reason why this shouldn't go in?

mwiencek:
@derat Nope, looks like it just got forgotten about since it was submitted right before holidays (and I hadn't responded to reosarevok's comment).

> Do we have any other cookies that would benefit from the same thing?

The `musicbrainz_server_session` cookie is already HttpOnly, so I think that only leaves beta, lang, filter, and some client ones related to guess-case, which I don't think need this attribute.

> Should we have a test that checks this is not undone?

Added a test.

yvanzo (approved):
Added test has a good coverage.

LGTMBDNT :ship:

reosarevok (approved):

PR #2749, opened 2022-11-22
MBS-12722: Add IdRef to the other databases whitelist

reosarevok:
### Implement MBS-12722

This is basically a French authority control page, kind of ISNI-like.


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2352,6 +2352,14 @@ limited_link_type_combinations: [
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://www.ibdb.com/broadway-cast-staff/antonin-leopold-dvorak-447817',
   },
+  // IdRef
+  {
+                     input_url: 'http://idref.fr/172248248',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://www.idref.fr/172248248',
+       only_valid_entity_types: ['artist', 'label', 'place', 'series', 'work'],
```
yvanzo, line 2361: It seems to be too restrictive. I added a comment to the ticket.

yvanzo (commented):
LGTMBDNT except that it seems to be too restrictive about entity types.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Added all but event so far, see ticket - can add event too if you're sure.

reosarevok: My bad, we don't have an `otherdatabases` rel for areas so I took that one off too :) I marked in the commit message that areas are supported but I'm not adding an `otherdatabases` rel for them just for this - I don't expect our area editors are going to add these by hand anyway, so if we change our mind we can import them from Wikidata later on.

reosarevok:

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: Perfect, thank you!

yvanzo (approved):
LGTMBDNT :shipit:

PR #2753, opened 2022-11-25
Avoid using multiple with in misc files

reosarevok:
https://metacpan.org/pod/Perl::Critic::Policy::Moose::ProhibitMultipleWiths

Since this involves a lot of files, doing it in parts to ensure it's easier to find what broke if something does - this is for the rest of non-Controller:: files.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2752

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
@@ -91,8 +91,10 @@ role {
         }
     };
 
-    method valid => sub {
-        my ($self) = @_;
+    # We would probably want 'override' here but 'around' is needed
+    # to work around method name conflicts
+    around valid => sub {
+        my ($orig, $self) = @_;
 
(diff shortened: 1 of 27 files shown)


reosarevok:

lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
```diff
@@ -91,8 +91,8 @@ role {
         }
     };
 
-    method valid => sub {
-        my ($self) = @_;
+    around valid => sub {
+        my ($orig, $self) = @_;
 
         return 1 unless $self->operator_cardinality($self->operator);
```
mwiencek, line 97: This line should probably be replaced with `return 0 unless $self->$orig;`, since the base method checks cardinality. (Otherwise, there's no point to using `around`, and `override` should be used instead.)

mwiencek (approved):

PR #2756, opened 2022-11-30
MBS-12725: Add "Refine this search" link on Votes by [user] page

reosarevok:
### Implement MBS-12725

This is not an exact match, since the edit search cannot currently match the "time of last vote by the selected user" edit sort `/votes` uses, but the content itself should be the same and allowing to refine it further seems sensible.


reosarevok:

mwiencek (approved):

PR #2759, opened 2022-12-02
MBS-12735: Handle Boomplay URL for both editing and rendering

yvanzo:
# Problem: MBS-12735

Boomplay URLs are neither normalized nor displayed in the sidebar.

# Solution

Display in the sidebar and handle editing as follows:

* Auto-select “free streaming” relationship type
  (See the rationale in comments to the ticket)
* Normalize URLs (see test examples) to the following canonical form
  `https://www.boomplay.com/<type>/<id>` used on the website and in
  their link sharing feature.
* Validate canonical URLs with the corresponding entity type only:
  * `albums` with MB Release
  * `artists` with MB Artist
  * `songs` with MB Recording

Specifically blocking `playlist` URLs which content is too volatile.

# Todolist for author

* [x] Locally test displaying Boomplay URLs in the sidebar
* [x] Locally test editing legitimate MB entities with Boomplay URLs
* [x] Locally test editing illegitimate MB entities with Boomplay URLs
  Note: It doesn’t block linking to MB works but it seems to be the expected behavior for now.


yvanzo:
@reosarevok: Is it just me or does editing work accept any Boomplay URLs even though it is supposed to be restricted to artists, recordings, and releases? (Even after fully reloading the page.)

I don’t see any issue in the added code so I wonder if it might be a more general bug or if I’m just missing something else.

reosarevok:
It's only restricted for entities that have `streamingfree` link types. That's how it has always worked AFAICT.

reosarevok (approved):
Tested locally, seems good :)

PR #2743, opened 2022-11-18
MBS-12445: Handle Audiomack URLs

reosarevok:
### Implement MBS-12445

Tested locally (plus added cleanup tests)

PR #2760, opened 2022-12-02
Fix an ISE when displaying Audiomack link in the sidebar

yvanzo:
# Problem

ISE when displaying a page with Audiomack URL in the sidebar

# Solution

Amend #2743 by importing the missing `l` sub from `Translation` package.

# Checklist for author

* [x] Tested locally by adding such link to a random artist and visiting its page


reosarevok (approved):
Sigh. Thanks.

PR #2761, opened 2022-12-05
MBS-12738 / MBS-12739: Don't assume "single" and "studio" are ETI unless inside parens 

reosarevok:
### Implement MBS-12738 / MBS-12739

These are common enough words in titles that we should not assume they're always ETI if at the end. Of course, it's mostly fine to assume that if inside parens, so this doesn't change that.

Same idea as https://github.com/metabrainz/musicbrainz-server/commit/901b8f278d7e8ff50c76721ca6aba245c474c650


mwiencek:
We have some existing guess case tests -- can you add a couple test cases for these?

reosarevok:
Done :)

mwiencek (approved):
Thanks!

PR #2764, opened 2022-12-05
MBS-12756: Convert Delete instrument page to React

reosarevok:
### Implement MBS-12756

On top of https://github.com/metabrainz/musicbrainz-server/pull/2763


yvanzo (approved):
Tested removing an instrument with/without relationship.
Pushed a small refactoring commit too.
:ship:

PR #2767, opened 2022-12-05
MBS-12768: Convert Delete relationship attribute type page to React

reosarevok:
### Implement MBS-12768

See sister PRs https://github.com/metabrainz/musicbrainz-server/pull/2767 and https://github.com/metabrainz/musicbrainz-server/pull/2771


yvanzo (approved):
The code looks good to me. The cancel addition is fine.

Tested locally by adding a relationship attribute “centenary” and removing it successfully.

:ship:

### Aside

I initially tested removing the existing attribute “anniversary” and it logically threw an ISE as it was still referenced by some relationship type. This is not an issue for now as the current relationship editor is also a developer (@reosarevok).

PR #2768, opened 2022-12-05
MBS-12769: Convert Delete relationship type page to React

reosarevok:
### Implement MBS-12769

See sister PRs https://github.com/metabrainz/musicbrainz-server/pull/2767 and https://github.com/metabrainz/musicbrainz-server/pull/2771


reosarevok:

yvanzo (approved):
LGTMBDNT. (Did test the sister pull request #2767 though.) :ship:

PR #2769, opened 2022-12-06
MBS-12630: Make the relationship type autocompletes more compact

mwiencek:
This fixes MBS-12630 by adding a "Show/hide descriptions" action at the end of link type and link attribute type (including instrument) autocomplete menus.  This preference is stored as a cookie so that it doesn't need to be constantly toggled off (as it defaults to true).

The above change by itself makes the menus a lot more compact, but in addition, I've also changed how link type results are displayed based on feedback from Aerozol by just showing "Forward link phrase / reverse link phrase" (so, no longer showing the actual type name, and no longer showing the "Forward link phrase:" and "Backward link phrase:" labels). Additionally, if the reverse link phrase is identical to the forward one once translated and stripped of attributes, only the forward one is shown.

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -18,6 +18,7 @@ import {
 import useOutsideClickEffect from '../hooks/useOutsideClickEffect.js';
 import {unwrapNl} from '../i18n.js';
 import clean from '../utility/clean.js';
+import getCookie from '../utility/getCookie.js';
 
 import {
   CLOSE_ADD_ENTITY_DIALOG,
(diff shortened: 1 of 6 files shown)


mwiencek:
image

image

reosarevok (approved):
Works fine, thanks!

Once the descriptions are not always there all the time, I don't even know if it's necessary to remove the `(help)` toggle on the side. But I don't mind either way, so up to you :)

mwiencek:
> Once the descriptions are not always there all the time, I don't even know if it's necessary to remove the (help) toggle on the side. But I don't mind either way, so up to you :)

It doesn't seem very useful to me anymore, since the descriptions are still shown in the list by default.  The "(help)" toggle only showed the description for the currently selected type, which you can see by just focusing the autocomplete.

If you mean keeping the "(help)" link instead of adding "Show/hide descriptions" menu items, that seemed confusing to me because clicking the link would have no visual effect unless the menu was open (and clicking it with the menu open would likely close the menu).

PR #2770, opened 2022-12-06
MBS-12780: Fix allowNew check for AddRelationship

reosarevok:
### Fix MBS-12780

This was checking for the `entity0` and `entity1` passed outside the `relationship` itself - which we were also adding to the relationship data for display. But we were only passing this if the ID was not defined in the first place, so this check was nonsensical AFAICT. In fact, I cannot find any reason to pass `entity0` and `entity1` outside the `relationship`, since we're already passing them inside it.

I'm also removing the `source_type` and `target_type` that I had added in 9b0eae3e837e848c5eecb71b574185a37a2e9387 since AFAICT we're again not using those, but the ones inside the `relationship` object.

I tested this with `/edit/92070356` (the one where both sides were removed that triggered the ticket), with a new URL being added to a new release (to make sure release preview *does* still say "will be created"), and with `/edit/15273082` (the edit that triggered https://github.com/metabrainz/musicbrainz-server/pull/1589) to make sure it still works for deleted reltypes. Please do have a quick think on whether there might be some other case I didn't think of that could break with this :)

### DIFF
--- a/lib/MusicBrainz/Server/Edit/Relationship/Create.pm
+++ b/lib/MusicBrainz/Server/Edit/Relationship/Create.pm
@@ -181,14 +181,14 @@ sub build_display_data
     my $loaded_entity_0 = $loaded->{$model0}{$entity0_gid_or_id} if $entity0_gid_or_id;
     my $entity0 = $loaded_entity_0 ||
         $self->c->model($model0)->_entity_class->new(
-            id => 0,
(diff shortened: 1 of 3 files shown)


mwiencek:
I tested it against older edits with removed entities and edit previews in the release editor, and it seems to work as expected.

mwiencek (approved):

PR #2771, opened 2022-12-06
MBS-12740: Convert Delete attribute page to React

reosarevok:
### Implement MBS-12740

Wasn't sure what's a good place for `AttributeT` so I dumped it into misc for now.

See sister PRs https://github.com/metabrainz/musicbrainz-server/pull/2767 and https://github.com/metabrainz/musicbrainz-server/pull/2768


yvanzo (approved):
LGTMBDNT. (Did test the sister pull request #2767 though.)

For `AttributeT` location, I would probably put it under `root/types/entity_type_attribute.js` but @mwiencek might have a better suggestion.

mwiencek:
If the type is only relevant to files under root/admin/attributes/, I'd put it in root/admin/attributes/types.js and import it where needed. root/types/ is more for global types that need to be used across many unrelated files

reosarevok:

PR #2772, opened 2022-12-06
MBS-12775: Convert the release Disc Ids page to React 

reosarevok:
### Implement MBS-12775

# Conversion changes
Since the release editor expects `medium.cdtocs` to be just an array of TOC strings, and this is the only place we seem to use
the full `MediumCDTOC`, I am passing the full `MediumCDTOC`s as an array via props.
Since we already load all the needed data for `isPerfectMatch` as part of that array, I am using the medium data in there rather than on the release mediums list to build the rows.

# Testing
Tested manually.
`/release/bb36d6af-8e22-4bb1-a789-291e9d0a19be/discids` seems like a good release to test - it has many mediums, some with and some without discids.
`/release/15f5c2cb-41ea-3630-a0f6-2f86785cf608/discids` has two discids for the same medium, one of which should be a perfect match.


mwiencek (approved):
Tested locally, seems to work well.

PR #2773, opened 2022-12-07
MBS-12779: Treat "n-bit" and "loop" as ETI if in parens

reosarevok:
### Implement MBS-12779

"Loop" seems to mostly be used in an ETI way when in parens (but is otherwise a very legitimate title word).
"8-bit" and the like are generally ETI in parens but it's unclear whether they should always be lowercase or not on titles; there's plenty of both usages online, so this only touches them if inside parens as well for now.


reosarevok:

mwiencek (approved):

PR #2774, opened 2022-12-07
MBS-12771: Convert Change release quality page to React

reosarevok:
### Implement MBS-12771


yvanzo (approved):
The code looks good to me.

Yes, the lack of localization for the data quality options in the editing interface was a bug. You can either mention it in the linked ticket or make a ticket about it.

Tested locally by changing data quality multiple times for the same release.

:ship:

reosarevok:
Left a comment in the ticket about untranslatable stuff.

PR #2775, opened 2022-12-08
MBS-12767: Convert relationship attribute type edit form to React

reosarevok:
### Implement MBS-12767


lib/MusicBrainz/Server/Controller/Relationship/LinkAttributeType.pm
```diff
@@ -153,6 +163,17 @@ sub edit : Chained('load') RequireAuth(relationship_editor)
             $c->response->redirect($c->uri_for_action('relationship/linkattributetype/show', [ $link_attr_type->gid ]));
         }
     }
+
+    $c->stash(
+        current_view => 'Node',
+        component_path => 'relationship/linkattributetype/EditRelationshipAttributeType',
+        component_props => {
+            disableCreditable => boolean_to_json($attribute_in_use && $link_attr_type->creditable),
+            disableFreeText => boolean_to_json($attribute_in_use),
```
mwiencek, line 172: We already disable the fields above:

```Perl
    if ($attribute_in_use) {
        $form->field('free_text')->disabled(1);
        if ($link_attr_type->creditable) {
            $form->field('creditable')->disabled(1);
        }
    }
```

Though it appears we don't serialize or use any kind of `disabled` attribute on the `FieldT` -- that seems like a good idea here.

reosarevok: I tried to do that with an extra commit here, see if that seems about right :)

PR #2776, opened 2022-12-09
MBS-12781: Avoid admin accidentally breaking options trees 

reosarevok:
### Fix MBS-12781

When editing attributes, relationship types and relationship attributes, you can change their parent type.
There's currently nothing stopping an admin from accidentally selecting the entity itself from the dropdown, which makes MB so confused it doesn't know where to show the entity - so it doesn't. Fun times.

I also tried blocking linking a type as a child of its own child, which also causes similar breaking loop messes. I don't think it's possible to do that completely with the tools we have now, though.
This code blocks a root from being set under any of its children and a direct parent from being set under its direct children, which should cover most possible cases; in a very deep tree `a->b->c->d` it's probably still possible to set `b` as a child of `d` but I don't think it's worth adding a new recursive query to get all child IDs just to avoid that scenario.


mwiencek:
A recursive query isn't much harder, so might be worth doing to prevent future mistakes?  (I didn't even know you could set a type as its own parent at present, but here we are. :))

```SQL
WITH RECURSIVE hierarchy(parent, children) AS (
  SELECT parent, ARRAY[id] AS children
    FROM link_type
  UNION ALL
  SELECT link_type.parent, hierarchy.children || link_type.id
    FROM link_type
    JOIN hierarchy ON link_type.id = hierarchy.parent
   WHERE link_type.parent != any(hierarchy.children)
)
SELECT 1
  FROM hierarchy
 WHERE parent = 123
   AND 456 = any(children)
 LIMIT 1;
```

I guess `Data::LinkType` doesn't consume `Data::Role::OptionsTree` for some reason, but we could just duplicate the query in `Data::LinkType` for now.

reosarevok:
Ok then! I updated the PR to use the recursive query - it was breaking for LinkAttributeType because that has a weird `_table`, but I just hardcoded that one.

lib/MusicBrainz/Server/Data/Role/OptionsTree.pm
```diff
@@ -32,6 +32,38 @@ sub get_tree {
...
+
+=cut
+
+sub is_child {
+    my ($self, $own_id, $possible_child_id) = @_;
+
+    # _table for link attribute types includes a bunch of joins we don't want
+    my $table = $self->_type eq 'link_attribute_type'
+        ? 'link_attribute_type'
+        : $self->_table;
```
mwiencek, line 47: Getting the table from `%ENTITIES` would be a bit more future-proof in case we add other joins in the future.

mwiencek (approved):
It would be nice to remove the link type duplication in the future (the form validation can probably already be a role), but this works for now. :)

lib/MusicBrainz/Server/Data/Role/OptionsTree.pm
reosarevok: That'd require the table to be on `%ENTITIES` in the first place, which seems fairly rare - but I guess I will just fix that.

reosarevok:
Updated this to use `%ENTITIES`. We could move the validation to a role indeed - the only reason not to might be that `LinkAttributeType` makes extra checks for instruments, but I guess that could be either a second `after validate` or a check for model inside the role. Which of the two sounds better?

reosarevok:

yvanzo:
Didn’t  review yet but this can be rebased after the pull request #2777.

mwiencek (approved):
Let me know if the changes I pushed look okay. I tested the new validation again for both link types & link attribute types, + the separate instrument validation.

PR #2777, opened 2022-12-12
MBS-12741: Convert Create/Edit attribute admin pages to React

reosarevok:
### Implement MBS-12741

# Changes during conversion
TT was happy to have a scrap heap of forms held together with wire and glue here, but I thought it would make a lot more sense to split the different forms into their own components and pick the right one according to the model.

I made it so that the entity type dropdown actually shows translatable entity names, rather than the type string.

# Testing
Manually, testing each of the different kinds of forms (Language, Script, SeriesType/CollectionType, MediumFormat, WorkAttributeType, AreaType for the default).


root/admin/attributes/types.js
```diff
@@ -9,54 +9,62 @@
...
-    +form: LanguageEditFormT,
-    +type: 'Language',
-  }>
+      +action: 'add' | 'edit',
+      +entityTypeSelectOptions:
+        {[entityType: CoreEntityTypeT]: CoreEntityTypeT},
+      +form: AttributeEditFormWithEntityTypeT,
+      +parentSelectOptions: SelectOptionsT,
+      +type: 'SeriesType',
+    }>
```
reosarevok, line 45: Huh, so even if they are the same, Flow doesn't allow using a union type?

root/admin/attributes/AttributeEditForm.js
```diff
@@ -18,40 +18,46 @@ import ScriptEditForm from './ScriptEditForm.js';
...
-    />
-  ) : (
-    <AttributeEditGenericForm
-      form={form}
-      parentSelectOptions={parentSelectOptions}
-    />
-  )
-);
+const AttributeEditForm = (
+  props: CreateOrEditAttributePropsT,
```
reosarevok, line 22: Does this work just because Flow doesn't check these as carefully as the previous way? Or is this a documented way to deal with this issue? :/

root/admin/attributes/types.js
mwiencek: Yes, not sure why, but it broke Flow's inference for some reason.

root/admin/attributes/AttributeEditForm.js
mwiencek: It's documented at https://flow.org/en/docs/types/unions/#toc-disjoint-object-unions

In order to refine an object type in this way, you have to check the property on the object (which only refines the variable holding the object).  If you destructure the object into separate variables first, Flow doesn't track that those variables have any logical relation to each other.

yvanzo (approved):
Code looks good. I tested all attributes manually too, even though it is impracticable to cover all the possible interactions. That reminded me of MBS-12781 too. Thanks! :ship:

PR #2778, opened 2022-12-13
MBS-12782: Block adding CritiqueBrainz links

reosarevok:
### Implement MBS-12782

Since we link to CB automatically from the Reviews tab, there's no point on letting users manually link to the reviews.

I expected that doing the string like that will ensure that the Reviews tab name uses whatever translation we already have, but I tried with Italian and it still showed "Reviews" despite us having a translation - is my understanding of how it should work wrong, or does it only work if the main string is translated? :)

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -708,6 +708,15 @@ export class _ExternalLinksEditor
                     Did you mean to paste something else?`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
+    } else if (isNewOrChangedLink && isCritiqueBrainz(link.url)) {
+      error = {
+        message: texp.l(
+          `Please don’t enter CritiqueBrainz links — reviews
+           are automatically linked from the “{reviews_tab_name}” tab.`,
+          {reviews_tab_name: l('Reviews')},
+        ),
+        target: URLCleanup.ERROR_TARGETS.URL,
+      };
     } else if (isNewOrChangedLink && isMalware(link.url)) {
       error = {
         message: l(`Links to this website are not allowed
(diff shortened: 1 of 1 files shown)


root/static/scripts/edit/externalLinks.js
```diff
@@ -705,6 +705,15 @@ class _ExternalLinksEditor
                     Did you mean to paste something else?`),
         target: URLCleanup.ERROR_TARGETS.URL,
       };
+    } else if (isNewOrChangedLink && isCritiqueBrainz(link.url)) {
+      error = {
+        message: texp.l(
+          `Please don’t enter CritiqueBrainz links — reviews
+           are automatically linked from the “{reviews_tab_name}” tab.`,
+          {reviews_tab_name: l('Reviews')},
+        ),
```
yvanzo, line 714: > I expected that doing the string like that will ensure that the Reviews tab name uses whatever translation we already have, but I tried with Italian and it still showed "Reviews" despite us having a translation - is my understanding of how it should work wrong, or does it only work if the main string is translated? :)

Your understanding is correct. I tested translating the new string and updating POT/PO files locally and “Reviews” remains untranslated here.

My guess is that it has something to do with component hydration. I didn’t find other cases of `l` being used in arguments of `texp.l` under `root/static/scripts`. It might be an undocumented use case if not a bug.

yvanzo (commented):
I confirm the localization issue.

The rest looks good to me, I even tested editing locally a release-group.

Not urgent as entering CritiqueBrainz URLs is hopefully quite rare.

yvanzo:
The updated commits are rebased on the latest `HEAD` of the branch `master` but don’t address the localization issue.

reosarevok:
Yeah. Let's see if @mwiencek has any ideas.

mwiencek (approved):

PR #2779, opened 2022-12-13
MBS-12746: Convert cdstub/browse to React

reosarevok:
### Implement MBS-12746

I added `getCDStubAge` to avoid doing the same thing in `CDStubSidebar` and `BrowseCDStubs`.

This seems to be the last use of `display_age_ago` so removing that macro.

The "today" hack used in the test was never going to work now that this is using JS. Hopefully calculating the expected year here in the same way `Plugin::Age` does will be consistently correct - I'm not bothering to use the plugin itself because I expect we'll remove it after we're done with the React conversion.

Tested by going to `/cdstub/browse` with pink data.


yvanzo (approved):
LGTMBDNT :ship:

PR #2781, opened 2022-12-14
MBS-12762: Convert Set RG cover art page to React

reosarevok:
### Implement MBS-12762


mwiencek:
I'm getting the following error locally when I navigate to a `/set-cover-art` page. Just me?

```
Couldn't render template "file error - release_group/set_cover_art.tt: not found"
```

root/static/scripts/release-group/components/SetCoverArtForm.js
```diff
@@ -0,0 +1,209 @@
...
+                    <p>
+                      {releaseDates ? (
+                        <>
+                          {addColonText(l('Date'))}
+                          {' '}
+                          {releaseDates}
+                          <br />
+                        </>
+                      ) : null}
+                      {releaseCountries ? (
```
mwiencek, line 123: This is testing the React element you created, which will always be truthy. You'll need to check `release.events` instead (same for other places here where you check a React element).

yvanzo:

yvanzo:

lib/MusicBrainz/Server/Controller/ReleaseGroup.pm
```diff
@@ -163,13 +167,26 @@ sub set_cover_art : Chained('load') PathPart('set-cover-art') Args(0) Edit
     my ($self, $c, $id) = @_;
 
     my $entity = $c->stash->{entity};
-    return unless $entity->can_set_cover_art;
+    unless ($entity->can_set_cover_art) {
+        # No point running anything else, just return and let it say nope
+        my %props = (
+            entity => $entity->TO_JSON,
```
mwiencek, line 173: This is missing several required props which causes the component to throw an error.

root/static/scripts/release-group/components/SetCoverArtForm.js
```diff
@@ -0,0 +1,231 @@
...
+import FieldErrors from '../../edit/components/FieldErrors.js';
+import HiddenField from '../../edit/components/HiddenField.js';
+
+component SetCoverArtForm(
+  allReleases: $ReadOnlyArray<ReleaseT>,
+  artwork: {[releaseId: number]: ArtworkT},
+  form: SetCoverArtFormT,
+) {
+  const originallySelected = form.field.release.value;
+  const [selected, setSelected] = React.useState(originallySelected);
```
mwiencek: `selectedRelease` is clearer to me (the TT file also used `selected_release`).
```suggestion
  const [selectedRelease, setSelectedRelease] = React.useState(originallySelected);
```

root/release_group/SetCoverArt.js
```diff
@@ -0,0 +1,54 @@
...
+          artwork={artwork}
+          form={form}
+        />
+      ) : (
+        <p>
+          {l(`No releases have cover art marked as "Front",
+              cannot set cover art.`)}
+        </p>
+      )}
+      {manifest(
```
mwiencek, line 45: This is missing some manifests that were present in the TT form, which means images currently don't load at all.
```
[%- script_manifest('common/loadArtwork.js', {async => 'async'}) -%]
[%- script_manifest('common/artworkViewer.js', {async => 'async'}) -%]
```

root/static/scripts/release-group/components/SetCoverArtForm.js
```diff
@@ -0,0 +1,231 @@
...
+        {exp.l(
+          `Only releases with a front cover on the {caa|Cover Art Archive}
+            can be selected.`,
+          {caa: '//coverartarchive.org'},
+        )}
+      </p>
+      <div className="row" id="set-cover-art-position-row">
+        <FieldErrors field={form.field.release} />
+        <div id="set-cover-art-images">
+          {allReleases.map((release, index) => {
```
mwiencek, line 228: While it only really affects RGs that have a lot of cover art, everything inside this `map` would make sense as a separate memoized component, since individual items whose selection doesn't change don't need to re-render.

root/static/scripts/release-group/components/SetCoverArtForm.js
```diff
@@ -0,0 +1,231 @@
...
+
+component SetCoverArtForm(
+  allReleases: $ReadOnlyArray<ReleaseT>,
+  artwork: {[releaseId: number]: ArtworkT},
+  form: SetCoverArtFormT,
+) {
+  const originallySelected = form.field.release.value;
+  const [selected, setSelected] = React.useState(originallySelected);
+
+  const selectCover = (event: SyntheticMouseEvent<HTMLDivElement>) => {
```
mwiencek, line 61: (If a separate component is added for the images, this should use `React.useCallback`.)

mwiencek (changes requested):
Submitting the actual form works fine, though there are some issues which are easier to spot with a local CAA setup. :)

> did not test submission since I don't have cover art set locally

See HACKING-IMG.md, though it's slightly outdated (running the indexer uses `poetry` now, for one, but that's in the artwork-indexer README).

reosarevok:
Made the required changes, plus set the CAA locally to test it a bit, which seemed to work.

root/static/scripts/release-group/components/SetCoverArtForm.js
reosarevok: Moved this into the separate component instead, which avoids the need for a callback AFAICT.

root/static/scripts/release-group/components/SetCoverArtForm.js
reosarevok: Moved to its own memoized component on a separate commit.

lib/MusicBrainz/Server/Controller/ReleaseGroup.pm
reosarevok: Made the props optional instead in Flow, and made the necessary changes to avoid errors and make Flow happy.

PR #2784, opened 2022-12-15
MBS-12753: Convert edit type documentation page to React

reosarevok:
### Implement MBS-12753

Much of the info in the table for this page is no longer relevant: voting period is now always 7, needed votes always 3, and every edit is accepted on expiration.
As such, I dropped those rows, and I'm adding a new row for whether an edit type is always an autoedit unless explicitly made votable (such as Add artist). I added a new condition for this, which is false unless set to true by `Role::AlwaysAutoEdit`.

By the way, did you know these pages actually try to load a wiki page with documentation for the edit type? I did not, and had never seen this done. It honestly might be better / easier to just hardcode this in the `Edit::` package or have a documentation table, but I guess it's nice we have something!

Suggested testing is `/doc/Edit_Types/314`, because I actually added a bit of documentation for that one to the wiki as a quick test, so you can actually check the wiki docs work. It also is a type which can be approved, but is not an autoedit by default, so it shows the table can be useful.

If you can think of more stuff that could be useful in this table, we could add it (either in this PR or later on, in which case please add a ticket).


lib/MusicBrainz/Server/Edit/Role/AlwaysAutoEdit.pm
```diff
@@ -7,6 +7,7 @@ around 'edit_conditions' => sub {
     my ($orig, $self, @args) = @_;
 
     my $conditions = $self->$orig(@args);
+    $conditions->{always_auto_edit} = 1;
```
mwiencek, line 10: I never liked the name of this package, since it's not always an auto-edit if you can still choose to make it votable. :)

root/doc/EditTypeIndex.js
```diff
@@ -0,0 +1,83 @@
...
+      <tbody>
+        <EditTypeProperty
+          label={l('Applies automatically')}
+          property={checkboxOrCross(editType.appliesAutomatically)}
+        />
+        <EditTypeProperty
+          className="even"
+          label={l('Can be approved')}
+          property={checkboxOrCross(editType.canBeApproved)}
+        />
```
mwiencek, line 77: "Applies automatically" is a useful addition. 👍

I feel like an "x" there can be misleading, since it may possibly still be applied automatically by auto-editors. Which is what "Can be approved" indicates, but in a weird, indirect way? (And a checkmark can be similarly misleading, since it may still be made votable by choice or by editor restrictions.) I guess it would be clearer to me if this were a single property...

"Can be applied automatically: Never/By normal editors/By auto-editors only".

("By auto-editors only" would imply it can be approved by them.)

Am I missing some case where these make sense separated?

lib/MusicBrainz/Server/Edit.pm
```diff
@@ -170,6 +170,7 @@ sub editor_may_edit {
 #  * votes - number of votes to consider the edit as unanimously accepted/rejected
 #  * expire_action - what do do with expired edits without votes
 #  * auto_edit - whether the edit can be automatically accepted for an autoeditor
+#  * always_auto_edit - whether the edit always applies automatically
```
mwiencek, line 173: For the same reason as above, I don't really like "always" in the comment here. Shouldn't it be "whether the edit applies automatically by default for normal editors"?

mwiencek (commented):
It'd be nice if we could at least clarify the comment below, if not rename "always_auto_edit." You can ignore the suggestion for combining the properties, though -- if you'd like to merge this as-is and agree they can be combined, I'll open a ticket instead.

> By the way, did you know these pages actually try to load a wiki page with documentation for the edit type? I did not, and had never seen this done.

I had no idea, either!

root/doc/EditTypeIndex.js
reosarevok: Probably not, so I made the change. This made me feel we need a second row "Required privileges" because "Can be applied automatically: By any editor" for `Add area` is kind of funny, but AFAICT we do not store the required privileges on the edit class itself... maybe we should?

lib/MusicBrainz/Server/Edit/Role/AlwaysAutoEdit.pm
reosarevok: I didn't rename the package for now, but I did rename the condition at least.

PR #2785, opened 2022-12-16
MBS-12754: Convert edit types documentation page to React

reosarevok:
### Implement MBS-12754

# Changes during conversion
Sorting the keys seems more sensible than looping, skipping if Historic, then requesting Historic.
This sorting also seems better to me than the original in that the previous DB sort the French Événement and Œuvre sorted last, while now they sort as E and O respectively; my understanding is that this is more correct.

# Testing

Tested the list manually to ensure the sort seems correct.


root/doc/EditTypeList.js
```diff
@@ -0,0 +1,65 @@
...
+
+const EditTypeList = ({
+  editTypesByCategory,
+}: EditTypeListProps): React$Element<typeof Layout> => {
+  const sortedCategories =
+    Object.keys(editTypesByCategory).sort(cmpCategories);
+  return (
+    <Layout fullWidth noIcons title={l('Edit Types')}>
+      <div className="wikicontent" id="content">
+        <h1>{l('Edit Types')}</h1>
```
mwiencek, line 41: sync strings with master
```suggestion
    <Layout fullWidth noIcons title={lp('Edit types', 'noun')}>
      <div className="wikicontent" id="content">
        <h1>{lp('Edit types', 'noun')}</h1>
```

mwiencek (approved):

PR #2786, opened 2022-12-19
MBS-12792: Hide ratings by spammer users

reosarevok:
### Implement MBS-12792

# Problem
While we block seeing the profiles of spammer users, their ratings are still visible. For cases where the user was specifically spamming ratings, this makes marking them as spammers kinda useless.

We also use the spammer ratings in the calculation of the average community rating, which is not ideal.

# Solution
Sadly, it turns out we will need a schema change (MBS-12794) to stop using spammer ratings (and tags) when calculating community results.

For now, this at least hides links to spammer profiles from the ratings page and makes it clear to users viewing from that page that the ratings might be dodgy. There's no indication of that on the sidebar, sadly, but it's something, and it probably doesn't make sense to write code to check for spammer ratings on all pages with sidebars if we plan to stop using them in calculations later on anyway.

# Testing
Tested manually. To test, you can use `/artist/9a694506-85b4-4610-ae05-f20934b9aabc/ratings` - to test with "legit" ratings too, when using pink, just temporarily add your own rating from mb.org.

I didn't add a more formal test since this is meant as a temporary hack.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Rating.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Rating.pm
@@ -28,8 +28,17 @@ sub ratings : Chained('load') PathPart('ratings')
 
     my @public_ratings;
     my $private_rating_count = 0;
+    my $spammer_rating_count = 0;
 
     for my $rating (@ratings) {
+        # We don't want spammer ratings, but they're still
+        # part of the calculated rating.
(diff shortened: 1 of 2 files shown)


reosarevok:

reosarevok:

mwiencek (approved):

PR #2787, opened 2022-12-23
MBS-12798: Cancel open edits of user when it's marked as spammer

reosarevok:
### Implement MBS-12798

We used to delete spammers, but now we leave them behind (marked as spammers) so that SpamBrainz can eventually learn from them.

When we delete an editor, their open edits get cancelled and their votes on open edits changed to Abstain. It seems sensible to do the same for spammers, since:
a) we used to remove them
b) their edits and votes are, by definition, dodgy.

Added a test for this.


reosarevok:

reosarevok:

yvanzo (commented):
Is it mailing spammers about cancelled edits? We don’t want to give hints to spammers or to spam a stolen email address in return.

I didn’t review it otherwise.

reosarevok:
No, we never send email about this AFAIK, it just automatically cancels and that's it.

yvanzo (changes requested):
This is a sensitive piece of code, please split the refactoring as a separate commit off from it.

reosarevok:
Ok, separated the two bits, assuming that's what you meant :)

mwiencek (approved):
conflicts have arised but the code lgtm.

lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -392,6 +392,8 @@ sub update_profile
 sub update_privileges {
     my ($self, $editor, $values) = @_;
 
+    my $should_cancel_edits = $values->{spammer};
```
yvanzo, line 395: That could be named `$should_cancel_edits_and_votes` for clarity.

yvanzo (approved):
Thanks for the neat commits.

Just one minor omission in variable name, see comment.

Otherwise LGTMBDNT thanks to the added test! :ship:

mwiencek (approved):

PR #2788, opened 2022-12-28
MBS-12801: Warn users about "Disc n" titles on release merge

reosarevok:
### Implement MBS-12801

# Problem
MBS-12034 made it mandatory to provide a confirmation when adding a "Disc 1" style medium name when editing a release. I did not remember, when implementing it, that there is a second place where you can set medium names: the release merge page. Right now, nothing blocks adding those from that page nor warns the user it might be a bad idea.

# Solution
I just implemented a basic warning, which is probably enough. We could also disable submission unless they confirm, by erroring at the Perl level, but this is less intrusive and hopefully does the trick still, since it appears just below the name row as soon as the user enters the triggering name.

image_2022-12-28_100817159

# Testing
Tested manually when merging two random releases, since the testing is only dependent on the entered medium name.

### DIFF
--- a/root/static/scripts/edit/components/ReleaseMergeStrategy.js
+++ b/root/static/scripts/edit/components/ReleaseMergeStrategy.js
@@ -21,6 +21,7 @@ import {ExpandedArtistCreditList}
 import Warning from '../../common/components/Warning.js';
 import formatTrackLength
   from '../../common/utility/formatTrackLength.js';
+import isUselessMediumTitle from '../utility/isUselessMediumTitle.js';
 
 import FormRowSelect from './FormRowSelect.js';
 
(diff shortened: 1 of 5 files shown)


root/static/scripts/edit/components/ReleaseMergeStrategy.js
```diff
@@ -40,6 +40,33 @@ const mergeStrategyOptions = {
   ],
 };
 
+// Please keep this in sync with root/static/scripts/release-editor/fields.js
```
mwiencek, line 43: Seems like like we should just make it a utility function and call `hasUselessMediumTitle(self.name())` in the computed observable. :)

mwiencek (approved):

PR #2789, opened 2022-12-29
MBS-12802: Use localize_error for artist credit errors

reosarevok:
### Fix MBS-12802

# Problem
Trying to submit (or seed, reportedly, haven't tested that directly) an artist credit field where there's a string with brackets as either the artist credit (with the artist field empty) or the artist (unlinked) raises an error telling the user that's wrong, as expected (e.g. `Artist "[whatever]" is unlinked, please select an existing artist.`) . FormHandler passes that to maketext, which chokes on the brackets (a restricted notation for it, see https://perldoc.perl.org/Locale::Maketext#BRACKET-NOTATION) and ISEs.

# Solution
I generalized the override we used elsewhere to skip `maketext` as `localize_error`, and then used it for artist credits as well.

While I was here, I simplified the check that raises the errors, since `$name` always exists if `$artist_name` does.

# Testing
Manually. You can try to add a standalone recording with a `[whatever]` unlinked artist, and/or with a `[whatever]` artist credit and an empty artist field.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
+++ b/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
@@ -39,7 +39,7 @@ around 'validate_field' => sub {
 
         my $artist_id = Text::Trim::trim $_->{'artist'}->{'id'};
         my $artist_name = Text::Trim::trim $_->{'artist'}->{'name'};
-        my $name = Text::Trim::trim $_->{'name'} || Text::Trim::trim $_->{'artist'}->{'name'};
+        my $name = Text::Trim::trim $_->{'name'} || $artist_name;
 
         if ($artist_id && $name)
         {
@@ -51,13 +51,13 @@ around 'validate_field' => sub {
                 l('Please add an artist name for {credit}',
                   { credit => $name }));
         }
-        elsif (! $artist_id && ( $name || $artist_name ))
+        elsif (! $artist_id && $name )
         {
             # FIXME: better error message.
             $self->add_error(
                 l('Artist "{artist}" is unlinked, please select an existing artist. ' .
                   'You may need to add a new artist to MusicBrainz first.',
-                  { artist => ($name || $artist_name) }));
+                  { artist => $name }));
         }
         elsif (!$artist_id)
         {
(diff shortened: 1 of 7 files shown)


mwiencek:
FormHandler runs the messages through maketext by default, but you can override it - https://metacpan.org/dist/HTML-FormHandler/view/lib/HTML/FormHandler/Manual/Intro.pod#Localization (grep for `localize_meth` in our codebase).  Adding `localize_meth` to the appropriate fields here is probably the "correct" solution (though I wish there was a way to change it globally, since I don't think we want maketext anywhere, really).

reosarevok:
I tried to add `localize_meth => \&localize_error` as generalized in the second commit now, but even adding it to all three fields `names`, `names.name` and `names.artist` still threw the `maketext` error. Any idea what am I missing?

mwiencek:
This works for me:

```diff
diff --git a/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm b/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
index 2a11063a1c..7376510c1f 100644
--- a/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
+++ b/lib/MusicBrainz/Server/Form/Field/ArtistCredit.pm
@@ -11,6 +11,7 @@ extends 'HTML::FormHandler::Field::Compound';
 use MusicBrainz::Server::Edit::Utils qw( clean_submitted_artist_credits );
 use MusicBrainz::Server::Entity::ArtistCredit;
 use MusicBrainz::Server::Entity::ArtistCreditName;
+use MusicBrainz::Server::Form::Utils qw( localize_error );
 use MusicBrainz::Server::Translation qw( l );

 has_field 'names'             => ( type => 'Repeatable', num_when_empty => 1 );
@@ -39,8 +40,6 @@ around 'validate_field' => sub {
         my $artist_id = Text::Trim::trim $_->{'artist'}->{'id'};
         my $artist_name = Text::Trim::trim $_->{'artist'}->{'name'};
         my $name = Text::Trim::trim $_->{'name'} || Text::Trim::trim $_->{'artist'}->{'name'};
-        # Need to escape any square brackets for maketext
-        $name =~ s/([\[\]])/~$1/g;

         if ($artist_id && $name)
         {
@@ -136,6 +135,10 @@ sub json {
     return to_json({names => $names});
 }

+sub build_localize_meth {
+    return \&localize_error;
+}
+
 =head1 COPYRIGHT AND LICENSE

 Copyright (C) 2011 MetaBrainz Foundation
```

I guess `localize_meth` has to be defined on the field that sets the errors, not the field that contains the errors.

mwiencek (approved):

PR #2790, opened 2023-01-03
MBS-12799 / MBS-7028: Extra options for filters

reosarevok:
### Implement MBS-12799 / MBS-7028

# Improvements
This adds two new filtering options: video/not video for recordings and disambiguation for all entities.

It also makes the flow typing more robust and organizes the code a bit more (I don't dare say simplifies it, but it's less spaghettish at least).

# Testing
By hand (mostly on Nine Inch Nails recordings), plus added tests.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Filter/Recording.pm
+++ b/lib/MusicBrainz/Server/Form/Filter/Recording.pm
@@ -22,7 +22,7 @@ has_field 'video' => (
 );
 
 sub filter_field_names {
-    return qw/ name artist_credit_id video /;
+    return qw/ disambiguation name artist_credit_id video /;
 }
 
 sub options_artist_credit_id {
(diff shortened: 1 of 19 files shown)


lib/MusicBrainz/Server/Form/Filter/Recording.pm
```diff
@@ -27,11 +33,19 @@ sub options_artist_credit_id {
     ];
 }
 
+sub options_video {
+    return [
+        { value => 1, label => l('Video only') },
+        { value => 2, label => l('Non-video only') },
```
mwiencek, line 39: I feel like video should be plural, but I'm also not entirely sure.

mwiencek (approved):

lib/MusicBrainz/Server/Form/Filter/Recording.pm
mwiencek: I do think that "Videos only" and "Non-videos only" sound better than "Video only" and "Non-video only," since it's dealing with multiple results. Similar to a sign that says "Pedestrians only" vs. "Pedestrian only."

But your alternative suggestions sound fine too.

reosarevok: Oh, I see how you're looking at it - I was looking at it as in "video is a flag, so it is either video flag or no video flag". Ok, actually it makes sense to me, let's do that.

PR #2792, opened 2023-01-04
MBS-12809: Pass minimal existing data for Historic/RemoveDiscID

reosarevok:
### Fix MBS-12809

We were trying to build a new CDTOC from the discid alone, which actually goes against our definition of a `CDTOC` entity. This was now breaking in `TO_JSON` since it now expects a `track_offset`.

Rather than making that optional everywhere, I'm just passing a "minimal-cdtoc" JSON object with the data we have and allowing `CDTocLink` to use it - it does not need anything else, in any case.

### DIFF
--- a/root/static/scripts/common/components/CDTocLink.js
+++ b/root/static/scripts/common/components/CDTocLink.js
@@ -13,7 +13,11 @@ import entityHref from '../utility/entityHref.js';
 
 type Props = {
   +anchorPath?: string,
-  +cdToc: CDTocT,
+  +cdToc: {
+    +discid: string,
+    +entityType: 'cdtoc',
+    ...
+  },
   +content?: string,
   +subPath?: string,
 };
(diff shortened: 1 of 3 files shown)


reosarevok:

root/static/scripts/common/components/CDTocLink.js
```diff
@@ -13,7 +13,11 @@ import entityHref from '../utility/entityHref.js';
 
 type Props = {
   +anchorPath?: string,
-  +cdToc: CDTocT,
+  +cdToc: {
+    +discid: string,
+    +entityType: 'cdtoc',
+    ...
```
mwiencek, line 19: this probably doesn't need to be inexact?

mwiencek (approved):

root/static/scripts/common/components/CDTocLink.js
reosarevok: Isn't that needed if I also want to allow passing proper `CDTocT`s?

mwiencek: Yeah, sorry, I forgot we still need to support those.

PR #2793, opened 2023-01-04
MBS-12815: Support immutable URL edit search predicate

reosarevok:
### Fix MBS-12815

Relationships involving URLs have an edit search link when the relationship has pending edits, like for any other entities. Problem is, we do not intentionally support URL as a predicate at all.

In fact, this search was working, because Entity also created an (undocumented) predicate for URL, but this had no label and a search field.
URLs are not searchable, so the search being there was misleading, and having no label leads to it being listed as ID.

I decided to implement a basic, uneditable option to properly support this use case, which shows the URL as text.
I made it support "is not" as well because is trivial, although I'm not sure if it's useful tbh.

This is only shown if URL is being seeded as an option to begin with; it's not selectable for users.

This is all a big hack that should be replaced by MBS-12122 but it feels better than the current situation until then.

This can be tested with `/search/edits?auto_edit_filter=&order=desc&negation=0&combinator=and&conditions.0.field=release&conditions.0.operator=%3D&conditions.0.name=The%20Miracle%20(deluxe%20edition)&conditions.0.args.0=3712343&conditions.1.field=url&conditions.1.operator=%3D&conditions.1.name=https%3A%2F%2Fshop.emi.com%2Fartists%2F*%2F*%2F-The-Miracle-Deluxe-Collector-s-Edition%2F7I9H0000000&conditions.1.args.0=10357826&conditions.2.field=type&conditions.2.operator=%3D&conditions.2.args=90%2C233&conditions.2.args=91&conditions.2.args=92&conditions.3.field=status&conditions.3.operator=%3D&conditions.3.args=1&field=Please+choose+a+condition`

For what it's worth, I'm not sure what causes the link in prod to give the URL the name `function n(){return arguments.length>0?(n.isDifferent(t,arguments[0])&&(n.valueWillMutate(),t=arguments[0],n._latestValue=t,n.valueHasMutated()),this):(s.dependencyDetection.registerDependency(n),t)}` but that part of the issue seems solved in beta / master.


mwiencek (approved):
This appears to work well. I'm a bit confused what more is required for MBS-12122 beyond making the URL predicate a text input that you can edit, and having the predicate on the server use the URL text for the lookup instead of an ID. It seems like you implemented most of it already, but I'm probably misunderstanding or underestimating what needs to be done.

reosarevok:
Well, I thought we would want to have a URL search so you can only choose URLs that actually exist in the DB? But I didn't honestly consider the idea of making it take text instead of an ID... if that seems safe, it might work.

mwiencek:
> Well, I thought we would want to have a URL search so you can only choose URLs that actually exist in the DB? But I didn't honestly consider the idea of making it take text instead of an ID... if that seems safe, it might work.

If it's not in the DB, wouldn't the edit search just fail, which is fine? A URL search field might tell you that sooner, but.

reosarevok:
You know what, you're right, that's a lot better than nothing. I implemented that, let me know if it looks good :)

reosarevok:

PR #2796, opened 2023-01-05
MBS-12737: Properly check backwardness with isRelationshipBackward

reosarevok:
### Fix MBS-12737

The overconfident check we had was hitting issues for recordings creating during the NGS migration which had the same rowid as the associated works.

We have a perfectly working `isRelationshipBackward` check, so this just changes the code to use it.

Tested manually with the examples on the ticket.


mwiencek (approved):
thank you!

PR #2797, opened 2023-01-05
MBS-12787: Editing a relationship to add instruments causes duplication

mwiencek:
Fixes https://tickets.metabrainz.org/browse/MBS-12787

If a "members" relationship is added without any instruments, submitted, and then edited to add instruments, then the relationship will be duplicated for each instrument (or vocal) instead of adding at least one to the existing relationship.  ("members" is just an example, as this could apply to any relationship type supporting instruments or vocals.)

The solution is, if the existing relationship has no instruments or vocals, to take the first new instrument or vocal that would otherwise be split, and merge it into the existing relationship's attributes.

The "first" instrument or vocal chosen is the one with the lowest link attribute type ID, hence why the added test expects vocals to be chosen for merging over drums.

### DIFF
--- a/root/static/scripts/relationship-editor/utility/splitRelationshipByAttributes.js
+++ b/root/static/scripts/relationship-editor/utility/splitRelationshipByAttributes.js
@@ -78,6 +78,7 @@ export default function splitRelationshipByAttributes(
     tree.toArray(origRelationship ? origRelationship.attributes : null)
       .filter(isInstrumentOrVocal),
   );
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
FWIW, adding more than one instrument/vocal in prod does not split the rels until they are submitted (they just are shown all on the original existing rel) but it might actually be more clear this way and is probably intentional?

This doesn't modify the existing rel either if *adding* a new rel, but that's consistent with prod and seems sensible.

mwiencek:
Yeah, I was thinking it'd be clearer to show how they are actually submitted.

PR #2798, opened 2023-01-06
MBS-12709: Do not copy series part number with "Add another [entity]"

reosarevok:
### Implement MBS-12709

While it's generally fairly useful to preload a new relationship with the attributes of the existing one, this is never the case with series part numbers - you almost always want each relationship to have a different number. As such, this stops them from being preloaded.

Tested manually.


mwiencek (approved):
LGTM assuming tests pass

reosarevok:

PR #2801, opened 2023-01-09
MBS-12806: Do not show "Play on LB" button for empty releases

reosarevok:
### Implement MBS-12806

# Problem
The new "Play on ListenBrainz" button is appearing on every MusicBrainz release, even those which have no tracks at all. This just wastes the user's time if they click it (plus LB returns an ISE at the moment, but even if they returned a better "this is empty" page it would still be useless).

# Solution
I couldn't find any good, straightforward way to check for emptiness with the current code, so I added a new `has_no_tracks` method that just checks whether mediums exist at all, and if they do, whether all have 0 tracks.

This still shows the button if at least one medium on the release has tracks, since that should still be playable (it currently isn't because of a LB-side error which hopefully will be resolved soon).

# Testing
Tested with `/release/8a5285e9-4ce6-4aef-b77b-458d8f25f00f`, `/release/a1960f31-a4b6-4371-8275-18834115f82b` and `/release/0ee90972-72e8-4ef4-872e-5f31f0df6944`

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Release.pm
+++ b/lib/MusicBrainz/Server/Entity/Release.pm
@@ -198,6 +198,26 @@ has 'cover_art_presence' => (
     is => 'rw'
 );
 
+=head2 has_no_tracks
+
+Returns true if the release has no tracks at all, or false
+if at least one medium on the release has one track or more.
+
+=cut
+
+sub has_no_tracks
+{
+    my ($self) = @_;
+    my @mediums = $self->all_mediums;
+    return 1 if !@mediums;
+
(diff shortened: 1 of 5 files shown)


reosarevok:

reosarevok:

yvanzo (approved):
I agree that the new `has_no_tracks` is more efficient than checking whether `all_tracks` returns nothing. :100: 
Tested manually with the same releases (the first one has an empty medium and a medium with tracks, the second one has only one empty medium, and the last one seems to just be a big release with many mediums and tracks) using the production standby database (as not available in sample dump).
It is small enough to sneak it into beta today. :snake:

PR #2802, opened 2023-01-09
MBS-12827: Run our sanitize function on new usernames

mwiencek:
Fixes https://tickets.metabrainz.org/browse/MBS-12827

This shows an error on the registration form (and should also show one on the admin edit user form) if the name doesn't match the sanitized version of the name.  I imagine this will mostly be accidentally triggered by people adding extraneous whitespace characters. Suggestions for how to best word the error message are appreciated.

If an unsanitized username somehow makes it past our form validation, the editor data layer performs the same check and possibly dies.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -18,6 +18,7 @@ use MusicBrainz::Server::Data::Utils qw(
     hash_to_row
     load_subobjects
     placeholders
+    sanitize
     type_to_model
 );
 use MusicBrainz::Server::Constants qw( :edit_status :privileges );
@@ -246,11 +247,22 @@ sub find_subscribers
     $self->query_to_list_limited($query, [$editor_id], $limit, $offset);
 }
 
+sub _die_if_username_invalid {
+    my $name = shift;
+    my $sanitized_name = sanitize($name);
+
+    die 'Invalid user name' if (
+        $name ne $sanitized_name ||
+        $sanitized_name =~ qr{^deleted editor \#\d+$}i
+    );
+}
+
 sub insert
 {
     my ($self, $data) = @_;
 
-    die 'Invalid user name' if $data->{name} =~ qr{^deleted editor \#\d+$}i;
+    _die_if_username_invalid($data->{name});
+
     my $plaintext = $data->{password};
     $data->{password} = hash_password($plaintext);
     $data->{ha1} = ha1_password($data->{name}, $plaintext);
@@ -321,6 +333,10 @@ sub update_password
 sub update_profile
 {
     my ($self, $editor, $update) = @_;
+
+    if (defined $update->{username}) {
+        _die_if_username_invalid($update->{username});
+    }
 
     my $row = hash_to_row(
         $update,
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -253,7 +253,8 @@ sub _die_if_username_invalid {
 
     die 'Invalid user name' if (
         $name ne $sanitized_name ||
-        $sanitized_name =~ qr{^deleted editor \#\d+$}i
+        $sanitized_name =~ qr{^deleted editor \#\d+$}i ||
+        $sanitized_name =~ qr{://}
     );
 }
 
--- a/lib/MusicBrainz/Server/Form/Utils.pm
+++ b/lib/MusicBrainz/Server/Form/Utils.pm
@@ -207,7 +207,10 @@ sub validate_username {
     if (defined $username) {
         unless (defined $previous_username && $editor_model->are_names_equivalent($previous_username, $username)) {
             my $sanitized_name = sanitize($username);
-            if ($username ne $sanitized_name) {
+            if (
+                $username ne $sanitized_name ||
+                $sanitized_name =~ qr{://}
+            ) {
                 $self->add_error(l('This username contains invalid characters. (Check for consecutive spaces.)'));
             }
             if ($username =~ qr{^deleted editor \#\d+$}i) {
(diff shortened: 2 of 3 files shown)


lib/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -246,11 +246,22 @@ sub find_subscribers
     $self->query_to_list_limited($query, [$editor_id], $limit, $offset);
 }
 
+sub _die_if_username_invalid {
+    my $name = shift;
+    my $sanitized_name = sanitize($name);
+
+    die 'Invalid user name' if (
+        $name ne $sanitized_name ||
+        $sanitized_name =~ qr{^deleted editor \#\d+$}i
```
mwiencek, line 255: This is now different in that the deleted editor check is performed on the sanitized version of the name, so I suppose it could ISE with "Invalid user name" if the form validation only checks the name.

lib/MusicBrainz/Server/Form/Utils.pm
```diff
@@ -205,6 +206,10 @@ sub validate_username {
 
     if (defined $username) {
         unless (defined $previous_username && $editor_model->are_names_equivalent($previous_username, $username)) {
+            my $sanitized_name = sanitize($username);
+            if ($username ne $sanitized_name) {
+                $self->add_error(l('This username contains invalid characters. (Check for leading, trailing, or consecutive spaces.)'));
```
reosarevok, line 211: The only improvement I could think of would be to give the user the sanitized username back. Something like:

`This username contains invalid characters. The closest valid form of the username is {sanitized_username} - feel free to use that`

But dunno, maybe it's weird.

lib/MusicBrainz/Server/Data/Editor.pm
reosarevok: Wouldn't that get caught by the first (sanitized != name) check anyway at the form level?

reosarevok (commented):
This seems good, but please add tests

lib/MusicBrainz/Server/Data/Editor.pm
mwiencek: Yes, that sounds right actually.

lib/MusicBrainz/Server/Form/Utils.pm
mwiencek: Only issue I see is that the sanitized name might be taken.  I did notice that the form trims the username anyway, so as far as whitespace is concerned, only consecutive spaces should need to be checked.

reosarevok: You *could* check for that and give them a separate message if it is taken, FWIW. We already have a check after all :)

t/lib/t/MusicBrainz/Server/Controller/User/Register.pm
```diff
@@ -74,6 +74,24 @@ test 'Trying to register with an invalid name' => sub {
...
+        'register.username' => "very\rnew\nlines",
+        'register.password' => 'foo',
+        'register.confirm_password' => 'foo',
+        'register.email' => 'foobar@example.org',
+    });
+    like($mech->uri, qr{/register}, 'stays on registration page');
+    $mech->content_contains('username contains invalid characters', 'form has error message for newlines in username');
+
+    $mech->submit_form( with_fields => {
+        'register.username' => "consecutive  spaces",
```
reosarevok, line 88: This line is failing the no needless interpolation critic test :)

reosarevok (changes requested):

t/lib/t/MusicBrainz/Server/Controller/User/Register.pm
yvanzo: Here is the patch:

```suggestion
        'register.username' => 'consecutive  spaces',
```

lib/MusicBrainz/Server/Form/Utils.pm
mwiencek: Right, but I'm not sure if the situation is common enough to warrant having a custom message for it.  They may not even want to use the sanitized name we suggest. :)

mwiencek:
You can test this manually by typing control characters directly into the username field.  (How to do this [depends on your environment](https://superuser.com/questions/59418/how-to-type-special-characters-in-linux), but ctrl+shift+u works for me.)  You can also open the web inspector and change the username field to a textarea; this will allow you to enter newlines without any special keyboard shortcuts.

lib/MusicBrainz/Server/Form/Utils.pm
yvanzo: Implementing a custom message would be over-engineered indeed. The new check is mainly to prevent entering control characters, I don’t expect genuine users to see it; See on IRC.

yvanzo (approved):
Tested locally using the `textarea` trick. Code looks good too, even the error message. Thank you! :speedboat:

lib/MusicBrainz/Server/Form/Utils.pm
```diff
@@ -207,7 +207,10 @@ sub validate_username {
     if (defined $username) {
         unless (defined $previous_username && $editor_model->are_names_equivalent($previous_username, $username)) {
             my $sanitized_name = sanitize($username);
-            if ($username ne $sanitized_name) {
+            if (
+                $username ne $sanitized_name ||
+                $sanitized_name =~ qr{://}
```
reosarevok, line 212: "Check for consecutive spaces" doesn't make sense for this error, does it :)

yvanzo: It does for genuine mistyping. All occurrences of `://` are ill-intentioned; See the IRC backlogs.

reosarevok: I know, but if some artist or whatnot tries to make their homepage their username, they'll be very confused :)

PR #2803, opened 2023-01-10
MBS-12828: Skip removing collection collaborator

yvanzo:
# Problem: MBS-12828

Before this patch, the script would crash when trying to remove any collaboration collaborator.

# Solution

Since collection collaborators are unlikely spam accounts, these should be skipped instead of removed.

# Checklist for author

* [x] Successfully tested with an editor set as collaborator of a public collection


reosarevok (approved):
We should probably have tests for this script, but this seems trivial.

yvanzo:
Applied the patch to the cron production container already.

PR #2804, opened 2023-01-10
MBS-2349: Allow merging RGs when merging releases

reosarevok:
### Implement MBS-2349

# Problem
We used to have a checkbox before NGS came which allowed editors to enter a merge for the associated release groups when merging releases, and I still don't know why it was removed. I *think* it's because ocharles thought it was better to just let orphaned RGs get autoremoved, but that goes against the whole merge rather than delete ethos.

Readding the checkbox would allow editors to stop leaving orphaned RGs behind, without them having to go merge them by hand separately (which is enough effort to often think "you know what, whatever, they'll be garbage-collected anyway").

# Solution
This (re-)adds a checkbox "Also merge the associated release groups" when merging releases. If selected, a second "Merge release groups" edit will be entered when submitting the "Merge releases edit" (under the original editor's account since it's entered by their choice). Both edits are linked to each other by ModBot edit notes.

# Testing
So far I've tested this manually and it does what we expect it to, so putting it up to see if the idea seems sensible to others.

# Actions
* [x] Write a test.
* [x] Make the ModBot notes translatable (I guess? Although the credit edit notes when editing an artist name are not translatable).
* [x] Figure out how to make these edits also appear on the "Thank you, your edit(s)" message, which they do not right now.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -4,6 +4,7 @@ use MooseX::MethodAttributes;
 
 use MusicBrainz::Server::Constants qw(
     :edit_type
+    $EDITOR_MODBOT
     $MAX_INITIAL_MEDIUMS
     $MAX_INITIAL_TRACKS
 );
@@ -43,7 +44,7 @@ with 'MusicBrainz::Server::Controller::Role::Collection' => {
 };
 
 use List::AllUtils qw( first nsort_by uniq );
-use MusicBrainz::Server::Translation qw( l );
+use MusicBrainz::Server::Translation qw( l N_l );
 use MusicBrainz::Server::Validation qw(
     is_integer
     is_positive_integer
@@ -53,6 +54,7 @@ use POSIX qw( ceil );
 use MusicBrainz::Server::Data::Utils qw(
     artist_credit_to_ref
     boolean_to_json
+    localized_note
     partial_date_to_hash
 );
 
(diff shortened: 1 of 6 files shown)


mwiencek:
> So far I've tested this manually and it does what we expect it to, so putting it up to see if the idea seems sensible to others.

It looks sensible to me. I tested it locally it seemed to work fine, too. But a test would be nice.

> Make the ModBot notes translatable (I guess? Although the credit edit notes when editing an artist name are not translatable).

Yeah, it should use `localized_note`. I don't think we ever converted all the messages to use it.

> Figure out how to make these edits also appear on the "Thank you, your edit(s)" message, which they do not right now.

You have to use `_insert_edit` for that, AFAIK.

lib/MusicBrainz/Server/Controller/Role/Merge.pm
```diff
@@ -194,29 +199,86 @@ role {
...
+                            $self->_extra_entity_data($c, $form, $entity_id{$_}),
+                        }, @old_ids ],
+                        (map { $_->name => $_->value } $form->edit_fields),
+                        $self->_merge_parameters($c, $form, $entities),
+                    },
+                    on_creation => sub {
+                        my ($edit, $form) = @_;
+
+                        if ($c->namespace eq 'release' && $form->field('merge_rgs')->value) {
+                            my $edit_id = $edit->id;
```
mwiencek, line 231: It's sometimes hard to avoid adding entity-specific code into roles, but I think in this case it should be relatively easy to add something like an empty `_merge_on_creation` method that gets called here (kinda like `_merge_parameters` or `_extra_entity_data`) that `Controller::Release` can override with all this code. That helps keep the role easier to follow as it'll mostly contain code that applies to all entities as intended.

lib/MusicBrainz/Server/Controller/Role/Merge.pm
```diff
@@ -194,29 +199,86 @@ role {
...
+
+                            my $new_rg_id = $new->release_group->id;
+                            # We want to make sure we're merging only different RGs!
+                            my @old_entities = map +{
+                                id => $entity_id{$_}->release_group->id,
+                                name => $entity_id{$_}->release_group->name,
+                            },
+                            grep { $entity_id{$_}->release_group->id != $new_rg_id }
+                            @old_ids;
```
mwiencek, line 240: I think you can simplify this slightly by throwing a `map { $entity_id{$_}->release_group }` in there. :)
```suggestion
                            my @old_entities = map +{ id => $_->id, name => $_->name },
                                grep { $_->id != $new_rg_id }
                                map { $entity_id{$_}->release_group }
                                @old_ids;
```

mwiencek (approved):
Thanks for this patch, it works great! Can you take a look at the small refactoring suggestions before merging? If not, perhaps add a TODO so we can try to improve it later.

lib/MusicBrainz/Server/Controller/Role/Merge.pm
reosarevok: I did this (in a separate commit for ease of review). Tests pass but maybe it can be simplified / improved further, please have a quick check :)

mwiencek:
@brainzbot test this please

mwiencek (approved):
Thanks for the refactoring!

lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -386,6 +388,61 @@ sub _merge_parameters {
...
+sub _merge_on_creation {
+    my ($self, $c, $edit, $form, $new, $old_ids, $entity_id) = @_;
+    if ($form->field('merge_rgs')->value) {
+        my $edit_id = $edit->id;
+
+        my $new_rg_id = $new->release_group->id;
+        # We want to make sure we're merging only different RGs!
+        my @old_entities = map +{ id => $_->id, name => $_->name },
+            grep { $_->id != $new_rg_id }
+            map { %$entity_id{$_}->release_group }
```
mwiencek, line 400:
```suggestion
            map { $entity_id->{$_}->release_group }
```
(IMO this variable should be renamed to `entities_by_id` or something similar, too.)

mwiencek (approved):
re-approving just because

PR #2807, opened 2023-01-19
MBS-12837: Add autoselect + sidebar for Tsutaya URLs

julian45:
# Problem

Not necessarily a problem, but this would implement [MBS-12837](https://tickets.metabrainz.org/browse/MBS-12837), which would autoselect and add a sidebar spot for URLs from Tsutaya, a Japanese chain of retailers which markets CDs and includes good metadata in their listings.

# Solution

These commits _should_ work relatively smoothly to integrate Tsutaya URLs with releases. This is my first-ever dive into this codebase, so I may have bungled the regexes in particular. I referenced #2296 while putting this together, as well as some changes made in [this Bitbucket PR](https://bitbucket.org/metabrainz/musicbrainz-server/pull-requests/1500) for CDJapan. Please let me know if I missed something!

# Action

Since, as mentioned above, this is my first piece of work in this codebase, I may have misinterpreted some of the code I was referencing, so I'd appreciate it if someone could give particular attention to regexes where they're present.

Thank you for your time!


reosarevok:
@brainzbot, add to whitelist

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4892,6 +4892,14 @@ limited_link_type_combinations: [
              input_entity_type: 'artist',
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://www.abc.net.au/triplejunearthed/artist/sophisticated-dingo/',
+},
```
reosarevok, line 4895: Super minor, but this closing bracket (and the next) are wrongly indented :)

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4892,6 +4892,14 @@ limited_link_type_combinations: [
...
+  // Tsutaya
+  {
+                     input_url: 'https://shop.tsutaya.co.jp/cd/product/4562494355418/',
+             input_entity_type: 'release',
+    expected_relationship_type: 'mailorder',
+            expected_clean_url: 'https://shop.tsutaya.co.jp/cd/product/4562494355418/',
+       only_valid_entity_types: ['release'],
```
reosarevok, line 4902: You are not actually restricting it to releases though :) If you want to do so (and it probably makes sense) you need to add a `validate` function that only returns `true` for `case LINK_TYPES.mailorder.release` - you can follow the one for YesAsia.

The other `mailorder` options are artist and label. If the site has pages for artists and/or labels, then we should probably allow those too :) If not, just don't validate them (YesAsia for example seems to have no label pages so we don't do a `case LINK_TYPES.mailorder.label` at all).

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
```diff
@@ -0,0 +1,21 @@
+package MusicBrainz::Server::Entity::URL::Tower;
```
reosarevok, line 1: You forgot to change the package name :)

reosarevok (changes requested):
Good start! Requested a few changes.

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
julian45: D'oh! 🤦‍♂️ I was probably looking at the wrong window while I was typing.

root/static/scripts/tests/Control/URLCleanup.js
julian45: Got it, thanks! Tsutaya does have artist pages, so I'll add a case for that as well.

root/static/scripts/tests/Control/URLCleanup.js
julian45: Good catch! This is an easy fix, so I can definitely adjust this.

julian45:
Thanks so much for your review and helpful comments! I'll be away from my main computer for the next few days, but I should be able to make the requested changes by next Monday or thereabouts.

reosarevok:
No rush! BTW, regarding the error with the CircleCI pipeline you might want to try the following: https://support.circleci.com/hc/en-us/articles/360043002793-Troubleshooting-CircleCI-Access-After-Enabling-Github-SSO (we've seen this happen once before, and it seems it was caused by the user having deleted their CircleCI account at some point, but logging in with Github solved it).

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
julian45: This is done.

root/static/scripts/tests/Control/URLCleanup.js
julian45: This is done.

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
```diff
@@ -0,0 +1,21 @@
...
+
+sub sidebar_name { 'TSUTAYA' }
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2023 MetaBrainz Foundation
```
yvanzo, line 16: @julian45, since this is your first contribution and this line is copied from another file, I have to make sure that it is intentional.

This line is assigning copyright for this contribution to the MetaBrainz Foundation, which is the preferred way as it allows us to enforce the license effectively when needed.

But it would be totally fine too if you personally preferred to add your own name (real or nickname) on another line to have explicit attribution beyond the git log.

In either case, you remain the copyright holder of your contribution.

julian45: Would [this](https://github.com/metabrainz/musicbrainz-server/pull/2807/commits/9f5b869ea69a78569e7fa77f9528cf056608daed) work?

root/static/scripts/tests/Control/URLCleanup.js
julian45: How does [this](https://github.com/metabrainz/musicbrainz-server/pull/2807/commits/a0989edd2862296b0b4f53a402dc03fa85e9d15b) look? The regex may be a bit funky, which is related to the difference between Tsutaya's URL format and those of YesAsia and Tower Records (both of whose validators I referenced in building the original commit and this one).

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4826,6 +4826,35 @@ const CLEANUPS: CleanupEntries = {
       return url;
     },
   },
+  'tsutaya': {
+    match: [new RegExp('^(https?://)?shop\\.tsutaya\\.co\\.jp/', 'i')],
+    restrict: [LINK_TYPES.mailorder],
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/cd\/product\/(\d+)(?:\/.*)?$/, 'https://shop.tsutaya.co.jp/cd/product/$1/');
```
yvanzo, line 4833: It removes subpaths `\/.*` only whereas everything `.*$` that comes after the numeric ID `\d+` can be ignored (while still writing with a final slash `/`):

```suggestion
      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/cd\/product\/(\d+).*$/, 'https://shop.tsutaya.co.jp/cd/product/$1/');
```

Details: The numeric ID `\d+` is fully matched because the quantifier `+` is _greedy_ unless specified _lazy_ in JavaScript; Actually all quantifiers are; See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers for details.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4826,6 +4826,35 @@ const CLEANUPS: CleanupEntries = {
...
+      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/dir_result\.html\?searchType\=3\&artistCD\=(\d+)\&artistName\=(.*)?$/, 'https://shop.tsutaya.co.jp/dir_result.html?searchType=3&artistCd=$1');
```
yvanzo, line 4834: Same as above:

```suggestion
      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/dir_result\.html\?searchType\=3\&artistCD\=(\d+).*$/, 'https://shop.tsutaya.co.jp/dir_result.html?searchType=3&artistCd=$1');
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4826,6 +4826,35 @@ const CLEANUPS: CleanupEntries = {
...
+      return url;
+    },
+    validate: function (url, id) {
+      const m = /^https:\/\/shop\.tsutaya\.co.jp\/(dir_result|cd)\.*$/.exec(url);
+      if (m) {
+        const suffix = m[1];
```
yvanzo, line 4840: There is a buggy backslash `\` in front of the special character dot `.` at the end of the matching pattern. That is the only reason the tests don’t pass successfully.

Note also that it matches only the most distinctive part of valid URLs, but the pattern could be as complete as patterns in the `clean` function are:

```suggestion
      const m = /^https:\/\/shop\.tsutaya\.co.jp\/(?:(cd)\/product\/\d+\/|(dir_result)\.html\?searchType=3\&artistCd=\d+)$/.exec(url);
      if (m) {
        const suffix = m[1] || m[2];
```

yvanzo (changes requested):

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: It looks nice. There is only a small bug and possible improvements. See the other conversation https://github.com/metabrainz/musicbrainz-server/pull/2807#discussion_r1087706318 for details.

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
yvanzo: Not really, commit https://github.com/metabrainz/musicbrainz-server/pull/2807/commits/9f5b869ea69a78569e7fa77f9528cf056608daed doesn’t follow the current format (when optionally specifying your name in the code):

```
Copyright (C) 2023 MetaBrainz Foundation
Copyright (C) 2023 <Your Name>
```

Sorry if my above comment wrongly let you think that I required you to make a change here, I did not, I just gave you the legal information.

You can either remove the line you just added, or follow the above template.

It is perfectly fine not to have a second line of Copyright notice. This is what most contributors do anyway, including me even when I was just volunteering. Note that you are already credited for your contributions through the git log.

julian45: Got it, thank you for clarifying. This has been taken care of in 57a972f.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4826,6 +4826,35 @@ const CLEANUPS: CleanupEntries = {
       return url;
     },
   },
+  'tsutaya': {
+    match: [new RegExp('^(https?://)?shop\\.tsutaya\\.co\\.jp/', 'i')],
+    restrict: [LINK_TYPES.mailorder],
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/cd\/product\/(\d+).*$/, 'https://shop.tsutaya.co.jp/cd/product/$1/');
+      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/dir_result\.html\?searchType\=3\&artistCD\=(\d+).*$/, 'https://shop.tsutaya.co.jp/dir_result.html?searchType=3&artistCd=$1');
```
yvanzo, line 4834: Typo spotted through the failed CI test:
```suggestion
      url = url.replace(/^(?:https?:\/\/)?shop\.tsutaya\.co\.jp\/dir_result\.html\?searchType\=3\&artistCd\=(\d+).*$/, 'https://shop.tsutaya.co.jp/dir_result.html?searchType=3&artistCd=$1');
```

yvanzo (changes requested):
Last bug apparently.

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
yvanzo: You linked to a commit about cleanup. The “Drafted by” line is still there.

root/static/scripts/edit/URLCleanup.js
julian45: Good catch! I just saw that failure and was about to look here.

lib/MusicBrainz/Server/Entity/URL/Tsutaya.pm
julian45: I definitely wrote a line that fixed this earlier this morning, but it looks like it didn't get pushed in the moment. I guess it may have gotten lost between making changes in both my local copy and through GitHub. 4e6029abfa should have it. Sorry about that!

yvanzo:

root/static/scripts/tests/Control/URLCleanup.js
julian45: Since all bugs and improvements related to cleanup and validation have been cleaned and tested, I'm going to "resolve" this for now.

yvanzo (changes requested):
I tested it locally (because the automated tests do not cover URL display) and almost everything worked but the website’s icon which was not shown. This is because the image file has the wrong name extension `.jpg` instead of `.png` even though the content of the file is correctly in PNG format.

Please fix the image file name extension and it should hopefully be fully ready for good.

julian45:
Done in abfb40ecc2.

yvanzo (approved):
Thank you!

reosarevok (approved):
Fixed the bracket indenting with an extra commit, tested, and it seems ready.

yvanzo:
Thank you @julian45, it should be available in the next beta update, likely on next Monday.

PR #2808, opened 2023-01-20
MBS-10609: Actually check for edit notes by beginners

reosarevok:
### Fix MBS-10609

# Problem
The edit search filter for edit notes by a beginner editor was returning edits which did not have an edit note at all.

On inspection of the code, I found that we were completely replacing the search condition in `Predicate::Role::User` when the operator was `limited`, so this search was becoming simply "Editor is a beginner".

# Solution

Originally I just added a new operator `beginner_note_author` and added a new `elsif` to handle it, but @mwiencek suggested in a comment a way we'd be able to use the passed template clauses, which seems to work nicely, so I ended up using that.

# Testing

Tested by hand only.

We don't seem to currently have any decent tests for edit search ensuring the correct results are returned (only some tests that check validation of fields). Ideally we'd change that, although a proper set of edit search tests would probably require putting together a dataset with a fair amount of edits to test on.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -411,7 +411,13 @@ sub update_privileges {
                 + ($values->{spammer}               // 0) * $SPAMMER_FLAG;
 
     Sql::run_in_transaction(sub {
-        $self->sql->do('UPDATE editor SET privs = ? WHERE id = ?', $privs, $editor->id);
+        $self->sql->do(
+            'UPDATE editor SET privs = ? | (privs & ?) WHERE id = ?',
+            $privs,
+            # Preserve the value of the beginner flag.
+            $BEGINNER_FLAG,
+            $editor->id,
+        );
     }, $self->sql);
 }
 
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
@@ -48,15 +48,15 @@ role {
         if ($self->operator eq 'me' || $self->operator eq 'not_me') {
             $query->add_where([ $sql, [ $self->user->id ] ]);
         } elsif ($self->operator eq 'limited') {
-            $query->add_where([
-              "EXISTS (
-                SELECT 1
-                FROM editor
-                WHERE id = edit.editor
-                AND (privs & $BEGINNER_FLAG) > 0
-              )",
-              [ ],
-            ]);
+            my $beginner_sql = <<~'SQL';
+                SELECT id
+                  FROM editor beginner
+                 WHERE (privs & ?) > 0
+                SQL
+
+            $sql = $template_clause =~
+                s/ROLE_CLAUSE\(([^)]*)\)/$1 IN ($beginner_sql)/r;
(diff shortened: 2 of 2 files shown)


mwiencek:
It feels like we should be able to make use of the `ROLE_CLAUSE` stuff that already exists, to make the `limited` operator work for both cases. Otherwise sharing the predicate at all feels a bit odd if we have to add exceptions. Would something like this work?

```Perl
my $beginner_sql = q{
  SELECT id
    FROM editor beginner
   WHERE id != ?
     AND deleted = FALSE
     AND (
        member_since > NOW() - INTERVAL '2 weeks'
        OR NOT EXISTS (
            SELECT 1
            FROM edit e2
            WHERE e2.editor = beginner.id
            AND e2.autoedit = 0
            AND e2.status = ?
            OFFSET 9
        )
     )
};
$sql = $template_clause =~
    s/ROLE_CLAUSE\(([^)]*)\)/$1 IN ($beginner_sql)/r;
$query->add_where([ $sql, [ $EDITOR_MODBOT, $STATUS_APPLIED ] ]);
```

reosarevok:
Your suggestion seemed to work fine, at least based on the first couple pages of results for each... we really don't have a good way to test these searches, do we.

mwiencek:
> I kinda got the feeling (when rebasing/adapting https://github.com/metabrainz/musicbrainz-server/pull/2812) that using role clause like this makes things slower - is IN a very slow check?

It depends on how postgres plans the query, so they'd have to be compared with `EXPLAIN ANALYZE`. Did you notice a slowdown compared to the previous query?

reosarevok:
I *thought* it was slower but I'm not sure, neither seemed very fast. Probably worth playing with analyses.

yvanzo:
Converted to draft as it requires further investigation with `EXPLAIN ANALYZE`.

mwiencek:
I finally got around to running some performance tests here for two different queries/searches on three different commits (six queries total). The two searches were: (1) the one linked in the ticket (let's call this one "restricted," since it included an ID range), and (2) another without the edit ID range included (so, all edits with edit notes by beginners; let's call this one "all"). Each query was run three times on hendrix, and I took the best execution time of the three.

**master (broken):**
restricted: 0.449 ms
all: 4023.730 ms

**`beginner_note_author` operator:**
restricted: 0.098 ms
all: 15,037.324 ms

**`limited` operator (using `ROLE_CLAUSE`):**
restricted: 0.123 ms
all: 14,445.989 ms

So: the current implementation in this PR does not appear to be any slower than the previous one introducing a separate `beginner_note_author` operator. Both are significantly slower than the current (broken) implementation in master when querying for all edits; however they at least do not time out. It will likely be difficult to improve the performance further without having a materialized table storing beginner editor IDs.

mwiencek (approved):

reosarevok:
Should this be merged, or kept for a schema change where we introduce a view for beginner editor IDs?

reosarevok:
I fixed this already when implementing https://github.com/metabrainz/musicbrainz-server/pull/3180

reosarevok:
No, I'm an idiot :)

mwiencek (approved):
Tested the new code using `$BEGINNER_FLAG` locally, worked fine.

mwiencek:
@brainzbot test this please

PR #2809, opened 2023-01-23
MBS-12169: OpenAPI docs for WS/2 [WIP]

reosarevok:
### Implement MBS-12169

# Problem
Our API documentation sucks. It lives in three massive wiki pages, it's a huge wall of text, it does not allow people to play with it, and it's not connected to the repo in any way, so it's very easy to forget to update it when something changes in the API itself.

# Solution
Creating an OpenAPI definition that is stored in the main repository makes it easier to keep it in sync with the code itself. Even though the way the API works (with massive lists of `inc=` and `query=` options) is not the easiest thing to document, it's still a lot easier to document each possible value using tables here, listing only the values that are valid for the specific query in question. I've set this up with a Redocly config for VS Code since that allows us to validate and lint the docs, but for actual documentation presentation we probably want a Swagger instance where people can actually compose and try the API calls directly from the docs.

# Checklist for author
* [ ] Figure out with @MonkeyDo how to set up a test Swagger viewer for MB.


mwiencek (approved):
I played around with this locally, and don't think it can be overstated what a massive improvement this is over the current docs. :+1: 

How do you want to proceed with this? Since it doesn't conflict with any existing code and can be expanded incrementally, I'm fine with merging this to MBS as a work-in-progress and continuing it in separate pull requests (as you prefer). One PR per entity type probably makes sense.

 * Once the docs are complete, we can figure out the "[s]erve the bundled JSON from StaticBrainz rather than the /static folder" part; I don't think that is urgent yet.
 * It would be nice if the interactive parts used the local server rather than hard-coding musicbrainz.org, but I guess that would add some kind of build step. Seems like another thing we can figure out later.

PR #2810, opened 2023-01-23
MBS-12851: Hide ratings/tags stats for deleted users

reosarevok:
### Implement MBS-12851

# Problem

We blank tags and ratings as part of the user deletion process, so showing stats for their ratings and tags is useless in any case. I'd argue it's also kind of misleading, since it might give the impression they had never tagged/rated before (since their added entity data for example still shows the correct numbers on the adjacent statistics table).

# Solution

As suggested by @Aerozol, rather than deleting the whole table, we specify that the data is deleted:
Screenshot from 2023-02-21 16-51-09

# Testing

Tested manually: both that a deleted editor, `/user/Deleted%20Editor%20%23667117`, shows no stats and says "Deleted", and that a non-deleted editor, `/user/reosarevok`, does still show them.

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -793,7 +793,13 @@ const UserProfileStatistics = ({
             {hasPublicTags ? (
               <>
                 <UserProfileProperty name={l('Tags upvoted')}>
-                  {$c.user && upvotedTagCount > 0 ? exp.l(
+                  {user.deleted ? (
+                    <abbr
+                      title={l('Tags are removed when an editor is deleted.')}
+                    >
+                      {lp('Deleted', 'tags')}
+                    </abbr>
+                  ) : $c.user && upvotedTagCount > 0 ? exp.l(
                     '{count} ({view_url|view})',
                     {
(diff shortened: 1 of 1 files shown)


yvanzo (commented):
I would suggest to rather inform the visitor that tags and ratings have been deleted too if any, as it isn’t documented otherwise.

reosarevok:
Hmm, that might work too. Would you replace the number with "Deleted" or the whole table with some infobox?

yvanzo:
The latter.

Aerozol:
What about:

image

With a rollover like: “Secondary data is removed when a editor is deleted”. Seems clear without cluttering the table?

Replace ‘removed’ with ‘deleted’ if wanted.

Aerozol:
Or maybe this is clearer, showing what’s gone (same rollover/tooltip as in comment above):

image

Or:

image

The tooltip can do the heavy lifting.

reosarevok:
Ok, sounds good! I updated it to look kinda like the first example on your last comment :)

yvanzo (approved):
Thank you both! :ship:

PR #2813, opened 2023-01-25
MBS-12847: Allow searching for edits without any notes 

reosarevok:
### Implement MBS-12847

On top of https://github.com/metabrainz/musicbrainz-server/pull/2808 to avoid possible conflicts.

# Problem
Edits without notes are a common mark of either beginners not quite sure of what they are doing (common risk of human error) or expert editors a bit too sure of what they are doing (also a common risk of human error). Despite that, we do not currently have any way to search for them specifically, AFAICT.

# Solution
This adds an option under "Edit Note Authors" (seemed the best fit) with the wording "do not exist (edit has no notes)". It might read a bit strangely, but it's the best I could think of - improvement suggestions welcome.

It also adds a new vote suggestion for beginner edits without notes, and another for destructive edits without notes. These seem like the most obvious use cases where a careful voter could be of use.

# Testing
Tested by hand (checked the results and they indeed had no edit notes) - it's also a very straightforward condition check.
Tested the new vote suggestion searches from `/vote`.

### DIFF
--- a/root/edit/search_macros.tt
+++ b/root/edit/search_macros.tt
@@ -357,6 +357,7 @@
                    [ 'not_subscribed', l('do not include editors in my subscriptions') ]
                    [ 'limited', l('include a beginner editor') ]
                    [ 'not_edit_author', l('include someone other than the edit author') ]
+                   [ 'nobody', l('do not exist (edit has no notes)') ]
                 ], field_contents) %]
   [% ELSE %]
     [% operators([ [ '=', l('is') ]
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/Role/User.pm
@@ -33,6 +33,7 @@ role {
             'not_me' => 0,
             'limited' => 0,
             'not_edit_author' => 0,
+            'nobody' => 0,
         );
     };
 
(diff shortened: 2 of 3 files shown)


reosarevok:

reosarevok:

root/edit/search_macros.tt
```diff
@@ -357,6 +357,7 @@
                    [ 'not_subscribed', l('do not include editors in my subscriptions') ]
                    [ 'limited', l('include a beginner editor') ]
                    [ 'not_edit_author', l('include someone other than the edit author') ]
+                   [ 'nobody', l('do not exist (edit has no notes)') ]
```
mwiencek, line 360: Maybe "include nobody" to match the language of the other options?

I was wondering if it'd be more useful to implement this as "Edit note count" or "Edit note author count" parameters instead (which you could specify 0 for), but this seems fine too.

reosarevok: "edit notes include nobody" seems super confusing compared to "edit notes do not exist" though, which is why I did it this way originally - it at least feels like English :sweat_smile:

mwiencek: It's "note authors include nobody" though, not "notes include nobody." :)

I think both sound kinda weird TBH, though we can always change the string later.

mwiencek (approved):
Tested locally, worked well for me.

root/edit/search_macros.tt
reosarevok: Ah, true - both do sound weird (it's the crappy situation of trying to build "real sentence" edit searches, which also translate horribly - we should probably change the approach entirely).

PR #2815, opened 2023-01-28
Document React state management

mwiencek:
# Problem

We are lacking developer documentation on how to best manage React state for complex pages.  This can also help in understanding how existing code, like the relationship editor, is structured.

# Solution

Adds a section about React state management to HACKING.md and an example component in root/static/scripts/tests/examples/todo-list/.  I'm not sure if the example is helpful or necessary -- if not we can just remove it.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -551,6 +551,222 @@ Common instructions for porting:
 
  7. All components should be named following the `UpperCamelCase` convention.
 
+React state management
+----------------------
+
+For simple components with a small amount of isolated state,
+`React.useState` works great.
+
+For complex pages (let's call these apps), where an action in one deeply-
+nested part of the component tree can affect the state in a sibling or
+higher-up part of the tree (e.g. in the relationship editor), it's
+recommended to have one [reducer](https://react.dev/learn/extracting-state-logic-into-a-reducer)
+in the very top component and pass all state downward as props.  React calls
+this concept [lifting state up](https://react.dev/learn/sharing-state-between-components)
+to create a "top-down" data flow.
+
+While having all state flow in one direction greatly simplifies reasoning
+about how changes are propagated and how state is kept in sync across many
+components, it also creates two issues that you need to manage:
+
+  * The first issue is related to performance: if all state comes from the
+    top component, then React has to re-render the entire page each time
+    any piece of state changes.  There are steps to address this, but they
+    must work together (i.e. one step on its own won't help, but will when
+    combined with the rest).
+
+    1. Make sure your components don't get too large and have a clear
+       purpose. If you can identify parts of a component that can render
+       (update) less often than the other parts, consider splitting those
+       parts into separate components.
+
+    2. Make sure every component only receives the props it actually needs.
+       For example, consider a component that receives the number of elements
+       of a list as a prop, but then only performs a boolean check on the
+       number, like `numItems > x`.  In this case you should just pass the
+       result of `numItems > x` as a prop.
+
+    3. Make liberal use of
+       [React.memo](https://react.dev/reference/react/memo),
+       especially for components that have complicated render functions.
+       Assuming you also follow the previous point (passing only the
+       essential information used by the component), then `React.memo` will
+       ensure that rendering can be skipped when the props don't change.
+       The cost of React comparing the props for a small sub-tree will be
+       much less than the cost of rendering the entire tree again.
+
+    It's highly recommended to install the
+    [React Developer Tools](https://react.dev/learn/react-developer-tools)
+    browser extension to identify areas where components are updating
+    unnecessarily.  With the tools pane open, click on the settings cog
+    and make sure the following checkboxes are enabled:
+
+      * "Highlight updates when components render," under "General."
+
+      * "Record why each component rendered while profiling," under
+         "Profiler."
+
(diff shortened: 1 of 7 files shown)


HACKING.md
```diff
@@ -520,6 +520,228 @@ Common instructions for porting:
...
+For simple components with a small amount of isolated state,
+`React.useState` works great.
+
+For complex pages (let's call these apps), where an action in one deeply-
+nested part of the component tree can affect the state in a sibling or
+higher-up part of the tree (e.g. in the relationship editor), it's
+recommended to have one [reducer](https://beta.reactjs.org/learn/extracting-state-logic-into-a-reducer)
+in the very top component and pass all state downward as props.  React calls
+this concept [lifting state up](https://reactjs.org/docs/lifting-state-up.html)
+to create a [top-down data flow](https://reactjs.org/docs/state-and-lifecycle.html#the-data-flows-down).
```
reosarevok, line 535: The first link says "The new docs will soon replace this site, which will be archived" and points to https://beta.reactjs.org/learn/sharing-state-between-components - not sure if you feel that's better?

HACKING.md
```diff
@@ -520,6 +520,228 @@ Common instructions for porting:
...
+    1. Make sure your components don't get too large and have a clear
+       purpose. If you can identify parts of a component that can render
+       (update) less often than the other parts, consider splitting those
+       parts into separate components.
+
+    2. Make sure every component only receives the props it actually needs.
+       A contrived example of this might be a component that receives the
+       number of elements of a list as a prop, but then only performs a
+       boolean check on the number, like `numItems > x`.  In this case you
+       should just pass the result of `numItems > x` as a prop.
```
reosarevok, line 556: I don't find the example too contrived - in fact, I wouldn't be surprised if some of our code might be passing down the actual list for this check, not even just the number :D

HACKING.md
```diff
@@ -520,6 +520,228 @@ Common instructions for porting:
...
+       especially for components that have complicated render functions.
+       Assuming you also follow the previous point (passing only the
+       essential information used by the component), then `React.memo` will
+       ensure that rendering can be skipped when the props don't change.
+       The cost of React comparing the props for a small sub-tree will be
+       much less than the cost of rendering the entire tree again.
+
+    It's highly recommended to install the
+    [React Developer Tools](https://beta.reactjs.org/learn/react-developer-tools)
+    browser extension to indentity areas where components are updating
```
reosarevok, line 569: identify?

HACKING.md
```diff
@@ -520,6 +520,228 @@ Common instructions for porting:
...
+    Even though there's only one "real" reducer, we don't have to make it
+    handle every single action in the app on its own. We can still define
+    separate reducers for each child component, and simply call them from the
+    parent reducer.
+
+    A general outline of the types and functions you'd typically define in
+    a component file are described below. These are specific to the file's
+    component, and handle all the actions/state used by that component. The
+    parent doesn't have to worry about the child's actions unless it needs
+    to.
```
reosarevok, line 607: It doesn't have to unless it has to? :) What do you have in mind?

HACKING.md
```diff
@@ -520,6 +520,228 @@ Common instructions for porting:
...
+        etc., along with any additional data needed by the action.
+
+      * `type StateT`: A read-only type that describes the state which is
+        needed to display the component and dispatch actions.
+
+      * `type PropsT`: In many cases, the only props you need are `dispatch`
+        (`(ActionT) => void`) and `state` (`StateT`), though it also makes
+        sense to pass static values that never change as props instead of
+        including them in `state`. If you need these props in the `reducer`
+        function, having then in `state` is convenient, though.
```
reosarevok, line 621:
```suggestion
        function, having them in `state` is convenient, though.
```

reosarevok (commented):
This sounds good in general, I guess I will try it next time I'm writing or editing a component like this and try to figure out what I miss and still needed to ask about :)

HACKING.md
mwiencek: It felt weird linking to something with beta in the subdomain, but some of the docs only existed on the beta site, so.  I guess I'll just use beta for everything and fix the links later if they break.

HACKING.md
mwiencek: Basically what is explained further down for `update-child`:

```js
          /*
           * You could even have another switch statement here on
           * childAction.type, in case you need to handle particular actions
           * in the parent reducer (because they affect state in the parent
           * that the child doesn't know about).
           */
```
But I'll just remove this sentence because it's confusing and doesn't add anything.

HACKING.md
```diff
@@ -551,6 +551,222 @@ Common instructions for porting:
...
+        make a copy:
+
+        ```js
+        const newState = {...state};
+        // ... and while handling some action:
+        newState.someList = [...newState.someList, newItem];
+        ```
+
+        If you need to make complex updates to deeply-nested properties,
+        use the `mutate-cow` library.
```
mwiencek, line 669: There used to be a sentence here suggesting to use `createFastObjectCloneFunction`, but that should rarely be necessary unless it's performance-critical code. I replaced it with a suggestion to use mutate-cow.

reosarevok: :cow:

PR #2816, opened 2023-01-30
MBS-12857 / MBS-12858: Details for recent entities in autocompletes

reosarevok:
### Fix MBS-12857 / Implement MBS-12858

# Problem
When using recent entities in an autocomplete search, the data returned is sometimes not enough to be sure of which entities they actually are. This is specially true of areas and places, which have containment info when you search for them, but no longer have it when loaded as part of the list of recent entities.

Additionally, in beta, a regression has been added where the entity type is also not saved (or rather, not loaded) in the recent entities list.

# Solution
Since beta uses the `WS::js` `/entities` endpoint to load all the recent items from their ID, I added a few extra conditional loads there, for entity types on all entities with them and for area and containment for areas/places.

# Testing
Manually, by searching for items in Add relationship popups and then reloading the page and checking what data the recent entities have. Almost any search should do the trick for types; I tested release groups separately since those have more complex types. For area containment you obviously will want to search for places and areas (you can do all of these from `/artist/create` if you use the artist area search for areas).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/WS/js.pm
+++ b/lib/MusicBrainz/Server/Controller/WS/js.pm
@@ -622,8 +622,14 @@ sub entities : Chained('root') PathPart('entities') Args(2)
         }
     }
 
+    my @entities = values %{$results};
+
     if ($ENTITIES{$type_name}{artist_credits}) {
-        $c->model('ArtistCredit')->load(values %{$results});
+        $c->model('ArtistCredit')->load(@entities);
+    }
+
+    if ($ENTITIES{$type_name}{type}) {
+        $c->model(type_to_model($type_name) . 'Type')->load(@entities);
     }
 
     while (my ($id, $entity) = each %{$results}) {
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Controller/WS/js.pm
```diff
@@ -626,6 +626,22 @@ sub entities : Chained('root') PathPart('entities') Args(2)
         $c->model('ArtistCredit')->load(values %{$results});
     }
 
+    if ($ENTITIES{$type_name}{type}) {
+        $c->model(type_to_model($type_name) . 'Type')->load(
+            values %{$results},
```
mwiencek, line 631: Can we store `my @entities = values %{$results};` above the artist_credits stuff above now that it's repeated 5 times?

lib/MusicBrainz/Server/Controller/WS/js.pm
```diff
@@ -626,6 +626,22 @@ sub entities : Chained('root') PathPart('entities') Args(2)
...
+    }
+
+    if ($type_name eq 'area') {
+        $c->model('Area')->load_containment(values %{$results});
+    }
+
+    if ($type_name eq 'place') {
+        $c->model('Area')->load(values %{$results});
+        $c->model('Area')->load_containment(map { $_->area } values %{$results}
+        );
```
mwiencek, line 642: (`);` can go on the previous line)
```suggestion
        $c->model('Area')->load_containment(map { $_->area } @entities);
```

mwiencek (approved):
lgtm but suggested some minor cleanups

mwiencek (approved):

PR #2818, opened 2023-02-02
MBS-12859: Allow relationship credit to be the same as name 

reosarevok:
# Problem MBS-12859
The new relationship editor ignores / silently drops entity credits for relationships when they match the name. This is a break with the way the previous relationship editor worked, and it was a feature, not a bug: it was being intentionally used before to ensure credits remained as printed even if the artist name itself changed (for example, a Russian orchestra credited by a Latin name, where we want the printed Latin credit to stay the same even if someone correctly improves the artist name to the original Russian later).

A minor secondary issue found while testing the change is that https://github.com/metabrainz/musicbrainz-server/pull/2805 only applied to entities without a credit, since `EntityLink` hides the disambiguation when custom content is passed unless explicitly told not to (probably so that it won't show disambiguations in things like tab titles).

# Solution
I removed the extra code that made it so that credits matching the name would be ignored. Now you can again intentionally enter a credit matching the entity name. I also made it so the checkbox for changing artist credits for other uses of the entity will always show if there's a credit, even if it's the same as the entity name; this way, it's trivial to blank all credits matching the entity name if the user decides that's the right thing to do.

I also explicitly told `EntityLink` to always show disambiguations for the relationship preview entities.

# Testing
Manually, by adding credits same as (and different from) the entity name and seeing whether they actually were retained.

# Comments
The preview now can be a bit confusing when removing or adding a credit that matches the entity name. In general, it might make sense to have a way to tell that an entity has a credit but it matches its name, so I added MBS-12862 for that. It seems outside the scope of this PR though, since we already have the issue elsewhere.

### DIFF
--- a/root/static/scripts/relationship-editor/components/DialogEntityCredit.js
+++ b/root/static/scripts/relationship-editor/components/DialogEntityCredit.js
@@ -104,6 +104,7 @@ const DialogEntityCredit = (React.memo<PropsT, void>(({
   let changeCreditsSection;
   if (
     state.creditsToChange ||
+    state.creditedAs !== '' ||
     state.creditedAs !== origCredit.current
   ) {
     changeCreditsSection = (
--- a/root/static/scripts/relationship-editor/components/DialogEntityCredit.js
+++ b/root/static/scripts/relationship-editor/components/DialogEntityCredit.js
@@ -71,8 +71,7 @@ const DialogEntityCredit = (React.memo<PropsT, void>(({
   state,
   targetType,
 }: PropsT): React.MixedElement => {
-  const creditedAsOrName = state.creditedAs || entityName;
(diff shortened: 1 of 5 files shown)


reosarevok:

mwiencek:
There's one issue which I think the Selenium tests failure is related to.  If you edit a relationship with an existing entity credit and clear the credit, it doesn't show the "Change credits for other ..." checkbox.

reosarevok:
Whoops. Ok, let's see if selenium is happy now.

yvanzo (approved):
I’ve tested filling artist credit to artist name, and blanking artist name from artist credit. Both resulted into appropriate raw edit data.

mwiencek:
I updated the PR to add a basic Selenium test, will merge now.

PR #2819, opened 2023-02-02
MBS-12863: Tooltips too far to the left

mwiencek:
# Problem MBS-12863

# Solution

This fixes the tooltip positioning in the external links editor and relationship dialog. An override has been added for the batch tool buttons, to keep the existing 50% positioning, since 100% makes it possible for the last button's tooltip to clip outside the viewport.

### DIFF
--- a/root/static/styles/relationship-editor.less
+++ b/root/static/styles/relationship-editor.less
@@ -1,7 +1,15 @@
 #content.rel-editor #tracklist td {vertical-align: top;}
 #content.rel-editor #tracklist td.midcol {vertical-align: middle; font-size: @small-text;}
 #content.rel-editor #tracklist input[type="checkbox"] {vertical-align: middle;}
-#content.rel-editor #batch-tools {width: 100%; margin: 0.8em;}
+#content.rel-editor #batch-tools {
+    width: 100%;
+    margin: 0.8em;
+    /*
+     * Center the tooltips so that they're less likely to clip
+     * outside the viewport.
+     */
+    .tooltip-container { left: 50%; }
+}
 #content.rel-editor div.ars {padding: 0; margin: 0.5em 0; margin-left: 1.5em;}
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
Ideally I guess we'd still keep them in the center of the release editor bit, but this is an improvement.

What is the 50% / 100% of right now on the external links editor? I guess it's not doable to position it at 50% right from the start of the ? icon (so the enter of it)?

yvanzo (approved):
Successfully tested in the “External links” section of the release editor, in the relationship editor (for release and release group), in the dialog box opened from the “Batch-add a relationship to recordings” button: It no longer hides the `?` button. :100: 

> It makes the tooltip on last the batch tool button slightly worse in that it's cut off if the window is small enough

It was not cut off by chance. It wrongly assumed that it was not localized or that the length of the button text and the length of the tooltip text will keep the same ratio when localized.

mwiencek:
> What is the 50% / 100% of right now on the external links editor? I guess it's not doable to position it at 50% right from the start of the ? icon (so the enter of it)?

It's positioned within the containing block (.tooltip-wrapper).  So it doesn't know where the icon is within the containing block.

reosarevok:
Unless changing that so that it knows is easy, just merge as-is, I guess. Alternatively, just make the position a parameter and override the default for one of the two.

mwiencek:
I added an override for all of the batch tool buttons (all will stay at 50%).  Does that seem okay?

reosarevok:
Yes :)

yvanzo:
It’s now hiding half of the label, but it still is an improvement over the current situation.

reosarevok:
That is how it was when first implemented, I think it's fine to momentarily hide part of the label on hover given it goes away when you stop hovering :) In any case, probably worth merging this now and seeing what people have to say.

PR #2820, opened 2023-02-04
MBS-12861: Improve "Two relationships with the same key" message

mwiencek:
# Problem

It's difficult to debug issues like MBS-12860 or MBS-12861 without more information.

# Solution

 * Improves the "Two relationships with the same key" error message by outputting a human-readable display of which two relationship states conflict (implemented by `prettyPrintRelationshipState`).

 * Adds a `_lineage` property to `RelationshipStateT` which is a list of strings that describe where the state came from.  It's set to an initial value whenever state is created ("loaded from database," "seeded," "added") and appended to whenever `cloneRelationshipState` is used to modify it.

   The lineage data will be outputted in the human-readable display above.


mwiencek:
Since I'm not actually sure how to trigger this error, I could only test it by modifying the source to throw the error on page load:

```diff
diff --git a/root/static/scripts/release/components/ReleaseRelationshipEditor.js b/root/static/scripts/release/components/ReleaseRelationshipEditor.js
index 5817e15724..d185f5d700 100644
--- a/root/static/scripts/release/components/ReleaseRelationshipEditor.js
+++ b/root/static/scripts/release/components/ReleaseRelationshipEditor.js
@@ -488,7 +488,7 @@ function* getAllRelationshipEdits(
     ) {
       const relationshipKey = getRelationshipKey(relationship);
       const seenRelationship = seenRelationships.get(relationshipKey);
-      if (seenRelationship) {
+      if (true) {
         /*
          * It's normal to come across the same relationship twice if a
          * recording or work is reused.  But they should always reference
@@ -498,7 +498,7 @@ function* getAllRelationshipEdits(
         if (seenRelationship !== relationship) {
           throw new Error(
             'Two relationships with the same key:\n\n' +
-            prettyPrintRelationshipState(seenRelationship) + '\n' +
+            prettyPrintRelationshipState(relationship) + '\n' +
             prettyPrintRelationshipState(relationship)
           );
         }
```

reosarevok (commented):
The idea seems good - would it make sense to add a test that actually checks the lineage is added and followed correctly? Or would that be too complex since it changes based on many things?

mwiencek:
> The idea seems good - would it make sense to add a test that actually checks the lineage is added and followed correctly? Or would that be too complex since it changes based on many things?

Hmm, well there are too many combinations to test it completely, and since it's only for debugging I think it's probably more effort than it's worth (given that there are more pressing issues).

reosarevok (approved):
Let's put this on beta so we can hopefully figure out what's still triggering this error then.

PR #2821, opened 2023-02-05
MBS-12860: Fix "Two relationships with the same key" error

mwiencek:
Likely fixes MBS-12860

This reverts commit 6230a40b799f0879408fe03d79d8fce83293c0d3.

That commit caused new relationship IDs to become reused by memoizing the initial relationship data seeded to dialogs.  It can be safely reverted in full because it was an unnecessary optimization in the first place.

You can trigger the reported error by adding two area relationships to any entity in the release relationship editor.  You can trigger it for target types other than area too, but the catch is that the memoized ID only changes when the target type changes, *after* the dialog is closed. So to trigger it for artist relationships from a fresh page, you'd have to add three artist relationships to the same entity.

While I can't be 100% certain this is the only cause of the key error, it seems pretty likely that it's the most common one.

Testing:

I added a Selenium test that performs the above steps and checks that the edits are submitted correctly. Without the patch applied, it fails due to the key error alert.

### DIFF
--- a/root/static/scripts/common/components/ButtonPopover.js
+++ b/root/static/scripts/common/components/ButtonPopover.js
@@ -33,8 +33,7 @@ type PropsT = {
   +wrapButton?: (React.MixedElement) => React.MixedElement,
 };
 
-const ButtonPopover: React.AbstractComponent<PropsT, mixed> =
-React.memo((props: PropsT): React.MixedElement => {
+const ButtonPopover = (props: PropsT): React.MixedElement => {
   const {
     buttonContent,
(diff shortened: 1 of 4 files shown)


mwiencek:
BTW, you don't even have to add identical relationships differing only by date to reproduce this.  If I add two release relationships with different link types, it also errors.  (Will add a test for this too.)

reosarevok (approved):
Reverting a dodgy commit seems trivial enough and the added test is nice, thanks. We should put this out and see if anyone keeps hitting the issue after it.

PR #2822, opened 2023-02-05
MBS-12832: Improve edit submission batching in the release relationship editor

mwiencek:
# Fix MBS-12832

## Problem

Prior to this commit, the release relationship editor would group edit submission requests by entity, and submit them at a rate of one entity per second; so if you made 100 edits to one recording it would submit them all in one request, but if you made 1 edit each to 100 recordings it would take roughly 100 seconds to complete.

("Per second" is a rough estimate, since what it actually does is wait for each request to complete before starting the next one, and additionally wait 500 ms before each request.)

## Solution

The implementation in this commit is more consistent and much faster in the average case.  Requests are no longer grouped by entity; we simply submit 25 edits per request, for all requests.

## Testing

I made 100+ edits to a release locally and watched the network request log in the browser development tools, which showed each request submitting 25 edits (other than the last one, of course).  The Selenium tests which submit a smaller number of edits also pass.

### DIFF
--- a/root/static/scripts/common/utility/natatime.js
+++ b/root/static/scripts/common/utility/natatime.js
@@ -1,0 +1,26 @@
+/*
+ * @flow strict
+ * Copyright (C) 2023 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export default function* natatime<T>(
+  n: number,
+  values: Iterable<T>,
+): Generator<Array<T>, void, void> {
+  invariant(n > 0);
+  let chunk: Array<T> = [];
+  for (const value of values) {
+    chunk.push(value);
+    if (chunk.length === n) {
+      yield chunk;
+      chunk = [];
+    }
+  }
+  if (chunk.length) {
+    yield chunk;
+  }
+}
--- a/root/static/scripts/release/components/ReleaseRelationshipEditor.js
+++ b/root/static/scripts/release/components/ReleaseRelationshipEditor.js
@@ -45,6 +45,7 @@ import {
 import clean from '../../common/utility/clean.js';
 import isDatabaseRowId from '../../common/utility/isDatabaseRowId.js';
 import isDateEmpty from '../../common/utility/isDateEmpty.js';
+import natatime from '../../common/utility/natatime.js';
 import {uniqueNegativeId} from '../../common/utility/numbers.js';
 import setMapDefault from '../../common/utility/setMapDefault.js';
 import sleep from '../../common/utility/sleep.js';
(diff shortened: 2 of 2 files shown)


root/static/scripts/common/utility/natatime.js
```diff
@@ -0,0 +1,26 @@
+/*
+ * @flow strict
+ * Copyright (C) 2023 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+export default function* natatime<T>(
```
mwiencek, line 10: I named this after https://metacpan.org/pod/List::MoreUtils#natatime-EXPR,-LIST

root/static/scripts/release/components/ReleaseRelationshipEditor.js
```diff
@@ -406,14 +407,7 @@ async function submitWorkEdits(
...
 ): Generator<
-  {
-    +edits: Generator<
-      [Array<RelationshipStateT>, WsJsEditRelationshipT],
-      void,
-      void,
-    >,
-    +entity: CoreEntityT,
-  },
+  [Array<RelationshipStateT>, WsJsEditRelationshipT],
```
mwiencek, line 410: We didn't actually use the `entity` property for anything even before this patch, so I simplified the generator type significantly.

mwiencek:
> The implementation in this commit is more consistent and much faster in the average case. Requests are no longer grouped by entity; we simply submit 25 edits per request, for all requests.

One exception to this is work creation edits.  Those always have to be submitted before the relationships are, and I believe we create all the works in a single request.

reosarevok:
> I believe we create all the works in a single request.

Could this potentially be problematic in a huge release with all-new works?

reosarevok:
You did not add a Testing section to the PR - what testing did you do? I assume you did some with more than 25 edits, but it'd be good to know :)

mwiencek:
> Could this potentially be problematic in a huge release with all-new works?

It hasn't been a reported issue yet AFAIK, but potentially.  I think it'd make sense to batch those into smaller requests too.

> You did not add a Testing section to the PR - what testing did you do? I assume you did some with more than 25 edits, but it'd be good to know :) The idea of doing fixed sets of 25 seems good, anyway. I expect if one in the set of 25 hits an error, the 25 will be rolled back and shown to the user or?

Added a testing section now.  If an error is hit, the submission should stop and show the error to the user since e5711a9ff48dbe5914947774a7f544be8f5a07c2 (and they should be able to continue the submission where it left off by removing the problematic edits if possible).  And yeah, all of the edits in the batch that errored will be rolled back since [they're submitted together in a single transaction](https://github.com/metabrainz/musicbrainz-server/blob/b6cd917/lib/MusicBrainz/Server/Controller/WS/js/Edit.pm#L746).

reosarevok (approved):
Based on the testing, it seems good to put this in beta to let users at it. A quick code check looks fine to me.

PR #2824, opened 2023-02-07
MBS-12874: Don't split new relationships with one instrument or vocal

mwiencek:
# Fix MBS-12874

## Problem

When you edit a relationship, we look at all the instruments and vocals added and try to split each one into a separate relationship.  That's handled by `splitRelationshipByAttributes`.

If the relationship being edited is new, i.e. not submitted yet, then prior to this commit we would incorrectly "split" this relationship into a new one (a clone) even if it only had a single instrument or vocal attribute.  The correct course of action is to simply return the same relationship back.

One side effect of not doing that was having triggered an error where the same relationship state is tried to be removed twice: once to replace the existing relationship state as edited by the dialog, and another to remove it again when the bogus "split" relationship (which is identical to the original other than by ID) is merged into the original by `mergeRelationship`.  The second removal fails because the relationship no longer exists at that point.

### Steps to reproduce

1. Open the relationship editor for an entity (try a recording).
2. Add an artist relationship of type "instrument" with a single instrument, or of type "vocal" with a single vocal.
3. Edit the relationship you just added and enter a date.

An error should occur once you accept the dialog after step 3.

## Solution

Modify `splitRelationshipByAttributes` so that it doesn't attempt to split new relationships with a single instrument or vocal, preserving the initial relationship state.  Since no clone is created, nothing is attempted to be merged into the original relationship.

## Testing

Tested manually using the steps above.  Added a Selenium test and expanded the unit tests for `splitRelationshipByAttributes` to cover this case.


reosarevok:
The general idea seems sensible. The whole of `splitRelationshipByAttributes` could probably use step-by-step comments to clarify exactly what it does and does not, but that's probably true of most of our code.

reosarevok:
While testing this, I found another bug that I expect is also related with the splitting code: MBS-12885

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
@reosarevok it's actually a similar issue, and the fix for it also fixes MBS-12874 without changing `splitRelationshipByAttributes` at all.  But the change to `splitRelationshipByAttributes` still makes sense IMO, so I'm just pushing it as a separate commit.

reosarevok (approved):
Tested it and it works nicely now for both cases, plus having the two tests is neat.

PR #2825, opened 2023-02-07
MBS-12886: Restrict Bandcamp /(album|track) to "get the music" rels

reosarevok:
### Implement MBS-12886

# Problem
While Bandcamp /track and /album URLs can match a fair bunch of "get the music" relationships, they do not match "paid streaming" nor any non-"get the music" rels. We didn't use to be able to do this kind of restriction, but we can now (e.g. https://github.com/metabrainz/musicbrainz-server/pull/2321 which did something similar for Soundcloud).

Having a lot of possible combinations is kind of messy and hard to follow in the `restrict` section of the handler, as exemplified by that same Soundcloud PR.

# Solution
The first commit here refactors the Soundcloud implementation with an `anyCombinationOf` that does what the name claims, allowing any combination of the relationship types provided to it. The commit also extends the Soundcloud allowed combinations since the previous limitation seemed to be just "we didn't find more on a quick check" and mostly due to the absurdity of adding all combinations in the previous style.

The second commit uses `anyCombinationOf` to limit the Bandcamp options:
* It restricts Bandcamp `/album` (release) pages to releases only, and to any combination of free download, purchase for download, purchase for mail order and stream for free.
* It restricts Bandcamp `/track` (album track, sometimes single release) pages to releases and recordings only, and to any combination of free download, purchase for download, and stream for free (plus mail order for releases just in case). It doesn't make sense to support mail order for recordings - if you can purchase it physically, surely it's a release, anyway.
* Since `/merch` links are mail-order only, rather than complicating this even more I just split them into their own cleanup like we do with `/campaign` ones, allowing release mail order only.
* It moves the checks that blocked adding artist/label Bandcamp pages at the release/recording level into the actual Bandcamp cleanup, which makes a lot more sense.

# Testing
Manually, plus with the usual handler tests.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -1019,6 +1041,64 @@ const CLEANUPS: CleanupEntries = {
...
+        case LINK_TYPES.downloadpurchase.recording:
+        case LINK_TYPES.streamingfree.recording:
+          if (/^(https?:\/\/)?([^\/]+)\.bandcamp\.com\/?$/.test(url)) {
+            return {
+              error: exp.l(
+                `This is a Bandcamp profile, not a page for a specific
+                 recording. Even if it shows a single recording right now,
+                 that can change when the artist releases another.
+                 Please find and add the appropriate recording page
+                 (“{single_url_pattern}”)
```
yvanzo, line 1054: Should it be a recording or a release?
Any actual example at hand?

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1068,6 +1148,23 @@ const CLEANUPS: CleanupEntries = {
...
+      return url.replace(/^(?:https?:\/\/)?([^\/]+)\.bandcamp\.com\/merch\/([^\/]+)?.*$/, 'https://$1.bandcamp.com/merch/$2');
+    },
+    validate: function (url, id) {
+      switch (id) {
+        case LINK_TYPES.mailorder.release:
+          return {result: /^https:\/\/[^\/]+\.bandcamp\.com\/merch\/[\w-]+$/.test(url)};
+      }
+      return {result: false, target: ERROR_TARGETS.ENTITY};
+    },
+  },
```
yvanzo, line 1167: Any current example?

The test URL `https://non-serviam-records.bandcamp.com/merch/the-howling-void-megaliths-of-the-abyss-digipak` is not found.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -5873,6 +5857,47 @@ entitySpecificRules.release_group = function (url) {
...
+    }
+
+    const resultOption: {
+      [entityType: RelatableEntityTypeT]: EntityTypes,
+    } = {};
+    resultOption[entityType] = option;
+    result.push(resultOption);
+  }
+
+  return result;
```
yvanzo, line 5942: It seems to be a pretty complicated implementation but I don’t have suggestion at hand.

yvanzo (commented):
Adding a function for any type combination is great!
Made a few comments, not sure about the test coverage. LGTMBDNT otherwise.

root/static/scripts/edit/URLCleanup.js
reosarevok: This is shown specifically when trying to add a link to a recording, so the only things allowed should be links like https://pyrrhonband.bandcamp.com/track/the-cost-of-living

root/static/scripts/edit/URLCleanup.js
reosarevok: Happy to get any suggestions to simplify it, but this is what I came up with that worked after fighting it for a while back in the day. Maybe there's a much simpler option though :)

root/static/scripts/edit/URLCleanup.js
yvanzo: Can Bandcamp profile points to track?
Any actual example at hand?

root/static/scripts/edit/URLCleanup.js
yvanzo: You know _NO FEAR_. Correct, I found https://web.archive.org/web/20191110201058/https://non-serviam-records.bandcamp.com/merch/the-howling-void-megaliths-of-the-abyss-digipak

root/static/scripts/edit/URLCleanup.js
reosarevok: It can, if just one single is all there is. I've seen it in the past, but not finding any examples right now :)

yvanzo: Isn’t single matching release rather than recording?

Also it seems less effort to add the Bandcamp profile URL to the corresponding MB artist first, and then to try to find the correct Bandcamp URL for the current recording/release.

reosarevok: Well, if it is still /track, then it belongs at both the release and recording level, really. But if you think the error message is not great, feel free to suggest improvements - this is not a new message, but the same we already had, so any improvements probably should be a separate commit anyway though :)

yvanzo: It’s not great but it’s not about this pull request indeed. :)

yvanzo (approved):
Thanks! :ship:

root/static/scripts/edit/URLCleanup.js
mwiencek: You can actually combine things into a single loop, but I don't know if it makes things much clearer; here's the result:

```diff
diff --git a/root/static/scripts/edit/URLCleanup.js b/root/static/scripts/edit/URLCleanup.js
index 49069c188f..e401c48c8f 100644
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -5911,34 +5911,24 @@ entitySpecificRules.release_group = function (url) {
  * 1, 2, 3, [1, 2], [1, 3], [2, 3], [1, 2, 3]
  */
 function anyCombinationOf(
-  entityType: RelatableEntityTypeT,
+  entityType: string,
   types: $ReadOnlyArray<string>,
 ): $ReadOnlyArray<EntityTypesMap> {
-  const validCombinations = Array(1 << types.length).fill().map(
-    (e1, i) => types.filter((e2, j) => i & 1 << j),
-  );
-
   const result = [];
-  for (const combination of validCombinations) {
-    let option;
+  const numCombinations = (1 << types.length) - 1;
+  for (let i = 1; i <= numCombinations; i++) {
+    const combination = types.filter((e2, j) => i & 1 << j);
     if (combination.length === 0) {
       // The empty subset is technically valid, but not of interest to us
       continue;
     }
     if (combination.length === 1) {
       // One-type subsets are expected as a string, not a 1-element array
-      option = combination[0];
+      result.push({[entityType]: combination[0]});
     } else {
-      option = combination;
+      result.push({[entityType]: combination});
     }
-
-    const resultOption: {
-      [entityType: RelatableEntityTypeT]: EntityTypes,
-    } = {};
-    resultOption[entityType] = option;
-    result.push(resultOption);
   }
-
   return result;
 }
```

reosarevok: @yvanzo: opinion? ^ I can commit this change if you like it more.

yvanzo: Check the timeline, I approved the pull request before @mwiencek posted his suggestion, just didn’t know he was still at this issue, I have no issue with it.

PR #2826, opened 2023-02-08
MBS-12870: Trap focus inside the Popover component

mwiencek:
# Fix MBS-12870

## Problem

Pressing tab on the last element inside a Popover or shift-tab on the first element will close the dialog.

## Solution

Sets the Dialog trapFocus prop to true in the Popover component so that tab/shift-tab cycles.

## Testing

Tested manually only.

Note that unlike the Modal component, clicking outside the Popover will not return focus to it, so you can still click on links etc. outside of it if needed.


reosarevok (approved):
Works nicely. This affects all popovers, but I checked and the other places where we use them (editing URLs, the guess case options) all have ways to get out of them by keyboard and in all cases this actually seems to be an improvement over the popover closing by accident.

PR #2827, opened 2023-02-08
MBS-12872: Autocomplete2: Don't focus the first item unless searching

mwiencek:
# Fix MBS-12872

## Problem

Trying to tab through the relationship dialog will automatically select the first item from every "Recent items" menu.

Relatedly, I found it annoying that if you get "(No results)," then you can't tab out of the menu when it's highlighted.

## Solution

Only highlight the first item if a search is performed.

Change "noop" actions like "(No results)" so that they close the menu instead.

## Testing

Tested manually, both with core entity lookups and static item lookups (link type, instrument, vocal).

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -551,11 +551,9 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
 
       case 'Enter':
       case 'Tab': {
-        if (isOpen) {
+        if (isOpen && highlightedItem) {
           event.preventDefault();
-          if (highlightedItem) {
-            selectItem(highlightedItem);
-          }
+          selectItem(highlightedItem);
         }
         break;
       }
@@ -822,6 +820,7 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
             ? 'visible'
             : 'hidden',
         }}
+        tabIndex="-1"
       >
         {disabled ? null : menuItemElements}
       </ul>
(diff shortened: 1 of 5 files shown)


root/static/scripts/common/components/Autocomplete2.js
```diff
@@ -822,6 +820,7 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
             ? 'visible'
             : 'hidden',
         }}
+        tabIndex="-1"
```
mwiencek, line 823: For some reason the other changes made it so that the `<ul>` itself was tabbable, but only for instruments and vocals (not for the relationship type or target entity). Not sure why, but it's easy enough to remove it from the tab index so that it never happens in either case.

mwiencek:
This might cause some Selenium tests to fail since they might rely on the old behavior of tab selecting the first item in all cases.  Will check those tomorrow.

reosarevok (approved):
A quick test seems to work nicely, so this seems fine to merge once the Selenium tests are fixed.

mwiencek:
@brainzbot test this please

mwiencek:
I ran all of
t/selenium/Release_Relationship_Editor.json5
t/selenium/Recording_Edit_Form.json5
t/selenium/Artist_Edit_Form.json5
t/selenium/Event_Edit_Form.json5
t/selenium/Genre_Edit_Form.json5
locally and they pass, so I *think* the tests are fine, actually.  This will be stuck in the Selenium build queue for another few hours, so I might just merge and re-check the tests on the beta branch later.

PR #2828, opened 2023-02-08
MBS-12888: Check ended when copying end date

reosarevok:
### Implement MBS-12888

# Problem
Currently, when using the copy date button to copy a date (for example in the alias editor), the "Ended" checkbox is not checked.

This is mostly just a visual issue, since having an end date implies ended and the checkbox is actually checked when removing the new date (this last bit might be unintentional though). In any case, it'd still a lot more clear if the (automatically disabled) ended checkbox is shown as checked rather than unchecked while frozen.

# Solution
In the reducer, I added the same check and dispatch call for `set-ended` on `case 'copy-date'` that we already had in the manual change dispatch, `case 'update-end-date'`.

I noticed while doing this that the check was broken, in that `isDateEmpty` would still claim the date is not empty for a case such as `{year: ''}`. I improved the checking in another commit, although this currently does not affect anything much since we don't actually *unset* ended if the end date gets blanked.

# Testing
Manually, in the alias editor, by adding a start date and copying to the end date and making sure the frozen checkbox is shown as checked.

# Related discussion
Should we actually unset the ended checkbox when the user empties the end date? I have seen a fair amount of cases where the end date is removed but ended is kept by accident, but I don't know how common the use case "remove the date because this is not actually ended" is vs "remove the date because it's unclear, but we know it is ended". I *think* I'd prefer to make the user actively check ended again if they want to mark it after emptying the field, but what do you feel?

### DIFF
--- a/root/static/scripts/common/utility/isDateEmpty.js
+++ b/root/static/scripts/common/utility/isDateEmpty.js
@@ -17,5 +17,5 @@ export default function isDateEmpty(
   date: ?PartialDateT | ?PartialDateStringsT,
 ): boolean %checks {
   return (date == null) ||
-    (date.year == null && date.month == null && date.day == null);
+         (empty(date.year) && empty(date.month) && empty(date.day));
 }
--- a/root/static/scripts/common/utility/isDateEmpty.js
+++ b/root/static/scripts/common/utility/isDateEmpty.js
@@ -7,12 +7,6 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
-declare type PartialDateStringsT = {
-  +day?: string,
-  +month?: string,
-  +year?: string,
-};
-
 export default function isDateEmpty(
   date: ?PartialDateT | ?PartialDateStringsT,
 ): boolean %checks {
--- a/root/static/scripts/edit/components/DateRangeFieldset.js
+++ b/root/static/scripts/edit/components/DateRangeFieldset.js
@@ -137,13 +137,21 @@ export function runReducer(
       const year = String(beginDateFields.year.value ?? '');
       const month = String(beginDateFields.month.value ?? '');
       const day = String(beginDateFields.day.value ?? '');
+      const newEndDate: PartialDateStringsT =
+        {day: day, month: month, year: year};
       runFormRowPartialDateReducer(
         subfields.end_date,
         {
-          date: {day: day, month: month, year: year},
+          date: newEndDate,
           type: 'set-date',
         },
       );
+      if (!isDateEmpty(newEndDate)) {
+        runReducer(
+          state,
+          {enabled: true, type: 'set-ended'},
+        );
+      }
       validateDatePeriod(state);
       applyAllPendingErrors(state);
       break;
--- a/root/types/entity.js
+++ b/root/types/entity.js
@@ -96,6 +96,12 @@ declare type PartialDateT = {
   +year: number | null,
 };
 
+declare type PartialDateStringsT = {
+  +day?: string,
+  +month?: string,
+  +year?: string,
+};
+
 declare type TypeRoleT<T> = {
   +typeID: number | null,
   +typeName?: string,


mwiencek (approved):
worksforme, thanks 👍

mwiencek:
> Should we actually unset the ended checkbox when the user empties the end date?

I'd probably unset it, since we can't be certain that leaving it checked is correct, but the user can always re-check it if needed.

reosarevok:
I think I agree with that, but given we don't do that right now elsewhere anyway, that seems like something we can make a ticket for and implement later on if @Aerozol agrees :)

Aerozol:
That’s a tricky one. It would be useful to unset it in a lot of situations, but I am wary of unsetting things that aren’t where the user has clicked. It wont be uncommon for the user to unset the end date but the artist has still ended.
If it’s ticked + greyed out when an end date is entered that might work (clearing the end date re-activates the field - I’m not sure whether it should also unset it)

reosarevok:
> If it’s ticked + greyed out when an end date is entered that might work (clearing the end date re-activates the field - I’m not sure whether it should also unset it)

That's what happens now :) At the moment clearing the date reactivates, and I'm also not sure whether it should unset it. I kinda feel that since it was often filled *because* there was an end date, it makes sense to unset it, but I think either way has chances of being confusing.

yvanzo:
@Aerozol wrote:
> The ideal solution for me would be that setting an end date greys out the ended checkbox, but doesn’t actually ‘check’ it. If it’s greyed out/an end date is set then it is checked when the edit is submitted. But until then unsetting the end date just ungreys the ended box - returning it to its previous status.
> Does that make sense?

It is starting to make more sense to me. I would add as a complementary suggestion to refine the label of the ended checkbox to be specifically about unknown end date. For example: “This relationship has ended at an unknown year.”

reosarevok:
>Hmm I didn’t see that this was for the alias editor specifically sorry 

It's not, specifically - it's for the React-based date code. So the alias editor, and in beta also the relationship editor, and eventually every place with a date field :) 

About the suggestions for the checkbox:
Right now "ended" just means "it has ended, whether we know when or not", which is why it's checked automatically when entering an end date. I guess the suggestion is that the database / API "ended" flag would still mean that, but the UI one would have its meaning changed to "I don't know the date but I know it's ended"? Isn't the clash a bit confusing? But maybe it's fine since we don't often call the checkbox just "ended" anyway.

If we go with that, the ended flag should probably be *hidden* entirely if an end date is set, rather than frozen/greyed out, since it would be entirely irrelevant to that use case?

yvanzo:
> but the UI one would have its meaning changed to "I don't know the date but I know it's ended"?
Yes, that is exactly what it is useful for, and the label change I suggested too.

> Isn't the clash a bit confusing?
Not really, it is the purpose of the UI to make it easier to edit.
As mentioned above, the code for the `ended` checkbox would then have to be renamed in the code.
It’s probably more useful for API/DB programming to keep having a simple “ended” column.
The two can coexist as long as it is properly named and documented.

> If we go with that, the ended flag should probably be _hidden_ entirely if an end date is set, rather than frozen/greyed out, since it would be entirely irrelevant to that use case?

It won’t be _irrelevant_ if it the end date was wrong and voluntarily erased. It can probably be compared with the checkbox for unknown medium format in the release editor, which is frozen/greyed out when a format is selected. I don’t know wheter it is relevant or not to follow the same behavior here to avoid potential visual disturbances. I would leave it to the expert.

reosarevok:
> It won’t be irrelevant if it the end date was wrong and voluntarily erased.

Sure, I thought that as soon as the date is erased, then the checkbox should appear again :) But maybe that's more confusing.

Aerozol:
What I’m proposing is:

1. No end date set: User can check or uncheck the ‘ended’ checkbox as wanted
2. The user sets an end date. The ended checkbox is now greyed out. Pressing ‘submit’ submits the checkbox as checked.
3. If the user unsets the end date (before pressing submit) the checkbox status returns to 1.

I think we’ve gone on different tangents here? Now I’m not sure anymore either :P

reosarevok:
What should happen if an end date was already set before, and the user deletes it?

yvanzo:
@Aerozol wrote:
>  *e.g. no new database stuff required

No “new databases stuff” is required either with the previous proposition that was to specify in the editing UI to check the box when it is known to be ended but the end year is unknown.

@reosarevok wrote:

> What […] if […]?

Exactly, the latest proposition by @Aerozol - which is theoretically fine - doesn’t cover this specific use case that you initially questioned in the section “Related discussion”. Now it really looks like “tangents” that go around in circles. :rofl: I encourage you to move it to community forums if you want a broader input from actual users since it is about UI/UX only.

Aerozol:
I’m still wary of unsetting stuff where the user hasn’t clicked.
But a few sleeps later, and taking into account that you have seen this happen a few times reosarevok, I think we can unset the end date if the user removes the release date (in a new editing session to the one where it was added, as opposed to my other discussion which was talking about the same editing session, sorry for derailing…)
Particularly if we consider an incorrect ‘ended’ status to be worse than an incorrect ‘still active’.

yvanzo:
@Aerozol wrote:
> taking into account that you have seen this happen a few times reosarevok

Having already seen @reosarevok saying “a bunch” for “two”, I would really like to know what “a few times” or more exactly “a fair amount” means in this case.

> as opposed to my other discussion which was talking about the same editing session, sorry for derailing…

It wasn’t completely off-topic as the UI has to keep handling both use cases correctly even if we change the behavior for one use case only.

> Particularly if we consider an incorrect ‘ended’ status to be worse than an incorrect ‘still active’.

That is probably true.

reosarevok:
> Having already seen @reosarevok saying “a bunch” for “two”, I would really like to know what “a few times” or more exactly “a fair amount” means in this case.

It's certainly happened more than twice just for me alone :) I haven't made a study, but I'd be surprised if I'm the only one who forgets it.

PR #2829, opened 2023-02-08
MBS-12879: Block Bandcamp /videoframe links 

reosarevok:
### Fix MBS-12879

# Problem
We got a report that we were mangling Bandcamp `/videoframe` links, since the cleanup was removing the video ID from them.

On researching it, yvanzo found that these seem to be intended as Bandcamp's inner working links, as they are, as far as we have found, meant to be shown in an `iframe` in `/track` pages (notice that the links are not even listed under the appropriate artist.bandcamp.com subpage).

# Solution
Since linking to the `/track` page seems a lot better, this blocks adding `/videoframe` pages and suggests using the `/track` link instead.

I shortcut the cleanup code when these links are found so that we don't run it and remove the query parameter needed for accessing it, since that's quite confusing. I considered cleaning them up to HTTPS but since we're blocking them anyway, I guess it doesn't matter :)

# Testing
Manually, plus added a test.

# Notes
On top of https://github.com/metabrainz/musicbrainz-server/pull/2825 since the implementation depends on us actually validating Bandcamp links for recordings and releases.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1120,12 +1120,34 @@ const CLEANUPS: CleanupEntries = {
       },
     ],
     clean: function (url) {
+      // To reject /videoframe links without breaking the ?video_id parameter
+      if (/^https?:\/\/bandcamp.com\/videoframe/.test(url)) {
+        return url;
+      }
       url = url.replace(/^(?:https?:\/\/)?([^\/]+\.)?bandcamp\.com(?:\/([^?#]*))?.*$/, 'https://$1bandcamp.com/$2');
       url = url.replace(/^https:\/\/([^\/]+)\.bandcamp\.com\/(?:((?:album|track)\/[^\/]+))?.*$/, 'https://$1.bandcamp.com/$2');
       url = url.replace(/^https:\/\/bandcamp\.com\/(?:discover|tag)\/([^\/]+).*$/, 'https://bandcamp.com/discover/$1');
       return url;
     },
     validate: function (url, id) {
+      if (/^https:\/\/(?:[^\/]+\.)?bandcamp\.com\/videoframe/.test(url)) {
+        return {
+          error: exp.l(
+            `Please do not add “{blocked_url_pattern}” links,
+             link to the appropriate “{wanted_url_pattern}” page instead.`,
+            {
+              blocked_url_pattern: (
+                <span className="url-quote">{'/videoframe'}</span>
+              ),
+              wanted_url_pattern: (
+                <span className="url-quote">{'/track'}</span>
+              ),
+            },
+          ),
+          result: false,
+          target: ERROR_TARGETS.URL,
+        };
+      }
       switch (id) {
         case LINK_TYPES.bandcamp.artist:
           if (/^https:\/\/[^\/]+\.bandcamp\.com\/(album|track)/.test(url)) {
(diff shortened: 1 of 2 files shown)


yvanzo (commented):
Is the first commit about SoundCloud an oversight?

There also is an unrelated change to WD/WP comment.

Last, I cannot understand/remember how `/campaign` & `/merch` are supposed to work. I just tested adding a `/campaign` link to a release in production and it doesn’t seem to use the `bandcampcampaign` rule: The error message seems to come from the `bandcamp` rule. Or am I missing something?

reosarevok:
Sorry, this should probably be marked as a draft. See https://github.com/metabrainz/musicbrainz-server/pull/2825 which this is on top of (and explains the campaign and merch bits further) first, and explains the SoundCloud bits.

reosarevok:
The stuff this was based on was merged a while ago but I never took this out of draft. Rebased and did that now.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1120,11 +1120,25 @@ const CLEANUPS: CleanupEntries = {
...
       url = url.replace(/^https:\/\/([^\/]+)\.bandcamp\.com\/(?:((?:album|track)\/[^\/]+))?.*$/, 'https://$1.bandcamp.com/$2');
       return url;
     },
     validate: function (url, id) {
+      if (/^https:\/\/(?:[^\/]+\.)?bandcamp\.com\/videoframe/.test(url)) {
+        return {
+          error: l(
+            `Please do not link to /videoframe links directly,
+             link to the appropriate track page instead.`,
+          ),
```
yvanzo, line 1137: I was trying to improve the formatting, and ended up rephrasing it in a generic way too:

```suggestion
          error: exp.l(
            `Please do not add “{blocked_url_pattern}” links,
             link to the appropriate “{wanted_url_pattern}” page instead.`,
            {
              blocked_url_pattern: (
                <span className="url-quote">{'/videoframe'}</span>
              ),
              wanted_url_pattern: (
                <span className="url-quote">{'/track'}</span>
              ),
            },
          ),
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -917,6 +917,17 @@ limited_link_type_combinations: [
...
        only_valid_entity_types: ['release'],
   },
+  {
+                     input_url: 'https://bandcamp.com/videoframe?video_id=188124/',
+             input_entity_type: 'recording',
+    expected_relationship_type: undefined,
+       input_relationship_type: 'streamingfree',
+       only_valid_entity_types: [],
+                expected_error: {
+                                  error: 'do not link to /videoframe links',
```
yvanzo, line 927: To match the suggestion in the previous comment:

```suggestion
                                  error: 'Please do not add',
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -917,6 +917,17 @@ limited_link_type_combinations: [
...
+                                  target: 'url',
```
yvanzo, line 928: Definitely the appropriate target here! :dart:

yvanzo (commented):
LGTMBDNT. Just suggesting (untested) changes to the source string.

root/static/scripts/edit/URLCleanup.js
reosarevok: This seems good - we might want to reuse it for some other strings later but we can start with this.

yvanzo (approved):
:ship:

PR #2830, opened 2023-02-09
Relationship editor required field style redux

mwiencek:
# Problem

In https://community.metabrainz.org/t/help-test-the-relationship-editor-on-beta-musicbrainz-org/621824 the discussion about how required fields should be displayed seemed to settle around showing both a red highlight (already done) and an asterisk (implemented in this commit, for color blind users if nothing else).

# Solution

Before:
Screen Shot 2023-02-08 at 8 36 41 PM

After:
Screen Shot 2023-02-08 at 8 36 58 PM

## Testing

I checked the display in Firefox, Safari, and Chromium on macOS, and it seemed fine in all of them.

### DIFF
--- a/root/static/scripts/relationship-editor/components/DialogPreview.js
+++ b/root/static/scripts/relationship-editor/components/DialogPreview.js
@@ -219,7 +219,7 @@ const DialogPreview = (React.memo<PropsT>(({
       ) : newRelationship ? (
         relationshipPreview(newRelationship, 'add-relationship')
       ) : (
-        <p>
+        <p className="required-fields-note">
           {l('Please fill out all required fields.')}
         </p>
       )}
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -744,6 +744,7 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
         aria-expanded={isOpen ? 'true' : 'false'}
         aria-haspopup="listbox"
         aria-owns={menuId}
+        className={state.required ? 'required' : undefined}
         role="combobox"
       >
         <input
@@ -751,6 +752,7 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
           aria-autocomplete="list"
           aria-controls={menuId}
           aria-labelledby={labelId}
+          aria-required={state.required ? 'true' : 'false'}
           autoComplete="off"
           className={
             (
--- a/root/static/styles/forms.less
+++ b/root/static/styles/forms.less
@@ -217,6 +217,18 @@ div.required {
     font-weight: bold;
 }
 
+/* Used to indicate required fields. */
+.asterisk {
+    content: "*";
+    left: -0.75em;
+    top: -0.25em;
+}
+
+div.required::before {
+    position: absolute;
+    .asterisk();
+}
+
 textarea {
     font-family: sans-serif;
     font-size: @medium-text;
--- a/root/static/styles/relationship-editor.less
+++ b/root/static/styles/relationship-editor.less
@@ -65,7 +65,7 @@ div.relationship-dialog, div.work-dialog {
             vertical-align: top;
             text-align: right;
             width: 25%;
-            padding-right: 0.5em;
+            padding-right: 0.75em;
             &.required {font-weight: bold;}
         }
         td.fields {
@@ -128,6 +128,7 @@ div.relationship-dialog {
     input[type="checkbox"] {
         vertical-align: middle;
     }
+    p.required-fields-note::before { .asterisk(); }
     div.buttons {
         width: 100%;
     }


reosarevok:
> (Note, the asterisks to the left of the fields go away once you fill them out.)

Why? It seems useful to have a clearer marking of requiredness than bold, which isn't very obvious at all. I could see them being red at first and turning black when filled, but honestly I'd probably just leave them be there in black all the time (and maybe even get rid of the bolding).

root/static/styles/widgets.less
```diff
@@ -303,6 +303,11 @@ div.autocomplete2 {
         display: flex;
         flex-direction: row;
 
+        &.required::before {
+            position: absolute;
+            .asterisk();
+        }
+
```
reosarevok, line 310: We're probably going to want this for *every* required field, not just autocomplete ones, so I would move this to a more generic place (`forms.less`?)

We don't have to implement it everywhere *yet* I guess, but it wouldn't hurt either if trivial. I'd possibly just put it next to 

```
label.required,
div.required {
    font-weight: bold;
}
```
in `forms.less` already, if that doesn't break anything.

root/static/scripts/relationship-editor/components/DialogPreview.js
```diff
@@ -219,7 +219,7 @@ const DialogPreview = (React.memo<PropsT>(({
       ) : newRelationship ? (
         relationshipPreview(newRelationship, 'add-relationship')
       ) : (
-        <p>
+        <p className="required-fields-note">
           {l('Please fill out all required fields.')}
         </p>
```
reosarevok, line 224: I think this'd be enough:

```suggestion
        <p>
          {l('Please fill out all required (*) fields.')}
        </p>
```

mwiencek:
> Why? It seems useful to have a clearer marking of requiredness than bold, which isn't very obvious at all. I could see them being red at first and turning black when filled, but honestly I'd probably just leave them be there in black all the time (and maybe even get rid of the bolding).

Hmm, I figured for the same reason the field doesn't stay red after you fill it in: we no longer need to draw attention to it.  It felt odd to show them when editing a relationship that had the fields filled already, at least, but maybe that's just me.  Getting rid of the bold if we always show the asterisks makes sense to me, though.

root/static/scripts/relationship-editor/components/DialogPreview.js
mwiencek: A doubt I have with that is that the \*s next to the fields are displayed with CSS (in order to position them more easily), so won't be announced by screen readers.  If we were to add `aria-required` on the inputs, which I can do but haven't done yet, then we would also want the \*s to remain visual-only so that we don't announce two different "required" indicators to the user.

Including the \* in this message where it will be announced might be confusing, then, because the user won't be able to find which fields have a \* next to them.

reosarevok: Hmm, ok, I see that point, although I feel the screen reader user might be able to figure out that the aria required is the equivalent of that?

My worry here is that the asterisk *before* the error message is kinda confusing, but maybe it's just me.

reosarevok:
> I figured for the same reason the field doesn't stay red after you fill it in: we no longer need to draw attention to it.

But we always indicate what fields are required with the bold font, so isn't this basically the same thing? Obviously we no longer need the strong red highlight, which suggests something is wrong, but having a clear marking of a required field seems useful all the time.

root/static/scripts/relationship-editor/components/DialogPreview.js
mwiencek: I was copying what https://tickets.metabrainz.org/browse/MBS-11847?focusedCommentId=66587&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-66587 used, which I see as being styled like a footnote (where the asterisk would normally come first).

mwiencek:
> But we always indicate what fields are required with the bold font, so isn't this basically the same thing?

Well, the bold is a lot less conspicuous to me (it just looks like a design choice, whereas the \* draws my attention).

> Obviously we no longer need the strong red highlight, which suggests something is wrong, but having a clear marking of a required field seems useful all the time.

It's definitely annoying when a form doesn't tell you what fields are required before you submit it, but I don't see that as being the case here, since we give four visual indicators when a required field is empty (asterisk, red highlight, preview message, disabled submit button).

However I'm not sure what the best practice is usability-wise, since I mostly see the \* used on static forms. @Aerozol would you prefer we keep the asterisk visible at all times?

Aerozol:
@mwiencek I think having it static/unchanging may be less distracting than the little ‘*’ appearing and disappearing when you do stuff.

mwiencek:
Updated this to leave the \* visible at all times and to add the `aria-required` attribute.

I left the labels bold for now, since we still do that on other forms; and to be honest when I tried changing it, the dialog looked oddly barren and maybe slightly less readable, so I wasn't sure about it anymore.

reosarevok:

reosarevok (approved):
This looks good to me - my only doubt is whether the asterisk should also be shown then by the "Related type" header, since that's technically also required (although I guess since it can't be unset, it will always be filled). I'm fine with merging for now and expanding it later to other required fields around MB anyway.

PR #2832, opened 2023-02-09
MBS-12641: Relationship dialog initial focus improvements

mwiencek:
# Fix MBS-12641

## Problem

When opening the relationship editor dialog (whether for a new or existing relationship), we sometimes focus fields that are pre-filled and unlikely to be changed.  (See MBS-12641 and also [salo.rock's comment on the forums](https://community.metabrainz.org/t/help-test-the-relationship-editor-on-beta-musicbrainz-org/621824/34).)

## Solution

We now choose which field to initially focus based on the following priorities:

 * If no target type is preselected, focus that.

   This is not the same as the field being empty, since you can't empty the target type field. It's about whether the target type is determined by the type of dialog being opened.

   I renamed `defaultTargetType` to `preselectedTargetType` in the code to make this clearer.

 * If no link type is preselected, focus that.

 * Otherwise, focus the target entity.

## Testing

Tested manually with each of the various types of "add" and "edit" buttons, and added a Selenium test which does the same (including for a seeded relationship).

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -332,6 +332,8 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
   const inputTimeout = React.useRef<TimeoutID | null>(null);
   const containerRef = React.useRef<HTMLDivElement | null>(null);
   const shouldUpdateScrollPositionRef = React.useRef<boolean>(false);
+  const recentItemsPromise =
+    React.useRef<Promise<$ReadOnlyArray<OptionItemT<T>>> | null>(null);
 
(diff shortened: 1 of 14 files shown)


reosarevok:
A quick test seems to work - do check the Selenium issues though. I found easy ways to test focus on the target type selector ("Add new relationship") and the target entity selector (+ button by an existing rel). What'd be a place where the focus would go to the relationship type selector?

reosarevok:
I'm assuming you tried the seeding and it worked? :)

mwiencek:
> I'm assuming you tried the seeding and it worked? :)

Yes, and there's also a Selenium test for that case.

reosarevok (approved):
Once tests are fixed this seems good to me.

PR #2833, opened 2023-02-09
MBS-12853: Disallow IPI that is all zeros

reosarevok:
### Implement MBS-12853

# Problem
A user recently entered an all-zeros IPI in [edit #96404467](https://musicbrainz.org/edit/96404467), possibly in order to fill the field with *something*. I'm pretty sure it is not a valid IPI, so we should probably just block it in the future

# Solution
Add a check to `is_valid_ipi` to return false if the IPI is all zeros. Since `format_ipi` specifically zero-fills the IPI to 11 digits, we can be sure it should always be 11.

# Testing
Added a new check to the `is_valid_ipi` test.

### DIFF
--- a/lib/MusicBrainz/Server/Validation.pm
+++ b/lib/MusicBrainz/Server/Validation.pm
@@ -146,6 +146,7 @@ sub format_iswc
 sub is_valid_ipi
 {
     my $ipi = shift;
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
LGTMBDNT :shipit:

PR #2834, opened 2023-02-09
MBS-12890: Show disambiguation in label/genre autocompletes

reosarevok:
### Fix MBS-12890

# Problem
Labels and genres with disambiguation comments do not show it when searching in an autocomplete2 inline search (for example to add relationships).

# Solution
Running the labels and genres through `formatGeneric` will give them the bare minimum that they should have: disambiguation and primary alias when relevant. We could also add types to labels (and artists) but that's not a regression so it can be done later.

# Testing
Manually, with **PMDC** for label search (easy from a release) and **bolero** for genre search (easy from an instrument).


yvanzo (approved):
Good catch! :100: :ship:

PR #2835, opened 2023-02-10
MBS-12852 / MBS-12893: Add events and places to country stats page

reosarevok:
### Implement MBS-12852 / MBS-12893

# Problem
While events and places are very directly associated to the areas they happen in, our country stats page at [`/statistics/countries`](https://musicbrainz.org/statistics/countries) has no country level stats for them. This seems like it would be interesting data to see.

While testing the code, I found a bug with release counts for countries with 0 releases, where the country would be skipped in the results rather than return 0 like for other entity counts. This is probably only an issue with sample data, since I'm pretty sure the production database has releases for all countries.

# Solution
I added both of the entity types to the stats. Place is a trivial implementation since they work in the same way as artists and labels. For events I copied the relevant parts of the query we use to get events with `find_by_area`.

For the found issue with release counts, I just changed a `LEFT JOIN` to a `FULL OUTER JOIN` like on other similar queries to ensure all countries will get a result.

# Testing
Manually, with (fairly limited) sample data. I checked that the numbers matched the amount of entities under the appropriate area page for a few cases.

### DIFF
--- a/root/statistics/stats.js
+++ b/root/statistics/stats.js
@@ -433,6 +433,11 @@ const stats = {
     color: '#e8ab08',
     label: l('Events'),
   },
+  'count.event.country.null': {
+    category: 'event-countries',
+    color: '#ff0000',
+    label: l('Events with no country set'),
+  },
   'count.event.type.null': {
     category: 'event-types',
     color: '#ff0000',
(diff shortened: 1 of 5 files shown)


mwiencek (approved):

root/statistics/stats.js
```diff
@@ -543,6 +548,11 @@ const stats = {
     color: '#bc0a0a',
     label: l('Places'),
   },
+  'count.place.country.null': {
+    category: 'pñace-countries',
```
mwiencek, line 552: (don't forget to push the fix for this though)

PR #2837, opened 2023-02-13
MBS-12898: Autoselect Apple Music rel for artists/labels

reosarevok:
### Implement MBS-12898

# Problem
Apple Music artist/label pages work in a very similar way to Bandcamp artist/label pages, in that the music can be available for streaming, or download, or both, or neither, and the contents can often change so that the right relationship would also change fairly often. As with Bandcamp though, the most important thing to denote is that the artist/label has an Apple Music page at that URL, so a new Apple Music relationship type has been added for artists and labels mirroring the Bandcamp one.

The current autoselect code though still selects purchase for download / streaming page, meaning the new relationship types cannot yet be used.

# Solution
This follows the Bandcamp model of autoselect and autoselects the new Apple Music relationship types instead.

# Action
* [ ] Run MBBE-70 either just after this is in beta or just after this is in production to migrate existing links

# Documentation
* [ ] After beta release, set the Apple Music link on [Cruz Cafuné](https://musicbrainz.org/artist/34652ef4-3f17-4eed-ba15-237c49373329) to the right type and use it as an example for [the label reltype](https://musicbrainz.org/relationship/debf36e1-b0fa-4e6c-987e-4248bf050fd8).
* [ ] After beta release, set the Apple Music link on [XL Recordings](https://musicbrainz.org/label/14221f01-8939-4ea0-b8f1-b5a21beae80a) to the right type and use it as an example for [the artist reltype](https://musicbrainz.org/relationship/64785d6c-2eeb-4f86-9418-b6c2d6c53c13).

# Testing
Updated the URL handling tests accordingly.

# Notes
On top of https://github.com/metabrainz/musicbrainz-server/pull/2825 since it uses the multi-option code introduced there.


yvanzo (commented):
This is overruling MBS-11938 and the two meeting discussions about it. The stance of not taking a decision after the second meeting is a quite unique case of not closing this ticket as “Won’t Do” rather than acknowledging that it shouldn’t be done this way for the given cons.

reosarevok:
The reason to have its own rel for this at the artist level is not related to that but with the Bandcamp case - this can be sometimes stream, sometimes download and sometimes both and we should not expect users to keep those changes up to date at the artist level.

That ticket is why this was only added for artist/label and not for recording / release where we are instead going with a multiple select option (again like with Bandcamp) :)

reosarevok:
That ticket was about splitting things that belong under one specific rel type into further sub types for every site. This is specifically a site that does not work under the current structure in that way and needs a separate relationship. The relationship has also already existed for months, and its use is being blocked by this.

Of course moving everything under get the music makes a lot of sense to you since you seem to want to delete all per-site relationship types anyway, but I have seen nobody else ask for such a change and a separate rel for this was explicitly requested (STYLE-801). As such, I would be thankful if I was allowed to continue doing my job as style leader by implementing the second part of that ticket. Alternatively, if you would like to do the job of style leader, we can talk about it and see what Rob thinks about it - it's not something I love doing and if you're excited about taking these sorts of decisions I wouldn't mind you taking over it.

PR #2838, opened 2023-02-13
MBS-12899: Add TMDB to the otherdbs whitelist for artists

reosarevok:
### Implement MBS-12899

# Problem
Several editors asked to be able to link to TMDB / themoviedb.org. This is for film (no music videos) but it has a lot of artists who are also in MB and biographical data for them. so I guess we can link to it in order to have the ID mapping if nothing else.

# Solution
Added the site to the other DBs whitelist, including sidebar and whatnot. For now allowing linking only to artists since it seems like the only clear match.

# Testing
Added a test for cleanup and autoselect for artists, and a second test to check linking films to works is blocked.


yvanzo (approved):
Code looks good.
Successfully tested it by linking a random artist to Madonna’s TMDB page and having it displayed it in the sidebar.

:warning: Since it is another database, the whitelist documentation page needs to be updated!

reosarevok:
Thanks, I had forgotten about that page again... I wish these things were part of the code. https://wiki.musicbrainz.org/index.php?title=Other_Databases_Relationship_Type%2FWhitelist&type=revision&diff=76187&oldid=75810 (not yet transcluded)

PR #2839, opened 2023-02-13
MBS-12891 / MBS-7971: Show more info for inline artist/label search 

reosarevok:
### Implement

# Problem
Inline search for labels and artists often offers frustratingly little info, especially when compared with something like a work search. Often knowing the type or the dates for an artist / label will automatically make it obvious it's the right choice or, at the very least, disqualify it from the possibilities, meaning there's one less entry the editor needs to check more carefully - but neither of these are present, while they are for other kinds of entities.

# Solution
This adds both artist/label dates and type, when present, in a new row for Autocomplete2.

image_2023-02-13_200517643

If all we know is an artist/label is ended, this currently doesn't indicate anything. If you think it should, we can make the change!

# Testing
By hand, with both direct and indexed search, searching artists and labels and making sure they show the info as expected.

# Notes
On top of https://github.com/metabrainz/musicbrainz-server/pull/2834 since that also touches label autocompletes.

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2/formatters.js
+++ b/root/static/scripts/common/components/Autocomplete2/formatters.js
@@ -24,6 +24,7 @@ import bracketed, {bracketedText} from '../../utility/bracketed.js';
 import formatDate from '../../utility/formatDate.js';
 import formatDatePeriod from '../../utility/formatDatePeriod.js';
 import formatTrackLength from '../../utility/formatTrackLength.js';
+import isDateEmpty from '../../utility/isDateEmpty.js';
 import CountryAbbr from '../CountryAbbr.js';
 
 import type {
@@ -107,6 +108,8 @@ function formatGeneric(
 function formatArtist(artist: ArtistT) {
   const sortName = artist.sort_name;
   let extraInfo;
+  const secondInfoLine = [];
+
   if (
     sortName &&
     sortName !== artist.name &&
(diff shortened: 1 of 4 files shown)


root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -117,7 +119,23 @@ function formatArtist(artist: ArtistT) {
       info.unshift(sortName);
     };
   }
-  return formatGeneric(artist, extraInfo);
+
+  if (nonEmpty(artist.typeName)) {
+    secondInfoRow.push(lp_attributes(artist.typeName, 'artist_type'));
+  }
+
+  if (artist.begin_date || artist.end_date) {
```
mwiencek, line 127: I'd recommend `isDateEmpty` here in case we get a partial date object with all nulls somewhere

root/static/scripts/common/components/Autocomplete2/formatters.js
```diff
@@ -107,6 +107,8 @@ function formatGeneric(
 function formatArtist(artist: ArtistT) {
   const sortName = artist.sort_name;
   let extraInfo;
+  const secondInfoRow = [];
```
mwiencek, line 110: maybe s/row/line/ since the display function is called showExtraInfoLine?

mwiencek (approved):
Works great, left two small suggestions

PR #2840, opened 2023-02-14
MBS-12829: Use link type phrases in autocomplete input value

mwiencek:
# Fix MBS-12829

## Problem

When you select a link type in the relationship dialog, we set the autocomplete input value to the link type name rather than the phrases we showed in the menu. For example, if you select "recording engineer for / recording engineer," we set the input value to just "recording."

## Solution

When a selection is made, the input value is now the same "forward phrase / reverse phrase" text we showed in the menu.

I decided not to change the input value based on the direction of the relationship, because the forward/reverse link phrases aren't meant to be read in sentence order like the long link phrases are, so I don't think it makes the selection easier to understand. It's also harder to implement.

## Testing

Tested manually both by searching for a link type and selecting it from the results, and selecting a link type from the recent items list (which use different code paths to generate the item names). There is one other code path where we set the input value: when switching the target type and only one link type is available (e.g. clicking "Add relationship"  on a recording and changing the target type to "work"; the one available link type will be automatically selected). The input value correctly shows the phrases in this case, too.

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2/formatters.js
+++ b/root/static/scripts/common/components/Autocomplete2/formatters.js
@@ -315,14 +315,8 @@ function formatLinkAttributeType(
   );
 }
 
-function formatLinkType(
-  linkType: LinkTypeT,
-  showDescriptions: ?boolean,
-) {
-  const description = stripHtml(linkType.l_description);
-  const isGroupingType = empty(description);
-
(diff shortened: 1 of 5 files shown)


reosarevok:

reosarevok (approved):
Tested this locally, worked nicely, thanks!

PR #2841, opened 2023-02-14
MBS-12901: Show user website and bio on the admin user views

reosarevok:
### Implement MBS-12901

# Problem
It's trivial to find out (with the IP lookup) what users share an IP with a known spammer. But the information on that view is pretty useless to figure out if any of those accounts are themselves spammers. Compared with the (much more useful) beginner editor report, the main things missing there are the user bio and website.

# Solution
This adds the bio and website to the `UserList` component used by `IpLookup`, `EmailSearch` and `PrivilegeList`. While the info is probably less obviously useful in the last two, I don't think it ever hurts, and it can sometimes still help, so I am not restricting it to `IpLookup` only.

# Testing
Manually, from the `EmailSearch` component (since I don't have IPs stored locally) - since the code used is the same, it should work on the other two as well.

### DIFF
--- a/root/admin/components/UserList.js
+++ b/root/admin/components/UserList.js
@@ -28,9 +28,11 @@ const UserList = ({users}: Props): React.Element<'table'> => {
         <tr>
           <th>{l('Editor')}</th>
           <th>{l('Member since')}</th>
+          <th>{l('Website')}</th>
           <th>{l('Email')}</th>
           <th>{l('Verified on')}</th>
           <th>{l('Last login')}</th>
+          <th>{l('Bio')}</th>
         </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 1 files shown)


root/admin/components/UserList.js
```diff
@@ -65,6 +70,9 @@ const UserList = ({users}: Props): React.Element<'table'> => {
                 formatUserDate($c, user.last_login_date)
               ) : null}
             </td>
+            <td>
+              {nonEmpty(user.biography) ? user.biography : null}
```
mwiencek, line 74: this will show raw HTML but I think that's fine on admin pages

mwiencek (approved):
LGTM

BTW, I think `PrivilegeList` in the description should be `PrivilegeSearch`.

root/admin/components/UserList.js
reosarevok: Yeah, that's what we show on the report :)

PR #2842, opened 2023-02-14
MBS-12873, MBS-12875: Avoid unloaded tracks when batch editing release relationships

yvanzo:
# Problems MBS-12873 & MBS-12875

Since the new React implementation of the release relationship editor avoid loading all the tracks for the big releases by default, the batch tools for changing credits in relationships (MBS-12873) and for linking to recordings and works (MBS-12875) don’t work nicely as described and as it used to be in the previous implementation of release relationship editor.

# Solution

Disable every checkbox related to these tools until all the related tracks have been loaded. Accompany such disabled checkbox with an help message.

It just uses a computation of the loading status already used for the release overview page.

The release relationship editor is expected to re-render only one more time after having loaded all the missing tracks.

The longer-term solution is to load all tracks when ticking such checkbox instead of requiring the user to chase the mediums with missing tracks.

# Testing

* [x] Updated existing tests without testing the new behavior.
  _I don’t think that it is needed to have automated tests at this point since it is not a definitive behavior._
* [x] Tested MBS-12873 manually using sample data with:
  * two releases:
    * `/release/32a6fcc2-bd3c-467b-956c-d207fe5bc367/edit-relationships` which has one partially loaded medium,
    * `/release/b4d66536-cbf1-4540-81cf-60f3b64ba78e/edit-relationships` which has a last completely unloaded medium;
  * to verify:
    * batch-adding a relationship to recordings,
    * batch-adding a relationship to works,
    * adding a relationship with artist credits to a recording,
    * adding a relationship with artist credits to a work,
    * adding a relationship with artist credits to the release,
    * adding a relationship with artist credits to the release group.
* [x] Tested MBS-12875 manually using the same releases to verify every selection checkbox before/after loading all tracks.

# Action

1. <del>Create a new (non-beta) ticket for the definitive solution.</del> See MBS-12907

### DIFF
--- a/root/static/scripts/release/components/ReleaseRelationshipEditor.js
+++ b/root/static/scripts/release/components/ReleaseRelationshipEditor.js
@@ -137,6 +137,7 @@ import {
   getMediumTracks,
   isMediumExpanded,
   runLazyReleaseReducer,
+  useReleaseHasUnloadedTracks,
   useUnloadedTracksMap,
 } from './TracklistAndCredits.js';
 
@@ -1387,6 +1388,8 @@ const MediumRelationshipEditors = React.memo(({
 }: MediumRelationshipEditorsPropsT) => {
   const hasUnloadedTracksPerMedium =
     useUnloadedTracksMap(release.mediums, loadedTracks);
+  const hasUnloadedTracks =
+    useReleaseHasUnloadedTracks(hasUnloadedTracksPerMedium);
   const mediumElements = [];
   for (const [medium, recordingStates] of tree.iterate(mediums)) {
     mediumElements.push(
@@ -1405,6 +1408,7 @@ const MediumRelationshipEditors = React.memo(({
         medium={medium}
         recordingStates={recordingStates}
         release={release}
+        releaseHasUnloadedTracks={hasUnloadedTracks}
         tracks={getMediumTracks(loadedTracks, medium)}
       />,
     );
@@ -1419,6 +1423,7 @@ type TrackRelationshipsSectionPropsT = {
   +loadedTracks: LoadedTracksMapT,
   +mediums: MediumStateTreeT,
   +release: ReleaseWithMediumsT,
+  +releaseHasUnloadedTracks: boolean,
   +selectedRecordings: tree.ImmutableTree<RecordingT> | null,
   +selectedWorks: tree.ImmutableTree<WorkT> | null,
 };
@@ -1430,6 +1435,7 @@ const TrackRelationshipsSection = React.memo(({
   loadedTracks,
   mediums,
   release,
+  releaseHasUnloadedTracks,
   selectedRecordings,
   selectedWorks,
 }: TrackRelationshipsSectionPropsT) => {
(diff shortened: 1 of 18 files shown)


mwiencek:
Thanks! This looks good on first glance. Passing the `releaseHasUnloadedTracks` prop through many layers does look a bit tedious, so might be a good situation to use [React context](https://beta.reactjs.org/learn/passing-data-deeply-with-context), similar to how we have `RelationshipSourceGroupsContextT`.  (This new context would only need to be provided in the `ReleaseRelationshipEditor` component.)

But if you prefer we can look into that idea later.

yvanzo:
I agree that a context could be used here. I mainly followed the code using `hasUnloadedTracks` for `MediumTable` here which was less tedious than getting into React Context at the same time.

We can mention using it for the definitive solution at the latest.

reosarevok (commented):
This seems good on a quick test - I would maybe consider adding an on-hover test to the disabled checkboxes saying "You must load this medium before selecting this" or something, to complement the message on top of the editor, but not a must either I don't think.

root/static/scripts/release/components/ReleaseRelationshipEditor.js
```diff
@@ -1481,8 +1487,20 @@ const TrackRelationshipsSection = React.memo(({
...
+            releaseHasUnloadedTracks={releaseHasUnloadedTracks}
             workSelectionCount={workCount}
           />
+          {releaseHasUnloadedTracks ? (
+            <div className="form-help">
+              <p>
+                {l(
+                  `Some tracks/mediums haven’t been loaded yet.
+                   If you want to make batch operations on all tracks,
+                   please fully load all mediums beforehand.`,
```
yvanzo, line 1499: Attempted improvements:
- Locate things involved in “batch operations”
- Loading is needed even to select “some mediums” not just “all tracks”

```suggestion
                  `Some tracks/mediums haven’t been loaded yet.
                   If you want to select some mediums below
                   in order to use with the batch-editing tools above,
                   please load all the tracks from these mediums beforehand.`,
```

reosarevok:
Oh, also, I'm an absolute idiot and/or blind, so I only now realized that the release relationship editor *already* has the expand/collapse all links. I wonder if "please fully load all mediums beforehand" in the new error could actually be a "link" that sends a dispatch in the same way "Expand all mediums" does, but since the two are so close to each other, it might be fine.

I still don't understand how I didn't notice the links until now!

yvanzo (commented):
I considered adding tool-tip conditionally but I didn’t find an easy way to do it. I guess that we need a generic component for that as it can be used in some other places such as the checkbox for unknown medium format in the release editor. Since it goes beyond this pull request and it is not a must, I just reported this improvement suggestion as MBS-12906.

I’m also making the help message more explanatory, see suggestion.

yvanzo:
To reply to your second comment: No because expanding all mediums won’t load the last tracks of the mediums that have more than 100 tracks.

Again, this pull request is just a temporary solution. I just reported the definitive solution I mentioned earlier as MBS-12907.

reosarevok (approved):
Tested it and it works fine.

PR #2843, opened 2023-02-14
MBS-11770: Always list autocomplete additional data for all entities

reosarevok:
### Fix MBS-11770

# Problem
Autocomplete2 loads useful additional data for works (related artists), recordings (ISRCs and RGs), releases (labels and release events) and events (related artists and places) when searching, but this data is missing from the recent entries when pasting an MBID and/or when reloading the page.

# Solution
This makes sure the data is also loaded on those two cases.

# Testing
By hand, with sample data, searching for "a" and picking some of the first results that have the extra data to be tested (since the data is always shown on first search, it's easy to find good testing cases).


mwiencek (approved):
Tested the event, recording, release, and work recent item menus and it works well.  Would be nice to have a centralized method for loading this information in all of the endpoints at some point.

PR #2844, opened 2023-02-14
MBS-12904: Fix another error submitting a recording artist relationship

mwiencek:
# Fix MBS-12904

## Problem

You can reproduce MBS-12904 as follows:

1. Open /recording/create and enter a title.
2. Search for a recording artist that isn't already in the recent entities list, and select it.
3. Add a relationship to the same artist you selected, choosing it from the recent entities list.
4. Enter an edit note and submit the form.

The recording will be submitted, but the relationship you added won't.

## Solution

This issue is very similar to [MBS-12694](https://tickets.metabrainz.org/browse/MBS-12694), but involves selecting the recording artist from a search first in order to push it onto the recent entities list.

The issue was that, when displaying search results, jQuery UI would modify the entity JSON to add undefined `label` and `value` properties as part of its "normalization" process. These undefined properties would cause `compactEntityJson` to throw an error.

I removed the normalization code from jQuery UI because its behavior is gross and we don't appear to need it anywhere. We override `_renderItem` everywhere, and don't require `label` or `value` to display anything that I can see.

## Testing

Added a Selenium test case which fails without the fix applied.

Manually tested other components using the old autocomplete code, to make sure they still function: the artist credit editor, the tag editor, and the "edit note author" field in the edit search all have working autocompletes.

### DIFF
--- a/root/static/lib/jquery.ui/ui/jquery-ui.custom.js
+++ b/root/static/lib/jquery.ui/ui/jquery-ui.custom.js
@@ -5055,9 +5055,6 @@ $.widget( "ui.autocomplete", {
 	},
 
 	__response: function( content ) {
-		if ( content ) {
-			content = this._normalize( content );
-		}
 		this._trigger( "response", null, { content: content } );
 		if ( !this.options.disabled && content && content.length && !this.cancelSearch ) {
 			this._suggest( content );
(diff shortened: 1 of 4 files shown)


root/static/lib/jquery.ui/ui/jquery-ui.custom.js
```diff
@@ -5055,9 +5055,6 @@ $.widget( "ui.autocomplete", {
 	},
 
 	__response: function( content ) {
-		if ( content ) {
```
mwiencek, line 5058: I don't mind making changes to this file because jQuery UI is deprecated in our codebase (new components just use React), so we don't plan on upgrading it.

mwiencek: We do the same with knockout already, for similar reasons: https://github.com/mwiencek/knockout/commits/mb (new uses of knockout are deprecated, and we are stuck on a commit in between releases and were never able to upgrade due to breaking changes, plus we have a couple custom fixes applied to the linked branch).

reosarevok (approved):
LGTMBDNT

PR #2845, opened 2023-02-15
MBS-12896: Show admins whether user is banned from profile

reosarevok:
### Implement MBS-12896

# Problem
Right now, visiting the user page for a spammer editor will clearly identify them as spammers, but someone who has just had their edit or edit notes privileges disabled, or is set as untrusted, is not visible to the admin until they go to the edit user page. 

# Solution
Added a second `Warning` for admins in the same place as the spammer one which specifies which restrictions a given user has (editing disabled, notes disabled, untrusted).

# Testing
Manually, by changing the privs of a random editor in the sample data.

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -19,7 +19,8 @@ import Warning from '../static/scripts/common/components/Warning.js';
 import {FLUENCY_NAMES} from '../static/scripts/common/constants.js';
 import * as TYPES from '../static/scripts/common/constants/editTypes.js';
 import {compare} from '../static/scripts/common/i18n.js';
-import commaList from '../static/scripts/common/i18n/commaList.js';
+import commaList, {commaListText}
+  from '../static/scripts/common/i18n/commaList.js';
 import commaOnlyList from '../static/scripts/common/i18n/commaOnlyList.js';
 import expand2react from '../static/scripts/common/i18n/expand2react.js';
 import bracketed, {bracketedText}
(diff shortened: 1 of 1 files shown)


root/user/UserProfile.js
```diff
@@ -855,6 +858,16 @@ const UserProfile = ({
...
+  const restrictions = [];
+  if (isEditingDisabled(user)) {
+    restrictions.push(l('Editing/voting disabled'));
+  }
+  if (isAddingNotesDisabled(user)) {
+    restrictions.push(l('Edit notes disabled'));
+  }
+  if (isUntrusted(user)) {
+    restrictions.push(l('Untrusted'));
+  }
```
mwiencek, line 870: I'd check `adminViewing` before pushing anything onto `restrictions` at all, since it isn't used otherwise and ensures we don't accidentally show those values anywhere we're not supposed to.
```suggestion
  if (adminViewing) {
    if (isEditingDisabled(user)) {
      restrictions.push(l('Editing/voting disabled'));
    }
    if (isAddingNotesDisabled(user)) {
      restrictions.push(l('Edit notes disabled'));
    }
    if (isUntrusted(user)) {
      restrictions.push(l('Untrusted'));
    }
  }
```

reosarevok: Commited the suggestion, we can squash on merge if it seems otherwise fine.

root/user/UserProfile.js
```diff
@@ -882,6 +897,20 @@ const UserProfile = ({
             />
           ) : null}
 
+          {restrictions.length && adminViewing ? (
+            <Warning
+              message={
+                texp.l(
+                  `This user’s editing rights have been restricted.
+                   Active restrictions: {restrictions}`,
```
mwiencek, line 905:
```suggestion
                   Active restrictions: {restrictions}.`,
```
:)

mwiencek (approved):
lgtmbdnt

PR #2846, opened 2023-02-15
MBS-12908: Do not reuse "cancelled"/"time" for artist-event rels

reosarevok:
### Implement MBS-12908

# Problem
When adding events and creating a new artist relationship based on an existing one that has the attributes "cancelled" and/or "time", those get prefilled for the new relationship. This is usually a desired behaviour, but it's unlikely that most artists will be performing at the same time and one being cancelled doesn't mean others are likely to be, so these two are likely to mostly get in the way.

# Solution
When cloning the relationship data, drop these two attributes.

# Testing
Manually, from `/event/create`.


reosarevok:

mwiencek:
@brainzbot retest this please

mwiencek (approved):

reosarevok:

yvanzo:
@brainzbot, retest this please, kisses

PR #2847, opened 2023-02-16
Fix MBS-11529 / MBS-12903: "Guess feat. artists" issues

mwiencek:
# Fix MBS-11529 / MBS-12903

## Problem

An exception is thrown on beta if you try to use the "Guess feat. artists" button outside of the release editor (MBS-12903).

Additionally, if we fix the exception then it has weird behavior (MBS-11529) where it removes the primary artist.

## Solution

The exception is fixed by making sure the `guessFeat` module is using the same source entity object as `initializeArtistCredit`.

The weird behavior is fixed by syncing the knockout and React state correctly.

## Testing

Tested manually on recording pages. A Selenium test has been added for MBS-11529 (which would have also failed due to MBS-12903 in either case).

Also quickly checked that the "guess feat." button still works in the release editor for the release title and track titles.

### DIFF
--- a/root/static/scripts/common/entity.js
+++ b/root/static/scripts/common/entity.js
@@ -10,6 +10,9 @@ import ko from 'knockout';
 import * as ReactDOMServer from 'react-dom/server';
 
 import formatLabelCode from '../../../utility/formatLabelCode.js';
+import getRelatedArtists from '../edit/utility/getRelatedArtists.js';
+import isEntityProbablyClassical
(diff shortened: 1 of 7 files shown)


mwiencek:
@brainzbot test this please

reosarevok (approved):
Tested in both the recording and release creation pages and both seem to work as expected. Hopefully we can soonish get run of the knockout code and make this less brittle too :)

PR #2848, opened 2023-02-16
MBS-11770 (erratum): Load work languages for recently used works

reosarevok:
### Fix MBS-11770 (redux)

# Problem
I missed work languages in https://github.com/metabrainz/musicbrainz-server/pull/2843. We show these for direct search results and when pasting an MBID, but we were not loading them for recent entries on reload.

# Solution
Also load work languages in the `/entities` endpoint.

# Testing
Manually by searching, selecting a work with a language, and then reloading to make sure the language still shows.


yvanzo (approved):
Both proofread and double-tested as release date is approaching. :speedboat:

reosarevok:
Hopefully this is the last bit we're missing, but if there's more it can always be merged for the next beta since a bit of data missing in recent entities has been an issue already in prod (by now, I think beta shows *more* data than prod).

PR #2850, opened 2023-02-17
MBS-12911, MBS-12928: Can't link two new works together

mwiencek:
# Fix MBS-12911 & MBS-12928

## Problem

See the description of MBS-12911.

## Solution

Newly-created works don't have gids, so we have to compare by id instead in order to avoid "entities in a relationship cannot be the same."

Fixing this also revealed another situation where the "two relationships with the same key" error can be triggered; in order to resolve that, we have to keep track of the new relationship states as they are updated in the update-submitted-relationships action, so that subsequent updates to the same relationship are operating on the latest version of its state.

## Testing

Added a Selenium test that links two newly-created works together with the "later version" AR, and checks that all edits are submitted successfully.

### DIFF
--- a/root/static/scripts/release/components/ReleaseRelationshipEditor.js
+++ b/root/static/scripts/release/components/ReleaseRelationshipEditor.js
@@ -1204,20 +1204,30 @@ const reducer = reducerWithErrorHandling<
         );
       }
 
+      const updatedRelationshipsByKey =
+        new Map<string, RelationshipStateT>();
+
       const updateRelationshipState = function (
         relationship: RelationshipStateT,
         callback: ({...RelationshipStateT}) => void,
       ) {
+        const currentRelationship = updatedRelationshipsByKey.get(
+          getRelationshipKey(relationship),
+        ) ?? relationship;
         const newRelationship =
-          cloneRelationshipState(relationship);
+          cloneRelationshipState(currentRelationship);
         newRelationship._lineage = [
           ...newRelationship._lineage,
           'submitted',
         ];
         callback(newRelationship);
+        updatedRelationshipsByKey.set(
+          getRelationshipKey(newRelationship),
+          newRelationship,
+        );
         updates.push(
           {
-            relationship,
+            relationship: currentRelationship,
             throwIfNotExists: false,
             type: REMOVE_RELATIONSHIP,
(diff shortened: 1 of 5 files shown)


root/static/scripts/release/components/ReleaseRelationshipEditor.js
```diff
@@ -1342,7 +1352,18 @@ const reducer = reducerWithErrorHandling<
...
+                      entity0.entityType === 'work' &&
+                      entity0.id === oldWork.id
+                    ) {
+                      newWorkRelationship.entity0 = newWork;
+                    } else if (
+                      entity1.entityType === 'work' &&
+                      entity1.id === oldWork.id
+                    ) {
+                      newWorkRelationship.entity1 = newWork;
+                    }
```
mwiencek, line 1366: Oh, another problem here was that this assumed the newly-created work was always `entity1` on the relationship, which is false.

reosarevok: I'm not sure I follow here. Above it has `const oldWork = relationship.entity1` but here you're comparing `entity0` and `entity1` with `oldWork`. Why is it safe to assume earlier that `oldWork` is always `entity1`, but not here?

reosarevok (changes requested):
Test is failing and it seems like a proper failure. Also the way we do this kinda confuses me, maybe you can explain :)

root/static/scripts/release/components/ReleaseRelationshipEditor.js
mwiencek: `relationship` above is always a recording-work relationship (passed from `getWorkEditsForEntity`). The code here is updating all relationships associated with the old work, not just recording-work ones. (I added some comments saying the same.)

mwiencek: But work-work relationships are the only real case where it can also be entity0, since work sorts last for every other entity-type pair of course.

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

reosarevok (approved):
Tested locally again just in case, seems to work, plus tests pass now.

PR #2851, opened 2023-02-18
MBS-12921: Can't search for works in the "Add related work" dialog

mwiencek:
# Fix MBS-12921

## Problem

Clicking "Add related work" and attempting to search for the pre-filled work name does not work in most cases.

## Solution

The Autocomplete2 component was attempting to show recent items instead of performing a search, because of an incorrect check where it assumed recent items would show even if the input was non-empty (they won't).

The reason the issue only manifests in the "Add related work" dialog is because we pre-fill the input value with the recording name there.

I've also changed the set-menu-visibility action to only open the menu if there are actually items; prior to this commit you could see an empty "menu" open (i.e. a gray line) beneath the input, and hitting arrow-down to move the item position would throw an exception because there are no items to highlight.

## Testing

Added a Selenium test which fails without the fix applied.

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -19,6 +19,7 @@ import useOutsideClickEffect from '../hooks/useOutsideClickEffect.js';
 import {unwrapNl} from '../i18n.js';
 import clean from '../utility/clean.js';
 import getCookie from '../utility/getCookie.js';
+import isBlank from '../utility/isBlank.js';
 
 import {
   CLOSE_ADD_ENTITY_DIALOG,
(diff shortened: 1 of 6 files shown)


mwiencek:
I've also included a fix for MBS-12922 here (with another Selenium test) since it doesn't work without the fix for MBS-12921.

reosarevok (changes requested):
The Selenium error seems to be because the test data is missing the `karaoke` attribute for the `performance` relationship for some reason? Not sure where that needs to be added.

The CircleCI error seems spurious (that test passed locally five times in a row) and in any case unrelated, but not sure why that would display (sort?) differently all of a sudden to cause a failure.

The code seems sensible, and testing locally works, so once the tests pass I think this is fine to merge.

mwiencek:
The Selenium error is likely because the performance link type was still cached from my sample database, which has the karaoke attribute. We may have to add the karaoke attribute later, but since it's not required now (and it's addition may cause other tests to fail), I just removed the karaoke bits from the link phrases in the edit data.

PR #2853, opened 2023-02-20
MBS-12924: Fix page freeze after selecting a link type on Android

mwiencek:
# Fix MBS-12924

## Problem

Selecting a link type in the relationship dialog on Android freezes the page.

## Solution

The `set-recent-items` dispatch causes an infinite loop of re-renders in Chrome on Android. Delaying the dispatch fixes it. I haven't determined why this doesn't occur on desktop, because the React internals are quite opaque to me.

The `cancelled` addition isn't strictly needed, but it's always a good idea to clean up your effects.

## Testing

Tested manually in the Android Emulator.

### DIFF
--- a/docker/musicbrainz-tests/run_selenium_tests.sh
+++ b/docker/musicbrainz-tests/run_selenium_tests.sh
@@ -51,7 +51,10 @@ sv start artwork-indexer artwork-redirect ssssss
 # Compile static resources.
 sudo -E -H -u musicbrainz yarn
 sudo -E -H -u musicbrainz make -C po all_quiet deploy
-NODE_ENV=test WEBPACK_MODE=development NO_PROGRESS=1 \
+NODE_ENV=test \
+     WEBPACK_MODE=development \
+     MUSICBRAINZ_RUNNING_TESTS=1 \
+     NO_PROGRESS=1 \
      sudo -E -H -u musicbrainz carton exec -- ./script/compile_resources.sh default tests
 
 # Add mbtest host alias to work around NO_PROXY restriction.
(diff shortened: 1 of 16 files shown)


reosarevok (approved):
Tested on test.mb on both Android and PC, works on both now AFAICT.

mwiencek:
@brainzbot test this please

reosarevok:

reosarevok (changes requested):
Selenium error is "Cannot read properties of undefined (reading 'setInputValueForReact')" so clearly that's connected to the latest changes. Is there a specific way we should be testing the bigger changes in the last commit, by the way?

mwiencek:
> Selenium error is "Cannot read properties of undefined (reading 'setInputValueForReact')" so clearly that's connected to the latest changes.

Should be a simple fix at least. I got so unlucky with the number of difficult and novel errors triggered by this PR!

> Is there a specific way we should be testing the bigger changes in the last commit, by the way?

That affects any case where the menu would be shown without performing a search. So:

 * Focusing the input when it's empty or when there was a previous search performed without an item being selected.
 * Pressing the down arrow.
 * Clicking the search icon.

reosarevok:

mwiencek:
With the latest commit ("Consistently serialize id as a number") the failing test passes for me locally three tries in a row. (Additionally, the release editor doesn't even use Autocomplete2 yet, so I'm not sure what's going on there.) I'll ask Jenkins again.

@brainzbot test this please

mwiencek:
Test are passing now, and I've tested this a bunch more locally, so I think it's fine to try it on beta. If it causes any new autocomplete regressions I'll add some tests for those.

PR #2856, opened 2023-02-21
MBS-12925: Allow submitting work dialogs with Enter

reosarevok:
### Fix MBS-12925

# Problem
Most pop-up dialogs in the release relationship editor can be submitted with Enter without issues, but that is not the case for the work dialogs (batch-create and edit), where pressing Enter does nothing.

# Solution
This follows the example of `RelationshipDialogContent` by submitting on pressing Enter unless it is pressed while on a button (to avoid an Enter press on "Cancel" submitting the form).

I turned the batch-create fragment into a `div` to support `onKeyDown`; I added the `"form"` class because it seemed sensible for parity.

This means Enter no longer opens the dropdowns for work types, but that's already the case for the target type dropdowns in the other dialogs, so it does not seem problematic to me (if anything, it's more consistent). Space still opens them fine.

# Testing
Tested manually, while pressing Enter on (AFAICT) every part of each form, and especially making sure pressing Enter on Cancel did not actually apply the changes.

### DIFF
--- a/root/static/scripts/release/components/BatchCreateWorksDialog.js
+++ b/root/static/scripts/release/components/BatchCreateWorksDialog.js
@@ -29,6 +29,8 @@ import DialogLinkType, {
   createInitialState as createDialogLinkTypeState,
   updateDialogState as updateDialogLinkTypeState,
 } from '../../relationship-editor/components/DialogLinkType.js';
+import useDialogEnterKeyHandler
+  from '../../relationship-editor/hooks/useDialogEnterKeyHandler.js';
 import type {
   BatchCreateWorksDialogStateT,
 } from '../../relationship-editor/types.js';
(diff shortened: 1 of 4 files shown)


root/static/scripts/release/components/BatchCreateWorksDialog.js
```diff
@@ -182,8 +182,22 @@ const BatchCreateWorksDialogContent = React.memo<
     sourceDispatch,
   ]);
 
+  const handleKeyDown = React.useCallback((event) => {
```
mwiencek, line 185: I think it'd make sense to add a custom hook for this, like `useDialogEnterKeyHandler` or something similarly named, which returns the `React.useCallback`.

Then you can just do `const handleKeyDown = useDialogEnterKeyHandler(acceptDialog);` in all three places.

mwiencek (approved):
Works for me

PR #2857, opened 2023-02-21
MBS-12932: Always show [No lyrics] in work languages

reosarevok:
### Fix MBS-12932

# Problem
Work language `[No lyrics]` shows correctly on the main list, but it regresses to `No linguistic content` for recent entities on reload.

# Solution
This uses `localizeLanguageName` when formatting recent items for `Autocomplete2`. Luckily, we know when something is set to the key `language-lyrics` so we don't need to work too hard to figure it out.

Using `localizeLanguageName` here also means languages in the dropdown are actually translated also when they're in recent entities after a reload, which AFAICT they were not before.

# Testing
Manually, by picking the `[No lyrics]` language on the batch-create works dialog, reloading, and making sure it still shows the right name (and also when using the French translation).

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2/recentItems.js
+++ b/root/static/scripts/common/components/Autocomplete2/recentItems.js
@@ -10,6 +10,7 @@
 import * as Sentry from '@sentry/browser';
 
 import {MAX_RECENT_ENTITIES} from '../../constants.js';
+import localizeLanguageName from '../../i18n/localizeLanguageName.js';
 import linkedEntities from '../../linkedEntities.mjs';
 import isDatabaseRowId from '../../utility/isDatabaseRowId.js';
 import isGuid from '../../utility/isGuid.js';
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/components/Autocomplete2/recentItems.js
```diff
@@ -179,6 +186,7 @@ export async function getOrFetchRecentItems<+T: EntityItemT>(
 ): Promise<$ReadOnlyArray<OptionItemT<T>>> {
   const ids = _getRecentEntityIds(key);
   const cachedList: Array<OptionItemT<T>> = [...getRecentItems<T>(key)];
+  const isLanguageForWorks = key === 'language-lyrics';
```
mwiencek, line 189: I was gonna suggest passing this to the other `getEntityName` call too, but I doubt there will ever be a case where the languages aren't loaded in `linkedEntities`.

Can we move this under `if (ids.size) {` then, since it's not used outside of that scope? (It's a tiny thing, but can be annoying when there are irrelevant variables in scope when using the step debugger.)

mwiencek (approved):
This works well and I don't see a better solution than checking the key.

PR #2858, opened 2023-02-21
MBS-12919: Show warning icon when rel has pending edits

reosarevok:
### Fix MBS-12919

# Problem
In the previous relationship editor, we used to show (previous) pending edits by highlighting the entity in orange, but that's now how changes made during the active editor session are shown so that doesn't work anymore. We should still have some way of showing the users that edits are pending for the rel, so they can check them before they accidentally delete it, for example (right now they will get notified of pending changes if they open the edit relationship popup, but not if they just enter a removal).

# Solution
I decided to show a small warning sign, with the same pending edits link on hover as we use once you in the relationship editing popup:

Peek 2023-02-21 17-48

# Testing
Manually, as per above example.

### DIFF
--- a/root/static/scripts/relationship-editor/components/RelationshipItem.js
+++ b/root/static/scripts/relationship-editor/components/RelationshipItem.js
@@ -10,6 +10,7 @@
 import * as React from 'react';
 import * as tree from 'weight-balanced-tree';
 
+import warningIconUrl from '../../../images/icons/warning.png';
 import ButtonPopover from '../../common/components/ButtonPopover.js';
 import DescriptiveLink from '../../common/components/DescriptiveLink.js';
 import {bracketedText} from '../../common/utility/bracketed.js';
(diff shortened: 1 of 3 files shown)


root/static/scripts/relationship-editor/components/RelationshipItem.js
```diff
@@ -272,6 +277,25 @@ const RelationshipItem = (React.memo<PropsT>(({
...
+          {hasPendingEdits && nonEmpty(openEditsLink) ? (
+            <>
+              {' '}
+              <Tooltip
+                content={exp.l(
+                  'This relationship has {edit_search|pending edits}.',
+                  {edit_search: openEditsLink},
+                )}
+                target={
+                  <img
```
mwiencek, line 289: The icon looks a lot better aligned for me with `vertical-align: middle` (in both Firefox and Chromium).

mwiencek (approved):
Works for me

PR #2860, opened 2023-02-21
MBS-12918: Reset credit string when updating target type

yvanzo:
# Problem MBS-12918

In the React relationship editor, when selecting a new type of target entity, the previous credit string was wrongly used for relationship preview in the case where the target entity does not support credit.

# Solution

Reset the credit string when updating target type.

# Testing

Tested manually by following the steps provided in the ticket.

Added a CI test on dialog state.

### DIFF
--- a/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
+++ b/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
@@ -419,6 +419,8 @@ export function reducer(
         });
       }
 
+      newTargetState.creditedAs = '';
+
       newState.targetEntity = newTargetState;
       break;
     }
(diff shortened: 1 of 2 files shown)


yvanzo:
@brainzbot, please retest this!

yvanzo:
@brainzbot, please retest this again

reosarevok (approved):
Tested locally, works nicely. Not entirely sure if having a test of `update-target-type` inside the `action: set-credit` test is clear? We could add a new set. But not the end of the world either :)

yvanzo:
Rebased on latest head master which passed Selenium tests successfully.

mwiencek (approved):
Yes, I think a separate `action: update-target-type` test makes sense, since the added test doesn't invoke the `set-credit` action.

mwiencek (approved):

PR #2862, opened 2023-02-22
MBS-12902: Add pagination to admin email/IP search

reosarevok:
### Implement MBS-12902

# Problem
In some cases, there's over 100 accounts using a dodgy IP or the same email pattern. In those cases, we want admins to be able to see (and possibly ban) them all, but the searches are currently limited to 100 results.

# Solution
This adds pagination to both email and IP lookup. The IP pagination is fairly trivial. The email one is more complicated, since the email address is posted when submitting the form. We need to store it somewhere (I put it in `$c->session` at the suggestion of @mwiencek) and use it when navigating through the pages, since otherwise the results will be lost as soon as we navigate to page 2.

# Testing
Email search tested manually with a specific user email (one result shows correctly, no pagination UI is shown) and with a wider search (`mailbox\.org`) with 10 pages of results (pagination works fine and the form contains the searched regex at all times).

IP lookup not tested because I don't have any IP data locally AFAICT.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Admin.pm
+++ b/lib/MusicBrainz/Server/Controller/Admin.pm
@@ -161,15 +161,23 @@ sub email_search : Path('/admin/email-search') Args(0) RequireAuth(account_admin
     my ($self, $c) = @_;
 
     my $form = $c->form(form => 'Admin::EmailSearch');
-    my @results;
-    my $searched = 0;
+    my $results;
+    my $stored_email = $c->session->{admin_searched_email};
 
     if ($c->form_posted_and_valid($form, $c->req->body_params)) {
         try {
-            @results = $c->model('Editor')->search_by_email(
-                $form->field('email')->value // '',
-            );
-            $searched = 1;
+            delete $c->session->{admin_searched_email};
+            my $searched_email = $form->field('email')->value;
+            $results = $self->_load_paged($c, sub {
+                $c->model('Editor')->search_by_email(
+                    $searched_email // '',
+                    shift,
+                    shift,
+                );
+            });
+            if ($searched_email) {
+                $c->session->{admin_searched_email} = $searched_email;
+            }
         } catch {
             my $error = $_;
             if ("$error" =~ m/invalid regular expression/) {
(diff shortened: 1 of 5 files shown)


reosarevok:

reosarevok:

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -161,15 +161,22 @@ sub email_search : Path('/admin/email-search') Args(0) RequireAuth(account_admin
     my ($self, $c) = @_;
 
     my $form = $c->form(form => 'Admin::EmailSearch');
-    my @results;
-    my $searched = 0;
+    my $results;
+    my $stored_email = $c->session->{searched_email};
 
     if ($c->form_posted_and_valid($form, $c->req->body_params)) {
         try {
```
mwiencek, line 168: Maybe we should `delete $c->session->{searched_email};` here so that if an error occurs it doesn't revert to the previously saved search when you reload the form?

lib/MusicBrainz/Server/Controller/Admin.pm
```diff
@@ -161,15 +161,22 @@ sub email_search : Path('/admin/email-search') Args(0) RequireAuth(account_admin
     my ($self, $c) = @_;
 
     my $form = $c->form(form => 'Admin::EmailSearch');
-    my @results;
-    my $searched = 0;
+    my $results;
+    my $stored_email = $c->session->{searched_email};
```
mwiencek, line 165: While there isn't really a chance of conflict here, making the key slightly more specific would be nice since these are global variables you can access from any action.
```suggestion
    my $stored_email = $c->session->{admin_searched_email};
```

mwiencek (approved):
Since these are admin-facing pages only, I think it's fine to test the IP changes on beta.

lib/MusicBrainz/Server/Controller/Admin.pm
reosarevok: I guess that would make sense, will do.

PR #2863, opened 2023-02-23
MBS-12622: Show if URL / URL relationship has pending edits in links editor

reosarevok:
### Implement MBS-12622

# Problem
We now show a small warning icon when a relationship has pending edits in the relationship editor (see https://github.com/metabrainz/musicbrainz-server/pull/2858), but nothing is shown in the link editor if a URL relationship has pending edits. We should probably show the same icon there (by each specific relationship under the link).

We don't show at the moment if the entity itself in the relationship has open edits, but we should probably still show something for URLs when the URL itself is being edited, at least, since that can sneakily make any relationships entered/changed wrong as soon as the edit applies.

# Solution
This implements a tooltip in the same way as https://github.com/metabrainz/musicbrainz-server/pull/2858. For URL relationships it literally reuses the code (I moved it to a reusable component). For pending URL edits it points to `/url/mbid/open_edits`, which is the best I could think of (I don't think we have a trivial way to see only the kinds of edits which trigger `url.editsPending`).

Screenshot from 2023-02-23 14-03-39

# Testing
Manually, by making open/votable changes to both a URL string *and* the dates on a URL relationship and making sure the warnings show (see image above) and the pending links point to the expected places.


reosarevok:

root/static/scripts/relationship-editor/utility/getOpenEditsLink.js
```diff
@@ -8,15 +8,28 @@
...
 
 import isDatabaseRowId from '../../common/utility/isDatabaseRowId.js';
+import type {
+  LinkRelationshipT,
+} from '../../edit/externalLinks.js';
 import type {
   RelationshipStateT,
 } from '../types.js';
 
+
```
mwiencek, line 18: is adding two newlines after the all the imports something you want to enforce with eslint? I don't mind either way but noticed you tend to add it

mwiencek (approved):
Tested locally, works great.

root/static/scripts/relationship-editor/utility/getOpenEditsLink.js
reosarevok: Huh. I'm not sure I even do it consciously - I don't know if it's better and I might even rather enforce just one and stop me from doing this :D

reosarevok: I removed the extra lines here - seems https://github.com/import-js/eslint-plugin-import/pull/1933 has been there for three years but is unmerged, so there's currently no `exact` option. Not sure if there's something we can do but I commented anyway.

PR #2864, opened 2023-02-23
MBS-12936: Don't update Autocomplete2 scroll position on item mouseover

mwiencek:
# Fix MBS-12936

## Problem

See the description of MBS-12936.

## Solution

In d4b15e9de756331755850ee6fb8479c0f42be14b I had intended to fix an issue where highlighting an item lower down in the list (thereby updating the scroll position), closing the menu (e.g. by hitting escape), and then re-opening the menu would not bring the scroll position back to the top of the list.  While I did fix that, there were some unintended effects when using a mouse: hovering over any item would cause the menu to scroll towards that item.

The actual intent is that we should only update the scroll position when using the arrow keys or when opening the menu. Obviously if a user can mouseover an item, then it's already in view, and we don't need to mess with their manual scrolling.

## Testing

Tested manually with a trackpad.  Added a Selenium test.

### DIFF
--- a/root/static/scripts/common/components/Autocomplete2.js
+++ b/root/static/scripts/common/components/Autocomplete2.js
@@ -337,7 +337,7 @@ const Autocomplete2 = (React.memo(<+T: EntityItemT>(
   const inputTimeout = React.useRef<TimeoutID | null>(null);
   const containerRef = React.useRef<HTMLDivElement | null>(null);
   const prevIsOpen = React.useRef<boolean>(false);
-  const prevHighlightedIndex = React.useRef<number>(-1);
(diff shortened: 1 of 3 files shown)


mwiencek:
This is deployed on http://test.musicbrainz.org/

t/selenium/Release_Relationship_Editor.json5 passes locally, trying tests again

@brainzbot test this please

yvanzo (approved):
Successfully tested with mouse wheel.

Successfully tested with arrow keys, but when the cursor is left over an item, then it can loop up and down in a limited range of items. I wouldn’t consider that as a blocking issue as moving the cursor out of the way is a straightforward workaround.

PR #2865, opened 2023-02-23
MBS-12937: Changing credits should work without modifying the source relationship

mwiencek:
# Fix MBS-12937

## Problem

Ticking "Change credits for other ... relationships on the page" and submitting the dialog has no effect unless you make other changes to the relationship.

## Solution

Fixes the check in `update-relationship-state` to proceed with updating entity credits even if the source relationship did not change.

## Testing

Tested manually and added a unit test.


yvanzo (approved):
Proofread the code changes.
Tested with the steps provided in ticket, using artist instead of instrument.
:ship:

PR #2866, opened 2023-02-23
MBS-12935: Readd type help in relationship editor

yvanzo:
# Problem MBS-12935

Since the autocomplete menu doesn’t show if there is a focused item, the relationship type’s description is no longer always visible.

Another issue was that the link to the full documentation page for the selected relationship type was no longer available.

# Solution

- Readd the same description box as before under the relationship type selector
- Readd the “(help)” toggle link but place it in the “Relationship” headline (instead of at the right of the selector as before) to follow the same style as for the “Attributes” headline
- As a bonus, add a link to the tree documentation page when no relationship type is selected.

# Testing

Tested manually with Artist->Work and Work->Recording.

No added CI test but it is likely needed.

# Draft progress

* [x] Check CI tests


root/static/scripts/relationship-editor/components/DialogLinkType.js
```diff
@@ -278,6 +281,12 @@ const DialogLinkType = (React.memo<PropsT>(({
     });
   }, [dispatch, source]);
 
+  const entityType0 = autocomplete.items[0].entity?.type0 ||
+    autocomplete.items[1].entity?.type0;
+  const entityType1 = autocomplete.items[0].entity?.type1 ||
+    autocomplete.items[1].entity?.type1;
+  const linkType = autocomplete.selectedItem?.entity;
```
mwiencek, line 288: This should be a more reliable way of getting type0/type1:
```suggestion
  const linkType = autocomplete.selectedItem?.entity;
  const entityType0 = linkType?.type0;
  const entityType1 = linkType?.type1;
```

yvanzo: Except that `entityType0` and `entityType1` are for use when `linkType` is not defined as no relationship type is selected, see the second commit’s message and bonus in pull request’s description.

mwiencek: In that case I think I'd pass `targetType` to `DialogLinkType` since we already have the source type too.

yvanzo: Is there another place where the two types are alphabetically ordered already? It seems to be in `LinkTypeAutocomplete`’s `id` (according to React Developer Tools) but I failed to trace it back in the code.

mwiencek: I think the link type ID comes from `getRelationshipKey`.  In a few other places I see we just sort the source/target types, like in `getDialogLinkTypeOptions`, `hasDialogLinkTypeOptions`, and `editorMayEditTypes` in exportTypeInfo.js

yvanzo: Thank you it is much cleaner now; See https://github.com/metabrainz/musicbrainz-server/pull/2866/commits/5cc69d12fc807389451d69d98bed8ae5979a440b?diff=split&w=1

mwiencek:
The Selenium failure looks unrelated, though the new browser console logs at least give a clue about what happened:
```
  [browser console log] [SEVERE] http://mbtest:5000/ws/js/artist?q=nine+inch+nails&page=1&direct=false - Failed to load resource: the server responded with a status of 500 (Internal Server Error)
```
Unfortunately we don't have access to the plackup logs to see what the ISE was. I pushed a commit here which will hopefully allow us to copy all of the runit service logs to the Jenkins build artifacts.

mwiencek:
(Marking as ready for review, since the most recent CI tests passed. Just running them once more to test the svlog build artifacts copying.)

mwiencek (commented):
The code looks good but I see one issue with the placement of the link: if you click the general "Add relationship" button (where focus is initially placed on the entity type), you now have to tab over the "help" link to get to the relationship type, where previously it was only a single tab press. This will probably annoy some people.

A quick fix would be to remove these links from the tab index (`tabIndex="-1"`) but we would lack an accessible way to click them without a mouse. Maybe a general "Show help" button at the bottom of the dialog that toggles everything would work.

yvanzo:
I followed your general idea and some @Aerozol’s input to replace these with a general “(help)” toggle link in the top right corner so that, using keyboard navigation, it isn’t required to skip it but it stays reachable.
See message to commit https://github.com/metabrainz/musicbrainz-server/pull/2866/commits/681e3d1a80042feaf630fc787fc650282de5587b and comments to the ticket MBS-12935 for details.

root/static/scripts/relationship-editor/components/DialogLinkType.js
```diff
@@ -291,6 +298,35 @@ const DialogLinkType = (React.memo<PropsT>(({
         <div aria-atomic="true" className="error" role="alert">
           {error}
         </div>
+        {isHelpVisible ? (
+          <div className="ar-descr">
+            {linkType === undefined ? (
+              <>
+                {exp.l(
+                  `{url|Documentation} about
+                     selectable relationship types.`,
```
reosarevok, line 307: Hmm. This is technically not always true (for example, if you open a recording series, and try to link it to a RG, we will not show "part of" as selectable, but it will obviously still be listed in the relationships link).

Also, showing "Documentation about selectable relationship types" as a help sentence seems a bit weird - if we keep something like this I might go with "See the {url|Documentation about selectable/possible/available relationship types}"

I'm actually thinking that, given this is shown if a type is not selected yet, maybe the help should be something like this?:
"Select the type of relationship between {edited_entity} and the target ({url|$type-$type relationship list}}"

But obviously not a blocking issue, so we could merge like this for now.

reosarevok (changes requested):
This clashes weirdly with the pending edits warning:

image

Could you take a quick look at that? (just make sure the rel has an open edit to test). Compare with master:

image

I left another comment, but as mentioned on the comment, I'm fine with merging this and improving wording later.

yvanzo:
Good catch! Fixed the clash with pending edits’ warning by setting `display` to `inline-block` for `dialog-titlebar` class; See [change](https://github.com/metabrainz/musicbrainz-server/compare/681e3d1a80042feaf630fc787fc650282de5587b..3e98142193e267e0fa893450fbe55592fe3579b0). Style taken from `buttons` class used at the bottom of the dialog box.

root/static/scripts/relationship-editor/components/DialogLinkType.js
yvanzo: That’s fair. I rephrased it as a gentle sentence, avoided using the part about “target” which is in code only, and linked using the same “(more documentation)” title as when a relationship type is selected; See [change](https://github.com/metabrainz/musicbrainz-server/compare/3e98142193e267e0fa893450fbe55592fe3579b0..c6b17ab4f9fd7961f7e3d106ea816330c80ec02d).

yvanzo:
Rebased on latest `master`’s `HEAD`.

reosarevok (approved):
Seems good now, thanks!

PR #2867, opened 2023-02-24
MBS-12931: Hide (+) button if no relationship type

yvanzo:
# Problem MBS-12931

When creating a new work for an artist, the relationships group editor is pre-filled with a relationship to the artist, but the relationship type is not selected. This is up to the user to select it. There is no point to allow adding a relationship of the same type if it isn’t defined yet, as there already is a button to add a new relationship of any type.

This issue has been uncovered by invariant error message in development mode.

# Solution

Just hide the (+) button as long as no relationship type is selected.

# Testing

Tried to reproduce it with the same steps as in the ticket.
Played with the relationship editor a bit.


reosarevok:

yvanzo:
That’s because the (+) is part of the right-aligned row header.
I don’t think it is a problem either.

mwiencek:
+1 though I'd prefer if we also looked into the underlying cause of the invariant error, since it could be indicative of other issues. But I can take a look at it later and open a separate PR if I find anything.

yvanzo:

PR #2871, opened 2023-02-28
ReleaseEventsDiff: Add missing nameVariation prop

mwiencek:
# ReleaseEventsDiff: Add missing nameVariation prop

## Problem

https://sentry.metabrainz.org/organizations/metabrainz/issues/1078098/?project=12

## Solution

Add the missing `nameVariation` prop to `ReleaseEventsDiff`. It's always `false` because the underlying content is `sideACountry.name` in either case.

I'm not 100% sure if this fixes every event logged to Sentry.

## Testing

Tested by loading /edit/97380335 with my local server connected to floyd.


reosarevok (approved):
LGTMBDNT

yvanzo (approved):
LGTMBDNT. Decided to release production without it as it isn’t much impacting users and had no beta test period.

PR #2872, opened 2023-02-28
MBS-11565: Use less hopeful image not available message

reosarevok:
### Implement MBS-11565

# Problem
Our current fallback message if a CAA image fails to load is "_Image not available yet, please try again in a few minutes._"

A lot of the times, the image is not actually going to appear in a few minutes, but having that in the text makes people think the longer delay is uncommon and needs reporting. Sadly, it isn't.

# Solution
This changes the "in a few minutes" wording to the more realistic "later". This will hopefully limit user's optimism and thus limit the amount of reports we have to answer with "shrug, it's the IA".

Since this message is also shown when images have been processed, but the IA is down, this also removes the "yet", which is weird when shown for an image that already worked before.

# Testing
None, since it's a string change.

### DIFF
--- a/root/components/Artwork.js
+++ b/root/components/Artwork.js
@@ -47,7 +47,7 @@ export const ArtworkImage = ({
       data-large-thumbnail={artwork.large_ia_thumbnail}
       data-message={nonEmpty(message)
         ? message
-        : l('Image not available yet, please try again in a few minutes.')}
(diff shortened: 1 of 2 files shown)


reosarevok:

mwiencek (approved):
Short of much longer help text or a wiki page we can link to that explains the situation better, this seems better than the current misleading message at least.

PR #2874, opened 2023-02-28
MBS-12939: Report for non-bootleg releases with bootleg label

reosarevok:
### Implement MBS-12939

# Reasoning
Bootleg labels pretty much only put out bootlegs. As such, this report lets editors find non-bootleg releases which have a bootleg label as a release label, which should almost always be wrong in some way.

# Testing
With sample data, which had a few matches already.


yvanzo (approved):
The report seems to be relevant, and you experienced it even with sample data. Ideally it should also be tested for performance against a full database but the query seems simple enough and limited enough to not impact much the running time of daily cron job.

LGTMBDNT :ship:

PR #2875, opened 2023-02-28
MBS-12946: Don't put Tooltip div inside span

reosarevok:
### Fix MBS-12946

# Problem
The pending edits tooltip on artist credits was showing twice and weirdly broken. On checking it a bit more, I found out it was a hydration error: `Expected server HTML to contain a matching <div> in <span>`. Apparently, `ArtistCreditLink` was trying to put the `Tooltip` `<div>` inside a `<span>`. This is invalid HTML, and React was not happy.

# Solution
I made `Tooltip` accept a `ContainerElement` prop like `CollapsibleList` does, where we can specify we want `Tooltip` to use `<span>` containers when we are using it in an inline situation.

While `externalLinks` (via `HelpIcon`) and `RelationshipItem` were not currently causing issues, both of them had `Tooltip` inside inline elements (`<label>` and `<span>` respectively) so I also changed it to use `<span>`s there.

The tooltip content needs to `display` as `block` when it's a `<span>` so that it's actually usable and readable, so added that to `tooltip.less`.

# Testing
I tested manually on the example given on the ticket, plus tested that the two other places I changed to use <span> still look fine (the links editor question marks, and the relationship editor pending edits warning signs).

### DIFF
--- a/root/static/styles/tooltip.less
+++ b/root/static/styles/tooltip.less
@@ -46,6 +46,7 @@
             background: @text-white;
             border: 1px solid @medium-border;
             box-shadow: 0 2px 5px @medium-border;
+            display: inline-block;
             font-size: @default-text-size;
             font-weight: normal;
             padding: 5px 10px;
--- a/root/static/scripts/edit/components/HelpIcon.js
+++ b/root/static/scripts/edit/components/HelpIcon.js
@@ -27,7 +27,7 @@ const HelpIcon = ({
   <Tooltip
     content={content}
     target={
-      <div
+      <span
         className="img icon help"
         style={ICON_STYLE}
       />
--- a/root/static/scripts/edit/components/Tooltip.js
+++ b/root/static/scripts/edit/components/Tooltip.js
@@ -22,8 +22,8 @@ type TooltipProps = {
 const Tooltip = ({
   content,
   target,
-}: TooltipProps): React.Element<'div'> => {
-  const containerRef = React.useRef<HTMLDivElement | null>(null);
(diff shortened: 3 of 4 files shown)


root/static/styles/tooltip.less
```diff
@@ -46,6 +46,7 @@
             background: @text-white;
             border: 1px solid @medium-border;
             box-shadow: 0 2px 5px @medium-border;
+            display: block;
```
mwiencek, line 49: Can we use `inline-block` instead, or does that not display correctly?  Using display: block inside an inline element still seems odd even if it doesn't break anything.

root/static/scripts/common/components/ArtistCreditLink.js
```diff
@@ -47,6 +47,7 @@ export const MpIcon = (hydrate<MpIconProps>('span.ac-mp', (
 
   return (
     <Tooltip
+      ContainerElement="span"
```
mwiencek, line 50: Is there any situation where we *need* divs, and spans with inline-block wouldn't work?

(unrelated, but I also find it strange to capitalize these variables, which don't refer to classes or components - they are just storing the string name of an element)

yvanzo (approved):
LGTMBDNT.

The implementation details don’t matter much to me as in the longer term we should consider using [`react-tooltip`](https://react-tooltip.com) instead, like BB and LB already do.

mwiencek (approved):
I checked the 5 places we use `<Tooltip />` (plus a few using `<HelpIcon />`) and didn't immediately notice any visual bugs.

PR #2876, opened 2023-03-01
MBS-12841 / MBS-8242: Link to edit and relationship docs from edits

reosarevok:
### Implement MBS-12841, MBS-8242

# Problem
Edit listings could always benefit from more ways to help voters make the right choice. Right now, there's two pieces of documentation we have and would almost certainly be good to show, but we are not linking in any way. One is the edit type documentation for the edit (the data there is not great, but it's being improved with https://github.com/metabrainz/musicbrainz-server/pull/2784 and should be improved further by actually documenting each edit type). The other is the documentation for relationship types involved in relationship edits (which usually have guidelines! which we don't have a way to access from the edit!).

# Solution
The second commit here adds a link to the doc page for the specific edit type in question to the header of every edit, be it on an edit list or on the edit page itself. This replaces the generic "Edit Types" link on the sidebar (not shown on edit lists) that I remove in the first commit.

The last commit adds a link to the doc page(s) for the relationship type(s) involved in a relationship edit (add, edit or remove). Rather than making it float by each relationship, I put it on the top right of the changes display, since @Aerozol suggested we should aim to have help boxes on the top right.

While testing that commit I found that types weren't being loaded properly for relationship removals - fixed with the third commit.

# Testing
Manually, using a bunch of relevant edits that happened to already be on my local sample data history from my previous tests for other PRs. A good testing suggestion would honestly just be to play with edit searches, but make sure you navigate to the edit pages as well to check the help works there too.


yvanzo:
Didn’t review yet, just noted that it now has conflicts.

yvanzo (changes requested):
Linking to the documentation is a must have, thank you for dealing with this.

At first glance, there are some issues with the current submission:

1. It doesn’t seem helpful to remove the link to the documentation for all edit types in the sidebar.
2. When an edit is replacing a relationship by another one of a different type, it is not clear which relationship type the help is about. There should probably be two help buttons in such case.
3. The help icon is too far away. It should probably just be by the text as in the external links editor.
4. The text in the tooltip should directly show the type’s description as in the external links editor.

Also do not hesitate to ask our design consultant some input on this.

yvanzo:
> since @Aerozol suggested we should aim to have help boxes on the top right.

Just reading this. I still think that icons are too far. @Aerozol what did you think about this issue?

yvanzo:
> 1. What is the use of linking to a generic list when we already link to the actual relevant type? (from where you can get to the generic list with one more click if needed)

Saving one click. The sidebar is not crowded either.

> 2. Did you test such edits? :)

Not this exact kind of edit, I trusted your screenshot.

reosarevok:
I just don't see when that main list is *ever* relevant to any specific edit, honestly :)

Aerozol:
Thanks both!
3. I agree that on its own the help icon looks far away from the text in question, but I think long-term it would be user-friendly to have all the help/doc materials on the right. There is room there, so we can be quite detailed with what we eventually add, and users will eventually know where to look. Here’s a professional mockup of what I mean:
image
3b. Perhaps custom styling the [?] box to fit in with the background a bit more, and maybe be a bit bigger, would help? So it’s not ‘floating’ on it’s own as much. Happy to deliver some designs.
4. I’m not sure I can see what’s in the tooltip currently - this isn’t on test or beta is it, for me to test?

yvanzo:
I’ve just put this branch on `test.mb.o`. Here are a few example pages to look into:
* [Search for “Edit relationship” edits](https://test.musicbrainz.org/search/edits?auto_edit_filter=&order=desc&negation=0&combinator=and&conditions.2.field=id&conditions.2.operator=BETWEEN&conditions.2.args.0=80427619&conditions.2.args.1=100279891&conditions.1.field=artist&conditions.1.operator=%3D&conditions.1.name=Ozzy+Osbourne&conditions.1.args.0=188&conditions.3.field=type&conditions.3.operator=%3D&conditions.3.args=91%2C234)
* [Edit changing the type of a relationship](https://test.musicbrainz.org/edit/100279891)
* [External links editor](https://test.musicbrainz.org/release-group/04773d95-c015-4ed3-a114-021391c6af7a/edit#external-links-editor-container) (the help button is not new to this pull request)

Note that there is no such help button in relationship editor at the moment.

Aerozol:
Thanks yvanzo!
I don’t mind it on the right, but this situation is definitely confusing (and a bit ugly):
image

Maybe it is worth spending longer on the design… something like:
image

4. yes, it would be to have: “See the documentation for this edit type: [Edit relationship]“, instead of just “See the documentation for this edit type”

reosarevok:
> it would be to have: “See the documentation for this edit type: [Edit relationship]“, instead of just “See the documentation for this edit type"

Hmm. I was only showing the types for relationships, because for edits there's only ever one type per edit and the type name is in large font on the heading. But if you feel it's useful for that, we could add it?

Aerozol:
Thanks reo!
re. 4, my gut feeling is that even with the text repeated, it is good practice to let someone see where a link is sending them. Maybe it will be annoying - I haven’t mocked this up and imagination is never the same as the real thing. So I’m leaning towards having the edit type in the rollover.

yvanzo:
You missed the point 4. What is in the tooltip for relationship type in the external links editor is the **description** of the relationship type, not just its name, so the user doesn’t even to click the link in most cases.

reosarevok:
I did miss that! I would like people to actually go and read the guidelines too, but it might be good to show more here - that just means we need to load it first though since I guess it's not loaded at the moment (unlike when editing).

reosarevok:
Changed this to actually show more info about the relationships:

example

It also now does "See the documentation for this edit type: [Edit relationship](https://musicbrainz.org/doc/Edit_Types/91)." on the edit type popup, as requested by @Aerozol.

The UI still seems kind of atrocious, though. @Aerozol, wanna play with it? :) (I understand you can now run code locally).

MonkeyDo:
I'm barging with no information and having not been part of the discussion, so sorry for that.

IMO the help icons should be as close as possible to what they apply to. In this case, that means next to "edit relationship" and "relationship" respectively, like so:
image

The icons floating to the right by themselves seem separated from what they describe, at least to my eyes.


I also asked in chat if the question mark icon was already a set style, because the color square it sits in makes it harder to blend into whatever component  it's on.
If it's possible to change the question mark icon I would suggest something like this: 
image

Aerozol:
Oof, sorry for forgetting this PR... thanks monkey! Now that the roll-over is much more useful I think having it next to the text makes sense.

FWIW my intention with having things on the right was to give us more real estate for the eventual dream of something like (also an ugly mockup, but hopefully you see what I mean):
image

My thinking was to eventually be able to have the same links for all elements, without making the UI go crazy, and maybe make them easily toggleable as well. But I agree that for now monkey's solution looks nice!

reosarevok:
Would it make sense to modernize the ugly question mark also elsewhere? If so, wanna choose and maybe give me a design? :D

Aerozol:
Yes to changing it everywhere!

Much as I would like to design something super flash, we should probably keep it subtle/simple - how about this, based on monkey's example: https://fontawesome.com/v5/icons/question-circle?f=classic&s=light

MonkeyDo:
Looks a little big perhaps, I would suggest 16px -> 14px, unless of course other icons of the sort on the website are 16px.

Would also suggest adding a margin-right (the same 3px as margin-left) for the relationship icon (the lower one):

BEFORE:
image

AFTER:
image

Aerozol:
I wonder if we are going to get some pushback re. the repetition of these - which might end up being solved by moving them all to the right again/into their own space. But I am keen to go ahead with this for now!

reosarevok:
`16px` is the default for icons, but we already override that for some that don't look that great at that size, so I can override it here too.

reosarevok:
Made the suggested changes, same link as before should work :)

PR #2877, opened 2023-03-02
MBS-12848: Reject inc=various-artists without inc=releases

reosarevok:
### Fix MBS-12848

# Problem
The `various-artists` include is used to request only releases where the artist is in the tracks but not the release (so, generally VA or featured releases). As such, it only makes sense if releases are being requested to begin with. Currently, `various-artists` was being accepted as an include even without the accompanying `releases` include, but was then silently ignored, which is not good UX.

# Solution
This treats `various-artists` as an `extra-inc` for `releases`, meaning we only accept passing it if it will actually do something (i.e., if `releases` is also passed), and we reject it with some more info otherwise.

# Testing
Locally by calling the API with/without `releases`, plus made sure the `LookupArtist` tests using `various-artists` still pass.

### DIFF
--- a/lib/MusicBrainz/Server/WebService/Validator.pm
+++ b/lib/MusicBrainz/Server/WebService/Validator.pm
@@ -53,7 +53,10 @@ our %relation_types = (
 our %extra_inc = (
     'recordings' => [ qw( artist-credits puids isrcs ) ],
     'recording-rels' => [ qw( artist-credits ) ],
-    'releases' => [ qw( artist-credits discids media type status ) ],
+    'releases' => [ qw(
+        artist-credits discids media
+        status type various-artists
+    ) ],
     'release-rels' => [ qw( artist-credits ) ],
     'release-groups' => [ qw( artist-credits type ) ],
     'release-group-rels' => [ qw( artist-credits ) ],
--- a/lib/MusicBrainz/Server/WebService/Validator.pm
+++ b/lib/MusicBrainz/Server/WebService/Validator.pm
@@ -51,16 +51,28 @@ our %relation_types = (
 # request for a recording or a request with inc=recordings.  This hash
 # helps validate the second case (inc=recordings).
 our %extra_inc = (
-    'recordings' => [ qw( artist-credits puids isrcs ) ],
-    'recording-rels' => [ qw( artist-credits ) ],
-    'releases' => [ qw(
-        artist-credits discids media
-        status type various-artists
-    ) ],
-    'release-rels' => [ qw( artist-credits ) ],
-    'release-groups' => [ qw( artist-credits type ) ],
-    'release-group-rels' => [ qw( artist-credits ) ],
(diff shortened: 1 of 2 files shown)


lib/MusicBrainz/Server/WebService/Validator.pm
```diff
@@ -207,7 +219,13 @@ sub validate_inc
     my %extra;
     for my $i (@inc)
     {
-        map { $extra{$_} = 1 } @{ $extra_inc{$i} } if (defined $extra_inc{$i});
+        if (defined $extra_inc{$i}) {
+            my @available_extra_incs = (
+                @{ $extra_inc{$i}{generic} // [] },
+                @{ $extra_inc{$i}{$resource} // [] },
+            );
+            map { $extra{$_} = 1 } @available_extra_incs;
```
mwiencek, line 227: Since you're already changing it,`for` is a lot clearer than using a `map` for side effects (TBH I thought you added a perlcritic rule for that already).

lib/MusicBrainz/Server/WebService/Validator.pm
```diff
@@ -224,8 +242,9 @@ sub validate_inc
         if (!exists $acc{$i} && !exists $extra{$i})
         {
             my @possible = grep {
-                my %all = map { $_ => 1 } @{ $extra_inc{$_} };
-                exists $all{$i}
+                my %all = map { $_ => 1 } @{ $extra_inc{$_}{generic} };
+                my %per_resource = map { $_ => 1 } @{ $extra_inc{$_}{$resource} };
```
mwiencek, line 246: Like above, I think you will need the `// []` here to avoid the dreaded "Can't use an undefined value as an ARRAY reference", unless I'm missing a reason that wouldn't be hit here.

mwiencek (commented):
Makes sense. This is technically a breaking change, but upon checking the logs for the past few days, it doesn't appear to be a common inc parameter. Additionally, I didn't see any requests specifying it that didn't also specify `inc=releases`.

reosarevok:
Maybe I'm wrong, but I'd also generally be of the opinion that breaking something that is already not working as the user probably expects might be better than leaving it be unchanged. But yeah, I doubted this would be much of an issue either way.

lib/MusicBrainz/Server/WebService/Validator.pm
reosarevok: I think you're thinking of https://github.com/metabrainz/musicbrainz-server/pull/2715

mwiencek:
> Maybe I'm wrong, but I'd also generally be of the opinion that breaking something that is already not working as the user probably expects might be better than leaving it be unchanged. But yeah, I doubted this would be much of an issue either way.

TBH it probably depends on how many requests it unintentionally breaks. Sometimes people include inc params they don't use/need without knowing what the correct usage is (maybe because our documentation wasn't clear to begin with).

mwiencek (approved):

PR #2878, opened 2023-03-03
MBS-12958: Block voting without a confirmed email

reosarevok:
### Fix MBS-12958

# Problem
For some reason, editors without a confirmed email cannot leave notes, but they can vote. I'm pretty sure this was an oversight rather than an intended ability, and it was made visible by us removing the beginner restrictions, which would have blocked most editors without an email from voting since it's hard to get 10 accepted edits when you cannot enter any.

# Solution
This specifically checks for `email_confirmation_date` when checking if a user may vote, both on the Perl level (actually blocking the votes being submitted) and on the JS level (hiding the interface to submit votes).

I'm also blocking the "My vote" info on edit list headers for editors who cannot vote, since it makes no sense for them. This is consistent with the `EditIndex` page, where we already hid it.

# Testing
Manually with a newly created account, plus I'm restoring the edit page test we had for beginners, but making it just for editors without a confirmed email instead.

### DIFF
--- a/lib/MusicBrainz/Server/Edit.pm
+++ b/lib/MusicBrainz/Server/Edit.pm
@@ -146,6 +146,7 @@ sub editor_may_vote {
     return (
         $self->is_open &&
         defined $editor &&
+        $editor->email_confirmation_date &&
         $editor->id != $self->editor_id &&
         !$editor->is_bot &&
         !$editor->is_editing_disabled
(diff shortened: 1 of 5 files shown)


yvanzo (approved):
LGTMBDNT. Just proofread the code and test, and thoroughly searched for similar oversights after changes in latest release. :hotsprings:  :adhesive_bandage: :rocket:

PR #2879, opened 2023-03-03
MBS-12955: Display beginner flag in vote tally 

reosarevok:
### Implement MBS-12955

# Problem
Since beginner users can now vote, it seems worth showing their beginner status also by their vote listing(s),in the same way we show them by their edit note(s). This makes it easier to understand a bad vote might be due to the user's inexperience. Also, a suspicious amount of beginner votes on an edit can be a mark of sockpuppetry.

# Solution
This just reuses the existing `EditorTypeInfo` component. This will also show if the voter is a bot (which is not allowed, so also worth knowing).

# Testing
Manually, with `/edit/97436260` in floyd data (which has a beginner editor and edit note author, making sure that nothing broke there, plus three beginner voters and at least one non-beginner one).

### DIFF
--- a/root/edit/EditIndex.js
+++ b/root/edit/EditIndex.js
@@ -24,6 +24,7 @@ import formatUserDate from '../utility/formatUserDate.js';
 
 import EditHeader from './components/EditHeader.js';
 import EditNotes from './components/EditNotes.js';
+import EditorTypeInfo from './components/EditorTypeInfo.js';
 import EditSidebar from './components/EditSidebar.js';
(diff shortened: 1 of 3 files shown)


yvanzo (approved):
LGTMBDNT. IIUC it will also show bot votes (which are not supposed to happen anyway). :ship:

reosarevok:
Yeah, it will :) If those happen, then we should be told, because we theoretically block bot votes at the Perl level :)

PR #2881, opened 2023-03-03
Upgrade Flow to 0.197.0

mwiencek:
First step to getting Flow up-to-date.

0.197.0 is the first version where `inference_mode=lti` is available in a non-experimental form, but I haven't enabled it in this PR because it'll require a lot more changes.

### DIFF
--- a/root/components/RelationshipsTable.js
+++ b/root/components/RelationshipsTable.js
@@ -296,7 +296,6 @@ const RelationshipsTable = ({
         ...PagedLinkTypeGroupT,
         +phrase: string,
       }>> = Object.values(targetTypeGroup)
-        // $FlowIgnore[incompatible-call]
         .map((group: PagedLinkTypeGroupT) => ({
           ...group,
           phrase: getLinkPhraseForGroup(group),
--- a/root/static/scripts/common/linkedEntities.mjs
+++ b/root/static/scripts/common/linkedEntities.mjs
@@ -144,7 +144,7 @@ export function mergeLinkedEntities(
       if (hasOwnProperty.call(linkedEntities, type)) {
         Object.assign(linkedEntities[type], entities);
       } else if (type in linkedEntities) {
-        // $FlowIgnore[prop-missing]
+        // $FlowIgnore[incompatible-type]
         linkedEntities[type] = entities;
       } else {
         throw new Error(
--- a/root/static/scripts/common/linkedEntities.mjs
+++ b/root/static/scripts/common/linkedEntities.mjs
@@ -95,7 +95,6 @@ export type LinkedEntitiesT = {
   work_type: {
     [workTypeId: string]: WorkTypeT,
   },
-  ...
 };
 
 // $FlowIgnore[method-unbinding]
@@ -144,8 +143,14 @@ export function mergeLinkedEntities(
     for (const [type, entities] of Object.entries(update)) {
       if (hasOwnProperty.call(linkedEntities, type)) {
         Object.assign(linkedEntities[type], entities);
-      } else {
+      } else if (type in linkedEntities) {
+        // $FlowIgnore[prop-missing]
         linkedEntities[type] = entities;
+      } else {
+        throw new Error(
+          JSON.stringify(type) +
+          ' is not a valid type assignable to linkedEntities',
+        );
       }
     }
   }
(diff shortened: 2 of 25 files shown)


root/static/scripts/common/linkedEntities.mjs
```diff
@@ -144,8 +143,14 @@ export function mergeLinkedEntities(
...
         Object.assign(linkedEntities[type], entities);
-      } else {
+      } else if (type in linkedEntities) {
+        // $FlowIgnore[prop-missing]
         linkedEntities[type] = entities;
+      } else {
+        throw new Error(
+          JSON.stringify(type) +
+          ' is not a valid type assignable to linkedEntities',
+        );
```
reosarevok, line 153: Did you check that we actually have all the types we're currently using listed under `linkedEntities` now? (not that long ago a few were missing, IIRC)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -5550,7 +5549,6 @@ function testAll(tests: $ReadOnlyArray<RegExp>, text: string) {
   return false;
 }
 
-// $FlowIssue[incompatible-type]: Array<mixed>
 const CLEANUP_ENTRIES: Array<CleanupEntry> = Object.values(CLEANUPS);
```
reosarevok, line 5552: Is `Array<CleanupEntry>` still needed here even if `CLEANUP` has `+[type: string]: CleanupEntry`?

root/components/RelationshipsTable.js
```diff
@@ -296,7 +296,6 @@ const RelationshipsTable = ({
         ...PagedLinkTypeGroupT,
         +phrase: string,
       }>> = Object.values(targetTypeGroup)
-        // $FlowIgnore[incompatible-call]
         .map((group: PagedLinkTypeGroupT) => ({
```
reosarevok, line 299: Is the type specification needed here, or is it just left for clarity, or you just forgot to remove it?

reosarevok (approved):
There's a couple places where we have `bigint` in the DB, not sure if there's a point to specify they are bigints in Flow?

mwiencek:
> There's a couple places where we have bigint in the DB, not sure if there's a point to specify they are bigints in Flow?

Outside of some materialized tables and replication functions, are `cover_art` and `event_art` the only ones?

That's definitely already broken (unless Perl is accidentally outputting them as strings). Changing the type to `bigint` wouldn't be right at the moment, because `JSON.parse` doesn't actually parse bigints (without a custom "reviver" function). So they actually are numbers lacking in precision (unless Perl, strings, etc.), not bigints. I think the only sane thing to do is force the artwork IDs to output as strings and change the type to `string`. But I'll punt the issue for now and hope we don't actually use those IDs for anything important.

reosarevok:
> I'll punt the issue for now and hope we don't actually use those IDs for anything important

Not without a ticket, no you won't (wagging of fingers involved).

But it does sound sensible with that caveat.

mwiencek:
> Not without a ticket, no you won't (wagging of fingers involved).

I have some good news: the [function we use to generate new cover art IDs](https://github.com/metabrainz/musicbrainz-server/blob/d2f3041c93fb491417bf0fa626ee6a03b7139fb8/lib/MusicBrainz/Server/Data/CoverArtArchive.pm#L47) is based on the number of seconds since the Unix epoch times one hundred, while [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) in JavaScript is  9007199254740991. If my arithmetic entered into WolframAlpha is correct, this means we have approximately "26 × time since the beginning of the last glacial period ( ≈ 110000 yr )" to open a ticket. 🥳

root/static/scripts/common/linkedEntities.mjs
mwiencek: I just checked all the calls to `add_linked_entity` and `link_entity` in Perl and don't see any that aren't listed.

root/static/scripts/edit/URLCleanup.js
mwiencek: Looks like it's not needed in Flow 0.197.0 at least. I'm not sure if it's needed under the new inference mode though and it can't hurt to keep, so I'll leave it alone for now.

root/components/RelationshipsTable.js
mwiencek: Same comment as above. Not needed in the old inference mode, but not certain it's not needed in the new one yet.

root/static/scripts/common/linkedEntities.mjs
mwiencek: Apparently I missed some Entity::Role::Type types, but for some reason the tests didn't fail until https://github.com/metabrainz/musicbrainz-server/pull/2884. I added them in that PR.

PR #2884, opened 2023-03-07
Upgrade to Flow 0.201.0

mwiencek:
https://medium.com/flow-type/local-type-inference-for-flow-aaa65d071347

One notable change related to our react-table usage was that I had to change the `<Table>` component into a `useTable` hook, because Flow still doesn't allow you to pass a type argument to React components. But it doesn't matter for us if it's a component or not.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -24,7 +24,6 @@ autoimports=true
 exact_by_default=true
 format.bracket_spacing=false
 format.single_quotes=true
-inference_mode=lti
 module.name_mapper.extension='less' -> '<PROJECT_ROOT>/root/flowstub.js.flow'
 react.runtime=automatic
 
--- a/.flowconfig
+++ b/.flowconfig
@@ -17,7 +17,7 @@
 
 [lints]
 unnecessary-invariant=error
-unused-promise-in-async-scope=error
+unused-promise=error
 
 [options]
 autoimports=true
--- a/.flowconfig
+++ b/.flowconfig
@@ -24,7 +24,7 @@ autoimports=true
 exact_by_default=true
 format.bracket_spacing=false
 format.single_quotes=true
-inference_mode=constrain_writes
+inference_mode=lti
 module.name_mapper.extension='less' -> '<PROJECT_ROOT>/root/flowstub.js.flow'
 react.runtime=automatic
 
--- a/lib/MusicBrainz/Server/Entity/AliasType.tt
+++ b/lib/MusicBrainz/Server/Entity/AliasType.tt
@@ -2,19 +2,17 @@
     the benefit of people looking at the auto-generated files. To regenerate
     them from this template, run ./script/generate_code.pl. -%]
 # Automatically generated, do not edit.
-package MusicBrainz::Server::Data::[% model %];
+package MusicBrainz::Server::Entity::[% model %];
 
 use Moose;
 
-extends 'MusicBrainz::Server::Data::Entity';
+extends 'MusicBrainz::Server::Entity::AliasType';
 
-with 'MusicBrainz::Server::Data::Role::AliasType';
+with 'MusicBrainz::Server::Entity::Role::OptionsTree' => {
+    type => '[% model %]',
+};
 
-sub _entity_class { 'MusicBrainz::Server::Entity::[% model %]' }
-
-sub _table { '[% table %]' }
-
-sub _type { '[% alias_type %]' }
+sub entity_type { '[% alias_type %]' }
 
 __PACKAGE__->meta->make_immutable;
 
@@ -24,7 +22,7 @@ no Moose;
 
 =head1 COPYRIGHT AND LICENSE
 
-Copyright (C) 2016 MetaBrainz Foundation
+Copyright (C) 2023 MetaBrainz Foundation
 
 This file is part of MusicBrainz, the open internet music database,
 and is licensed under the GPL version 2, or (at your option) any
--- a/root/components/list/ReleaseList.js
+++ b/root/components/list/ReleaseList.js
@@ -10,6 +10,7 @@
 import * as React from 'react';
 
 import {CatalystContext} from '../../context.mjs';
+import useTable from '../../hooks/useTable.js';
 import * as manifest from '../../static/manifest.mjs';
 import filterReleaseLabels
   from '../../static/scripts/common/utility/filterReleaseLabels.js';
@@ -30,7 +31,6 @@ import {
   removeFromMergeColumn,
   taggerColumn,
 } from '../../utility/tableColumns.js';
-import Table from '../Table.js';
 
 type Props = {
   ...InstrumentCreditsAndRelTypesRoleT,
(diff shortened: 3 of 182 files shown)


reosarevok:

root/report/components/ArtistList.js
```diff
@@ -11,28 +11,33 @@ import * as React from 'react';
...
 import PaginatedResults from '../../components/PaginatedResults.js';
-import Table from '../../components/Table.js';
+import useTable from '../../hooks/useTable.js';
 import {
   defineEntityColumn,
   defineTextColumn,
 } from '../../utility/tableColumns.js';
 
-type Props<D: {+artist: ?ArtistT, ...}> = {
+type NullableArtistResultT = {+artist: ?ArtistT, ...};
```
reosarevok, line 20: Why separate this here, but not elsewhere? (Release, etc?). It seems sensible to me but probably better to do it consistently.

root/static/scripts/guess-case/MB/GuessCase/Handler/Base.js
```diff
@@ -34,7 +34,13 @@ class GuessCaseHandler {
 
   // Member functions
 
-  checkSpecialCase(): number {
+  checkSpecialCase(
+    /*
+     * The `inputString` argument is specified to satisfy Flow.
+     */
+    // eslint-disable-next-line no-unused-vars
+    inputString?: string,
```
reosarevok, line 42: Is this a Flow bug of some sort or? Would it be better to flowissue it?

root/utility/generateFlowType.js
```diff
@@ -161,7 +161,7 @@ exports.generateFlowType = async function (
     +isEditDataTypeInfo: boolean,
   } */,
 )/*: Promise<string> */ {
-  const seenTypes = new Set();
+  const seenTypes = new Set/*:: <string> */();
```
reosarevok, line 164: Why the flow /:: syntax here?

root/static/scripts/work/edit.js
```diff
@@ -66,16 +79,20 @@ const workLanguageOptions: MaybeGroupedOptionsT = {
   options: scriptArgs.workLanguageOptions,
 };
 
-const store = createStore(function (state: WorkForm = form, action) {
+// eslint-disable-next-line func-call-spacing
```
reosarevok, line 82: Is this to avoid some weird indenting catch-22?

.flowconfig
```diff
@@ -17,14 +17,14 @@
...
-unused-promise-in-async-scope=error
+unused-promise=error
 
 [options]
 autoimports=true
 exact_by_default=true
 format.bracket_spacing=false
 format.single_quotes=true
-inference_mode=constrain_writes
+inference_mode=lti
```
reosarevok, line 27: I understand this is now the default, should we still specify it?

lib/MusicBrainz/Server/Entity/AliasType.tt
```diff
@@ -0,0 +1,31 @@
...
+
+__PACKAGE__->meta->make_immutable;
+
+no Moose;
+
+1;
+
+=head1 COPYRIGHT AND LICENSE
+
+Copyright (C) 2012 MetaBrainz Foundation
```
reosarevok, line 25: 2023?

root/entity/Aliases.js
```diff
@@ -14,7 +14,7 @@ import AliasesComponent from '../components/Aliases/index.js';
 import chooseLayoutComponent from '../utility/chooseLayoutComponent.js';
 
 type Props = {
-  +aliases: $ReadOnlyArray<AliasT>,
+  +aliases: $ReadOnlyArray<AliasT<mixed>>,
```
reosarevok, line 17: Would it make sense to have an `AliasTypeT` with the possible string values like we have for other `TypeT`s? I guess that'd clash with `AliasTypeT = OptionTreeT<'alias_type'>` - just feels weird to have `mixed` here.

reosarevok (commented):
Thanks for all this work! It generally looks great to me, left a few comments for now. Also, some wild conflicts appeared.

root/components/list/ReleaseList.js
```diff
@@ -163,24 +169,24 @@ const ReleaseList = ({
...
-        ...(mergeForm && releases.length > 2 ? [removeFromMergeColumn] : []),
-      ];
+        releaseLanguageColumn,
+        instrumentUsageColumn,
+        typeColumn,
+        statusColumn,
+        ($c.session?.tport == null) ? null : taggerColumn,
+        showRatings ? ratingsColumn : null,
+        (mergeForm && releases.length > 2) ? removeFromMergeColumn : null,
+      ].filter(Boolean);
```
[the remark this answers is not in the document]
mwiencek: It wasn't strictly required, but was semi-required, because Flow now wants us to annotate the empty array literals `[]`, which would've made things a lot uglier here. Changing those to null and filtering them out ended up being a lot cleaner.

root/report/components/ArtistList.js
mwiencek: I think I started with this file first, and wanted to create separate nullable and non-nullable types to account for the fact that `existingArtistItems` filters out the null ones already, and then change the `columns` definitions so that they don't have to unnecessarily check `result.artist` for null everywhere. Then I realized it was a lot of work and that we should probably just filter out undef items on the server. So I reverted most of the changes and probably forgot this one.

reosarevok: Feel free to change this back if you want then :) (and add a ticket for the filter, possibly?)

root/static/scripts/work/edit.js
mwiencek: I think it's actually an eslint bug.  I can add a comment above.

lib/MusicBrainz/Server/Entity/AliasType.tt
mwiencek: The base code was pretty much all copied from lib/MusicBrainz/Server/Entity/AliasType.pm which has the same copyright, but I suppose we could update it for the template changes.

root/utility/generateFlowType.js
mwiencek: 'Cause the rest of the file is still using at (and I don't remember the reason, so didn't want to break anything).

root/static/scripts/guess-case/MB/GuessCase/Handler/Base.js
mwiencek: Not a bug. The error it gives is
```js
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ root/static/scripts/guess-case/MB/GuessCase/Main.js:151:44

Cannot call handler.checkSpecialCase because no arguments are expected by function type [1]. [extra-arg]

     root/static/scripts/guess-case/MB/GuessCase/Main.js
     148│        * a special case, fetch the correct format, if the
     149│        * returned case is indeed a special case.
     150│        */
     151│       const num = handler.checkSpecialCase(inputString);
     152│       const output = handler.isSpecialCase(num)
     153│         ? handler.getSpecialCaseFormatted(inputString, num)
     154│         // if it was not a special case, start guessing

     root/static/scripts/guess-case/MB/GuessCase/Handler/Base.js
 [1]  37│   checkSpecialCase(): number {
      38│     return this.specialCaseValues.NOT_A_SPECIALCASE;
      39│   }



Found 1 error
```

mwiencek: I changed this so that the argument is specified within a Flow comment instead (which eslint doesn't see).

mwiencek:
@brainzbot test this please

reosarevok (approved):
LGTMBDNT

PR #2885, opened 2023-03-07
MBS-12552 (pre): Refine some Flow.js types

yvanzo:
# Problem

In preparation for MBS-12552, many occurrences of generic types such as `CoreEntityT` are abusively used. It’s better than `any` but it can be refined.

# Solution

- Introduce `EntityWithAliasesT` and `EntityWithArtistCreditsT`;
- Replace some (but not all) occurrences of `CoreEntityTypeT` with these;
- Drop unneeded `Core` from `CollectableCoreEntityT`.

# Testing

Not needed but ESLint and Flow, as just changing types.

### DIFF
--- a/root/types/entity.js
+++ b/root/types/entity.js
@@ -41,7 +41,7 @@ declare type CoreEntityRoleT<+T> = {
   +relationships?: $ReadOnlyArray<RelationshipT>,
 };
 
-declare type CollectableCoreEntityT =
+declare type CollectableEntityT =
   | AreaT
   | ArtistT
   | EventT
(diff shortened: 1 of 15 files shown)


root/types/entity.js
```diff
@@ -41,7 +55,7 @@ declare type CollectableCoreEntityT =
   | WorkT;
 
 declare type NonUrlCoreEntityT =
-  | CollectableCoreEntityT
+  | CollectableEntityT
   | GenreT;
```
mwiencek, line 59: I'm not sure I like defining this in terms of collectable entities to begin with, but it's unrelated to the PR.

mwiencek (approved):

root/types/entity.js
yvanzo: Completely agree on that.

PR #2886, opened 2023-03-08
Allow for JSX in .mjs files 

reosarevok:
# Problem
We have JSX in several `.mjs` files, and this was triggering an eslint error.

# Solution
Tell `jsx-filename-extension` that it is ok to have JSX in `.mjs` files.

I also changed the formatting of one rule because VSCode was telling me it was incorrect YAML formatting - although it did seem to still work anyway?

# Testing
The errors went away.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -375,6 +375,5 @@ rules:
       ["^\\.\\.(?!/?$)", "^\\.\\./?$"],
       # Other relative imports. Put same-folder imports and `.` last.
       ["^\\./(?=.*/)(?!/?$)", "^\\.(?!/?$)", "^\\./?$"],
-    ],
-  }]
+    ]}]
   simple-import-sort/exports: warn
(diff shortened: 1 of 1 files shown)


.eslintrc.yaml
```diff
@@ -375,6 +375,5 @@ rules:
       ["^\\.\\.(?!/?$)", "^\\.\\./?$"],
       # Other relative imports. Put same-folder imports and `.` last.
       ["^\\./(?=.*/)(?!/?$)", "^\\.(?!/?$)", "^\\./?$"],
-    ],
```
kellnerd, line 378: A trailing comma after the last array element is at least no valid JSON (which is a subset of YAML), maybe that's the reason?

reosarevok: Maybe! But it was still working, so I guess the parser ignored the issue :) Anyway, this should be better.

kellnerd: Yeah, I've also seen similar issue with JSON files where the linter complains about comments but the parser accepts JSONC.

mwiencek (approved):

PR #2887, opened 2023-03-08
Use React$ global types instead of React. ones

reosarevok:
# Problem
We are unnecessarily importing React in hundreds of files just to use one or another React type. Hopefully webpack is dropping all the unnecessary imports, but it's not 100% certain and it can be easily avoided.

# Solution
Use the global React$ Flow types when we're only importing React for types.

# Testing
Flow tests still pass.

# Comments
Should we also import only the parts of React we are actually using? (many files only use `useContext`, `useEffect` or whatnot).


mwiencek (approved):
Surprised at how many lines this removes. OK to merge given that tests pass, though I'd like to merge https://github.com/metabrainz/musicbrainz-server/pull/2884 first if possible so I don't have to fix conflicts again. :)

PR #2888, opened 2023-03-08
MBS-12978 / MBS-12980: Even more changes for DNB links

reosarevok:
### Fix MBS-12978 / Implement MBS-12980

# Problem
https://d-nb.info/gnd/25995-0 is apparently a valid DNB link, and outsidecontext let me know that the 9/10 digit restriction is only for artist entries anyway. Also, they seem to use HTTPS for all their permalinks now.

# Solution
outsidecontext suggested we just use `[0-9X-]+` instead for these, which hopefully will be the end of our travails with this mess. I also set everything to HTTPS now.

# Testing
Updated the tests accordingly.


reosarevok:

yvanzo (approved):
LGTBDNT. The only changed file is mostly covered by CI tests anyway. I also checked that `Entity::URL::DNB` and `Entity::URL` would not cause any breakage.

Also added a patch to make it effective immediately in the website without having users to wait for MBBE-72 to be completed.

reosarevok:

PR #2894, opened 2023-03-10
MBS-12984: Link zoomable image in the documentation

yvanzo:
# Problem MBS-12984

Links on images in the transcluded documentation are currently removed, because images didn't have to be zoomed anyway.  The rare large images (such as database schema) were linked through additional links that had to be added manually in the wiki. Now that there are more large images, it is necessary to allow zooming in.

# Solution

This patch allow linking a thumbnail to the original image simply by adding the option "class=zoomable" to the wanted "Image" WikiLinks.

It's cheaper than MediaWiki's MultiMedia Viewer extension but it works.

# Testing

Tested manually on the page `/doc/MusicBrainz_Database/Schema`, using the remote production standby database and a local flushed Redis cache.

There is no CI test for it, and it wouldn’t be worth adding any given that it's just a bandage.

### DIFF
--- a/lib/MusicBrainz/Server/Data/WikiDoc.pm
+++ b/lib/MusicBrainz/Server/Data/WikiDoc.pm
@@ -31,10 +31,24 @@ sub _fix_html_links
 
     my $href = $node->attr('href') || '';
 
-    # Remove broken links & links to images in the wiki
     if ($href =~ m,^(?:https?:)?//$wiki_server/(File|Image):, || $class =~ m/new/)
     {
-        $node->replace_with($node->content_list);
+        my $child = $node->getFirstChild();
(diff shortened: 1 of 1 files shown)


reosarevok (commented):
It certainly works better than what we have now. Given it doesn't zoom in a bubble but navigates to a new page, I might prefer it to open in a new tab, but it's not too bad as is.

Do you have examples of "Remove broken links & links to images in the wiki otherwise" for testing? (since I understand all images on the tested page are zoomable)

yvanzo:
Updated to open in a new tab.

See https://musicbrainz.org/doc/WikiDocs for an example of image with removed link.

reosarevok (approved):
Works nicely, thanks!

PR #2895, opened 2023-03-13
MBS-12994: Treat email as case-insensitive for admin email search

reosarevok:
### Implement MBS-12994

# Problem
While technically email addresses are [meant to be case-sensitive](https://blog.teknkl.com/email-addresses-case-sensitive-but-anyway/) the whole internet treats them as if the were not. Just today I was going to answer to a user who wanted to be deleted that we did not have them on the DB to begin with, until I realized they had typed their email with different caps in different places and I notice we do check our emails case-sensitively in `search_by_email`. We don't elsewhere (our `is_email_used_elsewhere` comparison does `lower()` for comparison).

# Solution
Turn the regex matching in `search_by_email` to case-insensitive.

# Testing
Manually, making sure a search for my email but CAPITALIZED still matches my user(s) now. Also added a test case to Data::Editor.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Editor.pm
+++ b/lib/MusicBrainz/Server/Data/Editor.pm
@@ -198,7 +198,7 @@ sub search_by_email {
 
     my $query = 'SELECT ' . $self->_columns .
         ' FROM ' . $self->_table .
-        q" WHERE (regexp_replace(regexp_replace(email, '[@+].*', ''), '\.', '', 'g') || regexp_replace(email, '.*@', '@')) ~ ?" .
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
Code change looks good, did not test but test coverage looks good too. :100: :ship: 

As there are such comments in other tests, do you think we should generalize `diag` usage?

reosarevok:
> As there are such comments in other tests, do you think we should generalize `diag` usage?

Yes! We use a few of those, but we probably have a ton more places where we could be doing it. There's also `note` but that does not get printed with `prove` so I think `diag` is better.

PR #2896, opened 2023-03-13
MBS-12552 (pre): Refine most Flow.js types

yvanzo:
# Problem

The JS code is using the catch-all type `CoreEntityT` in many places where only a small subset of it is needed. It makes it difficult to know what is exactly expected from this kind of entity.

Better typing the JS code should help with better classifying entity types, as a prerequisite to MBS-12552.

# Solution

This branch has been edited by the three of us, @mwiencek, @reosarevok and I, even though it is not ready yet, to have a common place for conversations.

It follows the pull request #2885.

Types being added by this pull request:
* `CollectableEntityTypeT`
* `EditableEntityT`
* `EditableEntityTypeT`
* `EntityWithAliasesTypeT`
* `EntityWithSeriesT`
* `EntityWithSeriesTypeT`
* `EntityWithWikipediaExtractT`
* `MergeableEntityT`
* `MergeableEntityTypeT`
* `MinimalAnnotatableEntityT`
* `MinimalEntityT`
* `MinimalEntityWithWikipediaExtractT`
* `MinimalTaggableEntityT`
* `NonReleaseRelatableEntityT`
* `NonReleaseRelatableEntityTypeT`
* `NonUrlRelatableEntityT`
* `NonUrlRelatableEntityTypeT`
* `PropsWithEntity`
* `RelatableEntityT`
* `RelatableEntityTypeT`
* `SubscribableEntityT`
* `SubscribableEntityTypeT`
* `SubscribableEntityWithSidebarT`
* `TaggableEntityT`
* `TaggableEntityTypeT`

# Checklist for author

A few “Core” occurrences still remain under the `root` directory.

# Actions

Next steps are to replace these latest occurrences and finally remove the *Core* type definitions.


root/types/entity.js
```diff
@@ -7,6 +7,20 @@
...
+  | GenreT
+  | InstrumentT
+  | LabelT
+  | PlaceT
+  | RecordingT
+  | ReleaseT
+  | ReleaseGroupT
+  | SeriesT
+  | WorkT;
+
```
yvanzo, line 23: @reosarevok wrote in https://github.com/metabrainz/musicbrainz-server/commit/f4065f0bf562734bc4111d60301db8f66fd6dd50#r103558097:

> There's already `AnnotatedEntityT` in `annotation.js`

root/static/scripts/common/utility/entityHref.js
```diff
@@ -19,7 +19,7 @@ type LinkableEntity =
   | {+entityType: 'editor', +name: string, ...}
   | {+entityType: 'isrc', +isrc: string, ...}
   | {+entityType: 'iswc', +iswc: string, ...}
-  | {+entityType: CoreEntityTypeT, +gid: string, ...}
+  | {+entityType: RelatableEntityTypeT, +gid: string, ...}
```
yvanzo, line 22: @reosarevok wrote in https://github.com/metabrainz/musicbrainz-server/commit/14988ae5210b93d1e841f121a215f35d620af210#r103558464:
> Genre for example was linkable and a "CoreEntity" before it supported relationships. Although we might not implement things like that in the future, and I guess it would be ok to just have any other entities like that be added here separately like `collection` and `link_type`.

@bitmap answered:
> Right. I'm just asserting that all relatable entities are linkable, not that they comprise all or most linkable entities. (But I also wouldn't have considered genre a core entity until they supported relationships, which I guess helps prove the vagueness of the term.)

That’s about commit https://github.com/metabrainz/musicbrainz-server/pull/2896/commits/63fc9201b9fc417861cb2bb82accd3b419ab831a.

root/static/scripts/common/constants.js
```diff
@@ -17,9 +17,22 @@ export const EMPTY_PARTIAL_DATE: PartialDateT = Object.freeze({
   year: null,
 });
 
+type N_l_T = () => string;
```
yvanzo, line 20: @reosarevok wrote in https://github.com/metabrainz/musicbrainz-server/commit/91b0577b7cb3869b9f69ee2c77be4281f02b3ca2#r103558776
> If we're doing this, put it somewhere where it's reusable, because we have several other places in the code with `() => string` types specifically for `N_l`.

That’s about commit https://github.com/metabrainz/musicbrainz-server/pull/2896/commits/8d530b8a7b2ea0781dbbc6479b85d366bd63be35 by @mwiencek.

root/types/wsjs.js
```diff
@@ -9,7 +9,7 @@
 
 declare type WsJsRelationshipEntityT =
   | {
-      +entityType: NonUrlCoreEntityTypeT,
+      +entityType: NonUrlRelatableEntityTypeT,
       +gid: string,
       +name: string,
     }
```
yvanzo, line 15: There is a resolved (?) conversation started by @reosarevok in https://github.com/metabrainz/musicbrainz-server/commit/91b0577b7cb3869b9f69ee2c77be4281f02b3ca2#r103559074 about this part of commit https://github.com/metabrainz/musicbrainz-server/pull/2896/commits/8d530b8a7b2ea0781dbbc6479b85d366bd63be35 by @mwiencek.

root/static/scripts/common/utility/entityHref.js
yvanzo: It has been further discussed on IRC and agreed it was an non-issue as it can be extended using union type if needed when a non-relatable entity type has to use this. That was already the case in many components using `| CollectionT` for example.

yvanzo (commented):
The changes look good to me so far.

* All `CoreEntity(Type)T` have been removed;
* A lot of types have been refined (even though it might be possible to further refine a some of those);
* Code has even been fixed in a few cases!

The only drawback is the duplication of type definition with  `entities.mjs` but it has been agreed on tracking it with MBS-13003 to fix it later on.

_Edit: I will apply the amending commits and rebase if approved._

root/static/scripts/tests/relationship-editor.js
```diff
@@ -851,7 +851,7 @@ function addRelationships(
 
 function addRelationshipsToRelease(
   rootState: ReleaseRelationshipEditorStateT,
-  source: CoreEntityT,
+  source: RelatableEntityT,
```
reosarevok, line 854: Commit message says Editable but the changes are to Relatable :)

root/static/scripts/common/components/CommonsImage.js
```diff
@@ -14,7 +14,7 @@ import entityHref from '../utility/entityHref.js';
 
 type Props = {
   +cachedImage: ?CommonsImageT,
-  +entity: CoreEntityT,
+  +entity: NonUrlRelatableEntityT,
```
reosarevok, line 17: Why do we even keep this code around?

yvanzo: I guess that the hope is that Commons will eventually overcome the issue that led us to disable it on our website.  It can be still used by local mirrors.

root/static/scripts/tests/relationship-editor.js
yvanzo: That was commit 7a641bc0585c057ac8283744e3385e0ff50dd398. Now fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/2896/commits/6ba688d993613f0a5a57fb7747f7f8c9e4770a61.

mwiencek (approved):
I reviewed all of the new commits and everything looks good to me save for any squashing that still needs to be done. Thanks a lot for looking into the rest of the `CoreEntityT` uses!

reosarevok (approved):
I checked this a couple days ago and it also seemed fine to merge to me - anything else left can be improved later :)

yvanzo:
Squashed the amending commits and rebased on `master` branch.

PR #2898, opened 2023-03-14
MBS-12989 / MBS-12998: Edit list quick links improvements

reosarevok:
### Fix MBS-12989 / MBS-12998

# Problem
The quick links section has a few issues:
* The refine search link on the subscribed editors page does not set the "I haven't voted" predicate (MBS-12989)
* The edit search pages are not supposed to link to themselves, but they do (no ticket)
* The quick links sometimes (like for Open edits) start with a stray | separator (MBS-12998)

# Solution
The first commit adds the missing predicate to the `refine_url_args` of `subscribed_editors`.

The second commit sets `isSearch` when calling `EditList` from `search_results.tt`, which I think was always supposed to happen - we literally didn't set `isSearch` anywhere at the moment.

The third commit builds an array of any appropriate quick links, then joins them with the | separator as needed, as we do with `FooterSwitch` in `ArtistIndex`, rather than trying to guess when a separator will be needed. This also allowed me to group the conditions a bit better, which hopefully is clearer.

# Testing
Manually, navigating between all the different edit lists, when logged in and logged out.

### DIFF
--- a/root/edit/components/ListHeader.js
+++ b/root/edit/components/ListHeader.js
@@ -22,13 +22,13 @@ type Props = {
   +username?: string,
 };
 
-const ListHeader = ({
+const QuickLinksSection = ({
   entity,
   isSearch = false,
   page,
   refineUrlArgs,
   username,
-}: Props): React$Element<'table'> => {
+}: Props): React$Element<'td'> => {
   const $c = React.useContext(CatalystContext);
   const isSecureConnection = $c.req.secure;
   const protocol = isSecureConnection ? 'https://' : 'http://';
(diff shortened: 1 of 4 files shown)


reosarevok:

jesus2099:
I tested the stray pipe problem and it's gone!

yvanzo (approved):
Much better code. Tested manually. Updated the already linked ticket MBS-12998 to cover hiding the link “Search for edits” too (as in the current second commit).
:ship:

Related to this second commit (but that can be addressed in a later pull request): The paragraph “No edits were found matching your query. Please try again.” that is now displayed (and was never displayed at least since the React conversion) doesn’t seem to be useful. It says the same thing as the “Found 0 edits” which is already shown. The link named “try again” actually blanks search conditions, while it would expect the same search to be tried again. I would rather just remove this paragraph.

reosarevok:
You're right about that no results message actually, it does more harm than good. Added a commit dropping it.

root/edit/components/ListHeader.js
jesus2099, line 1: Super!
I wanted to do that but I didn't recognise the JavaScript I know.
In JavaScript I usually use `join()` but this `.js` file uses some high tech code I didn't understand, so I was clueless and could only repeat exact same existing code.

PR #2901, opened 2023-03-15
MBS-12942: Also indicate if target has pending edits in relationship editor 

reosarevok:
### Implement MBS-12942

# Problem
The old relationship editor highlighted entities with pending edits, but the new one does not. We now indicate *relationships* with pending edits with an icon, but there's nothing indicating an *entity* has pending edits.

# Solution
@Aerozol came up with a pair of icons to represent entity and relationship pending edits, so I added a second tooltip option for when the entity has pending edits. See images below. This one links to the `/open_edits` page for the entity.

# Testing
By hand on a fake release I changed so that it has both types of pending edits.

# Example
image_2023-06-13_142000825

On top of https://github.com/metabrainz/musicbrainz-server/pull/2964 to use the other icons in the new set already.


reosarevok:

Aerozol:
Nice!!
I think the icons are too prominent atm. Could the icons be made slightly smaller, so the yellow circle is the same size as the circle around the (+) icon?

reosarevok:
Hmm. The height of both those icons is supposed to be 16px right now...

Aerozol:
huh, maybe the other icons have whitespace around them? It’s only a few px but it makes a difference:
image

reosarevok:
@Aerozol: can you play with it a bit and see what the preferred height would be for you? 15? 14?

atj:
Apologies if you've already discussed this, but is there any way we can convey the meaning of this icon a bit better? I tried adding the following but you can't really make it out at 16px:

open_edits_for_entity-new

reosarevok:
Hovering will open a tooltip anyway, so I don't think it's the end of the world if we cannot make it more obvious than it is now at that size :)

mwiencek (approved):

PR #2902, opened 2023-03-16
MBS-12976: Changing a work can cause duplication/key errors

mwiencek:
# Fix MBS-12976

## Problem

See the steps in MBS-12976, though you may not always be able to reproduce it: it depends on the relative sort orders of the names and IDs of the works you choose.

## Solution

This was caused by d95b3d3769dc2f494df2751988b8e3bee2c5fcf1. There were other comparators used for updating the `relatedWorks` entries besides that one. I've tried to make them call each other here to avoid other issues in the future.

## Testing

Added a unit test which reproduced the duplication issue.


mwiencek:
@brainzbot test this please

PR #2903, opened 2023-03-16
MBS-12990: Don't pad the legend relationship editor fieldset

reosarevok:
### Fix MBS-12990

# Problem
The "Relationships" heading when editing an entity is indented further right than all other `fieldset` legends.

# Solution
These legends are all supposed to be aligned, but the `padding` being applied to the whole `fieldset` was moving this one further right. I was planning to just add `padding` to the `table` inside, but I guess it would also make sense to keep errors padded, so I added an extra `div` for it.

# Testing
Manually, just making sure the alignment is correct.


yvanzo (approved):
LGTMBDNT :ship:

PR #2904, opened 2023-03-17
MBS-12956: Allow searching for edits with votes by beginners

reosarevok:
### Implement MBS-12956

# Problem
Beginner editors can now vote, but there's no way to search for edits where they've voted. This is useful because a surprising amount of voting beginners have turned out to be sockpuppets, but also in general to be able to double-check beginners' votes since it's expected they're still learning.

# Solution
This adds a "limited" option for Voter in the same way we have it for editor (based on the query from https://github.com/metabrainz/musicbrainz-server/pull/2808). This can already be useful since it's mostly relevant for the last few days of edits, but it times out quite easily for bigger searches - we should consider materializing the concept of "beginner" properly in the database with triggers, especially since we're using it in many places.

# Testing
Manually, limiting the edits to the last 10 days to make sure it did not time out.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate/Voter.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate/Voter.pm
@@ -3,7 +3,7 @@ use Moose;
 use namespace::autoclean;
 use Scalar::Util qw( looks_like_number );
 
-use MusicBrainz::Server::Constants qw( :vote );
+use MusicBrainz::Server::Constants qw( :vote $EDITOR_MODBOT $STATUS_APPLIED );
 use MusicBrainz::Server::Validation qw( is_database_row_id );
 
 with 'MusicBrainz::Server::EditSearch::Predicate';
@@ -31,6 +31,7 @@ sub operator_cardinality_map {
         'not_me' => undef,
         'subscribed' => undef,
         'not_subscribed' => undef,
+        'limited' => undef,
     );
 };
 
@@ -43,6 +44,7 @@ sub valid {
 sub voter_clause {
     my $self = shift;
     my $sql = 'vote.editor ';
+    my @args;
 
     if ($self->operator eq 'subscribed') {
         $sql .= 'IN (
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/EditSearch/Predicate/Voter.pm
```diff
@@ -43,32 +44,65 @@ sub valid {
...
-    } else {
+        push @args, $self->voter_id;
+    } elsif ($self->operator eq 'me') {
         $sql .= '= ?';
+        push @args, $self->user->id;
+    } elsif ($self->operator eq 'not_me') {
+        $sql .= '!= ?';
+        push @args, $self->user->id;
+    } elsif ($self->operator eq 'limited') {
+        # Please keep the beginner logic in sync with Report::LimitedEditors and Entity::Editor
```
yvanzo, line 76: If the goal is to prevent the different parts of the code to get out of sync, a similar comment should probably be added in these other parts of the code.

lib/MusicBrainz/Server/EditSearch/Predicate/Voter.pm
```diff
@@ -43,32 +44,65 @@ sub valid {
...
+                    member_since > NOW() - INTERVAL '2 weeks'
+                    OR NOT EXISTS (
+                        SELECT 1
+                          FROM edit e2
+                         WHERE e2.editor = beginner.id
+                           AND e2.autoedit = 0
+                           AND e2.status = ?
+                        OFFSET 9
+                    )
+                )
```
yvanzo, line 92: As this query is known to time out as-is, it should be limited to the last week for example.

yvanzo (changes requested):
> a surprising amount of voting beginners have turned out to be sockpuppets

Do you mean _unsurprising_? Made a comment to MBS-10614 which has some watchers about partly reverting beginner’s restrictions.

Otherwise, this search should probably be added to voting suggestions.

lib/MusicBrainz/Server/EditSearch/Predicate/Voter.pm
reosarevok: It doesn't feel right to have queries restricted to a hardcoded amount of time in the query itself - it seems better then to add a default time predicate to searches as you suggested at some point in the past :)

In any case, even if we hardcoded it, this wouldn't be hardcoded here, but would probably have to be some sort of more generic flag that would be then used by EditSearch::Query, maybe? So we would need to talk about how to do it more widely.

reosarevok:
> Do you mean unsurprising?

No, I did mean surprising, because they started doing it very fast as soon as it was made possible. I doubt these random sockpuppeteers were following our development, so I can only assume that means that they were actually trying stuff like this for a long time and we have tons of sockpuppet accounts around from before the change which were created to vote and abandoned when they realized they couldn't...

reosarevok:
Since we blocked beginners from being able to vote again, this is no longer needed.

PR #2905, opened 2023-03-20
MBS-13005: Remove Videogam.in links from the sidebar

reosarevok:
### Implement MBS-13005

# Problem
The Videogam.in page ended in 2017, but we still have all the (mostly useless) code for it.

# Solution
To be safe and since we plan to leave the links around as ended, not removing the URLCleanups whitelisting, but there seems to be no point to keep supporting the sidebar display, favicon and other "cosmetic" bits.

# Testing
Added one link locally, made sure it showed before this patch but not after, and that it could still be edited (set as ended) without errors.


yvanzo (approved):
Thanks, we can skip the beta period for this small remove-only locally-tested patchset. :ship:

PR #2906, opened 2023-03-20
MBS-12800: Exclude cancelled releases from first_release_date

reosarevok:
### Implement MBS-12800

# Problem
A release marked as cancelled by definition does not have a release date and as such should not be used when calculating the first release date of recordings or release groups connected to it.

# Solution
This amends the `get_release_first_release_date_rows` function to ignore releases with status 6 (cancelled). It also updates the `a_upd_release` function to recalculate first dates when a status is changed to or from 6 (cancelled).

# Checklist for author
* [x] Write an update script to amend the current data for releases marked as cancelled during the schema change.

This is a very small amount of all releases, so probably better to have a custom script that only updates rows related to these specific releases, instead of running `BuildMaterializedTables` to regenerate the whole thing. The simplest way seems to be:
1) Delete the cancelled release rows from `release_first_release_date`.
2) Run something like this to update release group dates:
```
  SELECT set_release_group_first_release_date(id) FROM (
    SELECT id
    FROM release_group
    WHERE EXISTS (
      SELECT TRUE
      FROM release
      WHERE status = 6
      AND release.release_group = release_group.id
    )
  )
```
3) Run something like this to update recording dates:
```
  SELECT set_releases_recordings_first_release_dates(ids) FROM (
    SELECT array_agg(id)
    FROM release
    WHERE status = 6
  )
```
* [x] Test the migration actually works.

# Testing
Added PGTAP tests to see cancelled release status affects dates as expected.

I did a basic migration test with sample local data: I set one release as cancelled and 2020 in one RG, another in the same RG as official and 2021, with a shared recording too. Before upgrade.sh both RG and recording showed 2020 as the date, after both show 2021, so that basic case at least works.


reosarevok:
Marking for review so the code can be checked before we start testing migrations with it.

yvanzo (changes requested):
The code looks good but it requires a test plan.

admin/sql/CreateFunctions.sql
```diff
@@ -1066,7 +1079,13 @@ BEGIN
...
         ) all_dates
         WHERE ' || condition ||
-        ' ORDER BY release, year NULLS LAST, month NULLS LAST, day NULLS LAST';
+        ' AND NOT EXISTS (
+          SELECT TRUE
+            FROM release
+           WHERE release.id = all_dates.release
+             AND status = 6
+        )
+        ORDER BY release, year NULLS LAST, month NULLS LAST, day NULLS LAST';
```
mwiencek, line 1088: The added subquery doesn't significantly affect the performance of the query. Planning and execution time are still both sub-1ms on floyd. (Tested with both cancelled and non-cancelled release IDs.)

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+UPDATE release SET status = 6 WHERE id = 3;
+
+SELECT results_eq(
+  'all_release_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint)
+  $$,
+  'Release dates as expected after marking release as cancelled'
```
yvanzo, line 237: It’s not about the “release dates”, it’s about the first release date for releases of the same group.
```suggestion
  'Release’s first release (of the same group) dates as expected after marking release as cancelled'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'release_group_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint)
+  $$,
+  'RG dates as expected after marking release as cancelled'
```
yvanzo, line 247: To clarify it is not about any other date (such as for RG relationships):
```suggestion
  'RG (first release) dates as expected after marking release as cancelled'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'recording_1_first_release_date',
+  'VALUES (1989::smallint, null::smallint, null::smallint)',
+  'Recording date as expected after marking release as cancelled'
```
yvanzo, line 253: To clarify it is not about any other date (such as for recording relationships):
```suggestion
  'Recording (first release) date as expected after marking release as cancelled'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+
+SELECT results_eq(
+  'all_release_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint),
+      (3::integer, 1970::smallint, 11::smallint, null::smallint)
+  $$,
+  'Release dates as expected after marking release as official again'
```
yvanzo, line 266: Same as above:
```suggestion
  'Release’s first release (of the same group) dates as expected after marking release as official again'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'release_group_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1970::smallint, 11::smallint, null::smallint)
+  $$,
+  'RG dates as expected after marking release as official again'
```
yvanzo, line 276: Same as above:
```suggestion
  'RG (first release) dates as expected after marking release as official again'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'recording_1_first_release_date',
+  'VALUES (1970::smallint, 11::smallint, null::smallint)',
+  'Recording date as expected after marking release as official again'
```
yvanzo, line 282: Same as above:
```suggestion
  'Recording (first release) date as expected after marking release as official again'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+UPDATE release SET status = 6, release_group = 1 WHERE id = 3;
+
+SELECT results_eq(
+  'all_release_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint)
+  $$,
+  'Release dates as expected after marking release as cancelled and changing its RG'
```
yvanzo, line 294: Same as above:
```suggestion
  'Release’s first release (of the same group) dates as expected after marking release as cancelled and changing its RG'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'release_group_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint)
+  $$,
+  'RG dates as expected after marking release as cancelled and changing its RG'
```
yvanzo, line 304: Same as above:
```suggestion
  'RG (first release) dates as expected after marking release as cancelled and changing its RG'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'recording_1_first_release_date',
+  'VALUES (1989::smallint, null::smallint, null::smallint)',
+  'Recording date as expected after marking release as cancelled and changing its RG'
```
yvanzo, line 310: Same as above:
```suggestion
  'Recording (first release) date as expected after marking release as cancelled and changing its RG'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+
+SELECT results_eq(
+  'all_release_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint),
+      (3::integer, 1970::smallint, 11::smallint, null::smallint)
+  $$,
+  'Release dates as expected after marking release as official again'
```
yvanzo, line 323: Same as above:
```suggestion
  'Release’s first release (of the same group) dates as expected after marking release as official again'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'release_group_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1970::smallint, 11::smallint, null::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint)
+  $$,
+  'RG dates as expected after marking release as official again'
```
yvanzo, line 333: Same as above:
```suggestion
  'RG (first release) dates as expected after marking release as official again'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'recording_1_first_release_date',
+  'VALUES (1970::smallint, 11::smallint, null::smallint)',
+  'Recording date as expected after marking release as official again'
```
yvanzo, line 339: Same as above:
```suggestion
  'Recording (first release) date as expected after marking release as official again'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+
+SELECT results_eq(
+  'all_release_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1989::smallint, null::smallint, null::smallint),
+      (3::integer, 1970::smallint, 11::smallint, null::smallint)
+  $$,
+  'Release dates as expected after changing release RG back'
```
yvanzo, line 352: Same as above:
```suggestion
  'Release’s first release (of the same group) dates as expected after changing release RG back'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'release_group_first_release_dates',
+  $$
+    VALUES
+      (1::integer, 1999::smallint, 1::smallint, 4::smallint),
+      (2::integer, 1970::smallint, 11::smallint, null::smallint)
+  $$,
+  'RG dates as expected after changing release RG back'
```
yvanzo, line 362: Same as above:
```suggestion
  'RG (first release) dates as expected after changing release RG back'
```

t/pgtap/first_release_dates.sql
```diff
@@ -225,6 +225,149 @@ SELECT results_eq(
...
+);
+
+SELECT results_eq(
+  'recording_1_first_release_date',
+  'VALUES (1970::smallint, 11::smallint, null::smallint)',
+  'Recording date as expected after changing release RG back'
```
yvanzo, line 368: Same as above:
```suggestion
  'Recording (first release) date as expected after changing release RG back'
```

yvanzo (commented):
Thanks for having added tests.
Just made marginal suggestions.
LGTMBDNT otherwise.

admin/sql/CreateFunctions.sql
```diff
@@ -479,6 +479,19 @@ BEGIN
         PERFORM dec_ref_count('artist_credit', OLD.artist_credit, 1);
         PERFORM inc_ref_count('artist_credit', NEW.artist_credit, 1);
     END IF;
+    IF (
+        NEW.status IS DISTINCT FROM OLD.status AND
+        (NEW.status = 6 OR OLD.status = 6)
+    ) THEN
```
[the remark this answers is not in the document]
reosarevok: It might be possible, it just seems like it would be less clear. I expect that's the reason we have two different checks for other things already (there's multiple `NEW.release_group != OLD.release_group` and `NEW.artist_credit != OLD.artist_credit`).

t/pgtap/first_release_dates.sql
reosarevok: IIRC this is actually about the earliest date for *this specific release* (for cases when there's several regions with different dates), or am I remembering wrong? (so it's not about a group).

admin/sql/CreateFunctions.sql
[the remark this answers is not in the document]
yvanzo: There is only one other `NEW.status IS DISTINCT FROM OLD.status`.

t/pgtap/first_release_dates.sql
yvanzo: Good catch. Then maybe something like that instead?
```suggestion
  'Release’s first release event dates as expected after marking release as cancelled'
```

admin/sql/CreateFunctions.sql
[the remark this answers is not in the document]
reosarevok: Yes, I know. My point is that this function already duplicates other checks for clarity when it could group them to save one check, so it seems sensible to keep prioritizing clarity over grouping for this new case.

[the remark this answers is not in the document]
yvanzo: OK, it doesn’t seem to be a valid point. Grouping conditions would not only be a gain of performances, it would also be a gain of clarity as you won’t have to skim the whole function for repeated conditions. Even if it was about to make anything less clear, comments can always be added to fix that.

reosarevok:

admin/sql/CreateFunctions.sql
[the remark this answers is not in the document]
mwiencek: IIUC, this check is `A && B` while the one below is `A || C || D`.  So combining the `A` checks means you'll have to duplicate the `INSERT INTO artist_release_group_pending_update` line.  When I was originally writing those materialized tables, I found it clearer to keep all the conditions related to updating them together, because I could quickly check that all the cases are covered in one place.  I didn't worry about performance so much because comparing two integers in PL/pgSQL should be extremely fast.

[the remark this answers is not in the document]
mwiencek: > No INSERT duplication is needed, you can just have A && B nested under A || C || D.

I thought the point was to avoid testing A twice.

mwiencek (approved):
LGTM, though a couple queries in the upgrade script can be simplified slightly.

reosarevok:

admin/sql/CreateFunctions.sql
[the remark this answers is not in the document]
yvanzo: `A && B` would not tested twice when `A || C || D` is false, which is a very (possibly the most) common case.

But this discussion already wasted more energy than it would save. Maybe one day someone will care improving either performance or clarity. Now that we are pressed by time for the schema change, this is not a showstopper for merging this pull request.

yvanzo (approved):
:ship:

PR #2908, opened 2023-03-21
MBS-13008: Add URL cleanup for Genie

Maxr1998:
Genie is a Korean music streaming platform that also allows purchasing albums for download. The URL cleanup rules remove unnecessary segments and convert mobile URLs to their desktop counterparts. Album URLs, artist and song pages are supported and validated. Search results will explicitly be prevented from being added.

Since it may happen that albums are only for streaming, we keep the option to set just the streamingpaid relationship to be on the safe side.

Closes MBS-13008.


yvanzo (approved):
Brilliant, thank you! :sparkles:

PR #2909, opened 2023-03-23
MBS-12794: Don't use spammer tags/ratings when calculating tag counts / rating averages

reosarevok:
### Implement MBS-12794

# Problem
The ratings and tags of accounts marked as spammers are still being used when calculating aggregate/average ratings and tags for entities. We don't want to delete the ratings and tags, for two reasons: 1) They can be useful for SpamBrainz; 2) If it turns out we were wrong to set them as spammers, we don't want to have lost all their data. But we don't want spammy tags and ratings to be shared with others and affect the community's averages.

# Solution
This adds a function (and trigger) `update_tags_and_ratings_for_spammer` to remove the tags and ratings by the spammer user from the calculated averages when the flag is added. For ratings, the average is just recalculated. For tags, it's a bit more complicated, and we need to basically emulate what happens when a tag is added/removed.

The functions for adding, editing and removing ratings and tags are also amended to do nothing if the editor is a spammer; this is mostly a safety fallback, since in theory at least any user marked as spammer is unable to log in and gets automatically logged out when the flag is set. 

# Checklist for author
* [ ] Write an update script to amend the current tag / rating data for editors already marked as spammers during the schema change.
* [ ] Add spammer cases to the existing rating/tag tests and add a test for setting/unsetting the flag.
* [ ] Test the migration actually works.

# Testing
None yet.


reosarevok:
Closing for now since we decided not to do this yet and maybe look at a better option.

PR #2910, opened 2023-03-25
MBS-13011: Block short links: spotify.link

reosarevok:
### Implement MBS-13011

# Problem
`spotify.link` links are just shortened redirects to a specific Spotify page, so we should do the same with them as we do with all other redirect shorteners like Amazon's and Apple's: block, and ask the user to follow the redirect.

# Solution
As above :) 

# Testing
Added a test - I also had to fix the error checking code to support React errors.


yvanzo (approved):
Look good to me, including the conversion of error message in tests.
Did not test, but added test offers a good coverage.
:ship:

PR #2912, opened 2023-03-29
MBS-13016: Allow, clean-up, and display WorldCat Entities URL

yvanzo:
# Problem MBS-13016

WorldCat Identities has been replaced with WorldCat Entities; See the ticket for more info.

MusicBrainz has ~50K “other database” links WorldCat, and mostly to Identities but not only.

The problem is that WorldCat Entities are hosted under a different domain name, so its URLs are currently neither allowed nor displayed in MusicBrainz.

# Solution

Allow WorldCat Entities URLs, clean those up to follow their “Copy URI” format, and display in the sidebar under the same name “WorldCat” but using a different favicon.

No change is required to the documentation which neither mention nor link to the discontinued Identities app.

Note that it keeps allowing for now the old WorldCat Identities URLs to be entered, because they are redirected to the new WorldCat Entities, and because WorldCat Identities IDs can still be found in many places such as WikiData for example.

# Testing

Added CI tests for allowing and cleaning up.
Tested adding the example link to an artist and looking at it in the sidebar.


reosarevok (approved):
LGTMBDNT. Can the old links be converted automatically, or do we need a bot to follow each redirect and update like that? In either case it should probably be an MBBE ticket.

reosarevok:

yvanzo:
> Can the old links be converted automatically, or do we need a bot to follow each redirect and update like that? In either case it should probably be an MBBE ticket.

This is detailed in comment to the ticket already.

reosarevok:
Sorry, I guess I expected a linked MBBE ticket to exist already and that's what I went check, rather than the comment. My bad :) 

For the last bit of the comment, that'd be great also for Wikipedia -> Wikidata, at least, so it would be good to have a generic "replacement finder" :)

yvanzo:
No problem, I should also have put it under a section “Actions” in the PR description.

Yes, having _a generic “replacement finder”_ is the idea and would have a lot of use cases. Unfortunately, it isn’t straightforward at the moment because the current specification of  `fetch` intentionally prevents getting the redirect destination from the browser. That’s the main reason I didn’t elaborate this idea so far. Some browsers are making progress towards it, but the polyfills are still not supporting it.

PR #2913, opened 2023-04-07
MBS-13025: Escape link symbols in setlist syntax

yvanzo:
# Problem MBS-13025

There is currently no way to escape squared brackets `[` and `]` from the `[mbid|name]` link syntax in event’s setlist. Rendering of squared brackets most often fails when such link is in the same line.

Note: It isn’t new to the React implementation, it was the same using Template Toolkit.

# Solution

Support the HTML entities for squared brackets, and for the ampersand `&` since it starts HTML entities.

This is the easiest to implement since it does not require to change the regular expression that captures links.

This is also the most natural (already tried for the event linked in the ticket, and already used in 3 setlists) and the most common (supported by both MediaWiki syntax and Markdown syntax).

It recommends using `&lsqb;`, `&rsqb;`, and `&amp;` HTML entities, but it also supports all equivalent HTML entities which will be documented in https://musicbrainz.org/doc/Event/Setlist and which are taken from https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references

# Testing

- Extended the JavaScript CI test for
  * rendering all the HTML entities
  * decoding HTML entity only once, e.g. `&amp;amp;` (setlist) → `&amp;` (plain text) → `&amp;amp;` (HTML)
- Manually tested the full process of submitting and displaying a setlist event with such HTML entity by copying the setlist from https://musicbrainz.org/event/13c739b2-e39d-4106-9cd4-75062c2fca7d into a new event (in local development database).

# Checklist for author

* [ ] Verify that the current occurrences of these entities are used in the same goal:
      ```sql
      SELECT setlist FROM event WHERE setlist ~ '&(#(38|91|93|x(26|7b|7d))|lsqb|rsqb|lbrack|rbrack);'
      ``` 
* [ ] Update the [`Event/Setlist`](https://wiki.musicbrainz.org/Event/Setlist) wiki page

# Action

1. Make updated WikiDocs pages visible on MusicBrainz website


root/event/edit_form.tt
```diff
@@ -20,7 +20,8 @@
         [% END %]
         <p>
             [%- l('Add "@ " at line start to indicate artists, "* " for a work/song, "# " for additional info (such as "Encore").
-                   [mbid|name] allows linking to artists and works.') -%]
+                   [mbid|name] allows linking to artists and works.
+                   Escape the symbol "[" using the HTML entity "&lrqb;", "]" using "&lsqb;", "&" using "&amp;".') -%]
```
reosarevok, line 24: Written like this it sounds a bit like you need to escape [] in [mbid|name] to me, but maybe it's not confusing for people? We could prepend "To show some symbols as part of the setlist text you will need to escape them:"

reosarevok (approved):
LGTMBDNT

reosarevok:

reosarevok (approved):
LGTMBDNT

PR #2914, opened 2023-04-10
MBS-12370: Rename schema change upgrade script suffixes to clearly indicate which nodes they run on

mwiencek:
# Fix MBS-12370

## Problem

(Refer to the description of MBS-12370.)

## Solution

Rename mirror -> all
Rename extensions -> all_extensions
Rename standalone -> master_and_standalone

The ticket suggested master_or_standalone, and it means the same thing to me, but 'and' sounded slightly better since 'or' can be inclusive or exclusive but 'and' is always inclusive.

I also changed the indentation of the items in upgrade.json to be more consistent and not be dependent on key length.

## Testing

Did not test, but these changes will be tested later as part of our general schema change testing. (I won't be testing any of the older schema upgrades, though.)


yvanzo (approved):
The new names look good to me.

However the patches split and/or re-indent a lot of lines, but not all lines either. I would suggest either to not split/re-indent at all, or to do it for all lines in a first separate commit for a more useful history.

mwiencek:
> However the patches split and/or re-indent a lot of lines, but not all lines either. I would suggest either to not split/re-indent at all, or to do it for all lines in a first separate commit for a more useful history.

Agreed, I pushed the commit "Reformat upgrade.json" to run the whole file through `jq .` first.

yvanzo (approved):
I checked `musicbrainz-docker` too, it’s just using `upgrade.sh` as is.
:ship:

mwiencek:
I merged this to master and created a new `schema-change-2023-q2` branch off of master just now.

PR #2915, opened 2023-04-12
Specified the OS requirements

dertuxmalwieder:
# Problem

According to INSTALL.md, Ubuntu/Debian are not strictly required, even Solaris *could* work. The README contradicted INSTALL.md.

# Solution

Generalized the README.

# Checklist for author

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Full installation instructions are available in [INSTALL.md](INSTALL.md).
 
 General Prerequisites:
 
-* Ubuntu/Debian
+* a Unix-like operating system (e.g. Debian GNU/Linux or Ubuntu)
 * Node (at least version 16)
 * Perl (at least version 5.30)
 * PostgreSQL (at least version 12)
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Full installation instructions are available in [INSTALL.md](INSTALL.md).
 
 General Prerequisites:
 
-* a Unix-like operating system (e.g. Debian GNU/Linux or Ubuntu)
+* Ubuntu/Debian (or some other Unix-like operating system at your own risks)
 * Node (at least version 16)
 * Perl (at least version 5.30)
 * PostgreSQL (at least version 12)


README.md
```diff
@@ -41,7 +41,7 @@ Full installation instructions are available in [INSTALL.md](INSTALL.md).
 
 General Prerequisites:
 
-* Ubuntu/Debian
+* a Unix-like operating system (e.g. Debian GNU/Linux or Ubuntu)
```
yvanzo, line 44: That would be nice but it is untrue.

For example, it is known to **not** fully work on FreeBSD, from MBS-8041.

The `INSTALL.md` file mentions that _Mac OS X will work just fine, if you're prepared to potentially jump through some hoops_, which @mwiencek is doing for development, but it is not fully covered by the instructions.

However Ubuntu is the only platform that we use for production (so it is guaranteed to be compatible), that we recommend for mirrors and development setup through [our Docker Compose project](https://github.com/metabrainz/musicbrainz-docker), and that we provide full instructions for installing from source (which is not even recommended from [MusicBrainz Server / Setup](https://musicbrainz.org/doc/MusicBrainz_Server/Setup) doc).

At best, we can say that it is very likely to work on other Linux distributions, and possibly on other Unix-like systems, if you are able to help yourself.

yvanzo (commented):
Thanks for your time submitting this pull request.

P.S. I deleted “irrelevant parts of” the template.

README.md
dertuxmalwieder: None of the Unix-like operating systems that run on my servers support Docker. However, *in general*, Ubuntu and Debian are not “strict requirements” either. How could this be written so that it’s *true*?

yvanzo: We don’t support other platforms because we don’t all use these, so we are **not** be able to check that our patches would not break the compatibility with these other platforms.

For example, I cannot assert that it will work on Solaris. However patches that help with improving the overall portability are welcome. We can also accept separate platform-specific instructions such as a `INSTALL_SOLARIS.md` file for example.

For now, I would suggest to mention other platforms very cautiously:

```suggestion
* Ubuntu/Debian (or some other Unix-like operating system at your own risks)
```

dertuxmalwieder: Good point.

mwiencek (approved):
Makes sense. It works without much effort on macOS, and I've done it on BSD in the past mostly with success. If people don't mind getting it working on their own, then I don't see a reason not to mention that they can.

PR #2916, opened 2023-04-19
MBS-13035: Better "no results" message for CDTOC release search

reosarevok:
### Implement MBS-13035

# Problem
The release search when trying to attach a CDTOC is limited to the 300 releases best matching the entered name and only from those does it look for mediums with the right number of tracks. This can be confusing: see MBS-12971 where the editor knew the release existed, and did not understand why the search was not returning it.

# Solution
This changes the message to a more concrete one that specifies users should be as specific as possible, and also directs them to an indexed search they can refine if needed.

# Testing
Manually to ensure the search link works properly.

### DIFF
--- a/root/cdtoc/AttachCDTocToRelease.js
+++ b/root/cdtoc/AttachCDTocToRelease.js
@@ -39,6 +39,9 @@ component AttachCDTocToRelease(
     ? lp('Move Disc ID', 'header')
     : lp('Attach CD TOC', 'header');
   const cdTocTrackCount = cdToc.track_count;
+  const searchLink = '/search?query=' +
+    form.field.query.value +
+    '&type=release&method=indexed';
 
   return (
     <Layout fullWidth title={title}>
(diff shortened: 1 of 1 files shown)


root/cdtoc/attach_filter_release.tt
```diff
@@ -61,7 +63,7 @@
             [%- IF was_mbid_search;
                   l('We couldn’t find a release matching that MBID.');
                 ELSE;
-                  l('No results found. Try refining your search query.');
+                  l('No results found. For performance issues, this search checks only a limited amount of releases closest to your entered query. To ensure a better result, search for the full release title or as close to it as possible. You can also {search_link|search manually} and paste the link to the release here.', { search_link => { href => search_link, target => '_blank' } });
```
yvanzo, line 66: For readability and to limit i18n breakage and redundancy, I suggest to keep the initial paragraph (for the short answer), and to add the new text in a separate paragraph (for the long answer).

About the added text itself now:

1. The first added sentence doesn’t really makes clear that the title is searched at first, and that other search conditions are tested on these limited amount of intermediate results.
2. The second added sentence is more about trying to work around the explained limitation than actually _ensuring a better result_.
3. The third and last added sentence provides a link for a search for release by name. It is not clearly worded. Would it be possible to pass some other conditions such as CD format and whatnot? See for example: https://musicbrainz.org/search?query=release%3A%28%22Never+Gonna+Give+You+Up%22%29+AND+format%3ACD&type=release&limit=25&method=advanced

yvanzo (commented):
This is conflicting with your own (currently draft but almost finished) pull request #2642 which converts this template to React.

It might also be simpler to create an advanced search link with escaped title using JavaScript as proposed in comment.

root/cdtoc/attach_filter_release.tt
reosarevok: For 3. we could, but the system doesn't map entirely great right now, because a) we don't always know it's a CD (other formats also support discids, including no format set) and b) the main thing we could use to search (number of tracks) clashes because we want number of *cdtoc* tracks (without data tracks) but indexed search just has total number :)

reosarevok: I reworded this a bit, at least (and moved it to the new React page)

root/cdtoc/AttachCDTocToRelease.js
```diff
@@ -121,13 +124,31 @@ component AttachCDTocToRelease(
...
+                  <p>{l('We couldn’t find a release matching that MBID.')}</p>
+                ) : (
+                  <>
+                    <p>
+                      {l('No results found. Try refining your search query.')}
+                    </p>
+                    <p>
+                      {l(
+                        `For performance reasons, this search only checks the
+                         apropriateness of a limited amount of releases with
```
mwiencek, line 137:
```suggestion
                         appropriateness of a limited amount of releases with
```

mwiencek (commented):
This looks fine to merge until we investigate the performance issue more closely.

mwiencek (approved):

PR #2917, opened 2023-04-20
MBS-12749: Convert cdtoc/attach_filter_artist to React

reosarevok:
### Implement MBS-12749

# Comments
* Renamed the file to `SelectArtistForCDToc` since you don't actually attach from this component (compare `AttachCDTocToRelease` in https://github.com/metabrainz/musicbrainz-server/pull/2642).
* We might want to change the old-school direct search here to an actual autocomplete using indexed search. That would also solve MBS-12974. It's not trivial (IIRC we didn't have documentation for how to do this?) so I haven't done it as part of the conversion. It would probably also need a way to actually check the artist pages, since we now can "click the artist’s name to get more info".

# Testing
Searched for "Sleater Kinney" (one result, no paging), "Slater" (bunch of results, two pages) and "Slaterrrrrrr" (no results). Also tried the Add release link to ensure it seeds the release editor with discid info *and* the release artist string.


yvanzo (approved):
LGTMBDNT :ship:

PR #2918, opened 2023-04-22
Increased default font size and modified home page

santiagofn:
# Problem

- Font size of 12px is too small. It hurts readability and thus accessibility. In fact,[ the MetaBrainz system design proposes to use a font size of 16px](https://github.com/metabrainz/design-system/blob/master/guidelines/design-guidelines.md#accessibility) (although with a different font family).
- The home page's organization of the content doesn't work on mobile and has some technical problems in its implementation.

# Solution

- Changed default font size to 14px. Manually tested navigating the site, I haven't seen anything broken.
- Home page:
  - Used CSS grid to organize its content and media queries to better adapt to smaller screens (still not there yet, thanks menu)
  - Moved the style file next to the .js one (it's the default in React/Angular/Vue projects and it's a mental step on the future change of using CSS isolation)
  - Cleanup of styles and divs

## Screenshots (desktop)

Before
imagen

After
imagen

Before
imagen

After
imagen

### Screenshots (mobile)

Before
imagen

After
imagen

imagen

# Checklist for author

* [x] Read the [Contributing Guidelines](https://github.com/metabrainz/musicbrainz-server/blob/master/CONTRIBUTING.md)
* [x] Delete irrelevant parts of this template

PR #2920, opened 2023-04-24
MBS-13042: Increase default font size (12 -> 14px)

reosarevok:
### Implement MBS-13042

# Problem
A fair amount of users (including me) seem to always use MBS zoomed in to at least 110% just because of how small the default font is. We should have a more user friendly default font size that makes zooming unnecessary for at least a fair amount of users.

# Solution
Apparently 16px is the recommended accesibility default, but given our current design might not accommodate it too well, this changes the default from 12px to 14px for now, which should already be fairly noticeable.

As a note, rem values in here are based on HTML font size, which is @default-text-size (set in layout.less).

I used @santiagofn's approximated pixel size comments from https://github.com/metabrainz/musicbrainz-server/pull/2918 rather than calculating my own.

# Testing
Moved around a bit locally (found that buttons were still small, added a commit for that). Put this on test.mb so we can check around more, and we should then put it on beta if it seems to work so that someone will inevitably find some issues :)


reosarevok:
@brainzbot: retest this please

PR #2921, opened 2023-04-24
MBS-13041: Fix broken Utaten favicon

reosarevok:
### Fix MBS-13041

# Problem
Utaten favicons are broken.

# Solution
We had a 32 px link but we had not specified 32 on favicons.less, so we were trying to link to a -16.png file. This just specifies that.

# Testing
Manually by checking the favicon now appears when adding a random link to https://utaten.com/lyric/mi23041713/ to a work in sample data.


yvanzo (approved):
:ship:

PR #2922, opened 2023-04-27
MBS-13045: Display artist rename relationship on overview

reosarevok:
### Implement MBS-13045

# Problem
We have a renamed relationship for artists now, but we don't display it on the artist overview (like legal name and other identities). We do show it for label, which also has renamed.

# Solution
This just effectively copies the label code to display the same for artist. Random fake example:
image_2023-04-27_092758443

# Testing
Manually by adding a random rel on sample data (see image above).


yvanzo (approved):
Proof-read the code by comparing it with the same code for label rename relationship.

Tested similarly (on a random pair of artists in sample data).

Some CI test would be nice to have but not a showstopper as this is a small display feature and this page is under redesign anyway.

:ship:

PR #2925, opened 2023-04-27
MBS-13049: Add Uta-Ten songWriter links to artist

reosarevok:
### Implement MBS-13049

# Issue
Uta-Ten has `/songWriter` pages that are separate from `/artist`, but they should probably still be linkable to MB artist pages.

# Testing
Added a test to the handling tests suite.


yvanzo (approved):
Code looks good to me, and test coverage seems sufficient. :ship:

PR #2928, opened 2023-05-02
MBS-13052: Support intl-XX/ path segment in Spotify URLs

reosarevok:
### Implement MBS-13052

# Problem
Apparently some languages are now adding a segment like intl-de or intl-it to the Spotify URLs. These redirect to the basic URL if the site is set to English, and the basic URL to them if not. As such, the easiest way to deal with them seems to be to just drop them in cleanup.

# Solution
This just drops the new path segments in cleanup, We need to also support them when checking whether a URL is `spotify` or `spotifyuseraccount`, of course.

# Testing
Added / modified a test for `spotify` and one for `spotifyuseraccount`.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -4524,13 +4524,14 @@ const CLEANUPS: CleanupEntries = {
   },
   'spotify': {
     match: [new RegExp(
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
Changes make sense, code looks good, and CI tests have a decent coverage.
C, C & C :speedboat:

PR #2929, opened 2023-05-04
MBS-9396: Properly parse seeded release group URLs

reosarevok:
### Fix MBS-9396

# Problem
Seeding URLs for the release group editor was not working at all - no error, nothing, just ignored completely. This has apparently been broken since at least 2017.

# Solution
I figured out that the seeded links for RGs were not getting parsed at all right now because of the wrong sourceType string expectation. This is yet another victim of the "release_group" vs "release-group" clash, where we need to special-case everything. Sigh. I added such special-casing now.

# Testing
Manually, by comparing `/release-group/create?edit-release-group.url.0.text=http://www.example.com/&edit-release-group.url.0.link_type_id=1` with `/artist/create?edit-artist.url.0.text=http://www.example.com/&edit-artist.url.0.link_type_id=1` and checking the seeding result is the same for both now.

Added a basic RG creation test to match other entity types (but that does not allow testing URL seeding) plus a tiny Selenium test to test URL seeding.

# Comments
This is very much not the first time this kind of thing has bitten us in the ass. Is there a real reason why we have to use `release_group` in some places and `release-group` in others? (and, IIRC, `rg` in yet some others, also special-cased).

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -179,8 +179,11 @@ export class _ExternalLinksEditor
           }
         }
       } else {
+        const seededSourceType = sourceType === 'release_group'
+          ? 'release-group'
+          : sourceType;
         const seededLinkRegex = new RegExp(
-          '(?:\\?|&)edit-' + sourceType +
+          '(?:\\?|&)edit-' + seededSourceType +
             '\\.url\\.([0-9]+)\\.(text|link_type_id)=([^&]+)',
           'g',
         );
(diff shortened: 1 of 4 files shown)


yvanzo (commented):
This form is missing CI tests under `t/lib/t/MusicBrainz/Server/Controller/ReleaseGroup/Create.pm` whereas some other entity creation forms have CI tests.

yvanzo:

yvanzo:

yvanzo:
Both runs failed on the added test:
> TimeoutError: Waiting for element to be located By(css selector, #external-link-0.td:nth-child(3) input)

yvanzo (approved):
LGTM and the added tests’ coverage and documentation are convincing.

Thank you very much! :ship:

> Is there a real reason why we have to use `release_group` in some places and `release-group` in others?

Those two characters are very useful to describe our mood when we find out such cause for a bug: `-_-`

PR #2931, opened 2023-05-05
MBS-7385: Use EntityLink over release name in medium append UI

reosarevok:
### Implement MBS-7385

# Problem
We were just printing "`was medium 1 on release Whatever`" on the UI for picking medium positions and titles when merging releases and appending mediums. This means release disambiguations were not shown here, even though they can be useful when merging different releases of the same or a similar title.

# Solution
This patch changes the text name to an `EntityLink` for the release, which both shows disambiguations and provides
an actual link to the release itself in case the user wants to reopen it to check something.

# Testing
Manually, to make sure the disambiguation now appears (compare below).

Old:
Screenshot from 2023-05-05 11-57-53

New:
image_2023-05-05_122653983

### DIFF
--- a/root/static/scripts/edit/components/ReleaseMergeStrategy.js
+++ b/root/static/scripts/edit/components/ReleaseMergeStrategy.js
@@ -129,7 +129,11 @@ const ReleaseMergeStrategy = ({
                           {
                             name: medium.name,
                             position: medium.position,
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
A 9yo issue resolved with such a straightforward change! :bowling: 

LGTMBDNT :ship:

PR #2934, opened 2023-05-08
MBS-13057 / MBS-13058: Improve the application/revoke-access page

reosarevok:
### MBS-13057 / MBS-13058

# Problem
Right now once the user clicks through to the `revoke-access` page, there's no longer any info printed about what access they are revoking.

The user is expected to click blindly to revoke access to whatever it is and might end up revoking the wrong thing if they misclicked in the applications page.

Additionally, the user can modify the URL to end up trying to revoke something completely different (which is more likely to ISE than not since it will ISE unless they guess an existing application ID and scope for their tokens).

# Solution
This just lists the application name and its permissions when asking for confirmation for the revoking.

For the made-up URLs ISE, given we already have the `check_granted_token` method which takes the same parameters, we can check whether there's something to delete before we render the form. I chose a basic 404 error with custom message rather than a `NotFound` page because the only way a user should get here anyway is if they edit the URL by hand so it's probably not worth a better error really.

# Testing
Manually, making sure the page 404s when it should and displays the app/permission data when it shouldn't (before/after for that below).

Screenshot from 2023-05-08 18-07-00
image_2023-07-24_225903016

# Notes
The first commit renames the files (and one component) under `/applications` because as one of the first sets of files converted to React they didn't match the way we usually do this elsewhere nowadays.


reosarevok:

lib/MusicBrainz/Server/Controller/Account.pm
```diff
@@ -752,6 +752,17 @@ sub revoke_application_access : Path('/account/applications/revoke-access') Args
...
     my $form = $c->form( form => 'Confirm' );
+
+    my $token_exists = $c->model('EditorOAuthToken')->check_granted_token(
+        $c->user->id,
+        $application_id,
+        $scope,
+    );
+    $c->detach(
+        '/error_404',
+        [ l('There is no OAuth token with these parameters.') ]
```
yvanzo, line 763: About the commit message rather than this line of code:

> the only way a user should get here anyway is if they edit the URL by hand

or if they reopened a tab revoking an already revoked token.

root/account/applications/RevokeApplicationAccess.js
```diff
@@ -8,18 +8,32 @@
...
     form={form}
-    question={l(
-      `Are you sure you want to revoke this application's access?`,
+    question={texp.l(
+      `You’re about to revoke {app}’s permissions to: {permissions}.
+       Are you sure you want to revoke this application's access?`,
+      {
+        app: application.name,
+        permissions: commaOnlyListText(permissions.map(permission => (
+          ACCESS_SCOPE_PERMISSIONS[permission]()
```
yvanzo, line 34: Permissions can be pretty long to read. It would probably easier to have these as an itemized list since this page is nearly empty and can afford some more lines, or at the very least in italic (**untested** suggestion).

```suggestion
        permissions: <i>{commaOnlyListText(permissions.map(permission => (
          ACCESS_SCOPE_PERMISSIONS[permission]()}</i>
```

root/account/applications/RevokeApplicationAccess.js
```diff
@@ -8,18 +8,32 @@
...
   form,
+  permissions,
 }: Props): React$Element<typeof ConfirmLayout> => (
   <ConfirmLayout
     form={form}
-    question={l(
-      `Are you sure you want to revoke this application's access?`,
+    question={texp.l(
+      `You’re about to revoke {app}’s permissions to: {permissions}.
+       Are you sure you want to revoke this application's access?`,
```
yvanzo, line 30: The first (new) sentence is using typographic apostrophe, so the second one should now use the same.
```suggestion
       Are you sure you want to revoke this application’s access?`,
```

yvanzo (changes requested):
Please make the list of permissions somehow more readable.

root/account/applications/RevokeApplicationAccess.js
reosarevok: I thought we were not changing apostrophes when it'd break translations?

reosarevok: Oh, wait, I'm breaking it anyway - although with your new suggestion maybe I could not do so, I'll check.

root/account/applications/RevokeApplicationAccess.js
reosarevok: That's a very good point, and I didn't think about the fact that `question` supports proper React anyway, so I amended it, this seems a lot better to me:

image_2023-07-24_225730694

yvanzo: “ListenBrainz Prod” looks like a label name :smile:

yvanzo (approved):
Nice! :ship:

PR #2936, opened 2023-05-11
MBS-13079: Add statistics for release groups with no type set

reosarevok:
### Implement MBS-13079

# Problem
For some reason, while we were calculating no-type stats for other types we were not doing it for RG primary nor secondary types.

# Solution
This just follows the general style we use for other entity types (more specifically, I based it on area_type).

# Testing
Generated stats locally. For secondary types, where the query is less obvious, I made sure the "None" result in a sample db matched the result of:

```
SELECT count(*) FROM release_group rg WHERE NOT EXISTS (
    SELECT 1 FROM release_group_secondary_type_join rgj WHERE rgj.release_group = rg.id
)
```

### DIFF
--- a/lib/MusicBrainz/Server/Data/Statistics.pm
+++ b/lib/MusicBrainz/Server/Data/Statistics.pm
@@ -1117,12 +1117,13 @@ my %stats = (
         CALC => sub {
             my ($self, $sql) = @_;
 
-            my $data = $sql->select_list_of_lists(
-                'SELECT type.id, COUNT(rg.id) AS count
-                 FROM release_group_primary_type type
-                 LEFT JOIN release_group rg ON rg.type = type.id
-                 GROUP BY type.id',
-            );
+            my $data = $sql->select_list_of_lists(<<~'SQL');
+                         SELECT COALESCE(type.id::text, 'null'),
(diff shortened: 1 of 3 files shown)


lib/MusicBrainz/Server/Data/Statistics.pm
```diff
@@ -1138,14 +1139,16 @@ my %stats = (
         CALC => sub {
             my ($self, $sql) = @_;
 
-            my $data = $sql->select_list_of_lists(
-                'SELECT type.id, COUNT(rg.id) AS count
-                 FROM release_group_secondary_type type
-                 LEFT JOIN release_group_secondary_type_join type_join
```
reosarevok, line 1144: I can't see why this used to be a left join, can you?

mwiencek: I suppose in case a secondary type is completely unused, it'll log 0 uses?  Probably only relevant on standalone/test DBs.

reosarevok: I guess that might be it. I'll make it a left join again then I guess.

mwiencek (approved):

PR #2937, opened 2023-05-12
MBS-12107: Enable replication with dbmirror2 by default

mwiencek:
* `REPLICATION_USE_DBMIRROR2` has been flipped to `1` in lib/DBDefs/Default.pm, which (1) enables dumping of v2 packets by default in ExportAllTables, and (2) enables the downloading of v2 packets by default in LoadReplicationChanges.

 * Scripts that accept a `--dbmirror2` flag (ImportReplicationChanges, InitDb.pl, ProcessReplicationChanges) now have that flag default to true. `--nodbmirror2` can be specified where it needs to be disabled. (Currently the JSON dumps and sitemaps generation still use the old packets.)

### DIFF
--- a/lib/DBDefs/Default.pm
+++ b/lib/DBDefs/Default.pm
@@ -50,9 +50,9 @@ sub REPLICATION_TYPE { RT_STANDALONE }
 sub REPLICATION_ACCESS_TOKEN { '' }
 
 # On RT_MASTER, setting this to 1 generates '-v2' packets from dbmirror2
-# (in addition to normal dbmirror packets). On RT_SLAVE, downloads the v2
-# packets when running replication.
-sub REPLICATION_USE_DBMIRROR2 { 0 }
+# (in addition to the old "v1" dbmirror packets). On RT_MIRROR, downloads the
+# v2 packets when running replication.
+sub REPLICATION_USE_DBMIRROR2 { 1 }
 
 ################################################################################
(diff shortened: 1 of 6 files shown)


mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

lib/DBDefs/Default.pm
```diff
@@ -52,7 +52,7 @@ sub REPLICATION_ACCESS_TOKEN { '' }
 # On RT_MASTER, setting this to 1 generates '-v2' packets from dbmirror2
 # (in addition to normal dbmirror packets). On RT_SLAVE, downloads the v2
```
reosarevok, line 53: If dbmirror2 is going to be the "normal" going forward, maybe we should change this comment :)

reosarevok (approved):
LGTMBDNT

reosarevok:
Why do "JSON dumps and sitemaps generation still use the old packets"? Is it non-trivial to switch?

mwiencek:
> Why do "JSON dumps and sitemaps generation still use the old packets"? Is it non-trivial to switch?

It probably won't be super difficult, but it's non-trivial at least, since we actually extract and parse the files inside the packet to read the changes.

yvanzo (approved):
LGTMBDNT

PR #2938, opened 2023-05-13
MBS-13087: Change bitwise AND to OR for combining privileges

the-good-boy:
### Fix MBS-13087

The ``PUBLIC_PRIVILEGE_FLAGS `` has been wrongly constructed using the bitwise AND operator. It should be constructed using the bitwise OR operator instead.

### DIFF
--- a/lib/MusicBrainz/Server/Constants.pm
+++ b/lib/MusicBrainz/Server/Constants.pm
@@ -358,12 +358,12 @@ Readonly our $EDITING_DISABLED_FLAG         => 1024;
 Readonly our $ADDING_NOTES_DISABLED_FLAG    => 2048;
 Readonly our $SPAMMER_FLAG                  => 4096;
 # If you update this, also update root/utility/sanitizedEditor.js
-Readonly our $PUBLIC_PRIVILEGE_FLAGS        => $AUTO_EDITOR_FLAG &
-                                               $BOT_FLAG &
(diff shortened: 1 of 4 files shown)


kellnerd (approved):
LGTM as per [my comment on IRC](https://chatlogs.metabrainz.org/libera/bookbrainz/msg/5173122/) after going through the related code.
This bug probably went unnoticed so far because the privileges of a sanitized editor aren't used anywhere throughout the codebase.
The privileges only seem to be of relevance in places where an unsanitized editor instance is used, e.g. https://github.com/metabrainz/musicbrainz-server/blob/e6cf1f7377e76f7e9dfabe11d69f6dbbe8c49800/root/user/UserProfile.js#L57

yvanzo:
@brainzbot, please add to whitelist.

reosarevok:
This seems sensible to me. We should probably be testing this somewhere, not quite sure where though :) (seems there's no tests for `sanitizedEditor` currently?)

reosarevok:
Pushed a second commit that adds an actual test for `sanitizedEditor` (including a test for auto-editor privs that failed before your commit but passes now).

reosarevok (approved):

reosarevok:
@kellnerd did miss one use of the privileges that meant we actually had a small display bug, which this PR now fixes :)

mwiencek (approved):

PR #2939, opened 2023-05-15
MBS-13083: Fix code comment

yvanzo:
# Problem MBS-13083

Minor display regression since `v-2023-05-02`: A code comment is wrongly displayed in the form to remove entities.

# Solution

Fix the comment syntax by moving the template directive end tag marker `-%]` to the next line.

# Testing

Tested manually by opening the removal form for a recording from the sample database: http://localhost:5000/recording/11e9d7dd-0ca9-42aa-bd2d-7b69ecd12003/delete

### DIFF
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -377,7 +377,8 @@
   <div id="external-links-editor-container"></div>
 [%- END -%]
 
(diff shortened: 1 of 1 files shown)


reosarevok (approved):
My bad, thanks!

PR #2940, opened 2023-05-15
Bump DB_SCHEMA_SEQUENCE, ACTIVE_SCHEMA_SEQUENCE to 28

mwiencek:
# Problem

`DB_SCHEMA_SEQUENCE` should point to 28 following the schema change today.

`ACTIVE_SCHEMA_SEQUENCE` is currently unused, but should also point to 28.

# Solution

Bump both of these values to 28.

Last year's commit is at https://github.com/metabrainz/musicbrainz-server/commit/0b7ff5556524e68371a38734eee6d86d18ceffee in case you want to compare.

# Testing

No testing was done.

### DIFF
--- a/docker/config/services/dev.musicbrainz-server.json
+++ b/docker/config/services/dev.musicbrainz-server.json
@@ -1,7 +1,7 @@
 {
   "DBDefs": {
     "CATALYST_DEBUG": "0",
-    "DB_SCHEMA_SEQUENCE": "27",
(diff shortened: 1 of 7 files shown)


yvanzo (approved):
Double-checked other `27` occurrences.

We forgot updating schema diagrams.
They will have to be updated later on are we don’t have time now.

LGTMBDNT otherwise! :ship:

PR #2941, opened 2023-05-15
Fix innocuous warning in LoadReplicationChanges

yvanzo:
# Problem

Small regression in `master`/`beta`/`production`

The following warning was output when running this script:

    print (...) interpreted as function
    at ./admin/replication/LoadReplicationChanges line 357.

# Solution

Drop unneeded parenthesis.

# Testing

* [x] Ran on local mirror

# Action

- Propagate to `beta` and `production` branches (used for tagging today’s release).
- No need to deploy in containers though.


mwiencek (approved):
LGTMBDNT.

PR #2942, opened 2023-05-16
LoadReplicationChanges: Exit code for schema sequence mismatch

mwiencek:
# Problem

After LoadReplicationChanges reaches the first packet of the next schema sequence, it outputs "You must upgrade your database in order to apply this replication packet" and exits with exit code 1. This in turn causes the incremental JSON dump scripts to fail, with admin/RunIncrementalJSONDump not rsyncing any dumps created prior to the failure.

# Solution

To avoid this, I'm using a different exit code (3) to indicate a schema sequence mismatch, and then ignoring this code in `JSONDump::Incremental`.

# Testing

I didn't test the changes at all because it's hard to reproduce without another schema-27 JSON dumps setup.  So hopefully they are straightforward.

### DIFF
--- a/admin/replication/LoadReplicationChanges
+++ b/admin/replication/LoadReplicationChanges
@@ -244,7 +244,7 @@ unless ($SCHEMA_SEQUENCE == $iSchemaSequence)
         $iSchemaSequence,
         ;
     print localtime() . " : You must upgrade your database in order to apply this replication packet\n";
-    exit 1;
+    exit 3;
 }
 
 # Read REPLICATION_SEQUENCE.  Check that it matches the one we thought we'd
--- a/lib/MusicBrainz/Script/JSONDump/Incremental.pm
+++ b/lib/MusicBrainz/Script/JSONDump/Incremental.pm
@@ -192,8 +192,17 @@ sub run_impl {
                     @replicate_args = (qw( carton exec -- ), @replicate_args);
                 }
 
-                system(@replicate_args) == 0
-                    or die "Replication failed (exit code $CHILD_ERROR)";
+                system(@replicate_args);
+                my $exit_code = $CHILD_ERROR >> 8;
+                # An exit code of 3 indicates a schema sequence mismatch, i.e.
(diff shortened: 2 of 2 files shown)


admin/replication/LoadReplicationChanges
```diff
@@ -244,7 +244,7 @@ unless ($SCHEMA_SEQUENCE == $iSchemaSequence)
         $iSchemaSequence,
         ;
     print localtime() . " : You must upgrade your database in order to apply this replication packet\n";
-    exit 1;
+    exit 3;
```
yvanzo, line 247: Same as in commit 4f53741a59307f597c2e7274464f6f8cc780d7dc can be useful here:

```suggestion
    # See NOTE-SEQM-1 in Script::JSONDump::Incremental
    exit 3;
```

lib/MusicBrainz/Script/JSONDump/Incremental.pm
```diff
@@ -192,8 +192,17 @@ sub run_impl {
                     @replicate_args = (qw( carton exec -- ), @replicate_args);
                 }
 
-                system(@replicate_args) == 0
-                    or die "Replication failed (exit code $CHILD_ERROR)";
+                system(@replicate_args);
+                my $exit_code = $CHILD_ERROR >> 8;
+                # An exit code of 3 indicates a schema sequence mismatch, i.e.
```
yvanzo, line 197: This goes with the above.

```suggestion
                # NOTE-SEQM-1:
                # An exit code of 3 indicates a schema sequence mismatch, i.e.
```

yvanzo (commented):
The comment is very explanatory but chances are it will be missed when reading/modifying `LoadReplicationChanges`. Suggested to use a key to refer to comment.

LGTMBDNT otherwise.

admin/replication/LoadReplicationChanges
mwiencek: Thanks!

yvanzo (approved):
:ship:

PR #2943, opened 2023-05-16
MBS-13085: Clarify "share email" report checkbox 

reosarevok:
### Implement MBS-13085

# Problem
We have a "Reveal your email" checkbox when reporting users, but the people who get this report have access to your email anyway, so this is misleading.

# Solution
Improved the wording as suggested by @Aerozol.

# Testing
Manually (trivial change).

### DIFF
--- a/root/user/ReportUser.js
+++ b/root/user/ReportUser.js
@@ -164,7 +164,7 @@ const ReportUser = ({
                 )}
               </p>
             }
-            label={l('Reveal my email address')}
+            label={l('Receive email updates about this report')}
             uncontrolled
           />
 


root/user/ReportUser.js
```diff
@@ -154,13 +154,11 @@ const ReportUser = ({
...
-                   regarding this report, you can uncheck the checkbox
-                   above.
+                  `If you don’t want our admins to contact you 
+                   regarding this report, uncheck the checkbox above.
                    <br />
-                   We recommend leaving it checked, so that you can be
-                   contacted if the report is resolved or the admins
-                   need more information.`,
+                   Allowing admins to contact you
+                   makes your report more likely to be resolved.`,
```
yvanzo, line 161: The previous version was more clear about our recommendation to not uncheck it.

The first sentence of the new version can be taken as an advice to uncheck it, the user might not even go to the second sentence.

If it absolutely needs to be simplified, I would suggest to swap sentences (and add a linking word such as “But” in between). This way it would start with what this is useful for, and then add that it is still possible to opt-out if wanted.

yvanzo (commented):
* :100: to the first commit;
* :question: to the second one.

root/user/ReportUser.js
reosarevok: We talked on IRC and realized there was a miscommunication between me, @Aerozol and @yvanzo about this bit, so dropped this commit for now. We can revisit this part later.

yvanzo (approved):
Acknowledged by @Aerozol on IRC: “just changing the first bit of text […], that was the important bit”.
:ship:

PR #2944, opened 2023-05-16
Reword: Use the same writing as in other component

yvanzo:
# Problem

A pair of confirmation messages were hardly shortened, using inappropriate typewriter apostrophe, needlessly duplicating messages to be translated.

# Solution

Use “Yes, I am sure” for all.

# Testing

Manually opened both pages locally using sample data.

### DIFF
--- a/root/admin/LockedUsernameUnlock.js
+++ b/root/admin/LockedUsernameUnlock.js
@@ -35,7 +35,7 @@ const LockedUsernameUnlock = ({
       <form method="post" name="confirm">
         <FormCsrfToken form={form} />
         <FormSubmit
-          label="Yes, I’m sure"
+          label="Yes, I am sure"
           name="confirm.submit"
           value="1"
         />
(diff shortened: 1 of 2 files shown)


root/admin/LockedUsernameUnlock.js
```diff
@@ -35,7 +35,7 @@ const LockedUsernameUnlock = ({
       <form method="post" name="confirm">
         <FormCsrfToken form={form} />
         <FormSubmit
-          label="Yes, I’m sure"
+          label="Yes, I am sure"
```
reosarevok, line 38: We might as well put this into l() then since it's an existing translation?

yvanzo: It’s an admin form though, we shouldn’t really translate it.

reosarevok: As you prefer :) I think we're very inconsistent right now with translating some and not others, so it'd be good to eventually drop those if we decide not to translate admin forms as a rule.

yvanzo: Maybe we should have a dummy `l_admin` that does nothing but prevents messages from being added to POT files.

reosarevok (approved):

yvanzo:
Selenium test failed with the usual unrelated async error, ignoring it as I tested the two pages manually.

PR #2945, opened 2023-05-16
Refactor ProcessReplicationChanges

yvanzo:
# Problem

Some variable names are not really telling much about this piece of code, leading sometimes to confusion. For example, see https://github.com/metabrainz/musicbrainz-server/pull/2919#discussion_r1183566776.

# Solution

In [`ProcessReplicationChanges`](https://github.com/metabrainz/musicbrainz-server/blob/v-2023-05-15-schema-change/admin/replication/ProcessReplicationChanges), rename the variables:
* [x] `$short_sql` into `$show_short_sql` (to avoid the confusion with any SQL connection)
* [x] `$long_sql` into `$show_long_sql` (to avoid the confusion with any SQL connection)
* [x] keep `$sql` (since it is used as a generic connection as usual)
* [x] `$sql2` into `$fetching_change_sql`
* [x] `$mirror` into `applying_change_sql`

### DIFF
--- a/admin/replication/ProcessReplicationChanges
+++ b/admin/replication/ProcessReplicationChanges
@@ -72,7 +72,7 @@ my $sql = Sql->new($c->conn);
 
 # Connection wrapper used only for looping, through a cursor, over each
 # pending_data change for the transaction (X) ID obtained from the above.
-my $sql2 = Sql->new($c->conn);
+my $fetching_change_sql = Sql->new($c->conn);
 
 # Connection wrapper used only for applying all of the changes obtained from the
 # above in a single but separate transaction for each transaction (X) ID.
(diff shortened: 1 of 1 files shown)


mwiencek:
Looks like `$sql` is used for a variety of things, while `$sql2` is only used for making a cursor to loop over the pending_data changes for a particular xid. Maybe change `$sql2` to `$cursor_sql` or `$pending_data_sql` or something?

I'm not sure what to call `$sql` at the moment. (In fact all three handles point to the mirror, so `$mirror` isn't exactly any more descriptive than `$sql`.)

yvanzo:
Thanks, it is probably fine to keep `$sql` if this is a generic connection as usual.
It just leaves renaming `$sql2` and `$mirror` with meaningful identifiers.

mwiencek:
OK, here's what I can gather:

 * `$sql` is used for looping over all of the ordered transaction IDs (also any any other misc. queries at the beginning/ending of the script).
 * `$sql2` is used for looping over each pending_data change for the transaction ID obtained from `$sql`.
 * `$mirror` is used for applying all of the changes obtained from `$sql2` in a single transaction (but a separate transaction for each transaction ID)

I tried asking ChatGPT to suggest some names for these for fun, and it came up with these:

 * `$sql` -> `$pending_txn_iterator`
 * `$sql2` -> `$change_set_iterator`
 * `$mirror` -> `$mirror_txn_applier`

Not terrible IMO.

yvanzo:
- Added code comments based on your explanations

And based on these:

- Renamed `$sql2` to `$fetching_change_sql`
- Renamed `$mirror` to `$applying_change_sql`

ChatGPT suggestions were not too bad but didn’t follow the naming convention for `Sql` objects in the rest of the code.

reosarevok (approved):
I think the comments make stuff a lot more clear and the names themselves are already a good improvement. Thanks!

Left one extra suggestion but if you don't agree just ignore :) 

Sorry for missing the request until now!

reosarevok:

mwiencek (approved):

PR #2946, opened 2023-05-17
MBS-13088: Standardize series type bubble

reosarevok:
### Fix MBS-13088

# Problem
The bubble for series types doesn't even appear at first as it should, and eventually appears empty.

# Solution
Rather than fidgeting with the non-standard implementation we had for series, I just changed it to match the other type bubbles (based it on `place`).

# Testing
Manually, by going to `/series/create` and checking the type docs appear.


yvanzo (approved):
Code changes look good. 
Tested with `/series/create` and non-empty `/series/d977f7fd-96c9-4e3e-83b5-eb484a9e6582/edit` given that the type of non-empty series can be changed only to keep gathering the same type of entity.

:ship:

PR #2948, opened 2023-05-18
Avoid area bubble crashing on missing iso_3166_1_codes

reosarevok:
# Problem
Area bubbles for artists (and I assume label too) are crashing when trying to fill them using the area data stored in the form since that is missing `iso_3166_1_codes`.

# Solution
Using `{this}` rather than `{this.toJSON()}` meant we were hitting the issue why we originally added the `iso_3166_1_codes` assignment to `toJSON`, so use `this.toJSON()` instead. In order to be able to display containment with `toJSON` we need to actively set it in `toJSON` as well.

# Testing
Manually with [the case reported by CatQuest](https://tickets.metabrainz.org/browse/MBS-8774?focusedCommentId=68134&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-68134) plus making sure the rest of areas (e.g. when searching) still worked fine.

### DIFF
--- a/root/static/scripts/common/entity.js
+++ b/root/static/scripts/common/entity.js
@@ -273,7 +273,12 @@ import MB from './MB.js';
     toJSON() {
       return Object.assign(
         super.toJSON(),
-        {iso_3166_1_codes: this.iso_3166_1_codes || []},
+        {
+          containment: this.containment || [],
+          iso_3166_1_codes: this.iso_3166_1_codes || [],
+          iso_3166_2_codes: this.iso_3166_2_codes || [],
+          iso_3166_3_codes: this.iso_3166_3_codes || [],
+        },
       );
     }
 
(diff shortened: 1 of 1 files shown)


root/static/scripts/common/entity.js
```diff
@@ -273,15 +273,18 @@ import MB from './MB.js';
     toJSON() {
       return Object.assign(
         super.toJSON(),
-        {iso_3166_1_codes: this.iso_3166_1_codes || []},
+        {
+          containment: this.containment || [],
+          iso_3166_1_codes: this.iso_3166_1_codes || [],
```
mwiencek, line 278: For safety we might want to add lines for the `*_2_*` and `*_3_*` properties too. Since, if those exist, they would be accessible from the instance but not the `toJSON` return value.

mwiencek (approved):
LGTMBDNT

PR #2949, opened 2023-05-19
MBS-12211: Avoid crash when URL has been removed

reosarevok:
### Fix MBS-12211

# Problem
`UrlList` reports were working as expected when the non-URL entity was gone, but were crashing when the URL itself was gone. 

# Solution
For a case where there's no URL, there's no point on showing the entity grouping at all, so this just returns the new `RemovedUrlRow` component in those cases.

# Testing
Manually, ensuring the reported `/report/DiscogsLinksWithMultipleArtists?filter=0&page=2` page now loads:
image_2023-05-19_152835349


yvanzo (approved):
LGTMDBNT :ship:

PR #2950, opened 2023-05-22
Split and improve JS utilities tests

reosarevok:
# Problem
All of our JavaScript utilities tests currently live in one large `/utilities.js` file. This causes at least two problems:
1) It's hard to see at a glance which files have no coverage at all (this means a fair amount of utilities are not tested and we don't even know it, leading to issues such as https://github.com/metabrainz/musicbrainz-server/pull/2938).
2) It's hard to follow and understand the large test file.

# Solution
This splits the large `/utilities.js` file into different files for each file we want to test, to keep a 1:1 mapping that makes it simpler to see the status of things at a glance than the alternative I could think of (thematic files such as `/utilities/dates.js` having all functions that deal with dates).

Since the new files, unlike the old one, use `@flow`, some changes are sometimes needed to actually satisfy Flow requirements. I've similarly fixed eslint issues on the files.

# Testing
The tests hopefully test themselves.

# Comment
The plan is to follow this refactoring with actually adding new tests for as yet untested files, but this PR is large as is :)


root/static/scripts/edit/utility/dates.js
```diff
@@ -11,9 +11,9 @@ import getDaysInMonth from '../../../../utility/getDaysInMonth.js';
 import parseInteger from '../../common/utility/parseInteger.js';
 
 type PartialDateWithStringsT = {
-  +day: StrOrNum | null,
-  +month: StrOrNum | null,
-  +year: StrOrNum | null,
+  +day?: StrOrNum | null,
+  +month?: StrOrNum | null,
+  +year?: StrOrNum | null,
```
mwiencek, line 16: these should be `+day: ?StrOrNum,` etc. to allow the property to be null or undefined (rather than allowing the property to be omitted, or null if present)

root/static/scripts/common/utility/formatDate.js
```diff
@@ -11,14 +11,14 @@ import ko from 'knockout';
 
 import {fixedWidthInteger} from './strings.js';
 
-function formatDate(date: PartialDateT | null): string {
+function formatDate(date?: PartialDateT | null): string {
```
mwiencek, line 14: Last time I checked, Flow doesn't correctly interpret `date?:` to mean that the argument can be omitted but can't be undefined (it allows passing undefined here), unlike the object type suggestion I made, but I'd still do the same as there for clarity, assuming the intent is the same. :)

```js
date: ?PartialDateT
```

root/utility/formatUserDate.js
```diff
@@ -152,8 +152,24 @@ type FormatUserDateOptions = {
   timezone?: string,
 };
 
+type formatUserDateContext = {
```
mwiencek, line 155: can we make the type name capitalized here like the rest?

root/static/scripts/edit/utility/dates.js
reosarevok: Shouldn't it be `+day?: ?StrOrNum` then? `dates.isDateValid({year: undefined, month: undefined, day: undefined})` and `dates.isDateValid({year: 2000})` are both expected to be `true` by the test.

root/static/scripts/common/utility/formatDate.js
reosarevok: I think in this case the intention is clearly to allow it being set as `undefined` rather than nothing at all being passed to `formatDate` so I made the change. The *partial date* sections though have the same issue as on my comment above.

root/static/scripts/edit/utility/dates.js
mwiencek: Yeah, `+day?: ?StrOrNum` is correct if both are valid.

reosarevok: I'll type them as the test expects for now. Is there a way to automatically let Flow search for unneeded undefined / null options, for the time when we have everything typed?

root/static/scripts/edit/utility/fullwidthLatin.js
```diff
@@ -15,11 +15,11 @@
  */
 const fullwidthLatinRegexp = /[\u3000\uFF01-\uFF60\uFFE0-\uFFE6]/;
 
-export const hasFullwidthLatin = function (str: string): boolean {
+export const hasFullwidthLatin = function (str?: string | null): boolean {
```
mwiencek, line 18: I would write the three of these as `str: ?string`.

mwiencek (approved):

PR #2951, opened 2023-05-24
MBS-13015: Tracks with only URL relationships don't show in the relationship editor

mwiencek:
# MBS-13015

## Problem

As the title says, tracks whose recordings have only URL relationships will be invisible in the release relationship editor. See the examples in the ticket.

In the code, appending such recordings to `recordingsWithNoRelationships` in the `else` branch of `if (recording.relationships?.length) ... else ...` is incorrect because URL relationships will be included in that length check.

## Solution

This commit modifies `updateRelationships` to return whether any updates actually occurred, which is a more direct way of determining whether any non-URL recording relationships were processed, without having to loop over them more than once.

## Testing

Added a Selenium test.


reosarevok:

reosarevok (changes requested):
Seems like you forgot to push the test? :) 

Tested locally and it seems to work fine, FWIW.

reosarevok (approved):

PR #2954, opened 2023-05-26
MBS-12552 (pre): Nail last "core" Flow.js types

yvanzo:
# Problem

In preparation for MBS-12552, a few occurrences of the generic types `Core*T` still remain.

# Solution

Finally drop these after being discussed at yesterday’s team meeting.

Also add a note to the JS class `CoreEntity` which will disappear along with knockout.js.

# Testing

Rely on CI Flow tests only.

### DIFF
--- a/entities.json
+++ b/entities.json
@@ -391,8 +391,7 @@
                 "extra_fks": {
                     "release_label": "label"
                 }
-            },
-            "manual": true
+            }
         },
         "report_filter": true,
         "reviews": true,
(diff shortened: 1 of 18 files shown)


reosarevok (approved):
The first commit here reminded me of an issue we found, so I fixed my own mess with 45a4efc8289b28446f83126837289b7efb687a5b.

Otherwise LGTMBDNT

yvanzo (commented):
Thanks for the additional patch. I just checked that it now matches Perl code by grepping for `Controller::Role::Delete`.

PR #2956, opened 2023-05-29
MBS-13104: Unblock and handle RYM interview links

reosarevok:
### Implement MBS-13104

# Problem
When setting validation for Rate Your Music we (probably inadvertently) blocked linking to other non-database content by Rate Your Music, mainly artist interviews they publish every once in a while (such as https://rateyourmusic.com/feature/sonemic-interview-vylet-pony/).

# Solution
This specifically ignores `/feature` links in the main RYM cleanup and adds a second one to autoselect them to interview for artist and block them elsewhere. I checked that the rest of features don't seem to be a good fit anywhere right now - we can always unlock and validate more options later if needed. I wasn't sure whether to autoselect only based on the `sonemic-interview-` start of the URL subsection, but I'm not confident enough that it will never change, so I didn't go there for now.

# Testing
Added a test, also tested manually by adding an interview link and an artist link to the same artist and making sure they work fine.

# Comments
I also changed "Rateyourmusic" to "Rate Your Music" in a second commit since they don't seem to use the weird caps stuff anymore.


lib/MusicBrainz/Server/Entity/URL/RateYourMusic.pm
```diff
@@ -1,11 +1,20 @@
...
 
 extends 'MusicBrainz::Server::Entity::URL';
 with 'MusicBrainz::Server::Entity::URL::Sidebar';
 
-sub sidebar_name { 'Rateyourmusic' }
+sub sidebar_name {
+    my $self = shift;
+
+    if ($self->url =~ m{^https?://(?:www.)?rateyourmusic.com/feature/}i) {
+        return l('Interview at Rate Your Music');
```
yvanzo, line 13: `Rate Your Music` should not be localizable as they don’t have any localized name.
```suggestion
        return l(
            'Interview at {external_website_name}',
            { external_website_name => 'Rate Your Music' }
        );
```
As a bonus, this message can be reused with other websites in the future.
(We should probably have the same with streaming and others.)

yvanzo (commented):
There is no `rateyourmusic.com/feature` URL in the database for now, so the first commit seems to be fine about allowing it for interview even if unsure.

Made a comment about new translatable message in second commit.

LGTMBDNT otherwise.

lib/MusicBrainz/Server/Entity/URL/RateYourMusic.pm
reosarevok: We discussed this on IRC but sadly while I liked the idea I realized it doesn't work for languages such as Estonian (probably neither for Finnish and other languages that modify word endings with cases) since those depend on the specific name of the site.

yvanzo (approved):
:ship:

PR #2958, opened 2023-05-31
MBS-13105 / MBS-13106: Fix ISEs on attribute deletion pages

reosarevok:
### Fix MBS-13105 / MBS-13106

# Problem
Trying to delete any language caused an ISE, because it was trying to check the no longer existing `language` column of the `work` table. Trying to delete any script not in use caused a different ISE (this ISE also applied to languages but was hidden there by the first one) since the code expected a non-existing `has_children` method.

# Solution
Fix the language `in_use` query to actually look at the `work_language` table. Add dummy `has_children` methods for language and script (which always return false).

# Testing
Manually, plus I added reasonably exhaustive tests for all three cases of language, script and standard attribute (the latter had no ISEs right now, but we might as well).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
+++ b/lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
@@ -1,6 +1,8 @@
 package MusicBrainz::Server::Controller::Admin::Attributes;
 use Moose;
 use namespace::autoclean;
+use utf8;
+
 use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 
 use MusicBrainz::Server::Translation qw( l );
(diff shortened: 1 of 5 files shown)


reosarevok:

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
```diff
@@ -116,7 +116,17 @@ sub edit : Chained('attribute_base') Args(1) RequireAuth(account_admin) SecureFo
...
+    my $attr = $c->model($model)->get_by_id($id)
+        or $c->detach(
+            '/error_404',
+            [ l(
+                'Found no attribute of type “{model}” with ID “{id}”.',
+                {
+                    id => $id,
+                    model => $model,
+                },
+            ) ],
```
yvanzo, line 128: This message is for admin only so it doesn’t need to be translated. **Untested** suggestion:

```suggestion
            [ "Found no attribute of type “$model” with ID “$id”." ],
```

yvanzo (approved):
Added tests are brilliant. Made a minor suggestion, LGTMBDNT otherwise :ship: 

Besides, maybe we should have edits for changes made to languages/scripts.

lib/MusicBrainz/Server/Controller/Admin/Attributes.pm
reosarevok: It works fine as long as we also `use utf8;` :) Will do that.

yvanzo (approved):
Thanks :ship:

PR #2959, opened 2023-06-01
MBS-13111 / MBS-13116: Hide list of autoeditor election voters during vote

reosarevok:
### Fix MBS-13111 / Implement MBS-13116

# Problem
The list of who has voted and when (even if with votes marked as "Private") being shown during the voting period has no actual benefit, and the drawback of allowing people who can see the vote tally (proposer, seconders and candidate) to figure out who voted what by comparing the vote tally with the vote list repeatedly upon changes (there's 4 Yes votes, then on refresh there's 4 Yes and 1 No, clearly the latest voter voted No).

The only real usefulness I can see in showing the voter list during the election period is that it allows voters to see what they have voted in case they want to change it, so that needs to be kept somehow.

Edit: another useful point was brought up in a ticket comment: since we don't show the total number of voters nor the number of abstentions in the tally, the voter list is the only way to know the amount of Abstain votes during the vote itself.

# Solution
This hides the voters list during the election period and indicates the current vote if any of the page visitor with a line above the voting buttons.

image_2023-06-01_141125826

The second commit adds "Total votes" and "Abstentions" to the tally, which allows people with tally viewing privs to still know the number of abstentions during the vote, and makes it so people don't need to do math in general to find out those numbers:

image_2023-06-01_151831048

# Testing
Manually, by creating an election in sample data and voting on it.

### DIFF
--- a/root/elections/ElectionDetails.js
+++ b/root/elections/ElectionDetails.js
@@ -22,6 +22,8 @@ type PropsT = {
 
 const ElectionDetails = ({election}: PropsT): React$MixedElement => {
   const $c = React.useContext(CatalystContext);
+  const voteCount = election.votes.length;
+  const abstainCount = voteCount - (election.yes_votes + election.no_votes);
   return (
     <>
       <h2>{l('Details')}</h2>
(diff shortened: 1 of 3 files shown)


yvanzo (approved):
@reosarevok president! Oops, too soon…

LGTMBDNT :shipit:

PR #2960, opened 2023-06-01
Remove Entity::CoreEntity

reosarevok:
# Description
We decided to get rid of `CoreEntity`, and all CoreEntity users are also `Relatable`, so merging it into `Relatable` and just extending `Entity` with that.

Doing this breaks the `sub name` in `Entity::URL`, (seems it gets overriden by `Role::Name`) so changing it to `has 'name'` with a `default`, which still works.

# Testing
Manually by browsing around a bit (which found the URL issue defined above). Putting this up to let CI tests run and find more issues if any.


yvanzo (approved):
LGTMBDNT

It makes it even easier to understand the connection between `name` in `Entity::Role::Name` and `Entity::URL`.

Thank you! :ship:

PR #2962, opened 2023-06-05
MBS-13120: Fix note-too-old check for modifications

mwiencek:
# MBS-13120

# Problem

The test "Can’t remove too old note" for `Controller::Edit::ChangeNote` started failing recently.  After investigating, it seems likely that the same check could fail in production too.

# Solution

According to
https://metacpan.org/pod/DateTime#$dt-%3Esubtract_datetime($datetime) (which from the source code appears to be what `-` is overloaded to), the returned duration "may have deltas for months, days, minutes, seconds, and nanoseconds." The key word being "may."  When I reproduced the test failure in `Controller::Edit::ChangeNote` locally, the duration object had values for months, minutes, and seconds, but not days.  This may vary depending on what time the test is run.

The correct API for this appears to be `delta_days`, which fixes the test for me locally.

# Testing

Checked that the failing test passes locally.


lib/MusicBrainz/Server/Controller/EditNote.pm
```diff
@@ -105,8 +105,8 @@ sub detach_if_cannot_change {
     }
 
     my $now = DateTime->now( time_zone => $edit_note->post_time->time_zone );
-    my $note_age = $now - $edit_note->post_time;
-    my $is_note_too_old = $note_age->{days} > 0;
+    my $note_age = $now->delta_days($edit_note->post_time);
+    my $is_note_too_old = $note_age->days > 0;
```
yvanzo, line 109: Doesn’t it prevent changing an edit note posted at 23:59:59 after 1 second?

yvanzo (commented):
I understand that the bug is both unlikely to be exploited and offer very limited opportunities anyway as changes would still be logged in the database. But it seems that the current patch might create another bug. More specific CI tests are probably needed.

lib/MusicBrainz/Server/Controller/EditNote.pm
reosarevok: If that's the case, then https://metacpan.org/pod/Date::Calc#3) should be useful.

The main issue with testing date differences is that it usually requires wrapping the Perl tests in a wrapper that sets a static date (we have one test that does that somewhere) but I guess we could do that :)

reosarevok (approved):
Seems sensible, thanks for the test too! Amended the commit to use ’ rather than ' like with other strings in that test.

yvanzo (approved):
LGTMBDNT :ship:

PR #2963, opened 2023-06-06
MBS-12026: Still change deleted editor votes if they lack privs

reosarevok:
### Fix MBS-12026

# Problem
We are being so respectful of the "Editing and voting disabled" flag that we are following it even when the editor is deleted and we want their votes to be changed to Abstain. The change to Abstain gets blocked because they don't have the right to vote at all, duh.

# Solution
This allows passing an `$override_privs` boolean to `enter_votes` which will skip checking whether the editor has voting rights. In order to pass that I needed to stop passing `@votes` as an array and pass an array ref instead, hence the other changes in this commit.

# Testing
Added a test for this situation, which didn't pass before the change but does now.


lib/MusicBrainz/Server/Data/Vote.pm
```diff
@@ -40,10 +40,10 @@ sub _column_mapping
 
 sub enter_votes
 {
-    my ($self, $editor, @votes) = @_;
+    my ($self, $editor, $votes, $override_privs) = @_;
```
mwiencek, line 43:
```suggestion
    my ($self, $editor, $votes, %opts) = @_;
```
It's a little more self-documenting if you accept `%opts` and then pass `override_privs => 1` at the call site.

lib/MusicBrainz/Server/Data/Vote.pm
```diff
@@ -59,9 +59,11 @@ sub enter_votes
             # not sufficient to filter the vote because the actual approval is happening elsewhere
             confess 'Unauthorized editor ' . $editor->id . ' tried to approve edit #' . $_->{edit_id};
         }
-        @votes = grep {
-            $_->{vote} == $VOTE_APPROVE || $edits->{ $_->{edit_id} }->editor_may_vote_on_edit($editor)
-        } @votes;
+        unless ($override_privs) {
```
mwiencek, line 62:
```suggestion
        unless ($opts{override_privs}) {
```

mwiencek (commented):
I was going to suggest just stripping the privileges from the editor object (since it's being deleted anyway), but I guess it also depends on `has_confirmed_email_address` and `is_bot`?

mwiencek (approved):

PR #2965, opened 2023-06-13
Add static.metabrainz.org to CSP header

mwiencek:
staticbrainz.org now redirects to static.metabrainz.org, which breaks CSP on "secure" forms (e.g. the login page). A side effect of that is that CSS, JS, and images can't load on these pages without fixing the header.

This doesn't adjust any irombook code for now.  As part of merging this, we should also update docker-server-configs to use `static.metabrainz.org` for `STATIC_RESOURCES_LOCATION`; but for now both domains are whitelisted in the header so that either case works while the configuration is being updated.

### DIFF
--- a/lib/MusicBrainz/Server.pm
+++ b/lib/MusicBrainz/Server.pm
@@ -692,13 +692,15 @@ sub set_csp_headers {
         'script-src',
         q('self'),
         qq('nonce-$globals_script_nonce'),
-        'staticbrainz.org'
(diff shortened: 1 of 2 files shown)


mwiencek:
I've deployed this on test.mb alongside the necessary DBDefs changes: try it out at https://test.musicbrainz.org/login (make sure you are logged out).

mwiencek:
@brainzbot test this please

PR #2967, opened 2023-06-14
MBS-13131: Move staticbrainz to static.metabrainz for instruments

reosarevok:
### Implement MBS-13131

# Description
We've changed the way the static resources work, so we need to update the expected URLs. The already existing URLs will be updated separately with a one-off bot run (MBBE-80).

# Testing
Just the basic (updated) URL test.


mwiencek (approved):

reosarevok:
@brainzbot retest this please

PR #2968, opened 2023-06-14
admin/run: fix log path and automatically create it

mwiencek:
# Problem

sitemaps cron has not been running for quite some time.  I'm not sure for how long, but at some point the logs volume was cleared, which breaks admin/run.

To check what was failing, I ran this inside the container:

```sh
root@0ca0886613a0:/home/musicbrainz/musicbrainz-server# ./admin/run admin/cron/daily-sitemaps.sh
run: '/root/log/daily-sitemaps' doesn't exist
Make sure to run in the appropriate container.
```

This is odd to me because the script is supposed to `sudo` as the musicbrainz user if it exists.  In that case it should also use /home/musicbrainz as the log path, since that's the user it's invoking the passed-in script as.

Next, I tried the same command as the musicbrainz user:

```sh
root@0ca0886613a0:/home/musicbrainz/musicbrainz-server# su musicbrainz
musicbrainz@0ca0886613a0:~/musicbrainz-server$ ./admin/run admin/cron/daily-sitemaps.sh
run: '/home/musicbrainz/log/daily-sitemaps' doesn't exist
Make sure to run in the appropriate container.
```

Since the log volume was cleared at some point, these directories don't exist, and the script fails.  This was never an issue before, because our old crontab automatically created the log path:

```
-10 0 * * * . /etc/container_environment.sh; mkdir -p $HOME/log/daily-sitemaps && cd $HOME/musicbrainz-server && carton exec -- admin/cron/daily-sitemaps.sh > $HOME/log/daily-sitemaps/$(date --utc +\%FT\%TZ).log 2>&1
```

It appears that the intent of the script is to error in case you're running the command inside the wrong container, but I think that rarely happens in practice.  What is clear is that we don't actively monitor the sitemaps cron logs for failures, so I'd much prefer the script be made more resilient.

# Solution

<del>
I've made the following changes to resolve these issues:

 * Storing the username we're invoking the command as in `$username` and putting logs in `~$username`.
 * Creating `$log_dir_path` instead of erroring.
 * Ensuring the created log path and file are owned by `$username`.
</del>

Address the case of missing log directories:

 * Creating the log directories when starting the container with an init script.
   (Keep erroring if any log directory is missing)

Address the case of running as `root`:

 * Exit with error if there is no `musicbrainz` user.
 * Run everything as `musicbrainz` user, including redirects and piped commands.
 * Use the HOME directory of `musicbrainz` user specifically.

# Testing

<del>
I've copied the changes into the musicbrainz-sitemaps container and checked that invoking `./admin/run admin/cron/daily-sitemaps.sh` works. (Only tested as the root user.)
</del>

See below comment https://github.com/metabrainz/musicbrainz-server/pull/2968#pullrequestreview-1613034283

### DIFF
--- a/admin/run
+++ b/admin/run
@@ -50,28 +50,27 @@ command_full_path=$(which "$1")
 command_verbatim=$1
 shift
 
-pre_command='carton exec --'
-username="$(id -u -n)"
-if id -u 'musicbrainz' &>/dev/null && [[ "$username" != 'musicbrainz' ]]
-then
-  pre_command="sudo --preserve-env --set-home -u musicbrainz -- $pre_command"
-  username='musicbrainz'
-fi
-
 if [[ $command_full_path =~ admin/cron/.*\.sh$ ]]
 then
   log_dir_name=$(echo "$command_full_path" | sed 's_^.*admin/cron/\(.*\)\.sh$_\1_')
 else
   log_dir_name='occasionally'
 fi
-
-log_dir_path="$(eval echo ~"$username/log/$log_dir_name")"
-$pre_command mkdir -p "$log_dir_path"
-
+log_dir_path="$HOME/log/$log_dir_name"
+if ! [[ -d "$log_dir_path" ]]
+then
+  echo >&2 "$SCRIPT_NAME: '$log_dir_path' doesn't exist"
+  echo >&2 "Make sure to run in the appropriate container."
+  exit 73 # EX_CANTCREAT
+fi
 log_timestamp=$(date --utc +%FT%TZ)
 log_file_path="$log_dir_path/$log_timestamp.log"
-# ensure log is owned by $username
-$pre_command touch "$log_file_path"
+
+pre_command='carton exec --'
+if id -u 'musicbrainz' &>/dev/null && [[ "$(id -u -n)" != 'musicbrainz' ]]
+then
+  pre_command="sudo --preserve-env --set-home -u musicbrainz -- $pre_command"
+fi
 
 echo "Running '$command_verbatim $*' and logging at '$log_file_path'..."
 if [[ -t 1 ]]
--- a/admin/run
+++ b/admin/run
@@ -65,6 +65,7 @@ fi
 
 if [[ $command_full_path =~ admin/cron/.*\.sh$ ]]
 then
+  # shellcheck disable=SC2001,SC2016
   log_dir_name=$(echo "$command_full_path" | sed 's_^.*admin/cron/\(.*\)\.sh$_\1_')
 else
   log_dir_name='occasionally'
--- a/admin/run
+++ b/admin/run
@@ -57,6 +57,12 @@ command_full_path=$(which "$1")
 command_verbatim=$1
 shift
 
+if [[ "$(id -u -n)" != 'musicbrainz' ]]
+then
+  exec sudo --preserve-env --set-home -u musicbrainz -- \
+    "${BASH_SOURCE[0]}" "$command_verbatim" "$@"
+fi
+
 if [[ $command_full_path =~ admin/cron/.*\.sh$ ]]
 then
   log_dir_name=$(echo "$command_full_path" | sed 's_^.*admin/cron/\(.*\)\.sh$_\1_')
@@ -74,11 +80,6 @@ log_timestamp=$(date --utc +%FT%TZ)
 log_file_path="$log_dir_path/$log_timestamp.log"
 
 pre_command='carton exec --'
-if [[ "$(id -u -n)" != 'musicbrainz' ]]
-then
-  pre_command="sudo --preserve-env --set-home -u musicbrainz -- $pre_command"
-fi
-
 echo "Running '$command_verbatim $*' and logging at '$log_file_path'..."
 if [[ -t 1 ]]
 then
--- a/admin/run
+++ b/admin/run
@@ -56,7 +56,7 @@ then
 else
   log_dir_name='occasionally'
 fi
-log_dir_path="$HOME/log/$log_dir_name"
+log_dir_path="/home/musicbrainz/log/$log_dir_name"
 if ! [[ -d "$log_dir_path" ]]
 then
   echo >&2 "$SCRIPT_NAME: '$log_dir_path' doesn't exist"
--- a/admin/run
+++ b/admin/run
@@ -43,6 +43,13 @@ then
   exit 64 # EX_USAGE
 fi
 
+if ! id -u 'musicbrainz' &>/dev/null
+then
+  echo >&2 "$SCRIPT_NAME: user 'musicbrainz' doesn't exist"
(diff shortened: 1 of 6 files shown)


admin/run
```diff
@@ -50,27 +50,28 @@ command_full_path=$(which "$1")
 command_verbatim=$1
 shift
 
+pre_command='carton exec --'
+username="$(id -u -n)"
+if id -u 'musicbrainz' &>/dev/null && [[ "$username" != 'musicbrainz' ]]
+then
+  pre_command="sudo --preserve-env --set-home -u musicbrainz -- $pre_command"
+  username='musicbrainz'
+fi
```
yvanzo, line 59: There is no installation requirement for having a `musicbrainz` user at the moment. For example `musicbrainz-docker` doesn’t have any even though it really should.

yvanzo (commented):
<del>The user change would make make `admin/run` unusable for all `musicbrainz-docker` and direct development installations not having a `musicbrainz` user.</del>

Running commands in the wrong container happened before and this is the reason for this check.

Another way out would be to make the container create all the appropriate log directories when starting up.

mwiencek:
> The user change would make make admin/run unusable for all musicbrainz-docker and direct development installations not having a musicbrainz user.

The help text of the script says "This script is meant to be used in production only."  So I don't think it's meant to be used in those environments (or the documentation is wrong). In production at least there will always be a musicbrainz user.

admin/run
mwiencek: This also only runs if the `id -u 'musicbrainz'` check passes, so I don't believe I added any new requirements to the script. If there's no musicbrainz user, it uses whatever name it got from `username="$(id -u -n)"`.

mwiencek:
> Running commands in the wrong container happened before and this is the reason for this check.
>
> Another way out would be to make the container create all the appropriate log directories when starting up.

I'm fine with having the container create the directories instead, though I think we should still fix the issues with running the script as root.

mwiencek:
> @yvanzo
[Partly revert](https://github.com/metabrainz/musicbrainz-server/pull/2968/commits/16b609b29431588c7f2f4a14c5ce3b0dc1e0d92f) https://github.com/mwiencek/musicbrainz-server/commit/e4b8517b06d2ed79b6922909b35034c23a91cabd

I think it reverts a bit too much, since I was trying to fix this issue:

> the script is supposed to sudo as the musicbrainz user if it exists. In that case it should also use /home/musicbrainz as the log path

but it looks like log_dir_path will again hardcode `$HOME` i.e. /root

yvanzo:
Fixed it in a simplified way in https://github.com/metabrainz/musicbrainz-server/pull/2968/commits/92ab423229a5b3eca8a6c69d32b609d42d6ed43f

mwiencek:
Thanks, the simplified fix looks good

Should we just have the script error if the musicbrainz user doesn't exist, then? (So, only support running as root or musicbrainz.)

> fixup 16b609b: Revert checking ownership

iirc without that the log files will be owned as root, since sudo doesn't apply to any of the piped/redirected commands - but maybe it can be simplified as above by hardcoding the ownership to musicbrainz

yvanzo:
> Should we just have the script error if the musicbrainz user doesn't exist, then? (So, only support running as root or musicbrainz.)

Now done in commit https://github.com/metabrainz/musicbrainz-server/pull/2968/commits/03d272be0726b6a0f828b7a42d304e95724e185d

> > fixup [16b609b](https://github.com/metabrainz/musicbrainz-server/commit/16b609b29431588c7f2f4a14c5ce3b0dc1e0d92f): Revert checking ownership
> 
> iirc without that the log files will be owned as root, since sudo doesn't apply to any of the piped/redirected commands - but maybe it can be simplified as above by hardcoding the ownership to musicbrainz

Thanks, I missed this issue. Just added commit https://github.com/metabrainz/musicbrainz-server/pull/2968/commits/f8592b8c07aa7da32733b52fabd77821608c7d69 to make the script run itself with musicbrainz user.

admin/run
```diff
@@ -50,13 +50,26 @@ command_full_path=$(which "$1")
...
+if ! id -u 'musicbrainz' &>/dev/null
+then
+  echo >&2 "$SCRIPT_NAME: user 'musicbrainz' doesn't exist"
+  echo >&2 "Make sure to run in a production container."
+  exit 67 # EX_NOUSER
+fi
+
+if [[ "$(id -u -n)" != 'musicbrainz' ]]
+then
+  sudo --preserve-env --set-home -u musicbrainz -- ${BASH_SOURCE[0]}
```
mwiencek, line 62: Thanks, I didn't think of that. :)

Does that need `exec`, or should the current script exit after this line?

yvanzo: I also forgot to path the arguments. All fixed in https://github.com/metabrainz/musicbrainz-server/pull/2968/commits/4c482cbbcc7a1259c46dd66c84d9162a55a3dfc7

mwiencek:
Regarding the create_log_directories script, it would be nice if we could run this automatically on container startup (since the advantage of creating the dirs in admin/run is that it didn't require any manual intervention).

For Docker images based on phusion/baseimage, like here, we can [place the script in `/etc/my_init.d`](https://github.com/phusion/baseimage-docker#running-scripts-during-container-startup).

However, since the script requires an argument, we'd also need to read the container type from somewhere. Perhaps set an environment variable like `MB_CONTAINER_TYPE` in the Dockerfile, and have the script read that?

yvanzo:
Rich idea, implemented in commit https://github.com/metabrainz/musicbrainz-server/pull/2968/commits/2854891545cd0f8b4246dbc417b631ba168417b5.

docker/templates/Dockerfile.json-dump.m4
```diff
@@ -13,6 +13,12 @@ COPY docker/musicbrainz-json-dump/crontab /var/spool/cron/crontabs/musicbrainz
 RUN chown musicbrainz:musicbrainz /var/spool/cron/crontabs/musicbrainz && \
     chmod 600 /var/spool/cron/crontabs/musicbrainz
 
+ENV MB_CONTAINER_TYPE json-dump
```
mwiencek, line 16: the script appears to use `MUSICBRAINZ_SERVICE` now (not sure which you prefer)

yvanzo: Not sure why I typed a different name in the script. Now fixed.

yvanzo (approved):
I tested it in `musicbrainz-sitemaps`: It correctly created the missing log directory `occasionally` on container. Running `admin/run env` in the container works for both users `root` and `musicbrainz`. The hourly cron has been successfully logged into the usual file. Created directories and files belong to `musicbrainz` user as expected.

:ship:

mwiencek:
Thanks for all the improvements, and for testing it. The code LGTM. 🚢

PR #2969, opened 2023-06-15
Block notes on add note

reosarevok:
# Description
Whether notes can be added was being checked only at the `insert_votes_and_notes` level, but the equivalent `enter_votes` checks at a lower level, which seems a lot safer (we have plenty of calls for `add_note` outside of that method). Just in case ModBot is not otherwise considered able to add notes for whatever reason, this gives ModBot free pass (since we very often call `add_note` for ModBot directly).

# Testing
Added a test for add_note specifically, made sure other tests still pass.

On top of https://github.com/metabrainz/musicbrainz-server/pull/2966 since that significantly changes the relevant test file.

### DIFF
--- a/lib/MusicBrainz/Server/Data/EditNote.pm
+++ b/lib/MusicBrainz/Server/Data/EditNote.pm
@@ -8,7 +8,11 @@ use MusicBrainz::Server::Email;
 use MusicBrainz::Server::Data::Utils qw(
     placeholders
 );
-use MusicBrainz::Server::Constants qw( :vote $LIMIT_FOR_EDIT_LISTING );
+use MusicBrainz::Server::Constants qw(
+    :vote
+    $EDITOR_MODBOT
+    $LIMIT_FOR_EDIT_LISTING
+);
 
 extends 'MusicBrainz::Server::Data::Entity';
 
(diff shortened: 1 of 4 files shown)


yvanzo:
It has to be rebased and conflicts resolved to be ready for review.

reosarevok:
I did that now. IIRC @mwiencek wasn't sure this was a good idea since it moves the check away from the controller, so requesting review to see if he thinks it'll be a problem.

lib/MusicBrainz/Server/Data/EditNote.pm
```diff
@@ -73,10 +77,17 @@ sub insert
 sub add_note
 {
     my ($self, $edit_id, $note_hash) = @_;
+
+    my $edit = $self->c->model('Edit')->get_by_id($edit_id)
+               or die "Edit $edit_id does not exist!";
+    my $note_editor_id = $note_hash->{editor_id};
+    my $is_modbot = $note_editor_id == $EDITOR_MODBOT;
+    my $note_editor = $self->c->model('Editor')->get_by_id($note_editor_id);
+    return unless $is_modbot || $edit->editor_may_add_note($note_editor);
```
mwiencek, line 86: This PR seems fine, but yes, ideally this is already prevented at the controller level. A check in the data layer is useful, but I would explicitly error, because it indicates that the controller (or other script) is doing something wrong. (In the suggestion below, I also skipped loading/checking `$note_editor` at all if we already know it's ModBot.)
```suggestion
    unless ($note_editor_id == $EDITOR_MODBOT) {
        my $note_editor = $self->c->model('Editor')->get_by_id($note_editor_id);
        $edit->editor_may_add_note($note_editor)
            or die "Editor $note_editor_id may not add a note to edit $edit_id";
    }
```
```

mwiencek (approved):

yvanzo (approved):
Moving the check  to the controller is a good move indeed!

LGTM and the test’s coverage and documentation are convincing.

Feel free to add a ticket if you think that editors need to be informed of this change.

:ship:

PR #2971, opened 2023-06-16
MBS-13094: Allow admins to browse edit note changes

reosarevok:
### Implement MBS-13094

# Problem
Admins should be able to see the edit history for any edit, since otherwise a user could, for example, insult another and then edit the note so that the email is sent but others don't see.

# Solution
This mostly follows what we do for annotations, but most of it does not support translated strings since it's admin-only. Translations are added for stuff that will show in otherwise translated pages.

### Edit view
image_2023-06-26_162838967

### History page
Screenshot from 2023-06-26 16-26-34

### Change page
Screenshot from 2023-06-26 16-26-53

# Checklist for author
* [x] Add tests for these pages

# Testing
Manually navigating around after changes, plus added a test file checking that the pages are blocked for non-admins and show the expected data for admins.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/EditNote.pm
+++ b/lib/MusicBrainz/Server/Controller/EditNote.pm
@@ -6,9 +6,12 @@ use namespace::autoclean;
 use DateTime;
 use List::AllUtils qw( first_index );
 use MusicBrainz::Server::Constants qw( $CONTACT_URL $EDITOR_MODBOT );
+use MusicBrainz::Server::ControllerUtils::JSON qw( serialize_pager );
 use MusicBrainz::Server::Data::Utils qw( load_everything_for_edits );
+use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array );
 use MusicBrainz::Server::Validation qw( is_database_row_id );
 use MusicBrainz::Server::Translation qw( l );
+use utf8;
 
 extends 'MusicBrainz::Server::Controller';
 with 'MusicBrainz::Server::Controller::Role::Load' => {
@@ -22,7 +25,12 @@ sub _load
 {
     my ($self, $c, $edit_note_id) = @_;
     return unless is_database_row_id($edit_note_id);
-    my $edit_note = $c->model('EditNote')->get_by_id($edit_note_id);
+    my $edit_note = $c->model('EditNote')->get_by_id($edit_note_id)
+        or $c->detach(
+            '/error_404',
+            [ "Found no edit note with ID “$edit_note_id”." ],
+        );
+
     $c->model('Editor')->load($edit_note);
     $c->model('EditNoteChange')->load_latest($edit_note);
     return $edit_note;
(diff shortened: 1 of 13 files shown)


reosarevok:

root/edit_note/EditNoteChange.js
```diff
@@ -0,0 +1,76 @@
...
+      {nonEmpty(change.reason)
+        ? <p>{change.reason}</p>
+        : <p className="small">{'No reason entered.'}</p>}
+      <p className="small">
+        <a href={`/edit-note/${change.edit_note_id}/changes`}>
+          {'See all changes for this edit note'}
+        </a>
+        {' / '}
+        <a href={noteUrl}>
+          {'Go back to the edit note'}
```
yvanzo, line 69: Edit note change history can be accessed from a list of edits, not necessarily from the edit page itself, so “back” isn’t necessarily correct.

```suggestion
          {'Go to the edit note'}
```

root/edit_note/EditNoteHistory.js
```diff
@@ -0,0 +1,104 @@
...
+        <EditNoteHistoryTable changes={changes} />
+      </PaginatedResults>
+    ) : (
+      <p>
+        {'This edit note has no change history.'}
+      </p>
+    )}
+    <p className="small">
+      <a href={noteUrl}>
+        {'Go back to the edit note'}
```
yvanzo, line 98: Same as above
```suggestion
        {'Go to the edit note'}
```

t/lib/t/MusicBrainz/Server/Controller/EditNote/NoteHistory.pm
```diff
@@ -0,0 +1,319 @@
...
+use utf8;
+
+use MusicBrainz::Server::Test qw( html_ok test_xpath_html );
+
+with 't::Context', 't::Mechanize';
+
+=head1 DESCRIPTION
+
+This checks that edit notes can be deleted and modified, and the restrictions
+on who and when can do so.
```
yvanzo, line 17: Probably an oversight here, the description is the same as for `ChangeNote`.
```suggestion
This checks that edit note history can be viewed by admins and only by admins.
```

yvanzo (changes requested):
Great stuff overall. Code doesn’t embed a single bit of old edit notes in usual edit pages and lists, so no risk at leaking data. Tested locally as admin, non-admin and logged out visitor. Checked CI tests coverage too. Just spotted a very few easy-to-fix points.

lib/MusicBrainz/Server/Controller/EditNote.pm
```diff
@@ -22,7 +25,15 @@ sub _load
...
-    my $edit_note = $c->model('EditNote')->get_by_id($edit_note_id);
+    my $edit_note = $c->model('EditNote')->get_by_id($edit_note_id)
+        or $c->detach(
+            '/error_404',
+            [ l(
+                'Found no edit note with ID “{id}”.',
+                { id => $edit_note_id },
+            ) ],
+        );
+
```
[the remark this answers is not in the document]
reosarevok: If an admin navigated by hand to a page like `/edit-note/20000000/changes`. Given it's admin only, I'll change this to not be translatable, like on another recent PR :)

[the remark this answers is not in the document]
yvanzo: Yes

yvanzo (approved):
:ship:

PR #2973, opened 2023-07-01
Special purpose language: No linguistic content (zxx)

jesus2099:
- Like `[Artificial (Other)]` (`qaa`)
- Like `[Multiple languages]` (`mul`)
- Like `[No lyrics]` (same `zxx`)

`No linguistic language` (`zxx`) should be renamed to `[No linguistic language]` (`zxx`) and should thus appear in release language list (select combo) after ~~`[Multiple languages]` (Frequently used)~~ `[Artificial (Other)]`, instead of between `Niuean` and `Nogai`.

I guess it also requires a Database edit of the stored label in languages table.

Maybe I am missing some files.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Controller/WS/js/Edit.pm
+++ b/t/lib/t/MusicBrainz/Server/Controller/WS/js/Edit.pm
@@ -117,7 +117,7 @@ test 'previewing/creating/editing a release group and release' => sub {
     like($html, qr/Vision Creation Newsun/, 'preview has release name');
     like($html, qr/limited edition/, 'preview has release comment');
     like($html, qr/4943674011582/, 'preview has barcode');
-    like($html, qr/No linguistic content/, 'preview has language');
+    like($html, qr/\[No linguistic content\]/, 'preview has language');
     like($html, qr/Symbols/, 'preview has script');
     like($html, qr/Official/, 'preview has release status');
     like($html, qr/1999-10-27/, 'preview has release date');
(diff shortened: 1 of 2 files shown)


reosarevok:
For what it's worth, the po files should not be changed by hand, and the language renaming, if we decide to do it, should be done by an admin via the site (languages are not hardcoded in code).

jesus2099:
I have created https://tickets.metabrainz.org/browse/MBS-13152 for this, then.
Are there any files that should be kept in this PR or do I cancel it?

admin/SetLanguageFrequencies
```diff
@@ -13,11 +13,11 @@ use Try::Tiny;
 
 use constant MAX_TOP_LANGUAGES => 20;
 
-# Always include the six official UN languages and the "Multiple languages"
+# Always include the six official UN languages, the "Multiple languages" and the "No linguistic content" / "No lyrics"
 # pseudo language.
 #
 use constant PRESELECTED_LANGUAGES => qw(
-    ara eng fra spa rus zho mul
```
reosarevok, line 20: This would set no linguistic content into the top "frequently used" list, which I understand from the ticket you don't think is needed?

reosarevok (commented):
The changes to tests seem fine if we change the name from the DB, but I think everything else is unwanted.

admin/SetLanguageFrequencies
jesus2099: Exacly y'oure righjt!

jesus2099: Reverted. :)

jesus2099:
I have removed `*.po` files from PR and reverted `admin/SetLanguageFrequencies`.
Now there are only 3 file changes remaining:

- `po/languages.pot`
- `t/lib/t/MusicBrainz/Server/Controller/WS/js/Edit.pm`
- `t/sql/initial.sql`

Maybe the `.pot` file should not be changed either?

reosarevok:
Yes, the `.pot` file is also for translations and will be changed automatically.

jesus2099:
I see I forgot a `po/languages_notrim.pot`.
I have reverted `po/languages.pot` now there are only 2 remaining changes:

- `t/lib/t/MusicBrainz/Server/Controller/WS/js/Edit.pm`
- `t/sql/initial.sql`

Are they OK?
If so, I have another PR where I must do some squash, so I will have to re-learn how to do that shortly and I will be able to make those 2 commits (more easy to follow-up now) into only 1.

jesus2099:
I have changed 2 commits into 1.

reosarevok (approved):
This is good now I think.

PR #2974, opened 2023-07-03
MBS-13149 / MBS-13153 / MBS-13156 / MBS-13157: More guess case changes

reosarevok:
### Fix MBS-13153 and MBS-13157, implement MBS-13149 and MBS-13156

# Problem
The words "remaster" ("remastered") and "refix" are pretty much always ETI and should be treated as such by guess case. Same for "re-edit", which we already try to cover, but buggily.
While looking into these I also noticed we don't use global regexes enough, causing issues if the same word appears twice (uncommon, but not impossible).

# Solution
I added all of the new words to the list of words which are put into parens if seen at the end of the title, and lowercased if in parens.
I fixed the re-edit check and made it behave differently inside and outside parens.
I made the a cappella and OC ReMix checks global so that it checks more than once per title.

# Testing
Added a basic test for each case.


reosarevok:

reosarevok:

reosarevok:

mwiencek (approved):

PR #2975, opened 2023-07-04
MBS-13092: Support Apple Music song links

reosarevok:
### Implement MBS-13092

# Problem
We were blocking song links for Apple Music, but I can't see any real reasons to. There's two kinds of song links - one that selects the track on an album (kind of like our track links in MB) and one that goes to a /song page for the track itself.

# Solution
This allows entering `/song` links on recordings, and converts `/album/albumid?i=trackid` links to /song ones, since there's no reason not to convert them and allow them, as far as I can tell.
I split the regexes a bit more for clarity.

# Testing
URL tests added.


reosarevok:

yvanzo (approved):
Maybe `/song` URLs didn’t exist when you initially implemented support for Apple Music?

The conversion of album links with queried song is great.

Didn’t test but CI test coverage is enough here.

:ship:

PR #2977, opened 2023-07-04
MBS-13098: Process cover art params for seeding

derat:
Update MusicBrainz::Server::Controller::Release's add_cover_art() method to pass query parameters to the form's process() method when handling GET requests so that the edit note textarea and make-votable checkbox can be seeded via add-cover-art.edit_note and add-cover-art.make_votable, respectively.

This matches what MusicBrainz::Server::Controller's edit_action() method does for other forms that are seedable via query params.

Also update add_cover_art.tt to remove query parameters from the form's "action" URL so that they won't be double submitted in the query string and the request body.

---

This was definitely at the upper end of the time-spent-to-lines-of-code ratio. 😬 I haven't written any Perl in a long time and haven't used Catalyst/HTML::FormHandler/etc. before, so it took me a lot of time to figure out what was different between the add-cover-art form and other forms that support seeding (like adding artists). Hopefully I'm not doing anything wrong.

I didn't see anything in the musicbrainz-docker repo about emulating the CAA, and I couldn't get the instructions in `HACKING-CAA.md` working under mb-docker, so I ended up just hacking `lib/MusicBrainz/Server/Controller/WS/js.pm` and `root/static/scripts/edit/MB/CoverArt.js` locally to skip uploading. This change works for me as expected after doing that.

I also wasn't sure what to do with regard to tests. One option might be to make `t/lib/t/MusicBrainz/Server/Controller/Release/EditCoverArt.pm` pass a query string and check that the resulting page contains the seeded value. Another option might be updating `t/selenium/CAA.json5`. Should I be looking into adding either of those changes to this? I haven't tried running tests yet and I'm not sure how they work with musicbrainz-docker (if at all).

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -361,7 +361,7 @@ sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
         }
     } elsif ($c->form_posted) {
         $c->response->status(500);
-    } elsif (!$c->form_posted && %{ $c->req->query_params }) {
+    } elsif (%{ $c->req->query_params }) {
         # Process query parameters to support seeding fields.
         my $merged = { ( %{$form->fif}, %{$c->req->query_params} ) };
         $form->process( params => $merged );
--- a/lib/MusicBrainz/Server/Controller/Release.pm
+++ b/lib/MusicBrainz/Server/Controller/Release.pm
@@ -361,6 +361,11 @@ sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
         }
     } elsif ($c->form_posted) {
         $c->response->status(500);
+    } elsif (!$c->form_posted && %{ $c->req->query_params }) {
+        # Process query parameters to support seeding fields.
+        my $merged = { ( %{$form->fif}, %{$c->req->query_params} ) };
+        $form->process( params => $merged );
+        $form->clear_errors;
     }
 
     if ($returning_json) {
--- a/root/release/add_cover_art.tt
+++ b/root/release/add_cover_art.tt
@@ -18,7 +18,9 @@
     }).fail(function () { $(".caa-warning").parent().toggle(); });
   </script>
 
-  <form id="add-cover-art" class="cover-art" action="[% c.req.uri %]" method="post" >
+  [%# Remove query params used to seed fields to avoid a "field must contain a single
+      value" error due to submitting duplicate parameters via the body later. %]
+  <form id="add-cover-art" class="cover-art" action="[% c.req.uri.replace('\?.*', '') %]" method="post" >
     [%- USE r = FormRenderer(form) -%]
 
     <div class="add-files row">
--- a/root/release/add_cover_art.tt
+++ b/root/release/add_cover_art.tt
@@ -20,7 +20,7 @@
 
   [%# Remove query params used to seed fields to avoid a "field must contain a single
       value" error due to submitting duplicate parameters via the body later. %]
-  <form id="add-cover-art" class="cover-art" action="[% c.req.uri.replace('\?.*', '') %]" method="post" >
+  <form id="add-cover-art" class="cover-art" action="[% c.req.uri.path %]" method="post">
     [%- USE r = FormRenderer(form) -%]
 
     <div class="add-files row">
--- a/t/selenium/CAA.json5
+++ b/t/selenium/CAA.json5
@@ -3,7 +3,8 @@
   commands: [
     {
       command: 'open',
-      target: '/release/dd245091-b21e-48a3-b59a-f9b8ed8a0469/add-cover-art',
+      // Seed the edit note field via a query parameter.
+      target: '/release/dd245091-b21e-48a3-b59a-f9b8ed8a0469/add-cover-art?add-cover-art.edit_note=seeded+note',
       value: '',
     },
     {
@@ -17,6 +18,11 @@
       command: 'check',
       target: "xpath=((//ul[contains(@class, 'checkboxes')])[1]//input[contains(@type, 'checkbox')][contains(@class, 'type')])[1]",
       value: '',
+    },
+    {
+      command: 'check',
+      target: "id=id-add-cover-art.edit_note",
+      value: 'seeded note',
     },
     {
       command: 'sendKeys',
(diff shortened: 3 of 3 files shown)


reosarevok:
@brainzbot, add to whitelist

reosarevok:
I was thinking that we already had a way to seed because @ROpdebee has a script that allows seeding this page, but that is actually [implemented on top of the page](https://github.com/ROpdebee/mb-userscripts/blob/1831d6ea8cdc70774e12c1fc8ff76f198b0a995d/src/mb_enhanced_cover_art_uploads/docs/seeding.md?plain=1#L11) (that's why it also can seed the image itself).

What does this allow seeding, exactly? We should document it. Only edit_note and make_votable then?

Updating the selenium test to seed something and ensure it works seems like a good idea (our selenium tests are fairly flaky, but they usually eventually pass).

derat:
Thanks for taking a look!

Yes, this just adds support for seeding `edit_note` and `make_votable`. I mentioned that in the commit description; do you mean it should also be documented somewhere under <https://musicbrainz.org/doc>? As far as I'm aware, seeding all forms besides the release editor is currently undocumented. :-(

I'll look into updating the Selenium test. <https://ci.metabrainz.org/job/musicbrainz-server/9309/> looks like it was interpreted as a failure, but I suspect that that was spurious.

reosarevok:
If that's all, then it's already documented as much as we currently do :) The ticket itself though should specify which fields are being made seedable though.

@yvanzo, @mwiencek: we should figure out where and how to document what fields can be seeded in forms and how maybe? :)

derat:
I've updated the CAA Selenium test to seed the edit note. I wasn't able to run Selenium tests locally (not sure if they don't get along with musicbrainz-docker or if there's some other issue; `MB_SERVER_ROOT` seems to be set incorrectly when I run outside of Docker), so I'm just hoping for the best with the automated build.

t/selenium/CAA.json5
```diff
@@ -152,6 +158,17 @@
         release: 'https://musicbrainz.org/release/dd245091-b21e-48a3-b59a-f9b8ed8a0469',
       },
     },
+    // Verify that the seeded edit note was used.
```
derat, line 161: The `/edit/<ID>/data` endpoint doesn't seem to include the original edit note, so I think I need to load the page in order to confirm that the seeded value was used. Let me know if I'm missing something, though.

derat:
It took a few tries, but I think the Selenium test change that I made is passing now (the failure in <https://ci.metabrainz.org/job/musicbrainz-server/9315/console> seems like unrelated flakiness).

Regarding documentation for non-release seeding, I'm happy to write a wiki page if it'd help. Should I file a ticket to track this or just create a page under my account and then ping you when it's ready for review?

derat:
Any other comments on this?

reosarevok:
Summer is a slow time, so let's wait for others to see this :)

reosarevok:
Hi! We had a small talk about it with @yvanzo and we thought we could start a page like https://wiki.musicbrainz.org/Development/Seeding which would (eventually) detail what can be seeded to each page in MB :) If you are willing to take a look at it and start it, that'd be sweet, we can help fill it too. Relationship editor seeding docs should then move from /Development/Release_Editor_Seeding to /Development/Seeding/Release_Editor and also be linked from this new page, I think (I can do the moving once we get started on the page).

derat:
Thanks, sounds good! I've created https://wiki.musicbrainz.org/Development/Seeding and will add what I know about each seedable form.

reosarevok:
I moved the release editor page to a subpage and amended the wording / link, as suggested :)

derat:
Thanks! I've added some detailed info about the "Add Artist" form to https://wiki.musicbrainz.org/Development/Seeding. If the approach that I took seems reasonable to you, I'll do the same thing for the other forms.

I'm not sure what would be best for fields that take database IDs. I could link to my generated file (<https://codeberg.org/derat/yambs/src/branch/main/seed/enums.go>) or the godoc version (e.g. <https://pkg.go.dev/codeberg.org/derat/yambs/seed@main#LinkType>), but I'm not sure if there's a more canonical source anywhere (besides the database itself).

derat:
Hi, just checking in again. Any comments, either on this change or on <https://wiki.musicbrainz.org/Development/Seeding>?

yvanzo:
I moved your added section to a subpage under <https://wiki.musicbrainz.org/Development/Seeding/Artist_Editor>.

derat:
Thanks, @yvanzo! I fixed what seemed like a minor typo about POST-ing to the artist form -- as far I'm aware, it's seeded via GETs, and POSTs  instead create/edit artists.

root/release/add_cover_art.tt
```diff
@@ -18,7 +18,9 @@
     }).fail(function () { $(".caa-warning").parent().toggle(); });
   </script>
 
-  <form id="add-cover-art" class="cover-art" action="[% c.req.uri %]" method="post" >
+  [%# Remove query params used to seed fields to avoid a "field must contain a single
+      value" error due to submitting duplicate parameters via the body later. %]
+  <form id="add-cover-art" class="cover-art" action="[% c.req.uri.replace('\?.*', '') %]" method="post" >
```
mwiencek, line 23: I'm guessing we can just use `c.req.uri.path` here to omit the query parameters, but I haven't tested it.
```suggestion
  <form id="add-cover-art" class="cover-art" action="[% c.req.uri.path %]" method="post">
```

lib/MusicBrainz/Server/Controller/Release.pm
```diff
@@ -361,6 +361,11 @@ sub add_cover_art : Chained('load') PathPart('add-cover-art') Edit {
         }
     } elsif ($c->form_posted) {
         $c->response->status(500);
+    } elsif (!$c->form_posted && %{ $c->req->query_params }) {
```
mwiencek, line 364: nitpick, but based on the `elsif ($c->form_posted)` above think `!$c->form_posted` is guaranteed?
```suggestion
    } elsif (%{ $c->req->query_params }) {
```

mwiencek (approved):
lgtm overall, thanks for the docs and for updating the selenium tests!

lib/MusicBrainz/Server/Controller/Release.pm
derat: Ah, sorry for missing that. Yep, that makes sense.

root/release/add_cover_art.tt
derat: Thanks, I was hoping that something like that was available. I can't test this right now either, but I think it'll be exercised by the Selenium test.

derat:
Thanks for approving this! I've applied your suggestions, and the tests look like they passed. Does anything else need to be done or can this be merged now?

yvanzo (approved):
Thank you, I double-checked the code too including the committed suggestions but could not test it locally.

PR #2978, opened 2023-07-06
Beta: Fix log flood regression since #2902

yvanzo:
# Problem

Since the commit e38d8e95e6ab06888c956df668a4cb5617a3b4ea (pull request #2902) is in prod (`v-2023-06-07`), search queries are triggering two warnings for each result in logs, that is up to 200 log lines for each query, resulting in needless GBs of logs.

# Solution

This patch simply checks if defined before enforcing integer conversion, that is the same approach for `position` and `release_id` as for `format_id` and `track_count`, thus stopping the log flood.

However by doing so, it partly reverts the commit e38d8e95e6ab06888c956df668a4cb5617a3b4ea as it unsyncs the Flow type `MediumT` for its `position` and `release_id` properties. It is unclear whether these should be nullable or if a new type should be used instead since these are null only for search results. Beyond Flow types, the property `mediums_loaded` in Perl is misleading too. A larger refactoring might be needed to sanitize the code.

# Testing

Manually tested by:
* visiting http://localhost:5000/taglookup?artist=Tracy%20Chapman&release=Tracy%20Chapman&track=&tracknum=&duration=&filename=&tport=8000 and checking server logs,
* trying to reproduce MBS-12976

Note: We don’t have CI tests for these warnings in logs.

# Draft progress

* [x] Further check code to determine if undefined `position` is legitimate.
* [x] Further check code to determine if undefined `release_id` is legitimate.
* [x] Further trace code to find out potential implications of this patch.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Medium.pm
+++ b/lib/MusicBrainz/Server/Entity/Medium.pm
@@ -172,14 +172,16 @@ around TO_JSON => sub {
 
     my $track_count = $self->track_count;
     my $format_id = $self->format_id;
+    my $position = $self->position;
+    my $release_id = $self->release_id;
 
     my $data = {
         %{ $self->$orig },
         cdtocs      => [map { $_->cdtoc->toc } $self->all_cdtocs],
         format      => $self->format ? $self->format->TO_JSON : undef,
         format_id   => defined $format_id ? (0 + $format_id) : undef,
-        position    => (0 + $self->position),
-        release_id  => (0 + $self->release_id),
+        position    => defined $position ? (0 + $position) : undef,
+        release_id  => defined $release_id ? (0 + $release_id) : undef,
         track_count => defined $track_count ? (0 + $track_count) : undef,
     };
 


reosarevok:
At the very least then `MediumT` should be changed to allow these to be null :) The flow change shows a lot of places where we expect `position` to always exist, and a couple places for `release_id`.

The tag lookup results' "mediums" are just this:
```
  'mediums' => [
                 bless( {
                          'durations_loaded' => 0,
                          'has_loaded_tracks' => 0,
                          'format' => bless( {
                                               'name' => '12" Vinyl'
                                             }, 'MusicBrainz::Server::Entity::MediumFormat' ),
                          'track_count' => 11
                        }, 'MusicBrainz::Server::Entity::Medium' )
               ]
```

The TagLookup display code doesn't seem to use mediums at all, so we might even consider dropping the mediums before converting the releases to JSON. Alternatively, I think we might want to try and load their position and release ID.

yvanzo:
No ticket needed as there is actually no change to the internal data, these properties are undefined only for search results, it is just typing soup.

@reosarevok: Yes, more refactoring is needed, but this is beyond this pull request which is just about to stop flooding the logs, which is also why I targeted `beta` specifically.

@mwiencek: This is partly reverting your patch, unsyncing `position` and `release_id` in `MediumT` Flow.js type. It keeps the integer conversion even though I didn’t check it is actually needed.

reosarevok:
The integer conversion should be kept for sure, that is a common cause of warnings otherwise :) 

Can you maybe leave a comment on the flow type then saying that this is incorrect for TagLookup? I'm not too happy with knowingly having incorrect flow typings but we can fix that later as long as we have something reminding us it needs to be fixed later :)

reosarevok:
I added MBS-13161 for a proper look at this after we patch the flood - @mwiencek, should we just run `load_related_info` when doing an external search for releases instead, or does that sound too demanding? It would at least mean we have more or less the expected info everywhere.

yvanzo:
> should we just run `load_related_info` when doing an external search for releases instead, or does that sound too demanding?

Loading useless data just to satisfy a type misfit doesn’t seem a good refactoring.

reosarevok (approved):
This is probably fine to merge for now to avoid the warning flood, assuming @mwiencek cannot think of a reason it'd cause problems elsewhere, but we should try to figure out what to do with the typing and the expectations for the release after this :)

yvanzo:
I specifically checked that it has no actual implications beyond the typing soup. I’m merging it now so it can make its way to beta. The review request was more on the refactoring part which is now tracked with the ticket you added.

PR #2979, opened 2023-07-07
MBS-13160: Rename "legal name" on artist page to "performance name of"

reosarevok:
### Implement MBS-13160

# Problem
We renamed the relationship text for "performs as" with STYLE-2328 but the display on the artist overview page was still hardcoded to "Legal name".

# Solution
This renames the hardcoded text to "Performance name of" (as in the relationship) in the case where we follow this relationship (which no longer mandates it must be a legal name). "Legal name" is kept for actual legal name alias display.

I wasn't sure what'd be a good name for the no-longer-`legalName`, ended up going with `baseName` since it's the base of the grouping of names for the artist.

# Testing
Manually, on the artist from the ticket.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -263,26 +263,26 @@ sub show : PathPart('') Chained('load')
 
     my $coll = $c->get_collator();
     my @identities;
-    my $legal_name_artist_aliases;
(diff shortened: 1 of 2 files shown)


mwiencek (approved):
> I wasn't sure what'd be a good name for the no-longer-legalName, ended up going with baseName since it's the base of the grouping of names for the artist.

How about `coreName`? (jk)

PR #2980, opened 2023-07-10
MBS-13168: Add autoselect for BookBrainz series

reosarevok:
### Implement MBS-13168

# Description
We now have a BookBrainz relationship for series too, so we should handle it like the others.

# Testing
Added a basic test - we don't do much validation with these so it's very basic.


yvanzo (approved):
LGTMBDNT

PR #2981, opened 2023-07-10
MBS-13162: Handle Threads links

reosarevok:
### Implement MBS-13162

# Description
This adds `threads.net` (Meta's new Twitter-like) links to the sidebar, and cleans them to use `https://www.` which seems to be the default. It also selects (mandatory) social media for any entity types that support it (currently artist, event, label, place and series), while selecting stream for free for recording and blocking other free streaming entities.
Social media links must be user pages, while streaming links must be threads ("tweets").

This matches what we do with Twitter, which this is mostly a straight clone of.

# Testing
Added the usual basic URL tests for both kinds of links.


yvanzo (commented):
Tested with artist for whom this patch indeed selects - as mentioned in the commit message - but also restricts to social network - which is fine but not mentioned.

About recordings, Threads supports videos up to 5 minutes (for example <https://www.threads.net/t/CucwR6erBPo/>) and it doesn’t seem that any other relationship type than `streamingfree` would make sense.

reosarevok:
I understand the suggestion is we should support posts, which I guess seems sensible enough and closer to what we do with Twitter anyway. Implemented that, see if this seems better now.

yvanzo:

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4690,6 +4690,34 @@ const CLEANUPS: CleanupEntries = {
...
+    match: [new RegExp('^(https?://)?([^/]+\\.)?threads\\.net/', 'i')],
+    restrict: [{...LINK_TYPES.streamingfree, ...LINK_TYPES.socialnetwork}],
+    clean: function (url) {
+      return url.replace(
+        /^(?:https?:\/\/)?(?:www\.)?threads\.net(?:\/#!)?\//,
+        'https://www.threads.net/',
+      );
+    },
+    validate: function (url, id) {
+      const isAProfile = /^https:\/\/www\.threads\.net\/(@[^/]+)$/.test(url);
```
yvanzo, line 4703: Capturing group is not needed here:
```suggestion
      const isAProfile = /^https:\/\/www\.threads\.net\/@[^/]+$/.test(url);
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4690,6 +4690,34 @@ const CLEANUPS: CleanupEntries = {
...
+      const isAThread = /^https:\/\/www\.threads\.net\/t\/[\w\d]+?$/.test(url);
```
yvanzo, line 4704: 1. The quantifier `?` is not wanted here as the thread identifier is mandatory.
2. The expression would not work with thread identifier containing a hyphen `-`.
   Maybe just use the same expression as above?
```suggestion
      const isAThread = /^https:\/\/www\.threads\.net\/t\/[^/]+$/.test(url);
```

yvanzo (changes requested):
The newly added validation makes sense.
Commented on two implementation details.
LGTMBDNT otherwise.

yvanzo (approved):

PR #2982, opened 2023-07-13
Upgrade Flow to 0.214.0

mwiencek:
Upgrades Flow to the latest version, so that we can make use of new features being added.

I changed the Babel parser to hermes, which is the parser recommended by the Flow team, so that the types guards syntax (which I've used here) can be parsed.

Similarly, for ESLint I've changed the parser to hermes-eslint (also recommended by the Flow team), so that ESLint can parse the new syntax too. However, there are some issues with it:
 * eslint-plugin-flowtype isn't supported, so I had to remove it.
 * Running eslint gives me an error in places-map.js which I'm not sure how to debug.

Perhaps we should check if the latest Babel ESLint parser supports type guards yet and stick with that.


reosarevok:

yvanzo:
> Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@babel/plugin-syntax-hermes-parser' imported from /home/musicbrainz/musicbrainz-server/babel-virtual-resolve-base.js

reosarevok:
I got the same error, I think the problem is that it should be `'babel/plugin-syntax-hermes-parser'` (no @ on this package) but that caused a different error for me and I ended having to add it to the packages later like we do with the `istanbul` parser. Not sure if that's even correct though.

reosarevok:
Seems this is still getting the `places-map.js` error :(

mwiencek:
Reported it upstream at https://github.com/facebook/hermes/issues/1088

reosarevok:
I see that hermes is supposed to be fixed in a week or so so I guess we should wait with this until then? :)

mwiencek:
Yeah, makes sense to wait. I did apply the patch from the linked issue to here (via `patch-package`) just to see if the tests pass otherwise.

reosarevok (commented):
Old comments from when I first reviewed this that I never submitted, apparently. Please ignore if no longer relevant.

yvanzo (approved):
LGTMBDNT :passenger_ship:

PR #2983, opened 2023-07-13
MBS-13167: Show last granted token date for auth applications

reosarevok:
### Implement MBS-13167

# Problem
It seems useful to know when was the last time a specific application / scope had a token granted; that is likely to help users revoke unneeded ones (I for example currently seem to have 7 ListenBrainz tokens, some of which were granted in 2015).

# Solution
This adds a column in the authorized applications table for the `granted` token column, which displays the last time a token was granted for this specific application + scope combo. I stopped returning the `refresh_token` here since it seems we do not use it at all.

# Testing
Manually.


yvanzo (approved):
LGTMBDNT :ship:

PR #2984, opened 2023-07-13
MBS-13169: Use SVG over PNG for new icons

reosarevok:
### Implement MBS-13169

# Problem
It seems like there's no real benefit to using the PNG versions of the new icons over just using SVG directly, and we've had complaints that they are blurry on high DPI screens.

# Solution
This drops the PNGs and uses SVG directly (set to 16px width for the `background-image` ones - others were already set at 16px height since they're not quite square).

# Testing
Manually, by checking edit pages and the relationship editor.


yvanzo (approved):
> there's no real benefit to using the PNG

PNG worked fine and was supported by a wider range of browsers and devices. Nowadays SVG is supported everywhere too; see <https://caniuse.com/?search=svg>. The SVG files provided seem to be simple enough to not increase too much the rendering load.

> on high DPI screens

The benefit of SVG is not only on high DPI screens, its scalability makes it look nicer when resized on any screen, that’s a step forward for mobile support too.

> Testing
> Manually

Tested the same way with recent FF/Linux. We already use SVG for entity type icons. I don’t expect any specific issue with these new SVG icons, but beta will help with finding any and it should be minor.

Just requesting @mwiencek’s input on browsers coverage.

mwiencek (approved):
We should be mindful of keeping the svgs optimized, but based on the caniuse link from @yvanzo I don't see any issue with using them for simple icons across the board.

PR #2985, opened 2023-07-14
MBS-13171: Add support for Boomkat URLs

atj:
This commit adds support for cleaning, validating, autoselecting and normalising Boomkat URLs added to artists, labels and releases.


root/static/scripts/common/constants.js
```diff
@@ -80,6 +80,7 @@ export const FAVICON_CLASSES = {
   'bigcartel.com': 'bigcartel',
   'bookbrainz.org': 'bookbrainz',
   'books.apple.com': 'applebooks',
+  'boomkat': 'boomkat',
```
reosarevok, line 83:
```suggestion
  'boomkat.com': 'boomkat',
```
Although this whole system is a mess, but still better to follow the general style :)

root/static/styles/favicons.less
```diff
@@ -29,6 +29,7 @@
 .favicon("beatport", 32);
 .favicon("bigcartel", 32);
 .favicon("bookbrainz", 32);
+.favicon("boomkat", 16);
```
reosarevok, line 32: 16 is the default so it does not need specifying (is there a reason you went for 16 though? we usually do 32 but probably fine either way)

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1337,6 +1337,80 @@ const CLEANUPS: CleanupEntries = {
...
+    restrict: [
+      {
+        artist: [
+          LINK_TYPES.downloadpurchase.artist,
+          LINK_TYPES.mailorder.artist,
+        ],
+        label: [
+          LINK_TYPES.downloadpurchase.label,
+          LINK_TYPES.mailorder.label,
+        ]
```
reosarevok, line 1351: This is missing a dangling comma and triggering a linting error :)

reosarevok (changes requested):

root/static/scripts/edit/URLCleanup.js
yvanzo: Same but actionable suggestion:
```suggestion
        ],
```

root/static/styles/favicons.less
atj: Strangely they only offer a 16px favicon and I couldn't find any other copies of the same icon linked anywhere in the page source:

https://boomkat.com/assets/favicon-eec454f2397be06df3606b27e5be871d3847fffbfdcc012af30e26439aec6ed4.ico

I've removed the default parameter value.

root/static/scripts/common/constants.js
atj: Sorry, missed that for some reason! Fixed...

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1337,6 +1337,80 @@ const CLEANUPS: CleanupEntries = {
...
+    select: function (url, sourceType) {
+      const m = /^https:\/\/boomkat\.com\/(artists|labels|products)\//.exec(url);
+      if (m) {
+        const prefix = m[1];
+        switch (prefix) {
+          case 'artists':
+            if (sourceType === 'artist') {
+              return [
+                LINK_TYPES.downloadpurchase.artist,
+                LINK_TYPES.mailorder.artist,
```
reosarevok, line 1369: Are things always available for download *and* mailorder? (this kind of thing is why we usually end up adding separate relationships for different stores, since in most cases you cannot know that an artist *does* have both purchase and mail order without checking further and even if they do now that might easily change... but the situation is always messy)

atj: Yes, this is a fair point. It depends whether the artist / label has digital or physical releases available to purchase for mail order. I've seen cases where only a physical release is available but a digital version is available on other sites.

I guess I should just remove the select function entirely?

yvanzo: Yes in this case removing the `select` function will do.

atj: OK, I removed the `select` function and updated the tests.

yvanzo (approved):
Tested locally, even though the relationship type isn’t auto-selected, the selector shows the two appropriate options only.

@reosarevok: Once the changes work for you feel free to merge it.

PR #2986, opened 2023-07-18
MBS-13173: Also allow daily.bandcamp.com interviews

reosarevok:
### Fix MBS-13173

# Problem
Bandcamp publishes interviews with artists seemingly quite often, but we were automatically selecting "bandcamp" for those.

# Solution
This moves `daily.bandcamp` URLs to a separate handling section, so we can deal with them more properly and select interview when appropriate.

While doing this I noticed that newer album reviews are no longer listed under the date as we expected, but under an `album-of-the-day` path. As such, I made both interview and review support these sorts of paths.

Finally, since we are no longer doing any RG-level selecting on the standard bandcamp handling, non-daily bandcamp URLs need to be blocked separately on RGs. AFAICT no non-daily bandcamp URL would be acceptable for RG (and the previous setup blocked them all) so this is a blanket ban.

# Testing
Added a test for the interviews, another for the new kind of review URLs. The RG blocking is tested by the `davidmandelberg.bandcamp.com` RG test (which broke until I reimplemented the blocking explicitly).

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -965,21 +965,16 @@ const CLEANUPS: CleanupEntries = {
   },
   'bandcamp': {
     match: [new RegExp(
-      '^(https?://)?([^/]+\\.)?bandcamp\\.com(?!/campaign/)',
+      '^(https?://)?(((?!daily)[^/])+\\.)?bandcamp\\.com(?!/campaign/)',
       'i',
     )],
     restrict: [{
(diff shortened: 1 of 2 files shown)


derat:
Thanks for fixing this so quickly! I'm envious of all those negative lookaheads; too much Go lately. :-)

yvanzo (approved):
Handling Bandcamp Daily separately is a good move.
Didn’t test manually but the CI test coverage is enough here.

:ship:

PR #2988, opened 2023-07-21
MBS-13183: Add CORS support to Wikipedia extracts

yvanzo:
# Problem

Allow ListenBrainz to easily use Wikipedia extracts retrieved by MusicBrainz server.

# Solution MBS-13183

Supporting CORS is one way out, it doesn’t require tinkering with Redis cache.

# Checklist for author

* [x] Test manually using cURL
* [x] Add CI test

# Action

1. Ping LB team when available


mwiencek (approved):
LGTM.  In the future it would be nice to reuse `Controller::Role::CORS` somehow (perhaps turning it into a parameterized role that specifies the method name to wrap).

PR #2989, opened 2023-07-21
MBS-13177: Show label events on label rels page

reosarevok:
### Fix MBS-13177

# Problem
Label-Event relationships are not shown anywhere on the label page. 

# Solution
Label-Event rels were already being loaded (unpaged), but not being shown since `Relationships.js` did not use them. This was not a problem until recently because we didn't have any. Since the only one we now have is `label presents event`, which seems more akin to recording/release rels with cardinality 1:0 than the other rels we show outside the rels table, I'm moving this to load them as a paged subset and showing them in `RelationshipsTable.js` instead. 

I considered a separate events page but events are nowhere near as relevant to most labels as to artists, so I think the generic rels table is more than good enough. 

# Testing
Manually, with the label provided in the ticket (`/label/e8fb3a04-7d34-42fe-a155-bd8c51c96db8/relationships`) and `pink` data.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Label.pm
+++ b/lib/MusicBrainz/Server/Controller/Label.pm
@@ -12,10 +12,10 @@ with 'MusicBrainz::Server::Controller::Role::Load' => {
         default => ['url'],
         subset => {
             show => ['artist', 'label', 'url'],
-            relationships => [qw( area artist event instrument label place series url )],
+            relationships => [qw( area artist instrument label place series url )],
         },
         paged_subset => {
-            relationships => [qw( recording release release_group work )],
+            relationships => [qw( event recording release release_group work )],
         },
     },
 };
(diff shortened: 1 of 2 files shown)


yvanzo (changes requested):
For the record, this change follows the creation of the first event-label relationship type “presented by” last week for STYLE-443. Thus it’s not part of sample data dump yet.

Tested locally by creating a random event-label relationship type, linking an event to a label and visiting their Relationships tabs.

Your decision to list these relationships as paginated in label page and not paginated for event page makes perfect sense but isn’t really explained. The rationale should be fully provided in pull request’s description and commit message. While you are at it, please fix commit message and pull request’s description to mention “event-label” rather than label-event relationship type.

It looks good to me otherwise.

yvanzo (approved):
Thanks! Someone in the future will thank you again. :ship:

PR #2991, opened 2023-08-02
MBS-13201: Add ORCID to other databases whitelist and handle links

reosarevok:
### Implement MBS-13201

# Description
ORCID identifiers are given to researchers - this includes music researchers (like musicologists and some academically trained musicians) as well as musicians that also happen to be scientists for their day jobs. 

As far as I can tell, only researchers (so, artists) get ORCIDs, so every other entity gets an ENTITY error.

The regex for the valid profile URLs is based on https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier

For the sidebar, as per https://info.orcid.org/brand-guidelines/ they seem to prefer the logo + the ID itself, which we could do, but I don't think the ORCID logo is recognizable outside a pure academic context, so I chose to do logo + ORCID for now and hope they don't mind.

# Testing
Manually plus added a test for the cleanup and another to make sure search links are rejected.


yvanzo (approved):
The code looks good but only then I looked into this database and its addition seems to be useless; See my comment to the ticket.

reosarevok:
Guess it doesn't make sense to keep this open if there's no path for it to be merged anyway.

PR #2993, opened 2023-08-02
MBS-13202: Add LibraryThing to other databases whitelist and handle links

reosarevok:
### Implement MBS-13202

# Description
LibraryThing is a site for (mostly) books, including audiobooks, although they also have music to some degree.

As far as I can tell, they have pages for `author` (artist), `nseries` (series), and `work`, which can be a book (MB work) but also a music album (MB RG), so I allowed `work` on both levels.

# Testing


reosarevok:
I have only seen book series so far but I wouldn't be surprised if they have album/RG ones too, yes.

reosarevok:
Documented under https://wiki.musicbrainz.org/Other_Databases_Relationship_Type/Whitelist#.28Audio.29books

PR #2994, opened 2023-08-03
MBS-13204: Add Goodreads to other databases whitelist and handle links

reosarevok:
### Implement MBS-13204

# Description
Goodreads is a site for books, including audiobooks.

As far as I can tell, they have pages for `author` (artist), `book` (edition, so release) and `series` (series). There is a `work` page (https://www.goodreads.com/work/editions/194190) but as far as I can tell that's only used to list all editions and has barely any info of its own, so I'm not adding support for it now. It would be linked to works if we later decide to use it.

The default link for `series` does not include `/show`, while it does for `author` and `book`. Since `series` links *do* work with `/show` I used it consistently for all cleaned up links.

# Testing


yvanzo (approved):
LGTMBDNT :ferry: 

Same as for other pull requests about documentation.

reosarevok:
Documented under https://wiki.musicbrainz.org/Other_Databases_Relationship_Type/Whitelist#.28Audio.29books

PR #2996, opened 2023-08-04
MBS-13210 / MBS-13211 / MBS-13212 / MBS-13213 / MBS-10977: Handle art gallery links

reosarevok:
### Implement MBS-13210 / MBS-13211 / MBS-13212 / MBS-13213 / MBS-10977

# Description
We now have an "art gallery" relationship for artists, so this handles all the art gallery sites that our users had requested. Implementation is very similar, see each commit for exact details.

For the two most used ones (pixiv and DeviantArt, both with many hundred uses) I also added sidebar display. For the rest I did not yet, we can add it later if they become more common.

# Testing
Added tests for all of the different sites

# Actions
1. Migrate existing links (MBBE-81) and I guess clean them up too - this does not need to happen before we merge this though. If we migrated them by hand the cleanup would happen automatically for many cases, but IIRC that's not the case with bot editing.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -905,6 +908,47 @@ const CLEANUPS: CleanupEntries = {
...
+          };
+        }
+        if (hardcodedPaths.includes(userName)) {
+          return {
+            result: false,
+            target: ERROR_TARGETS.URL,
+          };
+        }
+        return {result: true};
+      }
```
yvanzo, line 948: `hardcodedPaths` are used in a single place and at last, so they can just be defined there.

```suggestion
      const m = /^https:\/\/www\.artstation\.com\/([^\/]+)$/.exec(url);
      if (m) {
        const userName = m[1];
        if (userName === 'search') {
          return {
            error: noLinkToSearchMsg(),
            result: false,
            target: ERROR_TARGETS.URL,
          };
        }
        const hardcodedPaths = [
          'artwork',
          'blogs',
          'challenges',
          'hire',
          'jobs',
          'learning',
          'marketplace',
          'prints',
          'schools',
          'studios',
        ];
        if (hardcodedPaths.includes(userName)) {
          return {
            result: false,
            target: ERROR_TARGETS.URL,
          };
        }
        return {result: true};
      }
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1316,6 +1360,44 @@ const CLEANUPS: CleanupEntries = {
...
+            result: false,
+            target: ERROR_TARGETS.URL,
+          };
+        }
+        if (hardcodedPaths.includes(userName)) {
+          return {
+            result: false,
+            target: ERROR_TARGETS.URL,
+          };
+        }
```
yvanzo, line 1395: Same as above

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1990,6 +2072,45 @@ const CLEANUPS: CleanupEntries = {
...
+        return {result: true};
+      }
```
yvanzo, line 2110: Same as above

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2268,6 +2389,44 @@ const CLEANUPS: CleanupEntries = {
...
```
yvanzo, line 2426: Same as above

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1990,6 +2072,45 @@ const CLEANUPS: CleanupEntries = {
       return {result: false, target: ERROR_TARGETS.URL};
     },
   },
+  'deviantart': {
+    match: [new RegExp('^(https?://)?(www\\.)?deviantart\\.com', 'i')],
+    restrict: [LINK_TYPES.artgallery],
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?deviantart\.com\/([^\/]+).*$/, 'https://www.deviantart.com/$1');
```
yvanzo, line 2079: It should also drop query and fragment:

```suggestion
      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?deviantart\.com\/([^\/?#]+).*$/, 'https://www.deviantart.com/$1');
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -1990,6 +2072,45 @@ const CLEANUPS: CleanupEntries = {
...
+        'about',
+        'core-membership',
+        'forum',
+        'groups',
+        'join',
+        'shop',
+        'team',
+        'users',
+      ];
+      const m = /^https:\/\/www\.deviantart\.com\/([^\/]+)$/.exec(url);
```
yvanzo, line 2093: To match the suppresion of query and fragment as per previous comment:
```suggestion
      const m = /^https:\/\/www\.deviantart\.com\/([^\/?#]+)$/.exec(url);
```

yvanzo (changes requested):
Looks good overall but a few pretty minor points.

yvanzo (approved):
Thanks! LGTMBDNT :ship:

PR #2998, opened 2023-08-07
MBS-13215 - Update the Twitter/X logo

Aerozol:
Replaced old twitter icon with their current ‘X’ icon.


reosarevok:
But it's sooooo ugly :(

reosarevok (approved):
I guess we should bow to the whims of the nutjob anyway.

Aerozol:
Thanks! They made their ugly-looking bed, now they can lie in it 😑

PR #3000, opened 2023-08-07
MBS-13216: Add VNDB to other databases whitelist and handle links

reosarevok:
### Implement MBS-13216

# Description
VNDB is a site for visual novels, which apparently often include voice actors that also often sing - in any case, there's 50 of these already in the DB and several users asking for it to be otherdbs rather than "fanpage" as it seems has been used now.

As far as I can tell, they have pages for "staff" (`/s`, artist), "character" (`/c`, artist), "producer" (`/p`, label) and "visual novel" (`/v`, which would probably be works, but we don't actually use works for games right now and this is not a soundtrack page, so I'm blocking for now).

# Testing


yvanzo (approved):
LGTMBDNT :ship:

Same as for other pull requests: the documentation needs to be updated too.

reosarevok:
Documented under https://wiki.musicbrainz.org/Other_Databases_Relationship_Type/Whitelist#Video_game_music

PR #3001, opened 2023-08-07
MBS-13217: Fix displaying cover art edits

yvanzo:
# Problem MBS-13217

Two glitches were affecting the display of cover art edits.

1. When a release was merged, the release field was wrongly displayed as not existing anymore.
2. When a release was merged or deleted, the cover art field was wrongly claimed as not displayable (but for "Add cover art" edits).

# Solution

This patch loads two different releases instead, one for release link and the other for cover art image, and sets release MBID/GID if merged.

# Work in progress

* [ ] Fix `ReorderCoverArt` too
* [ ] Test manually displaying
  * [ ] “Add cover art” edit with merged release.
  * [ ] “Add cover art” edit with deleted release.
  * [ ] “Edit cover art” edit with merged release.
  * [ ] “Edit cover art” edit with deleted release.
  * [x] “Remove cover art” edit with merged release: [edit #97086435](http://localhost:5000/edit/97086435).
  * [ ] “Remove cover art” edit with deleted release.
  * [ ] “Reorder cover art” edit with merged release.
  * [ ] “Reorder cover art” edit with deleted release.

# Notes

Adding CI tests is a separate ticket: MBS-11490.


lib/MusicBrainz/Server/Edit/Release/AddCoverArt.pm
```diff
@@ -105,11 +105,19 @@ sub build_display_data {
...
     my $loaded_release = $loaded->{Release}{ $self->data->{entity}{id} };
-    my $release = $loaded_release ||
+
+    # The release used to link (if it still exists) from the release field.
+    my $field_release = $loaded_release ||
         Release->new(
-            id => $self->data->{entity}{id},
             name => $self->data->{entity}{name},
+            # Allow linking to release redirecting to still existing release.
+            $self->c->model('Release')->get_by_gid($self->data->{entity}{mbid}) ? (
```
mwiencek, line 114: I remembered that the `foreign_keys` method also supports loading entities by MBID, IIRC. So I believe we can just replace the reference to `$self->data->{entity}{id}` in `foreign_keys`, and also in the `$loaded_release` assignment above, to point to the `mbid` property instead (unless I am missing something).

mwiencek (commented):
Does this still need further testing? The code looks good to me, though I believe it can be simplified by using the MBID in `foreign_keys`, if I am not mistaken.

PR #3002, opened 2023-08-08
 MBS-13218: Add Jaxsta to other databases whitelist and handle links

reosarevok:
### Implement MBS-13218

# Description
Jaxsta is a database for "official music credits".

As far as I can tell, they have pages for `profile` (artist and label), `recording`, `release` and `work` (which map to ours). Recording, release (for singles) and work pages can contain (licensed) lyrics but that's not their primary function, so using `otherdatabases` by default still seems preferable; for works I also allow manually selecting lyrics in addition, as we do for Mainly Norfolk. For releases that is restricted to singles and for recordings we don't have an equivalent lyrics option anyway.

# Testing


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2910,7 +2910,42 @@ const CLEANUPS: CleanupEntries = {
...
+          case 'recording':
+            if (sourceType === 'recording') {
+              return LINK_TYPES.otherdatabases.recording;
+            }
+            break;
+          case 'release':
+            if (sourceType === 'release') {
+              return LINK_TYPES.otherdatabases.release;
+            }
+            break;
```
yvanzo, line 2939: This is not needed (but for `work`) because there is only one possible relationship type (through `restrict`) for each of these entity types.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -2850,6 +2850,42 @@ limited_link_type_combinations: [
...
+            expected_clean_url: 'https://jaxsta.com/recording/5b01140a-a82f-5f5c-aa81-93d61de00f94',
+       only_valid_entity_types: ['recording'],
+  },
+  {
+                     input_url: 'https://jaxsta.com/release/66512305-5fb7-5d43-a194-d5ed56c16cab/7c5528fe-96f7-5c0c-9af2-9b756b909436/variants',
+             input_entity_type: 'release',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://jaxsta.com/release/66512305-5fb7-5d43-a194-d5ed56c16cab/7c5528fe-96f7-5c0c-9af2-9b756b909436',
+       only_valid_entity_types: ['release'],
+  },
```
yvanzo, line 2881: This is linking to a “release variant” which matches our release.
Their “release” actually matches our release group.
This should be explained in ticket/pull request description/commit message.

Also it seems that it would be helpful to allow linking `https://jaxsta.com/release/66512305-5fb7-5d43-a194-d5ed56c16cab` for example, maybe formatted as `https://jaxsta.com/release/66512305-5fb7-5d43-a194-d5ed56c16cab/group/variants`, to a release group in case the current releases in the release group don’t match any of the current variants in Jaxsta.

yvanzo (changes requested):
It looks good overall but made two comments.

Also, same as for other pull requests: the documentation has to be updated too.

root/static/scripts/tests/Control/URLCleanup.js
reosarevok: Hmm. I didn't notice there was a way to link to a specific RG-level page, because just doing https://jaxsta.com/release/66512305-5fb7-5d43-a194-d5ed56c16cab automatically loads a variant (hence adding it as release only). But we also don't want to clean up the variant, so I guess we should only convert to https://jaxsta.com/release/66512305-5fb7-5d43-a194-d5ed56c16cab/group/variants if it's given without a second UUID, right? (ideally we'd just clean up differently depending on whether the user is at a RG page or a release one, but I guess we don't have any way to do that...)

root/static/scripts/edit/URLCleanup.js
reosarevok: Hmm, ok. If that's the case we should also drop the extra stuff for `mainlynorfolk` then, that's what I was following :)

yvanzo: Yes.

root/static/scripts/tests/Control/URLCleanup.js
yvanzo: There is no way to do it from their website, I edited the URL by hand. I thought it would be reasonable to replace the second UUID with `group` and add the subpath `/variants` but I now realize that the function `cleanup` only depends on the URL and not on the entity type being edited. So let’s give up this idea for now but still further explain the release mapping in commit message.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2972,6 +2972,50 @@ const CLEANUPS: CleanupEntries = {
...
+              target: ERROR_TARGETS.ENTITY,
+            };
+          case LINK_TYPES.otherdatabases.recording:
+            return {
+              result: type === 'recording',
+              target: ERROR_TARGETS.ENTITY,
+            };
+          case LINK_TYPES.otherdatabases.release:
+            return {
+              result: type === 'release',
```
yvanzo, line 3022: The second UUID should also be present.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -2972,6 +2972,68 @@ const CLEANUPS: CleanupEntries = {
...
+      }
+      return false;
+    },
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?jaxsta\.com\/([^#?]+).*$/, 'https://jaxsta.com/$1');
+      url = url.replace(/^https:\/\/jaxsta\.com\/(\w+)\/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})(\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})?.*$/, 'https://jaxsta.com/$1/$2$3');
+      return url;
+    },
+    validate: function (url, id) {
+      const m = /^https:\/\/jaxsta\.com\/(\w+)\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}(?:\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})?$/.exec(url);
```
yvanzo, line 3001: To allow testing the presence of the second UUID:

```suggestion
      const m = /^https:\/\/jaxsta\.com\/(\w+)\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}(\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})?$/.exec(url);
```

yvanzo (approved):
Thanks. Just one last point on validating the double-UUID for Jaxsta variant/MB release. LGTMBDNT otherwise :ship:

reosarevok:
Documented under https://wiki.musicbrainz.org/Other_Databases_Relationship_Type/Whitelist#General_pages

PR #3005, opened 2023-08-09
MBS-11947: Stop hiding focus indicators on links

derat:
# MBS-11947

Remove the 'outline-style: none' CSS rule on a:hover to allow the UA to draw focus indicators on links. This was added by 2f155ec2 in 2010, but it impairs keyboard accessibility for the site and it's unclear why it was needed.

### DIFF
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -32,8 +32,7 @@ a {
         text-decoration: none;
     }
 
-
-    &:active, &:focus {
+    &:active {
         outline-style: none;
     }
 


yvanzo (commented):
Thank you for looking into this.

Did you test this patch locally? How and on which page? Is there any expected visual difference?

mwiencek (approved):
I had wondered before why this didn't work on MB.  Thanks for actually taking the time to figure it out. 👍  Tested it locally.  I can't think of any purpose this would serve (certainly none that would override accessibility concerns).

yvanzo (approved):
LGTMBDNT.

You did a good job by locating the CSS bit in cause and finding the original commit, unfortunately it’s not saying the reason for this change. I guess it was something related to some browser of that time.

derat:
Thanks for the quick review and merge!

PR #3006, opened 2023-08-09
MBS-13207 (II): Make recording focusable when adding release

derat:
# MBS-13207

Manually focus the first radio button when Tab is typed in the recording name input field at /release/add#recordings.

By default, only the currently-checked radio button is keyboard-focusable, making it impossible to use the keyboard to select an existing recording.

### DIFF
--- a/root/static/scripts/tests/release-editor/seeds/mbs-13207.html
+++ b/root/static/scripts/tests/release-editor/seeds/mbs-13207.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <form action="/release/add" method="POST">
+      <input type="hidden" name="name" value="MBS-13207" />
+      <input type="hidden" name="artist_credit.names.0.artist.name" value="Nine Inch Nails" />
+      <input type="hidden" name="artist_credit.names.0.mbid" value="b7ffd2af-418f-4be2-bdd1-22f8b48613da" />
+      <input type="hidden" name="mediums.0.format" value="Digital Media" />
+      <input type="hidden" name="mediums.0.track.0.name" value="Piggy" />
+      <input type="hidden" name="mediums.0.track.0.length" value="264000" />
+      <input type="hidden" name="edit_note" value="Testing MBS-13207" />
+      <button type="submit">Submit</button>
+    </form>
+  </body>
+</html>
--- a/root/release/edit/recordings.tt
+++ b/root/release/edit/recordings.tt
@@ -163,7 +163,7 @@
         [% l('Search:') %]
         <span class="autocomplete">
           [% React.embed(c, 'static/scripts/common/components/SearchIcon') %]
-          <input type="text" class="name" data-bind="autocomplete: { entity: 'recording', currentSelection: recording, lookupHook: $root.recordingAssociation.autocompleteHook($data) }" />
+          <input type="text" class="name" data-bind="autocomplete: { entity: 'recording', currentSelection: recording, lookupHook: $root.recordingAssociation.autocompleteHook($data) }" data-keydown="keydownEvent" />
         </span>
       </p>
 
--- a/root/static/scripts/release-editor/bubbles.js
+++ b/root/static/scripts/release-editor/bubbles.js
@@ -123,6 +123,34 @@ class RecordingBubble extends MB.Control.BubbleDoc {
   moveToTrack(track, stealFocus) {
     this.show(track.bubbleControlRecording, stealFocus);
   }
+
+  keydownEvent(data, event) {
+    /*
+     * Manually advance the focus to the first recording radio button if the
+     * Tab key is pressed in the recording name text input. By default, only
+     * the currently-checked radio button (which defaults to "Add a new
+     * recording" at the end of the list when entering a new tracklist) is
+     * keyboard-focusable. See MBS-13207.
+     *
+     * TODO: This only makes it possible to select the first radio button.
+     * Consider doing something similar to support tabbing to other unchecked
+     * radio buttons.
+     */
+    const noMods = !event.altKey && !event.ctrlKey && !event.metaKey &&
+      !event.shiftKey;
+    if (event.key === 'Tab' && noMods && !event.isDefaultPrevented()) {
+      const radio = document.querySelector(
+        '#recording-assoc-bubble input[name="recording-selection"]',
+      );
+      if (radio) {
+        radio.focus();
+        return false;
+      }
+    }
+
+    // Instruct Knockout to not call preventDefault.
+    return true;
+  }
 }
 
 releaseEditor.recordingBubble = new RecordingBubble('Recording');
--- a/t/selenium.mjs
+++ b/t/selenium.mjs
@@ -692,6 +692,7 @@ const seleniumTests = [
     name: 'release-editor/MBS-12077.json5',
     sql: 'vision_creation_newsun.sql',
   },
+  {name: 'release-editor/MBS-13207.json5', login: true},
   {
     name: 'Check_Duplicates.json5',
     login: true,
--- a/t/selenium/release-editor/MBS-13207.json5
+++ b/t/selenium/release-editor/MBS-13207.json5
@@ -1,0 +1,70 @@
+{
+  title: 'MBS-13207',
+  commands: [
+    {
+      command: 'open',
+      target: '/static/scripts/tests/release-editor/seeds/mbs-13207.html',
+      value: '',
+    },
+    {
+      command: 'clickAndWait',
+      target: 'css=button[type=submit]',
+      value: '',
+    },
+    {
+      command: 'pause',
+      target: '500',
+      value: '',
+    },
+    {
+      command: 'click',
+      target: 'css=a[href="#recordings"]',
+      value: '',
+    },
+    {
+      command: 'click',
+      target: 'css=button.edit-track-recording',
+      value: '',
+    },
+    {
+      command: 'pause',
+      target: '1000',
+      value: '',
+    },
+    // Check that there's a radio button for at least one existing recording
+    // in addition to the "Add a new recording" button -- if there's only one
+    // button, then we can't verify the fix.
+    {
+      command: 'assertEval',
+      target: 'document.querySelectorAll("#recording-assoc-bubble input[name=recording-selection]").length > 1',
+      value: 'true',
+    },
+    {
+      command: 'sendKeys',
+      target: 'css=#recording-assoc-bubble input.name',
+      value: '${KEY_TAB}',
+    },
+    // The first recording's radio button should receive the focus.
+    // Normally, only checked radio buttons are keyboard-focusable.
+    {
+      command: 'assertEval',
+      target: 'document.activeElement === document.querySelector("#recording-assoc-bubble input[name=recording-selection]")',
+      value: 'true',
+    },
+    {
+      command: 'open',
+      target: '/',
+      value: '',
+    },
+    {
+      command: 'handleAlert',
+      target: 'accept',
+      value: '',
+    },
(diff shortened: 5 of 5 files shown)


derat:
First time in this code, so it's possible that there's a better place to be doing this. I'm also still not sure how to run Selenium tests under musicbrainz-docker (or whether it's even possible), so there's a good chance that I'll need to debug the new test that I'm adding. :-)

I think that the js-perl-and-pgtap test failure was spurious; it failed immediately with a docker.io authentication error.

yvanzo:
> I'm also still not sure how to run Selenium tests under musicbrainz-docker (or whether it's even possible)

No worry, it isn’t possible at the moment.

derat:
> > I think that the js-perl-and-pgtap test failure was spurious
> 
> It’s not, these are `eslint` errors; see the logs:

Thanks, I'll fix those!

(I was referring to this CircleCI Dockerhub outage, which looks like it's since been resolved: https://status.circleci.com/incidents/9c1j07z24vxf)

root/static/scripts/tests/release-editor/seeds/mbs-13207.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <form action="/release/add" method="POST">
```
derat, line 4: I'd hardcoded the action URL to `https://musicbrainz.org/release/add` so I could test the seeding manually against a server with working search, but I forgot to change it back to `/release/add` before uploading the change. 😬

derat:
Phew, I think this is finally ready for review after a lot of futzing around with a new test.

I might create one or two followup changes to try to add additional ~hacks~ refinements to further improve how the radio buttons behave, e.g. making the up and down arrows not automatically advance to the next track and allowing recordings beyond the first one to be focused using the keyboard.

derat:
Would this change be accepted if it also handled moving the focus backward and to other radio buttons?

Porting the bubble to React is probably more than I'm up for with my current level of (non-)familiarity with this code; I was just trying to make a minimal change to allow the keyboard to be used for the common case of selecting the best recording for each track. (And unless moving to React also involves using something else instead of `<input type="radio">`, I suspect that a similar level of hackiness will still be needed there to work around the default browser behavior.)

After doing some digging, I think that the weird Up/Down arrow key focus behavior is due to the change associated with [MBS-7237](https://tickets.metabrainz.org/browse/MBS-7237), and specifically this code:

https://github.com/metabrainz/musicbrainz-server/blob/8f799067b1a8d7c70737ffc66696731b2d3c2611/root/static/scripts/edit/MB/Control/Bubble.js#L41-L43

My interpretation is that when the selected recording changes, `show` is called with the track's `Edit` button, which results in `deferFocus` being called to focus the first input in the bubble, i.e. the autocomplete text field.

From the original issue, it sounds like this was intended to only be used with the "Next" and "Previous" buttons in the tracklist and recordings editors (although as far as I can tell, it isn't running in the tracklist editor anymore). If there are no objections, I'll look into making a separate change to prevent the refocus when the radio buttons are clicked.

derat:
I think I figured out the cause of the text field stealing the focus and have opened #3008 with a proposed fix.

root/static/scripts/tests/release-editor/seeds/mbs-13207.html
yvanzo: You can switch “direct search” to use the entity search field in local development setup.

yvanzo (approved):
Thanks, the code looks good to me. We also want to drop Knockout but I guess it’s better than using jQuery.

I successfully tested this patch in combination with #3008 by duplicating the release `http://localhost:5000/release/0d9e2d13-3a3b-4823-b7bf-1f992697c527` locally in sample database.

derat:
Thanks for taking a look! After #3008 is in, I'll try to update the Selenium test that's added by this change to also verify that the arrow keys move the focus between the radio buttons.

mwiencek (commented):
Thanks for the PR. https://github.com/metabrainz/musicbrainz-server/pull/3008 was a definite improvement, but this one is a bit harder to follow, so I want to make sure I understand the issue.

> By default, only the currently-checked radio button is keyboard-focusable, making it impossible to use the keyboard to select an existing recording.

As one of your comments on the ticket says, it *is* possible with the up/down arrow keys, correct?  It seems to be for me.

I'm not sure it's a good idea to override what radio option is initially focused when tabbing to the group (deviating from the selected one being focused by default), since this won't be consistent with other websites or other pages on MB itself.

The main issues seem to be that (1) having to tab through any links from prior options before getting to the selected option is annoying, and that (2) "pressing an arrow key selects the adjacent input and then automatically moves on to the next track."

First, I'm not sure how to reproduce (2) after trying for a bit. Can you provide some step-by-step instructions?

For (1), this does seem to be a downside of using links inside option labels, which I don't think is a good practice in general, but is what we're stuck with for now.

As a middle ground, could we keep the user agent's default behavior of focusing the currently-selected option, while still skipping over any links prior to the selected option as this patch does?

derat:
The main editing pain point that I'm trying to address is when I'm adding a new release to an existing RG using a seeded (rather than cloned) tracklist. The first recording listed for each track is typically the correct one in that case, but I don't have a practical way to select it without moving the mouse to click on it. Even if it were easier to jump to the initially-selected radio button, that would be "Add a new recording" at the end of the list, so it'd probably take one or more up-arrow presses afterward to move to the first recording (and @yvanzo mentioned on IRC yesterday that there seems to be an existing issue where the recording list is sometimes reordered when radio buttons are suggested; this seems to primarily occur when moving away from "Add a new recording").

I considered proposing new UI to automatically select the first recordings for all tracks, but that seems a bit risky; I'm hesitant to assign tracks to recordings without user intervention.

I think that (2) might've been a misinterpretation on my part -- I suspect that I mistook the reordering, layout shift, and focus theft (corrected by #3008) as the next track getting selected.

For (1), if skipping over the links entirely is desirable, I think it'd be cleaner to do so by setting a negative `tabindex` on them. There may be some a11y concerns around doing that (but I think that the same concerns may apply if skipping them via JS in response to Tab key events in the search input).

yvanzo:
I did not open a ticket yet because this bug cannot be reproduced in production due to the search field stealing the focus when pressing arrow keys. But here is how to reproduce:

1. Run MusicBrainz Server code with commits from both pull requests #3006 and #3008 and a sample database
2. Open `http://localhost:5000/release/add`
3. Name the release `The Downward Spiral`
4. Select [`Nine Inch Nails`](http://localhost:5000/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da) for the artist (Switch the search to “direct search” if you don’t have a running and up-to-date Solr instance)
5. Select [`The Downward Spiral`](http://localhost:5000/release-group/7c4cab8d-dead-3870-b501-93c90fd0a580) for the release group
6. Go to the next tab `Release Duplicates`
7. Select the first release [`The Downward Spiral`](http://localhost:5000/release/0d9e2d13-3a3b-4823-b7bf-1f992697c527) _(‘Digital Audio Technologies Australia’ edition)_
8. Go to the next tab `Tracklist`
9. Check “I don’t know” about the format of the first medium/tracklist.
10. Go to the next tab `Recordings`
11. Click on the button “Edit” for the 1st track “Mr Self Destruct” which should open a new dialog on the right side
12. Select the option “Add a new recording” in this dialog which should pop up 3 new options for suggested recordings
13. Navigate up and down using arrow keys, the ordering of the options will unexpectedly change at some point.

mwiencek:
> The first recording listed for each track is typically the correct one in that case, but I don't have a practical way to select it without moving the mouse to click on it. Even if it were easier to jump to the initially-selected radio button, that would be "Add a new recording" at the end of the list, so it'd probably take one or more up-arrow presses afterward to move to the first recording

What if we moved "Add a new recording" from the bottom to the top?  Since it'll be both the first option and also selected by default, you'd only have to press the down arrow key once to select the first suggested recording.

> and @yvanzo mentioned on IRC yesterday that there seems to be an existing issue where the recording list is sometimes reordered when radio buttons are suggested; this seems to primarily occur when moving away from "Add a new recording"

That sounds annoying.  One possibility is that it found some new suggestions and inserted them into the list.  Another is that it's the same list of recordings but we're not using a consistent sort function.

> I considered proposing new UI to automatically select the first recordings for all tracks, but that seems a bit risky; I'm hesitant to assign tracks to recordings without user intervention.

Agreed, it would likely encourage people to select incorrect matches without thinking.

> For (1), if skipping over the links entirely is desirable, I think it'd be cleaner to do so by setting a negative tabindex on them. There may be some a11y concerns around doing that (but I think that the same concerns may apply if skipping them via JS in response to Tab key events in the search input).

If we skipped over them with JS, I think the idea is that you'd still be able to tab to an option's links after selecting it with the arrow keys (since the radio button will become focused and the links will come next in the tab index).  But it may not matter if you think moving the "Add a new recording" option to the top would suffice.

derat:
> What if we moved "Add a new recording" from the bottom to the top? Since it'll be both the first option and also selected by default, you'd only have to press the down arrow key once to select the first suggested recording.

I thought about this, but there are some potentially tricky bits:

1. The focus is initially set to the search input, so the interaction needed to choose the first recording would still be a bit complicated (Tab, Down, Enter). Definitely easier than what it is right now, though.
2. I wasn't sure if there are any cases where "Add a new recording" isn't selected by default, in which case it might be weird to put it at the top of the list. I thought that this might happen when cloning the tracklist from an existing release (in which case the track-recording associations are already there), but "Add a new recording" is actually the _only_ radio button in that case (surprisingly to me), so maybe it's fine.

> That sounds annoying. One possibility is that it found some new suggestions and inserted them into the list. Another is that it's the same list of recordings but we're not using a consistent sort function.

This is pretty easy to repro now on beta.musicbrainz.org, so I'll file a bug about it soon if no one else gets to it first.

> > For (1), if skipping over the links entirely is desirable, I think it'd be cleaner to do so by setting a negative tabindex on them. There may be some a11y concerns around doing that (but I think that the same concerns may apply if skipping them via JS in response to Tab key events in the search input).
> 
> If we skipped over them with JS, I think the idea is that you'd still be able to tab to an option's links after selecting it with the arrow keys (since the radio button will become focused and the links will come next in the tab index). But it may not matter if you think moving the "Add a new recording" option to the top would suffice.

I think that moving "Add a new recording" top the top first is worth a shot. I was worried that the recording-list-gets-reordered issue discussed above might make it a non-starter, but I don't see the reordering happen on beta.musicbrainz.org when no recording has yet been selected, so hopefully it won't cause problems here.

derat:
I've filed MBS-13242 to track the recordings being reordered when clicking "Add a new recording".

mwiencek (changes requested):
(Marking this as "changes requested" to make the status more clear for other maintainers.) From the latest comment it was agreed that moving "Add a new recording" to the top of the list is worth a try. I'd prefer that instead of changing the default radio button behavior.

derat:
Thanks, @mwiencek; I was meaning to get back to this soon. :-)

Does something like this look reasonable?

image

It maybe feels strange to have the "Suggested recordings:" heading appear midway through the list of radio buttons, but it  also feels strange to have it appear above the "Add a new recording" option.

Also, any preference regarding updating this PR vs. creating a new PR? (Personally, I'd lean toward creating a new one when going in a different direction like this.)

mwiencek:
@derat That looks reasonable to me (assuming they're still in the same radio button group). I suppose we could try it on beta and see what others think.

I'm not fussed about opening a new PR, but if you'd prefer, go for it. :) It's possible that the previous discussion here may be distracting or hard to follow given the new direction.

derat:
I'm closing this PR for now; hopefully #3069 will be enough.

PR #3007, opened 2023-08-10
Fix CI tests: Replace Perl short `$!` with long `$OS_ERROR`

yvanzo:
# Problem

In Perl `$!` is a short name for `$OS_ERROR` and the linter doesn’t allow short names, breaking CI tests.

It happened because I did not rebase the pull request #2350 on `master` before merging, and the Perl linter setup changed in the meantime in `master` branch, thus the CI tests did pass for this pull request even though they would not have if rebased.

# Solution

Replace all instances of Perl `$!` with its long name `$OS_ERROR`; See https://perldoc.perl.org/perlvar#$OS_ERROR

# Notes

- `admin/functions.sh` is not tested as it is a Bash file embedding a Perl snippet; Seems worth replacing too.
- `script/reset_selenium_env.sh` uses `$!` but it is a Bash variable for the PID of last executed background task.

### DIFF
--- a/admin/functions.sh
+++ b/admin/functions.sh
@@ -16,7 +16,7 @@ make_temp_dir()
                         DIR => shift(),
                         CLEANUP => 0,
                         TMPDIR => 1,
-                ) or die $!;
+                ) or die $!; # OS_ERROR
                 print $dir;
         ' "$@"
     ` || exit $?
--- a/lib/MusicBrainz/Script/JSONDump.pm
+++ b/lib/MusicBrainz/Script/JSONDump.pm
@@ -121,7 +121,7 @@ EOF
              catdir($mbdump->output_dir, $dump_fname)) or die $OS_ERROR;
     }
 
-    unlink "$dump_fpath.lock" or die $!;
+    unlink "$dump_fpath.lock" or die $OS_ERROR;
 
     return;
 }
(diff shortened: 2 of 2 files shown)


admin/functions.sh
```diff
@@ -16,7 +16,7 @@ make_temp_dir()
                         DIR => shift(),
                         CLEANUP => 0,
                         TMPDIR => 1,
-                ) or die $!;
+                ) or die $OS_ERROR;
```
reosarevok, line 19: Did you test this? It might not work without `use English` unless it's being embedded somewhere that already does that?

yvanzo: Good catch, just commenting instead in commit https://github.com/metabrainz/musicbrainz-server/pull/3007/commits/1052216aa96379b38c0fac991915a9b3f7b03df1.

reosarevok: You forgot to actually change the line :) But I agree with the idea - once you actually add a comment instead, this is ready to merge.

yvanzo: Yup, forgot to commit the change; now did in commit https://github.com/metabrainz/musicbrainz-server/pull/3007/commits/49e8b4f64c8bc1b3d53ca7e078baeb371bc3e853

reosarevok (approved):

PR #3009, opened 2023-08-11
MBS-13227: Omit "music quoted in" rels in tracklists

derat:
# MBS-13227
Avoid displaying "music quoted in" work-work relationships in tracklists. Works linked by these relationships are typically unrelated to the track in question, and other similar relationships (based on, arrangements, medleys, etc.) are already omitted.

### DIFF
--- a/root/components/GroupedTrackRelationships.js
+++ b/root/components/GroupedTrackRelationships.js
@@ -108,6 +108,8 @@ const irrelevantLinkTypes: {
     [315, false], // revisions
     [316, false], // orchestrations
     [350, false], // arrangements
+    [1046, true], // works quoting music from this
+    [1047, true], // works quoting lyrics from this
   ]),
 };
 


derat:
I tested this locally by creating a new work W1, adding a relationship stating that it quotes another work W0, and then checking that a release with a recording of W0 no longer lists the relationship to W1 in its tracklist.

(Please let me know if I should be including this in the original PR message rather than as a comment. I've been deleting the PR template since I'm hopefully already adequately explaining the change's intent in the commit message, but I'm still not 100% sure on MB's desired workflow.)

derat:
I also wasn't sure if I should be adding a new Selenium test for a change like this, or if there's an existing test that I should be updating -- I didn't see anything obvious among the filenames in t/selenium.

reosarevok (changes requested):
You will want to do the same with "lyrics quoted in" (`1047`), since those are a pair and there's the same issue with that one anyway :) Other than that this seems sensible.

reosarevok:
Generally it would be good to have a PR template that does something like:

# Problem
What is wrong

# Solution
What way you picked to solve what is wrong (it makes sense to separate the two because then people can understand the problem first, then propose a better solution for it if they have one).

# Testing
What did you do to test this works.

reosarevok:
I don't think we are testing this right now. We probably *should* be testing this, but I don't think it's fair to ask you to deal with that when we have not :)

derat:
Thanks, added 1047 and re-tested.

reosarevok (approved):

derat:
(Sorry, I fail at GitHub's wonky PR workflow -- I'm not sure how to mark your requested change as complete now that I've force-pushed the updated version.)

Noted, re the PR template; I'll follow that format for future changes.

yvanzo:
@derat: I’ve added the ticket reference in PR description because it is autolinked so making it easier to open the ticket.

derat:
> @derat: I’ve added the ticket reference in PR description because it is autolinked so making it easier to open the ticket.

Thanks, I noticed that on an earlier change and didn't realize it was you. :-) I'll include those in future PR descriptions as well.

PR #3010, opened 2023-08-11
MBS-13225: Allow skipping Confirm Form Submission page

derat:
# MBS-13225

When /release/add is being seeded via a cross-site POST request, automatically submit the "Confirm Form Submission" form if the request included a "skip_confirmation" query parameter set to "1". This allows editors to go straight to the seeded "Add Release" page without needing an extra click.

# Problem

Users need to click through an extra "Confirm Form Submission" interstitial whenever they seed a release via a POST request to `/release/add`. This page still needs to be loaded to ensure that the `session` and `remember_login` cookies are passed (due to `SameSite=strict` being used), but making the user interact with it seems unnecessary for preventing CSRF attacks since the actual edit submission happens via a different `/ws/js/edit/create` path.

# Solution

The form can automatically submit itself when it's being loaded for `/release/add`. Clients must include a `skip_confirmation` query parameter set to `1` in order to auto-submit to avoid causing regressions in user scripts or extensions that already automatically submit the form.

# Testing

I've manually tested this change's behavior locally by modifying my yambs seeder to pass `skip_confirmation=1` and verifying that the confirmation screen is skipped. (I also had to hack `set_csp_headers` in `lib/MusicBrainz/Server.pm` to bail out early to actually get scripts to run; I'm not sure if there's some issue in musicbrainz-docker that's preventing CSP hashes from getting updated when I make local changes.)

I also verified that the screen isn't skipped when I omit `skip_confirmation` for `/release/add` requests, and that `skip_confirmation` doesn't skip the confirmation screen for `/artist/create` POST requests (which would permit CSRF attacks), and that I get the login page as before when starting in a logged-out state (and that I end up at the seeded form after logging in).

I added Selenium tests for this, but then I realized that they probably won't work since the tests' seeding forms are presumably loaded from the same origin as the server, resulting in the confirmation screen never getting shown. Please let me know if I'm mistaken there.

### DIFF
--- a/root/main/ConfirmSeed.js
+++ b/root/main/ConfirmSeed.js
@@ -29,6 +29,17 @@ const ConfirmSeed = ({
 }: Props): React$Element<typeof Layout> => {
   const $c = React.useContext(SanitizedCatalystContext);
   const title = l('Confirm Form Submission');
+
+  /*
+   * Automatically submit the form if /release/add (where POST requests are
+   * used to seed fields rather than to create edits) was requested and the
+   * client indicated that they want the form to be skipped. See MBS-13225.
+   */
+  const url = new URL($c.req.uri);
+  const autoSubmit =
+    url.pathname === '/release/add' &&
+    url.searchParams.get('skip_confirmation') === '1';
+
   return (
     <Layout fullWidth title={title}>
       <h1>{title}</h1>
@@ -50,7 +61,7 @@ const ConfirmSeed = ({
       </p>
       <form method="post">
         {postParameters ? <PostParameters params={postParameters} /> : null}
-        <ConfirmSeedButtons />
+        <ConfirmSeedButtons autoSubmit={autoSubmit} />
       </form>
       {manifest.js('confirm-seed', {async: 'async'})}
     </Layout>
--- a/root/static/scripts/main/components/ConfirmSeedButtons.js
+++ b/root/static/scripts/main/components/ConfirmSeedButtons.js
@@ -7,33 +7,50 @@
  * later version: http://www.gnu.org/licenses/gpl-2.0.txt
  */
 
+import * as React from 'react';
+
 import DBDefs from '../../common/DBDefs-client.mjs';
 
-const ConfirmSeedButtons = (): React$MixedElement => (
-  <>
-    <button type="submit">
-      {l('Continue')}
-    </button>
-    <button
-      className="negative"
-      onClick={() => {
-        if (history.length > 1) {
-          history.back();
-        } else {
-          window.location.replace(
-            window.location.protocol + '//' +
-            DBDefs.WEB_SERVER,
-          );
-        }
-      }}
-      type="button"
-    >
-      {l('Leave')}
-    </button>
-  </>
-);
+type PropsT = {
+  +autoSubmit: boolean,
+};
+
+const ConfirmSeedButtons = ({
+  autoSubmit,
+}: PropsT): React$MixedElement => {
+  const submitRef = React.useRef<HTMLButtonElement | null>(null);
+  React.useEffect(() => {
+    if (autoSubmit) {
+      submitRef.current?.click();
+    }
+  });
+
+  return (
+    <>
+      <button ref={submitRef} type="submit">
+        {l('Continue')}
+      </button>
+      <button
+        className="negative"
+        onClick={() => {
+          if (history.length > 1) {
+            history.back();
+          } else {
+            window.location.replace(
+              window.location.protocol + '//' +
+              DBDefs.WEB_SERVER,
+            );
+          }
+        }}
+        type="button"
+      >
+        {l('Leave')}
+      </button>
+    </>
+  );
+};
 
 export default (hydrate(
   'span.buttons.confirm-seed',
   ConfirmSeedButtons,
-): React$AbstractComponent<{}, void>);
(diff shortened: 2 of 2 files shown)


root/main/ConfirmSeed.js
```diff
@@ -29,6 +29,17 @@ const ConfirmSeed = ({
 }: Props): React$Element<typeof Layout> => {
   const $c = React.useContext(SanitizedCatalystContext);
   const title = l('Confirm Form Submission');
+
```
derat, line 32: I spent a lot of time trying to submit the form directly from this component rather than clicking the button in `ConfirmSeedButtons`, but `React.useEffect` calls that I add here strangely never seem to execute. This is my first time using React, so presumably there's some subtlety that I'm missing.

root/main/ConfirmSeed.js
```diff
@@ -29,6 +29,17 @@ const ConfirmSeed = ({
...
+  /*
+   * Automatically submit the form if /release/add (where POST requests are
+   * used to seed fields rather than to create edits) was requested and the
+   * client indicated that they want the form to be skipped. See MBS-13225.
+   */
+  const url = new URL($c.req.uri);
+  const autoSubmit =
+    url.pathname === '/release/add' &&
+    url.searchParams.get('skip_confirmation') === '1';
```
derat, line 41: I'm requiring this parameter to make the new behavior opt-in, but I'm not sure whether there are actually any clients that try to click through the confirmation page. Let me know if you think it's safe to do this unconditionally.

derat, line 41: I used a body parameter for this initially, but it resulted in an "unknown parameter" error further down the line from the release form. Let me know if it seems like it'd be better to use a body parameter here but then strip it out so it doesn't get re-posted.

root/main/ConfirmSeed.js
```diff
@@ -29,6 +29,17 @@ const ConfirmSeed = ({
...
   const $c = React.useContext(SanitizedCatalystContext);
   const title = l('Confirm Form Submission');
+
+  /*
+   * Automatically submit the form if /release/add (where POST requests are
+   * used to seed fields rather than to create edits) was requested and the
+   * client indicated that they want the form to be skipped. See MBS-13225.
+   */
+  const url = new URL($c.req.uri);
+  const autoSubmit =
```
derat, line 39: Would it be better to make this determination in Perl and pass it to the component as a parameter?

root/main/ConfirmSeed.js
```diff
@@ -50,7 +61,7 @@ const ConfirmSeed = ({
       </p>
       <form method="post">
```
derat, line 62: Would it be better to remove query parameters here and set the `action` attribute to the bare path?

root/static/scripts/main/components/ConfirmSeedButtons.js
```diff
@@ -7,33 +7,50 @@
...
+  +autoSubmit: boolean,
+};
+
+const ConfirmSeedButtons = ({
+  autoSubmit,
+}: PropsT): React$MixedElement => {
+  const submitRef = React.useRef<HTMLButtonElement | null>(null);
+  React.useEffect(() => {
+    if (autoSubmit) {
+      submitRef.current?.click();
```
derat, line 24: @mwiencek, in the issue you mentioned using a delay before submitting the form. Please let me know if that's actually necessary.

root/main/ConfirmSeed.js
mwiencek: That would be because `ConfirmSeedButtons` is "hydrated" on the client (effects and event handlers are run in the browser), while `ConfirmSeed` is only run server-side.

root/main/ConfirmSeed.js
mwiencek: I think it doesn't make much difference.

root/main/ConfirmSeed.js
mwiencek: Opt-in for now seems safer at least, and doesn't cost much to support.  We can always remove it later if it's not needed.

root/main/ConfirmSeed.js
mwiencek: It shouldn't cause any issues to submit with `skip_confirmation` included AFAIK.

root/main/ConfirmSeed.js
derat: Thanks, that was my guess. :-)

root/static/scripts/main/components/ConfirmSeedButtons.js
derat: Thanks, I've added the dependency list and confirmed that this still works.

root/main/ConfirmSeed.js
derat: (resolving this since it sounds from another comment like it's fine to use a query param)

mwiencek (approved):
LGTMBDNT (will be easier to test on beta)

yvanzo (approved):
LGTMBDNT :ship:

derat:
Thanks!

root/static/scripts/main/components/ConfirmSeedButtons.js
derat: Sorry, I forgot to update this PR after addressing this suggestion. I've created #3018 to pass the dependency list.

PR #3011, opened 2023-08-14
MBS-2604: Infer link direction between people and groups

derat:
# MBS-2604

Make the relationship editor automatically infer the link direction for certain artist-artist relationships between individuals and groups (including orchestras and choirs).

For example, when a "member of" relationship with a person target is added to a group artist, it is automatically reversed.

This is intended to reduce the need to click the "Change direction" button and to make errors less common. Editors are still free to manually change the link direction before saving the relationship.

# Problem

Artist-artist relationships often default to being in the wrong direction. For example, when editing a group to add a "member of" relationship with a person as the target, the link direction needs to be manually reversed. Editors frequently forget to do this, resulting in incorrect data being entered.

# Solution

This change updates `RelationshipDialogContent.js` to infer the correct link direction whenever the link type or target entity MBID is modified.

# Testing

I've manually verified that relationships are automatically reversed now in the case described above (linking a member via a band's page), and that they aren't in the opposite scenario (linking a band via a member's page). If the approach that I'm taking seems reasonable, I'll work on adding Selenium tests that exercise this code (and updating existing tests, if needed).

### DIFF
--- a/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
+++ b/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
@@ -22,6 +22,10 @@ import {
   indexItems,
 } from '../../common/components/Autocomplete2/searchItems.js';
 import {
+  ARTIST_GROUP_TYPES,
+  ARTIST_TYPE_PERSON,
+} from '../../common/constants.js';
+import {
   createNonUrlRelatableEntityObject,
   createUrlObject,
 } from '../../common/entity2.js';
@@ -201,6 +205,53 @@ export function createInitialState(props: PropsT): RelationshipDialogStateT {
   };
 }
 
+/*
+ * Possibly updates newState.backward based on the source and target
+ * entities and the link type. The link direction is updated only in
+ * cases where the correct value can be inferred, e.g. "member of"
+ * relationships between person and group artist entities. See MBS-2604.
+ */
+function inferLinkDirection(
+  newState: {...RelationshipDialogStateT},
+  source: RelatableEntityT,
+): void {
+  const target = newState.targetEntity.target;
+  const linkTypeId =
+    newState.linkType.autocomplete.selectedItem?.entity?.id ?? null;
+
+  if (
+    source.entityType === 'artist' &&
+    target.entityType === 'artist' &&
+    linkTypeId !== null &&
+    personGroupLinkTypeIds.has(linkTypeId)
+  ) {
+    const isSourcePerson = source.typeID === ARTIST_TYPE_PERSON;
+    const isSourceGroup =
+      source.typeID !== null && ARTIST_GROUP_TYPES.has(source.typeID);
+
+    const isTargetPerson = target.typeID === ARTIST_TYPE_PERSON;
+    const isTargetGroup =
+      target.typeID !== null && ARTIST_GROUP_TYPES.has(target.typeID);
+
+
+    if (isSourcePerson && isTargetGroup) {
+      newState.backward = false;
+    } else if (isSourceGroup && isTargetPerson) {
+      newState.backward = true;
+    }
+  }
+}
+
+// Link types for artist-artist relationships between people and groups.
+const personGroupLinkTypeIds = new Set([
+  53, // collaborator
+  103, // member
+  305, // conductor
+  855, // composer-in-residence
+  895, // founder
+  965, // artistic director
+]);
+
 function updateDialogStateForTargetTypeChange(
   newState: {...RelationshipDialogStateT},
   oldTargetType: RelatableEntityTypeT,
@@ -322,6 +373,18 @@ export function reducer(
           action.source,
         );
       }
+
+      /*
+       * Avoid unnecessary calls when update-target-entity actions are
+       * mysteriously dispatched even when the target didn't change, e.g.
+       * a null-to-null update when clicking the artist selector.
+       */
+      const oldTargetGid = state.targetEntity.target.gid;
+      const newTargetGid = newState.targetEntity.target.gid;
+      if (oldTargetGid !== newTargetGid) {
+        inferLinkDirection(newState, action.source);
+      }
+
       break;
     }
 
@@ -400,6 +463,7 @@ export function reducer(
           ...newState.sourceEntity,
           error: getSourceError(action.source, null),
         };
+        inferLinkDirection(newState, action.source);
       }
 
       break;
--- a/root/static/scripts/common/constants.js
+++ b/root/static/scripts/common/constants.js
@@ -54,6 +54,15 @@ export const VOCAL_ROOT_ID = 3;
 export const AREA_TYPE_COUNTRY = 1;
 
 export const ARTIST_TYPE_PERSON = 1;
+export const ARTIST_TYPE_GROUP = 2;
+export const ARTIST_TYPE_ORCHESTRA = 5;
+export const ARTIST_TYPE_CHOIR = 6;
+
+export const ARTIST_GROUP_TYPES: Set<number> = new Set([
+  ARTIST_TYPE_CHOIR,
+  ARTIST_TYPE_GROUP,
+  ARTIST_TYPE_ORCHESTRA,
+]);
 
 export const CONTACT_URL = 'https://metabrainz.org/contact';
 
--- a/t/selenium.mjs
+++ b/t/selenium.mjs
@@ -635,6 +635,7 @@ async function handleCommand({command, target, value}, t) {
 /* eslint-disable sort-keys */
 const seleniumTests = [
   {name: 'Create_Account.json5'},
+  {name: 'MBS-2604.json5', login: true},
   {name: 'MBS-5387.json5', login: true},
   {name: 'MBS-7456.json5', login: true},
   {name: 'MBS-9548.json5'},
--- a/t/selenium/MBS-2604.json5
+++ b/t/selenium/MBS-2604.json5
@@ -1,0 +1,59 @@
+{
+  title: 'MBS-2604: Infer link direction between people and groups',
+  commands: [
+    // Link a person from a group's page and check that the relationship
+    // direction defaults to person being the member of the group.
+    {
+      command: 'open',
+      target: '/artist/b7ffd2af-418f-4be2-bdd1-22f8b48613da/edit',
+      value: '',
+    },
+    {
+      command: 'click',
+      target: 'css=#relationship-editor button.add-item',
+      value: '',
+    },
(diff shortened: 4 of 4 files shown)


root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
```diff
@@ -322,6 +377,13 @@ export function reducer(
           action.source,
         );
       }
+
+      const oldTargetGid = state.targetEntity.target.gid;
+      const newTargetGid = newState.targetEntity.target.gid;
+      if (oldTargetGid !== newTargetGid) {
```
derat, line 383: For reasons that I don't understand (unrelated to this change), `reducer` appears to be called multiple times whenever a change is made. For example, when I select an entity, I see two seemingly-identical `update-target-entity` calls come in changing the target from null to the selected entity. Is that expected?

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
```diff
@@ -201,6 +205,57 @@ export function createInitialState(props: PropsT): RelationshipDialogStateT {
   };
 }
 
+/*
+ * Possibly updates newState.backward based on the source and target
+ * entities and the link type. The link direction is updated only in
+ * cases where the correct value can be inferred, e.g. "member of"
+ * relationships between person and group artist entities. See MBS-2604.
+ */
+function inferLinkDirection(newState: {...RelationshipDialogStateT}): void {
```
derat, line 214: I wasn't sure if this function can live here or if it should be in a different file. Conceptually, it doesn't feel very different from the existing `updateDialogStateForTargetTypeChange` function that's defined after it.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
```diff
@@ -201,6 +205,57 @@ export function createInitialState(props: PropsT): RelationshipDialogStateT {
...
+
+  if (
+    source.entityType === 'artist' &&
+    target.entityType === 'artist' &&
+    linkTypeId !== null &&
+    personGroupLinkTypeIds.has(linkTypeId)
+  ) {
+    const sourcePerson = source.typeID === ARTIST_TYPE_PERSON;
+    const sourceGroup =
+      source.typeID !== null && groupArtistTypeIds.has(source.typeID);
```
derat, line 228: I haven't used Flow before, but I didn't see any obvious way to do something similar to TypeScript's non-null assertion operator... hence all these null checks.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
```diff
@@ -201,6 +205,57 @@ export function createInitialState(props: PropsT): RelationshipDialogStateT {
...
+    if (sourcePerson && targetGroup) {
+      newState.backward = false;
+    } else if (sourceGroup && targetPerson) {
+      newState.backward = true;
+    }
+  }
+}
+
+// Artist types corresponding to groups.
+const groupArtistTypeIds = new Set([
```
derat, line 243: Would it be better to make this a generic helper function that may be more likely to get updated if new group subtypes are added in the future? If so, where should it be?

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
yvanzo: It is probably safer to use our utility `nonEmpty`here.

derat: Sure, either way's fine with me -- `nonEmpty` has the advantage of being a bit shorter. :-)

I've made this change locally but I'll hold off on force-pushing since I assume it'll throw away all of these comments (sigh, GitHub).

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
yvanzo: This constant could be defined in `constants.js` instead.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
```diff
@@ -201,6 +205,57 @@ export function createInitialState(props: PropsT): RelationshipDialogStateT {
...
+
+    const targetPerson = target.typeID === ARTIST_TYPE_PERSON;
+    const targetGroup =
+      target.typeID !== null && groupArtistTypeIds.has(target.typeID);
+
+    if (sourcePerson && targetGroup) {
+      newState.backward = false;
+    } else if (sourceGroup && targetPerson) {
+      newState.backward = true;
+    }
```
yvanzo, line 238: Just nitpicking: Boolean variables could be more clearly named after statements than objects; for example `isSourcePerson`.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
yvanzo: It is a good place for now.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
yvanzo: I didn’t find why so it’s probably preferable to document it in the code itself as follows.

```suggestion
      // For some reason this update function may be called twice for a single change
      if (oldTargetGid !== newTargetGid) {
```

root/static/scripts/relationship-editor/types.js
```diff
@@ -216,8 +216,8 @@ export type DialogLinkTypeStateT = {
 
 export type DialogSourceEntityStateT = $ReadOnly<{
   ...DialogEntityCreditStateT,
-  +entityType: RelatableEntityTypeT,
   +error: React$Node,
+  +source: RelatableEntityT,
```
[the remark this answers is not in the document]
yvanzo: It seems fine but should be clarified that it doesn’t reflect the state of the entity in the editing page. The entity type cannot be changed in the entity editing page but the other fields can usually be changed, so the `source` here is actually reflecting the entity in its initial state before being edited. Maybe rename it to something like `frozenSource`?

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
yvanzo: Sorry but after double-checking this change isn’t even needed.

yvanzo (approved):
@derat: Thank you, I tested it locally and it works great. I tried to answer your questions. Feel free to make some of these minor changes.

@mwiencek: Should a CI test be added using Selenium here?

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
derat: Just to clarify, here are all the updates that I'm seeing:

When first clicking "Add relationship":
- update-link-type (no change in type)
- update-target-entity (null to null)
- update-link-type (no change)
- update-target-entity (null to null)

When clicking the "Relationship type" selector to open the dropdown:
- update-link-type (no change)
- update-link-type (no change)

When hovering over the "member of" item:
- update-link-type (no change)
- update-link-type (no change)

When clicking "member of":
- update-link-type (type changed)
- update-link-type (type changed)

When clicking the "Artist" selector to open the dropdown:
- update-target-entity (null to null)
- update-target-entity (null to null)

When hovering over an artist:
- update-target-entity (null to null)
- update-target-entity (null to null)

When clicking on an artist:
- update-target-entity (null to selected MBID)
- update-target-entity (null to selected MBID)

So this check prevents unnecessary calls in the null-to-null update-target-entity updates, but even when the type or artist actually changes I still hear about it twice, and I'm not sure if there's any way to detect that case. I suspect that something may be set up incorrectly on the React side, but I don't have enough experience with React to debug that any further.

root/static/scripts/relationship-editor/types.js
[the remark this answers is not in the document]
derat: Makes sense! I've renamed the property to `frozenSource` and added a comment.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
derat: I've added a comment describing what I'm seeing.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
derat: Sure, done.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
derat: Done.

derat:
> @mwiencek: Should a CI test be added using Selenium here?

I'll look into making sure that this is exercised by a Selenium test.

derat:
Okay, I've added a (speculative, since I'm using musicbrainz-docker) test for this. Let me know if it'd be better if it went through and created edits instead of just checking the preview text.

Some previous runs have failed, but I haven't yet determined if that's due to my change vs. flakiness. Please let me know if there's a link I should be clicking or some text in the console output that I can search for to find what actually failed.

root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
mwiencek: At a glance it sounds like the [expected behavior of strict mode in development](https://react.dev/reference/react/StrictMode#reference).

derat: Thanks, that was it -- I no longer see the duplicate actions after temporarily commenting out the strict mode tags. I've updated this comment to just refer to the null-to-null updates that the if statement is skipping.

derat:
@mwiencek, please let me know if you have any comments on the way that I'm doing this. I think that it's ready to go in from my perspective (the test that I added is passing).

derat:
@mwiencek helpfully pointed out on IRC that `DialogActionT` already contains the source entity for `update-target-entity` and `update-target-type` actions, so I've simplified this change to just use that.

yvanzo:

derat:

derat:
Maybe tomorrow will be a better day for running Selenium tests. :-/

mwiencek:
@brainzbot test this please

reosarevok:

yvanzo:

yvanzo:
Less crosswind today apparently.

derat:
Thanks for all the help in getting this merged!

PR #3012, opened 2023-08-16
MBS-13237: Clean up Twitch mobile links

reosarevok:
### Implement MBS-13237

# Problem
Twitch mobile links (such as https://m.twitch.tv/marimbamatt913) are rejected.

# Solution
These mobile links equivalent to the usual www. links (such as https://www.twitch.tv/marimbamatt913) and should just be converted into them, so this just does that.

# Testing
Added a test.


yvanzo (approved):
LGTMBDNT :ship:

PR #3013, opened 2023-08-16
MBS-13234 / MBS-13235: Handle ticketing links

reosarevok:
### Implement MBS-13234 / MBS-13235

# Description
We now have a "ticketing" relationship for artists, events, labels, places and series, so this handles a couple ticketing sites that our users had requested (a smaller one, MBS-13233, is not yet implemented since it's unclear if we want to clean and autoselect every country's ticketing platforms). Implementation is very similar, see each commit for exact details.

I also added sidebar display for these two large sites.

# Testing
Added tests, plus manual testing that the favicons appear when adding links to artist.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -5293,6 +5338,39 @@ const CLEANUPS: CleanupEntries = {
       return {result: isAProfile, target: ERROR_TARGETS.URL};
     },
   },
+  'ticketmaster': {
+    match: [new RegExp('^(https?://)?(www\\.)?ticketmaster\\.com', 'i')],
+    restrict: [LINK_TYPES.ticketing],
+    clean: function (url) {
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?ticketmaster\.com\/(?:[\w-]+\/)?(artist|event|venue)\/([0-9A-F]+).*$/, 'https://www.ticketmaster.com/$1/$2');
```
yvanzo, line 5345: Wasn’t there some redirect happening when removing the slug name from Ticketmaster URLs?

yvanzo (changes requested):
This is handling their US websites only but they have websites in other countries too. See https://www.livenationentertainment.com/global-sites/ (which is a list for both Ticketmaster and Live Nation as the company behind those merged in 2010.)

These websites seem to not share the same data as the same artist ID may refer to different artists from a country website to another. There might be some exceptions though such as Ticketmaster US and Canada websites that you can switch from one to the other from their page footer.

root/static/scripts/edit/URLCleanup.js
reosarevok: I think what you're thinking about is the redirect when moving from `concerts.livenation` to `www.livenation` :)

yvanzo: No, it was about the examples you shared on IRC.

reosarevok: Yeah, but those were event examples for livenation IIRC, not ticketmaster ones?

reosarevok: Yeah, but those were event examples for livenation IIRC, not ticketmaster ones?

yvanzo: Correct.

yvanzo (commented):
The last commit https://github.com/metabrainz/musicbrainz-server/commit/f749a9cc7a571182ae821801740fbcf63fb87711 for Live Nation global sites other than US `.com` looks good at first glance.

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3653,17 +3653,24 @@ const CLEANUPS: CleanupEntries = {
...
+      '(?:[a-z]{2,3}?\\.)?[a-z]{2,4}/',
       'i',
     )],
     restrict: [LINK_TYPES.ticketing],
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?concerts\.livenation\.com\/(?:[\w\d-]+\/)?event\/([0-9A-F]+).*$/, 'https://concerts.livenation.com/event/$1');
       url = url.replace(/^(?:https?:\/\/)?(?:www\.)?livenation\.com\/(artist|event|venue)\/([0-9a-zA-Z]+).*$/, 'https://www.livenation.com/$1/$2');
+      // International sites use somewhat different formats
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?livenation\.((?:[a-z]{2,3}?\.)?[a-z]{2,4})\/(show|venue)\/([0-9a-zA-Z]+).*$/, 'https://www.livenation.$1/$2/$3');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?livenation\.((?:[a-z]{2,3}?\.)?[a-z]{2,4})\/artist-[\w\d-]+-([0-9]+).*$/, 'https://www.livenation.$1/artist-$2');
```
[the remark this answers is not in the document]
reosarevok: Is it worth making the regex more complex by explicitly excluding `.com` if it doesn't cause issues? I guess we could, just not sure if it's needed :)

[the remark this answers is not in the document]
yvanzo: No, it just didn’t seem to be intended.

PR #3014, opened 2023-08-17
MBS-13165: Allow X ending for IdRef IDs

reosarevok:
### Fix MBS-13165

# Problem
IdRef URLs can actually end on X, but we're expecting only digits.

# Solution
Wikidata suggests it's always 8 digits, then a digit or X: https://www.wikidata.org/wiki/Property:P269

That said, because we've been burned before when trusting WD too much, I'm just allowing any number of digits and Xs - I don't expect this has much of a chance to lead to false positives unless someone starts making up IdRef IDs for some reason.

# Testing
Added a small test with an url with X ending.


yvanzo (approved):
LGTMBDNT :ship:

PR #3015, opened 2023-08-17
MBS-13214 / MBS-13238: Improve Open Library handling

reosarevok:
### Fix MBS-13214 / Implement MBS-13238

# Problem
A link without an ending slug, such as `https://openlibrary.org/works/OL20723256W?edition=`, does not get cleaned up. Same for any `/publishers` link (valid for labels, for example publishing audiobooks).

No validation happens at all for any of these links either.

# Solution
The first ticket here makes it so that cleanup does not depend on whether a slash appears after the OL ID, since that does not seem relevant to anything really (in fact, it seems some places, like the IA site, specifically link to pages without an ending slash and that might be a slightly preferable permalink).

The second ticket also allows cleaning up publisher links. Those don't use OL IDs for some reason, just plain text slugs, but they can certainly still be useful (some audiobook publishers have hundreds of entries).

The third ticket adds validation, limiting `/authors` links to artists, `/publishers` links to labels, `/books` links to releases (for audiobooks) and `/works` links to works. The only other links in the DB right now are 4 links to `/subjects` which seems to be for stuff books are *about* and I think it doesn't really make sense to link to anyway.

# Testing
Expanded the existing tests to also check validation, and added `/publishers` tests (including one with a comma and one with a hyphen to make sure we can deal with those fine).

### DIFF
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -4170,9 +4170,10 @@ limited_link_type_combinations: [
             expected_clean_url: 'https://openlibrary.org/books/OL8993487M/',
   },
   {
-                     input_url: 'https://openlibrary.org/works/OL82592W/',
+                     input_url: 'https://openlibrary.org/works/OL20723256W?edition=',
              input_entity_type: 'work',
     expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://openlibrary.org/works/OL20723256W/',
   },
   // Operabase
   {
--- a/root/static/scripts/tests/Control/URLCleanup.js
+++ b/root/static/scripts/tests/Control/URLCleanup.js
@@ -4162,30 +4162,35 @@ limited_link_type_combinations: [
                      input_url: 'https://openlibrary.org/authors/OL23919A/',
              input_entity_type: 'artist',
     expected_relationship_type: 'otherdatabases',
+       only_valid_entity_types: ['artist'],
   },
   {
                      input_url: 'http://openlibrary.org/publishers/Penguin_Books,_Limited',
              input_entity_type: 'label',
     expected_relationship_type: 'otherdatabases',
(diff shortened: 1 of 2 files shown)


root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4163,6 +4163,24 @@ limited_link_type_combinations: [
              input_entity_type: 'artist',
     expected_relationship_type: 'otherdatabases',
   },
+  {
+                     input_url: 'https://openlibrary.org/publishers/ISIS_Audio_Books',
+             input_entity_type: 'label',
+    expected_relationship_type: 'otherdatabases',
+            expected_clean_url: 'https://openlibrary.org/publishers/ISIS_Audio_Books',
+  },
```
yvanzo: This test is pretty redundant with the next one which also has underscores (`_`) in the slug name. I would suggest to drop this one which has no comma (`,`) in it. Note that `expected_clean_url` is tested to be stable through the function `clean` so the scheme `https` is tested too by the next test.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4162,41 +4162,48 @@ limited_link_type_combinations: [
...
     expected_relationship_type: 'otherdatabases',
             expected_clean_url: 'https://openlibrary.org/works/OL20723256W/',
+       only_valid_entity_types: ['work'],
   },
   {
                      input_url: 'https://openlibrary.org/works/OL82592W/',
              input_entity_type: 'work',
     expected_relationship_type: 'otherdatabases',
+       only_valid_entity_types: ['work'],
   },
```
yvanzo, line 4207: This test (not new to this pull request) is redundant with the two previous ones as it has the same `input_url` as their `expected_clean_url` which goes through the same tests. I would suggest to drop it too.

yvanzo (approved):
Thank you! The commit messages are very clear, the fixes seem sensible.  and the test coverage looks good too so I didn’t test it manually. Just spotted a couple of redundant tests, otherwise good to ship! :ferry:

PR #3016, opened 2023-08-17
MBS-13214 (II): Clean Open Library URLs with no trailing slash

reosarevok:
### Implement MBS-13214 (second part)

# Description
The Internet Archive and OL itself seem to prefer the no trailing slash format, so it probably makes sense for us to do the same.

# Testing
Modified the tests as needed.


yvanzo (approved):
LGTMBDNT :ship:

PR #3017, opened 2023-08-18
Fix translatable messages

yvanzo:
# Problem

Noticed a few minor issues while translating new messages in `beta`.

# Solution

See commit messages.

### DIFF
--- a/root/account/ChangePassword.js
+++ b/root/account/ChangePassword.js
@@ -66,7 +66,7 @@ const ChangePasswordPageContent = ({
         <FormRowText
           autoComplete="username"
           field={form.field.username}
-          label={l('Username:')}
+          label={addColonText(l('Username'))}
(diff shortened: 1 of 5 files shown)


reosarevok (changes requested):
`root/static/scripts/account/components/RegisterForm.js` and `root/user/Login.js` also do `l('Username:')` :)  (they also do `l('Password:)` but we don't seem to have a translation for that without the colon so maybe it's better not to touch that one)

yvanzo:
@reosarevok: Forgot to push these two other files too, thanks for noticing. Fixed in commit https://github.com/metabrainz/musicbrainz-server/pull/3017/commits/44462d1d4ba0115e02227dfe90ff7d79425135c1.

reosarevok (approved):

PR #3018, opened 2023-08-18
MBS-13225: Add deps to useEffect call in ConfirmSeedButtons.

derat:
I'm not sure if it has any effect in practice, but fix an oversight from 6f0b967e3d2f.

---

It looks like I forgot to amend my commit to include a change suggested by @mwiencek, namely passing a dependency list to the `React.useEffect()` call that clicks the submit button. I'm including the original ticket in the commit title, although I'm not sure if it's necessary (or desirable). I had already manually tested this locally.


reosarevok:

yvanzo (approved):
Thank you for the follow-up!

PR #3023, opened 2023-08-30
MBS-13250: Don't crash on missing text content on setlist

reosarevok:
### Fix MBS-13250

# Problem
After a setlist was added to an event, it started crashing with an error about trying to run `replace` on `undefined`.

# Solution
I found that this was caused by the setlist linking to works without work titles like this:
` * [58b51a5a-279b-476d-b27e-f44402faa675]`

We do explicitly support empty content, so this was just a bug - this just makes it so we don't try to run `replace` on `undefined` content.

# Testing
Added one row like this to the test setlist in our `formatSetlist` test.

### DIFF
--- a/root/static/scripts/common/utility/formatSetlist.js
+++ b/root/static/scripts/common/utility/formatSetlist.js
@@ -36,7 +36,10 @@ const linkRegExp =
  * all equivalent syntaxes are supported and documented for convenience.
  * https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references
  */
-function decodeSomeHTMLEntities(content: string) {
+function decodeSomeHTMLEntities(content: ?string) {
+  if (content == null) {
+    return '';
+  }
   return content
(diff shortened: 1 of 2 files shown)


root/static/scripts/common/utility/formatSetlist.js
```diff
@@ -37,6 +37,9 @@ const linkRegExp =
  * https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references
  */
 function decodeSomeHTMLEntities(content: string) {
+  if (!content) {
+    return '';
+  }
```
mwiencek, line 42: Even empty strings should have the replace method, so I guess the parameter type of `content: string` is also wrong and can be fixed?

reosarevok: Yeah, it was `undefined`, good point.

mwiencek (approved):

PR #3026, opened 2023-08-30
MBS-13260: Upgrade Node.js to v18

mwiencek:
# MBS-13260: Upgrade Node.js to v18

From the ticket: "MusicBrainz Server depends on Node.js 16 which is currently still an LTS in maintenance phase. Its EOL is the 11th September 2023. So it should be replaced with the next active LTS which is Node.js 18."

Since we plan to require v20 when that reaches LTS, and given the short notice, I haven't updated the requirements in INSTALL.md yet (though I've noted that we run v18 in production); I've also modified the CircleCI config to run tests against the previous musicbrainz-tests image which contains Node.js v16 in order to avoid any breakage in the meantime.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,12 @@ version: 2.1
 executors:
   musicbrainz-tests:
     docker:
+      - image: metabrainz/musicbrainz-tests:v-2023-08
+        user: root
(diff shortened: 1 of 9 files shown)


mwiencek:
@brainzbot test this please

yvanzo (approved):
Thanks a lot! :ship:

PR #3027, opened 2023-09-04
MBS-10629: Handle Anghami URLs (autoselect, validation, cleanup, sidebar)

reosarevok:
### Implement MBS-10629

# Description
Anghami is an Arab-world-centric streaming service for which we have around 1500 links already, so it makes sense to support it properly.

The service has a free, ad-supported tier, so it seems like a good fit for the "stream for free" relationship.

There are four main types of links we should support: `/artist`, `/album` (release), and `/song` and `/video` (recording).

As with other streaming sites, I made the sidebar text "Stream at Anghami".

# Testing
Added the usual tests plus some manual checking for favicons and such.


yvanzo (approved):
LGTMBDNT, just checked code and test URLs. :ship:

PR #3028, opened 2023-09-05
MBS-13224: Show edit/note restrictions from user profile

reosarevok:
### Implement MBS-13224

# Problem
We have for a long time chosen not to show whether a user was blocked from editing or leaving edit notes publicly in order to avoid publicly shaming editors.
Sadly, this means already banned editors get reported again, kind of wasting the time of both re-reporters and admins.
Additionally, it's unfair to the user *not* to display this info, since other editors can demand the user react to their edit notes without knowing they cannot.

# Solution
This shows active editing/voting an edit note restrictions for the user to other logged in users only and only when they visit their profile, which is the least shaming way of doing it but still visible if someone is thinking of re-reporting.

The "Untrusted" restriction is explicitly not shown to other users. "Untrusted" is usually set as a "hidden" "minor" restriction for users who just need an extra eye on their edits, and we don't even notify the restricted user of it, so it wouldn't make sense to notify others.

# Testing
Manually, plus added a test for this (and while at it turned another pointless test for this page into one that does something, on a separate commit).

### DIFF
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -282,7 +282,7 @@ const UserProfileInformation = ({
 
         {/* Only relevant to logged in users, avoid more public shame */}
         {viewingUser && restrictions.length ? (
-          <UserProfileProperty name={l(addColonText('Active restrictions'))}>
+          <UserProfileProperty name={l(addColonText('Restrictions'))}>
             {commaOnlyListText(restrictions)}
           </UserProfileProperty>
         ) : null}
--- a/root/user/UserProfile.js
+++ b/root/user/UserProfile.js
@@ -21,7 +21,8 @@ import * as TYPES from '../static/scripts/common/constants/editTypes.js';
 import {compare} from '../static/scripts/common/i18n.js';
 import commaList, {commaListText}
   from '../static/scripts/common/i18n/commaList.js';
-import commaOnlyList from '../static/scripts/common/i18n/commaOnlyList.js';
+import commaOnlyList, {commaOnlyListText}
+  from '../static/scripts/common/i18n/commaOnlyList.js';
 import expand2react from '../static/scripts/common/i18n/expand2react.js';
 import bracketed, {bracketedText}
   from '../static/scripts/common/utility/bracketed.js';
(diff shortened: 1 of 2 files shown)


reosarevok:

mwiencek (approved):
Tested locally and works well. I wonder if it'd make more sense to move this info underneath/closer to the "Report this user for bad behavior" link? That way it's less prominent if you're just looking at the profile info, but more prominent if you're looking to report the user?

reosarevok:
Hmm. The only issue with that is it being less visible for the "why is this person not answering my notes" use case :) But maybe? @Aerozol: opinions? Currently looks like this:

image_2023-09-05_213116964

yvanzo:
> visible for the "why is this person not answering my notes" use case

It seems a pretty common use case to care about indeed. We can keep this info under “General information” but also add a hint next to the report link. For example:

    Report this user for bad behavior (despite the above active restrictions)

Aerozol:
It looks fine to me (nitpick, I think ‘Active restrictions’ could be cut down to ‘Restrictions’).

I would probably advise against adding more text to ‘Report this user for bad behaviour’, if we don’t have to. Perhaps if someone has active restrictions we can add an asterisk, or a [?] box with a mouseover saying ‘user has current restrictions, see above’. But perhaps that’s best left for a follow-up ticket if reo finds that users are often reported repeatedly, after already being restricted.

root/user/UserProfile.js
```diff
@@ -277,6 +280,13 @@ const UserProfileInformation = ({
           {memberSince}
         </UserProfileProperty>
 
+        {/* Only relevant to logged in users, avoid more public shame */}
+        {viewingUser && restrictions.length ? (
+          <UserProfileProperty name={l(addColonText('Active restrictions'))}>
```
yvanzo, line 285: Suggestion by @Aerozol in an actionnable way:
```suggestion
          <UserProfileProperty name={l(addColonText('Restrictions'))}>
```

yvanzo:
Applied @Aerozol’s suggestion to cut down the wording.

Opened the follow-up ticket MBS-13267 for adding reminders when reporting a user that has restrictions, so we can merge the current changes already.

PR #3029, opened 2023-09-06
MBS-13268: Stop cleaning up CDBaby links

reosarevok:
### Implement MBS-13268

# Problem
CD Baby is gone anyway, so added URLs nowadays will be from old data. As mentioned in our forum
(https://community.metabrainz.org/t/cd-baby-store-closes-on-31-march/467757/12) we often could benefit from not cleaning those since the cleanup can make it harder to find archived versions of the data, and the "cleaned" URL is also broken by now anyway.

# Solution
This removes cleanup/validation, but keeps the autoselection of the cdbaby reltype for artists.

# Testing
None, other than keeping the cdbaby artist test but only with autoselection testing.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -1889,38 +1889,12 @@ const CLEANUPS: CleanupEntries = {
       return {result: false, target: ERROR_TARGETS.URL};
     },
   },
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
LGTMBDNT. Thanks! :ship:

PR #3031, opened 2023-09-06
MBS-13195: More useful sort for applications list

reosarevok:
### Implement MBS-13195

# Problem
The current sort of the authorized applications list (by application ID) means nothing to users, so it feels effectively random. 

# Solution
The two best options seemed to be granted token time (does not require extra joins, but keeps similar applications separate) or application name (groups similar applications; needs one extra join).

It feels to me that the benefit of name is worth the join on this table since it shouldn't add a significant delay (application is relatively small). I'm doing a second level of sorting on token date (newest first) since it seems to also make the table nicer when there's several entries for a name.

# Testing
Manually with my own authorized apps (which include duplicate app names)


reosarevok:

mwiencek (approved):
Did not test, but the query looks sane.

PR #3033, opened 2023-09-08
MBS-13244: Add standalone recording filter to artist index 

reosarevok:
### Implement MBS-13244

# Problem
When we fall back to showing recordings for an artist overview, we still use the release group filter, which unsurprisingly is useless.

# Solution
This makes it so that if the artist has no RGs, the filter loaded becomes the recording one, which actually filters the returned list of standalone recordings as well.

# Testing
Manually, with the artist mentioned in the ticket (`/artist/c48c24dd-d3d5-4021-9098-6ef2190d8351`) for recordings and Nine Inch Nails to make sure it still works for RGs.


mwiencek (approved):

PR #3035, opened 2023-09-11
MBS-13145: Also show collection tab to logged out viewers

reosarevok:
### Implement MBS-13145

# Problem
We currently require login to see any data about collections for a specific entity, even when the entity is in a public collection that can otherwise be seen when logged out.
I don't think this is really intentional, but more of a side effect of the way the collections page / sidebar got implemented. I asked @Aerozol and he also suggested that this data should be shown to everyone, since it's interesting to showcase.

# Solution

This shows only the "in X collections" link on the sidebar for logged out users (all others seem irrelevant). It also allows them to see the `/collections` page for the entity, showing only public collections.
We still want to pass `show_private_only` for `_all_non_visible_collections` when there's no user, since otherwise it will not restrict it to non-visible collections, but we don't have an editor ID to pass; I chose to just pass `'none'` as a string, then make it skip the ID-related conditions if that is passed.

# Testing
Manually, by navigating to an entity in one of my public collections, seeing the sidebar shows it's in one collection, clicking through to the collections page and seeing it correctly links to my collection. Also checked the page is linked but empty when an entity is in 0 collections.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Collection.pm
+++ b/lib/MusicBrainz/Server/Data/Collection.pm
@@ -274,10 +274,13 @@ sub find_by {
         push @args, $editor_id, $editor_id;
     } elsif ($editor_id = $opts->{show_private_only}) {
         push @conditions, 'editor_collection.public = false';
-        push @conditions, 'editor_collection.editor != ?';
-        push @conditions, 'NOT EXISTS (SELECT 1 FROM editor_collection_collaborator ecc
-                            WHERE ecc.collection = editor_collection.id AND ecc.editor = ?)';
-        push @args, $editor_id, $editor_id;
+        # We skip this if there's no logged in editor
+        unless ($editor_id eq 'none') {
+            push @conditions, 'editor_collection.editor != ?';
+            push @conditions, 'NOT EXISTS (SELECT 1 FROM editor_collection_collaborator ecc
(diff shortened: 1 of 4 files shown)


lib/MusicBrainz/Server/Data/Collection.pm
```diff
@@ -274,10 +274,13 @@ sub find_by {
         push @args, $editor_id, $editor_id;
     } elsif ($editor_id = $opts->{show_private_only}) {
```
mwiencek, line 275: OK, I hadn't looked at this code in a while and was confused what this was for (was wondering why we are returning all private collections other than an editor's own to that editor) but it's just to get the "hidden" count. In the future, I feel it'd be safer to have a separate method that just returns the count, since the actual collections aren't used AFAICT.

mwiencek (approved):

lib/MusicBrainz/Server/Data/Collection.pm
reosarevok: You do have a very good point - pushed a second commit here that should do the trick.

mwiencek: didn't test the new method but lgtm :shipit:

mwiencek (approved):

PR #3037, opened 2023-09-12
MBS-13252: Report: Show notes links on non-Broadcast releases

reosarevok:
### Implement MBS-13252

# Description
Show notes is a relationship meant to link podcasts and similar broadcast releases to their, well, show notes. Users do seem to use it quite a bit to link to anything that has anything to do with any release, whether it is a podcast or not, probably because it sounds more generic than other choices. This should make it easier for editors to find these bad edits, fix them, and maybe let the user know so they stop doing it.

# Testing
Ran the report against the sample DB, found a few cases, all seemingly correct(ly incorrect).


mwiencek (approved):
Did not test, but the query looks straightforward.

PR #3038, opened 2023-09-18
MBS-13278: Show label type descriptions on edit form

reosarevok:
### Implement MBS-13278

# Problem
People are using the wrong label types relatively often (see MBS-13277 for an example); while it's probably not only because descriptions are missing, it certainly wouldn't hurt to have them and might avoid a fair amount of the misuse.

# Solution
This just uses the same `typeBubble` solution we already use for other entities; the main reason we didn't implement it here before is that we don't actually have descriptions for any of the types, which we need to add (little by little probably as to avoid overwhelming sir as much as possible).

# Action
1. Add descriptions to all label types

# Testing
Manually, on test data where I added a few descriptions.


mwiencek (approved):
Tested locally by updating some label type descriptions in the DB.

PR #3039, opened 2023-09-18
MBS-8952: Can rename recordings via autocomplete

mwiencek:
# MBS-8952

## Problem

If you select any recording which was obtained from the search server in a "Suggested recordings:" list in the release editor, then updating the recording autocomplete text will allow you to change the recording's name and unintentionally submit edits doing the same.

## Solution

I was able to reproduce this using the steps in https://tickets.metabrainz.org/browse/MBS-13021.

The primary issue is that recordings from the search server don't include a database row ID, so the `if (currentSelection.id)` check modified here didn't clear the selection when modifying text. The `else` branch updated the recording name with the entered text and notified all subscribers of the autocomplete observable.

Checking for `currentSelection.gid` there fixes the bug. I also changed the edit generation code to revert any recording changes if the necessary settings to update them aren't enabled.

# Testing

Tested locally using a local search server.

### DIFF
--- a/root/static/scripts/common/MB/Control/Autocomplete.js
+++ b/root/static/scripts/common/MB/Control/Autocomplete.js
@@ -288,7 +288,11 @@ $.widget('mb.entitylookup', $.ui.autocomplete, {
      * entity (like release group types).
      */
 
-    if (currentSelection.id) {
+    if (
+      // entities from the search server will not have a numeric id
+      currentSelection.id ||
+      currentSelection.gid
+    ) {
       this.currentSelection(this._dataToEntity({name: name}));
     } else if (currentSelection.name !== name) {
       currentSelection.name = name;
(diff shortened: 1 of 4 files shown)


reosarevok:
I understand some tests are coming? This looks good otherwise :)

mwiencek:
Trying to figure out a heisenbug in the added Selenium test. It passes locally and on paco (the same host and Docker image that the tests use!) when run manually. Perhaps some state is leaking between different test runs.

mwiencek:
> Perhaps some state is leaking between different test runs.

Yes, the track parser cookies.

mwiencek:
@brainzbot test this please

reosarevok (approved):
Code LGTM, test seems to make sense AFAICT and passes.

PR #3040, opened 2023-09-20
Clarify translation steps for beta release

reosarevok:
It was previously unclear where update_pot.sh should be run now, and how .pot changes would reach beta.


yvanzo (approved):
Thank you! :ship:

PR #3041, opened 2023-09-20
MBS-13279: Don't consider the empty string an email address

reosarevok:
### Fix MBS-13279

# Problem
We use `has_email_address` as a way to see whether the user has an email set that we can write to, but it's currently a Moose predicate for `email`. This means that if `email` is set to *anything*, it will return true, including the empty string (which we have for a couple thousand accounts) and, theoretically at least, `undef`. See https://metacpan.org/dist/Moose/view/lib/Moose/Manual/Attributes.pod#Predicate-and-clearer-methods

# Solution
I changed this to just use the predicate as the first check; if email is set it will make sure it's defined and not the empty string.

# Testing
Manually, checking the user mentioned on the ticket now shows `(none)` for their email.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Editor.pm
+++ b/lib/MusicBrainz/Server/Entity/Editor.pm
@@ -6,7 +6,11 @@ use Authen::Passphrase;
 use DateTime;
 use Encode;
 use MusicBrainz::Server::Constants qw( $PASSPHRASE_BCRYPT_COST );
-use MusicBrainz::Server::Data::Utils qw( boolean_to_json datetime_to_iso8601 );
+use MusicBrainz::Server::Data::Utils qw(
+    boolean_to_json
+    datetime_to_iso8601
+    non_empty
+);
 use MusicBrainz::Server::Entity::Preferences;
 use MusicBrainz::Server::Entity::Types qw( Area ); ## no critic 'ProhibitUnusedImport'
 use MusicBrainz::Server::Entity::Util::JSON qw( to_json_array to_json_object );
@@ -128,8 +132,13 @@ sub public_privileges {
 has 'email' => (
     is        => 'rw',
     isa       => 'Str',
-    predicate => 'has_email_address',
 );
+
+sub has_email_address
+{
+    my $self = shift;
+    return non_empty($self->email);
+}
 
 sub has_confirmed_email_address
 {


lib/MusicBrainz/Server/Entity/Editor.pm
```diff
@@ -128,9 +128,17 @@ sub public_privileges {
 has 'email' => (
     is        => 'rw',
     isa       => 'Str',
-    predicate => 'has_email_address',
+    predicate => 'has_email_set',
 );
 
+sub has_email_address
```
yvanzo, line 134: Has this exact change really been tested?
Where is `has_email_set` defined?
Where is `has_email_address` now used?

yvanzo (commented):
The general idea looks good but there might be an oversight in the implementation.

lib/MusicBrainz/Server/Entity/Editor.pm
reosarevok: `has_email_address` is used in the exact same places it was before. It just adds the extra checks to it.

`has_email_set` is defined here :) See the link about predicates in the PR description.

yvanzo: The link doesn’t mention any `has_email_set`.

reosarevok: `has_email_set` (same as `has_email_address` before) is just the name given to check the `predicate` for `email` which is just a default "method [that] tells you whether or not a given attribute is currently set"

yvanzo: Tentative to clarify the commit message:

> Previously `has_email_address` was defined as a predicate for whether or
> not the email address was set, even if set as an empty string.
> 
> This patch renames the predicate to `has_email_set` and defines
> `has_email_address` as a subroutine that also checks whether or not the
> email is non-empty string.
> 
> For reference, see the documentation for Moose predicates at
> https://metacpan.org/dist/Moose/view/lib/Moose/Manual/Attributes.pod#Predicate-and-clearer-methods

mwiencek (approved):

yvanzo (approved):
Typo “defnes” in the commit message, otherwise LGTMBDNT. :ship:

PR #3044, opened 2023-09-21
MBS-13281: Don't check for mail confirmation for beginners

reosarevok:
### Implement MBS-13281

# Problem
Checking for email confirmation in `is_limited` probably made sense when we called this "Limited user" (unconfirmed users are quite limited) but it doesn't make sense now that we explicitly use `is_limited` for beginners and claim everywhere the user is "new to MusicBrainz", since a user can remove their email even with thousands of edits.

# Solution
This just removes the no `email_confirmation_date` condition for `is_limited`. `editorMayVote` already checks for `email_confirmation_date` separately. Elsewhere, we seem to be using `is_limited` just for beginner checks. `Predicate::Role::User` already was not looking at mail confirmation.

# Testing
Tested with user `test`, which now shows as "Normal User"

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Editor.pm
+++ b/lib/MusicBrainz/Server/Entity/Editor.pm
@@ -193,10 +193,7 @@ sub is_limited
     return
         !($self->id == $EDITOR_MODBOT) &&
         !$self->deleted &&
-        ( !$self->email_confirmation_date ||
-          $self->is_newbie ||
-          !$self->has_ten_accepted_edits
-        );
+        ($self->is_newbie || !$self->has_ten_accepted_edits);
 }
 
 has birth_date => (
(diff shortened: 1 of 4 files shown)


mwiencek:
We currently force showing the "new notes have been left..." banner based on the `is_limited` check. But seems like it'd make sense to still check `has_confirmed_email_address` in that case, since the user may not be receiving email notifications about the edits otherwise.

reosarevok:
They cannot *make* edits without an email, but I guess we might want that to still show notifications for existing edits.

mwiencek:
The only other consequential check I can see is where we use `is_limited` to restrict showing collection descriptions and editor bios/homepages to prevent spam, but it seems sufficient for that, since the main concern is with new accounts.

mwiencek (approved):

PR #3046, opened 2023-09-21
MBS-13155: Don't consider all unselected labels dupes

reosarevok:
### Fix MBS-13155

# Problem
`releaseLabelKey` automatically defaults to '' as the label key if no ID exists, meaning two labels such as "A" and "B" get seen as a dupe until an actual label with an ID is selected. This has confused users in the past, since as far as they could tell the labels are different.

# Solution
This just makes it so that a label name is also taken into account when calculating the key. I chose to add a `name-` prefix to it to avoid the very unlikely but possible situation where a "label name" that matches the other label's row ID would trigger a "duplicate" error.

# Testing
Manually, by entering two labels (A and B) without selecting a search result and checking the duplicate is no longer shown, but checking it's still shown if I enter A twice. Also added test cases (and improved the relevant test a bit).

### DIFF
--- a/root/static/scripts/common/utility/releaseLabelKey.js
+++ b/root/static/scripts/common/utility/releaseLabelKey.js
@@ -15,7 +15,9 @@ export default function releaseLabelKey(releaseLabel: ReleaseLabelT): string {
   let result = '';
   const label = ko.unwrap(releaseLabel.label);
   if (label) {
-    result += (label.id || '');
+    result += (
+      label.id || (nonEmpty(label.name) ? 'name-' + label.name : '')
+    );
   }
   result += '\0';
   result += clean(ko.unwrap(releaseLabel.catalogNumber));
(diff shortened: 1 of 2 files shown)


root/static/scripts/common/utility/releaseLabelKey.js
```diff
@@ -15,7 +15,7 @@ export default function releaseLabelKey(releaseLabel: ReleaseLabelT): string {
   let result = '';
   const label = ko.unwrap(releaseLabel.label);
   if (label) {
-    result += (label.id || '');
+    result += (label.id || nonEmpty(label.name) ? 'name-' + label.name : '');
```
mwiencek, line 18: `||` has higher precedence than `?` so I'm not sure this is what was intended, and probably indicates a test is missing (same label name but different ids?)
```suggestion
    result += (label.id || (nonEmpty(label.name) ? 'name-' + label.name : ''));
```

reosarevok: Good point. Added an extra test.

mwiencek (approved):

PR #3047, opened 2023-09-27
MBS-13284: Don't require edit note for alias removals

reosarevok:
### Implement MBS-13284

# Problem
This was originally added on my request over 10 years ago, but while it makes sense to require a note for all entity removals and other similarly destructive edits, alias removals just drop an MBID-less text field that is trivial to re-add, and is not significantly different from dropping a disambiguation or other similar text field.

# Solution
Just drop the `requires_edit_note` argument from `delete_alias`.

# Testing
Amusingly, we had more tests checking for this edit note requirement than most other more important ones...
I removed most, but left one testing it's no longer required.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Role/Alias.pm
+++ b/lib/MusicBrainz/Server/Controller/Role/Alias.pm
@@ -135,7 +135,6 @@ sub delete_alias : Chained('alias') PathPart('delete') Edit
     cancel_or_action($c, $edit, $self->_aliases_url($c), sub {
         $self->edit_action($c,
             form => 'Confirm',
-            form_args => { requires_edit_note => 1 },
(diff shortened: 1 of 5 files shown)


mwiencek:
I still see some code in the tests like:
```Perl
                    'confirm.edit_note' =>
                        q(Some edit note since it's required)
```
which can probably be removed (indirectly testing that an edit note is not required) or at the very least re-worded.

yvanzo (changes requested):
See the above comment. LGTMBDNT otherwise!

reosarevok:
Updated these.

mwiencek (approved):

yvanzo (approved):
Thank you! :ship:

PR #3048, opened 2023-10-03
Remove all uses of the smartmatch operator and switch feature

mwiencek:
The smartmatch operator (`~~`) and switch feature (`given`/`when`) are both deprecated as of Perl 5.38.

https://perldoc.perl.org/perl5380delta#Switch-and-Smart-Match-operator

### DIFF
--- a/contrib/slowdown.psgi
+++ b/contrib/slowdown.psgi
@@ -1,7 +1,6 @@
 #!/usr/bin/env perl
 use strict;
 use warnings;
-use feature "switch";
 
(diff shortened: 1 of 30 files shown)


reosarevok (approved):
LGTMBDNT

reosarevok:
Now I tested a little bit - but this affects all sort of things all around the site so I expect there's no actual good way to test everything properly other than expecting the tests test the most important things and any small issue there might be is caught in beta tbh.

PR #3049, opened 2023-10-04
MBS-13307: Fix ListHeader quick links key warning

reosarevok:
### Fix MBS-13307

# Problem
Loading any page using `ListHeader` (such as edit searches) was logging a warning that the `<td>` in `QuickLinksSection` should have a key.

# Solution
I looked at the code and it seemed stupid to have the `<td>` in the `QuickLinksSection` return to begin with, so rather than adding a key, I am just returning the actual quick links and keeping the `<td>` as part of the `ListHeader` structure.

# Testing
Checked the edit search header still loaded and no longer logged a warning.


root/edit/components/ListHeader.js
```diff
@@ -174,15 +174,13 @@ const QuickLinksSection = ({
...
-    <td>
-      {quickLinks.reduce((accum: Array<React$Node>, link, index) => {
-        accum.push(link);
-        if (index < (quickLinks.length - 1)) {
-          accum.push(' | ');
-        }
-        return accum;
-      }, [])}
-    </td>
+    quickLinks.reduce((accum: Array<React$Node>, link, index) => {
```
mwiencek, line 177: Not sure I understand why this works, since components that return arrays still need keys on the elements in the array (or at least they should). Try editing the `MyButton` example on https://react.dev/learn and change the `return (...)` to `return [...]`.

mwiencek: I tried to debug this and I honestly can't tell if it's a React bug that the key warning isn't logged here -- it definitely logs it if you return an array elsewhere, so it seems like to intent is that these should have keys at least.

mwiencek (approved):

PR #3050, opened 2023-10-05
[WIP] [EXPERIMENTAL] edit search and other list improvements

mwiencek:


mwiencek:
The commit "[remove limit for edit listings](https://github.com/metabrainz/musicbrainz-server/pull/3050/commits/1fd3f6aa5868d936469cd343309aa8e6c95e4ccb)" only improves one query I tested, and introduces performance issues on others. Interestingly, the query it improves goes from timing out to finishing quickly, while the others go from finishing quickly to timing out. :shrug:

I'll try to parameterize whether the `LIMIT 500` is included, so we can disable it for the one query I tested. For now I'll close this and salvage the remaining commits into another branch.

PR #3052, opened 2023-10-09
MBS-13317: Remove Kget.jp from lyrics whitelist

reosarevok:
### Description
The kget.jp site apparently closed in March. Since it's a lyrics site and no longer licensed I will remove the existing URLs with MBBE-82 rather than setting them as ended.

This removes all our handling for the site, including the favicon etc.


yvanzo (approved):
LGTMBDNT :ship:

For the record, the only related documentation page [Lyrics whitelist](https://musicbrainz.org/doc/Style/Relationships/URLs/Lyrics_whitelist) has already been updated accordingly.

PR #3055, opened 2023-10-19
Normalize plural form and capitalization of Various Artists

yvanzo:
# Problem

The special-purpose artist entity “[Various Artists](https://musicbrainz.org/artist/89ad4ac3-39f7-470e-963a-56509c546377)” is sometimes written in its singular form or in lowercase. Inconsistently using different forms isn’t always clear that it specifically refers to this SPA entity. It also makes translation slightly more complex. (Noticed while reviewing the translation glossary.)

# Solution

Use the most common form “Various Artists” all the time.

# Action

1. Normalize the two occurrences of “various artists” in the description of the release group secondary type “compilation”.

### DIFF
--- a/root/artist/ArtistIndex.js
+++ b/root/artist/ArtistIndex.js
@@ -103,14 +103,14 @@ const FooterSwitch = ({
     if (showVA) {
       links.push(
         <a href={`${artistLink}?va=1`} key="show-va">
-          {l('Show official various artist release groups')}
+          {l('Show official various artists release groups')}
         </a>,
       );
     }
     if (showAllVA) {
       links.push(
         <a href={`${artistLink}?all=1&va=1`} key="show-all-va">
-          {l('Show all various artist release groups')}
+          {l('Show all various artists release groups')}
         </a>,
       );
     }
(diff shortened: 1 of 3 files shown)


reosarevok:
Hmm. This is not quite the same, now that I think about it, is it? The `va` filters don't (only) show RGs by Various Artists, they show any track appearances where the RG artist is not the same as the track artist (so they'll also show featured tracks on single-artist albums, for example, which are not credited to Various Artists).

Not sure what's a good wording for this, FWIW, since it's not like it being lowercase makes that more clear.

reosarevok:
An IRC discussion got to ""Show appearances on official release groups by other artists / Various Artists" as possibly the shortest option that is still both true and clear. That'd then be ""Show appearances on all release groups by other artists / Various Artists", "This artist only has appearances on unofficial release groups by other artists / Various Artists", etc. for RGs.

For releases, I guess something similar like "Showing appearances on releases by other artists / Various Artists", "This search only found appearances on releases by other artists / Various Artists", etc.? Does that sound right?

The statistics do seem to refer to Various Artists specifically and the change there seems entirely correct.

PR #3056, opened 2023-10-19
Normalize "pending edits" to "open edits" in UI

reosarevok:
# Description
"pending edits" are, in the way we are using the term, exactly the same as "open edits". As such, this normalizes the user-facing strings so that we always use "open edits" when we mean that.

We could arguably use "pending edits" for edits that still have not been entered, like the ones that will be created by an open release editor or relationship editor session.
That would seem like a much better use of the term, but sadly might require a huge refactoring since we use "edits_pending" in our schema for open edits.

# Testing
None, but this changes strings only. I made sure this should be all of the translatable "pending" strings with a Weblate search.

### DIFF
--- a/root/static/scripts/common/components/ArtistCreditLink.js
+++ b/root/static/scripts/common/components/ArtistCreditLink.js
@@ -46,12 +46,12 @@ export const MpIcon = (hydrate<MpIconProps>('span.ac-mp', (
   return (
     <Tooltip
       content={exp.l(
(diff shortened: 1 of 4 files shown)


mwiencek (approved):
> We could arguably use "pending edits" for edits that still have not been entered, like the ones that will be created by an open release editor or relationship editor session.

This makes sense to me.

PR #3059, opened 2023-10-20
Upgrade Flow to 0.219.2 and make a few small type improvements

mwiencek:
This PR simply upgrades Flow to the latest version, continuing the progress made with https://github.com/metabrainz/musicbrainz-server/pull/2982.

The only real problematic version was 0.217.0, which removed support for variance annotations on functions. Removing those required fixing some new type errors.

There are a few additional, miscellaneous commits making small improvements to the types:

 * "Replace $MakeReadOnly type with one using infer" makes use of the new [conditional types](https://flow.org/en/docs/types/conditional/) feature, replacing the use of `$Call`, which was less obvious.
 * "Remove uses of any in SeriesIndex.js" is an example of using [mapped types](https://flow.org/en/docs/types/mapped-types/).

### DIFF
--- a/root/static/scripts/common/i18n/expand2.js
+++ b/root/static/scripts/common/i18n/expand2.js
@@ -34,14 +34,14 @@ export interface VarArgsClass<+T> {
   has(name: string): boolean,
 }
 
-export class VarArgs<+T, +U = T> implements VarArgsClass<T | U> {
+export class VarArgs<+T> implements VarArgsClass<T> {
   +data: VarArgsObject<T>;
 
   constructor(data: VarArgsObject<T>) {
     this.data = data;
   }
 
-  get(name: string): T | U {
+  get(name: string): T {
     return this.data[name];
   }
 
@@ -53,7 +53,6 @@ export class VarArgs<+T, +U = T> implements VarArgsClass<T | U> {
 export type Parser<+T, -V> = (VarArgsClass<V>) => T;
 
 const EMPTY_OBJECT = Object.freeze({});
-const EMPTY_VARARGS = new VarArgs<empty, empty>(EMPTY_OBJECT);
 
 type State = {
   /*
@@ -187,7 +186,7 @@ export function parseContinuousString<V>(
   );
 }
 
-export const createTextContentParser = <+T, V>(
+export const createTextContentParser = <T, V>(
   textPattern: RegExp,
   mapValue: (string) => T,
 ): Parser<T | string | NO_MATCH, V> => () => {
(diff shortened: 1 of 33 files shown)


root/static/scripts/common/i18n/expand2.js
```diff
@@ -167,11 +167,11 @@ export function parseContinuous<T, U, V>(
 
 function concatStringMatch(
   accum: string | NO_MATCH,
-  match: string | NO_MATCH,
+  match: string,
 ): string {
   return (
     (gotMatch(accum) ? accum : '') +
-    (gotMatch(match) ? match : '')
```
reosarevok, line 174: What's the change here? (since this file is seeing a bunch of changes that are not just type improvements, it should probably be mentioned somewhere :) )

root/static/scripts/common/i18n/expand2.js
```diff
@@ -214,9 +214,6 @@ export const createVarSubstParser = <T, V>(
   },
 );
 
-export const parseStringVarSubst: Parser<string | NO_MATCH, mixed> =
```
reosarevok, line 217: I see this is moving to `expand2react` / `expand2text,` what's the reasoning? It's not possible to support both types? It's more clear? Both?

reosarevok (commented):
Checked up to the last version upgrade for now.

root/static/scripts/common/i18n/expand2.js
mwiencek: I thought it was related to the 0.217.0 upgrade but you're right, it's not. I moved it to a separate commit. :)

root/static/scripts/common/i18n/expand2.js
mwiencek: At one point it was to resolve some type errors which I also thought were cause by the 0.217.0 upgrade, but it seems I can revert this without breaking anything, so I did that.

reosarevok (approved):
Seems fine and passes tests. Are there further improvements we can do (in subsequent PRs) to our types now?

root/static/scripts/edit/utility/linkPhrase.js
```diff
@@ -96,24 +95,26 @@ class PhraseVarArgs<T>
...
 
 const textI18n: LinkPhraseI18n<string> = {
   commaList: commaListText,
+  defaultValue: '',
   displayLinkAttribute: displayLinkAttributeText,
   expand: expand2textWithVarArgsClass,
 };
 
-function _getAttributesByRootName<T>(
-  i18n: LinkPhraseI18n<T | string>,
```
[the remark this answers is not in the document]
mwiencek: I guess it was used at some point (looking at the git history) but I'm not sure which commit left it in.

PR #3060, opened 2023-10-23
Remove unneeded abbreviations

reosarevok:
# Description

As per the [design system guidelines](https://github.com/metabrainz/design-system/blob/master/guidelines/style-guidelines.md) we should not be using `i.e.` and `e.g.` since they are not as clear as "that is" and something like "such as" or "for example", respectively.

Changed all user-facing places I saw except for some API error strings using `e.g.` in case someone actually depends on the text of those for validation.
Separately changed all appearances I saw on other translatable places, such as relationship descriptions; appearances of `e.g.` on entity type descriptions left untouched for now to avoid causing issues with SIR.

# Testing
None, but these are string changes only.

# Notes
On top of https://github.com/metabrainz/musicbrainz-server/pull/3058 where `e.g.` was brought up.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,21 +41,22 @@ dependency updates, unnoticeable refactoring, and so on.
 
 #### Title
 
-Describe your change with a short imperative title, e.g.
+Describe your change with a short imperative title, such as
 
 > Change small unnoticeable bits
 
 If your change resolves a ticket (see [above](#ticket)), please make sure you
 prefix your pull request title with `MBS-XXX: ` in order for our issue tracker
-to link your pull request to that ticket, e.g.
+to link your pull request to that ticket, such as in
 
 > MBS-1234567: Change things relevant to users
 
-If it **partially resolves** a ticket, use parenthesis, e.g.
+If it **partially resolves** a ticket, use parenthesis, such as in
 
 > MBS-1234567 (I): Make first part of needed changes
 
-If your change relate to **several tickets**, separate these with commas, e.g.
+If your change relate to **several tickets**, separate these with commas,
+such as in
 
 > MBS-1234567, MBS-2345678: Change two related things at once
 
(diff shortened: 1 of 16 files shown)


CONTRIBUTING.md
```diff
@@ -41,29 +41,29 @@ dependency updates, unnoticeable refactoring, and so on.
...
 
 If your change resolves a ticket (see [above](#ticket)), please make sure you
 prefix your pull request title with `MBS-XXX: ` in order for our issue tracker
-to link your pull request to that ticket, e.g.
+to link your pull request to that ticket, such as
 
 > MBS-1234567: Change things relevant to users
 
-If it **partially resolves** a ticket, use parenthesis, e.g.
+If it **partially resolves** a ticket, use parenthesis, such as
```
mwiencek, line 54: Maybe it's just me or my dialect of English, but when I read "such as," it puts more focus on the object rather than verb, which sounds weird in some cases.

For example, instead of "use e.g. x," I hear "parentheses e.g. x," which is odd because "x" is not an example of parentheses, it's an example of a particular usage which contains parentheses, if that makes sense. :)


It sounds better to me with "in":
```suggestion
If it **partially resolves** a ticket, use parentheses, such as in
```

However, there are at least some that sound better without "in," like the one above:

```
Describe your change with a short imperative title, such as
> Change small unnoticeable bits
```

This one sounds good because in "title such as x," "x" is indeed a title.

reosarevok: That actually seems sensible to me, made a few amendments.

lib/MusicBrainz/Server/Form/Field/URL.pm
```diff
@@ -23,7 +23,7 @@ sub validate
     my $url = $self->value;
     $url = URI->new($url)->canonical;
 
-    return $self->add_error(l('Enter a valid url e.g. "http://google.com/"'))
+    return $self->add_error(l('Enter a valid URL, such as "http://google.com/"'))
```
mwiencek, line 26: one more annoying suggestion
```suggestion
    return $self->add_error(l('Enter a valid URL, such as “http://google.com/”'))
```

mwiencek (approved):

PR #3061, opened 2023-10-24
MBS-13333: Replication is blocked by tag duplicates

mwiencek:
Fixed packets will be published in conjunction with this commit.

`--ignore-conflicts` is needed for people with aborted loads, since adding broken packet numbers to `$known_broken_sequences` has the effect of truncating the currently-loaded data.

### DIFF
--- a/admin/replication/LoadReplicationChanges
+++ b/admin/replication/LoadReplicationChanges
@@ -158,6 +158,14 @@ if ($dbmirror2) {
 my %known_broken_sequences = (
     # MBS-9366
     '104949' => 1,
+    # MBS-13333
+    '162718' => 1,
+    '162719' => 1,
+    '162720' => 1,
+    '162721' => 1,
+    '162722' => 1,
+    '162723' => 1,
+    '162724' => 1,
 );
 my $aborted_sequence;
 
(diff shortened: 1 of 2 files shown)


admin/replication/LoadReplicationChanges
```diff
@@ -175,6 +183,12 @@ if (
         $sql->do('TRUNCATE dbmirror_pending CASCADE');
         $sql->auto_commit(1);
         $sql->do('TRUNCATE dbmirror_pendingdata CASCADE');
+        $sql->auto_commit(1);
+        $sql->do('TRUNCATE dbmirror2.pending_data CASCADE');
+        $sql->auto_commit(1);
+        $sql->do('TRUNCATE dbmirror2.pending_keys CASCADE');
+        $sql->auto_commit(1);
+        $sql->do('TRUNCATE dbmirror2.pending_ts CASCADE');
```
mwiencek, line 191: After these tables are cleared, the broken packet will be re-downloaded ("in case the packet was fixed", per the comment above) and applied again. Since the previously-aborted load will have applied any changes up until the first broken one, we also need a way to skip insertions that conflict with existing primary keys. The `--ignore-conflicts` flag is *only* passed if both `$aborted_sequence` and `$known_broken_sequences{$aborted_sequence}` are defined.

yvanzo (approved):
The correct target branch is `production` indeed.

Changes look good to me.

Successfully tested both:
* resuming replication stuck on half-applied broken packet
* running replication from the latest (Sat) full dump

Feel free to update your commit message (with your own comment, packet IDs…) if needed. :ship:

PR #3062, opened 2023-10-25
MBS-13336: Add a script to rebuild all indexes using collations

mwiencek:
# Problem

MBS-13336

# Solution

Provide a script should be run after a system upgrade whenever the versions of libc/icu have changed, in order to avoid index corruption.

A future improvement may be to allow specifying which collation to rebuild indexes for, but it doesn't hurt to rebuild all of them (except for taking longer).

### DIFF
--- a/admin/RebuildIndexesUsingCollations.pl
+++ b/admin/RebuildIndexesUsingCollations.pl
@@ -1,0 +1,31 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+use MusicBrainz::Server::Context;
+
+my $c = MusicBrainz::Server::Context->create_script_context(database => 'MAINTENANCE');
+my $dbh = $c->dbh;
+
+my $indexes = $c->sql->select_list_of_hashes(<<~SQL);
+    SELECT n.nspname AS schema_name, i.relname AS index_name
+      FROM pg_index AS idx
+      JOIN pg_class AS c ON idx.indrelid = c.oid
+      JOIN pg_class AS i ON idx.indexrelid = i.oid
+      JOIN pg_namespace AS n ON n.oid = c.relnamespace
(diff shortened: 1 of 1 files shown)


mwiencek:
Where would be a good place to document this? We don't really have any "maintenance" documentation at the moment.

reosarevok:
That kind of sounds like we *should* have that documentation? :)

mwiencek:
Yeah, that's why I was wondering where to put it. :) MAINTENANCE.md, and link it from INSTALL.md, I guess?

yvanzo (commented):
> whenever the versions of libc/icu have changed

For Postgres? For MusicBrainz Server? Both?

mwiencek:
>> whenever the versions of libc/icu have changed
>
> For Postgres? For MusicBrainz Server? Either? Both?

On the system -- so, for example, if glibc or libicuXX are updated in the Docker image. The collations in PostgreSQL created using these libraries will have been with an older version than is now present on the system.

This doesn't particularly matter for MusicBrainz Server; some CPAN modules need to be rebuilt, like `Unicode::ICU::Collator`, but in production that's part of the Docker image. PostgreSQL collations, on the other hand, are part of the database.

yvanzo:
> On the system

Which one(s)? I’m asking because Postgres and MusicBrainz Server are not running in the same container, so not on the same system.

mwiencek:
> Which one(s)? I’m asking because Postgres and MusicBrainz Server are not running in the same container, so not on the same system. That’s the case in production. That’s also the case in MusicBrainz Docker (Compose) mirrors.

For the issue I'm trying to resolve here, the versions of glibc/libicuXX are only relevant on the system/container running PostgreSQL.

yvanzo (approved):
Thank you! I guess that it has to be run on mirrors too, thus resolving MBVM-90.

Tested running:
* locally with latest sample data in 20s.
* on wolf with up-to-date full mirror in 48m.
* on a dedicated VM with up-to-date database-only mirror in 3m.

mwiencek:
I have further improvements in the works (documentation and script options), but I'll merge this and submit them as a separate PR.

PR #3063, opened 2023-10-26
MBS-13334: Also find emails with capitalization differences in find_by_email

reosarevok:
### Fix MBS-13334

# Problem
We just had a support case where the user already had an account, but `lost_username` claimed the email was not in use. When the user, confused, tried to create a new account, he got an email (using `is_email_used_elsewhere`) saying that the email is already in use. It seems the issue was due to capitalization differences (the email was stored in the DB with some capitalized letters)

# Solution
We had a better query in `is_email_used_elsewhere` so this just reuses effectively the same query for `find_by_email`.

Using `_get_by_keys` ran `load_preferences`, which this does not. Since we don't seem to use the preferences for our use of `find_by_email` I did not bother loading them with the new method.

# Testing
Added tests for these two email checks, and I improved and documented the `Data::Editor` test while at it so it wouldn't be a huge mess.

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Data/Editor.pm
+++ b/t/lib/t/MusicBrainz/Server/Data/Editor.pm
@@ -277,46 +277,56 @@ $editor_data->load_preferences($alice);
 is($alice->preferences->public_ratings, 0, 'load preferences');
 is($alice->preferences->datetime_format, '%m/%d/%Y %H:%M:%S', 'datetime_format loaded');
 is($alice->preferences->timezone, 'UTC', 'timezone loaded');
-
-my @editors = $editor_data->find_by_subscribed_editor(2, 10, 0);
-is($editors[1], 1, 'alice is subscribed to one person ...');
-is($editors[0][0]->id, 1, '          ... that person is new_editor');
-
-
-@editors = $editor_data->find_subscribers(1, 10, 0);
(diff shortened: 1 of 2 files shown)


t/lib/t/MusicBrainz/Server/Data/Editor.pm
```diff
@@ -174,11 +178,39 @@ test 'find_by_email and is_email_used_elsewhere' => sub {
...
+
+    @editors = $editor_data->find_by_email('editor@example.com');
+    is(scalar(@editors), 1, 'An editor was found searching with normal caps');
+    is($editors[0]->id, $new_editor_2->id, 'The right editor was found');
+
+    note('We check is_email_used_elsewhere shows the email as being in use');
+    ok(
+        $editor_data->is_email_used_elsewhere(
+            'EDITOR@EXAMPLE.COM',
+            $new_editor_2->id + 1,
```
mwiencek, line 204: Not sure where `$new_editor_2->id + 1` is coming from. Or is that supposed to just represent something like `$nonexistent_id`?

reosarevok: Yeah, it's just meant to be "someone else" since `is_email_used_elsewhere` needs an editor ID that wants to use the email. Maybe I just should make it a `$someone_else_id`.

mwiencek (approved):

PR #3065, opened 2023-10-31
MBS-13261: Upgrade the required version of Node.js to 20

mwiencek:
# Problem

See MBS-13261 and https://blog.metabrainz.org/2023/09/04/musicbrainz-website-mirror-will-require-node-js-20-starting-from-november-13-2023/.

# Solution

 * Upgrades Node.js to v20.9.0
 * Upgrades our Docker base images to jammy
 * Upgrades our version of Perl in production to 5.34 (as a consequence of the jammy upgrade)
 * Runs our CircleCI tests against both 5.30 and 5.34 (replacing the Node v16 tests)


reosarevok:

reosarevok:
Seems the chrome-for-testing change is not good enough? There's still selenium complaints about a mismatch.

mwiencek:
Yeah, I think I have to modify selenium.mjs to allow configuring where the chrome binary is located (seems a symbolic link did not work). Once that's in place, there shouldn't be a version mismatch.

docker/Dockerfile.perl5.30.tests
```diff
@@ -0,0 +1,295 @@
...
+        ca-certificates \
+        curl \
+        gnupg && \
+    cat nodesource_pubkey.txt | gpg --dearmor -o /etc/apt/trusted.gpg.d/nodesource.gpg && \
+    cat pgdg_pubkey.txt | gpg --dearmor -o /etc/apt/trusted.gpg.d/pgdg.gpg && \
+    cat yarn_pubkey.txt | gpg --dearmor -o /etc/apt/trusted.gpg.d/yarn.gpg && \
+    rm nodesource_pubkey.txt pgdg_pubkey.txt yarn_pubkey.txt && \
+    echo "deb [signed-by=/etc/apt/trusted.gpg.d/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list && \
+    echo "deb [signed-by=/etc/apt/trusted.gpg.d/pgdg.gpg] http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list && \
+    echo "deb [signed-by=/etc/apt/trusted.gpg.d/yarn.gpg] https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
```
atj, line 26:
```suggestion
    cp nodesource_pubkey.txt /etc/apt/keyrings/nodesource.asc && \
    cp pgdg_pubkey.txt /etc/apt/keyrings/pgdg.asc && \
    cp yarn_pubkey.txt /etc/apt/keyrings/yarn.asc && \
    rm nodesource_pubkey.txt pgdg_pubkey.txt yarn_pubkey.txt && \
    echo "deb [signed-by=/etc/apt/keyrings/nodesource.asc] https://deb.nodesource.com/node_20.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list && \
    echo "deb [signed-by=/etc/apt/keyrings/pgdg.asc] http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list && \
    echo "deb [signed-by=/etc/apt/keyrings/yarn.asc] https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
```

mwiencek:
@brainzbot test this please

mwiencek:
I was finally able to build production Docker images from this branch and deploy them on the test server: https://test.musicbrainz.org/.

mwiencek:
There was a very strange issue where `yarn` would not install all packages inside the production container, even after clearing both the Yarn cache and node_modules/. I ended up just upgrading Yarn to v4, which worked fine. But, like upgrading Node.js, this may also require manual intervention from the user.

Yarn is no longer installed using the system package manager, but instead via `corepack`:

```sh
# from the musicbrainz-server checkout
$ sudo corepack enable
```

This reads `packageManager` from package.json, and installs the specified version of Yarn from (I believe) their servers. It targets /usr/local/bin by default, hence why `sudo` may be needed. You can also specify the installation directory via `--install-directory`. If Yarn v1 is still required on the system, that's probably necessary.

INSTALL.md
```diff
@@ -92,9 +92,10 @@ Prerequisites
     This is only needed where it exists, so a warning about the package not being
     found is not a problem.
 
-    Next you need Yarn to install the JS dependencies. There are a variety of
-    installation methods described on their website, located here:
-    https://yarnpkg.com/en/docs/install
+    Next you need Yarn to install the JS dependencies. This can be obtained via
+    corepack, which is bundled with Node.js:
+
+        sudo corepack enable
```
reosarevok, line 98: "You can also specify the installation directory via `--install-directory`. If Yarn v1 is still required on the system, that's probably necessary" from the PR comments - should this be documented here then? How common is yarn v1 nowadays?

mwiencek: > How common is yarn v1 nowadays?

I have no idea, but v1 is the only version available on npm or apt, so I imagine quite common still

reosarevok (approved):
AFAICT this works now (also in test).

PR #3066, opened 2023-11-01
Use "remove" consistently for non-editor cases

reosarevok:
# Description
We generally use "remove" for things like entities, and "delete" (which is almost the same but has a slightly stronger vibe) for editors only. This changes a few translatable "delete" strings that were actually meant for non-editors, so should probably use "remove".

# Testing
None, but this changes strings only.

### DIFF
--- a/root/relationship/linktype/RelationshipTypeInUse.js
+++ b/root/relationship/linktype/RelationshipTypeInUse.js
@@ -21,7 +21,7 @@ const RelationshipTypeInUse = ({
       <h1>{l('Relationship Type In Use')}</h1>
       <p>
         {texp.l(
-          `The relationship type "{type}" can’t be removed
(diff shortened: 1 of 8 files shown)


root/relationship/linktype/RelationshipTypeInUse.js
```diff
@@ -21,7 +21,7 @@ const RelationshipTypeInUse = ({
       <h1>{l('Relationship Type In Use')}</h1>
       <p>
         {texp.l(
-          `The relationship type "{type}" can’t be deleted
+          `The relationship type "{type}" can’t be removed
```
mwiencek, line 24: Might as well, while the string is being changed:
```suggestion
          `The relationship type “{type}” can’t be removed
```

mwiencek (approved):

PR #3067, opened 2023-11-01
Replace xgettext-js with forked copy using hermes-parser

mwiencek:
# Problem

xgettext-js uses @babel/parser, which doesn't support new Flow syntax, and I'm not sure it (Babel) will be updated any time soon.

# Solution

Replace xgettext-js with a forked copy that uses hermes-parser instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "tap-junit": "3.1.0",
     "tape": "4.7.0",
     "utf8": "2.1.2",
-    "xgettext-js": "3.0.0"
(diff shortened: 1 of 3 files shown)


reosarevok (approved):
Tested with new string changes and it seems to generate the files correctly.

yvanzo (approved):
Successfully tested locally with both code and standby database. :ship:

Side-note: We are still having this issue where the lines are not wrapped the same depending on who is generating POT files, but it’s not new to this patch.

PR #3068, opened 2023-11-02
MBS-4822: Drop duplicate colon strings 

reosarevok:
### Implement MBS-4822

# Description
This looks at what seem to be all or at least most cases where we have two separate strings, `X` and `X:`, when we could use `addColon`, and uses it (generally `add_colon` or `addColonText` for Perl and JS respectively). When I saw cases where the strings had different meanings I added context using `lp` - I also did that to the existing strings without colons when relevant. There's of course a lot more places where context would be very useful, this is not meant to be an exhaustive solution for MBS-5380.

# Testing
None, but effectively string changes only.


reosarevok:

mwiencek (approved):

reosarevok:

PR #3069, opened 2023-11-02
MBS-13207 (II): Move "Add a new recording" above suggestions

derat:
# MBS-13207 (II)

Make "Add a new recording" appear above suggested recordings rather than below them in the release editor. This is intended to make it possible to select recordings using the keyboard (which is currently challenging due to W3C-mandated radio button focus behavior).

# Problem

When adding a release, it's currently very difficult to assign existing recordings to tracks using only the keyboard.

After opening the recording-selection bubble, the focus is initially assigned to the search input. The Tab key cannot be used to advance the focus to an existing recording's radio button. Instead, the user must advance the focus through all of the recording, artist, and release group links before focusing the "Add a new recording" radio button (currently at the bottom of the table), at which point they can use the Up arrow key to select the desired recording.

This is due to W3C-mandated behavior: per https://www.w3.org/wiki/RadioButton, the Tab key can only be used to focus the currently-selected radio button.

# Solution

This change moves "Add a new recording" to the top of the table. This radio button is selected by default if a recording has not already been assigned, so the user can just press Tab once to focus it and then use the arrow keys to select the desired recording.

Please see #3006 (a different approach) for earlier discussion.

# Actions

I'm not running a search server and just have test data loaded, so I've only been able to test this by using direct search to manually add recordings to the bubble. If possible, please manually test this using a more-realistic setup.

### DIFF
--- a/root/release/edit/recordings.tt
+++ b/root/release/edit/recordings.tt
@@ -169,9 +169,20 @@
 
       <table data-bind="with: target">
         <tbody>
+          <tr>
+            <td>
+              <input name="recording-selection" value="" type="radio" id="add-new-recording" data-bind="checked: hasNewRecording, checkedValue: true" />
+            </td>
+            <td colspan="3">
+              <label for="add-new-recording">
+                [% l('Add a new recording') %]
+              </label>
+            </td>
+          </tr>
+
           <!-- ko if: loadingSuggestedRecordings -->
           <tr>
-            <td colspan="4" style="padding-bottom: 1em; vertical-align: top;" class="loading-message">
+            <td colspan="4" style="background-position: left 1em; padding-bottom: 1em; padding-top: 1em; vertical-align: top;" class="loading-message">
               [% l('Looking for suggested recordings...') %]
             </td>
           </tr>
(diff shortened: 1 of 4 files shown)


root/release/edit/recordings.tt
```diff
@@ -169,9 +169,20 @@
...
+              <label for="add-new-recording">
+                [% l('Add a new recording') %]
+              </label>
+            </td>
+          </tr>
+
           <!-- ko if: loadingSuggestedRecordings -->
           <tr>
-            <td colspan="4" style="padding-bottom: 1em; vertical-align: top;" class="loading-message">
+            <td colspan="4" style="background-position: left 1em; padding-bottom: 1em; padding-top: 1em; vertical-align: top;" class="loading-message">
```
derat, line 185: I'm not super-happy with this, but I didn't see a better way to handle the loading message being smooshed up against the "Add a new recording" row.

I don't want to set `padding-bottom` on "Add a new recording" since the extra spacing remains after this message disappears. When I set `padding-top` here, I also need to update `background-position` so that the spinner added by the `loading-message` class is aligned with the text.

Other possibilities might be changing `display` so I can set `margin-top`, or adding an empty row within the `loadingSuggestedRecordings` condition just to add spacing. Let me know if either of those sound better.

derat:
@mwiencek, @reosarevok, and @yvanzo, you all had thoughts about #3006. Mind also taking a look at this approach? Thanks!

mwiencek (approved):
Tested locally and it feels like an improvement. I briefly looked at the loading spinner CSS but didn't find any nicer solution (except perhaps moving it to release-editor.less to limit the amount of inline CSS).

derat:
Thanks! I considered putting the CSS in a `.less` file, but I was worried that it might be even more confusing to have some of the padding/margin rules there and others in `recordings.tt`. I'm happy to try moving all of `#recording-assoc-bubble`'s inline styles into `release-editor.less` if it doesn't seem too out-of-scope for this change, though.

mwiencek:
Yeah, let's leave it alone for now. :) Thanks for following up on this!

PR #3070, opened 2023-11-02
Use "create", "add", "enter" in more specific ways

reosarevok:
# Description
We seem to use "create" for many things we have other words for. This uses "create" just for user accounts. Edits are "entered", which is the terminology we already use most of the time. Other things are "added" (especially things like entities).

# Testing
None, but it's just strings being changed


root/user/UserCollections.js
```diff
@@ -223,7 +223,7 @@ const UserCollections = ({
         </p>
       )}
       {viewingOwnProfile ? (
-        <p><a href="/collection/create">{l('Create a new collection')}</a></p>
+        <p><a href="/collection/create">{l('Ad a new collection')}</a></p>
```
kellnerd, line 226:
```suggestion
        <p><a href="/collection/create">{l('Add a new collection')}</a></p>
```

mwiencek (approved):

PR #3072, opened 2023-11-04
MBS-5987: Add "Submit votes & edit notes" to top

derat:
# MBS-5987

Add a second "Submit votes & edit notes" button near the top of the /edits page so that users don't need to scroll to the end of a long page when voting and commenting on recent edits.

# Problem

When viewing an entity's list of edits, the open edits are at the top of the page. It's unintuitive and a bit tedious to need to scroll all the way to the bottom of the page to submit votes and comments on open edits.

# Solution

Add a second button near the top of the page. I had to move the `form` element outside of `PaginatedResults` in order for the button to be a child of the form.

# Testing

I verified that both the top and bottom buttons work when viewing edits while logged in, and that the page renders as before when not logged in.

# Action

N/A

### DIFF
--- a/root/edit/components/EditList.js
+++ b/root/edit/components/EditList.js
@@ -98,8 +98,13 @@ const EditList = ({
 
       {edits.length ? (
         <div id="edits">
-          <PaginatedResults guessSearch={mightBeMoreEdits} pager={pager}>
-            <form action="/edit/enter_votes" method="post">
+          <form action="/edit/enter_votes" method="post">
+            <PaginatedResults guessSearch={mightBeMoreEdits} pager={pager}>
+              {$c.user ? (
+                <div className="align-right clear-both row no-label">
+                  <FormSubmit label={l('Submit votes & edit notes')} />
+                </div>
+              ) : null}
               {edits.map((edit, index) => (
                 <ListEdit
                   edit={edit}
@@ -116,8 +121,8 @@ const EditList = ({
                   <FormSubmit label={l('Submit votes & edit notes')} />
                 </div>
               ) : null}
-            </form>
-          </PaginatedResults>
+            </PaginatedResults>
+          </form>
         </div>
       ) : null}
 
--- a/root/static/styles/layout.less
+++ b/root/static/styles/layout.less
@@ -147,6 +147,8 @@ pre code {
 .align-left { text-align: left !important; }
 .align-top { vertical-align: top; }
 
+.clear-both { clear: both !important; }
+
 /*
 ** Header
 */


root/static/styles/layout.less
```diff
@@ -147,6 +147,8 @@ pre code {
 .align-left { text-align: left !important; }
 .align-top { vertical-align: top; }
 
+.float-right { float: right; }
```
derat, line 150: Please let me know if it'd be better to add a dedicated rule in `root/static/styles/edit.less`. I didn't see any obvious place for it there, so I wasn't sure if adding a selector like `#edits .top-submit` (and setting the appropriate class above) would be too gross.

root/edit/components/EditList.js
```diff
@@ -98,8 +98,14 @@ const EditList = ({
 
       {edits.length ? (
         <div id="edits">
-          <PaginatedResults guessSearch={mightBeMoreEdits} pager={pager}>
-            <form action="/edit/enter_votes" method="post">
+          <form action="/edit/enter_votes" method="post">
```
derat, line 101: Just as a heads-up, this might cause problems if there are plans for `PaginatedResults` to include another form within the paginator for e.g. jumping directly to a typed-in page. As far as I know, HTML5 doesn't allow nesting `form` elements (see https://html.spec.whatwg.org/multipage/forms.html#the-form-element).

derat:

root/edit/components/EditList.js
```diff
@@ -98,8 +98,14 @@ const EditList = ({
...
+            {$c.user ? (
+              <FormSubmit
+                className="float-right"
```
yvanzo, line 104: It aligns the button to the right but it is missing vertical margins. Would it be better to use the same `<div className="align-right row no-label">` as for the submission button at the bottom?

root/edit/components/EditList.js
yvanzo: For now it doesn’t seem to be causing any problem.

yvanzo (commented):
Thank you for this contribution, straightforward to review.

root/static/styles/layout.less
yvanzo: See https://github.com/metabrainz/musicbrainz-server/pull/3072#discussion_r1382393723

root/edit/components/EditList.js
derat: Maybe I'm misunderstanding, but I believe that that results in the button being in its own row, which wastes a bit of vertical space:

image

Or I guess it could go under the paginator:

image

Let me know if you prefer either of those.

yvanzo: > that results in the button being in its own row, which wastes a bit of vertical space

Yes, I didn’t notice this issue (waste of space) but it is how it is currently displayed, right?

The margin issue I mentioned is occurring on non-paginated lists of edits.

yvanzo: To clarify, I don’t think that having separate rows is a major issue. But if you intend to address it too (on top of MBS-5987), please make it a separate commit and changes it for both buttons (at the top and at the bottom).

derat: Got it; thanks for the explanation. I've amended the commit to place the button on its own row below the paginator.

root/edit/components/EditList.js
kellnerd: You could use the button's [form attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form) instead to assign it to a form which is not the button's ancestor.

derat: True, thanks. I think that that'd require a little bit of plumbing since the attribute doesn't seem to be exposed by the `FormSubmit` component.

yvanzo (approved):

derat:
Thanks for reviewing! Can this be merged now?

yvanzo:
> Let me try to promote myself to be an auto-editor in my dev setup

```sh
sudo docker-compose exec db psql -d musicbrainz_db -U musicbrainz -c "UPDATE editor SET privs = privs | 1 WHERE name = 'derat'"
```

derat:
Thanks! Is there anything more that I need to do to get the "Vote on all edits" radio buttons to show up? I'm seeing four open edits at `/artist/65f4f0c5-ef9e-490c-aee3-909e7ae6b2ab/open_edits` along with "Add Note", "Approve edit", etc. auto-editor buttons, but no vote-on-all radio buttons (even when using the `master` branch). I can see that `MB.Control.EditList` is getting called, so I'm a bit confused. I'm not very familiar with jQuery, though.

derat:
Summarizing IRC discussion: I wasn't getting the vote-all radio buttons due to `editorMayVoteOnEdit` in root/edit/components/Vote.js seeing that these were my own edits. The overall-vote div is using `float: right` to appear to the right of the paginator, so I added `clear: both` to the button's wrapper.

root/static/styles/layout.less
```diff
@@ -147,6 +147,8 @@ pre code {
 .align-left { text-align: left !important; }
 .align-top { vertical-align: top; }
 
+.clear-both { clear: both !important; }
```
derat, line 150: I added `!important` here to override `float: left` from the more-specific `form div.row` selector in root/static/styles/forms.less. I assume that anyone using this class actually wants it to take precedence, and there's precedent in the `align-left` class a few lines above this one. :-/

mwiencek (approved):

PR #3074, opened 2023-11-05
MBS-13347: Relationship Type edit search times out

mwiencek:
# Problem

MBS-13347

The use of `LIMIT 500` inside an edit search sub-query using the Relationship Type predicate causes PostgreSQL to ignore `edit_data_idx_link_type` and make use of a much less efficient query plan.

# Solution

This commit adds a flag to `EditSearch::Predicate` which can be used to disable the addition of `$LIMIT_FOR_EDIT_LISTING` if the given predicate is present.

### DIFF
--- a/lib/MusicBrainz/Server/EditSearch/Predicate.pm
+++ b/lib/MusicBrainz/Server/EditSearch/Predicate.pm
@@ -98,4 +98,6 @@ sub valid {
     return 1;
 }
 
+sub disable_limit_for_edit_listing { 0 }
+
 1;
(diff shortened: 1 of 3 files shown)


mwiencek:
Relevant search with which you can test against the production standby: http://localhost:5000/search/edits?auto_edit_filter=&order=desc&negation=0&combinator=and&conditions.0.field=editor&conditions.0.operator=!%3D&conditions.0.name=CatCat&conditions.0.args.0=38393&conditions.5.field=link_type&conditions.5.operator=%3D&conditions.5.args=896

yvanzo (approved):
In the commit message, it should be explicitly mentioned that this flag is specifically set for edit searches involving a condition on the type of the relationship being edited.

Looks good otherwise.

Did test locally with production standby database.

:ship:

PR #3077, opened 2023-11-06
MBS-13349: Support LibraryThing disambiguation URLs

reosarevok:
### Fix MBS-13349

# Problem
LibraryThing uses -1, -2, etc. after the author name for authors of the same name (such as https://www.librarything.com/author/alexandermichelle-1). We currently (accidentally) block these.

# Solution
It seems sensible and safe to allow hyphens alongside the already allowed numbers and letters, so this just does that.

# Testing
Added a test with the URL given above.


yvanzo (approved):
LGTMBDNT, added test covers it anyway. :ship:

PR #3078, opened 2023-11-06
MBS-13346: Support GTIN-14 barcodes

reosarevok:
### Implement MBS-13346

# Problem
GTIN-14 identifiers ("barcodes") seem to be becoming more common in digital storefronts and the like. Since these are the identifiers sent by the distributors, we should keep them, even they are generally just zero-padded EANs. Right now the release editor considers these invalid and asks the user to check the box though.

# Solution

`is_valid_ean` on the Perl side already accepts GTIN-14 so nothing to change there - this mostly just changes the release editor validation and adds a basic test for it.

I added a `GTIN` type to `Entity::Barcode` since I guess it cannot hurt. Not sure if we use these at all but :)

# Testing
Manually in the release editor, plus added a test.

### DIFF
--- a/lib/MusicBrainz/Server/Entity/Barcode.pm
+++ b/lib/MusicBrainz/Server/Entity/Barcode.pm
@@ -14,6 +14,7 @@ sub type {
     return 'EAN' if length($self->code) == 8;
     return 'UPC' if length($self->code) == 12;
     return 'EAN' if length($self->code) == 13;
+    return 'GTIN' if length($self->code) == 14;
 }
 
 sub format
--- a/root/static/scripts/tests/release-editor/validation.js
+++ b/root/static/scripts/tests/release-editor/validation.js
@@ -143,3 +143,19 @@ validationTest((
 
   t.ok(validation.errorsExist());
 });
+
+validationTest((
+  'Barcode validation'
+), function (t) {
+  t.plan(5);
+
+  const release = releaseEditor.rootField.release();
+  const field = release.barcode;
+
+  t.ok(field.validateCheckDigit('00810121774182'), '0-padded GTIN-14');
+  t.ok(field.validateCheckDigit('12345678901231'), 'GTIN-14');
+  // Adding '0' + as it matches what the actual release editor code does
+  t.ok(field.validateCheckDigit('0' + '0810121774182'), '0-padded EAN-13');
+  t.ok(field.validateCheckDigit('0' + '9399431762528'), 'EAN-13');
+  t.ok(field.validateCheckDigit('00' + '810121774182'), 'UPC');
+});
--- a/root/static/scripts/release-editor/fields.js
+++ b/root/static/scripts/release-editor/fields.js
@@ -932,12 +932,12 @@ class Barcode {
   }
 
   checkDigit(barcode) {
-    if (barcode.length !== 12) {
+    if (barcode.length !== 13) {
       return false;
     }
 
     let calc = 0;
-    for (let i = 0; i < 12; i++) {
+    for (let i = 0; i < 13; i++) {
       calc += parseInt(barcode[i], 10) * this.weights[i];
     }
 
@@ -946,8 +946,8 @@ class Barcode {
   }
 
   validateCheckDigit(barcode) {
-    return this.checkDigit(barcode.slice(0, 12)) ===
-            parseInt(barcode[12], 10);
+    return this.checkDigit(barcode.slice(0, 13)) ===
+            parseInt(barcode[13], 10);
   }
 
   writeBarcode(barcode) {
@@ -956,7 +956,7 @@ class Barcode {
   }
 }
 
-Barcode.prototype.weights = [1, 3, 1, 3, 1, 3, 1, 3, 1, 3, 1, 3];
+Barcode.prototype.weights = [3, 1, 3, 1, 3, 1, 3, 1, 3, 1, 3, 1, 3];
 
 fields.Barcode = Barcode;
 
--- a/root/static/scripts/release-editor/validation.js
+++ b/root/static/scripts/release-editor/validation.js
@@ -190,7 +190,7 @@ utils.withRelease(function (release) {
       ),
     );
   } else if (barcode.length === 12) {
-    if (field.validateCheckDigit('0' + barcode)) {
+    if (field.validateCheckDigit('00' + barcode)) {
       field.message(l('The barcode you entered is a valid UPC code.'));
       searchExistingBarcode(field, barcode, release.gid());
     } else {
@@ -209,7 +209,7 @@ utils.withRelease(function (release) {
       );
     }
   } else if (barcode.length === 13) {
-    if (field.validateCheckDigit(barcode)) {
+    if (field.validateCheckDigit('0' + barcode)) {
       field.message(l('The barcode you entered is a valid EAN code.'));
       searchExistingBarcode(field, barcode, release.gid());
     } else {
@@ -219,9 +219,20 @@ utils.withRelease(function (release) {
         doubleCheckText,
       );
     }
+  } else if (barcode.length === 14) {
+    if (field.validateCheckDigit(barcode)) {
+      field.message(l('The barcode you entered is a valid GTIN code.'));
+      searchExistingBarcode(field, barcode, release.gid());
+    } else {
+      field.error(
+        l('The barcode you entered is not a valid GTIN code.') +
+        ' ' +
+        doubleCheckText,
+      );
+    }
   } else {
     field.error(
-      l('The barcode you entered is not a valid UPC or EAN code.') +
+      l('The barcode you entered is not a valid UPC, EAN or GTIN code.') +
       ' ' +
       doubleCheckText,
     );


root/static/scripts/tests/release-editor/validation.js
```diff
@@ -143,3 +143,17 @@ validationTest((
...
+validationTest((
+  'Barcode validation'
+), function (t) {
+  t.plan(4);
+
+  const release = releaseEditor.rootField.release();
+  const field = release.barcode;
+
+  t.ok(field.validateCheckDigit('00810121774182'), '0-padded GTIN-14');
+  t.ok(field.validateCheckDigit('0810121774182'), '0-padded EAN-13');
```
yvanzo, line 156: Do you have any actual GTIN-14 example to test the added weight used in this computation?

yvanzo (approved):
Actual test is missing, LGTMBDNT otherwise :ship:

lib/MusicBrainz/Server/Entity/Barcode.pm
```diff
@@ -14,6 +14,7 @@ sub type {
     return 'EAN' if length($self->code) == 8;
     return 'UPC' if length($self->code) == 12;
     return 'EAN' if length($self->code) == 13;
+    return 'GTIN' if length($self->code) == 14;
```
mwiencek, line 17: I guess the idea is that a GTIN can be less than 14 digits, but if it's 14 digits then it's definitely a GTIN?

reosarevok: Pretty much.

root/static/scripts/tests/release-editor/validation.js
reosarevok: I didn't, but we do have one in the `is_valid_ean` test which might be made up but I might as well reuse.

yvanzo: I couldn’t find one either, but [Wikipedia article](https://en.wikipedia.org/wiki/Global_Trade_Item_Number) says that it is _used for wholesale shipments, not retail point of sale_. That might be the reason why we can’t find actual examples. Would it be more relevant to just stripe the leading zeroes if there are too many characters?

reosarevok: There's a lot of discussion in https://musicbrainz.org/edit/105112178 which led to this - my basic train of thought is that we should store the code without stripping the zeros if that's how the distributors are sending it, because we do not strip the zeros of EANs that are just 0 + UPC, and it seems intentional that they do it this way.

yvanzo: > There's a lot of discussion in https://musicbrainz.org/edit/105112178

In the case of this edit, it isn’t clear what is the source for this 14th zero because the initial edit [#105086824](https://musicbrainz.org/edit/105086824) is referring to Atisket and Deezer which both provides a GTIN-12 only. Has this data been picked up automatically by Atisket from Spotify which is providing a GTIN-14 for this release indeed (see [cached json](https://atisket.pulsewidth.org.uk/cached/spotify/albums_3J353DwtMfUV3V6t9vRPCk.json))?

> it seems intentional that they do it this way

It seems to just be a bug from Spotify API as GTIN-14 are “not [for] retail point of sale”.

reosarevok: Look at the whole discussion - most storefronts seem to keep it and only Deezer drops them. For what it's worth, our JSON-LD also uses "gtin-14" although I guess maybe we should change that to https://schema.org/gtin instead

Basically, since we do not drop a leading zero for an EAN even though it's just padding, I do not see a reason we would do it for this either.

yvanzo: > Look at the whole discussion

I did, and many notes started to question the relevance of the 14th digit for individual releases after the ticket has been opened. Actually the edit vote even ended in favor of removing the two leading zeroes from the barcode of this release, including the author of Atisket which has been used for submitting the initially zero-padded barcode.

Look also at my whole comment, including the links that explains GTIN-14. It is meant for logistics, not for point of sale. There will never be a GTIN-14 for an individual release starting with anything else than 0 by definition.

> most storefronts seem to keep it and only Deezer drops them

That’s an interpretation, we could rather say that Deezer API returns the most human-readable and still appropriate format for a point of sale whereas Apple Music and Spotify APIs are adding zeroes to pad to a unique-length but inappropriate format for a point of sale. It doesn’t seem to be known so far for other platforms.
* I’m not exactly sure why these two APIs do that, possibly to communicate with some legacy applications that are more about logistics. Also APIs are backends, not _storefronts_, so not exactly meant for end-users anyway.
* Also I’m not exactly sure why Atisket submits the padding zeroes to the MB release editing form while it doesn’t show those on its own page ([see above](https://github.com/metabrainz/musicbrainz-server/pull/3078#discussion_r1383916116)). The input of @atj on this point would be interesting.

> our JSON-LD also uses "gtin-14" although I guess maybe we should change that to https://schema.org/gtin instead

Good catch! The field `gtin14` has been added in 2014 through the commit https://github.com/metabrainz/musicbrainz-server/commit/cff781fde2293458e785522919d1a1558de165ab which unfortunately doesn’t explain the reason behind this choice. Also it is misleading (or even bugged?) as it actually contains the raw barcode instead of a zero-padded 14-digits code. We will have to improve (or fix?) that in a way or another once a decision has been made.

> Basically, since we do not drop a leading zero for an EAN even though it's just padding, I do not see a reason we would do it for this either.

The reason would be that any GTIN-14 exposed in some API is not meant for point of sale/end-consumer, and it isn’t exposed in any webpage or printed on any physical release either. On the contrary EAN-13 is meant for point of sale, and it is printed on many physical releases. EAN-13 isn’t necessarily starting with a zero either.

Side-note: Our release search by barcode ignores leading zeroes anyway, so it won’t be an issue in either case.

reosarevok: Honestly, I just don't see the benefit of telling people they cannot enter what is there when it is there. We could automatically drop the first zero, although that would be kind of off if the situation suggested in https://musicbrainz.org/edit/105112178#note-105112178-32 is a correct understanding. It just seems that changing the input without asking the user can be more confusing that it's worth for very little benefit, especially since as you said, a search will find the release either way.

atj: > Also I’m not exactly sure why Atisket submits the padding zeroes to the MB release editing form while it doesn’t show those on its own page ([see above](https://github.com/metabrainz/musicbrainz-server/pull/3078#discussion_r1383916116)). The input of @atj on this point would be interesting.

Yes this looks like a bug. AFAICT the original submitter looked up the release using a Deezer URL, which means it is used as the primary source. a-tisket will zero pad UPCs up to 14 characters when searching for releases due to the discrepancies between Deezer, Spotify and iTunes, and this is how the Spotify release was found:

* https://open.spotify.com/search/upc:810121774182 - no result
* https://open.spotify.com/search/upc:0810121774182 - no result
* https://open.spotify.com/search/upc:00810121774182 - 1 result

However the UPC submitted to the release editor  is zero padded to 14 characters:

https://atisket.pulsewidth.org.uk/?preferred_countries=GB%2CUS%2CDE&deez_id=461360545

The UPC should be set to the value returned by the primary vendor so this behaviour is incorrect.

FWIW this PR seems correct, in that MB should recognise and validate GTIN-14s. I'm not sure discussions around whether the zero padded form should be considered canonical or not are relevant to the PR but it would probably be worth clarifying in the style guidelines.

yvanzo: Some discoveries:

1. According to the official [GS1 standard](https://ref.gs1.org/standards/genspecs/):
   1. GTIN-14s are 14-digit GTINs that don’t start with 0,
      (Wikipedia article on GTIN can be misleading about this)
      and are _not permitted on retail consumer trade items_.
   2. _The presence or lack of these leading zeroes_
      (those between 8, 12, 13, and 14 digits)
      _does not change the GTIN concerned._
      _GTINs may be stored with or without_ (these) _leading zeroes in the same database field,_
      _depending on the requirements of the particular application._
      (However one shouldn’t remove just two leading zeroes of a GTIN-12 for example.)

2. Most online music platforms don’t even refer to GTIN or 14-digit codes:
   
   * [7digital](https://docs.7digital.com/reference/release-details): The `barcode` example is an EAN-13
   * [Bandcamp](https://get.bandcamp.help/hc/en-us/articles/360007802754-Does-Bandcamp-supply-UPCs-and-ISRCs): “UPC”
   * [CDBaby](http://cdbaby.com/cd-baby-cost/barcode-for-music/): “UPC”
   * [deezer-python](https://deezer-python.readthedocs.io/en/stable/api_reference/resources/album.html#deezer.Album.upc): `upc`
   * [Jamendo](https://help-artists.jamendo.com/hc/en-us/articles/360006659178-ISRC-and-UPC-EAN-codes-on-Jamendo): “UPC/EAN”
   * [Jaxsta](https://support.jaxsta.com/portal/en/kb/articles/upc-or-ean): “UPC or EAN”
   * [SoundCloud](https://developers.soundcloud.com/docs/api/explorer/open-api): _Schema > Playlist_ has an `ean` field
   * [TuneCore](https://support.tunecore.com/hc/en-us/articles/115006499567): “EAN/UPC”

   Since I don’t have keys for these APIs, I used Atisket to search its 3 only supported platforms, but it might be erroneous:

   * Deezer API seems to accept any length and to return what they have (12-digit and 13-digit codes)
   * iTunes API seems to accept any length and to return what the user requested
   * Spotify API only accepts 14-digit codes and only returns 14-digit codes

   Spotify seems to be the exception here.

The two above put together, the GS1 standard allows:
* all of the above platforms to keep using EAN and/or UPC
* Deezer to return 12-digit and 13-digit codes
* iTunes to return whatever the user requested
* Spotify to return 14-digit zero-padded codes

However, by denying 12-digit and 13-digit codes for search, Spotify doesn’t conform to the GS1 standard.

It was needed to validate 14-digit zero-padded GTINs, but it isn’t appropriate to validate GTIN-14s.

I’ve opened the pull request https://github.com/metabrainz/musicbrainz-server/pull/3155 to address some of these and related issues.

lib/MusicBrainz/Server/Entity/Barcode.pm
yvanzo: Fun fact: GTIN was previously named [EAN-14](http://www.transbar.fr/choisir/choisir_ean14.html?lg=en).

This subroutine is dead code anyway. To be removed in PR https://github.com/metabrainz/musicbrainz-server/pull/3155 / commit https://github.com/metabrainz/musicbrainz-server/pull/3155/commits/25e436078251648727d4b111596684d5a853cf34.

PR #3079, opened 2023-11-07
Allow connecting to any database with SYSTEM credentials

mwiencek:
# Problem

Sometimes it's necessary to connect to the database specified by `MAINTENANCE` or `READWRITE` using `SYSTEM` credentials (because certain objects owned by a superuser need to be updated).

# Solution

This is now possible by connecting to `SYSTEM_MAINTENANCE`, `SYSTEM_READWRITE`, or any variant of `SYSTEM_FOO`, where `FOO` is the name of a database defined in lib/DBDefs.pm.

This magic is only performed if `SYSTEM_FOO` isn't otherwise defined.

# Testing

 * Tested the InitDb.pl changes by running `dropdb -U postgres musicbrainz_test && ./script/create_test_db.sh`.
 * Tested the psql changes by running `./admin/psql --system MAINTENANCE`.
 * ~Did not test the DisableAllTriggers changes.~ [See below](#issuecomment-1798292409).

### DIFF
--- a/admin/psql
+++ b/admin/psql
@@ -26,9 +26,26 @@ GetOptions(
 
 my $key = shift // 'READWRITE';
 
-if ($help) {
-    die "Usage: psql [database]\n\n  database -- the DBDefs name of the database to connect to. (default is READWRITE)\n";
-}
+die <<EOF if $help;
+Usage: $0 [options] [database]
+
+Arguments:
+
+    database    The name of the database connection definition to use,
+                as registered in DBDefs.
+                (default: READWRITE)
+
+Options:
+
+    --system    Partly overwrite the database connection definition
+                with the properties (usually username/password)
+                of the database connection definition SYSTEM.
+                It allows connecting to any database as superuser.
+             
+Environment:
+
+    PGPASSWORD  If set, it is used as password to connect to the database.
+EOF
 
 my $db = Databases->get($key) or die "No such database '$key'\n";
 
--- a/admin/psql
+++ b/admin/psql
@@ -33,8 +33,7 @@ if ($help) {
 my $db = Databases->get($key) or die "No such database '$key'\n";
 
 if ($system) {
-    my $sys_db = Databases->get('SYSTEM');
-    $db = $db->meta->clone_object($sys_db, database => $db->database);
+    $db = Databases->get('SYSTEM_' . $key);
 }
 
 $ENV{'PGPASSWORD'} = $db->password;
--- a/lib/MusicBrainz/Server/DatabaseConnectionFactory.pm
+++ b/lib/MusicBrainz/Server/DatabaseConnectionFactory.pm
@@ -67,8 +67,18 @@ sub get {
 
     my $database = $databases{$name};
 
-    if ($name eq 'MAINTENANCE' && !defined($database)) {
-        $database = $databases{READWRITE};
+    unless (defined $database) {
+        if ($name eq 'MAINTENANCE') {
+            $database = $databases{READWRITE};
+        } elsif ($name =~ /^SYSTEM_(.+)$/) {
+            my $base_dbdef_key = $1;
+            my $system = $class->get('SYSTEM');
+            $database = $system->meta->clone_object(
+                $system,
+                database => $class->get($base_dbdef_key)->database,
+            );
+            $class->register_database($name, $database);
+        }
     }
 
     return $database;
--- a/admin/DisableAllTriggers
+++ b/admin/DisableAllTriggers
@@ -10,14 +10,7 @@ use DBDefs;
 use aliased 'MusicBrainz::Server::DatabaseConnectionFactory' => 'Databases';
 use MusicBrainz::Server::Constants qw( @FULL_TABLE_LIST );
 
-my $readwrite_db = Databases->get('READWRITE');
-my $system_db = Databases->get('SYSTEM');
-my $sysmb_db = $system_db->meta->clone_object(
-    $system_db,
-    database => $readwrite_db->database,
-);
-Databases->register_database('SYSMB', $sysmb_db);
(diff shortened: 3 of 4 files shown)


reosarevok (approved):
LGTMBDNT. Is there a particular reason you did not test the trigger change? :)

mwiencek:
> Is there a particular reason you did not test the trigger change? :)

Mainly because there's no script to re-enable them. :) (And the script does not allow specifying the database.)

But, I just hand-edited it to use `SYSTEM_TEST`, and verified that it ran without errors.

admin/psql
```diff
@@ -33,8 +33,7 @@ if ($help) {
 my $db = Databases->get($key) or die "No such database '$key'\n";
 
 if ($system) {
-    my $sys_db = Databases->get('SYSTEM');
-    $db = $db->meta->clone_object($sys_db, database => $db->database);
+    $db = Databases->get('SYSTEM_' . $key);
 }
```
yvanzo, line 53: For how to use this script, I appended the commit https://github.com/metabrainz/musicbrainz-server/pull/3079/commits/f41729a8c47b5c177b17a84f9f7bb5edff1ce836.

yvanzo (commented):
Is that a proper way to test it?

```sh
admin/psql --system <<< 'SELECT USER'
```

> Sometimes it's necessary to connect to the database specified by MAINTENANCE or
READWRITE using SYSTEM credentials

1. Do you have actual examples at hand to provide in the commit message?
2. Are “credentials” intended to be username/password or the whole entry as in the code?

mwiencek:
> Is that a proper way to test it?
> ```
> admin/psql --system <<< 'SELECT USER'
> ```

Yes, it should normally print `postgres` (depending on the `SYSTEM` user).

> Do you have actual examples at hand to provide in the commit message?

This is mainly code refactoring. (The `psql --system` flag has already existed for a while.) So the only examples are the code which the commit itself is changing. :)

> Are “credentials” intended to be username/password or the whole entry as in the code?

That's a great question. I would have thought yes, but as you can see in the three files changed here, we actually copy the `SYSTEM` database in full (using its connection details too) and only modifying the database on it. This is also what the new code does, so I think it makes the most sense for compatibility reasons.

lib/MusicBrainz/Server/DatabaseConnectionFactory.pm
```diff
@@ -67,8 +67,18 @@ sub get {
...
+    unless (defined $database) {
+        if ($name eq 'MAINTENANCE') {
+            $database = $databases{READWRITE};
+        } elsif ($name =~ /^SYSTEM_(.+)$/) {
+            my $database_name = $1;
+            my $system = $class->get('SYSTEM');
+            $database = $system->meta->clone_object(
+                $system,
+                database => $class->get($database_name)->database,
+            );
```
yvanzo, line 79: It seems to pick the properties from the `SYSTEM` database connection definition but to overwrite `database` (name in PostgreSQL) with the one from the (expected) database connection definition to use. If so, it is indeed just doing the same as before.

I updated `admin/psql` help message accordingly in commit https://github.com/metabrainz/musicbrainz-server/pull/3079/commits/8b1565ddcea71eeb76e1e6a7cb5bf5240529912b.

yvanzo (commented):
> This is mainly code refactoring.

So does it really “allow” anything new?

> So the only examples are the code which the commit itself is changing. :)

Okay, I thought that it was a prerequisite for some further changes.

mwiencek:
> So does it really “allow” anything new?

In a way it does, since previously this had to be done in code by manually cloning the Moose objects and overwriting the properties. With this change, you could for example pass `SYSTEM_READWRITE` to any script that accepts a `--database` flag, which wasn't possible previously.

> Okay, I thought that it was a prerequisite for some further changes.

It was part of another branch that I scrapped, but I thought it might be useful on its own, so I submitted it as a standalone patch.

lib/MusicBrainz/Server/DatabaseConnectionFactory.pm
```diff
@@ -67,8 +67,18 @@ sub get {
 
     my $database = $databases{$name};
 
-    if ($name eq 'MAINTENANCE' && !defined($database)) {
-        $database = $databases{READWRITE};
+    unless (defined $database) {
+        if ($name eq 'MAINTENANCE') {
+            $database = $databases{READWRITE};
+        } elsif ($name =~ /^SYSTEM_(.+)$/) {
+            my $database_name = $1;
```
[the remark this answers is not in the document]
mwiencek: It's the first match group in the regular expression above, so for example, if it matches `SYSTEM_FOO` then `$database_name` will be `'FOO'`.

I agree that the distinction between `$name` and `$database_name` is not clear here. Perhaps the latter should be `$target_name` or `$underlying_name` or something like that.

lib/MusicBrainz/Server/DatabaseConnectionFactory.pm
mwiencek: I would change "poweruser" to "superuser" since that is what the PostgreSQL documentation uses, but looks good otherwise. :+1:

lib/MusicBrainz/Server/DatabaseConnectionFactory.pm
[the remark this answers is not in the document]
yvanzo: It’s actually the key under which the definition for the database connection has been registered.
```suggestion
            my $dbdef_key = $1;
```

yvanzo (approved):
Okay so the option `--system` remains for compatibility but from now on we could just give it the argument `SYSTEM_FOO` instead.

LGTM besides the potential naming improvements! :ship:

mwiencek:
Applied the suggested `$base_dbdef_key` name, thanks!

PR #3081, opened 2023-11-07
MBS-13326: Add documentation bubble for the artist sort name field

reosarevok:
### Implement MBS-13326

# Problem
"Sort name" is a non-obvious field. Recently we've seen people reading it as "short name", but even without that, it's often filled wrong.

# Solution
There seems to be entirely no downsides to adding a bubble with a short explanation and a link to the guidelines, so I'm doing that and hoping it's enough to clarify the usage.

Screenshot from 2023-11-07 22-18-41

# Testing
Manually, just to make sure the bubble appears and goes away as expected.

### DIFF
--- a/root/components/forms.tt
+++ b/root/components/forms.tt
@@ -359,7 +359,7 @@
       <p>[% l('The sort name is an artist name variant used when sorting alphabetically.
                For usage information check the {sort_name_doc|sort name guidelines}.',
             { sort_name_doc => doc_link('Style/Artist/Sort_Name') }) %]</p>
-      <p>[% l('Last Name, First Name” and “Group, The” are common examples of sort names.') %]</p>
+      <p>[% l('“Last Name, First Name” and “Group, The” are common examples of sort names.') %]</p>
     </div>
 [%- END -%]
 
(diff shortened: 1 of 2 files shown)


reosarevok:
I like the wording, so updated the PR / image.

loucasal:
I think the example line was meant to read "Last Name, First Name" rather than the other way around.

root/components/forms.tt
```diff
@@ -354,6 +354,15 @@
     </div>
 [%- END -%]
 
+[%- MACRO sort_name_bubble BLOCK -%]
+    <div class="bubble" id="sort-name-bubble">
+      <p>[% l('The sort name is an artist name variant used when sorting alphabetically.
+               For usage information check the {sort_name_doc|sort name guidelines}.',
+            { sort_name_doc => doc_link('Style/Artist/Sort_Name') }) %]</p>
+      <p>[% l('Last Name, First Name” and “Group, The” are common examples of sort names.') %]</p>
```
mwiencek, line 362:
```suggestion
      <p>[% l('“Last Name, First Name” and “Group, The” are common examples of sort names.') %]</p>
```

mwiencek (changes requested):

mwiencek (approved):

PR #3082, opened 2023-11-07
MBS-13273: Update track credits with matching prefixes

derat:
# MBS-13273

When release artist credits are updated, also update track credits that start with the old release credits (instead of only ones that exactly match the release credits).

For example, if the release credits are changed from "A" to "B", a track that was credited to "A feat. C" will be updated to "B feat. C".

# Problem

When updating the artist credits on a release, track credits that exactly match the old release credits are automatically updated to use the new credits, but tracks that also include featured artists are not updated.

It's easy for an editor not to notice that some of the tracks weren't updated if the old and new artists have the same name. For example, see [edit #103710653](https://musicbrainz.org/edit/103710653) (later fixed by [edit #103710691](https://musicbrainz.org/edit/103710691)).

# Solution

Suppose a track's credits include T artists and the old release credits include R artists. If the track's reduced credits don't exactly match the release credits and T > R, check if the first R artists from the track credits match the old release credits. If so, update the first R artists in the track credits to instead use the new release credits.

Some examples:
* If a release is updated from `A` to `B` and a track credits `A feat. F`, update the track to credit `B feat. F`.
* If a release is updated from `A & B` to `A & C` and a track credits `A & B feat. F & G`, update the track to credit `A & C feat. F & G`.

# Action

N/A

### DIFF
--- a/root/static/scripts/common/immutable-entities.js
+++ b/root/static/scripts/common/immutable-entities.js
@@ -33,7 +33,26 @@ export const isCompleteArtistCredit =
     ac.names.every(hasArtist);
 
 export const reduceArtistCredit =
-  (ac: ArtistCreditT): string => ac.names.reduce(reduceName, '');
+  (ac: ArtistCreditT): string => reduceArtistCreditNames(ac.names);
+
+/*
+ * Joins the supplied credits into a single string, using credited-as names if
+ * set or artist names otherwise. If dropFinalJoinPhrase is true, the final
+ * join phrase is omitted from the returned string.
+ */
+export function reduceArtistCreditNames(
+  names: $ReadOnlyArray<ArtistCreditNameT>,
+  dropFinalJoinPhrase?: false,
+): string {
+  let s = names.reduce(reduceName, '');
+  if (dropFinalJoinPhrase && names.length > 0) {
+    const finalJoinPhrase = names[names.length - 1].joinPhrase;
+    if (finalJoinPhrase) {
+      s = s.slice(0, s.length - finalJoinPhrase.length);
+    }
+  }
+  return s;
+}
 
 export const isComplexArtistCredit = function (ac: ArtistCreditT): boolean {
   const firstName = ac.names[0];
--- a/root/static/scripts/release-editor/init.js
+++ b/root/static/scripts/release-editor/init.js
@@ -13,11 +13,15 @@ import {
   artistCreditsAreEqual,
   hasVariousArtists,
   reduceArtistCredit,
+  reduceArtistCreditNames,
 } from '../common/immutable-entities.js';
 import MB from '../common/MB.js';
 import {getSourceEntityData} from '../common/utility/catalyst.js';
 import clean from '../common/utility/clean.js';
-import {cloneObjectDeep} from '../common/utility/cloneDeep.mjs';
+import {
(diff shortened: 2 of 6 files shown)


derat:
@reosarevok I'm planning to add Selenium tests for this, but I figured I'd upload it first to get some early feedback about whether this approach seems reasonable.

On the ticket, I suggested only updating tracks if the old and new release credits match when they're reduced, but I _think_ that checking whether the release credit is a prefix of the track credit should be safe. Please let me know if you can think of some cases where this behavior would be undesirable, though.

derat:
@reosarevok, does this seem okay to you (or is there someone else I should ask)? Thanks!

reosarevok:
The idea seems fine to me - I'm sure there's a few rare cases where it won't be ideal (for example, if the artist for that specific track was indeed A) but it's rare enough I don't think it's worth worrying about.

One thing I would possibly consider is making `reduceArtistCreditPrefix` into `reduceArtistCreditSlice` and allowing it to take any specific slice rather than only the start of the array. If there's any case you can think of where the non-"prefix" bits of the ACs could change, then we could do two calls to `reduceArtistCreditSlice` and make sure the "prefix" is changed and the "suffix" is unchanged, but it might not be needed in this case.

derat:
Thanks. The "artist for that specific track was indeed A" case isn't a new one, right? As I understand it, the existing code will already change "A" to "B" in that case; it just won't change e.g. "A feat. F" to "B feat. F".

I considered having a function like `reduceArtistCreditSlice` (which could then just be called by the existing `reduceArtistCredit` function), but I wasn't sure if the "drop the final join phrase" behavior was too subtle to include. Any thoughts about that? I could move the removing-final-join-phrase code into the caller at root/static/scripts/release-editor/init.js, but I thought it might be better to keep the complexity in a place like immutable-entities.js where it's easier to unit-test. Or maybe it'd be best to just give it a signature like `reduceArtistCreditNames(names: $ReadOnlyArray<ArtistCreditNameT>, dropFinalJoinPhrase?: boolean)`... (Sorry about typos; I'm more familiar with TypeScript than with Flow.)

derat:
I've added a `reduceArtistCreditNames` function and unit and Selenium tests. (I suspect it'll take me a few more iterations to get the Selenium test passing on CircleCI...)

derat:

derat:
@reosarevok, at your convenience, I think this is ready for review now. I'll retry the Selenium tests in a bit -- I've seen the same failure the last two runs, but it looks like the server is returning a 500 in response to a request for `http://mbtest:5000/ws/js/label?q=nothing&page=1&direct=false` in the `The Downward Spiral` test, so it's not obvious to me how it could be related to this change.

derat:

mwiencek (approved):
Thanks @derat, I've tested this and it works well. Added tests look good. I'll let @reosarevok approve too just in case he had anything to add.

derat:
@reosarevok, did you have any other comments, or can this be merged?

reosarevok:
I need to test it (I will do that later today) but assuming I don't find anything we can merge it for the next beta release :)

reosarevok (approved):
Now tested locally, no complaints at all.

derat:
Thanks!

PR #3084, opened 2023-11-09
MBS-11934: Add report for releases without CAA

reosarevok:
### Implement MBS-11934

# Description
This report lists all releases that have no images in the Cover Art Archive at all. In its default state that's a ton (around 1.5 million), but this is mostly useful when filtered for subscriptions, allowing users to quickly find all the music by their favourite artists or labels that is missing cover art.

# Testing
I tested locally (on sample) that the report runs and displays fine. That does not have CAA set up, but I quickly tested the query on jimmy to make sure it seemed sensible there and it gave a number of results similar to substracting total releases vs releases with cover art in our statistics from yesterday :)

### DIFF
--- a/root/report/ReleasesWithoutCaa.js
+++ b/root/report/ReleasesWithoutCaa.js
@@ -1,0 +1,47 @@
+/*
+ * @flow strict
+ * Copyright (C) 2023 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import ReleaseList from './components/ReleaseList.js';
+import ReportLayout from './components/ReportLayout.js';
+import type {ReportDataT, ReportReleaseT} from './types.js';
+
+const ReleasesWithoutCaa = ({
+  canBeFiltered,
+  filtered,
+  generated,
+  items,
+  pager,
+}: ReportDataT<ReportReleaseT>): React$Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={exp.l(
+      `This report shows releases that have no cover art in the Cover Art
+       Archive. Given that most releases have some form of cover art, the
+       vast majority of releases in this report should have artwork 
+       {caa_how_to|uploaded to the Cover Art Archive}.
+       Note that the cover art uploaded for a release must always exactly
+       match the actual art for that specific release (for example,
+       they should have the same barcode, format, etc.).`,
+      {caa_how_to: '/doc/How_to_Add_Cover_Art'},
+    )}
+    entityType="release"
+    extraInfo={l(
+      `We strongly suggest restricting this report to entities
+       in your subscriptions only for a more manageable list of results.`,
+    )}
+    filtered={filtered}
+    generated={generated}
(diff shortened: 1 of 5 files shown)


root/report/ReleasesWithoutCaa.js
```diff
@@ -0,0 +1,40 @@
...
+  pager,
+}: ReportDataT<ReportReleaseT>): React$Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows releases which have no cover art in the Cover Art
+       Archive at all. Most releases have some sort of cover art, so the
+       vast majority of these should eventually get some art uploaded to the
+       Cover Art Archive. We suggest restricting this report to your
+       subscriptions for a more manageable list of results.`,
```
yvanzo, line 28: The second phrase here (_Most releases…_) is problematic as it encourages attaching cover art without caution. It happens too frequently that editors are wrongly attaching the cover art from a release (no matter whether it was or wasn’t in the MB database) to another release (in the MB database).

The third phrase is very relevant but should probably be even stronger as this report will always be huge.

yvanzo (commented):
I’m not too sure about this. I thought that reports were expected to be practically emptied with some editor’s work. It doesn’t seem to be the case here. Made a couple of comments just in case.

reosarevok:
Reports are meant to help editors edit - if this helps editors edit, I cannot see any downsides unless it is very slow and database-intensive, which I don't think this is :) (we have things like the admin-only beginner editor report which certainly do not have any expectation of becoming entirely empty, for example)

root/report/ReleasesWithoutCaa.js
reosarevok: I extended the first bits, added a link to the how to docs (seemed useful) and split the last phrase onto a second help row, with stronger wording - does that seem better?

root/report/ReleasesWithoutCaa.js
```diff
@@ -0,0 +1,46 @@
...
+}: ReportDataT<ReportReleaseT>): React$Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={exp.l(
+      `This report shows releases which have no cover art in the Cover Art
+       Archive at all. Most releases have some sort of cover art, so the
+       vast majority of these should eventually get some art
+       {caa_how_to|uploaded to the Cover Art Archive}.
+       Do make sure to only upload images that apply to this specific release
+       (same barcode, format, etc.).`,
```
atj, line 29: My attempt at improving the wording. Not entirely happy with it but I think it reads a bit better:

```suggestion
      `This report shows releases that have no cover art in the Cover Art
       Archive. Given that most releases have some form of cover art, the
       vast majority of releases in this report should have artwork 
       {caa_how_to|uploaded to the Cover Art Archive}.
       Note that the cover art for a release must always exactly match the
       actual art for that specific release (e.g. same barcode, format, etc.).`,
```

reosarevok: We try to avoid `e.g.` (we had a PR recently changing all those to "such as") but in general this seems decent to me. @mwiencek, what do you think?

mwiencek: Sounds good to me. :+1:

yvanzo (approved):
Tested  locally with sample data. Thanks for the improvements! :ship:

PR #3087, opened 2023-11-09
Fix formatting OAuth URI error messages

yvanzo:
# Problem

* `<code>` was wrongly displayed in the error messages for the callback URL when registering an application as a developer.
* The example URL was needlessly part of the message to be translated.

# Solution

* Replace `<code>` with double quotes.
* Use a variable placeholder for the example URL.

# Checklist for author

* [x] Testing the form `http://localhost:5000/account/applications/register` with the callback URL `javascript:boo` and either type of application.

### DIFF
--- a/lib/MusicBrainz/Server/Form/Application.pm
+++ b/lib/MusicBrainz/Server/Form/Application.pm
@@ -48,15 +48,16 @@ sub validate
             );
         } elsif ($self->field('oauth_redirect_uri')->value !~ $OAUTH_WEB_APP_REDIRECT_URI_RE) {
             $self->field('oauth_redirect_uri')->add_error(
-                l('Redirect URL scheme must be either <code>http</code> or ' .
-                  '<code>https</code> for web applications.')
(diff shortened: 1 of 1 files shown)


reosarevok (approved):
Hmm. I would have thought this would need `use utf8` on the file, but I guess not since you tested it? Looks good to me otherwise. 

There's another string with `<code>` in `Controller::Account` - does that also break then? I wouldn't mind changing `<code>{email}</code>` to `“{email}”` anyway.

yvanzo:
Good catch. It was probably using [utf8 from Field::URL](https://github.com/metabrainz/musicbrainz-server/blob/2468b81109bc8d6702e564b11371a8aaad71a773/lib/MusicBrainz/Server/Form/Field/URL.pm#L5) you committed recently. Even though it was correctly displayed, I just added an explicitly use of utf8 in Form::Application where the characters are written.

yvanzo:
As for the other `<code>` occurrence, I just tested it locally and it works fine. I guess the issue is with `add_error` only.

PR #3088, opened 2023-11-10
MBS-13362: Handle Spotify prerelease URLs

yvanzo:
# Problem MBS-13362

Spotify prerelease URLs were unknowingly blocked from being added to the external links of upcoming releases.

Note: Prerelease Spotify URLs point to countdown pages until the time the album is released, then it redirects to the album, so they allow streaming the music of the album once the countdown is reached.

For reference: https://artists.spotify.com/blog/countdown-pages-get-fans-hyped-for-your-new-album-spotify-stream-on

# Solution

Auto-selecting the streaming relationship type for Spotify streaming URLs, as it eventually allows to stream it.

Ideally the relationship begin date should be set to the countdown date.

# Checklist for author

* [x] Added CI test

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -5216,7 +5216,7 @@ const CLEANUPS: CleanupEntries = {
             };
           case LINK_TYPES.streamingfree.release:
             return {
-              result: prefix === 'album',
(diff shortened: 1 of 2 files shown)


reosarevok (approved):
While it's not quite a streaming link, since it will redirect into one eventually it probably makes sense to allow linking to it.

PR #3089, opened 2023-11-11
Add a tool that can execute any file via Webpack

mwiencek:
# Problem

Sometimes I want to write a quick, one-off script that imports some of our JavaScript modules and executes some functions in them. This is very difficult because the modules probably rely on Flow syntax and Webpack auto-imports.

# Solution

Adds webpack/exec.mjs which can be used for this. Documents it and ./bin/sucrase-node (which is actually used to execute webpack/exec.mjs) in HACKING.md.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -571,6 +571,38 @@ modern browsers when compiling static resources by specifying the
 env MODERN_BROWSERS=1 ./script/compile_resources.sh
 ```
 
+Sometimes you may want to test how a certain module or function behaves
+locally, in Node.js, without setting up a whole test suite or Webpack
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
LGTM. Tested the example locally: It returns the expected result (even though preceded with a warning which is not new to this pull request):

    (node:1778) [DEP0128] DeprecationWarning: Invalid 'main' field in '/musicbrainz-server/node_modules/po2json/package.json' of './lib/po2json'. Please either fix that or report it to the module author
    (Use `node --trace-deprecation ...` to show where the warning was created)
    
    1, 2, 3

PR #3090, opened 2023-11-12
MBS-13365: Flatten commaOnlyList result

mwiencek:
# Problem

MBS-13365

It turns out that React really doesn't like if you try to render a component hierarchy with a depth of, say, 2,000.

This was triggered in production by passing a list of close to 2,000 elements (the md5 hashes of IP addresses) to `commaOnlyList`.  Fortunately (or unfortunately for admins), it only triggered if you could see these IP hashes; and only if the user managed to browse the website under that many IPs.  (At least two active users have.)

# Solution

`commaOnlyList` works by iterating over the input items and feeding the current result (`rest`) back into `'{commas_only_list_item}, {rest}'`.  That created a new parent fragment for each item.  The solution is to allow arrays as input for variable substitutions, which are flattened into the current result.

`exp.l` still returns a fragment by default, because although React can render arrays of elements, I feared that returning them by default would trigger tons of missing key warnings (an issue which does not exist for fragments). Therefore it only returns an array if you specify `__wantArray: 'true'`.

While this commit fixes the issue by allowing us to render massive comma-separated lists, I don't think listing 2,000 IP hashes is actually useful.  But we would need to figure out a way to either paginate them, or sort and truncate them, which I did not feel like figuring out.


mwiencek:
Since this is only visible to admins, we could just *not* use `commaOnlyList`.  But I liked the idea of supporting arrays internally, thereby making the function more efficient, and it wasn't that hard to implement.

mwiencek:
Yes, I've copied the same changes over to `commaList`.

reosarevok (commented):
LGTMBDNT. Should we have different tests for the `wantArray` vs standard behaviour? I'm a bit confused about the new `expandHtml` test which seems to be the only appearance of `expandHtml` in this PR.

yvanzo (commented):
>  I feared that returning them by default would trigger tons of missing key warnings (an issue which does not exist for fragments).

Do you fear or did you experience it? I modified the code locally to force array and  tested area containment but that might not be the most relevant test.
If returning array is more efficient, should we generalize that as much as possible?

mwiencek:
> Do you fear or did you experience it? I modified the code locally to force array and tested area containment but that might not be the most relevant test.

In expand2react.js, I changed:

```js
    return result.length ? (
      result.length > 1
        ? (
          (args != null && args.get('__wantArray') === 'true')
            ? result
            : React.createElement(React.Fragment, null, ...result)
        )
        : result[0]
    ) : '';
```
to:
```js
    return result.length ? (
      result.length > 1
        ? result
        : result[0]
    ) : '';
```
and there were a lot of new key warnings logged on both the server and browser (try an artist page, for example).

mwiencek:
> Should we have different tests for the wantArray vs standard behaviour?

Added a few tests for `__wantArray`.

> I'm a bit confused about the new expandHtml test which seems to be the only appearance of expandHtml in this PR.

It's just testing that you can use arrays as variable substitutions. It uses `expandHtml` because it's checking the final HTML.

PR #3091, opened 2023-11-15
MBS-13117: Prevent localizing admin-only messages

yvanzo:
# Problem MBS-13117

Messages intended for admins only are needlessly extracted for translation.
Indeed admins are required to be English-speaking anyway.
So it is a waste of time for translators so far.

# Solution

* Prevent extracting any message from admin-only pages.
* Add and use `((t)exp).(N_)l[np]_admin` functions for admin-only messages in other pages.
  It makes explicit that some messages are intentionally not translated because these are intended for admin.
  This approach works well in JavaScript, should we follow the same in Perl (under `lib` only)?

# Draft progress checklist

* [x] Test navigating admin and non-admin pages with different UI languages
* [x] Decide on an approach for admin messages in Perl. See current occurrences:
  ```sh
  git grep -w '\(N_\)\?l[np]\?' $(git ls-files 'lib/**Admin**')
  ```
* [x] Implement the same approach in both JavaScript and Perl.
* [x] Uniformize the existing admin code to use this approach.
* [x] Re-test regenerating the file `po/mb_server_pot` <del>after the singular/plural issue has been resolved</del>.
* [x] Re-test navigating admin and non-admin pages with different UI languages

# Action

* Revert the last committed hack when resolving MBS-11178.

### DIFF
--- a/po/Makefile
+++ b/po/Makefile
@@ -10,13 +10,13 @@ GIT_LS_PERL:=git ls-files -- \
 	| $(SED_ESCAPE_SPACES)
 
 GIT_LS_JS:=git ls-files -- \
-    ':(top,glob)root/**/*.js' \
-    ':(top,glob,exclude)root/admin/**/*.js' \
-    ':(top,glob,exclude)root/static/scripts/tests/**/*.js' \
-    ':(top,glob,exclude)root/static/build/**/*.js' \
-    ':(top,glob,exclude)root/static/lib/**/*.js' \
-    ':(top,glob,exclude)root/statistics/**/*.js' \
-    | $(SED_ESCAPE_SPACES)
+	':(top,glob)root/**/*.js' \
+	':(top,glob,exclude)root/admin/**/*.js' \
+	':(top,glob,exclude)root/static/scripts/tests/**/*.js' \
+	':(top,glob,exclude)root/static/build/**/*.js' \
+	':(top,glob,exclude)root/static/lib/**/*.js' \
+	':(top,glob,exclude)root/statistics/**/*.js' \
+	| $(SED_ESCAPE_SPACES)
 
 GIT_LS_ADMINPERL:=git ls-files -- \
 	':(top,glob)lib/**/*Admin*.pm' \
@@ -32,10 +32,10 @@ GIT_LS_ADMINJS:=git ls-files -- \
 	| $(SED_ESCAPE_SPACES)
 
 GIT_LS_TT:=git ls-files -- \
-    ':(top,glob)root/**/*.tt' \
-    ':(top,glob,exclude)root/admin/**/*.tt' \
-    ':(top,glob,exclude)root/statistics/**/*.tt' \
-    | $(SED_ESCAPE_SPACES)
+	':(top,glob)root/**/*.tt' \
+	':(top,glob,exclude)root/admin/**/*.tt' \
+	':(top,glob,exclude)root/statistics/**/*.tt' \
+	| $(SED_ESCAPE_SPACES)
 
 GIT_LS_STATTT:=git ls-files -- \
 	':(top,glob)root/statistics/**/*.tt' \
--- a/po/Makefile
+++ b/po/Makefile
@@ -18,6 +18,19 @@ GIT_LS_JS:=git ls-files -- \
     ':(top,glob,exclude)root/statistics/**/*.js' \
     | $(SED_ESCAPE_SPACES)
 
+GIT_LS_ADMINPERL:=git ls-files -- \
+	':(top,glob)lib/**/*Admin*.pm' \
+	':(top,glob)lib/**/*Admin*/**/*.pm' \
+	| $(SED_ESCAPE_SPACES)
+
+GIT_LS_ADMINTT:=git ls-files -- \
+	':(top,glob)root/admin/**/*.tt' \
+	| $(SED_ESCAPE_SPACES)
+
+GIT_LS_ADMINJS:=git ls-files -- \
+	':(top,glob)root/admin/**/*.js' \
+	| $(SED_ESCAPE_SPACES)
+
 GIT_LS_TT:=git ls-files -- \
     ':(top,glob)root/**/*.tt' \
     ':(top,glob,exclude)root/admin/**/*.tt' \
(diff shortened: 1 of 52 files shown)


root/static/scripts/common/i18n/admin.js
```diff
@@ -0,0 +1,46 @@
...
+
+// eslint-disable-next-line no-unused-vars -- map context
+export const lp_admin = (key: string, context: string): string => (
+  cleanMsgid(key)
+);
+
+export {lp_admin as lp};
+
+export const N_l_admin = (key: string): (() => string) => (
+  () => l_admin(key)
```
mwiencek, line 31: Something I'm not sure I get is the point of having both `l_admin` and `N_l_admin` (and likewise for ln, lp), if both are functionally no-ops and the strings aren't meant to be extracted. I feel that it might make sense to rename `l_admin` to `N_l_admin` (and remove the current `N_l_admin`) to make it clear that these functions are always no-ops. (Returning a function isn't needed for the same reasons.)

root/admin/wikidoc/EditWikiDoc.js
```diff
@@ -8,6 +8,7 @@
  */
 
 import Layout from '../../layout/index.js';
+import {l} from '../../static/scripts/common/i18n/admin.js';
```
mwiencek, line 11: Any reason not to change the strings to use `l_admin` instead? (I thought it was because these ones are still meant to be extracted/localized, but they seem to be excluded from po/Makefile.)

root/static/scripts/common/i18n/admin.js
yvanzo: `l_admin` isn’t exactly no-op though, it is `cleanMsgid`.

mwiencek: That's right. I was thinking more in terms of no-localization. :)

I'm not that concerned about which name to use, anyway, though it still seems like we don't need both of these.

root/admin/wikidoc/EditWikiDoc.js
mwiencek: Keeping the code compact is a good point, but particularly for `l_statistics` which is longer and might be horrible to repeat in a file with many strings.

And I guess if we stick with `l`, `ln`, etc., we could still automatically check that they are only imported from `admin.js`.

yvanzo: > And I guess if we stick with `l`, `ln`, etc., we could still automatically check that they are only imported from `admin.js`.

I would love to see that, just not too sure how since those are provided for all files through webpack plugin config.

mwiencek: One option would be to have script/xgettext.js itself throw an error, since it has the paths of the files it's operating on.

But xgettext.js doesn't see where the functions are imported from. The best we could do there is error if *any* i18n functions are used in these files: so only raw strings would be allowed (and expand2react for the cases where substitution is needed).

root/static/scripts/common/i18n/admin.js
yvanzo: Okay, I dropped all the `N_*_admin`.

root/admin/wikidoc/EditWikiDoc.js
yvanzo: I expanded CircleCI tests to build a new target `test_no_admin` from `po/Makefile` that will fail if anything is extracted from admin-only files.

Consequently I just dropped this kind of import.

root/admin/PrivilegeSearch.js
```diff
@@ -41,109 +41,109 @@ const PrivilegeSearch = ({
...
+      <h1>{'Search users by privileges'}</h1>
 
       <form action="/admin/privilege-search" method="get">
         <p>
-          {l(`Select the flags you want to match. Editors which have
-              other flags in addition to the selected ones will also be shown,
-              unless you select “Exact match only” below.`)}
+          {`Select the flags you want to match. Editors which have
+            other flags in addition to the selected ones will also be shown,
+            unless you select “Exact match only” below.`}
```
mwiencek, line 52: No `l_admin` here?

po/Makefile
```diff
@@ -139,3 +158,38 @@ check_quiet:
 	@for file in $(POFILES); do \
 		msgfmt -o /dev/null -c $$file;\
 	done
+
+test_source: test_no_admin
+
+test_no_admin: $(ADMINPERL) $(ADMINTT) $(ADMINJS) $(wildcard extract_pot_templates)
+	@echo "Testing that no admin files contain any translatable messages (MBS-13117)";
+	@count=0; \
+	for file in $(ADMINPERL); do \
```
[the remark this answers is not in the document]
yvanzo: Good idea, moved in commit https://github.com/metabrainz/musicbrainz-server/pull/3091/commits/6e0619eba423e2c5d6e44ab42ba2857f1242525d

root/admin/EditBanner.js
```diff
@@ -22,21 +23,21 @@ type Props = {
...
 const EditBanner = ({form}: Props): React$Element<typeof Layout> => (
-  <Layout fullWidth title={l('Edit Banner Message')}>
+  <Layout fullWidth title="Edit Banner Message">
     <div id="content">
-      <h1>{l('Edit banner message')}</h1>
+      <h1>{'Edit banner message'}</h1>
       <p>
-        {l(`This will set the banner message that is shown at the top
-            of each page. An empty string removes the banner.`)}
+        {l_admin(`This will set the banner message that is shown at the top
```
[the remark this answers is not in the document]
yvanzo: Yes :)

root/static/scripts/common/i18n/expand2react.js
```diff
@@ -395,15 +400,33 @@ export const l = (
...
   key: string,
   context: string,
   args?: ?VarArgsObject<Input>,
 ): Output => expand2react(lpActual(key, context), args);
+
+export const lp_admin = (
+  key: string,
+  context: string,
+  args?: ?VarArgsObject<Input>,
+): Output => expand2react(lpAdminActual(key, context), args);
```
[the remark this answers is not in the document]
yvanzo: Good point, dropped in https://github.com/metabrainz/musicbrainz-server/pull/3091/commits/015e3e8438dcfbe9b0c259cab88f0b2eab5e031b

mwiencek (approved):
Looks great, thanks!

reosarevok (approved):
This seems good with the caveat that assuming we merge https://github.com/metabrainz/musicbrainz-server/pull/2480 first, which would seem the most sensible order, it will need some rebasing to "un-delete" some of the translation calls for the files that are no longer admin only (any forms would still be admin-only IIRC but since they won't be under /admin maybe they should use `l_admin` instead of nothing).

yvanzo:
Just rebased on latest `master` to resolve conflicts.

PR #3092, opened 2023-11-15
MBS-12239: Fix "too many URLs" in incremental sitemaps

mwiencek:
# Problem

MBS-12239

If the incremental sitemaps have not run for some time, then many changes can accumulate which are too numerous to bundle into a single sitemap without exceeding `$MAX_SITEMAP_SIZE`.

# Solution

I implemented batching code for the incremental sitemaps which is mostly copied from the overall sitemaps code.

# Testing

I copied these changes into the production musicbrainz-sitemaps container and ran the hourly script. It appears to have resolved the errors.

### DIFF
--- a/lib/MusicBrainz/Server/Sitemap/Incremental.pm
+++ b/lib/MusicBrainz/Server/Sitemap/Incremental.pm
@@ -7,14 +7,17 @@ use feature 'state';
 
 use Digest::SHA qw( sha1_hex );
 use JSON qw( decode_json );
-use List::AllUtils qw( partition_by );
+use List::AllUtils qw( min partition_by );
 use Moose;
 use Sql;
 
 use MusicBrainz::Server::Constants qw( entities_with );
 use MusicBrainz::Server::Context;
 use MusicBrainz::Server::Log qw( log_info );
-use MusicBrainz::Server::Sitemap::Constants qw( %SITEMAP_SUFFIX_INFO );
+use MusicBrainz::Server::Sitemap::Constants qw(
+    $MAX_SITEMAP_SIZE
+    %SITEMAP_SUFFIX_INFO
+);
 use MusicBrainz::Server::Sitemap::Builder;
 
 extends 'MusicBrainz::Server::Sitemap::Builder';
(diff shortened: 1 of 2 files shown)


mwiencek:
Hint: View the diff with "Hide whitespace" enabled.

lib/MusicBrainz/Server/Sitemap/Incremental.pm
```diff
@@ -217,52 +220,123 @@ sub post_replication_sequence {
...
+        #
+        # The batching code is mostly copied from `build_one_entity` in
+        # `Sitemap::Overall`. However, the logic to always put the last batch
+        # in its own sitemap is not performed here, since we're not as
+        # concerned about which sitemap a URL starts in for an *incremental*
+        # dump. (As mentioned previously, it's very unlikely there will be
+        # more than one batch, anyway, due to the code that attempts to merge
+        # adjacent ones.)
+
+        # Find the counts in each potential batch of 50,000.
```
yvanzo, line 236:
```suggestion
        # Find the counts in each potential batch of $MAX_SITEMAP_SIZE.
```

lib/MusicBrainz/Server/Sitemap/Incremental.pm
```diff
@@ -217,52 +220,123 @@ sub post_replication_sequence {
...
 
-        my %updates_by_suffix_key = partition_by {
-            $_->{sitemap_suffix_key}
-        } @{$all_updates};
+        my $batch = {count => 0, batches => []};
+        my @batches = ($batch);
+        for my $raw_batch (@$raw_batches) {
+            # Add this potential batch to the previous one if the sum will
+            # come out less than 50,000. Otherwise create a new batch and
+            # push the previous one onto the list.
```
yvanzo, line 259:
```suggestion
            # Add this potential batch to the previous one if the sum will
            # come out less than $MAX_SITEMAP_SIZE. Otherwise create a new
            # batch and push the previous one onto the list.
```

lib/MusicBrainz/Server/Sitemap/Incremental.pm
```diff
@@ -217,52 +220,123 @@ sub post_replication_sequence {
...
 
-        for my $suffix_key (sort keys %updates_by_suffix_key) {
-            my $updates = $updates_by_suffix_key{$suffix_key};
+        for my $batch_info (@batches) {
+            my $minimum_batch_number = min(@{ $batch_info->{batches} });
+
+            # The sitemaps.control columns will be NULL on first run, in
+            # which case we just select all updates for writing.
+            my $updates = $c->sql->select_list_of_hashes(
+                <<~"SQL",
```
yvanzo, line 278: Interpolation isn’t needed here.
```suggestion
                <<~'SQL',
```

lib/MusicBrainz/Server/Sitemap/Incremental.pm
```diff
@@ -217,52 +220,123 @@ sub post_replication_sequence {
...
+        # sitemaps have not run for some time and many changes have
+        # accumulated (MBS-12239).
+        #
+        # The batching code is mostly copied from `build_one_entity` in
+        # `Sitemap::Overall`. However, the logic to always put the last batch
+        # in its own sitemap is not performed here, since we're not as
+        # concerned about which sitemap a URL starts in for an *incremental*
+        # dump. (As mentioned previously, it's very unlikely there will be
+        # more than one batch, anyway, due to the code that attempts to merge
+        # adjacent ones.)
```
yvanzo, line 234: I’m not sure to understand the differences mentioned in the second sentence. What is to _put the last batch in its own sitemap_? Why are we _not as concerned about which sitemap a URL starts in for an *incremental* dump_?

Also about the sentence in parenthesis: What if there is more than one batch still? Where is _the code that attempts to merge adjacent ones_? (and which _ones_? batches?)

yvanzo (commented):
Made minor suggestions only. A bit lost with one of your comments in code.

lib/MusicBrainz/Server/Sitemap/Incremental.pm
mwiencek: > I’m not sure to understand the differences mentioned in the second sentence. What is to _put the last batch in its own sitemap_? Why are we _not as concerned about which sitemap a URL starts in for an _incremental_ dump_?

I tried to improve the comment in `Sitemap::Overall::build_one_entity` to explain the batching more, and also the comment here. Does it help at all?

> Also about the sentence in parenthesis: What if there is more than one batch still? Where is _the code that attempts to merge adjacent ones_? (and which _ones_? batches?)

Not sure I follow the first question (the PR makes it possible for there to be more than one batch). The code to merge adjacent batches for incremental dumps is below this comment (via the `for my $raw_batch (@$raw_batches)` loop). For the overall sitemaps, it's in `Sitemap::Overall::build_one_entity` which the comment here references.

lib/MusicBrainz/Server/Sitemap/Incremental.pm
mwiencek: It uses interpolation in the `FROM` table name: `sitemaps.${entity_type}_lastmod`

yvanzo: Oops, sorry!

lib/MusicBrainz/Server/Sitemap/Incremental.pm
yvanzo: To be honest, I’m not even sure what/why/how are batches to start with.  There seems to be many prerequisites missing to allow any understanding. This isn’t new to this pull request though as I’ve been under the same impression when reading `Sitemap::Overall::build_one_entity`. Maybe an architecture documentation in the wiki and/or a Markdown file would be helpful. I’ve just created MBS-13508 (and a _Sitemaps_ sub-component under _Back-end_) for whenever you’d like to make us understand how things are and work. :world_map:

yvanzo (approved):
This works at least and is running in production already, so let’s just merge this pull request and start a documentation later on with MBS-13508 as a reminder. :ship:

lib/MusicBrainz/Server/Sitemap/Incremental.pm
mwiencek: OK, thanks for the ticket. ;) The batching is just a way to split the sitemap files into smaller chunks. For example, the overall sitemap files `sitemap-artist-$batch.xml.gz` contain every artist URL in MusicBrainz...which is a lot of URLs.

Sitemap files have a `lastmod` timestamp associated with them which is stored in the [sitemap index](https://musicbrainz.org/sitemap-index.xml.gz). So if a particular batch hasn't changed since the previous build, we preserve its `lastmod` time. That indicates to crawlers that that particular file doesn't need to be downloaded or checked again. (You can imagine how this would be pointless if all artist URLs were contained in a single file.)

Additionally, in many sitemap files we have `lastmod` times for *the individual URLs*. These times indicate *when the JSON-LD on the page changed*.

But on pages which don't have JSON-LD, the sitemaps basically just list all the URLs for a particular path suffix. (In which case the `lastmod` of a batch only changes if a URL is removed, because of an entity being merged or deleted, or if a new URL is added, which can only happen in the very last batch.)

This PR adds batching to the incremental sitemaps, which output the URLs of pages whose JSON-LD has changed in the past hour. Normally there is not 50,000 such changes in an hour, except in the extreme case I mentioned. Since these URLs change hourly, the `lastmod` properties of the incremental batches don't help much (since they will change every hour anyway). The batching for these is mainly a way to work around a hard-coded limit (`$MAX_SITEMAP_SIZE`) without making that configurable.

PR #3093, opened 2023-11-15
MBS-13055, MBS-13340: Instrument search improvements

mwiencek:
# Problem

MBS-13055
MBS-13340


mwiencek:
With instrument aliases being outputted (as simple string lists), `/ws/js/type-info/link_attribute_type` has seen the following increase in size (gzip listed first):

Before: 96.28 kB (485.93 kB size)
After: 166.09 kB (654.97 kB size)

reosarevok (approved):

yvanzo (approved):
Successfully tested on sample data with `Guitar` for the case-sensitivity issue, and with added aliases of different kind for the search by alias. Added CI test looks good too! :ship: 

> you can't actually see what your search term is matching against , but neither could you with the (external) indexed search.

Yes you can’t see it through our Search API, but it can be seen in Solr admin console for debugging purpose, and there is a feature for [highlighting](https://solr.apache.org/guide/solr/latest/query-guide/highlighting.html) that we don’t use for now.

> the aliases are weighted the same as the instrument name now.

Note that the Solr indexed search currently weights the name with 2 and aliases with 1.2 but it seems to be rather arbitrary. See https://github.com/metabrainz/mbsssss/blob/v-2021-05-14/instrument/conf/request-params.xml#L5

reosarevok:
Ideally the weights would match the search server ones I guess, but this seems good enough for now and a big improvement so if that's not entirely trivial I'd suggest merging this and adding a ticket for the weight change.

yvanzo:
These weights are _arbitrary_ and compared to other search fields, so it doesn’t seem worth matching those here.

reosarevok:
Ok, as you prefer :) Just "name being stronger than alias" in some way seems maybe worth having as a concept, but yeah, certainly not a big deal.

yvanzo:
That would be assuming that the instrument is searched by its English name rather than its name in any language. It might make sense to differentiate non-primary aliases but those are very rare for instruments.
Those _arbitrary_ weights are the same for most entity types and is probably not very adapted to instruments.

PR #3094, opened 2023-11-17
MBS-13368: Properly compare series relationships if series is target

reosarevok:
### Fix MBS-13368

# Problem
We were only performing special series checks in compareRelationships in cases where the series entity is the source. This means entering relationships to series from other entities was entirely skipping this mechanism, so that different numbers for the same series got blocked as a duplicate relationship.

# Solution
This commit makes it so that the same mechanism is used if both targets are the same series, and the other conditions apply.

# Testing
Manually. We don't seem to have any tests for rejecting duplicate relationships (just for merging them); we probably should. @mwiencek, you might be a lot faster writing these than I would, could I convince you to?

### DIFF
--- a/root/static/scripts/relationship-editor/utility/compareRelationships.js
+++ b/root/static/scripts/relationship-editor/utility/compareRelationships.js
@@ -318,6 +318,28 @@ export default function compareRelationships(
     if (seriesItemCmp) {
       return seriesItemCmp;
     }
+  } else if (
+    targetIdCmp === 0 &&
+    targetA.entityType === 'series' &&
+    targetA.orderingTypeID === SERIES_ORDERING_TYPE_AUTOMATIC &&
+    linkTypeId != null &&
+    PART_OF_SERIES_LINK_TYPE_IDS.includes(linkTypeId)
+  ){
+    const seriesItemCmp = (
+      /*
+       * If the number attributes are different, the relationships would
+       * reference different rows in the link table, which is enough to
+       * establish uniqueness.
+       */
+      compareStrings(
+        getPaddedSeriesNumber(a),
+        getPaddedSeriesNumber(b),
+      ) ||
+      compareDatePeriods(a, b)
+    );
+    if (seriesItemCmp) {
+      return seriesItemCmp;
+    }
   }
 
   const linkOrderCmp = a.linkOrder - b.linkOrder;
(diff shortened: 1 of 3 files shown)


mwiencek:
Sorry if I'm being dense here...

It took me a while to understand this, since the ticket and PR descriptions are kinda unclear...but now I'm not sure if I understand it anymore, because the patch doesn't actually work for me. :laughing: 

Just to be clear: is this about adding the **same** item to the **same** series multiple times from the **item** edit page but with different number attributes? I honestly wasn't sure at first, because that seems like a fairly uncommon scenario -- but that's the only case where the series would be the target, and the relationships would only differ by number attribute, right?

Yet if I do exactly that, I still get "This relationship already exists." in the dialog with this patch.

root/static/scripts/relationship-editor/utility/compareRelationships.js
```diff
@@ -318,6 +318,28 @@ export default function compareRelationships(
     if (seriesItemCmp) {
       return seriesItemCmp;
     }
+  } else if (
+    targetA === targetB &&
```
mwiencek, line 322: Maybe this is the issue, because the objects differ? (Untested.)
```suggestion
    targetIdCmp === 0 &&
```

reosarevok: Hmm. The issue is what you suggested (it came up because two episodes of a podcast were released only bundled on one release, so the release needed to be linked as two numbers to the podcast series) and it actually did work when I was testing :) But it might be that the objects are not always the same? Maybe if one is not a full result of a search but saved as a recent item? I'll test.

reosarevok: Ok, it wasn't working now. I really dunno why it worked on my previous test then, but in any case your suggestion works and seems sensible, so pushed it.

mwiencek (approved):
Thanks, I tested it again and it works fine now. A `compareRelationships` test would be nice. You can wait for me to look into it if you want, but not sure how long it'll take. :)

reosarevok:
Tests added, at least some anyway.

mwiencek:
The tests look good, but ideally they'd check what the returned ordering is, not just `!== 0`. :)

reosarevok:
Fair - I was only using them to test they should not be considered the same, but I guess that's not all we use the function for, heh. Changed!

mwiencek (approved):
Thanks!

PR #3095, opened 2023-11-17
MBS-13366: Also show medium data on Edit medium preview

reosarevok:
### Implement MBS-13366

# Problem
While this data is obviously less important here than on a general edit listing, since we know at least what the release is, it seems like a good idea to still show what edit preview applies to what medium, especially for multi-medium releases.

# Solution
I couldn't think of any single good reason to drop this in previews (it doesn't even take much space) so I just dropped the difference entirely.

# Testing
Manually.

### DIFF
--- a/root/edit/details/EditMedium.js
+++ b/root/edit/details/EditMedium.js
@@ -455,14 +455,12 @@ const EditMedium = ({edit}: Props): React$MixedElement => {
 
   return (
     <table className="details edit-medium">
-      {edit.preview /*:: === true */ ? null : (
-        <tr>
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
Tested locally with the sample database.

It seems that it could be hidden in the most common case of previewing edit to one-medium release. But the current change is already an improvement over the current situation for previewing multi-medium release. :ship:

reosarevok:
I did consider that, and I would have gone for it if it was a larger block and got in the way - since it's so small, it feels that there's no significant drawback in just displaying it consistently.

PR #3096, opened 2023-11-17
Drop hard-coded IP address for incrental JSON dump

yvanzo:
# Problem

The IP address of the SSH server used to sync incremental JSON dumps could not be changed without changing it in the current container, and ultimately in the source code/image.

# Solution

This patch instead uses the hostname `mbincrementaljson.local` that must be defined in the container, for example by using the option `--add-host` of `docker run`.

It follows the same approach as for full export dump.

It intentionally targets the branch `production` for a deployment today as the SSH server will be moved.

# Action

1. Make sure to set `mbincrementaljson.local` through `--add-host` before deploying a new `json-dump` container.


mwiencek (approved):
Makes sense. Thank you!

PR #3098, opened 2023-11-18
MBS-13369 (1/3): Also allow Bugs! MV links on releases

Maxr1998:
# Problem MBS-13369 (1/3)

Currently, music videos from Bugs! are only allowed on recordings.

# Solution

Like YouTube videos, it would make sense to be able to add them to releases as well.

PR #3099, opened 2023-11-18
MBS-13369 (2/3): Add support for Melon MV links

Maxr1998:
# Problem MBS-13369 (2/3)

Currently, music videos on Melon are unsupported due to the existing link validation.

# Solution

This patch adds support to add those links to releases and recordings.


yvanzo (approved):
Same as https://github.com/metabrainz/musicbrainz-server/pull/3098#pullrequestreview-1739136317.

reosarevok (approved):
Seems fine to add these IMO.

PR #3101, opened 2023-11-18
MBS-13370: Rewrite more artist pages for Genie

Maxr1998:
# Problem

Genie has multiple tabs on their artist pages, the main artist page uses the format https://genie.co.kr/detail/artistInfo?xxnm=81599561, while other pages showing albums, songs and MVs differ slightly (e.g., https://genie.co.kr/detail/artistAlbum?xxnm=81599561).

# Solution

They should be rewritten to the (canonical) main artist page so that relationship types are auto-selected.

Closes MBS-13370.

# Testing

Added an extra test case that rewrites the albums page for an artist to their main page.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -2674,6 +2674,7 @@ const CLEANUPS: CleanupEntries = {
     },
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(?:(?:www|mw)\.)?genie\.co\.kr\//, 'https://www.genie.co.kr/');
+      url = url.replace(/^https:\/\/www\.genie\.co\.kr\/detail\/artist(?:Info|Album|Song|Mv)\?xxnm=(\d+)(?:[&#\/].*)?$/, 'https://www.genie.co.kr/detail/artistInfo?xxnm=$1');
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -2674,6 +2674,7 @@ const CLEANUPS: CleanupEntries = {
     },
     clean: function (url) {
       url = url.replace(/^(?:https?:\/\/)?(?:(?:www|mw)\.)?genie\.co\.kr\//, 'https://www.genie.co.kr/');
+      url = url.replace(/^https:\/\/www\.genie\.co\.kr\/detail\/artist(?:Info|Album|Song|Mv)\?xxnm=(\d+)(?:[&#\/].*)?$/, 'https://www.genie.co.kr/detail/artistInfo?xxnm=$1');
       url = url.replace(/^(https:\/\/www\.genie\.co\.kr\/detail\/(?:albumInfo\?ax|artistInfo\?xx|songInfo\?xg|mediaInfo\?xv)nm=\d+)(?:[&#\/].*)?$/, '$1');
```
reosarevok, line 2678: Both of these look at `artistInfo` - should we drop it from the second line? Is it needed twice?

Are there similar multiple pages for album, song, media?

Maxr1998: I think we can remove it from the second line, yes.
There are tabs on album and song details, but all those other tabs redirect to the artist page, so they're already handled by this.

Maxr1998: Cleaned up the RegEx, waiting for the tests to pass now.

reosarevok (approved):
Seems good now.

yvanzo (approved):
LGTMBDNT. Thank you again!

PR #3102, opened 2023-11-18
RemoveUnreferencedRows: Silence warning for already-deleted rows

mwiencek:
# Problem

We have the following useless warnings in our daily cron log:
```
2023-11-18T00:16:37.053937Z Use of uninitialized value $ref_count in numeric eq (==) at
/home/musicbrainz/musicbrainz-server/admin/cleanup/../../lib/MusicBrainz/Script/RemoveUnreferencedRows.pm line 94.
```

# Solution

If `$ref_count` is `undef`, then it means the row was already removed. This may be normal, as in the case of artist credits, but it's safe to skip here in any case. Log a message about it just in case it helps with debugging something later.

### DIFF
--- a/lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
+++ b/lib/MusicBrainz/Script/RemoveUnreferencedRows.pm
@@ -91,7 +91,19 @@ sub run {
                         $id,
                     );
 
-                    if ($ref_count == 0) {
+                    if (!defined $ref_count) {
+                        # The row was deleted by some other means.
+                        # This may be normal; for artist_credit, we do delete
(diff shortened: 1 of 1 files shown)


yvanzo (commented):
Do I understand correctly that this patch replaces a meaningless warning with a meaningful info message, and that skipping `id` here is harmless?

In any case it would be helpful to clarify this by a tiny bit, LGTMBDNT otherwise.

mwiencek:
@yvanzo Correct, I updated the code comment and PR description to try and clarify things a bit more.

reosarevok (approved):
LGTMBDNT

PR #3103, opened 2023-11-18
MBS-13300: Delete RG cover art if release is moved

mwiencek:
# Problem

MBS-13300

If a release group has manually set cover art, and the release from which the cover art was taken is moved to another RG, then the leftover entry in the `release_group_cover_art` table which we fail to delete prevents the otherwise-empty RG from being deleted.

# Solution

Delete unused rows from the `release_group_cover_art` table in `Data::Release::update`.

# Testing

### DIFF
--- a/t/lib/t/MusicBrainz/Server/Edit/Release/Edit.pm
+++ b/t/lib/t/MusicBrainz/Server/Edit/Release/Edit.pm
@@ -175,6 +175,36 @@ test 'A missing comment does not clear an existing one' => sub {
     is($release->comment, 'hello', 'comment is left unchanged');
 };
 
+test 'MBS-13300: Release group cover art is deleted if the release is moved' => sub {
+    my $test = shift;
+    my $c = $test->c;
+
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
LGTM, same about tests! :ship:

t/lib/t/MusicBrainz/Server/Edit/Release/Edit.pm
```diff
@@ -175,6 +175,36 @@ test 'A missing comment does not clear an existing one' => sub {
     is($release->comment, 'hello', 'comment is left unchanged');
 };
 
+test 'MBS-13300: Release group cover art is deleted if the release is moved' => sub {
```
reosarevok, line 178:
```suggestion
test 'MBS-13300: Release group cover art is unset if the release is moved' => sub {
```
more clearly shows that we don't expect the cover art itself to be deleted, just not set for the RG.

reosarevok (approved):

PR #3104, opened 2023-11-18
MBS-13372: Unused URL with gid redirect cannot be deleted

mwiencek:
# Problem

MBS-13372

An unused URL that has a GID redirect cannot be deleted due to the foreign key reference.

# Solution

Delete URL GID redirects in `Data::URL::delete`, the same as we do for other entities.

# Testing

### DIFF
--- a/lib/MusicBrainz/Server/Data/URL.pm
+++ b/lib/MusicBrainz/Server/Data/URL.pm
@@ -314,6 +314,7 @@ sub update
 
 sub delete {
     my ($self, @ids) = @_;
+    $self->remove_gid_redirects(@ids);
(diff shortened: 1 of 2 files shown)


yvanzo (approved):
LGTM and the added test’s coverage and documentation are convincing. :ship:

reosarevok (approved):
LGTM. Will this mean the next run of the delete script will automatically delete any URLs which had remained behind due to this and are still unused?

mwiencek:
> Will this mean the next run of the delete script will automatically delete any URLs which had remained behind due to this and are still unused?

Yes, but I think this was the only URL that had issues, since the script itself doesn't crash when it encounters a failure (and it was the only URL error that I saw logged).

PR #3106, opened 2023-11-20
Set -w (page width) on msgcat and msguniq

mwiencek:
# Problem

There is a persistent issue where running `./po/update_pot.sh` on different machines will re-wrap the strings to different page widths.

# Solution

This hardcodes `-w` on both msgcat and msguniq to 79, which seemed to produce the least number of changes when I re-ran `./po/update_pot.sh`.
    
However, I'm not sure if this actually resolves the issue, because there may be other tools and variables affecting things, and I'm only able to test it on my machine.

# Testing

Ran `./po/update_pot.sh` locally.

### DIFF
--- a/po/Makefile
+++ b/po/Makefile
@@ -1,5 +1,7 @@
 # Allows passing file names with spaces in the prerequisite lists.
 SED_ESCAPE_SPACES:=sed -e 's/ /\\ /g'
+# Controls the output page width of msgcat and msguniq.
+MSGWRAP:=79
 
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
Tested running `po/update_pot.sh` and it did generate the exact same `po/mb_server.pot` but for the header `POT-Creation-Date` as expected. Other files have had no change at all since they don’t have this header anyway. :100: :ship:

reosarevok (approved):
My pot file also changed only the timestamp, so this seems good.

PR #3107, opened 2023-11-21
MBS-13376: Handle metalmusicarchives.com links

reosarevok:
# Implement MBS-13376

# Description
These are the same family of sites as jazzmusicarchives and progarchives. As with those, we support artist pages and RG/album pages.

I found out about a search result pattern we weren't cleaning up for jazzmusicarchives either so a second commit here fixes that.

# Testing
Appropriate URL tests added.


root/static/scripts/edit/URLCleanup.js
```diff
@@ -3232,7 +3232,8 @@ const CLEANUPS: CleanupEntries = {
     match: [new RegExp('^(https?://)?(www\\.)?jazzmusicarchives\\.com', 'i')],
     restrict: [LINK_TYPES.otherdatabases],
     clean: function (url) {
-      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?jazzmusicarchives\.com\/([^#]+)(?:#.*)?$/, 'https://www.jazzmusicarchives.com/$1');
+      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?jazzmusicarchives\.com\/([^#?]+)(?:[#?].*)?$/, 'https://www.jazzmusicarchives.com/$1');
```
yvanzo, line 3235: The regular expression can additionally be simplified as follows:

```suggestion
      url = url.replace(/^(?:https?:\/\/)?(?:www\.)?jazzmusicarchives\.com\/([^#?]+).*$/, 'https://www.jazzmusicarchives.com/$1');
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -3232,7 +3232,8 @@ const CLEANUPS: CleanupEntries = {
...
+      url = url.replace(/^(https:\/\/www\.jazzmusicarchives\.com\/(\w+)\/(?:[\w%()-]+\/)?[\w%()-]*)\/$/, '$1');
```
yvanzo, line 3236: What is the purpose of this line? Removing the trailing slash?

yvanzo (commented):
Comments apply to both jazz/metal websites.

root/static/scripts/edit/URLCleanup.js
reosarevok: Yes, if you have a good way to drop the (optional) trailing slash in the previous line, that'd be fine as well of course :)

yvanzo: Did not test: 
```suggestion
      url = url.replace(/\/+$/, '');
```

yvanzo (approved):
LGTMBDNT :ship:

PR #3111, opened 2023-11-27
Update mutate-cow to v5

mwiencek:
# Problem

MBS-13382

In a couple of previous commits, I had removed usages of mutate-cow:
    
 * 98d8fbbe502fc04632e051268f3f935d1c408237
 * e056bea34cb0486532486939f18b46ea864bc603
    
Although the reasoning in those commits was that mutate-cow didn't help make the code any simpler, I was in fact hesitant to continue using it elsewhere too because of its reliance on `Proxy` objects, which I consider an unsafe, leaky abstraction. (See the Autocomplete2 commit above and try to guess why `unwrapProxy` was needed there, or what would have happened if you removed it. Flow would certainly not have indicated anything was wrong!)

# Solution

The issue with removing mutate-cow completely is that in the cases where it works, it's so *useful*. So I spent some free time over the weekend releasing a new version that removes its reliance on `Proxy` objects. Unfortunately, this completely changes the API, which is now much more visible: instead of receiving a `Proxy` which you can try to pretend doesn't exist (except when it blows up in your face), you receive a "context" object which has explicit `.get` and `.set` methods to update values.
    
Even though the API is a bit more verbose, using the library is not more verbose on the whole, as this commit demonstrates: we can remove a lot of type-level code, because separate "Writable" types are no longer required. The typing situation is much more strict: mutate-cow no longer blindly trusts a "Rw" type that you pass in (which could have been anything, really), and no longer hides the fact that you're not interacting with a real object. Which makes me a lot happier and more inclined to continue using the library in MBS!

As per usual disclaimer, no cows were harmed in the making of this PR, etc.

# Testing

Only manual testing on my local dev instance so far (will see what the automated tests say here):

 * Tested registering an application through /account/applications/register
 * Tested editing gender, location, languages, and fluency via /account/edit
 * Tested adding/editing a release alias
 * Tested editing all account preferences via /account/preferences
 * Tested adding/editing all fields of an empty release collection
 * Tested editing release/tracklist artist credits (including adding/removing AC names)
 * Tested adding/editing date periods in the external links editor
 * Tested using guess case on /recording/MBID/edit, including changing various GC options
 * Tested adding/editing/removing work lyrics languages


root/static/scripts/account/components/ApplicationForm.js
```diff
@@ -57,18 +57,19 @@ class ApplicationForm extends React.Component<Props, State> {
 
   handleOauthRedirectURIChange(e: SyntheticEvent<HTMLInputElement>) {
     const selectedOauthRedirectURI = e.currentTarget.value;
-    this.setState(prevState => mutate<State, _>(prevState, newState => {
-      newState.form.field.oauth_redirect_uri.value = selectedOauthRedirectURI;
-    }));
+    this.setState(prevState => mutate(prevState)
+      .set('form', 'field', 'oauth_redirect_uri', 'value',
+           selectedOauthRedirectURI)
```
mwiencek, line 62: BTW, the property names and values are still type-checked: so if you make a typo in one of the keys, or pass the wrong type of value at the end, Flow will error.

mwiencek:
Also BTW, in case you think https://github.com/immerjs/immer (which has an impressive 26.2k stars on GitHub compared to mutate-cow's 0) is a panacea:
 * It has poor Flow types which still require defining separate Writable types (which it blindly trusts)
 * Search "proxy revoked" in its issue tracker (to its credit, it handles object spreads better)

mwiencek:
@brainzbot test this please

root/static/scripts/work/edit.js
```diff
@@ -115,15 +105,15 @@ const store = createStore<WorkForm, ActionT, (ActionT) => empty>(function (
...
+    pushField(fieldCtx, null);
+  }).final();
 }
 
 function removeLanguageFromState(form: WorkForm, i: number): WorkForm {
-  return mutate<WritableWorkForm, _>(form, newForm => {
-    newForm.field.languages.field.splice(i, 1);
-  });
+  return mutate(form).update('field', 'languages', 'field', (fieldCtx) => {
+    fieldCtx.write().splice(i);
```
reosarevok, line 115: Is dropping `, 1` here (while it's left on other `splice()` calls) intentional?

t/selenium/Work_Editor.json5
```diff
@@ -60,7 +60,7 @@
     },
     {
       command: 'select',
-      target: 'name=edit-work.languages.1',
+      target: 'name=edit-work.languages.2',
```
reosarevok, line 63: Why did this change?

reosarevok (commented):
This seems nice in general, especially that we don't need writable copies of stuff anymore. Left a few comments I'm not sure about.

root/static/scripts/account/components/ApplicationForm.js
```diff
@@ -57,18 +57,19 @@ class ApplicationForm extends React.Component<Props, State> {
...
   handleOauthTypeChangeBound: (e: SyntheticEvent<HTMLSelectElement>) => void;
 
   handleOauthTypeChange(e: SyntheticEvent<HTMLSelectElement>) {
     const selectedOauthType = e.currentTarget.value;
-    this.setState(prevState => mutate<State, _>(prevState, newState => {
-      newState.form.field.oauth_type.value = selectedOauthType;
-    }));
+    this.setState(prevState => mutate(prevState)
+      .set('form', 'field', 'oauth_type', 'value', selectedOauthType)
+      .final());
```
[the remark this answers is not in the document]
mwiencek: It returns the final copy of the object with all the changes made (and if it object is frozen it re-freezes it). It's final 'cause it ends the mutation and is the final thing you call. :) Nothing stops you from passing the result to `mutate` again (since the point is that it can modify read-only objects).

root/static/scripts/edit/utility/subfieldErrors.js
```diff
@@ -7,29 +7,79 @@
...
+  }
+}
+
+function* iterContextSubfields(
+  formOrFieldCtx: CowContext<FormOrAnyFieldT>,
+): Generator<CowContext<AnyFieldT>, void, void> {
+  const formOrField = formOrFieldCtx.read();
+  switch (formOrField.type) {
+    case 'compound_field':
+      // $FlowIgnore[incompatible-type-arg]
```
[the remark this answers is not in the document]
mwiencek: Basically, checking `formOrField.type` doesn't refine `formOrFieldCtx`'s type parameter, so Flow thinks it can still refer to, e.g., a form rather than a compound field.

root/static/scripts/work/edit.js
mwiencek: It was a bug and also the reason for the test change (which I thought changed because `last_index` was previously mutable). Good catch!

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

yvanzo:
Does this patch provide syntactic sugar only or does it provide anything else? (such as compilation/runtime checks?)

> https://github.com/immerjs/immer […] has an impressive 26.2k stars on GitHub compared to mutate-cow's 0

1 now. Maybe your repository is just missing a logo? I’m sure that @Aerozol or @MonkeyDo would make one for you.

mwiencek:
> Does this patch provide syntactic sugar only or does it provide anything else? (such as compilation/runtime checks?)

It does provide better Flow types. The previous version required you to specify a writable type in the first type parameter, but didn't actually verify that it was correct. So previously you could do this without any errors:

```js
// mutate<Rw, Ro>
mutate<{haha: string}, FieldT<any>>(field, (newField) => { newField.haha = ''; })
```

While that is an extreme example, the point is that the writable types could become out of sync with the read-only ones without the type system noticing. In v5, this is not possible, because only the input type of `field` is used.

> 1 now. Maybe your repository is just missing a logo?

Haha, most likely, but I didn't mean to suggest that it deserves stars: just that more stars != better in every situation. I only mentioned the 0 stars for comical effect.

yvanzo:
Thank you, I understand that this is a comparison of the new version 5 with the previous version 4. I was asking about comparison to not using it at all.

> > 1 now. Maybe your repository is just missing a logo?
> 
> Haha, most likely, but I didn't mean to suggest that it deserves stars: just that more stars != better in every situation. I only mentioned the 0 stars for comical effect.

I know, I hope that mine was a good 1 too :clown_face:

yvanzo:
Should the following types be reworked to use `mutate-cow` instead?
* `WritableStateT` in `root/static/scripts/annotation/AnnotationHistoryTable.js`
* `WritableReleasePathsT` in `root/static/scripts/release/components/MediumRelationshipEditor.js`
* `WritableDatePeriodRoleT` in `root/static/scripts/tests/utility/age.js`

mwiencek:
> I was asking about comparison to not using it at all.

Ah, then you are right, it doesn't provide any type safety compared to not using it at all, just adds syntactic sugar to make updating read-only types easier.

It was really useful in the artist credit editor refactor, because it allowed me to reduce some code like this:

```js
// before
const newAcns = [...ac.names];
const newAcn = {...newAcns[index]};
newAcn.name = '';
newAcns[index] = newAcn;
const newAc = {...ac, names: newAcns};

// after
const newAc = mutate(ac).set('names', index, 'name', '');
```

> Should the following types be reworked to use mutate-cow instead?
> WritableStateT in root/static/scripts/annotation/AnnotationHistoryTable.js

We could use it here, but since only single field in the state is updated, it wouldn't simplify much. The `WritableStateT` usage also isn't unsafe: `const newState: WritableStateT = {...state};`.

> WritableReleasePathsT in root/static/scripts/release/components/MediumRelationshipEditor.js

Not this one, since mutate-cow doesn't work with `Map` (only plain objects).

> WritableDatePeriodRoleT in root/static/scripts/tests/utility/age.js

It appears that `entity` is only made writable here so that it can be modified before each test. I don't think mutate-cow would provide much benefit here; if we are concerned about `entity` being writable, we might as well pass a different object to each `age.hasAge` call.

Aerozol:
I (think) it was a joke re. logo, but FYI a mutated cow logo is a brief I would take on in a heartbeat 😄

yvanzo (approved):
Thanks for the detail explanations. Merging it for a quick beta release.

PR #3112, opened 2023-11-28
MBS-13388: Handle vk.gy links 

reosarevok:
### Implement MBS-13388

# Description
This is a visual kei database.

They have the following kinds of links I've found:
* `/artists` is for groups, so artist
* `/musicians` is for people, so artist
* `/labels` is for, well, labels
* `/releases` is for... yes, releases
* `/lives` is for events
* `/livehouses` (what?) is for places
* `/songs` is for... something in between work and recording, since lives are the same "song" but covers are a new one. As such, I left it unlinkable for now.

# Testing
The usual cleanup/validation tests, plus made sure the favicon actually shows (while looking at this I noticed the recently added metalmusicarchives favicon was missing its favicon class, added in a second commit).


yvanzo (approved):
The mapping is well documented in PR/commit, the code looks good and has good coverage, you did test the display so I didn’t need to. Thanks!

It doesn’t even conflict with our current work on localizable messages. :ship:

PR #3114, opened 2023-11-29
MBS-13392 / MBS-13393: Restore the external links editor submission upon form reload, and reduce the relationship data stored in sessionStorage

mwiencek:
# Problem

MBS-13392 - External links editor changes are lost if the form reloads

MBS-13393 - Can't update relationships or external links for entities with a lot of them

# Solution

For MBS-13392, instead of using `MB.formWasPosted`, which no longer exists, we can check the request method on the Catalyst context object instead.

For MBS-13393, see the associated commit.

# Testing

Tested manually. (I temporarily commented out some lines in validation.js to make submitting invalid forms easier.) Added a Selenium test for MBS-13392.

### DIFF
--- a/root/static/scripts/edit/externalLinks.js
+++ b/root/static/scripts/edit/externalLinks.js
@@ -87,7 +87,7 @@ export type LinkStateT = $ReadOnly<{
   ...DatePeriodRoleT,
   +deleted: boolean,
   +editsPending: boolean,
-  +entity0?:
+  +entity0:
     | RelatableEntityT
     | {
         +entityType: RelatableEntityTypeT,
@@ -96,9 +96,10 @@ export type LinkStateT = $ReadOnly<{
         +name?: string,
         +orderingTypeID?: number,
         +relationships?: void,
-      },
-  +entity1?: RelatableEntityT,
-  +pendingTypes?: $ReadOnlyArray<number>,
+      }
+    | null,
+  +entity1: RelatableEntityT | null,
+  +pendingTypes: $ReadOnlyArray<number> | null,
   +rawUrl: string,
   // New relationships will use a unique string ID like "new-1".
   +relationship: StrOrNum | null,
@@ -407,7 +408,7 @@ export class _ExternalLinksEditor
     }
     this.setState(prevState => {
       let newLinks = [...prevState.links];
-      newLinks[index] = {...newLinks[index], pendingTypes: undefined};
+      newLinks[index] = {...newLinks[index], pendingTypes: null};
       const emptyLinkIndex = newLinks.findIndex(link => {
         const isNewLink = !isPositiveInteger(link.relationship);
         return isNewLink && isEmpty(link);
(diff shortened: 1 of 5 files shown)


root/static/scripts/relationship-editor/utility/prepareHtmlFormSubmission.js
```diff
@@ -210,10 +213,20 @@ export default function prepareHtmlFormSubmission(
...
+          window.sessionStorage.setItem(
+            'relationshipEditorChanges',
+            JSON.stringify(compactEntityJson(changes)),
+          );
+        } catch (error) {
+          /*
+           * Don't prevent the form from submitting if we encounter
+           * an error; for instance, "quota exceeded," which we can't do
+           * anything about. Exceeding the storage quota should be much
+           * less likely now that MBS-13393 is mitigated.
```
reosarevok, line 226: What *would* happen if this got hit still somehow, and the form was submitted? Is there a risk of us interpreting something here as a reason to remove data?

reosarevok (commented):
Looking at Selenium, seems this broke some tests :) 

> For the external links editor, we don't appear to track which specific links were changed directly.

That sounds like something we should be doing. Is there a ticket at least? IIRC you wanted to refactor the links editor since it had some questionable choices in how some of it works, is this one of them?

root/static/scripts/edit/externalLinks.js
```diff
@@ -179,6 +182,10 @@ export class _ExternalLinksEditor
           }
         }
       } else {
+        /*
+         * If the form wasn't posted, extract seeded links from the URL
+         * query parameters instead.
+         */
```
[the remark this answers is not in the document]
mwiencek: Correct, it will restore all links that were on the page.

root/static/scripts/relationship-editor/utility/prepareHtmlFormSubmission.js
mwiencek: All of the form will be submitted normally, but if there is an error and the form reloads, the submitted relationships and/or external links won't be restored. Nothing will be removed, it'll just behave as if you loaded the edit page from scratch.

mwiencek:
> That sounds like something we should be doing. Is there a ticket at least? IIRC you wanted to refactor the links editor since it had some questionable choices in how some of it works, is this one of them?

We calculate all the changes in `getEditData`, and then on the server. I just meant the change status isn't stored directly on the link state objects. I'm not sure it's needed or would help with anything at the moment, just thought I'd mention why we don't save only the changed links. If I were refactoring the code I'd probably try to make it share a bit more code with the relationship editor, yes (i.e., it can be an alternative UI on top of the main relationship editor code).

mwiencek:
> Looking at Selenium, seems this broke some tests :)

It would appear so since they are external-links-editor related, though they pass for me locally, so not sure what's going on yet. I'll try re-running them...

@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
@brainzbot test this please

mwiencek:
Test are passing now. The problem was `submittedLinks` leaking between test runs, and the release editor trying to restore from it (it shouldn't).

yvanzo (approved):
Thank you! The code changes and the added test look good, the commit messages are very explanatory. Didn’t test manually as you did it already. :ship: 

Added the original reporter to ticket watchers for beta testing.

:ship:

PR #3119, opened 2023-12-07
webpack/exec.mjs fixes

mwiencek:
# Problems

1. The relative path in webpack/exec.mjs's shebang does not work on macOS:
   ```
   $ ./webpack/exec.mjs root/static/scripts/tests/i18n/expand2.js
   ./webpack/exec.mjs: line 2: /Applications: is a directory
   ./webpack/exec.mjs: line 3: syntax error near unexpected token `('
   ./webpack/exec.mjs: line 3: ` * Copyright (C) 2023 MetaBrainz Foundation'
   ```
 2. If the input script exits with a non-zero exit code, we throw an error without
logging any of the stdout/stderr that the script logged while running. This
means if you run a test file, you can't currently see which tests failed, which
isn't very useful.

# Solutions

 1. Adds a wrapper script, webpack/exec, which execs webpack/exec.mjs for us.
 2. Fixes the logging to come before we throw an error.


mwiencek:
Cancelled the Selenium tests since this doesn't affect anything there.

PR #3122, opened 2023-12-13
Fix some recently updated messages

yvanzo:
# Problem

When reviewing checks in Weblate, I found those (mostly recent) messages with issues.

# Solution

Only address messages that have been updated recently (or that already exist).

Additionally fixes three strings reported by Salo.Rock through Weblate.

# Note

Expected to be merged in `master`, then `beta` today.

### DIFF
--- a/root/cdtoc/lookup.tt
+++ b/root/cdtoc/lookup.tt
@@ -20,7 +20,7 @@
   [% IF cdstub %]
   <h2>[% l('CD stub found') %]</h2>
   <p>[% l('A CD stub was found that matches the disc ID you provided. If the below tracklist
-           appears correct, you may use it as a starting point as a new MusicBrainz release') %]</p>
+           appears correct, you may use it as a starting point as a new MusicBrainz release.') %]</p>
   <h3>[% l('{artist} - {name}', { artist => cdstub.artist,
                                   name => cdstub.title }) %]</h3>
   [% INCLUDE 'cdstub/cdstub.tt' %]
--- a/root/cdtoc/lookup.tt
+++ b/root/cdtoc/lookup.tt
@@ -20,7 +20,7 @@
   [% IF cdstub %]
   <h2>[% l('CD stub found') %]</h2>
   <p>[% l('A CD stub was found that matches the disc ID you provided. If the below tracklist
-           appears correct, you may use it as a starting point as a new MusicBrainz release.') %]</p>
+           appears correct, you may use it as a starting point for a new MusicBrainz release.') %]</p>
   <h3>[% l('{artist} - {name}', { artist => cdstub.artist,
                                   name => cdstub.title }) %]</h3>
   [% INCLUDE 'cdstub/cdstub.tt' %]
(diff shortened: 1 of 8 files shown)


root/static/scripts/account/components/PreferencesForm.js
```diff
@@ -189,7 +189,7 @@ class PreferencesForm extends React.Component<Props, State> {
         <fieldset>
           <legend>{l('Privacy')}</legend>
           <p>
-            {l('Allow other users to see:')}
+            {addColonText(l('Allow other users to see'))}
```
reosarevok, line 192: Should we always be using `addColonText` then even if there's not (and we don't expect there will be) a non-colon equivalent string? I'm fine with that, just wasn't sure that was a thing :)

root/cdtoc/lookup.tt
```diff
@@ -20,7 +20,7 @@
   [% IF cdstub %]
   <h2>[% l('CD stub found') %]</h2>
   <p>[% l('A CD stub was found that matches the disc ID you provided. If the below tracklist
-           appears correct, you may use it as a starting point as a new MusicBrainz release') %]</p>
+           appears correct, you may use it as a starting point as a new MusicBrainz release.') %]</p>
```
reosarevok, line 23: Hmm. Do you think it would be better "you may use it as a starting point `for` a new MusicBrainz release" if we're already changing the string?

reosarevok (approved):

root/static/scripts/account/components/ApplicationForm.js
```diff
@@ -93,7 +93,7 @@ class ApplicationForm extends React.Component<Props, State> {
         />
         <FormRowURLLong
           field={this.state.form.field.oauth_redirect_uri}
-          label={addColonText(l('Callback URL'))}
+          label={addColonText(l('Callback URI'))}
```
reosarevok, line 96: Not sure what this unifies with, we don't seem to have a `Callback URI` string at the moment (you mean the longer doc string below?) but it seems sensible, anyway.

root/static/scripts/account/components/PreferencesForm.js
yvanzo: Yes, even though it is probably not worth breaking all the existing translations for long-standing source strings at the moment.

root/cdtoc/lookup.tt
yvanzo: Yes, making an actionable suggestion:

```suggestion
           appears correct, you may use it as a starting point for a new MusicBrainz release.') %]</p>
```

root/static/scripts/account/components/ApplicationForm.js
yvanzo: Yes, it unifies with https://github.com/metabrainz/musicbrainz-server/blob/73c126c3ccf85b571f735e00a0fead613482bbd1/root/static/scripts/account/components/ApplicationForm.js#L104

It is technically both an URI and an URL, but we use URI in the code, and this is a tech field anyway.

PR #3123, opened 2023-12-13
MBS-13404: Avoid duplicating feat. artists

derat:
# MBS-13404

Fix a bug introduced by 2d82124e where tracklist artist credits could be duplicated when changing release artists.

When changing release credits to "A & B" while a track was already credited to "A feat. B", the track credits would be changed to "A & B feat. B". This change sets track credits to the new release credits (i.e. "A & B") if they already contain the same artists as the new release credits.

# Problem

As described at [this comment](https://tickets.metabrainz.org/browse/MBS-13273?focusedCommentId=70393&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70393), 2d82124eb732026305d74cb0074888ee460c4aa7 could result in secondary artists getting duplicated in cases where the release and track artists didn't match.

Specifically, the change was (by design) updating tracklist credits in cases where they had a prefix matching the old release credits. If the release credited "A" and the track credited "A feat. B", then changing the release credit to "C" would update the track credit to "C feat. B". The problem arose if the release was instead updated to credit "A & B": the track would end up being credited to "A & B feat. B" in that case due to the matching "A" prefix.

# Solution

This change adds an additional "do the track credits and new release credits include the same artists?" condition for setting the track credits to the new release credits. In the previous example, since "A feat. B" and "A & B" both include A and B, the track credits will be set to "A & B", which is presumably the editor's intent in this case.

It also updates the shared-prefix case to avoid updating track credits if they start with the same artists as the new release credits. When the track credits are "A & B feat. C" and the release credits are changed from "A" to "A & B", this prevents erroneous "A & B & B feat. C" track credits.

# Action

I did some manual testing by replicating the situation described in the ticket: I created a release credited to "A" with a track credited to "A & B", and checked that when I changed the release credit to "A feat. B" the track credit was updated to "A feat. B" rather than "A feat. B & B". I also checked that tracks credited to "A & B feat. C" would be left alone in this case rather than being updated to "A feat. B & B feat. C". More manual testing would be appreciated!

I also updated the existing Selenium test for MBS-13273 to include a track already credited to the new release artists to verify that it would be updated as expected. This isn't exactly the same as the situation described in the ticket (the test would probably need to create another release for that), but it should at least be exercising the new code path.

Please let me know if this seems to be getting too complicated. :-/

### DIFF
--- a/root/static/scripts/release-editor/init.js
+++ b/root/static/scripts/release-editor/init.js
@@ -16,6 +16,7 @@ import {
   reduceArtistCreditNames,
 } from '../common/immutable-entities.js';
 import MB from '../common/MB.js';
+import {arraysEqual} from '../common/utility/arrays.js';
 import {getSourceEntityData} from '../common/utility/catalyst.js';
 import clean from '../common/utility/clean.js';
 import {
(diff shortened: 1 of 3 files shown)


reosarevok:

reosarevok (approved):
LGTMBDNT

yvanzo:

derat:
Thanks for the quick review and merge, and sorry for missing this in the original patch!

PR #3127, opened 2023-12-16
MBS-13405: Also infer rel dirs when adding artists

derat:
# MBS-13405

When adding a new artist, infer the correct direction for an individual-to-group relationship from the target artist's type.

# Problem

54f9433c39c73c4390fcca9c525789db4dc21df1 made the "Edit artist" form infer the correct direction for individual-to-group artist relationships (e.g. member-of/members) based on the types (e.g. Person, Group, etc.) of the source and target entities. This didn't work when adding a new artist, since the source type is null.

# Solution

Even if the source type is unknown, it's possible to infer the correct direction based on the target entity's type. For example, a "member of" relationship is probably correct if the target is a group, and a "members" relationship is probably correct if the target is a person.

# Action

I did some manual testing, but more is always appreciated.

I also updated the existing Selenium test that I added for MBS-2604 to exercise the new behavior.

### DIFF
--- a/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
+++ b/root/static/scripts/relationship-editor/components/RelationshipDialogContent.js
@@ -225,6 +225,7 @@ function inferLinkDirection(
     linkTypeId !== null &&
     personGroupLinkTypeIds.has(linkTypeId)
   ) {
+    const isSourceUnset = source.typeID === null;
     const isSourcePerson = source.typeID === ARTIST_TYPE_PERSON;
     const isSourceGroup =
       source.typeID !== null && ARTIST_GROUP_TYPES.has(source.typeID);
@@ -233,10 +234,13 @@ function inferLinkDirection(
     const isTargetGroup =
       target.typeID !== null && ARTIST_GROUP_TYPES.has(target.typeID);
 
-
-    if (isSourcePerson && isTargetGroup) {
+    /*
+     * The source's type will be unset if the entity hasn't been added yet,
+     * so make an inference based on the target's type in that case.
+     */
+    if ((isSourcePerson || isSourceUnset) && isTargetGroup) {
       newState.backward = false;
-    } else if (isSourceGroup && isTargetPerson) {
+    } else if ((isSourceGroup || isSourceUnset) && isTargetPerson) {
       newState.backward = true;
     }
   }
--- a/t/selenium/Artist_Edit_Form.json5
+++ b/t/selenium/Artist_Edit_Form.json5
@@ -127,11 +127,6 @@
     },
     {
       command: 'click',
-      target: 'css=#add-relationship-dialog button.change-direction',
-      value: '',
-    },
-    {
-      command: 'click',
       target: 'css=#add-relationship-dialog button.positive',
       value: '',
     },
--- a/t/selenium/MBS-2604.json5
+++ b/t/selenium/MBS-2604.json5
@@ -55,5 +55,66 @@
       target: 'css=#add-relationship-dialog table.preview td',
       value: 'David Bowie is/was a member of Nine Inch Nails',
     },
+    // Start adding a new artist and check that relationship directions
+    // are inferred based on the target artist's type (see MBS-13405).
+    {
+      command: 'open',
+      target: '/artist/create',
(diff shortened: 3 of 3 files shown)


derat:
Just to mention it, there are some rare cases where the correct direction can't be inferred (e.g. when one group is being added as a member of another group). This change may reverse the default direction in some of those situations, but it seems like it's already essentially 50-50 whether the default is correct or not there, so I suspect that this won't make things worse.

I haven't looked into whether it'd be feasible to get the new source artist's type from the in-progress form (I think it seemed like it might be hard when I was working on #3011), but that may be another possibility. It would only help if the editor happened to select the type before adding the relationships, though.

derat:
@yvanzo, mind taking a look since you previously reviewed #3011? Thanks!

derat:

derat:

(My updated test passed in the last run, so it look like this is just flakiness.)

derat:
I think that the `Artist_Edit_Form` test was clicking the "change direction" button to work around the behavior that this change is trying to improve. :-P

reosarevok:
> I haven't looked into whether it'd be feasible to get the new source artist's type from the in-progress form (I think it seemed like it might be hard when I was working on https://github.com/metabrainz/musicbrainz-server/pull/3011), but that may be another possibility.

This should be doable at least once the form is converted to React - probably doable also now, just less trivial.

> It would only help if the editor happened to select the type before adding the relationships, though.

Given the way the form is structured, this should be the order these things happen in most of the time.

yvanzo (approved):
Thank you, the changes look code, and I tested it locally before finding out that your added Selenium tests covered it already.

yvanzo:
> > I haven't looked into whether it'd be feasible to get the new source artist's type from the in-progress form (I think it seemed like it might be hard when I was working on #3011), but that may be another possibility.
> 
> This should be doable at least once the form is converted to React - probably doable also now, just less trivial.

If it isn’t trivial now, better wait for the React conversion.

reosarevok:
> If it isn’t trivial now, better wait for the React conversion

I agree, but then we should make a ticket and list the artist form conversion as a dependency.

mwiencek:
> I haven't looked into whether it'd be feasible to get the new source artist's type from the in-progress form (I think it seemed like it might be hard when I was working on https://github.com/metabrainz/musicbrainz-server/pull/3011), but that may be another possibility.

Have you tried the "hack" of using querySelector to get the select `.value`? I think the selector should be stable.

derat:
> Have you tried the "hack" of using querySelector to get the select `.value`? I think the selector should be stable.

I think that I'd need to change `inferLinkDirection` to take a `sourceTypeID` parameter to use if `source.typeID` is null, and then somehow get the select value into the `reducer` function in RelationshipDialogContent.js.  Does that sound right?

I'm not very familiar with the relationship code (I remember struggling with it for #3011), but I suspect that doing that may require some ugly plumbing changes that I'd prefer to avoid, assuming that just looking at the target type is enough to choose the right default in most cases.

mwiencek:
> I think that I'd need to change inferLinkDirection to take a sourceTypeID parameter to use if source.typeID is null, and then somehow get the select value into the reducer function in RelationshipDialogContent.js. Does that sound right?

I was thinking you could get the select value directly in `inferLinkDirection`. Being a hack either way, I'm not sure there's any advantage to getting it in the reducer and passing it as a parameter. Or am I missing a reason it must be done in the reducer?

> I'm not very familiar with the relationship code (I remember struggling with it for https://github.com/metabrainz/musicbrainz-server/pull/3011), but I suspect that doing that may require some ugly plumbing changes that I'd prefer to avoid, assuming that just looking at the target type is enough to choose the right default in most cases.

I'm happy with the current changes if you'd prefer not to muck around with this code anymore. :)

derat:
> I'm happy with the current changes if you'd prefer not to muck around with this code anymore. :)

Thanks, I'd prefer for it to go in like this. `document.getElementById('id-edit-artist.type_id').value` is easy enough, but reaching out into the DOM from `inferLinkDirection` feels fragile and unexpected to me (and I'd rather not have the `git blame` evidence pointing back to me). :-P

mwiencek (approved):

PR #3129, opened 2023-12-18
Make even more admin strings untranslatable

reosarevok:


root/relationship/linktype/RelationshipTypeIndex.js
```diff
@@ -132,13 +132,13 @@ const RelationshipTypeIndex = ({
         {isRelationshipEditor($c.user) ? (
           <span className="buttons" style={{float: 'right'}}>
             <EntityLink
-              content={lp('Edit', 'verb, interactive')}
+              content={l_admin('Edit')}
```
mwiencek, line 135: I don't see `l_admin` imported in this file -- does that work?

mwiencek: I asked because some files do import it, and there's no entry for it in webpack/providePluginConfig.mjs.

However, it's declared as a global in root/vars.js, so Flow may think it's available when it's not.

If it must be imported, that declaration should be removed from root/vars.js.

reosarevok: I was tricked!

I don't see a reason why we would *not* want this to be a global, so rather than removing from `root/vars.js` I added it properly to `webpack/providePluginConfig.mjs`.

mwiencek (approved):

root/relationship/linktype/RelationshipTypeIndex.js
yvanzo: One reason would be that we rarely need it, but I have no idea of the impact of importing it globally.

mwiencek: It shouldn't have any impact (other than convenience). Webpack injects the imports as if you wrote them by hand.

mwiencek: Example I pulled from common-chunks...
```JS
/***/ "./root/layout/components/sidebar/SidebarProperties.js":
/*!*************************************************************!*\
  !*** ./root/layout/components/sidebar/SidebarProperties.js ***!
  \*************************************************************/
/***/ ((__unused_webpack_module, __webpack_exports__, __webpack_require__) => {

"use strict";
__webpack_require__.r(__webpack_exports__);
/* harmony export */ __webpack_require__.d(__webpack_exports__, {
/* harmony export */   "SidebarProperties": () => (/* binding */ SidebarProperties),
/* harmony export */   "SidebarProperty": () => (/* binding */ SidebarProperty)
/* harmony export */ });
/* harmony import */ var react_jsx_runtime__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(/*! react/jsx-runtime */ "./node_modules/react/jsx-runtime.js");
/* provided dependency */ var nonEmpty = __webpack_require__(/*! ./root/static/scripts/common/utility/nonEmpty */ "./root/static/scripts/common/utility/nonEmpty.js")["default"];
```

The `nonEmpty` import is commented with `/* provided dependency */`, but otherwise looks like any other import.

yvanzo: Thank you for checking it in detail!

PR #3131, opened 2023-12-22
Improve AC bubble wording

reosarevok:
# Description
Artist credits are not only for collaborations, so the current help is misleading rather than useful. @Aerozol proposed adding "name variations" which I guess would now cover most cases at least.

# Testing
None, but this is just a string change.

### DIFF
--- a/root/static/scripts/edit/components/ArtistCreditBubble.js
+++ b/root/static/scripts/edit/components/ArtistCreditBubble.js
@@ -57,7 +57,8 @@ const ArtistCreditBubble = ({
         <tr>
           <td colSpan="3" style={{paddingBottom: '1em'}}>
             {exp.l(
-              `Use the following fields to enter collaborations. See the
+              `Use the following fields to enter artist name variations
+               and multiple artist collaborations. See the
                {ac|Artist Credit} documentation for more information.`,
               {ac: '/doc/Artist_Credits'},
             )}


root/static/scripts/edit/components/ArtistCreditBubble.js
```diff
@@ -57,7 +57,8 @@ const ArtistCreditBubble = ({
         <tr>
           <td colSpan="3" style={{paddingBottom: '1em'}}>
             {exp.l(
-              `Use the following fields to enter collaborations. See the
+              `Use the following fields to enter artist name variations
+               and multiple artist collaborations. See the
```
mwiencek, line 61: I feel like that should be hyphenated so as not to be confused with multiple artist-collaborations. :)
```suggestion
               and multiple-artist collaborations. See the
```

mwiencek (approved):

root/static/scripts/edit/components/ArtistCreditBubble.js
reosarevok: I'm going to merge as is for now since I'm unconvinced by this, but if @Aerozol likes the suggestion I can push it later.

Aerozol: I don't think it makes it harder to read, so if it clarifies something it's probably an improvement... but I take reo's point from IRC that we would then have to change a _lot_ of strings (I haven't checked, but I assume so) to be consistent!

PR #3132, opened 2023-12-26
MBS-13415, MBS-13419: Add sidebar + autoselect & validation for OTOTOY URLs

julian45:
# Problem

Not a problem per se, but this implements [MBS-13415](https://tickets.metabrainz.org/browse/MBS-13415), which adds a sidebar spot for OTOTOY, a Japanese digital download service with helpful metadata and often high-resolution cover art.

# Solution

This seems a bit more straightforward than [my previous PR](https://github.com/metabrainz/musicbrainz-server/pull/2807) since OTOTOY URLs are already being validated and sorted into the right relationship types. I could, of course, be wrong.


reosarevok (approved):
This seems good :)

For what it's worth, OTOTOY links are not being validated in any way, they just have a very old implementation that accepts any OTOTOY link as a "purchase for download" link without further checks. We should probably improve it (for example by validating each type of link for artist/label/release as separated in MBS-12991) and if you feel like it, that'd be sweet, but it'd also be fine to just merge this as is and look into it later. I added MBS-13419 to track it, anyway.

julian45:
> We should probably improve it (for example by validating each type of link for artist/label/release as separated in MBS-12991) and if you feel like it, that'd be sweet, but it'd also be fine to just merge this as is and look into it later. I added MBS-13419 to track it, anyway.

I agree that we should be validating these link types. I'd be happy to work on that and will add some commits to this PR.

julian45:
I _think_ I handled the validation correctly in this commit, but I'm not sure — I don't often work with JavaScript, after all. 😅 

The regex in the `const m =` line seemed to be technically correct when I tested it in regex101, but the search groups I used may have resulted in some dicey checks in the `switch` statement below it.

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4405,6 +4405,16 @@ limited_link_type_combinations: [
              input_entity_type: 'release',
     expected_relationship_type: 'downloadpurchase',
   },
```
yvanzo, line 4407: Following https://github.com/metabrainz/musicbrainz-server/pull/3132/files#r1444824078, here is an applicable suggestion to test validation in the already existing test for release from OTOTOY.

```suggestion
       only_valid_entity_types: ['release'],
  },
```

root/static/scripts/tests/Control/URLCleanup.js
```diff
@@ -4405,6 +4405,16 @@ limited_link_type_combinations: [
...
+  {
+                     input_url: 'https://ototoy.jp/_/default/a/1064270',
+             input_entity_type: 'artist',
+    expected_relationship_type: 'downloadpurchase',
+  },
+  {
+                     input_url: 'https://ototoy.jp/labels/695414',
+             input_entity_type: 'label',
+    expected_relationship_type: 'downloadpurchase',
+  },
```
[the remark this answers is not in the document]
yvanzo: Same thing, as an applicable suggestion:

```suggestion
  {
                     input_url: 'https://ototoy.jp/_/default/a/1064270',
             input_entity_type: 'artist',
    expected_relationship_type: 'downloadpurchase',
       only_valid_entity_types: ['artist'],
  },
  {
                     input_url: 'https://ototoy.jp/labels/695414',
             input_entity_type: 'label',
    expected_relationship_type: 'downloadpurchase',
       only_valid_entity_types: ['label'],
  },
```

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4544,6 +4543,35 @@ const CLEANUPS: CleanupEntries = {
     ],
     restrict: [LINK_TYPES.otherdatabases],
   },
+  'ototoy': {
+    match: new RegExp('^(https?://)?([^/]+\\.)?ototoy\\.jp', 'i'),
+    restrict: [LINK_TYPES.downloadpurchase],
+    validate: function (url, id) {
+      const m = /^https:\/\/ototoy\.jp\/(?:(labels)\/\d+|(_\/default\/p)\/\d+|(_\/default\/a)\/\d+)$/.exec(url);
+      if (m) {
+        const suffix = m[1] || m[2] || m[3];
```
yvanzo, line 4552: It is a very good start Two points:

1. Major: It assumes that the URL is clean but there is no `clean` function. It may block entering some URLs that would just need a small cleanup. For example, see the rule `anidb`.

2. Minor: It can be simplified into:

```suggestion
      const m = /^https:\/\/ototoy\.jp\/(labels|_\/default\/[ap])\/\d+$/.exec(url);
      if (m) {
        const suffix = m[1];
```

yvanzo (changes requested):
Thank you for working on the validation. It goes into the right direction but still has a couple of issues, minor compared to what you‘ve done so far.

root/static/scripts/edit/URLCleanup.js
julian45: For the major point, I added a `clean` function in 4e2505f40e. I did some quick spot-checks in a JS console with it and it seemed like the function returned proper URLS, so hopefully we're good to go there.

For the minor point, I committed your suggestion as c094be0. Thanks!

reosarevok:
You should actually test your `clean` code too then :) Use `expected_clean_url` for that (and make some changes to the original URLs to make sure they get cleaned up as expected, such as using `http` to see if it gets changed to `https`, and something like `https://ototoy.jp/_/default/a/1064270#discography` to see if the `#discography` bit goes away).

root/static/scripts/edit/URLCleanup.js
```diff
@@ -4544,6 +4543,38 @@ const CLEANUPS: CleanupEntries = {
     ],
     restrict: [LINK_TYPES.otherdatabases],
   },
+  'ototoy': {
+    match: new RegExp('^(https?://)?([^/]+\\.)?ototoy\\.jp', 'i'),
+    restrict: [LINK_TYPES.downloadpurchase],
+    clean: function (url) {
+      return url.replace(/^(?:https:\/\/)?(?:www\.)?ototoy\.jp\/(labels|_\/default\/[ap])\/(\d+).*$/, 'https://ototoy.jp/$1/$2');
```
reosarevok, line 4550: You allow for `http` in `match` but do not actually clean up non-https links here, it seems.

yvanzo: If I understand correctly, this is just about the missing `?` after `https`:

```suggestion
      return url.replace(/^(?:https?:\/\/)?(?:www\.)?ototoy\.jp\/(labels|_\/default\/[ap])\/(\d+).*$/, 'https://ototoy.jp/$1/$2');
```

(Compare to the regular expression in `match`.)

julian45:
> You should actually test your `clean` code too then :) Use `expected_clean_url` for that (and make some changes to the original URLs to make sure they get cleaned up as expected, such as using `http` to see if it gets changed to `https`, and something like `https://ototoy.jp/_/default/a/1064270#discography` to see if the `#discography` bit goes away).

Done in c268bc9ec1.

yvanzo (approved):
Thanks a lot for your dedication! Both code and tests look good to me.

It happened very often that external websites change their presentation, their URL patterns, or that some other websites or software present these URLs in a way that break our URL handler. That’s one reason why we now systematically handle schemes, protocols, sub-paths, queries, and fragments, when possible: to prevent breakage.

reosarevok:
>  I've somehow managed not to find OTOTOY links in the wild like your example

That's entirely fair - I tried adding it to the URL by hand to see if it worked, which it did, but I don't think it's currently a link they offer directly :) But as @yvanzo said, it's good to be prepared, and since we generally clean up those things, we can test it here.

Thanks!

reosarevok (approved):

PR #3133, opened 2023-12-28
MBS-13413: Allow using "Add/Remove note" with keyboard

reosarevok:
### Fix MBS-13413

# Problem
In edit summaries, the "Add note" / "Remove note" toggle button is not reachable with keyboard navigation (it's completely skipped when tabbing).

# Solution
Since this is an `<a>` without href, it was being skipped by the tabindex, but we want it to be tabbable. There seems to be no reason why this should not be `<button>` instead, which makes it tabbable by default. It also means it responds to Enter and Space keypresses, which is nice.

# Testing
Manually, by tabbing around a list of open edits and ensuring the button is reachable and pressing Enter opens/closes the note field.

### DIFF
--- a/root/edit/components/EditSummary.js
+++ b/root/edit/components/EditSummary.js
@@ -56,9 +56,9 @@ const EditSummary = ({
       ) ? (
         <div className="cancel-edit buttons">
           {mayAddNote ? (
-            <a className="positive edit-note-toggle">
+            <button className="positive edit-note-toggle" type="button">
               {lp('Add note', 'interactive')}
-            </a>
+            </button>
           ) : null}
 
           {mayApprove ? (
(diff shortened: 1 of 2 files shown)


mwiencek:
Makes sense. But we should really be using...an actual button, which would be focusable by default, should not require any event handler, and would also work with space bar. I really have no idea why we're using an `<a>` here.

reosarevok:
Hmm, ok. Given all the other "buttons" on the `EditSummary` and `EditHeader` components seem to be `<a>`, I assumed - perhaps wrongly - that there was a reason for all of this. Should I drop this commit and change all of those to be buttons instead?

mwiencek:
If none of them are links then let's do that, though you may need to tweak the CSS so that it applies to button elements too.

reosarevok:
No, they are links, to things like `/edit/${edit.id}/cancel?${returnToCurrentPage($c)}`. Do you think in that case we should combine `button` and `a` elements here with the same style?

mwiencek:
OK, then the links should stay links, but I think add/remove note should still be a button. And yes, we can have the same styling for both.

mwiencek (approved):
Approving assuming that you've tested the new changes, because I haven't. :)

PR #3134, opened 2023-12-28
MBS-13414: Normalize x.com URLs to twitter.com

reosarevok:
### Implement MBS-13414

# Description
x.com URLs are simply redirects to the equivalent twitter.com URL, so we might as well clean them up.

# Testing
Added one test


mwiencek (approved):

yvanzo:
@reosarevok: There currently are 50 `x.com` URLs in the database so these should be converted by hand.

reosarevok:
I changed them now - I think that was all of them anyway.

PR #3135, opened 2023-12-29
MBS-13420: Support Bandcamp /discover tag links

reosarevok:
### Implement MBS-13420

# Problem
Bandcamp `/tag` links now redirect to a revamped `/discover` page, links which we don't currently support.

# Solution
This cleans up `/tag` links and drops unwanted extras from `/discover` ones (see test).

# Testing
Added one test for `/discover`, changed the existing `/tag` one to check the redirect.


mwiencek (approved):

yvanzo:
@reosarevok, did you check how many Bandcamp `/tag` links currently are in the database to see if it is worth creating an MBBE ticket for it?

reosarevok:
This actually broke a Selenium test, which we missed. Submitted a fix with https://github.com/metabrainz/musicbrainz-server/pull/3138

PR #3136, opened 2023-12-29
MBS-13327: Don't unlink recording from track when removing feat. artists

mwiencek:
# Problem

MBS-13327

# Solution

When deciding whether to unlink a recording, compare track names with featured artists removed.

Since `extractFeatCredits` is a heavy function to run each time a track name is modified (and is useless except when the title contains featured artists), I added `featQuickTestRegex` so that it can bail out early if it's sure there will be no matches.

# Testing

Tested locally with http://localhost:5000/release/f04eef04-06b7-498c-ae69-9a41e4d1686c/edit (on master, it unlinks the recordings).

Also added some tests for `similarTrackNames` and `watchTrackForChanges`.

### DIFF
--- a/root/static/scripts/edit/utility/guessFeat.js
+++ b/root/static/scripts/edit/utility/guessFeat.js
@@ -30,6 +30,12 @@ import getSimilarity from './similarity.js';
 
 /* eslint-disable sort-keys */
 const featRegex = /(?:^\s*|[,，－\-]\s*|\s+)((?:ft|feat|ｆｔ|ｆｅａｔ)(?:[.．]|(?=\s))|(?:featuring|ｆｅａｔｕｒｉｎｇ)(?=\s))\s*/i;
+/*
+ * `featQuickTestRegex` is used to quickly test whether a title *might*
+ * contain featured artists. It's fine if it returns false-positives.
+ * Please keep it in sync with `featRegex` above.
+ */
+const featQuickTestRegex = /ft|feat|ｆｔ|ｆｅａｔ/i;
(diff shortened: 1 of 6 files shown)


root/static/scripts/edit/utility/guessFeat.js
```diff
@@ -114,9 +120,13 @@ function extractBracketedFeatCredits(str, artists, isProbablyClassical) {
   }, {name: str, joinPhrase: '', artistCredit: []});
 }
 
-function extractFeatCredits(
+export function extractFeatCredits(
   str, artists, isProbablyClassical, allowEmptyName,
 ) {
+  if (!featQuickTestRegex.test(str)) {
```
mwiencek, line 126: I benchmarked `utils.similarTrackNames('Brooklyn Rumba', 'Brooklyn Rumba (feat. Dr. Israel & Marc Ribot)');` with and without this and it was only about 15% faster, so I don't think it would be noticeable when typing -- doesn't hurt, but can probably be removed.

reosarevok: What is the point of having a "quick test"? Like, if it works quicker and is good shouldn't we always use it, and if it gives worse results wouldn't it be problematic anyway?

reosarevok (approved):
Tested and works fine - also not sure we need a "fast match" though, it seems more problematic than useful?

root/static/scripts/edit/utility/guessFeat.js
reosarevok: Hmm, ok, I see. Well, I'm fine with leaving it in then if you feel the 15% improvement is significant enough - if not, I'd probably just use the old regex to limit complexity.

yvanzo (approved):
The optimization’s gain seems to be small but the complexity increase is small too. When this code will be running on tens of tracks at once, the cumulative gain may become noticeable. But I don’t mind either way, feel free to remove if you prefer. :ship:

PR #3138, opened 2024-01-04
Fix Selenium test broken by MBS-13420

reosarevok:
I missed this test when implementing https://github.com/metabrainz/musicbrainz-server/pull/3135


yvanzo (approved):
:speedboat:

PR #3139, opened 2024-01-04
MBS-13431: Fix using translations for expanded strings in statistics

yvanzo:
# Problem MBS-13431

Text that needs to be expanded - because it contains either variables or HTML tags - is incidentally not translated when rendering statistics pages.

This is because it was using `exp.l` or `texp.l` which match the translation domain `server` instead of `statistics`.

From a discussion on IRC, I found some additional issues with rendering translations from the statistics domain.

# Solution

Add `l_statistics` and `ln_statistics` to `exp` and `texp`, auto-import these new functions, parse these for string extraction, and use these instead of `l` and `ln` under `root/statistics/`.

Make use of it for rendering event descriptions.

Fix rendering translations in MusicBrainz history and statistics timeline.

# Testing

* [x] Using sample data (that has no statistics), open the page http://localhost:5000/statistics and check that the main paragraph is actually translated (in French, and probably in other production languages too).
* [x] Tested running `po/update_pot.sh`
* [x] Copying an event from the main database, open the page http://localhost:5000/history and check that the strings are actually translated (in Italian, which has the best coverage).
* [x] Tested Flow and ESLint on modified JS files.

### DIFF
--- a/root/static/scripts/common/i18n/expand2react.js
+++ b/root/static/scripts/common/i18n/expand2react.js
@@ -19,6 +19,10 @@ import {
   l_admin as lAdminActual,
   ln_admin as lnAdminActual,
 } from '../i18n/admin.js';
+import {
+  l_statistics as lStatisticsActual,
+  ln_statistics as lnStatisticsActual,
+} from '../i18n/statistics.js';
 
 import expand, {
   type NO_MATCH,
@@ -420,6 +424,11 @@ export const l_admin = (
   args?: ?VarArgsObject<Input>,
 ): Output => expand2react(lAdminActual(key), args);
 
+export const l_statistics = (
+  key: string,
+  args?: ?VarArgsObject<Input>,
+): Output => expand2react(lStatisticsActual(key), args);
+
 export const ln = (
   skey: string,
   pkey: string,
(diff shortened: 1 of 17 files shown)


reosarevok (approved):
Happy with this given you've tested it and it works, but I'd prefer we then drop the whole `import l_statistics as l` concept and just use the full `l_statistics` name elsewhere in these files, as it feels confusing rather than helpful right now. Happy to change that in a subsequent PR though.

yvanzo:
@reosarevok: I should have turned it into a draft when I started to work on this branch again after we briefly discussed it on IRC. I just pushed new commits and updated the PR description, please review these too.

I agree that the current aliased imports are not ideal, but I don’t see a better solution at the moment. The benefit is that it catches new `l` calls that would be unnoticed otherwise. I also agree on leaving this question out of this pull request.

reosarevok:
> I don’t see a better solution at the moment

What would be the problem with just not importing anything as `l` and just using `l_statistics` everywhere as it should be? It should then be obvious that any `l` use case in the stats files is wrong anyway, so it shouldn't be hard to catch them :) Fine to wait with that anyway, just not sure what I'm missing.

yvanzo:
Rebased on `master`

yvanzo:
Squashed amending commits.

yvanzo:
The problem is not having `l` or `l_statistics`. It is not having a mix of those which implies bugs.

I will try a patch with a CI test, but this PR is good to go already.

reosarevok:
Sure, but if we don't import `l` as `l_statistics`, then no use of `l` should ever be correct in the statistics files, right? So it should be trivial to find issues when reviewing them, but in any case CI won't hurt :) Thanks!

PR #3140, opened 2024-01-07
Fix deleted entity spacing in subscription email

derat:
Subscription emails are missing a space between names and disambiguation comments of deleted entities and contain an extra space after comments, e.g.

`Artist "Name"(comment)  - merged by edit #...`

Fix the spacing.

---

I didn't bother with filing an issue or filling out the PR template since this is a trivial change, but let me know if I should've.

`t/lib/t/MusicBrainz/Server/Email/Subscriptions.pm` seems like it didn't catch this since it only looks at a deleted collection.


yvanzo (approved):
Same, thank you!

PR #3142, opened 2024-01-09
Don't store incremental JSON dumps in a local volume

mwiencek:
# Problem

Currently, when incremental JSON dumps are generated, they are (1) copied to a local "FTP directory" (which, confusingly, is not our actual FTP server; it's just a local volume), and (2) rsynced to a (possibly) remote volume on the same host as the metabrainz.org container.

There is not really any reason to do (1) if (2) is succesful, because we presumably backup (2) already with borgmatic; there is no need to store these in three places.

This is more relevant now, because the incremental JSON dumps are being moved to hendrix, which is also where the metabrainz.org container is hosted. So without these changes, we'd be storing them twice on the same host!

# Solution

This updates admin/RunIncrementalJSONDump to NOT copy any dumps from the temporary dump directory to any local volume, only rsync the dumps directly from the temporary directory. This means we can no longer use rsync-fullexport-files, because that runs rsync with the `--delete` flag, and we do NOT want to delete any "remote" (but really local, now) incremental JSON dumps that do not exist in our temporary directory (read: all of them).

I've added separate `RSYNC_INCREMENTAL_JSON_DUMPS_*` environment variables to control the rsync invocation, which I think is clearer than trying to use scripts designed for fullexports anyway (which are intended to only store the few most recent dumps).

# Testing

These changes are currently applied in the musicbrainz-json-dump container on hendrix. I checked the hourly JSON dump logs for any errors and ensured the json-dump-* directories were synced to the sshd-musicbrainz-json-dumps-incremental containter correctly.

### DIFF
--- a/admin/RunIncrementalJSONDump
+++ b/admin/RunIncrementalJSONDump
@@ -23,14 +23,6 @@ source admin/config.sh
 . ./admin/functions.sh
 make_temp_dir
 
-DUMP_DIR="$JSON_DUMP_DIR"/incremental
-
-# Create necessary directories and set permissions. If we're not doing a full
-# dump then $DUMP_STAMP is empty, so that argument is ignored.
-
-mkdir -m "$JSON_DUMP_DIR_MODE" -p "$DUMP_DIR"
-chown "$JSON_DUMP_USER:$JSON_DUMP_GROUP" "$DUMP_DIR"
-
 # The database from which the dumps are generated is replicated, but created
 # as standalone - meaning it has foreign keys and triggers. Foreign keys are
 # needed for MusicBrainz::Server::Role::FollowForeignKeys to work. The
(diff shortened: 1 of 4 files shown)


yvanzo (approved):
Thanks for dealing with this major storage issue so quickly. :ship: 

In a few words, it’s a practical short-term solution, but probably not in the mid-term.

----

> There is not really any reason to do (1) if (2) is succesful, because we presumably backup (2) already with borgmatic; there is no need to store these in three places.

What if (2) fails? The dumps will be left in a temporary directory of the container `musicbrainz-json-dump` and lost if this container is removed (for updating to a new image version or for migrating to a new host). It seems to be the reason for (1): Ensuring a persistent storage of the freshly generated incremental JSON dumps.

But yes, even then:
* The storage in (1) wasn’t cleaned up once (2) succeeded.
* The only risk with skipping (1) is to have to regenerate recent incremental JSON dumps if missing, which is less an issue than the current lack of storage.

> This is more relevant now, because the incremental JSON dumps are being moved to hendrix, which is also where the metabrainz.org container is hosted.

Having all of these and the cron production too on the same host, seems to be addressing the current storage issue, but it will certainly cause performance/availability issues again in the mid-term.

Being on the same host shouldn’t be a requirement for this code to be reliable.

> we'd be storing them twice on the same host!

The basic issue here is having to carry ~100G of data at every single place, which can be addressed with MBS-13302 in the longer term.

mwiencek:
Thank you for the review!

> In a few words, it’s a practical short-term solution, but probably not in the mid-term.

Agreed, but I mainly just wanted to quickly address the dumps being stored twice on the same host.

> What if (2) fails? The dumps will be left in a temporary directory of the container musicbrainz-json-dump and lost if this container is removed (for updating to a new image version or for migrating to a new host).

Yes, this is an issue, but it also existed before: if the the dump script failed (which was far more likely than the rsync for (2) failing), then any already-completed dumps weren't being copied to (1), and it'd have to be done manually.

But yes, (1) was an extra fail-safe in case the actual rsync failed and the container was removed. That makes me wonder if we should target the temporary directory to another volume.

> The only risk with skipping (1) is to have to regenerate recent incremental JSON dumps if missing, which is less an issue than the current lack of storage.

Unfortunately I'm not sure this is possible, because generating an incremental dump for, e.g., replication packet 666 requires the database to be on packet 666.

> Having all of these and the cron production too on the same host, seems to be addressing the current storage issue, but it will certainly cause performance/availability issues again in the mid-term.

I believe there were two main performance/availability issues (correct me if I'm wrong):

 * On aretha (where the JSON dumps ran) the main issue was its slow disks, which is no longer a problem on hendrix.
 * On trille (where metabrainz.org used to run) it was a lack of disk space, which is also no longer a problem on hendrix *for now* (as there appears to be nearly 1TB of free space on hendrix compared to trille's 100GB), but it may become an issue in the future as the disk on hendrix becomes utilized more.

> Being on the same host shouldn’t be a requirement for this code to be reliable. [...] The basic issue here is having to carry ~100G of data at every single place, which can be addressed with MBS-13302 in the longer term.

Agreed that MBS-13302 is a good long-term solution to the storage issue.

yvanzo:
> Unfortunately I'm not sure this is possible, because generating an incremental dump for, e.g., replication packet 666 requires the database to be on packet 666.

If we have a full export anterior to replication packet 666, we can put a mirror database on packet 666.

mwiencek:
> If we have a full export anterior to replication packet 666, we can put a mirror database on packet 666.

That is true. Although...it also depends on values in the `json_dump` schema, which we don't export, and which only stores JSON data for packets since the most recent full JSON dump...so if the incremental script continues running during the next hour, after 666 had failed to sync, and then there is a full JSON dump at packet 669, we might be out of luck, I think.

PR #3143, opened 2024-01-12
MBS-13425: Normalize entities.oclc.org to id.oclc.org

reosarevok:
### Implement MBS-13425

# Problem
OCLC entity pages now redirect to an `entities.oclc.org` URL, which we currently don't recognize as OCLC at all. For example, https://id.oclc.org/worldcat/entity/E39PCYRCtHWqdGhtK9yGDjJKwK loads as https://entities.oclc.org/worldcat/entity/E39PCYRCtHWqdGhtK9yGDjJKwK.html

# Solution
This patch recognizes such URLs as OCLC ones, and cleans them up to use the `id.oclc.org` format since that's the result of their `Copy URI` button, which suggests it's the expected reference URI.

# Testing
Added one cleanup test, plus a quick manual check too.


yvanzo (changes requested):
The `Worldcat` entry in `%URL_SPECIALIZATIONS` of `MusicBrainz::Server::Data::URL` has to be updated too. LGTMBDNT otherwise.

reosarevok:
Good catch, updated :)

yvanzo (approved):
:ship:

PR #3147, opened 2024-01-17
MBS-13432: Ignore format chars when checking for useless notes and check them in more places

reosarevok:
### Implement MBS-13432

# Problem
Our useless edit note checks (`is_valid_edit_note` and `invalidEditNote`) only check edit notes that are required (so, release adds and destructive edits mostly), and they don't check for formatting characters, which means it's very easy to enter useless notes whether by mistake or, like in the case that prompted the ticket, as a strange attempt at vandalism.

# Solution
Three commits here that fix different parts of the problem:

The first makes the edit note checks ignore formatting characters when trying to determine whether a note contains any useful content or not. The affected characters are `ZERO WIDTH SPACE`, `SOFT HYPHEN` and the categories [`Cc`](https://www.compart.com/en/unicode/category/Cc), [`Cf`](https://www.compart.com/en/unicode/category/Cf) and [`Mn`](https://www.compart.com/en/unicode/category/Mn). These are not dropped from the notes, since in (non-empty) edit notes they can actually be perfectly fine and useful; they are only skipped when determining whether the rest of the note is useful.

The second makes it so that the validation code is not only run if the note is required, but in any case that uses `Form::Role::EditNote` (which mostly means entity edit forms) and in the release editor, as long as the note is not empty. Additionally, this fixes a kind-of-bug in the release editor where the messages for empty and invalid note would both be shown at the same time if the note was entirely empty - now only the empty note message is shown in that case.

The third also applies validation for invalid notes to the Form::Vote edit notes, which means invalid notes are also blocked when commenting on edits from edit pages or edit lists. The patch drops the invalid note, but still submits the vote, if any. This is the same as we already do for space-only notes that get trimmed to nothing, so it should at least be consistent.
Since an invalid edit note is either an honest mistake or vandalism, it seems fine not to notify the editor about the fact that it was dropped, which would be harder given that all the voting happens in `enter_votes` rather than the actual edit page or edit list itself.

# Testing
Manually, making sure that the invalid edits are blocked/ignored as intended, votes are still entered properly, and that notes that are valid are still submitted even if some notes in the same `enter_votes` call are invalid.

# Draft progress
* [x] Add some tests for `is_valid_edit_note` in `t/lib/t/MusicBrainz/Server/Validation.pm`
* [x] Add a test to make sure voting while submitting invalid notes does not break in the future.

### DIFF
--- a/lib/MusicBrainz/Server/Validation.pm
+++ b/lib/MusicBrainz/Server/Validation.pm
@@ -52,7 +52,10 @@ use Encode qw( decode encode );
 use Scalar::Util qw( looks_like_number );
 use Text::Unaccent::PurePerl qw( unac_string_utf16 );
 use MusicBrainz::Server::Constants qw( $MAX_POSTGRES_INT $MAX_POSTGRES_BIGINT );
-use MusicBrainz::Server::Data::Utils qw( contains_number );
+use MusicBrainz::Server::Data::Utils qw(
+    contains_number
+    remove_lineformatting_characters
+);
 use utf8;
 
 sub unaccent_utf16 ($)
@@ -315,6 +318,14 @@ sub is_valid_edit_note
 {
     my $edit_note = shift;
 
+    return 0 unless $edit_note;
+
+    # We don't want line format chars to stop an edit note from being "empty"
+    $edit_note = remove_lineformatting_characters($edit_note);
+    # Additional format chars that can cause unwanted negatives
+    $edit_note =~ s/[\p{Cf}\p{Mn}]//g;
+
+    # The note might now be entirely empty - if so, it is useless
     return 0 unless $edit_note;
 
     # An edit note with only spaces and / or punctuation is useless
--- a/t/lib/t/MusicBrainz/Server/Validation.pm
+++ b/t/lib/t/MusicBrainz/Server/Validation.pm
@@ -29,6 +29,7 @@ use MusicBrainz::Server::Validation qw(
     is_valid_partial_date
     is_database_row_id
     is_database_bigint_id
+    is_valid_edit_note
 );
 
 test 'Test trim_in_place' => sub {
@@ -276,6 +277,31 @@ test 'Test normalise_strings' => sub {
     is(normalise_strings('！＄０５ＡＺｂｙ'), '!$05azby', 'Fullwidth Latin to ASCII');
     is(normalise_strings('｡｢･ｱﾝ'), '。「・アン', 'Halfwidth Katakana/punctuation to fullwidth');
 
+};
+
+test 'Test is_valid_edit_note' => sub {
+    ok(!is_valid_edit_note(''), 'Empty edit note is invalid');
+    ok(is_valid_edit_note('This is a note!'), 'Standard edit note is valid');
+    ok(
+        !is_valid_edit_note('a'),
+        'Note made of just one ASCII character is invalid',
+    );
+    ok(
+        is_valid_edit_note('‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏漢'),
+        'Note made of just one kanji character is valid',
+    );
+    ok(
+        !is_valid_edit_note("  \t"),
+        'Note made of spaces and tabs is invalid',
+    );
+    ok(
+        !is_valid_edit_note("\N{ZERO WIDTH JOINER}\N{COMBINING GRAPHEME JOINER}\N{ZERO WIDTH JOINER}\N{COMBINING GRAPHEME JOINER}\N{ZERO WIDTH JOINER}\N{COMBINING GRAPHEME JOINER}"),
+        'Note made of format and join characters is invalid',
+    );
+    ok(
+        is_valid_edit_note("\N{ZERO WIDTH JOINER}\N{COMBINING GRAPHEME JOINER}abc\N{ZERO WIDTH JOINER}\N{COMBINING GRAPHEME JOINER}"),
+        'Note made of format and join characters plus text is valid',
+    );
 };
 
 1;
--- a/t/selenium/Vote_And_Note_Submission.json5
+++ b/t/selenium/Vote_And_Note_Submission.json5
@@ -1,0 +1,73 @@
+{
+  title: 'Vote and note submission',
+  commands: [
+    {
+      command: 'open',
+      target: '/user/editor1/edits',
+      value: '',
+    },
+    // We want to vote No on the first edit
+    {
+      command: 'click',
+      target: 'xpath=//*[@id="id-enter-vote.vote.0.vote-No"]',
+      value: '',
+    },
+    // We want to vote Yes on the second edit
+    {
+      command: 'click',
+      target: 'xpath=//*[@id="id-enter-vote.vote.1.vote-Yes"]',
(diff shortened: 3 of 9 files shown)


lib/MusicBrainz/Server/Validation.pm
```diff
@@ -317,6 +317,9 @@ sub is_valid_edit_note
 
     return 0 unless $edit_note;
 
+    # We don't want line format chars to stop an edit note from being "empty"
+    $edit_note =~ s/[\N{ZERO WIDTH SPACE}\N{SOFT HYPHEN}\p{Cc}\p{Cf}\p{Mn}]//g;
```
mwiencek, line 321: Can this use `remove_lineformatting_characters` for one step and then strip the extra stuff in another step? Just in case we update `remove_lineformatting_characters`in the future; I'd expect we'd want the same changes propagated here.

t/lib/t/MusicBrainz/Server/Validation.pm
```diff
@@ -278,4 +279,29 @@ test 'Test normalise_strings' => sub {
...
+        'Note made of just one kanji character is valid',
+    );
+    ok(
+        !is_valid_edit_note("  \t"),
+        'Note made of spaces and tabs is invalid',
+    );
+    ok(
+        !is_valid_edit_note('‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏ ‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏‍͏'),
```
mwiencek, line 298: Can we use `\x{NNNN}` escape sequences here so it's visible which characters are in the string?

t/selenium/Vote_And_Note_Submission.json5
```diff
@@ -0,0 +1,73 @@
...
+    // We want to vote Yes on the second edit
+    {
+      command: 'click',
+      target: 'xpath=//*[@id="id-enter-vote.vote.1.vote-Yes"]',
+      value: '',
+    },
+    // We want to leave a valid note on the first edit
+    {
+      command: 'click',
+      target: 'xpath=//*[@id="edits"]/form/div[2]/div[2]/div[2]/a[1]',
```
mwiencek, line 24: These are fine for now, but thought I'd note for the future: these types of selectors can be fragile, since if we ever change the structure of the HTML at all, or if the order of the elements changes, they will break. Where possible I stick with IDs/class names that are unlikely to change, and avoid enforcing a specific HTML structure by selecting direct descendants (so, preferring `//` instead of `/`, and specifying the least amount of descendants needed to identify the node).

mwiencek (commented):
Thanks for the added test!

t/selenium/Vote_And_Note_Submission.json5
mwiencek: For example, assuming this is for the add-note button: a good idea might be to add IDs to each of the edit-list* divs and do something like `xpath=//div[@id="edit-123"]//a[text()="Add note"])`.

(* Why is this called `edit-list` when it contains only a single edit? 😕)

reosarevok:

lib/MusicBrainz/Server/Validation.pm
reosarevok: Ok, but then I also need to keep a note to keep that one in sync with `invalidEditNote` :) Done.

t/lib/t/MusicBrainz/Server/Validation.pm
reosarevok: We can't, because Critic disagrees, but we can use `\N{}` named sequences which is a lot better.

t/selenium/Vote_And_Note_Submission.json5
reosarevok: I just used the xpath Chrome gave me for now. Better wouldn't hurt, although it seems "better IDs would be useful" is honestly true of most of our code (and tests) :)

mwiencek (approved):

PR #3148, opened 2024-01-22
MBS-13452: Add patronage autoselect for Buy Me a Coffee URLs

reosarevok:
### Implement MBS-13452

# Description
This autoselects buymeacoffee.com URLs as patronage and does some basic cleanup. We have only 30 or so of these so not adding them to the sidebar yet.

# Testing
Added a basic handling test.


yvanzo (approved):
LGTMBDNT. :100: agreed with not giving more time to it for now. :ship:

PR #3151, opened 2024-01-24
MBS-13463: Report for bootleg releases on non-bootleg label

reosarevok:
### Implement MBS-13463

# Reasoning
Non-bootleg labels pretty much never put out bootlegs. As such, this report lets editors find bootleg releases which have a non bootleg label as a release label, which should almost always be wrong in some way.

[no label] can of course also apply to bootlegs, so it is ignored.

This is basically the opposite of https://github.com/metabrainz/musicbrainz-server/pull/2874

# Testing
With sample data, which had 31 matches.

### DIFF
--- a/root/report/BootlegsOnNonBootlegLabels.js
+++ b/root/report/BootlegsOnNonBootlegLabels.js
@@ -1,0 +1,40 @@
+/*
+ * @flow strict
+ * Copyright (C) 2024 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import ReleaseList from './components/ReleaseList.js';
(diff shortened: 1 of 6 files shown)


root/report/BootlegsOnNonBootlegLabels.js
```diff
@@ -0,0 +1,40 @@
...
+  pager,
+}: ReportDataT<ReportReleaseT>): React$Element<typeof ReportLayout> => (
+  <ReportLayout
+    canBeFiltered={canBeFiltered}
+    description={l(
+      `This report shows releases that are set to status “Bootleg”, but have
+       at least one label in their labels list which is not “[no label]” nor
+       has the type “Bootleg Production”.
+       Other types of labels pretty much never release bootleg releases,
+       so chances are that either the label or the status are wrong.`,
```
mwiencek, line 28:
```suggestion
       so chances are that either the label or the status is wrong.`,
```

mwiencek (approved):
Looks good, with one minor grammatical suggestion.

I tested the query on hendrix and it completed in under 1s.

PR #3153, opened 2024-01-25
MBS-13402: Unbreak unsetting work order in release rel editor

reosarevok:
### Fix MBS-13402

# Problem
When trying to unset work orders by unchecking the "These relationships have a specific ordering" checkbox in the release relationship editor, it triggers the error `EDIT_RELATIONSHIPS_REORDER: invalid link_order`.

# Solution
Unsetting ordering sets orders to 0, but this only accepted positive integers with `is_database_row_id`. We could use `is_non_negative_integer` but I guess it was intentional to limit this to at most `MAX_POSTGRES_INT`, so this just allows 0 in addition to `is_database_row_id`.

# Testing
Manually, making sure the orders can now be unset (and set again).


mwiencek (approved):
Did this break with the new relationship editor, or was it always a bug? (I guess the latter, as I don't remember changing this code.)

PR #3154, opened 2024-01-27
MBS-13465: Prefer non-raw images when displaying front release group cover art

mwiencek:
# Problem

MBS-13465

If a release group does not have cover art explicitly set, then we select the first front image from the earlier release in the RG, including "Raw/Unedited" images.

# Solution

Before sorting by date, sort by the presence of "Raw/Unedited" in the cover art types, preferring front covers that do not have this type.

Note: A similar PR is needed for the artwork-redirect API (though it doesn't matter which is deployed first).

# Testing

Added an automated test.

### DIFF
--- a/lib/MusicBrainz/Server/Data/Artwork.pm
+++ b/lib/MusicBrainz/Server/Data/Artwork.pm
@@ -198,6 +198,8 @@ sub load_for_release_groups
         AND is_front = true
         AND cover_art_presence != 'darkened'
         ORDER BY release.release_group, release_group_cover_art.release,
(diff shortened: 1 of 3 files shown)


mwiencek:
artwork-redirect PR: https://github.com/metabrainz/artwork-redirect/pull/47

reosarevok (approved):
LGTMBDNT

yvanzo (approved):
I didn’t test it but `ASC` order is the default and the added test covers it well. Thanks for having also added other cover art types. :ship:

PR #3156, opened 2024-01-31
Add context to "Time" strings

reosarevok:
At least in Spanish, the time of an event would generally be translated in a significantly different way than the timecodes in a CDTOC.

### DIFF
--- a/root/main/error/components/ErrorEnvironment.js
+++ b/root/main/error/components/ErrorEnvironment.js
@@ -24,7 +24,7 @@ const ErrorEnvironment = ({
   return (
     <>
       <p>
-        <strong>{addColonText(l('Time'))}</strong>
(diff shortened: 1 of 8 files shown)


root/main/error/components/ErrorEnvironment.js
```diff
@@ -24,7 +24,7 @@ const ErrorEnvironment = ({
   return (
     <>
       <p>
-        <strong>{addColonText(l('Time'))}</strong>
+        <strong>{addColonText(lp('Time', 'error datetime'))}</strong>
```
yvanzo, line 27: Maybe even clarify the source string itself:

```suggestion
        <strong>{addColonText(lp('Date and time'))}</strong>
```

yvanzo (approved):
LGTM but the error string can probably be improved at the same time. :ship:

PR #3157, opened 2024-01-31
Remove colons from Type strings

reosarevok:
We already have a lot of Type strings without colons and these do not mean different things in different places.


yvanzo (commented):
Did you test these changes?

`addColonText` seems to be provided by Webpack plug-in, but it is still manually imported in 4 files.

reosarevok:
I did not, but AFAICT they are all safe string replacements. The 4 imports seem to be left behind from when we started using this with webpack, I removed them in a second commit.

yvanzo (approved):
Briefly tested entity pages and editing forms :ship:

PR #3159, opened 2024-01-31
Add context to "cancelled" for events

reosarevok:
We had context for the "cancelled" string for edits, but not for events (the only uses left without context).


yvanzo (approved):
:ship:

PR #3160, opened 2024-01-31
Drop contact email from strings

reosarevok:
We generally use `support@metabrainz` now, but in any case most of our strings direct people to the contact us page instead. Doing the same with these four stragglers. I also somewhat simplified the strings while at it, and changed some text that seemed to suggest the `account/edit` page would show the entered email (it does not, until it is verified).


yvanzo (approved):
LGTMBDNT :ship:

PR #3162, opened 2024-01-31
Add context to [none] barcodes

reosarevok:
We already have context for [none] strings for different entity types, but not when used to indicate no barcode.


yvanzo (approved):
:ship:

PR #3163, opened 2024-02-01
MBS-13472: Also update rel-editor entity name with guess case

reosarevok:
### Fix MBS-13472

# Problem
We update the entity name the relationship editor sees with `useEntityNameFromField` so that when editors rename the entity the current name in the form is shown on relationship previews and the like, but this depends on an event listener that was not getting triggered by guess case. As such, when using guess case the relationship preview would still show the old name.

# Solution
This just makes it so that the guess case button also triggers an input event, so that `useEntityNameFromField` will be able to use the new name. We pass a ref down to the `input` in `FormRowText` since that seems like the easiest way to trigger the event on the input later on.

# Testing
Manually, making sure that a new recording name now changes properly in relationship previews when affected by guess case.


mwiencek (approved):
Tested locally, works fine.

PR #3166, opened 2024-02-06
MBS-13474: Allow TikTok videos on recordings

reosarevok:
### Implement MBS-13474

# Description
We were turning these into profile links, but actually there's a fair amount of "TikTok music" being created that originates in and is mostly streamed at TikTok itself.
As such, it makes sense to allow linking recordings to `/video` TikTok links (/music links, which are used for videos that use licensed music, do not seem to belong in MusicBrainz).

# Testing
Added a test, plus checked manually that the links are blocked and allowed as expected.

### DIFF
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -5698,16 +5698,30 @@ const CLEANUPS: CleanupEntries = {
   },
   'tiktok': {
     match: [new RegExp('^(https?://)?(www\\.)?tiktok\\.com', 'i')],
-    restrict: [LINK_TYPES.socialnetwork],
+    restrict: [{...LINK_TYPES.streamingfree, ...LINK_TYPES.socialnetwork}],
     clean: function (url) {
       return url.replace(
-        /^(?:https?:\/\/)(?:www\.)?tiktok\.com\/@([\w.]+)(?:[\/?&#].*)?$/,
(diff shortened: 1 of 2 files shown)


root/static/scripts/edit/URLCleanup.js
```diff
@@ -5698,16 +5698,30 @@ const CLEANUPS: CleanupEntries = {
   },
   'tiktok': {
     match: [new RegExp('^(https?://)?(www\\.)?tiktok\\.com', 'i')],
-    restrict: [LINK_TYPES.socialnetwork],
+    restrict: [{...LINK_TYPES.streamingfree, ...LINK_TYPES.socialnetwork}],
     clean: function (url) {
       return url.replace(
-        /^(?:https?:\/\/)(?:www\.)?tiktok\.com\/@([\w.]+)(?:[\/?&#].*)?$/,
+        /^(?:https?:\/\/)(?:www\.)?tiktok\.com\/@([\w.]+(?:\/video\/\d+)?)(?:[\/?#].*)?$/,
```
yvanzo, line 5704: It is generally fine to keep `&` in the list of possible separators as some websites use it instead of `?` despite the fact that it doesn’t follow the [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).

No need to keep `&` here as TikTok seems to follow the standard. This regexp has been probably initially copied from a rule for another website that has this kind of issue.

yvanzo (approved):
LGTMBDNT :ship:

PR #3167, opened 2024-02-08
MBS-13424: Allow Spotify shop links

reosarevok:
### Implement MBS-13424

# Description
Much like Bandcamp, Spotify also allows selling merch now, including both physical and digital music, under https://shop.spotify.com/ (that link by itself doesn't work though, example from the ticket is https://shop.spotify.com/en-US/artist/4O15NlyKLIASxsJ0PrXPfz/store).

Only allowing these for releases and recordings since those seem like the most obviously useful options, and allowing either mail order or download purchase but not both since each item seems to be one or the other anyway.

# Testing
Added a test.


yvanzo (approved):
LGTMBDNT :ship:

PR #3168, opened 2024-02-08
Move track-artist-updating code into testable function

derat:
Move the release editor code that updates track credits when a release's credits are changed into a new standalone getUpdatedTrackArtists function, and add unit tests for it.

No functional changes are intended, but this should make it easier to make future improvements as discussed in MBS-13273.

# Problem

I'm planning to work on MBS-13273 some more, but testing the many edge cases via Selenium is painful.

# Solution

This change moves code for updating track credits out of the release editor's `init.js` file and into a separate Flow-checked utility file that can be unit-tested.

# Testing

I added a unit test that exercises a bunch of different cases. The existing release editor Selenium tests will hopefully pass, too.

### DIFF
--- a/root/static/scripts/common/immutable-entities.js
+++ b/root/static/scripts/common/immutable-entities.js
@@ -42,7 +42,7 @@ export const reduceArtistCredit =
  */
 export function reduceArtistCreditNames(
   names: $ReadOnlyArray<ArtistCreditNameT>,
-  dropFinalJoinPhrase?: false,
+  dropFinalJoinPhrase?: boolean = false,
 ): string {
   let s = names.reduce(reduceName, '');
   if (dropFinalJoinPhrase && names.length > 0) {
--- a/root/static/scripts/edit/utility/getUpdatedTrackArtists.js
+++ b/root/static/scripts/edit/utility/getUpdatedTrackArtists.js
@@ -1,0 +1,92 @@
+/*
+ * @flow strict-local
+ * Copyright (C) 2024 MetaBrainz Foundation
+ *
+ * This file is part of MusicBrainz, the open internet music database,
+ * and is licensed under the GPL version 2, or (at your option) any
+ * later version: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+import {reduceArtistCreditNames} from '../../common/immutable-entities.js';
+import {arraysEqual} from '../../common/utility/arrays.js';
+import {cloneArrayDeep} from '../../common/utility/cloneDeep.mjs';
+
+/*
+ * Returns true if the supplied credits include the same underlying artists.
+ */
+const sameArtists = (
+  a: $ReadOnlyArray<ArtistCreditNameT>,
+  b: $ReadOnlyArray<ArtistCreditNameT>,
+) => arraysEqual(a, b,
+  (a, b) => a.artist && b.artist && a.artist.gid === b.artist.gid);
+
+/*
+ * Returns updated artist credits for a track that's part of a release whose
+ * credits are being changed from oldArtists to newArtists.
+ *
+ * If the track's credits should not be changed, the original trackArtists
+ * array is returned.
+ */
+export default function getUpdatedTrackArtists(
+  trackArtists: $ReadOnlyArray<ArtistCreditNameT>,
+  oldArtists: $ReadOnlyArray<ArtistCreditNameT>,
+  newArtists: $ReadOnlyArray<ArtistCreditNameT>,
+): $ReadOnlyArray<ArtistCreditNameT> {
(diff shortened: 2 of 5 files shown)


root/static/scripts/common/immutable-entities.js
```diff
@@ -42,7 +42,7 @@ export const reduceArtistCredit =
  */
 export function reduceArtistCreditNames(
   names: $ReadOnlyArray<ArtistCreditNameT>,
-  dropFinalJoinPhrase?: false,
+  dropFinalJoinPhrase?: boolean = false,
```
derat, line 45: Whoops, I think this was a bug from one of my earlier changes. Still getting used to Flow...

root/static/scripts/edit/utility/getUpdatedTrackArtists.js
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * If the track's credits should not be changed, the original trackArtists
+ * array is returned.
+ */
+export default function getUpdatedTrackArtists(
+  trackArtists: $ReadOnlyArray<ArtistCreditNameT>,
+  oldArtists: $ReadOnlyArray<ArtistCreditNameT>,
+  newArtists: $ReadOnlyArray<ArtistCreditNameT>,
+): $ReadOnlyArray<ArtistCreditNameT> {
+  const oldArtistsReduced = reduceArtistCreditNames(oldArtists);
```
derat, line 35: This code previously only reduced these credits once per release, but now it does it once per track. I initially added support for passing in already-reduced credits, but it felt like premature optimization since we still need to reduce the track credits in the next statement.

root/static/scripts/edit/utility/getUpdatedTrackArtists.js
```diff
@@ -0,0 +1,92 @@
...
+    ) {
+      /*
+       * Replace the old release artist(s) with the new one(s) and restore
+       * the old join phrase following the release artist.
+       */
+      const names = cloneArrayDeep(newArtists)
+        .concat(trackArtists.slice(oldArtists.length));
+      names[newArtists.length - 1] = {
+        artist: names[newArtists.length - 1].artist,
+        joinPhrase: trackArtists[oldArtists.length - 1].joinPhrase,
```
derat, line 84: The old code was updated the `joinPhrase` property in the existing immutable object, but I think it was just able to get away with that by virtue of being in a non-Flow-checked file.

root/static/scripts/tests/edit/utility/getUpdatedTrackArtists.js
```diff
@@ -0,0 +1,133 @@
...
+        name('B', '2', ' feat. '),
+        name('C', '3'),
+      ],
+      oldRel: [name('A', '1', ' & ', 'a'), name('B', '2')],
+      newRel: [name('A', '1', ' & '), name('B', '2')],
+      want: [
+        /*
+         * TODO: As discussed in MBS-13273, the first artist's credited-as
+         * name should be removed here.
+         */
```
derat, line 119: This is something that I plan to change, but I figured it'd be good to add a test case now to document the existing behavior.

derat (commented):
@mwiencek and/or @reosarevok, mind taking a look since you reviewed the code that this is moving?

derat:

I think the failure was this, in the "Artist Edit Form" Selenium test, which seems unrelated to this change:

```
    ⨯ caught exception: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON at JSON.parse (<anonymous>) at handleCommand
```

root/static/scripts/tests/edit/utility/getUpdatedTrackArtists.js
```diff
@@ -0,0 +1,133 @@
...
+  const testCases = [
+    {
+      desc: 'artist change, track matches old',
+      track: [name('Old', '1')],
+      oldRel: [name('Old', '1')],
+      newRel: [name('New', '2')],
+      want: [name('New', '2')],
+    },
+    {
+      desc: 'artist change, track doesn\'t match old',
```
reosarevok, line 36: You can use unicode apostrophes rather than escaping the ASCII ones here.

mwiencek: The main benefit of that is if the test fails, you can grep for the string more easily, since what is printed will match the source code.

derat: Makes sense; thanks for the explanation. I've updated this.

mwiencek (approved):
Thanks for all of the added tests! This will definitely make further changes easier to be sure about.

derat:
@reosarevok, any other comments? If not, mind merging this? :-)

PR #3170, opened 2024-02-10
MBS-13481: Invalidate release on label changes

mwiencek:
# Problem

MBS-13481, plus another (pretty much harmless) issue with bogus keys being added to the cache.

# Solution

See associated commits for further details.

# Testing

Added some automated tests.

### DIFF
--- a/lib/MusicBrainz/Server/Data/ReleaseLabel.pm
+++ b/lib/MusicBrainz/Server/Data/ReleaseLabel.pm
@@ -41,7 +41,7 @@ sub _entity_class
 sub load
 {
     my ($self, @releases) = @_;
-    my %id_to_release = object_to_ids(@releases);
+    my %id_to_release = object_to_ids(grep { !$_->has_labels } @releases);
     my @ids = keys %id_to_release;
     return unless @ids; # nothing to do
     my $query = 'SELECT ' . $self->_columns . '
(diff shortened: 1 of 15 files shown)


lib/MusicBrainz/Server/Data/ReleaseLabel.pm
```diff
@@ -41,7 +41,7 @@ sub _entity_class
 sub load
 {
     my ($self, @releases) = @_;
-    my %id_to_release = object_to_ids(@releases);
+    my %id_to_release = object_to_ids(grep { !$_->has_labels } @releases);
```
reosarevok, line 44: I guess we don't need to worry about the labels having changed since they were loaded since this mostly only happens just after loading the release to begin with? Unless we load these on top of release data from edits or whatnot.

reosarevok (approved):
LGTMBDNT

yvanzo (approved):
Changes are sound but it is hard to say if it is complete.

Searched the code and mostly found the same classes as in the last commit, but it is difficult to check whether it is actually covering all the cases where deleting/invalidating cache would be needed.

Let’s deploy it now to both beta and production to address this reported bug subsequent to the PR #3130 which has been recently deployed to both websites too.

yvanzo:
For the record, it has been amended with the pull request https://github.com/metabrainz/musicbrainz-server/pull/3172 before releasing.

PR #3172, opened 2024-02-13
Don't delay cache additions until after the transaction ends

mwiencek:
# Problem

After searching for issues similar to [MBS-13481](https://tickets.metabrainz.org/browse/MBS-13481) (where we load additional data onto an entity which is then entered into the cache, but where changes to that additional data do not invalidate the entity's cache entry), I believe that there are simply far too many uncovered cases which we'd have to handle.

While storing more data in the cache and having intelligent invalidation logic for all the linked data sounds desirable, it's too many changes to manage for now, so we need to restore the previous behavior.

# Solution

This partially reverts https://github.com/metabrainz/musicbrainz-server/commit/82c79f531edad396c20a2fceb45a37de9035e055 and https://github.com/metabrainz/musicbrainz-server/commit/42be4bcf01f198530789772629093d1b69b853c5.

I realized (later than I should have) that we don't need to delay cache additions until after the database transaction ends. We may have had to in a previous implementation I had, but no longer.

The `post_txn_callbacks` ran regardless of whether or not the commit succeeded, so an entity would always be cached in the state it was initially loaded in. The only difference now is when that occurs. There's no type of race condition that isn't already handled:

 * If the entity is deleted from the cache right after it's added, then there is no issue, just like before.

 * If the entity is added to the cache right after it's deleted, then there is still no issue, because our existing code to delete the added entry if there exists a `recently_invalidated` key for the same still applies.

# Testing

The existing `*EntityCache` tests pass, and I've added a couple additional tests to ensure that neither release events nor release labels are entered into the cache with a release.

### DIFF
--- a/lib/MusicBrainz/Server/Data/ReleaseLabel.pm
+++ b/lib/MusicBrainz/Server/Data/ReleaseLabel.pm
@@ -136,7 +136,6 @@ sub insert
 
     my $release_id = $row->{release};
     $self->c->model('Series')->reorder_for_entities('release', $release_id);
-    $self->c->model('Release')->_delete_from_cache($release_id);
 
     return wantarray ? @created : $created[0];
 }
@@ -155,7 +154,6 @@ sub update
     );
 
     $self->c->model('Series')->reorder_for_entities('release', $release_id);
-    $self->c->model('Release')->_delete_from_cache($release_id);
 }
 
 sub delete {
@@ -167,7 +165,6 @@ sub delete {
     );
 
     $self->c->model('Series')->reorder_for_entities('release', @$release_ids);
-    $self->c->model('Release')->_delete_from_cache(@$release_ids);
 }
 
 __PACKAGE__->meta->make_immutable;
--- a/lib/MusicBrainz/Server/Data/ArtistCredit.pm
+++ b/lib/MusicBrainz/Server/Data/ArtistCredit.pm
@@ -16,11 +16,14 @@ with 'MusicBrainz::Server::Data::Role::EntityCache',
      'MusicBrainz::Server::Data::Role::PendingEdits' => {
         table => 'artist_credit',
      },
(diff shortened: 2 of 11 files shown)


mwiencek:
I have a fix for the `Data::Artist` failures, but the `Data::Series` ones require further investigation.

lib/MusicBrainz/Server/Data/ReleaseLabel.pm
```diff
@@ -136,7 +136,6 @@ sub insert
 
     my $release_id = $row->{release};
     $self->c->model('Series')->reorder_for_entities('release', $release_id);
-    $self->c->model('Release')->_delete_from_cache($release_id);
```
reosarevok, line 139: Does having this in preparation for that time when we get more intelligent invalidation logic actually hurt?

mwiencek: I don't know if/when that will happen, so it does hurt *slightly* in that it will increase the number of release cache misses for no benefit.

reosarevok: Oh, ok. I thought we'd want stuff to be uncached when it changes anyway, that's why I asked.

mwiencek: Just to be clear, we do, but in this case nothing is changing anymore. :) (Because release labels are no longer cached on the release.)

yvanzo (approved):
Thanks for the in-depth investigation of this cache issue even though it doesn’t come to the desired improvements. I agree that we cannot manage such changes at the moment.

Changes look good, commit messages are very explanatory, and I also quickly tested the website locally just in case.

Merging and deploying to both beta and production as requested on IRC. :ship: :b: :factory:

PR #3173, opened 2024-02-14
MBS-13486 / MBS-13487: Small CAA for pseudo-releases improvements

reosarevok:
### Implement MBS-13486, MBS-13487

# Problem
Our [guidelines](https://musicbrainz.org/doc/Style/Specific_types_of_releases/Pseudo-Releases) say pseudo-releases shouldn't generally get cover art (the only exception is if there's no official release and it helps figure out what the official release should be, after which they should be deleted from the pseudo-release). Despite that, we were still showing pseudo-releases in the "Releases without any art in the Cover Art Archive" report, suggesting cover art should be uploaded, and we had no way for users to know this was not the case without them going searching for a fairly obscure guideline.

# Solution
This does two things: it makes it so pseudo-releases no longer show in the report, and it adds a line to their cover art tab when there's no cover art yet specifying that cover art should generally not be uploaded and linking to the guidelines.

# Testing
Manually with sample data, by changing the status of a release without cover art and making sure it appeared in the report when it had no status or a status other than pseudo-release, but not when set to pseudo-release; and that the new text showed in the CAA tab when it was set to pseudo-release, but not when it had no status or a different one.

### DIFF
--- a/lib/MusicBrainz/Server/Report/ReleasesWithoutCAA.pm
+++ b/lib/MusicBrainz/Server/Report/ReleasesWithoutCAA.pm
@@ -11,11 +11,12 @@ sub query {<<~'SQL'}
       FROM (
         SELECT DISTINCT r.*
           FROM release r
-         WHERE NOT EXISTS (
+         WHERE (r.status IS DISTINCT FROM 4) -- skip pseudo-releases
+           AND NOT EXISTS (
             SELECT 1
               FROM cover_art_archive.cover_art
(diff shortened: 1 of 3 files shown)


mwiencek (approved):
LGTM, just a couple small suggestions

Aerozol:
Small thought - it might be worth adding a bullet point, at the top of the report page, saying that pseudo-releases are excluded.

PR #3174, opened 2024-02-14
Some eslint fixes for unfixed rules

reosarevok:
I got my VS Code eslint working again and I thought I'd fix a bunch of these.

### DIFF
--- a/root/edit/components/Vote.js
+++ b/root/edit/components/Vote.js
@@ -42,7 +42,7 @@ const VoteCheckbox = ({
     ? getLatestVoteForEditor(edit, user)
     : null;
   const checked =
-    (latestVote && latestVote.vote == props.value) ||
+    (latestVote && latestVote.vote === props.value) ||
     (!latestVote && props.value === EDIT_VOTE_NONE);
   return (
     <label htmlFor={`id-${name}-${label}`}>
--- a/root/static/scripts/edit/URLCleanup.js
+++ b/root/static/scripts/edit/URLCleanup.js
@@ -451,6 +451,7 @@ type CleanupEntries = {
   +[type: string]: CleanupEntry,
 };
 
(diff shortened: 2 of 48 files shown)


root/edit/components/Vote.js
```diff
@@ -42,7 +42,7 @@ const VoteCheckbox = ({
     ? getLatestVoteForEditor(edit, user)
     : null;
   const checked =
-    (latestVote && latestVote.vote == props.value) ||
+    (latestVote && latestVote.vote === props.value) ||
```
mwiencek, line 45: I was wondering if I had it like that because of Perl serialization issues. But I see we do `$self->vote + 0` in `Entity::Vote`, so it should be fine...

root/static/scripts/edit/URLCleanup.js
```diff
@@ -451,6 +451,7 @@ type CleanupEntries = {
   +[type: string]: CleanupEntry,
 };
 
+/* eslint-disable sort-keys */
```
mwiencek, line 454: We disable this rule so often that I wonder if there's even a point to have it enabled by default across the codebase.

mwiencek (approved):
LGTM. The `sort-keys` rule seems quite annoying, dunno if you feel it's useful? I also dislike how it doesn't allow me to have the `type` first in dispatch actions.

root/static/scripts/edit/URLCleanup.js
reosarevok: Well, I like that we have a consistent sorting, either because we sort for this, or because we unset it because we want a *different* consistent sorting :) But it's not the most important thing in the world.

reosarevok:
Merged this for now - if we decide to drop `sort-keys` in the future it should be easy to also drop the disabling and enabling lines anyway :)

PR #3175, opened 2024-02-15
MBS-13476: Show release label disambiguations on hover

reosarevok:
### Implement MBS-13476

# Problem
We hide disambiguations for labels on the release sidebar (I assume because of the lack of space) but this can be useful to quickly check if the right label is selected, so we should at least show it on hover.

# Solution
This adds a `showDisambiguationOnHover` option to `EntityLink`, which will either show the disambiguation as the whole hover if there's nothing else there or append it in brackets if there is (for the current use case `hover` should be always empty AFAICT).

# Testing
Manually, making sure the disambiguation shows if there's one, and no empty hover tooltip shows otherwise.

### DIFF
--- a/root/layout/components/sidebar/ReleaseSidebar.js
+++ b/root/layout/components/sidebar/ReleaseSidebar.js
@@ -228,6 +228,7 @@ const ReleaseSidebar = ({release}: Props): React$Element<'div'> | null => {
                       className="wrap-anywhere"
                       entity={releaseLabel.label}
                       showDisambiguation={false}
+                      showDisambiguationOnHover
                     />
                     <br />
                   </>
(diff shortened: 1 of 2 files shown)


root/layout/components/sidebar/ReleaseSidebar.js
```diff
@@ -228,6 +228,7 @@ const ReleaseSidebar = ({release}: Props): React$Element<'div'> | null => {
                       className="wrap-anywhere"
                       entity={releaseLabel.label}
                       showDisambiguation={false}
+                      showDisambiguationOnHover
```
mwiencek, line 231: Maybe the type of `showDisambiguation` could be `boolean | 'hover'`, even?

mwiencek (approved):

PR #3176, opened 2024-02-15
MBS-13441: Also accept localized ra.co links

reosarevok:
### Implement MBS-13441

# Description
RA also has localized links (see test for an example) which we were entirely ignoring. It seems that it is just a translation for the interface rather than a separate site, so it should be fine to just clean it up to the generic ra.co version.

# Testing
Changed a label link on the automated tests to the Spanish version to make sure it still gets cleaned up properly.


yvanzo (approved):
LGTMBDNT :ship:

PR #3177, opened 2024-02-16
MBS-13489: Update alias locales to Unicode CLDR 44

yvanzo:
# Problem MBS-13489

Keep up to date with Unicode CLDR for alias locales.

# Solution

- Bump Datetime-Locale from 1.39 to 1.40 (CLDR from 43 to 44.0.1);
- Rebuild the list of locales for aliases;
- Add Saraiki to the special cases of locale name to work around [CLDR-17385](https://unicode-org.atlassian.net/browse/CLDR-17385) reported upstream bug;
- Drop unneeded/defunct special cases of locale name;
- Fix sorting alias locales by name with the new special cases;
- Also rebuild CPAN snapshots, all minor seemingly impact-less updates;
- Replace the last occurrence of `sr_Latin_YU` in the database through edit [#108936635](https://musicbrainz.org/edit/108936635).

# Progress

* [x] Review [CLDR 44 release notes](https://cldr.unicode.org/index/downloads/cldr-44);
* [x] Rebuild CPAN snapshots
* [x] Review (seemingly impact-less) changes for all (minor) updates in CPAN snapshot files.
  Tip:
  ```shell
  git show --no-color | sed -n 's_^+  \([^ ]*\)-[0-9].*$_https://metacpan.org/dist/\1/changes_p' | sort -u | xargs firefox
  ```

# Testing

Manually:
* Opened an artist alias page;
* Edited an artist alias and checked some new locales are indeed in the list.

Also checked server logs.

# Action (after merge)

* Nothing about existing aliases as removed locale codes are no longer used in the database;
* Nothing about test image because changes to updated packages are minor enough.


yvanzo:
One noticeable change is that [CookieBaker now sets secure when SameSite=None](https://metacpan.org/dist/Cookie-Baker/changes) which is what we do already in beta.

reosarevok (approved):
LGTMBDNT

PR #3178, opened 2024-02-16
Amend #3091: Prevent git warning for ownership

yvanzo:
# Problem

Since PR #3091, CircleCI tests output contains:

```
fatal: detected dubious ownership in repository at '/home/musicbrainz/musicbrainz-server'
To add an exception for this directory, call:

        git config --global --add safe.directory /home/musicbrainz/musicbrainz-server
```

This is because `git grep` is run as `root` whereas the [files are owned by `musicbrainz`](https://github.com/metabrainz/musicbrainz-server/blob/v-2024-02-16/.circleci/config.yml#L27).

# Solution

Run as `musicbrainz` just like for all other test commands.

# Testing

Tested through SSH in CircleCI by running this modified command against `root/statistics` and `root/event`.

# Further action

New test images must be build to make use of this change.
However, this change alone is not worth building new test images.

### DIFF
--- a/docker/musicbrainz-tests/run_circleci_tests.sh
+++ b/docker/musicbrainz-tests/run_circleci_tests.sh
@@ -33,7 +33,7 @@ MUSICBRAINZ_RUNNING_TESTS=1 \
 
 sudo -E -H -u musicbrainz ./node_modules/.bin/flow --quiet
 sudo -E -H -u musicbrainz ./node_modules/.bin/eslint --max-warnings 0 .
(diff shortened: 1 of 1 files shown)


mwiencek (approved):
LGTMBDNT.

yvanzo:
Not sure why CI tests constantly failed  with perl 5.34 and not for perl 5.30.
Since this PR isn’t related to Perl, neither should be these failures hopefully.

PR #3179, opened 2024-02-20
MBS-13403: Drop unused priority column from link_type

reosarevok:
### Implement MBS-13403

# Description
The `priority` column was added in 2007 with https://github.com/metabrainz/musicbrainz-server/commit/94f59f6 and AFAICT never used for anything since. All our relationship types had the default `0` value except for one, which had `1` (I recently changed that to `0` as well)... but that changed nothing since, as mentioned, we do not use the value in any way.

I cannot drop this from the `Dict` in `AddLinkType` / `EditLinkType` since it fails to load old edits without it, so I left it in but added a comment explaining we no longer enter it.

# Testing
Checked that I can view add reltype edits from before this change. Checked that the add and edit reltype forms load fine, submit fine, and I can view the edits they generate fine.


yvanzo (approved):
LGTMBDNT

mwiencek (approved):

PR #3180, opened 2024-02-20
MBS-13492: Move beginner status to a privilege flag

reosarevok:
### Implement MBS-13492

# Problem
Anything that requires checking whether an editor is a beginner requires us to calculate beginnership every time, which is slow and cumbersome, and implemented in at least three different places in slightly different ways. Specifically, calculating beginnership as part of edit search conditions is awfully, terribly slow.

# Solution
Every new editor added via /register will get the `BEGINNER_FLAG` privilege by default, which will only be removed
when accepting edits once the number of edits + time passed requirements are fulfilled. The flag cannot be set/unset via Edit user, since even admins should have no reason to override the default mechanism for this.

We calculate this in `_do_accept`, rather than the edit queue itself, because approvals skip the edit queue but are still relevant for this check.

We run a one-off script to set the flag on any existing beginners.

# Testing
Registered a new user and saw it does get the Beginner flag. Added a test to make sure the `EditQueue` processing removes the beginner flag when appropriate.

# Draft progress
* [x] Add a cron job so that the beginner flag is eventually removed (@mwiencek suggested running this as part of `CheckVotes` on a per-edit basis, since we have the editor loaded and we can check their `privs` and `member_since` without having to query them, so we would only need to check if they have the required amount of edits and then unset their beginner flag).
* [x] Add a one-off script to set the beginner flag on every beginner account that already exists, after which they'll just work with the cron job like others.
* [x] Remove extra cruft from tests (such as SQL inserting 10 nonsense edits so that the editor is not a beginner anymore)

### DIFF
--- a/admin/sql/updates/20240221-mbs-13492.sql
+++ b/admin/sql/updates/20240221-mbs-13492.sql
@@ -1,0 +1,22 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+UPDATE editor
+   SET privs = privs | 8192 -- set new beginner flag
+ WHERE id != 4 -- avoid setting ModBot as beginner
+   AND NOT deleted
+   AND ( 
+        member_since > NOW() - INTERVAL '2 weeks'
+       OR
+        NOT EXISTS (
+          SELECT 1
+            FROM edit
+           WHERE edit.editor = editor.id
+             AND edit.autoedit = 0
+             AND edit.status = 2
+          OFFSET 9
+        )
+   );
+
+COMMIT;
--- a/lib/MusicBrainz/Server/Data/Edit.pm
+++ b/lib/MusicBrainz/Server/Data/Edit.pm
@@ -25,6 +25,7 @@ use MusicBrainz::Server::Constants qw(
     :edit_status
     :vote
     $AUTO_EDITOR_FLAG
+    $BEGINNER_FLAG
     $UNTRUSTED_FLAG
     $MINIMUM_RESPONSE_PERIOD
     $LIMIT_FOR_EDIT_LISTING
@@ -795,6 +796,34 @@ sub _do_accept
 
     my $status = try {
         $edit->accept;
+
+        # If this was not an autoedit and the editor is a beginner
+        # who can now lose the flag, do that
+        unless ($edit->auto_edit) {
+            my $editor_id = $edit->editor_id;
+            my $editor = $self->c->model('Editor')->get_by_id($editor_id);
+            if ($editor->is_beginner && !$editor->is_newbie) {
+                # We check if this will be the tenth accepted edit
+                my $has_nine_edits =
+                    $self->c->sql->select_single_value(<<~"SQL", $editor_id, $STATUS_APPLIED);
+                        SELECT 1
+                          FROM edit
+                         WHERE edit.editor = ?
+                           AND edit.autoedit = 0
+                           AND edit.status = ?
+                        OFFSET 8
+                        SQL
+
+                if ($has_nine_edits) {
+                    $self->c->sql->do(<<~"SQL", $BEGINNER_FLAG, $editor_id);
+                        UPDATE editor
+                           SET privs = privs - ?
+                         WHERE id = ?
+                        SQL
+                }
+            }
+        }
+
         return $STATUS_APPLIED;
     }
     catch {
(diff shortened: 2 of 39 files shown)


lib/MusicBrainz/Server/EditQueue.pm
```diff
@@ -149,6 +158,28 @@ sub _process_open_edit
         $self->log->debug("Applying edit #$edit_id\n");
         unless ($self->dry_run) {
             $self->c->model('Edit')->accept($edit);
+
+            # If the editor is a beginner who can now lose the flag, do that
```
mwiencek, line 162: Hmm, maybe this should be moved into `Edit::accept`? For example, if an auto-editor approves a beginner's edit.

mwiencek (commented):
> The flag cannot be set/unset via Edit user, since even admins should have no reason to override the default mechanism for this.

Would that affect bot users in any way?

reosarevok:
Well, we do not actually have any special mechanism for beginner bots at the moment :) I expect even just the test edits would be enough for a bot to stop being a beginner, in most cases.

lib/MusicBrainz/Server/EditQueue.pm
reosarevok: Oh, that's fair, I didn't consider that but those should also count. Will move it.

reosarevok: Hmm, but `accept` is also run for auto-edits, so we shouldn't *always* run this there. I guess we can pass a parameter for whether it should count or not.

mwiencek: Can't you just check `$edit->auto_edit`?

admin/sql/updates/20240221-mbs-13492.sql
```diff
@@ -0,0 +1,22 @@
+\set ON_ERROR_STOP 1
+
+BEGIN;
+
+UPDATE editor
```
mwiencek, line 5: I had a question about this query on IRC a while ago but I don't think it was addressed. Specifically the number of editors these conditions match seems to be unusually high:

```
musicbrainz_db=> SELECT count(*) FROM editor
 WHERE id != 4
   AND NOT deleted
   AND (
        member_since < NOW() - INTERVAL '2 weeks'
       OR
        NOT EXISTS (
          SELECT 1
            FROM edit
           WHERE edit.editor = editor.id
             AND edit.autoedit = 0
             AND edit.status = 2
          OFFSET 9
        ));
  count
---------
 2396138
(1 row)

musicbrainz_db=> select count(*) from editor;
  count
---------
 2402247
(1 row)
```
That would leave only 6,109 non-beginner editors. That doesn't seem right.

admin/sql/updates/20240221-mbs-13492.sql
```diff
@@ -0,0 +1,22 @@
...
+   SET privs = privs | 8192 -- set new beginner flag
+ WHERE id != 4 -- avoid setting ModBot as beginner
+   AND NOT deleted
+   AND ( 
+        member_since < NOW() - INTERVAL '2 weeks'
```
mwiencek, line 10: Oh, I think I see the issue right after leaving the previous comment.
```suggestion
        member_since > NOW() - INTERVAL '2 weeks'
```

mwiencek: Changing that gives 61,242 non-beginner editors.

lib/MusicBrainz/Server/Data/Edit.pm
```diff
@@ -795,6 +796,34 @@ sub _do_accept
...
+                # We check if this will be the tenth accepted edit
+                my $has_nine_edits =
+                    $self->c->sql->select_single_value(<<~"SQL");
+                        SELECT 1
+                          FROM edit
+                         WHERE edit.editor = $editor_id
+                           AND edit.autoedit = 0
+                           AND edit.status = $STATUS_APPLIED
+                        OFFSET 8
+                        SQL
```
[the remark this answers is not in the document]
reosarevok: Sure, I guess - what is the benefit anyway? We're fairly inconsistent with this.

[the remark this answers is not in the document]
mwiencek: Just anti-SQL-injection. These are supposed to be constants, but I'd rather be safe than sorry. :) Interpolating `$STATUS_APPLIED` does seem less bad than `$editor_id`, since it's actually `Readonly`, and `$editor_id` should be validated by Moose as an integer, but...

mwiencek (approved):

PR #3181, opened 2024-02-22
MBS-13495: Use RG table for RG disambiguation filter

reosarevok:
### Fix MBS-13495

# Problem
The RG diambiguation filter tries to access a non-existent `rg` table when filtering in fast mode (using `artist_release_group`).

# Solution
Just set `$needs_rg_table` to `1`, as we do with other similar filters but seemingly I just forgot to do here.

# Testing
Manually with r-o main db access with and without this (testing with a local sample db works either way since it uses the slow mode with the `release_group` table).


mwiencek (approved):
Good catch, thanks!

PR #3182, opened 2024-02-22
Add context to more Unknown strings

reosarevok:


mwiencek (approved):
Makes sense. I assume distinguishing between "entity type" (which I'd normally associated with a core entity name) and e.g., "artist type" doesn't matter much?

reosarevok:
I thought I wouldn't split those further because I expect "Type" to be the same gender/number in all languages whatever the type is for, but to be honest I'm not 100% sure :) (just seemed like it would be annoying to have to translate the same line once for each entity)

yvanzo (commented):
I agree with both of you that _entity type_ would more likely be associated with Artist, Event… and that there is no need to distinguish source strings depending on what kind of entity has this unknown type property.

Maybe just use _type_ instead?

PR #3184, opened 2024-02-26
Minor string improvements based on translation finds

reosarevok:


lib/MusicBrainz/Server/Form/User/Report.pm
```diff
@@ -37,7 +37,7 @@ Readonly our %REASONS => (
     'ignoring_guidelines' => N_l('Editor intentionally ignores accepted guidelines'),
     'enforcing_guidelines' => N_l('Editor is overzealous in enforcing guidelines as rules'),
     'voting' => N_l('Editor engages in overzealous or abusive yes/no voting'),
-    'other' => N_l('Editor has violated some other part of our Code of Conduct'),
```
mwiencek, line 40: Shouldn't it be capitalized since it's the title of a specific document? (That is, it's referring to MetaBrainz's Code of Conduct, not talking about codes of conduct in general.)

reosarevok: I dunno, we're not talking of it like a title, are we? It just feels a bit like "other part of our rules" or "other part of our guidelines". IIRC @Aerozol suggested lowercasing "Conduct" on the "About us" menu because of it not being a title :man_shrugging:

Aerozol: I think this was one of the other ways that could have gone either way? I know it's not helpful, but I don't mind either way...

reosarevok: Well, either we change it back to Code of Conduct everywhere, or we keep it as a generic "code of conduct" everywhere. @yvanzo, do I understand from your thumbs-up that you would prefer reverting the header change and using Code of Conduct everywhere? In that case I can drop that commit and revert the header instead :shrug:

yvanzo: > @yvanzo, do I understand from your thumbs-up that you would prefer reverting the header change and using Code of Conduct everywhere?

Yes as it clarifies that this isn’t just an idiomatic expression.

yvanzo (approved):
:ship:

lib/MusicBrainz/Server/Form/User/Report.pm
yvanzo: It also matches what we already have at https://metabrainz.org/code-of-conduct and in translation glossary.

PR #3186, opened 2024-02-26
MBS-13496: Support attributes with spaces in link phrase interpolation

reosarevok:
### Fix MBS-13496

# Problem
Using the new "a cappella" attribute is breaking link phrase interpolation, because spaces are not accepted as part of the interpolated "variable name". We already had several attributes with spaces in their name though, we just had gotten lucky we were not using any of them in link phrases until now.

# Solution
Just adding spaces to the allowed characters doesn't seem to cause any new issues, so it seems good enough for our needs.

# Testing
Added one test with spaces to the `expand2` tests.

# Further action

* [ ] Rename [the attribute](https://musicbrainz.org/relationship-attribute/4721e916-56c5-4e23-8828-33aedc7a8103) to "a cappella" and update the link phrases of [the relationship](https://musicbrainz.org/relationship/a3005666-a872-32c3-ad06-98af558e99b0).
* [ ] Check if the Weblate checks for interpolation need updating to support this change.


root/static/scripts/tests/i18n/expand2.js
```diff
@@ -46,6 +46,7 @@ test('expand2', function (t) {
   );
   expandText('An {apple_fruit}', null, 'An {apple_fruit}');
   expandText('An {apple_fruit}', {apple_fruit: 'apple'}, 'An apple');
+  expandText('An {apple fruit}', {'apple fruit': 'apple'}, 'An apple');
```
mwiencek, line 49: It'd be nice to have two more small tests for `condSubstStart` and `linkSubstStart`. :)

Also, I think `MusicBrainz::Server::Translation::expand` already supports this, but perhaps add a small test for the Perl side, too? There are some existing ones in t/lib/t/MusicBrainz/Server/Translation.pm.

reosarevok: Do I understand correctly that `condSubstStart` would be used in tests such as `expandText('{a:%{b:%|}%|}', {a: '0', b: ''}, '00');`? And is `linkSubstStart` tested in tests like this?

```
  expandHtml(
    'A {apple_fruit|<strong>{prefix} APPLE!</strong>}',
    {apple_fruit: 'http://www.apple.com', prefix: 'dang'},
    'A <a href="http://www.apple.com"><strong>dang APPLE!</strong></a>',
  );
```

reosarevok: I added a couple tests to the Perl file :)

yvanzo:

root/static/scripts/common/i18n/expand2.js
```diff
@@ -206,7 +206,7 @@ export const createTextContentParser = <T, V>(
   return mapValue(text);
 };
 
-const varSubst = /^\{([0-9A-z_]+)\}/;
+const varSubst = /^\{([0-9A-z_ ]+)\}/;
```
yvanzo, line 209: Beginning/ending spaces should probably not be allowed.

root/static/scripts/common/i18n/expand2.js
```diff
@@ -225,7 +225,7 @@ export const createVarSubstParser = <T, V>(
 export const parseStringVarSubst: Parser<string | NO_MATCH, mixed> =
   createVarSubstParser<string, mixed>(getString);
 
-const condSubstStart = /^\{([0-9A-z_]+):/;
+const condSubstStart = /^\{([0-9A-z_ ]+):/;
```
yvanzo, line 228: Same as above.

root/static/scripts/common/i18n/expand2react.js
```diff
@@ -53,7 +53,7 @@ type Output = Expand2ReactOutput;
 const textContent = /^[^<>{}]+/;
 const condSubstThenTextContent = /^[^<>{}|]+/;
 const percentSign = /(%)/;
-const linkSubstStart = /^\{([0-9A-z_]+)\|/;
+const linkSubstStart = /^\{([0-9A-z_ ]+)\|/;
```
yvanzo, line 56: Same as above.

root/static/scripts/common/i18n/expand2.js
reosarevok: Is there a reason why they shouldn't? I mean, I don't think we're going to *use* them, but what's the damage?

yvanzo: To prevent using them indeed. Damage is broken interpolation and translation. Same for underscores.

reosarevok: But that would only be a problem if we actually did add spaces at the beginning or end of variables, or a translator messed up (which is already the case even without the spaces) :) That said, if the regex can be easily changed to your preference, I have nothing against committing a suggestion for it, I just don't think we should complicate it too much otherwise.

yvanzo: Actually I don’t think that allowing spaces in variable names is a good approach at all, see [below](https://github.com/metabrainz/musicbrainz-server/pull/3186#issuecomment-1979380835).

reosarevok:
> A more durable implementation would be to use a slug of attribute name for variable name.
Do you mean that we should convert all attributes to what, replace spaces by hyphens and then check the variables? I'm not against that if it feels better, I guess, I just thought it might be more confusing but maybe not?

yvanzo:
> replace spaces by hyphens

By underscores, to follow the convention already in use for variables in the MusicBrainz Server code.

reosarevok:
Hmm. Well, we already support underscores, meaning it could theoretically clash between a name with a space and one with underscores, but we could probably make sure we don't mess that up at least :)

PR #3187, opened 2024-02-28
Limit documentation content to viewport width

derat:
Restrict MediaWiki content width to the viewport width to make it somewhat usable on mobile devices.

# Problem

Documentation pages set `min-width: 780px;` on `<body>`, which results in content being unusable on mobile devices with narrow screens -- the text is either tiny when zoomed out or multiple screen-widths wide when zoomed in. See the screenshots at <https://community.metabrainz.org/t/making-documentation-pages-usable-on-mobile-devices/602956>.

# Solution

Set `max-width: calc(100vw - 32px);` on the content so that it's limited to the viewport width even when the page itself is much wider (to accommodate the header).

Screenshot (Feb 28, 2024 07_27_15)

# Testing

Documentation pages look the same as before in wide windows, but the content is wrapped at the viewport width on mobile (in both Chrome dev tools and on a Pixel 7a).

### DIFF
--- a/root/static/styles/extra/wikidocs.less
+++ b/root/static/styles/extra/wikidocs.less
@@ -109,6 +109,12 @@
         font-size: @small-text;
     }
 
+    // MediaWiki content
+    .mw-parser-output {
+        // Make pages somewhat readable on mobile devices.
+        max-width: calc(100vw - 32px);
+    }
+
     // footer section
     .wikidocs-footer {
         border-top: 2px dotted @medium-border;


reosarevok:
When I try this in my Chrome dev for Pixel 7 I get this, which seems bad. What am I doing wrong?:

Screenshot from 2024-02-28 13-12-29

derat:
@reosarevok I see the same thing as you in DevTools; it seems like it doesn't include Android Chrome's logic for determining the initial zoom level. On my phone, I get the same overall layout, but the page is automatically zoomed to the transcluded content's width (like in the screenshot above). I just verified that Android Firefox 123.0 also auto-zooms to the content width.

mwiencek (approved):
Tested this on iOS 17.3 and it was certainly an improvement for the content portion (even if other parts of the page are still not responsive).

reosarevok:
Ok, if it helps in both Android and iOS and doesn't hurt in any way (which it shouldn't), let's get it merged. Hopefully we'll move to a better system later on anyway, but it's a help in the meantime.

PR #3189, opened 2024-02-28
MBS-12439: Handle VK music links

reosarevok:
### Implement MBS-12439

# Problem
We currently autoselect all VK links to social media, but for quite a while VK has also operated a streaming operation similar to Spotify.

# Solution
There seems to be no limit to the optional weirdness of the VK links.

This moves the VK music links to a separate cleanup. We support `/audio` and `/video` (recording), `/music/album` (release) and
`/artist` links, and try to clean up most known variations of them to end up with something mostly consistent.

This does not try to clean up / block numeric artist IDs, which redirect to a named artist page on desktop view (where we could
block and ask to follow the redirect) but stay numeric in mobile view for some reason. This means we might sometimes get duplicate numeric + named artist pages - we'll have to trust users to be mindful when adding these.

The "unofficial" music side of VK using `/wall` links (see ticket) is not handled here and can still be freely selected when needed. This does block `/music/playlist` links right now though (if it turns out to be unwanted we can look into handing those later on).

# Testing
Added a bunch of tests including some of the edge cases included in the ticket.


yvanzo (approved):
Thanks! Unlike VK whose URL patterns are difficult to follow, the comments and tests make this patch very easy to follow. The unrelated Selenium failure can be ignored. :ship:

PR #3191, opened 2024-02-28
MBS-7646: Show localized artist aliases in more places

derat:
# MBS-7646

Try to find artists' primary aliases in the user's locale and display them alongside disambiguation comments in search results and artist pages.

# Problem

Editors argue about whether Latin-script versions of artist names should be included in disambiguations (see e.g. https://musicbrainz.org/edit/109219215). Even when this is done, it's not a great approach since it doesn't help editors who are using UI languages based on other scripts. Without being able to see localized names in search results, editors either look at sort names (if they know about them) or end up creating duplicate entities.

# Solution

Localized versions of artist names are already stored as aliases, so automatically display those in italicized text at the beginning of the disambiguation comment when possible, e.g.

> Пётр Ильич Чайковский (_Pyotr Ilyich Tchaikovsky,_ Russian romantic composer)

I'm only including the alias when it doesn't already match the artist name, and I'm performing fallback to favor an exact language match before falling back to the generic language and then other variants (e.g. `en_UK` favors `en_UK`, then `en`, and then `en_US`). I'm not sure if the fallback is important given the current UI language options, though.

# Testing

I did manual testing by doing indexed and direct searches for `tchaikovsky` with various UI languages and checking that the appropriate alias is shown, and also checking that aliases are displayed in the header on artist pages.

I also added a new Selenium test.

### DIFF
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -146,9 +146,12 @@ after 'load' => sub
         }
     }
 
-    $c->model('ArtistType')->load($artist, map { $_->target } @{ $artist->relationships_by_type('artist') });
-    $c->model('Gender')->load($artist);
-    $c->model('Area')->load($artist);
+    my $lang = $c->stash->{current_language} // 'en';
+    # TODO: This loads the artist's aliases to look for a primary alias in the
+    # user's language, but note that aliases are also being loaded by the show
+    # subroutine to get the artist's legal name.
+    $c->model('Artist')->load_related_info([$artist], $lang);
+    $c->model('ArtistType')->load(map { $_->target } @{ $artist->relationships_by_type('artist') });
     $c->model('Area')->load_containment($artist->area, $artist->begin_area, $artist->end_area);
 };
 
--- a/lib/MusicBrainz/Server/Controller/Artist.pm
+++ b/lib/MusicBrainz/Server/Controller/Artist.pm
@@ -147,9 +147,6 @@ after 'load' => sub
     }
 
     my $lang = $c->stash->{current_language} // 'en';
-    # TODO: This loads the artist's aliases to look for a primary alias in the
-    # user's language, but note that aliases are also being loaded by the show
-    # subroutine to get the artist's legal name.
     $c->model('Artist')->load_related_info([$artist], $lang);
     $c->model('ArtistType')->load(map { $_->target } @{ $artist->relationships_by_type('artist') });
     $c->model('Area')->load_containment($artist->area, $artist->begin_area, $artist->end_area);
@@ -289,26 +286,22 @@ sub show : PathPart('') Chained('load')
     if (defined $base_name) {
         $c->model('Relationship')->load_subset(['artist'], $base_name);
         $c->stash( legal_name => $base_name );
-        my $aliases = $c->model('Artist')->alias->find_by_entity_id($base_name->id);
-        $c->model('Artist')->alias_type->load(@$aliases);
         my @aliases = uniq map { $_->name }
                       sort_by { $coll->getSortKey($_->name) }
                       # An alias equal to the artist name already shown isn't useful
                       grep { ($_->name) ne $base_name->name }
                       # A legal name alias marked ended isn't a current legal name
                       grep { !($_->ended) }
-                      grep { ($_->type_name // '') eq 'Legal name' } @$aliases;
+                      grep { ($_->type_name // '') eq 'Legal name' } @{ $base_name->aliases };
         $c->stash( legal_name_artist_aliases => \@aliases );
         $base_name_legal_name_aliases = \@aliases;
         push(@identities, $base_name);
     } else {
-        my $aliases = $c->model('Artist')->alias->find_by_entity_id($artist->id);
-        $c->model('Artist')->alias_type->load(@$aliases);
         my @aliases = uniq map { $_->name }
                       sort_by { $coll->getSortKey($_->name) }
                       # A legal name alias marked ended isn't a current legal name
                       grep { !($_->ended) }
-                      grep { ($_->type_name // '') eq 'Legal name' } @$aliases;
+                      grep { ($_->type_name // '') eq 'Legal name' } @{ $artist->aliases };
         $c->stash( legal_name_aliases => \@aliases );
         $legal_name_aliases = \@aliases;
     }
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -14,6 +14,7 @@ use MusicBrainz::Server::Data::Utils qw(
     add_partial_date_to_row
     conditional_merge_column_query
     contains_string
+    find_best_primary_alias
     hash_to_row
     load_subobjects
     merge_table_attributes
@@ -473,26 +474,16 @@ sub load_related_info {
     $c->model('Gender')->load(@artists);
     $c->model('Area')->load(@artists);
 
-    if (defined $user_lang) {
-        my $short_user_lang = substr($user_lang, 0, 2);
-        my $aliases = $c->model('Artist')->alias->find_by_entity_ids(
-            map { $_->id } @artists,
-        );
-        for my $artist (@artists) {
-            my $best_alias;
-            foreach my $alias (@{$aliases->{$artist->id}}) {
-                next if !defined $alias->locale || !$alias->primary_for_locale;
-                # If we find an exact match for the user's language, use it.
-                if ($alias->locale eq $user_lang) {
-                    $best_alias = $alias;
-                    last;
-                }
-                # Otherwise, favor more-generic aliases (e.g. "en" over "en_US").
-                if (substr($alias->locale, 0, 2) eq $short_user_lang &&
-                    (!defined $best_alias || length($alias->locale) == 2)) {
-                    $best_alias = $alias;
-                }
-            }
+    # Load and save aliases so Controller::Artist::show can use them later.
+    my $artist_aliases = $c->model('Artist')->alias->find_by_entity_ids(
+        map { $_->id } @artists,
+    );
+    for my $artist (@artists) {
+        my @aliases = @{ $artist_aliases->{$artist->id} };
+        $c->model('Artist')->alias_type->load(@aliases);
+        $artist->{aliases} = \@aliases;
+        if (defined $user_lang) {
+            my $best_alias = find_best_primary_alias(\@aliases, $user_lang);
             $artist->{primary_alias} = $best_alias->name if defined $best_alias;
         }
     }
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -478,13 +478,14 @@ sub load_related_info {
     my $artist_aliases = $c->model('Artist')->alias->find_by_entity_ids(
         map { $_->id } @artists,
     );
+    my @all_aliases = map { @$_ } values %$artist_aliases;
+    $c->model('Artist')->alias_type->load(@all_aliases);
     for my $artist (@artists) {
         my @aliases = @{ $artist_aliases->{$artist->id} };
-        $c->model('Artist')->alias_type->load(@aliases);
-        $artist->{aliases} = \@aliases;
+        $artist->aliases(\@aliases);
         if (defined $user_lang) {
             my $best_alias = find_best_primary_alias(\@aliases, $user_lang);
-            $artist->{primary_alias} = $best_alias->name if defined $best_alias;
+            $artist->primary_alias($best_alias->name) if defined $best_alias;
         }
     }
 }
--- a/lib/MusicBrainz/Server/Data/Artist.pm
+++ b/lib/MusicBrainz/Server/Data/Artist.pm
@@ -465,12 +465,37 @@ sub _hash_to_row
 }
 
 sub load_related_info {
-    my ($self, @artists) = @_;
+    my ($self, $artists_ref, $user_lang) = @_;
 
     my $c = $self->c;
+    my @artists = @{$artists_ref};
     $c->model('ArtistType')->load(@artists);
     $c->model('Gender')->load(@artists);
     $c->model('Area')->load(@artists);
+
+    if (defined $user_lang) {
+        my $short_user_lang = substr($user_lang, 0, 2);
+        my $aliases = $c->model('Artist')->alias->find_by_entity_ids(
+            map { $_->id } @artists,
+        );
+        for my $artist (@artists) {
+            my $best_alias;
+            foreach my $alias (@{$aliases->{$artist->id}}) {
+                next if !defined $alias->locale || !$alias->primary_for_locale;
+                # If we find an exact match for the user's language, use it.
+                if ($alias->locale eq $user_lang) {
+                    $best_alias = $alias;
+                    last;
+                }
+                # Otherwise, favor more-generic aliases (e.g. "en" over "en_US").
+                if (substr($alias->locale, 0, 2) eq $short_user_lang &&
+                    (!defined $best_alias || length($alias->locale) == 2)) {
+                    $best_alias = $alias;
+                }
+            }
+            $artist->{primary_alias} = $best_alias->name if defined $best_alias;
+        }
+    }
 }
 
 sub load_meta
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -419,7 +419,7 @@ sub schema_fixup_type {
 # matches up with the data returned from the DB for easy object creation
 sub schema_fixup
 {
-    my ($self, $data, $type) = @_;
+    my ($self, $data, $type, $user_lang) = @_;
 
     return unless (ref($data) eq 'HASH');
 
@@ -685,7 +685,7 @@ sub schema_fixup
 
             my %entity = %{ $rel->{$entity_type} };
 
-            $self->schema_fixup(\%entity, $entity_type);
+            $self->schema_fixup(\%entity, $entity_type, $user_lang);
 
             # The search server returns the MBID in the 'id' attribute, so we
             # need to delete that. (`schema_fixup` copies it to gid.)
@@ -716,13 +716,13 @@ sub schema_fixup
         next if $k eq '_extra';
         if (ref($data->{$k}) eq 'HASH')
         {
-            $self->schema_fixup($data->{$k}, $type);
+            $self->schema_fixup($data->{$k}, $type, $user_lang);
         }
         if (ref($data->{$k}) eq 'ARRAY')
         {
             foreach my $item (@{$data->{$k}})
             {
-                $self->schema_fixup($item, $type);
+                $self->schema_fixup($item, $type, $user_lang);
             }
         }
     }
@@ -738,6 +738,25 @@ sub schema_fixup
                     join_phrase => $namecredit->{joinphrase} || '' } );
         }
         $data->{'artist_credit'} = MusicBrainz::Server::Entity::ArtistCredit->new( { names => \@credits } );
+    }
+
+    if (defined $data->{aliases}) {
+        my $short_user_lang = substr($user_lang, 0, 2);
+        my $best_alias;
+        foreach my $alias (@{$data->{aliases}}) {
+            next if !defined $alias->{locale} || !$alias->{primary};
+            # If we find an exact match for the user's language, use it.
+            if ($alias->{locale} eq $user_lang) {
+                $best_alias = $alias;
+                last;
+            }
+            # Otherwise, favor more-generic aliases (e.g. "en" over "en_US").
+            if (substr($alias->{locale}, 0, 2) eq $short_user_lang &&
+                (!defined $best_alias || length($alias->{locale}) == 2)) {
+                $best_alias = $alias;
+            }
+        }
+        $data->{primary_alias} = $best_alias->{name} if defined $best_alias;
     }
 
     if ($type eq 'work') {
@@ -822,7 +841,7 @@ sub escape_query
 
 sub external_search
 {
-    my ($self, $type, $query, $limit, $page, $adv) = @_;
+    my ($self, $type, $query, $limit, $page, $adv, $user_lang) = @_;
 
     my $entity_model = $self->c->model( type_to_model($type) )->_entity_class;
     load_class($entity_model);
@@ -906,7 +925,7 @@ sub external_search
 
         foreach my $t (@{$data->{$xmltype}})
         {
-            $self->schema_fixup($t, $type);
+            $self->schema_fixup($t, $type, $user_lang);
             push @results, MusicBrainz::Server::Entity::SearchResult->new(
                     position => $pos++,
                     score  => $t->{score},
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -757,10 +757,10 @@ sub schema_fixup
         my $best_alias = find_best_primary_alias(\@aliases, $user_lang);
         $data->{primary_alias} = $best_alias->{name} if defined $best_alias;
 
-        # The aliases returned by the search API aren't Entity::Alias objects,
-        # so clear them to prevent validation from failing if this is an
-        # Entity::Artist.
-        $data->{aliases} = [];
+        # Save the new objects so validation will pass, but note that the search
+        # API doesn't include all attributes that are present in Entity::Alias,
+        # e.g. no IDs.
+        $data->{aliases} = \@aliases;
     }
 
     if ($type eq 'work') {
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -748,14 +748,19 @@ sub schema_fixup
     if (defined $data->{aliases}) {
         my @aliases = map {
             MusicBrainz::Server::Entity::Alias->new(
-                sort_name => $_->{sort_name} || '',
-                locale => $_->{locale} || '',
-                name => $_->{name} || '',
+                name => $_->{name} // '',
+                sort_name => $_->{sort_name} // '',
+                locale => $_->{locale} // '',
                 primary_for_locale => $_->{primary},
             )
         } @{ $data->{aliases} };
         my $best_alias = find_best_primary_alias(\@aliases, $user_lang);
         $data->{primary_alias} = $best_alias->{name} if defined $best_alias;
+
+        # The aliases returned by the search API aren't Entity::Alias objects,
+        # so clear them to prevent validation from failing if this is an
+        # Entity::Artist.
+        $data->{aliases} = [];
     }
 
     if ($type eq 'work') {
--- a/lib/MusicBrainz/Server/Data/Search.pm
+++ b/lib/MusicBrainz/Server/Data/Search.pm
@@ -11,6 +11,7 @@ use Data::Dumper;
 use Data::Page;
 use URI::Escape qw( uri_escape_utf8 );
 use List::AllUtils qw( any partition_by );
+use MusicBrainz::Server::Entity::Alias;
 use MusicBrainz::Server::Entity::Annotation;
 use MusicBrainz::Server::Entity::Area;
 use MusicBrainz::Server::Entity::AreaType;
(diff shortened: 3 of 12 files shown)


derat:
@mwiencek @reosarevok @yvanzo Early feedback appreciated. :-)

root/static/scripts/common/components/EntityLink.js
```diff
@@ -411,9 +411,18 @@ $ReadOnlyArray<Expand2ReactOutput> | Expand2ReactOutput | null => {
...
+    const primaryAlias = entity.primaryAlias ?? '';
+    if (comment || primaryAlias) {
+      const textParts = [];
+      if (primaryAlias) {
+        textParts.push(primaryAlias);
+      }
+      if (comment) {
+        textParts.push(comment);
+      }
+      const text = textParts.join(', ');
```
yvanzo, line 423: For a localized comma-separated list, import our `commaOnlyList` and use it as follows:
```suggestion
      const text = commaOnlyList(textParts);
```

yvanzo (commented):
> I'm uploading this to get some feedback about whether this seems like a reasonable approach to pursue.

Thank you for looking into this issue!

> Along with the TODOs in the code,

I didn’t review those.

> I suspect that this code doesn't yet work for indexed searches

I confirm, but Solr search returns with artist results with aliases, so it is a pure MBS issue to make it work. See for example [search API results with Pyotr Ilyich Tchaikovsky](https://musicbrainz.org/ws/2/artist?query=arid:9ddd7abc-9e1b-471d-8031-583bc6bc8be9&fmt=json).

> I've been testing it with direct searches in my local dev environment.

Works for me too: http://localhost:5000/search?query=Tchaikovsky&type=artist&method=direct

> I'm also not sure if aliases should be displayed in more (or fewer) places, and whether there's any way to generalize this to support additional entity types instead of only artists.

@Aerozol would be the right person for providing some guidance, especially now that he can test PRs locally! :rocket:

Aerozol:
I would love to test and feedback, but I think that would involve configuring the search server on my side? It was a bit of a rigmarole to get the test server running on Win at all so I'm a bit hesitant.

I would love to see a screenshot - this sounds really useful! I am keen to see how it's displayed.

derat:
@Aerozol no need for a search server; the localized names are shown in the results when you do a direct search (e.g. <http://localhost:5000/search?query=tchaikovsky&type=artist&limit=25&method=direct>).

There also isn't too much to see, though; the disambiguation is just rendered as "Pyotr Ilyich Tchaikovsky, Russian romantic composer" instead of "Russian romantic composer". :-)

Aerozol:
Sorry for the delay in comment, I finally (with further help from monkey) managed to wade through all my Win OS/Ubuntu VM issues, and 'test' this. Hopefully nobody was waiting for me!

This is a easy 'yes' from me. MB isn't mobile friendly anyway, so I don't see a downside in having this take up some of the empty horizontal space. And it is indeed a constant cause of friction re. disambiguation edits.

My only thought is that we may want to distinguish between the automatically displayed content by putting it in italics, in case users get confused, but it would also be uglier.

The TODO comments are too technical for me to address.

My only wish is that this was extended to more (all) entity types/searches. And some day we maybe automate even more of the disambiguation, like RYM does - e.g. 'rock group from Bulgaria' can all be automated. *cough* bumps into derat *cough*

Great stuff!

derat:
Thanks for taking a look!

I'll try to spend some more time soon working on the remaining issues, e.g. getting this to work for indexed search. It'd definitely be nice to support other entity types too, but I'll need to look into how hard it'll be to pass the aliases through there. And your suggestion to italicize the alias also seems like it'll help reduce confusion -- I'll try to see how that looks.

One other consideration is that there are probably a bunch of entities that already have Latin names in their disambiguations where this would result in near-duplicated text. IIRC I was thinking of avoiding displaying the alias if it's already present in the disambiguation or if it matches the entity name, but there are likely still many cases that would be missed due to alternate spellings, diacritics, etc.

Aerozol:
Ninja'd! I was just drafting another comment that we might run into some situations where there is some very repetitive text. Because of text in the disambiguation, but also where the primary alias is very similar to the artist name (I don't know if this avoids adding it if it's 100% the same?)

reosarevok:
Avoiding adding it if the same as artist name makes a lot of sense to me. I would *not* hide it if it's already in the disambiguation - it should not be there anyway, so this will make it more obvious so people can remove it :)

derat:
After some musicbrainz-docker help from @yvanzo in MBVM-95, I've started looking at this again. Here's a screenshot of what it looks like after italicizing the localized aliases:

image

Aerozol:
I like the italics :)
(I should have just tested it in the browser console before suggesting it, sorry!)

Another step towards clarity could be mouse-over text for the italics, something like: "This artists' primary alias in your [browser/account(?)] locale", but I don't think it's urgent.

root/static/scripts/common/components/EntityLink.js
derat: Thanks! I suspect that `commaOnlyList` can't be used now that I'm italicizing the alias (and the comma after it), but let me know if I'm wrong about that. Maybe I could do some hack like calling `commaOnlyList([alias, ' '])` and then putting that inside the `<i>`.

lib/MusicBrainz/Server/Controller/Artist.pm
```diff
@@ -146,9 +146,12 @@ after 'load' => sub
         }
     }
 
-    $c->model('ArtistType')->load($artist, map { $_->target } @{ $artist->relationships_by_type('artist') });
-    $c->model('Gender')->load($artist);
-    $c->model('Area')->load($artist);
+    my $lang = $c->stash->{current_language} // 'en';
+    # TODO: This loads the artist's aliases to look for a primary alias in the
+    # user's language, but note that aliases are also being loaded by the show
+    # subroutine to get the artist's legal name.
```
derat, line 152: Is loading the aliases twice like this a concern, or should I just delete this TODO?

lib/MusicBrainz/Server/Controller/Artist.pm
```diff
@@ -146,9 +146,12 @@ after 'load' => sub
         }
     }
 
-    $c->model('ArtistType')->load($artist, map { $_->target } @{ $artist->relationships_by_type('artist') });
-    $c->model('Gender')->load($artist);
-    $c->model('Area')->load($artist);
+    my $lang = $c->stash->{current_language} // 'en';
```
derat, line 149: I'm pretty clueless about how this `c` (context?) stash stuff works. I'm passing the language around since `current_language` didn't seem to be available in `load_related_info` and the indexed search code, but please let me know if there's an easier way to get at it there.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -465,12 +465,37 @@ sub _hash_to_row
...
     $c->model('Area')->load(@artists);
+
+    if (defined $user_lang) {
+        my $short_user_lang = substr($user_lang, 0, 2);
+        my $aliases = $c->model('Artist')->alias->find_by_entity_ids(
+            map { $_->id } @artists,
+        );
+        for my $artist (@artists) {
+            my $best_alias;
+            foreach my $alias (@{$aliases->{$artist->id}}) {
```
derat, line 483: This logic is essentially duplicated this file and Search.pm, but that may be unavoidable due to the alias objects being different (see e.g. `primary_for_locale` here vs. `primary` in Search.pm).

derat (commented):
I finally got this working for indexed search!

It's a bit weird in that I think it'll also work for non-artist entities in indexed search but only for artists everywhere else. Not sure if that's a concern; I suspect that getting primary aliases when loading other entity types from the DB will take a fair amount of additional code.

derat:
> Another step towards clarity could be mouse-over text for the italics, something like: "This artists' primary alias in your [browser/account(?)] locale", but I don't think it's urgent.

Thanks! I've added "Primary alias in your language" title text.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -465,12 +465,37 @@ sub _hash_to_row
...
+                next if !defined $alias->locale || !$alias->primary_for_locale;
+                # If we find an exact match for the user's language, use it.
+                if ($alias->locale eq $user_lang) {
+                    $best_alias = $alias;
+                    last;
+                }
+                # Otherwise, favor more-generic aliases (e.g. "en" over "en_US").
```
derat, line 490: `_with_primary_alias` in lib/MusicBrainz/Server/WebService/JSONSerializer.pm appears to also fall back to `en` and `en_.*`. This seems like it's probably desirable behavior for e.g. a Spanish user viewing a Japanese artist who only has an English alias, but it seems incorrect in many other cases.

This is something that's always bugged me a bit about aliases: there's no way to mark an alias as primary for all Latin-script languages, and there's also no locale associated with an entity's name, only with its aliases.

Let me know if there's something better I could be doing here. I think that what I'd really like would be to fall back to an alias using the same script as the user's language if the entity name isn't already using that script, but I'm not sure if that's possible.

derat:
The new tests pass:

```
MBS-7646: Show localized primary aliases alongside disambiguations                                                                                               
[17:16:41.997Z] sendKeys target="css=#headerid-query" value="tchaikovsky${KEY_ENTER}"  
[17:16:43.086Z] waitUntilUrlIs target="/search?query=tchaikovsky&type=artist&method=indexed" value=""  
[17:16:43.098Z] assertText target="css=#content table.tbl td a" value="Пётр Ильич Чайковский"  
  ✓ should be equal  
[17:16:43.131Z] assertText target="css=#content table.tbl td .comment" value="(Pyotr Ilyich Tchaikovsky, Russian romantic composer)"  
  ✓ should be equal  
[17:16:43.160Z] click target="css=input[name=\"method\"][value=\"direct\"]" value=""  
[17:16:43.223Z] click target="css=div.searchform button[type=\"submit\"]" value=""  
[17:16:44.225Z] waitUntilUrlIs target="/search?query=tchaikovsky&type=artist&limit=25&method=direct" value=""  
[17:16:44.238Z] assertText target="css=#content table.tbl td a" value="Пётр Ильич Чайковский"  
  ✓ should be equal  
[17:16:44.265Z] assertText target="css=#content table.tbl td .comment" value="(Pyotr Ilyich Tchaikovsky, Russian romantic composer)"  
  ✓ should be equal  
[17:16:44.291Z] click target="css=#content table.tbl td a" value=""  
[17:16:45.745Z] waitUntilUrlIs target="/artist/9ddd7abc-9e1b-471d-8031-583bc6bc8be9" value=""  
[17:16:45.758Z] assertText target="css=.artistheader a" value="Пётр Ильич Чайковский"  
  ✓ should be equal  
[17:16:45.791Z] assertText target="css=.artistheader .comment" value="(Pyotr Ilyich Tchaikovsky, Russian romantic composer)"  
  ✓ should be equal
```

derat:

derat:
@reosarevok and @yvanzo, I think this is ready for review now (when you have time). :-)

lib/MusicBrainz/Server/Data/Artist.pm
mwiencek: Perhaps we can build `Entity::Alias` objects in Search.pm so that they can be passed to a utility function that can be called from both files?

Technically there are separate alias classes per entity type (e.g., `Entity::ArtistAlias`), but the generic class `Entity::Alias` may suffice for the limited amount information we need to store to satisfy the utility code.

lib/MusicBrainz/Server/Data/Artist.pm
mwiencek: > I think that what I'd really like would be to fall back to an alias using the same script as the user's language

I'm not sure if we have any mapping between language (or territory) and likely script (is this available in the CLDR?), but perhaps @yvanzo has an idea.

I do wonder if `_with_primary_alias` could be modified to make use of the proposed utility function, so that it can benefit from any future improvements in this area too.

lib/MusicBrainz/Server/Controller/Artist.pm
mwiencek: It'd be nice if we could avoid it, since some artists have a ton of aliases. Perhaps add an `aliases` attribute to `Entity::Artist` to cache these on? Or store them in the stash somehow?

mwiencek (commented):
Thank you, I'm very happy with this improvement and it works well for me, although perhaps it could do with some light refactoring if you're up to it. :)

lib/MusicBrainz/Server/Controller/Artist.pm
derat: Thanks, I've written some code to store these in `Entity::Artist` that seems to work.

lib/MusicBrainz/Server/Data/Artist.pm
derat: This makes sense to me. Can you give me a pointer to some existing code that's manually creating entity objects? I've spun my wheels for a bit now without having any luck here -- my OO Perl knowledge is pretty much nonexistent (like the rest of my Perl knowledge). I think that the only fields that I care about are `locale`, `name`, and `primary_for_locale` (which needs to be remapped from `primary` in Search.pm).

lib/MusicBrainz/Server/Data/Artist.pm
derat: If it's okay with you, I'd prefer to avoid changing `_with_primary_alias` in this PR to reduce the (already high-seeming to me) odds that I break existing functionality. :-) Is adding a TODO okay for now?

derat (commented):
Thanks for taking a look!

lib/MusicBrainz/Server/Data/Artist.pm
derat: Never mind, I think I figured this out. :-)

lib/MusicBrainz/Server/Controller/Artist.pm
```diff
@@ -286,26 +286,22 @@ sub show : PathPart('') Chained('load')
     if (defined $base_name) {
         $c->model('Relationship')->load_subset(['artist'], $base_name);
         $c->stash( legal_name => $base_name );
-        my $aliases = $c->model('Artist')->alias->find_by_entity_id($base_name->id);
```
derat, line 289: Is it safe to assume that `load_related_info` will have saved the aliases to the `Artist` object whenever this is called, or should I include fallback code to try to load aliases if there aren't any?

It seems to work to the extent that I still see legal names when I view artist pages, at least. (Those weren't being shown with this change until I added the `$c->model('Artist')->alias_type->load(@aliases)` call in `load_related_info`.)

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -465,12 +466,27 @@ sub _hash_to_row
...
     $c->model('Gender')->load(@artists);
     $c->model('Area')->load(@artists);
+
+    # Load and save aliases so Controller::Artist::show can use them later.
+    my $artist_aliases = $c->model('Artist')->alias->find_by_entity_ids(
+        map { $_->id } @artists,
+    );
+    for my $artist (@artists) {
+        my @aliases = @{ $artist_aliases->{$artist->id} };
+        $c->model('Artist')->alias_type->load(@aliases);
```
derat, line 483: Would it be better to make a single call that loads the alias types for all of the aliases instead of doing it once per artist here?

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -740,6 +745,19 @@ sub schema_fixup
         $data->{'artist_credit'} = MusicBrainz::Server::Entity::ArtistCredit->new( { names => \@credits } );
     }
 
+    if (defined $data->{aliases}) {
+        my @aliases = map {
+            MusicBrainz::Server::Entity::Alias->new(
```
derat, line 750: This seems to work, but please let me know if I should be doing something else to create these objects.

lib/MusicBrainz/Server/Data/Utils.pm
```diff
@@ -770,6 +771,35 @@ sub contains_string {
...
+
+        # If we find an exact match for the user's language, use it.
+        return $alias if $alias->locale eq $lang;
+
+        # Otherwise, favor more-generic aliases (e.g. "en" over "en_US").
+        if (substr($alias->locale, 0, 2) eq $short_lang &&
+            (!defined $best || length($alias->locale) == 2)) {
+            $best = $alias;
+        }
+        # If we find an English alias, use it as a fallback. This is likely
```
derat, line 792: I figured it'd be worthwhile falling back to English here since it looks like all of the UI languages that are presently listed use the Latin script.

root/static/scripts/common/components/EntityLink.js
```diff
@@ -74,13 +74,23 @@ const iconClassPicker = {
...
-}: {+className: string, +comment: string}) => (
+}: {+alias?: string, +className: string, +comment: string}) => (
   <>
     {' '}
     <span className={className}>
-      {bracketed(<bdi key="comment">{comment}</bdi>)}
+      {bracketed(
+        <bdi key="comment">
+          {nonEmpty(alias) ? (
+            <i title={l('Primary alias in your language')}>
```
derat, line 87: Should I change this now that it's not necessarily the user's language (due to falling back to English)?

derat (commented):
I think this is ready for another look!

I pushed the updates as a separate commit so it's easier to see what's changed -- hopefully it's simple to squash everything when merging. <sub><sup>Sigh, Gerrit is so much better than GitHub's code review workflow. 🙃</sup></sub>

derat:
@mwiencek, mind taking another look?

lib/MusicBrainz/Server/Data/Artist.pm
mwiencek: > Would it be better to make a single call that loads the alias types for all of the aliases instead of doing it once per artist here?

Yes, it probably won't make a significant difference in production, but will where the PG connection latency is high (which happens when connecting via SSH tunnel, at least).

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -465,12 +466,27 @@ sub _hash_to_row
...
+        $artist->{aliases} = \@aliases;
```
mwiencek, line 484: Moose accessors can be called like setters:
```suggestion
        $artist->aliases(\@aliases);
```
Setting the property directly doesn't perform any validation or trigger other Moosey configuration on the attribute.

lib/MusicBrainz/Server/Data/Artist.pm
```diff
@@ -465,12 +466,27 @@ sub _hash_to_row
...
+        if (defined $user_lang) {
+            my $best_alias = find_best_primary_alias(\@aliases, $user_lang);
+            $artist->{primary_alias} = $best_alias->name if defined $best_alias;
```
mwiencek, line 487:
```suggestion
            $artist->primary_alias($best_alias->name) if defined $best_alias;
```

lib/MusicBrainz/Server/Data/Search.pm
```diff
@@ -740,6 +745,19 @@ sub schema_fixup
...
+                sort_name => $_->{sort_name} || '',
+                locale => $_->{locale} || '',
+                name => $_->{name} || '',
```
mwiencek, line 753: `//` can be used to only fall back to `''` if the LHS is `undef` (rather than any falsey value). I guess it would only matter if the name or sort name of the alias is `'0'` (unfortunately, `('0' || '') eq ''` evaluates to `1`).
```suggestion
                sort_name => $_->{sort_name} // '',
                locale => $_->{locale} // '',
                name => $_->{name} // '',
```

root/static/scripts/common/components/EntityLink.js
mwiencek: @reosarevok any suggestions? If we knew whether the alias was in English, I guess we could conditionally display "Primary alias in English" instead.

lib/MusicBrainz/Server/Entity/Artist.pm
```diff
@@ -25,6 +25,17 @@ has 'sort_name' => (
     isa => 'Str',
 );
 
+has 'aliases' => (
+    is => 'rw',
+    isa => 'ArrayRef[Any]',
```
mwiencek, line 30:
```suggestion
    isa => 'ArrayRef[MusicBrainz::Server::Entity::Alias]',
```

mwiencek (commented):
Thanks for the refactoring! Looking good.

lib/MusicBrainz/Server/Entity/Artist.pm
derat: Sorry, I forgot to leave a comment about this after uploading this commit. That's what I tried at first, but I got validation errors when performing indexed searches:

```
Attribute (aliases) does not pass the type constraint because: Validation failed for 'ArrayRef[MusicBrainz::Server::Entity::Alias]' with value [ { begin-date: undef, end-date: undef, gid: undef, id: 1, locale: "de", name: "Pjotr Iljitsch Tschaikowski", ... }, { begin-date: undef, end-date: undef, gid: undef, id: 1, locale: undef, name: "Tchaikovskij", ... }, { begin-date: undef, end-date: undef, gid: undef, id: 1, locale: undef, name: "Pyotr Il'yich Tchaikovsky", ... }, { begin-date: undef, end-date: undef, gid: undef, id: 1, locale: undef, name: "Iljitsch", ... }, { begin-date: undef, end-date: undef, gid: undef, id: 1, locale: "en", name: "Peter Tchaikovsky", ... }, { begin-date: undef, end-date: undef, gid: undef, id: 1, locale: undef, name: "Tschaikowski", ... }, ... ] at lib/MusicBrainz/Server/Data/Search.pm line 932
MusicBrainz::Server::Data::Search::external_search(?, ?, ?, ?, ?, ?, ?) called at lib/MusicBrainz/Server/Controller/Search.pm line 250
MusicBrainz::Server::Controller::Search::do_external_search(?, ?, ?, ?, ?, ?, ?, ?, ?, undef, ?, ?) called at lib/MusicBrainz/Server/Controller/Search.pm line 229
Catalyst::forward(?, ?) called at lib/MusicBrainz/Server/Controller/Search.pm line 62
Catalyst::dispatch(?) called at lib/MusicBrainz/Server.pm line 400
MusicBrainz::Server::__ANON__ at lib/MusicBrainz/Server.pm line 365
MusicBrainz::Server::with_translations(?, ?) called at lib/MusicBrainz/Server.pm line 401
Class::MOP::Method::Wrapped::__ANON__(?) called at lib/MusicBrainz/Server.pm line 415
Class::MOP::Method::Wrapped::__ANON__(?) called at lib/MusicBrainz/Server.pm line 522
```

I _think_ that it's happening because the addition of this field to `Entity::Artist` makes the aliases get validated, which then fails due to the search API not returning objects that adhere to `Entity::Alias` (e.g. no IDs).

Clearing `$data->{aliases}` in Server/Data/Search.pm's `schema_fixup` after I find the primary alias seems to prevent the validation error, but maybe that's too much of a time bomb for someone who ends up trying to use the aliases later. Then again, they probably wouldn't be able to use the aliases for much anyway due to the missing fields...

lib/MusicBrainz/Server/Data/Artist.pm
derat: Thanks, I've made this change for the next commit and it seems to work.

lib/MusicBrainz/Server/Data/Artist.pm
derat: Thanks, done.

lib/MusicBrainz/Server/Data/Search.pm
derat: Thanks, done.

lib/MusicBrainz/Server/Controller/Artist.pm
mwiencek: > Is it safe to assume that load_related_info will have saved the aliases to the Artist object whenever this is called, or should I include fallback code to try to load aliases if there aren't any?

It looks safe to me since `load_related_info` is called in `after 'load'`, while `show` uses `Chained('load')`.

root/static/scripts/common/components/EntityLink.js
derat: Maybe "Primary alias" would be good enough here. I think the most important bit is to let editors know where this came from in case they want to change it.

lib/MusicBrainz/Server/Entity/Artist.pm
mwiencek: Oh, I see. I was expecting that you'd have `$data->{aliases} = \@aliases;` after `@aliases` is declared, so that `Entity::Artist` receives the `Entity::Alias` objects when constructed. Or does that not work either?

lib/MusicBrainz/Server/Data/Utils.pm
derat: I think I mentioned it earlier, but a (future?) downside of this is that if e.g. a user selects Japanese as their UI language (I don't think they can at present), then they'll always see English aliases for Japanese artists, since I don't think there's any way to know that the artist entity's name is already in Japanese.

lib/MusicBrainz/Server/Entity/Artist.pm
derat: Huh, I thought I'd tried this, but maybe I used the array instead of an array ref -- it seems to work. I've updated the code and added a comment about how not all of the fields are present.

lib/MusicBrainz/Server/Data/Utils.pm
mwiencek: You could in theory check if there's a Japanese artist alias that exactly matches the artist name (but you're right, if that data doesn't exist).

derat: Yeah, in practice I don't think that many editors add primary aliases that are identical to entity names. https://musicbrainz.org/artist/9ddd7abc-9e1b-471d-8031-583bc6bc8be9/aliases has a zillion aliases, but of the two matching the "Пётр Ильич Чайковский" entity name, one is a non-primary legal name for Russian and the other is a search hint. Oddly, there's a primary artist name Russian alias of "Пётр Чайковский".

root/static/scripts/common/components/EntityLink.js
derat: @Aerozol too. I've shortened it to "Primary alias" for now, but it'd probably be possible to also pass through and display the alias's locale as well (at the cost of a bit more code).

Aerozol: Probably better to shorten it then, yeah - 'Primary alias' is still helpful in that it lets people know where the text comes from, and making the rollover too long seems annoying.
Some users will just have to figure out why it's sometimes in English and sometimes in their language, but that's not the most confusing thing in MB tbh :D

Aerozol: P.S. But if anyone can figure out how to say this in 5 words or less, I will give them a big smooch: "Primary alias in your language or English as a fallback"

derat: I was thinking maybe "Localized primary alias", but "localized" is probably meaningless to many users. And we'd need to argue about whether to spell it "localized" or "localised". :-)

root/static/scripts/common/components/EntityLink.js
```diff
@@ -74,13 +74,23 @@ const iconClassPicker = {
...
     {' '}
     <span className={className}>
-      {bracketed(<bdi key="comment">{comment}</bdi>)}
+      {bracketed(
+        <bdi key="comment">
+          {nonEmpty(alias) ? (
+            <i title={l('Primary alias')}>
+              {alias + (nonEmpty(comment) ? ', ' : '')}
+            </i>
+          ) : null}
```
mwiencek, line 90: Just realized this should probably be using `commaOnlyList`:

```suggestion
          {nonEmpty(aliasElement) ? (
            nonEmpty(comment) ? (
              commaOnlyList([aliasElement, comment])
            ) : aliasElement
          ) : comment}
```

Where `aliasElement` can be defined above, like:

```js
const aliasElement = nonEmpty(alias)
  ? <i title={l('Primary alias')}>{alias}</i>
  : null;
```

However it also requires changing the component definition to use `=> { ... }` instead of `=> ( ... )`.

derat: Thanks, done. I'd considered this earlier (@yvanzo gave the same suggestion), but I think I had worried that it'd look strange to have the comma and space in the non-italicized text node. Looks fine to me, though.

mwiencek (approved):
Thank you, a lot of people will appreciate this feature.

derat:
Thanks for all the help getting it merged!

lib/MusicBrainz/Server/Controller/Artist.pm
reosarevok: It works locally running the beta code, which suggests to me that the product of a prod load without aliases is being cached and causing the beta crash. If that's the case, we could either amend prod to save the info to the artist even if it doesn't use it, or amend beta temporarily to just show nothing rather than ISE if the data is missing. My preference would be the first, if we think it's safe (we're supposed to hotfix prod soon anyway for unrelated reasons)

derat: @mwiencek created #3252 to fix this.

derat: @mwiencek and @reosarevok, just to make sure I understand this properly, will the caching also cause issues with the `primary_alias` attribute on `Entity::Artist`? The alias is selected based on the UI language. Is that language incorporated into the cache keys, or will there be issues if an entity is cached for a request from a French-using user and then later reused for a request from an English-using user?

mwiencek: That shouldn't be an issue as entities are cached immediately before any additional data is loaded onto them. So no `primary_alias` value should be set on the cached copies.

PR #3194, opened 2024-03-01
Fix MBS-13500, MBS-13501, and a couple other small JSON dump issues

mwiencek:
# Problem

 * MBS-13500 - Incremental JSON dumps should not require foreign keys to be installed
 * MBS-13501 - Incremental JSON dumps disable mirror-only triggers
 * json_dump/CreatePrimaryKeys.sql is never called.
 * One of our queries references a non-existent table, `json_dump.delete_entities`.

This is only a very small part of the JSON dump changes I was working on, but I figured I'd submit them separately.

# Solution

See associated commits.

# Testing

I've been running these changes inside the production musicbrainz-json-dump container for the past week without issues. I've also updated the existing JSON dump and sitemap tests.


admin/RunIncrementalJSONDump
```diff
@@ -23,15 +23,6 @@ source admin/config.sh
 . ./admin/functions.sh
 make_temp_dir
 
-# The database from which the dumps are generated is replicated, but created
-# as standalone - meaning it has foreign keys and triggers. Foreign keys are
-# needed for MusicBrainz::Server::Role::FollowForeignKeys to work. The
-# triggers aren't needed and could (or should) be dropped, but it's safer to
-# make sure they are fully disabled here too. Leaving them enabled would
-# certainly cause the data to become corrupted and replication to break.
```
reosarevok, line 31: Do we need to disable *some* triggers so that they don't "cause the data to become corrupted and replication to break" or was this an unfounded fear all along?

admin/RunIncrementalJSONDump
```diff
@@ -23,13 +23,25 @@ source admin/config.sh
...
+
 echo Making incremental JSON dumps
 ./admin/DumpIncrementalJSON \
     --output-dir "$TEMP_DIR" \
     --compress \
     --database READWRITE \
+    --foreign-keys-dump "$FK_DUMP" \
     || exit $?
 
+rm -f "$FK_DUMP"
```
reosarevok, line 43: Is is at all potentially problematic that on error this will exit above (I assume) without removing the dump file?

reosarevok (commented):
This is marked as open, but you say "In a later commit, I will change this test file" on the message for 5b42a718f2c7adbc0dc2c72f853c9195c37e7797 - are you planning to do it as part of this PR or add a ticket? :)

mwiencek:
> This is marked as open, but you say "In a later commit, I will change this test file" on the message for https://github.com/metabrainz/musicbrainz-server/commit/5b42a718f2c7adbc0dc2c72f853c9195c37e7797 - are you planning to do it as part of this PR or add a ticket? :)

Not in this PR, but it's referencing a commit which will come in a later PR. I'll most like add a ticket for porting the incremental dumps to use dbmirror2 packets, which it's related to.

admin/RunIncrementalJSONDump
mwiencek: Well, I actually removed all triggers (except for the mirror-only ones) from the `musicbrainz_json_dump` DB, so there is nothing to disable anymore. :) And yeah, any trigger that modified replicated tables would cause replication to break.

yvanzo (commented):
Minor suggestions mostly.

admin/RunIncrementalJSONDump
mwiencek: Added `EXIT` traps.

yvanzo (approved):
Thank you! Commits are very explanatory. :ship:

PR #3196, opened 2024-03-05
MBS-13505: Split relationships on task attributes

reosarevok:
### Implement MBS-13505

# Problem
Task attributes, like instruments and vocals, should lead to split relationships, rather than be kept on new split relationships. This is needed for STYLE-2445 to lead to results that make sense.

If we add a new "member of" relationship with attributes "task: dancer", "instrument: piano" and "vocals: lead vocals":

* Currently, we end up with two split relationships, "task: dancer and piano" and "task: dancer and lead vocals".
* We want to end up with three split relationships, "task: dancer", "piano" and "lead vocals".

# Solution
Just add task to the list of things that we check for in `isInstrumentOrVocal` (which is changed to `isSplittableAttribute` because `isInstrumentOrTaskOrVocal` seemed too silly).

# Testing
Added a test for this case to the existing `splitRelationshipByAttributes` tests.

# Further action
Once this is released (on prod, probably) I'll actually implement STYLE-2445 adding the `task` attribute to `member of` relationships.

### DIFF
--- a/root/static/scripts/tests/relationship-editor.js
+++ b/root/static/scripts/tests/relationship-editor.js
@@ -304,7 +304,7 @@ test('merging duplicate relationships', function (t) {
 });
 
 test('splitRelationshipByAttributes', function (t) {
-  t.plan(16);
+  t.plan(19);
 
   const lyre = {
     type: {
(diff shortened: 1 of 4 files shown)


root/static/scripts/tests/relationship-editor.js
```diff
@@ -665,6 +665,55 @@ test('splitRelationshipByAttributes', function (t) {
     'second relationship contains drums',
   );
 
+  // This test ensures tasks are treated the same as instruments and vocals.
+
+  const task = {
+    text_value: 'stuffy stuff',
```
yvanzo, line 671: Can it be given a real example value for clarity? `'dancer'`?

reosarevok: Booooring. Sure.

reosarevok:
Yes, I did test locally :) Updated the commit message and ticket now.

root/static/scripts/tests/relationship-editor.js
yvanzo: `bird-disguised dancer`?

yvanzo (approved):
It’s show time! :dancer: :bird: :bird: :bird:

